PR #7025, opened 2023-11-15
[NVIDIA XLA] NFC Unify GPU backend configs

Tixxx:
This PR unifies all GPU backend configs into a single one, this is to make it easy to add generic attributes in the future.

PR #8452, opened 2024-01-12
Move stream attributes from hlo.proto to gpu backend config

Tixxx:
This is to follow up on this pr(https://github.com/openxla/xla/pull/7025) to move stream attributes from hlo.proto to gpuBackendConfig.


Tixxx:
@cezheng @cheshire  This is the follow up pr to remove stream attributes.

cheshire (approved):

PR #8459, opened 2024-01-13
[XLA:CPU][Bug-fix] Fix bias fusion to oneDNN mamtul.

mdfaijul:
This PR fixes a bug in the bias fusion to oneDNN matmul by not allowing fusion when the bias shape is non-compatible. Note that oneDNN matmul has a binary-add operation semantic for bias. This is more general than the usual bias having feature dimension only. While fixing the bug, this PR also does the following:

- Refactors pattern matching of bias fusion
- Adds more documentations in the source code
- Adds more checks for a valid fusion

### DIFF
--- a/xla/service/cpu/onednn_matmul_rewriter.cc
+++ b/xla/service/cpu/onednn_matmul_rewriter.cc
@@ -151,9 +151,9 @@ auto GELUActivation(HloInstruction* instr, HloInstruction** src) {
                    .WithOneUser());
 }
 
-// OneDNN matmul can fuse add operaton with automatic broadcasting along the
-// addend's dimensions that are 1s. When compatible Broadcast can be replaced by
-// Bitcast which is much cheaper. Computue new shape for the Bitcast.
+// OneDNN matmul can fuse add operation with automatic broadcasting along the
+// addend's dimensions that are 1s. When compatible, Broadcast can be replaced by
+// Bitcast, which is much cheaper. Compute new shape for the Bitcast.
 StatusOr<Shape> AdjustBiasShape(const HloInstruction* broadcast_instr,
                                 const Shape& dot_shape) {
   if (broadcast_instr->opcode() != HloOpcode::kBroadcast) {
@@ -162,8 +162,8 @@ StatusOr<Shape> AdjustBiasShape(const HloInstruction* broadcast_instr,
   }
   auto bcast = Cast<HloBroadcastInstruction>(broadcast_instr);
   Shape new_shape = bcast->shape();
-  // Broadcast instruction has "dimension" parameter along which its input's
-  // dim should not change. For exmaple,
+  // Broadcast instruction has "dimensions" parameter along which its input's
+  // dimensions should not change. For example,
   //      dot = f32[3,4,5,6] dot(...)
   //      arg = f32[3,6]{1,0} parameter(0)
   //      broad = f32[3,4,5,6]{3,2,1,0} broadcast(arg), dimensions={0,3}
@@ -191,8 +191,8 @@ StatusOr<Shape> AdjustBiasShape(const HloInstruction* broadcast_instr,
   }
   new_shape = ShapeUtil::DeleteDimensions(dims_to_delete, new_shape);
 
-  // New shape for bias should satisfy the condition: rank(new_shape) <=
-  // rank(dot).
+  // New shape for bias should satisfy the condition:
+  //   rank(new_shape) <= rank(dot).
   if (new_shape.rank() > dot_shape.rank()) {
     return absl::CancelledError(
         "Bias shape could not be adjusted for a fusion.");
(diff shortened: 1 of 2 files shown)


xla/service/cpu/onednn_matmul_rewriter.cc
```diff
@@ -178,41 +151,155 @@ auto GELUActivation(HloInstruction* instr, HloInstruction** src) {
...
+
   return new_shape;
 };
 
+inline bool IsOperandFusible(HloInstruction* operand, HloInstruction* dot) {
+  // Check if one of binary operand shape is compatible with matmul for fusion.
+  // An operand is fusable if
+  //    1. rank(oeprand) <= rank(dot) and
+  //    2. Starting from the last dim in backward direction, the dimension
+  //       size of operand is either 1 or same to dot.
```
penpornk, line 209: Nit: grammar/typos.
```suggestion
  // Check if the operand's shape is compatible with matmul for fusion.
  // An operand is fusable if
  //    1. rank(operand) <= rank(dot) and
  //    2. Starting from the last dim in backward direction, the dimension
  //       size of operand is either 1 or same to dot.
```

xla/service/cpu/onednn_matmul_rewriter.cc
```diff
@@ -178,41 +151,155 @@ auto GELUActivation(HloInstruction* instr, HloInstruction** src) {
...
+    int dot_idx = dot_dims.size() - 1;
+    for (; operand_idx >= 0; --operand_idx, --dot_idx) {
+      if (operand_dims[operand_idx] != 1 &&
+          operand_dims[operand_idx] != dot_dims[dot_idx])
+        return false;
+    }
+    return true;
+  } else {
+    return false;
+  }
```
penpornk, line 223: Nit: I think fewer nested code is easier to understand.
```suggestion
  if (operand_dims.size() > dot_dims.size()) return false;
  int operand_idx = operand_dims.size() - 1;
  int dot_idx = dot_dims.size() - 1;
  for (; operand_idx >= 0; --operand_idx, --dot_idx) {
    if (operand_dims[operand_idx] != 1 &&
        operand_dims[operand_idx] != dot_dims[dot_idx])
      return false;
  }
  return true;
```

xla/service/cpu/onednn_matmul_rewriter.cc
```diff
@@ -178,41 +151,155 @@ auto GELUActivation(HloInstruction* instr, HloInstruction** src) {
...
+      auto output_shape = instr->shape();
+      bool is_same_type = ShapeUtil::SameElementType(input_shape, output_shape);
+      bool has_equal_num_elems = ShapeUtil::ElementsIn(input_shape) ==
+                                 ShapeUtil::ElementsIn(output_shape);
+      bool has_rowmajor_layout =
+          IsRowMajor(input_shape) && IsRowMajor(output_shape);
+      return is_same_type && has_equal_num_elems && has_rowmajor_layout;
+    } else {
+      false;
+    }
```
penpornk, line 276: Same here. Early return may be a bit simpler.
```suggestion
    if (!instr) return false;
    auto input_shape = instr->operand(0)->shape();
    auto output_shape = instr->shape();
    bool is_same_type = ShapeUtil::SameElementType(input_shape, output_shape);
    bool has_equal_num_elems = ShapeUtil::ElementsIn(input_shape) ==
                                ShapeUtil::ElementsIn(output_shape);
    bool has_rowmajor_layout =
        IsRowMajor(input_shape) && IsRowMajor(output_shape);
    return is_same_type && has_equal_num_elems && has_rowmajor_layout;
```

xla/service/cpu/onednn_matmul_rewriter.cc
```diff
@@ -331,33 +417,41 @@ class OneDnnMatMulRewriteVisitor : public DfsHloRewriteVisitor {
...
-                           return !ShapeUtil::IsEffectiveScalar(ins->shape());
-                         })
-                         .WithOneUser())
-            .WithOneUser();
-
-    auto addend_reshape =
-        m::Bitcast(m::Op(&bitcast_input).WithOneUser()).WithOneUser();
+    // Try to do a fusion for Dot(onednn-matmul) + Add. However,
+    // HLO Add instruction might receive the addends after additional
+    // processing like Broadcast, Bitcast, Convert etc. applied to the raw
```
penpornk, line 422: Nit: grammar
```suggestion
    // processing like Broadcast, Bitcast, Convert, etc. is applied to the raw
```

xla/tests/onednn_matmul_test.cc
```diff
@@ -273,6 +275,27 @@ TEST_F(MatmulTest, SimpleTestF32TransposeBWithBiasAddFusion) {
...
+  const char* matmul_module_str = R"(
+  HloModule matmul.test.f32, entry_computation_layout={(f32[12288,2]{1,0},f32[2,1024]{1,0})->f32[32,384,1024]{2,1,0}}
+
+  ENTRY matmul.test.1 {
+    arg.0 = f32[12288,2]{1,0} parameter(0), parameter_replication={false}
+    arg.1 = f32[2,1024]{1,0} parameter(1), parameter_replication={false}
+    dot.0 = f32[12288,1024]{1,0} dot(arg.0, arg.1), lhs_contracting_dims={1}, rhs_contracting_dims={0}
+    reshape.0 = f32[32,384,1024]{2,1,0} reshape(dot.0)
+    constant.0 = f32[1,384,1024]{2,1,0} constant(15)
+    reshape.1 = f32[384,1024]{1,0} reshape(constant.0)
```
penpornk, line 288: Why do we need `reshape`? Can't we just make `constant.0` have 2 dimensions since the beginning?

penpornk (changes requested):
Thank you for the PR and sorry for the delay! I have minor comments.

xla/tests/onednn_matmul_test.cc
mdfaijul: This pattern arises in the BERT model. The purpose of this test is NOT to fuse the addend because oneDNN cannot support this fusion. The existing code crashes because we fused addend mistakenly. This PR essentially fixes this bug.

mdfaijul:
@penpornk Thanks for the review comments.  I have addressed those. Please take a look.

xla/tests/onednn_matmul_test.cc
penpornk: Thank you for the explanation! When you have time, would you mind making another PR annotating the test cases with the models they are from (in the comment, only for the tests that are from real model patterns)?

penpornk (approved):
Thank you for the quick changes!

xla/tests/onednn_matmul_test.cc
mdfaijul: Thanks! It is a great suggestion. We will make a PR.

PR #8480, opened 2024-01-15
[XLA:GPU] Explicitly set addrspace for SPIR target

Solaryee:
This is a sub PR to add SPIR target in openxla.
Instead of generic addrspace, SPIR target requires explicitly set global addrspace for kernel args/global variables, and private/local addrspace for local variables.


xla/service/llvm_ir/llvm_util.cc
```diff
@@ -377,8 +378,11 @@ llvm::AllocaInst* EmitAllocaAtFunctionEntryWithCount(llvm::Type* type,
   llvm::Function* function = b->GetInsertBlock()->getParent();
   b->SetInsertPoint(&function->getEntryBlock(),
                     function->getEntryBlock().getFirstInsertionPt());
+  llvm::Module* module = b->GetInsertBlock()->getModule();
+  // Explicitly set private addrspace for SPIR target.
+  int addrspace = llvm::Triple(module->getTargetTriple()).isSPIR() ? 5 : 0;
```
penpornk, line 383: Could you please document (in the comment) what address space 5 means for SPIR? It's specific to Intel Accelerators, right? It would be great if you could add the link to the specification, e.g., this 
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst#address-spaces

penpornk (changes requested):

xla/service/llvm_ir/llvm_util.cc
```diff
@@ -377,8 +377,9 @@ llvm::AllocaInst* EmitAllocaAtFunctionEntryWithCount(llvm::Type* type,
   llvm::Function* function = b->GetInsertBlock()->getParent();
   b->SetInsertPoint(&function->getEntryBlock(),
                     function->getEntryBlock().getFirstInsertionPt());
+  // Some backends requires explicit local addrspace.
```
penpornk, line 380: By adding a comment, I meant something like this. (Please also apply a code formatter.)
```suggestion
  // Some backends requires explicit local addrspace. addrspace 5 maps to Intel accelerators.
  // Source: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVRepresentationInLLVM.rst#address-spaces
```

xla/service/llvm_ir/llvm_util.cc
ShengYang1: This change is not related to addrspace representation in spir. Just as the title mentioned, we need explicitly set private addrspace(5) instead of generic addrspace(0, by default) for SPIR target, so I made this change.

More words about addrspace representations in spir. spir backend uses different addrspace representations [link](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/0aab1249acbfb8d206c24988f0974638c290e931/lib/SPIRV/SPIRVInternal.h#L181) with nvptx backend [link](https://llvm.org/docs/NVPTXUsage.html#address-spaces). To make xla codegen simple(avoiding changing addrspace based on device backend everywhere), I plan to use nvptx addrspace representations for spirv backend, and do a mapping from nvptx addrspace to spirv in SPIRV-LLVM-Translator, which only needs several lines changes. But this is not related to this PR, I will submit another PR to cover this.

penpornk: Got it. Thank you for the explanation!

penpornk (approved):

penpornk:
> This PR was rolled back in [a586c4f](https://github.com/openxla/xla/commit/a586c4fbd870587f77f0f9346c5cd0885979c80d)!

This was because it caused some OOM issues.

PR #8503, opened 2024-01-16
[ROCM] fixing build brake: 'operator=='

pemeliya:
The problem comes from this PR: https://github.com/openxla/xla/pull/8469

It seems that, CUDA compiler just gives a warning here:
_warning: defaulted comparison operators are a C++20 extension [-Wc++20-extensions]_
that's why it remained undiscovered. 
But on ROCM this generates a compile error as we are limited to C++17 now.


pemeliya:
already fixed..

PR #8506, opened 2024-01-16
[ROCm] Move device libs path initialization into CompileToHsaco

draganmladjenovic:
This fixes sporadic crashes in multithreaded_compilation_test_gpu.


penpornk (approved):
Thank you for the fix!

PR #8545, opened 2024-01-17
The CHECK macros can give warnings or errors on redefinition

elfringham:
These macros can be defined in more than one external project leading to warnings or errors as there is no guard against redefinition in abseil. So ensure that the abseil defines come first to prevent such issues.

### DIFF
--- a/xla/hlo/ir/hlo_computation.h
+++ b/xla/hlo/ir/hlo_computation.h
@@ -26,6 +26,8 @@ limitations under the License.
 #include "absl/container/flat_hash_map.h"
 #include "absl/container/flat_hash_set.h"
 #include "absl/functional/function_ref.h"
+#include "absl/log/check.h"
+#include "absl/log/log.h"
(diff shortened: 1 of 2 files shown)


xla/hlo/ir/hlo_computation.h
```diff
@@ -26,6 +26,8 @@ limitations under the License.
 #include "absl/container/flat_hash_map.h"
 #include "absl/container/flat_hash_set.h"
 #include "absl/functional/function_ref.h"
+#include "absl/log/check.h"
+#include "absl/log/log.h"
```
ddunl, line 30: I think this header needs corresponding target to be added in the BUILD file

elfringham: Done

akuegel (approved):

ddunl:
I'll work on rolling this forward. Somehow this broke TF's windows build

PR #8718, opened 2024-01-23
[NVIDIA XLA] Improve logic to avoid type promotion in float normalization

Tixxx:
We used to rely on the flag IsCollectiveCalledComputation to skip type promotion, but that mechanism is seen to be unsafe since a lot of places where computations get cloned don't set that flag. Computation deduplicator might also mess up the flag.
Instead, to be completely safe, we will change the behavior to do the following:
1. Remove 1 instance of ComputationDeduplicator which runs very early in the pipeline. It's currently run twice: once during simplification; once after all the pipelines and before IR emission. Keeping the last instance should be enough.
2. In the float normalization pass, we go through all computations and try to create a collection of all comps associated with collectives and use this collection in the floatNormalizationVisitor to avoid type promotion for certain computations. This is safer compared to the IsCollectiveCalledComputation flag since we are doing all analysis locally without the interference from other passes.

A follow-up change is to remove IsCollectiveCalledComputation which is only used in floatNormalization for now.

### DIFF
--- a/xla/service/float_normalization_test.cc
+++ b/xla/service/float_normalization_test.cc
@@ -691,7 +691,8 @@ TEST_F(FloatNormalizationTest, ConvertBeforeTuple) {
       F32);
 }
 
-TEST_F(FloatNormalizationNoComputeSupportTest, NoNormalizationForClonedModule) {
+TEST_F(FloatNormalizationNoComputeSupportTest,
+       NoNormalizationWithCollectiveCalledComputationSetToFalse) {
   auto module = CreateNewVerifiedModule();
   HloComputation::Builder sum_builder("sum");
   auto x = sum_builder.AddInstruction(HloInstruction::CreateParameter(
@@ -717,16 +718,31 @@ TEST_F(FloatNormalizationNoComputeSupportTest, NoNormalizationForClonedModule) {
                                       /*use_global_device_ids=*/false));
 
   auto computation = module->AddEntryComputation(builder.Build());
+  // Manually create a new computation which has the
+  // is_collective_called_computation flag set to false
+  // by default. It will replace the existing
+  // allreduce's to_apply region.
+  HloComputation::Builder sum2_builder("sum2");
+  auto x1 = sum2_builder.AddInstruction(HloInstruction::CreateParameter(
+      /*parameter_number=*/0, ShapeUtil::MakeShape(BF16, {}), "x1"));
+  auto y1 = sum2_builder.AddInstruction(HloInstruction::CreateParameter(
+      /*parameter_number=*/1, ShapeUtil::MakeShape(BF16, {}), "y1"));
+  sum2_builder.AddInstruction(HloInstruction::CreateBinary(
+      ShapeUtil::MakeShape(BF16, {}), HloOpcode::kAdd, x1, y1));
 
-  auto cloned_module = module->Clone();
-  // Since we skip processing to_apply region, nothing should change in the
-  // cloned module HLO.
-  EXPECT_FALSE(Normalize(cloned_module.get()));
-  HloInstruction* cloned_root =
-      cloned_module->entry_computation()->root_instruction();
-  EXPECT_EQ(cloned_root->shape().element_type(), BF16);
-  EXPECT_EQ(cloned_root->to_apply()->root_instruction()->opcode(),
-            HloOpcode::kAdd);
+  HloComputation* reduction2 =
+      module->AddEmbeddedComputation(sum2_builder.Build());
+
+  absl::flat_hash_map<HloComputation*, HloComputation*> replacement;
+  replacement[reduction] = reduction2;
+  module->ReplaceComputations(replacement);
+
+  EXPECT_FALSE(Normalize(module.get()));
+  // We don't rely on the flag to determine type-promotion anymore,
+  // allreduce's computation should not be promoted.
+  HloInstruction* root = module->entry_computation()->root_instruction();
+  EXPECT_EQ(root->shape().element_type(), BF16);
+  EXPECT_EQ(root->to_apply()->root_instruction()->opcode(), HloOpcode::kAdd);
 }
 
 }  // namespace xla
--- a/xla/service/float_normalization_test.cc
+++ b/xla/service/float_normalization_test.cc
@@ -691,4 +691,42 @@ TEST_F(FloatNormalizationTest, ConvertBeforeTuple) {
       F32);
 }
 
+TEST_F(FloatNormalizationNoComputeSupportTest, NoNormalizationForClonedModule) {
+  auto module = CreateNewVerifiedModule();
+  HloComputation::Builder sum_builder("sum");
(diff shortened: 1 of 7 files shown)


cheshire:
So is this change NFC? I'm a bit confused from the description : it does sound like you are changing the behavior?

cheshire:
Ah I understand now I think, is this for the regression from not compiling modules separately? Could we add tests then?

Tixxx:
> Ah I understand now I think, is this for the regression from not compiling modules separately? Could we add tests then?

This is not a NFC, i change the description a bit to reflect it.
I wasn't aware of the module splitting regression, this is to fix another internal error where allreduce's to_apply region is incorrectly promoted. Though it sounds like that issue is caused by the same root cause and therefore should be fixed by this pr. Let me add a test based on that regression then.

Tixxx:
@cheshire I updated the pr to remove the redundant instance of ComputationDeduplicator which runs at the very early stage in during compilation. Details are in the description. Can you take a look? Thanks.

xla/service/float_normalization_test.cc
```diff
@@ -691,4 +691,42 @@ TEST_F(FloatNormalizationTest, ConvertBeforeTuple) {
...
+  HloInstruction* crs = builder.AddInstruction(
+      HloInstruction::CreateAllReduce(bf16_shape_a, {a}, reduction,
+                                      /*replica_groups=*/{},
+                                      /*constrain_layout=*/false,
+                                      /*channel_id=*/std::nullopt,
+                                      /*use_global_device_ids=*/false));
+
+  auto computation = module->AddEntryComputation(builder.Build());
+
+  auto cloned_module = module->Clone();
```
joker-eph, line 721: Why do you need to clone here?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -858,8 +858,6 @@ absl::Status GpuCompiler::OptimizeHloModule(
       pipeline.AddPass<AlgebraicSimplifier>(layout_insensitive_algsimp_opts);
     }();
 
-    pipeline.AddPass<HloComputationDeduplicator>(
-        /*mark_fusion_duplications=*/false);
```
joker-eph, line 862: Is this intended as part of this PR?

xla/service/float_normalization_test.cc
Tixxx: The original failure is when the model is cloned, the IsCollectiveCalledComputation is not carried over properly which results in incorrect type-promotion. I'm adding that as a case here to verify cloning has no impact on the type-promotion bahavior.

xla/service/gpu/gpu_compiler.cc
Tixxx: Yes, this will dedup comps incorrectly, at least without [this](https://github.com/openxla/xla/pull/8800) it will.
We call flattenCallGraph multiple times anyway so this call is redundant.

xla/service/float_normalization_test.cc
joker-eph: That seems like a convoluted way to test this pass: can't we directly create the desired IR to test without cloning? Is it even valid IR to have it without the IsCollectiveCalledComputation carried?

What if he cloned method if fixed to carry over the IsCollectiveCalledComputation? Is this test still relevant?

xla/service/gpu/gpu_compiler.cc
joker-eph: That still seems like a separate change to me, which should thus have its own separate PR: is it possible to split this out?

xla/service/float_normalization_test.cc
Tixxx: The plan is to remove IsCollectiveCalledComputation completely, so I don't recommend trying to fix that.
That flag along with other flags like (isCustomCall and etc) are not part of the IR. They are only present in the HloInstruction object and relying on setters to set them correctly. The plan is to remove these flags and provide abstraction helpers to deduce from HLO graph directly.

xla/service/gpu/gpu_compiler.cc
Tixxx: Without this, some scenario will continue to fail since comps are still incorrectly comped. We can wait for the above fix to land then we can either leave this as is or remove it in another pr.

xla/service/float_normalization_test.cc
joker-eph: > The plan is to remove IsCollectiveCalledComputation completely

For all practical purpose here: this is a similar effect as to fixing the cloneModule though. Basically it means that "CloneModule"  is not relevant to this test as far as I can tell.

Thought experiment: let's say we remove these flags, is this test becoming useless?

> That flag along with other flags like (isCustomCall and etc) are not part of the IR. They are only present in the HloInstruction object and relying on setters to set them correctly. 

Ok, I guess we're using "part of the IR" differently here, but let's not get hung up on terminology: you have the HLOInstruction objects at hand here, so you could still set them in the test directly if it matters right? But does it matter for this test?

Tixxx: I see where you are coming at, in terms of exercising code path, this test doesn't really add additional coverage compared to other tests in this file. And yes, when we remove the flags, this test is useless.
In fact, all the failures we see now are caused by incorrect values of the flag. I can specifically set an incorrect flag in this test and verify the HLO is correct even with a wrong flag. But it'd need to removed when we remove the flag.

joker-eph: Right, so the advantage of setting `is_collective_called_computation_` flag explicitly is that indeed we'll be able to clean this up when the flag is removed. Instead of having an obsolete test with a weird cloneModule that in the future won't make sense.

Tixxx: Well it looks like there's no easy way to manually set the flag to false outside of the class. I have tried mocking HloComputation with GMOCK, but it wont work when using that computation. For now, I change the test behavior to create a computation for allreduce and then replace the computation with another one that has the flag set to false. It's not the best but more explicit than using module->clone.

xla/service/gpu/gpu_compiler.cc
jaro-sevcik: How about making float normalization itself duplicate the computation if it is shared between a collective and non-collective instruction? I have sketched that approach in a draft PR: https://github.com/openxla/xla/pull/8896

With that approach, float normalization will not rely on any sharing invariants.

What do you think?

Tixxx: I'm ok with either approach. One question I had is: I think we all agreed that collective comps cannot be shared with other ops, so if we actually see this case, we should just error out instead of trying to work around an illegal IR.
IMO, the correct flow should be that HloVerifier takes the responsibility to check for illegal IR, and then float norm pass just does a local analysis to skip type promotion.

Tixxx:
fixed with https://github.com/openxla/xla/pull/8896, closing this

PR #8725, opened 2024-01-23
Mark collective called computations in custom sharding lowering

jaro-sevcik:
When inlining a custom sharding code into the graph, cloned all-reduce computations must be explicitly marked as collective-called because the CloneWithNewOperands method clones the computation without marking it as collective-called. (There is code that marks all-reduce computations as collective-called in the HloAllReduceInstructionBase constructor, but that only marks the original reduce computation, not the cloned one.)

### DIFF
--- a/xla/python/custom_call_sharding.cc
+++ b/xla/python/custom_call_sharding.cc
@@ -95,6 +95,11 @@ HloInstruction* InlineHloComputation(HloInstruction* instruction,
       }
       auto* new_inst = builder->AddInstruction(
           inst->CloneWithNewOperands(inst->shape(), new_operands, &context));
+      if (HloAllReduceInstructionBase::ClassOf(new_inst)) {
+        // The cloning duplicated the computation for the new module, let us
+        // make sure the computation is marked as collective-called.
+        new_inst->to_apply()->SetCollectiveCallInstruction(new_inst);
+      }
       HloChannelInstruction* channel_instr =
           DynCast<HloChannelInstruction>(new_inst);
       if (channel_instr && channel_instr->channel_id().has_value()) {


jaro-sevcik:
This is a draft of a fix for a crasher (training PAX on multi-GPU), where custom sharding does not mark a computation as collective-called even though it is used by an all-reduce instruction. Do you think it makes sense to go in this direction? (Let me ask before embarking on writing a test for this.)

An **alternative** fix is to make `CloneWithNewOperands` automatically mark the computation as collective-called when cloning an all-reduce instruction. There is already an attempt to do just that, but it does not always work because the computation can be cloned later and thus lose the collective-called flag. An alternative collective-called marking could be done either in the common code ([`HloInstruction::CloneWithNewOperands`, after computation cloning](https://github.com/openxla/xla/blob/0ead25a4ab8bce6f071945082747a782080630ae/xla/hlo/ir/hlo_instruction.cc#L2466)) or in the computation duplicator (in [HloModule::DeepCloneComputation](https://github.com/openxla/xla/blob/0ead25a4ab8bce6f071945082747a782080630ae/xla/hlo/ir/hlo_module.cc#L1144)) or we could clone the computation eagerly in the [HloAllReduceInstructionBase constructor](https://github.com/openxla/xla/blob/0ead25a4ab8bce6f071945082747a782080630ae/xla/hlo/ir/hlo_instructions.cc#L959) if it comes from a different module, and *then* mark it (rather than marking the original computation, possibly in a different module). I prefer this last alternative because it is the most fool-proof one, but it does make the deep-cloning of computation happen in a different place for all-reduce instructions. All of these alternatives lend themselves to easier testing than this fix, but they are a bit more invasive in the core machinery.

### Invariant

There is also a high-level question on protecting invariants - it seems that the compiler (e.g., float_normalization) assumes that computations are not shared between collective and non-collective instructions, but that invariant does not seem to be explicitly checked. It just crashes later in the pipeline, when bad IR gets produced. For illustration, see [this gist](https://gist.github.com/jaro-sevcik/e44c44753e5bd39bec2a48a1ffffd741) for an example of bad hlo that crashes on both GPU and CPU.

### Some more details on the crasher

In the crasher, the incorrect marking from custom sharding code triggers the computation deduplicator pass to share the reduction computation between reduce and all-reduce instructions. Later, the reduce-scatter-creator pass marks the shared computation as is-collective-called again. That finally makes the float normalizer skip normalization of the reduction computation (which is a bf16 add), although it does normalize the calling reduce instruction (the return value and operands are all f32). So we end up with a f32 reduce with bf16 reduction computation.

cheshire (changes requested):
1) We usually use DynCast

2) What about tests?

kamaljeeti:
Hi @jaro-sevcik , can you have a look at the comments? Thanks.

jaro-sevcik:
Closing because it was superseded by https://github.com/openxla/xla/pull/8896

PR #8763, opened 2024-01-24
add mask support for flash attention

Cjkkkk:
* flash attention only support causal mask right now, cuDNN also supports take in a mask directly. Add mask support for flash attention.

### DIFF
--- a/xla/service/gpu/cudnn_fused_mha_rewriter.cc
+++ b/xla/service/gpu/cudnn_fused_mha_rewriter.cc
@@ -471,7 +471,9 @@ StatusOr<bool> IsFlashAttention(
   TF_RET_CHECK(hidden_dim.size() == 1);
   auto is_fixed_topology =
       (custom_call_name == kCudnnfMHAScaleBiasSoftmaxDropoutCallTarget ||
-       custom_call_name == kCudnnfMHAScaleBiasSoftmaxCallTarget);
+       custom_call_name == kCudnnfMHAScaleBiasSoftmaxCallTarget ||
+       custom_call_name == kCudnnfMHAScaleBiasMaskSoftmaxDropoutCallTarget ||
+       custom_call_name == kCudnnfMHAScaleBiasMaskSoftmaxCallTarget);
 
   auto is_seqlen_supported = seq_q[0] > 512 && seq_k[0] > 512 &&
                              seq_q[0] % 64 == 0 && seq_k[0] % 64 == 0;
@@ -736,7 +738,6 @@ MatchFwdResult MatchBmm1ScaleBiasMaskSoftmaxDropoutBmm2(
       matched_result.has_match = false;
       return matched_result;
     }
-    matched_result.is_causal_mask |= IsCausalMaskPattern(mask);
     if (has_dropout) {
       // Found BMM1 - (Scale) - (bias) - Mask - Softmax - dropout - BMM2
       matched_result.matched_custom_call_name =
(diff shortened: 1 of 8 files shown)


Cjkkkk:
@akuegel Hey Adrian, could you take a look at this PR?

xla/service/gpu/cudnn_fused_mha_rewriter.cc
```diff
@@ -736,7 +738,6 @@ MatchFwdResult MatchBmm1ScaleBiasMaskSoftmaxDropoutBmm2(
       matched_result.has_match = false;
       return matched_result;
     }
-    matched_result.is_causal_mask |= IsCausalMaskPattern(mask);
```
akuegel, line 739: Why do we need to remove this line?

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -2709,6 +2709,136 @@ class FlashAttentionBMMScaleBiasSoftmaxBMM : public MultiHeadedAttentionTest {
...
+    auto lhs_bmm1_literal =
+        GetInput4DLiteral<T>({2, 6, 1024, 64}, {3, 2, 1, 0});
+    auto rhs_bmm1_literal =
+        GetInput4DLiteral<T>({2, 6, 64, 1024}, {3, 2, 1, 0});
+    auto rhs_bmm2_literal =
+        GetInput4DLiteral<T>({2, 6, 1024, 64}, {3, 2, 1, 0});
+    auto mask_literal = GetMask4DLiteral({2, 6, 1024, 1024}, {3, 2, 1, 0});
+    auto do_literal = GetInput4DLiteral<T>({2, 6, 1024, 64}, {3, 2, 1, 0});
+    std::string hlo_string = "";
+    hlo_string =
```
akuegel, line 2834: Why not just directly assign?

xla/service/gpu/cudnn_fused_mha_rewriter.cc
Cjkkkk: Current CausalMask pattern match only match a causal mask that is added to the bmm1 output. Would be better to do this in a separate PR.

xla/service/gpu/tests/gpu_fused_mha_test.cc
Cjkkkk: Fixed.

xla/service/gpu/cudnn_fused_mha_rewriter.cc
Cjkkkk: Also require to rewrite some parts of rewriter logic to consider cases where bias and mask could both be causalMask.

akuegel: If I understand correctly, not removing this line would currently be an issue and you intend to add some additional logic in another PR? Sounds good.

akuegel (approved):

Cjkkkk:
@akuegel Not moving this line will not cause an issue as it will just not match. But remove this makes it clear that mask as causal mask is not supported right now.

akuegel:
> @akuegel Not moving this line will not cause an issue as it will just not match. But remove this makes it clear that mask as causal mask is not supported right now.

Makes sense, thanks for the explanation.

akuegel (approved):

Cjkkkk:
@akuegel Seems like the Google internal checks FAILED?

akuegel:
> @akuegel Seems like the Google internal checks FAILED?

That is not a real failure, it complained about a missing TESTED= tag. I have added it. Nothing for you to worry about :)
But the internal approval will probably only happen during US working hours.

PR #8800, opened 2024-01-25
Only deduplicate computations with the same caller opcode

jaro-sevcik:
This patch avoids using the unreliable IsCollectiveCalledComputation flag. Instead it only deduplicates computations with the same caller's opcode. This prevents deduplicating all-reduce and reduce computation, which is known to cause problems later in the pipeline (in float normalization).


xla/service/hlo_computation_deduplicator.cc
```diff
@@ -78,20 +78,42 @@ StatusOr<bool> HloComputationDeduplicator::Run(
     }
     return false;
   };
-  for (HloComputation* comp :
-       module->MakeComputationPostOrder(execution_threads)) {
+  // Let us collect the opcode of the callers of each computation. We will only
+  // deduplicate computations with the same caller opcode.
```
joker-eph, line 82: Nit: you're describing "what" the code does, but not "why": the latter is in general the most important form of documentation (the "what" can always be recovered from the code).
Can you elaborate on the invariant to preserve?

Also if different callers opcode can't share the same computation, that should be checked in the verifier: can we add this check?

xla/service/hlo_computation_deduplicator_test.cc
```diff
@@ -618,5 +618,35 @@ TEST_F(HloComputationDeduplicatorTest, DontDeduplicateReduceAllReduce) {
...
+    rd2 = s32[] all-reduce(Arg_1.1), to_apply=add.2
+    ROOT multiply.14 = s32[] multiply(rd1, rd2)
+  }
+  )";
+  auto computation_names = RunDeduplicatePass(text, /*expect_true=*/true);
+  EXPECT_EQ(computation_names.size(), 2);
+  for (auto name : computation_names) {
+    EXPECT_NE(name, "add.2");
+  }
+}
```
Tixxx, line 650: can you add another test with a reduce and an allreduce. it would be great to test that after the dedup pass, the hlo graph won't change.

joker-eph, line 650: Isn't your change about not deduplicating something that was deduplicated before? If so it's not clear to me what in this test shows the change? Wouldn't this test pass already before your change?

jaro-sevcik: Such a test is already here - see the DontDeduplicateReduceAllReduce test just above. Or have you had something else in mind?

(For completeness, the test above passed because of the IsCollectiveCalledComputation flag, now it passes because of the opcode comparison.)

jaro-sevcik: Right, my change is indeed about not deduplicating something that was deduplicated before, but as a somewhat of a side effect, it is also deduplicating something that was not deduplicated before (computations of two all-reduces). This unit test is documenting that.

xla/service/hlo_computation_deduplicator.cc
jaro-sevcik: As a newbie to XLA, I am on a thin ice when describing pipeline-wide invariants, but let me take a stab at a better comment.

xla/service/hlo_computation_deduplicator_test.cc
Tixxx: Yea similar to what @joker-eph  [suggested](https://github.com/openxla/xla/pull/8718#discussion_r1467118173) in the other PR. We'd need to be sure that this new mechanism is more reliable than using the IsCollectiveCalledComputation. I'm thinking something like this:
1. create a module with reduce and allreduce/reduceScatter
2. manually create a new computation(IsCollectiveCalledComputation is off by default)
3. replace the computation of allreduce with the one in 2 
4. run dedup pass and make sure HLO doesn't change.

nouiz: This is independent. It is a partial fix. The full fix will need many PRs:
- One of yours that you already have.
- this one.
- Others that add call to HLO and LLVM verifiers more frequently (this should have been catch up much earlier)
- And the refactoring of the attribute. Maybe we want them to be removed and inferred when needed or they should be cloned correctly.

@Tixxx what you describe shouldn't block this PR.

joker-eph: > my change is indeed about not deduplicating something that was deduplicated before,

I would expect a test for this though?

jaro-sevcik: I have added a test that checks the deduplicator with a wrongly set IsCollectiveCalledComputation flag now. Is that what you had in mind?

jaro-sevcik: Add the test. Is that what you had in mind?

jaro-sevcik (commented):
As requested by reviewers, I've added the motivating test that would fail without the patch (i.e., all-reduce and reduce sharing a cloned computation with a wrongly set IsCollectiveCalledComputation flag don't get dedup'ed).

It is not exactly the crashing scenario we see in Pax (where the cloned computation comes from a different module), but should be close enough.

jaro-sevcik:
For completeness, below is a small JAX repro of the crasher that we see on Pax (and that this patch partially fixes). This needs to be run on 8 gpu machine (I used 8xV100). Currently, this causes Jax crash. Before https://github.com/openxla/xla/commit/42d3fa191201941976cca9e37dc15e1abfc1190a, this program returned wrong result (because XLA miscompiles there).

```python
import jax
import jax.numpy as jnp
from jax.sharding import NamedSharding
from jax.experimental.custom_partitioning import custom_partitioning
from jax.experimental.pjit import pjit
from jax.sharding import PartitionSpec as P
from jax.sharding import Mesh

tp = jnp.bfloat16

def compute_shard(x):
    # This will create a reduction with bf16 add and an all-reduce
    # with bf16 add. Those will be inlined into the main function
    # by XLA. The inlining will mark both adds ad not-collective-called.
    s = jax.lax.reduce(x, jnp.array(0.0, dtype=tp), lambda a, b: a + b, [1])
    return jax.lax.psum(s, axis_name = 'x')

def partition(mesh, arg_shapes, result_shape):
    # Custom partitioner. The subtlety here is that the result is marked as
    # fully replicated. This is necessary for the repro because this will trigger
    # all-reduce -> reduce-scatter rewrite, which will in turn fix up the marking
    # of the all-reduce computation as collective-called after deduplication.
    # The collective-called mark will cause the float normalizer skip the reduce
    # computation even though it will normalize the reduce itself to f32.
    return mesh, compute_shard, NamedSharding(mesh, P(None, None)), (NamedSharding(mesh, P('x', None)),)

def infer_sharding_from_operands(mesh, arg_shapes, result_shape):
    # Dummy implementation, XLA does seem to call this on GPUs.
    return ((NamedSharding(mesh, P('x', None)),), NamedSharding(mesh, P(None, None)))

@custom_partitioning
def custom_op(x):
    # Let's just return a dummy value here, this is only used to infer shape,
    # values are ignored.
    return jnp.ones([8], dtype=tp)

custom_op.def_partition(
    infer_sharding_from_operands=infer_sharding_from_operands,
    partition=partition)

with Mesh(jax.devices(), ('x',)):
  x = jnp.ones([64, 32], dtype = tp).reshape([64, 32])
  # The correct result is [256, 256, 256, 256, 256, 256, 256, 256],
  # but JAX currently crashes on verification failure. Before
  # https://github.com/openxla/xla/commit/42d3fa191201941976cca9e37dc15e1abfc1190a,
  # this outputs wrong result - [0, 0, 0, 0, 0, 0, 0, 0, 0].
  print(pjit(custom_op, in_shardings=P('x'), out_shardings=P('x'))(x))
  ```

xla/service/hlo_computation_deduplicator_test.cc
Tixxx: Yep, that looks good. Thanks.

Tixxx (approved):

kamaljeeti:
Hi @cheshire, can you look into this once? Thanks.

jaro-sevcik:
Closing because it was superseded by https://github.com/openxla/xla/pull/8800

PR #8806, opened 2024-01-25
Prevent use after free in transpose_test

elfringham:
Ensure that parameters set in Options have a lifetime beyond the life of the Options as only a reference is taken.


ddunl (approved):

reedwm:
Looks like someone else coincidentally fixed this after you created this PR: 1f926834c8cffc06f94ba383a59d71067d3e3813. So I'll close this PR. Sorry about that, it is very annoying when this happens.

PR #8871, opened 2024-01-26
[XLA:GPU] Only allow cuDNN FMHA with CUDA >= 12

Cjkkkk:
* This PR is intended to add guard for using cuDNN FMHA with CUDA >= 12.


reedwm (approved):
Thanks for the fix!

PR #7963, opened 2023-12-20
[GPU] NCCL User Buffer Registration - Create CollectiveBFCAllocator in PJRT (3/3)

trevor-m:
If `collective_memory_size` is non-zero in the `GPUAllocatorConfig`, then a second BFCAllocator will be created for each device that uses `CollectiveMemoryAllocate` for its suballocator. For allocations with `memory_space == 1`, the allocators will be routed to this `CollectiveBFCAllocator` instead of the regular device memory allocator.

Main PR here: https://github.com/openxla/xla/pull/7843

PR #7962, opened 2023-12-20
[GPU] NCCL User Buffer Registration - Add CollectiveMemoryAllocate using ncclMemAlloc to SE (2/3)

trevor-m:
This PR adds new functions `CollectiveMemoryAllocate` and `CollectiveMemoryDeallocate` to the stream executor which are implemented using `ncclMemAlloc` and `ncclMemFree` for CUDA.

NCCL 2.19 is required for `ncclMemAlloc`/`ncclMemFree`

Main PR here: https://github.com/openxla/xla/pull/7843

PR #7843, opened 2023-12-15
[GPU] NCCL User Buffer registration (1/3)

trevor-m:
This PR enables XLA to take advantage of [NCCL User Buffer Registration](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html) in NCCL 2.19:
> User Buffer Registration is a feature that allows NCCL to directly send/receive/operate data through the user buffer without extra internal copy. It can accelerate collectives and reduce the resource usage (e.g. #channel usage).

NCCL supports this feature for all reduce, all gather, and reduce scatter. To use the feature, A buffer must be allocated with `ncclMemAlloc`. Then, the buffer must be registered using `ncclCommRegister` for all communicators. If capturing collectives using cuda graphs, `ncclCommRegister` can be skipped. Finally, when performing a collective using offsets to this buffer, all devices must have the same offset into the buffer.

This PR contains the following components to enable this feature:

1. Compiler
   - If flag `xla_gpu_enable_nccl_user_buffers` is true, a custom "colorer" function will be passed to `BufferAssignment` which will mark all buffers for all reduce, all gather, and reduce scatter to use alternate memory space `1`.
2. Runtime
   -  The StreamExecutorGpuClient will create a second `BFCAllocator` for each device which uses `ncclMemAlloc/ncclMemFree` as its suballocator. The amount of memory reserved can be configured using `GpuAllocatorConfig.collective_memory_size`.
   - `MultiDeviceAllocator` will read the `memory_space` value for each allocation and use it to route the allocation to the correct allocator. Since `memory_space` is not passed to the deallocate function, it uses a hash map to remember which memory space each address belongs to.
   - When running a collective, the first time a buffer is encountered for each communicator it is registered using `ncclCommRegister`. The collective buffers will use a slice of the `ncclMemAlloc`'d CollectiveBFCAllocator memory, but `ncclCommRegister` will automatically locate the base address and size to register the entire `ncclMemAlloc` memory. Once, cudagraph capture of nccl collectives is supported, this code can be removed.

NCCL 2.19 is required for `ncclCommRegister`.

PJRT Changes: https://github.com/openxla/xla/pull/7963
StreamExecutor Changes: https://github.com/openxla/xla/pull/7962

PR #8873, opened 2024-01-26
[GPU] NCCL User Buffers - collective memory allocation improvements

trevor-m:
Followup to https://github.com/openxla/xla/pull/7843 https://github.com/openxla/xla/pull/7962 https://github.com/openxla/xla/pull/7963

This PR aims to address the issue that `collective_memory_size` must be configured properly in order to use `--xla_gpu_enable_nccl_user_buffers=true`. 

* The default behavior sets the collective memory BFC allocator in "allow growth" mode. It will start without allocating any `ncclMemAlloc` memory and will grow as needed. If there are no collectives using the user buffers, no collective memory will ever be allocated.
* If the user instead wants to preallocate a block of memory like before, `XLA_PYTHON_CLIENT_COLLECTIVE_MEM_SIZE_MB` can be used.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -779,24 +779,17 @@ GetStreamExecutorGpuDeviceAllocator(
   }
 
   // Add any additional allocators for alternate memory spaces.
-  if (allocator_config.collective_memory_size != 0) {
-    bool preallocate;
-    TF_RETURN_IF_ERROR(tsl::ReadBoolFromEnvVar(
-        "TF_COLLECTIVE_MEMORY_PREALLOCATE", false, &preallocate));
-
-    for (const auto& ordinal_and_device : addressable_devices) {
-      TF_ASSIGN_OR_RETURN(
-          auto collective_bfc_allocator,
-          CreateCollectiveBFCAllocator(
-              ordinal_and_device.second->executor(),
-              /*allocator_memory=*/allocator_config.collective_memory_size,
-              /*preallocate=*/preallocate));
-      allocators.emplace_back(std::move(collective_bfc_allocator),
-                              ordinal_and_device.second->compute_stream(),
-                              /*memory_space=*/1);
-    }
+  for (const auto& ordinal_and_device : addressable_devices) {
+    TF_ASSIGN_OR_RETURN(
+        auto collective_bfc_allocator,
+        CreateCollectiveBFCAllocator(
+            ordinal_and_device.second->executor(),
+            /*memory_fraction=*/1.0 - allocator_config.memory_fraction,
+            allocator_config.collective_memory_size));
+    allocators.emplace_back(std::move(collective_bfc_allocator),
+                            ordinal_and_device.second->compute_stream(),
+                            /*memory_space=*/1);
   }
-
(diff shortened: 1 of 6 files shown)


trevor-m:
@PatriosTheGreat @cheshire @nouiz

nouiz:
If the 4GB is busted, which error is triggered? Can it print the WAR to set the env variable? This would give a better UI.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -778,13 +779,17 @@ GetStreamExecutorGpuDeviceAllocator(
...
+    bool preallocate;
+    TF_RETURN_IF_ERROR(tsl::ReadBoolFromEnvVar(
+        "TF_COLLECTIVE_MEMORY_PREALLOCATE", false, &preallocate));
+
     for (const auto& ordinal_and_device : addressable_devices) {
       TF_ASSIGN_OR_RETURN(
           auto collective_bfc_allocator,
           CreateCollectiveBFCAllocator(
               ordinal_and_device.second->executor(),
               /*allocator_memory=*/allocator_config.collective_memory_size,
```
PatriosTheGreat, line 791: Shouldn't we also subtract collective_memory_size from device available memory above at the BFC allocator in case of memory preallocation?
Currently we are relying on a fact that available device memory * allocator_config.memory_fraction > collective memory, which is a bit fragile.

trevor-m: I think this makes sense, because the buffers which are now going to the collective BFC pool previously were part of the standard BFC pool. This means we are not really reducing the memory available to the users model if the `XLA_PYTHON_CLIENT_COLLECTIVE_MEM_SIZE_MB` is tuned properly. However, if that value is set way higher than the user needs, they will end up starving the standard allocator which might cause more OOMs.

trevor-m:
> If the 4GB is busted, which error is triggered? Can it print the WAR to set the env variable? This would give a better UI.

The generic BFC allocator OOM message would appear:
https://github.com/openxla/xla/blob/main/third_party/tsl/tsl/framework/bfc_allocator.cc#L488

Since this code is used by all BFC allocators, maybe we can add a "help string" field in BFCAllocator which would be printed after the OOM message and allow it to be customized. @PatriosTheGreat thoughts?

trevor-m:
@PatriosTheGreat Hi Levon, I wanted to clarify that this PR is purely an improvement over the current code already in XLA and I think it's best to merge it. All of these changes only apply when `--xla_gpu_enable_nccl_user_buffers=true` is set which is not on by default. The nccl user buffer memory in XLA is already outside of the regular XLA reserved memory, so nothing is changing also in that regard.

PatriosTheGreat:
> @PatriosTheGreat Hi Levon, I wanted to clarify that this PR is purely an improvement over the current code already in XLA and I think it would be great if we can merge it. All of these changes only apply when `--xla_gpu_enable_nccl_user_buffers=true` is set which is not on by default. The nccl user buffer memory in XLA is already outside of the regular XLA reserved memory, so nothing is changing also in that regard.

Sorry I was a bit distracted by deterministic multi-host compilation. I'll try to merge this PR today maybe tomorrow.

PatriosTheGreat (approved):

PR #8896, opened 2024-01-29
Clone of collective computations during float normalization

jaro-sevcik:
If a computation is shared between collective and non-collective operation, this patch will make float normalization clone the shared computation and only normalize it for the non-collective operations. This prevents the compiler from wrongly normalizing all-reduce/reduce-scatter instruction's computation (e.g., if the computation is shared with a reduce instruction).

This aims to fix the bug described here: https://gist.github.com/jaro-sevcik/e50d90048eb7271a430d70e4d769945e

### DIFF
--- a/xla/service/float_normalization.cc
+++ b/xla/service/float_normalization.cc
@@ -18,18 +18,17 @@ limitations under the License.
 #include <vector>
 
 #include "absl/types/span.h"
-#include "tsl/platform/errors.h"
-#include "tsl/platform/logging.h"
 #include "xla/hlo/ir/dfs_hlo_visitor_with_default.h"
-#include "xla/hlo/ir/hlo_casting_utils.h"
 #include "xla/hlo/ir/hlo_computation.h"
 #include "xla/hlo/ir/hlo_instruction.h"
-#include "xla/hlo/ir/hlo_instructions.h"
 #include "xla/hlo/ir/hlo_opcode.h"
+#include "xla/service/call_graph.h"
 #include "xla/service/hlo_dce.h"
 #include "xla/service/tuple_simplifier.h"
 #include "xla/shape_util.h"
 #include "xla/xla_data.pb.h"
+#include "tsl/platform/errors.h"
+#include "tsl/platform/logging.h"
 
 namespace xla {
 
@@ -292,8 +291,12 @@ Status FloatNormalizationVisitor::ConvertCalledComputations(
   return OkStatus();
 }
 
-bool ShouldSkipCalledComputationsForInstruction(HloInstruction* hlo) {
-  return DynCast<HloAllReduceInstructionBase>(hlo) != nullptr;
+// Returns true if the called computations of the instruction should not
+// be touched by float normalization. In particular, we must not introduce
+// float conversions into collective reductions.
+bool ShouldAvoidNormalizingComputationsForInstruction(HloInstruction* hlo) {
+  return hlo->opcode() == HloOpcode::kAllReduce ||
+         hlo->opcode() == HloOpcode::kReduceScatter;
 }
 
 Status FloatNormalizationVisitor::HandleMultipleOutputs(HloInstruction* hlo) {
@@ -361,7 +364,7 @@ Status FloatNormalizationVisitor::HandleMultipleOutputs(HloInstruction* hlo) {
 
   std::vector<HloComputation*> low_precision_called_comps;
   for (auto* comp : hlo->called_computations()) {
-    if (ShouldSkipCalledComputationsForInstruction(hlo)) {
+    if (ShouldAvoidNormalizingComputationsForInstruction(hlo)) {
       continue;
     }
     bool comp_has_low_precision = false;
@@ -440,7 +443,7 @@ Status FloatNormalizationVisitor::HandleInstruction(HloInstruction* hlo) {
 
   std::vector<HloComputation*> low_precision_called_comps;
   for (auto* comp : hlo->called_computations()) {
-    if (ShouldSkipCalledComputationsForInstruction(hlo)) {
+    if (ShouldAvoidNormalizingComputationsForInstruction(hlo)) {
       continue;
     }
     bool comp_has_low_precision = false;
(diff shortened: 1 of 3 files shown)


xla/service/float_normalization.cc
```diff
@@ -290,6 +292,10 @@ Status FloatNormalizationVisitor::ConvertCalledComputations(
   return OkStatus();
 }
 
+bool ShouldSkipCalledComputationsForInstruction(HloInstruction* hlo) {
+  return DynCast<HloAllReduceInstructionBase>(hlo) != nullptr;
```
nouiz, line 296: Should we do that for all Collective and not just all-reduce?

xla/service/float_normalization_test.cc
```diff
@@ -586,6 +586,67 @@ TEST_F(FloatNormalizationNoComputeSupportTest,
   EXPECT_EQ(ShapeUtil::GetSubshape(crs->shape(), {1}).element_type(), BF16);
 }
 
+TEST_F(FloatNormalizationNoComputeSupportTest,
+       NormalizationClonesSharedApplyAllReduceAndReduce) {
+  auto module = CreateNewVerifiedModule();
```
nouiz, line 591: I personally prefer to build the graph with text HLO. I find it more readable.
Do as you prefer.

xla/service/float_normalization.cc
jaro-sevcik: Maybe. I had two reasons to do it his way:

- IsCollectiveCalledComputation is only set for HloAllReduceInstructionBase instructions and this patch is concerning with fixing the behavior of IsCollectiveCalledComputation.
- I am not sure what is the right predicate for all collectives (I know TJ had some way of doing this, but that actually required fixing the predicate, so I am reluctant to rely on that).

xla/service/float_normalization_test.cc
jaro-sevcik: I also prefer that, but this file avoids that way of building modules. When in Rome...

xla/service/float_normalization.cc
cheshire, line 296: Nit: we usually check opcode()

xla/service/float_normalization.cc
```diff
@@ -564,6 +570,58 @@ Status FloatNormalizationVisitor::Preprocess(HloInstruction* hlo) {
...
+  absl::flat_hash_set<HloComputation*> computations_to_skip;
+  // Let us first figure out what users each computation has.
+  struct ComputationUsers {
+    bool user_to_normalize_exists = false;
+    std::vector<HloInstruction*> users_to_skip_normalization;
+  };
+  absl::flat_hash_map<HloComputation*, ComputationUsers> computation_users;
+  for (auto& computation : computations_to_visit) {
+    for (auto* instruction : computation->instructions()) {
+      if (!instruction->has_called_computations()) continue;
```
cheshire, line 591: Nit: can we just remove this if-statement?

xla/service/float_normalization.cc
```diff
@@ -564,6 +570,58 @@ Status FloatNormalizationVisitor::Preprocess(HloInstruction* hlo) {
...
+      for (auto& called : instruction->called_computations()) {
```
cheshire, line 592: Nit: explicit types preferred, here it's not entirely obvious

xla/service/float_normalization.cc
```diff
@@ -290,6 +292,10 @@ Status FloatNormalizationVisitor::ConvertCalledComputations(
   return OkStatus();
 }
 
+bool ShouldSkipCalledComputationsForInstruction(HloInstruction* hlo) {
```
cheshire, line 295: Nit: docstring, possibly more self-descriptive name?

xla/service/float_normalization.cc
```diff
@@ -564,6 +570,58 @@ Status FloatNormalizationVisitor::Preprocess(HloInstruction* hlo) {
...
+              instruction);
+        } else {
+          computation_users[called].user_to_normalize_exists = true;
+        }
+      }
+    }
+  }
+
+  // If a computation is shared, create a clone, and replace it its the users.
+  for (auto [comp, users] : computation_users) {
```
cheshire, line 604: Nit: it's quite a bit of mental overhead to follow the code above and to understand what it actually does. WDYT about instead constructing a CallGraph (what passes usually do), and then simply iterating over `GetComputationCallers()` for each instruction? Then we don't need a 2-pass system.

cheshire (changes requested):

xla/service/float_normalization.cc
jaro-sevcik: I have just switched to explicit opcode, as @cheshire suggested.

xla/service/float_normalization.cc
```diff
@@ -564,6 +570,58 @@ Status FloatNormalizationVisitor::Preprocess(HloInstruction* hlo) {
...
 
+// We must avoid normalizing computations that have non-normalizing users
+// (e.g., all-reduce's computations should not be normalized). If a
+// computation is shared between normalizing and non-normalizing users, we will
+// clone the computation for the non-normalizing users so that it can be left
+// unchanged. This function clones the shared computations and returns the set
+// of non-normalizing computations that must be skipped by the visitor.
+absl::flat_hash_set<HloComputation*>
+CloneComputationsForNonNormalizingInstructions(
+    HloModule* module, std::vector<HloComputation*>& computations_to_visit) {
```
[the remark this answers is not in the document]
jaro-sevcik: This code is now gone because of the CallGraph change.

xla/service/float_normalization.cc
jaro-sevcik: Done.

xla/service/float_normalization.cc
jaro-sevcik: Replaced with CallGraph now.

xla/service/float_normalization.cc
jaro-sevcik: Changed the patch to use explicit types.

xla/service/float_normalization.cc
jaro-sevcik: If don't mind introducing the extra dependency and overhead, I am all for that.

That said, we still need a little 2 pass system (1 - to figure out if the computation is shared, 2 - to replace the cloned computation in instruction), but it does look a bit better.

jaro-sevcik (commented):
Thank you for the review, I have addressed the comments. Could you take another look?

cheshire (approved):

PR #8919, opened 2024-01-29
Turn the HLO verifier always on around float normalization on GPU

jaro-sevcik:
Currently, the GPU pipeline only turns the verifier around float normalization on only in debug. If float normalization generates ill-typed IR, the compiler can produce invalid code without anything failing. This patch turns the verification always on.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -1416,7 +1416,7 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
                          LayoutAssignment::InstructionCanChangeLayout)
                      .VerifyBroadcastDimensionsOrder()
                      .VerifyReshapeIsBitcast(),
-                 /*debug_only=*/false);
+                 /*debug_only=*/true);
 
   // Linearize collective schedule if online autotuning of convolutions is
   // enabled.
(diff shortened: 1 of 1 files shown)


xla/service/gpu/gpu_compiler.cc
```diff
@@ -1416,7 +1416,7 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
                          LayoutAssignment::InstructionCanChangeLayout)
                      .VerifyBroadcastDimensionsOrder()
                      .VerifyReshapeIsBitcast(),
-                 /*debug_only=*/true);
+                 /*debug_only=*/false);
```
nouiz, line 1419: I think this add the HloVerifier after each optimization pass.
Did you look at the history to see if this was explicitly added to speed up the compilation?
If so, may we should add only one pass late that is for non-debug?

akuegel: Indeed, the reason why it was turned off was due to compile time. The idea was that we would still run the verifier for tests, anyway. Unfortunately our tests don't cover everything. So I agree, adding the verifier just for just one pass is probably better.

jaro-sevcik: I have added the verifier to the end of this pipeline. Does that look reasonable?

(I have also verified locally that this would catch the PAX crasher that we had.)

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1469,6 +1469,17 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
   pipeline.AddPass<HloCSE>(/*is_layout_sensitive=*/true,
                            /*only_fusion_computations=*/false,
                            /*ignore_control_dependencies=*/true);
+
```
akuegel, line 1472: Maybe add this only in not-debug builds (wrapped in ```#if NDEBUG```)? For debug builds we would run the verifier before and after each pass, anyway.

jaro-sevcik: Good point! Done.

akuegel (approved):

PR #9042, opened 2024-02-01
[Draft][XLA:GPU] Add sycl runtime

Solaryee:
This PR aims to add sycl runtime support, we can run basic JAX GPU UTs with it. It includes:
1. sycl runtime crosstool build
2. sycl stream executor
3. spirv-llvm-translator
4. registration of spir_compiler/computation placer/transfer manager
5. refine build file to remove cuda specific changes(especially those cuda kernels)
6. fix CudaCubin to load spirv binary
7. fix metadata

This is a draft PR to show what is needed to run JAX on sycl runtime.

PR #9045, opened 2024-02-01
Add size for CudaCubinInMemory

Solaryee:
This is a sub PR to add SPIR target in openxla.
`CudaCubinInMemory` is an object represents compiled binary for different backends. `char* bytes` is enough for cuda backend, but spir requires additional `int size` to load the binary.


Zantares:
Sub PR of https://github.com/openxla/xla/pull/9042

kamaljeeti:
Hi @penpornk, any update on this PR? Thanks

dimitar-asenov (approved):

PR #9062, opened 2024-02-01
HLO evaluator: pass custom call handlers to embedded evaluators

sergachev:
Custom calls can also be used within fusions, but the evaluator needs this fix to support them, primarily to be used in tests. Example HLO can be found in the added test.


jvstokes (approved):
Thank you Ilia!!

PR #9137, opened 2024-02-02
[NVIDIA XLA GPU] Fix gcc build failure

Tixxx:
`DelinearizeInBoundsIndex` right now returns `llvm::SmallVector<AffineExpr, 4>`. The number of elements in the returned vector really depends on how many elements are in sizes and strides so it doesn't make much sense to put a fixed number, change it to return just an array with no constant size specified.
It looks like this only fails with gcc not clang.

### DIFF
--- a/xla/service/gpu/model/indexing_analysis.cc
+++ b/xla/service/gpu/model/indexing_analysis.cc
@@ -684,7 +684,7 @@ std::vector<int64_t> ToTransposeDimensions(const Layout& l) {
   return out;
 }
 
-llvm::SmallVector<AffineExpr, 4> DelinearizeInBoundsIndex(
+llvm::SmallVector<AffineExpr> DelinearizeInBoundsIndex(
     mlir::AffineExpr linear, absl::Span<const int64_t> sizes,
     absl::Span<const int64_t> strides) {
   llvm::SmallVector<AffineExpr> result;


xla/service/gpu/model/indexing_analysis.cc
```diff
@@ -684,7 +684,7 @@ std::vector<int64_t> ToTransposeDimensions(const Layout& l) {
   return out;
 }
 
-llvm::SmallVector<AffineExpr, 4> DelinearizeInBoundsIndex(
+llvm::SmallVector<AffineExpr, 6> DelinearizeInBoundsIndex(
```
andportnoy, line 687: The number of inlined elements is [decided based on the size of `T`](https://github.com/llvm/llvm-project/blob/0d6ed8399f35cc8b23a8e6d094379e3d0dd71c45/llvm/include/llvm/ADT/SmallVector.h#L1207) (= `AffineExpr` here), so seems to me a more robust fix would be to let the compiler calculate that number similar to the declaration of `result` in the function body:
```suggestion
llvm::SmallVector<AffineExpr> DelinearizeInBoundsIndex(
```
Otherwise if the size of `AffineExpr` changes in the future, the bug will resurface.

xla/service/gpu/model/indexing_analysis.cc
```diff
@@ -684,7 +684,7 @@ std::vector<int64_t> ToTransposeDimensions(const Layout& l) {
...
     mlir::AffineExpr linear, absl::Span<const int64_t> sizes,
     absl::Span<const int64_t> strides) {
   llvm::SmallVector<AffineExpr> result;
```
andportnoy, line 690: Alternatively, if 4 is actually the right number of element to inline (e.g. it's the max number of dimensions), update the declaration of `result` to match it:
```suggestion
  llvm::SmallVector<AffineExpr, 4> result;
```

Am I missing something? Does your interpretation match mine?

andportnoy (changes requested):

andportnoy:
Note XLA now has a `configure.py` that defaults to using Clang as the host compiler. A new `--use_clang` flag is being added to `build/build.py` in JAX: https://github.com/google/jax/pull/19633. NVIDIA JAX Toolbox will then also enable it: https://github.com/NVIDIA/JAX-Toolbox/pull/513. Soon these GCC incompatibility issues will not be breaking anymore.

xla/service/gpu/model/indexing_analysis.cc
Tixxx: sync'd offline, the actual number of this is 6 right now but it could grow in the future, so using a generic array type without a fixed range is more reasonable here.

Tixxx:
> Note XLA now has a `configure.py` that defaults to using Clang as the host compiler. A new `--use_clang` flag is being added to `build/build.py` in JAX: [google/jax#19633](https://github.com/google/jax/pull/19633). NVIDIA JAX Toolbox will then also enable it: [NVIDIA/JAX-Toolbox#513](https://github.com/NVIDIA/JAX-Toolbox/pull/513). Soon these GCC incompatibility issues will not be breaking anymore.

that'd be awesome to have!

zhenying-liu:
It is recommended to pass the SmallVector without the exact number of elements, according to the llvm doc: https://llvm.org/doxygen/classllvm_1_1SmallVector.html#details. Although it can pass clang compiler, this commit is still good to go.

reedwm (approved):
I don't understand why this was breaking before with gcc with an explicit value to N, but I'll approve anyway since we should fix the build quickly.

@Tixxx can you post what error you were getting with gcc?

andportnoy:
@reedwm https://github.com/NVIDIA/JAX-Toolbox/actions/runs/7754093006/job/21146722348#step:10:981
```
#15 120.7 external/xla/xla/service/gpu/model/indexing_analysis.cc:714:10: error: could not convert ‘result’ from ‘SmallVector<[...],6>’ to ‘SmallVector<[...],4>’
#15 120.7   714 |   return result;
#15 120.7       |          ^~~~~~
#15 120.7       |          |
#15 120.7       |          SmallVector<[...],6>
```

reedwm:
Interesting, I wonder why this works for clang. Maybe clang somehow has a different default default for N in SmallVector (4 instead of 6), or maybe clang somehow allows SmallVectors of different Ns to be assigned to each other.

Whatever the reason, good to fix this regardless.

andportnoy:
Minimal repro: https://godbolt.org/z/oK65d1zYv
|| C++17  | C++20 |
|-|-|-|
|GCC 12.3|   FAIL |  PASS|
|GCC 13.1 |  PASS | PASS|

Tixxx:
> Interesting, I wonder why this works for clang. Maybe clang somehow has a different default default for N in SmallVector (4 instead of 6), or maybe clang somehow allows SmallVectors of different Ns to be assigned to each other.
> 
> Whatever the reason, good to fix this regardless.

Thanks @andportnoy  for getting the error.
gcc is being too pedantic about the type here, ignoring the fact that the type itself is dynamic.
Anyway, thanks for the approval.

Tixxx:
@reedwm  A gentle ping to see if we could merge this soon.

ddunl:
is this still needed after https://github.com/openxla/xla/pull/9148?

reedwm:
No, looks like that change also fixes it by making the types match. So closing this PR. Sorry for not merging this earlier!

PR #9171, opened 2024-02-05
Add xla flag for fp8 gemm-gelu fusion

wenscarl:
For most cases, FP8 cublasLt gemm-gelu fusion is faster than non-fused cases but not all. Adding a switching flag.


reedwm (changes requested):
@wenscarl can you describe in what cases or models fusing gelu is slower?

@cheshire are you ok adding this flag?

wenscarl:
E.g. F8 gemm -> gelu ->q-> F8 gemm.
image
image
The decision to fuse gelu into gemm creates more cuda graph segments thus less optimal performance.

wenscarl:
Close. A third party error triggered perf regression, not XLA relevant.

PR #9205, opened 2024-02-06
[ROCm] Fix compilation errors in rocm_dnn

ekuznetsov139:
This fixes rocm_dnn after it got broken by the commit https://github.com/openxla/xla/commit/616486056afc4c97cc419d5cb8a033121c84e3d7.

It also corrects a small issue in ir_emitter_unnested.h (that header file needs to include rocm_config.h, otherwise some methods of IrEmitterUnnested will be missing their declarations).

### DIFF
--- a/xla/service/gpu/ir_emitter_unnested.h
+++ b/xla/service/gpu/ir_emitter_unnested.h
@@ -48,6 +48,11 @@ limitations under the License.
 #include "xla/status.h"
 #include "xla/statusor.h"
 
+#if TENSORFLOW_USE_ROCM
+// for TF_HIPBLASLT
+#include "rocm/rocm_config.h"
+#endif
+
 namespace xla {
 namespace gpu {
 
--- a/xla/stream_executor/rocm/rocm_dnn.h
+++ b/xla/stream_executor/rocm/rocm_dnn.h
@@ -22,6 +22,7 @@ limitations under the License.
 #include "absl/synchronization/mutex.h"
 #include "absl/types/span.h"
 #include "rocm/include/miopen/miopen.h"
+#include "xla/stream_executor/device_memory_allocator.h"
 #include "xla/stream_executor/dnn.h"
 #include "xla/stream_executor/plugin_registry.h"
 
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -427,6 +427,8 @@ cc_library(
         "//xla/service/gpu/runtime3:cub_sort_thunk",
         "//xla/service/gpu/runtime3:gpublas_lt_matmul_thunk",
         "//xla/service/gpu/runtime3:triangular_solve_thunk",
+    ]) + if_rocm_is_configured([
+        "@local_config_rocm//rocm:rocm_headers",
     ]),
 )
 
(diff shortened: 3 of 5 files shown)


xla/service/gpu/ir_emitter_unnested.h
```diff
@@ -48,6 +48,11 @@ limitations under the License.
 #include "xla/status.h"
 #include "xla/statusor.h"
 
+#if TENSORFLOW_USE_ROCM
+// for TF_HIPBLASLT
+#include "rocm/rocm_config.h"
```
akuegel, line 53: Would this need a corresponding BUILD file change?

xla/stream_executor/rocm/rocm_dnn.h
```diff
@@ -22,6 +22,7 @@ limitations under the License.
 #include "absl/synchronization/mutex.h"
 #include "absl/types/span.h"
 #include "rocm/include/miopen/miopen.h"
+#include "xla/stream_executor/device_memory_allocator.h"
```
akuegel, line 25: I think this needs a corresponding BUILD file update

ekuznetsov139: It builds as-is.

akuegel: Not in a strict mode. With layering check enabled, for every header include you need to have a corresponding dependency to a target that has it listed in its hdrs section.

akuegel: So the target you would need to depend on here is //xla/stream_executor:device_memory_allocator

xla/service/gpu/ir_emitter_unnested.h
ekuznetsov139: it's an existing dependency via service/gpu/ir_emitter_unnested -> service/gpu/gpu_executable -> @local_config_rocm//rocm:rocm_headers

akuegel: Yes, but it is better not to rely on transitive dependencies. layering_check will complain about that.

ekuznetsov139: How do I reproduce that complaint? I'm running
```
bazel test --config=rocm  \
    --build_tag_filters=-no_oss,-oss_excluded,-oss_serial,gpu,-no_gpu,-no_rocm --test_tag_filters=-no_oss,-oss_excluded,-oss_serial,gpu,-no_gpu,-no_rocm --test_output=errors --features=layering_check \
    --flaky_test_attempts=3 --jobs=150 --run_under=//tools/ci_build/gpu_build:parallel_gpu_execute  \
    --keep_going --nocache_test_results   --test_env=TF_GPU_COUNT=8 --local_test_jobs=32 \
    --    //xla/... //build_tools/... -//xla/stream_executor/cuda/...
```

and I'm not getting any warnings about dependencies.

akuegel: It is possible that in open source, layering_check does not work so well. Since we don't build ROCM internally, it probably doesn't matter, but from an engineering standpoint it makes sense to have direct dependencies for header includes. Otherwise the path to the dependency can be broken if the dependency is removed from the other target. Thanks for adding the dependencies :)

akuegel (approved):

PR #9246, opened 2024-02-06
Perform elementwise BF16 math natively in platforms that support it.

dimvar:
Perform elementwise BF16 math natively in platforms that support it, e.g., Hopper GPUs.

In xla/service/gpu, we check the platform using Cuda compute capability. We don't have access to that in llvm_ir, so we pass an optional bool to say whether native BF16 support is there. We make it optional with default to false, to avoid a very large refactoring that would touch most functions in llvm_ir.

### DIFF
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -3225,6 +3225,7 @@ cc_library(
         "//xla:xla_data_proto_cc",
         "//xla/hlo/ir:hlo",
         "//xla/service:float_support",
+        "//xla/stream_executor:device_description",
         "@com_google_absl//absl/log:check",
     ],
 )
--- a/xla/service/gpu/gemm_fusion_autotuner.cc
+++ b/xla/service/gpu/gemm_fusion_autotuner.cc
@@ -475,7 +475,8 @@ absl::StatusOr<std::unique_ptr<HloModule>> TritonGemmAutotuneExtractor(
 
   if (config.split_k > 1) {
     TF_RETURN_IF_ERROR(MakeDotSplitKBatch(cloned_dot_fusion, config));
-    GpuFloatSupport bf16_support(BF16);
+    GpuFloatSupport bf16_support(gpu_device_info.gpu_compute_capability(),
+                                 BF16);
     FloatNormalization float_normalization(&bf16_support);
     TF_RETURN_IF_ERROR(float_normalization.Run(new_module.get()).status());
     GpuInstructionFusion instruction_fusion(/*may_duplicate=*/false,
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -1323,9 +1323,9 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
                       GetAutotuneConfig(stream_exec, debug_options, options,
                                         gpu_target_config));
   // Lambdas and related constants:
-  const GpuFloatSupport bf16_support(BF16);
-  const GpuFloatSupport f8e5m2_support(F8E5M2, F16);
-  const GpuFloatSupport f8e4m3fn_support(F8E4M3FN, F16);
+  const GpuFloatSupport bf16_support(gpu_version, BF16);
+  const GpuFloatSupport f8e5m2_support(gpu_version, F8E5M2, F16);
+  const GpuFloatSupport f8e4m3fn_support(gpu_version, F8E4M3FN, F16);
   const FloatSupport f8e4m3b11fnuz_support(F8E4M3B11FNUZ, F16);
   const FloatSupport f8e5m2fnuz_support(F8E5M2FNUZ, F16);
   const FloatSupport f8e4m3fnuz_support(F8E4M3FNUZ, F16);
--- a/xla/service/gpu/gpu_float_support.cc
+++ b/xla/service/gpu/gpu_float_support.cc
@@ -73,6 +73,18 @@ bool GpuFloatSupport::IsSupported(const HloInstruction& hlo) const {
     // Other special ops.
     case HloOpcode::kBitcast:
       return true;
+    // Elementwise ops.
+    case HloOpcode::kAdd:
+    case HloOpcode::kSubtract:
+    case HloOpcode::kMultiply: {
+      if (LowPrecisionType() == BF16) {
+        auto* cuda_compute_capability =
+            std::get_if<se::CudaComputeCapability>(&compute_capability_);
+        return cuda_compute_capability != nullptr &&
+               cuda_compute_capability->IsAtLeastHopper();
+      }
+      return false;
+    }
     default:
       return false;
   }
--- a/xla/service/gpu/gpu_float_support.h
+++ b/xla/service/gpu/gpu_float_support.h
@@ -20,6 +20,7 @@ limitations under the License.
 
 #include "xla/hlo/ir/hlo_instruction.h"
 #include "xla/service/float_support.h"
+#include "xla/stream_executor/device_description.h"
 #include "xla/xla_data.pb.h"
 
 namespace xla {
@@ -27,9 +28,11 @@ namespace gpu {
 
 class GpuFloatSupport : public FloatSupport {
  public:
-  explicit GpuFloatSupport(PrimitiveType low_precision_type,
+  explicit GpuFloatSupport(se::GpuComputeCapability cc,
+                           PrimitiveType low_precision_type,
                            PrimitiveType high_precision_type = F32)
-      : FloatSupport(low_precision_type, high_precision_type) {}
+      : FloatSupport(low_precision_type, high_precision_type),
+        compute_capability_(cc) {}
 
   bool SupportsLowPrecisionOperand(const HloInstruction& hlo,
                                    int64_t operand_index) const override {
@@ -45,6 +48,8 @@ class GpuFloatSupport : public FloatSupport {
 
  private:
   bool IsSupported(const HloInstruction& hlo) const;
+
+  const se::GpuComputeCapability compute_capability_;
 };
 
 }  // namespace gpu
--- a/xla/tests/cpu_gpu_fusion_test.cc
+++ b/xla/tests/cpu_gpu_fusion_test.cc
@@ -887,6 +887,24 @@ ENTRY entry {
   EXPECT_TRUE(RunAndCompare(hlo, ErrorSpec{1e-5, 1e-5}));
 }
 
+XLA_TEST_F(CpuGpuFusionTest, DontUpcastBf16Small) {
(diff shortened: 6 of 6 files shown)


xla/service/gpu/tests/dynamic_update_slice_inplace_bf16.hlo
```diff
@@ -0,0 +1,221 @@
+// RUN: hlo-opt %s --platform=gpu --stage=llvm-before-optimizations --xla_gpu_target_config_filename=%S/../../../tools/hlo_opt/gpu_specs/%{GPU}.txtpb --split-input-file | FileCheck --check-prefixes=CHECK %s
```
cheshire, line 1: Maybe instead of opaque "GPU" specifically select the platform you're after (probably need both H100 and A100?)

xla/service/gpu/tests/dynamic_update_slice_inplace_bf16.hlo
```diff
@@ -0,0 +1,221 @@
...
+  p4 = s32[] parameter(4)
+  ROOT fusion_root_multiple = (bf16[10,11,12], bf16[8,11,12]) fusion(p0, p1, p2, p3, p4), kind=kLoop, calls=fused_computation
+}
+
+// -----
+
+// CHECK-LABEL: @fusion_root_multiple_transpose_bitcast
+// CHECK-SAME: %[[ARG0:[A-Za-z0-9]*]],
+// CHECK-SAME: %[[ARG1:[A-Za-z0-9]*]],
+// CHECK-SAME: %[[ARG2:[A-Za-z0-9]*]],
```
cheshire, line 58: Do we need to test everything? Maybe just matching a few operations will be enough?

xla/service/llvm_ir/ir_array.h
```diff
@@ -290,7 +291,8 @@ class IrArray {
   // available) or the multi-dimensional index should be used.
   llvm::Value* EmitReadArrayElement(const Index& index, llvm::IRBuilder<>* b,
                                     absl::string_view name = "",
-                                    bool use_linear_index = true) const;
+                                    bool use_linear_index = true,
+                                    bool native_bf16_support = false) const;
```
cheshire, line 295: Adding extra bool to all methods gets super clunky and fragile, maybe make it a class attribute passed to the constructor, WDYT?

cheshire (changes requested):
Left some nits.

Do we have micro-benchmarks/benchmarks with results?

xla/service/llvm_ir/ir_array.h
dimvar: I totally agree with this, and in fact, I had to do something similar on XLA:TPU in the past, to pass target information around. But it was a giant refactoring that took months.

Maybe it won't be as large for XLA:GPU, but it will still be pretty large, because llvm_ir/llvm_util.h contains free functions, not methods. So it will need to become a class and all uses will be changed.

I think for this contribution we should keep the blast radius limited. Then, I can ask more broadly if people want the larger refactoring, and do it if they agree it's a good idea. (Or even prototype some subset of it, to have something more concrete to show.)

WDYT?

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -3117,8 +3127,10 @@ absl::Status IrEmitterUnnested::EmitSelectAndScatter(
     llvm::Value* operand_address =
         operand_array.EmitArrayElementAddress(operand_index, &b_);
     llvm::AllocaInst* select_return_buffer = llvm_ir::EmitAllocaAtFunctionEntry(
-        llvm_ir::PrimitiveTypeToIrType(PRED, module_), "select_return_buffer",
-        &b_);
+        llvm_ir::PrimitiveTypeToIrType(
+            PRED, module_,
+            ir_emitter_context_->cuda_compute_capability().IsAtLeastHopper()),
```
cheshire, line 3132: Nit: /*has_bf16_support=*/

xla/service/gpu/tests/BUILD
```diff
@@ -652,8 +652,8 @@ xla_cc_test(
 )
 
 lit_test_suite(
-    name = "all_tests",
-    srcs = enforce_glob(
+    name = "lit_tests_ampere_or_rocm",
+    srcs = glob(
```
cheshire, line 656: Could we keep using enforce_glob? Why the rename?

xla/service/gpu/tests/BUILD
```diff
@@ -718,6 +715,24 @@ lit_test_suite(
     ],
 )
 
+lit_test_suite(
```
cheshire, line 718: Could use prev target and just pass GPU explicitly? Don't need to use the GPU renaming?

xla/service/llvm_ir/ir_array.cc
```diff
@@ -488,7 +488,8 @@ llvm::Value* IrArray::Index::Linearize(
 
 llvm::Value* IrArray::EmitArrayElementAddress(
     const IrArray::Index& index, llvm::IRBuilder<>* b, absl::string_view name,
-    bool use_linear_index, llvm::Value** is_high_order_bits) const {
+    bool use_linear_index, llvm::Value** is_high_order_bits,
+    bool native_bf16_support) const {
```
cheshire, line 492: Nit: `has_bf16_support` seems more self-descriptive?

xla/service/llvm_ir/ir_array.h
cheshire: Could be correct, but maybe also check how large a refactoring CL would be?

cheshire (changes requested):
Left some nits. Also would be nice to provide micro-benchmarks for expected speed-ups.

xla/service/gpu/tests/dynamic_update_slice_inplace_bf16.hlo
dimvar: fixed

xla/service/gpu/ir_emitter_unnested.cc
dimvar: fixed

xla/service/gpu/tests/BUILD
dimvar: This comment, and enforce_glob, and specify the GPU in the file are all connected, so I'll just reply here.

In the BUILD file, when "if_cuda_is_configured" is true, we can only specify one value for the GPU param. So we can't have both ampere and hopper. So, if we keep a single BUILD rule to test both, then we can keep specifying ampere in the BUILD file, and in some files in the glob change xla_gpu_target_config_filename to be hopper (and those tests won't run on rocm either). I think this can lead to confusion when someone looking at the BUILD rule thinks all tests are ampere, but in reality some are hopper. 

If we split into two rules, one for ampere and one for hopper, then we can't use enforce_glob, because it enforces that all ".hlo" files are included in the BUILD rule, which will no longer be the case.

xla/service/llvm_ir/ir_array.h
dimvar: Here is an incomplete proof of concept.
https://github.com/openxla/xla/commit/ca24dbd58d4b974601744b2be8b8dca0fba0ffc2

(Incomplete in the sense that it compiles, but the full refactoring needs further changes.)

I think it'd be useful to move forward with it, but since the full change is going to be large, it needs to be its own PR.

ezhulenev:
+1 to look into passing struct with options or keeping it as a member, but that can be done later

ezhulenev (approved):

tdanyluk (approved):

dimvar:
Thanks for approving. I rebased and the PR is ready to be merged.

dimvar:
Fixed some formatting issues.

cheshire (approved):

d0k:
We're actually ready to enable the bfloat16 type by default on all GPUs, LLVM is now set up to do it. I tried that last week in 97ed5c63dcf373c27d91ff6c784e6ad0ca364399 but it turns out that Aarch64 is still not ready and it's tricky to fix because of how that LLVM backend is set up, so XLA:CPU is out.

Having a different type just for H100+ is going to fall apart fairly quickly because we have big holes in CI coverage still. Did you run it through JAX' test suite?

cheshire (approved):

cheshire:
> Did you run it through JAX' test suite?

Good question! I'll test once CL is imported.

dimitar-asenov:
A number of JAX tests fail with this change on H100:

jax/tests:ann_test_gpu
jax/tests:api_test_gpu
jax/tests:export_harnesses_multi_platform_test_gpu
jax/tests:lax_autodiff_test_gpu
jax/tests:lax_numpy_reducers_test_gpu
jax/tests:lax_test_gpu
jax/tests:lax_vmap_test_gpu
jax/tests:nn_test_gpu
jax/tests:random_lax_test_gpu
jax/tests:scipy_signal_test_gpu
jax/tests:shape_poly_test_gpu
jax/tests:sparse_bcoo_bcsr_test_gpu
jax/tests:sparsify_test_gpu
jax/tests:state_test_gpu

In addition, we see this run-time check failure in one of our internal models:

```
INTERNAL: RET_CHECK failure (xla/service/gpu/gpu_compiler.cc:1636) !llvm::verifyModule(*llvm_module, &err_stream) Invalid LLVM IR before optimizations:
Both operands to a binary operator are not of the same type!
  %multiply.0 = fmul bfloat %15, i16 %constant_0
```

Please run the JAX tests and make sure this PR passes on them.

dimvar:
Turns out Ben's recent commit (https://github.com/openxla/xla/commit/a883b54870bc7ce24e510f941d458a354de67d66) has significant overlap with this PR, so the PR needs to be reworked.

The conversions to F32 on hopper are still there at head IIUC, so I'll remove these in my updated PR to get the speedups. If someone at Google is also looking into this, please let me know ASAP, so we don't duplicate work.

gpupuck:
@dimitar-asenov Hi! What are the test sets for your JAX CI, all the GPU tests?

dimitar-asenov:
I believe the JAX CI does run all the tests at https://github.com/google/jax/tree/main/tests on GPU, but I'm not sure it runs them on H100 yet. You might want to run those locally on a machine with H100.

ddunl (approved):

reedwm (approved):

jbcoe:
I submitted a modernized version of this PR. 

We had to relax tolerance for np jnp numerical comparison for bfloat16 in 
some JAX tests https://github.com/google/jax/commit/f60be7026449af166297a9e292fce4609b83248c

cheshire:
So can this PR be now closed as done? Or more features remain?

d0k:
This is done, the last remaining bits landed with 1612a95bb14900b98a3414cb9c2c11aa6c6b438e

PR #9254, opened 2024-02-06
[NVIDIA XLA] Use inner instruction for PGLE lookup

Tixxx:
If an instruction is wrapped using AsyncInstruction(i.e reduce-scatter-start), we need to use the inner instruction's name to look up profiled costs in PGLE since the profiled kernel name is per wrapped instruction, not the outer async instruction.


jurahul (approved):
LGTM, thanks!

Tixxx:
@jurahul can we try importing this PR? Thanks for the review.

ddunl (approved):

PR #9262, opened 2024-02-07
[XLA:GPU] Support more conv1d backward input/filter pattern match

ghost:
Conv1d always reshape to conv2d and make backward pattern can not match using current pattern matcher. There exist 2 situations:

(1) conv1d backward input pattern is reverse(reshape(filter)), which can reuse conv2d backward input patter match rule. But if run algsimp pass before conv rewrite pass, this pattern will change to reshape(reverse(filter)) so that current pattern matcher will fail. (2) conv backward filter pattern matcher require each kernel spatial dimension is lager than output spatil dimension. But it is not valid if conv1d reshape to conv2d. The reason is that there exist 1 size in kernel dimensions and output dimensions.

This PR aims to support this special case.

### DIFF
--- a/xla/service/gpu/gpu_conv_rewriter.cc
+++ b/xla/service/gpu/gpu_conv_rewriter.cc
@@ -43,6 +43,30 @@ namespace {
 using ConvolutionMatch = std::optional<
     std::tuple<Window, ConvolutionDimensionNumbers, HloInstruction*>>;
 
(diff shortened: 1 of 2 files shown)


kamaljeeti:
Hi @ezhulenev , can you look into this once? Thanks.

ezhulenev (approved):

PR #9398, opened 2024-02-10
[NVIDIA XLA] Copy control dependencies from original instruction to async start

Tixxx:
The control dependency indicates that if A depends on B, then A shouldn't start until B is done.
In CreateAsyncInstructions, the control deps are copied to the AsyncDone instruction which breaks the above pre-requisite. 
Control deps should be copied to AsyncStart.

### DIFF
--- a/xla/hlo/ir/hlo_computation.cc
+++ b/xla/hlo/ir/hlo_computation.cc
@@ -1068,7 +1068,15 @@ StatusOr<HloInstruction*> HloComputation::CreateAsyncInstructions(
   async_start->CopyBackendConfigFrom(instruction);
   async_done->set_metadata(instruction->metadata());
   async_done->CopyBackendConfigFrom(instruction);
-  TF_RETURN_IF_ERROR(async_start->CopyAllControlDepsFrom(instruction));
+  for (HloInstruction* control_pred : instruction->control_predecessors()) {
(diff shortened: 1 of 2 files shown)


xla/hlo/ir/hlo_computation.cc
```diff
@@ -1068,7 +1068,7 @@ StatusOr<HloInstruction*> HloComputation::CreateAsyncInstructions(
   async_start->CopyBackendConfigFrom(instruction);
   async_done->set_metadata(instruction->metadata());
   async_done->CopyBackendConfigFrom(instruction);
-  TF_RETURN_IF_ERROR(async_done->CopyAllControlDepsFrom(instruction));
+  TF_RETURN_IF_ERROR(async_start->CopyAllControlDepsFrom(instruction));
```
berkinilbeyi, line 1071: I think the right thing to do here would be to put the control predecessors to async_start and control successors to async_done

berkinilbeyi (changes requested):

xla/hlo/ir/hlo_computation.cc
Tixxx: ah good point, forgot that we had control successors.
updated

ezhulenev (approved):

berkinilbeyi (approved):

Tixxx:
@ddunl  could you help import this? thanks

PR #9400, opened 2024-02-10
Move Gt(Max) optimization after all other HandleCompare optimizations

apivovarov:
Changes:
1. Move `Gt(Max)` optimization after all other HandleCompare optimizations because other Gt optimizations are more valuable (such as `0 > unsigned  -> False`)
2. If `Gt(Max)` optimization was applied  - `return OkStatus()` instantly. (If `ReplaceInstruction` was called then `lhs/rhs` pointers initialized at the beginning of `HandleCompare` function might point to replaced operand. For consistency, we need to `return OkStatus()` right after `MarkAsChanged`)

@akuegel Could you review it?


akuegel (approved):
Thank you :)

PR #9433, opened 2024-02-12
[XLA:CPU][oneDNN] Align with oneDNN for supported data types in matmul

mdfaijul:
This PR updates the data type support based on CPU features and aligns with oneDNN.

### DIFF
--- a/xla/service/cpu/onednn_util.h
+++ b/xla/service/cpu/onednn_util.h
@@ -25,27 +25,23 @@ namespace cpu {
 
 inline bool IsSupportedType(xla::PrimitiveType dtype) {
   using tsl::port::CPUFeature;
-  bool is_supported = false;
   // TODO(intel-tf): Enable more types.
   switch (dtype) {
     case F32:
-      is_supported = true;
-      break;
+      return true;
     case BF16:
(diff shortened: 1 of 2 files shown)


xla/service/cpu/onednn_util.h
```diff
@@ -17,25 +17,35 @@ limitations under the License.
...
+#include "xla/xla_data.pb.h"
 
 namespace xla {
 namespace cpu {
 
 inline bool IsSupportedType(xla::PrimitiveType dtype) {
   using tsl::port::CPUFeature;
-  static bool is_bf16_supported = TestCPUFeature(CPUFeature::AVX512_BF16) ||
-                                  TestCPUFeature(CPUFeature::AMX_BF16);
+  bool is_supported = false;
```
penpornk, line 28: Nit: We can just get rid of this variable and just return the value/expression directly, e.g., `return true`, `return TestCPUFeature...`.

penpornk (changes requested):
Thank you for the PR!

xla/service/cpu/onednn_util.h
mdfaijul: Thanks! Corrected.

penpornk (approved):
Thank you for the changes!

PR #9509, opened 2024-02-14
[NVIDIA XLA] Support initializing communicators with custom configurations

Tixxx:
NCCL added support for giving custom configurations to communicators upon creation. [Here's](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcomminitrankconfig) the documentation for that.

We should utilize this feature to give flexibility for tuning in xla. This PR only exposes ways to configure maximum number of channels used by NCCL for different communicators, while it sets up the infra for exposing more configs in the future.

### DIFF
--- a/xla/service/gpu/nccl_api.cc
+++ b/xla/service/gpu/nccl_api.cc
@@ -299,7 +299,7 @@ class DefaultNcclApi final : public NcclApi {
 
   absl::StatusOr<std::vector<OwnedNcclComm>> CommSplit(
       absl::Span<const NcclCommHandle> comms, int32_t color,
-      absl::Span<const int32_t> keys) final;
+      absl::Span<const int32_t> keys, const Config& config) final;
 
   absl::Status CommAbort(NcclCommHandle comm) final;
   absl::Status CommFinalize(NcclCommHandle comm) final;
@@ -373,6 +373,12 @@ DefaultNcclApi::CommInitRanks(int32_t nranks, const NcclCliqueId& clique_id,
 
   ncclConfig_t comm_config = NCCL_CONFIG_INITIALIZER;
   comm_config.splitShare = config.split_share;
+  if (config.max_nchannels > 0) {
+    comm_config.maxCTAs = config.max_nchannels;
+    VLOG(1) << "Maximum number of channels for hash(id)="
+            << absl::HashOf(clique_id)
+            << " is set to: " << comm_config.maxCTAs;
+  }
 
   std::vector<OwnedNcclComm> comms;
   comms.reserve(ranks.size());
@@ -398,7 +404,7 @@ DefaultNcclApi::CommInitRanks(int32_t nranks, const NcclCliqueId& clique_id,
 
 absl::StatusOr<std::vector<NcclApi::OwnedNcclComm>> DefaultNcclApi::CommSplit(
     absl::Span<const NcclCommHandle> comms, int32_t color,
-    absl::Span<const int32_t> keys) {
+    absl::Span<const int32_t> keys, const Config& config) {
   VLOG(1) << absl::StreamFormat(
       "Split %d NCCL communicators using color %d and keys: [%s]", comms.size(),
       color, absl::StrJoin(keys, ","));
@@ -408,6 +414,16 @@ absl::StatusOr<std::vector<NcclApi::OwnedNcclComm>> DefaultNcclApi::CommSplit(
     return absl::InvalidArgumentError(
         absl::StrFormat("Comms and keys must have the same size, but %d != %d",
                         comms.size(), keys.size()));
+  }
+
+  ncclConfig_t comm_config = NCCL_CONFIG_INITIALIZER;
+  comm_config.splitShare = config.split_share;
+  // If max_nchannels is set, then we don't want to
+  // inherit from parent comm.
+  if (config.max_nchannels > 0) {
+    comm_config.maxCTAs = config.max_nchannels;
+    VLOG(1) << "CommSplit maximum number of channels "
+            << " is set to: " << comm_config.maxCTAs;
   }
 
   // In contrast to grouped initialization communicator splitting initializes
@@ -422,7 +438,7 @@ absl::StatusOr<std::vector<NcclApi::OwnedNcclComm>> DefaultNcclApi::CommSplit(
             << " and key " << keys[i];
     XLA_NCCL_RETURN_IF_ERROR(ncclCommSplit(Cast(comms[i]), color, keys[i],
                                            &split_comms_handles[i],
-                                           /*config=*/nullptr));
+                                           /*config=*/&comm_config));
   }
   TF_RETURN_IF_ERROR(GroupEnd());
 
@@ -607,5 +623,4 @@ DefaultNcclApi::DeregisterBuffer(NcclCommHandle comm,
       ncclCommDeregister(Cast(comm), reinterpret_cast<void*>(handle)));
 #endif  // NCCL_VERSION_CODE >= 21901
 }
-
 }  // namespace xla::gpu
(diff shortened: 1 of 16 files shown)


kamaljeeti:
Hi @cheshire, @ezhulenev , Please ignore this draft PR I will update once it's in open state.

xla/service/gpu/nccl_api.h
```diff
@@ -33,6 +33,21 @@ limitations under the License.
 
 namespace xla::gpu {
 
+//===----------------------------------------------------------------------===//
+// NcclCommConfig
+//===----------------------------------------------------------------------===//
+// NcclCommConfig contains configurations to initialize a communicator.
```
ezhulenev, line 39: nit: space before comment

xla/service/gpu/nccl_api.cc
```diff
@@ -321,6 +326,13 @@ class DefaultNcclApi final : public NcclApi {
 
   absl::StatusOr<NcclRegisteredBufferHandle> DeregisterBuffer(
       NcclCommHandle comm, NcclRegisteredBufferHandle handle) final;
+
+  absl::StatusOr<bool> IsInitConfigSupported() final;
```
ezhulenev, line 330: nccl_api_stub should implement new APIs

xla/service/gpu/nccl_api.h
```diff
@@ -226,8 +241,19 @@ class NcclApi {
...
+  // or not. Available after 2.17.0.
+  // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcomminitrankconfig
+  virtual absl::StatusOr<bool> IsInitConfigSupported() = 0;
+
+  // Creates a new communicator with custom configuration.
+  //
+  // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcomminitrankconfig
+  virtual absl::StatusOr<OwnedNcclComm> CommInitRankConfig(
+      int32_t nranks, const NcclCliqueId& clique_id,
+      int32_t rank, NcclCommConfig& config) = 0;
```
ezhulenev, line 255: instead of adding new api just pass config in std::optional with default nullopt

xla/service/gpu/nccl_api.h
```diff
@@ -33,6 +33,21 @@ limitations under the License.
...
 
+//===----------------------------------------------------------------------===//
+// NcclCommConfig
+//===----------------------------------------------------------------------===//
+// NcclCommConfig contains configurations to initialize a communicator.
+// It wraps the ncclConfig_t object. More configuration parameters
+// will be exposed in the future when necessary.
+
+struct NcclCommConfig {
+  static absl::StatusOr<NcclCommConfig> Create(int64_t max_nchannels);
```
ezhulenev, line 44: Looks like a simple struct would work fine here? Can be converted to factory later if needed.

xla/service/gpu/nccl_clique.h
```diff
@@ -147,7 +147,8 @@ struct NcclClique : public Lockable<NcclCliqueCommunicators, NcclCliqueName> {
 absl::StatusOr<std::shared_ptr<NcclClique::Lock>> AcquireNcclClique(
     RunId run_id, NcclCliqueKey clique_key,
     const NcclCliqueIdCallback& clique_id_callback, int32_t rank,
-    size_t num_local_participants, bool may_skip_rendezvous);
+    size_t num_local_participants, bool may_skip_rendezvous,
+    int64_t collective_max_nchannels, int64_t p2p_max_nchannels);
```
ezhulenev, line 151: pass `std::optional<NcclCommConfig>`? you have to construct it anyway, and nccl_api header dependency is already here

xla/service/gpu/nccl_clique_key.cc
```diff
@@ -37,8 +37,10 @@ namespace xla::gpu {
 //===----------------------------------------------------------------------===//
 
 NcclCliqueKey::NcclCliqueKey(std::vector<GlobalDeviceId> devices,
-                             int64_t stream_id)
-    : devices_(std::move(devices)), stream_id_(stream_id) {}
+                             int64_t stream_id, AsyncStreamKind stream_kind)
```
ezhulenev, line 40: this looks like a part of different change

xla/service/gpu/nccl_clique_key.h
```diff
@@ -73,14 +73,17 @@ inline uint64_t GetStreamId(
...
+  explicit NcclCliqueKey(
+      std::vector<GlobalDeviceId> devices, int64_t stream_id = 0,
+      AsyncStreamKind stream_kind = AsyncStreamKind::kCollective);
 
   absl::Span<const GlobalDeviceId> devices() const;
 
   // Returns the rank of the global device in the clique.
   std::optional<int64_t> rank(GlobalDeviceId id) const;
 
+  AsyncStreamKind GetStreamKind() const { return stream_kind_; }
```
ezhulenev, line 85: prefer `stream_kind` (lower case for simple getters/setters)

xla/service/gpu/nccl_api.h
Tixxx: sure, i will remove the factory for now.

xla/service/gpu/nccl_clique_key.cc
Tixxx: we need this when creating the clique key so we know what config to use when using the key to create comm.

xla/service/gpu/nccl_api.h
```diff
@@ -250,8 +270,12 @@ class NcclApi {
   // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcommderegister
   virtual absl::StatusOr<NcclRegisteredBufferHandle> DeregisterBuffer(
       NcclCommHandle comm, NcclRegisteredBufferHandle handle) = 0;
-};
 
+  // Returns if initializing communicator with custom configuration is supported
+  // or not. Available after 2.17.0.
+  // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcomminitrankconfig
+  virtual absl::StatusOr<bool> IsInitConfigSupported() = 0;
```
ezhulenev, line 277: Do we ever compile with NCCL that old? Definitely not at google. Should we just assume config is available?

Tixxx: I'm not sure about other users though but we always compile against the latest. If there's no concern on your side, i think we can make that assumption.

Tixxx:
@ezhulenev  i changed the commSplit API to take in a config too since we probably dont want to inherit some configs from parent.

xla/service/gpu/nccl_api.h
```diff
@@ -162,7 +163,7 @@ class NcclApi {
   // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcommsplit
   virtual absl::StatusOr<std::vector<OwnedNcclComm>> CommSplit(
       absl::Span<const NcclCommHandle> comms, int32_t color,
-      absl::Span<const int32_t> keys) = 0;
+      absl::Span<const int32_t> keys, const Config& config) = 0;
```
ezhulenev, line 166: make it optional? because I think the default is actually to reuse config from parent (because otherwise resources are not shared? are they?)

ezhulenev (approved):

xla/service/gpu/nccl_api.h
Tixxx: Since in xla we always set resource sharing to true, that behavior wont change. But making it optional makes sense, default we want is indeed inherit whatever parent has.

ezhulenev (approved):

PR #9593, opened 2024-02-16
[WIP] [ROCm] Moving blas::CallContext into NumericOptions

ekuznetsov139:
OpenXLA currently allows passing gradient flags for GEMM operations through the stack using an argument of type blas::CallContext.

This PR:
* Packs these flags into PrecisionConfig (at the high level) and NumericOptions (at the stream executor level)
* Makes sure that flags are passed correctly (particularly that they are reversed if GEMM operands are swapped)
* Adds unit tests covering the functionality (ROCm only)
* To that end, expands //xla/tests:dot_operation_test (for all backends)

I am somewhat unhappy with the method I am currently using to get data from stream executor back to the unit test, but I don't see a cleaner option - suggestions are welcome.

### DIFF
--- a/xla/service/gpu/matmul_utils.cc
+++ b/xla/service/gpu/matmul_utils.cc
@@ -298,14 +298,14 @@ absl::StatusOr<bool> CanFoldTransposeOperandIntoDot(const HloInstruction& dot,
     absl::Span<const int64_t> rhs_contracting_dims, const Shape& output_shape,
     double alpha_real, double alpha_imag, double beta,
     PrecisionConfig::Algorithm precision_algorithm,
-    std::optional<int64_t> algorithm, int64_t compute_precision, bool grad_x,
-    bool grad_y) {
+    std::optional<int64_t> algorithm, int64_t compute_precision,
+    const PrecisionConfig::Precision* input_precisions) {
   return GemmConfig::For(lhs_shape, lhs_batch_dims, lhs_contracting_dims,
                          rhs_shape, rhs_batch_dims, rhs_contracting_dims,
                          /*c_shape=*/output_shape, /*bias_shape_ptr=*/nullptr,
                          output_shape, alpha_real, alpha_imag, beta,
                          precision_algorithm, algorithm, compute_precision,
-                         grad_x, grad_y);
+                         input_precisions);
 }
 
 /*static*/ absl::StatusOr<GemmConfig> GemmConfig::For(
@@ -316,8 +316,8 @@ absl::StatusOr<bool> CanFoldTransposeOperandIntoDot(const HloInstruction& dot,
     const Shape* bias_shape_ptr, const Shape& output_shape, double alpha_real,
     double alpha_imag, double beta,
     PrecisionConfig::Algorithm precision_algorithm,
-    std::optional<int64_t> algorithm, int64_t compute_precision, bool grad_x,
-    bool grad_y) {
+    std::optional<int64_t> algorithm, int64_t compute_precision,
+    const PrecisionConfig::Precision* input_precisions) {
   absl::Span<const int64_t> lhs_col_dims = lhs_contracting_dims;
   TF_ASSIGN_OR_RETURN(
       std::vector<int64_t> lhs_row_dims,
@@ -427,10 +427,10 @@ absl::StatusOr<bool> CanFoldTransposeOperandIntoDot(const HloInstruction& dot,
                     {alpha_real, alpha_imag},
                     beta,
                     compute_precision,
+                    {input_precisions[0], input_precisions[1]},
                     precision_algorithm,
                     algorithm,
-                    grad_x,
-                    grad_y};
+                    std::nullopt};
 }
 
 namespace {
@@ -480,10 +480,9 @@ bool IsTf32Allowed(PrecisionConfig::Algorithm algorithm,
     vector_bias_shape = gemm->operand(vector_bias_index)->shape();
   }
 
-  auto attributes = gemm->frontend_attributes().map();
-  bool grad_x = (attributes["grad_x"] == "true");
-  bool grad_y = (attributes["grad_y"] == "true");
-
+  xla::PrecisionConfig::Precision input_precisions[2];
+  input_precisions[0] = config.precision_config().operand_precision(0);
+  input_precisions[1] = config.precision_config().operand_precision(1);
   int64_t precision = se::blas::kDefaultComputePrecision;
   for (auto operand_precision : config.precision_config().operand_precision()) {
     precision = std::max(precision, static_cast<int64_t>(operand_precision));
@@ -499,7 +498,7 @@ bool IsTf32Allowed(PrecisionConfig::Algorithm algorithm,
       /*bias_shape_ptr=*/
       vector_bias_shape ? &vector_bias_shape.value() : nullptr, output_shape,
       config.alpha_real(), config.alpha_imag(), config.beta(),
-      precision_algorithm, algorithm, precision, grad_x, grad_y);
+      precision_algorithm, algorithm, precision, input_precisions);
 }
 
 absl::StatusOr<GemmConfig::DescriptorsTuple> GemmConfig::GetMatrixDescriptors(
@@ -559,8 +558,7 @@ absl::Status DoGemmWithAlgorithm(const se::gpu::MatrixDescriptor& lhs,
                                  se::blas::AlgorithmType algorithm,
                                  se::blas::ComputePrecision compute_precision,
                                  const se::NumericOptions& numeric_options,
-                                 se::blas::ProfileResult* profile_result,
-                                 se::blas::CallContext context) {
+                                 se::blas::ProfileResult* profile_result) {
   CHECK(output.transpose == se::blas::Transpose::kNoTranspose);
   PrimitiveType lhs_type = primitive_util::NativeToPrimitiveType<Input>();
   PrimitiveType output_type = primitive_util::NativeToPrimitiveType<Output>();
@@ -585,13 +583,13 @@ absl::Status DoGemmWithAlgorithm(const se::gpu::MatrixDescriptor& lhs,
         rhs.cast<Input>(), rhs.leading_dim_stride, rhs.batch_stride, beta,
         &output_data, output.leading_dim_stride, output.batch_stride,
         output.batch_size, computation_type, algorithm, numeric_options,
-        profile_result, context);
+        profile_result);
   } else {
     return blas->BlasGemmWithAlgorithm(
         stream, lhs.transpose, rhs.transpose, output.m, output.n, output.k,
         alpha, lhs.cast<Input>(), lhs.leading_dim_stride, rhs.cast<Input>(),
         rhs.leading_dim_stride, beta, &output_data, output.leading_dim_stride,
-        computation_type, algorithm, numeric_options, profile_result, context);
+        computation_type, algorithm, numeric_options, profile_result);
   }
 }
 
@@ -605,8 +603,7 @@ absl::Status DoGemm(const se::gpu::MatrixDescriptor& lhs,
                     std::optional<se::blas::AlgorithmType> algorithm,
                     se::blas::ComputePrecision compute_precision,
                     const se::NumericOptions& numeric_options,
-                    se::blas::ProfileResult* profile_result,
-                    se::blas::CallContext context) {
+                    se::blas::ProfileResult* profile_result) {
   CHECK(output.transpose == se::blas::Transpose::kNoTranspose);
   se::DeviceMemory<Output> output_data(output.data);
   auto* blas = stream->parent()->AsBlas();
@@ -620,8 +617,7 @@ absl::Status DoGemm(const se::gpu::MatrixDescriptor& lhs,
   if (algorithm) {
     return DoGemmWithAlgorithm<Scale, Input, Output>(
         lhs, rhs, output, workspace, alpha, beta, stream, precision_algorithm,
-        *algorithm, compute_precision, numeric_options, profile_result,
-        context);
+        *algorithm, compute_precision, numeric_options, profile_result);
   }
 
   if (output.batch_size != 1) {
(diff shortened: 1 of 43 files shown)


xla/stream_executor/numeric_options.h
```diff
@@ -21,15 +21,26 @@ namespace stream_executor {
...
 
-  NumericOptions() : require_determinism(false), allow_tf32(true) {}
+  NumericOptions() : require_determinism(false), allow_tf32(true), grad_flags(0) {}
 
   // If true, the op must be deterministic
   bool require_determinism;
   // If true, float32 inputs can be rounded to TensorFloat-32 precision
   bool allow_tf32;
+
+  enum GradientFlags {
```
thomasjoerg, line 34: Please document the enum you introduced here in code comments. I'm not an expert on StreamExecutor or hipBLAS, but I think I should get an idea what the enum values are used for.

thomasjoerg (changes requested):

thomasjoerg:
@chsigg you know StreamExecutor really well. Can you review this PR please?

xla/xla_data.proto
```diff
@@ -931,7 +931,10 @@ message PrecisionConfig {
     // Each U8/S8 value in a tensor actually represents 2 nibble values.
     PACKED_NIBBLE = 3;
 
-    // Next: 4
+    ACTIVATION = 4;
```
thomasjoerg, line 934: Please add a comment to explain the new enum values. Are ACTIVATION and GRADIENT best added here? It seems orthogonal to precision scale from DEFAULT to HIGHEST in this enum.

xla/tests/dot_operation_test.cc
```diff
@@ -41,6 +41,118 @@ limitations under the License.
...
+        if(nd+2 >= y_shape.rank() && (y_shape.element_type()==F16 || y_shape.element_type()==BF16))
+          d = (d+7) & ~7;
+        size_y_ *= d;
+        nd++;
+      }
+      if(nontrivial_dims <= 1)
+        trivial_ = true;
+      size_y_ *= xla::primitive_util::BitWidth(y_shape.element_type())/8;
+    }
+    //printf("Shapes: %s, %s\n", x_shape.ToString().c_str(), y_shape.ToString().c_str());
```
chsigg, line 99: Does it make sense to remove these?

xla/tests/dot_operation_test.cc
```diff
@@ -41,6 +41,118 @@ limitations under the License.
...
+
+class GradFlagVerifier
+{
+  size_t size_x_=0, size_y_=0, flags_=0;
+  bool trivial_ = false;
+public:
+  GradFlagVerifier(XlaBuilder& builder, XlaOp& x, XlaOp& y, PrecisionConfig& cfg) {
+    cfg = xla::PrecisionConfigDEFAULT();
+#if TENSORFLOW_USE_ROCM
+    flags_ = rand() & 3;
```
chsigg, line 57: Do I understand correctly that you pick a random flag value?

xla/tests/dot_operation_test.cc
```diff
@@ -41,6 +41,118 @@ limitations under the License.
...
+
+      int flags = std::stoi(line.substr(lead.size(), sep1-lead.size()));
+      int size1 = std::stoi(line.substr(sep1+1, sep2-sep1-1));
+      int size2 = std::stoi(line.substr(sep2+1, sep3-sep2-1));
+
+      // Dot inputs may be swapped at some point before reaching rocblas
+      // or hipblaslt; if so, flags should be swapped too
+      int rev_flags = ((flags_&1)<<1) | ((flags_&2)>>1);
+      if(size_x_!=0 && size_y_!=0) {
+        EXPECT_TRUE((flags==flags_+256 && size1==size_x_ && size2==size_y_)
```
chsigg, line 142: What are these '256' constants? Maybe pull them out into a constant with a descriptive name?

third_party/tsl/tsl/protobuf/dnn.proto
```diff
@@ -177,6 +177,7 @@ message ConvolutionDescriptorProto {
   ConvolutionMode convolution_mode = 6;
   // Tensorflow node name, same as in NodeDef, for debugging purposes.
   string name = 7;
+  int32 grad_flags = 8;
```
chsigg, line 180: Please add documentation.

xla/service/gpu/matmul_utils.cc
```diff
@@ -636,25 +626,21 @@ absl::Status RunGemm(const GemmConfig& config, se::DeviceMemoryBase lhs_buffer,
...
-  se::blas::CallContext context = se::blas::CallContext::kNone;
-  if (config.grad_x) {
-    context = desc.operands_swapped ? se::blas::CallContext::kBackpropInput2
-                                    : se::blas::CallContext::kBackpropInput1;
-  }
-  if (config.grad_y) {
-    context = desc.operands_swapped ? se::blas::CallContext::kBackpropInput1
-                                    : se::blas::CallContext::kBackpropInput2;
+  if(desc.operands_swapped) {
+    numeric_options.grad_flags = ((numeric_options.grad_flags & 1)<<1)
```
chsigg, line 641: This is swapping the lowest two bits, right? Please pull out into a function with a descriptive name or add a comment.

xla/service/gpu/matmul_utils.h
```diff
@@ -43,6 +43,31 @@ limitations under the License.
...
+      }
+      if(value>=PrecisionConfig::ACTIVATION && value<=PrecisionConfig::GRADIENT)
+      {
+        if(grad_count<2 && value == PrecisionConfig::GRADIENT)
+          grad_flags |= 1 << grad_count;
+        grad_count++;
+      }
+    }
+    if(grad_count==2)
+      grad_flags |= 256;
```
chsigg, line 67: What is this '256' constant? Can it be pulled out into a constant?

xla/service/gpu/matmul_utils.h
```diff
@@ -43,6 +43,31 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+
+inline void GetXlaPrecisionConfigNumericFlags(const mlir::ArrayAttr& precision_config,
```
chsigg, line 47: How does this function relate to the one with the same name in xla/util.h?

chsigg (commented):
> I am somewhat unhappy with the method I am currently using to get data from stream executor back to the unit test

Can you elaborate?

xla/xla_data.proto
ekuznetsov139: Currently PrecisionConfig contains one field containing precision information for each input. This adds another field containing information whether this is a gradient or an activation. This is perhaps not the most natural place to keep this, but any alternative would require passing precision in one object and gradient flags in another object, which is precisely what I am trying to avoid.

xla/tests/dot_operation_test.cc
ekuznetsov139: Absolutely, once it's no longer marked as .work in progress

ekuznetsov139:
> > I am somewhat unhappy with the method I am currently using to get data from stream executor back to the unit test
> 
> Can you elaborate?

I am using printf inside StreamExecutor to print the values that reach it, capture stdout using googletest, and parse whatever is being printed. It seems a bit crude, but like I said, I don't see another easy way to get that information back to the unit test.

xla/service/gpu/matmul_utils.h
ekuznetsov139: It is the same function, but with a different data type (at some point precision config gets converted into mlir::ArrayAttr).

xla/service/gpu/matmul_utils.h
ekuznetsov139: It is an indication that the field has been initialized, I have it declared as NumericOptions::GF_Initialized (not sure if that declaration is visible at this point in the header). I'll try to get rid of 256.

xla/tests/dot_operation_test.cc
ekuznetsov139: Yes, would it be better to loop over all values (increasing execution time of the unit test 4x)?

xla/xla_data.proto
thomasjoerg: Please document the intended usage of this enum. Why are you trying to avoid adding a gradient flag?

ekuznetsov139:
The PR has been reworked:
* Instead of getting traces via stdout, I've added a specialized stream executor method for this purpose. It only has a single use right now, but I think it is a useful feature to have and more tests can be added on top of it.
* I've separated the Precision message into two parts: one for precision and one for gradients. It's a cleaner arrangement, but, unfortunately, it meant that I had to touch quite a few more files (e.g., there are many places where logic expects PrecisionConfig to contain an array of elementary values, and code that works for elementary values does not work for composite values.) I almost got it working, but there's some sort of internal compilation error that does not make total sense to me. 

I'm going to rebase this on top of latest trunk and then try to fix the compilation error.

xla/xla_data.proto
```diff
@@ -933,7 +933,18 @@ message PrecisionConfig {
...
+    ACTIVATION = 1;
+    GRADIENT = 2;
+  }
+
+  message PrecisionGradient {
+    Precision precision = 1;
+    Gradient gradient = 2;
+  };
+
+  repeated PrecisionGradient operand_precision = 1;
```
tdanyluk, line 947: I'm wondering if it would be a more compatible change to do this:
```
repeated Precision operand_precision = 1;
repeated Gradient operand_gradient = 1; (Is this a per operand thing BTW?)
```

I'm wondering how many things would the current change break internally. But maybe that's fine. Perhaps other Googlers can comment? @thomasjoerg @ddunl

thomasjoerg: As proposed the change is not backward compatible, you are right Tamas. I don't know what backward compatibility requirements this proto has TBH, but if in doubt, I'd prefer to keep compatibility the way you suggested.

xla/xla_data.proto
```diff
@@ -933,7 +933,18 @@ message PrecisionConfig {
 
     // Next: 4
   }
-  repeated Precision operand_precision = 1;
+  enum Gradient {
```
thomasjoerg, line 936: Gradient is a misleading name, because a Gradient can be an Activation or a Gradient. Can you please add documentation on the intended use?

xla/xla_data.proto
tdanyluk: One more reason to separate the gradient from the precision is that we actually try to move away from using precision values and we may change over to using "algorithm" values soon. So it's possible that in the future, the operand_precisions will not be set. (In the first step we will still allow setting them as a backwards compatibility measure.)

kamaljeeti:
HI @ekuznetsov139, , can you have a look at the comments? Thanks.

ekuznetsov139:
> HI @ekuznetsov139, , can you have a look at the comments? Thanks.

@kamaljeeti  I am now trying to merge part of this PR separately as #9827. I will revisit it and address the comments after that is done.

ekuznetsov139:
What I really need is to specify dynamic range information for inputs. Flagging an input as "Gradient" is essentially a shorthand for  "full float16 range of exponents." So, perhaps something like this would work?

``` 
 enum Precision {
    DEFAULT = 0;
    HIGH = 1;
    E4_BIAS10 = 2; // Values representable by F8_E4M3 bias 10
    E4_BIAS8 = 3; // Values representable by F8_E4M3 bias 8
    E4_BIAS6 = 4; // Values representable by F8_E4M3 bias 6
    ... etc ...
    E5; // Values representable by F8_E5M2 bias 16
    HIGHEST ;
    // Each U8/S8 value in a tensor actually represents 2 nibble values.
    PACKED_NIBBLE ;
  }
```
For gradients, I would set precision to E5.

This would be a ROCm-specific usage (at present, ROCm entirely ignores Precision). It would serve two purposes:

* On MI200/MI250, as now, GEMM and convolutions ops would switch internally to BF16 when one of the inputs is a gradient (due to the way MI200 tensor cores handle float16 denormals).
* On MI300, this allows us to do GEMMs and convolutions internally in float8 if requested by the user, possibly after applying bias scaling.

ekuznetsov139:
I need to carry information for both operands, it can be packed into a single Algorithm field (there needs to be dedicated logic to handle swapping operands anyway), but packing it will mean adding a large number of enum values to cover all permutations.

It would work nicely if, as mentioned in the RFC, Algorithm is a struct and possible  values of AType and BTYpe include at least F8_E4M3 and F8_E5M2 (and preferably multiple exponent bias values for F8_E4M3).

tdanyluk:
Thanks @ekuznetsov139!

Are these new types like "E4_BIAS10" also storage types? Or do we have a storage type of let's say F16 and we just cast it to E4_BIAS10 before calculation?

If these are the actual storage types, we could add them to the types here:
https://github.com/openxla/xla/blob/dd78dd8d709cf4bef10532129d265ae569d0d680/xla/xla_data.proto#L83

But if not, then probably you're right about the need to change the way algorithms work.
Do you expect the left and right operand to have different type in practice?

ekuznetsov139:
> Thanks @ekuznetsov139!
> 
> Are these new types like "E4_BIAS10" also storage types? Or do we have a storage type of let's say F16 and we just cast it to E4_BIAS10 before calculation?
> 
> If these are the actual storage types, we could add them to the types here:
> 
> https://github.com/openxla/xla/blob/dd78dd8d709cf4bef10532129d265ae569d0d680/xla/xla_data.proto#L83
> 
> But if not, then probably you're right about the need to change the way algorithms work. Do you expect the left and right operand to have different type in practice?
> 
> (FYI I'll be out for about a week, but I'll get back to this after that)

These are not actual storage types. They are hints informing the backend of the expected distribution of values. This would be used with F16 or BF16 storage types and cast into F8_E4M3 just before calculation. If bias is not 8, input operands must be multiplied or divided by some constant before conversion. 

In the forward pass, I expect that both operands will be either E4_BIAS8 or HIGHEST most of the time, but one or both might occasionally have a different value of bias. There's no reason to expect left and right operands to be the same type. In the backward pass, whichever input is the gradient will be E5.

ekuznetsov139:
I thought the exponent bias of F8E4M3 was 8, but that's a minor implementation detail.

PR #9638, opened 2024-02-19
[ROCm] Fix recent build failures

draganmladjenovic:


xla/stream_executor/rocm/rocm_executor.cc
```diff
@@ -464,7 +465,8 @@ absl::Status GpuExecutor::LoadModuleFromHsaco(const char* hsaco,
 // This is a non-essential operation; if there's a failure, proceed without
 // logging an error. It's nearly certain that in case of failures, we'd never
 // get here in the first place; these are very low-impact routines.
-void GpuExecutor::VlogOccupancyInfo(const Kernel& kernel,
+void GpuExecutor::VlogOccupancyInfo(const DeviceDescription& device_description,
+                                    const Kernel& kernel,
```
thomasjoerg, line 469: Nit: Is there an invariant the caller need to be aware of, now that `device_description` needs to be supplied separately from `kernel`?

thomasjoerg (approved):

xla/stream_executor/rocm/rocm_executor.cc
draganmladjenovic: Don't think so. At least for us it is not used.

PR #9666, opened 2024-02-20
[ROCm] Fused convolution+bias+activation

ekuznetsov139:
This PR enables MIOpen convolution+bias+activation fusion for ROCm and updates fusion unit tests accordingly.

### DIFF
--- a/xla/service/gpu/cudnn_fused_conv_rewriter.cc
+++ b/xla/service/gpu/cudnn_fused_conv_rewriter.cc
@@ -96,6 +96,10 @@ bool IsNonDepthwiseConvCustomCall(const HloInstruction* instr) {
   return IsConvCustomCall(instr) && !IsConvDepthwise(instr);
 }
 
+bool IsROCm(se::GpuComputeCapability cc) {
+  return std::holds_alternative<se::RocmComputeCapability>(cc); 
+}
+
 // elu, relu6, and leaky-relu activations are supported in cudnn via the
 // "runtime fusion" engine, which JIT compiles C++ code.  This can be slow to
 // compile, so we guard it with a debug option.
@@ -106,8 +110,12 @@ bool IsNonDepthwiseConvCustomCall(const HloInstruction* instr) {
 // Note that as of writing, xla_gpu_use_runtime_fusion is disabled by default
 // due to apparent bugs in cudnn 8.9.0.  See debug_options_flags.cc for details.
 bool ShouldUseCudnnRuntimeFusion(const DebugOptions& debug_opts,
-                                 se::CudaComputeCapability cc) {
-  return debug_opts.xla_gpu_use_runtime_fusion() && cc.IsAtLeast(7, 5);
+                                 se::GpuComputeCapability cc) {
+  const auto* cuda_cc = std::get_if<se::CudaComputeCapability>(&cc);
+  if(cuda_cc != nullptr)
+    return debug_opts.xla_gpu_use_runtime_fusion() && cuda_cc->IsAtLeast(7, 5);
+  else
+    return true;
 }
 
 bool IsSuitableForCudnnRuntimeFusion(HloInstruction* conv) {
@@ -984,7 +992,7 @@ absl::StatusOr<bool> FuseSideInputAlpha(HloComputation* comp) {
 }
 
 absl::StatusOr<bool> FuseElu(HloComputation* comp,
-                             se::CudaComputeCapability cc) {
+                             se::GpuComputeCapability cc) {
   if (!ShouldUseCudnnRuntimeFusion(comp->parent()->config().debug_options(),
                                    cc)) {
     return false;
@@ -1085,7 +1093,7 @@ absl::StatusOr<bool> FuseRelu(HloComputation* comp) {
 }
 
 absl::StatusOr<bool> FuseRelu6(HloComputation* comp,
-                               se::CudaComputeCapability cc) {
+                               se::GpuComputeCapability cc) {
   if (!ShouldUseCudnnRuntimeFusion(comp->parent()->config().debug_options(),
                                    cc)) {
     return false;
@@ -1134,7 +1142,7 @@ absl::StatusOr<bool> FuseRelu6(HloComputation* comp,
 }
 
 absl::StatusOr<bool> FuseLeakyRelu(HloComputation* comp,
-                                   se::CudaComputeCapability cc) {
+                                   se::GpuComputeCapability cc) {
   if (!ShouldUseCudnnRuntimeFusion(comp->parent()->config().debug_options(),
                                    cc)) {
     return false;
@@ -1254,7 +1262,10 @@ absl::StatusOr<bool> FuseConvertToF16(HloComputation* comp) {
   return changed;
 }
 
-absl::StatusOr<bool> FuseConvertToS8(HloComputation* comp) {
+absl::StatusOr<bool> FuseConvertToS8(HloComputation* comp,
+  se::GpuComputeCapability cc) {
+  if(IsROCm(cc))
+    return false;
   bool changed = false;
   for (HloInstruction* instr : comp->MakeInstructionPostOrder()) {
     HloInstruction* gte = nullptr;
@@ -1480,9 +1491,12 @@ absl::StatusOr<bool> CudnnFusedConvRewriter::Run(
     bool changed = false;
     // Rewrite FP8 convolutions and supported adjacent pointwise ops into a
     // ForwardGraph Custom Call.
-    TF_ASSIGN_OR_RETURN(changed, F8GraphConv(comp, compute_capability_));
-    if (changed) {
-      return changed;
+    if(!IsROCm(compute_capability_)) {
+      auto cc = std::get<se::CudaComputeCapability>(compute_capability_);
+      TF_ASSIGN_OR_RETURN(changed, F8GraphConv(comp, cc));
+      if (changed) {
+        return changed;
+      }
     }
     // Fuse "inside out" starting with the operations closest to the conv.
     TF_ASSIGN_OR_RETURN(changed, FuseRemoveConvertInConv(comp));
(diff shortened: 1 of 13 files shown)


xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
```diff
@@ -295,6 +319,105 @@ TEST_F(CudnnFusedConvRewriterTest, TestBias) {
...
+
+      conv = TYPE[1,<<<format>>>,64] convolution(input, filter), window={size=3x3 pad=1_1x1_1}, dim_labels=b01f_01io->b01f, feature_group_count=1
+      broadcasted_bias = TYPE[1,<<<format>>>,64] broadcast(bias), dimensions={3}
+      add1 = TYPE[1,<<<format>>>,64] add(conv, broadcasted_bias)
+      ROOT relu = TYPE[1,<<<format>>>,64] maximum(zeros, add1)
+    })";
+    absl::flat_hash_map<absl::string_view, absl::string_view> replacements;
+    replacements["<<<format>>>"] = "3,3";
+    replacements["TYPE"] = "f16";
+    const std::string hlo = absl::StrReplaceAll(code, replacements);
```
thomasjoerg, line 371: Since you are using the replacements just one, consider to just hard-code the test fixture.

xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
```diff
@@ -295,6 +319,105 @@ TEST_F(CudnnFusedConvRewriterTest, TestBias) {
...
+          while(counter < 0);
+        else if(id==3 && iter==0)
+          while(counter < 5);
+        EXPECT_TRUE(Run(std::move(modules[id*25+iter]), true));
+        fflush(stdout);
+      }
+    };
+    std::thread threads[4];
+    for(int i=0; i<4; i++)
+      threads[i] = std::thread(executor, i);
```
thomasjoerg, line 400: What is the motivation for the multi thread test? All threads run the same module, right?

thomasjoerg (changes requested):

xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
ekuznetsov139: This would potentially flag race conditions due to MIOpen (or CuDNN) executing identical fused operations in different threads.

xla/service/gpu/cudnn_fused_conv_rewriter.cc
```diff
@@ -94,8 +94,12 @@ bool IsNonDepthwiseConvCustomCall(const HloInstruction* instr) {
...
 // due to apparent bugs in cudnn 8.9.0.  See debug_options_flags.cc for details.
 bool ShouldUseCudnnRuntimeFusion(const DebugOptions& debug_opts,
-                                 se::CudaComputeCapability cc) {
-  return debug_opts.xla_gpu_use_runtime_fusion() && cc.IsAtLeast(7, 5);
+                                 se::GpuComputeCapability cc) {
+  auto pgcc = std::get_if<se::CudaComputeCapability>(&cc);
+  if(pgcc != nullptr)
+    return debug_opts.xla_gpu_use_runtime_fusion() && pgcc->IsAtLeast(7, 5);
+  else
+    return true;
```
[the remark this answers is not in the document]
ekuznetsov139: As I understand, runtime fusion simply means that kernels are being compiled on-demand at the time of call? ROCm convolutions, both fused and unfused, have been working this way since day one.

xla/stream_executor/rocm/rocm_dnn.cc
```diff
@@ -4656,6 +4739,408 @@ bool MIOpenSupport::DeriveOutputBatchDescriptor(
...
+    std::vector<std::unique_ptr<const dnn::FusedConvRunner>>* out_exec_plans) {
+
+  VLOG(2) << "MIOpenSupport::GetFusedConvolveRunners";
+  VLOG(2) << "filter_descriptor " << filter_descriptor.ndims();
+  if (filter_descriptor.ndims() > 2)
+    return tsl::errors::Unimplemented("MIOpen does not correctly fuse conv3d");
+
+  std::vector<dnn::AlgorithmDesc> algorithms;
+
+  CudaComputeCapability cuda_compute_capability;
```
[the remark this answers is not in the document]
ekuznetsov139: I believe this got here on accident, I'll remove it

xla/service/gpu/cudnn_fused_conv_rewriter.cc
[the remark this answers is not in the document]
thomasjoerg: That makes sense. I got confused by the name of the flag. It is cuDNN specific and should have that in the name, but that's outside of the scope of this PR of course.

xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
thomasjoerg: The race condition would need to make `Run(...)` return `false` for this test to catch it, which may or may not happen. Wouldn't running under ThreadSanetizer give you better chances of finding such issues? Also, this test class doesn't seem the right place to test MIOpen or cuDNN for thread safety, in my opinion.

xla/stream_executor/rocm/rocm_dnn.cc
```diff
@@ -4656,6 +4740,396 @@ bool MIOpenSupport::DeriveOutputBatchDescriptor(
...
+        biasActStatus = inplace_call(DeviceMemory<Eigen::bfloat16>(output_data), DeviceMemory<Eigen::bfloat16>(bias_data));
+      else 
+        return tsl::errors::Internal("Unsupported data type");
+
+      return ::tsl::OkStatus();
+  }
+
+  // Internal form of ToAlgorithmDesc without the StatusOr.
+  dnn::AlgorithmDesc MakeAlgorithmDesc() const {
+    return {algo_id_, /*tensor_ops_enabled_*/ true, workspace_size_};
```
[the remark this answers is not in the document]
ekuznetsov139: AMD completely ignores TF32 settings.

ekuznetsov139:
Bump

kamaljeeti:
Hi @tdanyluk , can you look into this once? Thanks.

xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
```diff
@@ -241,6 +252,19 @@ class CudnnFusedConvRewriterTest : public GpuCodegenTest {
...
+#if GOOGLE_CUDA
+  #if (CUDA_VERSION < 12000 || CUDNN_VERSION < 8900)
+    #define MAYBE_SKIP_TEST(CAUSE) \
+      if(absl::string_view(CAUSE) == "F8") \
+        GTEST_SKIP() << "FP8 convolutions require CUDA 12 and cuDNN 8.9.";
+  #else
+      #define MAYBE_SKIP_TEST(CAUSE)
+  #endif
+#else
+#define MAYBE_SKIP_TEST(CAUSE) \
```
tdanyluk, line 264: If you wrap the macro implementations in a `do { ... } while(0)`, then you can put a semicolon after the macro call, which looks more natural. E.g. `MAYBE_SKIP_TEST("F8");`
https://hownot2code.wordpress.com/2016/12/05/do-while-0-in-macros/

(I mean assuming that this works with gtest.)

tdanyluk (changes requested):
Thanks for the fixes and sorry for the delay on this.

Could you fix these 2 macro/ifdef related issues?

tdanyluk (approved):
Thanks for the changes, It's fine by me now, but I think it's better to get @thomasjoerg's approval too.

xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
ekuznetsov139: The test is still there, but it was changed to use RunAndCompare instead of Run, so it would catch mismatches. I don't see any other places where backend implementations of fused convolutions are tested or could be tested for concurrency issues. 

As it is, worst thing that might happen is that there might be an issue but the test won't catch it. I can certainly remove the test, but I'm not sure how dropping the test would improve matters.

xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
```diff
@@ -298,6 +331,148 @@ TEST_F(CudnnFusedConvRewriterTest, TestBias) {
...
+        else if(id==2) {
+          counter3++;
+          if(iter==0)
+            while(counter < 0);
+        } else if(id==3 && iter==0) {
+          while(counter < 5);
+        }
+        EXPECT_TRUE(RunAndCompare(std::move(modules[id*25+iter]),
+          ErrorSpec{0.01}));
+        fflush(stdout);
```
thomasjoerg, line 450: Nit: Is this required? It's not typically used in XLA tests.

xla/service/gpu/cudnn_fused_conv_rewriter_test.cc
thomasjoerg: Tests need to be maintained and in its current form this test will be hard to maintain. When this test breaks, the person looking at the breakage needs to right information to understand the issue.

* A code comment should explain the motivation for this test, because it is different from the non-threaded tests around it.
* Please explain where the race condition would occur. I'm not sure here - the threads don't seem to share anything and don't communicate.
* This test does not test XLA's conv rewriter, but MIOpen / cuDNN, right? This is not obvious and deserves a comment too. Would XLA actually call convs on multiple threads (inter-op parallelism)?

ekuznetsov139: The threads don't share anything and they don't communicate, but, at least in the MIOpen case, stream executor will cache and reuse the same "fusion plan" handle for operations with identical parameters (https://github.com/openxla/xla/blob/main/xla/stream_executor/rocm/rocm_dnn.cc#L560), and potentially the same handle may be used in parallel in different threads. CuDNN may be doing something similar at some level.

Since this PR is almost 2 months old by now, rather than argue the point further, I'm going to drop the test. If  race conditions are really possible, someone will have fun tracking them down the hard way.

kamaljeeti:
Hi @thomasjoerg , can you please look into this ? Thanks.

thomasjoerg (approved):

PR #9688, opened 2024-02-20
[XLA:CPU] Add F16 support for oneDNN matmul

kanvi-nervana:
This PR enables F16 support for dot operation when onednn is enabled and adds tests. In the existing pipeline, F16 dot is converted to FP32 through the change-op-data-type pass. With the changes in this PR, if onednn is enabled and the conditions for rewriting dot to onednn matmul are met then F16 dot is not upcast to FP32.


kamaljeeti:
Hi @penpornk , can you look into this once? Thanks.

penpornk (changes requested):
Thank you very much for the PR and sorry for the delay! I have one minor comment.

kanvi-nervana:
@penpornk I have made some changes to this PR, can you please re-review? Also, I do not see the test results anymore.

penpornk (approved):
Thank you for the change and sorry for the delay!

kanvi-nervana:
@penpornk Thanks for the review. There seem to be some internal failures which I cannot check. Can you please share the failure log? Thanks!

PR #9730, opened 2024-02-21
[ROCm] Fix build errors in rocm_diagnostics

draganmladjenovic:


kamaljeeti (approved):
Hi @draganmladjenovic , IMO changes seem fine to me, but can you please also mention what kind of build errors are fixed. (any memory leak, type mismatch or any other compiling error)

draganmladjenovic:
> Hi @draganmladjenovic , IMO changes seem fine to me, but can you please also mention what kind of build errors are fixed. (any memory leak, type mismatch or any other compiling error)

Sorry. Haven't checked further. Whatever changed it seems that using namespace std; is gone form some of the dependencies so this file no longer compiles due to all string uses w/o namespace prefix.

ddunl (approved):

PR #9827, opened 2024-02-22
[ROCm] Stream executor API logging API

ekuznetsov139:
Separated from https://github.com/openxla/xla/pull/9593.

This adds a low-overhead ability to log stream executor API calls, for testing purposes (unit tests can enable the functionality and use it to verify that high-level operations result in correct sequences of low-level calls). At present, this only covers BlasGemm / BlasLt calls on ROCm backend.

### DIFF
--- a/xla/stream_executor/gpu/gpu_executor.h
+++ b/xla/stream_executor/gpu/gpu_executor.h
@@ -319,6 +319,30 @@ class GpuExecutor : public StreamExecutor {
   int cc_major() const { return cc_major_; }
   int cc_minor() const { return cc_minor_; }
 
+  absl::StatusOr<std::vector<ApiTrace>> ExtractApiTrace() override {
+    absl::MutexLock lock(&logger_mu_);
+    return std::move(argument_logs_);
+  }
+
+  absl::Status RecordApiTrace(ApiTrace call) override {
+    absl::MutexLock lock(&logger_mu_);
+    if(std::holds_alternative<GemmCallTrace>(call) && 
+      (argument_logging_mode_ & kLogGemm)) {
+      argument_logs_.push_back(call);
+    }
+    return absl::OkStatus();
+  }
+
+  bool SetArgumentLoggingMode(uint64_t mode) override {
+    absl::MutexLock lock(&logger_mu_);
+    argument_logging_mode_ = mode;
+    return true;
+  }
+
+  uint64_t GetArgumentLoggingMode() const {
+    return argument_logging_mode_; 
+  }
+
  private:
   // Host callback landing routine invoked by CUDA.
   // data: User-provided callback provided to HostCallback() above, captured
@@ -438,6 +462,13 @@ class GpuExecutor : public StreamExecutor {
   // Memoized BLAS support object -- we only want to create this once when asked
   // for a BLAS interface.
   std::unique_ptr<blas::BlasSupport> blas_ ABSL_GUARDED_BY(mu_);
+
+  absl::Mutex logger_mu_;
+
+  mutable std::vector<ApiTrace> argument_logs_
+    ABSL_GUARDED_BY(logger_mu_);
+
+  uint64_t argument_logging_mode_ = 0;
 
   GpuExecutor(const GpuExecutor&) = delete;
   void operator=(const GpuExecutor&) = delete;
--- a/xla/stream_executor/rocm/hip_blas_lt.cc
+++ b/xla/stream_executor/rocm/hip_blas_lt.cc
@@ -418,6 +418,11 @@ absl::Status BlasLt::MatmulPlan::DoMatmul(
     std::optional<DeviceMemoryBase> workspace,
     std::optional<ScratchAllocator*> scratch_allocator,
     blas::ProfileResult* profile_result) const {
+  absl::Status status = blas_lt_ref_.parent_->RecordApiTrace(
+    StreamExecutorInterface::GemmCallTrace{
+      StreamExecutorInterface::GemmCallTrace::GemmType::kBlasLt, 
+      0, a.size(), b.size()});
+
   TF_ASSIGN_OR_RETURN(
       std::optional<gpu::GpuTimer> timer,
       gpu::GpuTimer::CreateIfNeeded(
--- a/xla/stream_executor/rocm/rocm_blas.cc
+++ b/xla/stream_executor/rocm/rocm_blas.cc
@@ -222,6 +222,26 @@ rocblas_side ROCMBlasSide(blas::Side side) {
   }
 }
 
+int DtypeSize(blas::DataType type) {
+  switch (type) {
+    case blas::DataType::kHalf:
+    case blas::DataType::kBF16:
+      return 2;
+    case blas::DataType::kFloat:
+      return 4;
+    case blas::DataType::kDouble:
+      return 8;
+    case blas::DataType::kInt8:
+      return 1;
+    case blas::DataType::kComplexFloat:
+      return 8;
+    case blas::DataType::kComplexDouble:
+      return 16;
+    default:
+      return 0;
+  }
+}
+
 absl::StatusOr<rocblas_datatype> AsRocBlasType(blas::DataType type) {
   switch (type) {
     case blas::DataType::kHalf:
@@ -432,6 +452,16 @@ Impl_DoBlasScal(wrap::rocblas_sscal, float,
  *    and ex functions expect the same type as the compute type (i.e. floats.)
  *
  **/
+using sei = StreamExecutorInterface;
+using GemmCallTrace = sei::GemmCallTrace;
+
+// Log the GEMM operation if the logging mode is enabled.
+void ROCMBlas::MaybeLogGemmOp(GemmCallTrace::GemmType op,
+                              blas::CallContext context, 
+                              uint64_t size1, uint64_t size2) {
+  auto status = parent_->RecordApiTrace(
+    GemmCallTrace{op, (int)context, size1, size2});
+}
 
 absl::Status ROCMBlas::DoBlasGemm(
     Stream *stream, blas::Transpose transa, blas::Transpose transb, uint64_t m,
@@ -439,6 +469,9 @@ absl::Status ROCMBlas::DoBlasGemm(
     const DeviceMemoryBase &a, int lda, const DeviceMemoryBase &b, int ldb,
     const void *beta, DeviceMemoryBase *c, int ldc,
     const NumericOptions &numeric_options, blas::CallContext context) {
+  MaybeLogGemmOp(GemmCallTrace::GemmType::kPlain, context, 
+    m*k*DtypeSize(dtype), n*k*DtypeSize(dtype));
+
   VLOG(1) << absl::StreamFormat(
       "doing rocBLAS GEMM: at=%d bt=%d m=%u n=%u "
       "k=%llu alpha=%p a=%p lda=%d b=%p ldb=%d beta=%p "
@@ -542,6 +575,8 @@ absl::Status ROCMBlas::DoBlasGemmWithAlgorithm(
                                   numeric_options, context));
 
   } else {
+    MaybeLogGemmOp(GemmCallTrace::GemmType::kPlain, context, m*k*DtypeSize(type_a),
+      n * k * DtypeSize(type_a));
     CheckPreconditions(transa, transb, m, n, k, type_a, lda, ldb);
     TF_ASSIGN_OR_RETURN(auto roc_type_a, AsRocBlasType(type_a));
     TF_ASSIGN_OR_RETURN(auto roc_type_c, AsRocBlasType(type_c));
@@ -601,6 +636,7 @@ absl::Status ROCMBlas::DoBlasGemmStridedBatchedWithAlgorithm(
         ldb, stride_b, beta, c, ldc, stride_c, batch_count, numeric_options,
         context));
   } else {
+    MaybeLogGemmOp(GemmCallTrace::GemmType::kStridedBatched, context, a.size(), b.size());
     VLOG(1) << absl::StreamFormat(
         "doing rocBLAS GEMM strided batched with Algorithm: at=%d bt=%d m=%u "
         "n=%u "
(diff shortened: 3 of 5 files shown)


xla/stream_executor/stream_executor_internal.h
```diff
@@ -335,6 +335,36 @@ class StreamExecutorInterface {
...
+    uint64_t size1, size2;
+  };
+  // This may be expanded as necessary to trace other calls
+  using CallArgument = std::variant<GemmCallRecord>;
+
+  // Retrieves and clears internal argument logs.
+  virtual void GetArgumentLogs(std::vector<CallArgument>& logs) {}
+
+  // The default 0 means no logging.
+  enum SE_ArgumentLogging {
```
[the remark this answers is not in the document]
klucke: There's no need to have the "SE_" prefix -- that's implied already from the namespace.  

Also prefer "ArgumentLoggingMode", since that's what this is.

ekuznetsov139:
Bump

kamaljeeti:
Hi @klucke , can you look into this once? Thanks.

xla/stream_executor/stream_executor_internal.h
```diff
@@ -336,11 +336,48 @@ class StreamExecutorInterface {
...
+  // This may be expanded as necessary to trace other calls
+  using ApiTrace = std::variant<GemmCallTrace>;
+
+  // Retrieves and clears internal argument logs.
+  virtual absl::StatusOr<std::vector<ApiTrace> > ExtractApiTrace() {
+    return absl::UnimplementedError("Not implemented");
+  }
+
+  // Bit mask
+  enum class SE_ArgumentLogging: uint64_t {
```
klucke, line 363: Just name this ArgumentLoggingMode.  The "SE_" prefix is redundant, as you're already in the stream_executor namespace.

xla/stream_executor/stream_executor_internal.h
```diff
@@ -336,11 +336,48 @@ class StreamExecutorInterface {
...
+    kNone = 0,
+    kGemm = 1<<0,
+  };
+
+  // Sets the argument logging mode. Returns true if 'mode' is valid.
+  virtual bool SetArgumentLoggingMode(SE_ArgumentLogging mode) { return false; }
```
klucke, line 369: Where's the intended eventual caller of this method?  I'm wondering if we need this all the way at the top level on StreamExecutorInterface.  Is this going to be ROCm-specific?  Can this just go on the rocm_blas class?

ekuznetsov139: The intended caller is here https://github.com/openxla/xla/pull/9593/files#diff-341de512548c6b2a2fa36f5208b4bc1274259307cebde81266e55cd36747b479R60

The initial use is ROCm-specific, but this is written in a way to be platform-independent, since this may be useful elsewhere and for other platforms.

xla/stream_executor/gpu/gpu_executor.h
```diff
@@ -429,6 +442,13 @@ class GpuExecutor : public internal::StreamExecutorInterface {
   absl::flat_hash_map<void*, Stream*> alive_gpu_streams_
       ABSL_GUARDED_BY(alive_gpu_streams_mu_);
 
+public:
```
klucke, line 445: The data members shouldn't be public.

xla/stream_executor/gpu/gpu_executor.h
```diff
@@ -321,6 +321,26 @@ class GpuExecutor : public internal::StreamExecutorInterface {
   int cc_major() const { return cc_major_; }
   int cc_minor() const { return cc_minor_; }
 
+  absl::StatusOr<std::vector<ApiTrace> > ExtractApiTrace() override {
```
ezhulenev, line 324: remove space between `> >`

xla/stream_executor/gpu/gpu_executor.h
```diff
@@ -321,6 +321,26 @@ class GpuExecutor : public internal::StreamExecutorInterface {
...
+    absl::MutexLock lock(&logger_mu_);
+    std::vector<ApiTrace> retval = argument_logs_;
```
ezhulenev, line 326: `retval = std::move(argument_logs_)`? or even return `std::move(...)`, isn't move semantics defined here as move and clear?

xla/stream_executor/rocm/rocm_blas.cc
```diff
@@ -539,6 +574,8 @@ absl::Status ROCMBlas::DoBlasGemmWithAlgorithm(
                                   numeric_options, context));
 
   } else {
+    MaybeLogGemmOp(GemmCallTrace::GemmType::kPlain, context, m*k*DtypeSize(type_a),
+      n*k*DtypeSize(type_a));
```
ezhulenev, line 578: linter will complain about missing spaces `n * k * ...`

xla/stream_executor/rocm/rocm_blas.h
```diff
@@ -38,6 +38,7 @@ limitations under the License.
 #if TF_HIPBLASLT
 #include "xla/stream_executor/rocm/hip_blas_lt.h"
 #endif
```
ezhulenev, line 40: newline before incluce

xla/stream_executor/stream_executor_internal.h
```diff
@@ -326,11 +326,48 @@ class StreamExecutorInterface {
...
+
+  // Retrieves and clears internal argument logs.
+  virtual absl::StatusOr<std::vector<ApiTrace> > ExtractApiTrace() {
+    return absl::UnimplementedError("Not implemented");
+  }
+
+  // Bit mask
+  enum class ArgumentLogging: uint64_t {
+    kNone = 0,
+    kGemm = 1<<0,
```
ezhulenev, line 355: just `1` until not other bit patterns are defined?

xla/stream_executor/gpu/gpu_executor.h
ekuznetsov139: in case of ```return std::move```, would that move still be protected by the mutex?

xla/stream_executor/rocm/rocm_blas.h
ekuznetsov139: Why?

xla/stream_executor/gpu/gpu_timer.cc
```diff
@@ -130,7 +130,11 @@ DeviceMemory<GpuSemaphoreState> GpuTimer::GpuSemaphore::device() {
     // Check the assumption that this device supports unified addressing,
     // otherwise skip the delay kernel
     TF_ASSIGN_OR_RETURN(int status, GpuDriver::GetDeviceAttribute(
+#if GOOGLE_CUDA
```
klucke, line 133: This seems unrelated to the central point of this PR.  It doesn't seem related to the logging API.

In addition, it's sprinkling #ifdefs throughout the code -- as a general rule, ifdefs should be avoided in this manner.  It would be far preferable to refactor the GpuTimer to be ROCm specific if there need to be differences between CUDA and ROCm.

ekuznetsov139: It is unrelated. It is a build fix. Unfortunately, since you guys don't require PRs to build (let alone pass unit tests) on ROCm before accepting them, the trunk of OpenXLA tends to be broken on ROCm much of the time, and I have a choice between including all build fixes in my PRs, or having them in the same broken state as the trunk (in which case, I can't even run unit tests on them).

The discussion about ifdefs would probably be better directed at the person who originally submitted the CUDA specific code into the repository, breaking the AMDGPU build (https://github.com/openxla/xla/blame/main/xla/stream_executor/gpu/gpu_timer.cc#L133).

klucke: If there's a build break, I would encourage you to file a separate PR to fix that build break before trying to submit an unrelated PR if that is causing you problems.  Adding multiple unrelated changes to a single PR can be problematic to effectively land.

As far as ROCm builds, I believe we are working with Chao Chen trying to get a ROCm build as part of openxla.

I can't change what was done in the past, but I can try and work with new PRs to try and fix things that were probably done incorrectly in the past.  I agree that having CUDA-specific code that breaks ROCm builds isn't great.  Fixing those build breaks is definitely something that should be done, but we should endeavor to fix them in the least brittle way to ensure it's easier to prevent those things from happening in the future.

i-chaochen: Hi @klucke  we already have a fixed PR on this build break, could you review and merge it, please? Thanks https://github.com/openxla/xla/pull/11205

kamaljeeti:
Hi @klucke , can you please review this PR and update any suggested changes? Thanks.

xla/stream_executor/gpu/gpu_executor.h
klucke: Yes, that move would still need to be protected by a mutex.

xla/stream_executor/gpu/gpu_executor.h
```diff
@@ -321,6 +321,26 @@ class GpuExecutor : public internal::StreamExecutorInterface {
...
+    absl::MutexLock lock(&logger_mu_);
+    argument_logging_mode_ = mode;
+    return true;
+  }
+
+  ArgumentLogging GetArgumentLoggingMode() const {
+    return argument_logging_mode_; 
+  }
+
+  std::vector<ApiTrace>& ArgumentLogs() { return argument_logs_; }
```
klucke, line 341: This use of argument_logs_ doesn't acquire the provided absl::Mutex.  Therefore it's not threadsafe.  Why does this function exist?

xla/stream_executor/gpu/gpu_executor.h
```diff
@@ -321,6 +321,26 @@ class GpuExecutor : public internal::StreamExecutorInterface {
...
+  absl::Mutex& LoggerMutex() { return logger_mu_; }
```
klucke, line 342: exposing internal Mutexes is error prone.  How do you do static analysis to ensure proper locking is always performed?

xla/stream_executor/gpu/gpu_executor.h
```diff
@@ -426,6 +446,12 @@ class GpuExecutor : public internal::StreamExecutorInterface {
   absl::flat_hash_map<void*, Stream*> alive_gpu_streams_
       ABSL_GUARDED_BY(alive_gpu_streams_mu_);
 
+  absl::Mutex logger_mu_;
+
+  mutable std::vector<ApiTrace> argument_logs_;
```
klucke, line 451: Annotate this member with ABSL_GUARDED_BY

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -390,6 +390,17 @@ absl::Status BlasLt::MatmulPlan::DoMatmul(
...
+  {
+    absl::MutexLock lock{&blas_lt_ref_.parent_->LoggerMutex()};
+    using sei = internal::StreamExecutorInterface;
+
+    if(blas_lt_ref_.parent_->GetArgumentLoggingMode() & sei::ArgumentLogging::kGemm) {
+      blas_lt_ref_.parent_->ArgumentLogs().push_back(
+        /* Known bug: blas::CallContext is not passed to this method */
+          sei::GemmCallTrace{sei::GemmCallTrace::GemmType::kBlasLt, 0,
+            a.size(), b.size()});
+    }
```
klucke, line 402: This should be replaced by a proper logging API on Stream executor rather than exposing details of the logging to the client.  What if we decided to replace the logging with a lockless implementation?  This also is hard to do static analysis on everywhere to prove the locking protocols are always correct.

xla/stream_executor/rocm/rocm_blas.h
```diff
@@ -199,6 +200,9 @@ class ROCMBlas : public blas::BlasSupport {
   // container holding solutions vector (to avoid reallocating it each time)
   std::vector<rocblas_int> solutions_;
 
+  void MaybeLogGemmOp(internal::StreamExecutorInterface::GemmCallTrace::GemmType op,
```
klucke, line 203: All functions should have a comment.

kamaljeeti:
Hi @klucke , could you please review the changes? Thanks

xla/stream_executor/stream_executor_interface.h
```diff
@@ -359,11 +359,51 @@ class StreamExecutorInterface {
...
+
+  // Sets the argument logging mode. Returns true if 'mode' is valid.
+  virtual bool SetArgumentLoggingMode(ArgumentLogging mode) { return false; }
+
  private:
   StreamExecutorInterface(const StreamExecutorInterface&) = delete;
   void operator=(const StreamExecutorInterface&) = delete;
 };
 
+inline bool operator&(StreamExecutorInterface::ArgumentLogging a,
```
klucke, line 402: This needs a comment for why it's necessary.  In general, it's problematic to bit-wise AND enum values together, as that creates undefined enum values, which is antithetical to making an enum class in the first place.

xla/stream_executor/stream_executor_interface.h
```diff
@@ -359,11 +359,51 @@ class StreamExecutorInterface {
...
+
+  // Retrieves and clears internal argument logs.
+  virtual absl::StatusOr<std::vector<ApiTrace> > ExtractApiTrace() {
+    return absl::UnimplementedError("Not implemented");
+  }
+  virtual absl::Status RecordApiTrace(ApiTrace call) {
+    return absl::UnimplementedError("Not implemented");
+  }
+
+  // Bit mask
```
klucke, line 388: Comments should be complete sentences, and add some "why" for the thing being described.  Please add some context for what this is used for, and why it's important for this to be a bit mask.

xla/stream_executor/stream_executor_interface.h
ekuznetsov139: What class/type should I use instead of enum? The intent is to allow logging to be enabled/disabled independently for individual submodules within StreamExecutor, with one flag per submodule.

klucke: If you want a bitmask, I think a uint64_t is the most appropriate thing, and probably a bunch of constexpr uint64_t for individual masks.  

I don't think it makes sense to use "enum class" for things that are bitmasks, because you can't really AND them together and produce valid enum class values unless you get super verbose (kFoo = 1, kBar = 2, kFooBar = 3, etc), which becomes basically unreadable IMO.

klucke (approved):

PR #9840, opened 2024-02-22
Fix typo in configure.py

mark14wu:
Fix missing lld_path argument in configure.py


ddunl (approved):
Thanks for catching this!

PR #9845, opened 2024-02-22
[GPU] Add flag to enable/disable per-stream NCCL communicators

trevor-m:
This PR adds a flag `--xla_gpu_enable_nccl_per_stream_comms` to allow us to disable per-stream communicators (it is enabled by default to retain current behavior).
Per-stream communicators can increase performance ~~if NCCL collectives are scheduled in parallel~~ by avoiding synchronization between streams. However, it costs a lot of GPU memory to create the extra communicators, and [using multiple communicators simultaneously can lead to hangs](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/communicators.html#using-multiple-nccl-communicators-concurrently).

Because of these drawbacks, we would like to add this flag to disable the feature in certain cases.

## Results with Pax GPT5B-PP on 2xH100 (NVLS enabled)

| enable_nccl_per_stream_comms | # Comms Created | Total Memory (MiB) | NCCL Memory (MiB) | Steps/sec |
| --- | --- |  --- |  --- |  --- |
| true (default, existing behavior) | 13 | 78366 | 21022 | 0.1254 |
| false (new with this PR) | 8 | 71830 | 14486 | 0.1285 |

In this case by disabling per-stream comms the memory used by NCCL decreased by 31% and the steps/sec actually increased by 2.47%.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -144,6 +144,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_shared_constants(true);
   opts.set_xla_gpu_enable_nccl_user_buffers(false);
   opts.set_xla_gpu_enable_nccl_comm_splitting(false);
+  opts.set_xla_gpu_enable_nccl_per_stream_comms(true);
 
   // Set 4GB space limit for redzone scratch allocator.
   opts.set_xla_gpu_redzone_scratch_max_megabytes(1LL << 12);
@@ -1192,6 +1193,14 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       debug_options->xla_gpu_enable_nccl_comm_splitting(),
       "Enables NCCL communicator splitting which allows sharing NCCL resources "
       "between different NCCL cliques."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_enable_nccl_per_stream_comms",
+      bool_setter_for(&DebugOptions::set_xla_gpu_enable_nccl_per_stream_comms),
+      debug_options->xla_gpu_enable_nccl_per_stream_comms(),
+      "A separate NCCL communicator will be created for each stream that a "
+      "NCCL collective is executed on. This can lead to higher performance if "
+      "NCCL collectives are issued concurrently at the cost of more GPU memory"
+      " usage."));
   flag_list->push_back(tsl::Flag(
       "xla_gpu_redzone_scratch_max_megabytes",
       int64_setter_for(
--- a/xla/service/gpu/nccl_collective_thunk.cc
+++ b/xla/service/gpu/nccl_collective_thunk.cc
@@ -68,6 +68,7 @@ namespace gpu {
 namespace {
 
 static constexpr int64_t kCollectiveMemorySpaceColor = 1;
+static constexpr int64_t kNoStreamId = 0;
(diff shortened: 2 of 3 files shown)


jurahul (commented):
From what I recall, per stream comms were added not to allow concurrent execution of collectives but because we cache and reuse the same communicator across multiple collectives. So if we have a sync and an async collective in the code (scheduled non-concurrently) and they use the same comm but are on 2 different streams, NCCL would add an implicit synchronization between the 2 streams and that regressed perf. Per stream comms try to work around that by not reusing the same comm on 2 different streams.

I am ok with this change, but just wanted to correct the reasoning, in the sense it affects back-to-back scheduled collectives as well as concurrently scheduled ones.

jurahul:
> @jurahul Thank you for the correction!
> 
> Could you help me understand when the sync would be added without per stream communicators? For example with this timeline, would the sync be added at step 3?
> 
> 1. sync collective on stream 0

My understanding is that here, where the same comm is used first on stream 0 and then on stream 1. XLA will add a 0 -> 1 sync, and I think NCCL detects internally that the same comm is used on s0 and s1 and hence adds another sync from 0->1 as a part of async collective start on stream1. Atleast that's my understanding.

> 2. async collective start on stream 1
> 3. compute ops on stream 0
> 4. async collective done on stream 1(wait for collective to finish)
> 5. sync collective on stream 0

trevor-m:
> > @jurahul Thank you for the correction!
> > Could you help me understand when the sync would be added without per stream communicators? For example with this timeline, would the sync be added at step 3?
> > 
> > 1. sync collective on stream 0
> 
> My understanding is that here, where the same comm is used first on stream 0 and then on stream 1. XLA will add a 0 -> 1 sync, and I think NCCL detects internally that the same comm is used on s0 and s1 and hence adds another sync from 0->1 as a part of async collective start on stream1. Atleast that's my understanding.
> 
> > 2. async collective start on stream 1
> > 3. compute ops on stream 0
> > 4. async collective done on stream 1(wait for collective to finish)
> > 5. sync collective on stream 0

Ah ok, that makes sense. When the same comm is used NCCL will add syncs and NCCL doesn't know that XLA is already syncing here.

jurahul:
> > > @jurahul Thank you for the correction!
> > > Could you help me understand when the sync would be added without per stream communicators? For example with this timeline, would the sync be added at step 3?
> > > 
> > > 1. sync collective on stream 0
> > 
> > 
> > My understanding is that here, where the same comm is used first on stream 0 and then on stream 1. XLA will add a 0 -> 1 sync, and I think NCCL detects internally that the same comm is used on s0 and s1 and hence adds another sync from 0->1 as a part of async collective start on stream1. Atleast that's my understanding.
> > > 2. async collective start on stream 1
> > > 3. compute ops on stream 0
> > > 4. async collective done on stream 1(wait for collective to finish)
> > > 5. sync collective on stream 0
> 
> Ah ok, that makes sense. When the same comm is used NCCL will add syncs and NCCL doesn't know that XLA is already syncing here.

The per-stream comm was an optimization to avoid this NCCL internal sync, and IIRC it showed ~5% improvement on some LLM inference workloads. But I understand the need to be memory footprint aware, so the flag makes sense. Infact, this concern was raised at that time by DeepMind team as well, but we decided to add a flag if the memory issue actually cropped up in a real use case. Approving the PR.

jurahul (approved):

PR #9856, opened 2024-02-22
Draft: [XLA] Control printing of a large GPU backend config field with a flag.

sergachev:
In the GPU backend cuDNN graphs can be attached in serialized form to HLO fusion instructions. Printing them by default is not informative and not necessary. In this regard they are similar to large constants, therefore the corresponding flag is reused.

The removal is implemented through protobuf reflection to avoid the dependency of the IR class on the backend-specific GPU proto.

### DIFF
--- a/xla/hlo/ir/hlo_instruction.cc
+++ b/xla/hlo/ir/hlo_instruction.cc
@@ -3429,6 +3429,42 @@ bool HloInstruction::IsCrossReplicaAllReduce() const {
   return false;
 }
 
+std::optional<std::string>
+HloInstruction::BackendConfigRep::GetFilteredRawString() const {
+  const tsl::protobuf::Message* msg = GetProtoPtr();
+  if (msg != nullptr &&
+      msg->GetDescriptor()->full_name() == "xla.gpu.GpuBackendConfig") {
+    const tsl::protobuf::FieldDescriptor* fusion_descriptor =
+        msg->GetDescriptor()->FindFieldByName("fusion_backend_config");
+    if (fusion_descriptor != nullptr &&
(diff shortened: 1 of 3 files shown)


thomasjoerg:
Hi Ilia, adding cuDNN-specific code to hlo_instruction.cc has a smell, in my opinion.

Can you tell more about the motivation for this change? You say the serialized graph is not informative. How would it look like? Is it human-readable at all? In which (debuggin) cases would we want to have the serialized cuDNN graph?

sergachev:
Hi Thomas,

> Is it human-readable at all?
Mostly not.

> In which (debuggin) cases would we want to have the serialized cuDNN graph?
Having it printed in logs - I'd say never. Having it dumped in files - maybe if we wanted to execute kernels compiled by cuDNN without compiling again. By default it's better to not have them even in the dump files.

Do you have better ideas how to have this field not printed by any hlo.ToString() call by default?

cheshire:
How big are those serialized graphs usually?

cheshire:
Thomas is right, hlo_instruction.cc/h should not have cuDNN-specific code.

cheshire:
> Mostly not.

Maybe this should be fixed then? IR should be hackable/human-readable.

sergachev:
> Thomas is right, hlo_instruction.cc/h should not have cuDNN-specific code.

It's actually not cuDNN-specific. It just removes field called
fusion.backend_config.fusion_backend_config.cudnn_fusion_config.serialized_graph. We can adjust it to be fusion.backend_config.fusion_backend_config.dnn_fusion_config.serialized_graph similar to the stream executor's class hierarchy.

sergachev:
Closing if favor of https://github.com/openxla/xla/pull/10165 after discussions.

PR #10037, opened 2024-02-27
[NVIDIA XLA] Copy backend config for derived instruction that has same opcode

Tixxx:
SetupDerivedInstruction copies many properties like frontend_attributes, metadata and etc, but it doesnt transfer over backend configs. If the derived instruction has the same opcode as the current one, the backend config is most likely applicable, this change copies over the backend config in that situation.

### DIFF
--- a/xla/hlo/ir/hlo_instruction.cc
+++ b/xla/hlo/ir/hlo_instruction.cc
@@ -2057,6 +2057,12 @@ void HloInstruction::SetupDerivedInstruction(
     derived_instruction->mutable_rare()->frontend_attributes.Clear();
     derived_instruction->mutable_rare()->statistics_viz.Clear();
   }
+  // If the derived instruction has the same opcode as current,
+  // then the backend config is also applicable.
+  if (opcode() == derived_instruction->opcode() && has_backend_config()) {
+    derived_instruction->set_raw_backend_config_string(
+        raw_backend_config_string());
+  }
 }
 
 bool HloInstruction::IsRoot() const {
--- a/xla/service/hlo_instruction_test.cc
+++ b/xla/service/hlo_instruction_test.cc
@@ -2805,5 +2805,41 @@ TEST_F(HloInstructionTest,
   EXPECT_EQ(num_conditional_branch_comp, branch_computations.size());
 }
 
+TEST_F(HloInstructionTest, BackendConfigCopiedToDerived) {
+  HloComputation::Builder b(TestName());
+  Shape shape = ShapeUtil::MakeShape(F32, {2, 2});
+  auto p0 = b.AddInstruction(HloInstruction::CreateParameter(0, shape, "p0"));
+  auto p1 = b.AddInstruction(HloInstruction::CreateParameter(0, shape, "p1"));
+  auto add = b.AddInstruction(
+      HloInstruction::CreateBinary(shape, HloOpcode::kAdd, p0, p0));
+
+  gpu::GpuBackendConfig gpu_config;
+  gpu_config.set_operation_queue_id(2);
+  TF_ASSERT_OK(add->set_backend_config(gpu_config));
+  auto add2 = b.AddInstruction(
+      HloInstruction::CreateBinary(shape, HloOpcode::kAdd, p0, p0));
+  add->SetupDerivedInstruction(add2);
+  auto backend_config = add2->backend_config<gpu::GpuBackendConfig>();
+  EXPECT_TRUE(backend_config.ok());
+  EXPECT_EQ(backend_config->operation_queue_id(), 2);
+}
+
+TEST_F(HloInstructionTest, BackendConfigNotCopiedToDerivedWithDiffOpcode) {
+  HloComputation::Builder b(TestName());
+  Shape shape = ShapeUtil::MakeShape(F32, {2, 2});
+  auto p0 = b.AddInstruction(HloInstruction::CreateParameter(0, shape, "p0"));
+  auto p1 = b.AddInstruction(HloInstruction::CreateParameter(0, shape, "p1"));
+  auto or1 = b.AddInstruction(
+      HloInstruction::CreateBinary(shape, HloOpcode::kOr, p0, p0));
+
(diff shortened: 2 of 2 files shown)


cheshire (approved):

dimitar-asenov:
We had to roll this back because it was causing Segmentation Faults. Here is an example stack trace:


```
SIGSEGV
    @     0x5647474bb1a7        928  FailureSignalHandler()
...
    @     0x564745477fa7         48  std::__u::default_delete<>::operator()()
    @     0x564745497223         80  mlir::MLIRContextImpl::~MLIRContextImpl()
    @     0x5647454923d6         32  mlir::MLIRContext::~MLIRContext()
    @     0x564734a587dd       1472  xla::gpu::CompileModuleToLlvmIr()
    @     0x564734a44299       1952  xla::gpu::GpuCompiler::CompileToBackendResult()
    @     0x564734a465d6       2976  xla::gpu::GpuCompiler::RunBackend()
    @     0x56473b45ede6        576  xla::gpu::AutotunerCompileUtil::Compile()
    @     0x56473b458107        192  xla::gpu::(anonymous namespace)::CompileMany()::$_0::operator()()
...
```

At this point we don't have a solid reproducer - working on it.

One speculation given the stack trace is that we might be calling double `free()`. Looking at the original change, the only way this might happen is if the backend string encodes some pointers which are freed. Do you know if this is the case?

sergeykozub (approved):

Tixxx:
> std::__u::default_delete<>::operator()()

Another instance is Triton kernels, they have IR embedded in backend config strings.

Tixxx:
> > We had to roll this back because it was causing Segmentation Faults. Here is an example stack trace:
> > ```
> > SIGSEGV
> >     @     0x5647474bb1a7        928  FailureSignalHandler()
> > ...
> >     @     0x564745477fa7         48  std::__u::default_delete<>::operator()()
> >     @     0x564745497223         80  mlir::MLIRContextImpl::~MLIRContextImpl()
> >     @     0x5647454923d6         32  mlir::MLIRContext::~MLIRContext()
> >     @     0x564734a587dd       1472  xla::gpu::CompileModuleToLlvmIr()
> >     @     0x564734a44299       1952  xla::gpu::GpuCompiler::CompileToBackendResult()
> >     @     0x564734a465d6       2976  xla::gpu::GpuCompiler::RunBackend()
> >     @     0x56473b45ede6        576  xla::gpu::AutotunerCompileUtil::Compile()
> >     @     0x56473b458107        192  xla::gpu::(anonymous namespace)::CompileMany()::$_0::operator()()
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > At this point we don't have a solid reproducer - working on it.
> > One speculation given the stack trace is that we might be calling double `free()`. Looking at the original change, the only way this might happen is if the backend string encodes some pointers which are freed. Do you know if this is the case?
> 
> That's interesting, most of the fields allowed in the backend config are primitive boolean or numerical types and protos. One exception might be `CuDnnFusionConfig` which has a string that contains a serialized graph. I'm not sure why that would cause a double free in the compilation stage though. I think an HLO reproducer would be helpful.

Another instance is Triton kernels, they have IR embedded in backend config strings.

sanjoy:
@dimitar-asenov Do you get a different error if you run the internal reproducer with asan & msan?

dimitar-asenov:
I'm now able to reproduce this, but unfortunately I can't run ASAN or MSAN, so debugging will be manual.

dimitar-asenov:
This might also be a threading issue. If I launch the compilation with low values of  `--xla_gpu_force_compilation_parallelism` (e.g. 1 or 2), then the issue becomes much less likely.

@beckerhe suspects some thread unsafe code in `HloInstruction`. We're still looking into this.

Tixxx:
> This might also be a threading issue. If I launch the compilation with low values of `--xla_gpu_force_compilation_parallelism` (e.g. 1 or 2), then the issue becomes much less likely.
> 
> @beckerhe suspects some thread unsafe code in `HloInstruction`. We're still looking into this.

Thanks for digging into this. I see that he raw backend string is lazily initialized somewhere in HloInstruction, would using `CopyBackendConfigFrom` instead of `set_raw_backend_config_string` help with the seg fault?

dimitar-asenov:
> Thanks for digging into this. I see that he raw backend string is lazily initialized somewhere in HloInstruction, would using `CopyBackendConfigFrom` instead of `set_raw_backend_config_string` help with the seg fault?

Thank you, TJ. Indeed that appears to resolve the issue. I will merge the PR with the fix.

Tixxx:
> We merged this again. Hopefully there won't be any more issues this time. Thank you for your patience.

good to know that fixed the issue. Thanks for the help!
Do you need me to resolve the conflicts or is it already merged internally?

dimitar-asenov:
This was already merged:

https://github.com/openxla/xla/blob/main/xla/hlo/ir/hlo_instruction.cc#L2102-L2106

Closing the PR.

sanjoy:
@dimitar-asenov Is it possible to add a test in OSS that would have caught this issue?

dimitar-asenov:
@sanjoy Hey, sorry for the long delay on this. I will look into creating a test. The reproducers we used internally are fairly large and not suitable, so I will need some time to isolate the issue. I will keep you updated here.

sanjoy:
Hey @dimitar-asenov - any luck on getting an HLO test case?

dimitar-asenov:
Hi. No progress yet. I could not work on this in the last two weeks (OOO + other priorities).

dimitar-asenov:
Hi @sanjoy . I just submitted https://github.com/openxla/xla/commit/7e3a51b5437eab49880168de8e9f38c933445e50

It refactors the backend config, fixes some issues, and adds tests. This should prevent issues like the one we encountered while merging this PR. There's still a few sharp edges and we're looking into a better solution long-term.

Sorry for the delay on this.

sanjoy:
Thanks @dimitar-asenov !!

PR #10164, opened 2024-02-29
[XLA:CPU] Improve Trace emitter

sachinmuradi:
This PR improves tracing a bit.

Without this PR, you would see operations in trace view as :
image

This way , it s difficult to see a) target of custom call b) can't group operations based on type of hlo operation and hence makes op list huge

This PR addresses these problems and now you should see tensorboard trace view as : 
image


ddunl:
LGTM but interested in @penpornk 's review. Thanks for the PR!!

kamaljeeti:
Hi @penpornk , can you review this PR? Thanks.

xla/service/cpu/ir_emitter.cc
```diff
@@ -3373,7 +3373,17 @@ void IrEmitter::TracingState::EmitTracingStart(llvm::IRBuilder<>* b,
...
   }
-  auto* hlo_name = b->CreateGlobalStringPtr(hlo->name());
+
+  // Pass opcode as argument to TraceMe call
+  auto hlo_name_str = hlo->name();
+  if (hlo->opcode() == HloOpcode::kCustomCall) {
+    // For custom call, passing custom call target is more informative
+    hlo_name_str = hlo->custom_call_target();
+  } else {
+    hlo_name_str = HloOpcodeString(hlo->opcode());
```
penpornk, line 3383: I'm not sure about this part since it would mean we can't tell the difference between each of the same HLO instructions anymore, e.g., dot.123 vs dot.42, etc. Maybe it would be better to improve the trace summarizer instead?

xla/service/cpu/ir_emitter.cc
```diff
@@ -3373,7 +3373,17 @@ void IrEmitter::TracingState::EmitTracingStart(llvm::IRBuilder<>* b,
     fn->setDoesNotThrow();
     fn->setOnlyAccessesArgMemory();
   }
-  auto* hlo_name = b->CreateGlobalStringPtr(hlo->name());
+
+  // Pass opcode as argument to TraceMe call
+  auto hlo_name_str = hlo->name();
+  if (hlo->opcode() == HloOpcode::kCustomCall) {
+    // For custom call, passing custom call target is more informative
+    hlo_name_str = hlo->custom_call_target();
```
penpornk, line 3381: For custom call, it's probably still good to know the custom call id. Having `custom-call.<id>` as prefix should still be helpful, e.g., `custom-call.46.__onednn$matmul`.

penpornk (changes requested):
Thank you for the PR and sorry for the delay! This fell through the crack.

sachinmuradi:
@penpornk Thank you for review, I modified the code a bit , to encode more information to TraceMe API, hope this addresses your comments, let me know your thoughts on this.

I am trying to avoid the long list generated at ops summary table in trace view, which sometimes is very slow to scroll if we have many hlo ops. It may be better if we just show the hlo opcodes in table and show which type of op is spending more time in execution, this is similar to what I have seen when using Tensorflow - Tensorboard trace view.

As to distinguish the hlo op name, I encoded the hlo name and some metadata information as well, that way user will be able to point the exact hlo op name for that op type. Here is an example screenshot : 
 
**1) When clicked on reduce-window op in trace (TF-xla example) :** 

image

**2) When clicked on onednn matmul (TF-XLA example) :**

image

**3) Additional example of onednn softmax with JAX-XLA :** 

image

penpornk (approved):
This is great! Thank you very much!

PR #6595, opened 2023-10-27
Fix multi-GPU cublasLt crash

wenscarl:
The issue was that each device shared a algorithm_ in CublasLtMatmulThunk.  To fix, now each stream has a different algorithm cache. Fix #6370 cc @kaixih @philipphack

PR #10182, opened 2024-03-01
[ROCM] fixing hipblas-lt large startup overhead

pemeliya:
This PR https://github.com/openxla/xla/pull/6595 introduced **matmul_algorithm_cache_** which calls MatmulPlan::GetAlgorithms while holding a mutex. This is not bad on CUDA side but on ROCM, hipblas-lt has a large startup overhead, and when running this code with multiple GPUs, this startup overhead increases linearly with the number of GPUs.
However, **MatmulPlan::GetAlgorithms** can be run in parallel across different StreamExecutorsL anyway library functions are called under [per-instance mutex](https://github.com/ROCm/xla/blob/43c67a4d135b5e1e023c3e67b9062b9aa88d6ab8/xla/stream_executor/cuda/cuda_blas_lt.cc#L244).

I also did the same for **matmul_plans_cache_**, besides I added a multi-GPU matmul test which runs with cublas and cublas-lt libraries.


i-chaochen:
@xla-rotation gentle ping

sergeykozub (approved):

kamaljeeti:
Hi @ddunl , can you have a look for the checks which are failing above? Thanks.

ddunl (approved):

pemeliya:
I wonder if there are any internal issues with this PR ? ARM64 CPU build brake does not seem to be caused by these changes.
It is already taken by copybara service but not merged yet..

PR #10206, opened 2024-03-01
[XLA:GPU] Fix xla_gpu_dump_autotuned_triton_fusions flag interaction with fallback to cuBLAS.

sergachev:
Currently using the flag leads to crashes on checks in TritonGemmConfig::FromProto() when any GEMM fusion falls back to cuBLAS.


tdanyluk (approved):
Thanks for fixing this, haven't used that flag for a while.

sergachev:
Rebased.

PR #9778, opened 2024-02-21
Added flash attention with new cudnn frontend

shraiysh:
This is the second attempt at #9528.

PR #10232, opened 2024-03-04
[XLA:GPU] add cuDNN FMHA variable sequence length support in thunk lowering

Cjkkkk:
* cuDNN accepts two tensors seqlen_q and seqlen_k with size [batch, 1] which describes the number of non-padded token in each batch. This allows faster FMHA computation and padding mask generation inside cuDNN if needed.
* Currently not accessible through FMHA rewriter, accessible through JAX SDPA API.

### DIFF
--- a/xla/service/gpu/ir_emitter_unnested.cc
+++ b/xla/service/gpu/ir_emitter_unnested.cc
@@ -969,12 +969,14 @@ absl::Status IrEmitterUnnested::EmitFusedMHAThunk(
       bias_shape = bias->shape();
     }
 
-    bool has_seqlen_qk = 2 == (instr->operand_count() - 3 - has_mask - has_bias);
+    bool has_seqlen_qk =
+        2 == (instr->operand_count() - 3 - has_mask - has_bias);
     int64_t seqlen_qk_operand_index = 2 + has_mask + has_bias;
     if (has_seqlen_qk) {
       const HloInstruction* seqlen_q = instr->operand(seqlen_qk_operand_index);
       TF_ASSIGN_OR_RETURN(seqlen_q_slice, GetAllocationSliceForHlo(seqlen_q));
-      const HloInstruction* seqlen_k = instr->operand(seqlen_qk_operand_index + 1);
+      const HloInstruction* seqlen_k =
+          instr->operand(seqlen_qk_operand_index + 1);
       TF_ASSIGN_OR_RETURN(seqlen_k_slice, GetAllocationSliceForHlo(seqlen_k));
     }
   }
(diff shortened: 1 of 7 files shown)


xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -967,6 +968,17 @@ absl::Status IrEmitterUnnested::EmitFusedMHAThunk(
       TF_ASSIGN_OR_RETURN(bias_slice, GetAllocationSliceForHlo(bias));
       bias_shape = bias->shape();
     }
+
+    bool has_seqlen_qk =
```
akuegel, line 972: Maybe rewrite `has_seqlen_qk` as:

```
bool has_seqlen_qk = seqlen_qk_operand_index == instr->operand_count() - 3;
```

I think that should be equivalent, because:

```
2 == (instr->operand_count() - 3 - has_mask - has_bias); <-> 2 + has_mask + has_bias == intr->operand_count() - 3
```

Given this rewrite, I am wondering what the last operand of `instr` is? It seems `seqlen_qk_operand_index` is for q, `selen_qk_operand_index + 1` is for k.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -1079,6 +1092,15 @@ absl::Status IrEmitterUnnested::EmitFusedMHABackwardThunk(
     fwd_output_shape = instr->operand(input_index++)->shape();
   }
 
+  BufferAllocation::Slice seqlen_q_slice, seqlen_k_slice;
+  bool has_seqlen_qk = 2 == (instr->operand_count() - input_index);
```
akuegel, line 1096: Same as above, please rewrite to make it easier to understand.

xla/service/gpu/ir_emitter_unnested.cc
akuegel: Also, in case this is a bug, and it should in fact be checking

```
has_seqlen_qk = seqlen_qk_operand_index == instr->operand_count() - 2
```

then, add a test that exposes this bug.

Cjkkkk: Fixed. operand order should be Q, K, V, mask*, bias*, seqlen_q*, seqlen_k* with last 4 being optional. Not sure how to test this without cudnn graph generation(will be in a separate PR). The new cuDNN front-end PR is still pending https://github.com/openxla/xla/pull/9778.

akuegel: Thanks for the fix, and I trust that you will add a test once it is possible.

akuegel (approved):

xla/service/gpu/ir_emitter_unnested.cc
Cjkkkk: Will do. Thanks for the quick approval.

PR #10291, opened 2024-03-06
[ROCM] fixing pmap failure on multiple gpus (forward convolution)

### LINKED ISSUE #14582: On ROCm, MIOpen can't find invoker for convolution forward when using pmap with 8 devices, but works fine with 1 device
### Description

Hi!

I'm working with AMD GPUs in a cluster environment.

When using pmap with >1 GPUs MIOpen can't seem to find the correct invoker for the convolution forward and raises the following error:

```
MIOpen Error: /long_pathname_so_that_rpms_can_package_the_debug_info/data/driver/MLOpen/src/ocl/convolutionocl.cpp:658: No invoker was registered for convolution forward. Was find executed?
```

But it runs just fine with a single GPU. This can also be reproduced with the following code snippet.

```python
from flax import jax_utils
import jax
import flax.linen as nn
import jax.numpy as jnp
rng = jax.random.PRNGKey(0)
conv = nn.Conv(1, kernel_size=(7,7), strides=(2,2), padding="VALID")
var = conv.init(rng, jnp.ones((1, 14, 14, 1)))
x = jax.random.normal(rng, (16, 14, 14, 1))
def forward(batch):
    return conv.apply(var, batch)
### works
o = forward(x).shape

pforward = jax.pmap(forward, axis_name="batch")
x = jax.random.normal(rng, (8, 16, 14, 14, 1))

### doesn't work
print(pforward(x).shape)
```

I've attached full log files from test runs with 1 and 8 devices, 
[logfile_1gpu.txt](https://github.com/google/jax/files/10778977/logfile_1gpu.txt)
[logfile_8gpus.txt](https://github.com/google/jax/files/10778978/logfile_8gpus.txt)

which were collected as follows 
```bash
export MIOPEN_ENABLE_LOGGING=1
export MIOPEN_LOG_LEVEL=5
export MIOPEN_FIND_MODE=3

python jax_simple.py 2>&1 | tee logfile_8gpus.txt
CUDA_VISIBLE_DEVICES=0 python test.py 2>&1 | tee logfile_1gpu.txt
```

If it helps, official `jax.pmap` examples run just fine. Even flax mnist and imagenet examples train just fine, as long as 1 gpu is used. 

Thanks!

### What jax/jaxlib version are you using?

jax v0.4.3, jaxlib v0.4.3

### Which accelerator(s) are you using?

GPU (AMD MI250x)

### Additional system info

Python 3.9, slurm environment, ROCm 5.3.3

### NVIDIA GPU info

_No response_

pemeliya:
In this PR I fix [this issue](https://github.com/google/jax/issues/14582) for AMD GPUs.

The problem was that, in order to use MIOpen convolutions, previously one needs to call **GetMIOpenConvolveAlgorithms** to populate an intenal MIOpen cache of convolution invokers. So far this was implicitly done by conv_algorithm_picker.cc for **one stream executor** (single GPU). But when one tries to run it on multiple GPU devices, the above problem occurs.\

Note that, the fix applies to AMD GPUs only, hence no changes in behaviuor on CUDA side.

@xla-rotation: could you please take a look ?

### DIFF
--- a/xla/service/gpu/runtime/convolution_thunk.cc
+++ b/xla/service/gpu/runtime/convolution_thunk.cc
@@ -22,7 +22,9 @@ limitations under the License.
 #include "absl/status/status.h"
 #include "xla/service/buffer_assignment.h"
 #include "xla/service/gpu/gpu_conv_runner.h"
+#include "xla/service/gpu/stream_executor_util.h"
 #include "xla/stream_executor/stream_executor.h"
+#include "xla/stream_executor/scratch_allocator.h"
 #include "xla/util.h"
 
 namespace xla {
@@ -40,10 +42,11 @@ ConvolutionThunk::ConvolutionThunk(
       config_(std::move(config)) {}
 
 GenericConvRunner& ConvolutionThunk::GetOrCreateRunner(
-    const stream_executor::Stream* stream) {
+    const stream_executor::Stream* stream, bool *runner_created) {
   absl::MutexLock lock(&mu_);
   auto it = runner_cache_.find(stream);
-  if (it == runner_cache_.end()) {
+  *runner_created = (it == runner_cache_.end());
+  if (*runner_created) {
     it = runner_cache_
              .insert({stream, std::make_unique<GenericConvRunner>(config_)})
              .first;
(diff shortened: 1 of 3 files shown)


i-chaochen:
gentle ping @xla-rotation

ezhulenev (approved):

xla/service/gpu/runtime/convolution_thunk.cc
```diff
@@ -22,7 +22,9 @@ limitations under the License.
 #include "absl/status/status.h"
 #include "xla/service/buffer_assignment.h"
 #include "xla/service/gpu/gpu_conv_runner.h"
+#include "xla/service/gpu/stream_executor_util.h"
```
akuegel, line 25: This requires a corresponding BUILD file dependency that has this header in its hdrs section.

pemeliya: Sorry for the delay, I have fixed it

ddunl (approved):

PR #10297, opened 2024-03-06
[NVIDIA GPU] Unroll all nested loops for double buffering transformer

Tixxx:
The loop_double_buffer_transformer pass currently only unrolls the loops in the main computation.
There are some cases where the memcpies(which this pass is designed to eliminate) would exist in nested loops.
This pr changes the pass to look at all loops in the module include the nested ones and unroll them.

### DIFF
--- a/xla/service/gpu/loop_double_buffer_transformer_test.cc
+++ b/xla/service/gpu/loop_double_buffer_transformer_test.cc
@@ -470,6 +470,60 @@ ENTRY main {
   // associated computations.
   EXPECT_EQ(while_loops_callees.size(), 8);
 }
+
+TEST_F(GpuLoopDoubleBufferTransformerTest, NestedWhileLoopAreUnrolled) {
+  const char* const kModuleString = R"(
+HloModule loop_unrolling_nested_are_unrolled
+condition_nested {
+  input_tuple = (s32[]) parameter(0)
+  cond = s32[] get-tuple-element(input_tuple), index=0
+  trip_count = s32[] constant(10)
+  ROOT done = pred[] compare(cond, trip_count), direction=LT
+}
+body_nested {
+ input_tuple = (s32[]) parameter(0)
+ cond = s32[] get-tuple-element(input_tuple), index=0
+ one = s32[] constant(1)
+ cond_plus_1 = s32[] add(cond, one)
+ ROOT output = (s32[]) tuple(cond_plus_1)
+}
+condition {
+  input_tuple = (s32[]) parameter(0)
+  cond = s32[] get-tuple-element(input_tuple), index=0
+  trip_count = s32[] constant(10)
+  ROOT done = pred[] compare(cond, trip_count), direction=LT
+}
+body {
+  input_tuple = (s32[]) parameter(0)
+  ROOT output = (s32[]) while(input_tuple), condition=condition_nested, body=body_nested, backend_config={"known_trip_count":{"n":"11"}}
+}
+ENTRY main {
+ param_0 = (s32[]) parameter(0)
+ ROOT while = (s32[]) while(param_0), condition=condition, body=body, backend_config={"known_trip_count":{"n":"11"}}
+})";
+
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<xla::HloModule> module,
+                          ParseAndReturnVerifiedModule(kModuleString));
+  LoopDoubleBufferTransformer double_buffer;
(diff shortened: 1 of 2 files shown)


xla/service/gpu/loop_double_buffer_transformer_test.cc
```diff
@@ -470,6 +470,57 @@ ENTRY main {
   // associated computations.
   EXPECT_EQ(while_loops_callees.size(), 8);
 }
+
+TEST_F(GpuLoopDoubleBufferTransformerTest,
```
bchetioui, line 474: I suspect that the two tests above (`NestedWhileLoopRemainsFlattened*`) will start failing now. Did you check them?

cheshire (approved):

xla/service/gpu/loop_double_buffer_transformer_test.cc
```diff
@@ -470,6 +470,57 @@ ENTRY main {
...
+})";
+
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<xla::HloModule> module,
+                          ParseAndReturnVerifiedModule(kModuleString));
+  LoopDoubleBufferTransformer double_buffer;
+  EXPECT_THAT(double_buffer.Run(module.get()), IsOkAndHolds(true));
+
+  for (const HloComputation* computation : module->computations()) {
+    for (const HloInstruction* instr : computation->instructions()) {
+      if (instr->opcode() == HloOpcode::kWhile) {
```
bchetioui, line 513: Just to be sure, should we check that we still have two while loops being hit here? (I don't expect a problem here, just figure this is an additional expectation we have that we could document---fully removing the nested loop would also pass this test.)

bchetioui (changes requested):
Thanks @Tixxx! Can I ask you to provide some more context around this change?

~I don't expect this to be particularly controversial, I'd just want to understand what this makes better (I assume we're doing this for a reason). What are the new cases we are encountering, and what are the gains? Could we also document that in the commit message?~ (EDIT: sorry, I see now you already explained, this is enough for me.)

Another question I have is what are the long term plans regarding loop double buffering. Do we expect this to remain gated behind a flag, or do we have another plan? I suspect we do not like this situation of having to turn it on manually.

xla/service/gpu/loop_double_buffer_transformer_test.cc
```diff
@@ -470,6 +470,57 @@ ENTRY main {
   // associated computations.
   EXPECT_EQ(while_loops_callees.size(), 8);
 }
+
+TEST_F(GpuLoopDoubleBufferTransformerTest,
+       NestedWhileLoopRemainsFlattenedOddTripCountNestedUnrolled) {
```
bchetioui, line 475: Name is a bit confusing. Maybe we could rename it to something like `NestedWhileLoopIsAlsoUnrolled`?

xla/service/gpu/loop_double_buffer_transformer_test.cc
Tixxx: I think they should be okay. Unrolling the nested loops would only change the trip count, the number of condition and body callsites should not change?

xla/service/gpu/loop_double_buffer_transformer_test.cc
Tixxx: renamed

Tixxx:
> Thanks @Tixxx! Can I ask you to provide some more context around this change?
> 
> ~I don't expect this to be particularly controversial, I'd just want to understand what this makes better (I assume we're doing this for a reason). What are the new cases we are encountering, and what are the gains? Could we also document that in the commit message?~ (EDIT: sorry, I see now you already explained, this is enough for me.)
> 
> Another question I have is what are the long term plans regarding loop double buffering. Do we expect this to remain gated behind a flag, or do we have another plan? I suspect we do not like this situation of having to turn it on manually.

I think there's a separate workstream to enable some of these flags by default or get rid of them completely. We have turned this on by default in our jax toolbox containers that we ship to customers and there hasn't been any issue, so it's safe to say that this could be on by default. To save a bit on compilation time, we can also check to see if there are any operands of loop output tuple that are copies, then we only need to unroll loops when necessary.

xla/service/gpu/loop_double_buffer_transformer_test.cc
bchetioui: Hah, you're right, nevermind. I think the name of the tests above tripped me up---maybe I should change it.

bchetioui (approved):

PR #7854, opened 2023-12-18
[NVIDIA SPMD] Add HLO support to run windowed einsum in multiple streams

Tixxx:
This PR contains the HLO changes to be able to run windowed einsum in multiple cuda streams.
This adds a debug option to use multi-streaming for einsum, it will unroll the loop when turned on and mark the corresponding dots with different stream ids.
With the unrolled windowed einsum loop and overlapping multiple gemms, we get 2-3% speedup on the gpt-3 175b models.
There are some improvements to be made, major ones are:
1. redundant memcpy for each einsum loop, this can be mitigated by improving loop_double_buffer_transformer to unroll all nested while loops([pr](https://github.com/openxla/xla/pull/10297)), potential speedup for this is ~4% for 175b.
2. Some collective-permutes are scheduled after the async gemm, they won't be able to obtain enough SMs to run and are therefore partially exposed, this can be fixed by forcing them to be delayed in the latency hiding scheduler pass([pr](https://github.com/openxla/xla/pull/10316)), potential speedup with this along is 4% for 175b.

I'm working on both the optimizations above in separate PRs.

Detailed discussion [here](https://github.com/openxla/xla/discussions/8865)

PR #10316, opened 2024-03-07
[NVIDIA GPU] Add a mechanism in GpuAsyncTracker to force delaying scheduling of an instruction.

Tixxx:
latency hiding scheduler has a property for each HloScheduleNode called SetForceDelay, setting this to true for a node will result in forcing LHS to ignore all cost info and schedule the instruction to the very beginning of the instruction sequence.
This is useful in manually enforcing an instruction to run before anything else in the scheduled graph.
Passes can set `should_force_delay` attribute in the GpuBackendConfig to instruct how this instruction will be scheduled. It's set to false by default.
An example usage:
```
original sequence:

allreduce-start
allreduce-done
add0, backend_config={should_force_delay=true}
add1

schedule after LHS:
add0, backend_config={should_force_delay=true}
allreduce-start
add1
allreduce-done
```

### DIFF
--- a/xla/service/gpu/backend_configs.proto
+++ b/xla/service/gpu/backend_configs.proto
@@ -261,9 +261,15 @@ message GpuBackendConfig {
     CudnnfMHABackendConfig cudnn_fmha_backend_config = 9;
   }
 
-  // This attribute instructs the latency-hiding scheduler to delay
-  // scheduling this particular instruction. Note that setting this to
-  // true will make this instruction scheduled at the very beginning of
-  // the parent computation before every other nodes.
-  bool should_force_delay_schedule = 10;
+  // This attribute instructs the latency-hiding scheduler to
+  // schedule this particular instruction to the earliest position.
+  // Note that setting this to true will make this instruction scheduled
+  // at the very beginning of the parent computation before
+  // every other nodes.
+  // An example use case would be deciding to schedule between collective
+  // or an async compute. LHS might put either one at the first place
+  // depending on the cost, but it'd be more beneficial if the collective
+  // is always scheduled first as it's not SM-heavy.
+  // In this case we can use this flag to enforce the ordering.
+  bool force_earliest_schedule = 10;
 }
--- a/xla/service/gpu/backend_configs.proto
+++ b/xla/service/gpu/backend_configs.proto
@@ -260,4 +260,10 @@ message GpuBackendConfig {
 
     CudnnfMHABackendConfig cudnn_fmha_backend_config = 9;
   }
+
+  // This attribute instructs the latency-hiding scheduler to delay
+  // scheduling this particular instruction. Note that setting this to
+  // true will make this instruction scheduled at the very beginning of
+  // the parent computation before every other nodes.
+  bool should_force_delay_schedule = 10;
 }
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -399,6 +399,22 @@ class GpuAsyncTrackerBase : public AsyncTracker {
             !IsSyncCollective(&hlo)) ||
            IsAsyncComputeOp(hlo);
   }
+
+  void PostProcessScheduleGraph(
+      HloScheduleGraph* schedule_graph,
+      const LatencyEstimator* latency_estimator) const override {
+    for (auto inst : schedule_graph->GetOriginalInstrList()) {
+      if (inst->has_backend_config()) {
+        auto gpu_config = inst->backend_config<GpuBackendConfig>();
+        if (gpu_config.ok()) {
+          HloGraphNode& node = schedule_graph->GetNode(inst);
+          node.SetForceDelay(gpu_config->should_force_delay_schedule());
+          VLOG(5) << "Setting force delay for instruction: "
+                  << inst->ToString();
+        }
+      }
+    }
+  }
 };
 
 // GPU async tracker maps all collectives onto an async stream resource.
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -408,7 +408,7 @@ class GpuAsyncTrackerBase : public AsyncTracker {
         auto gpu_config = inst->backend_config<GpuBackendConfig>();
         if (gpu_config.ok()) {
           HloGraphNode& node = schedule_graph->GetNode(inst);
-          node.SetForceDelay(gpu_config->should_force_delay_schedule());
+          node.SetForceDelay(gpu_config->force_earliest_schedule());
           VLOG(5) << "Setting force delay for instruction: "
                   << inst->ToString();
         }
--- a/xla/service/gpu/gpu_hlo_schedule_test.cc
+++ b/xla/service/gpu/gpu_hlo_schedule_test.cc
@@ -1111,7 +1111,7 @@ ENTRY e {
 )"));
 }
 
-TEST_F(GpuHloScheduleTest, ProfileGuidedCostModelWithForceDelay) {
+TEST_F(GpuHloScheduleTest, ProfileGuidedCostModelWithForceEarliestSchedule) {
   const char* hlo_text = R"(
   HloModule AsyncAR
   apply_op {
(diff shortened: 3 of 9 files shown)


cheshire:
Are there motivating benchmark results? It's unclear how it should interact with schedule derived from PGO?

Tixxx:
> Are there motivating benchmark results? It's unclear how it should interact with schedule derived from PGO?

@cheshire this can help the collective matmul optimization achieve further speedups, I have added the potential speedup with this mechanism in-place in the main collective matmul [pr](https://github.com/openxla/xla/pull/7854). If we use this mechanism to fine-control the instruction order of some collective-permutes in the collective matmul loop, we could efficiently hide all communications which lead to 4% speedup on 175b model. This pr only exposes the mechanism to force delay instructions, I will open a separate pr to add the attributes to collective matmul instructions.

In terms of iteration with PGO, setting an instruction to force_delay will make LHS ignore the cost of that instruction in profile and always schedule it at the top of the computation.

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -88,6 +98,17 @@ absl::StatusOr<bool> HandleRsWindowedEinsumLoop(HloComputation* comp,
       ++stream_id;
       changed = true;
     }
+
+    // We need to enforce the first collective-permute to be always scheduled
+    // at the beginning of the loop.
+    HloInstruction* matched_cp;
+    if (Match(inst, m::CollectivePermute(
+                        &matched_cp, m::GetTupleElement(m::Parameter(), 2)))) {
+      // Dispatch the dot to additional compute stream.
```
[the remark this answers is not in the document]
Tixxx: ah yes, deleted it.

jurahul (approved):

xla/service/gpu/backend_configs.proto
```diff
@@ -260,4 +260,10 @@ message GpuBackendConfig {
 
     CudnnfMHABackendConfig cudnn_fmha_backend_config = 9;
   }
+
+  // This attribute instructs the latency-hiding scheduler to delay
+  // scheduling this particular instruction. Note that setting this to
+  // true will make this instruction scheduled at the very beginning of
+  // the parent computation before every other nodes.
+  bool should_force_delay_schedule = 10;
```
cheshire, line 268: Maybe just "force_delay_schedule"? Or "forced_schedule_last"? "should" leaves too much room to interpretation and I think we could be more self-descriptive.

Additionally, "force_delay" (at least to me) contradicts the description of "makes this instruction scheduled at the very beginning", maybe "force_schedule_first"?

I think a motivating use-case should also be mentioned in the comment.

xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -332,6 +332,22 @@ class GpuAsyncTrackerBase : public AsyncTracker {
...
            IsAsyncComputeOp(hlo);
   }
+
+  void PostProcessScheduleGraph(
+      HloScheduleGraph* schedule_graph,
+      const LatencyEstimator* latency_estimator) const override {
+    for (auto inst : schedule_graph->GetOriginalInstrList()) {
+      if (inst->has_backend_config()) {
+        auto gpu_config = inst->backend_config<GpuBackendConfig>();
+        if (gpu_config.ok()) {
```
cheshire, line 342: and it it isn't .ok() ? Are we fine with instructions having or not having GpuBackendConfig?

xla/service/gpu/gpu_hlo_schedule_test.cc
```diff
@@ -933,6 +933,75 @@ ENTRY e {
...
+        return absl::c_find_if(instruction_sequence,
+                               [hlo_name](HloInstruction* instruction) {
+                                 return instruction->name() == hlo_name;
+                               }) -
+               instruction_sequence.begin();
+      };
+  // Using the profile, LHS should schedule all computes between ar pair,
+  // but since dot0 is marked as force delay, it should be scheduled
+  // before ar-start now.
+  EXPECT_LT(get_index("dot0", main), get_index("ar-start", main));
```
cheshire, line 1002: Might be more explicit to verify the whole schedule?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -68,6 +68,16 @@ absl::Status UpdateDotAndConsumerConfig(HloInstruction* dot,
   return absl::OkStatus();
 }
 
+absl::Status SetForceDelayForInstruction(HloInstruction* instr,
+                                         bool force_delay) {
+  auto gpu_config = instr->backend_config<gpu::GpuBackendConfig>();
```
cheshire, line 73: This could be done in a separate PR?

xla/service/xla_aot_compile_test_autotune_results.prototxt
```diff
@@ -24,7 +24,7 @@ results {
 }
 results {
   device: "sm_6.0 with 17071734784B RAM, 56 cores, 1480500KHz clock, 715000KHz mem clock, 4194304B L2$"
-  hlo: "(f32[1,1,2,3]{3,2,1,0}, u8[0]{0}) custom-call(f32[1,2,4,4]{3,2,1,0}, f32[1,2,3,2]{3,2,1,0}), window={size=3x2}, dim_labels=bf01_oi01->bf01, custom_call_target=\"__cudnn$convForward\", backend_config={\"operation_queue_id\":\"0\",\"wait_on_operation_queues\":[],\"cudnn_conv_backend_config\":{\"activation_mode\":\"kNone\",\"conv_result_scale\":1,\"side_input_scale\":0,\"leakyrelu_alpha\":0}}"
+  hlo: "(f32[1,1,2,3]{3,2,1,0}, u8[0]{0}) custom-call(f32[1,2,4,4]{3,2,1,0}, f32[1,2,3,2]{3,2,1,0}), window={size=3x2}, dim_labels=bf01_oi01->bf01, custom_call_target=\"__cudnn$convForward\", backend_config={\"operation_queue_id\":\"0\",\"wait_on_operation_queues\":[],\"cudnn_conv_backend_config\":{\"activation_mode\":\"kNone\",\"conv_result_scale\":1,\"side_input_scale\":0,\"leakyrelu_alpha\":0},\"should_force_delay_schedule\":false}"
```
cheshire, line 27: Do we expect invalidation of all previously stored autotuning results?

cheshire (changes requested):

xla/service/xla_aot_compile_test_autotune_results.prototxt
Tixxx: If the autotuning result validator is doing an exact string match on gpu config, then yes, this would invalidate them. I changed this because `xla_aot_compile_test` is matching the results char by char, although I think only looking at relevant field should be enough.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: Sure, let me take the changes in this pass to a separate pr.

xla/service/gpu/gpu_hlo_schedule.cc
Tixxx: if the instruction doenst have gpu config, then we dont need to look at that, we will simply skip to the next instruction

xla/service/gpu/backend_configs.proto
Tixxx: Yea the name could be confusing. LHS is doing a bottom-up scheduling, so schedule last in LHS is the same as schedule first in execution order. I will change this to `force_earliest_schedule` then.

xla/service/gpu/gpu_hlo_schedule_test.cc
Tixxx: Sure, i will add verifications for other instructions.

xla/service/gpu/gpu_hlo_schedule.cc
cheshire: What instructions don't have a GPU config? I thought they all do?

cheshire (approved):

xla/service/gpu/gpu_hlo_schedule.cc
Tixxx: There are a few that dont have a gpu config, i.e `while` or some jax custom calls that only have backend_config as an encoded string.

Tixxx:
@cheshire is there anything blocking this pr? it'd be great to have this merged soon so i can open the followup pr.

cheshire:
```
[/tensorflow/compiler/xla/service/gpu/gpu_hlo_schedule_test.cc:1164]:19: error: unused variable 'entry' [-Werror,-Wunused-variable]
 1164 |   HloComputation* entry = module->entry_computation();
```

@Tixxx 

@ddunl would be great to unify compiler flags

Tixxx:
> gpu_hlo_schedule_test

Fixed it.

ddunl (approved):

cheshire:
Seems to be flaky TF tests, re-running.

PR #10334, opened 2024-03-07
[ROCM] fixing collective ops tests

pemeliya:
Here I fix collectives ops tests which were not taken care of for quite a some time.
- xla/tests/collective_ops_test_e2e: failed because collectives were made async by default in https://github.com/openxla/xla/commit/79ff5de7601d27e78fc1760b4ca20baded1f9139 while this test checks for sync/async collectives
- xla/tests/collective_ops_test: failed with errors like:
```
INTERNAL: Unsupported instruction opcode: collective-permute
``` 
because HLO passes were not run for a number of subtests. The point is we need [AsyncCollectiveCreator](https://github.com/ROCm/xla/blob/7ab393fc384651f367fdff4c3b1d5b1c4cbd9342/xla/service/gpu/gpu_compiler.cc#L1103) call in order to rewrite sync collectives into "start" and "done" pairs: e.g., collective-permute gets rewritten into 'collective-permute-start' and 'collective-permute-done'.

@xla-rotation: could you have a look, please ?

### DIFF
--- a/xla/tests/collective_ops_test_e2e.cc
+++ b/xla/tests/collective_ops_test_e2e.cc
@@ -78,6 +78,7 @@ class AsyncCollectiveOps : public CollectiveOpsTestE2E,
 
     // Enable or disable all async collectives based on test parameter.
     const bool enable_async = GetParam();
+    debug_options.set_xla_gpu_enable_async_collectives(false);
     debug_options.set_xla_gpu_enable_async_all_reduce(enable_async);
     debug_options.set_xla_gpu_enable_async_collective_broadcast(enable_async);
     debug_options.set_xla_gpu_enable_async_collective_permute(enable_async);
--- a/xla/tests/collective_ops_test.cc
+++ b/xla/tests/collective_ops_test.cc
@@ -480,7 +480,7 @@ XLA_TEST_F(CollectiveOpsTest, AllReduce_ThreeReplicaGroups) {
   TF_ASSERT_OK_AND_ASSIGN(
       std::vector<Literal> results,
       ExecuteReplicated(std::move(module), {&input_literal}, /*num_replicas=*/4,
-                        /*use_threads=*/true));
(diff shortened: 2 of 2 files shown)


xla/tests/collective_ops_test_e2e.cc
```diff
@@ -78,6 +78,7 @@ class AsyncCollectiveOps : public CollectiveOpsTestE2E,
 
     // Enable or disable all async collectives based on test parameter.
     const bool enable_async = GetParam();
+    debug_options.set_xla_gpu_enable_async_collectives(false);
```
dimitar-asenov, line 81: Shouldn't this be `enable_async`, instead of `false`? The comment above certainly suggests this.

If false, is right, please explain why.

dimitar-asenov (changes requested):

xla/tests/collective_ops_test_e2e.cc
pemeliya: Yes, this flag should be set to false. The point, is, it is set to **true** [by default](https://github.com/ROCm/xla/blob/4b8bcbc61b49855c2477ab23bba9c59ab81dd62e/xla/debug_options_flags.cc#L126).   which makes all collectives async by default. However, this test verifies whether a given collective op is sync or async (depending on the boolean test param). A collective op type is set by [GpuAsyncCollectiveAnnotator](https://github.com/ROCm/xla/blob/4b8bcbc61b49855c2477ab23bba9c59ab81dd62e/xla/service/gpu/gpu_compiler.cc#L1150) based on the combination of **xla_gpu_enable_async_collectives** and individual **xla_gpu_enable_async_<>** flags.

dimitar-asenov: Thanks for clarifying, I was under the wrong impression both flags need to be true, but I see now that only one being true is sufficient for enabling the async collective.

dimitar-asenov (approved):

PR #10341, opened 2024-03-07
Use vec.data() instead of &vec[0]

apivovarov:
I noticed that we use address of the first element of the vector to get pointer of the internal buffer - that is a known style "issue" in CodeChecker / clang-tidy. 

_'data' should be used for accessing the data pointer instead of taking the address of the 0-th element_

## readability-container-data-pointer
Finds cases where code could use data() rather than the address of the element at index 0 in a container. This pattern is commonly used to materialize a pointer to the backing data of a container. std::vector and std::string provide a data() accessor to retrieve the data pointer which should be preferred.

This also ensures that in the case that the container is empty, the data pointer access does not perform an errant memory access.

https://clang.llvm.org/extra/clang-tidy/checks/readability/container-data-pointer.html

### DIFF
--- a/xla/hlo/evaluator/hlo_evaluator.cc
+++ b/xla/hlo/evaluator/hlo_evaluator.cc
@@ -4284,7 +4284,7 @@ static absl::StatusOr<bool> GenerateReduceOutputElement(
         // Periodically compute partial sum to avoid linear_indices getting
         // large
         computed_result += *input_arg0->GetSumAsDouble(
-            absl::MakeConstSpan(&linear_indices[0], n_linear_indices));
+            absl::MakeConstSpan(linear_indices, n_linear_indices));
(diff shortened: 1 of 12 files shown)


SandSnip3r:
Your statement about the choice of the const or non-const function invocation is not correct. The choice to return const or non-const depends on the constness of the vector, not the LHS. This is the case for both `operator[]()` and `data()`. This change does not improve anything.

apivovarov:
Agree, thank you for the explanation on relationship btw vector constness and `data()` / `operator[]` result. I updated the PR reasoning. 

CodeChecker / clang-tidy report this style "issue" - _'data' should be used for accessing the data pointer instead of taking the address of the 0-th element_

dimitar-asenov (approved):
Thanks!

PR #10358, opened 2024-03-08
Fix use of bitwise and with boolean operands

apivovarov:
Clang diagnostics reported `use of bitwise '&' with boolean operands` issue

[-Wbitwise-instead-of-logical](https://clang.llvm.org/docs/DiagnosticsReference.html#id82)

Diagnostic text: warning: use of bitwise '&' with boolean operands


dimitar-asenov (approved):
Thanks!

PR #10372, opened 2024-03-08
[GPU] Change cuDNN GEMM fusion flag type to an integer.

sergachev:
This will allow enabling more fusions in parallel with rolling them out by default.

### DIFF
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -711,9 +711,6 @@ message DebugOptions {
   // are counted.
   int64 xla_gpu_skip_mlir_kernels = 282;
 
-  // Let GEMM fusion autotuning probe cuDNN as a backend.
-  bool xla_gpu_cudnn_gemm_fusion = 277;
-
   // Threshold to rewrite matmul to cuBLAS or Triton (minumum combined number of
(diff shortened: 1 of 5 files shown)


xla/xla.proto
```diff
@@ -711,15 +711,15 @@ message DebugOptions {
...
-  // Let GEMM fusion autotuning probe cuDNN as a backend.
-  bool xla_gpu_cudnn_gemm_fusion = 277;
-
   // Threshold to rewrite matmul to cuBLAS or Triton (minumum combined number of
   // elements of both matrices in non-batch dimensions to be considered for a
   // rewrite).
   int64 xla_gpu_gemm_rewrite_size_threshold = 283;
 
-  // Next id: 284
+  // Let GEMM fusion autotuning probe cuDNN as a backend.
```
jreiffers, line 719: Can you document the semantics of the integer here as well please? Let's also try to keep it more or less up-to-date in the future.

jreiffers (approved):

xla/xla.proto
sergachev: Done. OK.

jreiffers (approved):

ddunl (approved):

PR #10389, opened 2024-03-09
[NVIDIA] Enable autotune of Cublaslt Fp8 Gemm

kaixih:
This PR enables the autotune of cublaslt fp8 gemm by default.

cc. @reedwm @philipphack @wenscarl @nluehr


reedwm (approved):
So presumably the cuBLAS algorithms have no meaning on Ampere and later, but the cublas LT algorithms still do. Do you know why only cuBLAS LT has meaningful algorithms? Also does this improve performance at all?

In any case, this looks good to me so approving.

PR #10395, opened 2024-03-09
Handle missing collective_permute in collective-permute-motion

### LINKED ISSUE #10394: XLA crashes in collective-permute-motion when running JAX program (without any collective permute)
The Jax program below segfaults on a machine with 8 V100 GPUs. When looking at the XLA dumps, the last dump is from before collective-permute-motion. The problem seems to be that the [FindMovableClusterAtBodyRoot]( https://github.com/openxla/xla/blob/36b3b135bb75dd292c420f65d20b8925a306e709/xla/service/spmd/collective_permute_motion.cc#L91) function returns a cluster with the `collective_permute` field being null. This later crashes [here](https://github.com/openxla/xla/blob/36b3b135bb75dd292c420f65d20b8925a306e709/xla/service/spmd/collective_permute_motion.cc#L246), where the code assumes that returned cluster has the field set.

```python
import jax
from jax import lax
from jax._src import test_util as jtu
import jax.numpy as jnp
from jax._src.sharding_impls import (NamedSharding, PositionalSharding,
                                     SingleDeviceSharding, GSPMDSharding,
                                     TransferToMemoryKind,
                                     common_devices_indices_map)
from jax.sharding import PartitionSpec as P
import math
import numpy as np

def test():
  mesh = jtu.create_global_mesh((2,), ("x",))
  shape = (256, 128)
  np_inp = np.arange(math.prod(shape), dtype=np.float32).reshape(shape)
  s = NamedSharding(mesh, P("x"))
  inp = jax.device_put(np_inp, s)

  def f(x):
    def g(ys, _):
      y, _ = ys
      w = jnp.sin(y)
      return (w, jnp.sum(w)), None
    _, scan_out = jax.lax.scan(g, (x, np.array(1, dtype=np.float32)), [np_inp])[0]
    return scan_out

  f = jax.jit(jax.grad(f))
  f(inp)  # doesn't crash
 
test()
```

jaro-sevcik:
Currently, the `MoveCollectivePermutes` function expects that if `FindMovableClusterAtBodyRoot` function returns a cluster, then the cluster's `collective_permute` field is not null. However, if `FindMovableClusterAtBodyRoot` does not find a collective_permute on when traversing an input, it happily returns cluster with `collective_permute == null`. This patch just returns `nullopt` for a 'cluster' without collective-permute.

Fixes #10394.

### DIFF
--- a/xla/service/spmd/collective_permute_motion.cc
+++ b/xla/service/spmd/collective_permute_motion.cc
@@ -129,6 +129,9 @@ std::optional<MovableCluster> FindMovableClusterAtBodyRoot(
       }
     }
   }
+  if (cluster.collective_permute == nullptr) {
+    return std::nullopt;
+  }
   return cluster;
 }
 
--- a/xla/service/spmd/collective_permute_motion_test.cc
+++ b/xla/service/spmd/collective_permute_motion_test.cc
@@ -68,6 +68,40 @@ TEST_F(CollectivePermuteMotionTest, SimpleMove) {
   EXPECT_THAT(output, op::Multiply(select, select));
 }
 
+TEST_F(CollectivePermuteMotionTest, NoCollectivePermute) {
+  absl::string_view hlo_string = R"(
+  HloModule test
+  body {
+    loop_var = (s32[], f32[], f32[]) parameter(0)
+    constant.1 = s32[] constant(1)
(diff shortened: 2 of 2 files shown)


jvstokes (approved):
Thanks for patching this!

PR #10421, opened 2024-03-11
[ROCm] Add CudnnNormRewriter pass to AMDGPU compiler

hsharsha:


reedwm (changes requested):
Don't you need to support layer norm in stream executor, e.g. by overriding [`NormRunnerFromDesc`](https://github.com/openxla/xla/blob/e434d6b88784b2ed231581edf9519d0b3a64dae4/xla/stream_executor/dnn.h#L1716) in MIOpenSupport?

i-chaochen:
@reedwm thanks for the review. Yes, agree. We will withdraw this till the MIOpen has related support.

PR #10450, opened 2024-03-12
Generate async copies with multiple streams for offloading

zhenying-liu:
Improving offloading via async copies.

1. Added the annotation in copy-start and copy-done instructions after hlo-rematerialization pass. The annotation specifies which particular stream the copy-start and copy-done will be executed on.
2. Implemented multiple streams in copy_thunk: the main stream is for computation and the other stream is for memcpy. So the memcpy is really asynchronous. Without this change, all the operations are in the main stream and sequential.
3. Used events to wait for the memcpy to be done. At copy-start, the "other" stream will do memcpy and record this event. The main compute stream will wait for this event at copy-done for the "other" stream to complete.
4. Added the xla flags to enable offloading of hlo-rematerialization pass for GPUs. Right now, offloading is disabled. Offloading is utilized to generate copy-start/copy-done.


xla/service/gpu/stream_attribute_annotator.cc
```diff
@@ -130,26 +131,45 @@ absl::StatusOr<bool> StreamAttributeAnnotator::Run(
...
+                                  instr, instr_gpu_config.value()));
+          changed |= comp_result;
+        }
+
+        TF_ASSIGN_OR_RETURN(
+            bool user_result,
+            AnnotateStreamAttributesForUsers(instr, instr_gpu_config.value()));
+        changed |= user_result;
+      } else {
+        if (instr->opcode() == HloOpcode::kCopyStart) {
```
Tixxx, line 158: i think if you just move this before `if (instr->opcode() == HloOpcode::kFusion) {`, after setting the stream id for kCopyStart, `AnnotateStreamAttributesForUsers` should take care of the kCopyDone?

xla/service/gpu/stream_attribute_annotator.cc
```diff
@@ -130,26 +131,45 @@ absl::StatusOr<bool> StreamAttributeAnnotator::Run(
   XLA_VLOG_LINES(
       5, "StreamAttributeAnnotator::Run(), before:\n" + module->ToString());
   bool changed = false;
+  int64_t host_memory_color =
+      static_cast<int64_t>(stream_executor::MemoryType::kHost);
```
Tixxx, line 135: looks like this value is fixed to be 5? I think that has a higher chance to have a clash with other instructions then it would be serialized. Maybe use `hlo_query::NextChannelId(*module)` for setting the stream id?

xla/service/gpu/runtime/copy_thunk.h
```diff
@@ -18,62 +18,113 @@ limitations under the License.
...
+  DeviceHostCopyThunk(ThunkInfo thunk_info,
+                      const BufferAllocation::Slice &source_buffer,
+                      const BufferAllocation::Slice &destination_buffer,
+                      uint64_t mem_size,
+                      std::shared_ptr<CopyAsyncEvents> events,
+                      const HloInstruction *instr, bool device_to_host);
+  absl::Status ExecuteOnStream(const ExecuteParams &params) override;
+
+private:
+  const ExecutionStreamId stream_id_;
```
Tixxx, line 103: like above, base Thunk class already has this field.

xla/service/gpu/runtime/copy_thunk.cc
```diff
@@ -45,41 +43,112 @@ absl::Status DeviceToDeviceCopyThunk::ExecuteOnStream(
...
+//===----------------------------------------------------------------------===//
+DeviceHostCopyThunk::DeviceHostCopyThunk(
+    ThunkInfo thunk_info, const BufferAllocation::Slice &source_buffer,
+    const BufferAllocation::Slice &destination_buffer, uint64_t mem_size,
+    std::shared_ptr<CopyAsyncEvents> async_events, const HloInstruction *instr,
+    bool device_to_host)
     : DeviceToDeviceCopyThunk(thunk_info, source_buffer, destination_buffer,
-                              mem_size) {}
+                              mem_size),
+      stream_id_(thunk_info.execution_stream_id),
```
Tixxx, line 89: so when you call WithProfileAnnotation when creating the thunk in ir emitter, it will already set the execution stream id in the base Thunk class, you dont need to save another copy here. You can get the stream id by calling Thunk::execution_stream_id()

xla/service/gpu/stream_attribute_annotator.cc
zhenying-liu: @Tixxx I tried that and found AnnotateStreamAttributesForUsers() inserted an async-done as the user of copy-start and crashed the  HloVerifier. "The operand of a copy-done instruction needs to be copy-start, found async-done". Can I avoid the generation of async-done?

xla/service/gpu/runtime/copy_thunk.cc
zhenying-liu: Will remove.

xla/service/gpu/runtime/copy_thunk.h
zhenying-liu: Will remove

xla/service/gpu/stream_attribute_annotator.cc
zhenying-liu: Thanks. Will use NextChannelId() instead. AnnotateStreamAttributesForUsers() makes sense to propagate the channel id through the def-use chain.

xla/service/gpu/stream_attribute_annotator.cc
Tixxx: Are you talking about this pass StreamAttributeAsyncWrapper?  StreamAttributeAnnotator shouldnt add any additional async instructions, it only adds annotations. StreamAttributeAsyncWrapper will take the stream attribute and turn it into asyncstart/done. Hlo remat runs after StreamAttributeAsyncWrapper, so i dont see an issue there.

cheshire (changes requested):
Thanks for the PR ! 

1) Could you expand the description to describe what exactly is being done?
2) Could you consider splitting into multiple PRs?
3) Could you show the effect on benchmarks you're looking at?

zhenying-liu:
> 1. Could you expand the description to describe what exactly is being done?
Yes, explained the description.
> 2. Could you consider splitting into multiple PRs?
Yes, I can split it into 3 stacked PRs: 
   (1) Add the stream id annotation for copy-start/copy-done in backend-config.
   (2) Multiple streams and events implementation in copy_thunk and emitter.
   (3) Enable offloading with XLA flags.
> 3. Could you show the effect on benchmarks you're looking at?
For the unit test, I can see from the log that the memcpy is executed on a separate stream with nonzero id other than the main stream with id 0. I'm running the JAX code like ResNet-50 to see if offloading kicks in and memcpy is running on other streams.

cheshire:
Yes - would be great to split - thanks!

zhenying-liu:
> Yes - would be great to split - thanks!
OK, will do.

zhenying-liu:
Screenshot 2024-03-16 at 10 33 03 AM
The Nsys picture shows the two D2H and two H2D ran on Stream 254 highlighted in green circles, while the computations like wrapped_tanh and wrapped_add_3 ran on Stream 118. @cheshire This answers the third question of "Could you show the effect on benchmarks you're looking at?".

PR #10452, opened 2024-03-12
Hook up host offloading gpu with fixed layout

jaro-sevcik:
NOT FOR COMMIT

This hooks op the memory offloading passes to the gpu compiler and fixes layouts of the offloading ops so that we can run somewhat realistic examples (such as PaxML on single-GPU without jax.lax.scan).


jaro-sevcik:
Replaced by https://github.com/openxla/xla/pull/10495.

PR #10474, opened 2024-03-12
memcmp is not valid for comparing 4 bit types

elfringham:
To compare the subshape of a literal that is a 4 bit type it is not valid to compare the memory with memcmp as the upper nibble of each byte may not be initialised and result in a false failure of the comparison. Instead detect this situation and compare each nibble that will hold the 4 bit type.

### DIFF
--- a/xla/literal.cc
+++ b/xla/literal.cc
@@ -1844,6 +1844,15 @@ bool LiteralBase::Piece::EqualElements(const LiteralBase::Piece& other) const {
     CHECK(LayoutUtil::IsDenseArray(subshape()))
         << __func__ << " is only supported for dense arrays: " << subshape();
     CHECK_EQ(size_bytes_dense(), other.size_bytes_dense());
+    if (primitive_util::Is4BitType(subshape().element_type())) {
+      auto one_array = buffer();
+      auto two_array = other.buffer();
+      for (int64_t i = 0; i < size_bytes_dense(); ++i) {
+        if ((one_array[i] & uint8_t{0xf}) != (two_array[i] & uint8_t{0xf}))
+          return false;
+      }
+      return true;
+    }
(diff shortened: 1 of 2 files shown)


xla/literal.cc
```diff
@@ -1844,6 +1844,16 @@ bool LiteralBase::Piece::EqualElements(const LiteralBase::Piece& other) const {
     CHECK(LayoutUtil::IsDenseArray(subshape()))
         << __func__ << " is only supported for dense arrays: " << subshape();
     CHECK_EQ(size_bytes_dense(), other.size_bytes_dense());
+    auto one_array = buffer();
+    auto two_array = other.buffer();
```
reedwm, line 1848: Define these inside if-statement below

xla/literal.cc
```diff
@@ -1844,6 +1844,16 @@ bool LiteralBase::Piece::EqualElements(const LiteralBase::Piece& other) const {
...
+    if (primitive_util::Is4BitType(subshape().element_type())) {
+      for (int64_t i = 0; i < size_bytes_dense(); ++i) {
+        if ((*(one_array + i) & uint8_t{0xf}) !=
```
reedwm, line 1851: Use notation `one_array[i]` instead of `*(one_array + i)`.

reedwm (changes requested):
Thanks for the fix! It is weird how the upper-4 bits of int4 values are not set to any consistent value but that's how XLA works.

Is it possible to write a test for this? If it's difficult to reproduce within a unit test, I'm fine not having a test.

xla/literal.cc
elfringham: Done

xla/literal.cc
elfringham: Done

elfringham:
Changes done as requested including unit test added.

reedwm (approved):

PR #10489, opened 2024-03-13
[GPU] Fix command buffer support for cuDNN fusions.

sergachev:
CuDnnCmd is constructed before DnnGraph in CuDnnThunk is initialized so CuDnnCmd has to get  unique_ptr\<DnnGraph\>& instead of DnnGraph& at initialization.

Accordingly cuDNN thunks have to be initialized before command buffer ones to initialize graphs before they get captured.

Test CommandBuffersAreSupported used to not demonstrate the use of command buffers because the corresponding command buffer call used to be inlined and no command buffers were created. This is now cleaned up and does work as expected with minimal CUDA graph size set to 1 with a flag.


xla/service/gpu/runtime/command_buffer_cmd.h
```diff
@@ -769,7 +768,7 @@ class CuDnnCmd : public TracedCommandBufferCmd {
 
  private:
   std::vector<BufferAllocation::Slice> args_;
-  const se::dnn::DnnGraph& graph_;
+  const std::unique_ptr<se::dnn::DnnGraph>& graph_;
```
ezhulenev, line 771: Instead of taking a reference you can use `std::shared_ptr`, the fact that Thunks stay alive is mostly accidental because of the workaround for cupti+cuda graphs bugs, this will go away once CUDA fix will land

xla/service/gpu/gpu_executable.cc
```diff
@@ -472,8 +472,17 @@ absl::Status ExecuteThunks(
         &collective_cliques};
 
     tsl::profiler::TraceMe trace([&] { return "Thunks::Initialize"; });
+    // cuDNN fusion thunks have to be initialized before they can be captured
+    // into command buffers at the initialization of command buffer thunks.
     for (const std::unique_ptr<Thunk>& thunk : thunk_sequence) {
-      TF_RETURN_IF_ERROR(thunk->Initialize(initialize_params));
+      if (thunk->kind() == Thunk::kCuDnn) {
```
ezhulenev, line 478: If state is properly shared we don't need any special cases here

xla/service/gpu/runtime/cudnn_thunk.h
```diff
@@ -43,7 +43,7 @@ class CuDnnThunk : public Thunk {
   absl::Status Initialize(const InitializeParams&) override;
   absl::Status ExecuteOnStream(const ExecuteParams&) override;
 
-  const se::dnn::DnnGraph& graph() const { return *graph_; }
+  const std::unique_ptr<se::dnn::DnnGraph>& graph() const { return graph_; }
```
ezhulenev, line 46: How about returning `std::shared_ptr<LazyDnnGraph>` (or something like this), and whoever gets called first (Thunk or corresponding Command) will initialize it.

Basically command emitter should steal (or share) a state from a thunk

xla/service/gpu/runtime/command_buffer_cmd.h
sergachev: Done.

xla/service/gpu/gpu_executable.cc
sergachev: Done.

xla/service/gpu/runtime/cudnn_thunk.h
sergachev: I just added a call to Initialize() to ExecuteOnStream() - does this look reasonable to you? Body of Initialize() is already wrapped in call_once().

ezhulenev (approved):

cheshire (approved):

cheshire:
Build layering checks are failing: 

```
[/xla/service/gpu/fusions/cudnn_test.cc:19](:10: error: module //third_party/tensorflow/compiler/xla/service/gpu/fusions:cudnn_test_gpu_a100 does not depend on a module exporting 'third_party/tensorflow/compiler/xla/tests/filecheck.h'
```

sergachev:
Fixed the build file.

cheshire (approved):

sergachev:
Rebased.

ezhulenev (approved):

sergachev:
I don't know what's holding this one up; tried to update the build file.

PR #10495, opened 2024-03-13
Enforce same input/output layout for offloading ops

jaro-sevcik:
This patch makes sure that the host offloader will not introduce layout mismatches
after removing the offloading custom calls by constraining the layout assignment
to assign the same layout for the custom call's input and output.


jaro-sevcik:
Here is a link to a motivation/summary doc: https://docs.google.com/document/d/12LjzvR0foAMnkB7p7-EtiuwdhfmsRAsEWYFTGGiOs7U/edit?usp=sharing

ezhulenev (approved):

PR #10376, opened 2024-03-08
Fix expm1 inaccuracies on complex inputs with small absolute values. Add Cosm1.

pearu:
As in the title.

Tests and improvement reports are in https://github.com/google/jax/pull/20144.

Accuracy tests are enabled in https://github.com/google/jax/pull/20436

PR #10503, opened 2024-03-13
Fix log1p inaccuracies on complex inputs with large absolute values.

pearu:
As in the title.

Tests and improvement reports are in https://github.com/google/jax/pull/20144.

Accuracy tests are enabled in https://github.com/google/jax/pull/20436


pearu:
The same comment as in https://github.com/openxla/xla/pull/10376#issuecomment-1991067768 applies here as well: the CI testOnComplexPlane failures are expected and that should be fixed after https://github.com/google/jax/pull/20144 lands

cheshire (changes requested):
Could we test the improved accuracy?

pearu:
> Could we test the improved accuracy?

The accuracy of log1p is tested in https://github.com/google/jax/pull/20144 . See the description of that PR that reports the accuracy improvements for log1p.

tdanyluk:
@cheshire gentle reminder

cheshire:
> The accuracy of log1p is tested in https://github.com/google/jax/pull/20144 . See the description of that PR that reports the accuracy improvements for log1p.

Understood, so JAX test is an e2e integration test, but could we have an XLA unit test as well?

pearu:
> ... could we have an XLA unit test as well?

@cheshire Having the complex functions correctness tests in XLA rather than in JAX makes lots of sense.

That said, when testing functions for correctness, it is essential to have reference implementations that correctness are verified in a platform independent way. Unfortunately, there are not many libraries available that qualify as a reference library.

For instance, within the Python/JAX framework, one obvious choice would be numpy, however, as it turns out, numpy functions often rely on system math libraries and the function evaluation results can be different on different platforms. This is one important reason why in JAX tests we use mpmath as a reference library. While there exists a few bugs in mpmath itself that require workarounds, one of the most important properties of mpmath is that its functions evaluation results are platform independent. In https://github.com/pearu/complex_function_validation/, I have validated complex functions from a number of Python array libraries for correctness with a conclusion that mpmath is a reasonable reference library indeed.

When considering XLA, one of the obvious choices is libc++. Unfortunately, there exists incorrectness issues also in libc++ that must be tackled via correctness analysis and implementing workarounds before using it as a reference library. The current tests in XLA are very minimal (functions are tested on a couple of points only) and miss many incorrectness issues that the correctness analysis of JAX complex functions above also demonstrates.

In short term, one approach would be to generate tables of inputs and outputs to different complex functions (say, using mpmath) and use these tables as a reference database when testing XLA complex functions for correctness.

What do you think?

cheshire:
In this case having hardcoded expected output values along with the long comment as the one above indeed seems like the preferred option, thanks!

pearu:
> In this case having hardcoded expected output values along with the long comment as the one above indeed seems like the preferred option, thanks!

I suggest to do this in a follow up PR because it is relevant for other functions as well, and to avoid XLA and JAX repos going out of sync (https://github.com/google/jax/pull/20436 has landed that includes `xla_extension_version` update related to this PR).

cheshire:
Sorry what's the connection between going out of sync and having a unit test?

pearu:
> The current situation is a bit worse: XLA and JAX main branches are out of sync.

A quick solution to this is to create a JAX PR that disables log1p units tests. Although, it would also make sense to land this PR as we do test the related changes. And in the future, we'll implement the corresponding unit tests also in XLA. What do you think?

cheshire:
Sorry I'm still confused. This PR improves log1p handling in XLA, right? So it would be natural to accompany it with a unit test? The question regarding JAX and/or JAX/XLA interop is orthogonal?

pearu:
> Sorry I'm still confused. This PR improves log1p handling in XLA, right? 

Correct.

> So it would be natural to accompany it with a unit test?

True. I am working on it atm.

> The question regarding JAX and/or JAX/XLA interop is orthogonal?

There exists inference only when JAX implements tests for XLA features. AFAIK, this situation is not unique to this issue.

cheshire (approved):

reedwm:
@pearu, I am working on this merging this. When I ran `generate_complex_unary_op_samples.py`, I got a slightly different results for nonfinite inputs. For example, for the first row, I got the result:

```
/* 0 */ { { -inf, -inf }, { inf, nan }, 1.e+00f },
```

While in the PR, you have the result:

```
/* 0 */ { { -inf, -inf }, { inf, -3*pi/4 }, 1.e+00f },
```

Do you know why this discrepancy is occurring? I am not familiar with how log and log1p works on non-real inputs, especially non-finite non-real inputs. I installed the nightly versions of JAX following https://jax.readthedocs.io/en/latest/installation.html#nightly-installation and installed mpmath with `pip install mpmath`. Perhaps we shouldn't include nonfinite values in the complex_unary_op_samples.h?

pearu:
> Do you know why this discrepancy is occurring?

Yes. It is due to a mpmath bug that has a workaround when applying the following patch to JAX:
```
$ git diff
diff --git a/jax/_src/test_util.py b/jax/_src/test_util.py
index 63d9597f0..5a64fdce4 100644
--- a/jax/_src/test_util.py
+++ b/jax/_src/test_util.py
@@ -1746,6 +1746,22 @@ class numpy_with_mpmath:
   def expm1(self, x):
     return x.context.expm1(x)
 
+  def log1p(self, x):
+    ctx = x.context
+    # workaround mpmath bugs:
+    if isinstance(x, ctx.mpc):
+      if ctx.isinf(x.real) and ctx.isinf(x.imag):
+        pi = ctx.pi
+        if x.real > 0 and x.imag > 0:
+          return ctx.make_mpc((x.real._mpf_, (pi / 4)._mpf_))
+        if x.real > 0 and x.imag < 0:
+          return ctx.make_mpc((x.real._mpf_, (-pi / 4)._mpf_))
+        if x.real < 0 and x.imag < 0:
+          return ctx.make_mpc(((-x.real)._mpf_, (-3 * pi / 4)._mpf_))
+        if x.real < 0 and x.imag > 0:
+          return ctx.make_mpc(((-x.real)._mpf_, (3 * pi / 4)._mpf_))
+    return ctx.log1p(x)
+
   def log2(self, x):
     return x.context.ln(x) / x.context.ln2
```

This workaround is not in the JAX main yet but I'll add it there in a follow-up PR.

Notice that there is no need to re-generate complex_unary_op_samples.h unless generate_complex_unary_op_samples.py is updated with other complex function cases.

> Perhaps we shouldn't include nonfinite values in complex_unary_op_samples.h?

No, it is important to test complex functions also with nonfinite inputs as functions should return correct values at all points where the functions are defined.

pearu:
@reedwm what is the status of landing this PR?

reedwm:
Sorry for the delay. When merging the PR, there were some minor lint errors in complex_unary_op_samples.h that were only caught by the internal linter. This is common and typically the merger fixes before merging, but in this case generate_complex_unary_op_samples.py should also be updated to output a correctly-linted file. However, due to the JAX issue you described, I couldn't regenerate the correct file.

I think we should wait until the JAX fix is in, so that generate_complex_unary_op_samples.py correctly generates the file once this PR is merged. But if you want to get this in earlier, I can instead manually apply the patch. Let me know what you prefer.

pearu:
@reedwm Thanks for finding the fix to the lint issue! I think it would be preferred to land your fix as soon as possible (because there are several PRs blocked on this work) and as it is, even with manually modified complex_unary_op_samples.h that I can use as an example of the lint issue to update the generate script in a follow-up PR.

pearu:
The mpmath fix is submitted: https://github.com/google/jax/pull/20560

pearu:
> I think we should wait until the JAX fix is in...

The JAX fix has landed: https://github.com/google/jax/pull/20560

reedwm (approved):
Approving to re-import PR. I didn't take a close look since @cheshire already reviewed

reedwm:
> If not, could you provide the linter failure messages?

There is a missing include guard in the generated header file. This is easier to fix from my side though, as path names and therefore include guard names are slightly different internally. I'm currently working on merging this.

PR #10536, opened 2024-03-13
Size 1 Dimensions in Layer Norm Fusion

philipphack:
Enables the fusion of layer norm patterns with degenerate input dimensions of size 1 and with additional optional type conversions or reshapes adding or removing degenerate dimensions.

### DIFF
--- a/xla/service/gpu/cudnn_norm_rewriter.cc
+++ b/xla/service/gpu/cudnn_norm_rewriter.cc
@@ -172,6 +172,28 @@ bool CompatibleElementType(const HloInstruction* instr) {
   return element_type == BF16 || element_type == F16 || element_type == F32;
 }
 
+// Returns the dimensions associated with shape, adjusted for the removal of any
+// degenerate dimensions in shape.
+std::vector<int64_t> AdjustedDimensions(const Shape& shape,
+                                        absl::Span<const int64_t> dimensions) {
+  absl::flat_hash_map<int64_t, int64_t> dimension_map;
+  for (int64_t dimension = 0, non_degen_dimension = 0; dimension < shape.rank();
+       ++dimension) {
+    if (shape.dimensions(dimension) > 1) {
+      dimension_map.insert({dimension, non_degen_dimension});
+      non_degen_dimension++;
+    }
+  }
+  std::vector<int64_t> adjusted_dimensions;
+  for (int64_t dimension : dimensions) {
+    auto non_degenerate_dimension = dimension_map.find(dimension);
+    if (non_degenerate_dimension != dimension_map.end()) {
+      adjusted_dimensions.emplace_back(non_degenerate_dimension->second);
+    }
+  }
+  return adjusted_dimensions;
+}
+
 // Returns the dimensions of broadcast or reduction instructions, adjusted for
 // the removal of any degenerate dimensions in the output or input.
 std::vector<int64_t> AdjustedDimensions(const HloInstruction* instr) {
@@ -183,22 +205,7 @@ std::vector<int64_t> AdjustedDimensions(const HloInstruction* instr) {
   } else {
     return {};
   }
-  absl::flat_hash_map<int64_t, int64_t> dimension_map;
-  for (int64_t dimension = 0, non_degen_dimension = 0; dimension < shape.rank();
-       ++dimension) {
-    if (shape.dimensions(dimension) > 1) {
-      dimension_map.insert({dimension, non_degen_dimension});
-      non_degen_dimension++;
-    }
-  }
-  std::vector<int64_t> dimensions;
-  for (int64_t dimension : instr->dimensions()) {
-    auto non_degenerate_dimension = dimension_map.find(dimension);
-    if (non_degenerate_dimension != dimension_map.end()) {
-      dimensions.emplace_back(non_degenerate_dimension->second);
-    }
-  }
-  return dimensions;
+  return AdjustedDimensions(shape, instr->dimensions());
 }
 
 // Returns whether the HLO Computation applied by instr calculates the sum of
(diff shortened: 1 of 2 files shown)


philipphack:
CC @reedwm.

xla/service/gpu/cudnn_norm_rewriter.cc
```diff
@@ -930,23 +959,32 @@ class CudnnNormRewriterVisitor : public DfsHloRewriteVisitor {
...
-      std::vector<int64_t> non_norm_dims;
-      for (int64_t x_dim = 0; x_dim < x.Instr()->shape().rank(); ++x_dim) {
+      std::vector<int64_t> non_norm_dims, non_norm_dims_adjusted;
+      for (int64_t x_dim = 0, degen_dims = 0; x_dim < x.Instr()->shape().rank();
+           ++x_dim) {
         if (std::find(norm_dims.begin(), norm_dims.end(), x_dim) ==
             norm_dims.end()) {
           non_norm_dims.emplace_back(x_dim);
+          if (x.Instr()->shape().dimensions(x_dim) != 1) {
+            non_norm_dims_adjusted.emplace_back(x_dim - degen_dims);
```
reedwm, line 983: The logic for non_norm_dims_adjusted is a bit hard to follow. Can you create an overload of `AdjustedDimensions` that takes a `Shape` and `Span<int>` of dimensions (with a clear comment describing what it does), have the version taking an instruction call this alternative version, and also call the alternative version here on `non_norm_dims` to get `non_norm_dims_adjusted`?

xla/service/gpu/cudnn_norm_rewriter_test.cc
```diff
@@ -1088,6 +1375,125 @@ TEST_F(CudnnNormRewriterTest, LayerNormTrainBackward4D12) {
   TestNorm(hlo_text, optimized_hlo);
 }
 
+// Dbias not expected to be fused into norm gradient.
```
reedwm, line 1378: Comment this is because dbias is not fused when a reduction dimension is degenerate.

Also, why is Dbias not fused in this case?

xla/service/gpu/cudnn_norm_rewriter.cc
```diff
@@ -568,12 +589,13 @@ auto Expectation(UniqueHloInstruction* expectation, Pattern pattern) {
 template <typename Pattern>
 auto Expectation(UniqueHloInstruction* expectation, HloInstruction** reduce,
                  Pattern pattern) {
-  auto shared_subpattern = MultiplyAnyOrder(m::Broadcast(m::ConstantScalar()),
-                                            AddReduce(reduce, pattern))
-                               .WithPredicate([](const HloInstruction* instr) {
-                                 return CalculatesExpectation(instr);
-                               })
-                               .WithPredicate(expectation->capture_or_verify);
+  auto shared_subpattern = OptionalSupportedTransform(
```
reedwm, line 592: You add a lot of OptionalSupportedTransform calls. The PR description does state we now support "additional optional type conversions", but this also allows additional reshapes. The AdjustedDimensions function can take into account these reshapes, but I think it is not tested. Update the PR description to state additional reshapes removing/adding degenerate dimensions are allowed, and also add a test with such a reshape.

reedwm (changes requested):

xla/service/gpu/cudnn_norm_rewriter.cc
```diff
@@ -171,6 +172,42 @@ bool CompatibleElementType(const HloInstruction* instr) {
   return element_type == BF16 || element_type == F16 || element_type == F32;
 }
 
+// Returns the dimensions associated with shape, adjusted for the removal of any
+// degenerate dimensions in shape.
```
reedwm, line 176: Maybe add the sentences:

Specifically, for each dimension 'd' in dimensions, returns the new index of d if all dimensions of size 1 are removed from 'shape'. If d itself has size 1, it is not included in the returned vector.

xla/service/gpu/cudnn_norm_rewriter.cc
```diff
@@ -275,15 +309,12 @@ bool FindTarget(const HloInstruction* custom_call, const HloInstruction* instr,
...
-  // The original and reshaped shape must not have degenerate dimensions.
-  if (ShapeUtil::HasDegenerateDimensions(original_shape) ||
-      ShapeUtil::HasDegenerateDimensions(reshaped_shape)) {
-    return {};
-  }
-
   auto dimension_product =
       [](const Shape& shape,
          absl::Span<const int64_t> product_dimensions) -> int64_t {
+    if (product_dimensions.empty()) {
```
reedwm, line 315: Why return 0 here instead of 1?

xla/service/gpu/cudnn_norm_rewriter.cc
```diff
@@ -930,25 +969,30 @@ class CudnnNormRewriterVisitor : public DfsHloRewriteVisitor {
...
       // normalized are the leading dimensions.
-      std::vector<int64_t> non_norm_dims;
+      std::vector<int64_t> non_norm_dims, non_norm_dims_adjusted;
       for (int64_t x_dim = 0; x_dim < x.Instr()->shape().rank(); ++x_dim) {
         if (std::find(norm_dims.begin(), norm_dims.end(), x_dim) ==
             norm_dims.end()) {
           non_norm_dims.emplace_back(x_dim);
         }
       }
+      non_norm_dims_adjusted =
```
reedwm, line 993: Better to not declare this above but instead here, and do: `std::vector<int64_t>non_norm_dims_adjusted = AdjustedDimensions(x.Instr()->shape(), non_norm_dims);`

reedwm (changes requested):

xla/service/gpu/cudnn_norm_rewriter.cc
SandSnip3r: This amendment greatly improved my understanding of the function without needing to look at the actual implementation of the function.

reedwm (approved):

PR #10561, opened 2024-03-14
Verifier: Allow dynamic-slice and dynamic-update-slice change spaces

jaro-sevcik:
In preparation for host offloading with partial copies (using dynamic-update-slice and dynamic-slice instructions), this patch relaxes the verifier to allow different memory spaces for dynamic-update-slice and dynamic-slice instructions.


jaro-sevcik:
Obsolete, similar patch already landed.

PR #10620, opened 2024-03-15
Add missing unit test tags in `xla/service/gpu/BUILD`

hsharsha:


xla/service/gpu/BUILD
```diff
@@ -82,6 +81,7 @@ tf_proto_library(
 xla_cc_test(
     name = "backend_configs_test",
     srcs = ["backend_configs_test.cc"],
+    tags = tf_gpu_tests_tags(),
```
cheshire, line 84: If this test had no custom tags before, does it need them now?

cheshire (changes requested):

xla/service/gpu/BUILD
hsharsha: @cheshire  Since some of the unit tests do no have custom tags, they were not selected to run as part of CI as they were filtered out. Motivation for this PR is to identify such unit tests and run them as part of CI.

hsharsha:
Moving this PR to draft as we need to discuss on right custom tags and what unit tests we want to enable in CI.

PR #10628, opened 2024-03-15
Exclude FP8 Conversions in DotOperandConverter

philipphack:
Excludes conversions of dot operands between F8E4M3FN and F8E5M2 types in the DotOperandConverter pass.


philipphack:
CC @reedwm.

reedwm (approved):

PR #10631, opened 2024-03-16
[XLA:CPU][oneDNN] Minor fix in CPU bfloat16 support

yimeisun123:
Align the usage of oneDNN bf16 cpu float normalization with the same guarding condition as for oneDNN rewrites in compilation step, i.e., only when oneDNN is enabled and in non-AOT mode.

### DIFF
--- a/xla/service/cpu/cpu_compiler.cc
+++ b/xla/service/cpu/cpu_compiler.cc
@@ -712,8 +712,8 @@ Status CpuCompiler::RunHloPassesThroughLayoutAssn(
   // backend can support BF16/F8 operations without directly implementing a
   // BF16/F8 lowering for most ops.
   FloatSupport bf16_support(BF16);
-  CpuFloatSupport onednn_bf16_support(BF16);
 #if defined(INTEL_MKL) && defined(ENABLE_ONEDNN_V3)
+  CpuFloatSupport onednn_bf16_support(BF16);
   if (!is_aot_compile) {
     pipeline.AddPass<FloatNormalization>(&onednn_bf16_support);
   } else {
--- a/xla/service/cpu/cpu_compiler.cc
+++ b/xla/service/cpu/cpu_compiler.cc
@@ -711,12 +711,17 @@ Status CpuCompiler::RunHloPassesThroughLayoutAssn(
   // Convert BF16 and F8 operations to F32 and F16 respectively so that the CPU
   // backend can support BF16/F8 operations without directly implementing a
   // BF16/F8 lowering for most ops.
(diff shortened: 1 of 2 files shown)


tdanyluk (approved):
I'm not a CPU expert, but this looks fine. Thank you for the contribution.

xla/service/cpu/cpu_compiler.cc
```diff
@@ -711,12 +711,17 @@ Status CpuCompiler::RunHloPassesThroughLayoutAssn(
   // Convert BF16 and F8 operations to F32 and F16 respectively so that the CPU
   // backend can support BF16/F8 operations without directly implementing a
   // BF16/F8 lowering for most ops.
+  FloatSupport bf16_support(BF16);
+  CpuFloatSupport onednn_bf16_support(BF16);
```
penpornk, line 715: `CpuFloatSupport` is only defined when `INTEL_MKL` and `ENABLE_ONEDNN_V3` are defined. It must always be under the `#if defined(INTEL_MKL) && defined(ENABLE_ONEDNN_V3)` guard.

See an example error message in the [ARM CI](https://source.cloud.google.com/results/invocations/55cc0324-7842-4614-950b-2cddf4073439/log) failure:
```
xla/service/cpu/cpu_compiler.cc:715:3: error: unknown type name 'CpuFloatSupport'; did you mean 'FloatSupport'?
  715 |   CpuFloatSupport onednn_bf16_support(BF16);
      |   ^~~~~~~~~~~~~~~
      |   FloatSupport
./xla/service/float_support.h:28:7: note: 'FloatSupport' declared here
   28 | class FloatSupport {
      |       ^
1 warning and 1 error generated.
```

penpornk (changes requested):
Thank you for the PR!

penpornk:
For future PRs, please also add a description on what the PR does and what problem it addresses, e.g., for this PR, we could say that there was a mismatch in adding expanded bf16 dtype support for oneDNN rewrite: oneDNN custom calls aren't used in AOT mode so the expanded CpuFloatSupport (which is specific to oneDNN custom calls) shouldn't be used in AOT mode too.

xla/service/cpu/cpu_compiler.cc
yimeisun123: Thanks for this info. Updated with moving the onednn_bf16_support under the guard.

yimeisun123:
Updated description field too. Please let me know if additional things need to be updated. Thanks for your review and suggestions.

penpornk (approved):
Thank you for the fix!

PR #10636, opened 2024-03-16
Offloading 1/3: Add annotation for copy-start/copy-done

zhenying-liu:
Add stream id in the backend_config of copy-start instruction. The stream id is obtained from hlo_query::NextChannelId(). 
The corresponding copy-done instruction which is the use of copy-start instruction will be traversed and added the stream id in the backend_config too. This part is automatically done by the subsequent AnnotateStreamAttributesForUsers() existing in the function.
The bool data member copy_start_done_ is used to differentiate copy-start/copy-done from other collective instructions and go through two different paths.
https://github.com/openxla/xla/pull/10450 is split and the current PR is the first 1 out of 3 PRs.

### DIFF
--- a/xla/service/gpu/stream_attribute_annotator.cc
+++ b/xla/service/gpu/stream_attribute_annotator.cc
@@ -25,7 +25,7 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_instruction.h"
 #include "xla/hlo/ir/hlo_module.h"
 #include "xla/hlo/ir/hlo_opcode.h"
-#include "xla/hlo/utils/hlo_query.h",
+#include "xla/hlo/utils/hlo_query.h"
 #include "xla/service/gpu/backend_configs.pb.h"
 #include "xla/service/gpu/thunk.h"
 #include "xla/statusor.h"
@@ -41,15 +41,12 @@ namespace {
 bool IsOnlyRootNonDefaultStream(HloComputation* computation) {
   HloInstruction* root = computation->root_instruction();
   auto root_gpu_config = root->backend_config<GpuBackendConfig>();
-  // Disable the annotation if its root is copy-start
-  if (!root_gpu_config.ok() || root->opcode() == HloOpcode::kTuple ||
-      root->opcode() == HloOpcode::kCopyStart) {
+  if (!root_gpu_config.ok() || root->opcode() == HloOpcode::kTuple) {
     return false;
   }
   int64_t root_stream_id = root_gpu_config->operation_queue_id();
   VLOG(2) << "Found fusion computation's root stream id to be "
           << root_stream_id;
-
   if (root_stream_id == Thunk::kDefaultExecutionStreamId.value()) {
     return false;
   }
@@ -92,11 +89,16 @@ absl::StatusOr<bool> AnnotateStreamAttributesForInstruction(
 }
 
 absl::StatusOr<bool> AnnotateStreamAttributesForCopyStart(
-    HloInstruction* instr, int64_t channel_id) {
-  GpuBackendConfig gpu_backend_config;
-  gpu_backend_config.set_operation_queue_id(channel_id);
+    HloInstruction* instr, int64_t channel_id,
+    GpuBackendConfig& instr_gpu_config) {
+  // Do nothing if copy-start has already been annotated
+  if (instr_gpu_config.operation_queue_id() !=
+      Thunk::kDefaultExecutionStreamId.value()) {
+    return false;
+  }
+  instr_gpu_config.set_operation_queue_id(channel_id);
+  TF_RETURN_IF_ERROR(instr->set_backend_config(instr_gpu_config));
   VLOG(3) << "Add copy-start's backend config: " << channel_id;
-  TF_RETURN_IF_ERROR(instr->set_backend_config(gpu_backend_config));
   return true;
 }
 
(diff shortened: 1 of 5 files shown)


xla/service/gpu/stream_attribute_annotator.h
```diff
@@ -45,6 +45,8 @@ namespace xla::gpu {
 
 class StreamAttributeAnnotator : public HloModulePass {
  public:
+  explicit StreamAttributeAnnotator(bool copy_start_done = false)
```
cheshire, line 48: Nit: add a description of what the var is doing.

xla/service/gpu/stream_attribute_annotator.cc
```diff
@@ -125,20 +126,31 @@ absl::StatusOr<bool> StreamAttributeAnnotator::Run(
...
       }
-      // For fusion instruction, only annotate
-      // when the root of fusion is a single instruction
-      // running on non-default stream.
-      if (instr->opcode() == HloOpcode::kFusion) {
-        TF_ASSIGN_OR_RETURN(bool comp_result,
-                            AnnotateStreamAttributesForInstruction(
-                                instr, instr_gpu_config.value()));
-        changed |= comp_result;
+      if (!copy_start_done_) {
```
cheshire, line 136: NitL 4 indentations get hard to read, maybe extract into a function?

cheshire (changes requested):
Is it possible to add tests?

xla/service/gpu/stream_attribute_annotator.h
zhenying-liu: > Nit: add a description of what the var is doing.

Done.

xla/service/gpu/stream_attribute_annotator.cc
zhenying-liu: > NitL 4 indentations get hard to read, maybe extract into a function?

Done in a function.

xla/service/gpu/stream_attribute_annotator.h
```diff
@@ -45,6 +45,11 @@ namespace xla::gpu {
 
 class StreamAttributeAnnotator : public HloModulePass {
  public:
+  // copy_start_ is used to differentiate copy-start from
+  // other instructions as it doesn't need the annotations to split into
+  // AsyncStart and AsyncDone: copy-done is available already.
+  explicit StreamAttributeAnnotator(bool copy_start = false)
```
cheshire, line 51: Sorry I still don't follow. This is invoked from gpu_compiler.cc with copy_start set to `true`. Is it ever set to `false`? When do we invoke this with `true` and when do we invoke this with `false`?

cheshire (changes requested):

xla/service/gpu/stream_attribute_annotator.h
zhenying-liu: > This is invoked from gpu_compiler.cc with copy_start set to true. 
Yes, it needs to happen after HLO rematerialization pass in RunPostSchedulingPipelines() because rematerialization can create copy-start and StreamAttributeAnnotator adds annotation to copy-start when copy_start set to true.
> Is it ever set to false? When do we invoke this with true and when do we invoke this with false?
It is set to false by default in the existing code: https://github.com/openxla/xla/pull/10636/files#diff-fc02eb6aea06ad0d72011e9a64da28e8c6fae000a9e0cab9b15b40b4e914af4aL1201. It's called in OptimizeHloModule(). It's too early for us as rematerialization is not called yet and copy-start is not created.
So StreamAttributeAnnotator is called twice right now. This way, I don't have to change the original invocation. Including @Tixxx into the loop as he created this pass.

xla/service/gpu/stream_attribute_annotator.cc
```diff
@@ -140,6 +153,13 @@ absl::StatusOr<bool> StreamAttributeAnnotator::Run(
                                 instr, instr_gpu_config.value()));
         changed |= comp_result;
       }
+      if (instr->opcode() == HloOpcode::kCopyStart) {
```
cheshire, line 156: Nit: `else if`

xla/service/gpu/stream_attribute_annotator.cc
```diff
@@ -88,6 +91,15 @@ absl::StatusOr<bool> AnnotateStreamAttributesForInstruction(
   return true;
 }
 
+absl::StatusOr<bool> AnnotateStreamAttributesForCopyStart(
+    HloInstruction* instr, int64_t channel_id) {
+  GpuBackendConfig gpu_backend_config;
```
cheshire, line 96: Are we fine with throwing out existing `GpuBackendConfig`? What if some values are already there?

cheshire (changes requested):

xla/service/gpu/stream_attribute_annotator.cc
```diff
@@ -40,7 +41,9 @@ namespace {
 bool IsOnlyRootNonDefaultStream(HloComputation* computation) {
```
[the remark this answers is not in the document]
zhenying-liu: It doesn't need separate handling for copy-start.

xla/service/gpu/stream_attribute_annotator.cc
zhenying-liu: Yes. Using GpuBackendConfig from the instruction. If some values are already there, return false to make no changes.

xla/service/gpu/stream_attribute_annotator.cc
```diff
@@ -88,6 +91,15 @@ absl::StatusOr<bool> AnnotateStreamAttributesForInstruction(
   return true;
 }
 
+absl::StatusOr<bool> AnnotateStreamAttributesForCopyStart(
+    HloInstruction* instr, int64_t channel_id) {
+  GpuBackendConfig gpu_backend_config;
+  gpu_backend_config.set_operation_queue_id(channel_id);
+  VLOG(3) << "Add copy-start's backend config: " << channel_id;
+  TF_RETURN_IF_ERROR(instr->set_backend_config(gpu_backend_config));
+  return true;
```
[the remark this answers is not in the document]
zhenying-liu: Modified to return false if it's already annotated. Otherwise annotate and return true.

xla/service/gpu/stream_attribute_annotator.cc
zhenying-liu: Done.

cheshire (approved):

PR #10642, opened 2024-03-17
Make configure.py command visible

prrathi:


reedwm (approved):
Good catch!

ddunl:
Sorry for the delay here, our markdown formatter does something extremely surprising with this change, so having to patch it myself on our end. The patched version should merge ~soon (and will be attributed to you)

PR #10649, opened 2024-03-18
[ROCm] Triton in XLA for ROCm - ir_emitter_triton related changes.

zoranjovanovic-ns:
Second commit of the series for enabling Triton in XLA for ROCm.

### DIFF
--- a/xla/service/gpu/ir_emitter_triton.cc
+++ b/xla/service/gpu/ir_emitter_triton.cc
@@ -30,8 +30,6 @@ limitations under the License.
 #include <variant>
 #include <vector>
 
-#include "nvidia/include/NVGPUToLLVM/NVGPUToLLVMPass.h"
-#include "nvidia/include/TritonNVIDIAGPUToLLVM/Passes.h"
 #include "absl/algorithm/container.h"
 #include "absl/container/flat_hash_map.h"
 #include "absl/container/flat_hash_set.h"
@@ -54,6 +52,7 @@ limitations under the License.
 #include "llvm/TargetParser/Triple.h"
 #include "mlir/Conversion/AffineToStandard/AffineToStandard.h"  // from @llvm-project
 #include "mlir/Conversion/ArithToLLVM/ArithToLLVM.h"  // from @llvm-project
+#include "mlir/Conversion/ControlFlowToLLVM/ControlFlowToLLVM.h" // from @llvm-project
 #include "mlir/Conversion/IndexToLLVM/IndexToLLVM.h"  // from @llvm-project
 #include "mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h"  // from @llvm-project
 #include "mlir/Dialect/Affine/IR/AffineOps.h"  // from @llvm-project
@@ -90,6 +89,7 @@ limitations under the License.
 #include "mlir/Target/LLVMIR/Dialect/Builtin/BuiltinToLLVMIRTranslation.h"  // from @llvm-project
 #include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"  // from @llvm-project
 #include "mlir/Target/LLVMIR/Dialect/NVVM/NVVMToLLVMIRTranslation.h"  // from @llvm-project
+#include "mlir/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.h"  // from @llvm-project
 #include "mlir/Target/LLVMIR/Export.h"  // from @llvm-project
 #include "mlir/Transforms/Passes.h"  // from @llvm-project
 #include "xla/autotuning.pb.h"
@@ -144,6 +144,7 @@ limitations under the License.
 #include "triton/Dialect/Triton/Transforms/Passes.h"
 #include "triton/Dialect/TritonGPU/Transforms/Passes.h"
 #include "triton/Dialect/TritonNvidiaGPU/Transforms/Passes.h"
+
 
 namespace xla {
 namespace gpu {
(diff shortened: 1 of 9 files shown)


xla/service/gpu/ir_emitter_triton.cc
```diff
@@ -122,14 +126,21 @@ limitations under the License.
 #include "tsl/platform/path.h"
 #include "tsl/platform/status.h"
 #include "tsl/platform/statusor.h"
+#ifdef TENSORFLOW_USE_ROCM
```
ddunl, line 129: I think it'd be good if possible to minimize the number of ifdefs, so maybe ifdefs used everywhere at the top, then ```
#ifdef GOOGLE_CUDA
<includes> 
#ifdef TENSORFLOW_USE_ROCM
<includes>
``` or similar.

File: xla/service/gpu/ir_emitter_triton.cc [hunk 2]
```diff
@@ -744,11 +758,23 @@ absl::StatusOr<Value> EmitScope(
...
-    mlir::OpPassManager& pm, const se::CudaComputeCapability& cc,
+    mlir::OpPassManager& pm, const se::GpuComputeCapability& cc,
     const TritonGemmConfig& config,
     mt::nvidia_gpu::ClusterInfo& out_cluster_info) {
-  const int ccAsInt = cc.major * 10 + cc.minor;
+#else
+absl::Status CreateTritonPipeline(
+    mlir::OpPassManager& pm, const se::GpuComputeCapability& cc,
+    const TritonGemmConfig& config) {
+#endif
```
tdanyluk, line 770: In general I would recommend to reduce the number of ifdefs inside the code, to close to 0.
Could we have some kind of abstraction, like have a 
Something class, and a derived CudaSomething and RocmSomething and just use a runtime if based on whether the se::GpuComputeCapability is cuda or rocm?

tdanyluk (changes requested):

xla/service/gpu/ir_emitter_triton.cc
tdanyluk: I talked with my colleagues and a class is not really needed.
We could use something like this:

something.h:
```
int something();
```

something_cuda.cc:
```
int something() {
 // implemented with cuda
}
```

something_rocm.cc:
```
int something() {
 // implemented with rocm
}
```
And then build either something_rocm.cc or something_cuda.cc with a select.
So ir_emitter_triton.cc would remain mostly device vendor independent and just include the shared "something.h'.

zoranjovanovic-ns: Thank you for the proposal and clarification. I will change the implementation (based on proposal) and update PR.

tdanyluk: thank you!

xla/service/gpu/ir_emitter_triton.cc
```diff
@@ -408,12 +418,16 @@ Value EmitElementwise(ImplicitLocOpBuilder& b, absl::string_view libdevice_path,
...
-          inputs[0].getType(), inputs, "libdevice", libdevice_path,
-          ObtainDeviceFunctionName(dev_fn_id.value(),
-                                   hlo.shape().element_type(),
-                                   llvm::Triple("nvptx64-unknown-unknown")),
-          /*pure=*/true);
+        llvm::Triple triple("nvptx64-unknown-unknown");
+        if (std::holds_alternative<se::CudaComputeCapability>
+                  (device_info.gpu_compute_capability())) {
+          triple.setTriple("amdgcn-unknown-unknown");
+        }
```
tdanyluk, line 425: Is there an error here? It looks like you set an AMD triple when the GPU is NVIDIA.

tdanyluk (changes requested):
Thanks, it looks really good after the refactor.

xla/service/gpu/ir_emitter_triton.cc
zoranjovanovic-ns: Yes, I am about to push fixed version. Sorry for inconvenience.

xla/service/gpu/ir_emitter_triton.cc
```diff
@@ -148,6 +149,15 @@ using ::mlir::Type;
...
 
+
+std::string GetLibdevicePath(const HloModuleConfig& hlo_config,
+                             const se::DeviceDescription& device_info);
+
+absl::Status CreateTritonPipeline(
+    mlir::OpPassManager& pm, const se::GpuComputeCapability& cc,
+    const TritonGemmConfig& config,
+    mt::nvidia_gpu::ClusterInfo& out_cluster_info);
+
```
tdanyluk, line 160: It's an interesting choice that these are not in the header.
I mean I get the point that these are kindof private to ir_emitter_triton, but still, we usually put function declarations in headers.
If it's not too difficult, please try to do that.

tdanyluk (changes requested):
Hi. Sorry for getting back to you once more, I just noticed some things.

xla/service/gpu/ir_emitter_triton.cc
zoranjovanovic-ns: OK, I will made changes that you have suggested.

tdanyluk (approved):
Thanks again, looks great!

xla/service/gpu/ir_emitter_triton_cuda.cc
```diff
@@ -0,0 +1,126 @@
...
+}
+
+std::string GetLibdevicePath(const HloModuleConfig& hlo_config,
+                             const se::DeviceDescription& device_info) {
+  return nvptx::LibDevicePath(
+      hlo_config.debug_options().xla_gpu_cuda_data_dir());
+}
+
+}  // namespace gpu
+}  // namespace xla
```
ddunl, line 126: Could you add a newline at the end of this file (also `ir_emitter_triton_rocm.cc`)

zoranjovanovic-ns: Yes, I will do that.

ddunl (approved):
Thanks!!

i-chaochen:
Hi @xla-rotation  this PR is already in the copybara-service pipeline, but it has conflicts now. Shall we resolve and rebase it? or we can leave it?

ddunl:
I think you can leave it, we have to patch because Copybara behaved strangely on this one so I have to investigate that, fix and patch

zoranjovanovic-ns:
> I think you can leave it, we have to patch because Copybara behaved strangely on this one so we have to investigate that, fix and patch

Hi @xla-rotation, should I take some additional actions in order to have this merged?

penpornk:
@zoranjovanovic-ns Sorry for the delay! Could you please help resolve merge conflicts?

zoranjovanovic-ns:
Yes, I will rebase and push new version.

penpornk:
@zoranjovanovic-ns Thank you for checking and for merging conflicts! We are applying some more fixes to pass internal tests. I'll let you know if we need more help. :)

PR #10653, opened 2024-03-18
[ROCM] fixing build brake and temporarily disable delayed kernel for rocm

pemeliya:
This PR fixes the ongoing build brake after https://github.com/openxla/xla/commit/28cfdcd27859e79cf10bd92779bc29f6404c79a0

and also I temporarily disable delayed kernel for ROCM since it significantly increases running times (to be investigated).

@xla-rotation: would you please have a look ?

### DIFF
--- a/xla/stream_executor/gpu/BUILD
+++ b/xla/stream_executor/gpu/BUILD
@@ -330,6 +330,9 @@ gpu_only_cc_library(
     name = "gpu_timer",
     srcs = ["gpu_timer.cc"],
     hdrs = ["gpu_timer.h"],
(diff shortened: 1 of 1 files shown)


beckerhe (approved):
Looks good to me. But note that the change with the delay kernel get rolled back, so we won't be able to merge this before we roll forward the change - which we intent to do, so this PR will still be needed.

pemeliya:
ok I see, thanks for the info.. then let put this PR on hold until GpuTimer changes are merged

akuegel:
This landed as part of the roll forward: https://github.com/openxla/xla/commit/a240f804899c6f334a45ffeaf6444c7d0d1af767#diff-a2b6eae7abf00c57b546d42c4317e1a37d5165e0b6c375e58444ad054a70b2ca

PR #10668, opened 2024-03-18
[XLA:GPU] Fix cuDNN FMHA rewriter sequence length checks and Bump up minimum flash attn cuDNN version to 8.9.4

Cjkkkk:
* If only seqlen_q or seqlen_kv is larger than 512, it will be lowered to fused attn instead of flash attn. This is incorrect as fused attn does not support this case. Add checks to lower to flash attn if one seqlen is larger than 512.
* While fixing the seqlen checks, also relax the seqlen/head dim constraints to support any seqlen % 2 == 0, head_dim <= 128 and head_dim % 8 = 0. (Since 8.9.4).
* Consolidate multiple get seqlen/head dim from fused attn/flash attn checks into one `getBHSD` function and rewrite both fused attn/flash attn checks to make it easier to understand.
* Bump up minimum cuDNN version require for flash attn from 8.9.3 to 8.9.4. (More seqlen/head dim support and cross attn support).
* Remove cross attn checks from rewriter and rewriter test since it is now default supported with 8.9.4.
* Add one testcase to cover the first bullet point.

### DIFF
--- a/xla/service/gpu/cudnn_fused_mha_rewriter.cc
+++ b/xla/service/gpu/cudnn_fused_mha_rewriter.cc
@@ -324,10 +324,17 @@ std::vector<int64_t> GetDimensionVector(absl::Span<const int64_t> dimensions,
   return vec;
 }
 
-absl::StatusOr<std::vector<int64_t>> GetBHSD(HloInstruction* bmm_1,
-                                             HloInstruction* bmm_2,
-                                             bool need_canonicalization) {
-  // get bhsd from bmm1
+struct QKVLayout {
+  int64_t batch;
+  int64_t num_heads;
+  int64_t seqlen_q;
+  int64_t seqlen_kv;
+  int64_t hidden_dim;
+};
+
+absl::StatusOr<std::optional<QKVLayout>> GetQKVLayout(
+    HloInstruction* bmm_1, HloInstruction* bmm_2, bool need_canonicalization) {
+  // get layout from bmm1
   const DotDimensionNumbers& bmm1_dnums = bmm_1->dot_dimension_numbers();
   TF_ASSIGN_OR_RETURN(
       std::vector<int64_t> bmm1_s_q_dims,
@@ -360,7 +367,7 @@ absl::StatusOr<std::vector<int64_t>> GetBHSD(HloInstruction* bmm_1,
   TF_RET_CHECK(bmm1_s_kv.size() == 1);
   TF_RET_CHECK(bmm1_d.size() == 1);
 
-  // get bhsd from bmm2
+  // get layout from bmm2
   const DotDimensionNumbers& bmm2_dnums = bmm_2->dot_dimension_numbers();
   TF_ASSIGN_OR_RETURN(
       std::vector<int64_t> bmm2_lhs_non_contracting_dims,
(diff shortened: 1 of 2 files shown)


xla/service/gpu/cudnn_fused_mha_rewriter.cc
```diff
@@ -337,150 +324,128 @@ std::vector<int64_t> GetDimensionVector(absl::Span<const int64_t> dimensions,
...
-      GetDimensionVector(bmm_1->operand(0)->shape().dimensions(), seq_q_dims);
-
-  std::vector<int64_t> seq_k =
-      GetDimensionVector(bmm_1->operand(1)->shape().dimensions(), seq_k_dims);
+  // otherwise check if it is supported by regular attention
+  int64_t s_q = bhsd[2];
+  int64_t s_kv = bhsd[3];
+  int64_t hidden_dim = bhsd[4];
+  bool is_seqlen_supported = (s_q <= 512 && s_kv <= 512) &&
+                             (!is_training || s_q % 64 == 0 && s_kv % 64 == 0);
```
akuegel, line 425: Please put () around the extra && condition. This would trigger a compiler warning otherwise which we treat internally as error.

xla/service/gpu/cudnn_fused_mha_rewriter.cc
```diff
@@ -337,150 +324,128 @@ std::vector<int64_t> GetDimensionVector(absl::Span<const int64_t> dimensions,
...
-  // flash attention requires cuDNN 8.9.3 to run non-fused QKV
+absl::StatusOr<bool> IsFlashAttention(
+    std::vector<int64_t>& bhsd, bool is_training,
+    stream_executor::CudaComputeCapability cc,
+    stream_executor::dnn::VersionInfo cudnn_version) {
+  int64_t s_q = bhsd[2];
+  int64_t s_kv = bhsd[3];
+  int64_t hidden_dim = bhsd[4];
+  bool is_seqlen_supported = (s_q > 512 || s_kv > 512) &&
+                             (!is_training || s_q % 2 == 0 && s_kv % 2 == 0);
```
akuegel, line 439: Same here, ```|| (s_q % 2 == 0 && s_kv % 2 == 0)```

xla/service/gpu/cudnn_fused_mha_rewriter.cc
```diff
@@ -1239,10 +1204,18 @@ absl::StatusOr<bool> IsMHABlockSupported(
     return false;
   }
 
+  // get batch/num heads/sequence length/hidden dim from bmm1 and bmm2
+  // also make sure they are the same between bmm1 and bmm2
+  TF_ASSIGN_OR_RETURN(std::vector<int64_t> bhsd,
+                      GetBHSD(bmm_1, bmm_2, need_canonicalization));
+  if (bhsd.size() == 0) {
```
akuegel, line 1211: nit: check bhsd.empty() instead of size() == 0

xla/service/gpu/cudnn_fused_mha_rewriter.cc
```diff
@@ -337,150 +324,128 @@ std::vector<int64_t> GetDimensionVector(absl::Span<const int64_t> dimensions,
   return vec;
 }
 
-absl::StatusOr<bool> IsSupportedBMM1(const HloInstruction* bmm_1,
-                                     bool is_training) {
-  const DotDimensionNumbers& dot_dims_bmm1 = bmm_1->dot_dimension_numbers();
+absl::StatusOr<std::vector<int64_t>> GetBHSD(HloInstruction* bmm_1,
```
akuegel, line 327: Maybe return a struct so that you can name the different values that are returned? I think it would make it a bit easier to read the code.

akuegel (changes requested):

xla/service/gpu/cudnn_fused_mha_rewriter.cc
Cjkkkk: I created a QKVLayout struct and use std::optional for checking return status.

akuegel (approved):
Thank you

akuegel:
We have hit an error in a jax test:

```
Execution of replica 0 failed: INTERNAL: s_kv not a multiple of 64 or d not a multiple of 64 is not supported with cudnn version below 8.9.6
```

Does this make sense to you? I guess we need additional checks?

akuegel:
@xla-rotation please find a different reviewer for this PR, I will be out of office next week

Cjkkkk:
Confirmed and reproducible with 8.9.5.

reedwm:
@Cjkkkk the next steps are for you to fix the issue with cuDNN versions below 8.9.6, right?

Cjkkkk:
> @Cjkkkk the next steps are for you to fix the issue with cuDNN versions below 8.9.6, right?

Correct.

Cjkkkk:
@reedwm Hi Reed, I added a check and test to cover the case where head dim is not 64 or 128 for cuDNN < 8.9.6. Let me know if this fixed the internal test failure. Thanks!

reedwm (approved):
Approving to start merge process. I did not look at the full PR, but only changes since @akuegel last approved.

Cjkkkk:
Thanks, seems like another internal failure. @reedwm

PR #10301, opened 2024-03-06
[XLA:CPU][oneDNN] Convolution XLA HLO Pattern Matcher with oneDNN custom call rewrite

akhilgoe:
This PR enables oneDNN library call for the matched XLA HLO Convolution pattern through custom_call instruction. In particular, this PR:

1. Adds oneDNN convolution rewriter pass that will rewrite HLO Convolution to oneDNN Convolution.
2. Refactors backend config to enhance code reusability.
3. Adds a convolution test file to verify rewrite and execution result

PR #10687, opened 2024-03-19
[CPU] Add SimplifyFPConversions only if xla_allow_excess_precision

apivovarov:
Several weeks ago it was a change which enables "simplify-fp-conversions" pass in cpu_compiler.cc for intel cpus unconditionally.

[PR-8402](https://github.com/openxla/xla/pull/8402) - [XLA:CPU] [oneDNN] Enable Dot op (MatMul) in BF16 Type

I noticed the following issue with having "simplify-fp-conversions" pass in cpu_compiler.cc enabled unconditionally.

My model uses bf16 operators (e.g. convolution). I want to jit compile and run it on CPU preserving intermediate bf16 accuracy.

Cpu compiler uses`float-normalization-bf16` pass which converts bf16 convolution to f32_convolution + convert_to_bf16 + convert_to_f32. (because typical cpu does not support bf16 computation) 

Cpu compiler (on XEON) also uses `simplify-fp-conversions` pass which simplifies `f32_convolution + convert_to_bf16 + convert_to_f32` to just `f32_convolution`.

As the result - the whole model was converted to f32 precision internally and conversion to bf16 happens only at the very end.

In some cases we want to execute bf16 model on CPU but get results with accuracy similar to the case when it is executed on bf16 hardware.

To control the accuracy we can use debug_option `xla_allow_excess_precision`
By default it is true - hence, `simplify-fp-conversions` pass is enabled.

If we need to emulate bf16 computation on intel cpu we can set `XLA_FLAGS="--xla_allow_excess_precision=false"` - in this case `simplify-fp-conversions` will not be added to cpu_compiler pipeline. f32 ops results will be converted to bf16 immediately. This will preserve bf16 accuracy internally.

[gpu_compiler.cc](https://github.com/openxla/xla/blob/main/xla/service/gpu/gpu_compiler.cc#L1359) already enables `SimplifyFPConversions` pass only if `debug_options.xla_allow_excess_precision()` is true.

### DIFF
--- a/xla/service/cpu/cpu_compiler.cc
+++ b/xla/service/cpu/cpu_compiler.cc
@@ -905,14 +905,20 @@ Status CpuCompiler::RunHloPassesAfterLayoutAssn(
 #if defined(INTEL_MKL) && defined(ENABLE_ONEDNN_V3)
   // AOT compiled code runs in single thread.
   if (!is_aot_compile) {
+    auto debug_options = module->config().debug_options();
     // Run SimplifyFPConversions pass to simplify the BF16 pattern and make it
     // easier to match.
-    pipeline.AddPass<SimplifyFPConversions>();
+    // Remove `f32 -> bf16 -> f32` casts inserted by bf16 normalization.
+    if (debug_options.xla_allow_excess_precision()) {
+      pipeline.AddPass<SimplifyFPConversions>();
+    }
     pipeline.AddPass<OneDnnMatMulRewriter>(max_parallelism,
                                            compile_options.thread_pool);
     // Run SimplifyFPConversions pass again to remove redundant Convert ops
     // that may exist as a result of running OneDnnMatMulRewriter pass.
-    pipeline.AddPass<SimplifyFPConversions>();
+    if (debug_options.xla_allow_excess_precision()) {
+      pipeline.AddPass<SimplifyFPConversions>();
+    }
   }
 #endif  // INTEL_MKL && ENABLE_ONEDNN_V3
 


penpornk (approved):
Thank you very much for the PR and for the detailed explanation! I think this PR is safe to merge now since `xla_allow_excess_precision` is true by default. But let me cc the original authors so they know that the oneDNN matmul pattern match could break when the option is off too.

@TensorFlow-MKL, @mahmoud-abuzaina, @agramesh1: 
I don't think we should keep SimplifyFPConversions running when `xla_allow_excess_precision` is off (because FP32->BF16->FP32 should still give BF16 accuracy, the simplification would lead to FP32 precision). So if this gets in the way with oneDNN Dot op pattern matching, we may need to make the pattern match more resilient.

mahmoud-abuzaina:
@penpornk Thanks for the heads-up. Yes we think this PR is safe to merge.

agramesh1:
As a side note, there is a PR currently under review of adding conv include BF16 support 
https://github.com/openxla/xla/pull/10301

penpornk:
@mahmoud-abuzaina @agramesh1 Thank you for the quick reply!

> As a side note, there is a PR currently under review for adding conv include BF16 support
> https://github.com/openxla/xla/pull/10301

Sorry for the delay on https://github.com/openxla/xla/pull/10301. I'll review it soon.

apivovarov:
I noticed that `OneDnnMatMulRewriter` pass can cause the following issue with some models when jit compile for intel cpu
```
could not create a primitive descriptor for a matmul primitive
zsh: illegal hardware instruction
```
Have you seen such error before? 
@TensorFlow-MKL, @mahmoud-abuzaina, @agramesh1, @yimeisun123, @rfsaliev

mahmoud-abuzaina:
@apivovarov this error may be caused by several reasons. Can we know more about you're trying to run or get a reproducer? Also, did this issue start to happen only after this PR got merged or it is independent?

apivovarov:
independent. old OneDnnRewriter (Dec 4 version) worked fine. After it was refactored to OneDnnMatMulRewriter recently and moved to new place in cpu_compiler.cc couple models from our tests crashed the compilation. I'll try to prepare small model to repro.

PR #10711, opened 2024-03-19
[GPU][NFC] Clarify dependency of cuDNN fusion compilation on stream executor and cuDNN handle.

sergachev:
LocalCuDnnHandle can now be created only via CudnnAccess which ensures that main cuDNN handle is created first. 

DnnGraph methods using cuDNN handle now require DnnSupport&.


ezhulenev (approved):

sergachev:
Addressed test breakage.

ezhulenev (approved):

PR #10712, opened 2024-03-19
add all backend results to AutotuneResult proto

Amir-19:
This PR adds the following functionalities:

- add all backend results to AutotuneResult proto: best time for triton, cublas, and cudnn backends
- add fusion name to autotuning proto

### DIFF
--- a/xla/autotuning.proto
+++ b/xla/autotuning.proto
@@ -38,6 +38,13 @@ message AutotuneResult {
     DISQUALIFIED = 3;
   }
 
+  enum Backend {
+    UNKNOWN_BACKEND = 0;
+    TRITON = 1;
+    CUDNN = 2;
+    CUBLAS = 3;
+  }
+
   message FailureResult {
     FailureKind kind = 1;
     string msg = 2;
@@ -85,6 +92,16 @@ message AutotuneResult {
     int64 num_ctas = 7;
   }
 
+  message Option {
+    google.protobuf.Duration run_time = 1;
+    Backend backend = 2;
+    FailureResult failure = 3;
+  }
+
+  message TriedBackends {
+    repeated Option tried_backends = 1;
+  }
+
   int64 scratch_bytes = 8;
   google.protobuf.Duration run_time = 9;
 
@@ -98,7 +115,9 @@ message AutotuneResult {
     stream_executor.dnn.AlgorithmProto algorithm = 16;
   }
 
-  // Next ID: 17
+  string fusion_name = 18;
+  TriedBackends backends = 19;
+  // Next ID: 20
 }
 
 message AutotuningLog {
--- a/xla/service/gpu/gemm_fusion_autotuner.cc
+++ b/xla/service/gpu/gemm_fusion_autotuner.cc
@@ -994,6 +994,10 @@ absl::StatusOr<AutotuneResult> Execute(const AutotuneConfig& config,
   VLOG(2) << fusion->name() << ": time with cuBLAS: " << cublas_duration;
   AutotuneResult best;
   best.mutable_failure()->set_kind(AutotuneResult::DISQUALIFIED);
+
+  absl::Duration best_duration_cudnn = absl::InfiniteDuration();
+  absl::Duration best_duration_triton = absl::InfiniteDuration();
+
   if (!triton_results.empty()) {
     TF_ASSIGN_OR_RETURN(const AutotuneResult triton_best,
                         PickBestResult(triton_results, root.ToString(),
@@ -1001,6 +1005,8 @@ absl::StatusOr<AutotuneResult> Execute(const AutotuneConfig& config,
     VLOG(2) << "Best time with Triton: "
             << tsl::proto_utils::FromDurationProto(triton_best.run_time());
     best = triton_best;
+    best_duration_triton =
+        tsl::proto_utils::FromDurationProto(triton_best.run_time());
   }
   if (!cudnn_results.empty()) {
     TF_ASSIGN_OR_RETURN(const AutotuneResult cudnn_best,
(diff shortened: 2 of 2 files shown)


xla/autotuning.proto
```diff
@@ -38,6 +38,13 @@ message AutotuneResult {
     DISQUALIFIED = 3;
   }
 
+  enum Backend {
```
cheshire, line 41: But this is already encoded in the oneof key?

xla/autotuning.proto
```diff
@@ -98,7 +115,9 @@ message AutotuneResult {
     stream_executor.dnn.AlgorithmProto algorithm = 16;
   }
 
-  // Next ID: 17
+  string fusion_name = 18;
```
cheshire, line 118: Storing fusion_name is probably a separate change?

cheshire (changes requested):

tdanyluk (commented):
Thanks for this development

xla/autotuning.proto
elfiegg: The context of this development is that the XLA:GPU developers at Nvidia needs debugging information for a tuned fusion - the fusion name, each backend autotuner has tried (triton / cuDNN / cuBLAS) for the fusion and their run_time. As well as the best kernel with the fastest runtime. This helps us understand how each backend performs. 

We believe this would be helpful for upstreaming into XLA.

IIUC, AutotuneResults is a list that presents all tuned fusions. And each tuned fusion is presented by AutotuneResult. We believe this debugging information is tightly attached to each HLO fusion and should stick with AutotuneResult. 

WDYT?

tdanyluk: My point is that the AutotuneResults structure consists of a list of Entries (repeated Entry results) and each entry has one AutotuneResult currently. Instead of putting the results of multiple backends into the same AutotuneResult object, I think that we could add multiple AutotuneResult objects to the Entry (repeated AutotuneResult alternative_results), one for each backend. And keep "result" as the best result.

But I would to check with @cheshire also what does he recommend.

xla/autotuning.proto
Amir-19: do you mean a separate PR or commit?

xla/autotuning.proto
```diff
@@ -85,6 +92,16 @@ message AutotuneResult {
     int64 num_ctas = 7;
   }
 
+  message Option {
```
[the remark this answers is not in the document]
Amir-19: +

xla/autotuning.proto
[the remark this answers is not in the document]
Amir-19: the purpose of storing fusion_name is to have a connection between the result in the proto--time for running a fusion--and the dump for each of the fusions.

cheshire: PR.

xla/service/gpu/gemm_fusion_autotuner.cc
```diff
@@ -994,25 +994,54 @@ absl::StatusOr<AutotuneResult> Execute(const AutotuneConfig& config,
   VLOG(2) << fusion->name() << ": time with cuBLAS: " << cublas_duration;
   AutotuneResult best;
   best.mutable_failure()->set_kind(AutotuneResult::DISQUALIFIED);
+
+  absl::Duration best_duration_cudnn = absl::InfiniteDuration();
```
cheshire, line 998: Why do we need to separately store those best durations? I thought we just store best + all other durations?

xla/autotuning.proto
```diff
@@ -85,6 +92,16 @@ message AutotuneResult {
...
+    google.protobuf.Duration run_time = 1;
+    Backend backend = 2;
+    FailureResult failure = 3;
+  }
+
+  message TriedBackends {
```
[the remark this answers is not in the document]
cheshire: Also not sure why this separate message is required.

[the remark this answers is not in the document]
cheshire: Isn't tried_backends implicit in all stored results?

xla/autotuning.proto
cheshire: Isn't this implicitly selected based on the type of the value of `oneof key`?

cheshire (changes requested):

xla/service/gpu/gemm_fusion_autotuner.cc
Amir-19: the proto, in the current state, only stores information about the best backend. the changes we want to introduce, store best time for each of the backends and we can use the information to compare backends for each fusion.

cheshire: That seems redundant if you store *all* the intermediate results? You can always derive best-for-each from there?

xla/autotuning.proto
elfiegg: Yeah each HLO/fusion has a selected best backend result implicitly encoded by oneof Key. We want to include other not-selected backend information for debugging purpose and hope to get the info dumped along with the autotune result.

xla/service/gpu/gemm_fusion_autotuner.cc
Amir-19: I don't think we store all intermediate results--not in the proto certianly. Storing best time for each backend would give an overview of the performance of each backend for that particular fusion which is useful for development purposes as well.

xla/autotuning.proto
cheshire: Right, I understand and agree with the rationale, I'm just saying that I think you're adding redundant data which is already there on top of the required change. Also : didn't you do stochastic rounding with us before ? :P

elfiegg: Hehe hello George :) I'm making a mess somewhere else now! 
So IIUC, you're also suggesting we should make a repeated AutotuneResult list in the Entry of AutotuneResults? Maybe we should have all of your feedback earlier in the loop - next time :)

cheshire: Yes - let's try to go back to the drawing board on what are we aiming to achieve? The `AutotuningLog` already can have all the `AutotuningResult`s. Each autotuning result already know it's backend, based on the `key` value. AutotuningLog for each instruction already stored `instr` representation itself. Do we need any proto changes at all?

tdanyluk: (Hmm I didn't know about `AutotuningLog` when I wrote my comments, maybe that's more fitting than `AutotuneResults`.)

kamaljeeti:
Hi @Amir-19 , any update on this PR? Thanks.

Amir-19:
> Hi @Amir-19 , Gentle reminder!

sorry I must have missed the notification. I need to close this PR, I'm working on a new solution for this which uses AutotuningLog proto instead.

PR #10724, opened 2024-03-20
[NVIDIA GPU] Fix crash in thunk lowering

Tixxx:
send and recv will have tuple as output shape, the tuple contains a token field, we only want to use the memory space of the first element in the tuple which is the actual data, not the token element.


trevor-m (approved):
Nice catch, thanks @Tixxx

cheshire (approved):

jurahul (approved):

PR #10747, opened 2024-03-20
[XLA:CPU][oneDNN][Bugfix] Fix FLOPs count for dot in oneDNN MatMul Rewriter

mdfaijul:
This PR fixes a bug in MAC (multiply-accumulate) / FMA(fused-multiply-add)  count in dot operation when B matrix is transposed.


penpornk (approved):
Thank you for the fix!

mdfaijul:
@penpornk Thanks for approving the PR. I noticed that onednn_softmax_test was failing for timeouts. So tagged as "no_oss".

PR #10763, opened 2024-03-20
[XLA:GPU] Fix cuDNN FMHA fwd scale not passed into cuDNN

Cjkkkk:
* FMHA fwd scale is not passed into cuDNN in pattern bmm1-scale-softmax-(dropout)-bmm2. Work correctly on other patterns (with bias/mask).
* A quick fix is that I simply copy what works in bwd scale setting.
* Weirdly e2e run does not capture this since we do have cases where fwd scale is not 1.

### DIFF
--- a/xla/service/gpu/gpu_fused_mha_runner.cc
+++ b/xla/service/gpu/gpu_fused_mha_runner.cc
@@ -434,19 +434,9 @@ absl::Status RunGpuFMHABackwardImpl(const GpufMHABackwardParams &params,
   config.is_causal_mask = desc.is_causal_mask;
   const CudnnfMHABackendConfig &backend_config = desc.backend_config;
   config.algorithm = se::dnn::AlgorithmDesc(backend_config.algorithm());
-
-  auto assign_scale = [&]() {
-    config.fmha_scale.emplace(backend_config.fmha_scale());
-  };
-
-  auto assign_dropout_rate = [&]() {
-    config.dropout_rate.emplace(backend_config.dropout_rate());
-  };
-
-  auto assign_seed = [&]() { config.seed.emplace(backend_config.seed()); };
-  assign_scale();
-  assign_dropout_rate();
-  assign_seed();
+  config.fmha_scale.emplace(backend_config.fmha_scale());
+  config.dropout_rate.emplace(backend_config.dropout_rate());
+  config.seed.emplace(backend_config.seed());
   return config;
 }
 
@@ -587,19 +577,9 @@ absl::Status RunGpuFMHABackwardImpl(const GpufMHABackwardParams &params,
   config.is_causal_mask = desc.is_causal_mask;
   const CudnnfMHABackendConfig &backend_config = desc.backend_config;
   config.algorithm = se::dnn::AlgorithmDesc(backend_config.algorithm());
-
-  auto assign_scale = [&]() {
-    config.fmha_scale.emplace(backend_config.fmha_scale());
-  };
-
-  auto assign_dropout_rate = [&]() {
-    config.dropout_rate.emplace(backend_config.dropout_rate());
-  };
-
-  auto assign_seed = [&]() { config.seed.emplace(backend_config.seed()); };
-  assign_scale();
-  assign_dropout_rate();
-  assign_seed();
+  config.fmha_scale.emplace(backend_config.fmha_scale());
+  config.dropout_rate.emplace(backend_config.dropout_rate());
+  config.seed.emplace(backend_config.seed());
   return config;
 }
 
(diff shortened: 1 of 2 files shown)


xla/service/gpu/gpu_fused_mha_runner.cc
```diff
@@ -487,9 +435,18 @@ absl::Status RunGpuFMHABackwardImpl(const GpufMHABackwardParams &params,
...
+  auto assign_scale = [&]() {
+    config.fmha_scale.emplace(backend_config.fmha_scale());
+  };
+
+  auto assign_dropout_rate = [&]() {
+    config.dropout_rate.emplace(backend_config.dropout_rate());
+  };
+
+  auto assign_seed = [&]() { config.seed.emplace(backend_config.seed()); };
+  assign_scale();
```
akuegel, line 447: why not just inline these methods?

akuegel (commented):
Could you maybe add an end2end test that covers this?
If I understand the PR description correctly, you would have expected this problem to show up in some existing end2end test. Did you already debug why it doesn't show up? That might be quite interesting and point out another issue.

akuegel:
@xla-rotation I will be on vacation the next week, please re-assign to a different reviewer

Cjkkkk:
> Could you maybe add an end2end test that covers this? If I understand the PR description correctly, you would have expected this problem to show up in some existing end2end test. Did you already debug why it doesn't show up? That might be quite interesting and point out another issue.

Yeah, i initially thought it is because we don't have e2e test cases after I created this PR. At second look, there is test case should cover this, still debug why it is not catched.

cheshire:
+1 on testing, and Adrian's comment.

Cjkkkk:
To summarize, the initial tolerance threshold is too high, it does not trigger the issue. There is one test already in e2e that has bmm1-scale-softmax-bmm2 pattern that can trigger the issue if scale is increased to larger value. After the fix, this test should pass with large scale.

I inlined all calls to set config in both fwd and bwd.

I didn't just directly lower the tolerance threshold cuz there are 2 tests that will fail. I know how to fix this but intend to include that in a separate PR. 
@cheshire

akuegel (approved):
Thanks for adding test coverage by increasing the scale, and a separate PR to make it possible to use a lower tolerance value sounds good to me.

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -2770,7 +2770,7 @@ class FlashAttentionBMMScaleSoftmaxBMM : public MultiHeadedAttentionTest {
       Arg_0.1 = bf16[2,6,1024,64]{3,2,1,0} parameter(0), sharding={replicated}
       Arg_1.2 = bf16[2,6,64,1024]{3,2,1,0} parameter(1), sharding={replicated}
       dot.11 = bf16[2,6,1024,1024]{3,2,1,0} dot(Arg_0.1, Arg_1.2), lhs_batch_dims={0,1}, lhs_contracting_dims={3}, rhs_batch_dims={0,1}, rhs_contracting_dims={2}
-      constant.17 = bf16[] constant(2)
+      constant.17 = bf16[] constant(100)
```
akuegel, line 2773: It looks like with this change, we get a test failure for FlashAttentionBMMScaleSoftmaxBMM.Flash_Attention_Training_BMM1_Softmax_BMM2_BF16

```
Mismatches in shape (bf16[2,6,1024,64], bf16[2,6,1024,64], bf16[2,6,64,1024], bf16[2,6,1024,64]) (3145728 elements):
Array at shape index {2}, 
Mismatch count 46025 (5.8524%) in shape bf16[2,6,64,1024] (786432 elements), abs bound 0.0025, rel bound 1e-05
  actual     0.004272, expected    4.244e-05, index {1,2,17,590}, rel error      100, abs error  0.00424
  actual     -0.00415, expected   -4.125e-05, index {1,1,53,115}, rel error     99.9, abs error  0.00412
  actual     -0.00415, expected   -4.125e-05, index {0,3,57,689}, rel error     99.9, abs error  0.00412
  actual     -0.00415, expected   -4.125e-05, index {0,2,34,235}, rel error     99.9, abs error  0.00412
  actual     -0.00415, expected   -4.125e-05, index {0,2,3,432}, rel error     99.9, abs error  0.00412
```

Cjkkkk: I guess 100 is too large? Maybe we just reduce it to sth like 20 or so.

akuegel: Can you reproduce the test failure? If yes, I guess you can try to find a value that prevents the failure, but still would cause a failure without the fixes from this PR?

Cjkkkk: I rebased and I can reproduce this on A100. Magic number 37 is able to prevent the failure but will fail if the fix is removed.

akuegel (approved):
Thank you

PR #10788, opened 2024-03-21
[ROCM] fixing build brake: disabling cudnn_fusion_compiler for ROCM

pemeliya:
In this PR I disable yet cudnn_fusion_compiler to fix the build brake
These feature is to be enabled later.

@xla-rotation: woud you have a look please ?


xla/service/gpu/gpu_compiler.cc
```diff
@@ -1982,11 +1982,12 @@ absl::StatusOr<std::unique_ptr<Executable>> GpuCompiler::RunBackend(
     std::unique_ptr<HloModule> module, se::StreamExecutor* stream_exec,
     const CompileOptions& options) {
   Thunk::BinaryMap dnn_compiled_graphs;
+#if GOOGLE_CUDA  // Note: this feature is not yet available on ROCM
```
tdanyluk, line 1985: We should minimize ifdefs in the code.
We could for example try to move this into NvptxCompiler. (Maybe you'll have to add a new virtual method.)

xla/stream_executor/gpu/gpu_timer.cc
```diff
@@ -124,9 +130,13 @@ DeviceMemory<GpuSemaphoreState> GpuTimer::GpuSemaphore::device() {
   if (ShouldLaunchDelayKernel()) {
     // Check the assumption that this device supports unified addressing,
     // otherwise skip the delay kernel
+#if GOOGLE_CUDA
+    auto attr = CU_DEVICE_ATTRIBUTE_UNIFIED_ADDRESSING;
+#elif TENSORFLOW_USE_ROCM
+    auto attr = hipDeviceAttributeUnifiedAddressing;
+#endif
```
tdanyluk, line 137: I suggest to at least move this out to a function so that we don't have inline ifdefs in the code.

xla/stream_executor/gpu/gpu_timer.cc
```diff
@@ -52,12 +52,18 @@ absl::Duration RandomDuration() {
 }
 
 bool ShouldLaunchDelayKernel() {
+#if GOOGLE_CUDA
```
tdanyluk, line 55: Maybe this is fine. But if we could use a normal "if" that would be even better. Not sure if we have a runtime variable somewhere around here, which we could query for cuda vs rocm?

tdanyluk (changes requested):
Hi.
I know that this is meant as a build fix, but actually we would like to reduce the number of ifdefs in the code, so I added some comments.
LMK if it's not possible or you need this really urgently.

xla/stream_executor/gpu/gpu_timer.cc
pemeliya: sure, let me think about it.. Maybe I can get access to device_description of StreamExecutor to check it at runtime

xla/service/gpu/gpu_compiler.cc
pemeliya: yes, I have added a respective virtual function to NvPtxCompiler

xla/stream_executor/gpu/gpu_timer.cc
pemeliya: looks like gpu_timer has been rolled back again, hence this is not relevant for now..

xla/stream_executor/gpu/gpu_timer.cc
pemeliya: not relevant anymore..

tdanyluk (approved):
thank you

PR #10803, opened 2024-03-21
Expose CompiledMemoryStats on CPU and GPU

gspschmid:
[CompiledMemoryStats](https://github.com/openxla/xla/blob/01073cf10b57c64615483545bfb34fb9ae0d3c06/xla/pjrt/pjrt_executable.h#L269-L269) is supposed to summarize the amount of memory required to run a given executable. Buffer-related fields in CompiledMemoryStats are currently missing for both the GPU and the CPU backend. This PR computes the missing fields using the executable's underlying [BufferAssignment](https://github.com/openxla/xla/blob/50c6489cb058881cc65622605c9c55029abebc5b/xla/service/buffer_assignment.h#L491) / buffer allocations.

# Example
```python
import jax
import jax.numpy as jnp

def f(a, b):
    ms = jnp.sum(a, axis=1, keepdims=True) / a.shape[1]
    return a * jax.lax.rsqrt(ms) + b

a0 = jax.random.uniform(jax.random.PRNGKey(0), (128, 128))
b0 = jax.random.uniform(jax.random.PRNGKey(0), (128, 128))
f_jit = jax.jit(f, donate_argnums=(0,))

print(f_jit.lower(a0, b0).compile().memory_analysis())
# CompiledMemoryStats(
#     generated_code_size_in_bytes=5792,
#     argument_size_in_bytes=131072,
#     output_size_in_bytes=65536,
#     alias_size_in_bytes=65536,
#     temp_size_in_bytes=512,
#     host_generated_code_size_in_bytes=0,
#     host_argument_size_in_bytes=0,
#     host_output_size_in_bytes=0,
#     host_alias_size_in_bytes=0,
#     host_temp_size_in_bytes=0)
```
Note the use of `donate_argnums=(0,)` which allows XLA to modify the first input buffer in-place and reuse it as the output. This is reflected in `alias_size_in_bytes` -- without buffer donation that number is zero.

# Open questions
- Does this match the semantics for the TPU backend?
- `dynamic_cast`-ing on `gpu::Executable` ain't great. Do we want `Executable`s to expose their buffer allocations publicly (i.e. GpuExecutable's [GetAllocations](https://github.com/openxla/xla/blob/f6ce6c11ceac904c9c4840fcec5f1ee715291425/xla/service/gpu/gpu_executable.h#L164) but for all Executables)?

### DIFF
--- a/xla/pjrt/cpu/cpu_client.h
+++ b/xla/pjrt/cpu/cpu_client.h
@@ -570,6 +570,8 @@ class TfrtCpuExecutable final : public PjRtLoadedExecutable {
           "cpu_executable_ has no hlo_proto.");
     }
     memory_stats.serialized_hlo_proto = proto->SerializeAsString();
+    memory_stats.PopulateBufferStatsFromAllocations(
+        cpu_executable_.get()->GetAllocations());
     return memory_stats;
   }
 
--- a/xla/pjrt/pjrt_stream_executor_client.h
+++ b/xla/pjrt/pjrt_stream_executor_client.h
@@ -918,6 +918,8 @@ class PjRtStreamExecutorLoadedExecutable : public PjRtLoadedExecutable {
     if (proto != nullptr) {
       memory_stats.serialized_hlo_proto = proto->SerializeAsString();
     }
+    memory_stats.PopulateBufferStatsFromAllocations(
+        executables_[0]->executable()->GetAllocations());
     return memory_stats;
   }
 
--- a/xla/pjrt/pjrt_executable.cc
+++ b/xla/pjrt/pjrt_executable.cc
@@ -216,6 +216,111 @@ absl::StatusOr<ExecuteOptions> ExecuteOptions::FromProto(
   return options;
 }
 
+CompiledMemoryStatsProto CompiledMemoryStats::ToProto() {
+  CompiledMemoryStatsProto proto;
+  proto.set_generated_code_size_in_bytes(generated_code_size_in_bytes);
+  proto.set_argument_size_in_bytes(argument_size_in_bytes);
+  proto.set_output_size_in_bytes(output_size_in_bytes);
+  proto.set_alias_size_in_bytes(alias_size_in_bytes);
+  proto.set_temp_size_in_bytes(temp_size_in_bytes);
+  proto.mutable_hlo_proto()->ParseFromString(serialized_hlo_proto);
+  proto.set_host_generated_code_size_in_bytes(host_generated_code_size_in_bytes);
+  proto.set_host_argument_size_in_bytes(host_argument_size_in_bytes);
+  proto.set_host_output_size_in_bytes(host_output_size_in_bytes);
+  proto.set_host_alias_size_in_bytes(host_alias_size_in_bytes);
+  proto.set_host_temp_size_in_bytes(host_temp_size_in_bytes);
+  return proto;
+}
+
+CompiledMemoryStats CompiledMemoryStats::FromProto(
+    const CompiledMemoryStatsProto& proto) {
+  CompiledMemoryStats stats;
+  stats.generated_code_size_in_bytes = proto.generated_code_size_in_bytes();
+  stats.argument_size_in_bytes = proto.argument_size_in_bytes();
+  stats.output_size_in_bytes = proto.output_size_in_bytes();
+  stats.alias_size_in_bytes = proto.alias_size_in_bytes();
+  stats.temp_size_in_bytes = proto.temp_size_in_bytes();
+  stats.serialized_hlo_proto = proto.hlo_proto().SerializeAsString();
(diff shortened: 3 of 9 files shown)


gspschmid:
cc @yashk2810

cheshire:
Makes sense to me overall, @hawkinsp WDYT?

kamaljeeti:
Hi @hawkinsp , can you look into this once? Thanks.

kamaljeeti:
Hi @hawkinsp , Gentle reminder!

xla/pjrt/pjrt_stream_executor_client.h
```diff
@@ -850,6 +924,17 @@ class PjRtStreamExecutorLoadedExecutable : public PjRtLoadedExecutable {
     if (proto != nullptr) {
       memory_stats.serialized_hlo_proto = proto->SerializeAsString();
     }
+    // TODO: Should this code live in backend-specific subclasses?
+    // (There is TfrtCpuExecutable, but no equivalent for GPUs.)
+    if (auto exec = dynamic_cast<xla::cpu::CpuExecutable*>(
```
hawkinsp, line 929: The `StreamExecutor` code is unused on CPU these days. The implementation for CPU is here: https://github.com/openxla/xla/blob/b3304688637b6cdcd5b016d3044b4707b06b15d7/xla/pjrt/cpu/cpu_client.h#L549

Can you update that as well? I think it just involves moving your `Populate` method into a separate module and calling it from both places.

xla/pjrt/pjrt_stream_executor_client.h
```diff
@@ -850,6 +924,17 @@ class PjRtStreamExecutorLoadedExecutable : public PjRtLoadedExecutable {
     if (proto != nullptr) {
       memory_stats.serialized_hlo_proto = proto->SerializeAsString();
     }
+    // TODO: Should this code live in backend-specific subclasses?
```
hawkinsp, line 927: This code used to be used as part of multiple in-tree backends, but at this point the only in-tree backend is GPUs (NVIDIA and AMD). There's actually a question here I can't answer: is anyone using this code for an out-of-tree backend, other than GPU? It's conceivable that someone is using this as a basis of an out-of-tree PJRT plugin.

I might merge this with the `dynamic_cast` but leave a TODO to specialize this code once this question is resolved.

hawkinsp (changes requested):
This looks fine to me overall.

xla/pjrt/pjrt_stream_executor_client.h
```diff
@@ -838,6 +840,78 @@ class PjRtStreamExecutorLoadedExecutable : public PjRtLoadedExecutable {
...
+        if (is_host) {
+          stats.host_temp_size_in_bytes += size;
+        } else {
+          stats.temp_size_in_bytes += size;
+        }
+      }
+    }
+  }
+
   StatusOr<CompiledMemoryStats> GetCompiledMemoryStats() const override {
```
hawkinsp, line 915: One more nit: please don't place non-trivial function implementations in header files. I acknowledge it was like that when you found it but since we're making it even fancier let's put it in the .cc file.

xla/pjrt/pjrt_stream_executor_client.h
gspschmid: 👍 Added a comment to that effect.

Do I understand correctly that you're suggesting to remove the `dynamic_cast<xla::cpu::CpuExecutable*>` entirely?

xla/pjrt/cpu/cpu_client.h
```diff
@@ -555,6 +556,11 @@ class TfrtCpuExecutable final : public PjRtLoadedExecutable {
           "cpu_executable_ has no hlo_proto.");
     }
     memory_stats.serialized_hlo_proto = proto->SerializeAsString();
+    if (auto exec = dynamic_cast<xla::cpu::CpuExecutable*>(
```
gspschmid, line 559: I suspect this is always the case, but wasn't sure.

gspschmid:
@hawkinsp PTAL and let me know if this is what you intended for the CPU path!

hawkinsp (approved):

dimitar-asenov:
Our internal tests exposed some problems. Essentially we shouldn't use the GPU/CPU and downcast. Instead, let's go with your suggestion to extend the generic `Executable` interface with a virtual `GetAllocations()` which returns an empty span by default.

gspschmid:
Hi @dimitar-asenov, thanks for the heads-up! I updated the PR and added to `Executable` the `GetAllocations()` method that `GpuExecutable` already exposes to `Executable`. Let me know if you'd rather keep those separate and have a new method whose name is explicitly reflects that it is partial (may return empty spans).

gspschmid:
Thanks everyone for helping shepherd this through!

PR #10835, opened 2024-03-22
[GPU] Add support of select operation in cuDNN fusions.

sergachev:


xla/service/gpu/cudnn_fusion_compiler.cc
```diff
@@ -107,6 +108,10 @@ inline std::optional<fe::PointwiseMode_t> GetElementwiseMode(
       return m::POW;
     case HloOpcode::kRsqrt:
       return m::RSQRT;
+#if CUDNN_VERSION >= 90100
```
tdanyluk, line 111: I see that in the test, we have a runtime test: IsAtLeastCuDnn91(), could we use a similar test here to avoid preprocessor checks in the source?

sergachev: This pass aims to be compatible with deviceless AOT in the near future and not use a stream executor like IsAtLeastCuDnn91() does. I don't see target cuDNN version being available in deviceless compilation configuration yet.

The condition can also be removed, it just saves a bit of compilation time by not trying unsupported fusions.

tdanyluk: Thanks for the answer, I guess it's fine then.

tdanyluk (approved):

PR #10858, opened 2024-03-22
[NVIDIA GPU] Enforce first collective permute of collective matmul to always run at the top of loop.

Tixxx:
This is to port collective matmul related changes from this pr(https://github.com/openxla/xla/pull/10316).
We'd need the first collective permute of a collective matmul loop always run at the beginning of the loop.
We set the force_earliest_schedule flag  for that instruction to achieve so.
Force this schedule order would give about additional 3% speedup on top of existing collective matmul for GPT-3 530B model.

### DIFF
--- a/xla/service/gpu/gpu_windowed_einsum_handler.cc
+++ b/xla/service/gpu/gpu_windowed_einsum_handler.cc
@@ -62,6 +62,16 @@ absl::Status UpdateDotAndConsumerConfig(HloInstruction* dot,
   return absl::OkStatus();
 }
 
+absl::Status SetForceDelayForInstruction(HloInstruction* instr,
+                                         bool force_delay) {
+  auto gpu_config = instr->backend_config<gpu::GpuBackendConfig>();
+
+  gpu_config->set_force_earliest_schedule(force_delay);
+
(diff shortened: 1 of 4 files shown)


Tixxx:
@cheshire  Could you take a look at this PR? This is the split PR from https://github.com/openxla/xla/pull/10316

cheshire (approved):

Tixxx:
@cheshire  @ddunl  could you help check if this is failing because of some compiler warnings?

ddunl:
Should be all good, just will take a bit more time to submit as we have some breakage happening right now that will be fixed within ~hours. FYI about warnings though - our CI should check for warnings since https://github.com/openxla/xla/commit/7365a56c008ccdfb55e273a14b9d269c9175254e

Tixxx:
> Should be all good, just will take a bit more time to submit as we have some breakage happening right now that will be fixed within ~hours. FYI about warnings though - our CI should check for warnings since [7365a56](https://github.com/openxla/xla/commit/7365a56c008ccdfb55e273a14b9d269c9175254e)

Ah Nice! That's super useful.

PR #10875, opened 2024-03-23
[XLA:CPU][oneDNN][Bugfix] Fix BF16 oneDNN Layernorm

sachinmuradi:
OneDNN expects scale and bias to be in Float32.
This PR will convert BF16/FP16 scale and weights into Float32 (similar to whats done with mkl layernorm in tensorflow [here ](https://github.com/tensorflow/tensorflow/blob/3a5a9ec33c1194f8fdec2a243714055c0118b44a/tensorflow/core/kernels/mkl/mkl_layer_norm_op.cc#L138)) , without conversion there is accuracy mismatch with oneDNN layernorm.

### DIFF
--- a/xla/tests/onednn_layer_norm_test.cc
+++ b/xla/tests/onednn_layer_norm_test.cc
@@ -141,6 +141,75 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
   MatchOptimizedHlo(layer_norm_module_str, onednn_layer_norm_);
 }
 
+// Test case encountered in models like TFViTForImageClassification in
+// HuggingFace
+// (https://huggingface.co/docs/transformers/model_doc/vit#transformers.TFViTForImageClassification)
+TEST_F(LayerNormTest, LayerNormTest1_BF16) {
+  const char* layer_norm_module_str = R"(
+  HloModule layer_norm.test, entry_computation_layout={(bf16[160,197,768]{2,1,0}, bf16[768]{0}, bf16[768]{0})->bf16[160,197,768]{2,1,0}}
+  region_add {
+    Arg_0.7555 = f32[] parameter(0)
+    Arg_1.7556 = f32[] parameter(1)
+    ROOT add.7557 = f32[] add(Arg_0.7555, Arg_1.7556)
+  }
+  ENTRY main {
+    Arg_0.1 = bf16[160,197,768]{2,1,0} parameter(0), sharding={replicated}
+    Arg_0.2 = bf16[768]{0} parameter(1), sharding={replicated}
+    Arg_0.3 = bf16[768]{0} parameter(2), sharding={replicated}
+    convert.80 = f32[160,197,768]{2,1,0} convert(Arg_0.1)
+    constant.81 = f32[] constant(0)
+    convert.82 = f32[] convert(constant.81)
+    reduce.87 = f32[160,197]{1,0} reduce(convert.80, convert.82), dimensions={2}, to_apply=region_add
+    constant.88 = s32[] constant(768)
+    convert.89 = f32[] convert(constant.88)
+    broadcast.90 = f32[160,197]{1,0} broadcast(convert.89), dimensions={}
+    divide.91 = f32[160,197]{1,0} divide(reduce.87, broadcast.90)
+    convert.92 = bf16[160,197]{1,0} convert(divide.91)
+    reshape.93 = bf16[160,197,1]{2,1,0} reshape(convert.92)
+    reshape.94 = bf16[160,197]{1,0} reshape(reshape.93)
+    broadcast.95 = bf16[160,197,768]{2,1,0} broadcast(reshape.94), dimensions={0,1}
+    subtract.96 = bf16[160,197,768]{2,1,0} subtract(Arg_0.1, broadcast.95)
+    multiply.97 = bf16[160,197,768]{2,1,0} multiply(subtract.96, subtract.96)
+    convert.98 = f32[160,197,768]{2,1,0} convert(multiply.97)
+    constant.99 = f32[] constant(0)
+    convert.100 = f32[] convert(constant.99)
+    reduce.105 = f32[160,197]{1,0} reduce(convert.98, convert.100), dimensions={2}, to_apply=region_add
+    constant.106 = s32[] constant(768)
+    convert.107 = f32[] convert(constant.106)
+    broadcast.108 = f32[160,197]{1,0} broadcast(convert.107), dimensions={}
+    divide.109 = f32[160,197]{1,0} divide(reduce.105, broadcast.108)
+    convert.110 = bf16[160,197]{1,0} convert(divide.109)
+    reshape.111 = bf16[160,197,1]{2,1,0} reshape(convert.110)
+    constant.112 = bf16[] constant(1.002e-12)
+    broadcast.113 = bf16[160,197,1]{2,1,0} broadcast(constant.112), dimensions={}
+    add.114 = bf16[160,197,1]{2,1,0} add(reshape.111, broadcast.113)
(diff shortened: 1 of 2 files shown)


xla/tests/onednn_layer_norm_test.cc
```diff
@@ -141,6 +141,75 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
   MatchOptimizedHlo(layer_norm_module_str, onednn_layer_norm_);
 }
 
+// Test case encountered in models like TFViTForImageClassification in
+// HuggingFace
+// (https://huggingface.co/docs/transformers/model_doc/vit#transformers.TFViTForImageClassification)
+TEST_F(LayerNormTest, LayerNormTest1_BF16) {
+  const char* layer_norm_module_str = R"(
+  HloModule layer_norm.test, entry_computation_layout={(bf16[160,197,768]{2,1,0}, bf16[768]{0}, bf16[768]{0})->bf16[160,197,768]{2,1,0}}
```
penpornk, line 149: To shorten the test and make it easier to read, let's remove `entry_computation_layout`.
```suggestion
  HloModule layer_norm.test
```

xla/tests/onednn_layer_norm_test.cc
```diff
@@ -141,6 +141,75 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
...
+  region_add {
+    Arg_0.7555 = f32[] parameter(0)
+    Arg_1.7556 = f32[] parameter(1)
+    ROOT add.7557 = f32[] add(Arg_0.7555, Arg_1.7556)
+  }
+  ENTRY main {
+    Arg_0.1 = bf16[160,197,768]{2,1,0} parameter(0), sharding={replicated}
+    Arg_0.2 = bf16[768]{0} parameter(1), sharding={replicated}
+    Arg_0.3 = bf16[768]{0} parameter(2), sharding={replicated}
```
penpornk, line 158: Let's also remove all default tensor layouts. (Same for all instructions below.)
```suggestion
    Arg_0.1 = bf16[160,197,768] parameter(0), sharding={replicated}
    Arg_0.2 = bf16[768] parameter(1), sharding={replicated}
    Arg_0.3 = bf16[768] parameter(2), sharding={replicated}
```

xla/tests/onednn_layer_norm_test.cc
```diff
@@ -141,6 +141,75 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
...
+    convert.80 = f32[160,197,768]{2,1,0} convert(Arg_0.1)
+    constant.81 = f32[] constant(0)
+    convert.82 = f32[] convert(constant.81)
```
penpornk, line 161: Is this needed? `constant.81` is already `f32`. Let's simplify everything that is not needed for testing our pattern-matching.

xla/tests/onednn_layer_norm_test.cc
```diff
@@ -141,6 +141,75 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
...
+    reduce.87 = f32[160,197]{1,0} reduce(convert.80, convert.82), dimensions={2}, to_apply=region_add
+    constant.88 = s32[] constant(768)
+    convert.89 = f32[] convert(constant.88)
```
penpornk, line 164: Can we just make `constant.88` f32?
```suggestion
    constant.88 = f32[] constant(768)
```

xla/tests/onednn_layer_norm_test.cc
```diff
@@ -141,6 +141,75 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
...
+    broadcast.90 = f32[160,197]{1,0} broadcast(convert.89), dimensions={}
+    divide.91 = f32[160,197]{1,0} divide(reduce.87, broadcast.90)
+    convert.92 = bf16[160,197]{1,0} convert(divide.91)
+    reshape.93 = bf16[160,197,1]{2,1,0} reshape(convert.92)
+    reshape.94 = bf16[160,197]{1,0} reshape(reshape.93)
```
penpornk, line 169: Why do we need `reshape.93` and `reshape.94`? If this is not part of the tricky pattern we need to recognize, let's simplify it.

penpornk (changes requested):
Thank you for the PR!

> This PR will convert BF16/FP16 scale and weights into Float32 (similar to whats done with mkl layernorm in tensorflow [here ](https://github.com/tensorflow/tensorflow/blob/3a5a9ec33c1194f8fdec2a243714055c0118b44a/tensorflow/core/kernels/mkl/mkl_layer_norm_op.cc#L138)) , without conversion there is accuracy mismatch with oneDNN layernorm.

Could you clarify what you mean by "accuracy mismatch with oneDNN layernorm"? Aren't we calling oneDNN layernorm here? Or do you mean mismatch with the TF oneDNN layernorm op?

xla/tests/onednn_layer_norm_test.cc
sachinmuradi: Done

xla/tests/onednn_layer_norm_test.cc
sachinmuradi: Done

xla/tests/onednn_layer_norm_test.cc
sachinmuradi: Done

xla/tests/onednn_layer_norm_test.cc
sachinmuradi: Modifying constant to fp32 won't capture the exact HLO generated when running models, This test captures the exact case for reduce sum, coming from keras.LayerNormalization. As the divisor constant for mean calculation is always created as s32, we also expect the pattern to be s32 constant ->convert->broadcast.
Also we also observe the same HLO for other tests as well in this file.

xla/tests/onednn_layer_norm_test.cc
sachinmuradi: Same here, since we are trying to capture the exact HLO produced by keras.LayerNormalization, which generates 2 back to back reshapes. So yes, we are trying to match this particular pattern, which is also observed for other test cases in this file.

sachinmuradi:
> Thank you for the PR!
> 
> > This PR will convert BF16/FP16 scale and weights into Float32 (similar to whats done with mkl layernorm in tensorflow [here ](https://github.com/tensorflow/tensorflow/blob/3a5a9ec33c1194f8fdec2a243714055c0118b44a/tensorflow/core/kernels/mkl/mkl_layer_norm_op.cc#L138)) , without conversion there is accuracy mismatch with oneDNN layernorm.
> 
> Could you clarify what you mean by "accuracy mismatch with oneDNN layernorm"? Aren't we calling oneDNN layernorm here? Or do you mean mismatch with the TF oneDNN layernorm op?

Thank you @penpornk for review. By accuracy mismatch, I meant inaccurate results (oneDNN layernorm vs default). Yes, we are calling oneDNN layernorm.
Just to elaborate more ,in case of BF16/FP16 oneDNN LayerNorm, if we see BF16/FP16 scale or bias, we used to pass them as is , that is, BF16/FP16 arguments to oneDNN layernorm, which resulted into inaccurate result. Since oneDNN expects scale and bias for LayerNorm to be always in Float32, we add additional converts for BF16/FP16 scales and biases, so that they will be passed as float32 arguments for oneDNN LayerNorm, and final result will still be BF16/FP16. Hope this clarifies it.

penpornk (approved):
Thank you for the changes and clarifications!

xla/tests/onednn_layer_norm_test.cc
penpornk: I understand. But this means we'll have to update the test each time keras or XLA CPU passes make any changes that affect the HLO though.

PR #10878, opened 2024-03-23
Disable Fusion of FP8 Matrix Bias

philipphack:
Disables the fusion of a matrix bias into a GEMM Custom Call when the bias has an FP8 data type.


philipphack:
CC @reedwm.

cheshire (approved):

PR #10925, opened 2024-03-25
Include FP8 GEMMs in IsCublasGemm

philipphack:
Adds FP8 GEMM Custom Calls to the set of ops considered in function IsCublasGemm.

### DIFF
--- a/xla/service/gpu/cublas_cudnn.cc
+++ b/xla/service/gpu/cublas_cudnn.cc
@@ -30,7 +30,8 @@ namespace xla {
 namespace gpu {
 
 bool IsCublasGemm(const HloInstruction& hlo) {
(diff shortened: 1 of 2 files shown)


reedwm (changes requested):
Remove the now-redundant call to IsCublasLtMatmulF8 here:
https://github.com/openxla/xla/blob/4a268b783a448e462c2c1a0f8750fb808ad0ed22/xla/service/gpu/gemm_algorithm_picker.cc#L448

I looked at all other callers of `IsCublasGemm`, and they should be OK. Some cost-modeling code calls this which doesn't fully take into account the scales and epilogues, but should still be more accurate than before this change.

reedwm (approved):

PR #10930, opened 2024-03-25
[XLA:CPU] Bug fix for enabling F16 support

kanvi-nervana:
This PR fixes a bug for f16 matmul enabling.
Since dot is supported in f16, there is no need for adding an extra convert instruction after fusion similar to the bf16 case.
Also add missed changes in build file.

### DIFF
--- a/xla/service/cpu/onednn_matmul_rewriter.cc
+++ b/xla/service/cpu/onednn_matmul_rewriter.cc
@@ -19,6 +19,8 @@ limitations under the License.
 
 #include "xla/service/cpu/onednn_matmul_rewriter.h"
 
+#include "tsl/platform/logging.h"  // IWYU pragma: keep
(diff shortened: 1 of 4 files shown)


xla/service/cpu/onednn_matmul_rewriter.cc
```diff
@@ -520,26 +515,30 @@ class OneDnnMatMulRewriteVisitor : public DfsHloRewriteVisitor {
       // for bf16 case to avoid datatype mismatch.
```
penpornk, line 515: Please update the comment here as well (that convert for bf16 is optional now).

penpornk (changes requested):
Thank you for the PR!

kanvi-nervana:
@penpornk I have addressed the comment.

penpornk (approved):
Thank you for the change!

PR #10977, opened 2024-03-27
Enable Relu epilogue fusion for cublasLt matmul for training

wenscarl:
To enable ReLU epilogue fusion for CublasLt matmul for training, 2 pair of epilogues: (RELU_AUX, DRELU) and (BIAS_RELU_AUX, DRELU_BGRAD) are added. The RELU_AUX(or BIAS_RELU_AUX) epilogue for the forward matmul outputs a bitmask in the auxiliary buffer for the backward matmul to use. This PR only targets both non-fp8 and fp8 matmuls. @kaixih @philipphack @reedwm 

This PR aims to pattern match the following forward pass:
```
y = matmul1(x1, weight1)
y = relu(y)
y = matmul2(y, weight2)
```
and its corresponding backward pass
```
dy = grad_matmul2(dy, weight1)
dy = drelu(dy)
dy = grad_matmul1(dy, weight2)
```
Here matmul1, matmul2, grad_matmul1 and grad_matmul2 could also be fp8 matmuls and vector bias could be present. In the forward pass, ReLU is fused into matmul1 and in the backward pass, drelu is fused into grad_matmul2.
For the matmul operation with the first matrix having dimensions m=16384, k=n=12288, the performance comparison yields:

With Optimization:
Execution Time: 32.98 ms

Without Optimization:
Execution Time: 33.52 ms

### DIFF
--- a/xla/service/gpu/gemm_algorithm_picker.cc
+++ b/xla/service/gpu/gemm_algorithm_picker.cc
@@ -173,7 +173,7 @@ class GemmAutotuner {
     TF_ASSIGN_OR_RETURN(bool has_aux_input,
                         xla::gpu::gpublas_lt::EpilogueHasAuxiliaryInput(
                             backend_config.epilogue()));
-
+    CHECK(!(has_aux_input && has_aux_output));
     TF_ASSIGN_OR_RETURN(auto epilogue,
                         AsBlasLtEpilogue(backend_config.epilogue()));
 
--- a/xla/service/gpu/gemm_algorithm_picker.cc
+++ b/xla/service/gpu/gemm_algorithm_picker.cc
@@ -75,6 +75,8 @@ absl::StatusOr<BlasLt::Epilogue> AsBlasLtEpilogue(
       return BlasLt::Epilogue::kGELU;
     case GemmBackendConfig::GELU_AUX:
       return BlasLt::Epilogue::kGELUWithAux;
+    case GemmBackendConfig::RELU_AUX:
+      return BlasLt::Epilogue::kReLUWithAux;
     case GemmBackendConfig::BIAS:
       return BlasLt::Epilogue::kBias;
     case GemmBackendConfig::BIAS_RELU:
@@ -83,6 +85,12 @@ absl::StatusOr<BlasLt::Epilogue> AsBlasLtEpilogue(
       return BlasLt::Epilogue::kBiasThenGELU;
     case GemmBackendConfig::BIAS_GELU_AUX:
       return BlasLt::Epilogue::kBiasThenGELUWithAux;
+    case GemmBackendConfig::BIAS_RELU_AUX:
+      return BlasLt::Epilogue::kBiasThenReLUWithAux;
+    case GemmBackendConfig::D_RELU:
+      return BlasLt::Epilogue::kDReLU;
+    case GemmBackendConfig::D_RELU_BGRAD:
+      return BlasLt::Epilogue::kDReLUBGrad;      
     default:
       return Internal("Unsupported Epilogue.");
   }
@@ -161,25 +169,42 @@ class GemmAutotuner {
         bool has_aux_output,
         gpublas_lt::EpilogueHasAuxiliaryOutput(backend_config.epilogue()));
 
+    // auxiliary buffer is only used for either input or output.
+    TF_ASSIGN_OR_RETURN(bool has_aux_input,
+                        xla::gpu::gpublas_lt::EpilogueHasAuxiliaryInput(
+                            backend_config.epilogue()));
+
     TF_ASSIGN_OR_RETURN(auto epilogue,
                         AsBlasLtEpilogue(backend_config.epilogue()));
 
     se::DeviceMemoryBase a_scale_buffer, b_scale_buffer, c_scale_buffer,
         d_scale_buffer, d_amax_buffer, bias_buffer, aux_buffer;
 
+    TF_ASSIGN_OR_RETURN(
+        bool is_forward_epilogue,
+        xla::gpu::gpublas_lt::EpilogueForForward(backend_config.epilogue()));
+
     if (has_vector_bias) {
-      TF_ASSIGN_OR_RETURN(
-          bias_buffer,
-          CreateBuffer(gemm->operand(has_matrix_bias ? 3 : 2)->shape()));
+      Shape bias_buffer_shape =
+          is_forward_epilogue ? gemm->operand(has_matrix_bias ? 3 : 2)->shape()
+                              : gemm->shape().tuple_shapes(1);  // DRELU_BGRAD
+
+      TF_ASSIGN_OR_RETURN(bias_buffer, CreateBuffer(bias_buffer_shape));
     }
+
     if (has_aux_output) {
       TF_ASSIGN_OR_RETURN(aux_buffer,
                           CreateBuffer(gemm->shape().tuple_shapes(1)));
     }
+    if (has_aux_input) {
+      // The last operand is for ReLU bitmask for D_RELU epilogue
+      TF_ASSIGN_OR_RETURN(
+          aux_buffer,
+          CreateBuffer(gemm->operand(gemm->operand_count() - 1)->shape()));
+    }
 
     TF_ASSIGN_OR_RETURN(auto plan,
                         BlasLt::GetMatmulPlan(stream_, gemm_config, epilogue));
-
     TF_ASSIGN_OR_RETURN(auto algorithms, plan->GetAlgorithms());
 
     auto tuned_func = [&](const BlasLt::MatmulAlgorithm& algorithm)
--- a/xla/service/gpu/gemm_rewriter.cc
+++ b/xla/service/gpu/gemm_rewriter.cc
@@ -450,6 +450,13 @@ auto OptionalBitcast(HloInstruction **optional_bitcast, Pattern pattern) {
                                   std::move(pattern));
 }
 
+template <typename Pattern>
+auto OptionalGetTupleElement(HloInstruction **optional_get_tuple,
+                             Pattern pattern) {
+  return m::AnyOf<HloInstruction>(
+      m::GetTupleElement(optional_get_tuple, pattern, 0), std::move(pattern));
+}
+
 // The rewriting proceeds in a bottom-up way:
 //
 // (kDot A B) is rewritten into a (kCustomCall:gemm A B)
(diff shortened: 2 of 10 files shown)


xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+                        m::Slice(m::CustomCall({kCublasLtMatmulCallTarget})),
+                        m::Bitcast(m::CustomCall({kCublasLtMatmulCallTarget})),
+                        m::CustomCall({kCublasLtMatmulCallTarget})),
+                    m::Broadcast(m::ConstantScalar(0))))) {
+        maximum = user;
+        valid_relu_bwd = true;
+        break;
+      }
+    }
+    auto fwd_gemm_num_rows = fwd_gemm->shape().dimensions(0);
```
philipphack, line 1618: Is there a good reason to use auto here?

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+              << "not fused into Custom Call.";
+      return absl::OkStatus();
+    }
+
+    // Rewrite Fwd Relu with ReluAux
+    TF_ASSIGN_OR_RETURN(auto gpu_config,
+                        fwd_gemm->backend_config<GpuBackendConfig>());
+    GemmBackendConfig &config = *gpu_config.mutable_gemm_backend_config();
+    if (config.epilogue() == GemmBackendConfig::DEFAULT) {
+      config.set_epilogue(GemmBackendConfig::RELU_AUX);
```
philipphack, line 1635: Does this work when we fuse a matrix bias? Maybe add a test.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    } else if (config.epilogue() == GemmBackendConfig::BIAS) {
+      config.set_epilogue(GemmBackendConfig::BIAS_RELU_AUX);
+    } else {
+      return absl::OkStatus();
+    }
+    auto total_elements = [](const HloInstruction *gemm) {
+      int64_t num_e = 1;
+      for (int i = 0; i < gemm->shape().rank(); ++i) {
+        num_e *= gemm->shape().dimensions(i);
```
philipphack, line 1644: Use ShapeUtil::ElementsIn.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    std::unique_ptr<HloInstruction> output = fwd_gemm->CloneWithNewShape(
+        ShapeUtil::MakeTupleShape({fwd_gemm->shape(), mask_shape}));
+    TF_RETURN_IF_ERROR(output->set_backend_config(gpu_config));
+    TF_RETURN_IF_ERROR(SetName(output->GetModule(), output.get()));
+    HloInstruction *tuple_output =
+        fwd_gemm->parent()->AddInstruction(std::move(output));
+    TF_RETURN_IF_ERROR(ReplaceWithNewInstruction(
+        maximum, HloInstruction::CreateGetTupleElement(tuple_output, 0)));
+
+    HloInstruction *get_tuple1 = fwd_gemm->parent()->AddInstruction(
```
philipphack, line 1660: Is there another "get_tuple"?

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+        HloInstruction::CreateGetTupleElement(tuple_output, 1));
+
+    // Insert DRELU in backward matmul
+    std::vector<HloInstruction *> operands(bwd_gemm->operands().begin(),
+                                           bwd_gemm->operands().end());
+    operands.insert(operands.end(), get_tuple1);
+
+    HloInstruction *new_bwd_custom_call =
+        bwd_gemm->parent()->AddInstruction(HloInstruction::CreateCustomCall(
```
philipphack, line 1669: Replace with CloneWithNewOperands.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    // 3. Insert DRelu as epiloguefor backward matmul.
+    HloInstruction *fwd_gemm = nullptr;
+    HloInstruction *bwd_gemm = nullptr;
+    HloInstruction *maximum = nullptr;
+    HloInstruction *compare = nullptr;
+
+    if (!Match(instr,
+               m::Select(m::Compare(&compare,
+                                    m::CustomCall(&fwd_gemm,
+                                                  {kCublasLtMatmulCallTarget}),
```
philipphack, line 1589: Replace with CublasLtMatmul(fwd_gemm). Can we also do this for FP8?

philipphack (changes requested):

reedwm (changes requested):
So far, I only looked at gemm_rewriter.cc and only gave a few high level comments.

I'm not sure why you need separate logic for FP8, which you implemented in a different PR: https://github.com/openxla/xla/pull/10999. Is this because you want to fuse fp8_matmul+relu+quantize into a single cublas call, and the logic here cannot handle the quantize? Can you write out the expected pattern of the FP8 pattern matcher? cublasLT is only used for FP8 by default, so it's not particular useful to support this without FP8.

Also do you have performance numbers for a model where this improves performance on?

kamaljeeti:
Hi @wenscarl, can you have a look at the comments? Thanks.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+                                    m::Broadcast(m::ConstantScalar(0))),
+                         m::CustomCall(&bwd_gemm, {kCublasLtMatmulCallTarget})
+                             .WithOneUser(),
+                         m::Broadcast(m::ConstantScalar(0))))) {
+      return absl::OkStatus();
+    }
```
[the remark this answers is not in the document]
wenscarl: Yes. The description has been updated with targeted patterns.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+
+    bool valid_relu_bwd = false;
+
+    for (auto user : fwd_gemm->users()) {
+      if (user->opcode() == HloOpcode::kMaximum &&
+          Match(user,
+                m::MaximumAnyOrder(
+                    m::AnyOf<HloInstruction>(
+                        m::Slice(m::CustomCall({kCublasLtMatmulCallTarget})),
```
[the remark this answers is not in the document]
wenscarl: Yes. The cases are not supported because the bitmask may not directly useable by the backward matmuls.

xla/service/gpu/gemm_rewriter.cc
wenscarl: Matrix bias is not supported since cublasLt's epilogue D_RELU_BGRAD only supports vector bias.

wenscarl: No. hlo_verifier.cc will check the `output_to_operand_aliasing` for matrix bias C which is also D. If the output shape of forward matmul is a tuple, the check may fail.

xla/service/gpu/gemm_rewriter.cc
wenscarl: Fixed.

xla/service/gpu/gemm_rewriter.cc
wenscarl: Fixed.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+                        m::Bitcast(m::CustomCall({kCublasLtMatmulCallTarget})),
+                        m::CustomCall({kCublasLtMatmulCallTarget})),
+                    m::Broadcast(m::ConstantScalar(0))))) {
+        maximum = user;
+        valid_relu_bwd = true;
```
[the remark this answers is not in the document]
wenscarl: Fixed.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1572,7 +1572,147 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    auto total_elements = [](const HloInstruction *gemm) {
+      int64_t num_e = 1;
+      for (int i = 0; i < gemm->shape().rank(); ++i) {
+        num_e *= gemm->shape().dimensions(i);
+      }
+      return num_e;
+    };
+    Shape mask_shape =
+        ShapeUtil::MakeShape(PrimitiveType::U8, {total_elements(fwd_gemm)});
+    mask_shape.mutable_layout()->set_element_size_in_bits(1);
```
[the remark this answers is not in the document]
wenscarl: Fixed

wenscarl:
All issues have been addressed.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -729,78 +753,135 @@ absl::Status IrEmitterUnnested::EmitCublasLtMatmulThunk(
...
       BufferAllocation::Slice c_scale,
       GetAllocationSliceForHlo(instr->operand(a_scale_index + 2)));
   TF_ASSIGN_OR_RETURN(
       BufferAllocation::Slice d_scale,
       GetAllocationSliceForHlo(instr->operand(a_scale_index + 3)));
-#else  // TENSORFLOW_USE_ROCM
-  BufferAllocation::Slice c_scale;
-  BufferAllocation::Slice d_scale;
-#endif
+  BufferAllocation::Slice aux, bias, d_amax;
```
philipphack, line 809: Can we streamline this logic? E.g. the case when has_vector_bias is true is repeated 4 times.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1285,27 +1439,61 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
-            instr->shape(), gemm_and_damax, 0));
-    HloInstruction *damax = instr->AddInstruction(
-        HloInstruction::CreateGetTupleElement(damax_shape, gemm_and_damax, 1));
-
-    // Convert DAmax from FP32 to the requested type and elide reduce.
-    HloInstruction *damax_converted = instr->AddInstruction(
-        HloInstruction::CreateConvert(reduce_damax->shape(), damax));
-    TF_RETURN_IF_ERROR(ReplaceInstruction(reduce_damax, damax_converted));
-    TF_RETURN_IF_ERROR(ReplaceInstruction(instr, d));
+    if (!get_tuple_element) {
```
philipphack, line 1446: Can we move the logic for handling the case when get_tuple_element is not nullptr into one or two small conditionals instead of duplicating everything?

philipphack (changes requested):

xla/service/gpu/ir_emitter_unnested.cc
wenscarl: Fixed.

xla/service/gpu/gemm_rewriter.cc
wenscarl: Fixed.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -729,78 +753,103 @@ absl::Status IrEmitterUnnested::EmitCublasLtMatmulThunk(
...
+                                  ? GetAllocationSliceForHlo(instr, {1})
+                                  : GetAllocationSliceForHlo(
+                                        instr->operand(a_scale_index + 4)));
   }
 
-  BufferAllocation::Slice d_amax;
-  if (has_damax) {
-    TF_ASSIGN_OR_RETURN(d_amax, GetAllocationSliceForHlo(instr, {1}));
+  if (!is_relu_epilogue) {
+    bool has_damax = has_tuple_output;
```
philipphack, line 818: Do we need this variable?

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -729,78 +753,103 @@ absl::Status IrEmitterUnnested::EmitCublasLtMatmulThunk(
...
+    }
+    // d_amax not used.
+  } else if (epilogue == GemmBackendConfig::RELU_AUX ||
+             epilogue == GemmBackendConfig::BIAS_RELU_AUX) {
+    if (has_aux_output) {
+      TF_ASSIGN_OR_RETURN(aux, GetAllocationSliceForHlo(instr, {1}));
+    }
+    bool has_damax = has_tuple_output &&
+                     instr->shape().tuple_shapes().size() >
+                         (epilogue == GemmBackendConfig::RELU_AUX ? 2 : 1);
```
philipphack, line 838: Is there valid case with instr the shape of a tuple with one element?

xla/service/gpu/tests/gemm_rewrite_test.cc
```diff
@@ -3719,6 +3719,165 @@ ENTRY test {
       )");
 }
 
+TEST_F(CublasLtGemmRewriteTest, ReluActivationWithAux) {
```
philipphack, line 3722: Add FP8 tests.

philipphack (changes requested):

xla/service/gpu/ir_emitter_unnested.cc
wenscarl: I cannot think of one. In this case, it's 3 for RELU_AUX and 2 for BIAS_RELU_AUX.

xla/service/gpu/tests/gemm_rewrite_test.cc
```diff
@@ -3719,6 +3719,167 @@ ENTRY test {
...
```
reedwm, line 3722: In most real-world ML models, there is a path from the forward matmul to the backward matmul, since you need intermediate activation in the forward pass to compute the backward pass. This affects the order GemmRewriterVisitor processes nodes, which could affect correctness, so would be good to test this case.

Can you add an FP8 relu aux test that tests this, by passing the output of the first dot as an input to the second dot?

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -674,53 +674,77 @@ absl::Status IrEmitterUnnested::EmitCublasLtMatmulThunk(
...
-  xla::ShapeIndex output_index =
-      has_aux_output ? xla::ShapeIndex{0} : xla::ShapeIndex{};
+  TF_ASSIGN_OR_RETURN(
+      bool has_aux_input,
+      xla::gpu::gpublas_lt::EpilogueHasAuxiliaryInput(epilogue));
+  if (is_forward_mode) {
+    TF_RET_CHECK(instr->operand_count() ==
+                 2 + int{has_matrix_bias} + int{has_vector_bias});
+  } else {
+    TF_RET_CHECK(instr->operand_count() == 2 + int{has_aux_input});
```
reedwm, line 692: Is it possible to have both a bias and an aux input? Admittedly this would be unusual since we only expect aux inputs on the backwards pass, which typically doesn't have bias adds

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -161,25 +169,42 @@ class GemmAutotuner {
...
+
+      TF_ASSIGN_OR_RETURN(bias_buffer, CreateBuffer(bias_buffer_shape));
     }
+
     if (has_aux_output) {
       TF_ASSIGN_OR_RETURN(aux_buffer,
                           CreateBuffer(gemm->shape().tuple_shapes(1)));
     }
+    if (has_aux_input) {
+      // The last operand is for ReLU bitmask for D_RELU epilogue
```
reedwm, line 200: It's invalid to have both an aux input and output. CHECK or DCHECK this

reedwm (changes requested):

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -858,33 +865,170 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    } else if (config.epilogue() == GemmBackendConfig::BIAS) {
+      config.set_epilogue(GemmBackendConfig::BIAS_RELU_AUX);
+    } else {
+      return absl::OkStatus();
+    }
+
+    Shape mask_shape = ShapeUtil::MakeShape(
+        PrimitiveType::U8, {ShapeUtil::ElementsIn(fwd_gemm->shape()) / 8});
+    std::unique_ptr<HloInstruction> output = fwd_gemm->CloneWithNewShape(
+        ShapeUtil::MakeTupleShape({fwd_gemm->shape(), mask_shape}));
```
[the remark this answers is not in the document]
wenscarl: That case is handled after this function. Otherwise, it will be matmul->get_tuple_element.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -858,33 +865,170 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+                        m::ConstantScalar(0)))) {
+      TF_ASSIGN_OR_RETURN(auto gpu_config,
+                          gemm->backend_config<GpuBackendConfig>());
+      GemmBackendConfig &config = *gpu_config.mutable_gemm_backend_config();
+      if (config.epilogue() != GemmBackendConfig::D_RELU) {
+        return absl::OkStatus();
+      }
+      config.set_epilogue(GemmBackendConfig::D_RELU_BGRAD);
+      std::unique_ptr<HloInstruction> output = gemm->CloneWithNewShape(
+          ShapeUtil::MakeTupleShape({gemm->shape(), instr->shape()}));
```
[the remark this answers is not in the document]
wenscarl: If the output is a tuple, then it won't be gemm but get_tuple_element, thus failing the pattern matching.

xla/service/gpu/ir_emitter_unnested.cc
wenscarl: I cannot think of one case.

xla/service/gpu/tests/gemm_rewrite_test.cc
wenscarl: There is a [FP8 test](https://github.com/openxla/xla/pull/10977/files#diff-71bce65d79d6395cde36b49b4504c6b0c73694cee153e5685a8b46ff146f09e0R5595) for this purpose.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -858,33 +865,170 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    }
+
+    Shape mask_shape = ShapeUtil::MakeShape(
+        PrimitiveType::U8, {ShapeUtil::ElementsIn(fwd_gemm->shape()) / 8});
+    std::unique_ptr<HloInstruction> output = fwd_gemm->CloneWithNewShape(
+        ShapeUtil::MakeTupleShape({fwd_gemm->shape(), mask_shape}));
+    TF_RETURN_IF_ERROR(output->set_backend_config(gpu_config));
+    TF_RETURN_IF_ERROR(SetName(output->GetModule(), output.get()));
+    HloInstruction *tuple_output =
+        fwd_gemm->parent()->AddInstruction(std::move(output));
```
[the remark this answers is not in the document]
wenscarl: `parent()` is necessary to ensure the correctness.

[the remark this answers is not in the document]
philipphack: HloInstruction::AddInstruction calls HloComputation::AddInstruction. Why do we need to call parent here?

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -858,33 +865,170 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+  // and its corresponding backward pass
+  // dy1 = grad_matmul2(dy, weight2)
+  // dy2 = drelu(dy1)
+  // dy3 = grad_matmul1(dy2, weight1)
+  // In the forward pass, fuse ReLU into matmul1 as an epilogue and generate a
+  // bitmask for subsequent use in grad_matmul2. For the backward pass,
+  // streamline and combine the ReLU gradient computation with grad_matmul2 as
+  // an epilogue.
+  absl::Status HandleSelect(HloInstruction *instr) override {
+    HloInstruction *fwd_gemm = nullptr;
```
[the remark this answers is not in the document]
philipphack: Wouldn't it be better to use the same names everywhere instead of having two sets of names and an additional comment explaining how they are related?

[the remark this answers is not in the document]
wenscarl: Fixed.

xla/service/gpu/gemm_rewriter.cc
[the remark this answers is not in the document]
wenscarl: Ref [here](https://github.com/openxla/xla/blob/main/xla/service/gpu/gemm_rewriter.cc#L1663). Otherwise, the epilogue RELU_AUX is not seen. The `fwd_gemm` is later replaced by other operation.

kamaljeeti:
Hi @reedwm , can you please have a look into this PR? Thanks.

xla/service/gpu/gemm_rewriter.cc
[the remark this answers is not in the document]
reedwm: Where is this case handled in particular?

Also I think https://github.com/openxla/xla/pull/11514 causes the output of the gemm to always be a tuple now.

xla/service/gpu/gemm_rewriter.cc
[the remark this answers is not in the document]
reedwm: The reference link is out of date since the file has since been modfied. I'm still confused why we need to call parent() here

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -858,33 +865,176 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+  absl::Status HandleReduce(HloInstruction *instr) override {
+    HloInstruction *gemm = nullptr;
+    if (Match(instr,
+              m::Reduce(CublasLtMatmulMaybeF8(&gemm), m::ConstantScalar(0))
+                  .WithPredicate([](const HloInstruction *reduce) {
+                    HloComputation *reducer = reduce->to_apply();
+                    return (reducer->root_instruction()->opcode() ==
+                                HloOpcode::kAdd &&
+                            reduce->dimensions().size() == 1 &&
+                            reduce->dimensions()[0] == 0);
```
reedwm, line 1013: You need to check the both operands of the reduce are parameters

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -858,33 +865,176 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+        *bwd_gpu_backend_config.mutable_gemm_backend_config();
+    bwd_config.set_epilogue(GemmBackendConfig::D_RELU);
+    TF_RETURN_IF_ERROR(
+        new_bwd_custom_call->set_backend_config(bwd_gpu_backend_config));
+
+    return ReplaceInstruction(instr, new_bwd_custom_call);
+  }
+
+  // Attempt to match the backward pass of a vector bias addition only when
+  // there is a ReLU.
```
reedwm, line 1003: Why only when there is a ReLU? If we just didn't implement the case without a Relu, I would phrase this as:

> Attempt to match the gradient of a vector bias addition, which is a reduction. Currently we only match this when there is a Relu gradient before the bias addition gradient.

xla/service/gpu/gemm_rewriter.cc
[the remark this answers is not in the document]
reedwm: Good point. But note, as I said in a previous comment, I think the output will now always be a tuple.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1285,24 +1444,36 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+        instr->AddInstruction(HloInstruction::CreateGetTupleElement(
+            damax_shape, gemm_and_damax, (get_tuple_element ? 2 : 1)));
 
     // Convert DAmax from FP32 to the requested type and elide reduce.
     HloInstruction *damax_converted = instr->AddInstruction(
         HloInstruction::CreateConvert(reduce_damax->shape(), damax));
+
+    if (get_tuple_element) {
+      TF_RETURN_IF_ERROR(instr->parent()->ReplaceInstructionWithDifferentShape(
+          existing_gemm, gemm_and_damax));
```
reedwm, line 1475: I know I suggested this before, but I actually prefer not to replace existing_gemm with gemm_and_damax. This reason is that you change the dtype of the first output of existing_gemm, so if there are any users of this output, this replacement will fail. 

The old logic you have I don't think is correct anymore either since even gemms without a bitmask can output a tuple. Instead, I would check for users for each of the outputs of gemm (other than the first output and the amax output) and replace them a new GTE instruction

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1285,24 +1444,36 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+                          HloInstruction *get_tuple_element) {
     // Change the output shape of the Custom Call to tuple(D, DAmax).
     Shape damax_shape = ShapeUtil::MakeScalarShape(F32);
-    Shape tuple_shape =
-        ShapeUtil::MakeTupleShape({instr->shape(), damax_shape});
+
+    std::vector<Shape> tuple_shape_vec = {instr->shape(), damax_shape};
+    if (get_tuple_element) {
+      tuple_shape_vec.insert(tuple_shape_vec.begin() + 1,
+                             existing_gemm->shape().tuple_shapes(1));
```
reedwm, line 1455: I don't think this works if the tuple output has more than 2 outputs. You should handle an arbitrary number of outputs by copying the tuple shape, replacing the first shape with `instr->shape()`, and appending damax_shape.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -858,33 +865,176 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+      if (Match(op, m::MaximumAnyOrder(CublasLtMatmulMaybeF8(nullptr),
+                                       m::Broadcast(m::ConstantScalar(0))))) {
+        maximum = op;
+        break;
+      }
+    }
+    int64_t fwd_gemm_num_rows = fwd_gemm->shape().dimensions(0);
+    if (maximum == nullptr ||
+        !SupportsEpilogueFusion(fwd_gemm->shape().element_type()) ||
+        (fwd_gemm_num_rows % 128 != 0)) {
```
reedwm, line 943: No need for parenthesis around `fwd_gemm_num_rows % 128 != 0`

xla/service/gpu/tests/gemm_rewrite_test.cc
```diff
@@ -5431,6 +5592,230 @@ TEST_P(ParameterizedFp8GemmRewriteTest, ScaledABAlphaDF8) {
...
+      x_scale_bcast = f32[128,128] broadcast(x_scale), dimensions={}
+      y_scale_bcast = f32[128,128] broadcast(y_scale), dimensions={}
+      x_unscaled = f32[128,128] multiply(x_f32, x_scale_bcast)
+      y_unscaled = f32[128,128] multiply(y_f32, y_scale_bcast)
+      const.0 = f32[] constant(0)
+      bcast.0 = f32[128,128] broadcast(const.0), dimensions={}
+      dot = f32[128,128] dot(x_unscaled, y_unscaled), lhs_contracting_dims={1}, rhs_contracting_dims={0}
+      c = f32[128,128] parameter(8)
+      maximum.1 = f32[128,128] maximum(dot, bcast.0)
+      dot.1 = f32[128,128] dot(maximum.1, c), lhs_contracting_dims={1}, rhs_contracting_dims={0}
```
reedwm, line 5627: I don't think dot.1 has any purpose here. I previously suggested testing this by having a path from the forward dot to the backwards dot, but this doesn't accomplish that. Instead, you can replaces uses of `a` with `dot` to accomplish that.

reedwm (changes requested):

penpornk:
@wenscarl Could you please help address @reedwm's comments?

dimitar-asenov:
@wenscarl It seems that PR has been inactive for two months. Please provide an update if you still intend to merge it, otherwise we will close it after some time of further inactivity.

wenscarl:
This matmul-relu-matmul pattern is not so commonly seen among models. Closing.

PR #11094, opened 2024-03-30
[ROCm] enable rocm for se_gpu_pjrt_compiler_aot_test

i-chaochen:


penpornk (approved):
Thank you for the PR!

xla/pjrt/gpu/BUILD
```diff
@@ -309,10 +309,8 @@ xla_cc_test(
     name = "se_gpu_pjrt_compiler_aot_test",
     srcs = if_gpu_is_configured(["se_gpu_pjrt_compiler_aot_test.cc"]),
     tags = [
-        "config-cuda-only",
         "gpu",
         "no_oss",
-        "requires-gpu-nvidia",
```
ddunl, line 315: We need to keep this tag here because it has some internal meaning that ensures this test runs with a GPU available

i-chaochen: done

ddunl (approved):
Thanks!!

xla/pjrt/gpu/BUILD
```diff
@@ -342,5 +339,12 @@ xla_cc_test(
         "@tsl//tsl/platform:protobuf",
         "@tsl//tsl/platform:statusor",
         "@tsl//tsl/platform:test_main",
-    ],
+    ]
+    + if_cuda([
```
ddunl, line 343: My bad should've caught this last time, I'd expect both `if_cuda` and `if_rocm` to be the `..._is_configured` variants. But maybe there's a reason why not that I'm not understanding here

i-chaochen: they are from https://github.com/openxla/xla/commit/1f3764dc65bb7a8d444e37b640e09f722c9e609f I just use it...If you want to change to ```..._is_configured``` variants, I think it needs to change all related ones in this file.

ddunl: I see, I think that there is an issue with how these are used in the build file but that was a misdiagnosis of the internal error on my part. I think we need to keep the `config-cuda-only` tag too, but I think it's usage here is deceiving as it has no meaning in OpenXLA, our build scripts only look at `gpu` and `requires-gpu-nvidia`. So I think what's best is to keep that tag but wrap it in `if_google` to try to clarify the intent a bit. Sorry for the back and forth here.

i-chaochen: wrap it in `if_google` ? I don't see this tag in this BUILD file. Do you mean adding sth like this

```
load(
    "@tsl//tsl:tsl.bzl",
    "if_google"
)


xla_cc_test(
    name = "se_gpu_pjrt_compiler_aot_test",
    srcs = if_google(["se_gpu_pjrt_compiler_aot_test.cc"]),
    local_defines = if_cuda(["GOOGLE_CUDA=1"]) + if_rocm([
        "TENSORFLOW_USE_ROCM=1",
    ]),
)
```

ddunl: I mean like loading `if_google` as you have, but then in the tags have something like `tags = [...] + if_google(["config-cuda-only"])`. I'll send a change doing that for other targets as well, I think it helps make the intent clearer. We have a somewhat confusing situation where the set of the tags needed internally is different than externally.

i-chaochen: Do you mean sth like this?

```
xla_cc_test(
    name = "se_gpu_pjrt_compiler_aot_test",
    srcs = if_gpu_is_configured(["se_gpu_pjrt_compiler_aot_test.cc"]),
    tags = [
        "gpu",
        "no_oss",
        "requires-gpu-nvidia",
    ] + if_google(["config-cuda-only"]),
  deps= [....],
```

akuegel: Yes, I think that should be what we want. And sorry for the confusing tags. This specific tag is used to indicate that our internal CI needs to pass the --config=cuda flag when running the test (we don't test --config=rocm internally).

i-chaochen: Thanks for the clarification. Help now it's ok.

akuegel (approved):

PR #11164, opened 2024-04-03
[XLA:GPU] bump up minimum PTX ISA to be 8.1 for CUDA >= 12.1

Cjkkkk:
NV internal workloads facing an error rn: `error   : Feature 'Kernel parameter size larger than 4352 bytes' requires PTX ISA .version 8.1 or later`. Bumping up to PTX81 solved the issue however it requires minimum CUDA 12.1 to work. So for WAR, I added a check to use PTX81 if CUDA 12.1 is available.

### DIFF
--- a/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
+++ b/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
@@ -76,6 +76,10 @@ limitations under the License.
 #include "rocm/rocm_config.h"
 #endif
 
+#if GOOGLE_CUDA
+#include "third_party/gpus/cuda/include/cuda.h"
+#endif
+
 namespace xla {
 namespace gpu {
 namespace {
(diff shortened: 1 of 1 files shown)


cheshire:
@Artem-B WDYT? Should we check the target device we are compiling to as well?

Artem-B:
I think it should be fine.

The only downside will be that if user somehow picks up old ptxas (via PATH?) the failure reason will be hard to diagnose for the user. I think we did have run-time ptxas version checks somewhere. We should probably add a safeguard check there that if we build with CUDA-12, the ptxas must be the same version or newer.

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
```diff
@@ -294,6 +294,11 @@ std::unique_ptr<llvm::TargetMachine> NVPTXGetTargetMachine(
     const DebugOptions& debug_options) {
   // Figure out the exact name of the processor as known to the NVPTX backend
   // from the gpu_architecture flag.
+#if defined(GOOGLE_CUDA) && CUDA_VERSION >= 12010
```
beckerhe, line 297: Can you add an include to cuda.h to make sure CUDA_VERSION is available and that we don't rely on transitive includes?

Otherwise this might silently break in the future when some headers/includes get restructured.

Cjkkkk: Done.

beckerhe (approved):

PR #11182, opened 2024-04-03
[NVIDIA GPU] Optimize collective permute using memcpy

Tixxx:
When the source and target pair of a collective permute instruction are located within the same node, we will try to use cudaMemCpyPeerToPeer to send data instead of invoking nccl send and recv kernels. Because memcpy will use the copy engine on the GPU instead of launching a kernel, copy engine can fully saturate the bandwidth without occupying any SM.
Since memcpyP2p needs the destination pointer of the other device, we first communicate the pointer values with peer using nccl calls and use the pointer to invoke memcpy.
Some perf comparison:
sending a tensor of this size bf16[512,24576]{1,0}:
Using collective-permute: 445 us
using memcpy: 100us

### DIFF
--- a/xla/service/gpu/gpu_executable.cc
+++ b/xla/service/gpu/gpu_executable.cc
@@ -507,6 +507,19 @@ absl::Status ExecuteThunks(
 
     TF_RETURN_IF_ERROR(thunk->ExecuteOnStream(execute_params));
   }
+
+  {  // Cleans up any resource occupied by thunks.
+    Thunk::CleanupParams cleanup_params{
+        /*executor=*/executor,
+        /*collective_params=*/&collective_params,
+    };
+
+    tsl::profiler::TraceMe trace([&] { return "Thunks::Cleanup"; });
+    for (const std::unique_ptr<Thunk>& thunk : thunk_sequence) {
+      TF_RETURN_IF_ERROR(thunk->Cleanup(cleanup_params));
+    }
+  }
+
   return MaybeSyncAndProfile(run_options, std::move(execution_timer),
                              block_host_until_done ? main_stream : nullptr);
 }
--- a/xla/service/gpu/gpu_executable.cc
+++ b/xla/service/gpu/gpu_executable.cc
@@ -507,19 +507,6 @@ absl::Status ExecuteThunks(
 
     TF_RETURN_IF_ERROR(thunk->ExecuteOnStream(execute_params));
   }
-
-  {  // Cleans up any resource occupied by thunks.
-    Thunk::CleanupParams cleanup_params{
-        /*executor=*/executor,
-        /*collective_params=*/&collective_params,
-    };
-
-    tsl::profiler::TraceMe trace([&] { return "Thunks::Cleanup"; });
-    for (const std::unique_ptr<Thunk>& thunk : thunk_sequence) {
-      TF_RETURN_IF_ERROR(thunk->Cleanup(cleanup_params));
-    }
-  }
-
   return MaybeSyncAndProfile(run_options, std::move(execution_timer),
                              block_host_until_done ? main_stream : nullptr);
 }
--- a/xla/service/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/service/gpu/runtime/command_buffer_cmd.cc
@@ -1448,12 +1448,13 @@ absl::Status AllReduceCmd::Record(const Thunk::ExecuteParams& execute_params,
 
   // Today when recording collective operations into command buffers we always
   // use a sync mode and a stream id `0`.
-  TF_ASSIGN_OR_RETURN(NcclApi::NcclCommHandle comm,
+  std::pair<NcclApi::NcclCommHandle, bool> comm_pair;
+  TF_ASSIGN_OR_RETURN(comm_pair,
                       GetNcclComm(*execute_params.collective_params,
                                   *execute_params.collective_cliques,
                                   config().replica_groups, config().group_mode,
                                   /*stream_id=*/0, GetAsyncStreamKind()));
-
+  NcclApi::NcclCommHandle comm = comm_pair.first;
   // Use custom allocator for persistent execution plans.
   NcclApi::ScopedPersistentPlanAllocator scoped_allocator(
       comm, tsl::MakeRef<NcclApi::PersistentPlanAllocator>(
@@ -1516,12 +1517,13 @@ absl::Status ReduceScatterCmd::Record(
 
   // Today when recording collective operations into command buffers we always
   // use a sync mode and a stream id `0`.
-  TF_ASSIGN_OR_RETURN(NcclApi::NcclCommHandle comm,
+  std::pair<NcclApi::NcclCommHandle, bool> comm_pair;
+  TF_ASSIGN_OR_RETURN(comm_pair,
                       GetNcclComm(*execute_params.collective_params,
                                   *execute_params.collective_cliques,
                                   config().replica_groups, config().group_mode,
                                   /*stream_id=*/0, GetAsyncStreamKind()));
-
+  NcclApi::NcclCommHandle comm = comm_pair.first;
   // Use custom allocator for persistent execution plans.
   NcclApi::ScopedPersistentPlanAllocator scoped_allocator(
       comm, tsl::MakeRef<NcclApi::PersistentPlanAllocator>(
@@ -1581,12 +1583,13 @@ absl::Status AllGatherCmd::Record(const Thunk::ExecuteParams& execute_params,
 
   // Today when recording collective operations into command buffers we always
   // use a sync mode and a stream id `0`.
-  TF_ASSIGN_OR_RETURN(NcclApi::NcclCommHandle comm,
+  std::pair<NcclApi::NcclCommHandle, bool> comm_pair;
+  TF_ASSIGN_OR_RETURN(comm_pair,
                       GetNcclComm(*execute_params.collective_params,
                                   *execute_params.collective_cliques,
                                   config().replica_groups, config().group_mode,
                                   /*stream_id=*/0, GetAsyncStreamKind()));
-
+  NcclApi::NcclCommHandle comm = comm_pair.first;
   // Use custom allocator for persistent execution plans.
   NcclApi::ScopedPersistentPlanAllocator scoped_allocator(
       comm, tsl::MakeRef<NcclApi::PersistentPlanAllocator>(
@@ -1646,12 +1649,13 @@ absl::Status CollectiveBroadcastCmd::Record(
 
   // Today when recording collective operations into command buffers we always
   // use a sync mode and a stream id `0`.
-  TF_ASSIGN_OR_RETURN(NcclApi::NcclCommHandle comm,
+  std::pair<NcclApi::NcclCommHandle, bool> comm_pair;
+  TF_ASSIGN_OR_RETURN(comm_pair,
                       GetNcclComm(*execute_params.collective_params,
                                   *execute_params.collective_cliques,
                                   config().replica_groups, config().group_mode,
                                   /*stream_id=*/0, GetAsyncStreamKind()));
-
+  NcclApi::NcclCommHandle comm = comm_pair.first;
   // Use custom allocator for persistent execution plans.
   NcclApi::ScopedPersistentPlanAllocator scoped_allocator(
       comm, tsl::MakeRef<NcclApi::PersistentPlanAllocator>(
--- a/xla/service/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/service/gpu/runtime/command_buffer_cmd.cc
@@ -1448,13 +1448,12 @@ absl::Status AllReduceCmd::Record(const Thunk::ExecuteParams& execute_params,
 
   // Today when recording collective operations into command buffers we always
   // use a sync mode and a stream id `0`.
-  std::pair<NcclApi::NcclCommHandle, bool> comm_pair;
-  TF_ASSIGN_OR_RETURN(comm_pair,
+  TF_ASSIGN_OR_RETURN(NcclCommHandleWrapper comm_handle,
                       GetNcclComm(*execute_params.collective_params,
                                   *execute_params.collective_cliques,
                                   config().replica_groups, config().group_mode,
                                   /*stream_id=*/0, GetAsyncStreamKind()));
-  NcclApi::NcclCommHandle comm = comm_pair.first;
+  NcclApi::NcclCommHandle comm = comm_handle.comm_handle;
   // Use custom allocator for persistent execution plans.
   NcclApi::ScopedPersistentPlanAllocator scoped_allocator(
       comm, tsl::MakeRef<NcclApi::PersistentPlanAllocator>(
@@ -1517,13 +1516,12 @@ absl::Status ReduceScatterCmd::Record(
 
   // Today when recording collective operations into command buffers we always
   // use a sync mode and a stream id `0`.
-  std::pair<NcclApi::NcclCommHandle, bool> comm_pair;
-  TF_ASSIGN_OR_RETURN(comm_pair,
+  TF_ASSIGN_OR_RETURN(NcclCommHandleWrapper comm_handle,
                       GetNcclComm(*execute_params.collective_params,
                                   *execute_params.collective_cliques,
                                   config().replica_groups, config().group_mode,
                                   /*stream_id=*/0, GetAsyncStreamKind()));
-  NcclApi::NcclCommHandle comm = comm_pair.first;
+  NcclApi::NcclCommHandle comm = comm_handle.comm_handle;
   // Use custom allocator for persistent execution plans.
   NcclApi::ScopedPersistentPlanAllocator scoped_allocator(
       comm, tsl::MakeRef<NcclApi::PersistentPlanAllocator>(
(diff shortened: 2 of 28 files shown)


xla/service/gpu/gpu_executable.cc
```diff
@@ -507,6 +507,19 @@ absl::Status ExecuteThunks(
 
     TF_RETURN_IF_ERROR(thunk->ExecuteOnStream(execute_params));
   }
+
+  {  // Cleans up any resource occupied by thunks.
```
ezhulenev, line 511: ... any resources acquired by thunks

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -186,24 +246,46 @@ absl::Status RunCollectivePermute(
...
-  }
-
-  // Receive data from the source peer to the destination buffer.
-  if (source_id) {
-    TF_RETURN_IF_ERROR(nccl_api->Recv(dest_addr, buffer.element_type,
-                                      buffer.element_count, *source_id, comm,
-                                      &stream));
+  // If all peers are local, only get/send device pointer values and invoke
+  // memcpy.
+  if (use_memcpy) {
```
ezhulenev, line 262: I think you should be able to do this without any calls to NCCL, if all peers are local you can use `RendezvousOnce` (it also has a std::once_flag like construct to make sure it's executed once) to exchange pointers between all thunks participating in the send/recv, and  then immediately to memcpy

xla/service/gpu/runtime/nccl_collective_thunk.cc
```diff
@@ -244,7 +244,7 @@ static absl::StatusOr<NcclCliqueKey> GetNcclCliqueKey(
                        stream_kind);
 }
 
-absl::StatusOr<NcclApi::NcclCommHandle> GetNcclComm(
+absl::StatusOr<std::pair<NcclApi::NcclCommHandle, bool>> GetNcclComm(
```
ezhulenev, line 247: prefer struct to a std::pair, the meaning of a bool is hard to guess without reading implementation

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
ezhulenev: You don't even need rendezvous, you should be able to simply put it into a flat_hash_map? Add `AsyncValue<void*>` when requestion a pointer, and make it available when "publishing" a pointer?

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -217,6 +299,14 @@ absl::Status RunCollectivePermute(
                                   device_string);
     TF_RETURN_IF_ERROR(stream.MemZero(&dest_addr, dest_addr.size()));
   }
+  if (use_memcpy && target_id) {
+    // Need to block until the receive pointer buffer has been updated.
+    TF_RETURN_IF_ERROR(stream.BlockHostUntilDone());
```
ezhulenev, line 304: One problem with this is that it will very likely will lead to deadlocks in presence of pending collective operations on different ranks, and it's incompatible with cuda graphs. Is it ok to give up cuda graphs for send/recv?

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
```diff
@@ -46,26 +46,34 @@ class NcclCollectivePermuteStartThunk : public NcclCollectiveThunk {
...
   absl::Status RunNcclCollective(const ExecuteParams& params,
                                  se::Stream& stream,
-                                 NcclApi::NcclCommHandle comm) override;
+                                 NcclApi::NcclCommHandle comm,
+                                 bool is_local) override;
 
  private:
   const NcclP2PConfig config_;
   const Buffer buffer_;
+  std::unordered_map<int64_t, uint64_t> send_value_map_;
```
ezhulenev, line 67: unordered map doesn't guarantee pointer stability, you can easily get asan errors in deregister, you need absl::node_hash_map here

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
```diff
@@ -46,26 +46,34 @@ class NcclCollectivePermuteStartThunk : public NcclCollectiveThunk {
...
+  std::unordered_map<int64_t, uint64_t> recv_value_map_;
```
ezhulenev, line 68: Also I'm not sure if you need cleanup stage at all, you can maintain `absl::node_hash_map<se::StreamExecutor*, void*>` map that you need to deregister in destructor. NCCL clique locking guarantees that you can't execute this thunk concurrently on the same stream executor. Once you see a new instance of stream executor, you register a pointer with it, and issue send/recv-ptr using that pointer.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: Ok that AsyncValue approach really simplifies the code quite a bit.

xla/service/gpu/runtime/nccl_collective_thunk.cc
Tixxx: updated

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: Changing to AsyncValue got rid of this, now it blocks on the async event :). 
As for the cuda graph, most of the cases we dont get many CP kernels in the graph, it should be okay to exclude them from cuda graph. cuda graph might get rid of some launch overhead but compared to 3x speedup using p2p, it's a good tradeoff.

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
Tixxx: IIUC, references and pointers to elements stored in unordered map dont get invalidated except in the case of removal, only iterators are invalidated in cases of rehashing, so i think the stability should be guaranteed. However I looked into the implementation of node_hash_map, looks like it's more optimal in terms of storage so i changed it to node_hash_map anyway.

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
Tixxx: Yea, AsyncValue + map got rid of all the nccl calls so we dont need to register any host pointer anymore, therefore no need for cleanup or storing stream executor.

xla/service/gpu/runtime/nccl_all_reduce_thunk.h
```diff
@@ -77,7 +77,8 @@ class NcclAllReduceStartThunk : public NcclAllReduceReduceScatterThunkBase {
  protected:
   absl::Status RunNcclCollective(const ExecuteParams& params,
                                  se::Stream& stream,
-                                 NcclApi::NcclCommHandle comm) override;
+                                 NcclApi::NcclCommHandle comm,
```
ezhulenev, line 80: Maybe change everything to wrapper type everywhere? You have header with that struct included anyway

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
```diff
@@ -46,26 +50,31 @@ class NcclCollectivePermuteStartThunk : public NcclCollectiveThunk {
...
   absl::Status RunNcclCollective(const ExecuteParams& params,
                                  se::Stream& stream,
-                                 NcclApi::NcclCommHandle comm) override;
+                                 NcclApi::NcclCommHandle comm,
+                                 bool is_local) override;
 
  private:
   const NcclP2PConfig config_;
   const Buffer buffer_;
+  absl::node_hash_map<int64_t, AsyncValueRef<void*>> recv_ptr_map_;
```
ezhulenev, line 69: This map has to be guarded with a mutex because thunks run concurrently, also you pass it around by reference to few function, I'd do something like AsyncEvents to wrap the map and add synchronization: https://github.com/openxla/xla/blob/main/xla/service/gpu/runtime/nccl_collective_thunk.h#L133-L148

xla/service/gpu/runtime/nccl_collective_thunk.h
```diff
@@ -107,6 +107,19 @@ NcclCollectiveConfig GetNcclCollectiveConfigForMlir(
   return config;
 }
 
+// This wraps the ncclCommHandle object along with other information
+// that could be useful.
+struct NcclCommHandleWrapper {
+  NcclCommHandleWrapper(NcclApi::NcclCommHandle handle, bool is_local) :
```
ezhulenev, line 113: Isn't it auto generated? Definitely auto-generated by clang, but subtly different in gcc, I'd try to remove it if not needed.

xla/service/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -1448,12 +1448,12 @@ absl::Status AllReduceCmd::Record(const Thunk::ExecuteParams& execute_params,
...
   // Today when recording collective operations into command buffers we always
   // use a sync mode and a stream id `0`.
-  TF_ASSIGN_OR_RETURN(NcclApi::NcclCommHandle comm,
+  TF_ASSIGN_OR_RETURN(NcclCommHandleWrapper comm_handle,
                       GetNcclComm(*execute_params.collective_params,
                                   *execute_params.collective_cliques,
                                   config().replica_groups, config().group_mode,
                                   /*stream_id=*/0, GetAsyncStreamKind()));
-
+  NcclApi::NcclCommHandle comm = comm_handle.comm_handle;
```
ezhulenev, line 1456: I wonder can you do `TF_ASSIGN_OR_RETURN(auto& [comm, _], ...`)? If not that's certainly something to fix in macro definition (some time later)

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
Tixxx: I agree we need to lock when doing initialization since we only do rendezvous after initialization in gpu_executable. Once we properly initialize the map, we shouldn't need more locking there? the current_id is unique, so each thread will only modify their own entry in the map and the synchronization is ensured by the value which is an AsyncValueRef, the sending thread will not proceed until the receiving thread signals that the value is in available state.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -217,6 +257,21 @@ absl::Status RunCollectivePermute(
                                   device_string);
     TF_RETURN_IF_ERROR(stream.MemZero(&dest_addr, dest_addr.size()));
   }
+  if (use_memcpy && target_id) {
+    // Need to block until the receive pointer buffer has been updated.
+    if (recv_ptr_map[*target_id].IsUnavailable()) {
+      BlockUntilReady(recv_ptr_map[*target_id].GetAsyncValue());
```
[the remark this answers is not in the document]
Tixxx: unfortunately it doesnt supporting blocking on AsyncValueRef, i added a TODO for myself to add it later.

xla/service/gpu/runtime/nccl_all_reduce_thunk.h
Tixxx: updated

xla/service/gpu/runtime/nccl_collective_thunk.h
Tixxx: No it doesnt auto-gen the proper constructor for me though, it only creates copy, move and default constructor with no arguments. I have tried gcc and clang-17, maybe im missing some flags? gcc will definitely complain cuz it's pedantic :(. I have to keep it there.

xla/service/gpu/runtime/command_buffer_cmd.cc
Tixxx: it doesnt work for me, i have seen this for a pair or tuple, didnt know it would work for well defined struct.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -114,38 +132,53 @@ NcclCollectivePermuteStartThunk::NcclCollectivePermuteStartThunk(
...
+      comm_wrapper.comm_handle, device_string, current_id, use_memcpy,
+      recv_ptr_map_);
 }
 
 absl::Status RunCollectivePermute(
     NcclApi* nccl_api, NcclP2PConfig::SourceTargetMapEntry source_target,
     DeviceBufferPair& buffer, se::Stream& stream, NcclApi::NcclCommHandle comm,
-    absl::string_view device_string, int64_t current_id) {
+    absl::string_view device_string, int64_t current_id, bool use_memcpy,
+    absl::node_hash_map<int64_t, AsyncValueRef<void*>>& recv_ptr_map) {
```
ezhulenev, line 181: If you'll pass map by const-ref then I think it should correct wrt to data races, you just have to promise that all insertions into the map a properly guarded

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -192,18 +225,30 @@ absl::Status RunCollectivePermute(
...
-                                      &stream));
+  // If all peers are local, only get/send device pointer values and invoke
+  // memcpy.
+  if (use_memcpy) {
+    // If sending to another peer, get the pointer value of the src addr.
+    // Only change the pointer value when it's different from stored one.
+    if (source_id && recv_ptr_map[current_id].IsUnavailable()) {
+      VLOG(3) << "Using memcpy, initializing pointer: " << dest_addr.opaque()
+              << " current_id " << current_id;
+      recv_ptr_map[current_id].emplace(dest_addr.opaque());
```
ezhulenev, line 236: If somehow current_id will not be in the map, this will be a data race. If you use find and fail on missing id's then it should be safe.

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
```diff
@@ -46,26 +50,31 @@ class NcclCollectivePermuteStartThunk : public NcclCollectiveThunk {
...
+  absl::node_hash_map<int64_t, AsyncValueRef<void*>> recv_ptr_map_ ABSL_GUARDED_BY(mutex_);
+  bool p2p_memcpy_enabled_ = false;
 };
 
 absl::Status RunCollectivePermute(
     NcclApi* nccl_api, NcclP2PConfig::SourceTargetMapEntry source_target,
     DeviceBufferPair& buffer, se::Stream& stream, NcclApi::NcclCommHandle comm,
-    absl::string_view device_string, int64_t current_id);
+    absl::string_view device_string, int64_t current_id, bool use_memcpy,
+    absl::node_hash_map<int64_t, AsyncValueRef<void*>>& recv_ptr_map);
```
ezhulenev, line 77: `const ref`

xla/service/gpu/runtime/nccl_collective_permute_thunk.h
ezhulenev: Yes, looks like it should be safe, it's just not clear from the function signature because you pass mutable map around

ezhulenev: Also I'd add an alias for this map name, it's passed around in a few places, and it's very long type name

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -191,21 +164,21 @@ absl::Status NcclCollectivePermuteStartThunk::RunNcclCollective(
   const NcclP2PConfig::SourceTargetMapEntry source_target =
       NcclP2PConfig::GetSourceTarget(config_.id_to_source_target, current_id);
 
-  bool use_memcpy = is_local &&
-                    recv_value_map_.find(current_id) != recv_value_map_.end() &&
-                    send_value_map_.find(current_id) != send_value_map_.end() &&
+  bool use_memcpy = false;
+  use_memcpy = comm_wrapper.is_local &&
```
jprabhas, line 168: NIT: why do we need a separate initialization and assignment. The two can be combined

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: yup, updated.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: good catch, it's a peft-over from previous commit. deleted it.

ezhulenev (approved):

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -114,38 +132,53 @@ NcclCollectivePermuteStartThunk::NcclCollectivePermuteStartThunk(
...
   std::string device_string = GetDeviceString(*params.collective_params);
 
   const NcclP2PConfig::SourceTargetMapEntry source_target =
       NcclP2PConfig::GetSourceTarget(config_.id_to_source_target, current_id);
 
-  return ::xla::gpu::RunCollectivePermute(nccl_api(), source_target,
-                                          device_buffers[0], stream, comm,
-                                          device_string, current_id);
+  bool use_memcpy = comm_wrapper.is_local &&
+                    recv_ptr_map_.find(current_id) != recv_ptr_map_.end() &&
```
golechwierowicz, line 169: Since you're using ABSL_GUARDED_BY(mutex_) on recv_ptr_map you need to lock the mutex here, and on line 171 (as you pass it to the RunCollectivePermute).

Tixxx: thanks, i eventually created a wrapper class to wrap the lock and the map and use the lock wherever the map is looked up.

golechwierowicz (approved):

PR #11225, opened 2024-04-04
Fix order of operations in failing transpose test

hsharsha:


xla/service/gpu/fusions/transpose_test.cc
```diff
@@ -80,7 +80,7 @@ TEST_F(TransposeTest, ThreadIndexing021) {
         (d0, d1, d2, d3, d4, d5)[s0, s1, s2] -> (
           d3 floordiv 2,
           d0 floordiv 32 + s1 * 4,
-          (d3 mod 2) * 32 + d0 mod 32
+          d0 mod 32 + (d3 mod 2) * 32
```
hsharsha, line 83: Does the order of addition affect performance?

xla/service/gpu/fusions/transpose_test.cc
```diff
@@ -79,9 +79,13 @@ TEST_F(TransposeTest, ThreadIndexing021) {
...
           d3 floordiv 2,
-          d0 floordiv 32 + s1 * 4,
-          (d3 mod 2) * 32 + d0 mod 32
-        )
+          d0 floordiv 32 + s1 * 4,)"
+#ifdef GOOGLE_CUDA
+          R"((d3 mod 2) * 32 + d0 mod 32)"
+#elif TENSORFLOW_USE_ROCM
+          R"(d0 mod 32 + (d3 mod 2) * 32)"
+#endif
```
olegshyshkov, line 87: Do you understand what's different in ROCm codegen that causes this difference? The expressions are equivalent, so something the construction is not deterministic. I would rather fix that inside indexing maps.

kamaljeeti:
Hi @hsharsha , can you look into the comments? Thanks.

hsharsha:
> Hi @hsharsha , can you look into the comments? Thanks.

Looks like this is similar to https://github.com/openxla/xla/pull/11567/ Will use the same pattern matching once that PR is accepted

dimitar-asenov:
Hi @hsharsha It seems that #11567 is merged. Please take a look and adjust the current PR when you have the time. Thanks.

hsharsha:
I think there are changes made to indexing maps and string matching ir. I will close this PR as the test is passing now and will revisit if there is a failure in the future.

hsharsha:
I think there are changes made to indexing maps and string matching ir. I will close this PR as the test is passing now and will revisit if there is a failure in the future.

PR #11249, opened 2024-04-05
[XLA:GPU] Update cuDNN flash attention backward to use new cuDNN frontend (>1.0)

Cjkkkk:
* Replace complicated node by node flash attention cuDNN graph construction with simple cudnn sdpa call provided in new cuDNN frontend (>1.0).
* Remove softmax_sum/dq_accum tensors from the custom call outputs since it is not required anymore in new cuDNN frontend.
* Add a pass to query workspace from cuDNN graph instead of manual calculation as workspace size is subject to change.

### DIFF
--- a/xla/stream_executor/cuda/cuda_dnn.cc
+++ b/xla/stream_executor/cuda/cuda_dnn.cc
@@ -6231,147 +6231,6 @@ GetCudnnFusedMHABackwardOperationGraph(
   return std::make_unique<cudnn_frontend::OperationGraph>(std::move(op_graph));
 }
 
-// Returns a cudnn tensor that's the output of the bias addition op
-absl::StatusOr<cudnn_frontend::Tensor> CreateCudnnFlashAttentionBiasFwdTensor(
-    std::vector<cudnn_frontend::Operation>& ops, absl::Span<const int64_t> dims,
-    absl::Span<const int64_t> strides, dnn::DataType dtype,
-    cudnn_frontend::Tensor& input_tensor) {
-  // Create the bias tensor.
-  TF_ASSIGN_OR_RETURN(
-      auto bias_tensor,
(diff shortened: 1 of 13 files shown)


xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -10425,6 +9760,7 @@ absl::Status CudnnGraph::Build(dnn::DnnSupport& dnn_support,
   TF_ASSIGN_OR_RETURN(auto cudnn, cudnn_support.cudnn_->GetLocalHandle());
   RETURN_IF_CUDNN_FRONTEND_ERROR(
       graph_.build_plan_at_index(cudnn->handle(), plan_id));
+  // RETURN_IF_CUDNN_FRONTEND_ERROR(graph_.build_plans(cudnn->handle()));
```
akuegel, line 9763: I guess this line can be removed?

Cjkkkk: fixed.

akuegel (approved):

akuegel:
It looks like there is a missing dependency:

error: module //third_party/tensorflow/compiler/xla/service/gpu:cudnn_workspace_rewriter does not depend on a module exporting 'third_party/tensorflow/compiler/xla/service/gpu/cublas_cudnn.h'

Can you please add that?

akuegel (approved):

Cjkkkk:
@akuegel the google internal test failure should be fixed now.

akuegel (approved):

akuegel:
@Cjkkkk I am currently in the process of moving HLO passes to the transforms subdirectory. In this process I noticed that we are missing a test for the CudnnWorkspaceRewriter pass which was added in this PR (I missed this during the review). Can you please write a test?

PR #11299, opened 2024-04-08
[XLA:CPU] Enable XLA on Windows

mraunak:
This PR aims to enable the XLA test cases on the Windows Platform. The changes made:

1. Changed the .bazelrc file to use the correct toolchain and platform
This change will allow the user to successfully run XLA tests on the Windows platform using the Clang compiler using '--config=win_clang' in the bazel command

2. Added conditions to a few test cases to successfully run on the Windows platform
These test cases check the exit/termination status of a process
WIFEXITED is typically supported in POSIX-compliant operating systems like Unix and Linux to check if a process has terminated normally. WEXITSTATUS allows examining the termination status of child processes. However, these macros are not Windows compliant, hence the additional condition block was added to check the exit/termination status of process or child process for the Windows platform

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -437,8 +437,8 @@ build:avx_win --copt=/arch:AVX
 # Use Clang-cl compiler on Windows
 build:win_clang --copt=/clang:-Weverything
 build:win_clang --extra_toolchains=@local_config_cc//:cc-toolchain-x64_windows-clang-cl
-build:win_clang --extra_execution_platforms=//tools/toolchains/win:x64_windows-clang-cl
-build:win_clang --host_platform=//tools/toolchains/win:x64_windows-clang-cl
+build:win_clang --extra_execution_platforms=//tensorflow/tools/toolchains/win:x64_windows-clang-cl
+build:win_clang --host_platform=//tensorflow/tools/toolchains/win:x64_windows-clang-cl
 build:win_clang --compiler=clang-cl
 build:win_clang --linkopt=/FORCE:MULTIPLE
 build:win_clang --host_linkopt=/FORCE:MULTIPLE
--- a/.bazelrc
+++ b/.bazelrc
@@ -437,8 +437,8 @@ build:avx_win --copt=/arch:AVX
 # Use Clang-cl compiler on Windows
 build:win_clang --copt=/clang:-Weverything
 build:win_clang --extra_toolchains=@local_config_cc//:cc-toolchain-x64_windows-clang-cl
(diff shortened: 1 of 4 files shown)


.bazelrc
```diff
@@ -437,8 +437,8 @@ build:avx_win --copt=/arch:AVX
 # Use Clang-cl compiler on Windows
 build:win_clang --copt=/clang:-Weverything
 build:win_clang --extra_toolchains=@local_config_cc//:cc-toolchain-x64_windows-clang-cl
-build:win_clang --extra_execution_platforms=//tensorflow/tools/toolchains/win:x64_windows-clang-cl
-build:win_clang --host_platform=//tensorflow/tools/toolchains/win:x64_windows-clang-cl
+build:win_clang --extra_execution_platforms=//tools/toolchains/win:x64_windows-clang-cl
+build:win_clang --host_platform=//tools/toolchains/win:x64_windows-clang-cl
```
penpornk, line 441: @ddunl Could you please help take a look if these .bazelrc changes are okay?

penpornk (changes requested):
Thank you for the PR!

.bazelrc
ddunl: LGTM!

penpornk: Thank you! :)

mraunak:
Hi @penpornk please review the PR, I have made the changes

penpornk (approved):
Thank you for the changes!

.bazelrc
ddunl: On second thought I realized that because we still use TF's bazelrc, we may need to have these under a new config like `win_clang_xla`. Does that sound ok?

mraunak: Hi @ddunl, got your point. I will create a pull request to TF and update TF's .bazelrc with win_clang_xla config and revert the changes (back to build:win_clang --extra_execution_platforms=//tensorflow/tools/toolchains/win:x64_windows-clang-cl
) in XLA bazelrc. How often TF bazelrc is merged to XLA bazelrc? Do you have any documentation for XLA where we can update the steps to run XLA on Windows?

ddunl: I can merge the TF change with this change on our end, I just wanted to make sure that we can use a different config for XLA. I think right now we unfortunately just have to have an extra copy like `win_clang_xla` that has all the same options but just the correct toolchain paths for this repo. If that sounds reasonable then I can do that change with this PR internally.

mraunak: Yeah, that sounds good. Please go ahead and create win_clang_xla with this PR internally

mraunak:
Public PR: https://github.com/openxla/xla/pull/11299

PR #11307, opened 2024-04-08
[ROCm] Upstreaming old changes

mmakevic-amd:
This PR upstreams changes from [#2467](https://github.com/ROCm/tensorflow-upstream/pull/2467/commits/c3d966e559ffd13f6a473d1b57e606314a456ce8) and also upstreams changes to `gpu_layout_assignment.cc` introduced [here](https://github.com/ROCm/tensorflow-upstream/commit/b34b1eb35c095a16208356913a956256cc419e99) and [here](https://github.com/ROCm/tensorflow-upstream/commit/dfb7dd0b2f7e6c6bbcd3f1dbc9e299be9554b4a7).

### DIFF
--- a/xla/service/gpu/gpu_layout_assignment.cc
+++ b/xla/service/gpu/gpu_layout_assignment.cc
@@ -55,6 +55,8 @@ limitations under the License.
 #include "tsl/platform/errors.h"
 #include "tsl/platform/statusor.h"
 
+#include "xla/tsl/util/env_var.h"
+
 namespace xla {
 namespace gpu {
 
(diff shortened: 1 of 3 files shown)


xla/service/gpu/gpu_layout_assignment.cc
```diff
@@ -125,9 +127,10 @@ HeuristicLayoutAssignment(const HloInstruction* instr,
     return kAllNHWC;
   }
 
+  const bool isFloat16 = (input_ty == F16) || (input_ty == BF16);
```
golechwierowicz, line 130: CI is complaining about undefined variable if the non TENSORFLOW_USE_ROCM define is active for a CPU build. Please fix.

xla/service/gpu/gpu_layout_assignment.cc
```diff
@@ -137,9 +140,21 @@ HeuristicLayoutAssignment(const HloInstruction* instr,
       instr->shape().tuple_shapes(0).dimensions_size() != 4) {
     return kAllNCHW;
   }
+#elif TENSORFLOW_USE_ROCM
```
golechwierowicz, line 143: Are these ifdefs necessary? We're trying to minimise ifdefs from the codebase. Can you find a way to get rid of them? See https://github.com/openxla/xla/pull/10788 for inspiration.

mmakevic-amd: I've changed these to perform rocm vs cuda checks during runtime (with 'regular' if-else). Is that better? It also solves undefined variable error for CPU case.

golechwierowicz: Thank you, much better.

golechwierowicz (approved):

PR #10811, opened 2024-03-21
Restore ReshapeDecomposer And LayoutNormalization Passes in GPU Compiler

philipphack:
Restores individual ReshapeDecomposer and LayoutNormalization passes in the GPU compiler previously removed in #9852 and fixes failures in cudnn_norm_rewriter_test.cc.

PR #11309, opened 2024-04-08
Run algebraic simplifier after layout normalization

jaro-sevcik:
The newly introduced layout normalization pass (https://github.com/openxla/xla/pull/10811) leaves redundant bitcasts in the HLO graph. In particular, the bitcasts for the host buffer intializations confuse HostOffloadLegalize pass. The legalization pass tries to ensure that host-offloader’s alias analysis can treat dynamic-update-slice instructions as in-place by duplicating (for each use) the broadcast instructions that will be treated as buffer initializations. However, the layout normalization pass l introduces a single bitcast as a user of the broadcast instruction, so the broadcast does not get duplicated.

This patch runs algebraic simplifier after layout normalization to get rid of the redundant bitcasts.


jaro-sevcik:
Here is a link to a motivation/summary doc: https://docs.google.com/document/d/12LjzvR0foAMnkB7p7-EtiuwdhfmsRAsEWYFTGGiOs7U/edit?usp=sharing

cheshire (approved):

PR #11345, opened 2024-04-09
[XLA:GPU] Relax row reduction vectorization restriction

ghost:
Row reduction did not enable vectorization if num_threads_x <= WarpSize() previously due to half or more of the threads don't do anything. But for current tiling implementation, it is not right. Assume row reduction operand shape is [1024, 256] and num_threads_x is 32 (because of 32 * 16 > 256),  each thread can handle 8 scalar or 4 vec<T, 2>. 
It is better to judge whether (num_threads_x * vector_size) is larger than the reduced minor dimension.


cheshire:
Thanks ! Could you show some micro-benchmarks where it does make a difference?

lingzhi98:
import jax
import jax.numpy as jnp
key = jax.random.key(0)
a = jax.random.uniform(key, shape=[65536, 256], dtype=jax.dtypes.bfloat16)
jax.profiler.start_trace("./benchmark_data")
for _ in range(0, 20):
  b = jnp.sum(a, axis=1, initial=0)
jax.profiler.stop_trace()

This reduction op spend 0.03ms on GPU if not enable vectorization but 0.026ms if enable (test on A100 40GB machine), about 13% improvement. By the way, does openxla have a lightweight tool to test single op performance?

cheshire:
> openxla have a lightweight tool to test single op performance

Sadly we had it but it got lost with an infra migration, should be totally possible to re-add it though, it used to print to stderr time/percentage of roofline for each fusion.

Could you attach an HLO microbenchmark to the commit message with results before/after?

lingzhi98:
I'm not sure if I understand you correctly, please correct me if not.

xla/tools/data/reduce_sum.hlo
```diff
@@ -0,0 +1,14 @@
+HloModule reduce_sum
```
cheshire, line 1: What is the goal of having this HLO in this folder?

cheshire (approved):

lingzhi98:
hi @cheshire 
After relaxing row reduction vectorization restriction, I have found some tests in coalescing_analysis_test will fail. The reason is that coalescing analysis compute memory index for vectorization load incorrectly. There exist 2 bugs in AssignValuesToOuterLoopIVs func:
(1) Replace the last symbol value in affine map
(2) Set the last symbol replacement as itself but not symbol 0
Explain with a specific example: assume the input affine map is (d0)[s0, s1] -> (d0 * 2 + s0 * 64 + s1). Bug 1 will lead to (d0)[s0] -> (d0 * 2), which make memory access inconsistent with reality. If not bug 1 but bug 2, result will be (d0)[s0] -> (d0 * 2 + s1). After fix them, can get correct result (d0)[s0] -> (d0 * 2 + s0).
Can you take a look if have time? I believe this fix is correct, but I think I should tell the reviewer the reason for this latest commit. Thanks!
By the way, I find another bugs of coalescing analysis which are not related about this PR. It is more suitable to push them into another PR than this.

cheshire:
> After relaxing row reduction vectorization restriction, I have found some tests in coalescing_analysis_test will fail

@olegshyshkov could you take a look?

PR #11405, opened 2024-04-10
[gpu] Add address computation fusion for RS->DUS pattern

shraiysh:
This patch recognizes a reduce-scatter operation that is used by a dynamic-update-slice operation. It then fuses this operation to an address-computation fusion.
For example, the following pattern is extracted to a fusion:
```
reduce-scatter = f16[64,128] reduce-scatter(%p0), channel_id=64, replica_groups={{0,1}}, use_global_device_ids=true, dimensions={0}, to_apply=add
dynamic-update-slice = f16[128,128] dynamic-update-slice(%p1, reduce-scatter, %p3, constant.718)
```

### DIFF
--- a/xla/service/gpu/address_computation_fusion_rewriter.cc
+++ b/xla/service/gpu/address_computation_fusion_rewriter.cc
@@ -409,15 +409,19 @@ absl::StatusOr<bool> AddressComputationFusionRewriter::Run(
                       std::pair<UseDefDataflowPaths, DefUseDataflowPaths>>
       matches;
 
-  // Collect all potential custom call matches in the non-fusion computations.
+  // Collect all potential custom call and reduce-scatter --> dus matches in the
+  // non-fusion computations.
   for (HloComputation* computation : module->computations()) {
     if (computation->IsFusionComputation()) continue;
     for (HloInstruction* instr : computation->instructions()) {
-      if (IsLegacyCublasMatmul(*instr) ||
-          (IsCustomCall(instr, platform_name_))) {
-        UseDefDataflowPaths sliced_operand_paths = GetSlicedOperandPaths(instr);
-        bool has_sliced_operand_paths = sliced_operand_paths.size() > 1;
-
+      UseDefDataflowPaths sliced_operand_paths = {instr};
+      bool has_sliced_operand_paths = false;
+      if (IsLegacyCublasMatmul(*instr) || IsCustomCall(instr, platform_name_)) {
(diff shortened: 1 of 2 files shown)


cheshire:
Makes sense to me, @ezhulenev could you also LGTM?

shraiysh:
ping @ezhulenev for review.

xla/service/gpu/address_computation_fusion_rewriter.cc
```diff
@@ -409,15 +409,21 @@ absl::StatusOr<bool> AddressComputationFusionRewriter::Run(
...
           (IsCustomCall(instr, platform_name_))) {
-        UseDefDataflowPaths sliced_operand_paths = GetSlicedOperandPaths(instr);
-        bool has_sliced_operand_paths = sliced_operand_paths.size() > 1;
-
+        sliced_operand_paths = GetSlicedOperandPaths(instr);
+        has_sliced_operand_paths = sliced_operand_paths.size() > 1;
+      }
+      if (instr->opcode() == HloOpcode::kReduceScatter ||
+          IsLegacyCublasMatmul(*instr) ||
+          (IsCustomCall(instr, platform_name_))) {
```
sanjoy, line 426: Minor nit: do you need the parens outside the `IsCustomCall` call?

shraiysh: I dont think so. I will edit this, and change this in the condition on line 417 too.

ezhulenev (approved):

shraiysh:
@ezhulenev Please don't start the merge for this for now. I realized it would be a good idea to test running this e2e. I will test if this fused computation runs smoothly and update here to give a go ahead for the merge.

akuegel:
Sorry, I didn't see your latest reply here and approved the internal change, so it got merged already. But indeed it seems it triggers a problem in a Jax test:

```jaxlib.xla_extension.XlaRuntimeError: INTERNAL: AddressComputationFusion requires a CustomCall hero```

tests/pmap_test.py::VmapPmapCollectivesTest::testWithVmapReduceScatter

So I fear we will have to roll it back for now.

sanjoy:
@akuegel - is this the failing test: https://github.com/google/jax/blob/d18533d558addea0397ac273dbb5ed8f79a8d1ee/tests/pmap_test.py#L1

akuegel:
Yes, that is the one

PR #11422, opened 2024-04-10
[NVIDIA GPU] Remove control knobs for each individual async collective and use the global xla_gpu_disable_async_collectives

Tixxx:
Currently we have 1 global flag(xla_gpu_enable_async_collectives) to control whether we want to asynchronize collectives or not. This flag overrides all other control knobs for each collective. The usage of it is confusing, instead we introduce a new flag xla_gpu_disable_async_collectives which will consolidate all the async flags we have now. We remove xla_gpu_enable_async_collectives and all other individual control knobs.
Sample usage:
xla_gpu_disable_async_collectives=allreduce,reducescatter
disables async allreduce and reducescatter
By default it's empty which indicates enabling async for all collectives.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -126,12 +126,6 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_reduce_scatter_combine_by_dim(true);
 
   opts.set_xla_gpu_enable_async_collectives(true);
-  opts.set_xla_gpu_enable_async_all_reduce(true);
-  opts.set_xla_gpu_enable_async_all_gather(false);
-  opts.set_xla_gpu_enable_async_collective_broadcast(true);
-  opts.set_xla_gpu_enable_async_collective_permute(false);
-  opts.set_xla_gpu_enable_async_all_to_all(false);
-  opts.set_xla_gpu_enable_async_reduce_scatter(false);
 
   opts.set_xla_gpu_enable_reassociation_for_converted_ar(true);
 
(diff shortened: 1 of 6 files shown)


golechwierowicz (approved):
Thanks for the change, I wanted to do it myself for some time :)

Tixxx:
> Thanks for the change, I wanted to do it myself for some time :)

Thanks for the review @golechwierowicz , I'm improving this pr with a better consolidation. Will address your comments and push a new commit today.

Tixxx:
> > Thanks for the change, I wanted to do it myself for some time :)
> 
> Thanks for the review @golechwierowicz , I'm improving this pr with a better consolidation. Will address your comments and push a new commit today.

I have updated this with a better approach. Please take another look. Thanks.

xla/debug_options_flags.cc
```diff
@@ -982,42 +1016,14 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
...
-  flag_list->push_back(tsl::Flag(
-      "xla_gpu_enable_async_all_to_all",
-      bool_setter_for(&DebugOptions::set_xla_gpu_enable_async_all_to_all),
-      debug_options->xla_gpu_enable_async_all_to_all(),
-      "Converts synchronous all-to-all ops into asynchronous."));
+      "xla_gpu_disable_async_collectives",
+      setter_for_xla_gpu_disable_async_collectives,
+      collective_op_types_to_string(
+          debug_options->xla_gpu_disable_async_collectives()),
+      "The types of the commands that are recorded into command buffers. It"
```
golechwierowicz, line 1023: Please change description, seems copy pasted from command buffer flag code.

golechwierowicz (commented):
Please fix the flag description, then LGTM from my side.

@cheshire can you also take a look?

xla/debug_options_flags.cc
```diff
@@ -553,6 +545,48 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
...
+      void operator()(std::string* out, int type) const {
+        absl::StrAppend(out, DebugOptions::CollectiveOpType_Name(type));
+      }
+    };
+    return absl::StrJoin(collective_ops, ", ", Formatter());
+  };
+
+  // Custom parser for xla_gpu_disable_async_collectives.
+  auto setter_for_xla_gpu_disable_async_collectives =
+      [debug_options](const std::string& input) {
```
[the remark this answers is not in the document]
Tixxx: updated

xla/debug_options_flags.cc
Tixxx: good catch :) updated

Tixxx:
adding @frgossen  FYI

xla/debug_options_flags.cc
```diff
@@ -125,14 +125,6 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_all_gather_combine_by_dim(true);
   opts.set_xla_gpu_enable_reduce_scatter_combine_by_dim(true);
 
-  opts.set_xla_gpu_enable_async_collectives(true);
-  opts.set_xla_gpu_enable_async_all_reduce(true);
-  opts.set_xla_gpu_enable_async_all_gather(false);
-  opts.set_xla_gpu_enable_async_collective_broadcast(true);
-  opts.set_xla_gpu_enable_async_collective_permute(false);
-  opts.set_xla_gpu_enable_async_all_to_all(false);
-  opts.set_xla_gpu_enable_async_reduce_scatter(false);
```
frgossen, line 134: Are these default values are not preserved? IIRC, enabling all async collectives by default caused issues in the past.

frgossen (changes requested):

xla/debug_options_flags.cc
Tixxx: The current behavior is that `xla_gpu_enable_async_collectives` overrides all other control knobs, that has been true by default for a while. So this pr wont change the current default settings.

frgossen: Thanks for clarifying

frgossen (approved):

Tixxx:
@golechwierowicz @frgossen  could one of you help take a look the copybara failure? Thanks

frgossen (approved):

frgossen:
Internal tmp import issue. Rerunning it

akuegel (approved):

frgossen:
This breaks internal tests that are still using the old flags. Wdyt about just adding the flag and deprecating the individual ones for now?

Tixxx:
> This breaks internal tests that are still using the old flags. Wdyt about just adding the flag and deprecating the individual ones for now?

hmm I didn't know those flags were still set explicitly. Yea in that case it makes sense just deprecate them instead of removing them from xla.proto. Do you need me to make the changes?

akuegel:
> > This breaks internal tests that are still using the old flags. Wdyt about just adding the flag and deprecating the individual ones for now?
> 
> hmm I didn't know those flags were still set explicitly. Yea in that case it makes sense to just deprecate them instead of removing them from xla.proto. Do you need me to make the changes?

I think it is resolved already, the internal users of the flag were migrated to the new flags.

xla/debug_options_flags.cc
```diff
@@ -982,42 +1016,14 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
...
-      debug_options->xla_gpu_enable_async_all_to_all(),
-      "Converts synchronous all-to-all ops into asynchronous."));
+      "xla_gpu_disable_async_collectives",
+      setter_for_xla_gpu_disable_async_collectives,
+      collective_op_types_to_string(
+          debug_options->xla_gpu_disable_async_collectives()),
+      "This disables a certain set of async collectives and turn them into"
+      " synchornous ones. By default, this is empty which indicates enabling"
+      " async execution for all collectives. A sample usage is: "
+      " --xla_gpu_disable_async_collectives=ALLREDUCE,REDUCESCATTER"));
```
nouiz, line 1026: @Tixxx 
How to disable all collectives?
It would be good to have a list of all collective somewhere, or a special value like "ALL" to disable all of them.

PR #11449, opened 2024-04-11
[XLA:GPU] replace is_causal_mask with enum mask_type to support more cudnn mask generation

Cjkkkk:
* replace is_causal_mask in backend_config with enum mask_type.
* cudnn also supports padding mask and alibi mask.


xla/service/gpu/backend_configs.proto
```diff
@@ -226,8 +226,15 @@ message CudnnfMHABackendConfig {
   // Is flash attention
   bool is_flash_attention = 20;
 
-  // Is causal mask
-  bool is_causal_mask = 21;
```
golechwierowicz, line 230: In principle changing the proto field is not okay. Please add a new field, migrate the existing usages (in the same PR), and remove the field (in the subsequent PR).

Cjkkkk: Not sure if I understand this correctly, you mean just add `is_causal_mask` back in the backend_configs.proto but use `mask_type` for is_causal_mask?

akuegel (approved):

xla/service/gpu/backend_configs.proto
golechwierowicz: Yes, in general https://protobuf.dev/programming-guides/proto3/#deleting.

Here it could be ok, because I checked the usages internally, and compiler is self-contained but I'd rather be safe than sorry here.

Thanks for the change!

golechwierowicz (approved):

PR #11478, opened 2024-04-12
[XLA:GPU] add guards for flash attention graph with cuDNN >= 8.9.4

Cjkkkk:
* Build XLA with cuDNN 8.6 will cause compilation error because flash attention graph will use `CudnnfMHAUid` which is only defined with cuDNN > 8.8.
* Add a guard for flash attention graph with at least cuDNN 8.9.4. So the current logic would be only compile FMHA with cuDNN > 8.8 and only compile flash attention with cuDNN > 8.9.4.

### DIFF
--- a/xla/stream_executor/cuda/cuda_dnn.cc
+++ b/xla/stream_executor/cuda/cuda_dnn.cc
@@ -6348,6 +6348,7 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
     const std::optional<double> dropout_rate, const bool is_causal_mask) {
   using cudnn_frontend::graph::Tensor_attributes;
 
+#if CUDNN_VERSION >= 8904
   if (VLOG_IS_ON(4)) {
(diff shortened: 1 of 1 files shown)


akuegel (approved):

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6348,6 +6348,7 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
     const std::optional<double> dropout_rate, const bool is_causal_mask) {
   using cudnn_frontend::graph::Tensor_attributes;
 
+#if CUDNN_VERSION >= 8904
```
akuegel, line 6351: Just to double-check, version 8.9.4 works? Because in the PR description you are saying it needs to be > 8.9.4

Cjkkkk: Yes, >=8.9.4 should work, sorry about the confusion in the title.

akuegel (approved):

PR #11524, opened 2024-04-15
[ROCM] added flag limiting the number of gemm solutions to be checked & get real IDs for hipblas-lt

pemeliya:
In this PR I add a new XLA flag **xla_gpu_autotune_max_solutions** which limits the number of GEMM solutions to be using for GEMM autotuning. This is a temporary measure while rocBLAS returns way too many solutions which can significantly slow-down autotuning for real applications.

Besides, I also added a "workaround" to obtain the real solution IDs for hipblas-lt which might be later used for deterministic autotuning.

@xla-rotation: would you have a look please ?


jvstokes (approved):
Thank you!

PR #11551, opened 2024-04-16
Dimensional Ordering of Layer Norm Operands

philipphack:
Orders the dimensions of the operands of layer norm Custom Calls as required by cuDNN 9.1 and newer.


philipphack:
CC @reedwm.

reedwm (approved):

PR #11562, opened 2024-04-16
Add a missing dependency on EventInterface

ekuznetsov139:
This adds a missing dependency of xla/backends/interpreter/executor.h. Without it, I am seeing the following compiler error:
```
/usr/include/c++/9/bits/unique_ptr.h: In instantiation of _void std::default_delete<_Tp>::operator()(_Tp*) const [with _Tp = stream_executor::EventInterface]_:             
/usr/include/c++/9/bits/unique_ptr.h:292:17:   required from _std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = stream_executor::EventInterface; _Dp = std::default_delete<stream_executor::EventInterface>]_                                                                                                                                        
./xla/backends/interpreter/executor.h:161:12:   required from here
/usr/include/c++/9/bits/unique_ptr.h:79:16: error: invalid application of _sizeof_ to incomplete type _stream_executor::EventInterface_ 
```
I assume this is due to a difference in behavior between compilers (some compilers allow construction of a std::unique_ptr to an incomplete type, and some don't.)

### DIFF
--- a/xla/backends/interpreter/BUILD
+++ b/xla/backends/interpreter/BUILD
@@ -151,6 +151,7 @@ cc_library(
         "//xla:xla_data_proto_cc",
         "//xla/stream_executor",
         "//xla/stream_executor:stream_executor_interface",
+        "//xla/stream_executor:event_interface",
         "//xla/stream_executor/host:host_stream",
         "@com_google_absl//absl/functional:any_invocable",
         "@com_google_absl//absl/log",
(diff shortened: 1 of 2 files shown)


kamaljeeti (approved):

akuegel:
Already fixed in https://github.com/openxla/xla/pull/11557

PR #11563, opened 2024-04-16
[NVIDIA GPU] Improve GPU collective matmul to support all-gather having multiple users

Tixxx:
We have identified another optimization opportunity for gpt-3 using collective matmul, in the backward pass, the all-gather has multiple dot users but current spmd will duplicate multiple collective matmul loops. We'd like this transformation:
before:
```
  //                       input
  //                       /    |
  //                      /     |
  //                     AG    windowed loop
  //                     /
  //                    /
  //                   dot

```
after:
```
  //                       input
  //                       |
  //                       |
  //                     windowed loop
  //                       | 
  //                       |
  //                      dot
```
This is advantageous since the chained dot can fully utilize all the resource on the GPU while comm is hidden by the first collective matmul loop.

We introduced an option to turn off CM loop duplication in SPMD and rewrite the graph to desired pattern in the gpu_windowed_einsum_handler pass.

### DIFF
--- a/xla/service/gpu/gpu_windowed_einsum_handler.cc
+++ b/xla/service/gpu/gpu_windowed_einsum_handler.cc
@@ -143,6 +143,83 @@ absl::StatusOr<bool> HandleAgWindowedEinsumLoop(HloComputation* comp,
   return changed;
 }
 
+absl::Status ProcessWindowedEinsumLoopForActivationCaching(
+    GpuWindowedEinsumHandler::WindowedEinsumAgLoops& ag_loop) {
+  HloInstruction* loop = ag_loop.loop;
+  // Transform the while body to cache the allgathered result in the
+  // output buffer to be consumed by the dot
+  HloComputation* while_body = loop->while_body();
+  HloInstruction* input_gte;
+  for (HloInstruction* gte : while_body->parameter_instruction(0)->users()) {
+    if (gte->tuple_index() == 0) {
+      input_gte = gte;
+    }
+  }
+  // Get the output operand of the full buffer.
+  HloInstruction* root = while_body->root_instruction();
+  // The full buffer that we will use to cache the accumulated activation
+  // is the 4th operand in the output tuple.
+  int64_t full_cache_buffer_index = 3;
+  HloInstruction* full_buffer_output_gte =
+      root->mutable_operand(full_cache_buffer_index);
+  HloInstruction* new_full_buffer_output;
+  // Find the DUS in the loop body and re-use the slice indices
+  // This should just be a constant(0)
+  HloInstruction* dus_boundary_constant;
+  for (HloInstruction* inst : while_body->MakeInstructionPostOrder()) {
+    HloInstruction* slice_indices;
+    // If we have a DUS(PARAM,DS) pattern, we need to update the output
+    // buffer with the first slice.
+    if (Match(inst,
+              m::DynamicUpdateSlice(
+                  m::GetTupleElement(m::Parameter()), m::Op(),
+                  m::Constant(&dus_boundary_constant),
+                  m::Reshape(m::DynamicSlice(&slice_indices, m::Op(), m::Op())),
+                  m::Op()))) {
+      slice_indices = while_body->AddInstruction(HloInstruction::CreateReshape(
+          dus_boundary_constant->shape(), slice_indices));
+      VLOG(5) << "Created slice op for first slice: "
+              << slice_indices->ToString();
+      full_buffer_output_gte =
+          while_body->AddInstruction(HloInstruction::CreateDynamicUpdateSlice(
+              full_buffer_output_gte->shape(), full_buffer_output_gte,
+              input_gte,
+              {dus_boundary_constant, slice_indices, dus_boundary_constant}));
+    }
+    // If we have a DUS(DUS,DS) pattern, then the einsum loop is
+    // unrolled, we need to update the output buffer again with the
+    // second slice. Since the second slice will have different indices,
+    // we need to re-capture slice_indices.
+    if (Match(inst,
+              m::DynamicUpdateSlice(
+                  m::DynamicUpdateSlice(), m::Op(), m::Constant(),
+                  m::Reshape(m::DynamicSlice(&slice_indices, m::Op(), m::Op())),
+                  m::Op()))) {
+      slice_indices = while_body->AddInstruction(HloInstruction::CreateReshape(
+          dus_boundary_constant->shape(), slice_indices));
+      VLOG(5) << "Created slice op for second slice: "
+              << slice_indices->ToString();
+      // The slice we need this time is the output of the first
+      // collective-permute
+      HloInstruction* cp_output;
+      for (HloInstruction* gte_user : input_gte->users()) {
+        if (gte_user->opcode() == HloOpcode::kCollectivePermute) {
+          cp_output = gte_user;
+          break;
+        }
+      }
+      new_full_buffer_output =
+          while_body->AddInstruction(HloInstruction::CreateDynamicUpdateSlice(
+              full_buffer_output_gte->shape(), full_buffer_output_gte,
+              cp_output,
+              {dus_boundary_constant, slice_indices, dus_boundary_constant}));
+    }
+  }
+  TF_RETURN_IF_ERROR(root->ReplaceOperandWith(full_cache_buffer_index,
+                                              new_full_buffer_output));
+  return OkStatus();
+}
+
 class WindowedEinsumVisitor : public DfsHloRewriteVisitor {
  public:
   explicit WindowedEinsumVisitor(
@@ -157,30 +234,25 @@ class WindowedEinsumVisitor : public DfsHloRewriteVisitor {
     for (GpuWindowedEinsumHandler::WindowedEinsumAgLoops ag_loop :
          all_ag_loops_) {
       HloInstruction* loop = ag_loop.loop;
-      HloInstruction* lhs_ag = nullptr;
-      HloInstruction* rhs_ag = nullptr;
+      HloInstruction* ag_operand = nullptr;
 
-      if (Match(dot, m::Dot(m::AllGather(&lhs_ag), m::Op())) ||
-          Match(dot, m::Dot(m::Op(), m::AllGather(&rhs_ag)))) {
+      if (Match(dot, m::Dot(m::AllGather(&ag_operand), m::Op())) ||
+          Match(dot, m::Dot(m::Op(), m::AllGather(&ag_operand)))) {
         HloInstruction* windowed_lhs =
             loop->mutable_operand(0)->mutable_operand(0);
         HloInstruction* ag_with_shared_operand = nullptr;
-        if (lhs_ag && lhs_ag->mutable_operand(0) == windowed_lhs) {
-          ag_with_shared_operand = lhs_ag;
-        }
-
-        if (rhs_ag && rhs_ag->mutable_operand(0) == windowed_lhs) {
-          ag_with_shared_operand = rhs_ag;
+        if (ag_operand && ag_operand->mutable_operand(0) == windowed_lhs) {
+          ag_with_shared_operand = ag_operand;
         }
 
         if (!ag_with_shared_operand) {
-          return absl::OkStatus();
+          continue;
         }
 
         VLOG(5) << "Found all-gather that shares the same operand with a "
-                      "windowed einsum loop : "
-                   << loop->ToString();
-        int64_t cache_output_index = ag_with_shared_operand == lhs_ag ? 0 : 1;
+                   "windowed einsum loop : "
+                << loop->ToString();
+        int64_t cache_output_index = dot->operand_index(ag_with_shared_operand);
         HloComputation* comp = dot->parent();
         HloInstruction* new_gte = comp->AddInstruction(
             HloInstruction::CreateGetTupleElement(loop, 3));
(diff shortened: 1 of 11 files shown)


xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -142,6 +143,134 @@ absl::StatusOr<bool> HandleAgWindowedEinsumLoop(HloComputation* comp,
...
+  // and remove the all-gather.
+  absl::Status HandleDot(HloInstruction* dot) override {
+    CHECK_EQ(dot->opcode(), HloOpcode::kDot);
+    for (GpuWindowedEinsumHandler::WindowedEinsumAgLoops ag_loop :
+         all_ag_loops_) {
+      HloInstruction* loop = ag_loop.loop;
+      HloInstruction* lhs_ag = nullptr;
+      HloInstruction* rhs_ag = nullptr;
+
+      if (Match(dot, m::Dot(m::AllGather(&lhs_ag), m::Op())) ||
```
cheshire, line 163: Is it possible to reduce LHS/RHS code duplication?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -142,6 +143,134 @@ absl::StatusOr<bool> HandleAgWindowedEinsumLoop(HloComputation* comp,
...
+                      "windowed einsum loop : "
+                   << loop->ToString();
+        int64_t cache_output_index = ag_with_shared_operand == lhs_ag ? 0 : 1;
+        HloComputation* comp = dot->parent();
+        HloInstruction* new_gte = comp->AddInstruction(
+            HloInstruction::CreateGetTupleElement(loop, 3));
+        TF_RETURN_IF_ERROR(
+            dot->ReplaceOperandWith(cache_output_index, new_gte));
+        TF_RETURN_IF_ERROR(comp->RemoveInstruction(ag_with_shared_operand));
+        if (!ag_loop.consumed) {
```
cheshire, line 190: Maybe extract it into a function?

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -193,5 +197,146 @@ ENTRY main.9_spmd {
...
+  EXPECT_TRUE(changed);
+
+  HloInstruction* ag_loop =
+      FindInstructionByName(module->entry_computation(), "while");
+  HloInstruction* inst =
+      FindInstructionByName(module->entry_computation(), "dot.7");
+  // dot.7 should now consume output of the windowed einsum while loop.
+  EXPECT_TRUE(inst->operand(0)->opcode() == HloOpcode::kGetTupleElement);
+  EXPECT_TRUE(inst->operand(0)->tuple_index() == 3);
+  EXPECT_TRUE(inst->operand(0)->operand(0) == ag_loop);
```
cheshire, line 323: Is there an execution test?

cheshire (changes requested):

xla/service/gpu/gpu_windowed_einsum_handler.cc
frgossen, line 163: Use `continue` and reduce nesting?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -142,6 +143,134 @@ absl::StatusOr<bool> HandleAgWindowedEinsumLoop(HloComputation* comp,
...
+          HloInstruction* input_gte;
+          for (HloInstruction* gte :
+               while_body->parameter_instruction(0)->users()) {
+            if (gte->tuple_index() == 0) {
+              input_gte = gte;
+            }
+          }
+          // Get the output operand of the full buffer.
+          HloInstruction* root = while_body->root_instruction();
+          HloInstruction* full_buffer_output_gte = root->mutable_operand(3);
```
frgossen, line 203: Does this have a named accessor?

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -193,5 +197,146 @@ ENTRY main.9_spmd {
       cp1->backend_config<GpuBackendConfig>()->force_earliest_schedule());
 }
 
+TEST_F(GpuWindowedEinsumHanlderTest, AgLoopsMultipleConsumersAreChained) {
```
frgossen, line 200: Would it be possible to have a minimal test case?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -163,9 +292,37 @@ absl::StatusOr<bool> GpuWindowedEinsumHandler::Run(
       VLOG(5) << "Processing computation: " << comp->name();
       TF_ASSIGN_OR_RETURN(bool comp_result,
                           HandleAgWindowedEinsumLoop(comp, stream_id));
+      all_ag_loops_.push_back(
+          WindowedEinsumAgLoops(comp->WhileCallInstruction()));
       changed = comp_result;
     }
   }
+  // Now that we have processed all loops, we can check if there are any
```
frgossen, line 300: Can you explain in a bit more detail why this is beneficial?

kamaljeeti:
Hi @Tixxx , can you have a look at the comments? Thanks.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: updated

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
Tixxx: updated

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: updated

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
Tixxx: good point, cut down the test case hlo to the minimum.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: added more description on why this is better.

Tixxx:
HI @seherellis  would you be able to take a look at the changes in SPMD partitioner? Most of the logic is in the gpu pass. Thanks.

xla/service/spmd/dot_handler.cc
```diff
@@ -1909,22 +1909,56 @@ absl::StatusOr<HloInstruction*> PartitionBaseCase(
     }
   }
 
+  // If we see a dot that shares the same operand with a windowed einsum ag loop
+  // and disable_ag_rewrite_for_multiple_consumers is true. We skip rewriting
+  // the current dot.
+  bool should_skip_windowed_einsum = false;
+  if (options.disable_ag_rewrite_for_multiple_consumers) {
+    auto lhs_operand = lhs.hlo()->opcode() == HloOpcode::kReshape ||
```
seherellis, line 1917: Why are these instructions skipped? Please add a comment explaining why.

seherellis (changes requested):

xla/service/spmd/dot_handler.cc
Tixxx: added a comment for this logic. We skip those since they are optional and dont change the functional meaning of the ag-dot pattern, essentially we can support ag-(reshape)-dot.

xla/service/spmd/spmd_partitioner_test.cc
```diff
@@ -5222,6 +5238,39 @@ ENTRY entry {
...
+      PartitionComputation(hlo_string,
+                           /*num_devices=*/4,
+                           /*conv_halo_exchange_always_on_lhs =*/true,
+                           /*choose_faster_windowed_einsum =*/false,
+                           /*unroll_windowed_einsum =*/false,
+                           /*bidirectional_windowed_einsum=*/false,
+                           /*threshold_for_windowed_einsum_mib=*/0,
+                           /*skip_checking_windowed_einsum_users=*/true,
+                           /*disable_ag_rewrite_for_multiple_consumers=*/true));
+  EXPECT_EQ(CountInstructions(*module->entry_computation(), HloOpcode::kWhile),
```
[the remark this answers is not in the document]
Tixxx: Yes, added a comment to explain this.

xla/service/spmd/spmd_partitioner_test.cc
```diff
@@ -5222,6 +5238,39 @@ ENTRY entry {
...
 
+TEST_P(SpmdPartitioningTest, EinsumDisableRewriteForAgWithMultipleConsumers) {
+  absl::string_view hlo_string = R"(
+HloModule test, entry_computation_layout={(bf16[2,2048,24576]{2,1,0}, bf16[24576,98304]{1,0}, bf16[24576,98304]{1,0})->bf16[2,2048,98304]{2,1,0}}, num_partitions=4
+
+ENTRY main {
+  Arg_0.1 = bf16[2,2048,24576]{2,1,0} parameter(0), sharding={devices=[1,4,1]<=[4]}
+  Arg_1.2 = bf16[24576,98304]{1,0} parameter(1), sharding={devices=[1,4]<=[4]}
+  dot.5 = bf16[2,2048,98304]{2,1,0} dot(Arg_0.1, Arg_1.2), lhs_contracting_dims={2}, rhs_contracting_dims={0}, sharding={devices=[1,1,4]<=[4]}
+  copy = bf16[2,2048,98304]{2,1,0} copy(dot.5), sharding={devices=[1,1,4]<=[4]}
```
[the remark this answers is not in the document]
Tixxx: ah this is from my jax unit test, it adds a copy there for some reason, i will remove it.

Tixxx:
@seherellis  here's the e2e execution test, CollectiveOpsTestE2E.WindowedEinsumE2EAllgatherMultiConsumer under //xla/tests:collective_ops_test_e2e. It goes through the entire gpu pipeline including spmd partitioner and gpu CM handler.

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
Tixxx: CollectiveOpsTestE2E.WindowedEinsumE2EAllgatherMultiConsumer under //xla/tests:collective_ops_test_e2e

Tixxx:
@frgossen @cheshire  @seherellis  could you please take a look at the updated PR again? Thanks

thomasjoerg:
@frgossen friendly ping!

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -163,9 +294,40 @@ absl::StatusOr<bool> GpuWindowedEinsumHandler::Run(
...
+  //                      /     |
+  //                     AG    windowed loop
+  //                     /
+  //                    /
+  //                   dot
+  // to:
+  //                       input
+  //                       |
+  //                       |
+  //                     windowed loop
```
frgossen, line 315: How does this work? Are you somehow reusing AGs from the loop nest and pipe them through as a result?

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
frgossen: I still have a hard time following this transformation. If it is as simple as depicted in the diagram, can we have an additional test case with just those few operations? I think a hand-crafted HLO test case to document this transformation would really help.

frgossen (changes requested):

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: yes, we gather all slices when we communicate in a ring-fashion, after the whole loop is done, the output buffer will contain the complete activation. This achieves the same result as running AG before the collective matmul transformation.

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
Tixxx: So the pass alone is trying to do 2 things:
1. look for an already rewritten windowed einsum loop
2. check if the loop is sharing the same operand with another dot
When these 2 conditions are met, we will perform the transformation in the diagram, so we need to have a rewritten windowed einsum loop in the graph to test this pass. In order to test just the operations in the diagram, we'd need to call the SPMD partitioner in this test. I have written an e2e test for this here `CollectiveOpsTestE2E.WindowedEinsumE2EAllgatherMultiConsumer` under `//xla/tests:collective_ops_test_e2e`

frgossen (approved):
Thank you!

xla/service/spmd/spmd_partitioner_test.cc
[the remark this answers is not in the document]
seherellis: I just tried running this and the resulting HLO is something like:

...
while
...
dus = dynamic-update-slice(..)
ar = all-reduce(dus)
dot.2 = dot(all-reduce)
ROOT add.4 = add(gte_from_while, dot.2)

So I am not sure if this example represents your use case as the test in gpu_windowed_einsum_handler_test expects to see a different pattern with an all-gather.

Shouldn't we have the same HLO in these tests?

seherellis (changes requested):
Left a comment here about the tests:
https://github.com/openxla/xla/pull/11563/files#r1591989174

xla/service/spmd/spmd_partitioner_test.cc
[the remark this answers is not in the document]
Tixxx: ah ok, i think this is due to how base spmd partitioner deals with the sharding. the allgather is added when i use StatefulRngSpmdPartitioner. I can move the test over there

Tixxx:
> Left a comment here about the tests: https://github.com/openxla/xla/pull/11563/files#r1591989174

@seherellis  I moved the unit test to stateful_rng_spmd_partitioner_test, it should have the expected graph after partitioner.

seherellis (approved):
Approving to see the changes on our side..

xla/tests/collective_ops_test_e2e.cc
```diff
@@ -645,5 +645,81 @@ TEST_F(CollectiveOpsTestE2E, NoAllToAllDecomposition) {
   LiteralTestUtil::ExpectR1Equal<uint32_t>({20, 25, 21, 26}, results[1]);
 }
 
+TEST_F(CollectiveOpsTestE2E, WindowedEinsumE2EAllgatherMultiConsumer) {
```
seherellis, line 648: Could you please verify that this test goes through the newly added code path? 

If possible (in your testing infrastructure), you can dump the HLO after each pass and (manually) check the HLO right before and after the new pass. (You need environment variable XLA_FLAGS=--xla_dump_to=<directory> --xla_dump_hlo_pass_re=.* to dump the HLO after each pass.)

Tixxx: Thanks @seherellis  for pointing it out. I verified the hlo when i first added this test but I think there was an issue when I was asked to minimize this test, when i removed the gemm+rs from the original hlo, the allgather didnt happen on the shared operand for the second dot anymore, in turn spmd partitioner decided to put ag on the other non-shared operand. I have added back the gemm+rs to the hlo and now it triggers the code path.
I also reduced the size of tensors so this runs a bit faster.
I have also verified the hlo again that the pattern is expected.
It'd be good to keep the gemm+rs pattern in there since we didnt really have any e2e for that before.

seherellis (approved):

PR #11567, opened 2024-04-16
Allowing permutations of terms in //xla/service/gpu/fusions:transpose_mlir_test

ekuznetsov139:
This expands the pattern matcher used by  //xla/service/gpu/fusions:transpose_mlir_test, so that it reports a match even if some terms are permuted. (Without the change, I see two subtests of this test fail on ROCm. With the change, the test passes there.)


pifon2a (approved):
I think that the reason, why we get different permutations is because we are not using the upstream simplifier for affine maps/exprs `mlir::simplifyAffineExpr` in AffineExpr.h. That simplifier also sorts the terms. But it also does some ugly things and we cannot use it. I think the right fix would be to extract the sorting logic and apply it for the indexing maps. I will try to do that this/next week. In the meantime, let's land this one.

pifon2a (approved):

akuegel (approved):

PR #11568, opened 2024-04-16
Improve NVTX ranges emitted during compilation

olupton:
New ranges:
 - XlaBufferAssignment
 - XlaCompileCudnnFusion
 - XlaCompileGpuAsm
 - XlaCreateGpuExecutable
 - XlaDumpHloModule
 - XlaDumpLlvmIr
 - XlaEmitGpuAsm
 - XlaEmitLlvmIr
 - XlaMemoryScheduler
 - XlaOptimizeLlvmIr

Tweaked XlaCompile, XlaCompileBackend and XlaPass to improve consistency. To allow more explicit range names, the `llvm::Module` is now set to the corresponding HLO module name, where previously it was an empty string.

Because part of #9896 was reverted, `GOOGLE_CUDA` needs to be defined in more translation units to make the NVTX ranges appear. This is done in a separate PR, https://github.com/openxla/xla/pull/11611, on the request of @cheshire.

### DIFF
--- a/xla/service/service.cc
+++ b/xla/service/service.cc
@@ -748,7 +748,8 @@ absl::StatusOr<std::unique_ptr<Executable>> Service::BuildExecutable(
       module_proto.name());
 
   tsl::profiler::ScopedAnnotation annotation{[&] {
-    return absl::StrCat("XlaCompile:#module=", module_proto.name(), "#");
+    // module's unique_id is not available yet
+    return absl::StrFormat("XlaCompile:#module=%s#", module_proto.name());
   }};
 
   TF_ASSIGN_OR_RETURN(
@@ -771,9 +772,6 @@ absl::StatusOr<std::unique_ptr<Executable>> Service::BuildExecutable(
                                     std::move(module), executor, options));
   }
 
-  tsl::profiler::ScopedAnnotation backend_annotation{[&] {
-    return absl::StrCat("XlaCompileBackend:#module=", module_proto.name(), "#");
-  }};
   TF_ASSIGN_OR_RETURN(
       std::unique_ptr<Executable> executable,
       backend->compiler()->RunBackend(std::move(module), executor, options));
--- a/xla/service/BUILD
+++ b/xla/service/BUILD
@@ -556,6 +556,7 @@ cc_library(
         "@tsl//tsl/platform:path",
         "@tsl//tsl/platform:regexp",
         "@tsl//tsl/platform:status",
+        "@tsl//tsl/profiler/lib:scoped_annotation",
     ],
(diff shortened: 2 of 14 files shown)


xla/service/service.cc
```diff
@@ -748,7 +748,8 @@ absl::StatusOr<std::unique_ptr<Executable>> Service::BuildExecutable(
       module_proto.name());
 
   tsl::profiler::ScopedAnnotation annotation{[&] {
-    return absl::StrCat("XlaCompile:#module=", module_proto.name(), "#");
+    // module's unique_id is not available yet
```
olupton, line 751: This is a bit annoying, because it means not all `XlaCompile` annotations look the same as each other, but it looks messy to predict what the unique_id will be on this line.

cheshire (commented):
Overall looks cool! Two questions:

1) Do we need all the local_defines changes?

2) Do we need that many different markers (e.g. unclear if we need the one for creating an executable?)

olupton:
1. yes, #9896 was supposed to avoid the need for this, and it did land, but https://github.com/openxla/xla/commit/c4b031bf9522685f998ecbd12570f29b391495ec rolled back an important part of the changeset, so until that can be rolled forward again then we need `GOOGLE_CUDA` everywhere. IIUC that would need internal investigation at Google.

2. the markers I added were all motivated by being non-trivial in some profile or other that I checked, so I would be inclined to keep them all (e.g. in the profile I have open right now, `XlaCreateGpuExecutable` is bigger than `XlaOptimizeLlvmIr`)

cheshire:
> yes, https://github.com/openxla/xla/pull/9896 was supposed to avoid the need for this, and it did land, but https://github.com/openxla/xla/commit/c4b031bf9522685f998ecbd12570f29b391495ec rolled back an important part of the changeset, so until that can be rolled forward again then we need GOOGLE_CUDA everywhere. IIUC that would need internal investigation at Google

But do we need to do it in this PR, or could we do it in a separate one?

olupton:
> > yes, #9896 was supposed to avoid the need for this, and it did land, but [c4b031b](https://github.com/openxla/xla/commit/c4b031bf9522685f998ecbd12570f29b391495ec) rolled back an important part of the changeset, so until that can be rolled forward again then we need GOOGLE_CUDA everywhere. IIUC that would need internal investigation at Google
> 
> But do we need to do it in this PR, or could we do it in a separate one?

OK, I kept the code changes in this PR and opened https://github.com/openxla/xla/pull/11611 with the `GOOGLE_CUDA` changes.

cheshire (approved):

PR #11592, opened 2024-04-16
Add a TritonFusionAnalysis util to query an op's scope.

elfiegg:
Add a TritonFusionAnalysis util to query an op's scope.
This will be used in cuDNN compiler for deciding an op's dimensions in the followup PR.

### DIFF
--- a/xla/service/gpu/triton_fusion_analysis.cc
+++ b/xla/service/gpu/triton_fusion_analysis.cc
@@ -320,6 +320,17 @@ absl::Status TritonFusionAnalysis::ExecuteForDotFusion(
   return absl::OkStatus();
 }
 
+std::optional<TritonFusionAnalysis::Scope>
+TritonFusionAnalysis::QueryInstructionScope(const HloInstruction& hlo) const {
+  for (const Scope& scope : {Scope::LHS, Scope::RHS, Scope::OUTPUT}) {
+    if (iter_specs_.at(scope).count(&hlo) > 0) {
+      return scope;
+    }
+  }
+  LOG(WARNING) << "No scope for hlo: " << hlo.ToString();
+  return std::nullopt;
+}
+
 const TensorIterationSpec::DimIterationSpec* TritonFusionAnalysis::IterSpec(
     const TritonFusionAnalysis::Scope scope, const HloInstruction* hlo,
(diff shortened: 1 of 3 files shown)


xla/service/gpu/triton_fusion_analysis.cc
```diff
@@ -320,6 +320,16 @@ absl::Status TritonFusionAnalysis::ExecuteForDotFusion(
...
 }
 
+TritonFusionAnalysis::Scope TritonFusionAnalysis::QueryInstructionScope(
+    const HloInstruction& hlo) const {
+  for (const Scope& scope : {Scope::LHS, Scope::RHS, Scope::OUTPUT}) {
+    if (iter_specs_.at(scope).count(&hlo) > 0) {
+      return scope;
+    }
+  }
+  LOG(FATAL) << "No scope for hlo: " << hlo.ToString();
```
cheshire, line 330: LOG(FATAL) seems a bit much, maybe use optional? Do we really want to crash a process there?

cheshire (changes requested):

xla/service/gpu/triton_fusion_analysis.cc
elfiegg: Maybe we can leave to the util-user to decide whether to crash it. The util will just return nullopt if no scope is found. Done.

xla/service/gpu/triton_fusion_analysis.cc
```diff
@@ -320,6 +320,17 @@ absl::Status TritonFusionAnalysis::ExecuteForDotFusion(
...
 }
 
+std::optional<TritonFusionAnalysis::Scope>
+TritonFusionAnalysis::QueryInstructionScope(const HloInstruction& hlo) const {
+  for (const Scope& scope : {Scope::LHS, Scope::RHS, Scope::OUTPUT}) {
+    if (iter_specs_.at(scope).count(&hlo) > 0) {
+      return scope;
+    }
+  }
+  LOG(WARNING) << "No scope for hlo: " << hlo.ToString();
```
cheshire, line 330: Nit: could it be too much log spam? Either it's an error and should be propagated, or it isn't.

cheshire (approved):

PR #11597, opened 2024-04-17
Compile F32, BF16 and S32 constants in cuDNN graph.

elfiegg:
Support compiling F32, BF16 and S32 constants in cuDNN graph.
This PR is blocked until cuDNN frontend dependency is upgraded to v1.3.

### DIFF
--- a/xla/service/gpu/cudnn_fusion_compiler.cc
+++ b/xla/service/gpu/cudnn_fusion_compiler.cc
@@ -174,7 +174,7 @@ int FusionLevel(const HloInstruction& hlo) {
 class GemmDimensionAdapter {
   explicit GemmDimensionAdapter(const HloDotInstruction& dot,
                                 TritonFusionAnalysis analysis)
-      : analysis_(std::move(analysis)), dot_(dot) {};
+      : analysis_(std::move(analysis)), dot_(dot){};
 
  public:
   const TritonFusionAnalysis analysis_;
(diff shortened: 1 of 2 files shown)


xla/service/gpu/cudnn_fusion_compiler.cc
```diff
@@ -311,6 +311,34 @@ class GemmDimensionAdapter {
...
+  if (!ShapeUtil::IsScalar(hlo.shape())) {
+    VLOG(3) << "Currently only support fusing scalar in the graph";
+    return std::nullopt;
+  }
+  PrimitiveType constant_type = hlo.shape().element_type();
+  switch (constant_type) {
+    case BF16:
+      return graph.tensor(__nv_bfloat16(
+          hlo.literal()
+              .data<primitive_util::PrimitiveTypeToNative<BF16>::type>()[0]));
```
cheshire, line 326: Nit:

1. Could we avoid data duplication? Both `hlo.literal` and `graph.tensor` calls should not be duplicated.
2. `literal.GetFirstElement<NativeType>()` seems more suitable

xla/service/gpu/fusions/cudnn_test.cc
```diff
@@ -479,6 +479,32 @@ ENTRY e {
                             ErrorSpec{/*aabs=*/1e-3, /*arel=*/1e-3}));
 }
 
+TEST_F(CuDnnFusionLevel2Test, FuseConstantExecutesCorrectly) {
+  EXPECT_TRUE(RunAndCompare(R"(
```
cheshire, line 483: Shouldn't we have both execution tests, and fusion tests? (the latter checking that fusion has actually occurred?)

cheshire (changes requested):

xla/service/gpu/cudnn_fusion_compiler.cc
elfiegg: Done.

xla/service/gpu/fusions/cudnn_test.cc
elfiegg: The entire fusion HLO will be serialized into a cudnn graph, and if cudnn can't handle it, cudnn will return a rejection. I didn't get what "the latter checking that fusion has actually occurred" means? Maybe my naming is confusing, changed to ConstantExecutesCorrectly.

xla/service/gpu/triton_fusion_analysis.cc
```diff
@@ -320,6 +320,17 @@ absl::Status TritonFusionAnalysis::ExecuteForDotFusion(
...
 }
 
+std::optional<TritonFusionAnalysis::Scope>
+TritonFusionAnalysis::QueryInstructionScope(const HloInstruction& hlo) const {
+  for (const Scope& scope : {Scope::LHS, Scope::RHS, Scope::OUTPUT}) {
+    if (iter_specs_.at(scope).count(&hlo) > 0) {
+      return scope;
+    }
+  }
+  LOG(WARNING) << "No scope for hlo: " << hlo.ToString();
```
cheshire, line 330: Should this diff be here?

xla/service/gpu/fusions/cudnn_test.cc
cheshire: This PR is adding new fusion capability, right?

You are testing that constructed fusions are executing correctly, but you are not testing that those fusions (`fusion1` in your example) would actually be built?

cheshire (changes requested):

elfiegg:
Updated the title of the PR.

cheshire:
OK I think I understand: it adds cuDNN support, but does not construct such fusions (yet?).

Overall looks good modulo remaining comments.

elfiegg:
Done, thanks!

cheshire (approved):

PR #11603, opened 2024-04-17
[ROCm] Configure pjrt gpu plugin for rocm

mmakevic-amd:
Configure PJRT GPU plugin so it can be built for ROCm as well.


i-chaochen:
@xla-rotation please review it. Thanks!

ddunl (approved):

PR #11631, opened 2024-04-17
Cleanup passes names

apivovarov:
I found couple passes with simple `absl::string_view name()` implementation which were defined in cc files.
All other passes define `name()` in header file. (Unless name() impl requires if/switch block)

This PR makes simple `absl::string_view name()` implementation consistent across all passes.


akuegel (approved):
Thank you

apivovarov:
Seems that this PR is ready for merge.

PR #11636, opened 2024-04-18
Add docs/passes.md

apivovarov:
This PR adds passes.md document ([preview](https://github.com/apivovarov/openxla-xla/blob/passes_doc/docs/passes.md)) which lists all existing XLA passes (Total 182 passes)

The passes are structured as follows:
- XLA Service CPU passes
- XLA Service GPU passes
- XLA Service SPMD passes
- XLA Service passes
- XLA HLO Transforms passes
- XLA HLO Experimental passes
- XLA Tools passes

I also included descriptions for some of the first passes

Related Discussion: https://groups.google.com/a/openxla.org/g/openxla-discuss/c/17YoGTDW9ps/m/2lJMAvJkAQAJ?utm_medium=email&utm_source=footer

@cheshire @elliotenglish

### DIFF
--- a/docs/passes.md
+++ b/docs/passes.md
@@ -1,0 +1,1356 @@
+# XLA Passes
+
+The following describes existing XLA passes
+
+## XLA Service passes
+
+### algsimp
+
+See also
+[`xla::AlgebraicSimplifier`](https://github.com/openxla/xla/tree/main/xla/service/algebraic_simplifier.h).
+
+A pass which performs algebraic simplifications.
+
+### all-gather-bcast-reorder
+
+See also
+[`xla::AllGatherBroadcastReorder`](https://github.com/openxla/xla/tree/main/xla/service/all_gather_broadcast_reorder.h).
+
+A pass that reorders `all-gather(broadcast(x)) -> broadcast(all-gather(x))`. The intent is to reduce the size of `all-gather` when possible by doing an `all-gather` on the (smaller) pre-broadcasted data and then applying the broadcast.
+
+### all-gather-combiner
+
+See also
+[`xla::AllGatherCombiner`](https://github.com/openxla/xla/tree/main/xla/service/all_gather_combiner.h).
+
+Combines small non-dependent AllGather ops into larger combined AllGather ops. A typical AllGather implementation has a minimum latency-induced time for a AllGather op so a single combined op can be more efficient than many small ones.
+
+### all-reduce-combiner
+
+See also
+[`xla::AllReduceCombiner`](https://github.com/openxla/xla/tree/main/xla/service/all_reduce_combiner.h).
+
+Combines small non-dependent AllReduce ops into larger combined AllReduce ops. A typical AllReduce implementation has a minimum latency-induced time for a AllReduce op so a ingle combined op can be more efficient than many small ones.
+
+### all-reduce-contiguous
+
+See also
+[`xla::AllReduceContiguous`](https://github.com/openxla/xla/tree/main/xla/service/all_reduce_contiguous.h).
+
+Concatenates `all-reduce` operands together, so the `all-reduce` is performed over a single, contiguous buffer.
+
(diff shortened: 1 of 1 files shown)


cheshire:
High-level comments:

1) I don't think TODO stubs are useful.

2) Could we aim for something more coarse-grained? We don't need to (and probably don't want to, since they keep changing) document every pass, but we should document conceptual groups of passes.

apivovarov:
The list of passes is created automatically by a specific header files parser.
Currently it can extract:
- Pass Class name
- Pass name
- Link to header file

In order to automatically capture pass description we can agree that pass description will be a comment block above the class declaration.

Example (from [xla/service/all_reduce_folder.h](https://github.com/openxla/xla/tree/main/xla/service/all_reduce_folder.h)):
```cpp
// A pass that folds an all-reduce feeding into another all-reduce by expanding
// the replica groups. 
//
// Example:
//
//   ar0 = all-reduce(x) replica_groups={{0,1},{2,3},{4,5},{6,7}}
//   ar1 = all-reduce(all-reduce0) replica_groups={{0,2},{1,3},{4,6},{5,7}}
//
//  Can be combined into a single all-reduce:
//
//   ar1 = all-reduce(x) replica_groups={{0,1,2,3},{4,5,6,7}}
//
class AllReduceFolder : public HloModulePass {
 public:
  absl::string_view name() const override { return "all-reduce-folder"; }

  using HloPassInterface::Run;
  absl::StatusOr<bool> Run(
      HloModule* module,
      const absl::flat_hash_set<absl::string_view>& execution_threads) override;
};
```

Resulting markdown block for the Class above will be:
```md
### all-reduce-folder

See also
[`xla::AllReduceFolder`](https://github.com/openxla/xla/tree/main/xla/service/all_reduce_folder.h).

A pass that folds an all-reduce feeding into another all-reduce by expanding the replica groups.

Example:

\```cpp
ar0 = all-reduce(x) replica_groups={{0,1},{2,3},{4,5},{6,7}}
ar1 = all-reduce(all-reduce0) replica_groups={{0,2},{1,3},{4,6},{5,7}}
\```

Can be combined into a single all-reduce:

\```cpp
ar1 = all-reduce(x) replica_groups={{0,1,2,3},{4,5,6,7}}
\```

```

dimitar-asenov:
@cheshire In my understanding the entire list is auto-generated, based on what's in the code. So there's no mixing between automatic and manual documentation.

@apivovarov Is it possible to group the different passes, e.g. similarly to the PR description?

cheshire:
> @cheshire In my understanding the entire list is auto-generated, based on what's in the code. So there's no mixing between automatic and manual documentation.

Yes, so in this case I don't think it should be in the documentation folder, as at least currently we have no automatic job to keep it up-to-date.

dimitar-asenov:
@cheshire: would simply putting in somewhere near the code make you happy? That avoids the mixture you mentioned though it doesn't solve the problem of automatic generation. We might be able to enforce this with presubmit checks + manually running a script whenever a new pass is added.

penpornk:
> @cheshire: would simply putting in somewhere near the code make you happy? That avoids the mixture you mentioned though it doesn't solve the problem of automatic generation. We might be able to enforce this with presubmit checks + manually running a script whenever a new pass is added.

@cheshire We need your answer to make progress with this PR.

cheshire:
Nope with no enforcement mechanism to keep the documentation current I'd still prefer not to have it in favor of extended header comments (and considering contributing an auto-generated documentation).

dimitar-asenov:
@apivovarov Thank you for proposing this improvement to the XLA docs. As per George's comment above, it seems that we don't want to merge this PR as is currently with the main issue being lack of enforcement that the docs are current. If you want to come up with a solution to this, please mark the PR as draft in the meantime, and mark it again as non-draft once it's ready for review. If you want to abandon this change you can just close the PR.

PR #11643, opened 2024-04-18
Fix unexpected resursive call in ksl for loop

ghost:
ksl does not have this func: ForWithStatus(string_view, llvm::Value*, llvm::Value*, llvm::Value, bool, const std::function<Status(llvm::Value*, bool)>&), only has ForWithStatus(string_view, llvm::Value*, llvm::Value*, llvm::Value, const std::function<Status(llvm::Value*, bool)>&) which is the expected func call. So if use peel first iteration, will call ForWithStatus(string_view, llvm::Value*, llvm::Value*, llvm::Value, bool, const std::function<Status(llvm::Value*, llvm::Value*)>&) recursively (std::function<Status(llvm::Value*, bool)> will cast to std::function<Status(llvm::Value*, llvm::Value*)> implicitly on my running enviroment), which leads stack smash.

### DIFF
--- a/xla/service/cpu/tiled_dot_emitter.cc
+++ b/xla/service/cpu/tiled_dot_emitter.cc
@@ -360,8 +360,9 @@ void ColumnMajorMatrixVectorProductEmitter::EmitInnerLoopEpilogue(
   ksl_.For(
       "dot.inner.epilg.outer", /*start=*/current_tile_col,
       /*end=*/b_->CreateAdd(columns_llvm, current_tile_col),
-      /*step=*/1, /*peel_first_iteration=*/false,
-      [&](llvm::Value* col, llvm::Value* is_first_scalar_col) {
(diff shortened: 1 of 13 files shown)


jvstokes (approved):

jvstokes (changes requested):
Actually, could you please provide a test case which highlights what was failing before (and is fixed now?)

akuegel:
It seems right now, nobody is passing peel_first_iteration=true (if they did, it would crash). So should we instead just delete the dead, broken code? In case you are planning to use this feature, I agree that having a test that calls it with peel_first_iteration=true would be great.

lingzhi98:
Okay, I'll remove the dead code. Although I intend to use this feature, it seems like removing it would be better if the use of peel_first_iteration does not exist. I can achieve the same effect in another way.

akuegel (approved):

PR #11644, opened 2024-04-18
[XLA:GPU] move cublasLt creation to thunk initialization.

shawnwang18:
To avoid cublasLt and command buffer deadlock, move cublasLt creation to thunk initialization.


jvstokes (approved):
Thank you!

PR #11649, opened 2024-04-18
[GPU] Rename and better describe flags related to GEMM fusions.

sergachev:
Rename one flag for consistency because fusions can use different backends.

Also clarify descriptions of other 2 related flags.


jvstokes (approved):
Thank you!

PR #11652, opened 2024-04-18
[ROCm] Remove AllocateOwnedArray from rocm code.

zoranjovanovic-ns:


zoranjovanovic-ns:
Will create new one in order to triger rocm CI.

PR #11753, opened 2024-04-22
Including all the gpu versions mentioned in the backends attribute.

hmonishN:
Added all the gpu version mentioned in the backends field in BUILD files. The tests marked with gpu_v100, gpu_a100, gpu_h100, etc were not getting picked up by bazel during execution. The fix will include the tests and run them on respective gpu nodes.

### DIFF
--- a/xla/tests/build_defs.bzl
+++ b/xla/tests/build_defs.bzl
@@ -111,7 +111,7 @@ def xla_test(
         if backend == "cpu":
             backend_deps = ["//xla/service:cpu_plugin"]
             backend_deps += ["//xla/tests:test_macros_cpu"]  # buildifier: disable=list-append
-        elif backend == "gpu":
+        elif backend == "gpu" or backend == "gpu_a100" or backend == "gpu_v100" or backend == "gpu_p100" or backend == "gpu_h100":
             backend_deps = if_gpu_is_configured(["//xla/service:gpu_plugin"])
             backend_deps += if_gpu_is_configured(["//xla/tests:test_macros_gpu"])  # buildifier: disable=list-append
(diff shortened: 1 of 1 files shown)


hmonishN:
+ @ddunleavy@google.com

xla/tests/build_defs.bzl
```diff
@@ -111,7 +111,7 @@ def xla_test(
         if backend == "cpu":
             backend_deps = ["//xla/service:cpu_plugin"]
             backend_deps += ["//xla/tests:test_macros_cpu"]  # buildifier: disable=list-append
-        elif backend == "gpu":
+        elif backend == "gpu" or backend == "gpu_a100" or backend == "gpu_v100" or backend == "gpu_p100" or backend == "gpu_h100":
```
ddunl, line 114: I think it'd be nice to have a `gpu_backends` list at the top with `all_backends`, and then redefine like `all_backends = gpu_backends + ["cpu"] + plugins.keys()`

hmonishN: Incorporated review comments.

ddunl:
In parallel here I'm going to change the meaning of `tf_cuda_tests_tags` because I think this PR is exposing some inconsistencies in how we are using backends in `xla_test`, and also the associated tags for indicating different hardware. Another note - I think that we want to rename `gpu` -> `gpu_any` in this scheme to indicate when we don't care about the specific hardware used, just that we want a GPU. Sorry for the back and forth here!

ddunl:
Closing now that https://github.com/openxla/xla/commit/c0f10b31c05590250d6a02c4911d84d3937bf904 is in. Thanks for this PR!!

PR #11757, opened 2024-04-22
[GPU] Enable workspace support in cuDNN GEMM fusions.

sergachev:


xla/hlo/utils/hlo_query.h
```diff
@@ -125,6 +125,9 @@ int64_t NextChannelId(const HloModule& module);
 // rewritten into tuple shaped transfers.
 bool HasX64TransformedHostTransfer(const HloModule& module);
 
+// Detects `ROOT tuple(..., custom-call())` used to allocate workspace buffers.
+bool IsWorkspaceAllocationRoot(const HloInstruction& root);
```
cheshire, line 129: This is a necessary condition, is it sufficient though? Looks like a heuristic, in which case a more localized helper function could make more sense?

cheshire (changes requested):

xla/hlo/utils/hlo_query.h
sergachev: Added custom call target check and moved the function to another unit.

cheshire (approved):

PR #11760, opened 2024-04-22
[gpu] Change the implementation of GetFusionRoots to a simpler one

shraiysh:
Following the tuple -> get-tuple-element sequence for getting roots and remove allowing a duplicate root in case it is encountered twice in a row in the recursive search.

### DIFF
--- a/xla/service/gpu/gpu_fusible.cc
+++ b/xla/service/gpu/gpu_fusible.cc
@@ -880,16 +880,17 @@ size_t GetOutputSizeOfFusible(const HloInstruction& instr) {
 // Recursive helper for GetFusionRoots below.
 static void GetFusionRootsRec(const HloInstruction* root,
                               std::vector<const HloInstruction*>& out) {
-  if (root->opcode() == HloOpcode::kGetTupleElement) {
-    return GetFusionRootsRec(root->operand(0), out);
+  if (root->opcode() == HloOpcode::kGetTupleElement &&
+      root->operand(0)->opcode() == HloOpcode::kTuple) {
+    return GetFusionRootsRec(root->operand(0)->operand(root->tuple_index()),
+                             out);
+  } else if (root->opcode() == HloOpcode::kGetTupleElement) {
+    out.push_back(root->operand(0));
   } else if (root->opcode() == HloOpcode::kTuple) {
     for (int i = 0; i < root->operand_count(); i++) {
       GetFusionRootsRec(root->operand(i), out);
     }
   } else {
-    if (!out.empty() && out.back() == root) {
-      return;
-    }
     CHECK(!absl::c_linear_search(out, root))
         << "Fusion root contains instruction " << root->ToString()
         << " multiple times";
--- a/xla/service/gpu/gpu_fusible_test.cc
+++ b/xla/service/gpu/gpu_fusible_test.cc
@@ -1470,5 +1470,173 @@ ENTRY computation {
             HloInstruction::FusionKind::kInput);
 }
 
+TEST_F(GpuFusibleTest, GetFusionRoots1) {
+  auto module = ParseAndReturnVerifiedModule(R"(
+    HloModule test_module
+
+    fusion {
+      p0 = s32[] parameter(0)
+      custom-call = (bf16[], s32[]) custom-call(p0), custom_call_target="my_custom_call"
+      get-tuple-element.0 = bf16[] get-tuple-element(custom-call), index=0
+      get-tuple-element.1 = s32[] get-tuple-element(custom-call), index=1
+      ROOT tuple = (bf16[], s32[], s32[]) tuple(get-tuple-element.0, get-tuple-element.1, p0)
+    }
+
+    ENTRY entry{
+      p0 = s32[] parameter(0)
+      ROOT fusion = (bf16[], s32[], s32[]) fusion(p0), kind=kCustom, calls=fusion
+    }
+  )")
+                    .value();
+  auto called_computations =
+      module->entry_computation()->root_instruction()->called_computations();
+  ASSERT_EQ(called_computations.size(), 1);
+  auto fusion = called_computations.front();
+  auto roots = GetFusionRoots(*fusion);
+  auto custom_call = fusion->root_instruction()->operand(0)->operand(0);
+  auto parameter = fusion->root_instruction()->operand(2);
+  std::vector<const HloInstruction*> expected_roots{custom_call, custom_call,
+                                                    parameter};
+  EXPECT_EQ(roots, expected_roots);
+}
+
+TEST_F(GpuFusibleTest, GetFusionRoots2) {
+  auto module = ParseAndReturnVerifiedModule(R"(
+    HloModule test_module
+
+    fusion {
+      p0 = s32[] parameter(0)
+      custom-call.1 = bf16[] custom-call(p0), custom_call_target="my_custom_call1"
+      custom-call.2 = bf16[] custom-call(p0), custom_call_target="my_custom_call2"
+      ROOT tuple = (bf16[], bf16[], s32[]) tuple(custom-call.1, custom-call.2, p0)
+    }
+
+    ENTRY entry{
+      p0 = s32[] parameter(0)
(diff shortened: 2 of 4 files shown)


shraiysh:
This check causes the test added in multi_output_fusion_test.cc to fail.

xla/service/gpu/gpu_fusible.cc
```diff
@@ -885,12 +885,6 @@ static void GetFusionRootsRec(const HloInstruction* root,
       GetFusionRootsRec(root->operand(i), out);
     }
   } else {
-    if (!out.empty() && out.back() == root) {
-      return;
-    }
-    CHECK(!absl::c_linear_search(out, root))
```
akuegel, line 891: I think the check still makes sense, but in line 881 we assume that if we have a GetTupleElement op, it has as operand a Tuple op (which is not true in case of DynamicAddressFusion). So I think a better fix might be to change line 881 to ```if (root->opcode() == HloOpcode::kGetTupleElement && root->operand(0)->opcode() == HloOpcode::kTuple)```

See also https://github.com/openxla/xla/commit/2204f157da667491c1baa4f2f37a883cfc73068e for a similar recent fix.

shraiysh: Does this look okay? I removed the check from line 888 because it makes sense to not allow returning same values, but why we should be allowing them in adjacent elements in tuple is not clear to me.

akuegel: Yes, looks ok to me. I also don't understand why we try to allow a duplicate root in case it is encountered twice in a row in the recursive search. Seems arbitrary to me, and we should also check fail in that case.

akuegel (approved):
Thank you :)

xla/service/gpu/gpu_fusible.cc
```diff
@@ -878,16 +878,16 @@ size_t GetOutputSizeOfFusible(const HloInstruction& instr) {
 // Recursive helper for GetFusionRoots below.
 static void GetFusionRootsRec(const HloInstruction* root,
                               std::vector<const HloInstruction*>& out) {
-  if (root->opcode() == HloOpcode::kGetTupleElement) {
+  if (root->opcode() == HloOpcode::kGetTupleElement &&
+      root->operand(0)->opcode() == HloOpcode::kTuple) {
```
akuegel, line 882: Your test case GpuFusibleTest.GetFusionRootsWithMakeTupleGTESequence shows that we will need an additional change to make this work correctly. We should only access the part of the tuple that the get-tuple-element refers to. So:

```
return GetFusionRootsRec(root->operand(0)->operand(root->tuple_index()), out);
```

xla/tests/build_defs.bzl
```diff
@@ -111,7 +111,7 @@ def xla_test(
         if backend == "cpu":
             backend_deps = ["//xla/service:cpu_plugin"]
             backend_deps += ["//xla/tests:test_macros_cpu"]  # buildifier: disable=list-append
-        elif backend == "gpu":
+        elif backend == "gpu" or backend == "gpu_v100":
```
akuegel, line 114: It looks to me that this belongs into a separate PR (I think I have even already seen it floating around)

xla/service/gpu/multi_output_fusion_test.cc
```diff
@@ -2175,5 +2175,32 @@ ENTRY computation {
...
+    fusion {
+      p0 = s32[] parameter(0)
+      p1 = s32[32] parameter(1)
+      custom-call = (bf16[], s32[], u32[]) custom-call(p1), custom_call_target="my_custom_call"
+      get-tuple-element.0 = bf16[] get-tuple-element(custom-call), index=0
+      get-tuple-element.1 = s32[] get-tuple-element(custom-call), index=1
+      bitcast = s32[1] bitcast(get-tuple-element.1)
+      dynamic-update-slice = s32[32] dynamic-update-slice(p1, bitcast, p0)
+      get-tuple-element.2 = u32[] get-tuple-element(custom-call), index=2
+      ROOT tuple.30 = (bf16[], s32[32], u32[]) tuple(get-tuple-element.0 ,dynamic-update-slice, get-tuple-element.2)
```
akuegel, line 2191: nit: put the comma directly after get-tuple-element.0

xla/service/gpu/gpu_fusible_test.cc
```diff
@@ -1470,5 +1470,173 @@ ENTRY computation {
...
+      custom-call = (bf16[], s32[], u32[]) custom-call(p1), custom_call_target="my_custom_call"
+      get-tuple-element.0 = bf16[] get-tuple-element(custom-call), index=0
+      get-tuple-element.1 = s32[] get-tuple-element(custom-call), index=1
+      bitcast = s32[1] bitcast(get-tuple-element.1)
+      dynamic-update-slice = s32[32] dynamic-update-slice(p1, bitcast, p0)
+      get-tuple-element.2 = u32[] get-tuple-element(custom-call), index=2
+      tuple = (bf16[], s32[32], u32[]) tuple(get-tuple-element.0 ,dynamic-update-slice, get-tuple-element.2)
+      get-tuple-element.3 = bf16[] get-tuple-element(tuple), index=0
+      get-tuple-element.4 = u32[] get-tuple-element(tuple), index=2
+      ROOT tuple2 = (bf16[], s32[32], u32[]) tuple(get-tuple-element.3 ,dynamic-update-slice, get-tuple-element.4)
```
akuegel, line 1617: nit: put the comma directly after get-tuple-element.3

xla/service/gpu/gpu_fusible_test.cc
```diff
@@ -1470,5 +1470,173 @@ ENTRY computation {
...
+    fusion {
+      p0 = s32[] parameter(0)
+      p1 = s32[32] parameter(1)
+      custom-call = (bf16[], s32[], u32[]) custom-call(p1), custom_call_target="my_custom_call"
+      get-tuple-element.0 = bf16[] get-tuple-element(custom-call), index=0
+      get-tuple-element.1 = s32[] get-tuple-element(custom-call), index=1
+      bitcast = s32[1] bitcast(get-tuple-element.1)
+      dynamic-update-slice = s32[32] dynamic-update-slice(p1, bitcast, p0)
+      get-tuple-element.2 = u32[] get-tuple-element(custom-call), index=2
+      tuple = (bf16[], s32[32], u32[]) tuple(get-tuple-element.0 ,dynamic-update-slice, get-tuple-element.2)
```
akuegel, line 1614: nit: put the comma directly after get-tuple-element.0

xla/service/gpu/gpu_fusible_test.cc
```diff
@@ -1470,5 +1470,173 @@ ENTRY computation {
...
+    fusion {
+      p0 = s32[] parameter(0)
+      p1 = s32[32] parameter(1)
+      custom-call = (bf16[], s32[], u32[]) custom-call(p1), custom_call_target="my_custom_call"
+      get-tuple-element.0 = bf16[] get-tuple-element(custom-call), index=0
+      get-tuple-element.1 = s32[] get-tuple-element(custom-call), index=1
+      bitcast = s32[1] bitcast(get-tuple-element.1)
+      dynamic-update-slice = s32[32] dynamic-update-slice(p1, bitcast, p0)
+      get-tuple-element.2 = u32[] get-tuple-element(custom-call), index=2
+      ROOT tuple = (bf16[], s32[32], u32[]) tuple(get-tuple-element.0 ,dynamic-update-slice, get-tuple-element.2)
```
akuegel, line 1578: nit: put the comma directly after get-tuple-element.0

akuegel (changes requested):
Please also update the PR description, I think it still refers to the change that the CHECK was removed.

shraiysh:
I don't understand the things done by copybara here @akuegel. Was this merged in the previous state because of the approval, without waiting for nits to be addressed?

akuegel:
> I don't understand the things done by copybara here @akuegel. Was this merged in the previous state because of the approval, without waiting for nits to be addressed?

No, these were just presubmits which also create a PR. You can ignore this.

akuegel (approved):

PR #11761, opened 2024-04-22
[GPU] Support nccl comm splitting in multiprocess mode

trevor-m:
Currently `--xla_gpu_enable_nccl_comm_splitting` is a no-op unless using single process mode. This PR allows it to work in multiprocess mode by removing the IsLocal check and fixing the key in the rank -> comm map which was causing the following error:
```
7:  Communicator for rank 1 not found in a NCCL clique devices=[3,7]; stream=0
5: E0403 15:37:24.104311 3871038 pjrt_stream_executor_client.cc:2809] Execution of replica 0 failed: INTERNAL: Communicator for rank 1 not found in a NCCL clique devices=[1,5]; stream=0
6: E0403 15:37:24.104477 3870182 pjrt_stream_executor_client.cc:2809] Execution of replica 0 failed: INTERNAL: Communicator for rank 1 not found in a NCCL clique devices=[2,6]; stream=0
4: E0403 15:37:24.105872 3871021 pjrt_stream_executor_client.cc:2809] Execution of replica 0 failed: INTERNAL: Communicator for rank 1 not found in a NCCL clique devices=[0,4]; stream=0
```

PR #11762, opened 2024-04-22
[GPU] Enable nccl comm split by default

trevor-m:
With https://github.com/openxla/xla/pull/11761 NCCL comm split will be available for multiprocess applications. Using that change, I tested a variety of PAXML and T5X workloads. I found no crashes, hangs, or performance regressions and a good amount of memory savings depending on the model. I tested many more workloads aside from these four, but there was no interesting results since they only used a single communicator.

| Workload                                                                 | Mem diff (MiB) | Speedup |
| ------------------------------------------------------------------------ | ------------- | --------- |
| paxml llama70b evaluate 16xh100 bfloat16 bs-4 ici--1-8-1 dcn--1-2-1      | 0 | 0.9915 |
| paxml GPT3 train 8xh100 bfl16 bs-4 ici--4-1-2                            | \-320 | 1.0029 |
| paxml GPT3 train 8xh100 bfl16 bs-8 ici--1-4-2                            | \-1252 |  1.0026 |
| paxml GPT3pp train 16xh100 bf16 bs-8 ici--2-2-1-2                        | \-2990 | 0.9947 |

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -147,7 +147,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_nccl_termination_timeout_seconds(-1);
   opts.set_xla_gpu_enable_shared_constants(true);
   opts.set_xla_gpu_enable_nccl_user_buffers(false);
-  opts.set_xla_gpu_enable_nccl_comm_splitting(false);
+  opts.set_xla_gpu_enable_nccl_comm_splitting(true);
   opts.set_xla_gpu_enable_nccl_per_stream_comms(false);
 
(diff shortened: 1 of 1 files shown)


PatriosTheGreat (approved):
I checked our multi-host benchmarks and they seems to be good with this change

hawkinsp:
We're fairly confident this change introduced a hang in the JAX CI test suite and may have to revert it.

Repro, using JAX and the JAX tests:
```
XLA_PYTHON_CLIENT_ALLOCATOR=platform pytest -n 8 tests/pmap_test.py
```
which I ran on a VM with 4xT4 GPUs.

trevor-m:
@hawkinsp Thanks for letting me know about the hang. I'll try to investigate it.

PR #11849, opened 2024-04-24
Dot mixed type verifier

Bode2222:


xla/service/hlo_verifier.cc
```diff
@@ -214,6 +214,13 @@ Status ShapeVerifier::HandleDot(HloInstruction* dot) {
           dot->operand(0)->shape(), dot->operand(1)->shape(),
           dot->dot_dimension_numbers(),
           /*preferred_element_type=*/dot->shape().element_type(), sparsity));
+  if (!HasCompatibleElementTypes(dot->operand(0)->shape(),
```
akuegel, line 217: allow_mixed_precision is about floating point mixed types. But dot allows also other mixed types (e.g. u8 inputs and u32 output), so adding this check here would mean they would suddenly be considered invalid (but codegen works fine for them).

kamaljeeti:
Hi @Bode2222 , can you have a look at the comments? Thanks.

xla/service/hlo_verifier.cc
Bode2222: Will fix. Expect new review soon

Bode2222:
On it

xla/service/hlo_verifier.cc
akuegel: Still a lot of tests are failing. Note that allow_mixed_precision in the verifier is set to false by default, and tests usually don't override the default. So all those tests would have to be changed to set allow_mixed_precision to true. Also I don't really see the point if we then set allow_mixed_precision to true everywhere. Then this new check in the verifier will just be dead code?

Bode2222: You're right, it would be a lot of work to fix all the tests that rely on the fact that the verifier is broken. However, having the verifier work as intended is important for the customers and stakeholders who use xla. Possible solutions include setting `allow_mixed_precision` to true for the scope of the tests,  setting `allow_mixed_precision` to true globally or in the worst case scenario fixing the errant tests by hand. Could you please let me know which of these is the most/least desirable to you?

akuegel: I would guess setting allow_mixed_precision to true globally might be the best option. But then, the new check in the verifier will never trigger, or do I miss something? So I am not exactly sure why you want this check in the verifier. Essentially, for dot we always allow mixed precision, so why do we need code in the verifier to trigger if we don't allow mixed precision? Do you have a XLA backend where you do not want to allow mixed precision? In that case, the default verifier option should probably be derived from some option that the backend can configure.

Bode2222: I do have an XLA backend where I do not want to allow mixed precision. Having the backend set the default verifier option sounds like the best option. Lets do that.

kamaljeeti:
Hi @akuegel , can you look into this? Thanks.

akuegel:
> Hi @akuegel , can you look into this? Thanks.

As far as I understand, this still requires changes to the HloVerifier default, otherwise a lot of tests will be broken. I suggested to derive this default setting from a backend specific setting, but I must admit I am not 100% sure what the best way to do it is. But in any case, the PR like it is will not work, as it will make a lot of tests fail.

penpornk:
Since it has been over a month without any activity. Shall we close this PR for now and reopen it when it's ready for a re-review?

dimitar-asenov:
@Bode2222 This PR has been closed due to prolonged inactivity. If you still intend to merge this PR, please reopen it and provide an update.

PR #11895, opened 2024-04-25
Offloading 3/3: enable rematerialization using xla flags

zhenying-liu:
This enables activation offloading automatically on GPU through HLO Rematerialization.
There is one xla flag needed to trigger rematerialization offloading:
1. A bool to enable host memory offloading. It's false by default to disable the offloading.

The other information is achieved by querying the GPU device:
2. Memory bandwidth in bytes per second.
3. The number of floating point operations on a device.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -251,6 +251,8 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
 
   opts.set_xla_gpu_require_complete_aot_autotune_results(false);
 
+  opts.set_xla_gpu_enable_host_memory_offloading(false);
+
   return opts;
 }
 
@@ -1670,6 +1672,11 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
           &DebugOptions::set_xla_reduce_window_rewrite_base_length),
       debug_options->xla_reduce_window_rewrite_base_length(),
       "Base length to rewrite the reduce window to, no rewrite if set to 0."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_enable_host_memory_offloading",
+      bool_setter_for(&DebugOptions::set_xla_gpu_enable_host_memory_offloading),
+      debug_options->xla_gpu_enable_host_memory_offloading(),
+      "Whether to trigger host memory offloading on a device."));
 }  // NOLINT(readability/fn_size)
 
 // Allocates flag_values and flag_objects; this function must not be called more
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -2175,16 +2175,41 @@ absl::Status GpuCompiler::RunPostSchedulingPipelines(
   {
     HloPassPipeline pipeline("remat-pipeline");
 
-    HloCostAnalysis hlo_cost_analysis(ShapeSizeBytesFunction());
+    const bool enable_offloading = module->config()
+                                       .debug_options()
+                                       .xla_gpu_enable_host_memory_offloading();
+    float flops_per_sec = gpu_device_info.core_count() *
+                          gpu_device_info.fpus_per_core() *
+                          gpu_device_info.clock_rate_ghz() * /*giga*/1e+9 * 2;
+    int64_t host_memory_space_color =
+        static_cast<int64_t>(se::MemoryType::kHost);
+
     HloRematerialization::RematerializationModeConfig
         rematerialization_mode_config(/*recompute=*/true, /*compress=*/true,
-                                      /*host_offload=*/false);
+                                      /*host_offload=*/enable_offloading);
+    HloCostAnalysis::Options hlo_cost_analysis_options;
+    hlo_cost_analysis_options.shape_size = ShapeSizeBytesFunction();
+    std::optional<HloRematerialization::HostMemoryOffloadConfig>
+        offloading_config = std::nullopt;
+    if (enable_offloading) {
+      hlo_cost_analysis_options.set_flops_per_second(flops_per_sec);
+      hlo_cost_analysis_options.set_transcendentals_per_second(flops_per_sec);
+      offloading_config =
+          std::make_optional<HloRematerialization::HostMemoryOffloadConfig>(
+              /*host_memory_space=*/host_memory_space_color,
+              /*bandwidth_to_host_bytes_per_second=*/
+              gpu_device_info.memory_bandwidth(),
+              /*bandwidth_from_host_bytes_per_second=*/
+              gpu_device_info.memory_bandwidth());
+    }
+    HloCostAnalysis hlo_cost_analysis(hlo_cost_analysis_options);
     HloRematerialization::Options options(
         hlo_cost_analysis, rematerialization_mode_config,
         // Assume 75% of the total device memory is available for XLA.
         /*memory_limit_bytes=*/scheduler_mem_limit,
         /*block_size_limit=*/1, /*block_rematerialization_factor=*/1,
-        /*min_remat_size=*/0, /*compact_shape_function=*/nullptr);
+        /*min_remat_size=*/0, /*compact_shape_function=*/nullptr,
+        /*host_memory_offload_config=*/offloading_config);
     HloRematerialization::RematerializationSizes sizes;
     pipeline.AddPass<HloRematerialization>(options, sizes);
     pipeline.AddPass<StreamAttributeAnnotator>();
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -768,7 +768,10 @@ message DebugOptions {
   // Base length to rewrite the reduce window to, no rewrite if set to 0.
   int64 xla_reduce_window_rewrite_base_length = 293;
 
-  // Next id: 296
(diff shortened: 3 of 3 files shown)


zhenying-liu:
The memory bandwidth and flops numbers of Grace Hopper are taken from this doc: https://resources.nvidia.com/en-us-grace-cpu/grace-hopper-superchip. It disables HLO rematerialization offloading for GPUs by default, and can enable host memory offloading via those three XLA flags. Since hlo_rematerialization_test.cc already has the unit tests for enabling offloading, we didn't provide the unit test here.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2194,26 +2194,69 @@ absl::Status GpuCompiler::RunPostSchedulingPipelines(
...
+      }
+    } else {
+      // Enable host memory offloading in Rematerialization pass
+      HloCostAnalysis::Options hlo_cost_analysis_options;
+      hlo_cost_analysis_options.shape_size = ShapeSizeBytesFunction();
+      hlo_cost_analysis_options.set_flops_per_second(flops_per_sec);
+      hlo_cost_analysis_options.set_transcendentals_per_second(flops_per_sec);
+      HloCostAnalysis hlo_cost_analysis(hlo_cost_analysis_options);
+      HloRematerialization::RematerializationModeConfig
+          rematerialization_mode_config(/*recompute=*/false, /*compress=*/false,
```
SandSnip3r, line 2238: Why do you disable other forms of rematerialization when the host offloading strategy is enabled?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2194,26 +2194,69 @@ absl::Status GpuCompiler::RunPostSchedulingPipelines(
...
+    float offloading_bandwidth =
+        module->config()
+            .debug_options()
+            .xla_gpu_offloading_bandwidth_bytes_per_sec();
+    float flops_per_sec =
+        module->config().debug_options().xla_gpu_offloading_flops_per_sec();
+    int64_t host_memory_space_color =
+        static_cast<int64_t>(se::MemoryType::kHost);
+
+    if (offloading_memory_limit_gb == 0) {
```
SandSnip3r, line 2209: There is a lot of duplicate code here. You should have one bit of code which builds the HloRematerialization::Options (based on the flags, etc.), then have just a single place where the passes are added to the pipeline.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2194,26 +2194,69 @@ absl::Status GpuCompiler::RunPostSchedulingPipelines(
...
-    if (changed) {
-      VLOG(1) << "HloRematerialization saved "
-              << sizes.before_bytes - sizes.after_bytes << " bytes";
+    int64_t offloading_memory_limit_gb =
+        module->config().debug_options().xla_gpu_offloading_memory_limit_gb() *
+        1024 * 1024 * 1024;
+    float offloading_bandwidth =
+        module->config()
+            .debug_options()
+            .xla_gpu_offloading_bandwidth_bytes_per_sec();
```
SandSnip3r, line 2203: Does it really make sense to take this from a flag? Wouldn't this be queryable based on the current GPU?

SandSnip3r (changes requested):

xla/service/gpu/gpu_compiler.cc
zhenying-liu: Sure, recompute/compress remain true whether host offloading is on.

xla/service/gpu/gpu_compiler.cc
zhenying-liu: Thanks for pointing this out. It's changed accordingly.

xla/service/gpu/gpu_compiler.cc
zhenying-liu: Right. Removed this flag by querying the current GPU.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2189,66 +2189,44 @@ absl::Status GpuCompiler::RunPostSchedulingPipelines(
     int64_t offloading_memory_limit_gb =
         module->config().debug_options().xla_gpu_offloading_memory_limit_gb() *
         1024 * 1024 * 1024;
-    float offloading_bandwidth =
-        module->config()
-            .debug_options()
-            .xla_gpu_offloading_bandwidth_bytes_per_sec();
```
nouiz: Can you also infer flops_per_sec ?
gpu_device_info has fpus_per_core and other function like clock_rate_ghz, that maybe allow you to infer it.

zhenying-liu: Let me figure it out.

zhenying-liu: Modified and verified with a V100 locally.

xla/debug_options_flags.cc
```diff
@@ -1681,6 +1682,13 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       bool_setter_for(&DebugOptions::set_xla_gpu_use_memcpy_local_p2p),
       debug_options->xla_gpu_use_memcpy_local_p2p(),
       "Whether to use memcpy for local p2p communication."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_offloading_memory_limit_gb",
+      int64_setter_for(&DebugOptions::set_xla_gpu_offloading_memory_limit_gb),
+      debug_options->xla_gpu_offloading_memory_limit_gb(),
+      "GPU memory usage upper limit in GB to trigger host memory offloading."
```
nouiz, line 1689: The usage is strange.
Right now, there is already a memory setting for that pass via scheduler_mem_limit.
There is already some flags to adjust it.
What this new version will provide?
Why not make this a bool that default to false? This way, user don't need to find the default GPU memory threshold that is set to use it. So it  would be easier to use.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2210,22 +2210,44 @@ absl::Status GpuCompiler::RunPostSchedulingPipelines(
...
-                                      /*host_offload=*/false);
+                                      /*host_offload=*/enable_offloading);
+    HloCostAnalysis::Options hlo_cost_analysis_options;
+    hlo_cost_analysis_options.shape_size = ShapeSizeBytesFunction();
+    if (enable_offloading) {
+      hlo_cost_analysis_options.set_flops_per_second(flops_per_sec);
+      hlo_cost_analysis_options.set_transcendentals_per_second(flops_per_sec);
+    }
+    HloCostAnalysis hlo_cost_analysis(hlo_cost_analysis_options);
+    HloRematerialization::HostMemoryOffloadConfig offloading_config(
```
SandSnip3r, line 2233: If offloading is not enabled, you should not be constructing a host memory offloading config nor passing it to the HloRematerialization options.

I think it is completely ok if you do, but it's an absolute waste and maybe confusing to someone else.

xla/debug_options_flags.cc
zhenying-liu: OK, I changed it to bool. I thought this memory limit could be smaller than scheduler_mem_limit in order to kick in offloading earlier. I agree it's better to have one flag to control the memory limit.

xla/service/gpu/gpu_compiler.cc
zhenying-liu: > If offloading is not enabled, you should not be constructing a host memory offloading config nor passing it to the HloRematerialization options.
How to pass a std::nullopt and offloading_config in an implicit if statement to construct a remat pass? I couldn't solve the type mismatch. It'll be nice to do it.

xla/debug_options_flags.cc
zhenying-liu: I have to change it back to int64_t. The activation offloading memory limit is much more smaller than scheduler_mem_limit. The former could be just few tensors for offloading candidates in a module, while the latter is the whole GPU memory used for JAX. If we use the scheduler_mem_limit for offloading, we can not enable offloading.

xla/service/gpu/gpu_compiler.cc
SandSnip3r:
```
    HloCostAnalysis::Options hlo_cost_analysis_options;
    hlo_cost_analysis_options.shape_size = ShapeSizeBytesFunction();
    std::optional<HloRematerialization::HostMemoryOffloadConfig> offloading_config;
    if (enable_offloading) {
      hlo_cost_analysis_options.set_flops_per_second(flops_per_sec);
      hlo_cost_analysis_options.set_transcendentals_per_second(flops_per_sec);
      offloading_config.emplace(/*host_memory_space=*/host_memory_space_color,
                                /*bandwidth_to_host_bytes_per_second=*/
                                gpu_device_info.memory_bandwidth(),
                                /*bandwidth_from_host_bytes_per_second=*/
                                gpu_device_info.memory_bandwidth());
    }
    HloCostAnalysis hlo_cost_analysis(hlo_cost_analysis_options);
    ```
    Then pass offloading_config unconditionally to the options struct

zhenying-liu: Thanks so much for showing me how to write such a code. I made a similar version and verified.

SandSnip3r (approved):

PR #11908, opened 2024-04-26
[GPU][NFC] Annotate cuDNN graphs with operation names from HLO.

sergachev:


cheshire (approved):
Is there a way to serialize the cudnn you generate and test it somehow? If not I guess it's fine.

sergachev:
> Is there a way to serialize the cudnn you generate and test it somehow? If not I guess it's fine.

Not yet but will become available soon.

PR #11912, opened 2024-04-26
[GPU] Support pointwise floor and ceil in GEMM fusions.

sergachev:
For this to work with Triton openxla/triton will need to include https://github.com/openai/triton/commit/62706e8c518c8c56e56460a43732d8e375217860.

### DIFF
--- a/xla/service/gpu/cudnn_fusion_compiler.cc
+++ b/xla/service/gpu/cudnn_fusion_compiler.cc
@@ -71,6 +71,8 @@ inline std::optional<fe::PointwiseMode_t> GetElementwiseMode(
       return m::ABS;
     case HloOpcode::kAdd:
       return m::ADD;
+    case HloOpcode::kCeil:
+      return m::CEIL;
     case HloOpcode::kCompare:
       switch (instruction.comparison_direction()) {
         case Comparison::Direction::kEq:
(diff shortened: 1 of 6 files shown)


Moerafaat:
The change looks good, though the needed Triton change would need to be included as you mentioned. I will cherry-pick your change internally so that it reflects on OpenXLA/Triton and then you can merge this change.

Moerafaat:
@sergachev The required dependency change now applies on top of OpenXLA/xla so this should be unblocked.

Moerafaat (approved):

sergachev:
//xla/service/gpu:softmax_rewriter_triton_test fails, I'm looking at that

dimitar-asenov:
Thanks @sergachev. Please ping me when you fix the issue and I will merge the PR internally.

sergachev:
@dimitar-asenov it's fixed now

dimitar-asenov:
@sergachev Actually, one more thing: Could you please add the new ops to this list:

https://github.com/openxla/xla/blob/main/xla/service/gpu/triton_support_test.cc#L193-L201

sergachev:
OK, I added them, but could not test - this test does not build in OSS because of the known problem with bazel backend tags; if I hack them it fails on unknown EXPECT_OK and so on.

PR #11967, opened 2024-04-30
[XLA:GPU] Add cuDNN flash attention dbias computation and broadcast bias

Cjkkkk:
* add cuDNN flash attention dbias computation. make sure dbias only pattern match cases where reduction is done on batch dim. Only bias with [1, N, S, T] is supported.
* add cuDNN broadcast bias support. cuDNN supports broadcast bias, [1, N, S, T], [1, 1, S, T] and [B, 1, S, T] are supported.

### DIFF
--- a/xla/service/gpu/cudnn_fused_mha_rewriter_test.cc
+++ b/xla/service/gpu/cudnn_fused_mha_rewriter_test.cc
@@ -3155,13 +3155,6 @@ TEST_F(CudnnFusedMhaRewriterTestHloTest, BF16Bmm1BiasSoftmaxBmm2PatternDbias) {
                                          GetCudnnVersion()};
   TF_ASSERT_OK(RunHloPass(&fusedMhaRewriter, m.get()).status());
 
-  AlgebraicSimplifierOptions alg_sim_options;
-  alg_sim_options.set_supports_non_canonical_dots(false);
-  alg_sim_options.set_is_layout_sensitive(true);
-  alg_sim_options.set_enable_conv_operand_swap(false);
-  AlgebraicSimplifier alge_simp{alg_sim_options};
-  TF_ASSERT_OK(RunHloPass(&alge_simp, m.get()).status());
-
   ComputationLayout computation_layout(
       m->entry_computation()->ComputeProgramShape());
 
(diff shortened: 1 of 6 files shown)


xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -5324,13 +5324,30 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionBackwardOperationGraph(
...
         graph.tensor(Tensor_attributes()
                          .set_name("bias")
                          .set_dim(bias_descriptor->dimensions())
                          .set_stride(bias_descriptor->GetLogicalStrides())
                          .set_uid(CudnnfMHAUid::BIAS_ID));
     sdpa_backward_options.set_bias(bias_tensor);
+
+    // shapes [1, 1, s, s], [b, 1, s, s], [b, h, s, s] are not supported for
+    // dbias calculation but they are supported for forward bias calculation
+    if ((b == 1) && (n == q_n)) {
```
akuegel, line 5342: nit: remove the extra () around b == 1 and n == qn checks

xla/service/gpu/cudnn_fused_mha_rewriter_test.cc
```diff
@@ -3070,6 +3070,142 @@ TEST_F(CudnnFusedMhaRewriterTestHloTest, HeadDimNotMultipleOf64) {
...
+  CudnnFusedMHARewriter fusedMhaRewriter{GetCudaComputeCapability(),
+                                         GetCudnnVersion()};
+  TF_ASSERT_OK(RunHloPass(&fusedMhaRewriter, m.get()).status());
+
+  AlgebraicSimplifierOptions alg_sim_options;
+  alg_sim_options.set_supports_non_canonical_dots(false);
+  alg_sim_options.set_is_layout_sensitive(true);
+  alg_sim_options.set_enable_conv_operand_swap(false);
+  AlgebraicSimplifier alge_simp{alg_sim_options};
+  TF_ASSERT_OK(RunHloPass(&alge_simp, m.get()).status());
```
akuegel, line 3163: Please store just the simplified HLO in hlo_BF16Bmm1BiasSoftmaxBmm2Pattern_dbias instead of running AlgebraicSimplifier here. Sorry that I didn't notice this earlier.

akuegel (changes requested):
Sorry for late reply, we had a public holiday yesterday.

xla/service/gpu/cudnn_fused_mha_rewriter_test.cc
Cjkkkk: Fixed. Thanks!

akuegel: I think you didn't update the HLO with the HLO how it looks like after running the simplifier. Instead you updated the expected matching HLO. I thought you could update the input HLO, and leave the expected matching HLO the same.

Cjkkkk: these transpose/reshape that should be simplified are introduced inside the fmha rewriter. Running simplifier for the input HLO would not change anything cuz the transpose/reshape will be added after that.

akuegel (approved):
Sorry for misreading the code. Makes sense now :)

PR #12018, opened 2024-05-01
Make affine map constraint check erased

ghost:
The [ComputeExpressionRange](https://github.com/openxla/xla/blob/aeaab3a1adbe5193956b118265a65ea1d2585a45/xla/service/gpu/model/indexing_map.cc#L829) is only correct iff sub exprs of affine expr are unrelated to each other or have the same monotonicity. For example, the expr s0 + s1 * 32 - (s0 floordiv 32) * 32 (0 <= s0 <=1023, 0 <= s1 <=31) will get [-992, 9183] value range but the correct range is [0, 8191]. The wrong value range cause simplify arith pass erase cmp op to fail so that llvm loop-related optimization can not apply, such as loop vectorization.
The above expr is not generated by openxla: the original expr is s0 mod 32 + s1 * 32. But mlir canonicalizer pass replace mod by floordiv (details can be seen in [here](https://github.com/llvm/llvm-project/blob/main/mlir/lib/IR/AffineExpr.cpp#L1286)). So the simplest solution is add simplify affine pass at the beginning of pass pipeline, so the affine map constraint check can be erased immediately. It is too hard to make ComputeExpressionRange suitable for any affine expr.

### DIFF
--- a/xla/service/gpu/fusions/concatenate_mlir_test.cc
+++ b/xla/service/gpu/fusions/concatenate_mlir_test.cc
@@ -118,7 +118,7 @@ TEST_F(MlirConcatenateFusionTest, StandAloneConcatenate) {
 
     // CHECK: %[[IF_2:.*]] = scf.if
     // CHECK:   %[[VAL_2:.*]] = xla_gpu.pure_call @fused_computation_param1
-    // CHECK:   %[[OUTPUT_INDEX_2:.*]] = arith.index_castui %[[AFFINE_APPLY_2:.*]] {xla.range = [200 : index, 599 : index]} : i32 to index
+    // CHECK:   %[[OUTPUT_INDEX_2:.*]] = arith.index_castui %[[AFFINE_APPLY_1:.*]] {xla.range = [200 : index, 599 : index]} : i32 to index
     // CHECK:   %[[INSERTED_2:.*]] = tensor.insert %[[VAL_2:.*]] into {{.*}}[%[[OUTPUT_INDEX_2]]]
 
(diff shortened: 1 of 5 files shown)


jreiffers:
Thanks for the PR. We're aware of this issue with the MLIR simplifier. pifon2a@ is currently working on a fix. Can you explain in a bit more detail why we would want to do this? SimplifyAffine is already run later, why does it need to run this early as well?

jreiffers:
Thanks, we're aware of this, but we're going with a different approach (replacing affine.apply with a custom op that uses our simplifier in folding and canonicalization).

lingzhi98:
Ok, close this PR and wait for openxla update.

jreiffers:
FYI https://github.com/openxla/xla/pull/12071 is about to land, which should resolve this. Please let us know if you see any issues!

lingzhi98:
#12071 is not related about my issue.  What make the column reduction can not vectorize is [this constraint check](https://github.com/openxla/xla/blob/main/xla/service/gpu/fusions/mlir/elemental_hlo_to_mlir.cc#L1306). And current canonicalization patterns for applyIndexing op don't include the computation of 'xla.range' so that simplify arith pass [fail to erase the constraint check](https://github.com/openxla/xla/pull/12071/files#diff-ff6784a483fa92aa338b31f99d1058ee5c21f2247f30010f6b3f007f18befb08R52). I am not sure whether it is the expected result, so I choose [another solution](https://github.com/openxla/xla/pull/11018/commits/25540062d5eec5447295282fa7a52528e6d0837f) and has verified column reduction can vectorize now.

PR #12049, opened 2024-05-02
Detect when we are running on an NVIDIA simulator and check that we pull the device description from gpu_specs.

dimvar:
This helps when testing the compiler on new hardware platforms. The simulator may have nonsensical values for a few fields in the device description (such as device_memory_size == -1), so we want to use a hard-coded device description that contains the correct values.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -1570,7 +1570,20 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
     return Compiler::TargetConfig{gpu_target_config_proto};
   }
   if (executor) {
-    return Compiler::TargetConfig{executor};
+    Compiler::TargetConfig target_config = Compiler::TargetConfig{executor};
+    int64_t device_memory_size =
+      target_config.device_description.device_memory_size();
+    // Checking for device_memory_size == -1 is how we detect that we are
+    // running on Nvidia's software simulator. When running on simulation,
+    // the config from StreamExecutor is inaccurate, so we must load the
+    // hard-coded config from a file.
+    if (device_memory_size == -1) {
+      return absl::FailedPreconditionError(
+        "When running on an NVIDIA simulation device, you must use "
+        "--xla_gpu_target_config_filename to pass in target information. "
+        "The target config from StreamExecutor is inaccurate.");
+    }
+    return target_config;
   }
   return absl::InternalError(
       "Either GPU has to be attached, or --xla_gpu_target_config_filename "


cheshire:
Could you describe that simulator a bit more? Can it execute compiled PTX? If not, can we just provide a device description instead of running on a simulator?

dimvar:
> Could you describe that simulator a bit more? Can it execute compiled PTX? 

Yes. It is a software simulator of a GPU architecture, analogous to Google's simulators for TPU.

> can we just provide a device description instead of running on a simulator?

I don't understand what you mean here. Can you rephrase the question?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1549,22 +1549,51 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
...
+    if (device_memory_size == -1) {
+      // Checking for device_memory_size == -1 is how we detect that we are running on
+      // Nvidia's software simulator. When running on simulation, discard the config
+      // from StreamExecutor, which can be inaccurate, and load the hard-coded config
+      // from a file.
+      auto cuda_cc = target_config.device_description.cuda_compute_capability();
+      switch (cuda_cc.major) {
+        case se::CudaComputeCapability::AMPERE: {
+          return target_config_from_filename(
+            "xla/tools/hlo_opt/gpu_specs/a100_40.txtpb");
```
cheshire, line 1584: I'm not sure paths work this way: have you tried it from any folder?

cheshire (changes requested):

kamaljeeti:
Hi @dimvar , can you have a look at the comments? Thanks.

xla/service/gpu/gpu_compiler.cc
dimvar: Added a comment to clarify. This is assuming that the compiler is invoked from the XLA root directory. If that is not the case, the user can always use --xla_gpu_target_config_filename, which causes the conditional before this one to be taken.

(You may ask, shouldn't we just always use --xla_gpu_target_config_filename then? We could, but it's nice when the compiler can figure out the right thing without the user having to pass flags.)

dimvar:
> Hi @dimvar , can you have a look at the comments? Thanks.

Just replied.

xla/service/gpu/gpu_compiler.cc
cheshire: > This is assuming that the compiler is invoked from the XLA root directory

I'm not sure what do you mean? The compiler can be invoked from any directory, and the behavior should not be changing based on that.

cheshire: >  We could, but it's nice when the compiler can figure out the right thing without the user having to pass flags

But then you get a highly-counterintuitive behavior when it differs depending on which folder is the binary called.

dimvar: Alright, let's just use the flag for the simulation case as well, and we can drop this PR.

dimvar: Actually, rather than closing, I'd like to change this PR to detect Nvidia simulation as before, and check-fail if --xla_gpu_target_config_filename is empty, to remind the user to supply the file with the target config. Are you OK w that?

dimvar: Updated the code.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1571,22 +1571,38 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
   if (options.target_config.has_value()) {
     return *options.target_config;
   }
-  if (!debug_opts.xla_gpu_target_config_filename().empty()) {
```
cheshire, line 1574: Not sure I understand the point of the diff here: defining the function and then calling it seems orthogonal to the diff below?

cheshire (changes requested):

xla/service/gpu/gpu_compiler.cc
dimvar: Yes, it's separate. I extracted the function for the previous version of the PR, and left it in because it seems a little cleaner to me. Do you prefer it without the function?

cheshire: yes, it's always better to minimize the diff.

dimvar: done

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1586,7 +1586,17 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
...
   if (executor) {
-    return Compiler::TargetConfig{executor};
+    Compiler::TargetConfig target_config = Compiler::TargetConfig{executor};
+    int64_t device_memory_size =
+      target_config.device_description.device_memory_size();
+    // Checking for device_memory_size == -1 is how we detect that we are
+    // running on Nvidia's software simulator. When running on simulation,
+    // the config from StreamExecutor is inaccurate, so we must load the
+    // hard-coded config from a file.
+    CHECK_NE(device_memory_size, -1)
```
cheshire, line 1596: I don't think we want to check-fail, maybe error out with a Status instead? Also maybe put most of the comment message into the returned string?

cheshire (changes requested):

xla/service/gpu/gpu_compiler.cc
dimvar: done

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1570,7 +1570,20 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
...
+    Compiler::TargetConfig target_config = Compiler::TargetConfig{executor};
+    int64_t device_memory_size =
+      target_config.device_description.device_memory_size();
+    // Checking for device_memory_size == -1 is how we detect that we are
+    // running on Nvidia's software simulator. When running on simulation,
+    // the config from StreamExecutor is inaccurate, so we must load the
+    // hard-coded config from a file.
+    if (device_memory_size == -1) {
+      return absl::FailedPreconditionError(
+        "When running on an NVIDIA simulation device, you must use "
```
cheshire, line 1582: Could you specify in the error message that we've implied it to be an NVIDIA simulator from memory size being minus one? 

Additionally, taking a step back, is there a more principled way to do this? Maybe some CUDA function which returns a known sentinel value? Maybe we should pipe that through instead?

cheshire (changes requested):

xla/service/gpu/gpu_compiler.cc
dimvar: > Could you specify in the error message that we've implied it to be an NVIDIA simulator from memory size being minus one?

I intentionally left that out because it's an obscure implementation detail that should not be relevant to the user.

> Additionally, taking a step back, is there a more principled way to do this? Maybe some CUDA function which returns a known sentinel value? Maybe we should pipe that through instead?

It could also be detected via the device name (available from the field device_description_str). That was my first implementation (before sending out the PR). But the name of the simulator is an internal Nvidia name. I asked inside Nvidia and I didn't get permission to open source the name.

cheshire (approved):

PR #12142, opened 2024-05-03
[XLA:GPU] Use cached cuDNN handle for FMHA thunk

AyanmoI:
ExecuteOnStream was calling into cuDNNCreate rather than calling the cached handle. This was causing a hang since handle was being created with a NCCL call in flight.
This fix addresses a hang.


nouiz:
Note, this fix a Hang. Would be great to have that clear in the description.

chsigg (approved):

dimitar-asenov:
@AyanmoI Could you please rebase on top of the latest XLA code? Thanks!

AyanmoI:
Closing this PR since it is not required anymore. Other overarching changes has made this redundant for now.

PR #12154, opened 2024-05-04
Scaling without Type Conversion for FP8 GEMMs

philipphack:
Enables the fusion of the scaling of the result of an FP8 GEMM without subsequent type conversion into the Custom Call.

### DIFF
--- a/xla/service/gpu/gemm_rewriter.cc
+++ b/xla/service/gpu/gemm_rewriter.cc
@@ -167,6 +167,27 @@ HloInstruction *PadOperandToMultipleOf16(absl::Span<const int64_t> batch_dims,
   return PadOperandToTargetShape(padded_shape, x);
 }
 
+// Calculates the reciprocal of scalar when invert is true and converts to FP32.
+absl::StatusOr<HloInstruction*> InvertAndConvertScalar(HloInstruction* scalar,
+                                                       bool invert) {
+  if (!ShapeUtil::IsScalar(scalar->shape())) {
+    return absl::OkStatus();
+  }
+
+  if (invert) {
+    Literal one_literal = LiteralUtil::One(scalar->shape().element_type());
+    HloInstruction* one = scalar->parent()->AddInstruction(
+        HloInstruction::CreateConstant(one_literal.Clone()));
+    TF_ASSIGN_OR_RETURN(scalar, MakeBinaryHlo(HloOpcode::kDivide, one, scalar,
+                                              &scalar->metadata()));
+  }
+  if (scalar->shape().element_type() != F32) {
+    scalar = MakeConvertToHlo(scalar, F32, &scalar->metadata());
+  }
+
+  return scalar;
+}
+
 // Recursively collects unary, divide, dynamic-slice, pad, multiply or select
 // operands of instr and the index of the operand identifying the next op in the
 // sequence until an instruction with FP8 element type is reached. Returns an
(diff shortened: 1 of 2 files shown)


philipphack:
CC @reedwm.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1164,6 +1182,53 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+      return OkStatus();
+    }
+
+    // If necessary, invert the scaling factor of D and convert to F32.
+    if (instr->opcode() == HloOpcode::kDivide) {
+      Literal one_literal = LiteralUtil::One(d_scale->shape().element_type());
+      HloInstruction* one = existing_gemm->parent()->AddInstruction(
+          HloInstruction::CreateConstant(one_literal.Clone()));
+      d_scale = existing_gemm->AddInstruction(HloInstruction::CreateBinary(
+          d_scale->shape(), HloOpcode::kDivide, one, d_scale));
```
thomasjoerg, line 1217: When creating HLO instructions, please copy the metadata from the original instruction. See line 761 for an example.
Same below.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1256,19 +1321,21 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
-      Literal one_literal = LiteralUtil::One(d_scale->shape().element_type());
-      HloInstruction *one = instr->AddInstruction(
-          HloInstruction::CreateConstant(one_literal.Clone()));
-      d_scale = instr->AddInstruction(HloInstruction::CreateBinary(
-          d_scale->shape(), HloOpcode::kDivide, one, d_scale));
-    }
-    if (d_scale->shape().element_type() != F32) {
-      d_scale = instr->AddInstruction(HloInstruction::CreateConvert(
-          ShapeUtil::MakeScalarShape(F32), d_scale));
+    if (d_scale) {
```
thomasjoerg, line 1324: The same code appears above. Extract a helper? This function is very long already, hence, that would also help readability.

thomasjoerg (changes requested):

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1164,6 +1182,53 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    }
+
+    TF_ASSIGN_OR_RETURN(auto gpu_backend_config,
+                        existing_gemm->backend_config<GpuBackendConfig>());
+    const GemmBackendConfig& gemm_backend_config =
+        gpu_backend_config.gemm_backend_config();
+
+    // When the result of an FP8 GEMM is scaled but not type converted to FP8,
+    // the scaling factor must be forwarded to the Custom Call as a_scale
+    // (chosen here) or b_scale. The scaling factor is fused here when no input
```
reedwm, line 1198: Mention the motivation for using a_sacle is that cublas LT doesn't supoprt having a d_scale when the output is not FP8.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1164,6 +1182,53 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+    // scaling factors were fused during the creation of the Custom Call. When
+    // the maximum of the absolute value of the result of an FP8 GEMM is
+    // calculated and the result is scaled and type converted to FP8, the
+    // scaling of the result is fused in F8ConvertD.
+    const int a_scale_operand_idx = gemm_backend_config.beta() == 0.0 ? 2 : 3;
```
reedwm, line 1203: If beta is nonzero, we cannot fuse d_scale at all. Since if the cublas call includes a matrix addition, scaling the output is not equivalent to scaling the input.

I believe fusing d_sclae as a_scale is also not valid when there are certain activations, like gelu (relu is fine I think).

Add a test for the case where the output is non-FP8, scaled, and there is a matrix addition.

reedwm (changes requested):

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -381,15 +381,19 @@ auto CublasLtMatmul(HloInstruction **instr) {
...
 
 auto CublasLtMatmulMaybeF8(HloInstruction **instr) {
   return m::CustomCall(
       instr, {kCublasLtMatmulCallTarget, kCublasLtMatmulF8CallTarget});
 }
 
+auto GemmOrCublasLtMatmul(HloInstruction **instr) {
+  return m::CustomCall(instr, {kGemmCallTarget, kCublasLtMatmulCallTarget});
+}
+
```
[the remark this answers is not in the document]
philipphack: The idea was to preserve the ordering of the m::CustomCall matches from less to more general. Let me know if you want to keep the previous order.

reedwm (changes requested):

reedwm (approved):

reedwm (approved):

reedwm (approved):

PR #12224, opened 2024-05-07
[GPU] Fix handling of flags in the cuDNN FMHA test.

sergachev:
The test got broken by https://github.com/openxla/xla/commit/8799ff0f4b2aa3437a9963f7c9968ee361a62fd7, this commit fixes it.

### DIFF
--- a/xla/service/gpu/tests/gpu_fused_mha_test.cc
+++ b/xla/service/gpu/tests/gpu_fused_mha_test.cc
@@ -24,8 +24,6 @@ limitations under the License.
 #include "absl/algorithm/container.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/match.h"
-#include "absl/strings/str_cat.h"
-#include "absl/strings/str_format.h"
 #include "absl/strings/string_view.h"
 #include "xla/array4d.h"
 #include "xla/client/xla_builder.h"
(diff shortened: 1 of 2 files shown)


xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -91,7 +89,7 @@ class MultiHeadedAttentionTest : public GpuCodegenTest {
   DebugOptions GetDebugOptionsForTest() override {
     auto debug_options = HloTestBase::GetDebugOptionsForTest();
     debug_options.set_xla_gpu_enable_xla_runtime_executable(false);
-    debug_options.set_xla_gpu_enable_cudnn_fmha(false);
+    debug_options.set_xla_gpu_enable_cudnn_fmha(true);
```
thomasjoerg, line 92: Nit: The code below would be more concise and easier to read if you just pass the `bool config_with_fmha` into this function.

thomasjoerg (changes requested):

xla/service/gpu/tests/gpu_fused_mha_test.cc
sergachev: This function is an override, it does not accept parameters. Functions like ParseAndReturnVerifiedModule apply debug options set by GetDebugOptionsForTest automatically. You are right though, code can be simplified further, I did that.

thomasjoerg (approved):

PR #12238, opened 2024-05-07
Do not promote fp8 operands until GemmAutotuning has been done.

elfiegg:
Do not promote fp8 operands until GemmAutotuning has been done.
The PR doesn't change the current behavior as all fp8 gemms are currently processed by GemmRewriter(translated into cublasLT calls) and will not reach GemmAutotuner. 
We're planning to let fp8 gemm fusions go through the autotuner.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -56,12 +56,12 @@ limitations under the License.
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/Utils/SplitModule.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"  // from @llvm-project
-#include "mlir/IR/Builders.h"  // from @llvm-project
-#include "mlir/IR/BuiltinOps.h"  // from @llvm-project
-#include "mlir/IR/Diagnostics.h"  // from @llvm-project
-#include "mlir/IR/DialectRegistry.h"  // from @llvm-project
-#include "mlir/IR/OwningOpRef.h"  // from @llvm-project
-#include "mlir/Support/LLVM.h"  // from @llvm-project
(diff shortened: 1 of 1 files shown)


xla/service/gpu/gpu_compiler.cc
```diff
@@ -1369,15 +1369,18 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
   const FloatSupport f8e4m3b11fnuz_support(F8E4M3B11FNUZ, F16);
   const GpuFloatSupport f8e5m2fnuz_support(gpu_version, F8E5M2FNUZ, F16);
   const GpuFloatSupport f8e4m3fnuz_support(gpu_version, F8E4M3FNUZ, F16);
-  auto add_float_normalization = [&](HloPassPipeline& pipeline) {
+  auto add_float_normalization = [&](HloPassPipeline& pipeline,
+                                     bool fp8_support = false) {
```
tdanyluk, line 1373: Maybe change "bool fp8_support" to "bool promote_fp8", and remove the default value, to make it more explicit.

tdanyluk (commented):
Thanks.
I think it's fine as long as it really doesn't regress anything.
I just had one short comment.

mooskagh:
Looks good, but note that I'm currently enabling the fp8 fusions to go through Triton, meaning that:
* `add_float_normalization` won't normalize fp8 in any case.
* `GemmRewriter` won't have `fp8` param, it will do both rewrites at the same time (first time -- after `GemmFusion`)
* `GemmFusion` will fuse fp8 gemms.

mooskagh:
Thinking about it a bit more, I'm struggling a bit to understand what is the intended plan.

If we normalize fp8 in the first call but not in the second, what would be the example flow where it makes a difference?

PR #12297, opened 2024-05-09
Fix side output index computation for transpose fusion

ghost:
There exist bug in ComputeThreadIdToOutputIndexing func. Currently, this func can not calculate indexing map correctly for sided output. Fix it and add corresponding test.


lingzhi98:
And I find current mlir transpose emitter does not support 2-1-0 permutaion. Can I know openxla's plan for it?

xla/service/gpu/fusions/transpose_test.cc
```diff
@@ -246,6 +246,71 @@ TEST_F(TransposeTest, ThreadIndexingPartialBlock) {
       )"));
 }
 
+TEST_F(TransposeTest, ThreadIndexingSidedOutput) {
```
jreiffers, line 249: Nit: s/Sided/Side/

xla/service/gpu/fusions/transpose.cc
```diff
@@ -288,8 +288,14 @@ std::optional<IndexingMap> TransposeFusion::ComputeThreadIdToOutputIndexing(
   const auto& hero = analysis_.fusion_hero(root_index).instruction();
   const auto& root = analysis_.fusion_root(root_index).instruction();
   if (!GetDescriptionForTiledTransposeEmitter(root, hero)) {
-    // Non-transpose roots are elementwise by definition.
-    return ComputeThreadIdToInputIndexing(root_index, 0, ctx);
+    // The shape of non-transpose roots are bitcast compatible with the input
+    // shape of transpose heros.
```
jreiffers, line 292: Nit: s/heros/heroes/

xla/service/gpu/fusions/transpose_test.cc
```diff
@@ -80,7 +80,7 @@ TEST_F(TransposeTest, ThreadIndexing021) {
         (d0, d1, d2, d3, d4, d5)[s0, s1, s2] -> (
           d3 floordiv 2,
           d0 floordiv 32 + s1 * 4,
-          (d3 mod 2) * 32 + d0 mod 32
+          ### (d3 mod 2) * 32 + d0 mod 32 ###
```
jreiffers, line 83: What's this?

xla/service/gpu/fusions/transpose_test.cc
```diff
@@ -246,6 +246,71 @@ TEST_F(TransposeTest, ThreadIndexingPartialBlock) {
...
+                    .value();
+
+  auto* root = module->entry_computation()->root_instruction();
+  auto analysis = AnalyzeFusion(*root, device_info_);
+
+  TF_ASSERT_OK_AND_ASSIGN(auto fusion, GetTransposeFusion(analysis));
+  mlir::MLIRContext mlir_context;
+
+  EXPECT_THAT(
+      fusion->ComputeThreadIdToInputIndexing(1, 0, &mlir_context)->ToString(),
```
jreiffers, line 274: This function is currently a bit ill-defined. It claims to be for the operand of the hero, but that's not a meaningful concept for side outputs.

Your implementation is correct in that it returns the indexing for %bitcast0, but I don't see how that would ever be useful. What we actually need is the indexing for the root itself. I'm not entirely sure how to update the function specification to make that not a mess though.

Can you add a comment here explaining what's going on?

xla/service/gpu/fusions/transpose_test.cc
```diff
@@ -246,6 +246,71 @@ TEST_F(TransposeTest, ThreadIndexingPartialBlock) {
...
+        s0 in [0, 0]
+        s1 in [0, 7]
+        s2 in [0, 0]
+      )"));
+  EXPECT_THAT(
+      fusion->ComputeThreadIdToOutputIndexing(1, &mlir_context)->ToString(),
+      MatchIndexingString(R"(
+        (d0, d1, d2, d3, d4, d5)[s0, s1, s2] -> (
+          d3 floordiv 2,
+          ### (d0 floordiv 32) * 64 + d0 mod 32 + s1 * 256 + (d3 mod 2) * 32 ###
```
jreiffers, line 298: Again, probably needs to be removed?

xla/service/gpu/fusions/transpose_test.cc
jreiffers, line 274: Let's add a test that verifies that the input indexing for root 0 is the same as the input indexing for root 1 (composed with %bitcast0, so we get the indexing for %input).

jreiffers (commented):
Thank you!

xla/service/gpu/fusions/transpose_test.cc
akuegel: See https://github.com/openxla/xla/commit/a70d147e4e5dcb3824af1aaec91d5b070ae1389c which added this kind of matching

jreiffers: That's weird. We should probably fix that properly before it spreads everywhere. These expressions must have a canonical form.

xla/service/gpu/fusions/transpose_test.cc
lingzhi98: Yeah, this function is not used by side outputs. Why I add this test just follow previous test function, maybe I can remove it.

lingzhi98: I find there exist [usage](https://github.com/openxla/xla/blob/main/xla/service/gpu/model/coalescing_analysis.cc#L167) of this func by side output. So this func has meaning also for side outputs to analysis memory coalescing if side outputs has different input compared with real hero, see this hlo:
fusion {
      %input0 = f32[100,32,64] parameter(0)
      %input1 = f32[100,32,64] parameter(1)
      %bitcast0 = f32[100,64,32] bitcast(%input1)
      %transpose = f32[100,64,32] transpose(%input0), dimensions={0,2,1}
      %bitcast1 = f32[100,2048] bitcast(%bitcast0)
      ROOT %tuple = (f32[100,64,32], f32[100,2048]) tuple(%transpose, %bitcast1)
}
I think we should keep this check.

lingzhi98: They are not same in my opinion. The input indexing of root0 corresponds to [100, 32, 64] but root1 corresponds to [100, 64, 32]. If my understanding is wrong, please correct me.

jreiffers: I agree, the check should stay. That's why I asked for a brief explanation in a code comment.

jreiffers: Unless I'm missing something, the test compares the input indexing for root 1 to the output indexing for root 1. We should compare the input indexing for root 0 and 1.

xla/service/gpu/fusions/transpose_test.cc
jreiffers: Will need adjustment after https://github.com/openxla/xla/pull/12457 lands.

xla/service/gpu/fusions/transpose_test.cc
lingzhi98: I think you may misunderstand something. Line 273 compare the expected input indexing for root1 to actual input indexing for root1. If compare to output indexing for root1, the size of affine map results should be 3 not 2, can see Line 276.

jreiffers: My comment was only about **input** indexing, not about outputs.

lingzhi98: I may understand what you mean. You advice me add test to check input indexing for root0 in order to ensure ComputeThreadIdToInputIndexing func can get correct root0 input indexing result for the tested hlo with side output, right?

jreiffers: Yes, we should verify that the input for the transpose and the side output have the same indexing (so the reads will be shared).

lingzhi98: You can see if this test as you expected. https://github.com/openxla/xla/pull/12297/files#diff-77c8f926d853dcddb6d390925c7dd4bcc05f4a811a4cd6a3946689b3c3b58683R249

lingzhi98:
This PR is ready to merge. Have any other comments?

jreiffers:
Sorry for the delay, this fell of my radar. Looks good, thank you.

jreiffers (approved):

xla/service/gpu/fusions/transpose_test.cc
```diff
@@ -243,6 +243,101 @@ TEST_F(TransposeTest, ThreadIndexingPartialBlock) {
       )"));
 }
 
+TEST_F(TransposeTest, SameInputIndexingForRealHeroAndSideOutput) {
```
akuegel, line 246: It seems this test needs some adjustments in the test expectations after latest merge. In our imported version of this PR, the test is failing

akuegel (approved):

akuegel (approved):

lingzhi98:
Does this PR still fail? I see the status of feedback/copybara is failure. But I passed the tests on A100, can I know the detail information about failure?

akuegel:
> Does this PR still fail? I see the status of feedback/copybara is failure. But I passed the tests on A100, can I know the detail information about failure?

Seems to pass now internally. The open source CI is currently broken due to a config error, and we need to wait for someone in US time zone to fix it.

PR #12328, opened 2024-05-10
Make shared cache read/write logic more clearly for transpose mlir emitter

ghost:
Current transpose mlir emitter allocate shared cache with shape 32x1x32 for transpose 2-1-0. But the read indices of shared cache are {0, y, x} as [this line](https://github.com/openxla/xla/blob/main/xla/service/gpu/fusions/transpose_mlir.cc#L190) shows, which is not compatible with 32x1x32 shape. What's strange is that transpose 2-1-0 can run successfully using transpose mlir emitter. I find the reason is that lower tensor pass use [linear index](https://github.com/openxla/xla/blob/main/xla/service/gpu/fusions/mlir/lower_tensors.cc#L148) to access shared cache, which is lucky to get right result. For example, the strides of 32x1x32 are {32, 32, 1}, and the linear index of {0, y ,x} is 0 * 32 + y * 32 + 32.
I am not sure if it is as expected or just mistake. If reviewer think no need of this PR, feel free to close.

### DIFF
--- a/xla/service/gpu/fusions/transpose_mlir.cc
+++ b/xla/service/gpu/fusions/transpose_mlir.cc
@@ -178,14 +178,23 @@ IndexingMap GetSharedMemoryWriteIndexingMap(
     const IndexingMap& thread_id_indexing, int loop_dim) {
   auto* mlir_context = thread_id_indexing.GetMLIRContext();
 
+  AffineExpr c0 = mlir::getAffineConstantExpr(0, mlir_context);
   AffineExpr th_x = mlir::getAffineDimExpr(0, mlir_context);
   SmallVector<AffineExpr, 3> tile_sizes(3);
   mlir::bindSymbolsList(mlir_context, llvm::MutableArrayRef(tile_sizes));
+  SmallVector<AffineExpr, 3> shared_memory_indices = {
+      th_x.floorDiv(32) + 4 * tile_sizes[loop_dim], th_x % 32};
+  for (auto [index, range_val] :
+       llvm::enumerate(thread_id_indexing.GetRangeVars())) {
+    if (range_val.range.NumElements() == 1) {
+      shared_memory_indices.insert(shared_memory_indices.begin() + index, c0);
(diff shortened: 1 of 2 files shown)


jreiffers:
Good catch! There's another mistake here, the tile size should be 32x33, not 32x32. Could you fix this as well while you're here? Thanks.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -214,6 +212,9 @@ MlirTransposeFusion::WriteResult MlirTransposeFusion::EmitWriteToShMemMlir(
     ValueRange output_args) const {
   std::vector<int64_t> shmem_tensor_size(tiling_.GetBlockTileSize().begin(),
                                          tiling_.GetBlockTileSize().end());
+  shmem_tensor_size.erase(shmem_tensor_size.begin() +
+                          (permutation_ == Vector3({0, 2, 1}) ? 0 : 1));
```
jreiffers, line 216: Can we just keep this? The 1 doesn't really hurt, and this line is not trivial to understand without context.

lingzhi98: If we keep this, we should make [the logic](https://github.com/openxla/xla/pull/12328/files#diff-0210cd152294b1475469517ffba6b5ee578acc4e786b0dceeee6854aba445a29L190) here more complex, because c0 should place on different position. Which do you recommend?

jreiffers: I'd prefer that, if it's not too much logic. FYI we also have some plans to extend this emitter to support arbitrary transposes (not just ones that can be mapped to 3d).

lingzhi98:
> Good catch! There's another mistake here, the tile size should be 32x33, not 32x32. Could you fix this as well while you're here? Thanks.

Fix it.

jreiffers (approved):
Thank you!

PR #12438, opened 2024-05-13
[XLA:GPU] add force inline option to get better llvm splits

Cjkkkk:
Add option to XLA to enforce inlining before llvm splitModule or set preserveLocals=False to get more balanced splits in parallel compilation case.

Some data of GPT3 5B model with different setting:
```
Compilation: TSL:XlaCompile:#module=pjit__wrapped_step_fn,program_id=24#: 3.754429084s (parallel + inline) runtime: 1.0s
Compilation: TSL:XlaCompile:#module=pjit__wrapped_step_fn,program_id=24#: 4.676450341s (parallel + no inline) runtime: 1.0s
Compilation: TSL:XlaCompile:#module=pjit__wrapped_step_fn,program_id=24#: 3.051018704s (parallel + perserve_local=False) runtime: 1.4s
Compilation: TSL:XlaCompile:#module=pjit__wrapped_step_fn,program_id=24#: 4.862938161s (serial) runtime: 1.0s
```

However, the runtime per step of perserve_locals=False vs other three setup is 1.4s vs 1.0s.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -193,6 +193,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_while_loop_reduce_scatter_code_motion(false);
 
   opts.set_xla_gpu_collective_inflation_factor(1);
+  opts.set_xla_llvm_split_module_level(0);
 
   opts.set_xla_gpu_exhaustive_tiling_search(false);
 
@@ -1095,6 +1096,15 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       debug_options->xla_gpu_collective_inflation_factor(),
       "Inflation factor for collectives. If set to > 1, each XLA/GPU "
       "collective will execute multiple times (will yield incorrect results)"));
+
+  flag_list->push_back(tsl::Flag(
+      "xla_llvm_split_module_level",
+      int32_setter_for(&DebugOptions::set_xla_llvm_split_module_level),
+      debug_options->xla_llvm_split_module_level(),
+      "Decide how llvm modules are split: Whether to force inline or preserve "
+      "locals. 0 = no inline, 1 = force inline before split, 2 = no line and "
+      "PreserveLocals=False during split, default is 0"));
+
   flag_list->push_back(tsl::Flag(
       "xla_gpu_enable_reassociation_for_converted_ar",
       bool_setter_for(
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -193,7 +193,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_while_loop_reduce_scatter_code_motion(false);
 
   opts.set_xla_gpu_collective_inflation_factor(1);
-  opts.set_xla_llvm_split_module_level(0);
+  opts.set_xla_llvm_force_inline_before_split(true);
 
   opts.set_xla_gpu_exhaustive_tiling_search(false);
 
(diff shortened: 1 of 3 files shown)


xla/debug_options_flags.cc
```diff
@@ -1095,6 +1096,15 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
...
       "Inflation factor for collectives. If set to > 1, each XLA/GPU "
       "collective will execute multiple times (will yield incorrect results)"));
+
+  flag_list->push_back(tsl::Flag(
+      "xla_llvm_split_module_level",
+      int32_setter_for(&DebugOptions::set_xla_llvm_split_module_level),
+      debug_options->xla_llvm_split_module_level(),
+      "Decide how llvm modules are split: Whether to force inline or preserve "
+      "locals. 0 = no inline, 1 = force inline before split, 2 = no line and "
+      "PreserveLocals=False during split, default is 0"));
```
joker-eph, line 1106: Why isn't 1 the better default?

Cjkkkk: Was hoping google can benchmark their internal workloads to make sure it does not break anything so we can make it default.

Cjkkkk: I will rerun experiments to ensure the slowdown is actually there not because i was competing the dgx resources with someone else. Compiler team do mention that by not doing inline, some optimization opportunities are not available in ptxas (for example, loop invariant analysis would not move dowork() outside any loops, alias analysis will become more conservative). 

Another thing is that I was told maybe in some workloads people are willing to sacrifice some perf for faster compilation. But for sure, the 1.4x slowdown is too fast. I should be able to run this on other models as well to get more info on this option.

cheshire:
> Compilation: TSL:XlaCompile:#module=pjit__wrapped_step_fn,program_id=24#: 3.754429084 (parallel + inline)

What are the units, seconds?

Mentioning both runtime and compile time in the bug description is a bit confusing, what is the overall effect of the change? No runtime change and ~20% faster compilation?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1787,6 +1788,29 @@ GpuCompiler::CompileToTargetBinary(const HloModuleConfig& module_config,
...
+                                   func.getName().str(), func.getNumUses());
+        std::vector<llvm::CallInst*> calls_to_inline;
+        for (auto* user : func.users()) {
+          if (auto* call = llvm::dyn_cast<llvm::CallInst>(user)) {
+            calls_to_inline.push_back(call);
+          }
+        }
+        for (auto* call_to_inline : calls_to_inline) {
+          llvm::InlineFunctionInfo inline_function_info;
+          CHECK(llvm::InlineFunction(*call_to_inline, inline_function_info)
```
cheshire, line 1807: Not sure if we want to CHECK

cheshire: Return bad Status instead?

cheshire (changes requested):
Is it possible to test this?

joker-eph:
I really don't see a case for `preserveLocals`, unless we setup a ThinLTO-like backend which allows to cross-import and still inline.

Cjkkkk:
Update some more models compilation with this changes: https://docs.google.com/spreadsheets/d/1uIRf66UT9hOBOge3nvRZebDintgM0zmozNts0tOiXQA/edit?usp=sharing. Seems preserveLocals=False is not doing any better than parallel + inline. So i will just remove that.

xla/service/gpu/gpu_compiler.cc
Cjkkkk: I return absl:InternalError instead.

Cjkkkk:
@cheshire Hi, any updates on this?

cheshire (approved):

Issue #11772, opened 2024-04-23
[XLA:CPU] oneDNN Softmax gives inaccurate results


penpornk:
From https://github.com/google/jax/issues/20856,
`jax.nn.softmax` returns different results under `jax.jit` because the `jax.jit` version calls oneDNN custom call.

Code:
```
import jax
import jax.numpy as jnp

x = jax.random.normal(jax.random.key(0), (3, 1, 1))

def f(x):
    return jax.nn.softmax(x, axis=0)

print(f(x))
print(jax.jit(f)(x))
```

Results
```
[[[0.75250584]]

 [[0.0755428 ]]

 [[0.17195134]]]
[[[1.]]

 [[1.]]

 [[1.]]]
```

Need to investigate why the oneDNN softmax custom call version is giving all 1s.

jakevdp:
Please also note that this behavior is not observed at HEAD, because the HLO generated by `softmax` changed after https://github.com/google/jax/pull/20643 and so the rewrite rule no longer emits the custom call.

penumallas1261:
Hi @penpornk ,
I can see the latest code is fixed the issue and following is the [gist](https://colab.sandbox.google.com/gist/penumallas1261/fc148a51768fc8065cbb9fd052b829f8/untitled0.ipynb), Please feel free to close the issue if it's resolved.

penpornk:
Hi @penumallas1261,
Thank you for bringing this up! I believe the issue may have been hidden as oneDNN is temporarily disabled on JAX and in XLA. But it will take a while before we re-enabled it. So I think we can close this issue for now. We can reopen it later if/when we use oneDNN primitives by default again.

penumallas1261:
Internal reproducer submitted in cl/781834261.

PR #12537, opened 2024-05-15
[XLA:CPU][oneDNN][BugFix] Add axis support to oneDNN softmax rewriter

sachinmuradi:
This PR fixes :
XLA Issue : https://github.com/openxla/xla/issues/11772
corresponding JAX issue : https://github.com/google/jax/issues/20856

Accuracy fix :
This PR adds support for softmax axis other than -1, previously oneDNN softmax was always executed with axis = -1 (last dimension)
Accuracy issue is observed from JAX 0.4.22 till last release, with current main branch we don't see accuracy issue as the JAX softmax HLO pattern changed (https://github.com/google/jax/pull/20643) and HLO pattern is not re-written to oneDNN softmax. Hence, this PR also adjusts the oneDNN softmax pattern to recognize new HLO pattern and rewrite into oneDNN softmax custom call.


xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -61,11 +63,35 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+  auto is_valid_reduce_max = [](const HloInstruction* reduce) {
+    HloComputation* reducer = reduce->to_apply();
+    return (reducer->root_instruction()->opcode() == HloOpcode::kMaximum) &&
+           (reduce->dimensions().size() == 1) &&
+           (reduce->operand(1)->opcode() == HloOpcode::kConstant) &&
+           is_neg_inf_const_scalar(reduce->operand(1));
+  };
+
+  return m::AnyOf<HloInstruction>(
+      m::Maximum().WithBinaryOperandsAnyOrder(
```
bchetioui, line 84: Is this particular match (`max(reduce, const)`) tested anywhere?

If not, please add a test for it.

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -95,24 +121,25 @@ std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr) {
...
+  HloInstruction* reduce_instr;
 
   // Lower diamond
   if (!Match(instr,
              m::Divide(
                  m::Exp(&left_exponential, m::Op()),
                  m::Broadcast(m::Reshape(
                      m::Broadcast(OptionalConvert(m::Reshape(OptionalConvert(
-                         m::Reduce(OptionalConvert(
+                         m::Reduce(&reduce_sum,
```
bchetioui, line 134: Why do we match `max(reduce_max, min)` above but not `add(reduce_sum, 0)` here?

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -61,11 +63,35 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+  };
+
+  return m::AnyOf<HloInstruction>(
+      m::Maximum().WithBinaryOperandsAnyOrder(
+          m::Reduce(instr).WithPredicate(is_valid_reduce_max).WithOneUse(),
+          m::Constant().WithPredicate(is_neg_inf_const_scalar)),
+      m::Reduce(instr).WithPredicate(is_valid_reduce_max).WithOneUse());
+}
+
+std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr, int* axis) {
```
bchetioui, line 90: Can you clarify in the documentation what `instr` is supposed to correspond to? Is it the root of the operation?

Please also note that what we return is the `producer` of the pattern, and that the reduction axis is populated as a side effect on successful matches.

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -61,11 +63,35 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
   //
   // producer
   // |   \
-  // |  reduce_max
+  // |  reduce_max or max(reduce_max))
```
bchetioui, line 94: I see optional converts mentioned in the matcher but not mentioned in the documentation of this pattern. Could you please make sure that the pattern described here matches the pattern implementation?

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -125,29 +152,28 @@ std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr) {
...
+                         m::Broadcast(m::Reshape(m::Broadcast(
+                                          m::Reshape(m::Op(&reduce_instr)))))
+                             .WithOneUse())
                  .WithOneUse())) {
     return std::nullopt;
   }
 
-  if (left_producer != right_producer || left_producer->user_count() != 2) {
+  // Match the reduce max
+  if (!Match(reduce_instr, MaxReduce(&reduce_max))) return std::nullopt;
```
bchetioui, line 164: nit: please wrap the condition's body in braces:

```
if (x) { 
  return std::nullopt;
}
```

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -125,29 +152,28 @@ std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr) {
...
+
+  if (left_producer != reduce_max->operand(0) ||
+      left_producer->user_count() != 2) {
     return std::nullopt;
   }
 
+  if ((reduce_sum->dimensions()[0]) != (reduce_max->dimensions()[0])) {
```
bchetioui, line 171: nit: please remove the unnecessary parentheses

bchetioui (changes requested):

xla/tests/onednn_softmax_test.cc
```diff
@@ -99,6 +106,46 @@ TEST_F(OneDnnSoftmaxTest, SoftmaxFP32) {
...
+          Arg_0.6 = f32[] parameter(0)
+          Arg_1.7 = f32[] parameter(1)
+          ROOT maximum.8 = f32[] maximum(Arg_0.6, Arg_1.7)
+        }
+        region_1.17 {
+          Arg_0.18 = f32[] parameter(0)
+          Arg_1.19 = f32[] parameter(1)
+          ROOT add.20 = f32[] add(Arg_0.18, Arg_1.19)
+        }
+        ENTRY main.27 {
```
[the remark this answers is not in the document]
sachinmuradi: Done

xla/tests/onednn_softmax_test.cc
```diff
@@ -99,6 +106,46 @@ TEST_F(OneDnnSoftmaxTest, SoftmaxFP32) {
...
+          Arg_0.1 = f32[3,1,1]{2,1,0} parameter(0)
+          constant.4 = f32[] constant(-inf)
+          reduce.9 = f32[1,1]{1,0} reduce(Arg_0.1, constant.4), dimensions={0}, to_apply=region_0.5
+          constant.2 = f32[1,1]{1,0} constant({ {-inf} })
+          maximum.10 = f32[1,1]{1,0} maximum(reduce.9, constant.2)
+          reshape.11 = f32[1,1,1]{2,1,0} reshape(maximum.10)
```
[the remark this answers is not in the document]
sachinmuradi: Yes, we had to add reshapes in pattern coz the HLO Passes before onednn_ops_rewrite introduce some reshapes and also we saw hlo from jax also has reshapes and broadcasts. The test cases and pattern hence include reshapes.

xla/tests/onednn_softmax_test.cc
```diff
@@ -99,6 +106,46 @@ TEST_F(OneDnnSoftmaxTest, SoftmaxFP32) {
...
+          reduce.21 = f32[1,1]{1,0} reduce(exponential.16, constant.3), dimensions={0}, to_apply=region_1.17
+          reshape.22 = f32[1,1,1]{2,1,0} reshape(reduce.21)
+          broadcast.23 = f32[1,1,1]{2,1,0} broadcast(reshape.22), dimensions={0,1,2}
+          reshape.24 = f32[1,1]{1,0} reshape(broadcast.23)
+          broadcast.25 = f32[3,1,1]{2,1,0} broadcast(reshape.24), dimensions={1,2}
+          ROOT divide.26 = f32[3,1,1]{2,1,0} divide(exponential.16, broadcast.25)
+        }
+    )";
+
+  EXPECT_TRUE(RunAndCompare(hlo_string, ErrorSpec{1e-4, 1e-4}));
```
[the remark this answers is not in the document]
sachinmuradi: For every test , we check the numerical correctness as well as the custom call target to verify if the pattern matched via filecheck. Though it might be redundant to add check for just rewrite pass, I included them for every test, would that work ?

xla/tests/onednn_softmax_test.cc
```diff
@@ -99,6 +106,46 @@ TEST_F(OneDnnSoftmaxTest, SoftmaxFP32) {
     )";
 
   EXPECT_TRUE(RunAndCompare(hlo_string, ErrorSpec{1e-4, 1e-4}));
+  MatchOptimizedHlo(hlo_string, onednn_softmax_);
+}
+
+TEST_F(OneDnnSoftmaxTest, SoftmaxFP32_0) {
```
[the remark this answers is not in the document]
sachinmuradi: Done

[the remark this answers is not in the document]
sachinmuradi: Done, added parametrized test and also some specific tests that we observed when running models.

xla/service/cpu/ir_emitter.cc
```diff
@@ -2728,6 +2728,16 @@ Status IrEmitter::HandleOneDnnLayerNorm(HloInstruction* custom_call) {
 }
 
 Status IrEmitter::HandleOneDnnSoftmax(HloInstruction* custom_call) {
+  // OneDnnSoftmaxConfig.
+  auto typed_custom_call = Cast<HloCustomCallInstruction>(custom_call);
```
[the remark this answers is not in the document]
sachinmuradi: Caller [here](https://github.com/openxla/xla/blob/ea6c3d5916b1c5125f1d8d3e440cbbd9f62c2343/xla/service/cpu/ir_emitter.cc#L2772) for HandleOneDnnSoftmax, passes HloInstruction* as input. So we will either have to cast it to HloCustomCallInstruction here in function body or at function call. I kept it in function body, hope thats fine.

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -61,11 +63,35 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
   return shift;
 }
 
-std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr) {
+bool is_neg_inf_const_scalar(const HloInstruction* const_instr) {
```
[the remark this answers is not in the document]
sachinmuradi: Added check for opcode::kConstant.

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -61,11 +63,35 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+  if (!ShapeUtil::IsEffectiveScalar(const_instr->shape())) return false;
+  auto value = LiteralUtil::GetFirstScalarLiteral(const_instr->literal());
+  return literal_comparison::Equal(
+             value, LiteralUtil::MinValue(const_instr->shape().element_type()))
+      .ok();
+}
+
+auto MaxReduce(HloInstruction** instr) {
```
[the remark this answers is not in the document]
sachinmuradi: Added documentation to describe the pattern here

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -61,11 +63,35 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+  auto is_valid_reduce_max = [](const HloInstruction* reduce) {
+    HloComputation* reducer = reduce->to_apply();
+    return (reducer->root_instruction()->opcode() == HloOpcode::kMaximum) &&
```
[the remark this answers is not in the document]
sachinmuradi: Could you elaborate more on this comment ? The is_valid_reduce_max looks precise enough ,as in, it checks for Maximum Computation and 2nd input as constant scalar and is negative infinity.

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Yes, The test SoftmaxFP32OnAxisZero verifies this case.

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: We added max(reduce_max, -inf) based on the patterns we saw for actual jax models, we didn't encounter a case for later for jax/TF models.

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Done

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Done

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Done

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Done

xla/tests/onednn_softmax_test.cc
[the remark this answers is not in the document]
bchetioui: Thanks---but indeed, done as is, it is a bit redundant!

What I wanted to suggest is that you write tests that do not exercise the full pipeline, but only test the pass. I think there are three types of tests you want to have:

1. **Pass-specific tests**: tests that check that your pass itself matches what it should. Those tests should call into the pass and make sure that it rewrites the right patterns into the right expected custom call; 
2. **Pipeline test**: test that checks that the compiler pipeline calls your pass appropriately. This test should call one full compilation loop and make sure that the relevant pass is called;
3. **Numerics tests**: tests that your custom-calls produce the right numerics vs a reference implementation. These tests should execute already optimized code (post pattern matching) and compare them against a reference implementation corresponding to what should have been matched.

Today, (1), (2) and (3) are sort of all bundled together. I claim that it makes sense to separate the 3 types of tests into separate tests, for several reasons:

* Bundling all three types of tests together mean that we always pay more runtime cost than necessary for the same coverage.
* In general, (1) and (2) are tests that shouldn't rely on specific hardware, but (3) may require the test to run on a specific platform. Splitting the test targets means that we can save resources by default, and only specialise on the platform where necessary.
* Testing (2) by matching the existence of a custom call at the end of the pipeline is somewhat brittle and intrusive. Someone adding a pass that causes your input HLO to look different at the start of your pass might need to come back to fix these tests, even though they might have not broken anything. This means that we want to (at the very least) have as few of this type of tests as possible to minimize churns. In your case, I think that checking that the optimized HLO has gone through the right pass by looking at the module metadata would be enough.

The request here would thus be to split tests into these three categories. Does that make more sense?

xla/service/cpu/onednn_ops_rewriter.cc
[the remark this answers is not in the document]
bchetioui: As requested below, please wrap conditional blocks in braces.

We follow [Google's C++ style guide](https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching) which requests this is the case for all conditional blocks.

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -84,6 +86,9 @@ auto MaxReduce(HloInstruction** instr) {
       m::Reduce(instr).WithPredicate(is_valid_reduce_max).WithOneUse());
 }
 
+// Match the softmax patter with divide instruction as root node.
```
bchetioui, line 89: Use active voice (i.e. Match -> Matches) for function comments. See [style guide](https://google.github.io/styleguide/cppguide.html#Function_Comments).

Also: patter -> pattern.

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -84,6 +86,9 @@ auto MaxReduce(HloInstruction** instr) {
...
+// Here we pass 'instr' as root node and return the producer HloInstruction
```
bchetioui, line 90: Please make sure to format sentences correctly, e.g., here:

"return the producer HloInstruction[. The a]xis on which softmax is applied is stored in 'axis'".

xla/service/cpu/onednn_ops_rewriter.cc
bchetioui: Thank you this is clearer!

xla/service/cpu/onednn_ops_rewriter.cc
bchetioui: Thanks for explaining. Would you mind adding a comment with this note?

xla/tests/onednn_softmax_test.cc
```diff
@@ -27,208 +32,177 @@ limitations under the License.
...
+  std::tie(data_type, batch_size) = GetParam();
+  if (!IsSupportedType(data_type)) {
+    GTEST_SKIP() << "CPU does not support "
+                 << primitive_util::LowercasePrimitiveTypeName(data_type);
+  }
+  float atol = (data_type == F32) ? 1e-4 : 1e-2;
+  float rtol = (data_type == F32) ? 1e-4 : 1e-2;
+
+  const std::string softmax_hlo_template_string = R"(
+        HloModule softmax_module, entry_computation_layout={($0[$1,128,30522]{2,1,0})->$0[$1,128,30522]{2,1,0}}
```
bchetioui, line 85: nit: here and in other tests, I don't think the `entry_computation_layout` needs to be specified. Please drop it if unnecessary for the sake of brevity :)

xla/tests/onednn_softmax_test.cc
```diff
@@ -27,208 +32,177 @@ limitations under the License.
...
-            exponential.14 = f32[1,128,30522]{2,1,0} exponential(subtract.13)
-            constant.2 = f32[] constant(0)
-            reduce.19 = f32[1,128]{1,0} reduce(exponential.14, constant.2), dimensions={2}, to_apply=region_1.15
-            reshape.20 = f32[1,128,1]{2,1,0} reshape(reduce.19)
-            broadcast.21 = f32[1,128,1]{2,1,0} broadcast(reshape.20), dimensions={0,1,2}
-            reshape.22 = f32[1,128]{1,0} reshape(broadcast.21)
-            broadcast.23 = f32[1,128,30522]{2,1,0} broadcast(reshape.22), dimensions={0,1}
-            ROOT divide.24 = f32[1,128,30522]{2,1,0} divide(exponential.14, broadcast.23)
+        ENTRY main {
+            Arg_0 = $0[$1,128,30522]{2,1,0} parameter(0), sharding={replicated}
```
bchetioui, line 97: Is this sharding annotation necessary? If not, please remove it.

xla/service/cpu/ir_emitter.cc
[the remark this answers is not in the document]
bchetioui: I think that the right thing to do is to perform this cast in the caller---then the caller has the responsibility for ensuring that what they pass is a valid custom call. Is there a particular reason you think this is better as is?

xla/service/cpu/onednn_ops_rewriter.cc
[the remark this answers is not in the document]
bchetioui: My point is that you don't really check for a maximum computation here: you check that the root of the reducer computation is `maximum`, but it would match correctly a reduction computation such as e.g.

```
x = parameter(0)
y = parameter(1)
min_0 = minimum(x, y)
min_1 = minimum(y, x)
ROOT max = maximum(min_0, min_1)
```

which, would not be a `maximum` computation, would involve a `maximum` root, but still obey the monoid property (and hence is valid HLO to write).

Thus, I think your check for the reducer should be stricter, and ensure that the parameters to the `maximum` root are the parameters of the reducer computation---in any order.

Does this make sense?

xla/service/cpu/onednn_ops_rewriter.cc
bchetioui: Got it, thanks for explaining. Would you mind adding a comment in the code to explain that this is necessary for cases you saw in practice?

bchetioui (changes requested):
Thanks for addressing my comments @sachinmuradi! A few more, but this is already looking better!

xla/tests/onednn_softmax_test.cc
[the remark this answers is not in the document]
sachinmuradi: Makes sense and thank you for clarification. I modified the test to just check the rewrite pass.

xla/service/cpu/ir_emitter.cc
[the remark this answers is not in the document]
sachinmuradi: No particular reason, but I tried keep it consistent with other custom call handlers in the caller function body : https://github.com/openxla/xla/blob/b983bd61044bf52838661057f570e9e218803549/xla/service/cpu/ir_emitter.cc#L2762

xla/service/cpu/onednn_ops_rewriter.cc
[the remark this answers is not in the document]
sachinmuradi: Done

xla/service/cpu/onednn_ops_rewriter.cc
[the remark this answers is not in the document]
sachinmuradi: Yes you are right, thank you for catching it and explaining it. Agreed that it needs to have more precise check, I modified the code accordingly, hope thats fine

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Modified the description of pattern matching to address this

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Done

xla/service/cpu/onednn_ops_rewriter.cc
sachinmuradi: Done

xla/tests/onednn_softmax_test.cc
sachinmuradi: Done

xla/tests/onednn_softmax_test.cc
sachinmuradi: Removed it

sachinmuradi:
Thank you @bchetioui for review, I addressed more review comments, can you take a look ?

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -88,28 +146,33 @@ std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr) {
   // |   /
   // divide  // (instr parameter)
   //
-  // where both reductions occur only on the last axis.
+
+  // Though this is most common SoftMax pattern we see and The patterns added
```
bchetioui, line 150: The capitalization is not correct here, please make sure to follow the classical grammatical and typographical rules for English language.

But thinking a bit deeper, I think what you are saying can be summarized with just the following sentence:

```
This matcher covers the most common SoftMax patterns we have encountered in real-life models.
```

(unless I missed something).

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -58,11 +60,61 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+      m::Maximum().WithBinaryOperandsAnyOrder(
+          m::Reduce(instr).WithPredicate(is_valid_reduce_max).WithOneUse(),
+          pu::OptionalBroadcast(
+              m::Constant().WithPredicate(is_neg_inf_const_scalar))),
+      m::Reduce(instr).WithPredicate(is_valid_reduce_max).WithOneUse());
+}
+
+// Matches the softmax pattern with divide instruction as root node.
+// Here we pass 'instr' as root node and return the producer HloInstruction.
+// Tha axis on which softmax is applied is stored in 'axis'
```
bchetioui, line 112: Please end sentence with dot (`.`).

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -58,11 +60,61 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+  for (int idx = 0; idx < comp->num_parameters(); ++idx) {
+    if (comp->parameter_instruction(idx) != max_instr->operand(0) &&
+        comp->parameter_instruction(idx) != max_instr->operand(1)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+// Pattern to match AnyOf  Maximum(Reduce_max(...), -inf) OR Reduce_max(...)
```
bchetioui, line 92: Please end sentence with dot (`.`). (And, unless there's a justifiable reason for it, remove double space after `AnyOf`.)

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -58,11 +60,61 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+    return false;
+  };
+  auto value = LiteralUtil::GetFirstScalarLiteral(const_instr->literal());
+  return literal_comparison::Equal(
+             value, LiteralUtil::MinValue(const_instr->shape().element_type()))
+      .ok();
+}
+
+// This function is to make sure the reducer computation is strictly a max
+// reducer and root instruction (kMaximum) has same inputs as computation
```
bchetioui, line 77: "has [the] same inputs", and please end sentence with dot (`.`).

xla/service/cpu/ir_emitter.cc
```diff
@@ -2730,6 +2730,16 @@ absl::Status IrEmitter::HandleOneDnnLayerNorm(HloInstruction* custom_call) {
 }
 
 absl::Status IrEmitter::HandleOneDnnSoftmax(HloInstruction* custom_call) {
+  // OneDnnSoftmaxConfig.
```
bchetioui, line 2733: Can you explain the purpose of this comment? It is either not expansive enough (in which case, please elaborate it), or not useful (in which case, please remove it).

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -122,29 +185,30 @@ std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr) {
...
+             m::Subtract(m::Op(&left_producer),
+                         m::Broadcast(m::Reshape(m::Broadcast(
+                                          m::Reshape(m::Op(&reduce_instr)))))
+                             .WithOneUse())
                  .WithOneUse())) {
     return std::nullopt;
   }
 
-  if (left_producer != right_producer || left_producer->user_count() != 2) {
+  // Match the reduce max
```
bchetioui, line 196: Please end sentence with dot (`.`).

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -58,11 +60,61 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
   return shift;
 }
 
-std::optional<HloInstruction*> MatchSoftmax(HloInstruction* instr) {
+bool is_neg_inf_const_scalar(const HloInstruction* const_instr) {
+  if (const_instr->opcode() != HloOpcode::kConstant) {
+    return false;
+  };
```
bchetioui, line 66: Please remove trailing semicolon after conditional closing brace.

xla/service/cpu/onednn_ops_rewriter.cc
```diff
@@ -58,11 +60,61 @@ HloInstruction* FindLayerNormShift(HloInstruction* instr) {
...
+  if (!ShapeUtil::IsEffectiveScalar(const_instr->shape())) {
+    return false;
+  };
```
bchetioui, line 69: Please remove trailing semicolon after conditional closing brace.

bchetioui (changes requested):
Thanks for so carefully addressing my changes! This looks really good now.

I left a few small comments, but the most important one that I'd like you to fix is the testing, to make sure that we have tests for all the 3 categories that we outlined.

I also left a few comments regarding formatting and documentation. Please address them---but I will not block your change on that so long as formatting is roughly OK :)

penpornk:
Because the JAX release branch cut is coming soon, we made changes addressing @bchetioui's comments on formatting internally and merged this PR. 

@sachinmuradi Could you please help open a follow-up PR adding the following tests? Thank you very much!

> I think however now you're missing tests for the two other categories that I mentioned, i.e. **Pipeline test** and **Numerics test**.
> 
> Could you add a single test for each of these categories, such that we'll be all covered? :) (The rest looks great now!)

penpornk:
Hi @sachinmuradi, 
Any updates on the follow-up PR? :)

sachinmuradi:
> Hi @sachinmuradi, Any updates on the follow-up PR? :)

Hi @penpornk, Thanks for checking on this, want to assure you that I haven't forgotten about it and still have it in mind :) , due to internal high-priority releases that required immediate attention, I had to delay this PR temporarily. I ll get back to this and will create PR by this week, thank you and @bchetioui for your patience.

PR #12494, opened 2024-05-15
[ROCM] GPU performance cost model specs for MI300

Ruturaj4:
@ddunl  @xla-rotation  : would you take a look please?

PR #12550, opened 2024-05-16
[ROCM] Avoid hard coded specifications for mi300 transpose fusion

Ruturaj4:


akuegel (approved):

akuegel (approved):
Just one question: the PR description says we can avoid some hardcoded specifications, but it seems they are not removed in this PR? Is this planned in a followup PR, or do I misunderstand something?

xla/service/gpu/fusions/transpose.cc
```diff
@@ -77,6 +78,18 @@ Tiling ComputeTransposeTiling(const TransposeDescription& tiled_transpose) {
   absl::InlinedVector<int64_t, 4> num_threads{1, 1, WarpSize()};
```
i-chaochen, line 78: since `gpu_device_info.threads_per_warp()` is already there, could you replace this `WarpSize()` as well?

i-chaochen:
Thanks for the review! Yes, there are more changes on tiling code as well.

akuegel:
> Thanks for the review! Yes, there are more changes on tiling code as well.

Ok, I think I understand what was meant: before, it did always choose one specific tiling config. Now, it chooses a different one for mi300, but it still chooses the other hardcoded values in other cases. Somehow I thought i would see some hardcoded values being removed, but I guess we still need it in other cases.

i-chaochen:
Oh I guess you're refering this one? https://github.com/openxla/xla/pull/12494

PR #12552, opened 2024-05-16
Enable vectorization support for transpose mlir emitter

ghost:
This PR aims to enable vectorization support for transpose mlir emitter. Below is the transpose kernel performance data with or without vectorization:

Transpose 0-2-1:

bf16 [8, 6144, 4096] -> bf16 [8, 4096, 6144]     673us->634us    improve 6%
s8     [8, 6144, 4096] -> s8    [8, 4096, 6144]     567us->497us    improve 13%         

bf16 [128, 1024, 1024] -> bf16 [128, 1024, 1024]     444us->401us   improve 10%
s8     [128, 1024, 1024] -> s8    [128, 1024, 1024]     378us->327us   improve 13%

Transpose 2-1-0:

bf16 [1024, 1024, 128] -> bf16 [128, 1024, 1024]     643us->471us   improve 26%
s8     [1024, 1024, 128] -> s8    [128, 1024, 1024]     481us->349us   improve 27%

bf16 [8192, 128, 128] -> bf16 [128, 128, 8192]     534us->467us    improve 13%
s8     [8192, 128, 128] -> s8     [128, 128, 8192]     443us->333us   improve 25%

All the above data are collected on A100 40GB machine.

### DIFF
--- a/xla/service/gpu/fusions/transpose_mlir.cc
+++ b/xla/service/gpu/fusions/transpose_mlir.cc
@@ -109,7 +109,7 @@ int GetVectorSize(const HloFusionAnalysis& analysis,
         input_dims[2] % (WarpSize() * vector_size) == 0;
     // Enable vectorization iff the below conditions are met:
     // (1) The minor dim of pre transpose and post transpose should be
-    // divisable by warp_size * vector_size. (2) Device occupancy is high.
+    // divisible by warp_size * vector_size. (2) Device occupancy is high.
     if (is_aligned && threads_counts(vector_size) >= threads_max) {
       break;
     }
--- a/xla/service/gpu/fusions/transpose_mlir.cc
+++ b/xla/service/gpu/fusions/transpose_mlir.cc
@@ -119,7 +119,8 @@ int GetVectorSize(const HloFusionAnalysis& analysis,
   return vector_size;
 }
 
-Tiling ComputeTransposeTiling(const TransposeDescription& tiled_transpose) {
+Tiling ComputeTransposeTiling(const HloFusionAnalysis& analysis) {
+  const auto& tiled_transpose = analysis.tiled_transpose();
   constexpr int kNumRows = 4;
   static_assert(WarpSize() % kNumRows == 0);
 
@@ -157,7 +158,7 @@ Tiling ComputeTransposeTiling(const TransposeDescription& tiled_transpose) {
 
 MlirTransposeFusion::MlirTransposeFusion(const HloFusionAnalysis& analysis)
     : analysis_(analysis),
-      tiling_(ComputeTransposeTiling(analysis.tiled_transpose())) {
+      tiling_(ComputeTransposeTiling(analysis)) {
   ConstHloInstructionSet transposes_to_tile;
   int index = 0;
   for (auto [root, hero] :
--- a/xla/service/gpu/fusions/transpose_mlir.cc
+++ b/xla/service/gpu/fusions/transpose_mlir.cc
@@ -107,7 +107,7 @@ Tiling ComputeTransposeTiling(const HloFusionAnalysis& analysis) {
                         analysis.device_info().threads_per_core_limit();
   int64_t non_vectorized_threads =
       non_vectorized_tiling.GetNumThreadsPerBlock() *
-      non_vectorized_tiling.GetNumThreadsPerBlock();
+      non_vectorized_tiling.GetNumBlocks();
   bool is_aligned =
       input_dims[permutation[2]] % (WarpSize() * vector_size) == 0 &&
       input_dims[2] % (WarpSize() * vector_size) == 0;
--- a/xla/service/gpu/fusions/transpose_mlir.cc
+++ b/xla/service/gpu/fusions/transpose_mlir.cc
@@ -78,47 +78,6 @@ using mlir_converter::PartitionedComputation;
 constexpr int kPreTransposeMinorDim = 2;
 constexpr int kPreTransposeVectorizedDim = 3;
 
-int GetVectorSize(const HloFusionAnalysis& analysis,
-                  absl::InlinedVector<int64_t, 4> input_dims,
-                  Vector3 permutation, int64_t threads_per_block) {
-  int vector_size = 2;
-  int64_t threads_max = analysis.device_info().core_count() *
-                        analysis.device_info().threads_per_core_limit();
-
-  auto threads_counts = [&](int vec_size) {
-    int64_t blocks = 1;
-    for (int i = 0; i < permutation.size(); ++i) {
-      if (i == permutation.size() - 1) {
-        blocks *=
-            CeilOfRatio(input_dims[permutation[i]], WarpSize() * vec_size);
-        continue;
-      }
-      if (permutation[i] == input_dims.size() - 1) {
-        blocks *=
-            CeilOfRatio(input_dims[permutation[i]], WarpSize() * vec_size);
-        continue;
-      }
-      blocks *= input_dims[permutation[i]];
-    }
-    return blocks * threads_per_block;
-  };
-
-  while (vector_size != 1) {
-    bool is_aligned =
-        input_dims[permutation[2]] % (WarpSize() * vector_size) == 0 &&
-        input_dims[2] % (WarpSize() * vector_size) == 0;
-    // Enable vectorization iff the below conditions are met:
-    // (1) The minor dim of pre transpose and post transpose should be
-    // divisible by warp_size * vector_size. (2) Device occupancy is high.
-    if (is_aligned && threads_counts(vector_size) >= threads_max) {
-      break;
-    }
-    vector_size = (vector_size + 1) / 2;
-  }
-
-  return vector_size;
-}
-
 Tiling ComputeTransposeTiling(const HloFusionAnalysis& analysis) {
   const auto& tiled_transpose = analysis.tiled_transpose();
   constexpr int kNumRows = 4;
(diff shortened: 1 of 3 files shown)


xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -75,7 +75,52 @@ using mlir_converter::ApplyAffineMap;
...
+        blocks *=
+            CeilOfRatio(input_dims[permutation[i]], WarpSize() * vec_size);
+        continue;
+      }
+      blocks *= input_dims[permutation[i]];
+    }
+    return blocks * threads_per_block;
+  };
+
+  while (vector_size != 1) {
```
jreiffers, line 106: This is a bit confusing. Looks like this loop will run at most once?

```
// Enable vectorization iff the below conditions are met:
// (1) The minor dim of pre transpose and post transpose should be
// divisible by warp_size * vector_size. (2) Device occupancy is high.
if (is_aligned && thread_counts(2) >= threads_max)) {
  return 2;
}
return 1; 
```

Also, since vec_size is now a constant in the lambda, we can just make it a real constant.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -75,7 +75,52 @@ using mlir_converter::ApplyAffineMap;
...
+int GetVectorSize(const HloFusionAnalysis& analysis,
+                  absl::InlinedVector<int64_t, 4> input_dims,
+                  Vector3 permutation, int64_t threads_per_block) {
+  int vector_size = 2;
+  int64_t threads_max = analysis.device_info().core_count() *
+                        analysis.device_info().threads_per_core_limit();
+
+  auto threads_counts = [&](int vec_size) {
+    int64_t blocks = 1;
+    for (int i = 0; i < permutation.size(); ++i) {
```
jreiffers, line 90: This loop is very confusing. Why not just create a `Tiling` and ask for the actual thread counts?

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -113,14 +167,35 @@ MlirTransposeFusion::MlirTransposeFusion(const HloFusionAnalysis& analysis)
...
     ++index;
   }
   shmem_transposes_ = {transposes_to_tile.begin(), transposes_to_tile.end()};
+  const auto& tiled_transpose = analysis.tiled_transpose();
+  absl::c_copy(tiled_transpose.permutation, std::back_inserter(permutation_));
+  permutation_.insert(absl::c_find(permutation_, kPreTransposeMinorDim) + 1,
+                      kPreTransposeVectorizedDim);
+}
+
+// Return bitcast mapping bewteen block_tile and permuted_block_tile if
```
jreiffers, line 183: Nit: typo.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -113,14 +167,35 @@ MlirTransposeFusion::MlirTransposeFusion(const HloFusionAnalysis& analysis)
...
+// `is_reverse` is false Otherwise return the reverse mapping.
+IndexingMap GetPermuteBlockTileBitcastMap(
+    const Tiling& tiling, const absl::InlinedVector<int64_t, 4>& permutation,
+    MLIRContext* mlir_context, bool is_reverse = false) {
```
jreiffers, line 187: It's not very clear what "reverse" means. Consider giving this a more descriptive name, or creating two functions instead of one.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -91,11 +136,20 @@ Tiling ComputeTransposeTiling(const TransposeDescription& tiled_transpose) {
                                              transposed_dims[permutation[1]],
                                              transposed_dims[permutation[2]]};
 
+  int vector_size =
+      GetVectorSize(analysis, input_dims, permutation, kNumRows * WarpSize());
+
   // We tile along the minor dimensions pre- and post-transpose.
   absl::InlinedVector<int64_t, 4> tile_sizes{1, 1, 1};
-  tile_sizes[permutation[2]] = WarpSize() / kNumRows;
+  tile_sizes[permutation[2]] = WarpSize() * vector_size / kNumRows;
```
jreiffers, line 144: This doesn't look right. In the vectorized case you end up with:

```
tile_sizes {1, 16, 1, 2}
```

In the unvectorized case, you get

```
tile_sizes {1, 8, 1, 1}
```

So you actually do `vector_size^2` amount of work per thread.

xla/service/gpu/fusions/transpose_mlir.h
```diff
@@ -94,7 +94,7 @@ class MlirTransposeFusion : public MlirFusionEmitterBase {
  private:
   const HloFusionAnalysis& analysis_;
   Tiling tiling_;
-  Vector3 permutation_;
+  absl::InlinedVector<int64_t, 4> permutation_;
```
jreiffers, line 97: The vectorized dimension most likely shouldn't show up here. It's only relevant in tiling, not in this bitcast representation of the transpose. Why was this necessary to add?

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -131,14 +206,16 @@ std::optional<IndexingMap> MlirTransposeFusion::ComputeThreadIdToOutputIndexing(
...
+      GetPermuteBlockTileBitcastMap(tiling_, permutation_, mlir_context);
   auto permuted_tiled_shape =
       ShapeUtil::MakeShape(U8, Permute(tiling_.GetShape(), permutation_));
 
   auto map = ComposeIndexingMaps(
       GetIndexingMapForTiling(
-          block_offset, thread_offset, tiling_.GetNumThreadsPerBlock(),
-          tiling_.GetNumBlocks(), tiling_.GetThreadTileSize(),
-          permuted_tiled_shape.dimensions()),
+          block_offset, bitcast_indexing.GetAffineMap().compose(thread_offset),
```
jreiffers, line 216: This doesn't look quite right too me. Only the blocks should be permuted, the read/write pattern inside a block should be the same. Or am I missing something?

jreiffers (changes requested):
Thanks!

xla/service/gpu/fusions/transpose_mlir.cc
lingzhi98: See the below example:
If threads = [1, 4, 1, 32] and tile_sizes = [1, 16, 1, 2], we can see 4 * 16 == 32 * 2. As you mentioned, pre-transpose threads is same as post-transpose threads. If we set tile_sizes as [1, 8, 1, 2], each warp write 8 * 4 elements to memory, which is not same as 2 * 32.

xla/service/gpu/fusions/transpose_mlir.cc
lingzhi98: We use [128, 256, 512]->[128, 512, 256] as example. Due to we add vectorized dim as the last dim of shape, so the input and output shape of transpose are [128, 256, 256, 2] - > [128, 256, 2, 256]. If block is permuted, the range of block offset are [[0, 127], [0, 255], [0, 1], [0, 255]]. Due to thread offset is not permuted, so the range of it are [[0, 0], [0, 63], [0,0], [0, 63], [0, 1]]. You can see block offset and thread offset is not aligned. So we should bitcast thread offsets to what we really want.

xla/service/gpu/fusions/transpose_mlir.h
lingzhi98: Vectorized dim is added to tiling's shape, so we need to add it to permutation so that get correct permuted block offsets. I think the below example can help you understand what I mean.

xla/service/gpu/fusions/transpose_mlir.cc
lingzhi98: We use [128, 256, 512]->[128, 512, 256] as example. Due to we add vectorized dim as the last dim of shape, so the input and output shape of transpose are [128, 256, 256, 2] - > [128, 256, 2, 256]. If block is permuted, the range of block offset are [[0, 127], [0, 255], [0, 1], [0, 255]]. Due to thread offset is not permuted, so the range of it are [[0, 0], [0, 63], [0,0], [0, 63], [0, 1]]. You can see block offset and thread offset is not aligned. So we should bitcast thread offsets to what we really want.

xla/service/gpu/fusions/transpose_mlir.h
jreiffers: Please add the extra dimension when applying the permutation. There is no fundamental connection between the permutation and vectorization.

lingzhi98: I think it is reasonable to change permutation here. As you see, shape of tiling has been added extra dim, so permutation need extend to same rank. If we add the extra dimension when applying the permutation, we need add too much code because permutation is used in many places. If you worry about it conflicts with previous definition, can I just change variable name "permutation" to "tiling_permutation" ?

xla/service/gpu/fusions/transpose_mlir.cc
jreiffers: Hmm, but why move the vector dim in the first place?

jreiffers: Yes, of course it's caused by the permutation. My question as more about why the permutation set up to move the vector dimension in the first place. I don't really see the benefit, and it seems to make things a lot more complicated?

lingzhi98: I can not get correct block offsets without permutation (or you have another idea?). Block offsets [a, b, c, vec_dim] corresponds to [a, c, vec_dim, b].

jreiffers: Just to make sure we're talking about the same thing, let's look at a toy example with 4x4 blocks:

If we have a 4x4 block starting at [a, b, c] (ignoring the vector dimension because its offset is always 0/size always 1):

```
0 1 2 3
4 5 6 7
8 9 a b
c d e f
```

With vectorization, thread 0 will read 0 and 1 from global memory and write them to shared memory, and it will read 0 and 4 from shared memory and write it to the indices corresponding to 0 and 1 in block `[a, c, b]`. So I believe for the block offsets what you want is `[a, c, b, vec_dim]`, where `vec_dim` is irrelevant (again, because it's always 0).

lingzhi98: We can use more simple example to show what is the correct block offset:
x0 y0 z0 w0
x1 y1 z1 w1
x2 y2 z2 w2
Assume a = b = c = 0, x0 y0 belong to block0, z0 w0 belong to block1, x1 y1 belong to block2, z1 w1 belong to block3. The per transpose offset of block3 is [0, 1, 1, 0]. If [a, c, b, vec_dim] is correct, the post transpose offset of block3 is [0, 1, 1, 0]. It's obvious that the index of z1 is same as block3 offset. Assume a in range [0, 0], b in range [0, 2], c in range [0, 1], vec_dim in range [0, 1], so the linear index of z1 is 1 * 3 * 2 + 2 = 8. But the linear index of z1 should be 7 as the below example show:
x0 x1 x2
y0 y1 y2
z0 z1 z2
w0 w1 w2

jreiffers: I'm sorry, but I don't follow these explanations. Let me try to implement this myself, maybe I'll see the problem then.

lingzhi98: You can see what I say above. I use your conclusion, but get wrong block offset result. https://github.com/openxla/xla/pull/12552#discussion_r1603577281

jreiffers: https://github.com/openxla/xla/pull/12620 should work? Maybe I'm still missing something.

jreiffers: But I don't really like that either. The block offset logic is really weird. Hmm.

lingzhi98: Now I understand what you mean. If has this line https://github.com/openxla/xla/pull/12620/files#diff-0210cd152294b1475469517ffba6b5ee578acc4e786b0dceeee6854aba445a29R148, the block offset is correct. But I think your PR do the same thing as me, so your main problem now is to make the logic simpler and easier to understand, right?

jreiffers: I think mine is already a bit simpler, WDYT. I won't have time to look into replacing Tiling today, but once we do that, it'll be good, I believe. If you agree, would you mind merging the two PRs? Yours has the vector size logic and updated tests.

lingzhi98: Yeah, It seems your logic only change block offset computation but not thread offset computation.  You can merge them If your PR does the rest. Or you want me to merge ?

jreiffers: Gotcha, I changed the other implementation to always use 128 threads as well. The `Tiling` abstraction was the main source of complexity here, so I removed that. I took your tests and most of your heuristic, but added a guard against exceeding the shmem budget and enabled vectorization for block sizes not divisible by 64 (unless there was a good reason not to do that?) Still need to run this against a larger test suite, but feel free to take a look already (still https://github.com/openxla/xla/pull/12620).

lingzhi98: #12620 seems fine. I have no other advice. I also wanted to abandon tiling at first, but I am not sure if openxla will accept this seemingly big change. Now it seems like I can be a little bolder in my subsequent PRs.  Thank you for your patient reply.

lingzhi98: The last thing: Maybe we can enable vectorization only for <32 bit datatype. I observed performance regression for some f32 shape.

PR #12579, opened 2024-05-16
[ROCm]  Enable new mlir fusion emitters only for CUDA

hsharsha:


hsharsha:
@xla-rotation @ddunl  Please do the needful.

xla/service/gpu/fusions/fusions.cc
```diff
@@ -145,12 +145,12 @@ absl::StatusOr<std::unique_ptr<FusionInterface>> GetFusionEmitter(
     if (!opts.xla_gpu_enable_mlir_emitters()) {
       return false;
     }
-    CHECK(!check ||
```
ddunl, line 148: It seems like the check argument to the lambda is now unused after this change?

hsharsha: Addressed review comment in https://github.com/openxla/xla/pull/12579/commits/437ddd8670b45fbfdb9bda892388d2189ca91ffe

jreiffers:
Sorry, what's  the purpose of this?

hsharsha:
> Edited: ah, ok, this is about rocm. This is not the right way to do it - please just leave the flag set to false if you're running ROCM. You can also remove the check in IsHloConversionSupported if you have time to contribute a bit more. Unfortunately I don't have any rocm hardware, so I can't test this myself.

IsHloConversionSupported needs more work for ROCm and I feel it is better if we do it in a separate commit.

xla/service/gpu/fusions/fusions.cc
```diff
@@ -142,15 +142,17 @@ absl::StatusOr<std::unique_ptr<FusionInterface>> GetFusionEmitter(
                          ->config()
                          .debug_options();
   auto check_mlir_emitters = [&](bool check = true) {
+    if (!check) return false;
```
jreiffers, line 145: This doesn't match the previous logic - check was suppose to only skip the CHECK failure below (because there's no elemental lowering for scatter).

xla/service/gpu/fusions/fusions.cc
```diff
@@ -142,15 +142,17 @@ absl::StatusOr<std::unique_ptr<FusionInterface>> GetFusionEmitter(
...
     if (!opts.xla_gpu_enable_mlir_emitters()) {
       return false;
     }
-    CHECK(!check ||
-          mlir_converter::IsHloConversionSupported(
-              analysis.fusion(),
-              fusion_info.analysis().device_info().gpu_compute_capability()))
-        << "Unsupported fusion: "
-        << analysis.fusion_root(0).instruction().parent()->ToString();
+    if (!mlir_converter::IsHloConversionSupported(analysis.fusion(),
```
jreiffers, line 150: Let's add a separate flag like `xla_gpu_enable_mlir_emitters`, but to allow the fallback to legacy emitters. `xla_gpu_enable_mlir_emitter_fallback`?

jreiffers:
> IsHloConversionSupported needs more work for ROCm and I feel it is better if we do it in a separate commit.

Alright, let's land a version of this for now.

jreiffers (changes requested):

xla/service/gpu/fusions/fusions.cc
```diff
@@ -141,6 +141,9 @@ absl::StatusOr<std::unique_ptr<FusionInterface>> GetFusionEmitter(
                          .GetModule()
                          ->config()
                          .debug_options();
+#ifdef TENSORFLOW_USE_ROCM
```
jreiffers, line 144: Why do we need an ifdef for this?

hsharsha: mlir_converter::IsHloConversionSupported always returns false for ROCm. So for ROCm side we are yet to support mlir emitters.  So I added this until we support mlir emitters as tracked in https://github.com/ROCm/frameworks-internal/issues/8063

jreiffers: Why not just do it properly right away? See my other comment re `xla_gpu_enable_mlir_emitter_fallback`.

hsharsha: Does `xla_gpu_check_mlir_emitter_support` add better clarification?

hsharsha:
@jreiffers I have modified test `xla/service/gpu/fusions/mlir_emitter_test_base.cc` to set `xla_gpu_enable_mlir_emitters` only for NVIDIA

penpornk:
@jreiffers Could you please help take another look?

hsharsha:
@penpornk  this is stale and the API's have changed. I am working on it in https://github.com/openxla/xla/pull/13663 and it requires MLIR dialect lowering for ROCm.

penpornk:
@hsharsha I see. Thank you for the heads up!

PR #12845, opened 2024-05-20
[XLA:GPU] add workspace rewrite for FP8 gemm

shawnwang18:
Add  workspace rewrite for FP8 with amax fusion.

### DIFF
--- a/xla/service/gpu/backend_configs.proto
+++ b/xla/service/gpu/backend_configs.proto
@@ -104,6 +104,7 @@ message GemmBackendConfig {
 
   optional bool grad_x = 16;
   optional bool grad_y = 17;
+  bool damax_output = 18;
 }
 
 // Backend config for bitcast operation generated from MLIR MHLO dialect.
--- a/xla/service/gpu/gemm_rewriter.cc
+++ b/xla/service/gpu/gemm_rewriter.cc
@@ -1391,6 +1391,12 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
     HloInstruction *gemm_and_damax =
         instr->AddInstruction(existing_gemm->CloneWithNewShape(tuple_shape));
 
+    TF_ASSIGN_OR_RETURN(auto gpu_config,
+                        gemm_and_damax->backend_config<GpuBackendConfig>());
+    GemmBackendConfig &config = *gpu_config.mutable_gemm_backend_config();
+    config.set_damax_output(true);
+    TF_RETURN_IF_ERROR(gemm_and_damax->set_backend_config(gpu_config));
+
     // Obtain D and DAmax separately from the output tuple.
     HloInstruction *d =
         instr->AddInstruction(HloInstruction::CreateGetTupleElement(
@@ -2262,9 +2268,10 @@ class GemmWorkspaceRewriteVisitor : public DfsHloRewriteVisitor {
           has_aux_output,
           xla::gpu::gpublas_lt::EpilogueHasAuxiliaryOutput(epilogue));
 
-      if (!((has_aux_output && instr->shape().IsTuple() &&
-             instr->shape().tuple_shapes_size() == 2) ||
-            (!has_aux_output && instr->shape().IsArray()))) {
+      if (!((instr->shape().IsTuple() &&
+             instr->shape().tuple_shapes_size() ==
+                 has_aux_output + config.damax_output() + 1) ||
+            instr->shape().IsArray())) {
         return absl::OkStatus();
       }
     } else if (instr->custom_call_target() != kGemmCallTarget ||
@@ -2314,7 +2321,7 @@ class GemmWorkspaceRewriteVisitor : public DfsHloRewriteVisitor {
       custom_call->set_output_to_operand_aliasing({{{0}, {2, {}}}});
     }
 
-    if (has_aux_output) {
+    if (instr->shape().IsTuple()) {
       for (auto user : instr->users()) {
         auto user_get_tuple =
             dynamic_cast<HloGetTupleElementInstruction *>(user);
--- a/xla/service/gpu/ir_emitter_unnested.cc
+++ b/xla/service/gpu/ir_emitter_unnested.cc
@@ -55,7 +55,7 @@ limitations under the License.
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Linker/Linker.h"
-#include "mlir/AsmParser/AsmParser.h"  // from @llvm-project
+#include "mlir/AsmParser/AsmParser.h" // from @llvm-project
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"  // from @llvm-project
 #include "mlir/Dialect/MemRef/Transforms/Passes.h"  // from @llvm-project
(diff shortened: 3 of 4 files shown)


xla/service/gpu/gemm_rewriter.cc
```diff
@@ -2314,7 +2325,7 @@ class GemmWorkspaceRewriteVisitor : public DfsHloRewriteVisitor {
       custom_call->set_output_to_operand_aliasing({{{0}, {2, {}}}});
     }
 
-    if (has_aux_output) {
+    if (instr->shape().IsTuple()) {
       for (auto user : instr->users()) {
```
reedwm, line 2329: Can you replace this entire for-loop with `ReplaceInstr(instr, new_call)`? That should cause all the GetTupleElement instructions to refer to new_call insteade of instr.

I just noticed this but it's unrelated to the fix. So if this is replaced by `ReplaceInstr(instr, new_call)`, it should be done in a separate PR

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -55,17 +55,17 @@ limitations under the License.
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Linker/Linker.h"
-#include "mlir/AsmParser/AsmParser.h"  // from @llvm-project
-#include "mlir/Dialect/LLVMIR/LLVMDialect.h"  // from @llvm-project
+#include "mlir/AsmParser/AsmParser.h"               // from @llvm-project
```
reedwm, line 58: Why have this unnecessary formatting? Did an autoformatter do this?

reedwm (changes requested):

xla/service/gpu/backend_configs.proto
```diff
@@ -104,6 +104,7 @@ message GemmBackendConfig {
 
   optional bool grad_x = 16;
   optional bool grad_y = 17;
+  optional bool has_damax_output = 18;
```
philipphack, line 107: Why is this an optional bool? Do we need three different states?

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -2261,10 +2268,14 @@ class GemmWorkspaceRewriteVisitor : public DfsHloRewriteVisitor {
       TF_ASSIGN_OR_RETURN(
           has_aux_output,
           xla::gpu::gpublas_lt::EpilogueHasAuxiliaryOutput(epilogue));
+      if (config.has_has_damax_output()) {
```
philipphack, line 2271: See above. I suggest making this a non-optional bool or renaming it to damax_output.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -758,9 +758,8 @@ absl::Status IrEmitterUnnested::EmitCublasLtMatmulThunkF8(
 
   TF_ASSIGN_OR_RETURN(bool has_vector_bias,
                       xla::gpu::gpublas_lt::EpilogueAddsVectorBias(epilogue));
-  bool has_damax = instr->shape().IsTuple();
   xla::ShapeIndex output_index =
-      has_damax ? xla::ShapeIndex{0} : xla::ShapeIndex{};
+      instr->shape().IsTuple() ? xla::ShapeIndex{0} : xla::ShapeIndex{};
```
philipphack, line 762: Are there situations where an FP8 custom call does not return a tuple?

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -823,9 +823,7 @@ absl::Status IrEmitterUnnested::EmitCublasLtMatmulThunkF8(
   TF_RET_CHECK(!has_aux_output);
   std::optional<BufferAllocation::Slice> workspace_buffer;
   if (instr->shape().IsTuple() &&
-      (instr->shape().tuple_shapes_size() - has_aux_output - 1)) {
-    TF_RET_CHECK((has_aux_output && instr->shape().tuple_shapes_size() == 3) ||
-                 (!has_aux_output && instr->shape().tuple_shapes_size() == 2));
+      (instr->shape().tuple_shapes_size() - has_damax == 2)) {
```
philipphack, line 826: IIUC, the call always returns a workspace. Do we need these conditionals?

philipphack (changes requested):

xla/service/gpu/backend_configs.proto
shawnwang18: Without optional, it will require protobuf to initialize the has_damax_output to false. though protobuf does it, but I think have an optional state should make the code more clear.

xla/service/gpu/gemm_rewriter.cc
shawnwang18: Yeah, let's do it in a separate PR.

xla/service/gpu/ir_emitter_unnested.cc
shawnwang18: I fixed the code that assume the output shape is tuple.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -2261,10 +2268,14 @@ class GemmWorkspaceRewriteVisitor : public DfsHloRewriteVisitor {
       TF_ASSIGN_OR_RETURN(
           has_aux_output,
           xla::gpu::gpublas_lt::EpilogueHasAuxiliaryOutput(epilogue));
+      if (config.has_damax_output()) {
+        has_damax_output = config.damax_output();
```
reedwm, line 2272: No need to check if the config has damax_output, you can just set `has_damax_output` uncondtionally. 

I don't think you even need a `has_damax_output` local variable. You can just use `config.damax_output()` in the if-statement below

reedwm (changes requested):

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -822,10 +822,7 @@ absl::Status IrEmitterUnnested::EmitCublasLtMatmulThunkF8(
   BufferAllocation::Slice aux;  // Not used.
   TF_RET_CHECK(!has_aux_output);
   std::optional<BufferAllocation::Slice> workspace_buffer;
-  if (instr->shape().IsTuple() &&
-      (instr->shape().tuple_shapes_size() - has_aux_output - 1)) {
-    TF_RET_CHECK((has_aux_output && instr->shape().tuple_shapes_size() == 3) ||
-                 (!has_aux_output && instr->shape().tuple_shapes_size() == 2));
+  if (instr->shape().tuple_shapes_size() - config.damax_output() == 2) {
```
philipphack, line 825: Is there a known case where this is false? If not, this is probably not needed.

shawnwang18: I don't know a case, but the workspace rewrite has a guard like this: 

      if (!((instr->shape().IsTuple() &&
             instr->shape().tuple_shapes_size() ==
                 has_aux_output + config.damax_output() + 1) ||
            instr->shape().IsArray())) {
       return absl::OkStatus();
}

If gemm_rewrite happens to create a new case where tuple includes other outputs, then it is an unsupported case.

reedwm (approved):

PR #12866, opened 2024-05-20
[NVIDIA] Move collective pipeliner after post-layout opts

kaixih:
To perform the fp8 dot, we insert quant and dequant ops before the dot and rely on the gemm-rewriter to pattern match and rewrite to cublaslt calls for GPUs.

When this fp8 pattern appears in while loops that is qualified for the collective pipeliner (CP), the CP might disrupts the pattern by accidentally peeling the quant or dequant ops off the loop.

So, this PR moves the CP after the gemm-rewriter (when the newly-introduced `xla_gpu_collective_pipeliner_post_layout_optimization` is set), since this can save us from redesigning the CP to support this special case. 

Note, for the long-term solution, we should move the collectives to the "safer" places before applying the CP. For example, in the above use case, we can move the collectives in-between the quant and dequant and then the CP should be able to preserve the dequant->dot pattern. This is actually already supported as an integral part in the gemm-rewriter.

cc. @Tixxx @hx89 @nluehr @philipphack


kaixih:
@frgossen Can you help review?

frgossen (changes requested):
Thanks for explaining the motivation. From the description, it is not quite clear why this is needed. Can you expand the motivation, maybe with an example? What fp8 pattern appears in a while loop and how is it transformed vs should be transformed?  

Can you add test cases with the pattern variations you care about? I think that could also help understand the motivation behind this change.

kaixih:
Sure. Basically, the original graph looks like
```
while->q->dq->AG->dot
```
where `q` takes bf16 and outputs fp8 and `dq` takes fp8 and outputs bf16.
Then our gemm rewriter moves the AG and pattern matches dq->dot to cublaslt_fp8gemm, like:
```
# move AG:
while->q->AG->dq->dot
# pattern match:
while->q->AG->custom_call(=cublaslt-fp8)
```

Now, with the CP pass, we see that the original graph is transformed into:
```
q->dq->AG->while->dot 
```
where you can see the `q->dq->AG` is moved out of the loop. Then, the gemm-rewriter can no longer rewrites this dot to the custom call of cublaslt-fp8.

So, that is why this PR moves the CP after the gemm rewriter.

Please let me know if the above makes sense to you. At the same time, I am trying to extract the above from our real model training as a unit test to showcase the issue.

kaixih:
@frgossen Any update?

We have recently tested this in some real models and we can see the fp8 gemms are back.

bixia1:
Do you know why moving CP to post layout is not generally good?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1239,6 +1261,8 @@ absl::Status GpuCompiler::OptimizeHloModule(
       hlo_module, stream_exec, options, gpu_target_config,
       thread_pool.get_mutable()));
 
+  TF_RETURN_IF_ERROR(RunCollectivePipelinerPasses(hlo_module));
```
bixia1, line 1264: It is better to rename and make it clear this is post layout CP: RunPostLayoutCollectivePipelinerPasses

xla/service/gpu/gpu_compiler.cc
```diff
@@ -884,6 +868,44 @@ absl::Status RunCollectiveOptimizationPasses(
         /*reuse_pipelined_op_buffer=*/HloPredicateFalse};
     collectives_pipeline.AddPass<CollectivePipeliner>(config);
   }
+  return absl::OkStatus();
```
bixia1, line 871: line 924 AddP2PPipeliner(collectives_pipeline);
also invokes CP and can potentially cause the same problem, and is not protected by you flag. 
Can you also include this?

Tixxx: Would the p2p pipeliner have any dependency on other pipeliner passes? if not i'd suggest to not move it for now, the purpose for this is to not break the QDQ patterns of fp8 quantization, i'm not sure if that breakage will ever happen with pipeline parallelism since we likely wont shard the graph in the middle of a QDQ.

kaixih:
> Do you know why moving CP to post layout is not generally good?

I think theoretically it should be fine to move CP to post layout, since the CP is supposed to be agnostic of non-collective ops. The main concern (from @Tixxx ) might be some passes in layout (e.g. layout normalization) might alter the collectives through inserting copies or transposes, for instance. We are not 100% confident that this would pose a problem or not.

So, our plan is to keep the flag and set it to `True` by default. If no complaints are received, we will remove the flag. How do you think of the plan?

xla/service/gpu/gpu_compiler.cc
kaixih: Done. PTAL.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -884,6 +868,44 @@ absl::Status RunCollectiveOptimizationPasses(
...
+}
+
+absl::Status RunPostLayoutCollectivePipelinerPasses(
+    HloModule* hlo_module) {
+  const DebugOptions& debug_options = hlo_module->config().debug_options();
+  HloPassPipeline collectives_pipeline("collective-pipeliner-optimizations");
+  if (debug_options.xla_gpu_run_post_layout_collective_pipeliner()) {
+    TF_RETURN_IF_ERROR(
+        AddCollectivePipelinerPasses(debug_options, collectives_pipeline));
+    collectives_pipeline.AddPass<WhileLoopTripCountAnnotator>();
```
frgossen, line 881: Why does `WhileLoopTripCountAnnotator` appear here and not in the other case?

frgossen (changes requested):

xla/service/gpu/gpu_compiler.cc
Tixxx: We also call WhileLoopTripCountAnnotator at the end of the collective pipeline, i think any pass that changes the loop trip count would have to call this after it.

frgossen (approved):
Thanks!

frgossen:
Can you address the build failures?

kaixih:
@frgossen Rebased. PTAL.

(Only AMD ROCm is failing but the log implies it is irrelevant to this PR).

frgossen:
Looks like this runs into a merge conflict. Can you rebase?

kaixih:
@frgossen Rebased. PTAL.

frgossen:
Can you also add an end to end test since you are changing the default behaviour?

kaixih:
Do you mean a full model test? Or a unit test?

We encountered the original issue in a complex model. Not sure if it is simple to extract the pattern to a unit test.

@Tixxx Do you think any of your previous unit test can be used as a template to do the above test?

kaixih:
> Can you also add an end to end test since you are changing the default behaviour?

Do you have pointers how do we usually do the test on pass orders (like pass X should after pass Y)?

Tixxx:
> Do you mean a full model test? Or a unit test?
> 
> We encountered the original issue in a complex model. Not sure if it is simple to extract the pattern to a unit test.
> 
> @Tixxx Do you think any of your previous unit test can be used as a template to do the above test?

There's a test suite called collective_ops_test_e2e where we add this type of e2e tests. It runs the entire gpu compiler pipeliner which is what we want. I think in your case, you might want to take one of the unit test hlos and modify that to reflect the pattern in this issue and add that to collective_ops_test_e2e suite. You can toggle the debug option on and off. With this option on, module should have an fp8 gemm, no fp8 gemm otherwise

kaixih:
The unit test is added. [frgossen](https://github.com/frgossen) PTAL.

frgossen (changes requested):
LGTM! We're hitting another merge conflict though. Can you rebase?

kaixih:
@frgossen Done. PTAL.

frgossen (approved):
Thanskj!

frgossen:
Posting here, too. This needs another rebase, which I cannot easily do on my end :/

frgossen:
Sorry about this but it is hitting a merge conflict still.

kaixih:
@frgossen should I close this one, since my colleague helped rebasing it and created a new PR in [here](https://github.com/openxla/xla/pull/13920)?

frgossen:
Yes, if you prefer to move forward with 13920, please close this

kaixih:
Closing. Further discussion should be in 13920.

PR #12912, opened 2024-05-21
Overload operator<< for XLA_FFI_DataType

andportnoy:


xla/ffi/api/api.h
```diff
@@ -1313,7 +1313,8 @@ inline std::ostream& operator<<(std::ostream& os, const XLA_FFI_AttrType type) {
       auto* scalar = reinterpret_cast<XLA_FFI_Scalar*>(attr);         \
       if (XLA_FFI_PREDICT_FALSE(scalar->dtype != TYPE)) {             \
         return diagnostic.Emit("Wrong scalar data type: expected ")   \
-               << TYPE << " but got " << scalar->dtype;               \
+               << XLA_FFI_Get_DataType_Name(TYPE) << " but got "      \
```
ezhulenev, line 1316: Can you add `operator<<` overload to api.h and forward https://github.com/openxla/xla/blob/main/xla/ffi/api/ffi.h#L65-L102 to it instead, just to have only one ToString conversion

xla/ffi/api/c_api.h
```diff
@@ -170,6 +170,31 @@ typedef enum {
 } XLA_FFI_DataType;
 // LINT.ThenChange(ffi_test.cc)
 
+#define _(x) case XLA_FFI_DataType_##x: return #x;
```
ezhulenev, line 173: At a high level I want to keep c_api.h as only structs and define functions that work with these structs in:

1. ffi/api/api.h - if they can be shared between external/internal ffi implenentations
2. ffi/api/ffi.h - for external ffi
3. ffi/ffi.h - for internal ffi

Mainly because you can use "normal" C++

xla/ffi/api/api.h
andportnoy: Why forward btw and not just move the overload from `ffi.h` to `api.h`?

xla/ffi/api/c_api.h
andportnoy: Makes sense, I added this function in `c_api.h` because I was thinking the users of the C API would benefit from a function like that too since the source of truth enum is defined at the C level. Though users can build their own.
- `c_api.h` <- data type source of truth enum is here
- `api.h` (includes `c_api.h`) <- currently no ToString here
- `ffi.h` (includes `c_api.h` and `api.h`) <- only here

xla/ffi/api/api.h
andportnoy: Nevermind, the class DataType is only defined in `ffi.h`.

ezhulenev: Yes, it's an `xla::PrimitiveType` re-implementation for user that can't depend on xla

xla/ffi/api/c_api.h
ezhulenev: The idea is that end users should not really see (and care) about C API, XLA aims to provide a c++ wrapper that hides all the details and provides a nice header-only c++17 API, just like PJRT, we have C API monstrosity but it's wrapped into c++ classes that are easier to work with

andportnoy: Hmm, then why does the C API exist?

ezhulenev: It's a layer that defines ABI and allows compiling xla/jaxlib/custom calls with different compilers and different C++ versions, it's nearly impossible to safely pass C++ structs/classes between shared libraries.

xla/ffi/api/api.h
andportnoy: >add operator<< overload to api.h and forward

Done, please take a look.

ezhulenev (approved):

andportnoy:
@ezhulenev not sure what's wrong with AMD CI, it seems to be failing for a reason unrelated to the contents of this PR.

ddunl:
It's broken on HEAD right now, I should have a fix in shortly

i-chaochen:
> @ezhulenev not sure what's wrong with AMD CI, it seems to be failing for a reason unrelated to the contents of this PR.

Yes, one commits bypass the check and caused the issue.

https://github.com/openxla/xla/commit/9500b988b369f12fd1cc874bbf352de431ceeca9#diff-d47af227fffaf809a9fcac1c89d8850db4d281fbfd7df042d4bb08ee09377ad8R2430-R2433

ezhulenev (approved):

PR #12915, opened 2024-05-21
Handle unsigned MLIR integers during FFI attribute conversion

andportnoy:


ezhulenev (approved):

pparuzel:
Has cl/632222737 been somehow reverted? This code has a bug which the CL does not introduce where it calls `getInt` for an unsigned type but this should call `getUInt`.

Error produced:
```
int64_t mlir::IntegerAttr::getInt() const: (getType().isIndex() || getType().isSignlessInteger()) && "must be signless integer"
```

andportnoy:
@pparuzel thanks, strange that I didn't run into this. I tested this locally and I build JAX in debug mode, so I would have expected to have run into this [assert](https://github.com/llvm/llvm-project/blob/cc3b6c3ba9c0dd7df0fd7ac23c8609c4675dd62a/mlir/lib/IR/BuiltinAttributes.cpp#L353-L354).

Unless a fix is already in flight, I'll send a PR to use `getUInt()`.

pparuzel:
JAX does not yet have FFI changes yet so this path isn't tested through JAX.

I'm currently working on that google/jax#21162 here and so my tests started failing, that's how I found out. I prepared the fix at #12955

andportnoy:
> JAX does not yet have FFI changes so this path wouldn't be tested through JAX.

I ran into the lack of unsigned integer conversion while doing an [end-to-end custom call integration example](https://github.com/google/jax/pull/21363) and my tests _didn't_ fail, that's why I'm surprised I didn't see this assert :)

I assume my LLVM was somehow built with asserts disabled.

andportnoy:
I left a [comment](https://github.com/openxla/xla/commit/bbb202a01dc5d5533fab1a3a2f1319307d27427c#r142293091) in https://github.com/openxla/xla/pull/12955, I'm curious why you are removing conversion for single bit integers.

pparuzel:
> > JAX does not yet have FFI changes so this path wouldn't be tested through JAX.
> 
> I ran into the lack of unsigned integer conversion while doing an [end-to-end custom call integration example](https://github.com/google/jax/pull/21363) and my tests _didn't_ fail, that's why I'm surprised I didn't see this assert :)
> 
> I assume my LLVM was somehow built with asserts disabled.

Oh that's interesting then. You're probably right it could be "asserts disabled".

> I left a [comment](https://github.com/openxla/xla/commit/bbb202a01dc5d5533fab1a3a2f1319307d27427c#r142293091) in #12955, I'm curious why you are removing conversion for single bit integers.

That's because I thought we interpret `i1` as `mlir::BooleanAttr`. Let me look into this, good point.

PR #12940, opened 2024-05-22
[ROCm] Fix dot_bf16.hlo.test on ROCm

mmakevic-amd:
Added additional params for `hlo_lit_tests` as a workaround, so `mi200.txtpb` would be used in `dot_bf16.hlo.test` for rocm.

### DIFF
--- a/xla/service/gpu/tests/BUILD
+++ b/xla/service/gpu/tests/BUILD
@@ -807,9 +807,15 @@ lit_test_suite(
     args = if_cuda_is_configured([
         "--param=PTX=PTX",
         "--param=GPU=a6000",
+        "--param=GPU1=v100",
+        "--param=SM=CHECK-SM70",
+        "--param=GPU2=a100_80",
     ]) + if_rocm_is_configured([
         "--param=PTX=GCN",
         "--param=GPU=mi200",
+        "--param=GPU1=mi200",
+        "--param=GPU2=mi200",
+        "--param=SM=CHECK-SM80",
     ]),
     cfg = "//xla:lit.cfg.py",
     data = [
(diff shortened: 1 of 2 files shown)


akuegel (changes requested):
I think a better fix would be to add a feature condition to the runline. I have an example for ptx tests:

RUN: %if ptxas && !ptxas-12.0 %{ llc < %s -march=nvptx -mcpu=sm_20 | %ptxas-verify %}

So as you can see, the syntax is `%if feature %{ <command> %}`

The feature can be passed as "args" parameter to lit_test build macro.
So in the if_cuda_is_configured block of the args parameter, you could add "sm70", and then change the RUN line to: 

RUN: %if sm70 %{ hlo-opt %s --platform=gpu --stage=hlo --xla_gpu_target_config_filename=%S/../../../tools/hlo_opt/gpu_specs/v100.txtpb --split-input-file | FileCheck %s --check-prefixes=CHECK-SM70 %}

I think this is better than to run the same test twice on ROCM

mmakevic-amd:
@akuegel Thanks for the suggestion, I've switched to using feature conditions on RUN lines.

xla/service/gpu/tests/BUILD
```diff
@@ -807,15 +807,12 @@ lit_test_suite(
     args = if_cuda_is_configured([
         "--param=PTX=PTX",
         "--param=GPU=a6000",
-        "--param=GPU1=v100",
-        "--param=SM=CHECK-SM70",
-        "--param=GPU2=a100_80",
+        "--param=sm70",
```
akuegel, line 810: features need to be given without "--param=" prefix. Right now, the %if condition is never triggered, as the sm70/sm80 is not recognized as a feature.
So just adding "sm70", "sm80" should work.

akuegel (changes requested):

xla/service/gpu/tests/BUILD
akuegel: Ok, seems it is actually not that easy, it also needs other changes. Currently this would result in lit: error: unrecognized arguments: sm70
I think we need to let lit know that the feature sm70 exists. I am currently trying it out myself what needs to be done.

akuegel: Ok, seems it actually needs to be passed via --param=sm70 like you did. But what is missing is lines like this:

config.available_features.add("sm70")

These need to be added to xla/lit.cfg.py

akuegel: So just summarizing, everything in your changes is already ok, just the changes to xla/lit.cfg.py need to be added (for each feature, add a line `config.available_features.add(...)`

mmakevic-amd: > config.available_features.add("sm70")
>These need to be added to xla/lit.cfg.py

Hm, just adding these features will make if statement true, regardless of what is in the BUILD file. We'd need something like this in  lit.cfg.py:

```
if using-cuda:
   config.available_features.add("sm70")
   config.available_features.add("sm80")
if using-rocm:
  config.available_features.add("if_rocm")
```

And I'm not sure how I could get target device here

akuegel: Ok, then I guess unfortunately this feature stuff doesn't work. Sad :-(
Then I guess we need to go back to your original idea.

mmakevic-amd: Reverted everything

akuegel: Thanks, and sorry for the extra work. I hoped it was possible to fix this in a better way.

akuegel (approved):

PR #12964, opened 2024-05-22
[XLA:CPU][oneDNN] Block elementwise linear fusion for non-scalar multipliers

akhilgoe:
This PR addresses a potential bug by prohibiting Elementwise Linear operation for non-scalar multipliers. Additionally, the PR includes a test to ensure that the rewrite does not occur in such scenarios.

### DIFF
--- a/xla/service/cpu/onednn_matmul_rewriter.cc
+++ b/xla/service/cpu/onednn_matmul_rewriter.cc
@@ -731,7 +731,10 @@ class OneDnnMatMulRewriteVisitor : public DfsHloRewriteVisitor {
 
     if (Match(instr, pattern)) {
       std::vector<HloInstruction*> new_operands;
-      auto constant_value = *GetConstantValueAsFloat32(constant);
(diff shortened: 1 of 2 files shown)


penpornk (approved):
Thank you for the PR! Do you plan to support non-scalar constants in the futures? oneDNN binary post-ops support non-scalar operand, right?

akhilgoe:
> Thank you for the PR! Do you plan to support non-scalar constants in the futures? oneDNN binary post-ops support non-scalar operand, right?

Hi @penpornk, thank you for your review! Yes, oneDNN supports non-scalar multipliers for binary fusions. Currently, we see the elementwise linear case only in the case of Scaled Dot Product, so we have only enabled it for scalar multipliers. However, if you think that non-scalar mul fusion would benefit any models/use-cases then we can take an Action Item to enable it.

penpornk:
@akhilgoe Thank you for the quick reply! I don't have a use case for now. Just asking out of curiosity. :)

PR #12975, opened 2024-05-23
[XLA:CPU][oneDNN] Fix failures in onednn_matmul_tests

kanvi-nervana:
This PR fixes failures seen in onednn_matmul_tests


penpornk (commented):
Thank you for the PR! Could you please explain what the failures are and how the fix help? Does matching BF16 operand make the accuracy drop?

kanvi-nervana:
> Thank you for the PR! Could you please explain what the failures are and how the fix help? Does matching BF16 operand make the accuracy drop?

The failure is due to pattern mismatch because of which the fusion does not happen. This happened because 2 PRs went in a different order and there was no merge conflict. It fixes this [test](https://github.com/openxla/xla/blob/6ff6bd59a20fdabaf450134d6b449348a7a3c2ed/xla/tests/onednn_matmul_test.cc#L1288)
The change in threshold was missed in the original [PR](https://github.com/openxla/xla/pull/12500) and is not related to BF16 operand change.

penpornk (approved):
Thank you for the context!

PR #13018, opened 2024-05-23
[ROCm] Distinguish between AMD and NVIDIA GPUs with relevant tags

hsharsha:
Add `requires-gpu-amd` for AMD gpus and unit tests that require AMD backend. Use `requires-gpu-nvidia` for any nvidia gpu backend.

### DIFF
--- a/xla/tests/build_defs.bzl
+++ b/xla/tests/build_defs.bzl
@@ -27,7 +27,7 @@ AMD_GPU_DEFAULT_BACKENDS = ["amd_gpu_any"]
 
 _DEFAULT_BACKENDS = ["cpu"] + NVIDIA_GPU_DEFAULT_BACKENDS + AMD_GPU_DEFAULT_BACKENDS
 
-ALL_GPU_BACKENDS = NVIDIA_GPU_BACKENDS + AMD_GPU_DEFAULT_BACKENDS
+GPU_BACKENDS = NVIDIA_GPU_BACKENDS + AMD_GPU_DEFAULT_BACKENDS
 
 _ALL_BACKENDS = ["cpu", "interpreter"] + NVIDIA_GPU_BACKENDS + AMD_GPU_DEFAULT_BACKENDS + list(plugins.keys())
 
@@ -318,7 +318,7 @@ def xla_test_library(
         this_backend_copts = []
         if backend == "cpu":
             backend_deps = ["//xla/tests:test_macros_cpu"]
-        elif backend in ALL_GPU_BACKENDS:
+        elif backend in GPU_BACKENDS:
             backend_deps = ["//xla/tests:test_macros_%s" % backend]
         elif backend == "interpreter":
             backend_deps = ["//xla/tests:test_macros_interpreter"]
--- a/xla/tests/build_defs.bzl
+++ b/xla/tests/build_defs.bzl
@@ -5,14 +5,10 @@ load(
     "tf_gpu_tests_tags",
 )
 load("//xla:xla.bzl", "xla_cc_test")
-load(
-    "//xla/stream_executor:build_defs.bzl",
-    "if_gpu_is_configured",
-)
 load("//xla/tests:plugin.bzl", "plugins")
 
 # Possible backend values for the GPU family.
-GPU_BACKENDS = [
+NVIDIA_GPU_BACKENDS = [
     "gpu_any",
     "gpu_p100",
     "gpu_v100",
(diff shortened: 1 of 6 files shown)


hsharsha:
@ddunl  Let me know if this can be done in a better way?

hsharsha:
```
Executed 0 out of 957 tests: 957 tests pass.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
All tests passed but there were other errors during the build.
```
How can I restart  XLA Linux CPU test as all the tests are passing

xla/tests/build_defs.bzl
```diff
@@ -21,35 +17,39 @@ GPU_BACKENDS = [
...
+NVIDIA_GPU_DEFAULT_BACKENDS = [
     "gpu_any",
     "gpu_a100",
     "gpu_h100",
 ]
 
-_DEFAULT_BACKENDS = ["cpu"] + GPU_DEFAULT_BACKENDS
+AMD_GPU_DEFAULT_BACKENDS = ["amd_gpu_any"]
+
+_NVIDIA_DEFAULT_BACKENDS = ["cpu"] + NVIDIA_GPU_DEFAULT_BACKENDS
```
beckerhe, line 28: I think we should keep this as `_DEFAULT_BACKENDS` since it also contains `cpu` and it's just the "default" which happens to include the NVIDIA targets.

xla/tests/build_defs.bzl
```diff
@@ -168,12 +168,13 @@ def xla_test(
...
     if not backends:
-        backends = _DEFAULT_BACKENDS
+        backends = _NVIDIA_DEFAULT_BACKENDS
 
     # Expand "gpu" backend name to specific GPU backends and update tags.
     backends, disabled_backends, backend_tags, backend_args = \
-        prepare_gpu_backend_data(backends, disabled_backends, backend_tags, backend_args)
+        prepare_nvidia_gpu_backend_data(backends, disabled_backends, backend_tags, backend_args)
 
+    backends += AMD_GPU_DEFAULT_BACKENDS
```
beckerhe, line 177: This is adding the AMD default backends in all cases - which means the user can't overwrite it anymore. I think you rather have to add them in line 171.

xla/tests/build_defs.bzl
```diff
@@ -21,35 +17,39 @@ GPU_BACKENDS = [
...
     gpu_backend_tags = backend_tags.get("gpu", [])
-    for key in GPU_BACKENDS:
+    for key in NVIDIA_GPU_BACKENDS:
         new_backend_tags.setdefault(key, gpu_backend_tags[:])
 
     new_backend_args = {key: value for key, value in backend_args.items() if key != "gpu"}
     if "gpu" in backend_args:
-        for key in GPU_BACKENDS:
+        for key in NVIDIA_GPU_BACKENDS:
             new_backend_args.setdefault(key, backend_args["gpu"])
```
beckerhe, line 53: I believe this part of this function should also apply to amdgpu backends. Would you mind putting this into a separate function (or rather put the remaining part below into a separate function - whatever works best)?

xla/xla.bzl
```diff
@@ -71,10 +71,62 @@ def xla_cc_binary(deps = [], copts = tsl_copts(), **kwargs):
...
+        # Notably, a test_suite with `tests = []` is not empty:
+        # https://bazel.build/reference/be/general#test_suite_args and the default
+        # `tests = []` behavior doesn't respect `--build_tag_filters` due to
+        # b/317293391. For this reason, if we would create an empty `test_suite`,
+        # instead create a `cc_test` with no srcs that links against `main` to have
+        # more predictable behavior that avoids bugs.
+        if test_names:
+            native.test_suite(name = name, tags = kwargs['tags'], tests = test_names)
+        else:
+            native.cc_test(name = name, deps = ["@tsl//tsl/platform:test_main"])
```
beckerhe, line 125: This seems to be replicating a lot of logic that we also have in `xla_test`. I'm wondering if we could remove the `use_gpu` flag from `xla_cc_test` entirely and make all the users of this flag use the `xla_test` macro instead. Is that something you have looked at by chance?

xla/tests/build_defs.bzl
hsharsha: I need to do this after the call to prepare_nvidia_gpu_backend_data. Otherwise I need to compare for amd backend and continue.  I could not find a cleaner solution for this.

xla/xla.bzl
hsharsha: I thought about it but that is almost 550+ occurrence of  `xla_cc_test`. It would be better to do it as a separate cleanup commit to ease to reviewing process.

hsharsha: 70 of them with use_gpu = True mostly in 3 or 4 directories, which can be added with this change.

beckerhe: Hey, I went ahead and did the change in #13092. Once that has landed you should be able to rebase (hopefully rather easily).

beckerhe: The change just went in.

beckerhe (approved):

beckerhe:
Looks good to me.

We still need to see though whether the CI agrees. Not sure why it didn't start yet.

xla/tests/build_defs.bzl
```diff
@@ -21,35 +17,39 @@ GPU_BACKENDS = [
...
     "gpu_a100",
     "gpu_h100",
 ]
 
-_DEFAULT_BACKENDS = ["cpu"] + GPU_DEFAULT_BACKENDS
+AMD_GPU_DEFAULT_BACKENDS = ["amd_gpu_any"]
+
+_DEFAULT_BACKENDS = ["cpu"] + NVIDIA_GPU_DEFAULT_BACKENDS + AMD_GPU_DEFAULT_BACKENDS
+
+ALL_GPU_BACKENDS = NVIDIA_GPU_BACKENDS + AMD_GPU_DEFAULT_BACKENDS
```
beckerhe, line 30: Can you rename this variable to `GPU_BACKENDS`? We have some internal build rule that is trying to load a variable called `GPU_BACKENDS` from this file.

beckerhe:
I'm approving this to get a fresh version of the PR imported into our internal system. But the ROCm build still needs to be fixed before we can merge this.

beckerhe (approved):

beckerhe:
Hey, I'm trying to get this imported, but I will need to make some changes on top to make it work with our internal infra.

The biggest one: I have to rename `amd_gpu_any` to `gpu_amd_any` because we have some assumption that the string before the first underscore identifies the hardware class.

I hope that doesn't cause any issues.

PR #13065, opened 2024-05-24
[NVIDIA] Disable the bias reuse for fp8 dot

kaixih:
For some operations, we might want to do the `output to operand aliasing` to reuse the memory space. For example, in the dot operation, we can alias the output to the bias operand, when they have the same shape.

However, this aliasing is not always safe especially when the dot outputs fp8 outputs. In this case the bias is 2x larger than expected output and the HLO verifier will complain.

This PR fixes this issue by disabling the `output to operand aliasing`  for fp8 dot.

cc. @philipphack @hx89 @nluehr

### DIFF
--- a/xla/service/gpu/gemm_rewriter.cc
+++ b/xla/service/gpu/gemm_rewriter.cc
@@ -1377,8 +1377,9 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
 
     // The F8ConvertD may change the output dtype. We need to turn off the
     // output to operand aliasing when it happens.
-    const auto& maybe_bias = existing_gemm->operand(2);
-    if (!ShapeUtil::Equal(maybe_bias->shape(), new_gemm->shape())) {
+    if (gemm_backend_config.beta() != 0.0 &&
+        !ShapeUtil::Equal(existing_gemm->operand(2)->shape(),
+                          new_gemm->shape())) {
       xla::Cast<HloCustomCallInstruction>(new_gemm.get())
           ->set_output_to_operand_aliasing({});
     }
--- a/xla/service/gpu/gemm_rewriter.cc
+++ b/xla/service/gpu/gemm_rewriter.cc
@@ -1374,6 +1374,15 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
 
     std::unique_ptr<HloInstruction> new_gemm =
         existing_gemm->CloneWithNewShape(instr->shape());
+
+    // The F8ConvertD may change the output dtype. We need to turn off the
+    // output to operand aliasing when it happens.
+    const auto& maybe_bias = existing_gemm->operand(2);
+    if (!ShapeUtil::Equal(maybe_bias->shape(), new_gemm->shape())) {
+      xla::Cast<HloCustomCallInstruction>(new_gemm.get())
+          ->set_output_to_operand_aliasing({});
+    }
+
     TF_RETURN_IF_ERROR(ReplaceWithNewInstruction(instr, std::move(new_gemm)));
 
     VLOG(1) << "Conversion" << (reduce_damax ? " and amax calculation" : "")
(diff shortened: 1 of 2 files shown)


xla/service/gpu/tests/gemm_rewrite_test.cc
```diff
@@ -4967,6 +4967,52 @@ TEST_P(ParameterizedFp8GemmRewriteTest, UnscaledABUnscaledDMatrixBiasF8) {
...
+)";
+
+  RunAndFilecheckHloRewrite(
+      hlo_text,
+      GemmRewriter(CudaHopperOrRocmMI300(), GetToolkitVersion(),
+                   /*f8_rewrite=*/true),
+      R"(
+; CHECK-LABEL: ENTRY %test ({{.*}}) -> <<F8E4M3>>[16,64] {
+; CHECK:           custom_call_target="__cublas$lt$matmul$f8",
+; CHECK-NOT:       output_to_operand_aliasing
```
philipphack, line 5011: Do we need a new unit test? I think we could just add the check-not to e.g. ScaledABUnscaledDMatrixBiasF8.

kaixih: Emm, I think there are a couple of reasons. The `ScaledABUnscaledDMatrixBiasF8` doesn't do the `output to operand aliasing` in the first place. I think this is because the bias is the parameter of the xla function and it should be read-only. That is why I intentionally add  a constant one for the given bias in my unit test of above.
 
Also, even if we fix `ScaledABUnscaledDMatrixBiasF8` to show the `output to operand aliasing`, I am thinking it is actually a valid usage of `output to operand aliasing` and we shouldn't disable it  if we have a long-term solution. So, I created this test to better exhibit the behavior of "fp8 fused dot but should not use `output to operand aliasing`".

The long-term solution should be check the bias and output dtype size along with the shapes.

philipphack: Comparing the output and bias shapes wouldn't be difficult though. Could this be done here?

xla/service/gpu/tests/gemm_rewrite_test.cc
```diff
@@ -4967,6 +4967,52 @@ TEST_P(ParameterizedFp8GemmRewriteTest, UnscaledABUnscaledDMatrixBiasF8) {
       )");
 }
 
+// Do not reuse matrix bias for fp8 matmul output.
+TEST_P(ParameterizedFp8GemmRewriteTest, DoNotReuseBiasForF8Output) {
```
philipphack, line 4971: Align the name of this test with the other tests, e.g. toUnscaledABScaledDMatrixBiasNoAliasingF8 or similar.

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1526,7 +1526,8 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
     // true if those uses all come before this operation.  But copy-insertion
     // runs before scheduling, so it can't know and has to conservatively insert
     // copies.)
-    if (IsLegacyCublasMatmul(*fused_op) || can_overwrite_bias) {
+    if (!IsCublasLtMatmulF8(*fused_op) &&
```
philipphack, line 1529: Expand the comment above to include the case you're disabling aliasing for.

philipphack (changes requested):

reedwm (changes requested):

xla/service/gpu/gemm_rewriter.cc
```diff
@@ -1374,6 +1374,15 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
 
     std::unique_ptr<HloInstruction> new_gemm =
         existing_gemm->CloneWithNewShape(instr->shape());
+
+    // The F8ConvertD may change the output dtype. We need to turn off the
+    // output to operand aliasing when it happens.
+    const auto& maybe_bias = existing_gemm->operand(2);
```
philipphack, line 1380: Is there a good reason for using auto here? Since this is only used to access the shape, maybe have a Shape variable instead or remove it?

reedwm (approved):

kaixih:
@reedwm It seems one AMD test is failing but from the log, it complains the MLIR stuff and seems irrelevant to this PR. Can you help on merging?

PR #13106, opened 2024-05-28
Fix a bug in AsyncAllocator memory pools access setting.

buptzyb:
The original code uses the wrong parameter `i` for `cuDeviceCanAccessPeer()`, which can cause undefined behavior when `(*all_ids_)[i] != platform_device_id`, for example, virtual devices are used. Should use `(*all_ids_)[i].value()` to represent the previous pool id.

Also, add a judgment to skip the pool initialization process if it's already initialized previously. This can happen in some cases, such as when multiple virtual devices are created from one physical GPU, the virtual devices will actually share the same CUDA memory pool.

Same PR from TensorFlow repo: https://github.com/tensorflow/tensorflow/pull/67373.

### DIFF
--- a/xla/stream_executor/gpu/gpu_cudamallocasync_allocator.cc
+++ b/xla/stream_executor/gpu/gpu_cudamallocasync_allocator.cc
@@ -241,6 +241,22 @@ GpuCudaMallocAsyncAllocator::GpuCudaMallocAsyncAllocator(
   static auto* all_ids_ = new std::vector<tsl::PlatformDeviceId>();
   if (!create_new_pool_) {
     DCHECK(all_pools_->size() == all_ids_->size());
+
+    // If the pool_ is found in all_pools_, it means it has been initialized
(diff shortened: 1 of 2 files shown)


ezhulenev (approved):

akuegel (approved):
Approving again, the imported change has been garbage collected by now

akuegel:
@xla-rotation the internal version of this change has been garbage collected again. I didn't check what was blocking the merge.

penpornk:
@buptzyb Apologies for the delay! This branch now has conflicts. Could you please help resolve them? We'll look into merging the PR again soon after.

buptzyb:
@penpornk Hi, I resolved the conflicts. Thanks!

akuegel:
Can you please also write a test?

buptzyb:
> Can you please also write a test?

Add a test that creates two allocators on one GPU and lets them share the default pool. Without this fix the test would fail.

akuegel (approved):
Thank you

PR #13109, opened 2024-05-28
Add flag for termination on nccl error

jaro-sevcik:
This introduces a flag for termination on NCCL async error. With the flag on, XLA will terminate the process on NCCL error. With the flag off, the existing behavior should remain unchanged.

The patch is motivated by several problems:

- Without this patch, the heartbeat monitor only checks communicators that are currently not use by the running executable (because it obtains the communicators with TryAcquire). Since NCCL errors cause a hang in the running communicator, most failing communicators are locked, so their async errors just go undetected. As a result, XLA often hangs until Grpc timeout even in cases when ncclCommGetAsyncError would report an error.

- Ideally we would recover by aborting the faulty communicators, but that seems to be unreliable (aborts can cause hangs if NCCL currently hangs on a different communicator than the one being aborted). NCCL team is aware of this and working on a fix (https://github.com/NVIDIA/nccl/issues/1013). At the moment, there does not seem to be a reliable fast recovery mechanism short of process termination.

We propose to expose a flag for terminating the process on failure so that there is some way to detect and recover from a NCCL failure. Once the comm-abort works reliably, we will use that and propagate the error to the API user.

The patch is based on a PoC from pshamis@nvidia.com and vincentz@nvidia.com.

### DIFF
--- a/xla/service/gpu/runtime/nccl_clique.cc
+++ b/xla/service/gpu/runtime/nccl_clique.cc
@@ -98,6 +98,10 @@ static absl::Duration TerminateTimeout() {
                                   : absl::InfiniteDuration();
 }
 
+static bool TerminateOnNcclError() {
+  return xla::GetDebugOptionsFromFlags().xla_gpu_nccl_terminate_on_error();
+}
+
 //===----------------------------------------------------------------------===//
 // NcclClique
 //===----------------------------------------------------------------------===//
@@ -180,6 +184,13 @@ static absl::Status CheckComm(NcclApi::NcclCommHandle comm) {
 // Runs async check on all communicators in a clique.
 static void CheckClique(const NcclCliqueKey& clique_key,
                         NcclClique& lockable_clique) {
+  if (TerminateOnNcclError()) {
+    absl::Status status = lockable_clique.CheckAsyncErrors();
+    if (!status.ok()) {
+      LOG(FATAL) << "Terminating process due to async NCCL error: " << status;
+    }
+    return;
+  }
   if (NcclClique::Lock clique = lockable_clique.TryAcquire()) {
     VLOG(5) << "Checking NCCL clique " << clique_key.ToString()
             << " for async errors; num_communicators="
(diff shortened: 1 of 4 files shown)


NaiyerRizz:
@jaro-sevcik its showing that the branch has some conflicts, Can you please resolve those.

xla/service/gpu/runtime/nccl_clique.h
```diff
@@ -112,18 +125,29 @@ struct NcclCliqueName {
...
+  // error checks, the constructor intentionally leaks the reference
+  // to the communicators from an acquired lock.
   NcclClique(NcclCliqueKey clique_key, std::optional<NcclCliqueId> clique_id,
              absl::btree_map<int32_t, NcclApi::OwnedNcclComm> communicators)
-      : Lockable(std::move(clique_key), clique_id, std::move(communicators)) {}
+      : Lockable(std::move(clique_key), clique_id, std::move(communicators)),
+        async_error_checker_(Acquire()->GetChecker()) {}
 
   std::string DebugString() const;
+  absl::Status Check();
```
golechwierowicz, line 147: Can you also briefly describe the contract for this method?

jaro-sevcik: Done. Also renamed the method.

golechwierowicz (approved):

jaro-sevcik (commented):
Could you take another look, please?

xla/service/gpu/runtime/nccl_clique.cc
```diff
@@ -180,6 +184,13 @@ static absl::Status CheckComm(NcclApi::NcclCommHandle comm) {
 // Runs async check on all communicators in a clique.
 static void CheckClique(const NcclCliqueKey& clique_key,
                         NcclClique& lockable_clique) {
+  if (TerminateOnNcclError()) {
+    absl::Status status = lockable_clique.CheckAsyncErrors();
+    if (!status.ok()) {
+      LOG(FATAL) << "Terminating process due to async NCCL error: " << status;
```
cheshire, line 190: Can we propagate Status instead?

cheshire (changes requested):
What would it take to enable this by default?

xla/service/gpu/runtime/nccl_clique.cc
jaro-sevcik: Do you mean propagating to the user (such as JAX)?

If yes, then we have such effort in the works, lead by @Tixxx . This PR is a workaround until proper propagation is in place. As things currently stand, you will have to kill the process anyway because ncclCommAbort can hang, so there is no reliable recovery.

jaro-sevcik (commented):
> What would it take to enable this by default?

That depends on what behavior you are expecting.

In my opinion, a good default behavior is that the error bubbles up to the user (such as the Python/JAX program), and the user is able to recover - perhaps by calling distributed.shutdown and distributed.initialize in JAX.

Currently such a recovery is not quite possible until shutting down NCCL communicators gets more reliable.

cheshire (approved):

shamisp:
> > What would it take to enable this by default?
> 
> That depends on what behavior you are expecting.
> 
> In my opinion, a good default behavior is that the error bubbles up to the user (such as the Python/JAX program), and the user is able to recover - perhaps by calling distributed.shutdown and distributed.initialize in JAX.
> 
> Currently such a recovery is not quite possible until shutting down NCCL communicators gets more reliable.

Agreed. I also want to point out that original code flow (before the patch) is semantically incorrect. ncclCommGetAsyncError() implementation does not guarantee error propagation, which means not all ranks will observe an error code. In the same time ncclCommAbort is a collective call that all ranks must call otherwise it may hangs. Since the call  invoked from a branch, it is assumes error propagation through ncclCommGetAsyncError().  I would argue that the new behavior introduced in this patch is preferred over default.

akuegel (approved):

PR #13125, opened 2024-05-28
Make UNKNOWN the default XLA FFI error code

andportnoy:
Default error is defined as !success(), where success is communicated using XLA_FFI_Error_Code_OK with an integer value of 0. The value of !success() then is 1, corresponding to "CANCELLED" before this commit.

This commit swaps the values of CANCELLED and UNKNOWN since UNKNOWN is a better default/generic error type.


ezhulenev (commented):
We rely on enum casting to/from absl::StatusCode and keep it consistent here (https://github.com/abseil/abseil-cpp/blob/master/absl/status/status.h#L99-L268).

Which reminds me that this have to be tested!

andportnoy:
Ah got it, closing. Looks like I also misunderstood a method of the `Error` class so this PR is invalid anyway.

PR #13165, opened 2024-05-29
 [ROCm] Update ROCm version to 6.0.2

hsharsha:


hsharsha:
@ddunl  Do you think this would be a proper place to house amd ci scripts?

ddunl:
I think for now it could be best to leave them as is. I know that Intel is contributing some scripts as well, and I'm also moving around the XLA ones. I bet after all that happens we'll coalesce on a new location but I think for now where they are is fine. Unless there's a more specific motivation for the move in this PR

hsharsha:
@ddunl  Noted. I have kept the script as is but updated ROCm version as most of our containers are running 6.0.2

ddunl (approved):

PR #13190, opened 2024-05-29
Add pipelined while loop annotator

shraiysh:
This patch recognises pipelined while loop using the rotate right pattern.
It recognises a rotate-right pattern on sharded inputs and labels the
surrounding while loop as a pipelined while loop. This is an unsafe
optimization. This is hidden behind a debug flag so it wont be triggered in
TPU pipelines or other GPU pipelines unexpectedly.


xla/debug_options_flags.cc
```diff
@@ -985,6 +987,17 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
...
+      "xla_gpu_unsafe_pipelined_while_loop_annotator_rotate_right_pattern",
+      bool_setter_for(
+          &DebugOptions::
+              set_xla_gpu_unsafe_pipelined_while_loop_annotator_rotate_right_pattern),
+      debug_options
+          ->xla_gpu_unsafe_pipelined_while_loop_annotator_rotate_right_pattern(),
+      "If this option is true, then the while loop with rotate right "
+      "pattern will be considered a pipelined while loop and the "
+      "operations within the pipeline bubbles may be considered no-ops. This "
+      "is an unsafe flag."));
```
frgossen, line 1000: Can you mention that this changes the semantics of collective permute?

xla/service/spmd/spmd_partitioner_test.cc
```diff
@@ -2568,6 +2575,59 @@ ENTRY entry {
...
+                           /*unsafe_pipeline_attr=*/true));
+  const HloInstruction* whileOp =
+      module->entry_computation()->root_instruction();
+  const auto root =
+      whileOp->while_body()->GetInstructionWithName("concatenate");
+  auto rotate =
+      op::Concatenate(op::CollectivePermute(op::Slice()), op::Slice());
+  VLOG(0) << module->ToString();
+  EXPECT_THAT(root, AllOf(rotate, op::Shape("f32[3]")));
+  EXPECT_TRUE(
```
frgossen, line 2627: Can you also test that the same IR remains correct in case you run `PartitionComputation` with unsafe_pipeline_attr=false?

frgossen (changes requested):

xla/service/spmd/pipelined_while_loop_annotator.cc
```diff
@@ -0,0 +1,119 @@
...
+#include "xla/service/spmd/spmd_partitioner.h"
+#include "xla/service/spmd/spmd_partitioner_util.h"
+
+namespace xla {
+
+namespace {
+
+// This function is a duplicate from
+// StatefulRngSpmdPartitioner::GetSpmdPartitionerOptions
+spmd::SpmdPartitionerOptions GetSpmdPartitionerOptions(
```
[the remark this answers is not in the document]
shraiysh: Can we resolve this? It is outdated.

xla/service/spmd/pipelined_while_loop_annotator.h
```diff
@@ -0,0 +1,52 @@
...
+// This pass annotates while loops with rotate right patterns as pipelined while
+// loops. On a high level, this pass replicates the same logic that SPMD
+// partitioner uses for recognizing a rotate right pattern and instead of going
+// ahead with the partitioning, it labels the surrounding while loop with the
+// attribute "is_pipelined_while_loop".
+//
+// This is an unsafe transformation. There are cases where a rotate right
+// pattern within a while loop will not be a pipelined while loop. This lets the
+// rest of the passes ignore the computations in the pipeline bubble.
+class PipelinedWhileLoopAnnotator : public spmd::SpmdPartitioner {
```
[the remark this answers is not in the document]
shraiysh: Can we resolve this too (again, outdated).

xla/service/spmd/spmd_partitioner_test.cc
shraiysh: Done.

xla/debug_options_flags.cc
shraiysh: Done.

frgossen (approved):
Thank you!

Tixxx (approved):

cheshire:
Is it possible to do this transform outside of spmd_partitioner, so that pipeline-related hacks are kept together?

shraiysh:
> Is it possible to do this transform outside of spmd_partitioner, so that pipeline-related hacks are kept together?
It would get messy. We duplicate entire code of [preprocessing sharding](https://github.com/openxla/xla/blob/e08cf1fa044e33ae848f15a74d506c496eadce80/xla/service/spmd/spmd_partitioner.cc#L5115C22-L5115C40) and with that too, we have to be careful that it is okay if this gets called twice. Or we would have to extract this out of SPMD partitioner into a separate pass and careful about it (which again would make this patch bigger). We would also need the same options that SPMD partitioner uses to do this. Another thing is that the logic within [PreprocessHlos](https://github.com/openxla/xla/blob/e08cf1fa044e33ae848f15a74d506c496eadce80/xla/service/spmd/spmd_partitioner.cc#L5285-L5365) would have to either be duplicated(fragile) or extracted(hard). These are the reasons why I figured it might be better to add this small change behind an option for spmd_partitioner.cc. I am open to easier and better solutions, if any.

xla/service/spmd/spmd_partitioner.cc
```diff
@@ -4059,12 +4059,13 @@ absl::Status SpmdPartitioningVisitor::HandleWhile(HloInstruction* hlo) {
...
-        MakePartitionedShape(hlo->shape(), sharding), hlo->while_condition(),
-        hlo->while_body(),
-        GetPartitionedHlo(hlo->operand(0)).Reshard(sharding).hlo()));
-  });
+
+  HloInstruction* whileOp = b_.AddInstruction(HloInstruction::CreateWhile(
+      MakePartitionedShape(hlo->shape(), sharding), hlo->while_condition(),
+      hlo->while_body(),
+      GetPartitionedHlo(hlo->operand(0)).Reshard(sharding).hlo()));
+  hlo->SetupDerivedInstruction(whileOp);
```
cheshire, line 4067: So the delta here is calling `->SetupDerivedInstruction`? Could we move that into a separate CL?

cheshire (changes requested):
small nit, looks good overall

xla/service/spmd/spmd_partitioner.cc
shraiysh: Sure. Let me push another change before this merges.

frgossen (approved):

shraiysh:
The JAX linux cpu fail seems unrelated. Is this good to merge @frgossen ?

penpornk:
@shraiysh Could you please help merge conflicts?

frgossen:
Needed another rebase and internal fixed. Changed the flag to `xla_gpu_unsafe_pipelined_loop_annotator` for linting reasons. Trying to land it now

shraiysh:
> @shraiysh Could you please help resolve conflicts?

Does this need anything to be done from my side? I can't see a conflict (I don't fully understand copybara so please let me know if anything needs attention from my end.)

PR #13224, opened 2024-05-30
[XLA:GPU] Adding an option that allocates temp buffer through separate memory allocator. 

shawnwang18:
This PR will assign XLA temp buffer to a dedicated memory space, and uses a standalone cuda async allocator to allocate temp buffer, as temp buffer is always allocated & deallocated for each step, this will ensure that temp buffer can be allocated to a stable address, which is good for cuda-graph's perf.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -148,6 +148,8 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_nccl_comm_splitting(false);
   opts.set_xla_gpu_enable_nccl_per_stream_comms(false);
 
+  opts.set_xla_gpu_temp_buffer_use_separate_color(false);
+
   // Set 4GB space limit for redzone scratch allocator.
   opts.set_xla_gpu_redzone_scratch_max_megabytes(1LL << 12);
   opts.set_xla_gpu_redzone_padding_bytes(8 * 1024 * 1024);
@@ -1274,6 +1276,14 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       "Enables NCCL User Buffer Registration. collective_memory_size in the "
       "allocator config must also be set to a non-zero value that is large "
       "enough to meet peak collective memory usage."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_temp_buffer_use_separate_color",
+      bool_setter_for(
+          &DebugOptions::set_xla_gpu_temp_buffer_use_separate_color),
+      debug_options->xla_gpu_temp_buffer_use_separate_color(),
+      "Enables temp User Buffer Registration. Enable this flag will use a "
+      "separate cuda async memory allocator to allocate temp buffer, this will "
+      "allocate temp buffer to the fixed address on every iteration"));
   flag_list->push_back(tsl::Flag(
       "xla_gpu_enable_nccl_comm_splitting",
       bool_setter_for(&DebugOptions::set_xla_gpu_enable_nccl_comm_splitting),
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -96,7 +96,9 @@ limitations under the License.
 #include "xla/pjrt/gpu/nccl_id_store.h"
 #include "xla/pjrt/stream_executor_executable.pb.h"
 #include "xla/service/gpu/gpu_compiler.h"
+#include "xla/service/gpu/gpu_memory_space_assignment.h"
 #include "xla/xla.pb.h"
+#include "xla/debug_options_flags.h"
 #endif  // GOOGLE_CUDA || TENSORFLOW_USE_ROCM
 
 #if GOOGLE_CUDA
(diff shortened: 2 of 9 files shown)


xla/debug_options_flags.cc
```diff
@@ -1232,6 +1234,13 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       "Enables NCCL User Buffer Registration. collective_memory_size in the "
       "allocator config must also be set to a non-zero value that is large "
       "enough to meet peak collective memory usage."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_enable_temp_async_buffer",
+      bool_setter_for(&DebugOptions::set_xla_gpu_enable_temp_async_buffer),
+      debug_options->xla_gpu_enable_temp_async_buffer(),
+      "Enables temp User Buffer Registration. Enable this flag will use a "
+      "separate cuda async memory allocator to allocate temp buffer, this will "
+      "allocate temp buffer to the fix address on every iteration"));
```
cheshire, line 1243: Nit: "fixed"

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -789,59 +790,55 @@ namespace {
...
 
-absl::StatusOr<std::vector<se::MultiDeviceAdapter::AllocatorInfo>>
-CreateCudaAsyncAllocator(
-    se::Platform* platform,
-    const std::map<int, std::unique_ptr<LocalDeviceState>>& addressable_devices,
-    double memory_fraction, bool preallocate) {
-  CHECK_GT(addressable_devices.size(), 0);
-  std::vector<se::MultiDeviceAdapter::AllocatorInfo> allocators;
+absl::StatusOr<std::unique_ptr<se::GpuCudaMallocAsyncAllocator>>
+CreateCudaAsyncAllocator(const LocalDeviceState& device, double memory_fraction,
```
cheshire, line 794: Nit: could we separate changes to async allocator into a separate PR?

xla/service/buffer_assignment.cc
```diff
@@ -251,9 +251,9 @@ void BufferAllocation::AddAssignment(const HloValue& buffer, int64_t offset,
   CHECK_LE(offset + size, size_)
       << "LogicalBuffer " << buffer
       << " size out of range at offset: " << offset << " with size: " << size;
-  CHECK_EQ(buffer.color(), color())
-      << "Buffer color " << buffer.color() << " for buffer " << buffer
-      << " does not match allocation color " << color() << ".";
+  // CHECK_EQ(buffer.color(), color())
```
cheshire, line 254: Should this be commented out?

xla/service/buffer_assignment.h
```diff
@@ -570,7 +573,8 @@ class BufferAssignment {
 
   // Combines allocations of temporary buffers into one big BufferAllocation.
   void CombineTempAllocations(
-      const absl::flat_hash_set<BufferValue::Color>& private_stack_colors);
+      const absl::flat_hash_set<BufferValue::Color>& private_stack_colors,
+      bool enable_temp_user_buffer);
```
cheshire, line 577: Is this a GPU-specific option? Should it go into a GPU subclass?

cheshire (changes requested):
Could you add a more high-level description to the PR + tests?

xla/xla.proto
```diff
@@ -786,7 +786,12 @@ message DebugOptions {
   // a deterministic implementation.
   bool xla_gpu_exclude_nondeterministic_ops = 297;
 
-  // Next id: 300
+  // Enables temp User Buffer Registration. Enable this flag will use a 
+  // separate cuda async memory allocator to allocate temp buffer, this will 
+  // allocate temp buffer to the fix address on every iteration.
+  bool xla_gpu_enable_temp_async_buffer = 301;
+
+  // Next id: 302
```
tdanyluk, line 794: I think that I roughly understand this but the wording is not really clear to me.
IIUC the goal is to allocate a temp buffer which stays the same on every iteration? Why is it important that the buffer is "async"? 

Shouldn't we call this flag something that emphasizes the stable nature of the buffer?

xla/service/buffer_assignment.h
```diff
@@ -43,6 +43,8 @@ limitations under the License.
 
 namespace xla {
 
+inline constexpr int64_t kPersistentTempMemorySpaceColor = 100;
```
tdanyluk, line 46: Should this be next to kInvalidColor?
Why was the value 100 chosen? Are we sure that it doesn't collide with other values?

tdanyluk (commented):
Thanks, could you please elaborate what is the use case for this?

xla/service/buffer_assignment.h
shawnwang18: No particular reason for chosing 100 as the number, just randomly select a value that is not collide with others. Yes, it should be better that XLA put all available ids in a centralized places, but nccl memory space id was hardcoded in the code. Can we make this as TODO that will be fixed in a separate PR?

xla/xla.proto
shawnwang18: I changed the flag name to xla_gpu_temp_buffer_use_cuda_async_allocator, it requires to use the `cudaAsyncAllocator` because it supports using separate memory pool for allocation, which will just ensure that the allocated address happens to be the stable.

xla/service/buffer_assignment.h
shawnwang18: Currently this feature is only valid for CUDA devices. but in theory it should also work for ROCM or other devices, as long as we use a separate memory allocator for temp buffer. BufferAssignment pass is a general optimization pass, so currently I made a patch here.

xla/service/buffer_assignment.cc
shawnwang18: For temp buffer with new memory space color, it will no longer be valid. I changed the assertion like below: 
  
  if (!(IsPreallocatedTempBuffer() &&
        color() == kPersistentTempMemorySpaceColor)) {
    CHECK_EQ(buffer.color(), color())
        << "Buffer color " << buffer.color() << " for buffer " << buffer
        << " does not match allocation color " << color() << ".";
  }

xla/pjrt/gpu/se_gpu_pjrt_client.cc
shawnwang18: Thanks, async allocator was merged in a separate PR.

xla/debug_options_flags.cc
shawnwang18: fixed

xla/service/buffer_assignment.cc
```diff
@@ -2108,7 +2124,12 @@ BufferAssigner::CreateAssignment(
   for (const auto& [color, computations] : private_stacks) {
     private_stack_colors.insert(color);
   }
-  assignment->CombineTempAllocations(private_stack_colors);
+  const bool temp_buffer_use_async_allocator =
```
ezhulenev, line 2127: This should not be here as well, this should be passed from GPU compiler via arguments

xla/service/buffer_assignment.h
```diff
@@ -43,6 +43,8 @@ limitations under the License.
 
 namespace xla {
 
+inline constexpr int64_t kTempBufferUsingCudaAsyncAllocatorColor = 100;
```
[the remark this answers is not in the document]
shawnwang18: Fixed

xla/service/buffer_assignment.cc
shawnwang18: Fixed

tdanyluk:
I'm withdrawing from this CL, deferring to ezhulenev and cheshire

ezhulenev (approved):

xla/service/gpu/gpu_memory_space_assignment.h
```diff
@@ -29,6 +29,7 @@ namespace xla {
 namespace gpu {
 
 inline constexpr int64_t kCollectiveMemorySpaceColor = 1;
+inline constexpr int64_t kTempBufferMemorySpaceColor = 5;
```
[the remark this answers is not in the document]
shawnwang18: 0,1,5 should have be taken. I fixed it with value 2. 

Maybe best to another PR that puts all the memory space color value in a single place.

xla/service/buffer_assignment.cc
```diff
@@ -700,6 +703,17 @@ void BufferAssignment::CombineTempAllocations(
             temp_allocation.peak_buffers_.begin(),
             temp_allocation.peak_buffers_.end());
       }
+
+      if (temp_buffer_color.has_value()) {
+        // If the temp allocation does not have special color requirement, and
+        // we assign it to a new color type indicating that it will be allocated
+        // with separate allocator to persistent allocation.
+        VLOG(3) << "Using async allocator for temp allocations, size "
```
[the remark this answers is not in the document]
shawnwang18: fixed

ezhulenev (approved):

penpornk (changes requested):
We have a new policy requiring test updates when adding new features or making any behavioral changes. Could you please help add a test for this?

ezhulenev (approved):

ezhulenev (approved):

reedwm:
This is failing the test `pjrt_c_api_gpu_test.cc` internally since GOOGLE_CUDA and CUDA_VERSION are not defined in `se_gpu_pjrt_client.cc`, causing the error `CUDA async allocator requires CUDA >= 11.2`.

I'm not sure why GOOGLE_CUDA and CUDA_VERSION are not defined, nor how to fix this. I'm asking internally to see if anyone else knows.

PR #13241, opened 2024-05-30
[GPU] Fail on conflicting autotuning results.

sergachev:


cheshire (changes requested):
Could we add a test?

sergachev:
Done. While adding a test I found out that most tests in autotuner_util_test are invalid because
1) it's not a GPU test in the build file, optimizing this HLO https://github.com/openxla/xla/blob/b919db1e0e77fae8d22aec8176890735e5135c0a/xla/service/gpu/autotuner_util_test.cc#L53 does not use the AutotunerUtil
2) this line https://github.com/openxla/xla/blob/b919db1e0e77fae8d22aec8176890735e5135c0a/xla/service/gpu/autotuner_util_test.cc#L75 is happy with empty result proto, which serializes to a non-empty string "version: 3".

Fix: https://github.com/openxla/xla/pull/13372.

cheshire (approved):

tdanyluk (approved):

PR #13244, opened 2024-05-30
[ROCm]  Get rid of .bazelrc file in run_xla script

hsharsha:


ddunl (approved):

ddunl:
Is this ready to merge? Don't want to jump the gun in case there's a CI config change on your end this should be gated on

hsharsha:
> Is this ready to merge? Don't want to jump the gun in case there's a CI config change on your end this should be gated on

AMD CI Build failure is not due to this change. It is due to not using filters correctly. This can be merged.

PR #13253, opened 2024-05-31
Make correct fusion decision for inplace fusion

ghost:
There exist bug when making fusion decision for inplace fusion. Current implementation does not consider non-elementwise ops in consumer and will lead to inplace conflict in some specific case. Detailed information can be seen in the added test.

### DIFF
--- a/xla/service/hlo_dataflow_analysis.cc
+++ b/xla/service/hlo_dataflow_analysis.cc
@@ -1815,37 +1815,6 @@ bool HloDataflowAnalysis::DoesNotUseOperandBuffer(
 
 namespace {
 
-// Removes layers of tuple indirection introduced via 'tuple' and
-// 'get-tuple-element' instructions to more directly identify the source of the
-// given HLO value (identified by the given `ShapeIndex` into the output of the
-// given `HloInstruction`).
-//
-// e.g. for the following:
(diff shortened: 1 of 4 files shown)


akuegel (approved):
Thanks, just curious, did you find this just by inspecting the code, or by debugging wrong results?

lingzhi98:
By debugging wrong results caused by jax test https://github.com/google/jax/blob/b480fd55ab0f454a29c8d7e169ce77b8330bc2aa/tests/linalg_test.py#L1376. Due to priority fusion make decision with considering the device information, the reason of why openxla don't find this issue maybe no testing it on different machine? I find this failure on intel atsm machine, which is similar to nvidia a10.

akuegel:
> By debugging wrong results caused by jax test https://github.com/google/jax/blob/b480fd55ab0f454a29c8d7e169ce77b8330bc2aa/tests/linalg_test.py#L1376. Due to priority fusion make decision with considering the device information, the reason of why openxla don't find this issue maybe no testing it on different machine? I find this failure on intel atsm machine, which is similar to nvidia a10.

Yes, most likely it is due to a different device. I can confirm that linalg_test passes on the devices we test on (for example A100 GPUs).

PR #13254, opened 2024-05-31
[NVIDIA] Support pipelining collectives in backward direction with loop invariant instructions in chain

Tixxx:
In some quantization patterns like:
```
                constant
                      |
                broadcast
                  /       \
                /           \
           clamp         clamp
            /                  \
         dot            all-gather
```
The all-gathers cannot be pipelined due to the shared operand broadcast having multiple users and not all users are in all-gather's operand chain. The backward pipelining restricts this situation to avoiding exploding data dependencies across loop iterations.
An exception can be made if such operand chain only contains loop invariant instructions in this case: broadcast and constant.
This pr introduces a config to collective pipeliner backward direction to allow loop invariant instructions to be in the operand chain.

### DIFF
--- a/xla/service/collective_pipeliner.cc
+++ b/xla/service/collective_pipeliner.cc
@@ -511,12 +511,20 @@ std::optional<std::vector<HloInstruction*>> CollectIndependentOperandChain(
     const bool is_scalar_shaped =
         ShapeUtil::IsEffectiveScalar(chain_instr->shape());
     if (!all_users_in_chain) {
+      // Whether we should allow loop variant parameter in the operand chain of
+      // the collective.
+      bool allow_loop_variant_parameter_in_chain =
+          (chain_instr->opcode() != HloOpcode::kGetTupleElement ||
+           chain_instr->operand(0)->opcode() != HloOpcode::kParameter ||
+           !should_allow_loop_variant_parameter_in_chain(chain_instr));
+      // Whether we should allow loop invariant instructions in the operand
+      // chain of the collective.
+      bool add_loop_invariant_op_in_chain =
+          (should_add_loop_invariant_op_in_chain &&
+           loop_invariant_instructions.contains(chain_instr));
       if ((!loop_invariant_params.contains(chain_instr) && !is_scalar_shaped &&
-           (chain_instr->opcode() != HloOpcode::kGetTupleElement ||
-            chain_instr->operand(0)->opcode() != HloOpcode::kParameter ||
-            !should_allow_loop_variant_parameter_in_chain(chain_instr))) &&
-          !(should_add_loop_invariant_op_in_chain &&
-            loop_invariant_instructions.contains(chain_instr))) {
+           allow_loop_variant_parameter_in_chain) &&
+          !add_loop_invariant_op_in_chain) {
         return std::nullopt;
       }
     }
@@ -706,6 +714,7 @@ class WhileLoopAnalysis {
       bool should_allow_control_dependencies = false,
       bool should_add_loop_invariant_op_in_chain = false);
   HloInstruction* while_loop_instruction() const { return while_; }
+  void ExtractLoopInvariantOps();
 
  private:
   HloInstruction* while_;
@@ -727,6 +736,28 @@ int64_t WhileLoopAnalysis::GetDUSIndex(const HloInstruction* dus) const {
   auto it = dus_index_map_.find(dus);
   CHECK(it != dus_index_map_.end());
   return it->second;
+}
+
+void WhileLoopAnalysis::ExtractLoopInvariantOps() {
+  for (HloInstruction* inst :
+       while_->while_body()->MakeInstructionPostOrder()) {
+    if (inst->opcode() == HloOpcode::kConstant) {
+      invariant_loop_instructions_.insert(inst);
+      continue;
+    }
+    if (invariant_loop_instructions_.contains(inst)) {
+      continue;
+    }
+    // Nodes that only consume loop invariants are also invariants.
+    bool should_add = true;
+    for (const HloInstruction* operand : inst->operands()) {
+      should_add &= (invariant_loop_instructions_.contains(operand) ||
+                     invariant_loop_parameters_.contains(operand));
+    }
+    if (should_add) {
+      invariant_loop_instructions_.insert(inst);
+    }
+  }
 }
 
 bool WhileLoopAnalysis::ComputeLoopStatistics() {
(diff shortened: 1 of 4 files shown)


xla/service/collective_pipeliner.cc
```diff
@@ -508,10 +511,12 @@ std::optional<std::vector<HloInstruction*>> CollectIndependentOperandChain(
     const bool is_scalar_shaped =
         ShapeUtil::IsEffectiveScalar(chain_instr->shape());
     if (!all_users_in_chain) {
-      if (!loop_invariant_params.contains(chain_instr) && !is_scalar_shaped &&
-          (chain_instr->opcode() != HloOpcode::kGetTupleElement ||
-           chain_instr->operand(0)->opcode() != HloOpcode::kParameter ||
-           !should_allow_loop_variant_parameter_in_chain(chain_instr))) {
+      if ((!loop_invariant_params.contains(chain_instr) && !is_scalar_shaped &&
```
frgossen, line 514: Can yo unwrap this expression and give it some local names?

xla/service/collective_pipeliner.cc
```diff
@@ -796,6 +807,28 @@ bool WhileLoopAnalysis::ComputeLoopStatistics() {
       invariant_loop_parameters_.insert(loop_root->operand(i));
     }
   }
+
+  // Extract all loop invariant instructions.
```
frgossen, line 811: Extract this as a function?

xla/service/collective_pipeliner_test.cc
```diff
@@ -2228,5 +2229,90 @@ ENTRY main.3813_spmd {
   ASSERT_IS_OK(verifier.Run(module.get()).status());
 }
 
+TEST_F(CollectivePipelinerTest, PipelineBackwardIncludeInvariantInChain) {
```
frgossen, line 2232: I assume this is the case that you care about. Can we also add a minimal test case with nothing but a collective + bcast constant?

xla/service/collective_pipeliner_test.cc
```diff
@@ -2228,5 +2229,90 @@ ENTRY main.3813_spmd {
...
+          /*pipeline_use_tree=*/false,
+          /*process_different_sized_ops=*/false,
+          /*direction=*/CollectivePipeliner::PipeliningDirection::kBackward,
+          /*should_process=*/IsAllGather,
+          /*acceptable_formatting=*/HloPredicateTrue,
+          /*reuse_pipelined_op_buffer=*/HloPredicateTrue,
+          /*should_allow_loop_variant_parameter_in_chain=*/HloPredicateTrue,
+          /*postprocess_backward_peeled=*/std::nullopt,
+          /*postprocess_backward_rotated=*/std::nullopt,
+          /*should_add_loop_invariant_op_in_chain=*/true)
```
frgossen, line 2299: Can we also assert that this does not happen if `should_add_loop_invariant_op_in_chain` is false?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -869,7 +869,12 @@ absl::Status RunCollectiveOptimizationPasses(
         CollectivePipeliner::PipeliningDirection::kBackward,
         /*should_process=*/HloPredicateIsOp<HloOpcode::kAllGather>,
         /*acceptable_formatting=*/HloPredicateTrue,
-        /*reuse_pipelined_op_buffer=*/HloPredicateFalse};
+        /*reuse_pipelined_op_buffer=*/HloPredicateFalse,
+        /*should_allow_loop_variant_parameter_in_chain=*/HloPredicateFalse,
+        /*should_allow_control_dependencies=*/false,
+        /*postprocess_backward_peeled_op=*/std::nullopt,
+        /*postprocess_backward_rotated_op=*/std::nullopt,
+        /*acceptable_loop_invariant_op_in_chain=*/true};
```
frgossen, line 877: Should this be an XLA flag?

frgossen (changes requested):
Thank you!

xla/service/collective_pipeliner.cc
Tixxx: i extracted the partial logic out of this entire expression, let me know if i understood your intention correctly.

xla/service/collective_pipeliner.cc
Tixxx: uptdaed

xla/service/collective_pipeliner_test.cc
Tixxx: yup, i minimized this test a bit more.

xla/service/collective_pipeliner_test.cc
Tixxx: sure, let me change it.

xla/service/gpu/gpu_compiler.cc
Tixxx: the intention is to let this be controlled by the pipeliner flag and support pipelining loop invariant by default for ag for gpu. I dont think having a separate flag for just this is necessary but let me know if you think otherwise.

xla/service/collective_pipeliner.cc
```diff
@@ -508,10 +511,20 @@ std::optional<std::vector<HloInstruction*>> CollectIndependentOperandChain(
...
            chain_instr->operand(0)->opcode() != HloOpcode::kParameter ||
-           !should_allow_loop_variant_parameter_in_chain(chain_instr))) {
+           !should_allow_loop_variant_parameter_in_chain(chain_instr));
+      // Whether we should allow loop invariant instructions in the operand
+      // chain of the collective.
+      bool add_loop_invariant_op_in_chain =
+          (should_add_loop_invariant_op_in_chain &&
+           loop_invariant_instructions.contains(chain_instr));
+      if ((!loop_invariant_params.contains(chain_instr) && !is_scalar_shaped &&
+           allow_loop_variant_parameter_in_chain) &&
```
frgossen, line 526: nit: remove the unnecessary brackets

xla/service/collective_pipeliner.cc
```diff
@@ -718,6 +738,28 @@ int64_t WhileLoopAnalysis::GetDUSIndex(const HloInstruction* dus) const {
   return it->second;
 }
 
+void WhileLoopAnalysis::ExtractLoopInvariantOps() {
+  for (HloInstruction* inst :
+       while_->while_body()->MakeInstructionPostOrder()) {
+    if (inst->opcode() == HloOpcode::kConstant) {
```
frgossen, line 744: I think hlo_query has some util function for walking over instuctions of an op type.

frgossen (approved):
Thanks!

frgossen (changes requested):
Can you provide some data points on where this change improves performance? What models was this benchmarked on?

Tixxx:
Hi @frgossen sorry for the delay in response, i see that this is merged thanks.
For the perf issue, we are benchmarking on an MOE model for a customer. With this change, we see about 4% e2e speedup for now, but a couple of the pipelined AGs are not fully overlapped and we are investigating those. The expected speedup should be about 6-7% e2e.

frgossen:
Thank you!

PR #13263, opened 2024-05-31
[ROCm] Use NearOrEqual for floating point literal comparision

hsharsha:


tdanyluk:
Let's ask Reed maybe. He created this determinism test. Not sure if he meant bit-by-bit determinism or not.

penpornk:
@reedwm Could you please help take a look?

reedwm:
Yes, the test is intended to test bit-by-bit determinism, so changing this to `NearOrEqual` is not correct. I'll close this PR.

@hsharsha can you explain the motivation here? The test should be bit-by-bit deterministic, although I never tested on AMD GPUs.  Let me know if I can help in any way.

hsharsha:
@reedwm  On AMD we are getting close enough result but not bit equivalent. We will work on tuning our determinism. Thanks,

reedwm:
In that case, I would disable the tests on AMD GPUs. You can either reopen this PR and do that or create a new PR. 

I also recommend raising an error if  `--xla_gpu_deterministic_ops` or `--xla_gpu_exclude_nondeterministic_ops` is specified on AMD GPUs and a nondeterministic ops is used.

PR #13278, opened 2024-05-31
Fix _xla_send_recv_validation_attribute in loop-double-buffer-transformer

shraiysh:
This patch fixes the value of valid iteration for GPUs in the loop-double-buffer-transformer pass.

If the original iteration bounds for a particular GPU are {a,b} then the transformed attribute is:

  - If number of iterations is even:
    - For collective.permute.1: {floor((a+1)/2), floor(b/2)}
    - For collective.permute.2: {floor(a/2), max(0, floor((b-1)/2))}
  - If the number of iterations is odd: then the first iteration is peeled out
    - For collective.permute.1: {floor(a/2), max(0, floor((b-1)/2))}
    - For collective.permute.2: {max(0, floor((a-1)/2)), max(0, floor((b-2)/2)}

For collective.permute.peeled, only one instance of the collective is valid. It is valid for all
the devices that had iteration 0 as valid iteration earlier. So, if the attribute was
{{0,4},{0,5},{1,5},{1,6}} for the old instruction, then for the peeled instruction, the attribute
would be {{0,0},{0,0},{1,0},{1,0}} because it is valid for the first two devices, but invalid for
the rest.

More details for these are added in comments as docstrings.


cheshire:
> More details about reasoning for this are [here](https://docs.google.com/document/d/1piacngohC8ZIRXL9Dmx42mczgeOZb3x5lVlhLYIr-p8/)

Why not bring it in as an extended comment? We like those!

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,169 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
   }
 }
 
+absl::StatusOr<std::vector<std::pair<int64_t, int64_t>>> ParseVectorOfPairs(
+    std::basic_string<char>& str) {
```
cheshire, line 87: Nit: basic_string is a bit unusual, maybe absl::string_view?

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,169 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
   }
 }
 
+absl::StatusOr<std::vector<std::pair<int64_t, int64_t>>> ParseVectorOfPairs(
```
cheshire, line 86: Nit: pair<int64, int64> occurs quite a few times below, maybe do

`using Interval = pair<int64, int64>` ?

cheshire, line 86: Docstring?

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,169 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+    int64_t a = replica_group.replica_ids(0);
+    int64_t b = replica_group.replica_ids(1);
+    res.emplace_back(a, b);
+  }
+  return res;
+}
+
+absl::Status SetSendRecvValidationForPeeledInstr(HloInstruction* new_instr,
+                                                 HloInstruction* old_instr) {
+  assert(new_instr->opcode() == old_instr->opcode() &&
```
cheshire, line 103: We do not use `assert` inside, either CHECK if you want to crash the whole process (very rare) or TF_RET_CHECK to return Status.

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,169 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+         "cloned instruction and original instruction have different opcodes");
+  if (!HloPredicateIsOp<HloOpcode::kCollectivePermute,
+                        HloOpcode::kCollectivePermuteStart, HloOpcode::kSend,
+                        HloOpcode::kRecv>(old_instr)) {
+    return absl::OkStatus();
+  }
+
+  google::protobuf::Map<std::string, std::string> attribute_map =
```
cheshire, line 111: const ref?

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,169 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+      ParseVectorOfPairs(attribute_map.at(kSendRecvValidationAttr)));
+
+  uint64_t n_pairs = send_recv_validation_attr.size();
+  if (n_pairs == 0) {
+    return absl::OkStatus();
+  }
+  std::vector<std::pair<int64_t, int64_t>> send_recv_validation_attr_updated(
+      n_pairs, {1, 0});
+  bool is_forward_cycle = (send_recv_validation_attr[0].first == 0);
+  if (is_forward_cycle)
```
cheshire, line 131: Nit: please add braces, here and below.

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,169 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+  for (const ReplicaGroup& replica_group : replica_groups) {
+    TF_RET_CHECK(replica_group.replica_ids_size() == 2);
+    int64_t a = replica_group.replica_ids(0);
+    int64_t b = replica_group.replica_ids(1);
+    res.emplace_back(a, b);
+  }
+  return res;
+}
+
+absl::Status SetSendRecvValidationForPeeledInstr(HloInstruction* new_instr,
```
cheshire, line 101: Please add docstrings.

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,169 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+
+  if (send_recv_validation_attr.size() == 0) {
+    return absl::OkStatus();
+  }
+
+  std::vector<std::pair<int64_t, int64_t>> send_recv_iterations_old_instr,
+      send_recv_iterations_new_instr;
+  send_recv_iterations_old_instr.reserve(send_recv_validation_attr.size());
+  send_recv_iterations_new_instr.reserve(send_recv_validation_attr.size());
+  for (std::pair<int64_t, int64_t> pair : send_recv_validation_attr) {
```
cheshire, line 188: Nit: const ref, here and below/above.

xla/service/gpu/double_buffer_loop_unrolling_test.cc
```diff
@@ -812,6 +813,457 @@ ENTRY main {
...
+  }
+  ASSERT_EQ(collective_permutes.size(), 2);
+  ASSERT_TRUE(collective_permutes[0]->frontend_attributes().map().contains(
+      kSendRecvValidationAttr));
+  ASSERT_TRUE(collective_permutes[1]->frontend_attributes().map().contains(
+      kSendRecvValidationAttr));
+  EXPECT_EQ(collective_permutes[0]->frontend_attributes().map().at(
+                kSendRecvValidationAttr),
+            "{{0,3},{1,3},{1,4},{2,4}}");
+  EXPECT_EQ(collective_permutes[1]->frontend_attributes().map().at(
```
cheshire, line 870: Nit: maybe do a FileCheck test to give an easier at-a-glance representation of what happens to IR?

cheshire (changes requested):

tdanyluk:
I'm deferring to cheshire@ on this one.

shraiysh:
> Why not bring it in as an extended comment? We like those!

I added more details along with examples in the docstrings for the functions. I have also tried to address all your comments. Requesting review again.

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,214 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+          std::max(0.0, std::floor((b - 2) / 2.0)));
+    } else {
+      send_recv_iterations_new_instr1.emplace_back(std::floor((a + 1) / 2.0),
+                                                   std::floor(b / 2.0));
+      send_recv_iterations_new_instr2.emplace_back(
+          std::floor(a / 2.0), std::max(0.0, std::floor((b - 1) / 2.0)));
+    }
+  }
+
+  std::string iteration_instances_new_instr1 =
```
frgossen, line 246: This serialization appears twice here and also above. Extract as a function?

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,214 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+
+  FrontendAttributes attributes;
+  // Copy the attributes from old_instr attributes
+  for (const auto& [key, val] : old_instr->frontend_attributes().map()) {
+    if (key == kSendRecvValidationAttr) {
+      continue;
+    }
+    (*attributes.mutable_map())[key] = val;
+  }
+  std::string send_recv_validation_attr_str =
```
frgossen, line 156: This logica can be shared with below.

xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -83,6 +83,214 @@ void SetChannelIdForNewCollective(HloInstruction* new_instr,
...
+    if (key == kSendRecvValidationAttr) {
+      continue;
+    }
+    (*attributes.mutable_map())[key] = val;
+  }
+  (*attributes.mutable_map())[kSendRecvValidationAttr] =
+      iteration_instances_new_instr1;
+  old_instr->set_frontend_attributes(attributes);
+
+  // Do the same for new instr
```
frgossen, line 280: This looks similar. Can you extraxt this into a function `CopyAttributes` with a skip_key param. 
Or something like that..

frgossen (commented):
Looks very good. A few minor comments.

shraiysh:
Addressed comments. Requesting review again @frgossen @cheshire.

frgossen (approved):
Thanks!

frgossen:
double_buffer_loop_unrolling.cc fails to build. 
Error: error: no template named 'Map' in namespace 'google::protobuf'
  122 |   const google::protobuf::Map<std::string, std::string>& attribute_map =

shraiysh:
I changed that to `const auto&`. Hopefully that'll use the right type.

golechwierowicz (approved):

PR #13289, opened 2024-05-31
[XLA:CPU] [oneDNN] Fix accuracy issue with layernorm fusion

kanvi-nervana:
This PR fixes the accuracy issue seen with F16 ViT model by adding a default value for epsilon. It also adds a slight variation seen for TF layernorm pattern in XLA for F32,BF16,F16.

### DIFF
--- a/xla/tests/onednn_layer_norm_test.cc
+++ b/xla/tests/onednn_layer_norm_test.cc
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include "xla/test.h"
 #include "xla/tests/hlo_test_base.h"
+#include "xla/service/cpu/onednn_util.h"
 
 namespace xla {
 namespace {
@@ -106,6 +107,9 @@ TEST_F(LayerNormTest, LayerNormTest0_FP32) {
 }
 
 TEST_F(LayerNormTest, LayerNormTest0_BF16) {
+  if (!xla::cpu::IsSupportedType(PrimitiveType::BF16)) {
+    GTEST_SKIP() << "CPU does not support BF16.";
+  }
   std::string layer_norm_module_str =
       R"(HloModule layer_norm.test, entry_computation_layout={(bf16[84,197,768]{2,1,0}, f32[768]{0}, f32[768]{0})->bf16[84,197,768]{2,1,0}})" +
       common_hlo_region_ + R"(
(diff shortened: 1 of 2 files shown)


xla/tests/onednn_layer_norm_test.cc
```diff
@@ -141,10 +148,130 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
   MatchOptimizedHlo(layer_norm_module_str, onednn_layer_norm_);
 }
 
+TEST_F(LayerNormTest, LayerNormTest1_F16) {
```
penpornk, line 151: Could you please add a description above the test about what it tests for? E.g., how what it tests is different LayerNormTest0_F16, etc.

xla/tests/onednn_layer_norm_test.cc
```diff
@@ -141,10 +148,130 @@ TEST_F(LayerNormTest, LayerNormTest0_F16) {
...
+    subtract_1 = f16[2,4,8] subtract(broadcast_5, multiply_3)
+    ROOT add_1 = f16[2,4,8] add(multiply_2, subtract_1)
+  }
+ )";
+
+  EXPECT_TRUE(RunAndCompare(layer_norm_module_str, ErrorSpec{1e-2, 1e-2}));
+  MatchOptimizedHlo(layer_norm_module_str, onednn_layer_norm_);
+}
+
+// Test for reversed inputs
```
penpornk, line 209: Could you please mention the name of the variables corresponding to these reversed inputs in the module string? (As a comment in the code.) It's not clear to me.

penpornk (changes requested):
Thank you for the PR!

penpornk (approved):
Since JAX is cutting a new release soon, I'll approve the PR to get the fix in now. But we can still talk about the tests here and make additional changes in a follow-up PR. :)

xla/tests/onednn_layer_norm_test.cc
kanvi-nervana: While executing the ViT model in F16 we observed an accuracy issue which was due to the small epsilon value being constant folded. This PR is addressing that issue by adding a default value for epsilon and the test here is testing that scenario by not having an epsilon as opposed to the test above which contains an epsilon value.

xla/tests/onednn_layer_norm_test.cc
kanvi-nervana: It was also observed that one of the layernorms existing in ViT have a slightly different pattern and was not getting fused. It is addressed here https://github.com/openxla/xla/pull/13289/files#diff-df9572d46ba1c55b10c3e30eb623d0eebadcaa2e9340089d73b2e7f354c795b0R227 and the test is for this pattern change.

PR #13293, opened 2024-05-31
[ROCm] Add gpu_plugin dependency for AMD gpus.

hsharsha:
Without this patch amdgpu_compiler dependency was not included for gpu tests and as a result most unit tests failed on amd platform


hsharsha:
@ddunl  unit tests were failing due to non inclusion of amdgpu_compiler.

ddunl (approved):

Issue #13054, opened 2024-05-24
Wrong output from JAX test after onednn change


hawkinsp:
One of our JAX users reported a miscompilation in the following program when run on CPU:

```
import jax
import jax.numpy as jnp

w = jnp.array([1., 2., 3.], dtype=jnp.float32)
# w = jnp.array([10., 10., 10.], dtype=jnp.float32)  # this works
# w = jnp.array([1e-3, 1e-2, 1e-1], dtype=jnp.float32)  # this does not work
dim = len(w)
x = jnp.ones((dim,))
ys = jnp.ones((10**6, dim))
# ys = jnp.ones((10**5, dim)) # this works

def fun(x, y):
  diff = y - x
  out = jnp.dot(diff, w)
  # The following line should have no effect on the output or the derivatives
  # of this function. Removing it solves the bug.
  out = out + jnp.dot(jnp.dot(diff, jnp.zeros((dim, dim))), diff)
  return out

def mean_fun(x):
  outs = jax.vmap(lambda yy: fun(x, yy))(ys)
  return jnp.mean(outs)

jitted_grad = jax.jit(jax.grad(mean_fun))(x)
grad = jax.grad(mean_fun)(x)
print(f'Difference grads: {grad-jitted_grad}')
print(f'All close: {jnp.sum(jnp.abs(grad-jitted_grad)) < 1e-3}')
# Difference grads: [ 0.        -1.0000002 -2.0000007]
# All close: False
```

If working correctly, it should print:
```
Difference grads: [0. 0. 0.]
All close: True
```

I bisected the problem to XLA commit:
https://github.com/openxla/xla/commit/0dc5563a97d6fa4584b85cb29b4bfcf289604f0b

@jmaksymc

HLO dumps from that commit (bad) and the immediately preceding commit (good): [dumps.tgz](https://github.com/openxla/xla/files/15437598/dumps.tgz)

We should either fix or revert that commit.

mdfaijul:
@hawkinsp @penpornk Thanks for reporting this issue. We are working on it.

mdfaijul:
@hawkinsp @penpornk A PR fixing the bug has been submitted (https://github.com/openxla/xla/pull/13301)

penpornk:
[...]

PR #13301, opened 2024-06-01
[Bug-fix][XLA:CPU][oneDNN] Fix BINARY_ADD fusion to Dot

mdfaijul:
This PR fixes a bug reported for JAX (https://github.com/openxla/xla/issues/13054)

### DIFF
--- a/xla/tests/onednn_matmul_test.cc
+++ b/xla/tests/onednn_matmul_test.cc
@@ -1442,6 +1442,52 @@ TEST_F(MatmulTest, WeightsPrepackAndScratch) {
   )");
 }
 
+TEST_F(MatmulTest, ConsecutiveBinaryAdd) {
+  const char* matmul_module_str = R"(
+  HloModule matmul.test.f32
+    region_0.22 {
+    Arg_0.23 = f32[] parameter(0)
+    Arg_1.24 = f32[] parameter(1)
+  ROOT add.25 = f32[] add(Arg_0.23, Arg_1.24)
+  }
+
+  region_1.29 {
+    Arg_0.30 = f32[] parameter(0)
+    Arg_1.31 = f32[] parameter(1)
+    ROOT add.32 = f32[] add(Arg_0.30, Arg_1.31)
+  }
+
+  ENTRY main {
+    constant.2 = f32[] constant(1e-06)
+    broadcast.3 = f32[1000000] broadcast(constant.2), dimensions={}
+    constant.7 = f32[] constant(1)
+    broadcast.8 = f32[1000000,3] broadcast(constant.7), dimensions={}
+    Arg_0.1 = f32[3] parameter(0)
+    reshape.10 = f32[1,3] reshape(Arg_0.1)
+    broadcast.11 = f32[1,3] broadcast(reshape.10), dimensions={0,1}
+    reshape.12 = f32[3] reshape(broadcast.11)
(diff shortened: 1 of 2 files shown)


xla/tests/onednn_matmul_test.cc
```diff
@@ -1442,6 +1442,52 @@ TEST_F(MatmulTest, WeightsPrepackAndScratch) {
...
+
+  region_1.29 {
+    Arg_0.30 = f32[] parameter(0)
+    Arg_1.31 = f32[] parameter(1)
+    ROOT add.32 = f32[] add(Arg_0.30, Arg_1.31)
+  }
+
+  ENTRY main {
+    constant.2 = f32[] constant(1e-06)
+    broadcast.3 = f32[1000000] broadcast(constant.2), dimensions={}
```
penpornk, line 1462: I don't think the size needs to be this big to reproduce the failure. Would 10 work?

xla/tests/onednn_matmul_test.cc
```diff
@@ -1442,6 +1442,52 @@ TEST_F(MatmulTest, WeightsPrepackAndScratch) {
...
+    broadcast.8 = f32[1000000,3] broadcast(constant.7), dimensions={}
+    Arg_0.1 = f32[3] parameter(0)
+    reshape.10 = f32[1,3] reshape(Arg_0.1)
+    broadcast.11 = f32[1,3] broadcast(reshape.10), dimensions={0,1}
+    reshape.12 = f32[3] reshape(broadcast.11)
+    broadcast.13 = f32[1000000,3] broadcast(reshape.12), dimensions={1}
+    subtract.14 = f32[1000000,3] subtract(broadcast.8, broadcast.13)
+    constant.4 = f32[] constant(0)
+    broadcast.5 = f32[3,3] broadcast(constant.4), dimensions={}
+    dot.15 = f32[1000000,3] dot(subtract.14, broadcast.5), lhs_contracting_dims={1}, rhs_contracting_dims={0}
```
penpornk, line 1473: Can we reduce the ops to just necessary ops that reproduce the failure? I don't think all the dots are needed.

penpornk (changes requested):
Thank you for the fix! Could you please explain what is causing the issue, and how this fix addresses it? Is it because of rank mismatch + wrong auto broadcasting or something?

xla/tests/onednn_matmul_test.cc
kanvi-nervana: The bug seems to be seen with this particular case.

kanvi-nervana:
> Thank you for the fix! Could you please explain what is causing the issue, and how this fix addresses it? Is it because of rank mismatch + wrong auto broadcasting or something?

oneDNN expects Matmul followed by Bias-Add followed by Binary-Add. But, here Matmul is followed by Binary-Add and then by Bias-Add which oneDNN does not support. The fix here is extending the dimensions of the Bias-Add to a Binary-Add which is supported. As seen below

Screenshot 2024-06-03 at 4 02 50 PM

xla/tests/onednn_matmul_test.cc
kanvi-nervana: The issue is not reproducible with a smaller size.

penpornk (approved):
Thank you very much for the clarifications!

PR #13310, opened 2024-06-03
[NVIDIA GPU] Added a rewrite logic in gpu_windowned_einsum_handler to handle all2all

Tixxx:
Added a rewrite logic in gpu_windowned_einsum_handler to decompose a2a+gemm or gemm+a2a into smaller a2a+gemm/gemm+a2a to hide communuication overhead. Partial results will be aggregated at the end.
An example will be:
```

input
   |
a2a{replica_groups={{0,1}}}
   |
gemm
  
```

decomposed into
```

                  input
                 /         \
           slice1      slice2
              /               \
           a2a1             a2a2
            /                  \
       gemm1                 gemm2
                \           /
                    add
```

All partial gemms will be dispatched to parallel streams too to achieve gemm-gemm overlap.

Performance metrics:
For an unit with just a2a+gemm or gemm+a2a, we see from 5-15% speedup depending on the size by doing this type of composition.

### DIFF
--- a/xla/service/gpu/gpu_windowed_einsum_handler_test.cc
+++ b/xla/service/gpu/gpu_windowed_einsum_handler_test.cc
@@ -44,6 +44,16 @@ HloInstruction* FindInstructionByName(HloComputation* comp, std::string name) {
   return nullptr;
 }
 
+std::vector<HloInstruction*> GatherAllInstructions(const HloComputation& computation, HloOpcode opcode) {
+  std::vector<HloInstruction*> all_instructions;
+  for (const auto& instruction : computation.instructions()) {
+    if (instruction->opcode() == opcode) {
+      all_instructions.push_back(instruction);
+    }
+  }
+  return all_instructions;
+}
+
 TEST_F(GpuWindowedEinsumHanlderTest, AgLoopsHaveStreamIds) {
   constexpr absl::string_view kHloString = R"(
 HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(bf16[1,512,24576]{2,1,0}, bf16[24576,24576]{1,0})->bf16[2048,24576]{1,0}}, num_partitions=4
@@ -262,7 +272,6 @@ ENTRY main.12_spmd {
   bool changed;
   TF_ASSERT_OK_AND_ASSIGN(changed, gpu_handler.Run(module.get()));
   EXPECT_TRUE(changed);
-
   HloInstruction* ag_loop =
       FindInstructionByName(module->entry_computation(), "while");
   HloInstruction* inst =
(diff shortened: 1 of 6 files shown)


cheshire:
@golechwierowicz @frgossen could you take a look?

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -287,6 +296,71 @@ ENTRY main.12_spmd {
                       m::Op(), m::Op(), m::Op(), m::Op()),
                   m::Op())));
 }
+TEST_F(GpuWindowedEinsumHanlderTest, A2aGemmHaveStreamIds) {
+  constexpr absl::string_view kHloString = R"(
+HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(bf16[1,8192,32768]{2,1,0}, bf16[1,4,2048,8192]{3,2,1,0})->bf16[1,4,2048,32768]{3,2,1,0}}, num_partitions=8
+
+ENTRY main.9_spmd {
```
frgossen, line 303: Can you add filecheck tests. I think it would be easier to follow the transform then.

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -44,6 +44,16 @@ HloInstruction* FindInstructionByName(HloComputation* comp, std::string name) {
   return nullptr;
 }
 
+std::vector<HloInstruction*> GatherAllInstructions(const HloComputation& computation, HloOpcode opcode) {
```
frgossen, line 47: Can we reuse this from or move it to `utils/hlo_query.h`?

frgossen (changes requested):
Can you provide some experiment results where this transform is beneficial? 

Can we also add an e2e test with 2 GPUs?

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
Tixxx: updated to use filecheck.

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
Tixxx: this is removed in favor of using filecheck

Tixxx:
> Can you provide some experiment results where this transform is beneficial?
> 
> Can we also add an e2e test with 2 GPUs?

We are still running the full model experiment but from an unit with just a2a+gemm or gemm+a2a, we see from 5-15% speedup depending on the size.
I will add e2e tests.

frgossen:
Thanks. Can we add this measure and the full model experiment result to the PR description?

Tixxx:
@frgossen  I added a few e2e tests for all the variations that we care about. We have also identified some optimizations that could help improve perf for the full model run, currently we dont see significant speedup for the model e2e(projecting to see ~5% speedup). It'd be better for them to go in as separate PRs.

frgossen (approved):
Thanks for adding the tests!

golechwierowicz (approved):

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -27,6 +27,7 @@ limitations under the License.
 #include "xla/service/pattern_matcher_gmock.h"
 #include "xla/tests/hlo_test_base.h"
 #include "tsl/platform/statusor.h"
+#include "xla/tests/filecheck.h"
```
akuegel, line 30: This include requires a corresponding dependency to be added to the build file.

Tixxx: Updated

akuegel (approved):

PR #13391, opened 2024-06-04
[NVIDIA] Disable the bias reuse when gemm users > 1

kaixih:
We encountered a new use case where our reset of the `output to operand aliasing` won't be reached when the gemm users > 1 and that additional user does another amax-like ops.

This PR fix it and the corresponding unit test is included.

cc. @philipphack @reedwm @hx89

### DIFF
--- a/xla/service/gpu/tests/gemm_rewrite_test.cc
+++ b/xla/service/gpu/tests/gemm_rewrite_test.cc
@@ -6510,7 +6510,7 @@ TEST_P(ParameterizedFp8GemmRewriteTest, ScaledABScaledDReluActivationF8) {
       )");
 }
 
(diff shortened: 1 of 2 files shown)


xla/service/gpu/tests/gemm_rewrite_test.cc
```diff
@@ -6522,6 +6522,12 @@ TEST_P(ParameterizedFp8GemmRewriteTest, ScaledABScaledDMatrixBiasF8) {
   const char* hlo_text = R"(
     HloModule test
 
+    apply {
```
philipphack, line 6525: Change the name to ScaledABScaledDMatrixBiasWithDAmaxF8 or add this as a new test.

kaixih: Done.

reedwm (approved):

PR #13392, opened 2024-06-04
Remove Redundant Shape Comparison in Norm Rewriter

philipphack:
Removes a superfluous comparison of the shapes of the input and scale when matching and rewriting layer norm patterns.

### DIFF
--- a/xla/service/gpu/cudnn_norm_rewriter.cc
+++ b/xla/service/gpu/cudnn_norm_rewriter.cc
@@ -970,13 +970,6 @@ class CudnnNormRewriterVisitor : public DfsHloRewriteVisitor {
         VLOG(1) << "Layer norm input dimensions not supported.";
         return absl::OkStatus();
       }
-      for (int i = 0; i < norm_dims.size(); ++i) {
-        if (x.Instr()->shape().dimensions(norm_dims[i]) !=
-            scale->shape().dimensions(i)) {
-          VLOG(1) << "Layer norm input dimensions not supported.";
-          return absl::OkStatus();
-        }
-      }
 
       // Verify the broadcasts of scale and bias.
       if (!ShapeUtil::EqualIgnoringElementType(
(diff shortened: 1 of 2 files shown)


xla/service/gpu/cudnn_norm_rewriter.cc
```diff
@@ -970,13 +970,6 @@ class CudnnNormRewriterVisitor : public DfsHloRewriteVisitor {
         VLOG(1) << "Layer norm input dimensions not supported.";
         return absl::OkStatus();
       }
-      for (int i = 0; i < norm_dims.size(); ++i) {
```
bchetioui, line 973: Just to understand, is the claim that this is redundant because we check `broadcast_bias` below?

And in that case, I assume this is already covered by an existing test, right? (In which case, could you please link to it?)

Sorry, I'm not too familiar with that code, so I appreciate your help in making sure all of this is correct :)

philipphack: Yes, we verify the shape of the broadcasted scale [here](https://github.com/openxla/xla/blob/4c64418323220b2b6cd17c46e1ec15f59ecd74ac/xla/service/gpu/cudnn_norm_rewriter.cc#L982), and we separately also check the dimensions of the broadcast [here](https://github.com/openxla/xla/blob/4c64418323220b2b6cd17c46e1ec15f59ecd74ac/xla/service/gpu/cudnn_norm_rewriter.cc#L988).

Unit test [LayerNorm4D3IncorrectScaleBroadcast](https://github.com/openxla/xla/blob/4c64418323220b2b6cd17c46e1ec15f59ecd74ac/xla/service/gpu/cudnn_norm_rewriter_test.cc#L548) verifies that the norm is not rewritten when the scale has been broadcasted incorrectly.

bchetioui (approved):
Thanks a lot @philipphack!

bchetioui (approved):

PR #13425, opened 2024-06-05
[ROCM] gemm precision settings for autotuner

pemeliya:
Here we add a new flag **xla_gpu_autotune_gemm_rtol** which controls the relative precision used by the BufferComparator (defaults to **0.1**).

Also I added one more "paranoid" level 5 for **xla_gpu_autotune_level** which forces the autotuner to discard solutions with accuracy problems. Long time I was under impression that the autotuner already does it, however this is not the case as outlined [here](https://github.com/ROCm/xla/blob/6301f04c50c7637a65b3e0c6f40be628aa00947f/xla/service/gpu/stream_executor_util.cc#L640). BufferComparator just prints out the error message but **keeps wrong solutions** as possible candidates which could lead to a great confusion. So, the autotune level 5 is supposed to discard solutions with accuracy problems.

Besides, I also did some small refactoring on BufferComparator to simplify the source code and added **verbose** flag in order to mute error messages if needed.

@xla-rotation: could you please have a look?

### DIFF
--- a/xla/service/gpu/buffer_comparator.cc
+++ b/xla/service/gpu/buffer_comparator.cc
@@ -49,7 +49,8 @@ using ComparisonKernelT =
 
 struct ComparisonParams {
   double relative_tol = 0.1;
-  const Shape* shape = nullptr;
+  bool verbose = true;
+  const Shape *shape = nullptr;
   se::Stream* stream = nullptr;
   se::DeviceMemoryBase current{};
   se::DeviceMemoryBase expected{};
@@ -129,6 +130,7 @@ static absl::StatusOr<bool> HostCompare(const ComparisonParams& params) {
     return a;
   };
   int differences_seen = 0;
+
   for (int64_t i = 0; i < n && differences_seen < 10; ++i) {
     auto current_value = static_cast<ComparisonType>(host_current[i]);
     auto expected_value = static_cast<ComparisonType>(host_expected[i]);
@@ -148,11 +150,11 @@ static absl::StatusOr<bool> HostCompare(const ComparisonParams& params) {
         !(std::abs(current_value_canonical - expected_value_canonical) /
               (std::max(std::abs(current_value_canonical),
                         std::abs(expected_value_canonical)) +
-               1) <
-          params.relative_tol)) {
+               1) < params.relative_tol)) {
+      if(!params.verbose) return false; // Return immediately if not verbose.
       ++differences_seen;
       LOG(ERROR) << "Difference at " << i << ": " << current_value
-                 << ", expected " << expected_value;
+                   << ", expected " << expected_value;
     }
   }
   return differences_seen == 0;
@@ -180,7 +182,9 @@ static absl::StatusOr<bool> CompareEqualParameterized(
 absl::StatusOr<bool> BufferComparator::CompareEqual(
     se::Stream* stream, se::DeviceMemoryBase current,
     se::DeviceMemoryBase expected) const {
-  ComparisonParams params{relative_tol_, &shape_, stream, current, expected};
+
+  ComparisonParams params{
+    relative_tol_, verbose_, &shape_, stream, current, expected};
 
   switch (shape_.element_type()) {
 #if GOOGLE_CUDA  // not available for ROCm yet..
@@ -226,10 +230,9 @@ absl::StatusOr<bool> BufferComparator::CompareEqual(
   }
 }
 
-BufferComparator::BufferComparator(const Shape& shape,
-                                   const HloModuleConfig& config,
-                                   double tolerance)
-    : shape_(shape), config_(config), relative_tol_(tolerance) {
+BufferComparator::BufferComparator(const Shape& shape, double tolerance, 
+      bool verbose) : 
+        shape_(shape), relative_tol_(tolerance), verbose_(verbose) {
   // Normalize complex shapes: since we treat the passed array as a contiguous
   // storage it does not matter which dimension are we doubling.
   auto double_dim_size = [&]() {
--- a/xla/service/gpu/buffer_comparator.h
+++ b/xla/service/gpu/buffer_comparator.h
@@ -34,8 +34,8 @@ class BufferComparator {
   BufferComparator(const BufferComparator&) = delete;
   BufferComparator(BufferComparator&&) = default;
 
-  BufferComparator(const Shape& shape, const HloModuleConfig& config,
-                   double tolerance = 0.1);
+  explicit BufferComparator(const Shape& shape, double tolerance = 0.1,
+                   bool verbose = true);
 
   // Returns true if the two buffers compare equal. The definition of "equal"
   // is:
@@ -51,8 +51,8 @@ class BufferComparator {
                                     se::DeviceMemoryBase expected) const;
  private:
   Shape shape_;
-  HloModuleConfig config_;
-  double relative_tol_;
+  double relative_tol_;  // relative tolerance for comparison
+  bool verbose_;          // whether to print out error message on mismatch
 };
 
 namespace buffer_comparator {
--- a/xla/service/gpu/buffer_comparator_test.cc
+++ b/xla/service/gpu/buffer_comparator_test.cc
@@ -75,7 +75,7 @@ class BufferComparatorTest : public testing::Test {
         ShapeUtil::MakeShape(
             primitive_util::NativeToPrimitiveType<ElementType>(),
             {static_cast<int64_t>(current.size())}),
-        HloModuleConfig(), tolerance);
+        tolerance);
     return comparator
         .CompareEqual(stream.get(), current_buffer.memory(),
                       expected_buffer.memory())
@@ -394,8 +394,7 @@ TEST_F(BufferComparatorTest, BF16) {
       stream_exec_->AllocateArray<Eigen::bfloat16>(element_count));
   InitializeBuffer(stream.get(), BF16, &rng_state, rhs.memory());
 
-  BufferComparator comparator(ShapeUtil::MakeShape(BF16, {element_count}),
-                              HloModuleConfig());
+  BufferComparator comparator(ShapeUtil::MakeShape(BF16, {element_count}));
   EXPECT_FALSE(comparator.CompareEqual(stream.get(), lhs.memory(), rhs.memory())
                    .value());
 }
--- a/xla/service/gpu/gemm_algorithm_picker.cc
+++ b/xla/service/gpu/gemm_algorithm_picker.cc
@@ -92,13 +92,18 @@ class GemmAutotuner {
   se::Stream* stream_ = nullptr;
   bool deterministic_ops_ = false;
   size_t solutions_limit_ = 0;
+  size_t num_algorithms_left_ = 0;
 
  public:
   explicit GemmAutotuner(const AutotuneConfig& autotune_config)
       : autotune_config_(autotune_config) {}
 
+  size_t num_algorithms_left() const { return num_algorithms_left_; }
+
   absl::StatusOr<AutotuneResult> operator()(const HloInstruction* gemm,
                                             const AutotuneCacheKey& key) {
+
+    num_algorithms_left_ = 0;
     if (autotune_config_.IsDeviceless()) {
       // Return empty result, will tune at runtime.
       return AutotuneResult{};
@@ -274,7 +279,11 @@ class GemmAutotuner {
                               ShapeUtil::ByteSizeOf(output_shape)));
     }
 
-    BufferComparator comparator(output_shape, hlo_module_config);
+    // Do not print error messages if should_skip_wrong_results() is ON.
+    BufferComparator comparator(output_shape, 
+        hlo_module_config.debug_options().xla_gpu_autotune_gemm_rtol(),
+        /* verbose */!autotune_config_.should_skip_wrong_results()
+    );
     std::vector<AutotuneResult> results;
     results.reserve(algorithms.size());
     std::optional<int64_t> reference_algorithm;
@@ -307,6 +316,7 @@ class GemmAutotuner {
           absl::Milliseconds(profile_result.elapsed_time_in_ms()));
 
       if (!autotune_config_.should_check_correctness()) {
+        num_algorithms_left_++;
         continue;
       }
       TF_ASSIGN_OR_RETURN(
@@ -322,25 +332,35 @@ class GemmAutotuner {
         continue;
       }
 
+      num_algorithms_left_++; 
       if (!reference_algorithm) {
         TF_RETURN_IF_ERROR(stream_->Memcpy(&reference_buffer, OutputBuffer(),
                                            OutputBuffer().size()));
         reference_algorithm = profile_result.algorithm();
-      } else {
-        // Perform the comparison.
-        TF_ASSIGN_OR_RETURN(
-            bool outputs_match,
-            comparator.CompareEqual(stream_, /*current=*/OutputBuffer(),
+        continue;
+      } 
+      // Perform the comparison versus the reference algorithm.
+      TF_ASSIGN_OR_RETURN(
+          bool outputs_match,
+          comparator.CompareEqual(stream_, /*current=*/OutputBuffer(),
                                     /*expected=*/reference_buffer));
-        if (!outputs_match) {
-          LOG(ERROR) << "Results mismatch between different GEMM algorithms. "
-                     << "This is likely a bug/unexpected loss of precision.";
-          CHECK(!autotune_config_.should_crash_on_check_failure());
+      if (!outputs_match) {
+        LOG(ERROR) << "Results mismatch between different GEMM algorithms. "
+                   << "This is likely a bug/unexpected loss of precision.";
+        CHECK(!autotune_config_.should_crash_on_check_failure());
 
-          result.mutable_failure()->set_kind(AutotuneResult::WRONG_RESULT);
-          result.mutable_failure()->mutable_reference_gemm()->set_algorithm(
-              *reference_algorithm);
+        // By default, autotuner does NOT really skip wrong results, but 
+        // merely prints out the above error message: this may lead to a 
+        // great confusion. When should_skip_wrong_results() is set to true,
+        // solutions with accuracy problems will be disqualified.
+        auto kind = AutotuneResult::WRONG_RESULT;
+        if (autotune_config_.should_skip_wrong_results()) {
+          kind = AutotuneResult::DISQUALIFIED;
+          num_algorithms_left_--; // Decrement again since we disqualified it.
         }
+        result.mutable_failure()->set_kind(kind);
+        result.mutable_failure()->mutable_reference_gemm()->set_algorithm(
+              *reference_algorithm);
       }
     }  // for algorithms
 
@@ -373,13 +393,15 @@ class GemmAutotuner {
 // Do Gemm Autotune without stream executor. Use results from autotune cache
 // only.
 absl::StatusOr<bool> RunOnInstruction(HloInstruction* gemm,
-                                      const AutotuneConfig& config) {
+                                const AutotuneConfig& config,
+                                size_t *num_algorithms_left) {
   VLOG(3) << "Loading the autotune result of GemmThunk " << gemm->ToString();
 
   GpuBackendConfig gpu_config =
       gemm->backend_config<GpuBackendConfig>().value();
   GemmBackendConfig& backend_config = *gpu_config.mutable_gemm_backend_config();
 
+  *num_algorithms_left = 0;
   // Degenerate gemms replaced with memzero operation, no need to auto tune it.
   if (backend_config.alpha_real() == 0.0 &&
       backend_config.alpha_imag() == 0.0 && backend_config.beta() == 0.0) {
@@ -393,6 +415,7 @@ absl::StatusOr<bool> RunOnInstruction(HloInstruction* gemm,
                       AutotunerUtil::Autotune(
                           gemm, config, [&] { return autotuner(gemm, key); }));
 
+  *num_algorithms_left = autotuner.num_algorithms_left();
   auto old_algorithm = backend_config.selected_algorithm();
   bool update_algorithm =
       IsCublasLtMatmulF8(*gemm) ||
@@ -434,11 +457,15 @@ absl::StatusOr<bool> RunOnInstruction(HloInstruction* gemm,
 }
 
 absl::StatusOr<bool> RunOnComputation(HloComputation* computation,
-                                      AutotuneConfig config) {
+               AutotuneConfig config, size_t *num_algorithms_left) {
   bool changed = false;
+
   for (HloInstruction* instr : computation->instructions()) {
     if (IsCublasGemm(*instr)) {
-      TF_ASSIGN_OR_RETURN(bool result, RunOnInstruction(instr, config));
+      size_t num_left;
+      TF_ASSIGN_OR_RETURN(bool result, RunOnInstruction(instr, config, &num_left));
+      // Gathering statistics on the algorithms left after tuning (for testing)
+      *num_algorithms_left = std::max(*num_algorithms_left, num_left);
       changed |= result;
     }
   }
@@ -453,6 +480,7 @@ absl::StatusOr<bool> GemmAlgorithmPicker::Run(
   XLA_SCOPED_LOGGING_TIMER(
       absl::StrCat("GemmAlgorithmPicker for ", module->name()));
 
+  num_algorithms_left_ = 0;
   if (module->config().debug_options().xla_gpu_autotune_level() == 0) {
     VLOG(2) << "GEMM auto-tuning disabled, GemmAlgorithmPicker returning early";
     return false;
@@ -461,7 +489,8 @@ absl::StatusOr<bool> GemmAlgorithmPicker::Run(
   bool changed = false;
   for (HloComputation* computation :
        module->MakeNonfusionComputations(execution_threads)) {
-    TF_ASSIGN_OR_RETURN(bool result, RunOnComputation(computation, config_));
+    TF_ASSIGN_OR_RETURN(bool result, 
+                RunOnComputation(computation, config_, &num_algorithms_left_));
     changed |= result;
   }
   return changed;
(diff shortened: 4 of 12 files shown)


xla/service/gpu/buffer_comparator.cc
```diff
@@ -148,83 +139,78 @@ absl::StatusOr<bool> HostCompare(se::Stream* stream,
               (std::max(std::abs(current_value_canonical),
                         std::abs(expected_value_canonical)) +
                1) <
-          kTolerance)) {
+          (double)rtol_tolerance_)) {
```
tdanyluk, line 142: use static_cast?

xla/service/gpu/buffer_comparator.cc
```diff
@@ -148,83 +139,78 @@ absl::StatusOr<bool> HostCompare(se::Stream* stream,
...
   }
 }
 
-BufferComparator::BufferComparator(const Shape& shape,
-                                   const HloModuleConfig& config)
-    : shape_(shape), config_(config) {
+BufferComparator::BufferComparator(const Shape& shape, 
+    const HloModuleConfig& config, bool verbose)
+    : shape_(shape), 
+      rtol_tolerance_(config.debug_options().xla_gpu_autotune_gemm_rtol()),
```
tdanyluk, line 212: rtol_tolerance_ sounds strange because it means "relative tolerance tolerance", maybe use relative_tolerance_?

xla/service/gpu/buffer_comparator.h
```diff
@@ -43,11 +44,31 @@ class BufferComparator {
...
+    std::string_view kernel_name, void* kernel_symbol);
 
  private:
   Shape shape_;
-  HloModuleConfig config_;
+  float rtol_tolerance_;  // relative tolerance for comparison
+  bool verbose_;          // whether to print out error message on mismatch
+  se::Stream* stream_ = nullptr;
+  se::DeviceMemoryBase current_;
+  se::DeviceMemoryBase expected_;
```
tdanyluk, line 71: Could we not add these members please?
I think that Compare(a, b) is more readable then Compare() which reads a_ and b_.
In general in XLA multiple people prefer functional style now, because it shows dependencies instead of hiding them.
Also these are just used to do parameter passing in an implicit way. What if we wanted to make this class multi-threaded later? It wouldn't work in that case..

xla/service/gpu/buffer_comparator.cc
```diff
@@ -52,26 +50,21 @@ using ComparisonKernelT =
...
-                                          se::DeviceMemoryBase current,
-                                          se::DeviceMemoryBase expected,
-                                          const Shape& buffer_shape,
-                                          const HloModuleConfig& config,
-                                          std::string_view kernel_name,
-                                          void* kernel_symbol) {
-  se::StreamExecutor* executor = stream->parent();
+absl::StatusOr<bool> BufferComparator::DeviceCompare(
+              std::string_view kernel_name, void* kernel_symbol) {
+  se::StreamExecutor* executor = stream_->parent();
```
tdanyluk, line 55: Please keep this and similar functions static. (see other comment)

tdanyluk (commented):
Thank you, added some comments.

tdanyluk:
BTW it's not guaranteed that the reference result is the correct one, so not sure if this new level actually helps in all cases.
Please consider this.

xla/service/gpu/buffer_comparator.cc
pemeliya: yes makes sense :)

pemeliya:
> BTW it's not guaranteed that the reference result is the correct one, so not sure if this new level actually helps in all cases. Please consider this.
> 
> See this:
> 
> https://github.com/openxla/xla/blob/534ffbf30a14beb7c1d3c1548991a1881302afc3/xla/service/gpu/stream_executor_util.cc#L640

Yes. exactly, I have also mentioned this place in the original description. So, the user needs to know what he is doing: anyway the default autotune level is still 4 which means that solutions marked with WRONG_RESULT are kept in the list. On ROCM, we have a problem that the default (fallback) solution is accurate but additional ones are not due to whatever bugs in the underlying math libraries. But we still would like to use gemm autotuning in that case, and hence discard solutions with bad accuracy ..

xla/service/gpu/buffer_comparator.h
pemeliya: OK, then I will rewind some of my refactoring changes: perhaps, I can group same parameters into some struct in order to avoid passing the same set of params over again (to make code a bit less verbose).

xla/service/gpu/buffer_comparator.cc
```diff
@@ -182,49 +183,56 @@ static absl::StatusOr<bool> CompareEqualParameterized(
...
     se::DeviceMemoryBase expected) const {
+
+  ComparisonParams params{
+    .relative_tol = relative_tol_,
+    .verbose = verbose_,
+    .shape = shape_, // Keep only reference to the shape
+    .stream = stream,
+    .current = current,
+    .expected = expected,
+  };
```
tdanyluk, line 194: Named initializers are a C++20 construct and XLA is c++17, so this will fail our internal presubmit, can you change it please?

xla/service/gpu/buffer_comparator.cc
```diff
@@ -182,49 +183,56 @@ static absl::StatusOr<bool> CompareEqualParameterized(
...
+
+#define FUNC_NAME(x) #x, buffer_comparator::x()
```
tdanyluk, line 196: Sorry for adding one more comment, this macro which generates 2 values is really not a good fit for our code style, could you just write out the names twice, as it was?
https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros
"Try not to use macros that expand to unbalanced C++ constructs..."

xla/service/gpu/buffer_comparator.cc
```diff
@@ -41,37 +41,40 @@ limitations under the License.
...
 template <typename ElementT>
 using ComparisonKernelT =
     se::TypedKernel<se::DeviceMemory<ElementT>, se::DeviceMemory<ElementT>,
                     float, uint64_t, se::DeviceMemory<uint64_t>>;
 
+struct ComparisonParams {
+  float relative_tol;
+  bool verbose;
+  const Shape& shape;
+  se::Stream* stream;
```
tdanyluk, line 53: maybe set default values just to be super safe

tdanyluk (commented):
thanks for the changes

pemeliya:
> thanks for the changes

yep, I have fixed the issues you mentioned: I have also squashed everything into a single commit since frequent changes in xla.proto create lots of conflitcs

tdanyluk (approved):
thanks again!

golechwierowicz (changes requested):
Can you add a test to `xla/service/gpu/gemm_algorithm_picker_test.cc`?

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -279,6 +277,7 @@ class GemmAutotuner {
                               ShapeUtil::ByteSizeOf(output_shape)));
     }
 
+    VLOG(0) << "autotune_config_.should_skip_wrong_results() " << autotune_config_.should_skip_wrong_results();
```
akuegel, line 280: maybe use VLOG(2) instead? Doesn't seem it requires such high severity level.

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -330,38 +330,45 @@ class GemmAutotuner {
         continue;
       }
 
+      num_algorithms_left_++; // Increment the number of "valid" algorithms
```
akuegel, line 333: I think the comment is redundant here.

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -330,38 +330,45 @@ class GemmAutotuner {
...
+              *reference_algorithm);
       }
     }  // for algorithms
 
     absl::StatusOr<AutotuneResult> best =
         PickBestResult(results, gemm->ToString(), hlo_module_config);
     if (best.ok()) {
+      // Return a real algorithm ID if return_algo_index is false: 
+      // e.g., in case of legacy cublas tuning.
+      if(!return_algo_index) return best;
```
akuegel, line 370: nit: add a space between "if" and '('

penpornk:
@pemeliya Could you please help resolve conflicts as well?

pemeliya:
Sure, I am currently OOO but I will take care of this today evening.

penpornk:
@pemeliya Please take your time. Thank you very much! :)

pemeliya:
It seems that this PR https://github.com/openxla/xla/pull/13788 has already added tolerance parameter to BufferComparator and also made three template functions to members.. So I had to rebase it and make those template functions back to static. But actually, the changes were almost the same as in this PR (yet covered only BufferComparator). Also I had to squash everything into one commit to help resolving potential new conflicts.

xla/service/gpu/buffer_comparator_test.cc
```diff
@@ -52,6 +52,13 @@ class BufferComparatorTest : public testing::Test {
         stream_exec_(platform_->ExecutorForDevice(0).value()) {
   }
 
+  const DebugOptions& debug_options() const {
```
akuegel, line 55: It looks to me these two methods are unused?

xla/service/gpu/buffer_comparator_test.cc
```diff
@@ -99,6 +106,7 @@ class BufferComparatorTest : public testing::Test {
                                                           kDefaultTolerance);
   }
 
+  xla::HloModuleConfig module_config_{};
```
akuegel, line 109: This seems to be unused.

xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -52,31 +53,102 @@ class GemmAlgorithmPickerTest : public HloTestBase,
...
+    num_left1 = gpicker.num_algorithms_left();
+    if(num_left1 < 2) {
+      GTEST_SKIP() << "Too few algorithms left after the first step";
+    }
+  }
+
+  // Clear cache before the second run!
+  AutotunerUtil::ClearAutotuneResults();
+  {
+    // Run once again but now with autotune level 5 and embarassignly tight
```
akuegel, line 124: nit: embarassignly -> embarassingly

xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -52,31 +53,102 @@ class GemmAlgorithmPickerTest : public HloTestBase,
...
+    // Run once again but now with autotune level 5 and embarassignly tight 
+    // rtol which shall disqualify most of the algorithms.
+
+    // Note that, we have "two sources of truth" for GemmAlgorithmPicker: i.e.,
+    // debug_options are used to initialize both 'HloModuleConfig' and also 
+    // 'AutotuneConfig'..
+    debug_opts.set_xla_gpu_autotune_gemm_rtol(1e-12);
+    debug_opts.set_xla_gpu_autotune_level(5);
+    module_cfg.set_debug_options(debug_opts);
+    TF_ASSERT_OK_AND_ASSIGN(auto m,
```
akuegel, line 133: Instead of parsing again, you can modify the module config of the module:

```
m->mutable_config().set_debug_options(debug_opts);
```

xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -52,31 +53,102 @@ class GemmAlgorithmPickerTest : public HloTestBase,
...
+          GemmRewriter(gpu_comp(), /*toolkit_version=*/12040),
+          m.get()));
+
+    AutotuneConfig cfg{DeviceConfig{stream_exec(), nullptr}, debug_opts};
+    GemmAlgorithmPicker gpicker(cfg);
+    TF_ASSERT_OK_AND_ASSIGN(changed, RunHloPass(gpicker, m.get()));
+    ASSERT_TRUE(changed);
+    num_left2 = gpicker.num_algorithms_left();
   }
+  // Assert that we have much fewer algorithms left after the second run
```
akuegel, line 147: nit: remove the word "much". We only assert that we have fewer algorithms. End the line with '.'

xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -52,31 +53,102 @@ class GemmAlgorithmPickerTest : public HloTestBase,
...
+
+  // Clear cache before the second run!
+  AutotunerUtil::ClearAutotuneResults();
+  {
+    // Run once again but now with autotune level 5 and embarassignly tight 
+    // rtol which shall disqualify most of the algorithms.
+
+    // Note that, we have "two sources of truth" for GemmAlgorithmPicker: i.e.,
+    // debug_options are used to initialize both 'HloModuleConfig' and also 
+    // 'AutotuneConfig'..
```
akuegel, line 129: nit: remove duplicate '.' at the end of the line

xla/service/gpu/buffer_comparator_test.cc
pemeliya: indeed.. after the last small refactoring, module config is no longer needed, will take care of this today

xla/service/gpu/gemm_algorithm_picker_test.cc
pemeliya: Thanks for the hint

akuegel (approved):

xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -52,31 +53,101 @@ class GemmAlgorithmPickerTest : public HloTestBase,
...
+    TF_ASSERT_OK_AND_ASSIGN(
+      bool changed,
+      RunHloPass(
+          GemmRewriter(gpu_comp(), /*toolkit_version=*/12040),
+          module.get()));
+
+    AutotuneConfig cfg{DeviceConfig{stream_exec(), nullptr}, debug_opts};
+    GemmAlgorithmPicker gpicker(cfg);
+    TF_ASSERT_OK_AND_ASSIGN(changed, RunHloPass(gpicker, module.get()));
+    ASSERT_TRUE(changed);
```
reedwm, line 143: I'm seeing this assertion flakily fail on V100s. I think the issue is the algorithm doesn't necessarily change the second time you call `RunHloPass`, since the selected algorithm from the first run of GemmAlgorithmPicker may have chosen an algorithm that doesn't have accuracy issues.

Probably the best way to fix this is to copy the module before passing it to GemmAlgorithmPicker, so you can pass a fresh copy to both GemmAlgorithmPicker instances.

xla/service/gpu/buffer_comparator.h
```diff
@@ -30,8 +30,8 @@ class BufferComparator {
   BufferComparator(const BufferComparator&) = delete;
   BufferComparator(BufferComparator&&) = default;
 
-  BufferComparator(const Shape& shape, const HloModuleConfig& config,
-                   double tolerance = 0.1);
+  BufferComparator(const Shape& shape, double tolerance = 0.1,
```
reedwm, line 33: The BufferComparator constructor is also called in `conv_algorithm_picker.cc` and `gemm_fusion_autotuner.cc`. Please modify these files to take into account `HloModuleConfig` is no longer an argument.

Also mark the constructor as `explicit`

xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -24,6 +24,7 @@ limitations under the License.
 #include "xla/service/gpu/autotuner_util.h"
 #include "xla/service/gpu/backend_configs.pb.h"
 #include "xla/service/gpu/gemm_rewriter.h"
+#include "xla/service/gpu/variant_visitor.h"
```
reedwm, line 27: Add a `":variant_visitor",` dependency to the BUILD file [here](https://github.com/openxla/xla/blob/27aa23a8b48fbe805fa3c90b5652213b60dfd404/xla/service/gpu/BUILD#L1736)

xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -150,15 +212,6 @@ ENTRY main {
...
 TEST_P(GemmAlgorithmPickerTest, GetAlgorithmWithoutDevice) {
-  auto comp = backend()
-                  .default_stream_executor()
-                  ->GetDeviceDescription()
-                  .cuda_compute_capability();
-  if (comp.IsAtLeast(se::CudaComputeCapability::AMPERE)) {
-    GTEST_SKIP() << "Skipping this test for Ampere+ as it is supported and "
-                    "recommended with "
-                    "the Nvidia Volta+ GPUs.";
-  }
```
reedwm, line 215: Remove the blank line here

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -327,31 +330,45 @@ class GemmAutotuner {
...
-          result.mutable_failure()->set_kind(AutotuneResult::WRONG_RESULT);
-          result.mutable_failure()->mutable_reference_gemm()->set_algorithm(
-              *reference_algorithm);
+        // By default, autotuner does NOT really skip wrong results, but 
+        // merely prints out the above error message: this may lead to a 
+        // great confusion. When should_skip_wrong_results() is set to true,
+        // solutions with accuracy problems will be disqualified.
+        auto kind = AutotuneResult::WRONG_RESULT;
+        if (autotune_config_.should_skip_wrong_results()) {
+          kind = AutotuneResult::DISQUALIFIED;
```
reedwm, line 356: This is not causing any tests to fail, but what if the first algorithm is incorrect? Wouldn't this disqualify all the other algorithms since they are compared with the first, when it's the first algorithm we should disqualify?

I don't know an easy way to fix this unfortunately. @tdanyluk, any ideas?

reedwm (changes requested):
I ran into some internal lint/test issues when trying to merge this PR. Added line comments for these issues.

xla/service/gpu/gemm_algorithm_picker.cc
tdanyluk: Good question, I also asked about this in an earlier comment.
On ROCM, according to the author, the default (fallback) solution is always correct, so it sounds like this logic makes sense for their use-case.

If we really want to do this well, we could write a reference implementation that runs on CPU (such as the HLO interpreter) and compare the result with that. But the problem is that if the matrices are big, the runtime could be slow. We considered last year, that in some cases could try validating on some very sparse matrices, so that the reference computation would be fast, but it's not really easy, especially for fusions.

xla/service/gpu/buffer_comparator.h
pemeliya: Indeed, looks like these parts of conv_algorithm_picker and gemm_fusion_autotuner are excluded from build on ROCM.. Now I have fixed them on CUDA machine

xla/service/gpu/gemm_algorithm_picker.cc
pemeliya: Perhaps we can add an extended comment to debug_options.h where xla_gpu_autotune_level is defined ? Anyway, by default autotune_level is set to 4, hence no algorithms will be disqualified due to accuracy problems. If the user wants level 5, he should know what he is doing..

xla/service/gpu/gemm_algorithm_picker_test.cc
pemeliya: yes, I have tried running this test on V100 machine and removed **ASSERT_TRUE(changed);**. Not it seems to work fine: anyway, we do not care if GemmAlgorithmPicker has selected another algorithm: all that matters is that the number of algorithms used for autotuning during the second run (when **xla_gpu_autotune_level** is set to 5) is smaller that that of during the 1st run

xla/service/gpu/gemm_algorithm_picker.cc
dimitar-asenov: @tdanyluk is OOO.

@pemeliya Please add a more detailed explanation of the new level 5. In particular, explain that level 5 assumes a reference algorithm is correct.

Is this actually different to the code before this PR? Looking on the left it seems that we're anyway comparing to a reference? So in that sense, the PR doesn't change this behavior? Did I miss somethig?

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -327,31 +330,45 @@ class GemmAutotuner {
...
+              *reference_algorithm);
       }
     }  // for algorithms
 
     absl::StatusOr<AutotuneResult> best =
         PickBestResult(results, gemm->ToString(), hlo_module_config);
     if (best.ok()) {
+      // Return a real algorithm ID if return_algo_index is false: 
+      // e.g., in case of legacy cublas tuning.
+      if (!return_algo_index) return best;
```
dimitar-asenov, line 370: Is this really necessary? It would be better if we always return a consistent value. Can you make this PR work with the mapping code below and remove the `return_algo_index` parameter? If not, could you please explain why?

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -197,7 +202,7 @@ class GemmAutotuner {
     };
 
     return GetBestAlgorithm<BlasLt::MatmulAlgorithm>(
-        gemm, algorithms, gemm_config.beta, tuned_func);
+        gemm, algorithms, gemm_config.beta, true, tuned_func);
```
dimitar-asenov, line 205: I would really prefer to remove this parameter (see other comment). If it truly has to stay, then please add `/*return_algo_index=*/` in front.

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -242,21 +246,15 @@ class GemmAutotuner {
...
-    TF_ASSIGN_OR_RETURN(best_algorithm,
-                        GetBestAlgorithm<se::blas::AlgorithmType>(
-                            gemm, algorithms, gemm_config.beta, tuned_func));
-    if (best_algorithm.has_gemm()) {
-      int alg_idx = best_algorithm.gemm().algorithm();
-      best_algorithm.mutable_gemm()->set_algorithm(algorithms[alg_idx]);
-    }
-    return best_algorithm;
+    return GetBestAlgorithm<se::blas::AlgorithmType>(
+                      gemm, algorithms, gemm_config.beta, false, tuned_func);
```
dimitar-asenov, line 250: I would really prefer to remove this parameter (see other comment). If it truly has to stay, then please add `/*return_algo_index=*/` in front.

dimitar-asenov (changes requested):
This PR does at least 2 things, possibly 3, each of which should be in separate PRs. Please break it down to make it easier to check and submit. This is in accordance Google's best practices guide: https://google.github.io/eng-practices/review/developer/small-cls.html

Possible break-down:
- There are many NFC changes that should be submitted first. Some examples are:
    - Inside `GetBestAlgorithm`, there is a change to the if-condition in the loop. It removes an `else` branch and instead adds a `continue` statement.
    - The `BufferComparator` class changes
    - Test file cleanups and restructuring in `gemm_algorithm_picker_test.cc`.
    - The above are just examples, there's likely more NFC changes.
- Adding level 5 autotuning and the new `gemm_rtol` flag. I'm not sure if these need to be 1 or 2 changes, but consider breaking them down as much as possible. 

In general, please try to keep PRs very focused and minimal, which will make the approval process much smoother and quicker.

xla/service/gpu/gemm_algorithm_picker.cc
pemeliya: @dimitar-asenov, sorry for a delay in response. Yes, I have added a more detailed description into debug_options.cc of that new level 5. As I noted in the description of that PR, currently we have the autotune  level 4 set by default. Yes, we are still comparing to a reference solution, but solutions with numerical problems **are NOT discarded** with level 4: the autotuner just prints out a warning. The rationale behind is that the reference solution might **itself be wrong**, therefore the BufferComparator test failure is merely the message for developers that there is something wrong with the underlying math library. 

Yet, in our case, on ROCM, we are sure that the first reference solution is correct (we are specifically sort the list of solutions in such a way that the default one is always on top), therefore with level 5 we would like to really discard numerically wrong solutions.

xla/service/gpu/gemm_algorithm_picker.cc
pemeliya: yes, the point is, in case of cublas-lt, we really need to get **an index** of the best solutions from the list. Later on, this algorithm_idx is used to get the actual solution [here](https://github.com/ROCm/xla/blob/01c6ba775e824415b3117a6cdf92d57518b75cc7/xla/service/gpu/runtime/gpublas_lt_matmul_thunk.cc#L145). The reason for this is because, for cublas-lt, the algorithm is essentially **an opaque object** which cannot be stored inside an HLO.

In constast, for cublas, we just take the actual algorithm since cublas' algorithm is just a 32-bit integer. Previously, we had a double conversion in algorithm picker: first from index to solution, and then from solution back to index (in case of cublas-lt). And I just simplified this path.

pemeliya:
> This PR does at least 2 things, possibly 3, each of which should be in separate PRs. Please break it down to make it easier to check and submit. This is in accordance Google's best practices guide: https://google.github.io/eng-practices/review/developer/small-cls.html
> 
> Possible break-down:
> 
> * There are many NFC changes that should be submitted first. Some examples are:
>   
>   * Inside `GetBestAlgorithm`, there is a change to the if-condition in the loop. It removes an `else` branch and instead adds a `continue` statement.
>   * The `BufferComparator` class changes
>   * Test file cleanups and restructuring in `gemm_algorithm_picker_test.cc`.
>   * The above are just examples, there's likely more NFC changes.
> * Adding level 5 autotuning and the new `gemm_rtol` flag. I'm not sure if these need to be 1 or 2 changes, but consider breaking them down as much as possible.
> 
> In general, please try to keep PRs very focused and minimal, which will make the approval process much smoother and quicker.

OK, sure, I will think about how to split up this PR. The background history was that, originally it was just GemmAlgorithmPicker and BufferComparator changes, but then I was also asked to provide a test for that new features which resulted in a small gemm_algorithm_picker_test refactoring

pemeliya:
I have created two NFC PRs from that one: for GemmAlgorithmPicker and BufferComparator. Once those are merged, I will rebase this PR to keep only new features

pemeliya:
I have rebased the PR, now there shall only be new xla_gpu_autotune_level and xla_gpu_autotune_gemm_rtol changes

dimitar-asenov (approved):

dimitar-asenov:
Thank you! LGTM.

golechwierowicz (approved):

PR #13450, opened 2024-06-06
Fold reduce_window + slice[-1] to reduce + reshape

apivovarov:
In some cases framework can generate reduce_window + slice instead of simple reduce.

This PR extends [existing simplification](https://github.com/openxla/xla/blob/main/xla/service/algebraic_simplifier.cc#L7756) of `reduce_window` to `reduce` for the following patterns:

```
// Example 1 (one reduce dimension):
r = s32[2,8] reduce-window(s32[2,8] p, c), window={size=1x8 pad=0_0x7_0}
s = s32[2,1] slice(r), slice={[0:2], [7:8]}
// Can be folded to:
r = s32[2] reduce(s32[2,8] p, c), dimensions={1},
s = s32[2] reshape(r)
```
```
// Example 2 (two reduce dimensions):
p = s32[3,4,2]
r = s32[3,4,2] reduce-window(p, c), window={size=1x4x2 pad=0_0x3_0x1_0}
s = s32[3,1,1] slice(r), slice={[0:3], [3:4], [1:2]}
// Can be folded to:
r = s32[3] reduce(p, c), dimensions={1,2},
s = s32[3,1,1] reshape(r)
```

@DavidNorman @blakehechtman Could you review it?

### DIFF
--- a/xla/service/algebraic_simplifier.cc
+++ b/xla/service/algebraic_simplifier.cc
@@ -6568,6 +6568,88 @@ absl::Status AlgebraicSimplifierVisitor::HandleSlice(HloInstruction* slice) {
     }
   }
 
+  if (HloInstruction * reduce_window;
+      options_.enable_window_reduce_to_reduce_replacement() &&
+      hlo_instruction_utils::IsUnstridedSlice(slice) &&
+      Match(slice, m::Slice(m::ReduceWindow(&reduce_window).WithOneUse()))) {
+    // A reduce_window with window pad + slice[:,-1] can be expressed as
+    // reduce + reshape if all dimensions either have a window size of one or
+    // the entire dimension. No stride or dilation are expected. reduce_window
+    // pad should be present to make output shape equal to input shape.
+    // slice_limit[dim] should be equal to reduce_window shape[dim].
+    // slice_limit[dim] - slice_start[dim] should be equal to 1 for reduced dim
+    //
+    // The reshape is a bitcast since it adds one-sized dimensions. Often
+    // these ones are immediately removed as well with another reshape. The
+    // implementation of reduce tends to be slightly more efficient at
+    // reducing entire dimensions compared to reduce window.
+    //
+    //// Example 1:
+    // r = s32[2,8] reduce-window(s32[2,8] p, c), window={size=1x8 pad=0_0x7_0}
+    // s = s32[2,1] slice(r), slice={[0:2], [7:8]}
+    //// Can be folded to:
+    // r = s32[2] reduce(s32[2,8] p, c), dimensions={1},
+    // s = s32[2] reshape(r)
+    //
+    //// Example 2:
+    // p = s32[3,4,2]
+    // r = s32[3,4,2] reduce-window(p, c), window={size=1x4x2 pad=0_0x3_0x1_0}
+    // s = s32[3,1,1] slice(r), slice={[0:3], [3:4], [1:2]}
+    //// Can be folded to:
+    // r = s32[3] reduce(p, c), dimensions={1,2},
+    // s = s32[3,1,1] reshape(r)
+    auto effective_reduce_dims = [&] {
+      auto& window = reduce_window->window();
+      // reduce_window should have padding, but no Strides and Dilation
+      if (window_util::HasStride(window) || window_util::HasDilation(window) ||
+          !window_util::HasPadding(window)) {
+        return DimensionVector{};
+      }
+      auto rank = reduce_window->shape().dimensions_size();
+      auto& slice_starts = slice->slice_starts();
+      auto& slice_limits = slice->slice_limits();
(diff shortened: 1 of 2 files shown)


penpornk:
@blakehechtman Could you please help review this or assign the appropriate reviewer? Thank you!

xla/service/algebraic_simplifier.cc
```diff
@@ -6487,6 +6487,91 @@ absl::Status AlgebraicSimplifierVisitor::HandleSlice(HloInstruction* slice) {
...
+    // r = s32[3] reduce(p, c), dimensions={1,2},
+    // s = s32[3,1,1] reshape(r)
+    auto effective_reduce_dims = [&] {
+      auto& window = reduce_window->window();
+      // reduce_window should have padding, but no Strides and Dilation
+      if (window_util::HasStride(window) || window_util::HasDilation(window) ||
+          !window_util::HasPadding(window)) {
+        return DimensionVector{};
+      }
+      auto rank = reduce_window->shape().dimensions_size();
```
blakehechtman, line 6527: nit: pretty sure this is guaranteed for valid hlo

xla/service/algebraic_simplifier.cc
```diff
@@ -6487,6 +6487,91 @@ absl::Status AlgebraicSimplifierVisitor::HandleSlice(HloInstruction* slice) {
...
+        auto reduce_window_dim_size = reduce_window->shape().dimensions(i);
+        auto slice_dim_size = slice->shape().dimensions(i);
+        if (reduce_window_dim_size != slice_limits[i] ||
+            window.dimensions(i).padding_low() != slice_starts[i] ||
+            window.dimensions(i).padding_high() != 0) {
+          return DimensionVector{};
+        }
+        if (window_dim_size == 1 && reduce_window_dim_size == slice_dim_size &&
+            slice_starts[i] == 0)
+          continue;
```
blakehechtman, line 6546: nit: style guide says to use braces

xla/service/algebraic_simplifier.cc
```diff
@@ -6487,6 +6487,91 @@ absl::Status AlgebraicSimplifierVisitor::HandleSlice(HloInstruction* slice) {
...
+      auto& slice_starts = slice->slice_starts();
+      auto& slice_limits = slice->slice_limits();
+      DimensionVector reduce_dims;
+      for (auto i = 0; i < rank; ++i) {
+        auto window_dim_size = window.dimensions(i).size();
+        auto reduce_window_dim_size = reduce_window->shape().dimensions(i);
+        auto slice_dim_size = slice->shape().dimensions(i);
+        if (reduce_window_dim_size != slice_limits[i] ||
+            window.dimensions(i).padding_low() != slice_starts[i] ||
+            window.dimensions(i).padding_high() != 0) {
```
blakehechtman, line 6541: it isn't obvious to me why the presence of pad_high makes the transformation invalid given the other checks

blakehechtman (changes requested):

xla/service/algebraic_simplifier.cc
apivovarov: I made the following observations regarding window pad by looking at HLO examples which I have:
- pad_low is zero for all dims except of reduced dims. For reduced dims it is dim_size - 1
- pad_high is zero for all dims.

I tried to bake HLOs where pad_high is greater than zero - I can non active desired arithmetical result - where the last value is total sum of all values in reduced dimension.

If pad_high is non-zero then it is a different pattern and I need to look at it first to analyze what it is doing.
Because I only analyzed HLOs where pad_high is zero - I expect it to be zero for now.

If you have reduce-window+slice HLO examples which should be folded to simple reduce and they have pad_high != 0 in any of the window dimensions I can analyze them and make additional adjustments to the pattern code to make it more generic

xla/service/algebraic_simplifier.cc
apivovarov: fixed

xla/service/algebraic_simplifier.cc
apivovarov: I confirm - these two checks are done in hlo_verifier already. Removed them from here.

xla/service/algebraic_simplifier.cc
apivovarov: @blakehechtman, could you please take a look at the reply above when you have a moment? Thank you!

blakehechtman: that is fair, i was mostly imagining if the window was larger that the input size so the pad_high effecitvely did not do anything

blakehechtman (approved):

apivovarov:
Hi Penporn, I just want to check if this PR is good to merge. @penpornk

akuegel (approved):

PR #11514, opened 2024-04-15
Move cublasLT workspace allocation to buffer assignment

shawnwang18:
Move cublasLT workspace allocation to buffer assignment to support command buffer lowering

PR #13462, opened 2024-06-06
[ROCM][NFC] gpublas-lt refactoring after adding workspace and scratch allocator

pemeliya:
This PR https://github.com/openxla/xla/pull/11514 added workspace allocation to cublas-lt. Basically, it doubled the implementation of a number of functions in gpu/cu/hipblas-lt, i.e. now we have:

```
DoMatmul(..., std::optional<DeviceMemoryBase> workspace)
DoMatmul(..., std::optional<ScratchAllocator*> scratch_allocator)
DoMatmul(..., std::optional<DeviceMemoryBase> workspace, std::optional<ScratchAllocator*> scratch_allocator)
```
and the same holds for ```ExecuteOnStream```. This makes gpublas_lt interface barely readable.  The first two functions outlined above are just forwarding calls to the 3rd most generic one. Therefore, I do not see any need to implement these inside the derived classes, i.e. hip_blas_lt.h and cuda_blas_lt.h. Instead, forwarding can be handled in gpu_blas_lt.h interface.

@xla-rotation: could you please have a look?


tdanyluk (approved):
Thanks for the cleanup

PR #13467, opened 2024-06-06
[GPU] Relax the check for scheduled modules.

sergachev:
The current check is too strict as it does not allow running modules even with --xla_disable_all_hlo_passes.


tdanyluk (approved):

golechwierowicz:
Can you add a test to gpu_compiler_test.cc for this check relaxation? It's a functional change and now we have internal checks checking whether these are tested via unit tests.

golechwierowicz (approved):

sergachev:
done

golechwierowicz (approved):

golechwierowicz (approved):

PR #13482, opened 2024-06-06
[ROCm] Distinguish between NVIDIA and AMD gpu tags

hsharsha:


i-chaochen:
@ddunl @beckerhe  please take a look

ddunl (commented):
I think this looks good to me, I'm curious what the functional change is though in terms of tags/generated tests (if any). On a separate note, I think we should try to write tests for this at some point, no need to do it in this PR though.

hsharsha:
> I think this looks good to me, I'm curious what the functional change is though in terms of tags/generated tests (if any). On a separate note, I think we should try to write tests for this at some point, no need to do it in this PR though.

@ddunl  Previously backend_tags were processed by `prepare_nvidia_gpu_backend_data` and processed tags were rewritten back into backend_tags variable. This was then passed to `prepare_amd_gpu_backend_data`. As a result of preprocessing we lost `gpu` key and instead it was replaced by nvidia gpus. As a result any gpu backend tags that needed to be applied to amd gpus were lost.

hsharsha:
@ddunl Can you please take a look at this?

ddunl (approved):

ddunl:
I'm not quite spotting the bug, but there's an issue with e.g. `//xla/service/gpu/runtime:command_buffer_thunk_test_gpu_a100` where it should have a `requires-sm80-only` tag. You can see the expansion with `bazel query --output=build $TARGET_NAME`

ddunl:
The only ones that have this problem that I can tell are `//xla/service/gpu/runtime:dynamic_slice_thunk_test_gpu_{a100,v100}`, `//xla/service/gpu/runtime:dynamic_slice_thunk_test_gpu_{a100,v100}`, `//xla/service/gpu:cudnn_fused_conv_rewriter_test_gpu_a100`

hsharsha:
@ddunl the problem was happening in the merge phase of two dictionaries using | operator. nvidia_backend_tags was correctly populated for a100, v100 etc and the same key was overwritten by empty list in prepare_amd_backend_data

ddunl (approved):

hsharsha:
@ddunl I have added a new commit to this to handle disabled backends for AMD gpus

PR #13486, opened 2024-06-06
[ROCm] Add AMD_SERIALIZE_KERNEL environment for gpu_offloading_test

hsharsha:
This ensures kernel lauch are serialized.

### DIFF
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -3626,6 +3626,7 @@ xla_test(
     name = "gpu_offloading_test",
     srcs = ["gpu_offloading_test.cc"],
     backends = ["gpu"],
+    env = if_rocm_is_configured({"AMD_SERIALIZE_KERNEL" : "3"}),
     deps = [
         ":backend_configs_cc",
         ":horizontal_loop_fusion",


golechwierowicz (approved):

xla/service/gpu/BUILD
```diff
@@ -3626,6 +3626,7 @@ xla_test(
     name = "gpu_offloading_test",
     srcs = ["gpu_offloading_test.cc"],
     backends = ["gpu"],
+    env = if_rocm_is_configured({"AMD_SERIALIZE_KERNEL" : "3"}),
```
golechwierowicz, line 3629: Actually this change does not work in non-ROCM paths. They will evaluate to [] which is not a valid.

https://github.com/openxla/xla/blob/main/third_party/tsl/third_party/gpus/rocm/build_defs.bzl.tpl#L55-L63

Can you come up with another solution?

golechwierowicz (changes requested):
Requesting changes, breaks non-ROCM path

hsharsha:
> Requesting changes, breaks non-ROCM path

I am trying out a more fine grained wait for copy-start to wait until the computation is available before starting async copy-start

hsharsha:
This is addressed by https://github.com/openxla/xla/pull/13597/files

PR #13497, opened 2024-06-07
Swap inner and outer minor reduced dimension of tree reduction

ghost:
We observe larger n_div_k can improve tree reduction performance. Swap k and n_div_k if possible. See the below cases to get detailed information:
```bash
Case 0: 
ENTRY main {
      a = bf16[1024,50048] parameter(0)
      constant = bf16[] constant(0)
      ROOT reduce = bf16[1024] reduce(a, constant), dimensions={1}, to_apply=Add
})"
(1) k = 128, n_div_k = 391, runtime = 102us
(2) k = 391, n_div_k = 128, runtime = 132us
Even though (2) can be vectorized, but n_div_k is too small. (1) can get better performance than (2).

Case 1:
ENTRY main {
      a = bf16[10303,1024] parameter(0)
      constant = bf16[] constant(0)
      ROOT reduce = bf16[1024] reduce(a, constant), dimensions={0}, to_apply=Add
})"
(1) k = 64, n_div_k = 161, runtime = 33us
(2) k = 161, n_div_k = 64, runtime = 46us
ENTRY main {
      a = bf16[512,100005] parameter(0)
      constant = bf16[] constant(0)
      ROOT reduce = bf16[512] reduce(a, constant), dimensions={1}, to_apply=Add
})"
(3) k = 295, n_div_k = 339, runtime = 180us
(4) k = 339, n_div_k = 295, runtime = 183us
ENTRY main {
      a = bf16[1024,50257] parameter(0)
      constant = bf16[] constant(0)
      ROOT reduce = bf16[1024] reduce(a, constant), dimensions={1}, to_apply=Add
})"
(5) k = 140, n_div_k = 359, runtime = 104us
(6) k = 359, n_div_k = 140, runtime = 158us
Choose (1), (3) and (5) due to we want reduce more elements in the first reduction.

Case 2:
ENTRY main {
      a = bf16[1024,73984] parameter(0)
      constant = bf16[] constant(0)
      ROOT reduce = bf16[1024] reduce(a, constant), dimensions={1}, to_apply=Add
})"
(1) k = 256, n_div_k = 289, runtime=142us
(2) k = 289, n_div_k = 256, runtime=133us
Choose (2) due to k is almost as large as n_div_k and (2) can be vectorized.
```

### DIFF
--- a/xla/service/gpu/tree_reduction_rewriter.cc
+++ b/xla/service/gpu/tree_reduction_rewriter.cc
@@ -103,9 +103,11 @@ class ReductionRewriterVisitor : public DfsHloRewriteVisitor {
     bool maybe_vectorized = n_div_k % 2 == 0 && n % 2 == 0;
     // Swapping only affects row reduction vectorization.
     if (is_row_reduction && maybe_vectorized) {
-      // Swap if n_div_k is small enough or k is divisible by 2 also.
+      // Swap if n_div_k is small enough or k dim can be vectorized also.
       return n_div_k * 2 < k || k % 2 == 0;
     }
+    // If no vectorization, swapping can reduce memory store and the launch
+    // overhead of blocks.
     return true;
   }
 
--- a/xla/service/gpu/tree_reduction_rewriter.cc
+++ b/xla/service/gpu/tree_reduction_rewriter.cc
@@ -103,7 +103,7 @@ class ReductionRewriterVisitor : public DfsHloRewriteVisitor {
     bool maybe_vectorized = n_div_k % 2 == 0 && n % 2 == 0;
     // Swapping only affects row reduction vectorization.
     if (is_row_reduction && maybe_vectorized) {
-      // Swap if n_div_k is small enough even though vectorization may occur.
+      // Swap if n_div_k is small enough or k is divisible by 2 also.
       return n_div_k * 2 < k;
     }
     return true;
--- a/xla/service/gpu/tree_reduction_rewriter.cc
+++ b/xla/service/gpu/tree_reduction_rewriter.cc
@@ -104,7 +104,7 @@ class ReductionRewriterVisitor : public DfsHloRewriteVisitor {
     // Swapping only affects row reduction vectorization.
     if (is_row_reduction && maybe_vectorized) {
       // Swap if n_div_k is small enough or k is divisible by 2 also.
-      return n_div_k * 2 < k;
+      return n_div_k * 2 <= k;
     }
     return true;
   }
--- a/xla/service/gpu/tree_reduction_rewriter.cc
+++ b/xla/service/gpu/tree_reduction_rewriter.cc
@@ -98,15 +98,24 @@ class ReductionRewriterVisitor : public DfsHloRewriteVisitor {
     if (k > race_free_bound) {
       return false;
     }
-    // Rough conditions for row reduction vectorization, not mean that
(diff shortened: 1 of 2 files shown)


lingzhi98:
```bash
Case 3:
ENTRY main {
      a = bf16[1024,98304] parameter(0)
      constant = bf16[] constant(0)
      ROOT reduce = bf16[1024] reduce(a, constant), dimensions={1}, to_apply=Add
})"
(1) k = 256, n_div_k = 384, runtime=165us
(2) k = 384, n_div_k = 256, runtime=170us
Pre-swap and Post-swap can also vectorized, so we choose (1) because 384 > 256.
```
Update Case 3.

Note: All of the above performance data is collected on A100 40GB machine with MLIR emitter.

xla/service/gpu/tree_reduction_rewriter.cc
```diff
@@ -86,6 +86,29 @@ class ReductionRewriterVisitor : public DfsHloRewriteVisitor {
...
+    }
+    // Rough conditions for row reduction vectorization, not mean that
+    // vectorization will definitely occur.
+    bool maybe_vectorized = n_div_k % 2 == 0 && n % 2 == 0;
+    // Swapping only affects row reduction vectorization.
+    if (is_row_reduction && maybe_vectorized) {
+      // Swap if n_div_k is small enough or k is divisible by 2 also.
+      return n_div_k * 2 < k || k % 2 == 0;
+    }
+    return true;
```
cheshire, line 109: Why do we always swap in this case?

cheshire (changes requested):
Thanks a lot! 

Maybe it's possible to add some dedicated test cases, where we swap and don't swap the dimensions?

xla/service/gpu/tree_reduction_rewriter.cc
lingzhi98: Inner reduction has same amounts of load whether swap or not. In this case, vectorization can not enable and k is larger than n_div_k. After swapping, load spend same time, but the store can be reduced. And k is related about how many blocks will be launched. For example, assume row reduction input shape is [b, k, n_div_k], the launched blocks are O(b * k). After swapping, block launch overhead can be reduced also.

xla/service/gpu/tree_reduction_rewriter.cc
```diff
@@ -162,6 +195,10 @@ class ReductionRewriterVisitor : public DfsHloRewriteVisitor {
       }
     }
     uint64_t padded_n = n + minimum_padding;
+    if (ShouldSwapInnerAndOuterReducedMinorDimension(
+            best_k, padded_n / best_k, n, race_free_bound, is_row_reduction)) {
+      best_k = padded_n / best_k;
```
cheshire, line 200: Nit: the logic is a bit hard to follow, and the expression `padded_n / best_k` is referenced at least three times. Maybe we could create 2 variables to represent two new dimensions with self-descriptive names, and then std::swap them to make it obvious that they are swapped?

cheshire (changes requested):
Thanks again, one more nit

xla/service/gpu/tree_reduction_rewriter.cc
lingzhi98: Done. The code may be a little different from what you said, but I think it also meet what you want.

cheshire (approved):

PR #13526, opened 2024-06-07
Remove gpu_a100 and gpu_h100 from default nvidia gpu backends list

hmonishN:
For the xla_test targets  (e.g. test1 ) without backend field the targets are generated as test1_gpu_any, test1_gpu_a100, test1_gpu_h100, test1_gpu_amd_any, test1_cpu, and for test targets with just "gpu" backend fields, the targets are generated as test1_gpu_any, test1_gpu_a100, test1_gpu_h100. This causes repetitive tests execution in the CI job. This can be controlled with using test tags e.g. -xla_gpu_a100/-requires-sm-80/-requries-sm90-only,etc as in OSS CI but to run the tests on ampere gpu as default, we need to provide requires-gpu-nvidia (to include tests that have this tag mentioned in the target) and requires-sm-80, requires-sm-80-only for a100 test targets hence each test without backend field gets executed twice. 
Changing the default backend target to gpu_any and including the requires-sm-80,requires-sm-80-only tags ensures the test targets are executed once and a100/h100 specific tests are included using the tags (either xla_gpu_a100/xla_gpu_h100 or requires-sm-80/90/-only)

### DIFF
--- a/xla/tests/build_defs.bzl
+++ b/xla/tests/build_defs.bzl
@@ -17,11 +17,7 @@ NVIDIA_GPU_BACKENDS = [
 ]
 
 # The generic "gpu" backend includes the actual backends in this list.
-NVIDIA_GPU_DEFAULT_BACKENDS = [
-    "gpu_any",
-    "gpu_a100",
-    "gpu_h100",
-]
+NVIDIA_GPU_DEFAULT_BACKENDS = ["gpu_any"]
 
 AMD_GPU_DEFAULT_BACKENDS = ["gpu_amd_any"]
 


hmonishN:
Adding @ddunl

ddunl:
I think this change as is is not really possible to merge because our internal infra depends on this and functions differently. I agree with the spirit of the change though, because we also do these extra executions on our own external CI. I'll have to think about what a change that isn't disruptive to what happens internally would look like that would let us accomplish the same goal

ddunl:
I'm curious if there's a `bazel query` we could use to filter the targets, I think what we want to know is for all invocations of `xla_test`, what is the set of tests which have `backends` `gpu_any` and `gpu_$INSERT_GPU`, and then deduplicate those in the target patterns we run. Obviously this is less ideal than making some source level change, but I'd also argue that we want the default set of backends to be more than `gpu_any` in the first place

hmonishN:
I did try to use bazel query to get a superset of tests with tags gpu_any and gpu_a100 (for running on ampere) using intersect, i.e. bazel query 'attr("tags", "xla_gpu_any", //xla/...) intersect attr("tags", "xla_gpu_a100", //xla/...)' but this didn't give any result. Even if we use bazel query to filter out tests which are tagged with either one or both of gpu_any and gpu_a100 we will still be getting repetitive generated targets for default backends.

hmonishN:
I think having default backends as more generic i.e cpu and gpu makes sense and having different gpu vendor specific list with more detailed versions of gpus is helpful, if we make gpu versions as part of default gpu backend lisr, the list is just going to keep growing (if more nvidia/amd gpus version will be added) and more repetitive targets will be generated. By default we shouldn't be generating gpu vendor versions specific targets, we could though generate gpu vendor specific targets.

ddunl:
I'll let Henning be the tie breaker here. I think if we do this we should at least add back a100 and h100 to the targets which would have those backends removed after this change

sergey-kozub:
There is one important point to consider:

Every test with "gpu" backend generates several targets with suffixes like "_gpu_v100", "_gpu_a100", "_gpu_h100" and "_gpu_any". What is important to know is that all these tests run with the same architecture - the one of the machine the tests are running on.

I.e. if you run all of the above tests on Ampere (for example, when doing "blaze test //xla/service/gpu/..."), the same test will be run 4 times - which is completely unnecessary. What is more confusing, that even "_gpu_h100" will work on Ampere (will not fail).

I propose, instead of having 4 targets for each test (where only the tags are different, essentially), to have a single target with 4 tags. Then the CI could be set up to filter by those tags.

I'm working on a PR that introduces a new macro, "xla_gpu_test", which should make the transition easier, but I'll need your help from the Google side, as most of the "xla_test" macro is not visible in the OSS.

ddunl:
I think the problem is it's not obvious from my perspective how we could get this to cooperate with the way the google testing infra works internally. To me it seems they *must* be separate targets internally. We could have the internal/external xla_test implementations diverge, but I've been trying to work in the opposite direction. I do agree that right now there's a problem with running multiple copies of the same test on different hardware though.

I think the situation you describe where it the h100 tests are working on ampere we can solve in configure.py by generating a more opinionated set of tag filters based on the compute capability. Not running copies of tests is trickier, but I also think that can be solved in the `xla_test` implementation. (Also looking forward to hearing from @beckerhe on this when he's back from OOO next week)

sergey-kozub:
The worst of it is the "_gpu_any" suffix.
If a test exposes a bug which only manifests on Ampere, but doesn't manifest on Pascal/Volta, then using this suffix would result in a flaky test, as google CI infra will run it on _some_ available hardware, without guaranteeing the architecture.

I believe "_gpu_any" should be deprecated, and we should use a set of architecture-specific suffixes (p100|v100|a100|h100) which have their corresponding "requires-gpu-smXX-only" tags.
What do you think? Would that work?

ddunl:
That sounds good to me, it definitely seems preferable to have each target have a concrete set of targets to run on. I'll have to look into the origin of the gpu_any tag as I just kept it when merging parts of the internal `xla_test` and the external `xla_test`.

sergey-kozub:
I'll send a PR on Monday with the `xla_gpu_test` macro as I envision it (open to discussion).
I just need a few more hours to complete it and write a description.

sergey-kozub:
Please take a look at PR#13836
We can continue the discussion there.

dimitar-asenov:
I'm closing this PR since we can't really merge it as is. Further discussion in [PR# 13836](https://github.com/openxla/xla/pull/13836).

PR #13542, opened 2024-06-08
[NVIDIA] Change DCE to replay control deps for GTE-fusion simplification.

Tixxx:
We have seen a case when some optimization pipelines(ie tuple simplifier) add control deps to a GTE output of a fusion kernel to preserve execution order, DCE's un-used GTE simplification logic will error out trying to remove the GTE. Instead of replacing and removing, we can use ReplaceInstruction with relay_control_dependency set to true to move the control deps to the fusion instruction.


akuegel (approved):
Thanks for the fix.

PR #13555, opened 2024-06-09
Fix _xla_send_recv_validation in collective pipeliner

shraiysh:
This patch fixes the value of `_xla_send_recv_validation` attribute in
collective-pipeliner. Collective pipeliner peels one iteration outside
loop and updates the initialization variable of the loop. This means
that the trip count of the loop reduces.

More details about this are added as docstrings.


frgossen (changes requested):
How does this relate to your previous PR 12892? What aspect does it fix?

xla/service/collective_pipeliner.cc
```diff
@@ -1197,6 +1199,94 @@ HloInstruction* CreateZero(HloComputation* comp, const Shape& shape,
...
+      send_recv_validation_attr = std::vector<Interval>(n_pairs, {1, 0});
+      send_recv_validation_attr.front().first = 0;
+    }
+  } else {
+    return absl::UnimplementedError(
+        "Only forward and backward pipelining are handled for the "
+        "_xla_send_recv_validation attribute for collective permutes");
+  }
+
+  std::string send_recv_validation_attr_str =
```
frgossen, line 1272: This appears in multiple PRs. Can you extract it as a function and reuse it?

frgossen (changes requested):

shraiysh:
> How does this relate to your previous PR 12892? What aspect does it fix?

This is not about fixing a logical issue in the code. This fixes the value of the attribute when collective pipeliner is run. Collective pipeliner peels one iteration, where the peeled collective ends up before (or after) the loop if the pipelining direction is backward (or forward). This means that there is one less iteration and so the "valid iterations" should reflect that change. The peeled collective should also have a proper annotation indicating which device should run it.

On a related note, can you add @bixia1 as a review too? I think it would be good if she could check the testcases and give a go ahead for this fix.

shraiysh:
I think I need to fix the logic in this a bit. Converting to draft for sometime while I figure this out.

shraiysh:
This is ready for review now @bixia1 @frgossen

frgossen (approved):
Thanks!

frgossen:
This needs another rebase to pass the internal CI

shraiysh:
>This needs another rebase to pass the internal CI

Done. Thanks for the review!

bixia1 (approved):

shraiysh:
@frgossen any ideas why this isn't going through? The CI build failure seems unrelated.

akuegel (approved):

akuegel:
It looks like somehow the approval of @bixia1 didn't count. I approved again.

PR #13597, opened 2024-06-11
Activation offloading dependency fix: insert a wait

zhenying-liu:
To ensure proper synchronization for the asynchronous copy, this CL makes the other stream wait for the completion of the operation (res_3) in the main stream.
  %param_1 = f32[1024]{0} parameter(1)
  %param_0 = f32[1024]{0} parameter(0)
  %res_3 = f32[1024]{0} fusion(%param_1, %param_0), kind=kInput, calls=mul
  %copy-start = (f32[1024]{0:S(5)}, f32[1024]{0}, u32[]) copy-start(f32[1024]{0} %res_3)

### DIFF
--- a/xla/service/gpu/execution_stream_assignment.cc
+++ b/xla/service/gpu/execution_stream_assignment.cc
@@ -43,7 +43,6 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
   // each invocation of `async-start` will result in the target computation
   // being assigned a new `ExecutionStreamId`.
   ExecutionStreamId next_stream_id = ExecutionStreamId(1);
-  ExecutionStreamId copy_start_stream_id = ExecutionStreamId(0);
 
   // Each `Pending` item represents an `HloComputation` that needs to be
   // processed. We start with the entrypoint and add callees as we discover
@@ -74,13 +73,17 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
   // Assigns source and destination streams to an instruction and records it in
   // async_instructions_.
   auto assign_async_execution_streams =
-      [&](HloInstruction* instruction, ExecutionStreamId source_stream_id,
-          ExecutionStreamId destination_stream_id) {
+      [&](HloInstruction* instruction, ExecutionStreamId source_stream_id) {
         AsyncExecutionStreamIds streams;
         streams.source_stream_id = source_stream_id;
-        streams.destination_stream_id = destination_stream_id;
+        streams.destination_stream_id = next_stream_id;
 
         CHECK(async_instructions_.try_emplace(instruction, streams).second);
+
+        next_stream_id++;
+        if (next_stream_id.value() > options.number_of_execution_streams) {
+          next_stream_id = ExecutionStreamId(1);
+        }
       };
 
   while (!queue.empty()) {
@@ -91,20 +94,10 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
     // instructions. Asynchronous instructions will be handled afterwards.
     for (HloInstruction* instruction : pending.node->instructions()) {
       if (instruction->IsAsynchronous()) continue;
-      // The copy-start operation can only be in the main stream (stream 0).
       if (instruction->opcode() == HloOpcode::kCopyStart) {
-        ExecutionStreamId destination_stream_id = pending.stream_id;
-        CHECK_EQ(pending.stream_id, ExecutionStreamId(0));
-        if (copy_start_stream_id == ExecutionStreamId(0)) {
-          copy_start_stream_id = next_stream_id;
-          next_stream_id++;
-          if (next_stream_id.value() > options.number_of_execution_streams) {
-            next_stream_id = ExecutionStreamId(1);
-          }
-        }
-        destination_stream_id = copy_start_stream_id;
-        assign_async_execution_streams(instruction, pending.stream_id,
-                                       destination_stream_id);
+        // CopyStart is morally an async instruction, let us treat it
+        // as an async instruction.
+        assign_async_execution_streams(instruction, pending.stream_id);
       } else {
         CHECK(sync_instructions_.try_emplace(instruction, pending.stream_id)
                   .second);
@@ -120,11 +113,7 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
         CHECK_EQ(callsite.instruction()->opcode(), HloOpcode::kAsyncStart);
         enqueue_called_computations(callsite, next_stream_id);
         assign_async_execution_streams(callsite.instruction(),
-                                       pending.stream_id, next_stream_id);
-        next_stream_id++;
-        if (next_stream_id.value() > options.number_of_execution_streams) {
-          next_stream_id = ExecutionStreamId(1);
-        }
+                                       pending.stream_id);
       } else {
         // Synchronous calls will result in the called computations being
         // invoked using the same `ExecutionStreamId`.
--- a/xla/service/gpu/execution_stream_assignment.cc
+++ b/xla/service/gpu/execution_stream_assignment.cc
@@ -43,6 +43,7 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
   // each invocation of `async-start` will result in the target computation
   // being assigned a new `ExecutionStreamId`.
   ExecutionStreamId next_stream_id = ExecutionStreamId(1);
+  ExecutionStreamId copy_start_stream_id = ExecutionStreamId(0);
 
   // Each `Pending` item represents an `HloComputation` that needs to be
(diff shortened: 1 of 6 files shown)


hsharsha:
@zhenying-liu  Why is this still a draft?

zhenying-liu:
> @zhenying-liu Why is this still a draft?

I added the check to see if all the previous async events for copy-start/copy-done were completed. But I found 1 of 3 unit tests in gpu_offloading_tests.cc failed due to 2 unfinished events. I didn't find time to debug it yet.

hsharsha:
It is passing on ROCm side though with this fix. I was thinking of using WaitForStreamsThunk with copy_start operand instruction 
for annotation (in above example it is %fusion) and source_stream for copy_start stream and wait_for_stream a.k,a destination as operand instruction stream ( in the above example GetSyncStream(fustion_instr) )
```
AddThunkToThunkSequence(std::make_unique<WaitForStreamsThunk>(
              Thunk::ThunkInfo::WithProfileAnnotation(instr),
              streams.source_stream_id, streams.destination_stream_id))
```

zhenying-liu:
@hsharsha Adding the WaitForStreamsThunk is more traceable. I think it's a good idea! I plan to work on this PR tomorrow. 
I have a case for the weight offloading. When the source of copy-start is not an instruction calculated by the main stream, we don't need to insert this wait. We need to differentiate these two cases.
Below is the example:
```
HloModule jit__lambda_, is_scheduled=true, entry_computation_layout={(s32[4,1]{1,0:S(5)})->s32[4,1]{1,0}}, num_partitions=4, frontend_attributes={fingerprint_before_lhs="148a81385680930413c9450e8e5a42c8"}

ENTRY main.5_spmd {
 param.1 = s32[4,1]{1,0:S(5)} parameter(0), sharding={devices=[2,2]<=[4]}, metadata={op_name="x"}
 copy-start.1 = (s32[4,1]{1,0}, s32[4,1]{1,0:S(5)}, u32[]) copy-start(param.1), backend_config={"operation_queue_id":"1","wait_on_operation_queues":[],"force_earliest_schedule":false}
 ROOT copy-done.1 = s32[4,1]{1,0} copy-done(copy-start.1)
}
```

zhenying-liu:
> It is passing on ROCm side though with this fix. I was thinking of using WaitForStreamsThunk with copy_start operand instruction for annotation (in above example it is %fusion) and source_stream for copy_start stream and wait_for_stream a.k,a destination as operand instruction stream ( in the above example GetSyncStream(fustion_instr) )
> 
> ```
> AddThunkToThunkSequence(std::make_unique<WaitForStreamsThunk>(
>               Thunk::ThunkInfo::WithProfileAnnotation(instr),
>               streams.source_stream_id, streams.destination_stream_id))
> ```

I can't add it as thunk with minor changes because the copy-start instruction isn't inherited from an async instruction. The copy-start instruction doesn't have the member function of GetSyncStream() to get the stream id.

golechwierowicz:
@ezhulenev can you take look?

xla/service/gpu/runtime/copy_thunk.cc
```diff
@@ -125,6 +125,11 @@ absl::Status DeviceToHostCopyThunk::ExecuteOnStream(
   TF_ASSIGN_OR_RETURN(
       se::Stream * stream,
       GetStreamForExecution(Thunk::execution_stream_id(), params));
+  // Wait for the source to be ready on device unless it's an input parameter
+  if (instr_->operand(0)->opcode() != HloOpcode::kParameter) {
```
ezhulenev, line 129: It is unsafe to access `instr_` (it should be removed from thunk), because nothing guarantees that HloModule will be alive at run time, it's accidental that it's usually available.

Also instead of `stream->WaitFor` in a thunk, it should a be a `WaitFor` thunk in a sequence that enforces execution order and adds necessery stream syncs. It's a bug in IrEmitterUnnested that CopyThunk was emitted without a stream wait thunk.

ezhulenev: Basically `copy-start` and `copy-done` lowering are responsible for adding stream wait thunks to guarantee that nested copy is correctly syncrhonized

zhenying-liu: Thanks @ezhulenev for pointing out the difference between `stream->WaitFor` and a `WaitFor` thunk. I'll rewrite the code to add a `WaitFor` thunk.

xla/service/gpu/execution_stream_assignment.cc
```diff
@@ -78,8 +91,24 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
     // instructions. Asynchronous instructions will be handled afterwards.
     for (HloInstruction* instruction : pending.node->instructions()) {
       if (instruction->IsAsynchronous()) continue;
-      CHECK(sync_instructions_.try_emplace(instruction, pending.stream_id)
-                .second);
+      // The copy-start operation can only be in the main stream (stream 0).
+      if (instruction->opcode() == HloOpcode::kCopyStart) {
+        ExecutionStreamId destination_stream_id = pending.stream_id;
+        CHECK_EQ(pending.stream_id, ExecutionStreamId(0));
+        if (copy_start_stream_id == ExecutionStreamId(0)) {
```
jaro-sevcik, line 98: It should not be necessary to assign all copy-starts to the same stream, how about getting rid of `copy_start_stream_id`?

xla/service/gpu/execution_stream_assignment.cc
```diff
@@ -78,8 +91,24 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
...
+          copy_start_stream_id = next_stream_id;
+          next_stream_id++;
+          if (next_stream_id.value() > options.number_of_execution_streams) {
+            next_stream_id = ExecutionStreamId(1);
+          }
+        }
+        destination_stream_id = copy_start_stream_id;
+        assign_async_execution_streams(instruction, pending.stream_id,
+                                       destination_stream_id);
```
jaro-sevcik, line 107: I believe you should always take next_stream_id as destination_stream_id, just like we do for other async instructions.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2405,6 +2405,19 @@ absl::Status IrEmitterUnnested::EmitCopyStartThunk(
...
+        ExecutionStreamAssignment::AsyncExecutionStreamIds streams,
+        stream_assignment.GetAsyncExecutionStreamIds(copy_start_instr));
+    // Insert a waitFor() thunk for asynchronous memcpy only when the source
+    // and destination stream IDs differ. If the IDs are the same, the memcpy
+    // operation is synchronous within that stream.
+    if (streams.destination_stream_id != streams.source_stream_id) {
+      AddThunkToThunkSequence(std::make_unique<WaitForStreamsThunk>(
+          Thunk::ThunkInfo::WithProfileAnnotation(copy_start_instr),
+          streams.destination_stream_id, streams.source_stream_id));
+    }
```
jaro-sevcik, line 2420: You need to set the right execution stream on the thunk, just like we do on other async instructions.

```
   thunk->set_execution_stream_id(streams.destination_stream_id);
```

The same needs to be done for DeviceToHostCopyThunk below.

xla/service/gpu/execution_stream_assignment.cc
jaro-sevcik: If you do this, then you should be able to add the code for bumping next_stream_id into the lambda so that the lambda automatically uses next_stream_id as destination stream and bumps next_stream_id.

xla/service/gpu/execution_stream_assignment.cc
zhenying-liu: Done.

xla/service/gpu/execution_stream_assignment.cc
zhenying-liu: Done.

xla/service/gpu/ir_emitter_unnested.cc
zhenying-liu: Done.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2405,6 +2405,19 @@ absl::Status IrEmitterUnnested::EmitCopyStartThunk(
         static_cast<int>(stream_executor::MemoryType::kHost)));
   }
   if (is_dst_host_memory) {
+    const ExecutionStreamAssignment& stream_assignment =
```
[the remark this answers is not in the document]
zhenying-liu: Done.

jaro-sevcik:
Looks good to me.

@ezhulenev , could you take another look, please?

ezhulenev (approved):

PR #13603, opened 2024-06-11
NVTX: name threads, CUDA devices and CUDA streams

olupton:
This aims to improve the profiling experience. These names are shown in the Nsight Systems UI.

Device names:
Screenshot 2024-06-10 at 14 52 37

Stream names:
Screenshot 2024-06-10 at 14 53 25

Thread names:
Screenshot 2024-06-10 at 14 54 04

This also provides a missing link between replica IDs in the HLO and the physical devices in the profile.

### DIFF
--- a/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
+++ b/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
@@ -24,6 +24,7 @@ limitations under the License.
 #include <sys/syscall.h>
 
 #include "nvtx3/nvToolsExt.h"
+#include "nvtx3/nvToolsExtCuda.h"
 #include "nvtx3/nvToolsExtCudaRt.h"
 #include "nvtx3/nvToolsExtPayload.h"
 
@@ -42,6 +43,7 @@ std::optional<uint32_t> GetCurrentThreadId() {
 }  // namespace
 
 namespace tsl::profiler {
+static_assert(std::is_pointer_v<CUstream>);
 static_assert(std::is_pointer_v<nvtxDomainHandle_t>);
 static_assert(std::is_pointer_v<nvtxStringHandle_t>);
 
@@ -59,6 +61,10 @@ void NameCurrentThread(const std::string& thread_name) {
 
 void NameDevice(int device_id, const std::string& device_name) {
   nvtxNameCudaDeviceA(device_id, device_name.c_str());
+}
+
+void NameStream(StreamHandle stream, const std::string& stream_name) {
+  nvtxNameCuStreamA(reinterpret_cast<CUstream>(stream), stream_name.c_str());
 }
 
 void RangePop(ProfilerDomainHandle domain) {
--- a/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
+++ b/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
@@ -20,7 +20,10 @@ limitations under the License.
 #include <string_view>
 #include <type_traits>
 
+#include <sys/syscall.h>
+
 #include "nvtx3/nvToolsExt.h"
+#include "nvtx3/nvToolsExtCudaRt.h"
 #include "nvtx3/nvToolsExtPayload.h"
 
 namespace tsl::profiler {
@@ -31,6 +34,14 @@ ProfilerDomainHandle DefaultProfilerDomain() {
   static ProfilerDomainHandle domain =
       reinterpret_cast<ProfilerDomainHandle>(nvtxDomainCreateA("TSL"));
   return domain;
+}
+
+void NameCurrentThread(const char* thread_name) {
+  nvtxNameOsThreadA(syscall(SYS_gettid), thread_name);
+}
+
+void NameDevice(int device_id, const char* device_name) {
+  nvtxNameCudaDeviceA(device_id, device_name);
 }
 
 void RangePop(ProfilerDomainHandle domain) {
--- a/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
+++ b/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
@@ -16,6 +16,7 @@ limitations under the License.
 
 #include <cstddef>
 #include <cstdint>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <type_traits>
@@ -25,6 +26,20 @@ limitations under the License.
 #include "nvtx3/nvToolsExt.h"
 #include "nvtx3/nvToolsExtCudaRt.h"
 #include "nvtx3/nvToolsExtPayload.h"
+
+namespace {
+// Get the ID of the current thread following the convention for
+// nvtxNameOsThreadA:
+// https://nvidia.github.io/NVTX/doxygen/group___r_e_s_o_u_r_c_e___n_a_m_i_n_g.html
+// This convention may not match the one in tsl::Env::GetCurrentThreadId().
+std::optional<uint32_t> GetCurrentThreadId() {
+#ifdef __linux__
+  return syscall(SYS_gettid);
+#else
+  return std::nullopt;
+#endif
+}
+}  // namespace
 
 namespace tsl::profiler {
 static_assert(std::is_pointer_v<nvtxDomainHandle_t>);
@@ -36,12 +51,14 @@ ProfilerDomainHandle DefaultProfilerDomain() {
   return domain;
 }
 
-void NameCurrentThread(const char* thread_name) {
-  nvtxNameOsThreadA(syscall(SYS_gettid), thread_name);
+void NameCurrentThread(const std::string& thread_name) {
+  if (std::optional<uint32_t> tid = GetCurrentThreadId(); tid.has_value()) {
+    nvtxNameOsThreadA(*tid, thread_name.c_str());
+  }
 }
 
-void NameDevice(int device_id, const char* device_name) {
-  nvtxNameCudaDeviceA(device_id, device_name);
+void NameDevice(int device_id, const std::string& device_name) {
+  nvtxNameCudaDeviceA(device_id, device_name.c_str());
 }
 
 void RangePop(ProfilerDomainHandle domain) {
--- a/third_party/tsl/tsl/profiler/lib/nvtx_utils.h
+++ b/third_party/tsl/tsl/profiler/lib/nvtx_utils.h
@@ -34,6 +34,12 @@ using ProfilerDomainHandle = ProfilerDomain*;
 // Get the "TSL" domain if NVTX profiling is enabled, otherwise null
 ProfilerDomainHandle DefaultProfilerDomain();
 
+// Assign a human-readable name to the current thread
+void NameCurrentThread(const char* thread_name);
+
+// Assign a human-readable name to the given local device
+void NameDevice(int device_id, const char* device_name);
+
 // Register a string with the profiler/NVTX implementation for faster use
 StringHandle RegisterString(ProfilerDomainHandle, const std::string&);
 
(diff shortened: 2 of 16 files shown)


golechwierowicz:
@cheshire can you take a look? I'm not super familiar with nsys profiler, and cannot tell whether this improves internal profiling experience or not.

cheshire (approved):

xla/stream_executor/gpu/gpu_stream.cc
```diff
@@ -26,6 +26,9 @@ limitations under the License.
 #include "xla/stream_executor/platform.h"
 #include "xla/stream_executor/stream.h"
 
+#if GOOGLE_CUDA
```
jbaiocchi, line 29: why not use nvtx_utils.h?

jbaiocchi (commented):
Thanks, this seems very useful. I'm concerned about portability and the interactions with the thread interface in Env.

third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
```diff
@@ -33,6 +36,14 @@ ProfilerDomainHandle DefaultProfilerDomain() {
   return domain;
 }
 
+void NameCurrentThread(const char* thread_name) {
+  nvtxNameOsThreadA(syscall(SYS_gettid), thread_name);
```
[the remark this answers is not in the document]
olupton: The use of `syscall` is based on the NVTX documentation: https://nvidia.github.io/NVTX/doxygen/group___r_e_s_o_u_r_c_e___n_a_m_i_n_g.html#gaeb7d5b25e1147fc4aaf1f3acb8e719d0

I agree that it's not portable, but this is in a file that is -- I believe -- only compiled for CUDA/Linux, so I believe it is _sufficiently_ portable.

`Env::GetCurrentThreadId` is based on https://en.cppreference.com/w/cpp/thread/thread/get_id: https://github.com/openxla/xla/blob/f8f2e60035b7c37ec567e69b4e04201d5c1f37e6/third_party/tsl/tsl/platform/default/env.cc#L243

Given that the point is to associate a name with the thread ID convention used by the NVTX implementation, I believe it is important to align with the convention in the documentation.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1006,6 +1007,20 @@ absl::Status BuildDistributedDevices(
...
+        auto suffix =
+            absl::StrFormat(":#global=%d,local=%d,process=%d,slice=%d#",
+                            device_proto.global_device_id(),
+                            device_proto.local_device_ordinal(), node.node_id(),
+                            device_proto.slice_index());
+        // Name the device.
+        tsl::profiler::NameDevice(device_proto.local_device_ordinal(),
+                                  ("Xla" + suffix).c_str());
+        // Name the thread that launches work on this device.
+        local_device->execute_thread()->Schedule(
```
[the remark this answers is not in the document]
olupton: I believe it would be possible to tell the profiler/NVTX implementation about the name used in the Env interface.

As it stands, it would require a more significant reorganisation of the XLA code to ensure that an informative name is available at thread creation time. The existing names are not very detailed or informative. Alternatively, it would require a new API to set the name later than thread creation time.

xla/pjrt/local_device_state.cc
```diff
@@ -59,34 +59,43 @@ LocalDeviceState::LocalDeviceState(se::StreamExecutor* executor,
...
+    fixed_size_pool_usage_streams_.emplace_back(
+        create_stream(absl::StrFormat("Fixed size pool #%d", i)));
   }
   external_ready_event_streams_.reserve(kNumExternalReadyEventStreams);
   for (int i = 0; i < kNumExternalReadyEventStreams; ++i) {
-    external_ready_event_streams_.emplace_back(create_stream());
+    external_ready_event_streams_.emplace_back(
+        create_stream(absl::StrFormat("External ready event #%d", i)));
   }
   execute_thread_ =
```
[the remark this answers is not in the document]
olupton: Note that the information that we are trying to add to the thread names is not available until after https://github.com/openxla/xla/blob/f8f2e60035b7c37ec567e69b4e04201d5c1f37e6/xla/pjrt/gpu/se_gpu_pjrt_client.cc#L995-L998

xla/stream_executor/gpu/gpu_stream.cc
olupton: The first argument of `nvtxNameCuStreamA` is a `CUstream`: https://nvidia.github.io/NVTX/doxygen/group___r_e_s_o_u_r_c_e___n_a_m_i_n_g.html#gacada16c62463b268f8ede952ec77f766, which is available in this context as a `GpuStreamHandle`.

Doing this in nvtx_utils.h would I believe either require promoting some kind of abstract stream handle into TSL, or resorting to `void*`/similar; I thought the current approach was pragmatic.

third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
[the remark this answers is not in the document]
jbaiocchi: OK, would it be possible to create a file-local GetCurrentThreadId() and guard the syscall for Linux only?
Also, in comments explain why Env::GetCurrentThreadId() is not compatible with NTVX and reference the documentation to help a future maintainer adding support for other platforms. Thanks

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1006,6 +1007,20 @@ absl::Status BuildDistributedDevices(
...
         gpu_device_ids[device_proto.local_device_ordinal()] = global_device_id;
+        // Assign some descriptive names for profiling tools.
+        auto suffix =
+            absl::StrFormat(":#global=%d,local=%d,process=%d,slice=%d#",
+                            device_proto.global_device_id(),
+                            device_proto.local_device_ordinal(), node.node_id(),
+                            device_proto.slice_index());
+        // Name the device.
+        tsl::profiler::NameDevice(device_proto.local_device_ordinal(),
+                                  ("Xla" + suffix).c_str());
```
jbaiocchi, line 1018: absl::StrCat("Xla", suffix)

xla/pjrt/gpu/se_gpu_pjrt_client.cc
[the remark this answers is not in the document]
jbaiocchi: Please expand the comment and document why the name cannot be set at creation time and needs to be overridden here.

third_party/tsl/tsl/profiler/lib/nvtx_utils.h
```diff
@@ -34,6 +34,12 @@ using ProfilerDomainHandle = ProfilerDomain*;
 // Get the "TSL" domain if NVTX profiling is enabled, otherwise null
 ProfilerDomainHandle DefaultProfilerDomain();
 
+// Assign a human-readable name to the current thread
```
jbaiocchi, line 37: Define the interface to take const std::string& and call c_str() within the implementation.

xla/stream_executor/gpu/gpu_stream.cc
jbaiocchi: The advantage would be to have one file that mediates the interaction with NVTX.

nvtx_utils.h already forward declares structs like ProfilerDomain, can we do something similar for CUstream?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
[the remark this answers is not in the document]
olupton: Done

third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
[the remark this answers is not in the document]
olupton: Done.

xla/stream_executor/gpu/gpu_stream.cc
olupton: I believe this is a little different, in that the concrete meaning of `ProfilerDomain` is opaque everywhere outside nvtx_utils. Here there is a `CUstream` that comes from inside XLA's GPU abstraction layers that needs to be routed to NVTX.

Note that it is impractical to ensure `GOOGLE_CUDA` is defined in all situations that `nvtx_utils.h` is included. That reality was part of the motivation for the introduction of `ProfilerDomain` in https://github.com/openxla/xla/pull/9896.

I pushed a separate commit trading type safety for NVTX confinement, but it's just glorified `void*`. Let me know if you prefer this.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
olupton: Done

third_party/tsl/tsl/profiler/lib/nvtx_utils.h
olupton: Done

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1015,6 +1016,23 @@ absl::Status BuildDistributedDevices(
...
+                            device_proto.slice_index());
+        // Name the device.
+        tsl::profiler::NameDevice(device_proto.local_device_ordinal(),
+                                  absl::StrCat("Xla", suffix));
+        // Name the thread that launches work on this device. This is deferred
+        // until after ExchangeTopologies has been called so the global device
+        // id and slice index are known. These are not available when the thread
+        // is created.
+        local_device->execute_thread()->Schedule(
+            [name = "XlaLauncher" + suffix] {
```
jbaiocchi, line 1033: absl::StrCat("XlaLauncher", suffix)

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1015,6 +1016,23 @@ absl::Status BuildDistributedDevices(
...
+              tsl::profiler::NameCurrentThread(name.c_str());
```
jbaiocchi, line 1034: Don't call c_str() here.

third_party/tsl/tsl/profiler/lib/nvtx_utils.h
```diff
@@ -34,6 +34,19 @@ using ProfilerDomainHandle = ProfilerDomain*;
...
+
+// Assign a human-readable name to the given local device
+void NameDevice(int device_id, const std::string& device_name);
+
+struct Stream;
+// Opaque handle to an execution stream
+using StreamHandle = Stream*;
+
+// Assign a human-readable name to the given execution stream
+void NameStream(StreamHandle stream, const std::string& stream_name);
```
jbaiocchi, line 48: This version looks good to me, it is consistent with the rest of the code in this file. Thanks!

jbaiocchi (changes requested):
Thanks for the changes. Just fix the call to NameCurrentThread and it's good to go.

xla/stream_executor/gpu/gpu_stream.cc
olupton: Resolving based on https://github.com/openxla/xla/pull/13603#discussion_r1644594297

xla/pjrt/gpu/se_gpu_pjrt_client.cc
olupton: Done

xla/pjrt/gpu/se_gpu_pjrt_client.cc
olupton: Done

jbaiocchi (approved):

xla/pjrt/gpu/BUILD
```diff
@@ -116,6 +116,7 @@ cc_library(
         "@tsl//tsl/platform:status",
         "@tsl//tsl/platform:statusor",
         "@tsl//tsl/profiler/lib:connected_traceme",
+        "@tsl//tsl/profiler/lib:nvtx_utils",
```
akuegel, line 119: It seems adding this dependency unconditionally here causes a linker error if a target links in both nvtx_utils and nvtx_utils_libtpu. I could avoid this problem when I added this dependency inside if_cuda guard and guarded the include and the new code, but not sure this is the solution you would choose yourself.
We will revert this for now.

Issue #13606, opened 2024-06-11
[XLA:GPU] cuDNN norm rewriter triggers UAF when running with ASAN


bchetioui:
When using ASAN, running the `CudnnNormRewriterTest.LayerNormTrain2D1` test fails due to a use-after-free:

```
[ RUN      ] CudnnNormRewriterTest.LayerNormTrain2D1
=================================================================
==4616==ERROR: AddressSanitizer: heap-use-after-free on address 0x50400098d028 at pc 0x7f86ca1a44aa bp 0x7ffe1c0ec870 sp 0x7ffe1c0ec868
READ of size 8 at 0x50400098d028 thread T0
    #0 0x7f86ca1a44a9 in xla::gpu::(anonymous namespace)::CudnnNormRewriterVisitor::MatchLayerNorm(xla::HloInstruction*) xla/service/gpu/cudnn_norm_rewriter.cc:1117
    #1 0x7f86ca19feb1 in xla::gpu::(anonymous namespace)::CudnnNormRewriterVisitor::HandleAdd(xla::HloInstruction*) xla/service/gpu/cudnn_norm_rewriter.cc:886
...
```

Reproduction command (assuming I got the OSS path right):

```
bazel test //xla/service/gpu:cudnn_norm_rewriter_test_gpu_a100 --config=asan --config=cuda --compilation_mode=opt
```

Assigning to you @philipphack since I know you worked on this---but please feel free to re-assign if there is someone better to look at this.

philipphack:
It looks like asan isn't recognized as a valid configuration.

`ERROR: Config value 'asan' is not defined in any .rc file`

bchetioui:
@philipphack sorry about that.

Someone proposed a way to do this for OSS bazel [here](https://github.com/google/xls/issues/241#issue-775544387). Reproducing the proposed addition to `.bazelrc` below:

```
build:asan --strip=never
build:asan --copt -fsanitize=address
build:asan --copt -DADDRESS_SANITIZER
build:asan --copt -O1
build:asan --copt -g
build:asan --copt -fno-omit-frame-pointer
build:asan --linkopt -fsanitize=address
```

(I looked up the internal configuration, but it seems like I can't just paste it. Please let me know if you can reproduce with that---thanks a lot for your help.)

philipphack:
Thanks! Resolved in #13639.

bchetioui:
Thank you for the fix @philipphack! Closing.

PR #13639, opened 2024-06-11
Fix UAF in Norm Rewriter

philipphack:
Resolves a use-after-free when matching and rewriting layer norm patterns. See #13606.


reedwm (approved):
I assume the issue is because one of the HloInstructions passed to MatchNormFactor previously would be deallocated in MatchNormFactor? Or perhaps because norm_factor->users() would be modified inside MatchNormFactor?

In any case, MatchNormFactor should never be called twice for the same custom_call as this is also functionally incorrect, so this looks good to me.

philipphack:
MatchNormFactor replaces norm_factor which leads to an illegal access during the next step of the loop.

PR #13646, opened 2024-06-12
[ROCM] gemm_rewriter: bugfixing supported datatypes combinations

pemeliya:
This PR https://github.com/openxla/xla/pull/12553 introduced some a regression bug by enabling CUDA-specific and ROCM-specific datatype combinations on both platforms.

### DIFF
--- a/xla/service/gpu/gemm_rewriter.cc
+++ b/xla/service/gpu/gemm_rewriter.cc
@@ -2040,7 +2040,7 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
         {ComputationType::kF64, DataType::kComplexDouble, PrimitiveType::C128,
          PrimitiveType::C128, DataType::kComplexDouble},
     };
-    if (IsCuda(gpu_version_) &&
+    if (/*IsCuda(gpu_version_) &&*/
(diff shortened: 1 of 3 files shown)


golechwierowicz (approved):

golechwierowicz:
Can you write a test so the regression does not happen in the future?

pemeliya:
> Can you write a test so the regression does not happen in the future?

Yes, actually we found it out by //xla/tests/dot_operation_test failures on ROCM however this test is not included in AMD community build due to long running times. I will try to extend some of the existing AMD community tests to include this one.

golechwierowicz:
Can we first add a test to AMD CI build and add a unit test in gemm_rewriter_test.cc to cover this path?

pemeliya:
> Can we first add a test to AMD CI build and add a unit test in gemm_rewriter_test.cc to cover this path?

Yes, I have added a subtest to matmul_test which is a part of AMD CI and verified that it fails without my fix. And also added gemm_rewriter subtest to check for f64/c64 dots (which are not supported by hipblas-lt yet and hence shall be rewritten to legacy rocblas calls)

golechwierowicz (approved):

PR #13663, opened 2024-06-12
[ROCm] Disable fusion MLIR for ROCM

hsharsha:


golechwierowicz (changes requested):
1. Please update description of the PR with the description of the issue this diff is trying to solve.
2. Why is the ROCM path even executing when in `xla/service/gpu/fusions/mlir_emitter_test_base.h` device is set up as a CUDA device?

PR #13687, opened 2024-06-12
Fix simplify_fp_conversions_test on Hopper

sergey-kozub:
"simplify_fp_conversions_test" is an e2e test that verifies the whole compilation pipeline (all passes).

Originally it was supposed to check that there are no two consecutive converts left in the resulting HLO.
However, on Hopper (where multiply for bf16 is supported), no two consecutive converts appear, so the resulting HLO is different.

The test is updated to verify the Hopper e2e resulting HLO.


golechwierowicz:
Can you add a short description to a PR?

cheshire:
@sergey-kozub welcome to the codebase :P

golechwierowicz (approved):

chsigg (approved):

PR #13721, opened 2024-06-13
Fix BUILD files to allow building //xla/service/gpu/... in OSS

sergey-kozub:
Currently the following doesn't work in OSS:
$ bazel build //xla/service/gpu/...

with the PR applied, one can now build all targets.


ddunl:
Approving but I'm suspicious that deleting the gunit_for_test_only will work. But I'll get the internal tests running to find out at least

ddunl (approved):

ddunl:
Yeah it won't quite work like this, but after https://github.com/openxla/xla/pull/13746 which is really a Copybara change, we can submit the `service/gpu/tests` change

ddunl:
I think now if you rebase we can submit the remaining changes

sergey-kozub:
Something went wrong after merging. Created a new PR#13798

PR #13763, opened 2024-06-14
Add host memory space to entry_computation_layout

zhenying-liu:
During the transformation of MLIR to HLO, the host memory space was being dropped. This commit addresses that issue by utilizing utility functions to retrieve the memory space and layout information. It then passes this information to the module's entry computation.
With the input parameters receiving the correct host memory space, the commit enables proper generation of copy-start and copy-done operations for pinned host-to-device memory copies.


zhenying-liu:
Close this one and will reopen a new one by our team member.

PR #13765, opened 2024-06-14
Convolution Layout Normalization for Parameter Operands

philipphack:
Modifies the layout normalization for convolutions to support the direct operation on parameters and constants.

This enables running TestConvAmaxF8 and TestConvReluAmaxF8 in cudnn_fused_conv_rewriter_test.cc.

### DIFF
--- a/xla/service/gpu/conv_layout_normalization.cc
+++ b/xla/service/gpu/conv_layout_normalization.cc
@@ -136,15 +136,7 @@ absl::StatusOr<std::optional<HloInstruction*>> UpdateLayoutForCudnnConvolution(
     const Shape& s = op->shape();
     Shape s_reordered =
         ShapeUtil::MakeShapeWithDescendingLayoutAndSamePhysicalLayout(s);
-    HloInstruction* normalized_op = op->mutable_operand(0);
-    HloInstruction* new_op;
-    if (normalized_op->shape() == s_reordered) {
-      new_op = normalized_op;
-    } else {
-      new_op = MakeBitcastHlo(op, s_reordered);
-      performed_normalization = true;
-    }
-    normalized_operands.push_back(new_op);
+    normalized_operands.emplace_back(MakeBitcastHlo(op, s_reordered));
   }
 
   // Avoid replacing the Custom Call with an identical copy.
--- a/xla/service/gpu/conv_layout_normalization_test.cc
+++ b/xla/service/gpu/conv_layout_normalization_test.cc
@@ -23,7 +23,15 @@ namespace xla {
 namespace gpu {
 namespace {
 
-using ConvolutionLayoutNormalizationTest = HloTestBase;
+class ConvolutionLayoutNormalizationTest : public HloTestBase {
+ public:
+  se::CudaComputeCapability GetCudaComputeCapability() {
+    return backend()
+        .default_stream_executor()
+        ->GetDeviceDescription()
+        .cuda_compute_capability();
+  }
+};
 
 TEST_F(ConvolutionLayoutNormalizationTest, BackwardInput) {
   const char* hlo = R"(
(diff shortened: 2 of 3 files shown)


xla/service/gpu/conv_algorithm_picker.cc
```diff
@@ -675,9 +675,17 @@ absl::StatusOr<AutotuneResult> GpuConvAlgorithmPicker::AutotuneOneConvRunner(
 
   if (reference_result->has_value()) {
     XLA_SCOPED_LOGGING_TIMER_LEVEL("BufferComparator::CompareEqual", 2);
-    BufferComparator comparator(runtime_arguments.rz_buffers.output_shape(),
-                                runtime_arguments.hlo_module_config);
     for (int i = 0; i < result_buffers.size(); ++i) {
+      Shape output_shape;
+      if (runtime_arguments.rz_buffers.output_shape().IsTuple()) {
```
golechwierowicz, line 680: Can you separate this if-else logic of calculating output shape to a separate function in anonymous namespace? Also comment on the behavior like e.g. here

https://github.com/openxla/xla/blob/53f1bd856cb9d25f895490ab793afe124304f457/xla/service/gpu/ir_emitter_triton.cc#L762-L798

xla/service/gpu/conv_layout_normalization.cc
```diff
@@ -136,7 +136,12 @@ absl::StatusOr<std::optional<HloInstruction*>> UpdateLayoutForCudnnConvolution(
     const Shape& s = op->shape();
     Shape s_reordered =
         ShapeUtil::MakeShapeWithDescendingLayoutAndSamePhysicalLayout(s);
-    HloInstruction* normalized_op = op->mutable_operand(0);
+    HloInstruction* normalized_op;
```
golechwierowicz, line 139: Ditto on separate function.

philipphack:
PTAL.

cheshire (changes requested):
1. Is it possible to add tests?
2. Maybe split into two CLs? One for layout normalization, one for conv_picker? Both probably require tests actually.

philipphack:
@cheshire I added tests. Let me know if you want to split the change.

cheshire:
Yes, layout normalization changes IMO should be split from conv algorithm picker.

philipphack:
The algorithm picker changes are now in #14111.

dimitar-asenov:
@cheshire Could you please take another look?

golechwierowicz:
@cheshire gentle ping

xla/service/gpu/conv_layout_normalization.cc
```diff
@@ -136,7 +146,7 @@ absl::StatusOr<std::optional<HloInstruction*>> UpdateLayoutForCudnnConvolution(
     const Shape& s = op->shape();
     Shape s_reordered =
         ShapeUtil::MakeShapeWithDescendingLayoutAndSamePhysicalLayout(s);
-    HloInstruction* normalized_op = op->mutable_operand(0);
+    HloInstruction* normalized_op = MaybeOperand0(op);
```
cheshire, line 149: Might be more readable to just inline: ` = instr->operand_count() > 0 ? instr->operand(0) : instr`

cheshire: Also maybe add a comment why we may or may not see a unary instruction here?

cheshire (changes requested):

xla/service/gpu/conv_layout_normalization.cc
philipphack: I added a comment. Function MaybeOperand0 was added following a previous [reviewer request](https://github.com/openxla/xla/pull/13765#discussion_r1639498852).

cheshire: I still don't really follow why do we follow the first operand, why it's OK not to follow it, and what are we trying to do.

cheshire (changes requested):

xla/service/gpu/conv_layout_normalization.cc
philipphack: @cheshire the logic for accessing the first operand of the convolution operands was introduced [here](https://github.com/openxla/xla/commit/e8d3f0e06cbf51aaadd034a7d18b4ccd8e4249ab). I think the idea is to compare the shape of the operand to that of the operand's operand and insert a bitcast if necessary. However, when the convolution directly operates on a parameter or constant (e.g. TestConvAmaxF8 or TestConvReluAmaxF8 in cudnn_fused_conv_rewriter_test.cc), the operand of the convolution has no operand. These tests were previously not executed because the header defining CUDNN_VERSION had been removed from cudnn_fused_conv_rewriter_test.cc.

cheshire:
Sorry I still don't follow. Could you expand the PR description significantly?

Specifically, what isn't working now, why is it not working, what is this PR doing to fix that, how is it going about this, and what would be the new behavior (& delta vs. the previous behavior).

NaiyerRizz:
Hi @philipphack 
Can you please look into the above given comment. 
Thank you

xla/service/gpu/conv_layout_normalization.cc
reedwm: IIUC, currently this logic (without this PR) implicitly assumes that `normalized_op = op->mutable_operand(0)` is a bitcast instruction. If the input to this bitcast is already in the right shape and layout, it just sets the new operand to this input (`new_op = normalized_op`). Otherwise, it inserts another bitcast on top of the first.

I think we shouldn't implicitly assume there is a bitcast at all. For one, this is not true for FP8 scales, which don't have an operand, let alone a bitcast operand, since the scales are parameters or constants. Second, this seems like a potentially unsafe assumption in general, even for non-FP8 convs.

Instead, let's not get the operand at all. Just bitcast `op` to the correct shape unconditionally. If we insert a redundant bitcast, some future pass will fix it. So I think you can remove lines 149 to 158 in this PR and replace it with: `HloInstruction* new_op = MakeBitcastHlo(op, s_reordered)`.

@cheshire let me know if my understanding of the current logic is incorrect.

reedwm (changes requested):

reedwm (approved):

PR #13771, opened 2024-06-14
Fix algebraic_simplifier for rsqrt

apivovarov:
Some algsimp patterns used `IsPositive()` function instead of using `IsNonNegative()` to check its operands.

As a result, the patterns did not work when the operand was >= 0. However, they should.

### Fix for pattern `rsqrt(B) * rsqrt(B) => 1/B`
Issue: Pattern did not work for B >= 0, for example when B=abs(x).
Solution: Fixed by checking that B isNonNegative
Validation: 
- If `B==0` the `result is inf`
- If `B>0` the `result > 0`
- If `B is inf` the `result is 0`
- If `B is nan` the `result is nan`

### Fix for pattern `rsqrt(pow(A, -2)) => A`
Issue: Pattern did not work for A >= 0, for example when A=abs(x).
Solution: Fixed by checking that A isNonNegative.
Validation (before and after the simplification): 
- If `A==0` the result is `0` 
- If `A>0` the `result > 0`
- If `A is inf` the `result is inf`
- If `A is nan` the `result is nan`

Additional fix: Since we know that A is non-negative we can use A directly without wrapping it with `abs()`.

### Fix for pattern `rsqrt(1/A)) => sqrt(A)`
Issue: Pattern did not work for A >= 0, for example when A=abs(x).
Solution: Fixed by checking that A isNonNegative
Validation (before and after the simplification): 
- If `A==0` the result is `0` 
- If `A>0` the `result > 0`
- If `A is inf` the `result is inf`
- If `A is nan` the `result is nan`

### DIFF
--- a/xla/service/algebraic_simplifier_test.cc
+++ b/xla/service/algebraic_simplifier_test.cc
@@ -9122,141 +9122,109 @@ TEST_F(AlgebraicSimplifierTest, ReduceAddIsCommutative) {
               GmockMatch(m::Reduce(m::Parameter(0), m::ConstantScalar(0))));
 }
 
-TEST_F(AlgebraicSimplifierTest, RsqrtOfRPower) {
+// rsqrt(pow(x, -2)) => x, for x >= 0
+TEST_F(AlgebraicSimplifierTest, RsqrtOfRPower2) {
   const char* kModuleStr = R"(
     HloModule m
     test {
-      p0 = f32[128,32,2,112]{3,2,1,0} parameter(0)
-      p1 = f32[32]{0} parameter(1)
-      p2 = f32[32]{0} parameter(2)
-      c0 = f32[] constant(0.001)
-      c1 = s64[] constant(1)
-      custom-call.1 = (f32[128,32,2,112]{3,2,1,0}, f32[32]{0}, f32[32]{0}) custom-call(p0, p1, p2, c0, c1), custom_call_target="__cudnn$batchNormalizationForwardTraining"
-      get-tuple-element.1 = f32[128,32,2,112]{3,2,1,0} get-tuple-element(custom-call.1), index=0
-      get-tuple-element.2 = f32[32]{0} get-tuple-element(custom-call.1), index=1
-      get-tuple-element = f32[32]{0} get-tuple-element(custom-call.1), index=2
-      c2 = f32[] constant(-2)
-      broadcast = f32[32]{0} broadcast(f32[] c2), dimensions={}
-      power = f32[32]{0} power(get-tuple-element, broadcast)
-      rsqrt = f32[32]{0} rsqrt(f32[32]{0} power)
-      ROOT tuple = (f32[128,32,2,112]{3,2,1,0}, f32[32]{0}, f32[32]{0}) tuple(get-tuple-element.1, get-tuple-element.2, rsqrt)
+      p0 = f32[1,32] parameter(0)
+      abs0 = f32[1,32] abs(p0)
+      c1 = f32[] constant(-2.0)
+      broadcast = f32[1,32] broadcast(c1), dimensions={}
+      power = f32[1,32] power(abs0, broadcast)
+      ROOT rsqrt = f32[1, 32] rsqrt(power)
     }
   )";
   TF_ASSERT_OK_AND_ASSIGN(auto m, ParseAndReturnVerifiedModule(kModuleStr));
-  default_options_.set_cudnn_batchnorm_forward_training_metadata(
-      "__cudnn$batchNormalizationForwardTraining");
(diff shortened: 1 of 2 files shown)


xla/service/algebraic_simplifier_test.cc
```diff
@@ -9122,141 +9122,109 @@ TEST_F(AlgebraicSimplifierTest, ReduceAddIsCommutative) {
...
-  EXPECT_EQ(FindInstruction(m.get(), HloOpcode::kRsqrt), nullptr);
-
-  auto computation = m->entry_computation();
-  auto root = computation->root_instruction();
-  EXPECT_EQ(root->opcode(), HloOpcode::kTuple);
-  EXPECT_EQ(root->operand(2)->opcode(), HloOpcode::kDivide);
-  EXPECT_EQ(root->operand(2)->operand(0)->opcode(), HloOpcode::kBroadcast);
-  EXPECT_EQ(root->operand(2)->operand(1)->opcode(),
-            HloOpcode::kGetTupleElement);
+  ASSERT_THAT(m->entry_computation()->root_instruction(),
```
golechwierowicz, line 9210: s/ASSERT_THAT/EXPECT_THAT
?

xla/service/algebraic_simplifier_test.cc
```diff
@@ -9122,141 +9122,109 @@ TEST_F(AlgebraicSimplifierTest, ReduceAddIsCommutative) {
...
+    test {
+      p0 = f32[1,32] parameter(0)
+      c1 = f32[] constant(-2.0)
+      broadcast = f32[1,32] broadcast(c1), dimensions={}
+      power = f32[1,32] power(p0, broadcast)
+      ROOT rsqrt = f32[1, 32] rsqrt(power)
+    }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(auto m, ParseAndReturnVerifiedModule(kModuleStr));
+  ASSERT_FALSE(AlgebraicSimplifier(default_options_).Run(m.get()).value());
```
golechwierowicz, line 9158: If you rely on implication that "module not run => IR unchanged" then this assertion should become an expectation. 

s/ASSERT_FALSE/EXPECT_FALSE
?

golechwierowicz (commented):
Thanks for the diff and detailed explanation! I really like the test simplification.

I made small comments about ASSERT v. EXPECT.

apivovarov:
_`ASSERT_*` yields a fatal failure and returns from the current function, while `EXPECT_*` yields a nonfatal failure, allowing the function to continue running._

When `Run(m.get()).value()` returns `false`, it means that no changes are made.

`EXPECT_FALSE` is needed if we want to continue test execution even if `EXPECT_FALSE` fails. In this specific PR, all three `*_NegativeTestCase`s do not have any additional checks after `ASSERT_FALSE(Run())`. 
Therefore, both `ASSERT_FALSE(Run())` and `EXPECT_FALSE(Run())` will work identically because it will be the last executed check.

If we add additional checks after `ASSERT_FALSE(Run())` in the future, we can convert `ASSERT_FALSE` to `EXPECT_FALSE` to ensure these additional checks run even if `Run()` returns `true`.

xla/service/algebraic_simplifier_test.cc
apivovarov: see my comment regarding `ASSERT_*`  vs  `EXPECT_*`

xla/service/algebraic_simplifier_test.cc
apivovarov: see my comment regarding `ASSERT_*`  vs  `EXPECT_*`

golechwierowicz:
> _`ASSERT_*` yields a fatal failure and returns from the current function, while `EXPECT_*` yields a nonfatal failure, allowing the function to continue running._
> 
> When `Run(m.get()).value()` returns `false`, it means that no changes are made.
> 
> `EXPECT_FALSE` is needed if we want to continue test execution even if `EXPECT_FALSE` fails.
> 
> In this specific PR, all three `*_NegativeTestCase`s do not have any additional checks after `ASSERT_FALSE(Run())`.
> 
> Therefore, both `ASSERT_FALSE(Run())` and `EXPECT_FALSE(Run())` will work identically because it will be the last executed check.
> 
> If we add additional checks after `ASSERT_FALSE(Run())` in the future, we can convert `ASSERT_FALSE` to `EXPECT_FALSE` to ensure these additional checks are executed even if `Run()` returns `true`.

Functionally they do work identically with a current diff, but syntactically there used to be a style guide recommendation (I must admit, I did not find it in OSS) to model expectations of the end state as EXPECT, and use ASSERT as invariant checks for a proper test set up. That being said I think it's quite outdated, and since the functionally there is no difference I'll LGTM this.

golechwierowicz (approved):

PR #13781, opened 2024-06-14
[GPU] Let the on-disk kernel compilation cache grow.

sergachev:


xla/debug_options_flags.cc
```diff
@@ -1739,14 +1739,11 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
...
-      debug_options->xla_gpu_kernel_cache_file(),
-      "Path to a file to cache compiled kernels. If the file doesn't exist "
-      "write the compilation cache of the first compiled HLO module into it."
-      "Once the file exists, further compilations will read it to reuse "
-      "the kernels, but not write it. This behavior may change later."));
+  flag_list->push_back(
+      tsl::Flag("xla_gpu_kernel_cache_file",
+                string_setter_for(&DebugOptions::set_xla_gpu_kernel_cache_file),
+                debug_options->xla_gpu_kernel_cache_file(),
+                "Path to a file to cache compiled kernels."));
```
tdanyluk, line 1746: Some explanation about the usage of the file would still be helpful.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2003,34 +2010,45 @@ absl::StatusOr<GpuCompiler::BackendCompileResult> GpuCompiler::CompileAndLink(
...
-        VLOG(2) << "Stored " << cache.entries_size() << " / "
-                << binaries_to_cache.size();
+      if (stored_kernel_count > 0) {
+        TF_RETURN_IF_ERROR(
+            tsl::WriteStringToFile(tsl::Env::Default(), resolved_path,
+                                   disk_cache.SerializeAsString()));
+        VLOG(2) << "Stored " << stored_kernel_count << " / "
+                << binaries_to_cache.size() << " kernels in the cache file.";
       }
     }
```
tdanyluk, line 2053: Maybe this could go to 1 or 2 separate function(s). We could also consider this for the lines 1985-2017, but I can't add a comment there.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2003,34 +2010,45 @@ absl::StatusOr<GpuCompiler::BackendCompileResult> GpuCompiler::CompileAndLink(
...
-        auto it = entries->find(name);
-        if (it == entries->end()) {
-          continue;
+      VLOG(2) << "Using " << loaded_kernel_count << " / "
+              << current_cache.entries_size() << " cached kernels.";
+    }
+    if (!binaries_to_cache.empty()) {
+      CompilationCacheProto disk_cache;
+      if (cache_file_exists) {
+        // Load the unmodified existing cache file to augment it.
```
tdanyluk, line 2022: I think that now the code has the following overall logic:
1. load full cache in compile_module_to_llvm_ir.cc/LoadCache
2. filter out the needed kernels when doing `results.kernel_compilation_cache =
        ir_emitter_context.kernel_cache().Export();`
3. read the full cache again, add the new results to it, and write it out in gpu_compiler.cc

Would it be possible to simplify this flow or at least try to move all the cache reading and writing functions to one file, maybe kernel_reuse_cache.cc?
(e.g. why filter kernels if we load the discarded ones again? But maybe it makes sense from some perspective)
Now the logic of the caching is a bit hard to follow for me, because we have some kernel caching logic "here and there".

(This could be done in a separate CL potentially, if that makes sense.)

tdanyluk (commented):
Thanks, I think it's great, but added some comments.

xla/debug_options_flags.cc
sergachev: done

xla/service/gpu/gpu_compiler.cc
sergachev: Moved this part into UpdateDiskKernelCache(). The other part does not look clearer as a separate function - I tried.

xla/service/gpu/gpu_compiler.cc
sergachev: Filtering is done to avoid linking unused kernels potentially present in the cache into the current executable. Another complexity for the overall flow is that at present not all compiled kernels are cacheable because of how llvm::SplitModule() works, I'll submit a patch there making it more convenient.
I'll draw a diagram and see what can be simplified.

xla/service/gpu/kernel_reuse_cache.h
```diff
@@ -88,6 +92,14 @@ class KernelReuseCache {
   absl::flat_hash_set<std::string> hits_;
 };
 
+// Add kernels to the cache file. Binaries are taken from binaries_to_cache,
+// all other kernel properties are taken from current_cache.
+// do_append makes an existing file be loaded first.
+absl::Status UpdateDiskKernelCache(
+    absl::string_view path, bool do_append,
+    const CompilationCacheProto& current_cache,
+    absl::Span<const KernelReuseCache::NamedBinary> binaries_to_cache);
```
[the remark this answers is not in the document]
sergachev: done

tdanyluk (approved):
thanks

PR #13799, opened 2024-06-14
Enable dot algorithms for AMD GPUs

hsharsha:


xla/service/gpu/dot_algorithm_support_test.cc
```diff
@@ -143,8 +159,23 @@ TEST_P(DotAlgorithmSupportTest, AlgorithmIsSupportedFromCudaCapability) {
...
+    GetRocmComputeCapability().gfx9_mi100_or_later();
+    if (TF_ROCM_VERSION < params.min_rocm_version &&
+        (params.input_storage_type == F8E5M2 || params.input_storage_type == F8E4M3FN) &&
+        params.output_storage_type == BF16) {
+       GTEST_SKIP() << "TODO: Unsupported F8 to BF16 in ROCm version < 6.2";
+    }
+    if (params.backend_restriction == BackendRestriction::kTritonOnly) {
+        GTEST_SKIP() << "TODO: Triton unsupported in ROCm";
+    }
+#endif //GOOGLE_CUDA
```
tdanyluk, line 177: Is it possible to do everything without ifdefs? Just get the gpu_compute_capability() and check if it holds a cuda or rocm capability.

xla/service/gpu/dot_algorithm_support_test.cc
```diff
@@ -59,35 +63,38 @@ struct Sizes {
...
+        sizes(std::get<5>(t)),
+        min_rocm_version(std::get<6>(t)) {}
 };
 
 std::string TestParamsToString(
     const TestParamInfo<TestParams::TupleType>& info) {
   const TestParams params(info.param);
   return absl::StrFormat(
-      "%s_with_input_%s_output_%s_from_cc_%d_%d_%s_c_%d_nc_%d",
+      "%s_with_input_%s_output_%s_from_cc_%d_%d_%s_c_%d_nc_%d_rocm_%d",
```
tdanyluk, line 90: Nit: the rocm version kindof belongs next to the cuda version.
Something like this:
%s_with_input_%s_output_%s_from_cc_%d_%d_rocm_%d_%s_c_%d_nc_%d

tdanyluk (commented):
Thanks it's good, just added 2 comments

hsharsha:
@tdanyluk  Can you please review, I have addressed your review comments. Thanks

xla/service/gpu/dot_algorithm_support_test.cc
dimitar-asenov: This comment is marked as resolved, but I still see a lot of `#ifdef GOOGLE_CUDA` and `#elif TENSORFLOW_USE_ROCM`. These should not be needed. Please check the compute capability using run-time methods, not macros.

dimitar-asenov (changes requested):

dimitar-asenov:
@hsharsha Tamas is OOO so I'll take care of merging this PR. I left one request for a fix and then this should be good to go. Please ping me here when it's ready.

xla/service/gpu/dot_algorithm_support_test.cc
hsharsha: @dimitar-asenov  I am not sure if I can get rid of those `#ifdefs`. In ROCm we do not store ROCM_VERSION as part of device descriptor unlike CUDA. We use TF_ROCM_VERSION which is defined in rocm_configs.h file. I plan to change this but in a different commit.

dimitar-asenov: @hsharsha Please see the discussion and solution in https://github.com/openxla/xla/pull/13479. Let's wait for that to land then you can start using the same solution.

dimitar-asenov: That PR has landed, so you should be able to use that solution here.

hsharsha: @dimitar-asenov  Thanks for pointing me to that. I have updated this PR on the same lines of #13479

xla/service/gpu/dot_algorithm_support_test.cc
```diff
@@ -102,11 +105,18 @@ class DotAlgorithmSupportTest
...
+          .default_stream_executor()
+          ->GetDeviceDescription();
+  }
+  se::GpuComputeCapability GetGpuComputeCapability() {
+    return GetDeviceDescription()
+        .gpu_compute_capability();
+  }
+
+  bool IsCUDA() {
+      return std::holds_alternative<se::CudaComputeCapability>(GetGpuComputeCapability());
```
dimitar-asenov, line 119: Please don't define this function. See comment below.

xla/service/gpu/dot_algorithm_support_test.cc
```diff
@@ -143,8 +153,25 @@ TEST_P(DotAlgorithmSupportTest, AlgorithmIsSupportedFromCudaCapability) {
       primitive_util::LowercasePrimitiveTypeName(params.output_storage_type),
       params.sizes.contracting_size, params.sizes.non_contracting_size);
 
-  if (GetCudaComputeCapability().IsAtLeast(params.min_cuda_capability.major,
-                                           params.min_cuda_capability.minor)) {
+  bool is_algorithm_supported = false;
+  if (IsCUDA()) {
```
dimitar-asenov, line 157: Please use this form:
```
if ( const auto* ccc = std::get_if<se::CudaComputeCapability>(GetGpuComputeCapability())) {
...
} else if ( const auto* rcc = std::get_if<se::CudaComputeCapability>(GetGpuComputeCapability())) {
...
}
```

This is both reasonably compact and doesn't assume that there is only Cuda and RoCM.

dimitar-asenov (changes requested):

xla/service/gpu/dot_algorithm_support_test.cc
dimitar-asenov: NP, I left a small comment and then I will merge this. Sorry for the long back and forth.

xla/service/gpu/dot_algorithm_support_test.cc
hsharsha: Submitted a new commit

dimitar-asenov (approved):

PR #13808, opened 2024-06-14
FP8 Windowed Einsums with Multiple All-Gather Dots

philipphack:
Enables FP8 windowed einsums with all-gathers that have multiple dot users by shifting the dequantization of the FP8 operands to the output of the while loop.

### DIFF
--- a/xla/tests/collective_ops_e2e_test.cc
+++ b/xla/tests/collective_ops_e2e_test.cc
@@ -70,6 +70,31 @@ class CollectiveOpsTestE2E : public HloTestBase {
            GetDebugOptionsForTest().xla_gpu_enable_cublaslt();
   }
 
+  void CollectiveOpsVerifyF8Matmul(absl::string_view hlo_text,
+                                   const DebugOptions& options) {
+    if (!HasFp8Support()) {
+      return;
+    }
+    const int64_t kNumReplicas = 1;
+    const int64_t kNumPartitions = 4;
+
+    HloModuleConfig config =
+        GetModuleConfigForTest(/*replica_count=*/kNumReplicas);
+    config.set_debug_options(options);
+    config.set_num_partitions(kNumPartitions);
+    TF_ASSERT_OK_AND_ASSIGN(auto module,
+                            ParseAndReturnVerifiedModule(hlo_text, config));
+
+    TF_ASSERT_OK_AND_ASSIGN(auto executable,
+                            CreateExecutable(std::move(module),
+                                             /*run_hlo_passes=*/true));
+    EXPECT_TRUE(executable->has_module());
+    HloInstruction* gemm_op =
+        FindInstruction(&executable->module(), HloOpcode::kCustomCall);
+    EXPECT_THAT(gemm_op, NotNull());
+    EXPECT_EQ(gemm_op->custom_call_target(), "__cublas$lt$matmul$f8");
+  }
+
   absl::StatusOr<std::vector<Literal>> ExecuteReplicated(Executable* executable,
                                                          int64_t num_replicas) {
     DeviceAssignment device_assignment = MakeDeviceAssn(num_replicas);
@@ -825,6 +850,59 @@ ENTRY main.12 {
   // Custom Calls.
   CollectiveOpsCompareWindowedNonWindowed(kModuleReplicatedStr,
                                           /*disable_dot_merger=*/true);
+
+  // Verify the creation of FP8 GEMM Custom Calls on Hopper and newer
+  // architectures.
+  DebugOptions opts = GetDebugOptionsForTest();
+  opts.set_xla_gpu_threshold_for_windowed_einsum_mib(0);
+  opts.set_xla_gpu_multi_streamed_windowed_einsum(true);
+  opts.set_xla_gpu_graph_min_graph_size(200);
+  opts.set_xla_gpu_enable_triton_gemm(false);
+  opts.add_xla_disable_hlo_passes("dot-merger");
+  CollectiveOpsVerifyF8Matmul(kModuleReplicatedStr, opts);
+}
+
+TEST_F(CollectiveOpsTestE2EWindowedNonWindowed,
+       WindowedEinsumE2EAllGatherMultiConsumerF8) {
+  absl::string_view kModuleReplicatedStr = R"(
+HloModule windowed_einsum_e2e_all_gather_multi_consumer_f8, entry_computation_layout={(f8e4m3fn[2,16,48]{2,1,0}, f8e4m3fn[48,192]{1,0}, f8e4m3fn[48,192]{1,0}, bf16[], bf16[], bf16[])->bf16[2,16,192]{2,1,0}}, allow_spmd_sharding_propagation_to_parameters={false,false,false,false}, num_partitions=4
+
+ENTRY main {
+  rhs = f8e4m3fn[2,16,48]{2,1,0} parameter(0), sharding={devices=[1,4,1]<=[4]}
+  lhs0 = f8e4m3fn[48,192]{1,0} parameter(1), sharding={devices=[1,4]<=[4]}
+  scale_rhs = bf16[] parameter(3)
+  scale_lhs0 = bf16[] parameter(4)
+  scale_rhs_bcast = bf16[2,16,48]{2,1,0} broadcast(scale_rhs), dimensions={}
+  scale_lhs0_bcast = bf16[48,192]{1,0} broadcast(scale_lhs0), dimensions={}
+  rhs_bf16 = bf16[2,16,48]{2,1,0} convert(rhs)
+  lhs0_bf16 = bf16[48,192]{1,0} convert(lhs0)
+  rhs_scaled = bf16[2,16,48]{2,1,0} multiply(scale_rhs_bcast, rhs_bf16)
+  lhs0_scaled = bf16[48,192]{1,0} multiply(scale_lhs0_bcast, lhs0_bf16)
+  dot0 = bf16[2,16,192]{2,1,0} dot(rhs_scaled, lhs0_scaled), lhs_contracting_dims={2}, rhs_contracting_dims={0}
+  lhs1 = f8e4m3fn[48,192]{1,0} parameter(2), sharding={devices=[1,4]<=[4]}
+  scale_lhs1 = bf16[] parameter(5)
+  scale_lhs1_bcast = bf16[48,192]{1,0} broadcast(scale_lhs1), dimensions={}
+  lhs1_bf16 = bf16[48,192]{1,0} convert(lhs1)
+  lhs1_scaled = bf16[48,192]{1,0} multiply(scale_lhs1_bcast, lhs1_bf16)
+  dot1 = bf16[2,16,192]{2,1,0} dot(rhs_scaled, lhs1_scaled), lhs_contracting_dims={2}, rhs_contracting_dims={0}
+  ROOT add.8 = bf16[2,16,192]{2,1,0} add(dot0, dot1)
+} // main
+)";
+
+  // Disable the dot merger pass which can prevent the creation of FP8 GEMM
+  // Custom Calls.
+  CollectiveOpsCompareWindowedNonWindowed(kModuleReplicatedStr,
+                                          /*disable_dot_merger=*/true);
+
+  // Verify the creation of FP8 GEMM Custom Calls on Hopper and newer
+  // architectures.
+  DebugOptions opts = GetDebugOptionsForTest();
+  opts.set_xla_gpu_threshold_for_windowed_einsum_mib(0);
+  opts.set_xla_gpu_multi_streamed_windowed_einsum(true);
+  opts.set_xla_gpu_graph_min_graph_size(200);
+  opts.set_xla_gpu_enable_triton_gemm(false);
+  opts.add_xla_disable_hlo_passes("dot-merger");
+  CollectiveOpsVerifyF8Matmul(kModuleReplicatedStr, opts);
 }
 
 TEST_F(CollectiveOpsTestE2EWindowedNonWindowed,
@@ -980,19 +1058,7 @@ ENTRY entry {
   opts.set_xla_gpu_run_post_layout_collective_pipeliner(true);
   opts.set_xla_gpu_enable_pipelined_collectives(true);
   opts.set_xla_gpu_enable_triton_gemm(false);
-  config.set_debug_options(opts);
-  config.set_num_partitions(kNumPartitions);
-  TF_ASSERT_OK_AND_ASSIGN(
-      auto module, ParseAndReturnVerifiedModule(kModuleReplicatedStr, config));
-
-  TF_ASSERT_OK_AND_ASSIGN(auto executable,
-                          CreateExecutable(std::move(module),
-                                           /*run_hlo_passes=*/true));
-  EXPECT_TRUE(executable->has_module());
-  HloInstruction* gemm_op =
-      FindInstruction(&executable->module(), HloOpcode::kCustomCall);
-  EXPECT_THAT(gemm_op, NotNull());
-  EXPECT_EQ(gemm_op->custom_call_target(), "__cublas$lt$matmul$f8");
+  CollectiveOpsVerifyF8Matmul(kModuleReplicatedStr, opts);
 }
 
 TEST_F(CollectiveOpsTestE2E,
--- a/xla/service/gpu/transforms/windowed_einsum_handler.cc
+++ b/xla/service/gpu/transforms/windowed_einsum_handler.cc
@@ -48,15 +48,15 @@ namespace m = match;
 // and type conversions of FP8 operands into the bodies of their while loops,
 // i.e. rewrites
 //
-//   inputs --> dequant --> while loop {dynamic-slice/collective-permute/dot}
+//   inputs --> dequant --> while loop {collective-permute/dot/etc}
 //
 // into
 //
-//   inputs --> while loop {dequant --> dynamic-slice/collective-permute/dot}.
-absl::Status ShiftDequantizationF8(const HloComputation* comp,
-                                   const std::array<HloInstruction*, 2>& gte) {
-  HloInstruction* while_instr = comp->WhileCallInstruction();
-  if (!while_instr) {
+//   inputs --> while loop {dequant --> collective-permute/dot/etc}.
+absl::Status ShiftDequantizationF8(HloComputation* while_body) {
+  HloInstruction* while_instr = while_body->WhileCallInstruction();
+  // The input of the while loop will be modified and must have no other users.
+  if (!while_instr || while_instr->operand(0)->user_count() != 1) {
     return absl::OkStatus();
   }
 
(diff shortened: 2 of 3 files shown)


reedwm:
@philipphack I don't fully understand this pass. I pinged you, asking for help to understand it. Can you check your Google Chat pings? :)

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -526,35 +623,40 @@ absl::StatusOr<bool> GpuWindowedEinsumHandler::Run(
...
+  //                   /
+  //                  /
+  //                dot
   // to:
   //                       input
   //                       |
   //                       |
   //                     windowed loop
   //                       |
+  //                     DQ (FP8)
```
frgossen, line 649: What is DQ? Dequantize?

xla/tests/collective_ops_test_e2e.cc
```diff
@@ -781,5 +781,45 @@ ENTRY main.12 {
 
   CollectiveOpsCompareWindowedNonWindowed(kModuleReplicatedStr);
 }
+
+TEST_F(CollectiveOpsTestE2EWindowedNonWindowed,
```
frgossen, line 785: The pattern depicted in the comment includes only a few ops. Can the test case be simplified?

frgossen (changes requested):
Can you provide a little more context to this PR? 
- What performance improvements did you measure? 
- What patterns are you matching for? 

The test cases seem somewhat verbose to me but I might be wrong. The patterns you're matching for seem a lot simpler. Is it possible to condense the test cases to be minimal?

abhinavgoel95:
@frgossen this PR helps enable collective matmul for FP8 gemms. This is needed to overlap Tensor Parallel communication.
The expected speedup depends on the model configs.
But in the recent MoE models, we can expect 10-20% speedup.

xla/service/gpu/gpu_windowed_einsum_handler.cc
philipphack: Yes. I've edited the comment.

xla/tests/collective_ops_test_e2e.cc
philipphack: I made the input HLO less complex. The FileCheck in the unit test now compares only a part of the main computation.

philipphack:
@frgossen the patterns are FP8 windowed einsums with input dequantizations (i.e. type-conversion to a wider type like FP16 or BF16 and scaling) and all-gathers that have multiple dot users. The dequantization is moved from the all-gather of the additional dot to the output of that holds the cached all-gather result as indicated in the comment. This enables the creation of a specific FP8 GEMM custom call in gemm_rewriter.cc which replaces dequantization and dot.

(Also rebased to resolve conflicts.)

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -53,11 +53,11 @@ namespace m = match;
 // into
```
frgossen, line 53: `Enables the creation of FP8 GEMM Custom Calls for all-gather and
// reduce-scatter windowed einsums in gemm_rewriter.cc...` - Can we add a test that tests the two passes in combination?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -673,66 +638,142 @@ class WindowedEinsumVisitor : public DfsHloRewriteVisitor {
...
     // Now that we have processed all loops, we can check if there are any
     // allgather-dot pattern that we can optimize. We'd want to transform:
     //                       input
     //                       /    |
-    //                      /     |
-    //                     AG    windowed loop
-    //                     /
-    //                    /
-    //                   dot
+    //                dequantize  |
```
frgossen, line 645: Is dequantize optional or always needed to match?

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -831,6 +831,37 @@ ENTRY main.9_spmd {
   CollectiveOpsCompareWindowedNonWindowed(kModuleReplicatedStr);
 }
 
+TEST_F(CollectiveOpsTestE2EWindowedNonWindowed,
+       WindowedEinsumE2EAllgatherMultiConsumerF8) {
+  absl::string_view kModuleReplicatedStr = R"(
+HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(f8e4m3fn[2,16,48]{2,1,0}, f8e4m3fn[48,192]{1,0}, f8e4m3fn[48,192]{1,0}, bf16[], bf16[], bf16[])->bf16[2,16,192]{2,1,0}}, allow_spmd_sharding_propagation_to_parameters={false,false,false,false}, num_partitions=4
+
+ENTRY main.12 {
```
frgossen, line 839: IIUC, the pattern is `inputs --> dequant --> while loop {dynamic-slice/collective-permute/dot}` and here you are testing the dot case? 

- Is the test case minimal?
- Can we add test cases for the two other cases?

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -830,14 +830,113 @@ ENTRY main.9_spmd {
...
 
-TEST_F(GpuWindowedEinsumHanlderTest,
+TEST_F(GpuWindowedEinsumHandlerTest, AllGatherMultipleConsumersF8) {
+  constexpr absl::string_view kHloString = R"(
+HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(f8e4m3fn[2,512,24576]{2,1,0}, f8e4m3fn[24576,24576]{1,0}, f8e4m3fn[24576,24576]{1,0}, f32[], f32[], f32[])->f32[2,2048,24576]{2,1,0}}, num_partitions=4
+windowed_dot_general_body_ag {
+  param.1 = (f32[2,512,24576]{2,1,0}, f32[24576,24576]{1,0}, f32[2,2048,24576]{2,1,0}, f32[2,2048,24576]{2,1,0}, u32[]) parameter(0)
+  get-tuple-element.1 = f32[2,512,24576]{2,1,0} get-tuple-element(param.1), index=0
+  collective-permute = f32[2,512,24576]{2,1,0} collective-permute(get-tuple-element.1), channel_id=2, source_target_pairs={{0,3},{1,0},{2,1},{3,2}}
+  collective-permute.1 = f32[2,512,24576]{2,1,0} collective-permute(collective-permute), channel_id=3, source_target_pairs={{0,3},{1,0},{2,1},{3,2}}
```
frgossen, line 855: This is testing the collective permute case, is it? Can we also add minimal test cases for the two remaining cases?

frgossen (changes requested):

xla/service/gpu/gpu_windowed_einsum_handler.cc
philipphack: The comment was intended to indicate that there are dynamic-slice, collective-permute and dot instructions inside the all-gather and reduce-scatter loops. The instructions occur together and don't describe separate cases. I've edited the comment to make it clearer.

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -830,14 +830,113 @@ ENTRY main.9_spmd {
...
+; CHECK-DAG:         "wait_on_operation_queues":["[[OPQUEUEID1]]"],
+; CHECK-DAG:         "force_earliest_schedule":false}
 ; CHECK-NEXT:    [[CP1:[^ ]+]] = f32[2,512,24576]{2,1,0} collective-permute([[ADD5]]), channel_id=10
 ; CHECK-NEXT:  ROOT [[OUT:[^ ]+]] = (f8e4m3fn[2,2048,24576]{2,1,0}, f8e4m3fn[24576,24576]{1,0}, f32[2,512,24576]{2,1,0}, f32[2,512,24576]{2,1,0}, u32[], /*index=5*/f32[], f32[]) tuple([[GTE0]], [[GTE1]], [[ADD3]], [[CP1]], [[ADD0]], /*index=5*/[[GTE3]], [[GTE5]])
 )");
 }
 
-TEST_F(GpuWindowedEinsumHanlderTest,
+TEST_F(GpuWindowedEinsumHandlerTest, AllGatherMultipleConsumersF8) {
+  constexpr absl::string_view kHloString = R"(
```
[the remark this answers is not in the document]
philipphack: I think this test is minimal. This is essentially test AgLoopsMultipleConsumersAreChained with FP8 inputs and added dequantization (type conversion and scaling).

xla/service/gpu/gpu_windowed_einsum_handler.cc
philipphack: The dequantization step is optional. When the data type is FP8 and there’s a dequantization, it will be shifted as described in the comment. This doesn't interfere with other cases.

xla/tests/collective_ops_e2e_test.cc
philipphack: Yes, I think the test case is minimal. The comment in gpu_windowed_einsum_handler.cc was intended to indicate that there are dynamic-slice, collective-permute and dot instructions inside the all-gather and reduce-scatter loops. The instructions occur together and don't describe separate cases. I've edited the comment to make it clearer.

dimitar-asenov:
@frgossen @cheshire @reedwm Could you please review the latest version of this PR?

toli-y:
>  this PR helps enable collective matmul for FP8 gemms. This is needed to overlap Tensor Parallel communication. The expected speedup depends on the model configs. But in the recent MoE models, we can expect 10-20% speedup.

@abhinavgoel95 It's great that we expect 10-20% speedup. Can you please provide exact model and hardware configuration and the speedup you are seeing with this change?

abhinavgoel95:
@toli-y Are you able to run GPT-5B, FP8 precision, with ICI_MESH = [1,1,8], on 8 H100 GPUs?

philipphack:
@frgossen can you PTAL?

thomasjoerg:
@ezhulenev to nag reviewers or assign others.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -831,6 +866,48 @@ ENTRY main.9_spmd {
...
+
+  // Verify the creation of FP8 GEMM Custom Calls on Hopper and newer
+  // architectures.
+  DebugOptions opts = GetDebugOptionsForTest();
+  opts.set_xla_gpu_threshold_for_windowed_einsum_mib(0);
+  opts.set_xla_gpu_multi_streamed_windowed_einsum(true);
+  opts.set_xla_gpu_graph_min_graph_size(200);
+  opts.set_xla_gpu_enable_triton_gemm(false);
+  opts.add_xla_disable_hlo_passes("dot-merger");
+  CollectiveOpsVerifyF8Matmul(kModuleReplicatedStr, opts);
```
frgossen, line 908: Why are these options needed to test the windowed einsum handler? 
What does this have to do with the dot-merger? Why do you have to disable triton?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -94,54 +94,55 @@ absl::Status ShiftDequantizationF8(const HloComputation* comp,
...
-                                           &coll_perms[0], m::Op().Is(gte[0]))),
-                   m::Op().Is(gte[1]),
-                   m::DynamicUpdateSlice(
-                       m::DynamicUpdateSlice().WithOperand(
-                           1, m::Dot(&dots[0], m::Op().Is(gte[0]),
-                                     m::Op().Is(gte[1]))),
-                       m::Dot(&dots[1], m::Op(), m::Op().Is(gte[1])), m::Op(),
-                       m::Op(), m::Op()),
-                   m::Op(), m::Op()))) {
+  if (Match(while_root,
```
frgossen, line 115: I don't really understand what you are matching here exactly but it does look quite specific. Does this work on a wider range of models, e.g. will this apply if I run this on maxtext Llama 2?

frgossen (changes requested):
Sorry for the delay here. I still have a hard time following what this PR does and why it is an improvement as it adds quite a bit of complexity. 

Would it be possible to write down the motivation behind this change and the observed improvements? In particular, I would like to understand: 

- What the IR looks like before this pass and why that is a problem? 
- How this pass changes the IR and why that solves the problem? 
- Why this is the right place to fix the problem? 
- On what benchmarks/ workloads do you observe performance improvements? 
- Can you provide some relevant HLO that I can add to our benchmark suite to ensure this does not regress over time?

philipphack:
@frgossen the motivation for this change is the extension of the existing functionality for collective matmuls with multiple all-gather dots in the windowed einsum handler (see [this comment](https://github.com/openxla/xla/blob/e56c9c1c51cef21cda50eddc45f2a115836a9358/xla/service/gpu/gpu_windowed_einsum_handler.cc#L676)) to support FP8 inputs. In this case, the dot operates on a dequantization of an FP8 operand, i.e. a type conversion from FP8 to a wider type followed by a multiplication by a broadcast scaling factor. The current change adds functionality for moving the dequantization from between input and all-gather to between the windowed loop and the dot as explained in [the updated comment](https://github.com/openxla/xla/blob/6adbf31a147fd8f6fea2741fd00a02e27c203cea/xla/service/gpu/gpu_windowed_einsum_handler.cc#L641). The dequantization -> dot sequence is rewritten in gemm_rewriter.cc into a specific FP8 GEMM custom call. This change does not interfere with the existing functionality of the windowed einsum handler.

The HLO of WindowedEinsumE2EAllgatherMultiConsumerF8 in collective_ops_e2e_test.cc can also be used for monitoring performance (FP8 requires Hopper or newer architectures).

xla/tests/collective_ops_e2e_test.cc
philipphack: The purpose of this end-to-end test is to verify the sequence of windowed einsum handler and GEMM rewriter passes. The dot merger and Triton passes are disabled since they can potentially interfere with the outcome.

xla/service/gpu/gpu_windowed_einsum_handler.cc
philipphack: This matches the structure of the while loops of all-gather or reduce-scatter windowed einsum patterns. The patterns are generated by the SPMD dot handler and should not depend on the model.

frgossen: One example here: I find it strange that the pattern matches a specific size tuple. If you rely on the SPMD partitioner to generate this specific pattern that seems overly fragile. What if the partitioner changes? Do the transformations here depend on the number of values passed through the while loop?

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -813,6 +838,16 @@ ENTRY main.12 {
...
   CollectiveOpsCompareWindowedNonWindowed(kModuleReplicatedStr,
                                           /*disable_dot_merger=*/true);
+
+  // Verify the creation of FP8 GEMM Custom Calls on Hopper and newer
+  // architectures.
+  DebugOptions opts = GetDebugOptionsForTest();
+  opts.set_xla_gpu_threshold_for_windowed_einsum_mib(0);
+  opts.set_xla_gpu_multi_streamed_windowed_einsum(true);
+  opts.set_xla_gpu_graph_min_graph_size(200);
+  opts.set_xla_gpu_enable_triton_gemm(false);
```
frgossen, line 848: How is this related to triton gemms?

xla/tests/collective_ops_e2e_test.cc
frgossen: Do you know why they interfer with the outcome? Is there a fundamental reason or can this be implemented more generically so that it does compose well with the rest of XLA?

frgossen (changes requested):
I left a few comments but Reed will look into this in more detail. Some of the test cases still have value names based on the operations. I suspect some of these are quantization and dequantization. If you can give them meaningful names it might be easier to follow this.

xla/tests/collective_ops_e2e_test.cc
philipphack: The purpose of this test is to verify that gemm_rewriter.cc creates an FP8 cublasLT custom call. The Triton rewriter may attempt to alternatively rewrite this into a Triton GEMM, which would interfere with this objective.

philipphack:
@frgossen for some reason I can't directly respond to these two comments:

> Do you know why they interfer with the outcome? Is there a fundamental reason or can this be implemented more generically so that it does compose well with the rest of XLA?

The dot merger pass has a threshold for the input dimensions which means that it generally doesn't interfere in actual models. Since the test uses relatively small inputs, the pass is manually deactivated here to test the functionality.

> One example here: I find it strange that the pattern matches a specific size tuple. If you rely on the SPMD partitioner to generate this specific pattern that seems overly fragile. What if the partitioner changes? Do the transformations here depend on the number of values passed through the while loop?

I think the general purpose of the windowed einsum handler is to transform windowed einsums that were created by the SPMD dot handler. As such, it relies on the output of that pass by design. The alternative would be the direct modification of the dot handler which is already relatively complex.

xla/tests/collective_ops_e2e_test.cc
frgossen: Can you add a comment?

frgossen (changes requested):
> The dot merger pass has a threshold for the input dimensions which means that it generally doesn't interfere in actual models. Since the test uses relatively small inputs, the pass is manually deactivated here to test the functionality.

Can you add a comment? 

> I think the general purpose of the windowed einsum handler is to transform windowed einsums that were created by the SPMD dot handler. As such, it relies on the output of that pass by design. The alternative would be the direct modification of the dot handler which is already relatively complex.

If this pass is so tightly coupled with the dot handler, shouldn't there be tests that reflect this? Do you think it makes sense to add test cases that test the two passes together?

philipphack:
> If this pass is so tightly coupled with the dot handler, shouldn't there be tests that reflect this? Do you think it makes sense to add test cases that test the two passes together?

@frgossen that's what the end-to-end tests in collective_ops_e2e_test.cc do. The SPMD dot handler transforms the input HLO into a collective matmul/windowed einsum which is transformed again by the windowed einsum handler.

> Can you add a comment?

Done.

reedwm (changes requested):

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -48,16 +48,15 @@ namespace m = match;
...
 //
-//   inputs --> while loop {dequant --> ...}.
-absl::StatusOr<bool> ShiftDequantizationF8(HloInstruction* instr) {
-  HloComputation* comp = instr->parent();
+//   inputs --> while loop {dequant --> collective-permute/dot/etc}.
+absl::Status ShiftDequantizationF8(HloComputation* comp) {
   HloInstruction* while_instr = comp->WhileCallInstruction();
-  if (!while_instr) {
-    return false;
+  if (!while_instr || while_instr->operand(0)->user_count() != 1) {
```
reedwm, line 58: Comment why there is the check `while_instr->operand(0)->user_count() != 1`. E.g.:

The while-loop tuple operand must only be used by the while-loop, since we modify the tuple arguments

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -329,28 +326,27 @@ absl::StatusOr<bool> HandleRsWindowedEinsumLoop(HloComputation* comp,
       changed = true;
     }
   }
+  // If present, move the dequantization of FP8 operands of the dot into the
+  // while loop to allow gemm_rewriter.cc to rewrite into an FP8 Custom
```
reedwm, line 330: I should have commented before, but the triton pass, not just gemm_rewriter.cc, can rewrite into an FP8 matmul. Probably should be phrased as 

...to allow the dequantization and dot to be fused into an FP8 matmul, e.g. by gemm_rewriter.cc.

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -638,12 +637,15 @@ class WindowedEinsumVisitor : public DfsHloRewriteVisitor {
...
+    // windowed einsum loop to consume the output of the loop and remove the
+    // all-gather. Now that we have processed all loops, we can check if there
+    // are any allgather-dot pattern that we can optimize. We'd want to
+    // transform:
     //                       input
     //                       /    |
-    //                dequantize  |
-    //                  (FP8)     |
+    //               dequantize   |
+    //          (optional if FP8) |
```
reedwm, line 648: `(optional if FP8)` implies, to me at least, that the dequantize is only optional if FP8 (and otherwise required).

Change to `(optional, only if FP8)` or `(optional)` or `(only if FP8)`

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -48,16 +48,15 @@ namespace m = match;
...
-//   inputs --> dequant --> while loop {...}
+//   inputs --> dequant --> while loop {collective-permute/dot/etc}
 //
 // into
 //
-//   inputs --> while loop {dequant --> ...}.
-absl::StatusOr<bool> ShiftDequantizationF8(HloInstruction* instr) {
-  HloComputation* comp = instr->parent();
+//   inputs --> while loop {dequant --> collective-permute/dot/etc}.
+absl::Status ShiftDequantizationF8(HloComputation* comp) {
```
reedwm, line 56: Document `comp` is the body computation of the while-loop.

In fact, you get a `while_body` HloComputation* local variable below, which is exactly the same as `comp`. I would rename `comp` to `while_body` (which is obvious enough it doesn't have to be documented), and then remove the definition of `while_body` below.

reedwm (changes requested):

reedwm (approved):

PR #13813, opened 2024-06-15
[NVIDIA GPU] Assign a fixed index for cached activation

Tixxx:
gpu_windowed_einsum_handler pass has been re-using the empty buffer of the transformed while loop. This buffer is given by the spmd dot_handler pass. The shape of the buffer has changed from the allgathered shape of the sharded operand to the output shape of the dot which leads to a shape incompatibility error. To make the gpu handler completely safe, we will make a new element in the tuple to host the cached activation with the desired shape.
The slice index of where to write the slice into the full buffer also changes based on whether it's contracting or non-contracting dim is sharded. With the new element, we will need to determine the slice index ourselves in the handler pass.

### DIFF
--- a/xla/service/gpu/gpu_windowed_einsum_handler.cc
+++ b/xla/service/gpu/gpu_windowed_einsum_handler.cc
@@ -495,37 +495,31 @@ absl::Status ProcessWindowedEinsumLoopForActivationCaching(
               {dus_boundary_constant, slice_indices, dus_boundary_constant}));
     }
 
-    // If we have a DS(parameter_index1), then operands are sharded along the
-    // contracting dim. Slice indices will be the contracting dim's slices.
+    // If we have a Dot(DS(parameter_index1)), then operands are sharded along
+    // the contracting dim. Slice indices will be the contracting dim's slices.
     HloInstruction* slice_index;
     HloInstruction* ds_index_constant;
     HloInstruction* remainder;
     HloInstruction* ds_param;
     HloInstruction* ds_operand2;
     HloInstruction* ds_operand3;
-    // There will be 2 DSes for unrolled loops, match for each one.
-    if (Match(inst, m::DynamicSlice(&ds_param).WithPredicate(
-                        [&](const HloInstruction* instr) {
-                          bool op1_is_param =
-                              Match(instr->operand(0),
-                                    m::GetTupleElement(m::Parameter(), 1));
-                          return op1_is_param && instr->operands().size() >= 3;
-                        }))) {
-      ds_operand2 = ds_param->mutable_operand(1);
-      ds_operand3 = ds_param->mutable_operand(2);
-      bool matched_rhs_sharded_contracting =
-          Match(ds_operand2,
+    // There will be 2 dynamic-slices for unrolled loops, match for each one to
+    // get the slice index which will be used to write the corresponding
+    // received shard into cached activation buffer. For unrolled loops, we need
+    // to write to the final buffer twice per iteration, so we need to match for
+    // the correct slice index based on each DS.
+    if (Match(inst, m::Dot(m::Op(), m::DynamicSlice(&ds_param))) &&
+        Match(ds_param->operand(0), m::GetTupleElement(m::Parameter(), 1))) {
+      for (int64_t ds_op_i = 1; ds_op_i < ds_param->operands().size();
+           ds_op_i++) {
+        if (!Match(
+                ds_param->mutable_operand(ds_op_i),
                 m::Reshape(&slice_index, m::DynamicSlice(m::Constant(),
                                                          m::Op(&remainder)))) &&
-          Match(ds_operand3, m::Constant(&ds_index_constant));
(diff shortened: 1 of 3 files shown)


Tixxx:
@frgossen  can you take a look at this? This unblocks benchmarking from our side.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -659,6 +659,7 @@ class CollectiveOpsTestE2EWindowedNonWindowed : public CollectiveOpsTestE2E {
     auto opts = GetDebugOptionsForTest();
     opts.set_xla_gpu_threshold_for_windowed_einsum_mib(0);
     opts.set_xla_gpu_multi_streamed_windowed_einsum(true);
+    opts.set_xla_gpu_mlir_emitter_level(0);
```
frgossen, line 662: Why is this needed? Do you run into an issue with the MLIR emitters?

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -838,5 +837,82 @@ ENTRY main.9_spmd {
 )");
 }
 
+TEST_F(GpuWindowedEinsumHanlderTest,
+       AgLoopsMultipleConsumersAreChainedWithShardedContratingDim) {
+  constexpr absl::string_view kHloString = R"(
```
frgossen, line 842: Can you simplify the test case? From the PR description it reads like it is possible to reproduce this with a smaller example.

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -434,24 +488,76 @@ absl::Status ProcessWindowedEinsumLoopForActivationCaching(
...
           while_body->AddInstruction(HloInstruction::CreateDynamicUpdateSlice(
               full_buffer_output_gte->shape(), full_buffer_output_gte,
-              cp_output,
+              first_cp_output,
               {dus_boundary_constant, slice_indices, dus_boundary_constant}));
     }
+
+    // If we have a DS(parameter_index1), then operands are sharded along the
+    // contracting dim. Slice indices will be the contracting dim's slices.
+    HloInstruction* slice_index;
```
frgossen, line 500: You are matching a very specific pattern here. Are you sure this is the right thing to do to fix this bug or does the issue occur for other patterns, too?

frgossen (changes requested):

xla/tests/collective_ops_e2e_test.cc
Tixxx: i can remove this now. There was an issue with the mlir emitter level 1. We had a discussion on gchat, this has been changed to level 0 by default.

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
Tixxx: The scenario that this breaks is specific, we need this pattern to trigger the issue:

1. an already rewritten allgather+gemm loop(with either contracting or noncontracting dim being sharded)
2. another allgather that shares the same operand with the above loop
3. another gemm that consumes the above allgather
That's why the unit test looks complex but it actually is a minimum case to repro. We had an unit test before with noncontracting dim sharded, this newly added one is with contracting dim sharded.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: Yea it's a good point. I simplified it a bit and made it more generic. We still need to be specific when we look for the slice index when we update the final output buffer when the loop is unrolled. I added more comment to explain why we need to be specific in that case.

frgossen (approved):

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -434,24 +488,70 @@ absl::Status ProcessWindowedEinsumLoopForActivationCaching(
...
+    // the contracting dim. Slice indices will be the contracting dim's slices.
+    HloInstruction* slice_index;
+    HloInstruction* ds_index_constant;
+    HloInstruction* remainder;
+    HloInstruction* ds_param;
+    HloInstruction* ds_operand2;
+    HloInstruction* ds_operand3;
+    // There will be 2 dynamic-slices for unrolled loops, match for each one to
+    // get the slice index which will be used to write the corresponding
+    // received shard into cached activation buffer. For unrolled loops, we need
```
frgossen, line 508: Can you check this logic again? ds_operand2/3 are unused variables.

frgossen (changes requested):

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: ah forgot to remove them after cleanup. removed.

frgossen (approved):

Issue #13733, opened 2024-06-13
How to use PJRT C API v.2


janpfeifer:
hi, I'm forking the previous #7038 because that seems to get PJRT to work leveraging the C++ wrappers.

I'm trying to build a [PJRT wrap for Go](https://github.com/gomlx/gopjrt) leveraging purely the [PJRT C API](https://github.com/openxla/xla/blob/main/xla/pjrt/c/pjrt_c_api.h) to power an [ML framework for Go](https://github.com/gomlx/gomlx) (I've written here before so some may be familiar with the project. But hopefully the questions are generic for anyone building an ML for any X language (or anyone else needing compiled accelerator code).

**Question 1** and the most important (before I spend too much time on it): is using simply the `pjrt_c_api.h` file, doing a `dlopen` on the compiled plugins (from `openxla/xla` for now), and using it directly a valid approach ? Or does one need any of the helper/wrappers around it under `xla/xla/pjrt` ?  Are there specific considerations or limitations one should be aware of when taking this path?

For the time being in the GoMLX project I'll use XlaBuilder to generate the StableHLO (it provides me shape inference of the result of the ops). But later the goal would be to use purely the PJRT C API, if viable.

**Question 2**: While the pjrt_c_api.h header provides function definitions, there are several aspects that I found unclear and I was wondering if there is additional documentation. Some examples -- apologies if noob questions, but I usually don't interact at this level:

1. What are the available extension types and their purpose ?
2. Are there documentation on the CPU/GPU plugins (from OpenXLA/XLA repository) ?
3. Client/Multi-Node Semantics: How are PjRT clients and multi-node setups related? If there are multiple GPUs on a machine, should one client be created per GPU instance? Is this behavior consistent across different plugins?
4. Buffer Abstractions: What are the underlying abstractions for PjRT buffers? What synchronization mechanisms are in place (host-device, cross-device, cross-plugin)? How does the concept of tiling (mentioned in pjrt_c_api.h) work?
 
Are there any design documents or references (perhaps from JAX or TensorFlow) that would be helpful in understanding these aspects of the C API?
 
Btw, since I'm opening the issue, one potential (C vs C++) issue with `pjrt_c_api.h`: on [line 56](https://github.com/openxla/xla/blob/main/xla/pjrt/c/pjrt_c_api.h#L56) whenever I include it on the most trivial C file, the compile fails with:

```
./pjrt_c_api.h:56:3: error: unknown type name 'PJRT_Extension_Base'
```

[...]

PR #13819, opened 2024-06-15
Added `struct` in front of recursive struct PJRT_Extension_Base

janpfeifer:
In C the type alias of `struct PJRT_Extension_Base` to `PJRT_Extension_Base` (without "struct") is not created yet, and the .h file fails to compile (both in gcc and clang).

C++ compiles that because it does create the alias.

The change won't break C++, and it will allow the C version to work as well.

See #13733

Tested with `bazel test //xla/pjrt/c:pjrt_c_api_cpu_test`


cheshire (approved):

janpfeifer:
Failure on AMD ROCm seem very unrelated to this PR.

PR #13836, opened 2024-06-17
Introduce `xla_gpu_test` macro to avoid common boilerplate in BUILD files

sergey-kozub:
Provides a simpler interface to `xla_test` macro for GPU-only tests.

1) Wraps `srcs` in `if_gpu_is_configured` to avoid compilation on non-GPU backends (e.g. CPU).
Some test targets currently do this, but it's not consistent.

2) Similarly, wraps `deps` in `if_gpu_is_configured` to generate less dependencies on non-GPU backends (e.g. CPU).
Only a few test targets currently do this.

3) Consistently define GOOGLE_CUDA and TENSORFLOW_USE_ROCM We had a few issues due to missing definitions in the past.

4) Do not require listing specific architectures in `backends` Instead, enable shortcut that enables compatible cuda architectures: e.g. with_cuda = "ampere+"

5) Do not generate tests with "gpu_any" suffix.
This means "run on any compatible architecture" in the Google CI, which *will* result in flaky tests for any tests that check for SM version during runtime (e.g. if a test passes on Ampere but fails on Hopper, then "gpu_any" test will be flaky).

6) Hides the complexity of `xla_test` macro.
It is significantly different at Google because of TPU support, so using it is hard in OSS (as the macro source is not visible).

Eventually we should consider removing test suffixes for GPU tests, as "gpu_a100" and "gpu_h100" are essentially the same, and are only needed due to how Google internal CI operates (requires SM-specific tags for a test target).

I propose to use `xla_gpu_test` for all runtime tests (i.e. requiring a GPU to run the test) inside the //xla/service/gpu/... namespace.

### DIFF
--- a/xla/service/gpu/ir_emitter_triton_test.cc
+++ b/xla/service/gpu/ir_emitter_triton_test.cc
@@ -1425,7 +1425,7 @@ ENTRY entry {
 
   auto backend_config_or =
       triton_dot_fusion->backend_config<GpuBackendConfig>();
-  ASSERT_OK(backend_config_or);
+  TF_ASSERT_OK(backend_config_or);
   GpuBackendConfig& backend_config = *backend_config_or;
 
   FusionBackendConfig& fusion_backend_config =
@@ -1439,7 +1439,7 @@ ENTRY entry {
   config.set_num_warps(8);
   config.set_num_stages(4);
 
-  ASSERT_OK(triton_dot_fusion->set_backend_config(backend_config));
+  TF_ASSERT_OK(triton_dot_fusion->set_backend_config(backend_config));
 
   BlockLevelParameters block_level_parameters;
   block_level_parameters.num_ctas = 1;
@@ -1457,7 +1457,7 @@ ENTRY entry {
   config.set_block_n(128);
   config.set_block_k(128);
   block_level_parameters.num_stages = 1;
-  ASSERT_OK(triton_dot_fusion->set_backend_config(backend_config));
+  TF_ASSERT_OK(triton_dot_fusion->set_backend_config(backend_config));
 
   TF_ASSERT_OK_AND_ASSIGN(
       const auto result,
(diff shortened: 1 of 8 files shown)


sergey-kozub:
Discussion started in [PR#13526](https://github.com/openxla/xla/pull/13526)

xla/service/gpu/ir_emitter_triton_test.cc
```diff
@@ -1457,7 +1457,7 @@ ENTRY entry {
   config.set_block_n(128);
   config.set_block_k(128);
   block_level_parameters.num_stages = 1;
-  ASSERT_OK(triton_dot_fusion->set_backend_config(backend_config));
+  TF_ASSERT_OK(triton_dot_fusion->set_backend_config(backend_config));
```
dimitar-asenov, line 1460: These have been already fixed. Please sync to the latest version.

dimitar-asenov (commented):
While I'm not keen on introducing yet another `xla` specific BUILD rule, I do agree that this addresses a real issue. Overall I'm supportive of this change.

I'm not experienced enough with our OSS setup to foresee issues this might introduce. @ddunl and @beckerhe should have the final word.

ddunl:
Sorry forgot to comment here. I'm skeptical that we need a completely new macro, I think many of these changes we can do to xla_test right now (e.g. replacing `gpu_any` with something specific). My preference is a change which has a similar effect but keeps the same macro everywhere. Also interested in what @beckerhe has to say

sergey-kozub:
It is difficult to modify the `xla_test` macro in OSS, as it has hidden parts, every change to it would also require Copybara changes.
Adding a new GPU specific macro would help addressing GPU specific issues. Also it's easier to read and understand.
Having said that, if this PR is not approved - that's fine with me, just wanted to improve what we have now.

ddunl:
Agreed on the difficulty of editing from OSS, if we coalesce on what I'm suggesting then I'm happy to make the change to smooth over the difficulty. In particular I think the removal (or replacement) of `gpu_any` is uncontroversial, I just haven't made the change myself yet

sergey-kozub:
Marked as draft, I'll try to find another solution to the issues outlined in the description.

PR #13842, opened 2024-06-17
[ROCm] fixed build due to https://github.com/openxla/xla/commit/19c11…

zoranjovanovic-ns:
…baa83f31e25a3f841cf41fa47a53e8ca161

@xla-rotation 
https://github.com/openxla/xla/blob/main/xla/service/gpu/ir_emitter_triton_rocm.cc was partially updated according to the change, but there was missing include directive in it (because of which ROCm build is broken).


i-chaochen:
@ddunl @bchetioui Please check

ddunl (approved):

bchetioui:
Thanks for the fix, and sorry to have broken you!

PR #13879, opened 2024-06-18
[ROCm] Pin amdhip64 soversion in dso loader

draganmladjenovic:
This prevents us accidentally loading a second copy of HIP runtime in local_config_rocm. Do similar for rocblas to guard against ABI break in rocm 6.0.

### DIFF
--- a/third_party/tsl/third_party/gpus/rocm/rocm_config.h.tpl
+++ b/third_party/tsl/third_party/gpus/rocm/rocm_config.h.tpl
@@ -22,5 +22,7 @@ limitations under the License.
 #define TF_MIOPEN_VERSION %{miopen_version_number}
 #define TF_HIPRUNTIME_VERSION %{hipruntime_version_number}
 #define TF_HIPBLASLT %{hipblaslt_flag}
+#define TF_HIPRUNTIME_SOVERSION "%{hip_soversion_number}"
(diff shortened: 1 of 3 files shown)


dimitar-asenov:
@ddunl What is the right way to submit changes to `third_party/tsl`? Is it via a PR such as this, or should it be a PR on some other repo?

ddunl:
Confusingly really the right repo is TensorFlow for the `third_party/tsl/third_party/gpus`. We still have to patch unfortunately. Once I'm done moving TSL (which is my top priority next month as it's taken way too long) I'll start moving the stuff that we use from `third_party/tensorflow/third_party` to the right place and then we won't have to patch.

dimitar-asenov:
Thanks. I will approve this and we'll patch it internally.

dimitar-asenov (approved):

dimitar-asenov:
OK, my approval didn't do anything, because Copybara doesn't think this PR affects any internal files.

@ddunl To unblock this, what is the right way to proceed? I assume @draganmladjenovic needs to make an analogous PR to a different repo, should this be https://github.com/google/tsl or something else?

dimitar-asenov:
I'm going to see if I can submit an equivalent change internally, so that we don't have to jump through many hoops.

draganmladjenovic:
@dimitar-asenov  Thanks!

PR #13885, opened 2024-06-18
[XLA:GPU] Allow user to change command buffer trace cache capacity

shawnwang18:
Allow user to set the command buffer trace cache size, increasing the cache size may sometimes reduces the chances of doing command buffer tracing for updating command buffer instance.

### DIFF
--- a/xla/service/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/service/gpu/runtime/command_buffer_cmd.cc
@@ -38,6 +38,7 @@ limitations under the License.
 #include "absl/strings/string_view.h"
 #include "absl/synchronization/mutex.h"
 #include "absl/types/span.h"
+#include "xla/debug_options_flags.h"
 #include "xla/executable_run_options.h"
 #include "xla/ffi/call_frame.h"
 #include "xla/ffi/ffi_api.h"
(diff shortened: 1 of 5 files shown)


xla/xla.proto
```diff
@@ -824,7 +824,12 @@ message DebugOptions {
   // unsafe flag.
   bool xla_gpu_unsafe_pipelined_loop_annotator = 309;
 
-  // Next id: 310
+  // The command buffer trace cache size, increasing the cache size may "
+  //  "sometimes reduces the chances of doing command buffer tracing for "
+  //  "updating command buffer instance."
+  int64 xla_cmd_buffer_trace_cache_capacity = 310;
```
ezhulenev, line 830: `xla_cmd_buffer_trace_cache_size` as your refer to it as size anyway

xla/service/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -429,8 +430,13 @@ absl::Status TracedCommandBufferCmd::AddTracedCommandBuffer(
     const Thunk::ExecuteParams& execute_params,
     const RecordParams& record_params, se::CommandBuffer* command_buffer,
     absl::FunctionRef<absl::Status(se::Stream*)> trace) {
-  auto traced_cmd = record_params.state.GetOrCreate<TracedCommandBuffer>(
-      this, [&] { return std::make_unique<TracedCommandBuffer>(buffers()); });
+
+  const auto& debug_options = xla::GetDebugOptionsFromFlags();
```
ezhulenev, line 434: You should move `GetDebugOptionsFromFlags` into lambda because it can be very expensive and now you execute it on every graph re-capture

ezhulenev (approved):

reedwm:
@shawnwang18 can you add a test? Our internal infrastructure now requires all changes have a test, or requires us to add an explanation on why a test is unneeded. And of course, adding tests is good in general.

I have little context on this change, but `xla_cmd_buffer_trace_cache_size` seems to simply affects performance. So if directly testing the buffer trace cache size is difficult, I think writing a test that changes the flag and checks XLA still runs correctly is sufficient.

xla/xla.proto
```diff
@@ -826,7 +826,12 @@ message DebugOptions {
 
   string xla_gpu_per_fusion_autotune_cache_dir = 310;
 
-  // Next id: 311
+  // The command buffer trace cache size, increasing the cache size may "
```
reedwm, line 829: Remove double-quotation at end of line

shawnwang18: Fixed

ezhulenev (approved):

PR #13938, opened 2024-06-19
Set Asynchronous Custom Call Start Thunks

philipphack:
Corrects the types of the thunks emitted for asynchronous GEMM Custom Call starts.

### DIFF
--- a/xla/service/gpu/ir_emitter_unnested.cc
+++ b/xla/service/gpu/ir_emitter_unnested.cc
@@ -1727,14 +1727,14 @@ absl::Status IrEmitterUnnested::EmitAsyncCustomCallStart(
   }
 #if GOOGLE_CUDA || TF_HIPBLASLT
   if (IsCublasLtMatmul(*wrapped)) {
-    auto status = EmitGemmThunk(custom_call);
+    auto status = EmitCublasLtMatmulThunk(custom_call);
     if (status.ok()) {
       thunk_sequence_.back()->set_execution_stream_id(execution_stream_id);
     }
     return status;
   }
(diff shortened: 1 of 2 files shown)


cheshire:
Was it working before? What is the behavior change?

philipphack:
We want to emit cublasLT or cublasLT FP8 thunks based on the type of the wrapped call. Currently, all types of GEMM Custom Calls result in the emission of a legacy cublas thunk. This causes problems since legacy calls only support a subset of the features of cublasLT. For example, FP8 operands are not supported by legacy cublas and lead to an error without this change. Legacy thunks are still emitted [here](https://github.com/openxla/xla/blob/main/xla/service/gpu/ir_emitter_unnested.cc#L1728) when the type of the wrapped call matches.

cheshire:
> For example, FP8 operands are not supported by legacy cublas and lead to an error without this change

Is it possible to write a test then?

philipphack:
Yes, this is meant to be tested by WindowedEinsumE2EAllGatherAndReduceScatterF8 in collective_ops_test_e2e.cc. However, the dot merger pass prevented the GEMM rewriter from generating non-legacy custom calls. I've disabled the dot merger in the test. The test now fails without the changes in ir_emitter_unnested.cc.

cheshire:
But collectives test is an e2e one, perhaps we could have a more isolated test?

ezhulenev (approved):

penpornk:
@philipphack Could you please help rebase? collective_ops_test_e2e has been renamed.

philipphack:
@penpornk Done.

PR #13958, opened 2024-06-19
Prefer vector size as 4 for u8/s8 transpose

ghost:
We observe 4 bytes as per vectorized load can get better performance for transpose, which means 4xi8 is the better choice compared with previous 2xi8.
```
%fused_computation {
    %p0 = u8[8,4096,4096] parameter(0)
    %transpose = u8[8,4096,4096] transpose(%p0), dimensions={0,2,1}
 }
ENTRY main {
    %param = u8[8,4096,4096] parameter(0)
    ROOT %fusion = u8[8,4096,4096] fusion(%param), kind=kInput,
        calls=%fused_computation
}
improve from 270us to 214us, collected on A100 40GB machine.
```


lingzhi98:
The reason of why I remove the result check is that there exists gap between IR check and result check. The  IR check base on RTXA6000Device, but result check use current registered device. So we can find the result check does not run transpose kernel with expected vec size, due to the hlo is designed for RTXA6000Device and transpose emitter choose different vec size by device info. Alternatively, we can redesign the hlo tested for vectorized transpose, but it is hard to find the hlo which can run with expected vec size on any kind of device.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -126,13 +127,22 @@ MlirTransposeFusion::MlirTransposeFusion(const HloFusionAnalysis& analysis)
...
   // for large data types does not help (there's already enough parallelism).
   const auto& device = analysis_.device_info();
-  bool enough_work = Product(block_counts_) * kNumThreadsPerBlock >=
-                     4 * device.core_count() * device.threads_per_core_limit();
-  bool enough_shmem = shmem_usage * 4 <= device.shared_memory_per_block();
-  bool aligned_dims =
-      (input_shape_[2] % 2 == 0) && (input_shape_[permutation_[2]] % 2 == 0);
-  if (max_element_bytes < 4 && enough_work && enough_shmem && aligned_dims) {
-    compute_block_sizes(2);
+  int prefered_vec_size = kMaxVectorizedBytes / max_element_bytes;
```
jreiffers, line 130: Nit: `prefered_vec_size` is a slightly unfortunate name for a variable that gets modified (and has a typo). How about this?

```
for (int vec_size = kMaxVectorizedBytes / max_element_bytes;
     vec_size > 1; vec_size /= 2) {
  ...
}
```

jreiffers (changes requested):

jreiffers:
Looks good, just a nit.

jreiffers (approved):

jreiffers:
Thank you!

lingzhi98:
Can I know If there exists any plan to adjust the block size of transpose? I have seen high register pressure in some cases which cause vectorization can not get expected performance though still better than non-vectorization. I have implemented one approach to adjust block size based on registers_per_core and registers_usager_per_block to get performance speedup. If openxla has no plan now, I will test more case and submit it. If has, can I know the latest status?

jreiffers:
We're aware of the issue but no one is actively working on it, so contributions would be appreciated there. There is some work for better reduction block sizes, but it's early stages.

PR #13985, opened 2024-06-20
Removing spurious `option go_package` from autotuning.proto

janpfeifer:
Generally, **bazel** uses the `M` option for the Go proto generator -- see https://protobuf.dev/reference/go/go-generated/#invocation

Unfortunately, hard-coding `option go_package` here prevents other libraries from using the `.proto` file by compiling the Go bindings in their own package structure.

Removing it makes it friendlier for downstream Go users of the XLA library.

Plus I don't see it used anywhere in the library, so better remove things not used -- and it's not defined in any of the other `.proto` files.

### DIFF
--- a/xla/autotuning.proto
+++ b/xla/autotuning.proto
@@ -11,8 +11,6 @@ import "google/protobuf/any.proto";
 import "google/protobuf/duration.proto";
 import "tsl/protobuf/dnn.proto";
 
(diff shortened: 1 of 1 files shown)


ddunl:
It seems like these were originally added here: https://github.com/tensorflow/tensorflow/pull/36387/files#diff-f61248520e180396d513997bb629557d5b63af95bea51506a004584fc49302c7. I can't find any internal users either, so LGTM! Thanks!!

ddunl (approved):

PR #14073, opened 2024-06-22
Add select(compare(a, b, GT/GE), a, b) => or(a, b) to algsimp

apivovarov:
In one of the customer's HLOs (in the reduceOp computation function), I found the following pattern:
```c
p0 = pred[]{0} parameter(0)
p1 = pred[]{0} parameter(1)
compare = pred[]{0} compare(p0, p1), direction=GT
select = pred[]{0} select(compare, p0, p1)
```
It can be simplified to `logical_or`.

This PR adds the following patterns to algsimp
```c
select(compare(a, b, GT/GE), a, b) => or(a, b)
select(compare(a, b, LT/LE), a, b) => and(a, b)
select(compare(a, b, EQ), a, b) => b
select(compare(a, b, NE), a, b) => a

a,b ∈ PRED
```

### DIFF
--- a/xla/service/algebraic_simplifier.cc
+++ b/xla/service/algebraic_simplifier.cc
@@ -8077,6 +8077,33 @@ absl::Status AlgebraicSimplifierVisitor::HandleSelect(HloInstruction* select) {
                       select->mutable_operand(0)->shape(), HloOpcode::kNot,
                       select->mutable_operand(0)));
     }
+    // select(compare(a, b, GT/GE), a, b) => or(a, b)
+    // select(compare(a, b, LT/LE), a, b) => and(a, b)
+    // select(compare(a, b, EQ), a, b) => b
+    // select(compare(a, b, NE), a, b) => a
+    HloInstruction *compare, *lhs, *rhs;
+    if (Match(select, m::Select(m::Op(&compare), m::Op(&lhs), m::Op(&rhs))) &&
+        Match(compare, m::Compare(m::Op().Is(lhs), m::Op().Is(rhs)))) {
+      auto cmp_dir = compare->comparison_direction();
+      if (cmp_dir == ComparisonDirection::kGt ||
+          cmp_dir == ComparisonDirection::kGe) {
+        return ReplaceWithNewInstruction(
+            select, HloInstruction::CreateBinary(select->shape(),
+                                                 HloOpcode::kOr, lhs, rhs));
+      }
+      if (cmp_dir == ComparisonDirection::kLt ||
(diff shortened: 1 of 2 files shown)


xla/service/algebraic_simplifier_test.cc
```diff
@@ -736,6 +736,74 @@ TEST_F(AlgebraicSimplifierTest, SelectPredPred2) {
...
 }
 
+// select(compare(a, b, GT/GE), a, b) => or(a, b),   a,b ∈ PRED
+TEST_F(AlgebraicSimplifierTest, SelectGtCompare) {
+  const char* kModuleStr = R"(
+    HloModule m
+    test {
+      p0 = pred[8]{0} parameter(0)
+      p1 = pred[8]{0} parameter(1)
+      compare = pred[8]{0} compare(p0, p1), direction=GT
```
db7894, line 746: Maybe for thoroughness, worth testing both `GT` and `GE`? Same for below.

apivovarov: fixed

db7894 (approved):
Thanks for adding the other inequality tests, LGTM

xla/service/algebraic_simplifier.cc
```diff
@@ -8077,6 +8077,33 @@ absl::Status AlgebraicSimplifierVisitor::HandleSelect(HloInstruction* select) {
                       select->mutable_operand(0)->shape(), HloOpcode::kNot,
                       select->mutable_operand(0)));
     }
+    // select(compare(a, b, GT/GE), a, b) => or(a, b)
+    // select(compare(a, b, LT/LE), a, b) => and(a, b)
+    // select(compare(a, b, EQ), a, b) => b
+    // select(compare(a, b, NE), a, b) => a
+    HloInstruction *lhs, *rhs, *compare;
+    if (Match(select, m::Select(m::Compare(&compare, m::Op(&lhs), m::Op(&rhs)),
+                                m::Op(&lhs), m::Op(&rhs)))) {
```
akuegel, line 8086: I think this usage of the matcher does not make sure that the first operand of the compare is the same as the second operand of the select. Same for second operand of compare/third operand of select. This needs to be tested separately. Please add a test case where this would be caught.

apivovarov: Good catch! Fixed. 
Now I call `Match()` two times - for `select` and for `compare` ops separately. The second Match call uses `m::Op().Is(lhs), m::Op().Is(rhs)` to make sure the operands order in `compare` Op is the same as in `select` Op.

akuegel (approved):
Thank you

PR #14074, opened 2024-06-22
Add sqrt(x) * sqrt(x) => x to algsimp

apivovarov:
This PR adds pattern `sqrt(x) * sqrt(x) => x , for x >= 0`
Validation - the following checks are valid both before and after simplification.:
```c
If x == 0 the result is 0
If x > 0 the result > 0
If x is inf the result is inf
If x is nan the result is nan
```
Related PR - https://github.com/openxla/xla/pull/13771

### DIFF
--- a/xla/service/algebraic_simplifier.cc
+++ b/xla/service/algebraic_simplifier.cc
@@ -4569,6 +4569,13 @@ absl::Status AlgebraicSimplifierVisitor::HandleMultiply(
         HloInstruction::CreateUnary(multiply->shape(), HloOpcode::kExp, add));
   }
 
+  VLOG(10) << "trying transform [sqrt(x) * sqrt(x) => x], for x >= 0 "
+           << multiply->ToString();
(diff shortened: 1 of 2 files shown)


golechwierowicz (approved):

xla/service/algebraic_simplifier.cc
```diff
@@ -4569,6 +4569,13 @@ absl::Status AlgebraicSimplifierVisitor::HandleMultiply(
         HloInstruction::CreateUnary(multiply->shape(), HloOpcode::kExp, add));
   }
 
+  VLOG(10) << "trying transform [sqrt(x) * sqrt(x) => x], for x >= 0 "
+           << multiply->ToString();
+  if (Match(multiply, m::Multiply(m::Sqrt(m::Op(&a)), m::Sqrt(m::Op(&a)))) &&
```
akuegel, line 4574: This line has the same problem as in your other PR, you don't verify that a is the same. Can you please send a followup PR to fix this?

akuegel: Actually I found that the same issue already exists just below (since 2019), so I am going to fix this myself.

PR #14086, opened 2024-06-23
[MLIR][MHLO] Fix PadEmptyTensor pattern to return success()

apivovarov:
Currently, the `PadEmptyTensor` and `DynamicPadEmptyTensor` patterns incorrectly return `failure()` after `rewriter.replaceOpWithNewOp`. This appears to be a typo, as the usual practice is to return `success()` following `rewriter.replaceOpWithNewOp`.

### DIFF
--- a/xla/mlir_hlo/mhlo/IR/hlo_ops.cc
+++ b/xla/mlir_hlo/mhlo/IR/hlo_ops.cc
@@ -4219,16 +4219,16 @@ struct PadEmptyTensor : public OpRewritePattern<PadOp> {
 
     llvm::SmallVector<Value> reifiedShapes;
     if (failed(op.reifyReturnTypeShapes(rewriter, op.getOperands(),
-                                        reifiedShapes)))
+                                        reifiedShapes))) {
       return failure();
+    }
 
     auto dimsType = RankedTensorType::get({0}, rewriter.getIntegerType(64));
     auto broadcastDims =
         DenseIntElementsAttr::get(dimsType, SmallVector<int64_t, 1>{});
     rewriter.replaceOpWithNewOp<mhlo::DynamicBroadcastInDimOp>(
         op, op.getType(), padVal, reifiedShapes.front(), broadcastDims);
-
-    return failure();
+    return success();
   }
 };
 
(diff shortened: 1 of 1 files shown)


d0k (approved):

apivovarov:
Just want to check if this PR is ready to be merged? @d0k @penpornk

apivovarov:
Benjamin, Penporn, Can you look at this PR? It seems to be stuck. I think it is ready to be merged. @d0k @penpornk

d0k (approved):

d0k:
Looks like my approval isn't enough somehow? @penpornk do you know what's going on?

apivovarov:
I just rebased this PR branch from the latest main

akuegel:
I wonder whether it is possible to test this? Maybe we didn't have any test coverage for this pattern?

reedwm (changes requested):
We have  a presubmit that requires all changes have a test, or a justification for why no test is needed. Can a test be added?

Also any ideas why this is not already failing a test (or a real model)? I would expect returning `failure()` would cause something to fail. @d0k do you know

reedwm:
Ah @akuegel commented the same thing as I did right before me :)

akuegel:
> I wonder whether it is possible to test this? Maybe we didn't have any test coverage for this pattern?

I just checked, the test coverage is through https://github.com/openxla/xla/blob/main/xla/mlir_hlo/tests/Dialect/mhlo/canonicalize/canonicalize.mlir#L2362

This really executes the code, but somehow the test doesn't fail although we return failure. So that indicates that it may be difficult to test this behavior change.

akuegel:
> > I wonder whether it is possible to test this? Maybe we didn't have any test coverage for this pattern?
> 
> I just checked, the test coverage is through https://github.com/openxla/xla/blob/main/xla/mlir_hlo/tests/Dialect/mhlo/canonicalize/canonicalize.mlir#L2362
> 
> This really executes the code, but somehow the test doesn't fail although we return failure. So that indicates that it may be difficult to test this behavior change.

I have asked in our mlir chat room what is going on here. Maybe someone knows.

reedwm:
From the chat thread, it seems testing cannot be done without creating C++ code to run the pass. I don't think it's worth trying to test this, considering this change has no functional impact, and AFAIK it's very common to test passes via MLIR files without testing whether the pass returns success() vs failure(). I'll try merging.

PR #14087, opened 2024-06-23
Implement GetOutputMemoryKinds for GPU executables

jaro-sevcik:


jyingl3:
cc @yashk2810

jyingl3 (commented):
Just one minor suggestion.

xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3234,9 +3235,57 @@ PjRtStreamExecutorLoadedExecutable::GetHloModules() const {
...
+
+}  // namespace
+
 absl::StatusOr<std::vector<std::vector<absl::string_view>>>
 PjRtStreamExecutorLoadedExecutable::GetOutputMemoryKinds() const {
-  return Unimplemented("GetOutputMemoryKinds is not supported.");
+  TF_ASSIGN_OR_RETURN(auto shapes, GetOutputShapes());
+  TF_ASSIGN_OR_RETURN(PjRtMemorySpace * default_memory_space,
+                      addressable_devices()[0]->default_memory_space());
+  std::vector<std::vector<absl::string_view>> out;
```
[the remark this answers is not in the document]
jaro-sevcik: Done, thank you!

I have also rearranged the MemoryKindsFromShape function so that it constructs the one element vector directly, rather than by pushing.

jyingl3 (approved):

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -800,24 +800,35 @@ absl::StatusOr<std::unique_ptr<PjRtBuffer>> CreateDeviceBufferForTest(
   return input;
 }
 
+constexpr char const* kD2HProgram = R"(
```
yliu120, line 803: maybe constexpr char* is enough? or constexpr char kD2HProgram[]?

Please note that this is compile-time literal so no need for the extra const. It might be better to align with the other location of the code base.

jaro-sevcik: > Please note that this is compile-time literal so no need for the extra const. It might be better to align with the other location of the code base.

It is unclear what to align with - I actually took the same incantation that is already used in this very file (see line 99 above).

PR #14089, opened 2024-06-23
Handle allocation in memory space in GPU's BufferFromHostBuffer

jaro-sevcik:


xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -431,18 +432,46 @@ AllocateDestinationBuffer(
...
+      result = transfer_manager->HostShapeToDeviceShape(result);
+    }
+    return result;
+  };
+
+  TF_ASSIGN_OR_RETURN(
+      ScopedShapedBuffer dst_buffer,
+      transfer_manager->AllocateScopedShapedBuffer(
+          on_host_shape, se_client->allocator(),
+          local_device->local_device_id().value(), memory_space_shape_fn));
```
jyingl3, line 474: do you need to update https://github.com/openxla/xla/blob/b700711f832e87bcc2c3960f879e287f98b9737b/xla/stream_executor/stream_executor.h#L370-L378 as well? it looks like memory_limit_bytes check does not apply to kHostMemorySpace?

jyingl3:
cc @yashk2810 and @ezhulenev

xla/pjrt/pjrt_stream_executor_client.cc
jaro-sevcik: I am not sure if the limit applies or not. Arguably the change to checking the limit is independent of this patch - the StreamExecutor::AllocateArray method is already getting the memory space as argument, so it should do the appropriate checking if that's important.

How would you feel about not updating the AllocateArray method in this patch?

xla/pjrt/pjrt_stream_executor_client.h
```diff
@@ -376,6 +376,14 @@ class PjRtStreamExecutorClient : public PjRtClient {
       absl::AnyInvocable<void() &&> on_done_with_host_buffer,
       PjRtMemorySpace* memory_space, const Layout* device_layout) override;
 
+  absl::StatusOr<std::unique_ptr<PjRtBuffer>> BufferFromHostBuffer(
+      const void* data, PrimitiveType type, absl::Span<int64_t const> dims,
+      std::optional<absl::Span<int64_t const>> byte_strides,
+      HostBufferSemantics host_buffer_semantics,
+      absl::AnyInvocable<void() &&> on_done_with_host_buffer,
+      PjRtDevice* device, const Layout* device_layout,
+      PjRtMemorySpace* memory_space);
```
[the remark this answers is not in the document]
jaro-sevcik: I do not see how we could implement that variant and still support the mode with JAX memories disabled.

If JAX memories are disabled, then `memory_space` would have to be null, so the device-less overload would not know for which device to allocate. Or is there some way to extract the device from the other parameters even if `memory_space` is null?

jaro-sevcik (commented):
Thank you for the review. I have some questions (in the comments).

xla/pjrt/pjrt_stream_executor_client.h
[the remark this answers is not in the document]
jyingl3: JAX/jaxlib is device-agnostic and only calls methods in base pjrt_client.h. When the sharding has memory_kind, the variant with memory_space and layout is called https://github.com/openxla/xla/blob/223bb067ca99000aa25398b787a9be58e6199180/xla/python/pjrt_ifrt/pjrt_client.cc#L581, so I guess you will want to implement that variant to fully support memory space.

Sorry my initial comment was not clear. I am not suggesting removing the implementation of device only variant. Before PjRtDevice is migrated to PjRtMemorySpace, the two variants (1. device only 2. device and optional layout) are still needed. What I mean is to do some more refactoring based on current commit so that the three variants in the base pjrt_client.h is implemented, instead of adding a new variant to pjrt_stream_executor_client.h.

Please let me know if more clarification is needed!

xla/pjrt/pjrt_stream_executor_client.cc
jyingl3: I was thinking skip this check if the memory_space is kHostMemorySpace? Current memory_limit_bytes is only for device memory. It is possible that this check will fail at `static_cast<int64_t>(bytes) > memory_limit_bytes`, however, the bytes will be allocated from host memory and should not matter if it is larger than memory_limit_bytes. What do you think? 

@ezhulenev Any suggestion?

xla/pjrt/pjrt_stream_executor_client.h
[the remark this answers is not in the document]
jaro-sevcik: I am still confused, I am afraid. As far as I can see, we do implement the three variants already:

1. Device only at [xla/pjrt/pjrt_stream_executor_client.cc, line 1037](https://github.com/openxla/xla/pull/14089/files#diff-84be417aca12b5e1527c22c6fb6ad3d8fab372f91214bb1cc8e1d93b28b12f6cR1037).
2. Device + layout at [xla/pjrt/pjrt_stream_executor_client.cc, line 1050](https://github.com/openxla/xla/pull/14089/files#diff-84be417aca12b5e1527c22c6fb6ad3d8fab372f91214bb1cc8e1d93b28b12f6cR1050).
3. Memory space + layout at [xla/pjrt/pjrt_stream_executor_client.cc, line 1063](https://github.com/openxla/xla/pull/14089/files#diff-84be417aca12b5e1527c22c6fb6ad3d8fab372f91214bb1cc8e1d93b28b12f6cR1063).

In fact, the memory space + layout is the one that the newly introduced test exercises ([xla/pjrt/gpu/se_gpu_pjrt_client_test.cc, line 795](https://github.com/openxla/xla/pull/14089/files#diff-ce4a6f9e304d9bac11e6cacffa07b57f584ad705009e41aded11b4146ae536fdR795)).

All those variants eventually end up in this new variant because none of the existing variants seems to get the parameters that apply in all scenarios (in particular, the common variant needs both the memory space *and* device parameters). As a result, out of the three existing variants, there does not seem to be one generic one that the other two could relay to. Am I missing something?

xla/pjrt/pjrt_stream_executor_client.cc
jaro-sevcik: I don't disagree, I just think that should be a different PR. Would it help if I write such a PR now?

jaro-sevcik (commented):
Thank you for taking a look, more questions in the comments.

xla/pjrt/pjrt_stream_executor_client.cc
jyingl3: A different PR sounds good. Sorry I misunderstood the earlier reply.

xla/pjrt/pjrt_stream_executor_client.h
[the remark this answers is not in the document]
jyingl3: Sorry I misread some of the code. Does the new variant needs to be a public function of PjRtStreamExecutorClient? Can it be a free helper method that is not part of PjRtStreamExecutorClient? I just think adding one more variant is confusing.

I am fine with moving it to private if making it a free helper method does not worth the effort.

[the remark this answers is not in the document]
jaro-sevcik: A free helper function would need access to many internals of the client (`client()`, `thread_pool()`, `host_memory_allocator()`, `transpose_mu_` mutex, `ScopeHold::ToClosure` type, ...); it looks like it would be a major refactoring to make the method free. Also, PjrtStreamExecutorClient does not have any private section, only a protected section (even for what should normally be private fields). I hope it's ok if I make the method only protected and equip it with 'Internal' suffix.

jaro-sevcik (commented):
Could you take another look please?

jyingl3 (approved):

xla/pjrt/pjrt_stream_executor_client.cc
jaro-sevcik: Just to follow up here, it looks like the host memory allocator bypasses the AllocateArray method. Instead of AllocateArray, it call HostMemoryAllocate, see [here](https://github.com/openxla/xla/blob/5f4c1ebb3ef1b141e0f47452867c8c8dbc066983/xla/stream_executor/integrations/device_host_allocator.h#L59). 

As a result, I believe there is no need to worry about that limit check.

PR #14090, opened 2024-06-23
Propagate memory space from MLIR to the compiler on GPUs

jaro-sevcik:
Let us propagate memory space annotations on parameters and result from MLIR module to Hlo using existing utils functions (which are currently only used by TPUs). While we are there, we also hook up the layout mode annotations.

### DIFF
--- a/xla/pjrt/pjrt_stream_executor_client.cc
+++ b/xla/pjrt/pjrt_stream_executor_client.cc
@@ -3461,6 +3461,12 @@ PjRtStreamExecutorClient::Compile(mlir::ModuleOp module,
       /*use_tuple_args=*/options.parameter_is_tupled_arguments,
       /*return_tuple=*/false));
 
+  // If the compile options specify argument layout, then let's
+  // fall back to using the options to determine layouts.
+  if (options.argument_layouts) {
+    return Compile(xla_computation, options);
+  }
+
   TF_ASSIGN_OR_RETURN(std::vector<LayoutMode> arg_layout_modes,
                       GetArgLayoutModes(module));
   TF_ASSIGN_OR_RETURN(std::vector<LayoutMode> out_layout_modes,
--- a/xla/pjrt/pjrt_stream_executor_client.cc
+++ b/xla/pjrt/pjrt_stream_executor_client.cc
@@ -3439,6 +3439,27 @@ PjRtStreamExecutorClient::GetExecutableExtras(CompileOptions* options) {
 absl::StatusOr<std::unique_ptr<PjRtLoadedExecutable>>
 PjRtStreamExecutorClient::Compile(const XlaComputation& computation,
                                   CompileOptions options) {
+  auto input_options = options;
+  std::vector<const Shape*> argument_layout_pointers;
+  TF_RETURN_IF_ERROR(DetermineArgumentLayoutsFromCompileOptions(
+      computation,
+      [local_client = client()](Shape shape) {
+        return local_client->backend()
+            .transfer_manager()
+            ->ChooseCompactLayoutForShape(shape);
+      },
+      options.argument_layouts, &options.executable_build_options,
+      &argument_layout_pointers));
+  return Compile(computation, argument_layout_pointers,
+                 *(options.executable_build_options.result_layout()),
+                 input_options);
+}
+
+absl::StatusOr<std::unique_ptr<PjRtLoadedExecutable>>
+PjRtStreamExecutorClient::Compile(
+    const XlaComputation& computation,
+    const std::vector<const Shape*>& argument_layout_pointers,
+    const Shape& result_layout, CompileOptions options) {
   tsl::profiler::TraceMe traceme("PjRtStreamExecutorClient::Compile");
   VLOG(1) << "PjRtStreamExecutorClient::Compile";
   options.executable_build_options.set_process_index(process_index());
@@ -3457,16 +3478,7 @@ PjRtStreamExecutorClient::Compile(const XlaComputation& computation,
       addressable_device_logical_ids = extras.addressable_device_logical_ids;
   std::vector<PjRtDevice*>& addressable_devices = extras.addressable_devices;
 
-  std::vector<const Shape*> argument_layout_pointers;
-  TF_RETURN_IF_ERROR(DetermineArgumentLayoutsFromCompileOptions(
-      computation,
-      [local_client = client()](Shape shape) {
-        return local_client->backend()
-            .transfer_manager()
-            ->ChooseCompactLayoutForShape(shape);
-      },
-      options.argument_layouts, &options.executable_build_options,
-      &argument_layout_pointers));
+  options.executable_build_options.set_result_layout(result_layout);
 
   TF_ASSIGN_OR_RETURN(
       std::vector<std::unique_ptr<LocalExecutable>> local_executables,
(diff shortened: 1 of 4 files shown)


xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3393,7 +3405,32 @@ PjRtStreamExecutorClient::Compile(mlir::ModuleOp module,
...
+  TF_ASSIGN_OR_RETURN(std::vector<LayoutMode> out_layout_modes,
+                      GetOutputLayoutModes(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> arg_memory_spaces,
+                      GetArgMemoryKinds(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> out_memory_spaces,
+                      GetOutputMemoryKinds(module));
+
+  Shape result_layout;
+  ExecutableBuildOptions build_options = options.executable_build_options;
+  TF_ASSIGN_OR_RETURN(auto arg_shapes_and_layouts,
```
yashk2810, line 3420: why is it called `arg_shapes_and_layouts`? It includes the results too right? Call it something like: `arg_and_result_layouts`?

jaro-sevcik: Results are buried somewhere in [build options](https://github.com/openxla/xla/blob/f052a450327bc4fb73dfd3685b6e4d176859d157/xla/pjrt/utils.cc#L705). Updated the name to `arg_layouts_and_pointers`.

jaro-sevcik:
@yashk2810 Could you take another look, please?

xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3448,7 +3460,32 @@ PjRtStreamExecutorClient::Compile(mlir::ModuleOp module,
...
+                          [this](Shape shape) -> absl::StatusOr<Shape> {
+                            return this->client()
+                                ->backend()
+                                .transfer_manager()
+                                ->ChooseCompactLayoutForShape(shape);
+                          },
+                          build_options));
+
+  return Compile(xla_computation, arg_layouts_and_pointers.second,
+                 *(build_options.result_layout()), options);
```
yashk2810, line 3488: What's `build_options.result_layout`? The out_layouts we care about are the layouts and memory_spaces you calculated on line 3470 and 3466.

jaro-sevcik: For some reason, the LayoutModesToXla function stores the result into build_options, see here: https://github.com/openxla/xla/blob/5e5a794c256b91695f90f7502185aa56cf80b66c/xla/pjrt/utils.cc#L705

The layouts calculated on line 3470 and 3466 need to be converted to XLA, that's what LayoutModesToXla function is doing.

jaro-sevcik: TBH, the LayoutModesToXla function has some weird contract, maybe we should be using  LayoutModesToXlaShapes?

xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3448,7 +3460,32 @@ PjRtStreamExecutorClient::Compile(mlir::ModuleOp module,
...
+                      GetArgLayoutModes(module));
+  TF_ASSIGN_OR_RETURN(std::vector<LayoutMode> out_layout_modes,
+                      GetOutputLayoutModes(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> arg_memory_spaces,
+                      GetArgMemoryKinds(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> out_memory_spaces,
+                      GetOutputMemoryKinds(module));
+
+  Shape result_layout;
+  ExecutableBuildOptions build_options = options.executable_build_options;
```
jyingl3, line 3474: If LayoutModesToXla always updates options.build_options.result_layout() in place, you do not need to pass result_layout to Compile, and read it from options instead. It does change the assumption of the newly added Compile method but it may be more consistent if you are using LayoutModesToXla. What do you think?

jaro-sevcik: The original version of Compile seems to go out of its way to keep the options intact (it takes a copy of the options early on and that unchanged copy is stored into the executable), so I was concerned that changing the options might have some bad consequences. If you think it's ok to mutate the options, I am happy to take that path.

jaro-sevcik:
@jyingl3 Could you take another look?

I have removed the result_layout parameter (passing the result_layout in options) and made the new Compile method variant protected.

xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3448,7 +3460,32 @@ PjRtStreamExecutorClient::Compile(mlir::ModuleOp module,
...
+  TF_ASSIGN_OR_RETURN(std::vector<LayoutMode> arg_layout_modes,
+                      GetArgLayoutModes(module));
+  TF_ASSIGN_OR_RETURN(std::vector<LayoutMode> out_layout_modes,
+                      GetOutputLayoutModes(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> arg_memory_spaces,
+                      GetArgMemoryKinds(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> out_memory_spaces,
+                      GetOutputMemoryKinds(module));
+
+  Shape result_layout;
```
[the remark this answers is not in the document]
jaro-sevcik: Removed now.

xla/pjrt/pjrt_stream_executor_client.cc
jaro-sevcik: [Resolving, let's stick to LayoutModesToXla.]

xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3439,6 +3439,58 @@ PjRtStreamExecutorClient::GetExecutableExtras(CompileOptions* options) {
...
+                      GetArgLayoutModes(module));
+  TF_ASSIGN_OR_RETURN(std::vector<LayoutMode> out_layout_modes,
+                      GetOutputLayoutModes(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> arg_memory_spaces,
+                      GetArgMemoryKinds(module));
+  TF_ASSIGN_OR_RETURN(std::vector<MemorySpaceColor> out_memory_spaces,
+                      GetOutputMemoryKinds(module));
+
+  TF_ASSIGN_OR_RETURN(auto arg_layouts_and_pointers,
+                      LayoutModesToXla(
```
jyingl3, line 3474: Add a comment this method will update the result_layout in options.executable_build_options?

I will approve right after the comment is added (cannot approve right now as it will trigger pulling of the change)

jyingl3:
> @jyingl3 Could you take another look?
> 
> I have removed the result_layout parameter (passing the result_layout in options) and made the new Compile method variant protected.

Thank you! Just one more minor suggestion about adding some comments. Will approve as soon as the comment is added.

jaro-sevcik:
Thank you for the review, I've added the comment.

jyingl3 (approved):

reedwm (changes requested):
I can't yet merge since there are two test failures.

In `se_gpu_pjrt_client_test.cc`, `TF_ASSIGN_OR_RETURN` is called, but tests cannot return a Status. Use `TF_ASSERT_OK_AND_ASSIGN` instead.

The test `xla_client_test.py` is failing with

```
  File "/build/work/8a6b11f4211fb6bfb014b3e6f705065fc499/google3/runfiles/google3/third_party/tensorflow/compiler/xla/python/xla_client_test.py", line 799, in testSetArgumentLayoutsLegacy
    self.assertEqual(
AssertionError: Tuples differ: (2, 1, 0) != (1, 2, 0)

First differing element 0:
2
1

- (2, 1, 0)
+ (1, 2, 0)
``` 

@jaro-sevcik, can you fix these issues?

jaro-sevcik:
I have fixed the TF_ASSIGN_OR_RETURN and attempted a fix of the legacy layouts.

Unfortunately, I have not managed to run the xla_client_test.py test, but I have reproduced the problem with a similar C++ test (and added layout tests to this PR).

Could you please try again?

akuegel (approved):

PR #14092, opened 2024-06-24
NVTX: name threads, CUDA devices and CUDA streams

olupton:
See https://github.com/openxla/xla/pull/13603, which landed and got rolled back.
f75962e80d387f32dc9055cd1fff9029d97f0026 attempts to fix the issue described in https://github.com/openxla/xla/pull/13603#discussion_r1647078940.

### DIFF
--- a/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
+++ b/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
@@ -14,16 +14,37 @@ limitations under the License.
 ==============================================================================*/
 #include "tsl/profiler/lib/nvtx_utils.h"
 
+#include <sys/syscall.h>
+
 #include <cstddef>
 #include <cstdint>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <type_traits>
 
 #include "nvtx3/nvToolsExt.h"
+#include "nvtx3/nvToolsExtCuda.h"
+#include "nvtx3/nvToolsExtCudaRt.h"
 #include "nvtx3/nvToolsExtPayload.h"
+#include "third_party/gpus/cuda/include/cuda.h"
+
+namespace {
+// Get the ID of the current thread following the convention for
+// nvtxNameOsThreadA:
+// https://nvidia.github.io/NVTX/doxygen/group___r_e_s_o_u_r_c_e___n_a_m_i_n_g.html
+// This convention may not match the one in tsl::Env::GetCurrentThreadId().
+std::optional<uint32_t> GetCurrentThreadId() {
+#ifdef __linux__
+  return syscall(SYS_gettid);
+#else
+  return std::nullopt;
+#endif
+}
+}  // namespace
 
 namespace tsl::profiler {
+static_assert(std::is_pointer_v<CUstream>);
 static_assert(std::is_pointer_v<nvtxDomainHandle_t>);
 static_assert(std::is_pointer_v<nvtxStringHandle_t>);
 
@@ -31,6 +52,20 @@ ProfilerDomainHandle DefaultProfilerDomain() {
   static ProfilerDomainHandle domain =
       reinterpret_cast<ProfilerDomainHandle>(nvtxDomainCreateA("TSL"));
   return domain;
+}
+
+void NameCurrentThread(const std::string& thread_name) {
+  if (std::optional<uint32_t> tid = GetCurrentThreadId(); tid.has_value()) {
+    nvtxNameOsThreadA(*tid, thread_name.c_str());
+  }
+}
+
+void NameDevice(int device_id, const std::string& device_name) {
+  nvtxNameCudaDeviceA(device_id, device_name.c_str());
+}
+
+void NameStream(StreamHandle stream, const std::string& stream_name) {
+  nvtxNameCuStreamA(reinterpret_cast<CUstream>(stream), stream_name.c_str());
 }
 
 void RangePop(ProfilerDomainHandle domain) {
--- a/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
+++ b/third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
@@ -14,7 +14,9 @@ limitations under the License.
 ==============================================================================*/
 #include "tsl/profiler/lib/nvtx_utils.h"
 
+#ifdef __linux__
 #include <sys/syscall.h>
+#endif
 
 #include <cstddef>
 #include <cstdint>
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -963,6 +963,26 @@ GetStreamExecutorGpuDeviceAllocator(
                                                   std::move(allocators));
 }
 
+// Name the devices and threads that launch work on them. Note: the launcher
+// thread is only used if there are multiple devices driven by a single process.
+void NameDeviceAndLauncherThread(const LocalTopologyProto& node,
+                                 const DeviceProto& device_proto,
+                                 WorkerThread* launcher_thread) {
+  auto suffix = absl::StrFormat(":#global=%d,local=%d,process=%d,slice=%d#",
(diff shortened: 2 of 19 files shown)


third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
```diff
@@ -14,16 +14,37 @@ limitations under the License.
...
+#include "third_party/gpus/cuda/include/cuda.h"
+
+namespace {
+// Get the ID of the current thread following the convention for
+// nvtxNameOsThreadA:
+// https://nvidia.github.io/NVTX/doxygen/group___r_e_s_o_u_r_c_e___n_a_m_i_n_g.html
+// This convention may not match the one in tsl::Env::GetCurrentThreadId().
+std::optional<uint32_t> GetCurrentThreadId() {
+#ifdef __linux__
+  return syscall(SYS_gettid);
```
cheshire, line 39: Sorry I'm not an expert here, is there really no better way to get a thread id?

cheshire (changes requested):
Any ideas if it's possible to test this? Maybe now that we've separated interface, might make sense to have a "mock" one which would record the calls, and test that?

third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
olupton: See discussion on the original PR here: https://github.com/openxla/xla/pull/13603#discussion_r1640777378
tl;dr this comes from the NVTX documentation https://nvidia.github.io/NVTX/doxygen/group___r_e_s_o_u_r_c_e___n_a_m_i_n_g.html#gaeb7d5b25e1147fc4aaf1f3acb8e719d0

beckerhe (commented):
The build target change looks good to me. Thank you for that!

third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
```diff
@@ -14,16 +14,37 @@ limitations under the License.
 ==============================================================================*/
 #include "tsl/profiler/lib/nvtx_utils.h"
 
+#include <sys/syscall.h>
```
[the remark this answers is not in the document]
olupton: Good spot, done 👍

olupton:
@cheshire after the discussion on Tue, can we move ahead with this?
I just rebased to resolve a trivial conflict.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1030,6 +1031,23 @@ absl::StatusOr<DeviceTopologyPair> BuildDistributedDevices(
         TF_RET_CHECK(it->second != nullptr);
         local_device = std::move(it->second);
         gpu_device_ids[device_proto.local_device_ordinal()] = global_device_id;
+        // Assign some descriptive names for profiling tools.
```
cheshire, line 1034: This function is already massive. Maybe we could extract the profiler logic into a subroutine?

third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
```diff
@@ -33,6 +56,20 @@ ProfilerDomainHandle DefaultProfilerDomain() {
   return domain;
 }
 
+void NameCurrentThread(const std::string& thread_name) {
```
cheshire, line 59: Nit: convention is absl::string_view

cheshire (changes requested):
Thanks ! Minor nits.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
olupton: Sure, done in ae594fa862.

third_party/tsl/tsl/profiler/lib/nvtx_utils.cc
olupton: We need a C-style string for the NVTX API, so this would require an extra copy - I believe this is why the other `tsl::profiler` methods don't take `absl::string_view`, and I chose to be consistent here. I'll leave this for now but can change if you prefer.

(Of course, we don't often name devices/threads, so the cost of the copy is probably not relevant for these methods specifically)

cheshire (approved):

reedwm:
Is it feasible to test this? All changes requires a test or an explanation for why no test is necessary.

Is it possible to programmatically query the NVTX information, or can only NVIDIA tools like nsys do this?

olupton:
Rebased to resolve conflicts.

> Is it feasible to test this? All changes requires a test or an explanation for why no test is necessary.
> 
> Is it possible to programmatically query the NVTX information, or can only NVIDIA tools like nsys do this?

We discussed this in the meeting last Tue; in the shorter term we will make sure downstream NVIDIA CI is exercising this functionality.

To answer the question, it is _possible_ to subscribe to receive information passed via NVTX. [XLA knows how to do this for a single call](https://github.com/openxla/xla/blob/5a10dd1ba0b5af7b5eb501f5946ff269f0780aec/xla/backends/profiler/gpu/cupti_tracer.cc#L1109-L1126), but given that the motivation for these annotations is to improve the nsys experience, it would make more sense to test via nsys. IIUC this would take a bit of work from the Google side to get set up, and we said this PR would not be held up waiting for that.

cheshire (approved):

dimitar-asenov:
@olupton We would really like to have at least a basic test in this PR.

E.g. is it possible to expose an Nvidia getter API for a stream name so that you can write a simple test like:

```
stream->set_name("xyz");
CHECK_EQ(Nvidia_get_stream_name_api(se::gpu::AsGpuStreamValue(stream)), "xyz");
```

olupton:
> @olupton We would really like to have at least a basic test in this PR.
> 
> E.g. is it possible to expose an Nvidia getter API for a stream name so that you can write a simple test like:
> 
> ```
> stream->set_name("xyz");
> CHECK_EQ(Nvidia_get_stream_name_api(se::gpu::AsGpuStreamValue(stream)), "xyz");
> ```

I believe there is no such API. In general, the NVTX APIs are one-way: https://github.com/NVIDIA/NVTX?tab=readme-ov-file#give-dont-take and if you do not run the executable under a profiling tool, the calls do nothing and the stream name (in your example) is not saved anywhere. Obviously we could track it in the XLA-side implementation of the methods in nvtx_utils.h, but I believe that would just add code without adding value.

As I noted above:

> > Is it feasible to test this? All changes requires a test or an explanation for why no test is necessary.
> > Is it possible to programmatically query the NVTX information, or can only NVIDIA tools like nsys do this?
> 
> We discussed this in the meeting last Tue; in the shorter term we will make sure downstream NVIDIA CI is exercising this functionality.
> 
> To answer the question, it is _possible_ to subscribe to receive information passed via NVTX. [XLA knows how to do this for a single call](https://github.com/openxla/xla/blob/5a10dd1ba0b5af7b5eb501f5946ff269f0780aec/xla/backends/profiler/gpu/cupti_tracer.cc#L1109-L1126), but given that the motivation for these annotations is to improve the nsys experience, it would make more sense to test via nsys. IIUC this would take a bit of work from the Google side to get set up, and we said this PR would not be held up waiting for that.

I believe the most meaningful way of doing targeted unit-like testing of nvtx_utils in XLA would be to use `nsys`, as that is the tool (in the sense of the word used in https://github.com/NVIDIA/NVTX?tab=readme-ov-file#give-dont-take: subscriber that receives NVTX calls) we are trying to improve the profiling experience in with PRs like this. I am imagining something along the lines of building a small test executable and executing it under nsys, with separate validation of the results. i.e. execute `nsys profile -o test_case_${case} --export ./test_exe ${case}` for each test case, on a machine with at least 1 GPU, and then read + validate the exported profile data (sqlite, json, ..).

When we discussed this just over a week ago, I believe the conclusion was that that would take some infrastructure work on the Google side and that it would not hold up this PR. I am happy to provide the substance of the test cases/validators if you agree with this architecture and the scaffolding is provided. In the meantime, we said we will make sure downstream NVIDIA CI is exercising this functionality.

PR #14111, opened 2024-06-24
Tuple Outputs in Convolution Algorithm Picker

philipphack:
Adds support for tuple-shaped outputs in the convolution algorithm picker.

This enables running TestConvAmaxF8 and TestConvReluAmaxF8 in cudnn_fused_conv_rewriter_test.cc.

### DIFF
--- a/xla/service/gpu/autotuning/conv_algorithm_picker_test.cc
+++ b/xla/service/gpu/autotuning/conv_algorithm_picker_test.cc
@@ -25,6 +25,7 @@ limitations under the License.
 #include "xla/service/gpu/autotuning/autotuner_util.h"
 #include "xla/service/gpu/stream_executor_util.h"
 #include "xla/service/gpu/transforms/conv_rewriter.h"
+#include "xla/service/gpu/transforms/cudnn_fused_conv_rewriter.h"
 #include "xla/service/pattern_matcher.h"
 #include "xla/service/pattern_matcher_gmock.h"
 #include "xla/service/platform_util.h"
@@ -36,6 +37,10 @@ limitations under the License.
 #include "tsl/platform/statusor.h"
 #include "tsl/platform/test.h"
 
+#if GOOGLE_CUDA
+#include "third_party/gpus/cuda/include/cuda.h"
+#endif
+
 namespace xla::gpu {
 namespace {
 
@@ -43,6 +48,16 @@ namespace m = ::xla::match;
 
 class GpuConvAlgorithmPickerTest : public HloTestBase {
  public:
+  se::CudaComputeCapability GetCudaComputeCapability() {
+    return backend()
+        .default_stream_executor()
+        ->GetDeviceDescription()
+        .cuda_compute_capability();
+  }
+  stream_executor::dnn::VersionInfo GetDnnVersion() {
+    return GetDnnVersionInfoOrDefault(backend().default_stream_executor());
+  }
+
   GpuConvAlgorithmPickerTest() { AutotunerUtil::ClearAutotuneResults(); }
 };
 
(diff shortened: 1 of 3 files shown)


xla/service/gpu/conv_algorithm_picker_test.cc
```diff
@@ -36,13 +37,27 @@ limitations under the License.
 #include "tsl/platform/statusor.h"
 #include "tsl/platform/test.h"
 
+#if GOOGLE_CUDA
+#include "third_party/gpus/cuda/include/cuda.h"
```
cheshire, line 41: Do we need this header?

xla/service/gpu/conv_algorithm_picker.cc
```diff
@@ -731,10 +741,11 @@ absl::StatusOr<AutotuneResult> GpuConvAlgorithmPicker::AutotuneOneConvRunner(
   }
 
   if (reference_result->has_value()) {
-    XLA_SCOPED_LOGGING_TIMER_LEVEL("BufferComparator::CompareEqual", 2);
```
cheshire, line 734: Should we remove the timer?

xla/service/gpu/BUILD
```diff
@@ -2025,6 +2025,7 @@ cc_library(
 xla_test(
     name = "conv_algorithm_picker_test",
     srcs = if_gpu_is_configured(["conv_algorithm_picker_test.cc"]),
+    local_defines = if_cuda_is_configured(["GOOGLE_CUDA=1"]),
```
cheshire, line 2028: Would be ideal not to add NVIDIA-specific headers if possible to that file.

cheshire (changes requested):

xla/service/gpu/conv_algorithm_picker_test.cc
philipphack: We need the header (and the change to the BUILD file above) if we want to test FP8 convolutions since the constructor of the CudnnFusedConvRewriter pass requires the CUDA version.

cheshire: Right I see, I was thinking can we use runtime CUDA version? Or we actually need to use compile-time one?

philipphack: Can you provide reference to the runtime version? Other tests such as cudnn_fused_conv_rewriter_test.cc also use CUDA_VERSION.

dimitar-asenov:
@cheshire @reedwm Please have a look at this PR.

golechwierowicz:
@cheshire @reedwm gentle ping

xla/service/gpu/conv_algorithm_picker_test.cc
```diff
@@ -125,5 +140,105 @@ ENTRY main {
...
+
+  const se::GpuComputeCapability& cc = GetCudaComputeCapability();
+  bool changed = false;
+  TF_ASSERT_OK_AND_ASSIGN(changed, RunHloPass(GpuConvRewriter(cc), m.get()));
+  TF_ASSERT_OK_AND_ASSIGN(
+      changed, RunHloPass(CudnnFusedConvRewriter(GetCudaComputeCapability(),
+                                                 GetDnnVersion(), CUDA_VERSION),
+                          m.get()));
+
+  changed = false;
```
cheshire, line 197: Should we test the value of change in between those?

xla/service/gpu/conv_algorithm_picker_test.cc
```diff
@@ -125,5 +140,105 @@ ENTRY main {
...
+  ASSERT_EQ(results.results_size(), 1);
+  auto& result = *results.mutable_results(0)->mutable_result();
+  int64_t old_scratch_bytes = result.scratch_bytes();
+  int64_t new_scratch_bytes = old_scratch_bytes + 1;
+  result.set_scratch_bytes(new_scratch_bytes);
+
+  AutotunerUtil::ClearAutotuneResults();
+  TF_ASSERT_OK(AutotunerUtil::LoadAutotuneResults(results));
+
+  // Now send the same module through GpuConvAlgorithmPicker again.  The conv
```
cheshire, line 216: This test looks like it could benefit from being broken up into multiple ones.

cheshire (changes requested):

xla/service/gpu/conv_algorithm_picker_test.cc
philipphack: Done.

xla/service/gpu/conv_algorithm_picker_test.cc
philipphack: Do you have an idea how to split this? Currently, the test verifies that the convolution algorithm picker makes changes and that it correctly identifies the convolution and its scratch space when loading the results of a previous run, which requires running the algorithm picker and serializing the results. This is essentially the FP8 version of the existing SetAlgorithm unit test.

cheshire (changes requested):
Could you expand the change description?

Something along the lines of:

- What is not working now
- What is it doing now
- How are you addressing the issue
- How will it start working

The high-level picture is still eluding me somewhat, even after looking at the test.

philipphack:
@cheshire TestConvAmaxF8 and TestConvReluAmaxF8 in cudnn_fused_conv_rewriter_test.cc return tuples. The algorithm picker currently creates a [BufferComparator](https://github.com/openxla/xla/blob/360465c96889d9b653877a317c7ad0fa6bddd97c/xla/service/gpu/autotuning/conv_algorithm_picker.cc#L737) for the output shape of the convolution even though that class doesn't support tuples. The current change to conv_algorithm_picker.cc resolves this by separately creating a BufferComparator object for [each of the elements](https://github.com/openxla/xla/blob/2a8cf8862c4c99f79ecdac20cb89b2ab919cb136/xla/service/gpu/conv_algorithm_picker.cc#L751) of the output tuple.

xla/service/gpu/conv_algorithm_picker_test.cc
```diff
@@ -125,5 +140,114 @@ ENTRY main {
...
+  ASSERT_TRUE(changed);
+  changed = false;
+  DebugOptions opts = DefaultDebugOptionsIgnoringFlags();
+
+  AutotuneConfig cfg{DeviceConfig{stream_exec, nullptr}, opts};
+  TF_ASSERT_OK_AND_ASSIGN(changed,
+                          RunHloPass(GpuConvAlgorithmPicker(cfg), m.get()));
+  ASSERT_TRUE(changed);
+
+  AutotuneResults results;
```
reedwm, line 208: I don't understand the motivation of this test after this point. You serialize the results, increasing the scratch bytes, load the results, rerun the pass, and check it has the increased number of scratch bytes. Perhaps this is useful, but what does this have to do with this PR, or FP8?

reedwm (changes requested):

reedwm (approved):

xla/service/gpu/autotuning/conv_algorithm_picker_test.cc
```diff
@@ -25,6 +25,7 @@ limitations under the License.
 #include "xla/service/gpu/autotuning/autotuner_util.h"
 #include "xla/service/gpu/stream_executor_util.h"
 #include "xla/service/gpu/transforms/conv_rewriter.h"
+#include "xla/service/gpu/transforms/cudnn_fused_conv_rewriter.h"
```
akuegel, line 28: this requires a corresponding dependency in the BUILD file

xla/service/gpu/autotuning/conv_algorithm_picker_test.cc
```diff
@@ -36,13 +37,27 @@ limitations under the License.
 #include "tsl/platform/statusor.h"
 #include "tsl/platform/test.h"
 
+#if GOOGLE_CUDA
+#include "third_party/gpus/cuda/include/cuda.h"
```
akuegel, line 41: This requires a corresponding dependency in the BUILD file guarded with if_cuda_is_configured

PR #14158, opened 2024-06-25
Add kExp op to algsimp IsNonNegative

apivovarov:
`kExp` result is always non-negative (similar to `kAbs`).
We can add `kExp` to `IsNonNegative()` util function in algsimp

Test:
```python
import jax.numpy as jnp                                                                                                                                                                                                                 
x = jnp.array([0.0, -0.0, jnp.inf, -jnp.inf, jnp.nan, 0.1, -0.1, 3, -3]) 

jnp.abs(x)
# Array([0. , 0. , inf, inf, nan, 0.1, 0.1, 3. , 3. ], dtype=float32)

jnp.exp(x)
# Array([ 1. , 1. , inf, 0. , nan, 1.105171, 0.9048374, 20.085537, 0.04978707], dtype=float32)
```
Adrian, what you think? @akuegel

### DIFF
--- a/xla/service/algebraic_simplifier.cc
+++ b/xla/service/algebraic_simplifier.cc
@@ -203,41 +203,6 @@ static bool IsScalarConstantInf(const HloInstruction* hlo) {
                           LiteralUtil::MaxValue(hlo->shape().element_type()));
 }
 
-bool IsNonNegative(const HloInstruction* hlo,
-                   const AlgebraicSimplifierOptions& options) {
-  // Utility only handles real types.
-  if (IsAnyOperandComplex(hlo)) {
-    return false;
-  }
-  switch (hlo->opcode()) {
-    case HloOpcode::kMultiply: {
-      return hlo->operand(0) == hlo->operand(1);
-    }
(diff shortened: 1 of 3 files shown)


akuegel (commented):
Yes, good idea, in fact I thought it is already there when I wrote the test for https://github.com/openxla/xla/commit/31e4b5df8ebe8edd2d32fb8750571243f17294aa

But can you please also write a test for this?

apivovarov:
> Yes, good idea, in fact I thought it is already there when I wrote the test for [31e4b5d](https://github.com/openxla/xla/commit/31e4b5df8ebe8edd2d32fb8750571243f17294aa)
> 
> But can you please also write a test for this?

Apparently, the `isNonNegative()` function did not have direct unit tests.

I've added several tests for the `isNonNegative` function. 
To make it accessible in `algebraic_simplifier_test.cc`, I moved it into the `AlgebraicSimplifierVisitor` class and made it static.

akuegel (approved):
Thank you

NaiyerRizz (approved):

PR #14162, opened 2024-06-25
Add kMinimum case to IsNonNegative

apivovarov:
If both operands of kMinimum operation is non-negative then the result also should be non-negative.

Adrian, Can you review the PR? @akuegel


akuegel (commented):
Can you please add a test for this?

akuegel (approved):

apivovarov:
> Can you please add a test for this?

Added positive and negative test cases

NaiyerRizz (commented):
Looks good

PR #14177, opened 2024-06-26
Fix missing entries of device description proto

ghost:
Add registers_per_core_limit and registers_per_block_limit which were missed before into device description proto.

### DIFF
--- a/xla/stream_executor/device_description.proto
+++ b/xla/stream_executor/device_description.proto
@@ -44,9 +44,11 @@ message GpuDeviceInfoProto {
   float clock_rate_ghz = 13;
   int64 device_memory_size = 14;
   int32 shared_memory_per_block_optin = 15;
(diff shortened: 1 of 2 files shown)


xla/stream_executor/device_description.proto
```diff
@@ -44,9 +44,11 @@ message GpuDeviceInfoProto {
   float clock_rate_ghz = 13;
   int64 device_memory_size = 14;
   int32 shared_memory_per_block_optin = 15;
+  int64 registers_per_core_limit = 16;
```
reedwm, line 47: Don't change the proto numbers. See https://protobuf.dev/programming-guides/dos-donts/#reuse-number. Instead assign the new fields numbers 18 and 19.

reedwm (changes requested):

reedwm (approved):

PR #14222, opened 2024-06-26
1 create function to convert hloinstruction type to hlomodule type

dinodeep:
Implemented function that goes from `HloInstruction` type to `HloModule` type with simple iteration and some simple logging functionality for the auto-parallelism code

### DIFF
--- a/xla/service/BUILD
+++ b/xla/service/BUILD
@@ -677,6 +677,19 @@ xla_cc_test(
 )
 
 cc_library(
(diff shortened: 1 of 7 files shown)


cheshire:
CC @ezhulenev @penpornk 

Thanks! Could you provide more context for the CL? It's probably not mergeable as-is.

cheshire:
Closing as something  a lot more narrow/targeted would have to be opened.

Issue #14172, opened 2024-06-26
Build of target //xla/tools:run_hlo_module failed - no such target '@com_google_protobuf//:any_cc_proto'


apivovarov:
Build of target `//xla/tools:run_hlo_module` failed

`no such target '@com_google_protobuf//:any_cc_proto'`

Log:
```bash
INFO: Build options --compilation_mode, --copt, --cxxopt, and 1 more have changed, discarding analysis cache.
ERROR: /local/home/pivovaa/workspace/xla/xla/backends/interpreter/BUILD:67:11: no such target '@com_google_protobuf//:any_cc_proto': target 'any_cc_proto' not declared in package '' defined by /local/home/pivovaa/.cache/bazel/_bazel_pivovaa/0e997ddc2bf61ebd028b3d99dca42844/external/com_google_protobuf/BUILD.bazel (did you mean 'any_proto'? Tip: use `query "@com_google_protobuf//:*"` to see all the targets in that package) and referenced by '//xla/backends/interpreter:platform_id'
ERROR: Analysis of target '//xla/tools:run_hlo_module' failed; build aborted: 
INFO: Elapsed time: 0.756s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (17 packages loaded, 1603 targets configured)
```

Suspecting commit https://github.com/openxla/xla/commit/f50749ace34227bc038ffb1b6d33398b547ac435
David, can you have a look? @ddunl

ddunl:
Sorry about that! I'll make a fix tomorrow morning.

apivovarov:
Hi David, I just tried to build it again. I'm using the latest XLA
```bash
commit c53e9f4e48f11d0da3469c85a0692db7ebd2a5d3 (HEAD -> main, origin/main, origin/HEAD)
Author: Benjamin Kramer <kramerb@google.com>
Date:   Wed Jun 26 08:54:46 2024 -0700
```
Error: `no such target '@com_google_protobuf//:any_cc_proto': target 'any_cc_proto' not declared in`

Output:
```bash
bazel build  //xla/tools:run_hlo_module 

INFO: Build options --compilation_mode, --copt, --cxxopt, and 1 more have changed, discarding analysis cache.
[...]

PR #14224, opened 2024-06-27
Fix build error for LogicalIdThunk::buffer_uses()

apivovarov:
I tried to build `run_hlo_module` using GCC-13
```bash
bazel build //xla/tools:run_hlo_module
```
and faced build error
```bash
xla/service/cpu/runtime/logical_id_thunk.cc:113:18: error: no declaration matches 'xla::cpu::BufferUses<type> xla::cpu::LogicalIdThunk<type>::buffer_uses() const'
  113 | BufferUses<type> LogicalIdThunk<type>::buffer_uses() const {
      |                  ^~~~~~~~~~~~~~~~~~~~
```

This PR fixes it by using `Thunk::BufferUses` for the output type of `LogicalIdThunk<type>::buffer_uses()` impl.

`BufferUses` alias is declared in `class Thunk`

Related issue: https://github.com/openxla/xla/issues/14172
David, Pawel,  Can you review the fix? @ddunl @pparuzel

### DIFF
--- a/xla/service/cpu/runtime/logical_id_thunk.cc
+++ b/xla/service/cpu/runtime/logical_id_thunk.cc
@@ -109,11 +109,7 @@ LogicalIdThunk<logical_id_kind>::Execute(const ExecuteParams& params) {
 }
 
 template <LogicalIdKind logical_id_kind>
-using BufferUses = typename LogicalIdThunk<logical_id_kind>::BufferUses;
-
-template <LogicalIdKind logical_id_kind>
-BufferUses<logical_id_kind> LogicalIdThunk<logical_id_kind>::buffer_uses()
-    const {
+Thunk::BufferUses LogicalIdThunk<logical_id_kind>::buffer_uses() const {
   return {BufferUse::Write(logical_id_buffer_)};
 }
 


pparuzel (approved):
LGTM

Indeed, BufferUses should have been taken from the Thunk class itself. My bad and thanks for the quick fix!

reedwm:
Looks like e7c9cae60b0dddac7581375a0afc8f717849ea03 attempts to fix the same issue. @apivovarov, does the issue still occur?

apivovarov:
> Looks like [e7c9cae](https://github.com/openxla/xla/commit/e7c9cae60b0dddac7581375a0afc8f717849ea03) attempts to fix the same issue. @apivovarov, does the issue still occur?

yes, `//xla/tools:run_hlo_module` build still fails with error:
```bash
xla/service/cpu/runtime/logical_id_thunk.cc:115:29: error: no declaration matches 'xla::cpu::internal::BufferUses<logical_id_kind> xla::cpu::internal::LogicalIdThunk<logical_id_kind>::buffer_uses() const'
  115 | BufferUses<logical_id_kind> LogicalIdThunk<logical_id_kind>::buffer_uses()
```

I rebased my PR to apply my fixes on top of e7c9cae

reedwm (approved):
Weird, I would have thought the original code should work, given that `LogicalIdThunk<logical_id_kind>::BufferUses` is equivalent to `Thunk:BufferUses`. Maybe it's a bug in gcc, or maybe it's actually correctly giving an error due to some C++ rule I don't know about. In any case, good to fix it, and the new code is clearer as well.

PR #14255, opened 2024-06-27
Fix build issue in xla/hlo/utils/hlo_sharding_util_test.cc

apivovarov:
This PR fixes the following build error in `xla/hlo/utils/hlo_sharding_util_test.cc` (used GCC-13):
```bash
xla/hlo/utils/hlo_sharding_util_test.cc:538:41: error: call of overloaded 'TileAssignment(<brace-enclosed initializer list>)' is ambiguous
  538 |   TileAssignment tile_assignment({16, 4});
```
Reed, can you look at this build fix too? @reedwm

### DIFF
--- a/xla/hlo/ir/tile_assignment.h
+++ b/xla/hlo/ir/tile_assignment.h
@@ -173,6 +173,8 @@ class TileAssignment {
       : TileAssignment(std::make_shared<const Array<int64_t>>(
             std::initializer_list<int64_t>{1}, device_id)) {}
   explicit TileAssignment(IotaTileAssignment iota) : iota_(std::move(iota)) {}
+  explicit TileAssignment(std::initializer_list<int64_t> dims)
+      : iota_(IotaTileAssignment::Create(dims)) {}
   explicit TileAssignment(absl::Span<const int64_t> dims)
       : iota_(IotaTileAssignment::Create(dims)) {}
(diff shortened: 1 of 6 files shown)


apivovarov:
@akuegel Can you look at this build fix?

reedwm (changes requested):
Interesting, I can reproduce this with https://godbolt.org/z/MonhYaYzb. clang compiles this, but gcc complains. I wonder which compiler is right.

Anyway, I prefer adding a `TileAssignment` constructor taking a `std::initializer_list<int64_t>`:

```c++
explicit TileAssignment(std::initializer_list<int64_t> dims) : 
    TileAssignment(absl::Span<const int64_t>(dims)) {}
```

I think this will allow constructing TileAssignment like `TileAssignment({4, 1})`, which might be accidentally done in the future since we don't test with gcc.

reedwm:
Also the build failure is spurious, ignore it.

apivovarov:
- Added `TileAssignment(initializer_list)` constructor
- Replaced `TileAssignment((absl::Span<const int64_t>){a, b})` with `TileAssignment({a, b})`.

apivovarov:
Looks like this PR got lost. @reedwm

akuegel (approved):

PR #14264, opened 2024-06-28
[XLA:GPU] Add  debug info for command buffer trace cache

shawnwang18:


cheshire:
Looks good to me, but @ezhulenev should probably take a look as well.

xla/service/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -409,6 +410,9 @@ absl::StatusOr<se::CommandBuffer*> TracedCommandBuffer::GetOrTraceCommandBuffer(
     // pointer to cached command buffer.
     if (ABSL_PREDICT_TRUE(absl::c_equal(entries_[i].recorded_allocs, allocs) &&
                           entries_[i].command_buffer)) {
+      VLOG(6) << "Command buffer trace cache hit for command "
+              << (trace_cmd_.has_value() ? trace_cmd_.value()->ToString()
```
ezhulenev, line 414: You can do simply `this->ToString()` because it's a virtual API and all traced commands inherit from TracedCommandBuffer

shawnwang18: No, traced command does not inherit TracedCommandBuffer.

The inherit is like: 
class TracedCommandBufferCmd : public CommandBufferCmd
class GemmCmd : public TracedCommandBufferCmd

ezhulenev (approved):

PR #14274, opened 2024-06-28
[XLA:GPU] Disable cublasLT tracing for cuda version < 12030

shawnwang18:


ezhulenev (approved):

reedwm:
All changes need a test, or require us to provide an internal explation for why a test is unnecessary when merging.

Does this fail already on CUDA versions less than 12.3? If so, tests will already fail on such CUDA versions, so no test is necessary.

ezhulenev (approved):

PR #14293, opened 2024-06-28
Changes needed for future hardware compatibility.

dimvar:


cheshire:
Could we send/link a patch to upstream Triton?

NaiyerRizz:
Hi @dimvar 
Can you please look into the above comment from @cheshire

thomasjoerg:
@dimvar friendly ping!

dimvar:
Hi, sorry for the late reply, just got back from vacation. We've modified these commits since I sent this PR. Let's close it for now, and I may open another one with the new commits later.

PR #14295, opened 2024-06-28
[ROCm] fixed gcc build

i-chaochen:
gcc compilation will break  due to this change https://github.com/openxla/xla/commit/4d7ec1e49fb89185ecf0c46493f9fb6fcc6519cd

@xla-rotation

### DIFF
--- a/xla/literal.h
+++ b/xla/literal.h
@@ -360,12 +360,16 @@ class LiteralBase {
   // function which would lead to inconsistencies. Use Hash instead.
   template <typename H>
   friend H AbslHashValue(H state, const LiteralBase& value) {
+#if (!defined(__GNUC__)) // clang
     static_assert(false,
                   "Do not use Literal directly as a hash key, because it has "
                   "multiple definitions of equality - layout sensitive or "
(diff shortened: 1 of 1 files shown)


xla/literal.h
```diff
@@ -360,12 +360,16 @@ class LiteralBase {
   // function which would lead to inconsistencies. Use Hash instead.
   template <typename H>
   friend H AbslHashValue(H state, const LiteralBase& value) {
+#if (!defined(__GNUC__)) // clang
     static_assert(false,
```
ezhulenev, line 364: can you instead of disabling this check for gcc change it to `static_assert(sizeof(H) == 0, ...` (see https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2593r0.html)

i-chaochen: thanks for this link! looks `static_assert(always_false<T>::value);` is a more acceptable way... :)

i-chaochen:
thanks @ezhulenev for the suggestion! PTAL

i-chaochen:
This can be closed due to https://github.com/openxla/xla/pull/14296

PR #14342, opened 2024-07-01
Fix //xla/service/gpu:ir_emitter_triton_mem_utils_test in OSS

sergey-kozub:
Currently the test fails with:

xla/service/gpu/ir_emitter_triton_mem_utils_test.cc:48:10: fatal error: 'third_party/triton/include/triton/Dialect/Triton/IR/Dialect.h' file not found
   48 | #include "third_party/triton/include/triton/Dialect/Triton/IR/Dialect.h"

Also, using `cc_test` instead of `xla_cc_test` results in linker errors

### DIFF
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -705,7 +705,7 @@ cc_library(
     ],
 )
 
-cc_test(
+xla_cc_test(
     name = "ir_emitter_triton_mem_utils_test",
     srcs = if_cuda_is_configured(["ir_emitter_triton_mem_utils_test.cc"]),
     deps = [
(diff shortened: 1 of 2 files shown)


anlunx (approved):

dimitar-asenov:
I wonder what happened here, because that aspect of the test hasn't been changed recently. @sergey-kozub did this only recently start failing?

@ddunl did something change in how we translate paths between OSS and google3?

sergey-kozub:
> did this only recently start failing?
I believe it was always failing in OSS, we just don't run all the tests in OSS (for whatever reason)

dimitar-asenov (approved):

PR #14354, opened 2024-07-02
Fix //xla/service/gpu:autotuner_util_test in OSS

sergey-kozub:
Currently the test fails with:

external/com_google_googletest/googletest/include/gtest/gtest-message.h:103:3: note: candidate constructor not viable: no known conversion from 'const AutotuneResult' to 'const Message &' for 1st argument        
  103 |   Message(const Message& msg) : ss_(new ::std::stringstream) {  // NOLINT


tdanyluk (approved):
thanks.

PR #14359, opened 2024-07-02
Add Blackwell-related methods to CudaComputeCapability class

sergey-kozub:
Also remove unused methods `GetMaxResidentBlocksPerSM` and `GetMaxResidentWarpsPerSM`

### DIFF
--- a/xla/stream_executor/device_description_test.cc
+++ b/xla/stream_executor/device_description_test.cc
@@ -1,0 +1,37 @@
+/* Copyright 2024 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+#include "xla/stream_executor/device_description.h"
+
+#include "tsl/platform/test.h"
+
+namespace stream_executor {
+namespace {
+
+TEST(CudaComputeCapability, GenerationNumericTest) {
+    EXPECT_TRUE(CudaComputeCapability(7, 5).IsAtLeastVolta());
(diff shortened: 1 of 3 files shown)


akuegel (commented):
I am not 100% sure whether this would count as a NFC. If possible, could you potentially add a test that calls the new IsAtLeastBlackwell method?

sergey-kozub:
> could you potentially add a test that calls the new IsAtLeastBlackwell method?

I don't see any existing tests for `IsAtLeast(Volta|Ampere|Hopper)` methods.
Should I add those as well? Also, which test file should I put this to?

On another note, should we completely remove `GetMaxResidentBlocksPerSM` and `GetMaxResidentWarpsPerSM` methods, that are not used anywhere? They just add maintenance burden.

akuegel:
> > could you potentially add a test that calls the new IsAtLeastBlackwell method?
> 
> I don't see any existing tests for `IsAtLeast(Volta|Ampere|Hopper)` methods. Should I add those as well? Also, which test file should I put this to?

Yes, historically we were not so strict about tests, but now there is a presubmit check. Please put the test file in the same directory. Adding calls for the other IsAtLeast methods would be very nice, but I would also approve without, as they are not new.

> On another note, should we completely remove `GetMaxResidentBlocksPerSM` and `GetMaxResidentWarpsPerSM` methods, that are not used anywhere? They just add maintenance burden.

Completely agree, let's remove them. I checked that also internally they are marked as dead code.

sergey-kozub:
PTAL

akuegel (approved):
Thank you very much!

xla/stream_executor/device_description_test.cc
```diff
@@ -0,0 +1,37 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+#include "xla/stream_executor/device_description.h"
+
+#include "tsl/platform/test.h"
```
akuegel, line 17: It seems the build target is missing a dependency for this header include

sergey-kozub: Fixed.

akuegel (approved):

PR #14364, opened 2024-07-02
[ROCm] Add script to run multi gpu tests

hsharsha:
This is a rocm specific script housed under `build_tools/rocm`  It runs following distributed tests which require more >= 4 gpus and these tests are skipped currently in the CI due to tag selection. These tests are tagged either as manual or with oss
```
//xla/tests:collective_ops_e2e_test_gpu_amd_any 
//xla/tests:collective_ops_test_gpu_amd_any 
//xla/tests:replicated_io_feed_test_gpu_amd_any 
//xla/tools/multihost_hlo_runner:functional_hlo_runner_test_gpu_amd_any 
//xla/pjrt/distributed:topology_util_test 
//xla/pjrt/distributed:client_server_test
 ```
 Also these tests do not use `--run_under=//tools/ci_build/gpu_build:parallel_gpu_execute` with bazel which locks down individual gpus thus making multi gpu tests impossible to run
 
 Eventually we would like to enable these tests in a separate pipeline to get better test coverage

### DIFF
--- a/build_tools/rocm/run_xla.sh
+++ b/build_tools/rocm/run_xla.sh
@@ -15,6 +15,9 @@
 #
 # ==============================================================================
 
+# This script runs XLA unit tests on ROCm platform by selecting tests that are
+# tagged with requires-gpu-amd
+
 set -e
 set -x
 
--- a/build_tools/rocm/run_xla_multi_gpu.sh
+++ b/build_tools/rocm/run_xla_multi_gpu.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright 2024 The TensorFlow Authors. All Rights Reserved.
+#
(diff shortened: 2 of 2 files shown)


dimitar-asenov (changes requested):
Can you please add a comment on top of the script explaining in detail how to use it? E.g. what are the requirements for running it, what are the expected inputs, what tests are run, etc.

dimitar-asenov:
@hsharsha Thanks for adding the info in the description of the PR. I actually meant to have this as a inline comment on top of the script file, so that folks looking at the code will know what it is and how to use it, without looking at the Blame history. Could you please add it there?

dimitar-asenov (approved):

dimitar-asenov:
Thanks.

PR #14368, opened 2024-07-02
[XLA:CPU] Fix the detection of FP16 extension to AVX512

akhilgoe:
Currently, the check for AVX512_FP16 is not correct. This PR fixes it to match xbyak code.

### DIFF
--- a/third_party/tsl/tsl/platform/cpu_info.cc
+++ b/third_party/tsl/tsl/platform/cpu_info.cc
@@ -209,12 +209,14 @@ class CPUIDInfo {
     cpuid->have_rdseed_ = (ebx >> 18) & 0x1;
     cpuid->have_smap_ = (ebx >> 20) & 0x1;
 
+    const bool have_avx512bw = have_avx512 && ((ebx >> 30) & 0x1);
+
     cpuid->have_avx512f_ = have_avx512 && ((ebx >> 16) & 0x1);
     cpuid->have_avx512cd_ = have_avx512 && ((ebx >> 28) & 0x1);
     cpuid->have_avx512er_ = have_avx512 && ((ebx >> 27) & 0x1);
     cpuid->have_avx512pf_ = have_avx512 && ((ebx >> 26) & 0x1);
     cpuid->have_avx512vl_ = have_avx512 && ((ebx >> 31) & 0x1);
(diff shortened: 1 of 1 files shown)


dimitar-asenov (changes requested):
Can you please add a test that verifies the fix?

akhilgoe:
Thanks @dimitar-asenov for the review! Is there an existing test file that verifies CPU features? I am a little unsure about how to add a test for this change.

dimitar-asenov:
One option to add a test is to split the `Initialize` function into two independent parts:
- One that calls `GETCPUID` to get the underlying bits.
- One that interprets the result from the above.  I see there are four different calls to `GETCPUID` so there should be 4 different interpretation functions.

The `Initialize` function would call both of these new functions and work exactly as it does currently. A new test could simply call the separate interpretation function with custom provided `uint32` values for `eax`, `ebx`, `ecx` , and `edx` and verify that they are interpreted correctly. It's enough to add a test that only verifies the current fix and will be great if the tested `eax`, `ebx`, `ecx` , and `edx` inputs come from some real CPUs.

I know this may sound like overkill, but it will help make the code better. If we had tests like this previously, we may have avoided the error altogether.

What do you think?

akhilgoe:
Yes, that makes sense! I'll create a draft commit and will request for a review.

akhilgoe:
Hi @dimitar-asenov, I apologize for the delay. I got caught up with other tasks over the past few weeks. Just wanted to let you know that this is still on my radar, and I'll be pushing a commit soon that refactors the existing code and adds tests. Thank you for your patience!

sgerrard:
@akhilgoe checking in on the status of the PR?

dimitar-asenov:
@akhilgoe, could you please provide an update on this PR?

akhilgoe:
@sgerrard @dimitar-asenov very sorry about the delay! I have made the changes locally and am testing them out. As soon as I get the test report and all looks good, I'll push the changes. Thanks again for your patience!

akhilgoe:
Update: All looks good in my local testing, but to eliminate any concerns that the refactoring might have introduced errors, my team believes it is important to test this on our legacy hardware as well. Converting this PR to draft until I get approval to update it.

PR #14419, opened 2024-07-03
[GPU] Improve compiled kernel caching with new SplitModule mode.

sergachev:
The new RoundRobin mode distributes functions to modules more evenly, therefore there are more single-function cacheable modules. Without this the example in the added test would have only 2 kernels cached.

This needs https://github.com/llvm/llvm-project/commit/c02e8f762a410e55581866c43636efcd6504c1bd to get integrated before it can be submitted.


dimitar-asenov (approved):

sergachev:
rebased to a version with the required LLVM commit integrated

PR #14458, opened 2024-07-05
[XLA:GPU] use empty cuda-graph node as command buffer barrier node for CUDA_VERSION >= 12040

shawnwang18:
tested that cuda_12.4 have fixed the cuda graph issue that empty node can not be in conditioanl operator's sub-graph.

### DIFF
--- a/xla/service/gpu/runtime/command_buffer_cmd_test.cc
+++ b/xla/service/gpu/runtime/command_buffer_cmd_test.cc
@@ -231,6 +231,8 @@ TEST(CommandBufferCmdTest, MemcpyCmd) {
 }
 
 TEST(CommandBufferCmdTest, BarrierCmd) {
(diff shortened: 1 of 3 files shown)


cheshire (changes requested):
Any way we could test this?

shawnwang18:
> Any way we could test this?

I think running existing tests with cuda 12.4 should test it. For example: test https://github.com/openxla/xla/blob/d2d668321c5507a9563fc3a77a0a7a62de4e1a93/xla/service/gpu/runtime/command_buffer_thunk_test.cc#L849 should cover this case. 

google's CI currently uses 12.3?  can we also add new cuda release?

ezhulenev (approved):

NaiyerRizz (approved):

ezhulenev (approved):

PR #7179, opened 2023-11-21
[XLA:GPU] add back pointer in while body computation to trace back to while call

Cjkkkk:
add back pointer in while body computation to trace back to while call for data dependency analysis and pattern matching.

PR #14476, opened 2024-07-05
Fix while loop parsing in in protobuf format

shraiysh:
While loop computations were not being linked to the while operation. This patch fixes that and connects them.

### DIFF
--- a/xla/service/hlo_parser_test.cc
+++ b/xla/service/hlo_parser_test.cc
@@ -2541,6 +2541,14 @@ class HloParameterizedParserTest
           original,
           module->ToString(HloPrintOptions().set_print_large_constants(true)));
     }
+    for (HloComputation* computation : module->computations()) {
+      for (HloInstruction* instr : computation->instructions()) {
+        if (instr->opcode() == HloOpcode::kWhile) {
+          EXPECT_EQ(instr->while_body()->WhileCallInstruction(), instr);
+          EXPECT_TRUE(instr->while_body()->IsWhileBodyComputation());
+        }
+      }
+    }
   }
 };
 
--- a/xla/hlo/ir/hlo_instruction.cc
+++ b/xla/hlo/ir/hlo_instruction.cc
@@ -1184,6 +1184,8 @@ absl::StatusOr<std::unique_ptr<HloInstruction>> HloInstruction::CreateFromProto(
         TF_RET_CHECK(proto.called_computation_ids_size() == 2)
             << "While should have 2 called computation but has "
             << proto.called_computation_ids_size();
+        computation_map.at(proto.called_computation_ids(0))
+            ->SetWhileCallInstruction(instruction.get());
       }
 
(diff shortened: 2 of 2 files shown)


frgossen (changes requested):
Can you add a test case?

shraiysh:
Do you know where testcases for this would be added?

frgossen:
> Do you know where testcases for this would be added?

hlo_parser_test.cc

shraiysh:
@frgossen modified tests to check for this too.

frgossen (changes requested):
I looked a bit more and think the parser should actually not populate this because it is ambiguous.

Example: For a computation that is used as a body for two distinct while ops, there is no way to populate this field correctly. With this change, you will just override it and keep the last one which is at least misleading. This field can only be populated correctly after flattening the call graph.

This field is populated in `xla/service/flatten_call_graph.cc` together with other caller/callee pointers.

xla/service/hlo_parser_test.cc
```diff
@@ -2541,6 +2541,14 @@ class HloParameterizedParserTest
           original,
           module->ToString(HloPrintOptions().set_print_large_constants(true)));
     }
+    for (HloComputation* computation : module->computations()) {
```
frgossen, line 2544: I don't think this is the right place to add an assertion. 
Can you add a test below to make sure that the field is null after parsing?

frgossen (changes requested):

shraiysh:
I understand the concern, however, this is also done while [parsing HLO as text](https://github.com/openxla/xla/blob/15a888968d354c92b34d0d04cd962e8883ab4275/xla/service/hlo_parser.cc#L2155). The CreateWhileOp call sets this. I think we should fix that too in this case. Another thing is that some execution (for example, [address computation fusion here](https://github.com/openxla/xla/blob/15a888968d354c92b34d0d04cd962e8883ab4275/xla/service/gpu/fusions/custom.cc#L179) depend on this being already set up. Does this mean that we should run call graph flattening everytime we parse HLO even if it is already optimized?

frgossen:
@akuegel , you approved https://github.com/openxla/xla/pull/7179. Do you know what is the right thing to do here? 
- Populate the field and be aware that it could be ambiguous 
- Not populate it

akuegel:
> @akuegel , you approved #7179. Do you know what is the right thing to do here?
> 
> * Populate the field and be aware that it could be ambiguous
> * Not populate it

Yes, I approved it, and also suggested to use a back pointer. I was not aware of the difficulties with adding back pointers. We have an internal bug open for this. Not only while is affected, other ops with back pointers are potentially affected as well. Just for fusions we are on the safe side, because it is an op internal to XLA, so we control that there is always a 1:1 relationship between fusion instruction and fusion computation.

frgossen (approved):
Thanks@

PR #14477, opened 2024-07-06
Allow to build multihost_hlo_runner for CPU backend

apivovarov:
Currently, Bazel skips building `multihost_hlo_runner:hlo_runner_main` on CPU because the target is tagged with `tf_gpu_tests_tags`.

This tag should only be applied to the dedicated `hlo_runner_main_gpu` target.

The regular `hlo_runner_main` works fine on CPU (when `--xla_force_host_platform_device_count=N` is used)

@bchetioui @ddunl @beckerhe Can you have a look?


beckerhe (approved):
LGTM. Thank you for making this work!

PR #14478, opened 2024-07-06
IsNonNegative(constant(-0.0)) should be false

apivovarov:
Currently `AlgebraicSimplifierVisitor::IsNonNegative` returns true for `Constant(-0.0)`.

Why it is a problem.

Particular optimization patterns should only be applied when IsNonNegative is true.
e.g.
```
y = sqrt(x) * sqrt(x)
can be simplified to 
y = x , when x >=0
```
```
if x is -0.0 the results are different:
sqrt(-0.0) = -0.0 =>
y = sqrt(-0.0) * sqrt(-0.0) = 0.0

After simplification - different result
y = x = -0.0

if later we have smth like 1/y - the results will be completely different +inf and -inf
```

Another example
```
y = power(x, 0.5) can be simplified to y = sqrt(x)
but
power(-0.0, 0.5) = 0.0
sqrt(-0.0) = -0.0
```
Results are different too.

in C++ `0.0` is equal to `-0.0`.
Simple check for non-negativity `x >= 0.0` returns True for `-0.0`

To distinguish between `-0.0` and `0.0` we can use `std::signbit`.

Adrian, what you think? @akuegel

### DIFF
--- a/xla/service/algebraic_simplifier.cc
+++ b/xla/service/algebraic_simplifier.cc
@@ -534,7 +534,8 @@ bool AlgebraicSimplifierVisitor::IsNonNegative(
     }
     case HloOpcode::kConstant: {
       if (std::optional<double> value = GetConstantValue(hlo)) {
-        return *value >= 0.0;
+        // return false for -0.0, -Inf, NaNs and negative values
+        return !std::signbit(*value) && !std::isnan(*value);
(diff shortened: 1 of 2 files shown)


akuegel (approved):
Thank you, that change makes sense.

PR #13003, opened 2024-05-23
[ROCm] Switch on Triton feature for ROCm.

zoranjovanovic-ns:

PR #14605, opened 2024-07-08
[ROCm] Switch on Triton feature for ROCm.

zoranjovanovic-ns:
Last in series of commits to switch on Triton in XLA for ROCm.

This is new version of:
https://github.com/openxla/xla/pull/13003

Changes in third_party/triton/temporary/amd_pr7.patch are already merged on:
https://github.com/triton-lang/triton/pull/4238


dimitar-asenov:
@gflegar I see this PR has some Triton changes. How do we usually merge such PRs? Zoran says that these are already upstreamed, so does that mean we need to first integrate that Triton version and then merge this PR?

xla/service/gpu/triton_support.cc
```diff
@@ -628,6 +628,13 @@ absl::Status EnsureTritonSupportsComputeCapability(
     const se::GpuComputeCapability& gpu_compute_capability) {
   auto cuda_compute_capability =
       std::get_if<se::CudaComputeCapability>(&gpu_compute_capability);
+  auto rocm_compute_capability =
```
dimitar-asenov, line 631: This file is in flux and for the time being the support is expressed in two places:

- Here, where you modified the code to add support. This is the support for everything except MatMul.
- In the `legacy_triton` namespace above, where support for MatMul is defined. Consider updating functions in that namespace as well.

dimitar-asenov: Also, changes to this file normally mean that you have to modify the `triton_support_test.cc` file as well. Please add RoCM testing to that file.

dimitar-asenov (changes requested):

gflegar:
> @gflegar I see this PR has some Triton changes. How do we usually merge such PRs? Zoran says that these are already upstreamed, so does that mean we need to first integrate that Triton version and then merge this PR?

Yes, @vwbaker update Triton yesterday, so the upstream change should be included. We should rebase, remove the Triton patch and see if that works.

xla/service/gpu/triton_support.cc
zoranjovanovic-ns: There are some ROCM support in functions already, e.g.:
[CanTritonHandleGEMM](https://github.com/openxla/xla/blame/c2d607e21ab93d3bd20cfd9e781bcec79c9b925f/xla/service/gpu/triton_support.cc#L241C17-L241C36)
https://github.com/openxla/xla/blame/c2d607e21ab93d3bd20cfd9e781bcec79c9b925f/xla/service/gpu/triton_support.cc#L450C28-L450C28

Are you referring to this or some additional support should be implemented?


Also, there is some rocm support in triton_support_test.cc:
https://github.com/openxla/xla/blob/c2d607e21ab93d3bd20cfd9e781bcec79c9b925f/xla/service/gpu/triton_support_test.cc#L136

dimitar-asenov: Those are the ones I was referring to. If you think the existing coverage is enough, that's OK with me.

dimitar-asenov (approved):

dimitar-asenov:
@gflegar The PR has been updated and doesn't have any Triton patches any more. Can you please review it now? Thanks!

xla/service/gpu/triton_support.cc
zoranjovanovic-ns: Thank you for clarification.

gflegar (approved):
LGTM, thanks for polishing this up!

xla/service/gpu/fusions/triton.cc
```diff
@@ -114,6 +114,9 @@ absl::StatusOr<FusionEmissionResult> TritonFusion::Emit(
         triton_config.set_block_k(64);
         triton_config.set_block_n(64);
         triton_config.set_split_k(1);
+        triton_config.set_num_stages(1);
```
bchetioui, line 117: Is this necessary? I don't think `num_stages`, `num_warps`, and `num_ctas` are read from anywhere else than `block_level_parameters`---are they?

xla/service/gpu/ir_emitter_triton_rocm.cc
```diff
@@ -69,47 +72,50 @@ absl::Status CreateTritonPipeline(
...
   // TODO ROCm Check if we want to compare MI100 and greater
+  pm.addPass(mlir::createTritonAMDGPUOptimizeEpiloguePass());
   pm.addPass(mt::gpu::createTritonGPUOptimizeDotOperands({true}));
-  pm.addPass(mlir::createCSEPass());
-  pm.addPass(
-      mt::gpu::createTritonGPUPipeline({block_level_parameters.num_stages}));
-  pm.addPass(mt::gpu::createTritonGPUPrefetch());
-
-  // TODO ROCm Check if we want to compare MI100 and greater
+  if(block_level_parameters.num_stages == 0 &&
```
bchetioui, line 88: There is a missing space between `if` and the opening parenthesis. (Also below.)

bchetioui, line 88: What's the semantic meaning of `0` stages here and below?

This doesn't seem like it should be a valid value.

xla/service/gpu/ir_emitter_triton_test.cc
```diff
@@ -3279,6 +3282,10 @@ TEST_F(TritonGemmLevel2Test, SplitLHSInputOutputIsFused) {
   if (!SupportsBF16(GpuComputeComp())) {
     GTEST_SKIP() << "BF16 not supported.";
   }
+  if (std::holds_alternative<se::RocmComputeCapability>(GpuComputeComp())) {
+    GTEST_SKIP() << "Skipped until corresponding issue on ROCm is not fixed.";
```
bchetioui, line 3286: is not fixed -> is fixed

xla/service/gpu/triton_support.cc
```diff
@@ -628,6 +628,13 @@ absl::Status EnsureTritonSupportsComputeCapability(
...
+      std::get_if<se::RocmComputeCapability>(&gpu_compute_capability);
+  if (!cuda_compute_capability && !rocm_compute_capability) {
+    return absl::FailedPreconditionError(
+        "Triton support is only enabled for CUDA and ROCm GPUs.");
+  }
+
   if (cuda_compute_capability && !cuda_compute_capability->IsAtLeastAmpere()) {
     return absl::FailedPreconditionError(
         absl::StrCat("Triton support is only enabled for Ampere GPUs (compute ",
```
bchetioui, line 640: This error message might need a little reworking since it's now a bit wrong.

How about changing it to "CUDA Triton support is [...]"?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1376,13 +1376,15 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
...
 
     // Rewrite FP8 GEMMs ahead of Triton which currently lacks support for FP8
     // and may rewrite quantized FP8 GEMMs as higher-precision GEMMs.
     pipeline.AddPass<GemmRewriter>(gpu_version, GetToolkitVersion(),
                                    /*f8_rewrite=*/true);
-    if (debug_options.xla_gpu_enable_triton_gemm() && cuda_cc != nullptr &&
-        cuda_cc->IsAtLeast(se::CudaComputeCapability::AMPERE)) {
+    if (debug_options.xla_gpu_enable_triton_gemm() && ((cuda_cc != nullptr &&
+        cuda_cc->IsAtLeast(se::CudaComputeCapability::AMPERE)) ||
+        rocm_cc)) {
```
bchetioui, line 1387: nit: let's be consistent with the check for `cuda_cc` on the same line and write `rocm_cc != nullptr`?

bchetioui (changes requested):

xla/service/gpu/ir_emitter_triton_rocm.cc
zoranjovanovic-ns: Was using the same check as in corresponding Triton code:
https://github.com/openxla/triton/blob/7c6c2ef26430d34bfec657d6e2d957505425e577/third_party/amd/backend/compiler.py#L151

I can investigate this further to get understanding of semantic meaning.

xla/service/gpu/fusions/triton.cc
zoranjovanovic-ns: If value is not set here it fails following check:
https://github.com/openxla/xla/blob/8caa27b40e443672f2eac42a06cdbe7680547f16/xla/service/gpu/matmul_utils.cc#L824

xla/service/gpu/ir_emitter_triton_rocm.cc
zoranjovanovic-ns: Value zero represents AMD specific software pipelining / double buffering.

bchetioui: Thanks for clarifying that this is indeed intentional. I think that if this is in our codebase, we need to understand exactly what the implications are so that people unfamiliar with the code can maintain it. I think this is not enough as is, because the semantics are still not clear.

I would like us to do the following:

1. let's give a name to `0`, i.e. make it a `constexpr int <some semantically meaningful name> = 0;`;
2. comment on top of the definition what exactly `0` means here. Presumably, this triggers some specific behaviour that is not expressible using other values for `num_stages`---what is it?

xla/service/gpu/ir_emitter_triton_test.cc
```diff
@@ -5641,6 +5657,27 @@ CHECK: wgmma
 )");
 }
 
+// Test usage of default matmul config information.
+TEST_F(TritonGemmTest, TestNoAutotuner) {
```
bchetioui, line 5661: Thanks for adding the test! Unfortunately, it's not clear that this is testing the path we care about---nothing tells us that things were lowered through to Triton here.

Could you please also add a `MatchOptimizedHlo` that checks the relevant properties (that all the relevant parameters in the backend config are set) to make sure that the right path is exercised?

bchetioui (changes requested):
Thank you for your patience and for addressing my comments, this looks in good shape now. Please address my last comment, and after this I think we can merge!

xla/service/gpu/ir_emitter_triton_test.cc
zoranjovanovic-ns: Modified the test case.
The issue is present only on rocm because GemmFusionAutotuner pass is not added into pipeline at all.
For Cuda, even with xla_gpu_autotune_level set to 0, it will always provide some config, thus check will never fail.
That is the reason why test is skipped for cuda.

bchetioui (approved):
Nice, thank you!

PR #14641, opened 2024-07-09
Removed unused deps from bzl files

sergey-kozub:
Some third party dependencies are present both in //third_party and in //third_party/tsl/third_party, one notable example is llvm. This results in patches present in two places (only one set is actually applied), which could be confusing.

This PR removes some of the repeating dependencies and updates BUILD paths.

Also removes `_tf_bind` from workspace0.bzl, as this is a duplication of the rules added by `tsl_workspace0`


akuegel:
I would like to defer the review to @ddunl who is probably more familiar with this open source setup.

beckerhe:
> Replaces "@local_config_nccl//:nccl_headers" with ":nccl", as the former prevents the compilation in OSS

Interesting. This change results in a cyclic dependency on the CI. What exact error message did you get that triggered this change?

sergey-kozub:
> What exact error message did you get that triggered this change?

ERROR: /home/skozub/xla/xla/tsl/cuda/BUILD.bazel:278:11: no such target '@local_config_nccl//:nccl_headers': target 'nccl_headers' not declared in package '' defined by /home/skozub/.cache/bazel/deadbeef/external/local_config_nccl/BUILD (Tip: use `query "@local_config_nccl//:*"` to see all the targets in that package) and referenced by '//xla/tsl/cuda:nccl_stub'
ERROR: Analysis of target '//xla/tsl/cuda:nccl_stub' failed; build aborted: Analysis failed

When I run: bazel query "@local_config_nccl//:*"
I cannot see the nccl_headers target

I've removed this change from this PR, as it's not relevant and should probably go in a separate PR.

third_party/python_runtime/BUILD
```diff
@@ -1,8 +0,0 @@
-licenses(["notice"])  # New BSD, Python Software Foundation
-
-package(default_visibility = ["//visibility:public"])
-
-alias(
-    name = "headers",
-    actual = "@local_config_python//:python_headers",
-)
```
beckerhe, line 8: It seems jaxlib still needs this.

ddunl:
I don't think this change is possible to make from OSS because it requires a substantial change to the Copybara config to work. I'm planning on making a change similar to this but I think it requires a lot of movement internally. One quirk that is still leftover from when XLA was originally taken out of TF is that everything in `third_party/` and `third_party/tsl/third_party` is actually copied from `tensorflow/third_party` internally (gross!). So I agree with the spirit of this change but I don't think Copybara will be able to understand what's happening until I get the internal changes required for this in first.

sergey-kozub:
Marked this PR as draft.
Please fix the issue at some point, as having two LLVM deps is confusing.

PR #14689, opened 2024-07-09
[NVIDIA GPU] Fully unroll windowed einsum loops to hide DUS overheads

Tixxx:
The windowed einsum loops used to be unrolled by a factor of 2 to achieve overlap between 2 gemms. But that leaves some of the dynamic update slices at the end to be exposed.
This pr fully unrolls the loop so DUSes can be overlapped with independent gemms too.
To avoid the loop being inlined by while loop simplifier pass, we add the attribute "skip-simplify-while-loops/trip-count-one=true" to the fully unrolled loop.
Also a minor fix to while loop double buffering pass to skip unroll when trip count is 1.


xla/service/gpu/double_buffer_loop_unrolling.cc
```diff
@@ -533,9 +533,9 @@ absl::StatusOr<bool> DoubleBufferLoopUnrolling::Run(
   for (HloInstruction* while_instr : while_instrs) {
     TF_ASSIGN_OR_RETURN(WhileLoopBackendConfig config,
                         while_instr->backend_config<WhileLoopBackendConfig>());
-    if (!config.has_known_trip_count()) {
+    if (!config.has_known_trip_count() || config.known_trip_count().n() == 1) {
       VLOG(2) << while_instr->ToString()
```
jprabhas, line 537: why do we need to exclude while loops with trip count of 1 from unrolling?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -1142,6 +1143,44 @@ absl::StatusOr<bool> GpuWindowedEinsumHandler::Run(
...
+            loop, /*unroll_factor=*/-1, /*wrap_in_trivial_loop=*/true,
+            /*force_unroll=*/false));
+
+    if (result.unrolled) {
+      result.new_while_op->while_body()->SetAndSanitizeName(
+          absl::StrCat("unrolled_", original_body_name));
+      result.new_while_op->while_condition()->SetAndSanitizeName(
+          absl::StrCat("unrolled_", original_cond_name));
+      // The loop is fully unrolled but has a trip count of 1
+      // To prevent it from being inlined by while loop simplifier,
```
jprabhas, line 1173: whats the rationale for preserving while loops with trip count of 1?

frgossen (changes requested):
> Also a minor fix to while loop double buffering pass to skip unroll when trip count is 1. 

Can that be a separate PR? It might be easier to follow these changes.

xla/service/gpu/double_buffer_loop_unrolling.cc
Tixxx: since we dont have anything to unroll if the loop only has 1 iteration, we can skip the rest of the processing early.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: This is for containing the CM loop in a separate computation to get perfect overlap of computes and comms. If the loop is inlined, it might have undesired schedule.

Tixxx:
> > Also a minor fix to while loop double buffering pass to skip unroll when trip count is 1.
> 
> Can that be a separate PR? It might be easier to follow these changes.

Sure, will do.

Tixxx:
> Fully unrolling the loop solves some problems but it also makes it hard to debug IR since it's so much bigger than. I would like to understand the problem a bit better and why unrolling the loop is the only solution here. Can you provide a small test case to help understand the issue / solution?

To summarize the problem here, the collective matmul loop at a high level contains:
```
p2p_comm
gemm
DUS
```

We initially attempted to unroll by a factor of 2 so we can achieve comm-gemm overlap but also gemm-gemm overlap.
But at the loop boundary, we still have dynamic update slices exposed:
```
p2p_comm1
gemm1
DUS1
p2p_comm2
gemm2
DUS2
```
With fully unrolled loops, we will have all of the sharded gemms overlapped along with almost all DUSes overlapped with either comm or compute, the last DUS will still be exposed but it's an improvement compared to 2-factor unrolled loops.

frgossen:
Does it solve the problem if you move the DUS into the next iteration?

Tixxx:
> Does it solve the problem if you move the DUS into the next iteration?

It will solve the DUS issue but we still want to achieve maximum overlap between gemms, the current schedule for CM loops is:
```
p2p_comm1 p2p_comm2         | p2p_comm3  p2p_comm4
                        DUS |                       DUS 
gemm_1                      |  gemm_3
         gemm_2             |       gemm_4                      
                     (loop boundary)
```

With fully unrolled loops, it would be:
```
comm1 comm2 comm3  comm4
          DUS           DUS
gemm_1     gemm_3
      gemm_2    gemm_4                      
```

xla/service/gpu/double_buffer_loop_unrolling.cc
frgossen: If the trip count is 1, we could unroll it an get rid of the loop, no? 
Also, since you said this is only for loops with a trip count of up to 8, do you think it makes sense to limit this by 8? We could make that configurable in the pass.

frgossen: Is this change here related to the rest of the PR? If yes, how? If no, can it be a separate PR?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -567,6 +538,34 @@ bool ShouldAddToChain(const HloInstruction* inst) {
...
   }
 }
+absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
+  HloComputation* while_body = loop->while_body();
+  // This is to set force delay for the first collective permute so it can
+  // be scheduled always at the top of computation. The GTE index it's consuming
+  // is 2 for RS loop; 0 for AG loop.
+  int64_t force_delay_cp_gte_index =
+      while_body->name().find(
+          GpuWindowedEinsumHandler::kWindowedEinsumRsLoopName) == 0
```
frgossen, line 548: This seems fragile. Can we generalize this? What is the characteristic why you want to schedule it first? 

Why is the index 2 or 0? Does this rely on the previously generated IR?

xla/service/gpu/gpu_windowed_einsum_handler.cc
frgossen: Why do we get an undesired schedule if the loop is inlined? Can we investigate this? I don't think we should rely on not inlining a trivial loop.

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -1142,6 +1143,44 @@ absl::StatusOr<bool> GpuWindowedEinsumHandler::Run(
...
+    if (result.unrolled) {
+      result.new_while_op->while_body()->SetAndSanitizeName(
+          absl::StrCat("unrolled_", original_body_name));
+      result.new_while_op->while_condition()->SetAndSanitizeName(
+          absl::StrCat("unrolled_", original_cond_name));
+      // The loop is fully unrolled but has a trip count of 1
+      // To prevent it from being inlined by while loop simplifier,
+      // we add this attribute to it.
+      xla::FrontendAttributes attributes;
+      (*attributes.mutable_map())["skip-simplify-while-loops/trip-count-one"] =
```
frgossen, line 1176: What is this attribute needed for?

frgossen (changes requested):

xla/service/gpu/double_buffer_loop_unrolling.cc
Tixxx: The change in the file is from a merge conflict when i rebased. I will fix it.
I think setting an upper limit for the unroll makes sense. I can change that.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: The overlap of gemm+gemm and gemm+comm needs some explicit scheduling of instructions to maximize perf. If we inline this into the parent computation, other passes like collective combiner or fusions passes will fuse the CM instructions with others which introduce undesired data dependencies and schedule.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: This attribute is added by the while loop simplifier pass, it will skip inlining trivial loop if the loop has this attribute set to true. Since this has been in xla for a while and only used in while loop simplifier, I'm actually not sure if this is the best way to go about it.
Another alternative is to create a call instruction with the call region containing all unrolled instructions and remove the loop completely from the hlo. In that way, we dont have to worry about other loop optimizations passes affecting our CM computations.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: This is to enforce the order of comm-gemm and gemm+gemm so we get the best overlap, we manually control the order of instructions to be comm->gemm->comm->gemm.
The indices are fixed in the previously generated IR, when SPMD partitioner creates the CM loop, it sets these indices in the loop tuple.

penpornk:
Hi @Tixxx, Is the PR ready for another review or are you still working on it? (Asking just because some of @frgossen's questions haven't been answered.)

xla/service/gpu/gpu_windowed_einsum_handler_test.cc
```diff
@@ -662,11 +659,11 @@ ENTRY test_main {
 
   RunAndFilecheckHloRewrite(kHloString, GpuWindowedEinsumHandler(),
                             R"(
-; CHECK-LABEL: windowed_dot_general_body_ag
```
frgossen, line 665: This test case looks autogenerated to me an I have no chance of understanding this.

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -315,28 +317,13 @@ absl::StatusOr<bool> HandleRsWindowedEinsumLoop(HloComputation* comp,
       // while loop to allow gemm_rewriter.cc to rewrite into an FP8 Custom
       // Call.
       TF_RETURN_IF_ERROR(ShiftDequantizationF8(comp, gte));
-
```
frgossen, line 318: Is this removal related to the PR or an artifact?

xla/service/gpu/gpu_windowed_einsum_handler.cc
```diff
@@ -697,7 +696,7 @@ class WindowedEinsumVisitor : public DfsHloRewriteVisitor {
     // to be consumed by the dot. This is advantageous since the chained dot can
     // fully utilize all the resources on the GPU while comm is hidden by the
     // first collective matmul loop.
-    for (GpuWindowedEinsumHandler::WindowedEinsumAgLoops ag_loop :
+    for (GpuWindowedEinsumHandler::WindowedEinsumAgLoops& ag_loop :
```
frgossen, line 699: Can we factor out changes like this into a separate PR? Itr's unrelated and makes the diff hard to parse.

xla/service/gpu/gpu_windowed_einsum_handler.cc
frgossen: > If we inline this into the parent computation, other passes like collective combiner or fusions passes will fuse the CM instructions with others which introduce undesired data dependencies and schedule.

That means this is a workaround for other passes not doing the right thing, is it? Can we instead fix these other passes?

frgossen (changes requested):
Thanks for discussing this PR offline. It makes sense to unroll the loop, however, I still find this PR hard to review: 

- The test cases are big and autogenerated (?) 
- There are things in here that are not related to the change
- There are things in here that appear to be an artifact of some base PR?

Tixxx:
> That means this is a workaround for other passes not doing the right thing, is it? Can we instead fix these other passes?
It's more like avoiding the other optimization passes to mangle with CM schedule. A CM loop contains ops like DUS or DS, the fusion passes will naturally fuse those with other ops and create fusion kernels that wont make the gemm-gemm overlap very efficient. We could annotate the CM ops and make all other passes aware of those but it seems like an overkill. This approach is to make CM ops more contained. I think we discussed replacing the trivial loop with a generic call instruction, that refactoring is in progress and will be a follow-up pr to this.

xla/service/gpu/gpu_windowed_einsum_handler.cc
Tixxx: It's more like avoiding the other optimization passes to mangle with CM schedule. A CM loop contains ops like DUS or DS, the fusion passes will naturally fuse those with other ops and create fusion kernels that wont make the gemm-gemm overlap very efficient. We could annotate the CM ops and make all other passes aware of those but it seems like an overkill. This approach is to make CM ops more contained. I think we discussed replacing the trivial loop with a generic call instruction, that refactoring is in progress and will be a follow-up pr to this.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -282,7 +285,12 @@ absl::Status UpdateDotAndConsumerConfig(HloInstruction* dot,
   HloInstruction* updater = dot->users()[0];
   auto updater_gpu_config = updater->backend_config<gpu::GpuBackendConfig>();
   dot_gpu_config->set_operation_queue_id(stream_id);
-  updater_gpu_config->mutable_wait_on_operation_queues()->Add(stream_id);
+  const absl::flat_hash_set<int64_t> op_queue_id_set(
```
reedwm, line 288: Instead of converting to a set to check for stream_id, use `absl::c_linear_search(updater_gpu_config->wait_on_operation_queues(), stream_id)`

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -55,11 +58,11 @@ namespace m = match;
 // into
 //
 //   inputs --> while loop {dequant --> collective-permute/dot/etc}.
-absl::Status ShiftDequantizationF8(HloComputation* while_body) {
+absl::StatusOr<bool> ShiftDequantizationF8(HloComputation* while_body) {
```
reedwm, line 61: Document this function returns whether the computation is changed.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -299,78 +307,33 @@ absl::Status SetForceDelayForInstruction(HloInstruction* instr,
...
   // while loop to allow e.g. gemm_rewriter.cc to fuse the dequantization and
   // dot into an FP8 GEMM.
-  TF_RETURN_IF_ERROR(ShiftDequantizationF8(comp));
+  TF_ASSIGN_OR_RETURN(changed, ShiftDequantizationF8(comp));
   return changed;
 }
 
-absl::StatusOr<bool> HandleAgWindowedEinsumLoop(HloComputation* comp,
-                                                int64_t stream_id) {
+absl::StatusOr<bool> HandleAgWindowedEinsumLoop(HloComputation* comp) {
```
reedwm, line 325: Functions `HandleRsWindowedEinsumLoop` and `HandleAgWindowedEinsumLoop` are identical, and pretty much just call `ShiftDequantizationF8`. Remove both, inlining their logic into `WindowedEinsumHandler::Run`.

The `NumberOfInstructionsInComp(comp, HloOpcode::kDot) <= 1` check is no longer done for the call of `UpdateDotAndConsumerConfig`. I don't think this check is useful to guard the call of `ShiftDequantizationF8`. Either remove it outright, or move it to `WindowedEinsumHandler::Run`, only adding the computation to `all_windowed_einsum_loops` if the check holds.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -1163,19 +1154,20 @@ absl::StatusOr<bool> WindowedEinsumHandler::Run(
...
       changed = comp_result;
+      all_windowed_einsum_loops.push_back(comp->WhileCallInstruction());
     } else if (comp->name().find(kWindowedEinsumAgLoopName) == 0) {
       VLOG(5) << "Processing computation: " << comp->name();
-      TF_ASSIGN_OR_RETURN(changed, HandleAgWindowedEinsumLoop(comp, stream_id));
+      TF_ASSIGN_OR_RETURN(changed, HandleAgWindowedEinsumLoop(comp));
       all_ag_loops_.push_back(
           WindowedEinsumAgLoops(comp->WhileCallInstruction()));
+      all_windowed_einsum_loops.push_back(comp->WhileCallInstruction());
     }
```
reedwm, line 1171: Given my previous suggestion to inline `HandleRsWindowedEinsumLoop` and `HandleAgWindowedEinsumLoop`, I think it makes sense to merge the blocks `if (comp->name().find(kWindowedEinsumRsLoopName) == 0)` and `else if (comp->name().find(kWindowedEinsumAgLoopName) == 0)`

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -1186,6 +1178,44 @@ absl::StatusOr<bool> WindowedEinsumHandler::Run(
     changed |= visitor.changed();
   }
 
+  if (all_windowed_einsum_loops.size() > 0) {
+    TF_ASSIGN_OR_RETURN(bool applied_algsimp,
+                        AlgebraicSimplifier(AlgebraicSimplifierOptions())
+                            .Run(module, execution_threads));
+    changed |= applied_algsimp;
+    TF_ASSIGN_OR_RETURN(bool applied_cf,
+                        HloConstantFolding().Run(module, execution_threads));
```
reedwm, line 1187: I'm somewhat nervous that these passes may free or invalid some instructions in `all_windowed_einsum_loops`. I don't think AlgebraicSimplifier can touch while-loop, but I'd rather not rely on this assumption. And maybe `HloConstantFolding` can get rid of while-loops.

Why do you need these passes? Please comment this. And can you run them before creating `all_windowed_einsum_loops`? Or alternatively, you can clear `all_windowed_einsum_loops` and find and put the while-loop instructions back in the vector after running the passes.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -1186,6 +1178,44 @@ absl::StatusOr<bool> WindowedEinsumHandler::Run(
...
+    if (result.unrolled) {
+      result.new_while_op->while_body()->SetAndSanitizeName(
+          absl::StrCat("unrolled_", original_body_name));
+      result.new_while_op->while_condition()->SetAndSanitizeName(
+          absl::StrCat("unrolled_", original_cond_name));
+      // The loop is fully unrolled but has a trip count of 1
+      // To prevent it from being inlined by while loop simplifier,
+      // we add this attribute to it.
+      xla::FrontendAttributes attributes;
+      (*attributes.mutable_map())["skip-simplify-while-loops/trip-count-one"] =
```
reedwm, line 1211: Due to https://github.com/openxla/xla/commit/78418c6a4a60cff6b3ad538de43602d23dc2aaf5, change string to `"skip-simplify-while-loops_trip-count-one"`.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -1186,6 +1178,44 @@ absl::StatusOr<bool> WindowedEinsumHandler::Run(
...
+    VLOG(5) << "Processing " << loop->ToString() << " for unrolling.";
+    std::string original_body_name = std::string(loop->while_body()->name());
+    std::string original_cond_name =
+        std::string(loop->while_condition()->name());
+
+    TF_ASSIGN_OR_RETURN(
+        UnrollResult result,
+        WhileLoopUnroller::UnrollAndReturnReplacement(
+            loop, /*unroll_factor=*/-1, /*wrap_in_trivial_loop=*/true,
+            /*force_unroll=*/false));
```
reedwm, line 1200: Comment the motivation for unrolling the while-loop, and for not allowing it to be inlined.

Also is the motivation to overlap DUS at the end of each iteration with the matmul and/or collective at the start of the next iteration?

xla/service/while_loop_unroller.h
```diff
@@ -42,6 +42,15 @@ struct WhileLoopConfig {
   int64_t induction_var_idx;
 };
 
+// Config for unrollable while loops.
+struct UnrollResult {
+  // Whether it's unrolled.
+  bool unrolled = false;
+  // New unrolled while op that replaced the old one.
+  // For non-trivially wrapped case, this would be nullptr.
```
reedwm, line 50: This comment is unclear to me. What is the "non-trivially wrapped case"? This is only nullptr if either the while-loop was not unrolled or if `wrap_in_trivial_loop=false`, in which case there is no new while-op, right? If so, phrase something like

> If the while-loop has been unrolled and replaced with a new unrolled while-loop with a trip count of 1, this is the new while-loop. Otherwise this is nullptr.

xla/service/while_loop_unroller.h
```diff
@@ -117,6 +126,12 @@ class WhileLoopUnroller : public HloModulePass {
                                      bool wrap_in_trivial_loop = false,
                                      bool force_unroll = false,
                                      bool prepare = true);
+  // Similar to above, but returns the unrolled flag and
+  // the new unrolled while instruction.
+  static absl::StatusOr<UnrollResult> UnrollAndReturnReplacement(
```
reedwm, line 131: I don't think it's worth adding a new method. The existing Unroll method is only used [here](https://github.com/openxla/xla/blob/0fbf2e1eef1c45ade8ebc8e070b734d81112a30c/xla/service/hlo_unstacker.cc#L1422). Instead just return `UnrollResult` from Unroll and change the single current usage.

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
```diff
@@ -110,15 +110,15 @@ ENTRY test_main {
...
-      cp1->backend_config<GpuBackendConfig>()->force_earliest_schedule());
+  int64_t dot_count = 0;
+  for (HloInstruction* inst : ag_loop_body->MakeInstructionPostOrder()) {
+    if (inst->opcode() == HloOpcode::kDot) {
+      dot_count++;
+      EXPECT_GT(inst->backend_config<GpuBackendConfig>()->operation_queue_id(),
+                0);
+    }
+  }
+  EXPECT_EQ(dot_count, 4);
```
reedwm, line 121: You no longer check the `force_earliest_schedule` field on the dots and collective-permute. SHould you check this?

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
```diff
@@ -263,28 +264,23 @@ ENTRY main.12_spmd {
...
-                              .WithPredicate([](const HloInstruction* instr) {
-                                return instr->tuple_index() == 5;
-                              }),
-                          m::Op(), m::Op(), m::Op(), m::Op()),
-                      m::Op(), m::Op(), m::Op(), m::Op()))));
+  const HloInstruction* while_loop = inst->operand(0)->operand(0);
+  EXPECT_EQ(while_loop->opcode(), HloOpcode::kWhile);
+  HloComputation* while_body = while_loop->while_body();
+  int64_t dot_count = 0;
+  for (HloInstruction* ins : while_body->MakeInstructionPostOrder()) {
```
reedwm, line 276: You should still check there is a DUS chain, as was done before. This chain should still exist even with the while-loop being unrolled, right? In fact, I think the chain will be much longer this time due to the unrolling.

Also it would be good to check the all-gather is removed and replaced with the DUS chain output of the while-loop. But since this isn't already done, it's fine if this PR doesn't address this.

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
```diff
@@ -703,16 +702,75 @@ ENTRY test_main {
...
+; CHECK-NEXT:    [[CONVERT4:%[^ ]+]] = s32[] convert([[CLAMP1]])
+; CHECK-NEXT:    [[MUL4:%[^ ]+]] = s32[] multiply([[CONVERT4]], [[C6]])
+; CHECK-NEXT:    [[RESHAPE1:%[^ ]+]] = s32[] reshape([[MUL4]])
 ; CHECK-NEXT:    [[DUPDATESLICE1:%[^ ]+]] = f32[2,2048,24576]{2,1,0} dynamic-update-slice([[DUPDATESLICE0]], [[DOT1]], [[C0]], [[RESHAPE1]], [[C0]])
 ; CHECK-NEXT:    [[GTE6:%[^ ]+]] = f32[2,2048,24576]{2,1,0} get-tuple-element([[P0]]), index=3
-; CHECK-NEXT:    [[ADD3:%[^ ]+]] = u32[] add([[ADD1]], [[C3]])
-; CHECK-NEXT: ROOT [[OUT:%[^ ]+]] = (f8e4m3fn[2,512,24576]{2,1,0}, f8e4m3fn[24576,24576]{1,0}, f32[2,2048,24576]{2,1,0}, f32[2,2048,24576]{2,1,0}, u32[], /*index=5*/f32[], f32[]) tuple([[CP1]], [[GTE1]], [[DUPDATESLICE1]], [[GTE6]], [[ADD3]], /*index=5*/[[GTE3]], [[GTE4]])
+; CHECK-NEXT:    [[C7:%[^ ]+]] = u32[] constant(2)
+; CHECK-NEXT:    [[ADD3:%[^ ]+]] = u32[] add([[GTE7]], [[C7]])
+; CHECK-NEXT:    [[TUPLE0:%[^ ]+]] = (f8e4m3fn[2,512,24576]{2,1,0}, f8e4m3fn[24576,24576]{1,0}, f32[2,2048,24576]{2,1,0}, f32[2,2048,24576]{2,1,0}, u32[], /*index=5*/f32[], f32[]) tuple([[CP1]], [[GTE1]], [[DUPDATESLICE1]], [[GTE6]], [[ADD3]], /*index=5*/[[GTE3]], [[GTE4]])
```
reedwm, line 718: I don't think it's worth checking that the whole loop is unrolled properly for multiple tests, since it's so verbose. I would add just a single new test to do this.

You can just remove the CHECK-NEXTs  for the rest of the while-loop. No need to match the whole body. Or alternatively, reduce the num_partitions to 2.

reedwm (changes requested):

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: Running this is to prepare the loops for unrolling, the unroller util function determines loop iterator with a very specific pattern which expects alg simp and constant fold applied. I checked that both passes wont touch while loops but i will document the intention to run these passes.

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: life saver! i didnt know this is renamed.

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: Yes, this is to allow DUS to overlap with matmul and collective from other iterations, in practice, we actually can overlap the matmul AND DUS from current iteration with matmul and collective from other iterations.

reedwm: Is the DUS codegen-ed as a specific kernel? I would have expected the dot kernel to directly write into the correct place in the buffer, instead of writing to a temp buffer before being copied to the output buffer via DUS.

Tixxx: the DUS is code-gened as a part of a fusion kernel fused with some other trivial compute kernels. We have a dot+dus fusion to fuse this but it's not working reliably for collective matmul due to the way the DUS index is computed in the loop, that's being worked on separately from our side. Even with the DUS fused, we still need to unroll to overlap gemms across iterations to minimize wave quantization overhead.

reedwm: Thanks, that all makes sense

xla/service/while_loop_unroller.h
Tixxx: im hesitant to do that, this pass is not used anywhere in OSS but it's being actively maintained, so probably some other internal compilers are using this pass? this function is a public static function.

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
Tixxx: good catch, accidentally removed it.

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
Tixxx: i think the DUS output replacement is checked but allgather's presence is not. i can add the check here

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
Tixxx: makes sense, i removed all but one test

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: updated based on the suggestion

reedwm (approved):

PR #14692, opened 2024-07-09
[GPU][NFC] Skip tests requiring large shared memory on low-spec GPUs.

sergachev:


sergachev:
ping

xla/service/gpu/kernels/cutlass_gemm_fusion_test.cc
```diff
@@ -351,7 +359,13 @@ TEST_F(CutlassFusionTest, RowMajorGemmWithUpcastKernel) {
                                       error_spec, /*run_hlo_passes=*/false));
 }
 
+constexpr int kRequiredSharedMemorySize = 147456;
```
thomasjoerg, line 362: Is this constant derived from the test fixtures (i.e. HLO snippets) somehow? If so, let's make this more readable by specifying the calculation not just the result. It's also easy to forget to update this, should test fixtures change. Please keep it within the tests right next to the fixtures.

thomasjoerg (changes requested):

xla/service/gpu/kernels/cutlass_gemm_fusion_test.cc
sergachev: done

thomasjoerg (approved):

PR #14699, opened 2024-07-09
Add while-loop-trip-count-annotator before collective-permute-valid-iteration-annotator pass

shraiysh:
While loop trip count was not annotated to the while instruction, which meant that the collective-permute-valid-iteration-annotator could not deduce the valid iterations for collective-permute


frgossen (approved):
Thanks for fixing this!

SsomsakTH (approved):

PR #14759, opened 2024-07-10
Adjust block size of transpose by registers count

ghost:
Transpose mlir emitter is tiled-base ir emitter, which need large amount of registers and lead to high register pressure. We should choose threads_number_per_block carefully to avoid this situation. By evaluating the register usage at the level of hlo instructions, transpose can achieve better performance caused by better setting of threads_number_per_block. The following is a brief description of the evaluation method registers count:

We compute the indexing map of parameters of fusion. We assume the loop in transpose is fully unrolled, so the bytes_per_block used by param is the interval size of indexing map. And then we iterate hlo with the post order of fusion and compute the max bytes count:
```
instr_bytes = param_bytes
max_instr_bytes = instr_bytes
for instr in post_order:
  unused_bytes = 0
  output_bytes = 0
  if instr only involve index computation, like broadcast, bitcast:
    reuse operand bytes
  else:
    output_bytes = output bytes of instr
    for operand in instr.operands():
      unused_bytes += operand bytes
   if output can reuse any operand register:
     unused_bytes -= reused_operand_bytes
     output_bytes = 0
  instr_bytes += output_bytes
  max_instr_bytes = max(max_instr_bytes, instr_bytes)
  instr_bytes -= unused_bytes
// 32bit registers
max_registers = max_instr_bytes / 4
```
We test multiple transpose fusion on A100 40GB device, results can be seen in the below (* means which block size this PR choose):

```
Case 0:
HloModule Transpose
    %fused_computation {
      param_0 = u8[8,6144,4096]{2,1,0} parameter(0)
      transpose.31511.1 = u8[8,4096,6144]{2,1,0} transpose(param_0), dimensions={0,2,1}
    }
    ENTRY main {
      %p0 = u8[8,6144,4096]{2,1,0} parameter(0)
      ROOT %fusion = u8[8,4096,6144]{2,1,0} fusion(%p0), kind=kInput,
        calls=%fused_computation
    }

block_size     128     256*    512     1024
time (us)      318     310     320      353

Case 1:
HloModule Transpose
    %fused_computation {
      param_0 = u8[8,6144,4096]{2,1,0} parameter(0)
      convert.26818.3 = f32[8,6144,4096]{2,1,0} convert(param_0)
      param_1.4141 = bf16[8,1,4096]{2,1,0} parameter(1)
      bitcast.98403.6 = bf16[8,4096]{1,0} bitcast(param_1.4141)
      convert.29908.3 = f32[8,4096]{1,0} convert(bitcast.98403.6)
      broadcast.26006.5 = f32[8,6144,4096]{2,1,0} broadcast(convert.29908.3), dimensions={0,2}
      multiply.11854.5 = f32[8,6144,4096]{2,1,0} multiply(convert.26818.3, broadcast.26006.5)
      convert.26820.3 = bf16[8,6144,4096]{2,1,0} convert(multiply.11854.5)
      transpose.31547.1 = bf16[8,4096,6144]{2,1,0} transpose(convert.26820.3), dimensions={0,2,1}
    }
    ENTRY main {
      %p0 = u8[8,6144,4096]{2,1,0} parameter(0)
      %p1 = bf16[8,1,4096]{2,1,0} parameter(1)
      ROOT %fusion = bf16[8,4096,6144]{2,1,0} fusion(%p0, %p1), kind=kInput,
        calls=%fused_computation
    }

block_size     128     256    512*     1024
time (us)      569     559    544       748

Case 2:
HloModule Transpose
    %fused_computation {
      param_0 = u8[8,6144,4096]{2,1,0} parameter(0)
      transpose.31547.1 = u8[8,4096,6144]{2,1,0} transpose(param_0), dimensions={0,2,1}
      convert.26818.3 = f32[8,4096,6144]{2,1,0} convert(transpose.31547.1)
      param_1.4141 = bf16[8,1,4096]{2,1,0} parameter(1)
      bitcast.98403.6 = bf16[8,4096]{1,0} bitcast(param_1.4141)
      convert.29908.3 = f32[8,4096]{1,0} convert(bitcast.98403.6)
      broadcast.26006.5 = f32[8,4096,6144]{2,1,0} broadcast(convert.29908.3), dimensions={0,1}
      multiply.11854.5 = f32[8,4096,6144]{2,1,0} multiply(convert.26818.3, broadcast.26006.5)
      convert.26820.3 = bf16[8,4096,6144]{2,1,0} convert(multiply.11854.5)
    }
    ENTRY main {
      %p0 = u8[8,6144,4096]{2,1,0} parameter(0)
      %p1 = bf16[8,1,4096]{2,1,0} parameter(1)
      ROOT %fusion = bf16[8,4096,6144]{2,1,0} fusion(%p0, %p1), kind=kInput,
        calls=%fused_computation
    }

block_size     128     256    512     1024*
time (us)      526     528    526      507
```

### DIFF
--- a/xla/service/gpu/fusions/transpose_mlir.cc
+++ b/xla/service/gpu/fusions/transpose_mlir.cc
@@ -76,9 +76,9 @@ using mlir::tensor::ExtractOp;
 using mlir::tensor::InsertOp;
 using mlir_converter::ApplyIndexing;
 
-constexpr int kNumRows = 4;
 constexpr int kBaseBlockSize = WarpSize();
-constexpr int kNumThreadsPerBlock = 128;
+constexpr int kMinThreadsPerBlock = 128;
+constexpr int kMaxThreadsPerBlock = 1024;
 constexpr int kMaxVectorizedBytes = 4;
 
 }  // namespace
@@ -130,7 +130,7 @@ MlirTransposeFusion::MlirTransposeFusion(const HloFusionAnalysis& analysis)
   for (int vec_size = kMaxVectorizedBytes / max_element_bytes; vec_size > 1;
        vec_size /= 2) {
     int elems_per_thread = vec_size * vec_size;
-    bool enough_work = Product(block_counts_) * kNumThreadsPerBlock >=
+    bool enough_work = Product(block_counts_) * kMinThreadsPerBlock >=
                        elems_per_thread * device.core_count() *
                            device.threads_per_core_limit();
     bool enough_shmem =
@@ -142,6 +142,11 @@ MlirTransposeFusion::MlirTransposeFusion(const HloFusionAnalysis& analysis)
       break;
     }
   }
+
+  shmem_usage_per_block_ = shmem_usage * vector_size_ * vector_size_;
+  threads_per_block_ = ChooseSuitableThreadsNumber(device);
+  CHECK(threads_per_block_ % kBaseBlockSize == 0);
+  num_rows_ = threads_per_block_ / kBaseBlockSize;
 }
 
 std::optional<IndexingMap> MlirTransposeFusion::ComputeThreadIdToOutputIndexing(
@@ -179,7 +184,7 @@ std::optional<IndexingMap> MlirTransposeFusion::ComputeThreadIdToInputIndexing(
 }
 
 LaunchDimensions MlirTransposeFusion::launch_dimensions() const {
-  return LaunchDimensions(Product(block_counts_), kNumThreadsPerBlock);
+  return LaunchDimensions(Product(block_counts_), threads_per_block_);
 }
 
 IndexingMap MlirTransposeFusion::GetSharedMemoryIndexing(
@@ -187,8 +192,8 @@ IndexingMap MlirTransposeFusion::GetSharedMemoryIndexing(
   auto thread_offsets =
       Permute(GetThreadOffsets(ctx), read ? Vector3{0, 1, 2} : permutation_);
   return {mlir::AffineMap::get(6, 2, thread_offsets, ctx),
-          DimVarsFromTensorSizes({kNumThreadsPerBlock, 1, 1, 1, 1, 1}),
-          RangeVarsFromTensorSizes({block_size_ / kNumRows, vector_size_}),
+          DimVarsFromTensorSizes({threads_per_block_, 1, 1, 1, 1, 1}),
+          RangeVarsFromTensorSizes({block_size_ / num_rows_, vector_size_}),
           {}};
 }
 
@@ -368,7 +373,7 @@ llvm::SmallVector<mlir::AffineExpr, 4> MlirTransposeFusion::GetThreadOffsets(
       KernelFusionInterface::kIndexingMapThreadIdxDims[0], ctx);
   auto loop = mlir::getAffineSymbolExpr(0, ctx);
   auto vector = mlir::getAffineSymbolExpr(1, ctx);
-  int loop_stride = block_size_ * kNumRows;
+  int loop_stride = block_size_ * num_rows_;
   auto linear_index = loop * loop_stride + thread * vector_size_ + vector;
   return DelinearizeInBoundsIndex(linear_index, block_sizes_);
 }
@@ -389,8 +394,8 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
   IndexingMap result{
       mlir::AffineMap::get(6, 2, offsets, ctx),
       DimVarsFromTensorSizes(
-          {kNumThreadsPerBlock, 1, 1, Product(block_counts_), 1, 1}),
-      RangeVarsFromTensorSizes({block_size_ / kNumRows, vector_size_}),
+          {threads_per_block_, 1, 1, Product(block_counts_), 1, 1}),
+      RangeVarsFromTensorSizes({block_size_ / num_rows_, vector_size_}),
       {}};
   auto normalized_shape =
       input ? ShapeUtil::MakeShape(shape.element_type(), input_shape_)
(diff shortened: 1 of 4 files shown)


jreiffers:
Thanks for the PR. Since there's a lot of new logic here, reviewing it will probably take a while.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+
+struct HloBytesUsageDesc {
+  int tile_size;
+  int elem_bytes;
+  int ref_count;
+  int ref_id;
+  int unused_bytes;
+};
+
+std::optional<HloBytesUsageDesc> GetHloBytesUsageDesc(
```
jreiffers, line 423: This doesn't look like it's specific to transpose, is it? If not, let's move it to xla/service/gpu/model, since it seems related to cost modeling. Let's also add a bunch of comments, since as is, it's too hard to understand what's going on here.

I'll leave some comments inline to hopefully help a bit with readability.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+    HloInstructionAdaptor& instr_adaptor,
+    absl::InlinedVector<int, 2>& live_range,
+    absl::InlinedVector<HloBytesUsageDesc, 2>* instr_bytes_desc,
+    absl::flat_hash_map<const HloInstruction*, int>& instr_to_ids) {
+  auto* instr = &instr_adaptor.instruction();
+  // These Hlo which only involve index computation don't need to
```
jreiffers, line 429: This is not strictly true. It depends on the index computation. But you can say that registers used for indices are ignored.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+  int tile_size;
+  int elem_bytes;
+  int ref_count;
+  int ref_id;
+  int unused_bytes;
+};
+
+std::optional<HloBytesUsageDesc> GetHloBytesUsageDesc(
+    HloInstructionAdaptor& instr_adaptor,
+    absl::InlinedVector<int, 2>& live_range,
```
jreiffers, line 425: This should be const?

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+struct HloBytesUsageDesc {
+  int tile_size;
+  int elem_bytes;
+  int ref_count;
+  int ref_id;
+  int unused_bytes;
+};
+
+std::optional<HloBytesUsageDesc> GetHloBytesUsageDesc(
+    HloInstructionAdaptor& instr_adaptor,
```
jreiffers, line 424: HloInstructionAdaptors should be passed by value.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+    absl::InlinedVector<int, 2>& live_range,
+    absl::InlinedVector<HloBytesUsageDesc, 2>* instr_bytes_desc,
+    absl::flat_hash_map<const HloInstruction*, int>& instr_to_ids) {
```
jreiffers, line 427: Let's get rid of the `instrs_to_ids` structure and instead index the data structures directly by pointers. The added indirection doesn't seem to do anything useful here, but it's harder to read.

xla/service/gpu/fusions/transpose_mlir.cc
jreiffers, line 425: What does this mean? It seems to be a map from ID to ID, but I don't really understand how that is a live range.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+    auto& input_desc = (*instr_bytes_desc)[input_id];
+    if (instr_to_ids[instr] == live_range[input_id]) {
+      (*instr_bytes_desc)[input_desc.ref_id].ref_count--;
+    }
+    (*instr_bytes_desc)[input_desc.ref_id].ref_count++;
+    return HloBytesUsageDesc{0, input_desc.elem_bytes, 0, input_desc.ref_id, 0};
+  }
+  if (instr->IsElementwise()) {
+    int output_bytes = primitive_util::ByteWidth(instr->shape().element_type());
+    int max_tile_size = 0, unused_bytes = 0, reuse_desc_id = 0;
```
jreiffers, line 444: What do max_tile_size and unused_bytes mean here?

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
   return result;
 }
 
+namespace {
+
+struct HloBytesUsageDesc {
```
jreiffers, line 415: Can you add comments for all of these members? I don't really understand what they are.

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+        if (ref_desc.ref_count == 0) {
+          unused_bytes += ref_desc.tile_size * ref_desc.elem_bytes;
+        }
+      }
+      if (ref_desc.tile_size >= max_tile_size) {
+        // Dest can reuse src registers iff src has no next user and
+        // owned bytes is larger than needed bytes.
+        bool reuse_operand =
+            ref_desc.ref_count == 0 && output_bytes <= ref_desc.elem_bytes;
+        can_reuse = ref_desc.tile_size > max_tile_size
```
jreiffers, line 462: Why?

xla/service/gpu/fusions/transpose_mlir.cc
```diff
@@ -405,5 +410,376 @@ IndexingMap MlirTransposeFusion::GetIndexing(bool input,
...
+        if (ref_desc.ref_count == 0) {
+          instrs_bytes -= ref_desc.tile_size * ref_desc.elem_bytes;
+        }
+      }
+    }
+  }
+  return max_bytes_count;
+}
+
+std::optional<int> ComputeRegNumFromParamToTarget(
```
jreiffers, line 533: I don't really understand why indexing maps should be related to register pressure at all. Say we have a reduction in the prologue: the register use *might* depend a bit on the size of the reduction, but not significantly. For large reductions, it doesn't depend on it at all. For small reductions, where the loop is unrolled, we get something like

```
reducer(reducer(reducer(reducer(init, e[0]), e[1]), e[2]), e[3])
```

As you can see, this shouldn't need signficantly more registers than reducer(init, e[0]) on its own, since no intermediates need to be kept around.

jreiffers (changes requested):

xla/service/gpu/fusions/transpose_mlir.cc
lingzhi98: I focus on the indexing map of parameters but not other. For the above case you provide, I don't compute the registers usage of intermediates. And the indexing map of parameter help us to distinguish these two cases:
```
parameter -> transpose
parameter -> broadcast -> transpose
```
The second case need less registers than the first. And the indexing map is special, we only compute loaded elements in one block, you can see this line https://github.com/openxla/xla/pull/14759/files#diff-0210cd152294b1475469517ffba6b5ee578acc4e786b0dceeee6854aba445a29R647.

lingzhi98:
Why 

> This seems like a lot of added complexity for not very significant gains. Can the logic here be simplified a bit? If not, I find it hard to justify the maintenance cost given the speedups.
> 
> Can you also explain the measurements a bit? If this is about register pressure, why do the more complex fusions get larger block sizes?


No matter what block size we choose, the tile size of transpose is same https://github.com/openxla/xla/pull/14759/files#diff-0210cd152294b1475469517ffba6b5ee578acc4e786b0dceeee6854aba445a29R376. So the block size is larger, blocks per core is less, the tile elements loaded by per core is less too.
And I agree with you,  the logic of PR is too complex but performance gain is small. Maybe we need simple but useful heuristic here. I will try to find it and close this PR for now.

Issue #14352, opened 2024-07-02
Build fails


AleksKnezevic:
I'm trying to build xla from source for CPU following the instructions [here](https://github.com/openxla/xla/blob/main/docs/developer_guide.md) and it's failing with: 

```
xla/service/gpu/runtime/command_buffer_cmd_emitter.cc:32:10: fatal error: 'xla/service/gpu/runtime/gpublas_lt_matmul_thunk.h' file not found
   32 | #include "xla/service/gpu/runtime/gpublas_lt_matmul_thunk.h"
```

I'm on an ubuntu 20.04 machine running a TF docker as the guide recommends.  I have bazel 6.5.0 and clang 17. Is this config supported? Anything better to use? 

If I run build with `-s` I see that `-iquote .` is included, and `xla/service/gpu/runtime/gpublas_lt_matmul_thunk.h` is there. Here's the full subcommand output:
```
SUBCOMMAND: # //xla/mlir_hlo:transforms_passes [action 'Compiling xla/mlir_hlo/transforms/alloc_to_arg_pass.cc', configuration: 94361239bbff404c4f64fc2130c27f83352dcbf11f78849ccfba5713e8af8cb0, execution platform: @local_execution_config_platform//:platform]
(cd /root/.cache/bazel/_bazel_root/e4ab50d61a21943a819d1e092972a817/execroot/xla && \
  exec env - \
    CLANG_COMPILER_PATH=/usr/lib/llvm-17/bin/clang \
    DOCKER_CACHEBUSTER=1719550690218542678 \
    LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64 \
    PATH=/root/.cache/bazelisk/downloads/bazelbuild/bazel-6.5.0-linux-x86_64/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
    PWD=/proc/self/cwd \
    PYTHON_BIN_PATH=/usr/bin/python3 \
    TF2_BEHAVIOR=1 \
[...]

PR #14778, opened 2024-07-11
Enable xla cc tests in OSS XLA gpu jobs.

hmonishN:
Enable xla_cc_tests defined in //xla/service/gpu dir and subdirs by adding the gpu tags to them so that they can be picked up in OSS CI XLA gpu job. 
Currently these tests are not picked up in gpu job as they are missing any gpu tag specified as part of bazel test/build_tag_filters

### DIFF
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -15,6 +15,10 @@ load(
 )
 load(
     "@tsl//tsl/platform:build_config_root.bzl",
+    "tf_cuda_tests_tags",
+)
+load(
+    "@tsl//tsl/platform:build_config_root.bzl",
     "if_static",
 )
 load(
@@ -80,6 +84,7 @@ tf_proto_library(
 xla_cc_test(
     name = "backend_configs_test",
     srcs = ["backend_configs_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":backend_configs_cc",
         "//xla/hlo/ir:hlo",
@@ -187,6 +192,7 @@ xla_test(
 xla_cc_test(
     name = "gpu_copy_insertion_test",
     srcs = ["gpu_copy_insertion_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":buffer_sharing",
         "//xla:test",
@@ -251,6 +257,7 @@ cc_library(
 xla_cc_test(
     name = "target_util_test",
     srcs = ["target_util_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":target_util",
         "//xla/tests:xla_internal_test_main",
@@ -723,6 +730,7 @@ cc_library(
 xla_cc_test(
     name = "ir_emitter_triton_mem_utils_test",
     srcs = if_cuda_is_configured(["ir_emitter_triton_mem_utils_test.cc"]),
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":hlo_traversal",
         ":ir_emitter_triton",
@@ -1126,6 +1134,7 @@ cc_library(
 xla_cc_test(
     name = "ir_emission_utils_test",
     srcs = ["ir_emission_utils_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":hlo_traversal",
         ":ir_emission_utils",
@@ -1168,6 +1177,7 @@ cc_library(
 xla_cc_test(
     name = "reduction_utils_test",
     srcs = ["reduction_utils_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":reduction_utils",
         "//xla/hlo/ir:hlo",
@@ -1288,6 +1298,7 @@ cc_library(
 xla_cc_test(
     name = "triton_support_test",
     srcs = ["triton_support_test.cc"],
+    tags = tf_cuda_tests_tags(),
     shard_count = 20,
     deps = [
         ":gpu_device_info_for_tests",
@@ -1368,6 +1379,7 @@ cc_library(
 xla_cc_test(
     name = "triton_tiling_propagation_test",
     srcs = ["triton_tiling_propagation_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":triton_tiling_propagation",
         "//xla/tests:hlo_test_base",
@@ -1407,6 +1419,7 @@ cc_library(
 xla_cc_test(
     name = "triton_fusion_analysis_test",
     srcs = ["triton_fusion_analysis_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":gemm_fusion",
         ":triton_fusion_analysis",
@@ -1457,6 +1470,7 @@ cc_library(
 xla_cc_test(
     name = "gemm_fusion_test",
     srcs = ["gemm_fusion_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":cublas_padding_requirements",
         ":gemm_fusion",
@@ -1504,6 +1518,7 @@ cc_library(
 xla_cc_test(
     name = "gemv_rewriter_test",
     srcs = ["gemv_rewriter_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":gemv_rewriter",
         "//xla/hlo/ir:hlo",
@@ -1549,6 +1564,7 @@ cc_library(
 xla_cc_test(
     name = "split_k_gemm_rewriter_test",
     srcs = ["split_k_gemm_rewriter_test.cc"],
+    tags = tf_cuda_tests_tags(),
     deps = [
         ":matmul_utils",
         ":split_k_gemm_rewriter",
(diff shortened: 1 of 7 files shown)


hmonishN:
Adding @ddunl

ddunl:
Hmm.. It definitely is a bug that these are not running. I think the correct change is to remove the target filtering in the build script. I'm looking back at the history and it's not totally clear to me why we did this or what the alternative is. I think the maximally correct thing is to run these on CPU CI as they don't need a GPU to run. Could you change this PR to deleting the target exclusion here: https://github.com/openxla/xla/blob/main/build_tools/build.py#L248 ?

hmonishN:
I did experiment to see if these any of these tests needs an actual gpu or not. There were two findings:

1.  If we enable //xla/service/gpu/... in CPU build the build fails as a lot of cc_libraries/test target dependencies are guarded by if_gpu_is_configured flag e.g.: https://github.com/hmonishN/xla/blob/hmonish/enable_xla_cc_tests/xla/service/gpu/BUILD#L1717, https://github.com/hmonishN/xla/blob/hmonish/enable_xla_cc_tests/xla/service/gpu/BUILD#L1674, etc. and in sub dirs as well which causes build failures.

2. These tests needs actual gpu to run: 
- //xla/service/gpu/fusions:concatenate_test
- //xla/service/gpu/fusions:in_place_dynamic_update_slice_test
- //xla/service/gpu/fusions:input_slices_test
- //xla/service/gpu/fusions:loop_test
- //xla/service/gpu/fusions:scatter_test
- //xla/service/gpu/fusions:transpose_test
- //xla/service/gpu/fusions:triton_test
- //xla/service/gpu/model:coalescing_analysis_test
- //xla/service/gpu/model:gpu_cost_model_stats_collection_test
- //xla/service/gpu/model:gpu_indexing_performance_model_test
- //xla/service/gpu/model:gpu_performance_model_base_test
- //xla/service/gpu/model:gpu_performance_model_test
- //xla/service/gpu:fusion_merger_test
- //xla/service/gpu:gpu_aot_compilation_test
- //xla/service/gpu:multi_output_fusion_test
- //xla/service/gpu:priority_fusion_test
- //xla/service/gpu:softmax_rewriter_triton_test

Hence the safest way to enable them was to add them to GPU job.

ddunl:
I think we need to solve `1.`. I'm not sure I agree with 2 looking at a few examples. If "needs a gpu" -> needs a gpu because the build rules demand it (due to depending on something with `if_gpu_is_configured`) then I agree, but I don't think conceptually they need a GPU - I'd expect the GPU will be cold the whole time the test is running (but I could be mistaken)

hmonishN:
In the tests definition, the calls are made to fetch gpu information using stream executor and then computations are being done as well: e.g. https://github.com/openxla/xla/blob/main/xla/service/gpu/fusions/in_place_dynamic_update_slice_test.cc#L72
All the above tests don't run on CPU.

hmonishN:
Adding @shraiysh

ddunl:
It's not clear to me that this needs a GPU still, we run anything without `cuda_test_tags` internally on CPUs. I don't think computations are actually executed, it inspects a module, but I don't think execution is required. I don't think the stream_executor bit needs a GPU either because it's using https://github.com/openxla/xla/blob/7c5744ea674ca992204eccb4e18d24c62d37ca1a/xla/service/gpu/gpu_device_info_for_tests.cc#L16

shraiysh:
Thanks for pointing this out @ddunl. I went back and tried to run these tests without a GPU again and I realized that the way I got a list of these tests was inaccurate. I am now able to execute these tests without a GPU. To run these tests without a GPU, we still need to configure the project with GPU backend. I will sync up with @hmonishN to figure out the next steps here.

hmonishN:
Then we can make it part of the CPU job the only thing is we do set the backend config for CPU similar to rbe_linux_cpu, which is causing these tests to fail due to compilation error. 
I think first let's try to fix 1. issue and then make 2. tests part of CPU job with cuda config settings?

akuegel:
> Then we can make it part of the CPU job the only thing is we do set the backend config for CPU similar to rbe_linux_cpu(configure.py --backend CPU) , which is causing these tests to fail due to compilation error. I think first let's try to fix 1. issue and then make 2. tests part of CPU job with cuda config settings?

Note that recently I already did a few fixes to make the build inside xla/service/gpu directory work (https://github.com/openxla/xla/issues/14352). Did you try with latest XLA head? Internally I tried building everything inside gpu (including subdirectories) with letting if_cuda_is_configured and if_gpu_is_configured taking the false path, and there was no compile error anymore after my fixes.

akuegel:
> Hmm.. It definitely is a bug that these are not running. I think the correct change is to remove the target filtering in the build script. I'm looking back at the history and it's not totally clear to me why we did this or what the alternative is. I think the maximally correct thing is to run these on CPU CI as they don't need a GPU to run. Could you change this PR to deleting the target exclusion here: https://github.com/openxla/xla/blob/main/build_tools/build.py#L248 ? @thomasjoerg who may remember the motivation

I looked at the history of this build script, and the exclusion was there from the beginning in your initial change. My guess would be the motivation was that compilation failed, or that it was not clear that some tests inside gpu directory actually can run on CPU.

ddunl:
I was thinking of this: https://github.com/openxla/xla/commit/38c7199fa4793fd77aa257a49d845d38e7771066 I looked at the description internally but still haven't quite pieced together the problem. I think there was a different change that wouldn't pass CI without this one bc of changes in dependencies

hmonishN:
With the latest changes, I did some experiments for this and following are the findings:

1. including //xla/service/gpu/... in CPU build job fails for this dir: /xla/service/gpu/model/fuzztest with the error: "ERROR: /xla/service/gpu/model/fuzztest/BUILD:11:28: no such package 'tools': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - tools and referenced by '//xla/service/gpu/model/fuzztest:affine_grammar_source'
ERROR: Analysis of target '//xla/service/gpu/model/fuzztest:affine_grammar_source' failed; build aborted: Analysis failed"

2. Excluding the dir //xla/service/gpu/model/fuzztest/... from build makes the build pass for CPU.

3. Only following test fails if we include //xla/service/gpu/... in bazel test command for CPU backend (without actual GPU) and without GPU config, i.e. executing ./configure --backend CPU before bazel test command: 
  3.1 //xla/service/gpu:triton_support_test 81 tests passed, 2 skipped, 19 failed

akuegel:
> With the latest changes, I did some experiments for this and following are the findings:
> 
> 1. including //xla/service/gpu/... in CPU build job fails for this dir: /xla/service/gpu/model/fuzztest with the error: "ERROR: /xla/service/gpu/model/fuzztest/BUILD:11:28: no such package 'tools': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
> 
> * tools and referenced by '//xla/service/gpu/model/fuzztest:affine_grammar_source'
>   ERROR: Analysis of target '//xla/service/gpu/model/fuzztest:affine_grammar_source' failed; build aborted: Analysis failed"
> 
> 2. Excluding the dir //xla/service/gpu/model/fuzztest/... from build makes the build pass for CPU.
> 3. Only following test fails if we include //xla/service/gpu/... in bazel test command for CPU backend (without actual GPU) and without GPU config, i.e. executing ./configure --backend CPU before bazel test command:
>    3.1 //xla/service/gpu:triton_support_test 81 tests passed, 2 skipped, 19 failed

The fuzztest directory was added recently in https://github.com/openxla/xla/commit/f20aa9d3e801fba49a24d33af2eaa69e2de636b7
It seems we need to do some followup work to avoid a compile error on CPU.

Regarding triton_support_test, this is somewhat expected as I have stubbed out triton_ir_emitter in the build config without GPU configured. Maybe we can make it work, I was planning to refactor the build targets a bit.

akuegel:
I think I found the problem, the bazel macro that we are importing references the tools directory without project prefix, so bazel will try to find it inside the XLA project and fails. I have sent out a fix for google_fuzztest project, once that is merged, we can update XLA to use the new fixed version.

hmonishN:
Is there a PR for the fix? If its internal, can you let me know once that is merged.

akuegel:
> Is there a PR for the fix? If its internal, can you let me know once that is merged.

It was an internal change, essentially we need to change the config for the service that imports and exports to github so that it will map to the right dependency (with `@google_com_fuzztest` prefix).
After this is updated, we will have to update the XLA pin for google_fuzztest to use the new revision. I will let you know once that is done.

akuegel:
So the fuzztest repo has a fix now, but my attempt to update to the new revision is blocked, because fuzztest now also depends on Riegeli. I tried to disable this dependency, but then a Tensorflow presubmit failed that checks that blaze query works (which it will not because it will go into all selects, so even though the Riegeli dependency is "guarded", it will try to fetch it). https://github.com/openxla/xla/pull/15060 has the attempt.
We will now remove this fuzztest directory from open source.

hmonishN:
After removal of fuzztest dir the build passes now for CPU. The only test that is failing is //xla/service/gpu/fusions/triton:triton_support_test

hmonishN:
With the latest fixes https://github.com/openxla/xla/commit/6c5d5686515f4ea2e505b9f7dd85cbde78f51b4f the tests are enabled now as part of XLA CPU jobs. Closing this PR

PR #14780, opened 2024-07-11
[XLA:GPU][AutoSharding] Replace hashmap with btree map to determine the iteration order

### LINKED ISSUE #7248: Support autosharding in OSS
In order to support autosharding in OSS, we need to fix the non-determinism issues from using `absl::flat_hash_map`. The iteration order for `StableHashMap` in `auto_sharding_strategy.h` needs to become deterministic (internally it follows the insertion order, backed by linked hash map).

lausannel:
This PR addresses one of the problems mentioned in #7248, regarding the non-deterministic iteration order of `StableHashMap`.

To resolve this, this PR proposes the following changes:

- Replaced `::absl::flat_hash_map` with `::absl::btree_map` within the `StableHashMap` implementation.
- Added new tests to validate the deterministic iteration order of the `StableHashMap`.

### DIFF
--- a/xla/hlo/experimental/auto_sharding/auto_sharding_strategy.h
+++ b/xla/hlo/experimental/auto_sharding/auto_sharding_strategy.h
@@ -32,6 +32,7 @@ limitations under the License.
 #include "xla/service/hlo_value.h"
 #include "xla/shape_util.h"
 #include "absl/container/flat_hash_map.h"
+#include "absl/container/btree_map.h"
 #include "absl/container/flat_hash_set.h"
 
 namespace xla {
@@ -42,7 +43,7 @@ constexpr double kInfinityCost = 1e20;
 
 // Type alias
 template <typename Key, typename Value>
-using StableHashMap = ::absl::flat_hash_map<Key, Value>;
+using StableHashMap = ::absl::btree_map<Key, Value>;
 template <typename Key>
 using StableHashSet = ::absl::flat_hash_set<Key>;
 
--- a/xla/hlo/experimental/auto_sharding/auto_sharding_solver_test.cc
+++ b/xla/hlo/experimental/auto_sharding/auto_sharding_solver_test.cc
@@ -846,6 +846,20 @@ TEST(ScaleRequest, SkipsScaling) {
   EXPECT_THAT(request, ::testing::EqualsProto(expected_request));
 }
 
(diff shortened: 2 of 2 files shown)


xla/hlo/experimental/auto_sharding/auto_sharding_strategy.h
```diff
@@ -42,7 +43,7 @@ constexpr double kInfinityCost = 1e20;
 
 // Type alias
 template <typename Key, typename Value>
-using StableHashMap = ::absl::flat_hash_map<Key, Value>;
+using StableHashMap = ::absl::btree_map<Key, Value>;
```
dimitar-asenov, line 46: Please add a test that checks that `StableHashMap` now indeed has deterministic order. E..g a test where you insert 10 pairs in a random order and expect to read them in a known order.

dimitar-asenov (changes requested):

dimitar-asenov:
The description says that this fixes https://github.com/openxla/xla/issues/7248, but looking at that issue, it seems that there is more that needs to be done to fix it. Can you please explain if this PR really fixes the issue? If not, then please adjust the description.

xla/hlo/experimental/auto_sharding/auto_sharding_strategy.h
cheshire: TBH I think this is fine, we never had flat_hash_map here before (it's different internally), and not everything is deterministic yet, so this fix clearly makes it incrementally better.

cheshire:
> The description says that this fixes https://github.com/openxla/xla/issues/7248,

Agreed with the comment, would not mention the bug as "fixed".

xla/hlo/experimental/auto_sharding/auto_sharding_strategy.h
dimitar-asenov: I'm not opposed to the solution, just want to have a test that clearly states that whatever `StableHashMap` is defined to be, it should be deterministic.

lausannel:
@cheshire @dimitar-asenov Sorry for the lack of clarity. Thank you for your review comments. I have updated the PR description and added the corresponding tests. Could you take another look?

NaiyerRizz (approved):

dimitar-asenov (approved):

PR #14795, opened 2024-07-11
[GPU] Add methods for deterministic backend config fingerprinting.

sergachev:
Backend config being a proto can contain maps, which are not printed in deterministic order by default. The new methods allow printing a deterministic fingerprint of a backend config, which is implemented as an encoded string of a deterministic binary serialization of the backend config proto.


xla/hlo/ir/backend_config.cc
```diff
@@ -57,6 +60,13 @@ const std::string& BackendConfigWrapper::GetRawStringWithoutMutex() const {
   return raw_string_.empty() ? *kEmptyString : raw_string_;
 }
 
+absl::StatusOr<std::string> BackendConfigWrapper::GetFingerprint() const {
+  absl::MutexLock lock{&mutex_};
+  std::string result;
+  TF_RET_CHECK(tsl::SerializeToStringDeterministic(*proto_, &result));
```
dimitar-asenov, line 66: This may be `null`. The backend config is initialized with either one of `proto` or `raw_string` and the other one is not set unless you call a getter method for it.

To fix:
- Remove the locking of the mutex above and directly call the `GetProto()` method.
- Add a test that also tests the path where the backend config is created from a string, not a proto.

xla/hlo/ir/backend_config.cc
```diff
@@ -57,6 +60,13 @@ const std::string& BackendConfigWrapper::GetRawStringWithoutMutex() const {
...
+  return absl::BytesToHexString(result);
```
dimitar-asenov, line 67: BytesToHexString is not just a fingerprint (hash), it's a full, reversible encoding of the original data.

Can you explain what the use-case is? You might as well just return the result from `SerializeToStringDeterministic` or event the proto directly, at which point I'm not even sure we need this method.

dimitar-asenov (changes requested):

xla/hlo/ir/backend_config.cc
sergachev: How does one call GetProto(&proto) for an unknown proto? (tsl::protobuf::Message is abstract)

xla/hlo/ir/backend_config.cc
sergachev: >BytesToHexString is not just a fingerprint (hash), it's a full, reversible encoding of the original data.

A fingerprint does not have to be a hash. The current one may be a little more practical if someone wanted to load it back from an HLO dump.

> Can you explain what the use-case is?

We need structures like map<hlo_instruction, compiled_binary> where the binary also depends on the instruction's backend config.

> You might as well just return the result from SerializeToStringDeterministic

Yes, though that would be harder to use in debugging, because that contains non-printable binary sequences.
The current output of GetFingerprint() can for example be used in HLO dumps or in logging.

>or event the proto directly

What do you mean? We need some deterministic string representation which can be attached to the string representation of everything else in the HLO instruction.

xla/hlo/ir/backend_config.cc
sergachev: For our purpose fingerprinting of unknown protos created from strings is not required, I can just add a proto_ != nullptr check and a test if that's OK.

dimitar-asenov: This will be brittle. If we make changes to `BackendConfigWrapper` they should work in all cases. See my other comment.

xla/hlo/ir/backend_config.cc
dimitar-asenov: The HLO instruction can already print the JSON version of the backend config. This can be used for a full-round trip if necessary. From the text above I don't get the feeling that such roundtripping is essential to your use case.

Calling the current solution in this PR a "fingerprint" is misleading, because a fingerprint is commonly understood to be a just an "identifier", not a complete encoding of the full data. If we proceed with this PR as is, the HLO instruction dump would have two different ways to express the entire backend config, which is not what we want.

Is the main missing piece here the determinism of the encoding? Is the JSON encoding non-deterministic? If so, then we should rather find a way to make it deterministic, rather than introducing an extra encoding.

sergachev: >a fingerprint is commonly understood to be a just an "identifier"

See https://github.com/openxla/xla/blob/a2e537951445a1d558ddc31a3ae33c829c40c52c/xla/service/gpu/kernel_reuse_cache.cc#L70 for counterexample. Though I don't mind using a hash or renaming the method.

> the HLO instruction dump would have two different ways to express the entire backend config

There are many ways to print HLO: https://github.com/openxla/xla/blob/d8f3bc8212c6074c716c5fc57cbb918393ac92b8/xla/hlo/ir/hlo_instruction.h#L92

> Is the main missing piece here the determinism of the encoding? Is the JSON encoding non-deterministic?

Yes, see the description of the PR.

> we should rather find a way to make it deterministic

That would be nice, do you see a reasonable way to do so?

sergachev:
As discussed, I switched to WebSafeBase64Escape encoding, moved the code out of BackendConfigWrapper and templated the functions by proto kind.

xla/service/gpu/ir_emission_utils_test.cc
```diff
@@ -1032,5 +1035,45 @@ ENTRY main {
               IsOkAndHolds(true));
 }
 
+TEST_F(IrEmissionUtilsTest, ProtoFingerprintIsDeterministic) {
+  gpu::GpuBackendConfig proto;
+  auto& knobs = *proto.mutable_cudnn_fmha_backend_config()
+                     ->mutable_algorithm()
+                     ->mutable_tuning_knobs();
+  knobs[0] = 1;
```
dimitar-asenov, line 1043: Is testing just two knobs enough to establish determinism? Even with a non-deterministic implementation this will still be correct 50% of the time. Consider adding more, e.g. 10 or so.

xla/service/gpu/ir_emission_utils_test.cc
```diff
@@ -1032,5 +1035,45 @@ ENTRY main {
...
+  TF_ASSERT_OK_AND_ASSIGN(std::string fingerprint, GetProtoFingerprint(proto));
+  EXPECT_EQ(fingerprint, "Sg5CDCIECAAQASIECAIQAw");
+}
+
+TEST_F(IrEmissionUtilsTest, BackendConfigFingerprintIsDeterministic) {
+  gpu::GpuBackendConfig proto;
+  auto& knobs = *proto.mutable_cudnn_fmha_backend_config()
+                     ->mutable_algorithm()
+                     ->mutable_tuning_knobs();
+  knobs[0] = 1;
```
dimitar-asenov, line 1054: Same comment as above.

xla/service/gpu/ir_emission_utils.h
```diff
@@ -238,6 +239,31 @@ class DenseDataIntermediate {
...
+// Returns a deterministic encoded string representation of the backend config.
+template <typename ConfigType>
+absl::StatusOr<std::string> GetBackendConfigFingerprint(
+    const BackendConfigWrapper& wrapper) {
+  ConfigType proto;
+  TF_RETURN_IF_ERROR(wrapper.GetProto(&proto));
+  return GetProtoFingerprint(proto);
+}
+
+// Returs concatenated fingerprint of an HLO instruction without its backend
```
dimitar-asenov, line 255: `Returns` (missing n).

xla/service/gpu/ir_emission_utils_test.cc
```diff
@@ -1032,5 +1035,45 @@ ENTRY main {
...
+      GetBackendConfigFingerprint<GpuBackendConfig>(wrapper));
+  EXPECT_EQ(fingerprint, "Sg5CDCIECAAQASIECAIQAw");
+}
+
+TEST_F(IrEmissionUtilsTest,
+       InstructionFingerprintWithBackendConfigIsDeterministic) {
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnVerifiedModule(R"(
+ENTRY e {
+  ROOT _ = u8[0] custom-call(), custom_call_target="", backend_config={"cudnn_fmha_backend_config": {"algorithm": {"tuning_knobs": {"0": "1", "2": "3"}}}}
```
dimitar-asenov, line 1068: Same comment as above.

dimitar-asenov (changes requested):

xla/service/gpu/ir_emission_utils_test.cc
sergachev: done

xla/service/gpu/ir_emission_utils.h
sergachev: done

xla/service/gpu/ir_emission_utils_test.cc
sergachev: done

dimitar-asenov (approved):

PR #14829, opened 2024-07-11
[NVIDIA GPU] Flatten call graph after collective pipeliner pass

Tixxx:
After moving the collective pipeliner pass after post layout optimization passes, we need to call flatten call graph after pipeliners have run so call sites are unique. We didnt need to do this before because layout assignment(which is after previous location of collective pipeline) runs flattenCallGraph.

### DIFF
--- a/xla/tests/collective_ops_e2e_test.cc
+++ b/xla/tests/collective_ops_e2e_test.cc
@@ -981,5 +981,98 @@ ENTRY entry {
   EXPECT_THAT(gemm_op, NotNull());
   EXPECT_EQ(gemm_op->custom_call_target(), "__cublas$lt$matmul$f8");
 }
+
+TEST_F(CollectiveOpsTestE2E, PostLayoutCollectivePipelinerShouldFlattenCallGraph) {
+// The allgather in the loop has a nested while loop as its operand,
+// when the pipelining happens, the nested while loop will be peeled outside.
+// However, when a while is cloned, its callsites are still preserved which will
+// error out in alias analysis. When the graph is flattened, the error should not
+// happen.
+  absl::string_view kModuleReplicatedStr = R"(
+HloModule module
+
+while_cond {
+  param = (s32[], f32[2,128], f32[8,128], f32[8,128]) parameter(0)
+  gte = s32[] get-tuple-element(param), index=0
+  constant.1 = s32[] constant(3)
+  ROOT cmp = pred[] compare(gte, constant.1), direction=LT
+}
+
+while_nested_cond {
+  param.nested = (s32[], f32[2,128]) parameter(0)
+  gte.nested = s32[] get-tuple-element(param.nested), index=0
+  constant.nested = s32[] constant(3)
+  ROOT cmp.nested = pred[] compare(gte.nested, constant.nested), direction=LT
+}
+while_nested_body {
+  param.body_nested = (s32[], f32[2,128]) parameter(0)
+  gte.body_nested = s32[] get-tuple-element(param.body_nested), index=0
+  gte.2.body_nested = f32[2,128] get-tuple-element(param.body_nested), index=1
+
+  constant.body_nested = s32[] constant(1)
+  add.body_nested = s32[] add(gte.body_nested, constant.body_nested)
+  rsqrt.body_nested = f32[2,128] rsqrt(gte.2.body_nested)
+  ROOT tuple.body_nested = (s32[], f32[2,128]) tuple(add.body_nested, rsqrt.body_nested)
+}
+
+while_body {
+  param = (s32[], f32[2,128], f32[8,128], f32[8,128]) parameter(0)
+  get-tuple-element.394 = s32[] get-tuple-element(param), index=0
+  get-tuple-element.395 = f32[2,128] get-tuple-element(param), index=1
+  get-tuple-element.35 = f32[8,128] get-tuple-element(param), index=2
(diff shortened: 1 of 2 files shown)


frgossen (changes requested):
Can you add a test case to help understand why this breaks w/o extra flattening?

Tixxx:
> Why do we have two paths for pipelining the collectives anyway?
> 
> It seems to me this post-layout path was introduced on ["let's see if it breaks anything"](https://github.com/openxla/xla/pull/12866#issuecomment-2138240159) basis. Now, that we see there are some flaws to this, in my mind it makes sense to disable this by default.
> 
> Also please add a test for the case you provide a fix.

I think the plan was to only keep the pipeliner passes after all the post-layout optimizations have run. Pipeliners have a few side effects on some of the gemm fusions we have in post-layout optimizations, so we'd like to run them after target-specific fusions have happened which will result in simpler graphs. This is a bug that we didn't have full test coverage before, since there's no fundamental flaw of moving the pipeline, I think it'd be good to keep the original plan? We are still tracking the removal of the original pipeliners on our side.
I'm working on minimizing the model to have an e2e test for this case.

golechwierowicz:
> This is a bug that we didn't have full test coverage before, since there's no fundamental flaw of moving the pipeline, I think it'd be good to keep the original plan?

Fundamentally the plan sounds good, however I've seen one instance of RS not being pipelined recently (I will share the HLO soon-ish) so this is why I have some doubts, mostly how freely we can move passes around given the robustness of collective pipeliner.

I think ultimately we don't want to see additional patches to make some specific cases work in collective pipeliner as it is already quite complex.

Tixxx:
> > This is a bug that we didn't have full test coverage before, since there's no fundamental flaw of moving the pipeline, I think it'd be good to keep the original plan?
> 
> Fundamentally the plan sounds good, however I've seen one instance of RS not being pipelined recently (I will share the HLO soon-ish) so this is why I have some doubts, mostly how freely we can move passes around given the robustness of collective pipeliner.
> 
> I think ultimately we don't want to see additional patches to make some specific cases work in collective pipeliner as it is already quite complex.

I understand the concern. We are kind of choosing the less of 2 evils.
If we keep the pipeliner where it is, then we'd need a lot more logic in gemm rewriter to deal with pipelined QDQ nodes(which potentially could cause more bugs);
If we move pipeliner passes, we need to fix bugs like this before it stabilizes.
So given the concerns from your side, would keeping the 2 callsites of collective pipeliner and defaulting to the original callsite be a better option going forward?

We have noticed some issues with RS on our side too, some investigations suggest that we likely dont need to change pipeliner. Some transformations in gemm rewriter have side effects which can be avoided. Once you have the hlo, we can confirm if it's the same root cause.

Tixxx:
@frgossen @golechwierowicz  added an e2e test that repros the crash.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -981,5 +981,98 @@ ENTRY entry {
...
+TEST_F(CollectiveOpsTestE2E, PostLayoutCollectivePipelinerShouldFlattenCallGraph) {
+// The allgather in the loop has a nested while loop as its operand,
+// when the pipelining happens, the nested while loop will be peeled outside.
+// However, when a while is cloned, its callsites are still preserved which will
+// error out in alias analysis. When the graph is flattened, the error should not
+// happen.
+  absl::string_view kModuleReplicatedStr = R"(
+HloModule module
+
+while_cond {
```
frgossen, line 994: Is this a minimal test case reproducing teh issue?

frgossen (changes requested):

xla/tests/collective_ops_e2e_test.cc
Tixxx: Yes, we need the following conditions to repro the crash:
1. nested while loop
2. inner while loop being an operand of a pipelined collective which will be peeled out of the outer loop
The test just contains the hlo to meet the above conditions.

Tixxx:
> > So given the concerns from your side, would keeping the 2 callsites of collective pipeliner and defaulting to the original callsite be a better option going forward?
> 
> I'd say we need to align on this a little bit more, pipelines are kind of a mess right now and it would be good to not add additional complexity there. Maybe I will write a small doc on how I would see the future of pipelines related to horizontal scaling (pipeliners, collective rewriters etc.) so we can get a consensus.
> 
> For the time being let's apply your suggestion of defaulting to the old callsite.
> 
> Thanks for your input!

That sounds good. IMO collective optimizations ideally should happen after "local" optimizations i.e gemm fusions or kernel fusions, but that also means the current collective passes would need to handle additional ops like bitcasts or converts from float normalizations which adds more complexity. Having a doc to brainstorm sounds great to me.

golechwierowicz:
> > For the time being let's apply your suggestion of defaulting to the old callsite.
> 
> @golechwierowicz Can we put this in a separate pr?

Ok, let's put it in a separate PR.

golechwierowicz (approved):

Tixxx:
@golechwierowicz @frgossen  can you guys share more info on the revert? thanks.

golechwierowicz:
> @golechwierowicz @frgossen can you guys share more info on the revert? thanks.

The change is in after all https://github.com/openxla/xla/commit/045256bbef5ab4fea5a695efb37dd907d49c7129.

frgossen:
The issue was an unused variable, kNumPartitions

PR #14881, opened 2024-07-12
[GPU] Remove DRAM size from model_str in device description.

sergachev:
Exact DRAM size can slightly vary for the same GPU model so it shouldn't be used in model identification. The removed part of gpu_compiler_test_autotune_db.textproto illustrates that.

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -131,7 +131,7 @@ example looking like:
 ```
 version: 2
 results {
-  device: "sm_8.0 with 42331013120B RAM, 108 cores, 1410000KHz clock, 1215000KHz mem clock, 41943040B L2$"
+  device: "sm_8.0 with 108 cores, 1410000KHz clock, 1215000KHz mem clock, 41943040B L2$"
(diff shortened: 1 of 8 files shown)


derdrdirk (approved):

dimitar-asenov:
I just saw this PR. We're aware of the problem and I'm actually working on a different solution that also excludes the RAM, but also makes other changes to the cache string. It should be ready today.

dimitar-asenov:
FYI: the alternative PR is ready, but we need to wait until Thursday to submit it. Conceptually, it's very similar to this one.

dimitar-asenov:
Here is the current version of the other PR: https://github.com/openxla/xla/pull/14954

> Then what's the reason to prefer it?

The other PR doesn't change the model_str in cuda executor. It touches only the autotuning key and does a cleanup related to deviceless autotuning keys. It also does changes that are needed internally (and not visible on GitHub). At the point I became aware of your PR we already had the internal one nearly done. It's now fully ready but we have to wait for an extra internal approval until tomorrow.

PR #14899, opened 2024-07-15
Add TF_ prefix to ASSERT_OK in tests

jaro-sevcik:
Fixes OSS build breakage of `se_gpu_pjrt_client_test.cc`.


jaro-sevcik:
The `se_gpu_pjrt_client_test.cc` file used to build just fine before https://github.com/openxla/xla/commit/db472b8c3d83bc27b3c67067802b9a37ef7542e3. Let us keep it buildable in OSS. (Not sure why we don't build the file in OSS by default.)

thomasjoerg (approved):

PR #14956, opened 2024-07-16
[ROCM][NFC] gemm_algorithm_picker/test: refactoring in preparation for precision settings integration

pemeliya:
These is refactoring extracted from the original PR: https://github.com/openxla/xla/pull/13425

I have added **return_algo_index** parameter to GetBestAlgorithm function in order to avoid double conversion: algorithm -> index and then index -> algorithm. This part was quite misleading and hard to follow in the code as to why this was necessary.
The reason for this is because, cu/hipblasLt return an **opaque object** as an algorithm ID which cannot be saved within an HLO. Therefore, we needed an index of the algorithm here. In contrast, cu/hipblas return a 32-bit int algorithm ID which can readily be stoded in HLO (gemm_backend_config). 

Besides, I also simplified the respective test by unifiying some copy-paste stuff.

### DIFF
--- a/xla/service/gpu/gemm_algorithm_picker.cc
+++ b/xla/service/gpu/gemm_algorithm_picker.cc
@@ -345,8 +345,12 @@ class GemmAutotuner {
     absl::StatusOr<AutotuneResult> best =
         PickBestResult(results, gemm->ToString(), hlo_module_config);
     if (best.ok()) {
-      // Return a real algorithm ID if return_algo_index is false: 
-      // e.g., in case of legacy cublas tuning.
+      // Note that, cublas-lt returns an opaque object as an algorithm ID,
+      // therefore we need to convert it to the index from the list algorithms' 
+      // list (otherwise, we cannot store this ID inside a gemm_backend_config).
+      // In contrast, legacy cublas returns a 32-bit integer algorithm ID which
+      // can be readily stored inside an HLO (hence return_algo_index is false
+      // for cublas case).
       if (!return_algo_index) return best; 
       // Otherwise, map a real algorithm ID to its index among the results.
       for (size_t i = 0; i < results.size(); ++i) {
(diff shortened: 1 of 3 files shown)


xla/service/gpu/gemm_algorithm_picker_test.cc
```diff
@@ -60,15 +61,32 @@ class GemmAlgorithmPickerTest : public HloTestBase,
...
 
-    if (auto* procm = std::get_if<se::RocmComputeCapability>(&gpu_cc)) {
-      if (GetDebugOptionsForTest().xla_gpu_enable_cublaslt() &&
-          !procm->has_hipblaslt()) {
-        GTEST_SKIP() << "No gpublas-lt support on this architecture!";
-      }
-    }
+  void SetUp() override {
+    std::visit(VariantVisitor{
+      [](const se::CudaComputeCapability& cc) {
```
dimitar-asenov, line 76: Putting this in `SetUp` will cause all tests to be skipped on Ampere+. However, currently the `BlasGetVersion` test is not skipped for `Ampere`. Please make sure to preserve the original behavior.

xla/service/gpu/gemm_algorithm_picker.cc
```diff
@@ -352,6 +345,10 @@ class GemmAutotuner {
     absl::StatusOr<AutotuneResult> best =
         PickBestResult(results, gemm->ToString(), hlo_module_config);
     if (best.ok()) {
+      // Return a real algorithm ID if return_algo_index is false: 
+      // e.g., in case of legacy cublas tuning.
+      if (!return_algo_index) return best;
```
dimitar-asenov, line 350: Please provide additional comments here to explain why this is necessary. You can copy the explanation form the PR description, it's good to have this directly in the code.

dimitar-asenov (changes requested):

dimitar-asenov (approved):

PR #13108, opened 2024-05-28
[GPU] Shard GEMM fusion autotuning across multiple compilation processes.

sergachev:
This solves the problem of compiler processes compiling different programs due to fluctuations of measurements during autotuning and in addition makes some cases faster (depending on the number of fusions and the availability of CPU cores).

PR #14968, opened 2024-07-16
[GPU] Enable sharding of autotuning by default.

sergachev:
Requires https://github.com/openxla/xla/pull/14881

@PatriosTheGreat

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -273,7 +273,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
 
   opts.set_xla_use_shardonnay(false);
 
-  opts.set_xla_gpu_shard_autotuning(false);
+  opts.set_xla_gpu_shard_autotuning(true);
 
   opts.set_xla_gpu_per_fusion_autotune_cache_dir("");
 


thomasjoerg:
> Requires https://github.com/openxla/xla/pull/14881

@dimitar-asenov you said you are working on a different solution on this PR. Hence, bringing this one to your attention and adding you as reviewer.

dimitar-asenov:
> > Requires #14881
> 
> @dimitar-asenov you said you are working on a different solution on this PR. Hence, bringing this one to your attention and adding you as reviewer.

Thanks, Thomas. I believe https://github.com/openxla/xla/pull/14881 is just a bug fix that's needed here. I think the PR here should be reviewed independently.

frgossen (changes requested):
Can you provide a bit more context? Why was this not enabled, why do you like to enable this now?

sergachev:
See https://github.com/openxla/xla/pull/13108 for the implementation and the explanation what problem does it solve. It was not enabled so far because of the usual development scheme - commit under a flag, test, flip the flag. It has been discussed for a while, tested at NVIDIA and by @PatriosTheGreat.

dimitar-asenov (approved):

dimitar-asenov:
LGTM.

If someone is interested, this is tested here:
https://github.com/openxla/xla/blob/main/xla/tools/multihost_hlo_runner/functional_hlo_runner_test.cc#L270

dimitar-asenov:
We had to roll this back, because there were internal use cases where models would freeze (or take more than an hour longer than usual, but we didn't wait to see).

The specific line where we observed the issue is this:

https://github.com/openxla/xla/blob/main/xla/service/gpu/gemm_fusion_autotuner.cc#L1187

Calling `key_value_store.Get` inside `ExchangeResults` seems to indeed wait for a very long time when `absl::InfiniteDuration()` is provided.

I'm asking the team that observed this for a more concrete reproducer and will post here once we have more info.

PR #15027, opened 2024-07-17
Add Gloo support for MacOS.

heiner:
This is an alternative to https://github.com/openxla/xla/pull/7726.

Gloo supports MacOS, but requires using libuv as the transport mechanism.


third_party/gloo/gloo.BUILD
```diff
@@ -22,7 +22,7 @@ substitions = {
     "#cmakedefine01 GLOO_USE_REDIS": "#define GLOO_USE_REDIS 0",
     "#cmakedefine01 GLOO_USE_IBVERBS": "#define GLOO_USE_IBVERBS 0",
     "#cmakedefine01 GLOO_USE_MPI": "#define GLOO_USE_MPI 0",
-    "#cmakedefine01 GLOO_USE_LIBUV": "#define GLOO_USE_LIBUV 0",
+    "#cmakedefine01 GLOO_USE_LIBUV": "#define GLOO_USE_LIBUV (__APPLE__ ? 1 : )",
```
penpornk, line 25: Should we keep the original `0`?
```suggestion
    "#cmakedefine01 GLOO_USE_LIBUV": "#define GLOO_USE_LIBUV (__APPLE__ ? 1 : 0)",
```
Alternatively, would `"#define GLOO_USE_LIBUV __APPLE__"` have the same effect?

third_party/uv/uv.BUILD
```diff
@@ -0,0 +1,29 @@
+# Description:
+#   libuv
```
penpornk, line 2: Nit: Maybe add more details?
```suggestion
#   libuv is a cross-platform asynchronous I/O library.
```

third_party/uv/BUILD
```diff
@@ -0,0 +1 @@
+# libuv.
```
penpornk, line 1: Nit: Please use:
```suggestion
# copybara:uncomment package(default_applicable_licenses = ["//tensorflow:license"])
```

third_party/uv/workspace.bzl
```diff
@@ -0,0 +1,17 @@
+"""Provides the repository macro to import libuv."""
+
+load("//third_party:repo.bzl", "tf_http_archive", "tf_mirror_urls")
+
+def repo():
+    """Imports libuv."""
+
+    UV_VERSION = "v1.38.0"
```
penpornk, line 8: Is there a reason we are using v1.38.0 and not the latest version (v1.48.0)?

xla/python/xla.cc
```diff
@@ -274,9 +278,26 @@ NB_MODULE(xla_extension, m_nb) {
...
+        if (interface) {
+          uv_attrs.iface = *interface;
+        }
+        auto uv_device = gloo::transport::uv::CreateDevice(uv_attrs);
+        return std::make_shared<cpu::GlooCollectives>(std::move(gloo_kv_store),
+                                                      std::move(uv_device));
 #else   // __linux__
         throw xla::XlaRuntimeError(
-            "make_gloo_tcp_collectives only implemented for linux");
+            "make_gloo_tcp_collectives only implemented for linux and macos");
```
penpornk, line 300: We probably should rename the method to just `make_gloo_collectives`. But I'll handle this later since this involves changing JAX as well.

penpornk (changes requested):
Thank you for the PR!

heiner (commented):
> Hi @heiner
> Can you please sign CLA
> Thank you

Should be signed.

image

third_party/gloo/gloo.BUILD
heiner: Thanks! Leaving out the `0` was an oversight.

Re: `#define GLOO_USE_LIBUV __APPLE__` I'm not sure. I think `__APPLE__` is `1` when it is defined, but it's probably not `0` otherwise.

The other question is whether any of this is strictly necessary, running `git grep USE_LIBUV` in https://github.com/facebookincubator/gloo shows hits only in cmake files and `gloo/config.h.in`, but none in actual code.

third_party/uv/workspace.bzl
heiner: I chose the version used in the tests in Gloo: https://github.com/facebookincubator/gloo/blob/main/.circleci/config.yml#L36

It's likely newer versions will work just as well, happy to update.

heiner:
> Thank you for the PR!

Thanks for the quick turnaround time!

xla/python/xla.cc
heiner: Sounds good! Note that if the `libuv` dependency is okay to add (I wasn't sure about that, might be tricky in `google3`?), it's possible to use it for Linux too which would allow merging the code paths.

third_party/gloo/gloo.BUILD
penpornk: Got it. Thank you for investigating!

third_party/uv/workspace.bzl
penpornk: I see. Agreed on matching their version.

xla/python/xla.cc
penpornk: Yes, `libuv` is okay for google3. On the open source side, I think it's fine to add one more dependency. Cc'ing @ddunl just in case.

penpornk (approved):
Thank you for the quick responses!

xla/python/xla.cc
heiner: Excellent! In that case, the code in this file can probably be simplified to look more more or less like they did after https://github.com/openxla/xla/pull/7674, except for using the `uv` protocol instead of the `tcp` one.

ddunl:
Extra dependency looks fine to me: FYI for whoever lands this (maybe me tomorrow 🙂 ) it's probably easiest to submit a change first with changes to `tensorflow/third_party` + the associated `workspace.bzl` update in TF. The `workspace.bzl` change has to happen anyway to land this, and then Copybara won't choke on this change.

ddunl (approved):

mmanzoorTT:
@heiner I am trying building your branch on MacOS with `bazel build --test_output=all --spawn_strategy=sandboxed //xla/...` and I am getting this error

heiner:
Hey @mmanzoorTT,

you must be doing something wrong. On MacOS, `gloo/transport/tcp` should not be used. Instead, `gloo/transport/uv` should be. My original PR did that via the right BUILD file settings. Perhaps something is slightly different in your code?

mmanzoorTT:
Hey @heiner thanks for prompt reply. I am running `./configure.py --backend=CPU` and then `bazel build --test_output=all --spawn_strategy=sandboxed //xla/...`. 
is there any other way to configure for MacOS?

heiner:
Let me try. The way I tested things is via

```
python build/build.py --configure_only --bazel_options=--override_repository=xla=~/src/upstream/xla/ --disable_cuda
```

from JAX.

heiner:
I can reproduce this. Will look for a fix shortly.

heiner:
@mmanzoorTT can you confirm your command works for you without my change? I believe it might try to compile the `gloo_collectives_test` target (in `xla/pjrt/cpu/BUILD`) despite the `tags = ["nomac"]`.

mmanzoorTT:
@heiner I cannot build on MacOS with or without your changes with the given commands. I got the same error in both cases. 
Currently, I am trying these hardcoded changes. 
```
diff --git a/third_party/gloo/gloo.BUILD b/third_party/gloo/gloo.BUILD
index dcd436d332..15a239e139 100644
--- a/third_party/gloo/gloo.BUILD
+++ b/third_party/gloo/gloo.BUILD
@@ -23,10 +23,10 @@ substitions = {
     "#cmakedefine01 GLOO_USE_IBVERBS": "#define GLOO_USE_IBVERBS 0",
     "#cmakedefine01 GLOO_USE_MPI": "#define GLOO_USE_MPI 0",
     "#cmakedefine01 GLOO_USE_LIBUV": "#define GLOO_USE_LIBUV (__APPLE__ ? 1 : 0)",
-    "#cmakedefine01 GLOO_HAVE_TRANSPORT_TCP": "#define GLOO_HAVE_TRANSPORT_TCP 1",
+    "#cmakedefine01 GLOO_HAVE_TRANSPORT_TCP": "#define GLOO_HAVE_TRANSPORT_TCP 0",
     "#cmakedefine01 GLOO_HAVE_TRANSPORT_TCP_TLS": "#define GLOO_HAVE_TRANSPORT_TCP_TLS 0",
     "#cmakedefine01 GLOO_HAVE_TRANSPORT_IBVERBS": "#define GLOO_HAVE_TRANSPORT_IBVERBS 0",
-    "#cmakedefine01 GLOO_HAVE_TRANSPORT_UV": "#define GLOO_HAVE_TRANSPORT_UV 0",
+    "#cmakedefine01 GLOO_HAVE_TRANSPORT_UV": "#define GLOO_HAVE_TRANSPORT_UV 1",
     "#cmakedefine01 GLOO_USE_AVX": "#define GLOO_USE_AVX __AVX__",
 }
 
diff --git a/third_party/uv/uv.BUILD b/third_party/uv/uv.BUILD
index 75a2df39c4..fc8588e7fa 100644
--- a/third_party/uv/uv.BUILD
+++ b/third_party/uv/uv.BUILD
@@ -14,6 +14,7 @@ cc_library(
     srcs = glob(["src/*.c"]),
     hdrs = [
         "include/uv.h",
+        "include/uv/*.h",
     ],
     copts = [
         "-fexceptions",
diff --git a/xla/pjrt/cpu/BUILD b/xla/pjrt/cpu/BUILD
index eff9c1f550..43b2c8609c 100644
--- a/xla/pjrt/cpu/BUILD
+++ b/xla/pjrt/cpu/BUILD
@@ -320,7 +320,7 @@ xla_cc_test(
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/time",
         "@com_google_absl//absl/types:span",
-        "@gloo//:transport_tcp",
+        "@gloo//:transport_uv",
         "@tsl//tsl/lib/core:status_test_util",
         "@tsl//tsl/platform:env",
         "@tsl//tsl/platform:errors",
```
It is not done building yet but it seems to be working. I'll update once build process complete.

heiner:
So the problem isn't with this PR. Instead, the command you use to build on MacOS isn't the right one. We could fix the `@gloo//:transport_tcp` issue by not relying on that at all (that would require further changes however, the diff you showed is insufficient). However, you are likely to run into other problems since there's more things tagged with `nomac`.

Instead, I'd advise you to reach out to XLA developers re: how to build XLA for MacOS.

ddunl:
It seems likely that it's an oversight in configure.py that we don't add `nomac` to the tags to avoid. I'll add that to my todo list (but also feel free to send a PR!)

ddunl:
SGTM, I'll get the requisite change that updates TensorFlow in today and then we can merge this PR and go from there. Thanks!!

ddunl:
Sorry, getting held up on some licensing issues internally. I'm hoping to merge this tomorrow. There's no material issue with the license of course, but convincing the internal systems of that is a battle....

ddunl:
Woops, realized I merged this without the extra target in the uv BUILD file. Sorry about that! Fixing that now.

ddunl:
This PR is resulting in build failures for JAX, wondering if you have encountered these? ``` File "<string>", line 1, in <module>
  File "/private/tmp/jax/jax/__init__.py", line 25, in <module>
    from jax._src.cloud_tpu_init import cloud_tpu_init as _cloud_tpu_init
  File "/private/tmp/jax/jax/_src/cloud_tpu_init.py", line 17, in <module>
    from jax._src import config
  File "/private/tmp/jax/jax/_src/config.py", line 27, in <module>
    from jax._src import lib
  File "/private/tmp/jax/jax/_src/lib/__init__.py", line 87, in <module>
    import jaxlib.xla_client as xla_client
  File "/Users/ddunleavy/.pyenv/versions/3.10.14/lib/python3.10/site-packages/jaxlib/xla_client.py", line 33, in <module>
    from . import xla_extension as _xla
ImportError: dlopen(/Users/ddunleavy/.pyenv/versions/3.10.14/lib/python3.10/site-packages/jaxlib/xla_extension.so, 0x0002): symbol not found in flat namespace '_uv_async_init'```

heiner:
Hey David - sorry for the late response, I was traveling / relocating last week.

I haven't seen that issue before, could you share the steps you took to produce it?

Judging from the error message it seems `uv` wasn't linked into the `xla_extension.so` dynamic library, which it should have.

ddunl:
No worries, my attempt at a fix was to have this change only take affect for x86 builds - but the nightly hasn't reran yet so I'm not sure if the fix is correct. Commit is here: https://github.com/openxla/xla/commit/f0ca502aec33ac423a3ddf585b51847ee0c6fba8

heiner:
It's very strange that this wouldn't work on ARM. In fact, that seems somewhat unlikely to me.

I'm probably not testing things properly (again). Could you share a quick repro; specifically:

- What you call to compile XLA,
- what you run to see this failure?

It may well be a bug in my original PR.

ddunl:
I was testing using the JAX nightly setup and was able to repro the original error message - I used their build.py to generate this configure.bazelrc:
```
build --strategy=Genrule=standalone
build --action_env CLANG_COMPILER_PATH="/usr/bin/clang"
build --repo_env CC="/usr/bin/clang"
build --repo_env BAZEL_COMPILER="/usr/bin/clang"
build --copt=-Wno-error=unused-command-line-argument
build --copt=-Wno-gnu-offsetof-extensions
build:rocm --action_env TF_ROCM_AMDGPU_TARGETS="gfx900,gfx906,gfx908,gfx90a,gfx1030"
build --config=mkl_open_source_only
```

And the command I ran was ```./build/bazel-6.5.0-darwin-arm64 run --verbose_failures=true --override_repository=xla=/tmp/xla //jaxlib/tools:build_wheel -- --output_path=/private/tmp/jax/dist --jaxlib_git_hash=b64b19f1e3782f7486447467a7982163636f72c2 --cpu=arm64```


Let me know if there's any other info I can provide, I rarely run into issues on Mac so it's possible I could've done something wrong while debugging this.

heiner:
I can reproduce this. Looking into it now.

heiner:
My apologies David - I didn't test the original PR properly.

https://github.com/openxla/xla/pull/15784 fixes this issue. I simply didn't include every required file into the libuv compilation.

Probably the right thing to do is to fix and enable the Gloo tests on MacOS so this is detected as part of the build/deployment.

ddunl:
unfortunately I had to revert as the TensorFlow MacOS nightly is still broken. Will work on getting instructions for how to repro that build. Sorry about this!

heiner:
My apologies for sending you something broken. I'll work on a PR with a proper test case.

ddunl:
No worries, I'm still trying to dig into it as well but wanted to get builds back to green in the meantime

ddunl:
I think the best way forward is when we rollforward, we just enable for either x86 or ARM first. And if possible have logs showing working bazel build on JAX ```build/build.py --bazel_options=--override_repository=xla=/tmpfs/src/xla --bazel_options=--config=macos_cache_push```

Check for devices with build wheel installed ```python -c 'import jax; print(jax.default_backend()); print(jax.devices()); print(len(jax.devices()))'``` 

And pytest ```python -m pytest -n auto --tb=short tests examples```

This is what the nightly does so if all that is passing then I have high confidence it will work. Then we can try to swtich on for the other architecture, that way the rollback isn't as big.

heiner:
SGTM.

Could you share how it failed in the nightly build in its most recent version?

heiner:
This change seems to work for me: https://github.com/openxla/xla/pull/16014.

PR #15050, opened 2024-07-18
Quantized Collectives

philipphack:
Introduces a pass that can reduce the amount of data transferred in all-gather, all-to-all, collective-broadcast and collective-permute ops by exchanging the collective with a subsequent quantization or conversion to a narrower type.


frgossen (approved):
Very nice indeed! Thank you for adding this!

PR #15081, opened 2024-07-18
Adds a convenient python binding for obtaining instruction costs.

yliu120:
The python binding being added in this PR can help quick analysis on the cost of the HLO instructions with easy python code.

The binding is tested with a real tensorboard dir:

```
import jax.lib.xla_bridge as xb
client = xb.xla_client
costs = client.profiler.get_instructions_profile(tensorboard_dir)
for name, cost in costs:
  print(name, cost)
```

works.


hawkinsp (approved):
I'm fine adding this as an experimental API with no particular support expectations. It's not part of JAX's supported API and might go away with zero notice if it gets in the way of improving the profiler.

PR #15149, opened 2024-07-19
[XLA:GPU] print cudnn frontend check_support error message

Cjkkkk:


xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -8385,10 +8385,7 @@ absl::StatusOr<bool> CudnnGraph::Prepare(dnn::DnnSupport& dnn_support) {
   RETURN_IF_CUDNN_FRONTEND_ERROR(graph_.build_operation_graph(cudnn->handle()));
   RETURN_IF_CUDNN_FRONTEND_ERROR(
       graph_.create_execution_plans({cudnn_frontend::HeurMode_t::A}));
-  if (auto result = graph_.check_support(cudnn->handle()); result.is_bad()) {
-    VLOG(3) << result.get_message();
-    return false;
-  }
+  RETURN_IF_CUDNN_FRONTEND_ERROR(graph_.check_support(cudnn->handle()));
```
akuegel, line 8388: I think this is not just a NFC, as now in case result.is_bad() returns true, we will return absl::InternalError vs just returning false as before. Can you please add a test case that covers this?

Cjkkkk: we have something like this for all calls to `Prepare`:
```
TF_ASSIGN_OR_RETURN(bool supported, cudnnGraph.Prepare(dnn_support));
  if (!supported) {
    return absl::InternalError("cuDNN graph is not supported.");
  }
```
So both will be captured but i think we don't even need to check `supported` anymore.

Cjkkkk: if `Prepare` returns false, it will just be captured by
```
if (!supported) {
    return absl::InternalError("cuDNN graph is not supported.");
  }
```
which still return `absl::InternalError` anyway but without the real error message. So this PR doesn't change the expected behavior. The false return is required before by cudnn fusion to probe cudnn fusion support but it is not required anymore with the current code. @sergachev

akuegel: Thanks, that makes sense.

akuegel (approved):

PR #15163, opened 2024-07-21
[NVIDIA GPU] Skip processing for trip count 1 in loop double buffer unrolling

Tixxx:
A minor improvement to loop double buffer unrolling pass to skip processing for loops with trip count =1


frgossen (changes requested):
Thanks for fixing. Can you and a tiny test case?

Tixxx:
> Thanks for fixing. Can you and a tiny test case?

Added a small test case for that

frgossen (approved):
Thank you!

PR #15216, opened 2024-07-23
Make GpuExecutor::HostMemoryAllocate NUMA aware

olupton:
This improves the achieved throughput of D2H transfers, for example when checkpointing. For example, there is a ~2x improvement in throughput of overlapped D2H copies from 8xH100 on a DGX node.

Notes:
- `TENSORFLOW_USE_NUMA` is set unconditionally instead of being hidden behind an option; it's not clear from OSS-world if this is an important handle for Google internally.
- `stream_executor::StreamExecutor::HostMemoryDeallocate` now takes the allocation size; all call sites updated. This is required by the `tsl::port::NUMAFree` API.

### DIFF
--- a/xla/stream_executor/cuda/cuda_executor.cc
+++ b/xla/stream_executor/cuda/cuda_executor.cc
@@ -165,6 +165,11 @@ absl::Status GpuExecutor::Init() {
   cc_major_ = desc->cuda_compute_capability().major;
   cc_minor_ = desc->cuda_compute_capability().minor;
   numa_node_ = desc->numa_node();
+  if (numa_node_ < 0) {
+    LOG(WARNING) << "NUMA node could not be determined for device "
+                 << device_ordinal_
+                 << ", host memory allocations will not be NUMA-pinned";
+  }
   return absl::OkStatus();
 }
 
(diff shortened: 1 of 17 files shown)


thomasjoerg:
I'm not the best person to review this and George is OOO.
@ezhulenev to assign reviewers.

xla/stream_executor/cuda/cuda_executor.cc
```diff
@@ -866,8 +911,24 @@ GpuExecutor::CreateDeviceDescription(int device_ordinal) {
     builder.set_pci_bus_id(pci_bus_id);
 
     // Read the NUMA node corresponding to the PCI bus ID out of sysfs.
-    int numa_node = TryToReadNumaNode(pci_bus_id, device_ordinal);
-    builder.set_numa_node(numa_node);
+    std::optional<int> numa_node =
+        TryToReadNumaNode(pci_bus_id, device_ordinal);
+    if (numa_node.has_value()) {
+      if (*numa_node < 0) {  // See http://b/18228951 for details on this path.
```
olupton, line 917: Can a Googler check if this bug is still relevant? This "fudge failure to a valid value" logic makes the test logic a bit cumbersome.

ezhulenev: This is a bug from 2014 and I think most of the involved people are no longer at google

klucke: -1 for numa_node is still relevant, it's a kernel definition for "no numa node"

olupton: Right, we should certainly handle that gracefully (it's even the case that the CI runners hit). My intent with this PR is that if you don't know the numa node, you just allocate without pinning.

The thing that's a little awkward is that -1 gets massaged into a 0 in the device description, which is lossy. That shows up in the test logic where you don't know anymore whether you expect to get an unpinned allocation or an allocation pinned to node 0.

klucke (approved):

thomasjoerg (changes requested):

xla/stream_executor/rocm/rocm_executor.cc
```diff
@@ -455,6 +455,20 @@ void GpuExecutor::Deallocate(DeviceMemoryBase* mem) {
...
+GpuExecutor::HostMemoryAllocate(uint64_t size) {
+  auto* buffer = GpuDriver::HostAllocate(context_, size);
+  if (buffer == nullptr && size > 0) {
+    return absl::InternalError(
+        absl::StrFormat("Failed to allocate HostMemory of size %d", size));
+  }
+  return std::make_unique<HostMemoryAllocation>(buffer, size, this);
+}
+
+void GpuExecutor::HostMemoryDeallocate(void* location, uint64_t size) {
```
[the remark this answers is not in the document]
olupton: I pushed a change that tries to smooth this over by keeping the two signatures in parallel. I am not sure how to test the impact in TensorFlow.

This will not allow all combinations to work, as I think that could only be achieved by adding an extra runtime data structure tracking {ptr: size}, but it should allow old implementations of the old interface to continue working.

olupton:
The clang-format check is complaining about a line I did not touch.
When I run it locally, it wants to make the same change.
In this case, should I push the suggested change?

xla/stream_executor/rocm/rocm_executor.cc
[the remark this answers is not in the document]
klucke: I will attempt to fix up this CL internally, and try and get it submitted today.

PR #15291, opened 2024-07-25
[NVIDIA GPU] Add Bitcast to collective pipeliner acceptable users

terryysun:
Right now collective pipeliner will filter user type when determining if a value can be pushed to the next iteration of the loop. Bitcast is not in the acceptable users, which causes the collective pipeliner not to kick in under some situations. This PR adds Bitcast to acceptable users.


golechwierowicz:
Hey,

on which model do you see a lack of this support leading to collective pipeliner not kicking in?

In general I think it is a safe change, but please add a test.

terryysun:
> Hey,
> 
> on which model do you see a lack of this support leading to collective pipeliner not kicking in?
> 
> In general I think it is a safe change, but please add a test.

It's an internal implementation of MaxText, the absence of Bitcast caused collective pipeliner not to kick in in backward pass. Added a test for a situation where this change is useful.

xla/service/collective_pipeliner_test.cc
```diff
@@ -181,6 +181,63 @@ ENTRY entry {
...
+
+while_cond {
+  param = (s32[], bf16[3,8,128], bf16[3,8,128]) parameter(0)
+  gte = s32[] get-tuple-element(param), index=0
+  constant.1 = s32[] constant(3)
+  ROOT cmp = pred[] compare(gte, constant.1), direction=LT
+}
+
+while_body {
+  param = (s32[], bf16[3,8,128], bf16[3,8,128]) parameter(0)
```
frgossen, line 202: Can you simplify this test case and give meaningful names to the values? 
It looks to me that there are a lot of ops in this that are not related to pipelining and are not relevant to the test case. 
I think a comment could also help understand what is tested here.

frgossen (changes requested):
Thank you for adding the test case :)  
I think we can simplify it a bit still

golechwierowicz:
frgossen@ PTL

frgossen:
> @frgossen PTL

I don't think the last comments were addressed yet

terryysun:
> > @frgossen PTL
> 
> I don't think the last comments were addressed yet

I've simplified the test HLO with some comments and renamed the variables, @frgossen could you comments on things you find unaddressed? Thanks!

frgossen:
Sorry, I was expecting some reply to the comments, even just a simple `Done`. Based on the commits alone, I find it hard to tell when the PR is ready for the next round of review. 

Thanks for simplifying the test case. Will take another look!

xla/service/collective_pipeliner_test.cc
```diff
@@ -181,6 +181,59 @@ ENTRY entry {
...
+  parameter.0 = bf16[3,8,128] get-tuple-element(param), index=1
+  parameter.1 = bf16[3,8,128] get-tuple-element(param), index=2
+  constant.1 = s32[] constant(1)
+  next-loop-index = s32[] add(current-loop-index, constant.1)
+  constant.0 = s32[] constant(0)
+  sliced-parameter.1 = bf16[1,8,128] dynamic-slice(parameter.1, current-loop-index, constant.0, constant.0), dynamic_slice_sizes={1,8,128}
+
+  all-reduce.0 = bf16[1,8,128] all-reduce(sliced-parameter.1), replica_groups={}, to_apply=add, channel_id=1
+  bitcasted-all-reduce.0 = u16[3,8,128] bitcast(all-reduce.0)
+  all-reduce.1 = bf16[3,8,128] bitcast(bitcasted-all-reduce.0)
```
frgossen, line 216: This is the one that depends on a bitcast and you want to see it pipelined, right? 
To test your change, I think you need no more than these two ops in the loop, do you?

xla/service/collective_pipeliner_test.cc
```diff
@@ -181,6 +181,59 @@ ENTRY entry {
...
+  param = (s32[], bf16[3,8,128], bf16[3,8,128]) parameter(0)
+  gte = s32[] get-tuple-element(param), index=0
+  constant.1 = s32[] constant(3)
+  ROOT cmp = pred[] compare(gte, constant.1), direction=LT
+}
+
+while_body {
+  param = (s32[], bf16[3,8,128], bf16[3,8,128]) parameter(0)
+  current-loop-index = s32[] get-tuple-element(param), index=0
+  parameter.0 = bf16[3,8,128] get-tuple-element(param), index=1
```
frgossen, line 207: If these parameters are relevant to the test, can you give them a name that explains their role?

xla/service/collective_pipeliner_test.cc
```diff
@@ -181,6 +181,59 @@ ENTRY entry {
...
+  tuple = (s32[], bf16[3,8,128], bf16[3,8,128]) tuple(c0, p0, p0)
+  while = (s32[], bf16[3,8,128], bf16[3,8,128]) while(tuple), condition=while_cond, body=while_body
+  ROOT gte1 = bf16[3,8,128] get-tuple-element(while), index=1
+}
+)";
+  auto module = ParseAndReturnUnverifiedModule(hlo_string, config_).value();
+  EXPECT_TRUE(RunOptimizer(module.get(), /*last_run=*/true).value());
+  XLA_VLOG_LINES(1, module->ToString());
+  const HloInstruction* root = module->entry_computation()->root_instruction();
+  EXPECT_THAT(root, op::DynamicUpdateSlice(_, _, _, _, _));
```
frgossen, line 234: Can we check a bit more here? I think the test case should check the `all-reduce`s that you want to see pipelined.

frgossen (changes requested):
Thanks for simplifying the test. 
I think can still be simplified quite a bit more to test your change.

xla/service/collective_pipeliner_test.cc
terryysun: yes this is the part we want to test. but in order to trigger collective pipeliner, we need the pattern of `compute op` (this is what to overlap, dynamic-slice here) `-> collective op` (the other key part to overlap, all-reduce here) `-> elementwise formatting op` (what we are updating, bitcast here) `-> dynamic-update-slice` (because collective pipeliner will reuse the output buffer for intermediate result), I believe it's already the minimal set of ops for this pattern.

xla/service/collective_pipeliner_test.cc
terryysun: renamed

xla/service/collective_pipeliner_test.cc
terryysun: added

terryysun:
@frgossen could you take another look? Thanks!

frgossen (approved):

PR #15292, opened 2024-07-25
[NVIDIA GPU] Disable post layout assignment collective pipeliner by default

Tixxx:
This is to address the reduce-scatter pipelining issue after enabling post layout assignment collective pipeline by default.


golechwierowicz (approved):
Thank you!

PR #15303, opened 2024-07-25
A better python API for accessing profiled instructions

yliu120:
A previous PR https://github.com/openxla/xla/pull/15170 adds a python binding for accessing the profiled instruction.

Actually the API previously added contains repeated logic with `get_fdo_profile`, which works very similar as the added `get_instructions_profile`. The only difference is that `get_fdo_profile` writes to a string.

In essense, we only need to parse the proto to Python objects, if users really want to do so. 

This PR also adds a GPU test for the new API.


yliu120:
@hawkinsp Sorry that before I have chance to add the test, the previous PR was merged.

I made a new PR that simply parse the proto and I guess we don't need to repeat xplane -> ProfiledInstructionsProto in many places. I also added a test to guard this API.

Thanks Peter.

hawkinsp (approved):

yliu120:
Thanks. Just a note for the future outside-Google developer that needs to add tests for "no_oss" GPU tests.

I basically need to do,
1) comment out the "no_oss" tag
2) Add "xla_gpu_extension" and "gpu_plugin_deps" targets to the test target manually
3) Run `bazel test` with `--config=monolithic` to make it work.

xla/python/profiler.cc
```diff
@@ -270,22 +278,28 @@ void BuildProfilerSubmodule(nb::module_& m) {
...
-        for (const auto& c : profile_proto.costs()) {
-          results.emplace_back(c.name(), c.cost_us());
-        }
-        return results;
-      },
-      nb::arg("tensorboard_dir"));
+  // The difference between this API and `get_fdo_profile` is that this
+  // API parses the protos to in-memory Python objects so that they can be
+  // easily manipulated by the users.
+  nb::class_<ProfiledInstruction> profiled_instruction_class(
```
thomasjoerg, line 284: When attempting to merge this PR, the internal CI at Google asked me to double-check that no updates to xla_extension/*.pyi type stub files were forgotten. Such updates may be required to  keep the type defs in sync with the code.

Can you please update type stub files are confirm that no updates are needed please?

xla/python/profiler.cc
```diff
@@ -270,22 +278,28 @@ void BuildProfilerSubmodule(nb::module_& m) {
...
+  profiler.def("parse_profiled_instructions_proto",
+               [](nb::bytes profile) -> std::vector<ProfiledInstruction> {
+                 tensorflow::profiler::ProfiledInstructionsProto profile_proto;
+                 profile_proto.ParseFromString(
+                     std::string(profile.c_str(), profile.size()));
+
+                 std::vector<ProfiledInstruction> result;
+                 for (const auto& c : profile_proto.costs()) {
+                   result.push_back(ProfiledInstruction{
+                       .instruction_name = c.name(), .cost_us = c.cost_us()});
```
thomasjoerg, line 299: Designated initializers are incompatible with C++ standards before C++20.
Please avoid them in the OpenXLA repo at the moment.

thomasjoerg (changes requested):

penpornk:
@yliu120 Could you please help address @thomasjoerg's comments?

yliu120:
Oh, sorry I haven't keep track this PR for a while.

PR #15311, opened 2024-07-25
[ROCm] GPU/CPU unified memory for rocm

i-chaochen:
@xla-rotation

### DIFF
--- a/xla/stream_executor/rocm/rocm_driver.cc
+++ b/xla/stream_executor/rocm/rocm_driver.cc
@@ -1367,16 +1367,32 @@ struct BitPatternToValue {
 /* static */ void* GpuDriver::UnifiedMemoryAllocate(GpuContext* context,
                                                     uint64_t bytes) {
   ScopedActivateContext activated{context};
-
-  LOG(ERROR)
-      << "Feature not supported on ROCm platform (UnifiedMemoryAllocate)";
-  return nullptr;
+  hipDeviceptr_t result = 0;
+  // "managed" memory is visible to both CPU and GPU.
+  hipError_t res = wrap::hipMallocManaged(&result, bytes, hipMemAttachGlobal);  
+  if (res != hipSuccess){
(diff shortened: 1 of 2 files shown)


ezhulenev (approved):

xla/stream_executor/rocm/rocm_driver.cc
```diff
@@ -1367,16 +1367,32 @@ struct BitPatternToValue {
 /* static */ void* GpuDriver::UnifiedMemoryAllocate(GpuContext* context,
```
thomasjoerg, line 1367: Question regarding testing: Are there existing test that cover unified memory and should be enabled for ROCm?

thomasjoerg (changes requested):

xla/stream_executor/rocm/rocm_driver.cc
i-chaochen: I don't find any unified memory tests in XLA, could you point it to me, please? This is request from other users using the same feature in NV side. https://github.com/ROCm/ROCm/issues/3364

and it's same API in [cuda_driver.cc](https://github.com/openxla/xla/blob/main/xla/stream_executor/cuda/cuda_driver.cc#L1476-L1504).

thomasjoerg: Thanks for your quick reply. I checked test coverage stats and `GpuDriver::UnifiedMemoryAllocate` does not seem to have any. Maybe acceptable, because it's just a proxy for an platform-specific alloc.

thomasjoerg (approved):

xla/stream_executor/rocm/rocm_driver.cc
i-chaochen: I think it might have this kind of tests in JAX side, as it's a JAX feature. But I agree with you, it will be great if we can have related test to check this unified memory alloc in XLA.

PR #15317, opened 2024-07-25
Add rendezvous timeouts as new XLA flags.

heiner:
We found the need to increase the warn-stuck timeout at @xai-org.


xla/service/gpu/gpu_executable.cc
```diff
@@ -563,8 +565,14 @@ absl::Status RendezvousAfterInitialization(
       run_options->device_ordinal(),
       run_options->run_options().run_id().ToInt());
 
-  RendezvousSingle(rendezvous_name, rendezvous_key, num_local_participants,
-                   absl::Seconds(10), absl::Seconds(30));
+  RendezvousSingle(
+      rendezvous_name, rendezvous_key, num_local_participants,
+      absl::Seconds(debug_options
+                        ? debug_options->xla_gpu_executable_warn_stuck_timeout()
+                        : 10),
```
thomasjoerg, line 572: Is this code working as intended? The ternary checks if `debug_options` is not null. If so, it will pass the value of xla_gpu_executable_warn_stuck_timeout, which is a boolean, so either 0 or 1. Otherwise it defaults to 10.
Same question for the code below.

xla/debug_options_flags.cc
```diff
@@ -1830,6 +1833,16 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       "version checks must be done by the user (e.g. if you want to use "
       "separate caches for different versions of XLA, please use different "
       "directories). Default: no cache."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_executable_warn_stuck_timeout",
+      bool_setter_for(&DebugOptions::set_xla_gpu_executable_warn_stuck_timeout),
+      debug_options->xla_gpu_executable_warn_stuck_timeout(),
+      "Set timeout for RendezvousSingle stuck warning"));
```
thomasjoerg, line 1840: Consider to make the timeout seconds configurable. One can use 0 to indicate no timeout.

thomasjoerg (changes requested):

xla/service/gpu/gpu_executable.cc
heiner: My apologies. The flags should have been int32, not bool.

thomasjoerg: Thanks for the changes. I wondered if we can avoid repeating the default values here, but saw it's done in very much the same way in existing code too.

thomasjoerg (approved):

xla/service/gpu/gpu_executable.cc
```diff
@@ -478,7 +478,8 @@ absl::Status ExecuteThunks(
   // only in presence of collective cliques which means that we have collective
   // operations in the XLA operations that tend to cause deadlocks.
   if (!collective_cliques.empty()) {
-    TF_RETURN_IF_ERROR(RendezvousAfterInitialization(run_options));
+    TF_RETURN_IF_ERROR(
+        RendezvousAfterInitialization(run_options, debug_options));
```
thomasjoerg, line 482: Did this build for you? I'm seeing the following error:

```
no matching function for call to 'RendezvousAfterInitialization'
  482 |         RendezvousAfterInitialization(run_options, debug_options));
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

You probably want to add `debug_options` on line 353 too.

thomasjoerg (changes requested):

penpornk:
@heiner Could you please help address @thomasjoerg's comment and resolve merge conflicts?

thomasjoerg (approved):

nouiz:
@thomasjoerg Any idea what is blocking it? It was approved 3 days ago.

thomasjoerg:
I resolved the merge conflict.

PR #15092, opened 2024-07-18
Support cuDNN frontend scaled dot product attention for FP8. Part- 1(forward)

wenscarl:
Add cudnn frontend support of scaled dot product attention for FP8 forward. doc [here](https://github.com/NVIDIA/cudnn-frontend/blob/98ca4e1941fe3263f128f74f10063a3ea35c7019/docs/operations/Attention.md).

NOTE: this feature relies on cudnn-frontend v1.6.1 which is not in XLA yet.

PR #15331, opened 2024-07-25
Support cuDNN frontend scaled dot product attention for FP8. Part- 2(backward) 

wenscarl:
As the 2nd part of #15092.
NOTE: this feature relies on cudnn-frontend v1.6.1 which is not in XLA yet.


xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -5140,6 +5140,342 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
...
+
+  auto next_uid = [uid = 0]() mutable -> int { return CuDnnTensorUID(uid++); };
+
+  std::shared_ptr<Tensor_attributes> q_tensor =
+      graph.tensor(Tensor_attributes()
+                       .set_name("Q")
+                       .set_dim(q_descriptor.GetCudnnCompatibleDimensions(true))
+                       .set_stride(q_descriptor.GetCudnnCompatibleStrides(true))
+                       .set_uid(next_uid()));
+  auto dim = k_descriptor.GetCudnnCompatibleDimensions(true);
```
sergachev, line 5188: looks unused

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -5140,6 +5140,342 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
...
+    }
+  }
+
+  cudnn_frontend::graph::Graph graph;
+  dnn::DataType q_type = q_descriptor.type();
+  dnn::DataType k_type = k_descriptor.type();
+  dnn::DataType v_type = v_descriptor.type();
+  dnn::DataType o_type = o_descriptor.type();
+  if (!(q_type == k_type && k_type == v_type && v_type == o_type)) {
+    return absl::InternalError("Input datatypes do not match");
```
sergachev, line 5172: "o" is not an input

xla/stream_executor/cuda/cuda_dnn.h
```diff
@@ -713,6 +713,25 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionBackwardOperationGraph(
...
+    const float scale, const dnn::FMHAMaskKind mask_type);
+
+absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionBackwardF8OperationGraph(
+    dnn::DnnSupport& dnn_support, const dnn::MatmulTensorDescriptor& q_desc,
+    const dnn::MatmulTensorDescriptor& k_desc,
+    const dnn::MatmulTensorDescriptor& p_desc,
+    const dnn::MatmulTensorDescriptor& v_desc,
+    const dnn::MatmulTensorDescriptor& do_desc,
+    const dnn::TensorDescriptor& dq_desc, const dnn::TensorDescriptor& dk_desc,
+    const dnn::TensorDescriptor& dv_desc, double scale,
```
sergachev, line 732: Is there some reason for scale to be double here but float in GetCudnnFlashAttentionF8OperationGraph()?

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -5140,6 +5140,342 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
 #endif
 }
 
+absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionF8OperationGraph(
```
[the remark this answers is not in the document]
wenscarl: Yes.

sergachev (approved):

dimitar-asenov:
@philipphack: Gentle ping. Could you please provide an update for this PR, is it in review or blocked on something?

bchetioui:
Given that cuDNN's `FlashAttention` is meant to remain behind a flag (as discussed previously), I wonder whether it still makes sense to integrate this within XLA.

I believe that we should already support calls to scaled dot product attention through JAX directly, is that correct?

wenscarl:
> Given that cuDNN's `FlashAttention` is meant to remain behind a flag (as discussed previously), I wonder whether it still makes sense to integrate this within XLA.
> 
> I believe that we should already support calls to scaled dot product attention through JAX directly, is that correct?

Do you refer to https://github.com/jax-ml/jax/pull/22670? If so, jax's sdpa api still calls cudnn sdpa from XLA behind the scene. Plus, the forward pass PR is already merged.

xla/service/gpu/cublas_cudnn.cc
```diff
@@ -124,6 +127,22 @@ bool IsCustomCallToDnnNorm(const HloInstruction& hlo) {
   return target == kCudnnNormCallTarget;
 }
 
+bool IsFwdCustomCallTofMHAF8(const HloInstruction& hlo) {
+  if (hlo.opcode() != HloOpcode::kCustomCall) {
```
bchetioui, line 131: nit: here and below, we could maybe synthesize the function into a one-liner: `return hlo.opcode() == HloOpcode::kCudnnfMHASoftmaxF8CallTarget && hlo.custom_call_target() == kCudnnfMHASoftmaxF8CallTarget;`

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1102,6 +1102,212 @@ class FlashAttentionBMMScalePaddingMaskSoftmaxBMM
...
+  void TestImpl_Flash_Attention_Inference_BMM1_NoMask_Softmax_BMM2_F8() {
+    if (skip_reason_) GTEST_SKIP() << *skip_reason_;
+    if (GetDnnVersionInfoOrDefault(backend().default_stream_executor()) <
+        se::dnn::VersionInfo(9, 1, 0)) {
+      GTEST_SKIP() << "Flash Attention requires cuDNN >= 9.1.0.";
+    }
+    XlaBuilder builder(TestName());
+    std::string hlo_string_ref =
+        R"(
+    HloModule jit__unnamed_wrapped_function_, entry_computation_layout={(bf16[4,16,4,16]{3,2,1,0}, bf16[4,16,4,16]{3,2,1,0}, bf16[4,16,4,16]{3,2,1,0})->bf16[4,4,16,16]{3,1,2,0}}, allow_spmd_sharding_propagation_to_parameters={true,true,true}, allow_spmd_sharding_propagation_to_output={true}
```
bchetioui, line 1117: Can you explain what is the purpose of all the HLO leading to the custom call? Does it actually provide value? If not, should we just compare the upcasted custom call vs the reference one?

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1102,6 +1102,212 @@ class FlashAttentionBMMScalePaddingMaskSoftmaxBMM
...
+      divide.32 = bf16[4,16,4,16]{3,2,1,0} divide(Arg_2.3, broadcast.7)
+      call.41 = bf16[4,16,4,16]{3,2,1,0} call(divide.32, constant.5, constant.4), to_apply=clip.33
+      convert.42 = f8e4m3fn[4,16,4,16]{3,2,1,0} convert(call.41)
+      convert.43 = bf16[4,16,4,16]{3,2,1,0} convert(convert.42)
+      custom-call.21.0 = (f8e4m3fn[4,4,16,16]{3,1,2,0}, f32[1,1,1,1]{3,2,1,0}, f32[1,1,1,1]{3,2,1,0}, u8[16]{0}) custom-call(convert.18, convert.30, convert.42, broadcast.99, broadcast.99, /*index=5*/broadcast.99, broadcast.99, broadcast.99, broadcast.99), custom_call_target="__cudnn$fmhaSoftmax$f8", operand_layout_constraints={f8e4m3fn[4,16,4,16]{3,2,1,0}, f8e4m3fn[4,16,4,16]{3,2,1,0}, f8e4m3fn[4,16,4,16]{3,2,1,0}, f32[1,1,1,1]{3,2,1,0}, f32[1,1,1,1]{3,2,1,0}, f32[1,1,1,1]{3,2,1,0}, f32[1,1,1,1]{3,2,1,0}, f32[1,1,1,1]{3,2,1,0}, f32[1,1,1,1]{3,2,1,0}}
+      get-tuple-element.5.0 = f8e4m3fn[4,4,16,16]{3,1,2,0} get-tuple-element(custom-call.21.0), index=0
+      ROOT out = bf16[4,4,16,16]{3,1,2,0} convert(get-tuple-element.5.0)
+    } // main.106
+  )";  // NOLINT
+    EXPECT_TRUE(RunAndCompareTwoModules(hlo_string, hlo_string_ref,
```
bchetioui, line 1189: Do we need to run HLO passes here? If not, let's disable them in this call. (`/*run_hlo_passes=*/false`).

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1102,6 +1102,212 @@ class FlashAttentionBMMScalePaddingMaskSoftmaxBMM
...
+  }
+
+  void TestImpl_Flash_Attention_Bwd_BMM1_NoMask_Softmax_BMM2_F8() {
+    if (skip_reason_) GTEST_SKIP() << *skip_reason_;
+    if (GetDnnVersionInfoOrDefault(backend().default_stream_executor()) <
+        se::dnn::VersionInfo(9, 1, 0)) {
+      GTEST_SKIP() << "Flash Attention requires cuDNN >= 9.1.0.";
+    }
+    XlaBuilder builder(TestName());
+    // generate padding mask in cuDNN directly
```
bchetioui, line 1200: Are we actually doing any pattern matching here?

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1102,6 +1102,212 @@ class FlashAttentionBMMScalePaddingMaskSoftmaxBMM
...
+    // XLA pattern match does not support pattern matching padding mask
+    // so directly lower to custom call instead for reference
+    std::string hlo_string_ref = R"(
+    HloModule jit__unnamed_wrapped_function_, entry_computation_layout={(bf16[1,1,256,128]{3,2,1,0}, bf16[1,1,256,128]{3,2,1,0}, bf16[1,1,256,128]{3,2,1,0}, bf16[1,1,256,128]{3,2,1,0}, bf16[1,1,256,128]{3,2,1,0})->bf16[1,1,256,128]{3,1,2,0}}, allow_spmd_sharding_propagation_to_parameters={true,true,true,true,true}, allow_spmd_sharding_propagation_to_output={true}
```
bchetioui, line 1204: Same questions as above for the HLOs.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
```diff
@@ -276,6 +321,89 @@ absl::StatusOr<se::gpu::CudnnGraph> HloCustomCallToCuDnnGraph(
...
             dnn_mask_type, force_deterministic));
     return std::move(graph);
+  } else {
+    TF_ASSIGN_OR_RETURN(
+        auto gpu_config,
+        custom_call->backend_config<xla::gpu::GpuBackendConfig>());
+    xla::gpu::CudnnfMHABackendConfig &config =
+        *gpu_config.mutable_cudnn_fmha_backend_config();
+
+    int input_index = 0;
```
bchetioui, line 331: Can you explain why you use `input_index++` instead of just hardcoding the value of the argument in each access?

I think it makes the code harder to read, since to understand what indexing is being done, we need to track the variable all the increments/decrements that happened since declaration.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
```diff
@@ -276,6 +321,89 @@ absl::StatusOr<se::gpu::CudnnGraph> HloCustomCallToCuDnnGraph(
...
+    Shape d_output_shape = custom_call->operand(input_index++)->shape();
+
+    Shape bmm2_grad_gemm1_lhs_shape(config.intermediate_tensor_shape());
+    input_index++;
+
+    TF_ASSIGN_OR_RETURN(const CudnnfMHAKind kind,
+                        GetCudnnfMHAKind(custom_call));
+    TF_RET_CHECK(input_index == 6);
+
+    int output_index = 0;
```
bchetioui, line 349: Same question as for `output_index`.

bchetioui:
> Do you refer to https://github.com/jax-ml/jax/pull/22670? If so, jax's sdpa api still calls cudnn sdpa from XLA behind the scene. Plus, the forward pass PR is already merged.

You're right, that seems reasonable, thanks for the clarification.

bchetioui (changes requested):

bchetioui:
Gentle ping @wenscarl :)

xla/service/gpu/tests/gpu_fused_mha_test.cc
wenscarl: The HLO leading to the custom-call performs a "[cast-to-representable](https://github.com/google/praxis/blob/619fc3463d0eae7b24b0f4aa508280cba790666e/praxis/layers/injection/fp8_nvidia_gpu_test.py#L41)" operation, which adjusts the input to fit within the range that the FP8 data type can represent. Therefore, it's also necessary for the reference implementation to include this step in order to maintain numerical equivalence.

xla/service/gpu/tests/gpu_fused_mha_test.cc
wenscarl: No patten matching here.

xla/service/gpu/tests/gpu_fused_mha_test.cc
wenscarl: Yes. The input HLOs are before optimization ones.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
wenscarl: Removed input and output indexing as there are no variadic inputs or outputs.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
wenscarl: Removed input and output indexing as there are no variadic inputs or outputs.

xla/service/gpu/tests/gpu_fused_mha_test.cc
bchetioui: The purpose of the test seems to be to test that the emitted custom call is correct, right? Presumably, we are not trying to test the end-to-end compilation pipeline---or are we testing anything particularly useful here by running it?

Let's make sure to use already-optimized HLO here and in the other cases where we try to ensure correctness of the custom call, and only use before-optimizations HLO where it is necesary.

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1813,6 +1813,114 @@ XLA_TEST_F(FlashAttentionBMMScaleSoftmaxDropoutBMM,
...
+      ROOT get-tuple-element.8 = bf16[1,1,128,128]{3,2,1,0} get-tuple-element(custom-call.7), index=0
+    })";
+
+  std::string hlo_string = R"(
+HloModule jit__unnamed_wrapped_function_, entry_computation_layout={(bf16[1,1,128,128]{3,2,1,0}, bf16[1,1,128,128]{3,2,1,0}, bf16[1,1,128,128]{3,2,1,0}, bf16[1,1,128,128]{3,2,1,0}, bf16[1,1,128,128]{3,2,1,0}, f32[1,1,128]{2,1,0})->bf16[1,1,128,128]{3,2,1,0}}, allow_spmd_sharding_propagation_to_parameters={true,true,true,true,true,true}, allow_spmd_sharding_propagation_to_output={true}
```
bchetioui, line 1876: Note: in general it would be nice to trim HLOs as much as possible to simplify reading the tests later. Could you please do so in your newly added tests?

For instance, the part after the `HloModule <name>` on this line is not at all required for the test, so we could delete it.

In general, I'd suggest treating HLO test code as if it were real code (though SSA) and writing it such that it's easy to read. What I've found to be helpful is to remove `%` when they appear (they don't here!), trailing `.<integer>` at the end of variable names and rename those to give them meaningful names whenever possible, and re-ordering instructions for logical reasoning. Also, don't hesitate to add `// comments` when you need to explain logic in place, as those are supported by HLO.

Of course, we can be pragmatic about it, but the idea is that it should be easy for test readers to
1. easily understand why the test is structured as it is;
2. easily modify the test if they need to.

bchetioui (approved):
@wenscarl thank you for thoughtfully addressing the comments! Love the new tests, they look great!

I'd prefer if we ran without HLO passes here, but I suppose it doesn't hurt the test that much and improves readability to leave it as is---since the clipping logic can be called several times without duplication.

PR #15340, opened 2024-07-25
Fix an escaped error return in GpuCommandBuffer::Finalize()

shawnwang18:


ezhulenev (approved):

xla/stream_executor/gpu/gpu_command_buffer.cc
```diff
@@ -1006,6 +1006,8 @@ absl::Status GpuCommandBuffer::Finalize() {
       } else {
         TF_RETURN_IF_ERROR(retry);
       }
+    } else {
+      TF_RETURN_IF_ERROR(instantiated);
```
thomasjoerg, line 1010: This line could be removed without breaking any tests. Could you please add some coverage?

thomasjoerg (changes requested):

ezhulenev (approved):

xla/stream_executor/gpu/gpu_command_buffer.cc
thomasjoerg: I see that you made changes to `xla/service/gpu/runtime/command_buffer_cmd_test.cc`. However, the changes affect whitespace only. Can you please explain who these changes add test coverage? Thank you!

thomasjoerg (changes requested):

xla/stream_executor/gpu/gpu_command_buffer.cc
shawnwang18: I mean to bypass the testing rule,  this is quite obvious fix that only handles the error case,  is it really necessary to introduce a more complex unittest to test it?

thomasjoerg: > I mean to bypass the testing rule

Sorry, PRs are auto-formatted on import and whitespace-only changes disappear.

> is it really necessary to introduce a more complex unittest to test it?

The `GpuCommandBuffer::Finalize()` method is long and the code is branchy. Inserting a new `else` somewhere in-between did not seem obviously benign to me when I reviewed this PR. I checked and saw that the retry logic has no test coverage either. The problem here is that anybody refactoring this code in the future is much more likely to break it this way. That said, I don't know how hard it is to test this code. @klucke may be able to comment on this.

klucke: I have some ideas on how to make testing this file better in the long term, but in the short term, I don't have any great ideas.

PR #15444, opened 2024-07-29
Fixed some issues around compiling on Windows.

eaplatanios:
This PR fixes some issues I bumped into when trying to compile XLA on Windows. I still haven't gotten GPU support to work but I'm making progress. The CPU only version compiles fine after some of the changes in this PR. I'll point out some specific issues this PR fixes in comments.

There are also TSL-specific changes that are pulled in a separate PR (#15499).


third_party/tsl/third_party/gpus/cuda/build_defs.bzl.tpl
```diff
@@ -104,9 +104,16 @@ def if_cuda_newer_than(wanted_ver, if_true, if_false = []):
     wanted_major = int(wanted_ver.split('_')[0])
     wanted_minor = int(wanted_ver.split('_')[1])
 
-    configured_version = "%{cuda_version}"
-    configured_major = int(configured_version.split('.')[0])
-    configured_minor = int(configured_version.split('.')[1])
+    # Strip "64_" which appears in the CUDA version on Windows.
```
eaplatanios, line 107: This is mostly self-explanatory. In Windows, the CUDA version looks something like `64_121`. There are other parts of the build that already handle such version numbers but it was not being handled properly here.

xla/backends/profiler/gpu/cupti_buffer_events.cc
```diff
@@ -186,18 +186,18 @@ void AddGraphTraceActivityEvent(CuptiEventCollectorDelegate &collector,
...
-      .annotation = info.annotation,
-      .nvtx_range = info.nvtx_range,
-      .start_time_ns = graph_trace->start,
-      .end_time_ns = graph_trace->end,
-      .device_id = graph_trace->deviceId,
-      .correlation_id = graph_trace->correlationId,
-      .context_id = graph_trace->contextId,
-      .stream_id = graph_trace->streamId,
-      .graph_id = graph_trace->graphId,
+      /* .type = */ CuptiTracerEventType::CudaGraph,
```
eaplatanios, line 189: XLA is configured to build using C++ 17. However, this is a C++ 20 feature, resulting in the following error when trying to compile on Windows:

```
error C7555: use of designated initializers requires at least '/std:c++20'
```

xla/backends/profiler/gpu/cupti_buffer_events.h
```diff
@@ -56,7 +56,7 @@ struct MemcpyDetails {
   int8_t dst_mem_kind;
 
   // ID of the hardware channel on which this operation ran.
-  uint32_t channel_id = -1;
+  uint32_t channel_id = static_cast<uint32_t>(-1);
```
eaplatanios, line 59: This resulted in an implicit type narrowing error (I believe it was C2397). The explicit static cast fixes it.

xla/service/cpu/runtime/conv_impl.h
```diff
@@ -41,7 +41,7 @@ void EigenConv2DImpl(
     Eigen::Index padding_y_after, Eigen::Index lhs_x_dilation,
     Eigen::Index lhs_y_dilation, Eigen::Index rhs_x_dilation,
     Eigen::Index rhs_y_dilation, Eigen::Index feature_group_count,
-    std::optional<std::function<void()>> done_callback = std::nullopt) {
```
eaplatanios, line 44: This resulted in this error:

```
error C2765: 'tensorflow::xla::EigenConv2DImpl': an explicit specialization or instantiation of a function template cannot have any default arguments
```

I just removed a couple default arguments that were causing this error.

xla/pjrt/gpu/se_gpu_pjrt_compiler.cc
```diff
@@ -199,13 +199,17 @@ StreamExecutorGpuCompiler::Compile(CompileOptions options,
 #endif
 }
 
+#if TENSORFLOW_USE_ROCM
```
eaplatanios, line 202: The nested macro seems to not be supported by MSVC. Pushing the inner ifdef outside the other macro seems to work and doesn't change the behavior/functionality of the code here.

xla/service/gpu/kernels/cutlass_gemm_custom_kernel.cc
```diff
@@ -101,7 +101,11 @@ KernelArgsPacking ArgsPacking(int32_t m, int32_t n, int32_t k,
   // object constructed in the storage. For now we ignore it, and it's textbook
   // definition of UB, but for CUTLASS kernels we use today it's perfectly safe.
   struct Params {
+#if defined(_MSC_VER)
+    alignas(64) std::byte storage[1024];
+#else
     alignas(128) std::byte storage[1024];
+#endif
```
eaplatanios, line 108: This results in the following error on Windows:
```
error C2719: '<args_0>': formal parameter with requested alignment of 128 won't be aligned
```
cc @dimvar who previously made the change from `64` to `128`.

eaplatanios:
I also have some changes for `third_party/tsl`. I assume I should make those directly to the `openxla/tsl` repository? If I do that, how do I get the submodule here to update since it appears to be a copy of the code?

ezhulenev:
@ddunl

ddunl:
Thanks!! LGTM, for TSL changes I think it's easiest to open a separate PR on this repo which does the TSL edits in the copy. We are in limbo with TSL right now, some of it has been moved here to `xla/tsl` and I'm still moving the rest. It'd be good if the `third_party/gpus/cuda` change could be a part of that TSL PR also.

eaplatanios:
> Thanks!! LGTM, for TSL changes I think it's easiest to open a separate PR on this repo which does the TSL edits in the copy. We are in limbo with TSL right now, some of it has been moved here to `xla/tsl` and I'm still moving the rest. It'd be good if the `third_party/gpus/cuda` change could be a part of that TSL PR also.

Thanks that's super helpful! I'll go ahead and open that PR as well in this repo.

eaplatanios:
@ddunl I also opened [this one](https://github.com/openxla/shardy/pull/31) for shardy though I'm also not sure if that's the right place for it. And then there's a couple triton changes. I was planning to try and update `openxla/triton` after [this PR](https://github.com/triton-lang/triton/pull/4420) is merged (🤞) but please let me know if there's a better way to do that.

eaplatanios:
I removed the `third_party/tsl` changes from this PR. They're coming in a separate PR with all TSL-related changes.

ddunl:
I think that's the right place for Shardy (or at least Tom who is tagged on that PR will know). I think for Triton once that lands we should get those changes via integrate, I'm not sure if they would accept a PR to the fork. Please tag me on the TSL PR once it's created, and thanks again for the contribution!

ddunl (approved):

eaplatanios:
Great, thanks @ddunl! In that case, I'll tag you once the triton PR is merged so we can figure out how to bring that in. There's also some BUILD file changes that are necessary there but I think I can just put up a PR for those at `openxla/triton`. I'll do that later today.

There's the pending CUDA stubs issue. I'll see if I can find a workaround for that one. It appears to be the last remaining big issue since if I remove the stubs, it compiles fine but fails during linking for those symbols. There's a small other issue in the XLA repo around a large constant string in one file. I'll also try to put up a PR for that one soon.

xla/service/gpu/model/hlo_op_profiles_data.h
```diff
@@ -741,8 +741,8 @@ constexpr char kDeviceHloOpProfiles[] = R"pb(
         clock_cycles: 15
       }
     }
-  }
-
+  })pb"
```
eaplatanios, line 744: @ddunl fyi I pushed the changes to this file after you approved. I realized they were left out. They are meant to address this error in MSVC:

```
error C2026: string too big, trailing characters truncated
```

eaplatanios: `clang-format` made the diff massive but it's just breaking it down to a few concatenated strings as opposed to a single very long one.

xla/service/gpu/fusions/mlir/erase_dead_functions.cc
```diff
@@ -77,8 +77,7 @@ class EraseDeadFunctionsPass
 
 }  // namespace
 
-std::unique_ptr<mlir::OperationPass<mlir::ModuleOp>>
-CreateEraseDeadFunctionsPass() {
+std::unique_ptr<mlir::Pass> CreateEraseDeadFunctionsPass() {
```
eaplatanios, line 80: I pushed this after the approval to resolve a linking error. The header uses `mlir::Pass` and MSVC appears to be sensitive to that resulting in mismatching symbol names that fail to link.

xla/stream_executor/cuda/cuda_executor.cc
```diff
@@ -766,10 +766,6 @@ absl::StatusOr<std::unique_ptr<GpuEvent>> GpuExecutor::CreateGpuEvent(
   return std::move(gpu_event);
 }
 
-absl::StatusOr<std::unique_ptr<Event>> GpuExecutor::CreateEvent() {
```
eaplatanios, line 769: I also pushed this after the approval to fix the last remaining linking error. My understanding of what was going on is that `GPUExecutor` is not an abstract class as it's sometimes instantiated in the codebase directly. This means that it needs to implement `CreateEvent` which is a pure virtual function. Since both the CUDA and the ROCm executors used the same implementation, I moved it to `GPUExecutor` this resolved the linking error. Without this change, `cuda_executor.obj` was ending up with a symbol named `CreateEventA` and the linker was failing for places in the code that needed to link to `CreateEvent`. I'm not quite sure how that `A` ends up in there and what the name mangling scheme is since I'm not familiar with C++, but that's my best attempt at an explanation of what was going on.

PR #15447, opened 2024-07-29
[GPU] Add a flag for the timeout of sharded GEMM fusion autotuning

sergachev:
and set its default value to 1 hour.

### DIFF
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -878,7 +878,9 @@ message DebugOptions {
   // TODO(b/355487968): Remove this option when validation complete.
   bool xla_enable_command_buffers_during_profiling = 317;
 
-  // Next id: 318
+  int64 xla_gpu_sharded_autotuning_timeout_seconds = 318;
+
+  // Next id: 319
 
   // Extra options to pass to the compilation backend (e.g. LLVM); specific
   // interpretation of these values is left to the backend.
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -283,6 +283,8 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
 
   opts.set_xla_enable_command_buffers_during_profiling(false);
 
+  opts.set_xla_gpu_sharded_autotuning_timeout_seconds(3600);
+
   return opts;
 }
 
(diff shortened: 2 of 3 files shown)


xla/xla.proto
```diff
@@ -878,7 +878,9 @@ message DebugOptions {
   // TODO(b/355487968): Remove this option when validation complete.
   bool xla_enable_command_buffers_during_profiling = 317;
 
-  // Next id: 318
+  int64 xla_gpu_sharded_autotuning_timeout_seconds = 318;
```
dimitar-asenov, line 881: If we decide to have this flag, please polish it by:
- Add a comment explaining what it does and what the default value is. The comment should mention the flag that enables sharded autotuning.
- Move the line in the file to be just after the line that defines the autotuning enabled flag.

dimitar-asenov (changes requested):
Thanks for sending this in.

At this point I'm not sure we want to have this flag. Perhaps you can help me understand what's going on here. Here's a few questions and comment:

I'm not familiar with the sharded autotuning. What exactly does enabling it do? Parallelize the autotuning computation further to multiple GPUs/hosts or what?

What I've seen now multiple teams complain internally is that enabling sharded autotuning caused their models to freeze on loading. I would expect that this feature should work exactly the opposite way - allow models to load and compile faster. Can you provide some insights into what situations could cause the models to load so much slower than before?

The reason I feel we shouldn't have this new flag is that ideally, this should all just work optimally, without user involvement. Are there legitimate situations where sharded autotuning is expected to be slower?

Independent of all the things I mentioned above, a value of 3600 for this flag is still too large to allow us to enable sharded autotuning by default. I would prefer to find a fix for the underlying issue.

sergachev:
> What exactly does enabling it do? Parallelize the autotuning computation further to multiple GPUs/hosts or what?

Multiple hosts.

> Can you provide some insights into what situations could cause the models to load so much slower than before?
> Are there legitimate situations where sharded autotuning is expected to be slower?

That's not the expected behavior.

> this should all just work optimally, without user involvement.
>  I would prefer to find a fix for the underlying issue.

Debugging flags are generally not for end users. I'm adding this flag for debugging.

> a value of 3600 for this flag is still too large to allow us to enable sharded autotuning by default.

It's never expected to be hit unless something is broken. Hopefully 3600 s is still better than the current infinite timeout: https://github.com/openxla/xla/pull/13108#discussion_r1633784068

PR #15448, opened 2024-07-29
Fixed Windows compilation for Triton.

eaplatanios:
I have put up an [upstream PR](https://github.com/triton-lang/triton/pull/4416) as the comment mentions. I'm not sure if you want to merge this patch of if you instead want to update your upstream dependency once that PR is merged. If the latter, I'm not sure how to make that update myself in your fork.

EDIT: I put up [another PR](https://github.com/triton-lang/triton/pull/4420) and will be updating the patch here to include those changes unless there is a better recommended path for doing this.


eaplatanios:
I'll close this because iiuc the right move here is to update `openxla/triton` after my PRs are merged to `triton` and then update the commit ID referenced in this repository.

PR #15490, opened 2024-07-30
Do not fail GpuPerformanceWithCollectiveModel on Blackwell

sergey-kozub:
Add bandwidth data for Blackwell in the collective model (the actual numbers are unknown at this point, so just copied from Hopper).
Note that "kLowLatencyMaxBandwidths" value for Hopper was incorrect - fixed it.

Additionally, if nvml doesn't support the card, return "unsupported" instead of CHECK-failing.

### DIFF
--- a/xla/service/gpu/model/gpu_collective_performance_model.h
+++ b/xla/service/gpu/model/gpu_collective_performance_model.h
@@ -57,16 +57,16 @@ class GpuPerformanceWithCollectiveModel : public GpuPerformanceModelBase {
 
   // Table for max system bandwidths GB/s for using NCCL's low latency
   // algorithm. This is used for intra-node estimate.
-  static constexpr std::array<double, 3> kLowLatencyMaxBandwidths = {
-      39.0 /* Volta*/, 87.7 /* Ampere*/, 87.7 /* Hopper*/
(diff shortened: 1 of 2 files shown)


anlunx (approved):

xla/service/gpu/model/gpu_collective_performance_model.h
```diff
@@ -57,16 +57,16 @@ class GpuPerformanceWithCollectiveModel : public GpuPerformanceModelBase {
 
   // Table for max system bandwidths GB/s for using NCCL's low latency
   // algorithm. This is used for intra-node estimate.
-  static constexpr std::array<double, 3> kLowLatencyMaxBandwidths = {
-      39.0 /* Volta*/, 87.7 /* Ampere*/, 87.7 /* Hopper*/
+  static constexpr std::array<double, 5> kLowLatencyMaxBandwidths = {
+      39.0 /* Volta */,      87.7 /* Ampere */,    141.0 /* Hopper */,
+      141.0 /* Blackwell */, 141.0 /* next-gen */,
```
olegshyshkov, line 62: Could you please add a comment that this number is not final yet and needs to be updated after the Blackwell release?

olegshyshkov (approved):

PR #15557, opened 2024-07-31
Add OUTPUT_PATH environment for housing generated whl file

hsharsha:


hsharsha:
@ddunl Please see if we can accommodate this.

ddunl:
To clarify - does this fix when building with TF or JAX? I think I'm not completely understanding the motivation

hsharsha:
> To clarify - does this fix when building with TF or JAX? I think I'm not completely understanding the motivation

This is to build with TF so that we can house tf wheel packages in the directory listed with OUTPUT_PATH which can be passed to bazel using repo_env

ddunl:
Gotcha, @vam-google does this look good to you?

vam-google (commented):
This PR adds OUTPUT_PATH env variable to the corresponding repository rule, but does not use it in any way, so this PR as is should be a no-op. If that OUTPUT_PATH env variable is expected to be propagated/consumed somehow indirectly (i.e. basically magically) please explain how it is supposed to work, as as of now it seems no-op.

hsharsha:
> This PR adds OUTPUT_PATH env variable to the corresponding repository rule, but does not use it in any way, so this PR as is should be a no-op. If that OUTPUT_PATH env variable is expected to be propagated/consumed somehow indirectly (i.e. basically magically) please explain how it is supposed to work, as as of now it seems no-op.

I had a long standing PR at https://github.com/tensorflow/tensorflow/pull/66633
where tensorflow/tools/pip_package/utils/tf_wheel.bzl will consume OUTPUT_PATH variable and pass it onto setup.py to build wheel file.

Now with latest tensorflow python_repo.bzl is coming from XLA instead of tensorflow hence the corresponding change in XLA.

vam-google:
@hsharsha Please check my comment in TF's repo for that PR. IIUC, unfortunately your prs are trying to do what we explicitly forbid to do.

ddunl:
I'm closing for now based on the comment from @vam-google but feel free to reopen if anything is unclear

PR #15565, opened 2024-07-31
Fixed remaining issues for Windows CUDA support.

eaplatanios:
These are changes I included late yesterday in #15444 but I believe an earlier copy of that branch was picked up when merging so they were left out. cc @ddunl


xla/service/gpu/model/hlo_op_profiles_data.h
```diff
@@ -22,331 +22,961 @@ namespace gpu {
 // The data below is obtained with
 // xla/service/gpu/model:hlo_op_profiler_run
 
-constexpr char kDeviceHloOpProfiles[] = R"pb(
```
eaplatanios, line 25: They are meant to address this error in MSVC:
```
error C2026: string too big, trailing characters truncated
```
`clang-format` made the diff massive but it's just breaking it down to a few concatenated strings as opposed to a single very long one.

xla/stream_executor/cuda/cuda_executor.cc
```diff
@@ -809,10 +809,6 @@ absl::StatusOr<std::unique_ptr<GpuEvent>> GpuExecutor::CreateGpuEvent(
   return std::move(gpu_event);
 }
 
-absl::StatusOr<std::unique_ptr<Event>> GpuExecutor::CreateEvent() {
```
eaplatanios, line 812: My understanding of what was going on is that `GPUExecutor` is not an abstract class as it's sometimes instantiated in the codebase directly. This means that it needs to implement `CreateEvent` which is a pure virtual function. Since both the CUDA and the ROCm executors used the same implementation, I moved it to `GPUExecutor` this resolved the linking error. Without this change, `cuda_executor.obj` was ending up with a symbol named `CreateEventA` and the linker was failing for places in the code that needed to link to `CreateEvent`. I'm not quite sure how that A ends up in there and what the name mangling scheme is since I'm not familiar with C++, but that's my best attempt at an explanation of what was going on.

ddunl:
LGTM, just going to double check that CI agrees with these changes then I'll try to merge. Thanks!!

ddunl (approved):

ddunl:
Apologies for the delay in merging, can you rebase? Our internal CI was happy with everything but saw some files were out of date. Apologies again for the delay.

eaplatanios:
@ddunl unfortunately, as described [here](https://github.com/openxla/xla/pull/15518#issuecomment-2322061872) I decided to give up on this because it's much more work than I was anticipating and the workaround of using WSL is likely good enough for my current use case.

PR #15577, opened 2024-07-31
[PJRT:GPU] Add setting for mocked number of hosts per slice

jaro-sevcik:
With the existing `enable_mock_nccl` setting it is impossible to warm up compilation cache when there are multiple processes per node. This is because the cache key includes topology and GPU topology contains information about number of slices and number of hosts per slice. The current mocking of topologies always sets num_hosts_per_slice to 1. However, if you have multiple GPUs on a node and run a process-per-GPU then num_hosts_per_slice must be set to the number of GPUs.

This patch allows setting num_hosts_per_slice explicitly when creating the GPU client.

### DIFF
--- a/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
+++ b/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
@@ -83,8 +83,7 @@ PJRT_Error* PJRT_Client_Create(PJRT_Client_Create_Args* args) {
           {"node_id", PJRT_NamedValue_Type::PJRT_NamedValue_kInt64},
           {"num_nodes", PJRT_NamedValue_Type::PJRT_NamedValue_kInt64},
           {"enable_mock_nccl", PJRT_NamedValue_Type::PJRT_NamedValue_kBool},
-          {"mock_num_hosts_per_slice",
-           PJRT_NamedValue_Type::PJRT_NamedValue_kInt64},
+          {"mock_gpu_topology", PJRT_NamedValue_Type::PJRT_NamedValue_kString},
       });
   PJRT_RETURN_IF_ERROR(
       ValidateCreateOptions(create_options, kExpectedOptionNameAndTypes));
@@ -143,10 +142,10 @@ PJRT_Error* PJRT_Client_Create(PJRT_Client_Create_Args* args) {
       it != create_options.end()) {
     enable_mock_nccl = std::get<bool>(it->second);
   }
-  int mock_num_hosts_per_slice = 1;
-  if (auto it = create_options.find("mock_num_hosts_per_slice");
+  std::string mock_gpu_topology;
+  if (auto it = create_options.find("mock_gpu_topology");
       it != create_options.end()) {
-    mock_num_hosts_per_slice = std::get<int64_t>(it->second);
+    mock_gpu_topology = std::get<std::string>(it->second);
   }
 
   xla::GpuClientOptions options;
@@ -159,7 +158,7 @@ PJRT_Error* PJRT_Client_Create(PJRT_Client_Create_Args* args) {
       pjrt::ToCppKeyValueStore(args->kv_get_callback, args->kv_get_user_arg,
                                args->kv_put_callback, args->kv_put_user_arg);
   options.enable_mock_nccl = enable_mock_nccl;
-  options.mock_num_hosts_per_slice = mock_num_hosts_per_slice;
+  options.mock_gpu_topology = mock_gpu_topology;
   PJRT_ASSIGN_OR_RETURN(std::unique_ptr<xla::PjRtClient> client,
                         xla::GetStreamExecutorGpuClient(options));
   args->client = pjrt::CreateWrapperClient(std::move(client));
--- a/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
+++ b/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
@@ -142,7 +142,7 @@ PJRT_Error* PJRT_Client_Create(PJRT_Client_Create_Args* args) {
       it != create_options.end()) {
     enable_mock_nccl = std::get<bool>(it->second);
   }
-  std::string mock_gpu_topology;
+  std::optional<std::string> mock_gpu_topology;
   if (auto it = create_options.find("mock_gpu_topology");
       it != create_options.end()) {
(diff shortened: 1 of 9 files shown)


cheshire:
Looks reasonable to me, but I think @hawkinsp should LGTM as well.

penpornk:
@hawkinsp Could you please help take a look?

sgerrard:
@hawkinsp A friendly nudge to take a look, your review has been requested for this PR.

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -83,6 +83,8 @@ PJRT_Error* PJRT_Client_Create(PJRT_Client_Create_Args* args) {
           {"node_id", PJRT_NamedValue_Type::PJRT_NamedValue_kInt64},
           {"num_nodes", PJRT_NamedValue_Type::PJRT_NamedValue_kInt64},
           {"enable_mock_nccl", PJRT_NamedValue_Type::PJRT_NamedValue_kBool},
+          {"mock_num_hosts_per_slice",
```
cheshire, line 86: Probably makes more sense to have a single flag, with expected GPU topology? I'm also not sure what "slice" means, if we are talking about a number of hosts per slice?

cheshire (changes requested):

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
jaro-sevcik: Slice means a node/machine here.

For instance, if you have 8 nodes, each with 8 GPUs and those 8 GPUs are split between 2 processes on each node, then

number of slices = 8
number of hosts per slice = 2
number of devices per host = 4

The naming is taken from here: https://github.com/openxla/xla/blob/9544c5184db44e05b77b46637c3c98af812f26f5/xla/pjrt/gpu/gpu_topology.proto#L25

How do you imagine the single flag to look like? Something like mock_gpu_topology=8x2? Or do you want to encode the entire GPU topology proto in the flag? I.e. something like mock_gpu_topology=8x2x4x80xNVIDIA A100-SXM4-40GB (not a big fan of this one).

cheshire: > Slice means a node/machine here.

I'm not sure it does; @hawkinsp WDYT? Since we've discussed these namings recently.

> Something like mock_gpu_topology=8x2?

Exactly. Also given that hosts are organized into slices potentially up to 8x8x8

> I.e. something like mock_gpu_topology=8x2x4x80xNVIDIA A100-SXM4-40GB (not a big fan of this one).

This needs to be a separate flag, and I think orthogonal to this change.

I think it makes sense to have two modes:

 - Fully device-less mocking: does not exist now, will require supplying device spec with a flag
 - Mocking from 0th GPU: just specify topology.

jaro-sevcik: Currently, devices will be assigned the same slice index iff [their boot_id is the same](https://github.com/openxla/xla/blob/6fabf593785ae1391c40d01d60618c249ce5c9ea/xla/pjrt/distributed/topology_util.cc#L140). So in essence, a slice is currently implemented to correspond to the machine. I do realize that this might not be consistent with the [comment on the proto](https://github.com/openxla/xla/blob/6fabf593785ae1391c40d01d60618c249ce5c9ea/xla/pjrt/gpu/gpu_topology.proto#L23) "Devices on the same slice are connected by the fast network via NVLinks, which could be within a host or span across multiple hosts".

As for the `mock_gpu_topology`, should this replace the `num_nodes` parameter? Or are you imagining this would supersede `num_nodes`? (I.e., if topology is specified, then num-nodes is ignored or at least force to be consistent with the topology.) Currently, the code for building GPU topology assumes that [nodes and hosts are the same thing](https://github.com/openxla/xla/blob/6fabf593785ae1391c40d01d60618c249ce5c9ea/xla/pjrt/distributed/topology_util.cc#L259). Or are we assuming that hosts are not necessarily the same thing as nodes, so we just stamp out the GPU topology without any checks (and also bypassing the global topology proto)?

I am currently leaning towards creating the GPU topology from the provided mock topology and not insisting on num_nodes==num_hosts_per_slice*num_slices (even though you will then compile with topology that will never exist in practice, so the cache will never hit).

cheshire (commented):
How do we run and exercise this change?

jaro-sevcik:
Are you asking how it's going to be exposed in JAX? Here is a draft PR: https://github.com/google/jax/pull/23374

Or were you asking for some JAX scripts that make use of that option?

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
cheshire: > So in essence, a slice is currently implemented to correspond to the machine

As you specify further down, that seems to contradict the meaning of the word "slice" as used already in PjRT.

At least intuitively, I would interpret `--mock-gpu-topology=8x8` as 8 hosts with 8 GPUs each (and 8 slices I'd guess?)

> As for the mock_gpu_topology, should this replace the num_nodes parameter

Ideally the proto need to be redone, but I think the first number from mock GPU topology should specify num_nodes.

I'd ignore num_slices entirely, or just assume it's always equal to the number of hosts for the initial implementation.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1069,12 +1070,32 @@ absl::StatusOr<DeviceTopologyPair> BuildDistributedDevices(
 
   GlobalTopologyProto global_topology;
   if (enable_mock_nccl) {
+    int num_slices = num_nodes;
+    int num_hosts_per_slice = 1;
+    if (!mock_gpu_topology.empty()) {
+      std::vector<std::string> topology_components =
```
cheshire, line 1076: Maybe explicit `#slicesx#hostsx#gpus-per-host`?

cheshire (changes requested):

dimitar-asenov:
@hawkinsp Could you please take a look at this PR?

jaro-sevcik:
@cheshire Could you take another look, please?

cheshire (approved):

cheshire (approved):

PR #15584, opened 2024-07-31
Dequantization Patterns in Collective Quantizer

philipphack:
Extends the collective quantizer pass to exchange collectives with preceding dequantizations or type conversions to a wider type.

### DIFF
--- a/xla/service/collective_quantizer.cc
+++ b/xla/service/collective_quantizer.cc
@@ -15,6 +15,7 @@ limitations under the License.
 
 #include "xla/service/collective_quantizer.h"
 
+#include "xla/service/hlo_replication_analysis.h"
 #include "xla/service/pattern_matcher.h"
 #include "xla/shape_util.h"
 
@@ -22,6 +23,20 @@ namespace xla {
 namespace {
 
 namespace m = match;
+
+// Holds the ops of a subgraph describing quantization (conversion to a narrower
+// type after scaling by a broadcasted scalar and clamping), dequantization
+// (scaling by a broadcasted scalar after type conversion to a wider type) or
+// plain type conversion.
+struct ConversionSubgraph {
+  HloInstruction* convert = nullptr;
+  HloInstruction* binary = nullptr;
+  HloInstruction* clamp = nullptr;
+  HloInstruction* scale_bcast = nullptr;
+  // Unary instructions following a dequantization or preceding a quantization
+  // in top-down order (operand-to-user).
+  std::vector<HloInstruction*> unaries;
+};
 
 // Matches a broadcast of a scalar operand.
 template <typename... Args>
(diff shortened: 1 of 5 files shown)


xla/service/collective_quantizer.cc
```diff
@@ -50,6 +50,108 @@ auto ConvertToNarrowerType() {
...
+    }
+  }
+  return {};
+}
+
+// Returns true iff instr describes a dequantization, i.e. a multiplication or
+// division by a broadcasted scalar operating on a type conversion, or a plain
+// type conversion to a wider type. Unary bitcast, copy, reshape or slice ops
+// may follow the dequantization or type conversion.
+bool IsSupportedDequantization(HloInstruction* instr, HloInstruction** convert,
```
cheshire, line 119: Could we return an optional<struct> (with a new struct) instead of using out parameters to show captures?

xla/service/collective_quantizer.cc
```diff
@@ -50,6 +50,108 @@ auto ConvertToNarrowerType() {
...
+bool IsSupportedCollective(HloInstruction* instr) {
+  return instr->operand_count() == 1 &&
+         (instr->opcode() == HloOpcode::kAllGather ||
+          instr->opcode() == HloOpcode::kAllToAll ||
+          instr->opcode() == HloOpcode::kCollectiveBroadcast ||
+          instr->opcode() == HloOpcode::kCollectivePermute);
+}
+
+// Sequentially applies the ops in unary_ops to the output of instr.
+HloInstruction* ShiftUnaryOps(HloInstruction* instr,
```
cheshire, line 73: Maybe `ApplyUnary`?

xla/service/collective_quantizer.cc
```diff
@@ -50,6 +50,108 @@ auto ConvertToNarrowerType() {
   return m::Convert().WithPredicate(converts_to_narrower_type);
 }
 
+// Matches a type conversion to a type with a larger byte size than that of the
+// operand.
+auto ConvertToWiderType() {
```
cheshire, line 55: Nit: IsConvertToWiderType?

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1103,5 +1103,34 @@ ENTRY entry {
   EXPECT_EQ(all_to_all->shape().element_type(), BF16);
 }
 
+TEST_F(CollectiveOpsTestE2E, ConvertAllToAllCollectiveQuantizer) {
+  absl::string_view kModuleReplicatedStr = R"(
+HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(bf16[4,32,128]{2,1,0})->f32[4,32,128]{2,1,0}}, num_partitions=4
+ENTRY entry {
+  param = bf16[4,32,128]{2,1,0} parameter(0)
+  convert = f32[4,32,128]{2,1,0} convert(param)
+  ROOT all-to-all = f32[4,32,128]{2,1,0} all-to-all(convert), channel_id=1, replica_groups={{0,1,2,3}}, dimensions={1}
```
cheshire, line 1112: Q: Should we have some 2-GPU execution tests?

cheshire (changes requested):

xla/service/collective_quantizer.cc
philipphack: I added struct ConversionSubgraph.

xla/service/collective_quantizer.cc
philipphack: Done.

xla/service/collective_quantizer.cc
philipphack: This is meant to replace m::Convert in Match calls. Let me know if you want to rename it anyway.

xla/tests/collective_ops_e2e_test.cc
philipphack: This test verifies that the type of the all-to-all has changed from F32 to BF16. The HLO isn’t executed.

philipphack:
@cheshire can you PTAL?

xla/service/collective_quantizer.cc
```diff
@@ -50,14 +60,121 @@ auto ConvertToNarrowerType() {
...
+            unary_op->shape().layout().minor_to_major()),
+        {instr}));
+  }
+  return instr;
+}
+
+// Recursively collects unary, divide, or multiply operands of instr until a
+// conversion to a wider type is reached. Returns an empty vector when no
+// conversion is reached.
+std::vector<HloInstruction*> FindDequantizationSubgraphRecursive(
```
reedwm, line 98: Can you write this function in a similar style to the [corresponding function](https://github.com/openxla/xla/blob/676521539300ce2141c169b314d293bda920c277/xla/service/gpu/transforms/gemm_rewriter.cc#L200) in gemm_rewriter.cc? A few months ago, I refactored the function to make it more clear. I think you can just copy the new version with some minor changes, like using `std::vector<HloInstruction*>` instead of `std::vector<std::pair<HloInstruction *, int>>`.

In particular, the comment on the gemm_rewriter function is more explicit about what is done, and the order of HLO instructions doesn't need to be reversed, and the subgraph shouldn't be both passed-in and returned.

xla/service/collective_quantizer.cc
```diff
@@ -50,14 +60,121 @@ auto ConvertToNarrowerType() {
...
+      }
+    }
+  }
+  return {};
+}
+
+// Returns true iff instr describes a dequantization, i.e. a multiplication or
+// division by a broadcasted scalar operating on a type conversion, or a plain
+// type conversion to a wider type. Unary bitcast, copy, reshape or slice ops
+// may follow the dequantization or type conversion.
```
reedwm, line 128: Be more explicit about the fact that `instr` itself can refer to a unary op. E.g. by replacing the last sentence with

Also returns true if instr refers to the last instruction of a sequence of unary bitcast, copy, reshape or slice ops preceded by a dequantization.

xla/service/collective_quantizer.cc
```diff
@@ -50,14 +60,121 @@ auto ConvertToNarrowerType() {
...
+           m::MultiplyAnyOrder(&subgraph.binary, m::Convert(&subgraph.convert),
+                               ScalarBroadcast(&subgraph.scale_bcast))) ||
+       Match(candidate_subgraph[1],
+             m::Divide(&subgraph.binary, m::Convert(&subgraph.convert),
+                       ScalarBroadcast(&subgraph.scale_bcast))))) {
+    subgraph.unaries = {candidate_subgraph.begin() + 2,
+                        candidate_subgraph.end()};
+  } else if (candidate_subgraph.size() > 0 &&
+             Match(candidate_subgraph[0], m::Convert(&subgraph.convert))) {
+    subgraph.unaries = {candidate_subgraph.begin() + 1,
```
reedwm, line 152: According to [this StackOverflow quesiton](https://stackoverflow.com/questions/60112841/copying-structs-with-uninitialized-members), copying a struct with uninitialized members is undefined behavior. So initialize all members of subgraph to avoid this. The best way is to just set them to nullptr in the definition of ConversionSubgraph.

xla/service/collective_quantizer.cc
```diff
@@ -76,49 +193,132 @@ bool IsSupportedQuantization(HloInstruction* instr, HloInstruction** convert,
...
-                                                  ScalarBroadcast(scale_bcast)),
-                                        ScalarBroadcast()))))) {
-    unary_ops = {ops.begin(), ops.end() - 3};
-  } else if (ops.size() > 0 && Match(ops.back(), m::Convert(convert))) {
-    unary_ops = {ops.begin(), ops.end() - 1};
+             m::Convert(&subgraph.convert,
+                        m::Clamp(&subgraph.clamp, ScalarBroadcast(),
+                                 m::MultiplyAnyOrder(
+                                     &subgraph.binary, m::Op(),
+                                     ScalarBroadcast(&subgraph.scale_bcast)),
```
reedwm, line 207: Moving the quantization before the collective op results in incorrect behavior if the scale is not the same on each GPU. And same if the lower and higher clamp values are not the same on each GPU.

And similarly for the dequantization case.

xla/service/collective_quantizer.cc
```diff
@@ -76,49 +193,132 @@ bool IsSupportedQuantization(HloInstruction* instr, HloInstruction** convert,
...
-bool IsSupportedCollective(HloInstruction* instr) {
-  return instr->opcode() == HloOpcode::kAllGather ||
-         instr->opcode() == HloOpcode::kAllToAll ||
-         instr->opcode() == HloOpcode::kCollectiveBroadcast ||
-         instr->opcode() == HloOpcode::kCollectivePermute;
+absl::Status MatchDequantization(HloInstruction* instr, bool* changed) {
+  std::optional<ConversionSubgraph> subgraph =
+      IsSupportedDequantization(instr->mutable_operand(0));
+
+  if (subgraph.has_value()) {
```
reedwm, line 241: Return early if the subgraph does not have a value. And similarly in MatchQuantization.

xla/service/collective_quantizer.cc
```diff
@@ -50,14 +60,121 @@ auto ConvertToNarrowerType() {
...
 // or slice ops with one user may precede the quantization or type conversion.
-bool IsSupportedQuantization(HloInstruction* instr, HloInstruction** convert,
-                             HloInstruction** binary, HloInstruction** clamp,
-                             HloInstruction** scale_bcast,
-                             std::vector<HloInstruction*>& unary_ops) {
+std::optional<ConversionSubgraph> IsSupportedQuantization(
+    HloInstruction* instr) {
+  ConversionSubgraph subgraph;
   std::vector<HloInstruction*> ops;
   while (instr->user_count() <= 1) {
```
reedwm, line 179: Maybe it would be easier to match the convert by creating a `handleConvert` method and matching up through operands instead of down through users. That way its more consistent with with the DQ case, and you could share the `FindDequantizationSubgraphRecursive` method.

xla/service/collective_quantizer.cc
```diff
@@ -76,49 +193,132 @@ bool IsSupportedQuantization(HloInstruction* instr, HloInstruction** convert,
...
+          instr->AddInstruction(subgraph->clamp->operand(2)->CloneWithNewShape(
+              coll_operand->shape()));
+      new_clamp = instr->AddInstruction(subgraph->clamp->CloneWithNewOperands(
+          coll_operand->shape(),
+          {new_clamp_lower, new_binary, new_clamp_upper}));
+    }
+
+    // Move the collective past the conversion to the narrow type.
+    Shape new_convert_shape = ShapeUtil::ChangeElementType(
+        instr->operand(0)->shape(), subgraph->convert->shape().element_type());
```
reedwm, line 305: Use `coll_operand` instead of `instr->operand(0)`

xla/service/collective_quantizer.cc
```diff
@@ -50,14 +60,121 @@ auto ConvertToNarrowerType() {
...
+      return std::nullopt;
+    }
+  }
+  return std::make_optional<ConversionSubgraph>(subgraph);
+}
+
 // Returns true iff instr describes a quantization, i.e. a multiplication or
 // division by a broadcasted scalar followed by a clamp and a type conversion,
 // or a plain type conversion to a narrower type. Unary bitcast, copy, reshape
 // or slice ops with one user may precede the quantization or type conversion.
```
reedwm, line 174: In the comment, clarify that `instr` refers to the first, top-most such op in a quantization pattern. As contrasted with IsSupportedDequantization, where `instr` refers to the last, bottom-most op in the dequantization pattern.

xla/tests/collective_ops_e2e_test.cc
reedwm: I think we should have a test that actually executes HLO modified by the collective quantizer.

reedwm (changes requested):
In general, there is a good amount of duplication in logic to match and handle the Q and DQ, e.g. in collective quantizer and gemm_rewriter. Can some of this logic be moved to a shared location?

This doesn't necessarily have to be done in this PR but I think this would be a good idea in general.

xla/service/collective_quantizer.cc
```diff
@@ -92,9 +95,32 @@ HloInstruction* ApplyUnary(HloInstruction* instr,
...
+
+  TF_ASSIGN_OR_RETURN(
+      auto replication_analysis,
+      HloReplicationAnalysis::Run(module,
+                                  /*cross_partition_spmd=*/true));
+  if (!replication_analysis->HloInstructionIsReplicatedAt(instr, {})) {
+    return false;
+  }
+
+  return true;
```
reedwm, line 117: Instead of these lines, just do:

```c++
return replication_analysis->HloInstructionIsReplicatedAt(instr, {});
```

xla/service/collective_quantizer.cc
```diff
@@ -92,9 +95,32 @@ HloInstruction* ApplyUnary(HloInstruction* instr,
...
+}
+
+// Recursively collects and returns unary, divide, or multiply operands of instr
+// until a conversion to a wider type is reached. Returns the collected ops in
+// top-down order (operand-to-user). Returns an empty vector when no conversion
```
reedwm, line 122: Doesn't this return user-to-operand? It is reversed outside the function call, but this function itself returns user-to-operand.

xla/service/collective_quantizer.cc
```diff
@@ -23,13 +24,15 @@ namespace {
...
+// plain type conversion.
 struct ConversionSubgraph {
-  HloInstruction* convert;
+  HloInstruction* convert = nullptr;
   HloInstruction* binary = nullptr;
-  HloInstruction* clamp;
-  HloInstruction* scale_bcast;
+  HloInstruction* clamp = nullptr;
+  HloInstruction* scale_bcast = nullptr;
   std::vector<HloInstruction*> unaries;
```
reedwm, line 36: Document what unaries are, e.g. "List of instructions following a dequantization or preceding a dequantization, in top-down order (operand-to-user)"

xla/service/collective_quantizer.cc
```diff
@@ -279,45 +320,55 @@ absl::Status MatchQuantization(HloInstruction* instr, bool* changed) {
...
-          {new_clamp_lower, new_binary, new_clamp_upper}));
+  if (!subgraph.has_value()) {
+    return absl::OkStatus();
+  }
+
+  if (subgraph->scale_bcast) {
+    // The scale must be replicated across all devices.
+    TF_ASSIGN_OR_RETURN(
+        bool scale_is_replicated,
+        InstrIsReplicated(instr->parent()->parent(), subgraph->scale_bcast));
```
reedwm, line 331: you also should check the clamp values are replicated. Or for simplicity, check they are constant.

xla/service/collective_quantizer_test.cc
```diff
@@ -233,6 +233,30 @@ TEST_F(CollectiveQuantizerTest, AllGatherQuantizeUnary) {
 }
 
 TEST_F(CollectiveQuantizerTest, AllGatherQuantizeMultiUser) {
```
reedwm, line 235: Why explicitly have both a replicated and non-replicated test when both don't change the HLO for the same reason, which is because the all-gather has multiple users and not because the scale is non-replicated?

Instead, I'd have a test that normally would be rewritten, except for the fact the scale is non-replicated.

reedwm (changes requested):

xla/service/collective_quantizer.cc
```diff
@@ -33,6 +33,8 @@ struct ConversionSubgraph {
   HloInstruction* binary = nullptr;
   HloInstruction* clamp = nullptr;
   HloInstruction* scale_bcast = nullptr;
+  // Unary instructions following a dequantization or preceding a dequantization
```
reedwm, line 36: Replace "preceding a dequantization" with "preceding a *quantization*"

reedwm (changes requested):

reedwm (approved):

reedwm (approved):

PR #15604, opened 2024-08-01
[XLA:GPU] Remove gpu_performance_model_cache mutex locks in Priority fusion

Cjkkkk:
* remove mutex for gpu_performance_model_cache. let cache rewrites only happen in sequential order in main thread and make sure only read in worker threads.
* fix cache invalidation to only invalidate producer if it doesn't have any consumers after fusion.

some of the benchmarks using this PR: https://docs.google.com/spreadsheets/d/1EL35JeY9jouqTNvaF9J5jvoaRX9jEIMexNqM-Rk0fss/edit?usp=sharing. Check column B and G.

### DIFF
--- a/xla/service/gpu/priority_fusion.cc
+++ b/xla/service/gpu/priority_fusion.cc
@@ -165,7 +165,7 @@ class GpuPriorityFusionQueue {
     }
     // write here and only read in worker threads
     for (auto producer : instructions) {
-      updatePerformanceModelCache(producer);
+      UpdatePerformanceModelCache(producer);
     }
     ComputeAndSetPriorities(instructions);
   }
@@ -250,7 +250,7 @@ class GpuPriorityFusionQueue {
     return !current_consumers_.empty();
   }
 
-  void updatePerformanceModelCache(HloInstruction* producer) {
+  void UpdatePerformanceModelCache(HloInstruction* producer) {
     if (producer->opcode() == HloOpcode::kBitcast ||
         producer->opcode() == HloOpcode::kConstant ||
         !IsFusible(*producer)) {
@@ -260,15 +260,12 @@ class GpuPriorityFusionQueue {
     auto config = GpuPerformanceModelOptions::PriorityFusion(
         &fusion_analysis_cache_, &gpu_performance_model_cache_);
 
-    EstimateRunTimeData producer_runtime =
-        GpuPerformanceModel::EstimateRunTimeForInstructionCached(
-            producer, *device_info_, &cost_analysis_, config);
-    for (auto consumer : producer->users()) {
-      if (!IsFusible(*consumer)) continue;
-      EstimateRunTimeData consumer_runtime =
-          GpuPerformanceModel::EstimateRunTimeForInstructionCached(
-              consumer, *device_info_, &cost_analysis_, config);
+    if (auto cached_result = gpu_performance_model_cache_.Get(*producer)) {
+      return;
     }
+    auto runtime_data = GpuPerformanceModel::EstimateRunTimeForInstruction(
+        producer, *device_info_, &cost_analysis_, config);
+    gpu_performance_model_cache_.Set(*producer, runtime_data);
   }
 
   // Update priorities of all affected ops.
@@ -279,7 +276,7 @@ class GpuPriorityFusionQueue {
       TF_CHECK_OK(cost_analysis_.RevisitInstruction(instruction));
     }
     for (auto producer : to_update_priority_) {
-      updatePerformanceModelCache(producer);
+      UpdatePerformanceModelCache(producer);
     }
 
     ComputeAndSetPriorities(std::vector<HloInstruction*>{
@@ -428,8 +425,7 @@ class GpuPriorityFusionQueue {
         GpuPerformanceModel::EstimateRunTimesForPriorityFusion(
             producer, *device_info_, &cost_analysis_,
             GpuPerformanceModelOptions::PriorityFusion(
-                &fusion_analysis_cache_, &gpu_performance_model_cache_,
-                true /*gpu_performance_model_cache_read_only*/),
+                &fusion_analysis_cache_, &gpu_performance_model_cache_),
             producer->users());
 
     if (fusion_process_dump_) {
--- a/xla/service/gpu/priority_fusion.cc
+++ b/xla/service/gpu/priority_fusion.cc
@@ -252,8 +252,7 @@ class GpuPriorityFusionQueue {
 
   void UpdatePerformanceModelCache(HloInstruction* producer) {
     if (producer->opcode() == HloOpcode::kBitcast ||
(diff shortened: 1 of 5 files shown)


xla/service/gpu/priority_fusion.cc
```diff
@@ -247,13 +250,37 @@ class GpuPriorityFusionQueue {
     return !current_consumers_.empty();
   }
 
+  void updatePerformanceModelCache(HloInstruction* producer) {
```
olegshyshkov, line 253: Rename to `UpdatePerformanceModelCache`. Google C++ code style requires that function names start from a capital letter.

xla/service/gpu/priority_fusion.cc
```diff
@@ -247,13 +250,37 @@ class GpuPriorityFusionQueue {
...
+        producer->opcode() == HloOpcode::kConstant ||
+        !IsFusible(*producer)) {
+      return;
+    }
+
+    auto config = GpuPerformanceModelOptions::PriorityFusion(
+        &fusion_analysis_cache_, &gpu_performance_model_cache_);
+
+    EstimateRunTimeData producer_runtime =
+        GpuPerformanceModel::EstimateRunTimeForInstructionCached(
```
olegshyshkov, line 264: We can inline the `EstimateRunTimeForInstructionCached` here and that will eliminate the need to `gpu_performance_model_cache_read_only`.

If we do `gpu_performance_model_cache_->Get()` + `EstimateRunTimeForInstruction` + `gpu_performance_model_cache_->Set()` and only do `Get` in `EstimateRunTimesForPriorityFusion`.

xla/service/gpu/priority_fusion.cc
```diff
@@ -395,6 +418,8 @@ class GpuPriorityFusionQueue {
       return std::numeric_limits<Priority>::min();
     }
 
+    // everything should already be in performance model cache
```
olegshyshkov, line 421: This comment is not relevant here anymore.

olegshyshkov (approved):
Thank you!

xla/service/gpu/priority_fusion.cc
```diff
@@ -163,7 +163,10 @@ class GpuPriorityFusionQueue {
       }
       instructions.push_back(instruction);
     }
-
+    // write here and only read in worker threads
+    for (auto producer : instructions) {
```
olegshyshkov, line 167: This loop should run on all fusible instructions in the computation. `instructions` array doesn't have ROOT instructions, because they have no users. Subsequent `ComputeAndSetPriorities` will either crash or return an incorrect result.

xla/service/gpu/priority_fusion.cc
```diff
@@ -395,6 +418,7 @@ class GpuPriorityFusionQueue {
       return std::numeric_limits<Priority>::min();
     }
 
+    // never write cache here as we dont have locks anymore
```
olegshyshkov, line 421: I'm not sure what this comment is trying saying. Maybe it's not necessary?

xla/service/gpu/priority_fusion.cc
```diff
@@ -163,7 +163,10 @@ class GpuPriorityFusionQueue {
       }
       instructions.push_back(instruction);
     }
-
+    // write here and only read in worker threads
```
olegshyshkov, line 166: Could you ellaborate in the comment what do we write here? It's not clear from the current comment.

olegshyshkov (commented):
You need to update tests in `gpu_performance_model_test.cc` to use the cache and fill it with data for individual instructions.

Cjkkkk:
I fixed the unit test and loop through all fusible instructions in the computation to initialize the cache, could you take another look?

xla/service/gpu/priority_fusion.cc
```diff
@@ -164,6 +164,11 @@ class GpuPriorityFusionQueue {
       instructions.push_back(instruction);
     }
 
+    for (auto instruction : computation->MakeInstructionPostOrder()) {
+      if (instruction->IsFusible()) {
```
olegshyshkov, line 168: `IsFusible` is also checked inside `UpdatePerformanceModelCache`. It's enough to do this check only once. Either here or inside the function.

xla/service/gpu/priority_fusion.cc
```diff
@@ -247,13 +252,44 @@ class GpuPriorityFusionQueue {
...
+    auto config = GpuPerformanceModelOptions::PriorityFusion(
+        &fusion_analysis_cache_, &gpu_performance_model_cache_);
+
+    if (!gpu_performance_model_cache_.Get(*producer)) {
+      auto runtime_data = GpuPerformanceModel::EstimateRunTimeForInstruction(
+          producer, *device_info_, &cost_analysis_, config);
+      gpu_performance_model_cache_.Set(*producer, runtime_data);
+    }
+
+    for (auto consumer : producer->users()) {
```
olegshyshkov, line 274: It's not necessary to update costs for consumers here. If everything is done right, the function will be also called for all fusible instructions in the computation anyway.

olegshyshkov (approved):
Thank you for addressing comments!

xla/service/gpu/priority_fusion.cc
Cjkkkk: That's right. Except there is a case where bitcast can still have consumers after fusing as producer. Therefore, after cache invalidate, it can still be in the graph causing cache miss.

olegshyshkov: Is this because of `producer->opcode() == HloOpcode::kBitcast` check above? I think we can remove that check, it's not necessary.

Cjkkkk: https://github.com/openxla/xla/blob/0a4d157ba44b378221a915965036c0aff0120a70/xla/service/gpu/priority_fusion.cc#L241 here, bitcast can still have consumers after fusing with all its consumers so after cache invalidation, you wont find bitcast in the cache but it can still be required. I changed the logic to only invalidate if the producer has 0 consumers. See my latest commit.

xla/service/gpu/priority_fusion.cc
```diff
@@ -247,13 +247,31 @@ class GpuPriorityFusionQueue {
     return !current_consumers_.empty();
   }
 
+  void UpdatePerformanceModelCache(HloInstruction* producer) {
+    if (!IsFusible(*producer)) {
```
olegshyshkov, line 251: There is one tests failing in priority_fusion_test: `CanMergeTritonFusionWithBothProducerAndConsumer`. That's because we try to use the existing cost model for Triton fusion, but kCustom fusion is considered not fusible. It should be enough to add here: `if (!IsFusible(*producer) && !IsGenericTritonFusion(*producer))` as a temporary fix.

Cjkkkk: fixed.

olegshyshkov (approved):

PR #15615, opened 2024-08-01
[PJRT:GPU] Respect auto layout modes in MLIR compilation

jaro-sevcik:
Without this patch, auto layouts for arguments and results turn into default layouts (in the
DetermineArgumentLayoutsFromCompileOptions function). This patch avoids calling
DetermineArgumentLayoutsFromCompileOptions from the MLIR compilation method
so that empty layouts on shapes are preserved.

With this patch we should be able to enable four more layout tests in JAX on GPUs.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
@@ -1307,6 +1307,64 @@ TEST(StreamExecutorGpuClientTest, MlirResultHostMemorySpaceIsSetInHlo) {
   EXPECT_EQ(result_layout.memory_space(), Layout::kHostMemorySpace);
 }
 
+TEST(StreamExecutorGpuClientTest, MlirAutoResultLayoutIsSet) {
+  constexpr char kMlirWithParameterLayout[] =
+      R"(
+    func.func public @main(%arg0: tensor<2x4x2xi32> {
+            mhlo.layout_mode = "{2, 1, 0}"
+        }) -> (tensor<2x2x4xi32> {
+            jax.result_info = "",
+            mhlo.layout_mode = "auto"}) {
+      %0 = stablehlo.transpose %arg0, dims = [0, 2, 1]
+          : (tensor<2x4x2xi32>) -> tensor<2x2x4xi32>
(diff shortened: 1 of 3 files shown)


ezhulenev (approved):

nouiz:
The PR dot approved, but not merged. Any idea what is going on?

akuegel:
> The PR dot approved, but not merged. Any idea what is going on?

We always need approval by two different people, and only Eugene approved both on GitHub and internally. Therefore merging was blocked, it would have required a different person to approve internally.

@xla-rotation please follow up on getting this PR merged

akuegel (approved):
Approving again to recreate the internal change for merging.

akuegel:
@jaro-sevcik we get a test failure in 3 of the Jax tests:

INVALID_ARGUMENT: Cannot serialize ExecutableBuildOptions::layout_canonicalization_callback

The tests in question are:

https://github.com/google/jax/blob/main/tests/aot_test.py

https://github.com/google/jax/blob/main/tests/pgle_test.py

https://github.com/google/jax/blob/main/tests/compilation_cache_test.py

Any idea how to fix this?

jaro-sevcik:
Thank you for sharing the failures. I have a good idea where this is coming from. Let me fix this tomorrow.

jaro-sevcik:
@akuegel I have added a small (but somewhat ugly) fix to the PR. Could you take a look?

Alternatively, I could pass the layout callback to the CompileInternal method and set it on the compile options only after we take a copy of the compile options for the executable. Please let me know if that's preferable.

akuegel:
> @akuegel I have added a small (but somewhat ugly) fix to the PR. Could you take a look?
> 
> Alternatively, I could pass the layout callback to the CompileInternal method and set it on the compile options only after we take a copy of the compile options for the executable. Please let me know if that's preferable.

I will defer to others here, I am not so familiar with the PJRT stuff. @ezhulenev what do you think?

jaro-sevcik:
I have added the callback as an argument to the CompileInternal method. It does look cleaner.

@ezhulenev , could you take a look, please?

ezhulenev (approved):

PR #15630, opened 2024-08-01
[NVIDIA GPU] Use cuda runtime api to determine if 2 ranks are on the same host

Tixxx:
The current logic in nccl clique sets is_local to true by looking at the number of local participants and total devices in the clique. It's been used to determine if replica group is a local group, but this doesn't always translates to a local communicator.
i.e on a 8-gpu machine, if we have a group for collective permute
replica_group={{0,1},{1,0},{9,10},{10,9}}

In XLA's perspective, this is not a local replica group, but the communicators we create are only for rank (0,1) and (10,9), which are both local communicators in nccl perspective.
This pr uses the cuda runtime api to get the number of devices on a host and use the current rank id to determine if source and targets are located in the same host in collective permute thunk.

### DIFF
--- a/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
+++ b/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
@@ -161,7 +161,10 @@ NcclCollectivePermuteStartThunk::NcclCollectivePermuteStartThunk(
 absl::Status NcclCollectivePermuteStartThunk::Initialize(
     const InitializeParams& params) {
   TF_RETURN_IF_ERROR(NcclCollectiveThunk::Initialize(params));
-  TF_ASSIGN_OR_RETURN(device_count_, params.executor->GetDeviceCount());
+  const se::Platform* platform = params.executor->GetPlatform();
+  CHECK(platform != nullptr);
+  device_count_ = platform->VisibleDeviceCount();
+  CHECK(device_count_ > 0);
   VLOG(5) << "Local device count: " << device_count_;
 
   if (p2p_memcpy_enabled_) {
--- a/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
+++ b/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
@@ -59,23 +59,26 @@ bool IsLocalPeerTransfer(
     se::Stream& stream, const int64_t current_id, const int64_t device_count) {
   const std::optional<int64_t> source_id = source_target.source;
   const std::optional<int64_t> target_id = source_target.target;
-  // Since mixing nccl p2p with p2p memcopy will cause random deadlocks.
+  // Mixing nccl p2p with p2p memcopy will cause random deadlocks, namely
+  // when calling nccl call and cuda memcpy p2p together(which both are
+  // synchronizing devices), in this case if this rank is sending across host
+  // using a nccl call but receiving from a local peer which is going through
+  // cuda api, the deadlock could happen because nccl cannot ensure the
+  // order of cuda api calls.
   // We determine if it's a local peer by the following conditions:
   // 1. Both source and target IDs are present and they are within a node
-  // 2. Source ID is present, but target ID is not.
-  // 3. Target ID is present, but source ID is not.
+  // 2. Source ID is present and is within the node, but target ID is not
+  // present.
+  // 3. Target ID is present and is within the node, but source ID is not
+  // present.
   int64_t host_id = (current_id / device_count);
   if (source_id && target_id) {
     return (host_id == (*source_id / device_count)) &&
            (host_id == (*target_id / device_count));
   }
-  if (source_id) {
-    return (host_id == (*source_id / device_count));
-  }
-  if (target_id) {
-    return (host_id == (*target_id / device_count));
-  }
-  return false;
+  if (source_id && host_id != *source_id / device_count) return false;
+  if (target_id && host_id != *target_id / device_count) return false;
+  return true;
 }
 
 }  // namespace
@@ -158,7 +161,7 @@ NcclCollectivePermuteStartThunk::NcclCollectivePermuteStartThunk(
 absl::Status NcclCollectivePermuteStartThunk::Initialize(
     const InitializeParams& params) {
   TF_RETURN_IF_ERROR(NcclCollectiveThunk::Initialize(params));
-  device_count_ = params.executor->GetDeviceCount();
+  TF_ASSIGN_OR_RETURN(device_count_, params.executor->GetDeviceCount());
   VLOG(5) << "Local device count: " << device_count_;
 
   if (p2p_memcpy_enabled_) {
--- a/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
+++ b/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
@@ -56,7 +56,7 @@ absl::StatusOr<const int64_t> GetCurrentId(
 
 bool IsLocalPeerTransfer(
     const NcclP2PConfig::SourceTargetMapEntry& source_target,
-    se::Stream& stream, const int64_t current_id, const int64_t device_count) {
+    const int64_t current_id, const int64_t device_count) {
   const std::optional<int64_t> source_id = source_target.source;
   const std::optional<int64_t> target_id = source_target.target;
   // Mixing nccl p2p with p2p memcopy will cause random deadlocks, namely
@@ -65,17 +65,9 @@ bool IsLocalPeerTransfer(
   // using a nccl call but receiving from a local peer which is going through
   // cuda api, the deadlock could happen because nccl cannot ensure the
   // order of cuda api calls.
-  // We determine if it's a local peer by the following conditions:
-  // 1. Both source and target IDs are present and they are within a node
-  // 2. Source ID is present and is within the node, but target ID is not
-  // present.
-  // 3. Target ID is present and is within the node, but source ID is not
-  // present.
+  // We determine if it's a local peer if the source/target id is within a node
+  // if they are present.
   int64_t host_id = (current_id / device_count);
-  if (source_id && target_id) {
-    return (host_id == (*source_id / device_count)) &&
-           (host_id == (*target_id / device_count));
-  }
   if (source_id && host_id != *source_id / device_count) return false;
   if (target_id && host_id != *target_id / device_count) return false;
   return true;
@@ -190,7 +182,7 @@ absl::Status NcclCollectivePermuteStartThunk::RunNcclCollective(
   const NcclP2PConfig::SourceTargetMapEntry source_target =
       NcclP2PConfig::GetSourceTarget(config_.id_to_source_target, current_id);
   bool is_local_peer =
-      IsLocalPeerTransfer(source_target, stream, current_id, device_count_);
+      IsLocalPeerTransfer(source_target, current_id, device_count_);
   VLOG(5) << "Is local peer : " << (is_local_peer ? "true" : "false");
 
   bool use_memcpy = is_local_peer && recv_ptr_map_.IsInitialized(current_id) &&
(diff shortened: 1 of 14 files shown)


xla/stream_executor/stream_executor.h
```diff
@@ -324,6 +324,9 @@ class StreamExecutor {
   // Returns the memory limit in bytes supported by this executor.
   virtual int64_t GetMemoryLimitBytes() const = 0;
 
+  // Returns the total number of compute capable devices.
+  virtual int64_t GetDeviceCount() { return 0; };
```
frgossen, line 328: Why is this returning 0?

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -53,6 +53,31 @@ absl::StatusOr<const int64_t> GetCurrentId(
           : current_logical_id.computation_id;
   return current_id;
 }
+
+bool IsLocalPeerTransfer(
+    const NcclP2PConfig::SourceTargetMapEntry& source_target,
+    se::Stream& stream, const int64_t current_id, const int64_t device_count) {
+  const std::optional<int64_t> source_id = source_target.source;
+  const std::optional<int64_t> target_id = source_target.target;
+  // Since mixing nccl p2p with p2p memcopy will cause random deadlocks.
```
frgossen, line 62: Can you elaborate where you see deadlocks? Will this also happen if some nodes decide that their permute is within a host and some other nodes decide that their permute is across hosts?

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -53,6 +53,31 @@ absl::StatusOr<const int64_t> GetCurrentId(
...
+  // We determine if it's a local peer by the following conditions:
+  // 1. Both source and target IDs are present and they are within a node
+  // 2. Source ID is present, but target ID is not.
+  // 3. Target ID is presetn, but source ID is not.
```
frgossen, line 66: nit: present

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -53,6 +53,31 @@ absl::StatusOr<const int64_t> GetCurrentId(
...
+
+bool IsLocalPeerTransfer(
+    const NcclP2PConfig::SourceTargetMapEntry& source_target,
+    se::Stream& stream, const int64_t current_id, const int64_t device_count) {
+  const std::optional<int64_t> source_id = source_target.source;
+  const std::optional<int64_t> target_id = source_target.target;
+  // Since mixing nccl p2p with p2p memcopy will cause random deadlocks.
+  // We determine if it's a local peer by the following conditions:
+  // 1. Both source and target IDs are present and they are within a node
+  // 2. Source ID is present, but target ID is not.
```
frgossen, line 65: What does it mean if the source ID is present, but target ID is not? You can effectively not tell if this is within a host, can you?

frgossen (changes requested):

xla/stream_executor/stream_executor.h
Tixxx: since this is the base class for all back-end executors, i gave it the default behavior to return 0 since no back-end library is available at this point. we can also change this to -1 for an invalid value.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: the deadlocks could happen when calling nccl call and cuda memcpy p2p together(which both are synchronizing devices), in this case if this rank is sending across host using a nccl call but receiving from a local peer which is going through cuda api, the deadlock could happen because nccl cannot ensure the order of cuda api calls.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: updated.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: it means this rank is only receiving from a peer but not sending data to anybody. We can still use the source id and current id to determine that this is a local peer or not. Since there's no destination we need to send, it's still a local copy.

frgossen (changes requested):
Thanks for explaining the deadlock in more detail. Left a few comments and added @klucke  for the stream executor aspect of this PR

xla/stream_executor/stream_executor.h
Tixxx: yea , i notice that it returns 0 when the query of device count fails, so it makes sense to not return 0 here.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: sure updated.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -53,6 +53,31 @@ absl::StatusOr<const int64_t> GetCurrentId(
...
+    return (host_id == (*source_id / device_count)) &&
+           (host_id == (*target_id / device_count));
+  }
+  if (source_id) {
+    return (host_id == (*source_id / device_count));
+  }
+  if (target_id) {
+    return (host_id == (*target_id / device_count));
+  }
+  return false;
```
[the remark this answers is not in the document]
Tixxx: Yes, i simplified it a bit.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: added

xla/stream_executor/stream_executor.h
```diff
@@ -324,6 +324,11 @@ class StreamExecutor {
   // Returns the memory limit in bytes supported by this executor.
   virtual int64_t GetMemoryLimitBytes() const = 0;
 
+  // Returns the total number of compute capable devices.
+  virtual absl::StatusOr<int64_t> GetDeviceCount() {
```
klucke, line 328: There's already an interface defined in xla/stream_executor/platform.h to expose this number.  Platform::VisibleDeviceCount.  There are several places in XLA where this is already used.  It doesn't belong on StreamExecutor or its derived classes.

Tixxx: Thanks for the pointer. I updated the pr to use that and removed the new additions to SE.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -133,6 +161,12 @@ NcclCollectivePermuteStartThunk::NcclCollectivePermuteStartThunk(
 absl::Status NcclCollectivePermuteStartThunk::Initialize(
     const InitializeParams& params) {
   TF_RETURN_IF_ERROR(NcclCollectiveThunk::Initialize(params));
+  const se::Platform* platform = params.executor->GetPlatform();
+  CHECK(platform != nullptr);
+  device_count_ = platform->VisibleDeviceCount();
```
klucke, line 166: There could be several "visible" devices that aren't actually being targeted by the running XLA instance.  Is it always ok to assume that NCCL collectives are using all visible devices?

Tixxx: good point, i see that in pjrt client it always uses visible devices to build the distributed topology. But if xla decides to not use all visible devices, this logic in the pr wont be correct. Do you know if there's any way to get the set of targeted devices?

klucke: Maybe use LocalClient::device_count or Backend::device_count?  I think either of those would be correct.  (Backend::device_count is I think derived from LocalClient::device_count)

Tixxx: OK seems like those options are not easily accessible in gpu executable. I added a device count property in ExecutableRunOptions, and the thunk will just read the value from there.

Tixxx:
@frgossen @klucke  do you have more comments about this pr?

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -133,6 +161,11 @@ NcclCollectivePermuteStartThunk::NcclCollectivePermuteStartThunk(
 absl::Status NcclCollectivePermuteStartThunk::Initialize(
     const InitializeParams& params) {
   TF_RETURN_IF_ERROR(NcclCollectiveThunk::Initialize(params));
+  const se::Platform* platform = params.executor->GetPlatform();
```
klucke, line 164: I think the platform is no longer used.

klucke (commented):
I'll let frgossen comment on the collective portions.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: removed.

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -53,6 +53,34 @@ absl::StatusOr<const int64_t> GetCurrentId(
...
+  // 3. Target ID is present and is within the node, but source ID is not
+  // present.
+  int64_t host_id = (current_id / device_count);
+  if (source_id && target_id) {
+    return (host_id == (*source_id / device_count)) &&
+           (host_id == (*target_id / device_count));
+  }
+  if (source_id && host_id != *source_id / device_count) return false;
+  if (target_id && host_id != *target_id / device_count) return false;
+  return true;
```
frgossen, line 81: Same comment as on the earlier PR. I think this can be simplified. You can check src and target ID separately, if present. Also the comment can be simpler in the same way. 
You really just care if you're sending to a device within the same hiost or receiving from one or both, no?

frgossen (changes requested):
One comment. Looks good otherwise :)

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: forgot to remove the redundant check above. Updated the comment.

Tixxx:
@frgossen do you have more comments on this?

frgossen (approved):
Thanks!

sgerrard:
This code is not tested and it doesn't look like it is impacting any current tests (none were altered). Can you figure out a way to add testing?

Tixxx:
> This code is not tested and it doesn't look like it is impacting any current tests (none were altered). Can you figure out a way to add testing?

This can only be tested through multi-node setup and by looking at the profile to determine that p2p mem copy api has been called. It doesnt seem like there's existing infra in xla to test it. This was validated both manually and for e2e model benchmarking.
For unit testing, I think a better way is to have a jax unit test to run a cross-node collective permute with jax profiler. I can look into creating a jax unit test pr.

frgossen:
> > This code is not tested and it doesn't look like it is impacting any current tests (none were altered). Can you figure out a way to add testing?
> 
> This can only be tested through multi-node setup and by looking at the profile to determine that p2p mem copy api has been called. It doesnt seem like there's existing infra in xla to test it. This was validated both manually and for e2e model benchmarking. For unit testing, I think a better way is to have a jax unit test to run a cross-node collective permute with jax profiler. I can look into creating a jax unit test pr.

I agree with that atm. We are working on adding that sort of test infra but its not there yet. For the moment, we will see this in our benchmarks.

frgossen (changes requested):

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -133,6 +153,10 @@ NcclCollectivePermuteStartThunk::NcclCollectivePermuteStartThunk(
 absl::Status NcclCollectivePermuteStartThunk::Initialize(
     const InitializeParams& params) {
   TF_RETURN_IF_ERROR(NcclCollectiveThunk::Initialize(params));
+  device_count_ = params.local_device_count;
+  CHECK(device_count_ > 0);
```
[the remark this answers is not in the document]
Tixxx: updated

frgossen (approved):

PR #15636, opened 2024-08-01
[GPU] Expose caching DebugOptions to python

trevor-m:
Adds python bindings for `xla_gpu_kernel_cache_file`, `xla_gpu_enable_llvm_module_compilation_parallelism` and `xla_gpu_per_fusion_autotune_cache_dir`.

We would like to add some convenience  features to JAX which will enable all caches with one flag/option (will open PR for that soon). This change is necessary for that.

### DIFF
--- a/xla/python/xla_client_test.py
+++ b/xla/python/xla_client_test.py
@@ -3154,7 +3154,10 @@ module @jit__lambda_ attributes {mhlo.num_partitions = 1 : i32,
                          getattr(restored.executable_build_options, name),
                          msg=name)
 
-      for name in ("xla_cpu_enable_fast_math", "xla_test_all_input_layouts"):
+      for name in ("xla_cpu_enable_fast_math", "xla_test_all_input_layouts",
+                   "xla_gpu_kernel_cache_file",
+                   "xla_gpu_enable_llvm_module_compilation_parallelism",
+                   "xla_gpu_per_fusion_autotune_cache_dir"):
         self.assertEqual(
             getattr(options.executable_build_options.debug_options, name),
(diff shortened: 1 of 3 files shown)


cheshire (changes requested):
Is it possible to test those?

cheshire:
@skye @hawkinsp for PjRT API changes

trevor-m:
> Is it possible to test those?

Thanks for reviewing, added test to `xla_client_test`.

xla/python/xla_client_test.py
```diff
@@ -3137,6 +3137,9 @@ def testCompileOptionsSerialization(self):
       executable_build_options.num_partitions = 2
       executable_build_options.debug_options.xla_cpu_enable_fast_math = True
       executable_build_options.debug_options.xla_test_all_input_layouts = True
+      executable_build_options.debug_options.xla_gpu_kernel_cache_file = "/foo/bar"
```
skye, line 3140: I think you actually need to test these flags by adding them to L3157

skye (approved):
Looks good, please address the test comment though

xla/python/xla_client_test.py
trevor-m: Thanks for the review! Fixed this.

skye (approved):
Sweet thanks!

Uh @cheshire should you get this in or should I? Not sure how to do xla PRs but I can try...

nouiz:
What can we do to get this PR merged?
@cheshire @skye

cheshire (approved):

PR #15752, opened 2024-08-06
[XLA:GPU] Skip command buffer update for multi run from loop

shawnwang18:


NaiyerRizz:
Hi @shawnwang18 
Internal Cl are failing, Can you please look into it.
Thank you

ezhulenev:
This seems to be too brittle, in inference we might have multiple run ids executed concurrently. The problem is in `force_graph_update` and NCCL?

shawnwang18:
Ah, yes, it has a problem in inference run.  We might ends to a situation that some ranks was running the update, but some ranks do not, this is an issue for nccl update.

PR #15827, opened 2024-08-07
Do not default to sm_90a on non-Hopper platforms.

dimvar:


dimitar-asenov (changes requested):
I know this is a minor change, but please add a test.

dimvar:
Can you say more what kind of test you'd expect? This change is a no-op for publicly available platforms. It only takes effect on Blackwell onward.

dimitar-asenov:
Sorry I forgot to mention this.

A simple unit test of the `GetSmName` function would do. That function simply takes a compute capability as input and outputs a string. It should be trivial to simply call this function with a few different compute capabilities and ensure that the correct string is returned. The fact that some CCs are unreleased shouldn't matter.

You will also have to expose this function in the header file to make it testable, that should be fine.

dimvar:
Thanks for the clarification. Added a test. PTAL.

dimitar-asenov (approved):

dimitar-asenov:
Thank you for adding the test, much appreciated.

LGTM.

dimvar:
feedback/copybara failed but the "details" goes to an google-internal link to a CL. Can someone take a look?

dimitar-asenov:
I did not manage to submit this yesterday (Zurich time) and am going away for a week today. 

I believe the issue is that the test target in the `BUILD` file is missing a dependency on:

```
"//third_party/tensorflow/compiler/xla/stream_executor:device_description",
```

@golechwierowicz Could you please take care of submitting this?

PR #15831, opened 2024-08-07
Remove dup third_party/py/ml_dtypes as it is not used.

apivovarov:
Remove duplicate location of ml_dtypes - `third_party/py/ml_dtypes` as it is not used.

The actively used ml_dtypes is located in `third_party/tsl/third_party/py/ml_dtypes`


apivovarov:
Hi David, I think this PR may have been missed. Could you take a look? @ddunl

ddunl:
Ah it was, sorry about that. In principle this PR is correct, but the change do it internally is a bit awkward as this is generated in some sense by our Copybara config file. It will be fixed once TSL has been completely moved inside XLA which I'm currently working on. If there's a specific reason though why it would be helpful to not keep the duplicate around I can import this and then patch with the Copybara change. Let me know what you think!

apivovarov:
I'm working on adding support for the float8_e4m3 type in [ml_dtypes PR-161](https://github.com/jax-ml/ml_dtypes/pull/161/) and XLA. While doing so, I noticed that `ML_DTYPES_COMMIT` is defined in two places within XLA, but only one of them is actually used. This isn't a blocker for me; I just decided to clean it up.

Since the duplication will eventually be addressed, I think this PR can be closed.

PR #15839, opened 2024-08-07
[XLA:GPU] Bump minimum flash attention cuDNN version to 9.0.0

Cjkkkk:
* Bump flash attention minimum cudnn version to 9.0.0. This will simplify flash attention runtime and allow more upcoming enhancement like graph serialization. This also aligns with XLA's goal to deprecate cuDNN < 9.0.0 in the future.
* Add compute capability 8.6 and 8.9 support. (8.6/8.9 not supported with fused attn which is removed long time ago, 8.6 and 8.9 does support flash attention).
* Removed 2 tests for cuDNN < 8.9.6 where seq has to be multiple of 64. Such requirements aren't needed with >= 9.0.0.


Cjkkkk:
@sergachev

akuegel (approved):

PR #15399, opened 2024-07-27
[GPU] Add a pass converting specific custom calls to custom fusions.

sergachev:
This will let JAX users run chosen computations as cuDNN kernels.

PR #15861, opened 2024-08-08
Roll forward PR #15399

sergachev:
The previous version failed under `bazel test -c dbg ...`, the new one does not.


akuegel (approved):
Thank you :)

PR #15879, opened 2024-08-08
gemv rewrite pass independent of triton

wenscarl:
In the decoding stage of some MOE model inferences, XLA squeezes dimensions of size 1 when sequence length is 1. For example, it transforms a shape of [1, 4096] into [4096], resulting in a GEMV operation. When Triton GEMM is disabled, the GEMV rewriter is ineffective, which leads to failures in rewriting FP8 GEMM operations.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -1391,11 +1391,12 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
     const auto* cuda_cc = std::get_if<se::CudaComputeCapability>(&gpu_version);
     const auto* rocm_cc = std::get_if<se::RocmComputeCapability>(&gpu_version);
 
+    pipeline.AddPass<GemvRewriter>();
+
     if (debug_options.xla_gpu_enable_triton_gemm() &&
         ((cuda_cc != nullptr &&
           cuda_cc->IsAtLeast(se::CudaComputeCapability::AMPERE)) ||
          rocm_cc != nullptr)) {
-      pipeline.AddPass<GemvRewriter>();
       pipeline.AddPass<GemmFusion>(gpu_version);
     }
 
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -1391,13 +1391,17 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
     const auto* cuda_cc = std::get_if<se::CudaComputeCapability>(&gpu_version);
     const auto* rocm_cc = std::get_if<se::RocmComputeCapability>(&gpu_version);
 
-    pipeline.AddPass<GemvRewriter>();
-
-    if (debug_options.xla_gpu_enable_triton_gemm() &&
-        ((cuda_cc != nullptr &&
+    if (((cuda_cc != nullptr &&
           cuda_cc->IsAtLeast(se::CudaComputeCapability::AMPERE)) ||
          rocm_cc != nullptr)) {
-      pipeline.AddPass<GemmFusion>(gpu_version);
+      if (debug_options.xla_gpu_enable_triton_gemm()) {
+        pipeline.AddPass<GemvRewriter>();
+        pipeline.AddPass<GemmFusion>(gpu_version);
+      } else if (cuda_cc != nullptr &&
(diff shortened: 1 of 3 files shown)


anlunx (changes requested):
Why would we need GemvRewriter when triton is not enabled? You mentioned without GemvRewriter there would be a failure from rewriting fp8 gemms, could you add a test for that?

wenscarl:
> Why would we need GemvRewriter when triton is not enabled? You mentioned without GemvRewriter there would be a failure from rewriting fp8 gemms, could you add a test for that?

In the generating phase of inference with seq_len being 1, it's possible see  [m x n] @ [n] where seq_len dim gets squeezed by XLA passes. When fp8 quantization with direct quantization pattern(q-dot-dq) comes to play, triton gemm is often disabled. In such case, there is be error:
```
 Unexpected GEMM dtype f16 f16 bf16
 ```
 as a result of pattern matching failure from gemm rewriter and float normalization(fp8 type to f16). As for the test, gemv_rewriter_test already covers the functionality regardless of triton gemm. Where do you suggest to put the test?

**Review by anlunx — APPROVED** [review]

**Comment by penpornk** [comment]
@wenscarl XLA Linux CPU CI failed with the new test. Could you please help take a look? [Full log here](https://btx.cloud.google.com/invocations/7bfade3e-f460-4b40-abb7-9aabcbf8cb83/targets/tensorflow%2Fxla%2Flinux%2Fcpu%2Fbuild_cpu/log).

```
[ RUN      ] GemvRewriterTest.RewriteMatrixVectorMultiplicationToGemmPipeline
xla/service/gpu/transforms/gemv_rewriter_test.cc:64: Failure
Value of: (*result)->entry_computation()->root_instruction()
Expected: an HloInstruction:
 * with opcode bitcast AND
 * with operand 0 which is:
     an HloInstruction:
      * with opcode dot AND
      * with operand 0 which is:
          an HloInstruction with opcode parameter AND
      * with operand 1 which is:
          an HloInstruction:
           * with opcode bitcast AND
           * with operand 0 which is:
               an HloInstruction:
                * with opcode parameter AND
                * outputting
                    a shape:
                     * with element type F32 AND
                     * with dimensions [7] AND
           * outputting
               a shape:
                * with element type F32 AND
                * with dimensions [7,1] AND
 * outputting
     a shape:
      * with element type F32 AND
      * with dimensions [32]
  Actual: 0x55e0f13bf9b0 (of type xla::HloInstruction*), HloInstruction doesn't have opcode bitcast
in d = f32[32]{0} dot(f32[32,7]{1,0} p0, f32[7]{0} p1), lhs_contracting_dims={1}, rhs_contracting_dims={0}
[  FAILED  ] GemvRewriterTest.RewriteMatrixVectorMultiplicationToGemmPipeline (14 ms)
```

**Comment by wenscarl** [comment]
@penpornk could you help with the bazel test command to reproduce the error? 
I tried with the following bazel test command from the CI log,
```
bazel test --build_tag_filters=-no_oss,-gpu,-requires-gpu-nvidia,-requires-gpu-amd --test_tag_filters=-no_oss,-gpu,-requires-gpu-nvidia,-requires-gpu-amd --config=warnings --config=nonccl --config=rbe_linux_cpu --test_output=errors --verbose_failures --keep_going --nobuild_tests_only --profile=profile.json.gz --flaky_test_attempts=3 --jobs=150 --bes_upload_mode=fully_async -- //xla/
```
but seeing
```
ERROR: Error initializing RemoteModule
```
thanks!

bchetioui:
Hi @wenscarl,

We don't actually intend for the path of disabling Triton GEMMs to be supported in the longer term, so while fixing an immediate issue is fine in spirit, I'd like to understand why we need to do that to begin with. Could you please explain what is the issue that makes it necessary for you to disable Triton GEMMs? Thanks!

dimitar-asenov:
Hi @wenscarl, If you still intend to merge this PR, could you please provide the explanation that Benjamin asked for?

Also, I think the code in question might have changed since you originally created the PR. Please take a look and close the PR if it's no longer relevant.

wenscarl:
Apologies for the late response. Since #15879 and [Support fp8 direct quantization](https://github.com/google/praxis/pull/69) have been merged, the fp8 direct quantization pattern is now in use with Triton always enabled. As a result, this PR is no longer necessary and will be closed.

PR #15933, opened 2024-08-09
[XLA:GPU] remove some redundant mutex lock in fusion analysis cache invalidation

Cjkkkk:


olegshyshkov (changes requested):
Since we're changing the thread-compatibility expectations of the class, could you please update the comment in `fusion_analysis_cache.h`? To say that `Get` can be called concurrently, but `Invalidate` and `Clear` shouldn't.

Cjkkkk:
> Since we're changing the thread-compatibility expectations of the class, could you please update the comment in `fusion_analysis_cache.h`? To say that `Get` can be called concurrently, but `Invalidate` and `Clear` shouldn't.

Sure, I dont see them getting called concurrently anyway so i think we should just remove it.

olegshyshkov (approved):
Thank you!

PR #15987, opened 2024-08-12
[XLA:CPU][oneDNN] Add Bias and Binary Add fusions for oneDNN Convolutions

akhilgoe:
This PR adds support for Bias Add and Binary Add fusions with oneDNN convolution. In addition, it adds tests to test the functionality.


xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -586,148 +599,154 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
...
+    HloInstruction* new_instr;
+    // If matched pattern has custom-call -> bitcast -> add, then we need to
+    // insert bitcast after the new fusion to maintain the correct shape
+    // (new-custom-call -> bitcast). Also, this will optionally be followed
+    // by -> convert for bf16 case to avoid datatype mismatch.
+    if (optional_contraction_bitcast != nullptr &&
+        optional_contraction_bitcast->opcode() == HloOpcode::kBitcast) {
+      if (optional_contraction_convert != nullptr &&
+          optional_contraction_convert->opcode() == HloOpcode::kConvert) {
+        auto bitcast_call =
```
sdasgup3, line 723: Given

```c++
 bitcast_call = custom_call->AddInstruction(HloInstruction::CreateBitcast(
                ShapeUtil::ChangeElementType(
                    instr->shape(), custom_call->shape().element_type()),
                custom_call));
```

As the result shape of the `bitcast_call->shape->element_type == custom_call->shape->element_type`, can we conclude that  the dimensions of `instr->shape` == dimensions of `custom_call->shape` ? If yes, isn't the `bitcast_call` created here is a no-op.  

I was expecting something like following here

```c++
auto bitcast_call = custom_call->AddInstruction(
            HloInstruction::CreateBitcast(instr->shape(), custom_call));
```

akhilgoe: The dimensions of ```instr->shape()``` might not always be equal to the dimensions of ```custom-call->shape()```. A hypothetical example:

``` 

op1               op2                                                           op1               op2                                         op1               op2              addend 
 |                 |                                                             |                 |                                           |                 |                 |
 |_________________|                                                             |_________________|                                           |_________________|_________________|
          |                                                                               |                                                                      |
          |                                                                               |                                                                      |
     custom-call                                                                     custom-call          addend                                            custom-call    
      [1024,100]                                                                      [1024,100]           [100]                                            [1024,100]
          |                                                                               |                 |                                                    |
          |                                                                               |_________________|                                                    |
       bitcast    (broadcasted) addend                    =======>                                 |                        =======>                          bitcast
     [32,32,100]       [32,32,100]                                                                 |                                                        [32,32,100]
          |                 |                                                                  add (instr)
          |_________________|                                                                  [1024,100]             
                   |                                                                               | 
                   |                                                                               |
               add (instr)                                                                      bitcast
               [32,32,100]                                                                    [32,32,100]

                                                                        Note: oneDNN custom-calls can absorb non-broadcasted addends
     
```


Also, what you suggested in the last line is exactly what we have for the non-BF16 case.
https://github.com/openxla/xla/blob/8d10441bcf6c2e23fb2e17e6321bc31e29f0772b/xla/service/cpu/onednn_contraction_rewriter.cc#L734-L735


 The BF16 case gets a little tricky because currently we don't support elemental arithmetic operations like Add in the lower precision, as a result they are upcated to F32. Hence the pattern looks something like this:
Here, we need to change the element type of the generated ```bitcast``` so that it matches the ```custom-call``` dtype.
``` 

op1               op2                                        op1               op2              addend 
 |                 |                                          |                 |                 |
 |_________________|                                          |_________________|_________________|
          |                                                                     |
          |                                                                     |
     custom-call                                                           custom-call    
   BF16 [1024,100]                        =======>                       BF16 [1024,100]
          |                                                                     |
          |                                                                     |
       convert                                                               bitcast
    F32 [1024,100]                                                       BF16 [32,32,100]
          |                                                  
          |
       bitcast    (broadcasted and upcasted) addend                                                
    F32 [32,32,100]    F32 [32,32,100]                                                        
          |                 |                                                                  
          |_________________|                                                                             
                   |                                                                            
                   |                                                                            
               add (instr)                                                                
             F32 [32,32,100]                                                                  
                   |
                   |                                                     
                convert
            BF16 [32,32,100] 
```

Please let me know if this helps answer your questions.

thomasjoerg:
@penpornk Can you take a look or dispatch please?

sdasgup3 (approved):

PR #15998, opened 2024-08-12
[GPU] Fix kernel cache for loaded executables.

sergachev:
LoadCache() has to be called also in GpuThunkAotCompilationResult::LoadExecutable() to properly initialize the state of the name uniquer in the IR emitter context on this path to execution.

This makes the XLA kernel cache compatible with the JAX module cache (example test: `bazel test --test_env=XLA_FLAGS="--xla_gpu_enable_llvm_module_compilation_parallelism --xla_gpu_kernel_cache_file=/dev/shm/xla.kernel.cache" tests/compilation_cache_test_gpu`).


cheshire:
In the commit message, could you elaborate on what is the user-observable delta? (e.g. a sequence of steps X used to result in an undesirable state Y and now results in Z). You have that in tests, but it's a bit hard to connect this to what is user-observable (do we need a JAX test?)

sergachev:
The failing JAX test was `bazel test --test_env=XLA_FLAGS="--xla_gpu_enable_llvm_module_compilation_parallelism --xla_gpu_kernel_cache_file=/dev/shm/xla.kernel.cache" tests/compilation_cache_test_gpu` - see https://github.com/google/jax/pull/22899#discussion_r1708026237

xla/service/gpu/compile_module_to_llvm_ir.cc
```diff
@@ -50,13 +50,8 @@ limitations under the License.
 #include "xla/service/gpu/gpu_constants.h"
 #include "xla/service/gpu/gpu_executable.h"
 #include "xla/service/gpu/gpu_memory_space_assignment.h"
-#include "xla/service/gpu/ir_emitter_context.h"
 #include "xla/service/gpu/ir_emitter_unnested.h"
 #include "xla/service/gpu/metrics.h"
-#include "xla/service/gpu/runtime/conditional_thunk.h"
-#include "xla/service/gpu/runtime/sequential_thunk.h"
-#include "xla/service/gpu/runtime/thunk.h"
-#include "xla/service/gpu/runtime/while_thunk.h"
```
cheshire, line 59: Should those cleanups be in this commit?

cheshire (approved):

xla/service/gpu/compile_module_to_llvm_ir.cc
sergachev: It's tempting to include such trivial small cleanups while changing the files for other reasons (also because for us, external developers, submitting changes is a slower process). I'll try to separate them in the future.

PR #16037, opened 2024-08-13
[ROCm] Get platform name from `PlatformUtil` helper

hsharsha:


derdrdirk (approved):
LGTM

xla/stream_executor/gpu/BUILD
```diff
@@ -811,7 +811,17 @@ xla_test(
         "//xla/tools/hlo_opt:gpu_specs/h100_sxm.txtpb",
         "//xla/tools/hlo_opt:gpu_specs/p100.txtpb",
         "//xla/tools/hlo_opt:gpu_specs/v100.txtpb",
-    ],
+    ]) + if_rocm_is_configured([
+        "//xla/tools/hlo_opt:gpu_specs/mi200.txtpb",
+    ]),
+    local_defines = if_cuda_is_configured([
+        'GPU_SPEC_FILE_NAMES=(std::string[]){\\"a100_pcie_80\\", \\"a100_sxm_40\\", \
+         \\"a100_sxm_80\\", \\"a6000\\", \\"h100_pcie\\", \\"h100_sxm\\", \\"p100\\", \\"v100\\"}',
```
penpornk, line 819: The defined value needs to be wrapped in a quote, or we'd get an error.
```
attribute local_defines: definition contains too many tokens (found 8, expecting exactly one)
```

```suggestion
        'GPU_SPEC_FILE_NAMES="(std::string[]){\\"a100_pcie_80\\", \\"a100_sxm_40\\", \
         \\"a100_sxm_80\\", \\"a6000\\", \\"h100_pcie\\", \\"h100_sxm\\", \\"p100\\", \\"v100\\"}"',
```

xla/stream_executor/gpu/gpu_device_info_test.cc
```diff
@@ -47,7 +45,7 @@ TEST(DeviceInfoTest, DeviceInfoMatches) {
   }
 
   TF_ASSERT_OK_AND_ASSIGN(Platform * platform,
-                          PlatformManager::PlatformWithName("CUDA"));
+                          PlatformManager::PlatformWithName(PLATFORM_NAME));
```
beckerhe, line 48: You can use the `PlatformUtil` class to obtain the currently available GPU platform.

There is an example in this test: https://github.com/openxla/xla/blob/3c40cddc40fc43b197374825d954ff806c607a26/xla/stream_executor/gpu/gpu_kernel_test.cc#L42

xla/stream_executor/gpu/BUILD
```diff
@@ -811,7 +811,17 @@ xla_test(
...
+         'PLATFORM_NAME=\\"CUDA\\"'
+    ]) + if_rocm_is_configured([
+        'GPU_SPEC_FILE_NAMES=(std::string[]){\\"mi200\\"}',
```
penpornk, line 822: Nit: This one was fine because the defined value doesn't have spaces in between. But maybe wrap them in quotes too just to be safe?
```suggestion
        'GPU_SPEC_FILE_NAMES="(std::string[]){\\"mi200\\"}"',
```

penpornk (changes requested):

beckerhe (changes requested):
Hey Harsha,
we are trying to get away from some many compile time switches and I believe your change doesn't actually need any of the `if_cuda_is_configured`/`if_rocm_is_configured` guards.

Wouldn't it just work if you added your spec file to the list of data files, no matter if it's a ROCm or CUDA build? In that case the test would loop over all the specs but that shouldn't be a problem right?

Avoiding too many different build configuration will help us greatly maintaining the build and it means that it's less likely that the ROCm code path will break in the future.

penpornk (approved):
Nice! Thank you! :)
Maybe update the PR title as well?

PR #16072, opened 2024-08-14
Do not fail for dot ops with E5M2 inputs and BF16 output

sergey-kozub:
Prior to this PR, the following HLO fails to run:

```
HloModule test
ENTRY test {
  x = f8e5m2[32,32] parameter(0)
  y = f8e5m2[32,32] parameter(1)
  ROOT dot = bf16[32,32] dot(x, y),
    lhs_contracting_dims={1}, rhs_contracting_dims={0}, algorithm=dot_any_f8_any_f8_f32
}
```

This works as (f16 f16 bf16) GEMM config is not supported by cuBLASlt, but (bf16 bf16 bf16) is.

### DIFF
--- a/xla/service/gpu/transforms/gemm_rewriter.cc
+++ b/xla/service/gpu/transforms/gemm_rewriter.cc
@@ -2204,16 +2204,20 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
     return f32_dot;
   }
 
-  // Turns an F8 dot into an F16 dot, converting operands to F16 and
+  // Turns an F8 dot into an F16 dot, converting operands to F16 (or BF16) and
   // converting the output back to F8.
   absl::StatusOr<HloInstruction *> TurnF8DotIntoF16Dot(HloInstruction *instr) {
     DCHECK(IsF8Type(instr->operand(0)));
     DCHECK(IsF8Type(instr->operand(1)));
 
-    // Convert operands to F16
+    // If the output type is BF16, the input types have to be BF16 as well.
+    PrimitiveType conv_type =
+        instr->shape().element_type() == BF16 ? BF16 : F16;
+
+    // Convert operands to F16 (or BF16).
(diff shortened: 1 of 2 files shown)


akuegel (commented):
Can you please add a test?

akuegel:
> The test is already there (dot_algorithm_support_test), it was just disabled as it was failing.

Ah, right, I didn't spot that this is a hopper specific fix, and changing the tag of the test to "gpu" re-enables it for hopper.

akuegel (approved):

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -2204,16 +2204,20 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
-  // Turns an F8 dot into an F16 dot, converting operands to F16 and
+  // Turns an F8 dot into an F16 dot, converting operands to F16 (or BF16) and
   // converting the output back to F8.
   absl::StatusOr<HloInstruction *> TurnF8DotIntoF16Dot(HloInstruction *instr) {
     DCHECK(IsF8Type(instr->operand(0)));
     DCHECK(IsF8Type(instr->operand(1)));
 
-    // Convert operands to F16
+    // If the output type is BF16, the input types have to be BF16 as well.
+    PrimitiveType conv_type =
```
reedwm, line 2214: Why call this `conv_type`?

reedwm, line 2214: Why call this `conv_type`?

sergey-kozub: Naming is hard. Please suggest a better name.

reedwm: Ah sorry I misinterpreted the `conv` as `convolution`. But now realize it means `convert`.

I think this is fine.

PR #16095, opened 2024-08-14
Create overloads of cub::ThreadLoadVolatilePointer instead of template specializations

dimvar:
This avoids a build failure with an upcoming CUDA change.
Commit https://github.com/NVIDIA/cccl/commit/6dfc8dddaef5f01adeb683790bf0c8ec05302460 changed the method signature and breaks our specializations.
Specializing cub:: templates is against the CCCL guidelines: https://github.com/NVIDIA/cccl?tab=readme-ov-file#compatibility-guidelines

The workaround here was suggested by someone on the CUDA team.
It is also against the guidelines, since it is adding to the cub:: namespace, but it fixes the compilation failure.
The more futureproof solution is to open a feature request with them (https://github.com/NVIDIA/cccl/issues) to support loading of `Eigen::half` and `tsl::bfloat16` in CUB directly.

### DIFF
--- a/xla/service/gpu/gpu_prim.h
+++ b/xla/service/gpu/gpu_prim.h
@@ -44,10 +44,9 @@ __device__ __forceinline__ void ThreadStoreVolatilePtr<Eigen::half>(
       Eigen::numext::bit_cast<uint16_t>(val);
 }
 
-template <>
-__device__ __forceinline__ Eigen::half ThreadLoadVolatilePointer<Eigen::half>(
-    Eigen::half *ptr, Int2Type<true> /*is_primitive*/) {
-  uint16_t result = *reinterpret_cast<volatile uint16_t *>(ptr);
+__device__ __forceinline__ Eigen::half ThreadLoadVolatilePointer(
+    const Eigen::half *ptr, Int2Type<true> /*is_primitive*/) {
+  uint16_t result = *reinterpret_cast<volatile const uint16_t *>(ptr);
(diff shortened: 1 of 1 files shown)


cheshire (approved):

sgerrard:
We need a test here, looks like there aren't tests impacted by this change, which means a test needs to be added.

dimvar:
@sgerrard This PR is fixing a compile time failure. The existing tests should be WAI. What kind of new test are you thinking?

xla/service/gpu/gpu_prim.h
```diff
@@ -44,10 +44,9 @@ __device__ __forceinline__ void ThreadStoreVolatilePtr<Eigen::half>(
       Eigen::numext::bit_cast<uint16_t>(val);
 }
 
-template <>
-__device__ __forceinline__ Eigen::half ThreadLoadVolatilePointer<Eigen::half>(
```
sgerrard, line 48: What about a unit test ensuring this works as expected?

dimvar: I'm not sure how to write one. I'm replacing the template specialization with an overload, but the body of the function stays the same. I can't think of a test that behaves differently before and after my patch.

If you have an idea for a good test, let me know. Otherwise, IMO the PR is OK to merge as-is.

sgerrard: Alright, this one is ok to submit.

PR #16134, opened 2024-08-15
[XLA:GPU] Enable --xla_gpu_enable_nccl_comm_splitting=true by default

trevor-m:
Now that the hang with NCCL comm split is fixed #15935, I'm resubmitting #11762 to enable nccl comm split by default which will reduce the amount of memory used by NCCL communicators.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -149,7 +149,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_nccl_termination_timeout_seconds(-1);
   opts.set_xla_gpu_enable_shared_constants(true);
   opts.set_xla_gpu_enable_nccl_user_buffers(false);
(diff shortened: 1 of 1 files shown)


ezhulenev (approved):

penpornk:
@trevor-m Do we have a unit test specific to comm splitting yet? If so, could you please give me the target path (so I can refer to it when merging)? If not, would you mind helping add one?

trevor-m:
> @trevor-m Do we have a unit test specific to comm splitting yet? If so, could you please give me the target path (so I can refer to it when merging)? If not, would you mind helping add one?

There is the [pmap_test ](https://github.com/google/jax/blob/main/tests/pmap_test.py) in JAX but I don't think XLA itself has any unit tests for NCCL communicators. @ezhulenev do you know?

penpornk (approved):
Re-approving to re-trigger PR import process.

PR #16146, opened 2024-08-16
Add more checks for multiplication results

knightXun:
Add positive checks for multiplication results

This commit introduces a new function that determines if the result of a multiplication is a positive number. It handles the cases where both operands are negative or both are positive, ensuring that the result is treated as a positive number if the product of two negatives is involved.

The function checks the sign of the operands and applies the mathematical rule that the product of two negative numbers is positive. This is crucial for ensuring the correctness of calculations in various scenarios, such as financial calculations or scientific computations where the sign of the result is significant.

### DIFF
--- a/xla/service/algebraic_simplifier.cc
+++ b/xla/service/algebraic_simplifier.cc
@@ -128,41 +128,6 @@ bool IsAnyOperandComplex(const HloInstruction* hlo) {
   return false;
 }
 
-bool IsPositive(const HloInstruction* hlo,
-                const AlgebraicSimplifierOptions& options) {
-  // Utility only handles real types.
-  if (IsAnyOperandComplex(hlo)) {
-    return false;
-  }
-  switch (hlo->opcode()) {
-    case HloOpcode::kGetTupleElement: {
-      const HloInstruction* gte_operand = hlo->operand(0);
-      switch (gte_operand->opcode()) {
-        case HloOpcode::kCustomCall: {
-          const auto& target = gte_operand->custom_call_target();
-          return target ==
-                     options.get_cudnn_batchnorm_forward_training_metadata() &&
-                 hlo->tuple_index() == 2;
-        }
-        default:
-          return false;
-      }
-    }
-    case HloOpcode::kPower:
-    case HloOpcode::kAbs:
-    case HloOpcode::kRsqrt:
-    case HloOpcode::kSqrt:
-      return IsPositive(hlo->operand(0), options);
-
-    case HloOpcode::kMultiply: {
-      return hlo->operand(0) == hlo->operand(1) &&
-             IsPositive(hlo->operand(0), options);
-    }
-    default:
-      return false;
-  }
-}
-
 static bool IsScalarConstant(const HloInstruction* hlo) {
   return hlo->opcode() == HloOpcode::kConstant &&
          ShapeUtil::IsEffectiveScalar(hlo->shape());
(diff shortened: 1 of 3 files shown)


NaiyerRizz:
Hi @knightXun 
Can you please add a test case for the same.
Thank you

NaiyerRizz:
Hi @knightXun 
Can you please look into it as checks are failing  
Thank you

knightXun:
> Hi @knightXun Can you please look into it as checks are failing Thank you

ok, I will do this today, thanks for the hint

knightXun:
@NaiyerRizz nicely ping.

knightXun:
@loislo nicely ping.

knightXun:
@NaiyerRizz @loislo nicely ping.

NaiyerRizz:
Hi @penpornk 
Can you please have a look on this PR
Thank you

knightXun:
> Hi @penpornk Can you please have a look on this PR Thank you

Yeah，asian helps asian

knightXun:
@beckerhe Can you help me review this PR?

beckerhe:
Hi,
I'm not familiar enough with the `AlgebraicSimplifier` to do an effective review.

@reedwm can you take a look or forward this to someone?

xla/service/algebraic_simplifier.cc
```diff
@@ -519,6 +484,53 @@ int64_t GetReduceFlops(const HloInstruction* reduce) {
...
+          return target ==
+                     options.get_cudnn_batchnorm_forward_training_metadata() &&
+                 hlo->tuple_index() == 2;
+        }
+        default:
+          return false;
+      }
+    }
+    case HloOpcode::kExp:
+      return true;
```
reedwm, line 508: Exp can be zero if the argument is negative infinity. Similarly, for the added `kMultiply` case where the LHS and RHS are positive, the result can underflow to zero.

Hmmm, but I see existing cases have similar problems, even without this PR. E.g. kPower can also be zero if the RHS is negative infinity. Perhaps we don't care about cases where underflow happen, or where arguments are infinity? @akuegel, do you have an opinion here?

xla/service/algebraic_simplifier.cc
```diff
@@ -519,6 +484,53 @@ int64_t GetReduceFlops(const HloInstruction* reduce) {
...
+        return IsPositive(hlo->operand(0), options);
+      }
+      if (IsPositive(hlo->operand(0), options) &&
+          IsPositive(hlo->operand(1), options)) {
+        return true;
+      }
+      if (!IsNonNegative(hlo->operand(0), options) &&
+          !IsNonNegative(hlo->operand(1), options)) {
+        return true;
+      }
```
reedwm, line 526: Remove this case. If either the LHS or RHS is zero, the output will be zero, not positive.

reedwm (changes requested):

xla/service/algebraic_simplifier.cc
akuegel: Sorry, I somehow missed this question. Good point, I think we should not return true for kExp. And same for kPower. We can return true for them in the IsNonNegative function.
Also, good catch regarding underflow to zero, I think we should also not return true for kMultiply.

akuegel (changes requested):
@knightXun please move these checks to the IsNonNegative function. Like @reedwm pointed out, we can have zero results.

dimitar-asenov:
@knightXun I see this PR has been inactive for a while. If you're still interested in merging it, could you please do the requested changes above? Otherwise, we will close it after some more time of inactivity.

dimitar-asenov:
Closing this PR due to inactivity. If you're still interested in merging this, please update the PR and reopen it.

PR #13779, opened 2024-06-14
[ROCM] added memory management functions

pemeliya:
Here we enable XLA-managed workspace buffer for rocblas

@xla-rotation: would you please have a look ?

PR #16149, opened 2024-08-16
[ROCM] disabling setting workspace for rocblas

pemeliya:
After enabling workspaces on ROCM in https://github.com/openxla/xla/pull/13779, we have a problem with **//xla/tests:triangular_solve_test**. Note that, [triangular_solve_thunk](https://github.com/openxla/xla/blob/4cfd0f84e1fbd5e92c7b961d59650474e7f310a4/xla/service/gpu/runtime/triangular_solve_thunk.cc#L151) does **not** set a workspace while calling DoBlasTrsmBatched functions which is probably not needed on CUDA side.

Apparently, rocblas has a different behaviour: since rocblas_set_workspace was not called before TRSM routine, rocblas will try to reuse **previously set** workspace which might not be enough for that triangular solve. I fix it by setting workspace to **nullptr** if it is not explicitly provided (thereby forcing rocblas to use its internal scratch buffer). 

**UPDATE:** this feature seems to produce sporadic errors since rocblas_set_workspace uses blocking memory functions to manipulate scratch memory (which do not work under stream capture). **Disabled** until we have a clear idea on how to make it working on ROCM.

### DIFF
--- a/xla/stream_executor/rocm/rocm_blas.cc
+++ b/xla/stream_executor/rocm/rocm_blas.cc
@@ -359,24 +359,27 @@ absl::Status ROCMBlas::DoBlasInternalImpl(FuncT rocblas_func, Stream *stream,
 
   // set the atomics mode, leaving default to library
   bool allow_atomics = !OpDeterminismRequired();
-  rocblas_status ret;
   if (!allow_atomics) {
-    ret = wrap::rocblas_set_atomics_mode(blas_, rocblas_atomics_not_allowed);
+    auto ret = wrap::rocblas_set_atomics_mode(blas_, rocblas_atomics_not_allowed);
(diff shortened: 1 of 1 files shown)


sgerrard:
The changes look fine, but you have some build failures for AMD ROCm? Can you address those?

pemeliya:
Sure, we currently have some build-brakes on AMD side, once these are fixed, I will rebase my PR ..

i-chaochen:
@pemeliya @sgerrard this ROCm CI build break is already fixed last Friday and you just need to rebase it https://github.com/openxla/xla/pull/16152

sgerrard (approved):

loislo (approved):

PR #16153, opened 2024-08-16
[GPU] Change sharded GEMM autotuning timeout from infinity to 24 hours.

sergachev:
Fixes a problem with MPI: https://github.com/google/jax/issues/22995

### DIFF
--- a/xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
+++ b/xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
@@ -1201,7 +1201,7 @@ absl::Status ExchangeResults(KeyValueStoreInterface& key_value_store,
         std::string autotune_results_str,
         key_value_store.Get(
             absl::StrFormat("%s_%d_%d", kKeyPrefix, module_id, i),
-            absl::InfiniteDuration()));
+            absl::Hours(24)));
     TF_RETURN_IF_ERROR(
         AutotunerUtil::LoadAutotuneResults(autotune_results_str, true));
   }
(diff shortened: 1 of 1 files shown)


bchetioui (approved):
Thanks!

xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
```diff
@@ -1201,7 +1201,7 @@ absl::Status ExchangeResults(KeyValueStoreInterface& key_value_store,
         std::string autotune_results_str,
         key_value_store.Get(
             absl::StrFormat("%s_%d_%d", kKeyPrefix, module_id, i),
-            absl::InfiniteDuration()));
+            absl::Hours(24)));
```
nouiz, line 1204: Maybe add a comment to tell infinity cause issues with MPI, so that isn't added back later?

penpornk: @sergachev If you'd like at make another commit to add comment, I can wait for that before merging the PR.

sergachev: I don't see why would anyone try to increase it beyond 24 hours and don't see a necessity to document all possible values that would not work here.

bchetioui: > I don't see why would anyone try to increase it beyond 24 hours

Someone could do it just because `absl::InfiniteDuration()` seems to better capture the intention. There is also no theoretical limit for how long exhaustive autotuning can take, so documenting that we don't choose `absl::InfiniteDuration()` for a specific reason does seem worth doing to me.

sergachev: OK

bchetioui (approved):

xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
bchetioui: Thanks Ilia!

PR #16241, opened 2024-08-20
Align the scheduling name with the instruction name for constants

zhenying-liu:
This commit corrects the scheduling name after it is altered by SanitizeConstantNames, preventing assertion failures in the HLO verifier.

### DIFF
--- a/xla/service/gpu/transforms/sanitize_constant_names.cc
+++ b/xla/service/gpu/transforms/sanitize_constant_names.cc
@@ -64,6 +64,7 @@ absl::StatusOr<bool> SanitizeConstantNames::Run(
       // Register this new name with the module's instruction_name_uniquer to
       // avoid name collision that might happen in future.
       module->instruction_name_uniquer().GetUniqueName(instr->name());
+      instr->set_metadata_scheduling_name(instr->name());
       changed = true;
     }
   }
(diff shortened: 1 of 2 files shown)


dimitar-asenov (approved):

akuegel:
It seems that during the rematerialization pipeline, there is another name change that will trigger a verifier error:

```
FAILED_PRECONDITION: during context [hlo verifier]: Expected instruction name to remain the same. Was 'constant_1834_1' is 'constant_1834_1.1'.

Failed after remat-pipeline
```

golechwierowicz:
This should not be needed after https://github.com/openxla/xla/commit/ca5545b2585e151b243e1b94b39f0ff126536e35.

golechwierowicz (changes requested):
This should not be needed after https://github.com/openxla/xla/commit/ca5545b2585e151b243e1b94b39f0ff126536e35.

zhenying-liu:
> This should not be needed after [ca5545b](https://github.com/openxla/xla/commit/ca5545b2585e151b243e1b94b39f0ff126536e35).

Great to have a fix. Will close this (I found I didn't update the unit test with the scheduling names, but won't fix as it's not needed).

PR #16438, opened 2024-08-24
aarch64: implement onednn matmul operator with explicit reorders

snadampal:
I have added a new function for aarch64 mainly because the changes are spread all over and keeping it common with the existing one looked very convoluted.


snadampal:
**Alternative solutions tried:** I have tried to handle it at HLO rewrite stage by introducing `__onednn$matmul_reorder` customcall before `__onednn$matmul` but looks like the reorder custom call is not fully supported, so, it didn't lead to any working solution. 
If the reorder customcall is supported, please let me know if there is any reference on how to insert it into the HLO graph.

snadampal:
Appreciate if someone can review and provide feedback.

xla/service/cpu/onednn_matmul.cc
```diff
@@ -262,6 +262,160 @@ CreateOneDnnPrimDesc<dnnl::matmul::primitive_desc>(HloInstruction* instr) {
                               fused_shapes, matmul_config);
 }
 
+#ifdef __aarch64__
+ABSL_ATTRIBUTE_NO_SANITIZE_MEMORY void __xla_cpu_runtime_OneDnnMatMul(
```
penpornk, line 266: This function has a lot of code similar to the x86 one. Could we use a shared function and refactor just the code that differs to functions (and selectively call them from the shared function)?

penpornk (changes requested):
Very sorry for the delay!

cc: @milpuz01 @TensorFlow-MKL

xla/service/cpu/onednn_matmul.cc
snadampal: thanks, @penpornk , as I mentioned earlier, i tried to combine, but it became too convoluted. let me revisit it to see if I can split it into utility functions and reuse.

penpornk: Thank you! I think we can consolidate them more. Could you please help try if the following code works? Only one `__xla_cpu_runtime_oneDnnMatMul` function, no `#ifdef`s.
```c++
ABSL_ATTRIBUTE_NO_SANITIZE_MEMORY void __xla_cpu_runtime_OneDnnMatMul(
    void* result, void* scratch, void** args) {
  // args[0]: ptr to nargs
  // args[1]: ptr to ExecutableRunOptions
  // args[2]: ptr to OneDnnMatMulConfig
  // args[3...]: ptrs to operands
  int arg_indx = 0;
  const int64_t num_args = *(static_cast<int64_t*>(args[arg_indx++]));

  const xla::ExecutableRunOptions* run_options =
      static_cast<const xla::ExecutableRunOptions*>(args[arg_indx++]);
  auto thread_pool = CreateOneDnnThreadPool(
      run_options ? run_options->intra_op_thread_pool() : nullptr);
  engine cpu_engine(engine::kind::cpu, 0);
  auto onednn_stream = MakeOneDnnStream(cpu_engine, thread_pool.get());

  std::string config_str(static_cast<const char*>(args[arg_indx++]));
  OneDnnMatMulConfig matmul_config;
  matmul_config.ParseFromString(config_str);

  MemrefInfo input_minfo(args[arg_indx++]);
  MemrefInfo weights_minfo(args[arg_indx++]);
  MemrefInfo output_minfo(result);

  auto input_md = input_minfo.GetOneDnnMemDesc();
  auto weights_md = weights_minfo.GetOneDnnMemDesc();
  // Input and weights memory::desc need to be in correct layout before matmul
  // primitive descriptor is created.
  TRANSPOSE_LAST_TWO_DIMS_IF(
      matmul_config.transpose_a() && input_md.get_ndims() > 1, input_md);
  TRANSPOSE_LAST_TWO_DIMS_IF(
      matmul_config.transpose_b() && weights_md.get_ndims() > 1, weights_md);
  auto output_md = output_minfo.GetOneDnnMemDesc();
  auto weight_format = tsl::port::IsAarch64CPU() ? memory::format_tag::any
                                                 : memory::format_tag::ab;
  void* rhs_data = weights_minfo.Data();

  if (matmul_config.optimization_config().weights_prepacked()) {
    // Weight pre-packing is supported for 2D weights only.
    // Since prepacked weights array is flattened, try to infer the dims from
    // input and output.
    // TODO(intel-tf): Add support for prepacked weights for higher then 2D
    // array.
    weights_md =
        memory::desc({input_md.get_dims().back(), output_md.get_dims().back()},
                     weights_md.get_data_type(), weight_format);
  } else if (tsl::port::IsAarch64CPU()) {
    // Weights are not pre-packed, and this scenario requires
    // weights reordering on aarch64
    auto bias_md = dnnl::memory::desc{};
    if (absl::c_count(matmul_config.fusions().ops(), OneDnnFusionConfig::BIAS) >
        0) {
      MemrefInfo bias_minfo(args[arg_indx++]);
      bias_md = bias_minfo.GetOneDnnMemDesc();
    }
    auto reordered_weight_md = OneDnnMatMulOptWeightsDesc(
        cpu_engine, input_md, weights_md, bias_md, output_md);
    auto reordered_weight_shape =
        MemDescToXlaShapeFlattened(reordered_weight_md);
    rhs_data = Literal(reordered_weight_shape).untyped_data();

    auto weight_mem =
        dnnl::memory{weights_md, cpu_engine, weights_minfo.Data()};
    auto reordered_weight_mem =
        dnnl::memory{reordered_weight_md, cpu_engine, rhs_data};

    dnnl::reorder rdr{weight_mem, reordered_weight_mem};
    rdr.execute(onednn_stream, weight_mem, reordered_weight_mem);
    onednn_stream.wait();
    weights_md = reordered_weight_md;
  }
  const int64_t num_fused_operands = num_args - arg_indx;
  std::vector<memory::desc> fused_mds;
  std::vector<void*> fused_bufs;
  for (int64_t i = 0; i < num_fused_operands; ++i) {
    MemrefInfo operand_minfo(args[arg_indx++]);
    fused_mds.push_back(operand_minfo.GetOneDnnMemDesc());
    fused_bufs.push_back(operand_minfo.Data());
  }

  std::vector<std::pair<int, dnnl::memory>> postop_args;
  FusedOperandsRef fused_operands_ref{fused_bufs, postop_args};
  auto matmul_pd =
      CreateMatMulPrimDesc(cpu_engine, input_md, weights_md, output_md,
                           fused_mds, matmul_config, &fused_operands_ref);

  XLA_LIGHTWEIGHT_CHECK(num_args == arg_indx);

  auto lhs_mem = memory(input_md, cpu_engine, input_minfo.Data());
  auto rhs_mem =
      memory(matmul_pd->weights_desc(), cpu_engine, rhs_data);
  auto result_mem = memory(output_md, cpu_engine, output_minfo.Data());

  if (std::strstr(matmul_pd->impl_info_str(), "ref") != nullptr) {
    LOG(WARNING) << "[Perf]: MatMul reference implementation being executed";
  }

  auto matmul_prim = matmul(*matmul_pd);

  std::unordered_map<int, memory> matmul_args{{DNNL_ARG_SRC, lhs_mem},
                                              {DNNL_ARG_WEIGHTS, rhs_mem},
                                              {DNNL_ARG_DST, result_mem}};

  if (matmul_config.optimization_config().user_scratchpad()) {
    XLA_LIGHTWEIGHT_CHECK(scratch != nullptr);
    MemrefInfo scratch_minfo(scratch);
    auto scratchpad_md = matmul_pd->scratchpad_desc();
    auto scratch_mem = memory(scratchpad_md, cpu_engine, scratch_minfo.Data());
    matmul_args.insert({DNNL_ARG_SCRATCHPAD, scratch_mem});
  }

  matmul_args.insert(postop_args.begin(), postop_args.end());

  matmul_prim.execute(onednn_stream, matmul_args);
}
```

third_party/tsl/third_party/compute_library/set_openmp_runtime_off_by_default.patch
penpornk, line 1: https://github.com/tensorflow/tensorflow/pull/76831 has similar changes. It was [reverted](https://github.com/tensorflow/tensorflow/pull/77091) because it broke our ARM build (missing OpenMP symbols). I think we should isolate just this change to a separate PR and try to get it in first.

penpornk (changes requested):

third_party/tsl/third_party/compute_library/set_openmp_runtime_off_by_default.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/BUILD.bazel b/BUILD.bazel
+index f1766d9582..849a991cea 100644
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -56,7 +56,7 @@ bool_flag(
+ 
+ bool_flag(
+     name = "openmp",
+-    build_setting_default = True,
++    build_setting_default = False,
```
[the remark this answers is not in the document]
milpuz01: Hi @penpornk , as far as I recall for TF we have enabled to use custom scheduler as in https://github.com/oneapi-src/oneDNN/blob/main/src/cpu/aarch64/acl_threadpool_scheduler.cpp in order to be able to use Eigen threadpool so it doesn't really matter whether we build ACL with OpenMP or pthreads as custom scheduler will be used. I believe that ACL requires to be built with at least of one non-custom schedulers, but it doesn't end up being used.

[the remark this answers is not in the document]
snadampal: Hi @penpornk , @milpuz01 , this default setting is overwritten with the threadpool bazel config, so, runtime there is no issue. however, this default setting is unnecessarily creating linker dependency on libomp. which is creating issues with importing the TF library into other projects (for example TF Java), and in this JAX case, it was misleading. that's why I wanted to set it to off by default.

third_party/tsl/third_party/mkl_dnn/onednn_acl_add_sbgemm_matmul_primitive_definition.patch
```diff
@@ -0,0 +1,47 @@
...
+                               weights_md()->data_type, dst_md()->data_type)
+                     && platform::has_data_type_support(data_type::f16);
++            const bool is_fp32_bf16_ok
++                    = (utils::everyone_is(data_type::f32, src_md()->data_type,
++                               dst_md()->data_type, desc()->accum_data_type)
++                            && platform::has_data_type_support(data_type::f32)
++                            && utils::everyone_is(
++                                    data_type::bf16, weights_md()->data_type)
++                            && platform::has_data_type_support(
++                                    data_type::bf16));
```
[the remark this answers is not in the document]
snadampal: Arm team is working on oneDNN upgrade for TF. My understanding is once it's done for TF, it will be carried to OpenXLA.

third_party/tsl/third_party/mkl_dnn/onednn_acl_allow_blocked_weight_format_for_matmul_primitive.patch
```diff
@@ -0,0 +1,103 @@
...
+             const bool is_fp32_ok
+                     = utils::everyone_is(data_type::f32, src_md()->data_type,
+                               weights_md()->data_type, dst_md()->data_type,
+@@ -76,18 +78,20 @@ struct acl_matmul_t : public primitive_t {
+                     = utils::everyone_is(data_type::f16, src_md()->data_type,
+                               weights_md()->data_type, dst_md()->data_type)
+                     && platform::has_data_type_support(data_type::f16);
++            const bool is_weights_md_format_ok
++                    = utils::one_of(weights_format_kind_received,
++                            format_kind::any, format_kind::blocked);
```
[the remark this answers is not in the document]
snadampal: yes, here is the draft: https://github.com/oneapi-src/oneDNN/pull/2068

snadampal:
Hi @penpornk , thanks for your review.  Regarding the code consolidation, the main challenge to add the extra reorder logic as a conditional check is with the scope of the reordered weights memory descriptor.  in fact my first version was a  simple if/else, but later I had to change it to keep the reorder and primitive execute together to avoid destroying the reordered weights before the primitive is executed. If there is a way to handle the underlying buffer scope, we can attempt this approach. Please let me know if I'm missing anything here.

`Literal(reordered_weight_shape).untyped_data();`

snadampal:
Hi @penpornk , I have addressed the comments. regarding code refactoring and consolidation, I have created the reordered weights `Literal` unconditionally. Please review and check if this is acceptable for x86 platform. If not acceptable, separating the aarch64 and x86 logic is the only option I see. 
As you suggested I have raised another PR for ACL default settings for openmp: https://github.com/openxla/xla/pull/18488

xla/service/cpu/onednn_matmul.cc
```diff
@@ -302,6 +302,30 @@ ABSL_ATTRIBUTE_NO_SANITIZE_MEMORY void __xla_cpu_runtime_OneDnnMatMul(
...
       matmul_config.transpose_b() && weights_md.get_ndims() > 1, weights_md);
   auto output_md = output_minfo.GetOneDnnMemDesc();
+
+  // TODO(snadampal): allocate the reordered weights Literal conditionally
+  // only when the weights are already not pre-packed and only for aarch64
+  // currently it's allocated upfront because Literal object scope issues
+  auto bias_md = dnnl::memory::desc{};
+  if (absl::c_count(matmul_config.fusions().ops(), OneDnnFusionConfig::BIAS) >
+        0) {
+      MemrefInfo bias_minfo(args[arg_indx++]);
```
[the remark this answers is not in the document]
snadampal: Hi @akhilgoe , this is conditionally done only is there was a bias operand. so, the args array would have all the required indices.
to make sure i understand your point, could you please elaborate it?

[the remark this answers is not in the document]
akhilgoe: Sure, what I mean to say is that this `if-block` is only being used to get the bias md, but it has a side-effect, it is incrementing `arg_indx`, iow `arg_indx` now points to the argument after bias, and since you are not storing the pod data anywhere my concern is that the bias data is lost.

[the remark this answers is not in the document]
snadampal: I see your point now, thank you! I have updated the PR. Since the reorder is an additional primitive i have introduced, I'm just skipping the  index increment, so that the actual matmul primitive processing is not affected.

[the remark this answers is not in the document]
snadampal: Hi @akhilgoe, please let me know if this addresses your comment.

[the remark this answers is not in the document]
akhilgoe: Yes it does, thanks! 
Can you please run a model that will trigger the MM + Bias ( + activations) scenario and / or run tests under xla/service/cpu/tests/onednn_matmul_test.cc and confirm that everything runs as expected, thanks!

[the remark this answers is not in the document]
snadampal: Hi @akhilgoe , I have updated the PR and with this version all the unit tests are passing on AWS Graviton4 based c8g instances. appreciate if you could review one more time.

note: the one test failure from `onednn_matmul_test` ([  FAILED  ] MatmulTest.TestF32ConstantWeights (77 ms))
 is not related to this PR, it's failing on the mainline code as well, so, planning to take that up separately instead of blocking this PR.

~~~
//xla/service/cpu/tests:cpu_bytesizeof_test                     (cached) PASSED in 0.2s
//xla/service/cpu/tests:onednn_matmul_test                  (3/4 cached) FAILED in 1 out of 4 in 13.1s
  Stats over 4 runs: max = 13.1s, min = 2.0s, avg = 5.5s, dev = 4.5s
  /home/ubuntu/.cache/bazel/_bazel_ubuntu/6d062a7b894a0860cec805fab2fe3ab8/execroot/xla/bazel-out/aarch64-opt/testlogs/xla/service/cpu/tests/onednn_matmul_test/shard_1_of_4/test.log
//xla/service/cpu/tests:cpu_aot_export_test                              PASSED in 1.4s
//xla/service/cpu/tests:cpu_dyn_shape_test                               PASSED in 1.4s
//xla/service/cpu/tests:cpu_eigen_dot_operation_test                     PASSED in 1.3s
//xla/service/cpu/tests:cpu_external_constants_test                      PASSED in 1.3s
//xla/service/cpu/tests:cpu_fusion_test                                  PASSED in 1.2s
//xla/service/cpu/tests:cpu_infeed_test                                  PASSED in 1.2s
//xla/service/cpu/tests:cpu_intrinsic_test                               PASSED in 1.5s
//xla/service/cpu/tests:cpu_key_value_sort_test                          PASSED in 1.4s
//xla/service/cpu/tests:cpu_literal_caching_test                         PASSED in 1.4s
//xla/service/cpu/tests:cpu_noalias_test                                 PASSED in 1.2s
//xla/service/cpu/tests:cpu_outfeed_test                                 PASSED in 1.4s
//xla/service/cpu/tests:cpu_profiling_test                               PASSED in 1.4s
//xla/service/cpu/tests:cpu_spmd_compile_test                            PASSED in 1.4s
//xla/service/cpu/tests:cpu_topk_test                                    PASSED in 1.4s
//xla/service/cpu/tests:cpu_vectorization_test                           PASSED in 1.4s
//xla/service/cpu/tests:cpu_while_test                                   PASSED in 1.4s
//xla/service/cpu/tests:onednn_convolution_test                          PASSED in 1.4s
//xla/service/cpu/tests:onednn_layer_norm_test                           PASSED in 3.0s
//xla/service/cpu/tests:tree_reduction_rewriter_test                     PASSED in 1.4s
//xla/service/cpu/tests:onednn_softmax_test                              PASSED in 3.2s
  Stats over 4 runs: max = 3.2s, min = 1.3s, avg = 1.8s, dev = 0.8s

Executed 21 out of 22 tests: 21 tests pass and 1 fails locally.
~~~

snadampal:
Hi @penpornk , @akhilgoe , this version addresses all your feedback and also unit test failures, could you please review and approve if it looks good.

snadampal:
Hi @akhilgoe , @penpornk ,
I have also analyzed the `TestF32ConstantWeights` test failure and added my analysis in this issue: https://github.com/openxla/xla/issues/19416
It looks like a false failure, and anyway not related to this PR, so this can be merged independent of that.

snadampal:
Hi @akhilgoe , I have updated the PR to make it acl backend (based on bazel build config), so, it shouldn't add any overhead for x86. Could you please check if this looks okay for x86?

akhilgoe:
Thanks @snadampal! I can confirm that the regressions that we saw before are resolved now.

snadampal:
thanks for checking @akhilgoe. btw, now the cpu unit tests are passing 100% on AWS Graviton based EC2 instances. The rebase fixed the 1 failing test I observed before.

`bazel test --config mkl_aarch64_threadpool //xla/service/cpu/tests:*`
~~~
INFO: Build completed successfully, 8771 total actions
//xla/service/cpu/tests:cpu_aot_export_test                              PASSED in 1.5s
//xla/service/cpu/tests:cpu_bytesizeof_test                              PASSED in 0.2s
//xla/service/cpu/tests:cpu_dyn_shape_test                               PASSED in 1.5s
//xla/service/cpu/tests:cpu_eigen_dot_operation_test                     PASSED in 1.4s
//xla/service/cpu/tests:cpu_external_constants_test                      PASSED in 1.3s
//xla/service/cpu/tests:cpu_fusion_test                                  PASSED in 1.3s
//xla/service/cpu/tests:cpu_infeed_test                                  PASSED in 1.3s
//xla/service/cpu/tests:cpu_intrinsic_test                               PASSED in 1.6s
//xla/service/cpu/tests:cpu_key_value_sort_test                          PASSED in 1.4s
//xla/service/cpu/tests:cpu_literal_caching_test                         PASSED in 1.4s
//xla/service/cpu/tests:cpu_noalias_test                                 PASSED in 1.3s
//xla/service/cpu/tests:cpu_outfeed_test                                 PASSED in 1.4s
//xla/service/cpu/tests:cpu_profiling_test                               PASSED in 1.6s
//xla/service/cpu/tests:cpu_spmd_compile_test                            PASSED in 1.5s
//xla/service/cpu/tests:cpu_topk_test                                    PASSED in 1.4s
//xla/service/cpu/tests:cpu_vectorization_test                           PASSED in 1.6s
//xla/service/cpu/tests:cpu_while_test                                   PASSED in 1.6s
//xla/service/cpu/tests:onednn_convolution_test                          PASSED in 1.7s
//xla/service/cpu/tests:onednn_layer_norm_test                           PASSED in 3.1s
//xla/service/cpu/tests:tree_reduction_rewriter_test                     PASSED in 1.5s
//xla/service/cpu/tests:onednn_matmul_test                               PASSED in 5.7s
  Stats over 4 runs: max = 5.7s, min = 2.4s, avg = 4.0s, dev = 1.5s
//xla/service/cpu/tests:onednn_softmax_test                              PASSED in 3.4s
  Stats over 4 runs: max = 3.4s, min = 1.3s, avg = 1.9s, dev = 0.9s

Executed 22 out of 22 tests: 22 tests pass.
~~~

akhilgoe (approved):
Looks good to me from an x86 perspective, thank you!

snadampal:
Hi @penpornk , given it's approved for x86, I think there is no blocker for this, please let me know if you have any more comments.

snadampal:
Hi @penpornk, @Adam-Banas , this PR is ready for your final review and merge. It has been validated and approved from x86 and AWS Graviton CPUs. I have also rebased it to the latest commit.

ps: `xla.proto` file which is being changed actively by others and making me to rebase this PR every time. appreciate if this can be merged sooner.

xla/debug_options_flags.cc
```diff
@@ -877,6 +880,11 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
                 bool_setter_for(&DebugOptions::set_xla_cpu_use_mkl_dnn),
                 debug_options->xla_cpu_use_mkl_dnn(),
                 "Generate calls to MKL-DNN in the CPU backend."));
+  flag_list->push_back(
+      tsl::Flag("xla_cpu_use_mkl_dnn_with_acl",
+                bool_setter_for(&DebugOptions::set_xla_cpu_use_mkl_dnn_with_acl),
+                debug_options->xla_cpu_use_mkl_dnn_with_acl(),
+                "Generate calls to MKL-DNN using ACL (Arm Compute Library) in the CPU backend."));
```
penpornk, line 887: I don't think we need a `DebugOptions` flag for this. We just need to check whether the kernel is running on an aarch64 CPU. Please see my other comment on onednn_matmul.cc.

xla/service/cpu/onednn_matmul.cc
```diff
@@ -230,8 +236,49 @@ ABSL_ATTRIBUTE_NO_SANITIZE_MEMORY void __xla_cpu_runtime_OneDnnMatMul(
     // array.
     weights_md =
         memory::desc({input_md.get_dims().back(), output_md.get_dims().back()},
-                     weights_md.get_data_type(), memory::format_tag::ab);
+                     weights_md.get_data_type(), weight_format);
+  } else if (matmul_config.optimization_config().with_acl_backend()) {
```
penpornk, line 240: You can call `tsl::port::IsAarch64CPU()` from cpu_info.h instead.
https://github.com/openxla/xla/blob/29715b5b586204349a3dfbc6806f094133e8eff6/third_party/tsl/tsl/platform/cpu_info.h#L164

penpornk (changes requested):
I'm very sorry for the delay! I had been on vacation. 
I think the aarch64 check can just use an existing function instead of adding a new flag. 

The PR seems to have conflicts. Could you please help merge conflicts again?

xla/debug_options_flags.cc
snadampal: Hi @penpornk, the extra reorder is required only for ACL backend. So, it would be more appropriate to check for ACL backend instead of generic aarch64 platform check.

snadampal:
Hi @penpornk , I hope you had a fantastic time off! I made explicit ACL backend checks because the reorder is a requirement for the ACL kernels.  if this looks good, I will rebase and resolve the conflicts.

xla/debug_options_flags.cc
penpornk: `__xla_cpu_runtime_oneDNNMatMul` is only called from oneDNN integration code, and we always use ACL for oneDNN aarch64 backend. So I think it's fine to just detect aarch64 CPU.

xla/service/cpu/onednn_matmul.cc
penpornk: Please also add a comment:
```
// TODO(penporn): Make this code work with AOT cross-compilation, e.g., using
// x86 CPU to compile for aarch64 CPU, by passing the target machine
// architecture as a function parameter.
```

penpornk (changes requested):
@snadampal Thank you! :D
I still think it's fine to check if we are running on aarch64 CPU because of the following reasons:
1. `__xla_cpu_runtime_OneDnnMatMul` is only called from oneDNN integration code.
2. Our oneDNN integration for aarch64 backend always use ACL. 

What do you think?

In the future, we will need to change this check a bit to support ahead-of-time compilation for different architectures as well (e.g., using x86 CPU to compile for aarch64 -- is this case `IsAarch64CPU()` would return false, but we want it to be true). But this will require some more changes and will delay the PR further. So I think for now we can just say cross-compiling is not a supported use case for the oneDNN custom calls and just add a TODO.

snadampal:
>__xla_cpu_runtime_oneDNNMatMul is only called from oneDNN integration code, and we always use ACL for oneDNN aarch64 backend. So I think it's fine to just detect aarch64 CPU.

I have seen ArmPL also as a oneDNN backend for arm64, that's why I made it ACL specific. and also, ACL check might solve the AOT cross-compilation right, @penpornk ? 

but, if Open XLA is limiting the backend to ACL alone, I will simplify the code to check for arm64 platform.

snadampal:
adding @milpuz01 to comment on oneDNN backend options for arm64. Milos, please check the above thread.

milpuz01:
> adding @milpuz01 to comment on oneDNN backend options for arm64. Milos, please check the above thread.

There are at the moment two different ways to build oneDNN on AArch64 with and without ACL. When building with ACL we are using fixed format GEMM kernels that expect caller to reorder data before calling GEMM. So from my understanding conditional should be for ACL check and that should be sufficient, because if oneDNN is built without ACL it will fall to reference kernels and they do not need reorder.

penpornk:
@snadampal @milpuz01 Thank you for the quick reply!

Looking at XLA's config file:
https://github.com/openxla/xla/blob/e68425dd41bae0df8b28cb33089201d5129169fe/.bazelrc#L208-L217

`--config=mkl_aarch64` always build oneDNN with ACL. 

I had assumed that `--config=mkl_aarch64_threadpool` also uses ACL, but it seems it's not defined in the config. Can we just use ACL with `mkl_aarch64_threadpool` too? 

If building with ACL doesn't have any performance disadvantage (compared to building without ACL), it's better to just have one option (build with ACL).

penpornk:
@snadampal @milpuz01 I just dug a bit further, any configs with `--define=build_with_mkl_aarch64=true` also uses ACL:
1. The C define is guarded by `if_mkldnn_aarch64_acl`: 
https://github.com/openxla/xla/blob/e68425dd41bae0df8b28cb33089201d5129169fe/xla/tsl/tsl.bzl#L330
2. `if_mkldnn_aarch64_acl` is true whenever `build_with_mkl_aarch64` is defined.
https://github.com/openxla/xla/blob/e68425dd41bae0df8b28cb33089201d5129169fe/xla/tsl/mkl/build_defs.bzl#L145

Since both oneDNN aarch64 build configs use ACL, we can just check for aarch64 CPUs in oneDNN custom call.

snadampal:
Hi @penpornk , I have updated the PR to check for aarch64 instead of ACL. This is based on the alignment that XLA will be shipped with ACL for aarch64 builds.

penpornk (approved):
Thank you very much again for the changes!

snadampal:
Hi @NaiyerRizz, the PR has been tested and approved. Please let me know the next steps to get this merged.

snadampal:
Hi @penpornk , @NaiyerRizz , I see there are few commits from this PR still not merged. without them, the performance will regress on aarch64 linux platform. Can you please merge the rest of the commits as well?

third_party/tsl/third_party/mkl_dnn/onednn_acl_add_sbgemm_matmul_primitive_definition.patch
penpornk, line 1: This patch could not be applied to the source file. Did you generate the patch for oneDNN v3.2.1? OpenXLA is [using v3.2.1](https://github.com/openxla/xla/blob/33d3624f89d1f3a19f78868cc19b9791cc7323a4/third_party/tsl/workspace2.bzl#L169) for aarch64 backend. Could you please update the patch file?

penpornk (commented):
Apologies about the missing files. One of the patch files could not be applied. Could you please help update it? I'll add all the missing files after this patch file is fixed.

third_party/tsl/third_party/mkl_dnn/onednn_acl_add_sbgemm_matmul_primitive_definition.patch
snadampal: Yes, @penpornk , I created all the oneDNN patches on top of v3.2.1. (on top of commit: `7f38289d0e959ff69528cee456049d383885323a (tag: v3.2.1)`)

I'm able to apply the remaining 4 commits cleanly in the order mentioned here. Could you please share the errors you were getting while applying the patch?

~~~
commit f31c92afb4d7d421efef284ba47404e866a7a9c4 (HEAD -> aarch64_xla)
Author: Sunita Nadampalli <nadampal@amazon.com>
Date:   Fri Nov 8 18:08:58 2024 +0000

    onednn acl: add sbgemm definition for matmul primitive

commit c0fc5b9ccc4070051b13d6f29c0269f0eb1a343f
Author: Sunita Nadampalli <nadampal@amazon.com>
Date:   Fri Nov 8 18:07:54 2024 +0000

    onednn acl: add bf16 platform support check

commit 812668e3121347a7c8baaf3cefae9813990a9d4e
Author: Sunita Nadampalli <nadampal@amazon.com>
Date:   Fri Nov 8 18:05:21 2024 +0000

    onednn acl: fx segfault during post op execute

commit 612d01228645d46571b5847ee216fd11fd88e52c
Author: Sunita Nadampalli <nadampal@amazon.com>
Date:   Fri Oct 18 19:58:25 2024 +0000

    onednn acl: add blocked layout format support for matmul weight tensors

commit 30b32502b41f1b50e1ec0a068201bb6d19a88a62 (origin/main, origin/HEAD, main)
Author: Oleg Shyshkov <shyshkov@google.com>
Date:   Wed Dec 11 05:52:49 2024 -0800
~~~

penpornk: Thank you for the quick reply! The patches were just applied in the wrong order. They work now.
I believe all the missing files have gone in (https://github.com/openxla/xla/pull/20447). Could you please help double-check if things work as intended now?

snadampal: thanks, @penpornk , all the changes are merged and the performance is as expected!

penpornk: Thank you for checking! :)

penpornk:
Thank you very much for checking, @snadampal! :)

PR #16486, opened 2024-08-26
Remove copy-start from stream attribute annotator

zhenying-liu:
This change removes the copy-start instruction handling from StreamAttributeAnnotator, which was originally implemented in PR 10636. This removal is a direct consequence of PR 13597, which moved the responsibility of stream assignment for copy-start instructions to ExecutionStreamAssignment.


ezhulenev (approved):

ezhulenev:
I think that `StreamAttributeAnnotator` can be safely removed, because it never really worked and with `ExecutionStreamAssignment` + async thunks we have a correct solution for same problem

NaiyerRizz (approved):

Issue #16431, opened 2024-08-24
CUDA 12.6 uses PTX ISA version 8.5 not 8.6


sclarkson:
While running JAX compiled with CUDA 12.6, I get error messages to the effect of

```
'+ptx86' is not a recognized feature for this target (ignoring feature)
```

Per the table [here](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes-ptx-release-history), both CUDA 12.5 and 12.6 use PTX ISA 8.5, making the calculation below no longer correct.

https://github.com/openxla/xla/blob/ce2fffa8197cf8c8fd463b2f849e68746f4678f6/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc#L302-L318

cheshire:
@beckerhe WDYT?

beckerhe:
I never was a fan of the heuristic but I also don't have a good alternative approach.

For now I will cap the PTX version.

sclarkson:
Haven't tested but I'm reasonably sure that fixed it. Thanks.

PR #16504, opened 2024-08-27
[PTX] Use the correct ptx version (8.5) for CUDA 12.6

sergey-kozub:
For CUDA 12.6, PTX version is 8.5 (simple versioning rule heuristic used previously no longer works):
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes

For CUDA 12.6 and above, use the highest known PTX version (8.5).
Without this change, the following error is observed:
'+ptx86' is not a recognized feature for this target (ignoring feature)
See: https://github.com/openxla/xla/issues/16431

This PR also adds a basic test.

### DIFF
--- a/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
+++ b/xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
@@ -295,30 +295,6 @@ absl::Status NVPTXTargetModuleLinker(llvm::Module* module,
   return absl::OkStatus();
 }
 
-#ifdef GOOGLE_CUDA
-namespace {
-constexpr int kFallbackPtxVersion = 65;
-
-int DetermineHighestSupportedPtxVersionFromCudaVersion(
-    stream_executor::ToolVersion cuda_version) {
-  if (cuda_version[0] < 11) {
-    // For everything below CUDA 11 we just fall back to PTX 6.5.
-    // We don't support CUDA below 11 anymore.
-    return kFallbackPtxVersion;
-  }
-
-  // Mapping determined from
-  // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes
-  // Examples:
-  // CUDA 11.0 -> PTX 7.0
-  // CUDA 11.1 -> PTX 7.1
-  // CUDA 12.0 -> PTX 8.0
-  // CUDA 12.4 -> PTX 8.4 etc.
-  return (cuda_version[0] - 4) * 10 + cuda_version[1];
-}
-}  // namespace
-#endif
-
 std::unique_ptr<llvm::TargetMachine> NVPTXGetTargetMachine(
     llvm::Triple target_triple, se::CudaComputeCapability compute_capability,
     const DebugOptions& debug_options) {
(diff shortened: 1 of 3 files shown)


xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
```diff
@@ -313,8 +314,10 @@ int DetermineHighestSupportedPtxVersionFromCudaVersion(
   // CUDA 11.0 -> PTX 7.0
   // CUDA 11.1 -> PTX 7.1
   // CUDA 12.0 -> PTX 8.0
-  // CUDA 12.4 -> PTX 8.4 etc.
-  return (cuda_version[0] - 4) * 10 + cuda_version[1];
+  // CUDA 12.4 -> PTX 8.4
+  // This versioning scheme is valid until CUDA 12.6
+  int cuda_id = cuda_version[0] * 10 + cuda_version[1];
+  return cuda_id < 126 ? cuda_id - 40 : kMaxPtxVersion;
```
beckerhe, line 320: Can you do the comparison of Cuda versions in `ToolVersion` - like `cuda_version >= ToolVersion{12, 6}`?

Also can you put that into a separate `if`-statement that returns early? The current version is a bit hard to read?

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
```diff
@@ -708,6 +711,9 @@ absl::StatusOr<std::string> CompileToPtx(
   return ptx;
 }
 
+int PtxVersionFromCudaVersionForTest(se::ToolVersion tool_version) {
+  return DetermineHighestSupportedPtxVersionFromCudaVersion(tool_version);
+}
```
beckerhe, line 716: Why not expose the original function?

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib_test.cc
```diff
@@ -33,6 +38,44 @@ TEST(UtilsTest, TestGetSmName) {
   ASSERT_EQ(nvptx::GetSmName(cc_next), "sm_90");
 }
 
+struct VersionPair {
+  int ptx_version;
+  int cuda_version;
```
beckerhe, line 43: Can you use `ToolVersion` instead of an `int` for the CUDA version?

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib_test.cc
```diff
@@ -33,6 +38,44 @@ TEST(UtilsTest, TestGetSmName) {
...
+                             {77, 117},
+                             {78, 118},
+                             // CUDA 12
+                             {80, 120},
+                             {81, 121},
+                             {82, 122},
+                             {83, 123},
+                             {84, 124},
+                             {85, 125},
+                             {85, 126},
```
beckerhe, line 73: Nit: If you were reversing the order (CUDA version first, PTX version second), then it would feel more naturally like a mapping.

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
sergey-kozub: `ToolVersion` is defined in `xla/stream_executor/cuda/cuda_asm_compiler.h` (i.e. must live inside `#if GOOGLE_CUDA` conditional), which makes it hard to use. So I introduced another type alias, `nvptx::Version`, which is a simple int pair.

Replaced the `if` statements as you suggest above.

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc
sergey-kozub: Exposed the original function in the `nvptx` namespace.

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib_test.cc
sergey-kozub: ditto

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib_test.cc
sergey-kozub: Done.

beckerhe (approved):
Thank you!

sergey-kozub:
@beckerhe Why does github say "Merging is blocked (Merging can be performed automatically with 1 approving review)" after your approval?

beckerhe:
> @beckerhe Why does github say "Merging is blocked (Merging can be performed automatically with 1 approving review)" after your approval?

This is just because due to Copybara this is not the PR thet gets merged. The change is already being processed internally and should land in a few minutes.

PR #16520, opened 2024-08-27
[ROCM] ResetStream function for GemmAlgorithmPicker (BlasSupport interface)

pemeliya:
Here I added **ResetStream** function which sets the underlying stream for cublas/rocblas libraries to default stream 0.

This is useful for GemmAlgorithmPicker which uses a temporary stream object for autotuning. In rocblas, **rocblas_set_stream** function is **persistent**, meaning that once the stream value is set, it will be used in all subsequent computations until new stream value is set. 

In case of GemmAlgorithmPicker, we leave a **destroyed** stream object set into the math library. This does not produce any error behaviour but merely just a warning on ROCM side: "Stream Capture Check Failed".

With this new ResetStream function, one can reset the stream value in GemmAlgorithmPicker destructor. Potentially, it can also be useful in other places where temporary stream value is used.

Besides, I have also made some small code restructure for GemmAlgorithmPicker

@xla-rotation: could you have a look please?

### DIFF
--- a/xla/service/gpu/autotuning/gemm_algorithm_picker.cc
+++ b/xla/service/gpu/autotuning/gemm_algorithm_picker.cc
@@ -100,6 +100,8 @@ class GemmAutotuner {
   explicit GemmAutotuner(const AutotuneConfig& autotune_config)
       : autotune_config_(autotune_config) {}
 
+  const AutotuneConfig& config() const { return autotune_config_; }
+
   size_t num_algorithms_left() const { return num_algorithms_left_; }
 
   absl::StatusOr<AutotuneResult> operator()(const HloInstruction* gemm,
@@ -388,20 +390,18 @@ class GemmAutotuner {
                  << best.status();
     return AutotuneResult{};
   }  // GetBestAlgorithm
-};  // GemmAutotuner
+};  // class GemmAutotuner
 
 // Do Gemm Autotune without stream executor. Use results from autotune cache
 // only.
 absl::StatusOr<bool> RunOnInstruction(HloInstruction* gemm,
-                                      const AutotuneConfig& config,
-                                      size_t* num_algorithms_left) {
+                                      GemmAutotuner& autotuner) {
   VLOG(3) << "Loading the autotune result of GemmThunk " << gemm->ToString();
 
   GpuBackendConfig gpu_config =
       gemm->backend_config<GpuBackendConfig>().value();
   GemmBackendConfig& backend_config = *gpu_config.mutable_gemm_backend_config();
 
-  *num_algorithms_left = 0;
   // Degenerate gemms replaced with memzero operation, no need to auto tune it.
   if (backend_config.alpha_real() == 0.0 &&
       backend_config.alpha_imag() == 0.0 && backend_config.beta() == 0.0) {
@@ -409,13 +409,12 @@ absl::StatusOr<bool> RunOnInstruction(HloInstruction* gemm,
     return false;
   }
 
+  const AutotuneConfig& config = autotuner.config();
   AutotuneCacheKey key(config.GetModelStr(), *gemm);
-  GemmAutotuner autotuner(config);
   TF_ASSIGN_OR_RETURN(AutotuneResult algorithm,
                       AutotunerUtil::Autotune(
                           gemm, config, [&] { return autotuner(gemm, key); }));
 
-  *num_algorithms_left = autotuner.num_algorithms_left();
   auto old_algorithm = backend_config.selected_algorithm();
   bool update_algorithm =
       IsCublasLtMatmulF8(*gemm) ||
@@ -442,9 +441,8 @@ absl::StatusOr<bool> RunOnInstruction(HloInstruction* gemm,
 
     if (new_algorithm == old_algorithm &&
         backend_config.has_selected_algorithm()) {
-      // We don't need to update the backend config if
-      // the algorithm hasn't changed unless previously
-      // the algorithm wasn't set explicitly.
+      // We don't need to update the backend config if the algorithm was not 
+      // changed unless previously the algorithm wasn't set explicitly.
       return false;
     }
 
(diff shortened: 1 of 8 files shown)


xla/service/gpu/autotuning/gemm_algorithm_picker.cc
```diff
@@ -485,12 +488,14 @@ absl::StatusOr<bool> GemmAlgorithmPicker::Run(
     VLOG(2) << "GEMM auto-tuning disabled, GemmAlgorithmPicker returning early";
     return false;
   }
-
+  // Autotuner is created on demain only
```
beckerhe, line 491: on demand?

xla/service/gpu/autotuning/gemm_algorithm_picker.cc
```diff
@@ -98,6 +98,14 @@ class GemmAutotuner {
   explicit GemmAutotuner(const AutotuneConfig& autotune_config)
       : autotune_config_(autotune_config) {}
 
+  ~GemmAutotuner() {
+    if (stream_ != nullptr) {
+      if (auto blas = stream_->parent()->AsBlas()) blas->ResetStream();
+    }
+  }
+
+  const AutotuneConfig& config() { return autotune_config_; }
```
beckerhe, line 107: Can you make this getter `const`?

xla/service/gpu/autotuning/gemm_algorithm_picker.cc
```diff
@@ -98,6 +98,14 @@ class GemmAutotuner {
   explicit GemmAutotuner(const AutotuneConfig& autotune_config)
       : autotune_config_(autotune_config) {}
 
+  ~GemmAutotuner() {
+    if (stream_ != nullptr) {
+      if (auto blas = stream_->parent()->AsBlas()) blas->ResetStream();
+    }
```
beckerhe, line 104: I think what we actually want is that the stream notifies the Blas instance when it gets out of scope.

We already have a similar mechanism for DNN (For example: https://github.com/openxla/xla/blob/c2348031b2f8c7c1719eace85f112171f0e06d0a/xla/stream_executor/cuda/cuda_executor.cc#L499).

I'm not sure what architecture we should strive for here, so I'm going to ask a colleague who knows more about stream_executor and will come back to you ASAP.

beckerhe: Another thought: It seems we call `SetStream` in each call to `ROCMBlas::DoBlasInternalImpl`.
Couldn't we just reset the stream back to zero at the end of `ROCMBlas::DoBlasInternalImpl`?

This would be more in line with the API and less spaghetti-like communication between parts of the code base. I'm not sure though whether this has perfomance implications.

pemeliya: oh I think I like the first idea: we can add such a 'listener' function NotifyStreamDestroyed to Blas interface. Than we can get rid of a specific destructor for GemmAlgorithmPicker. I will try it out. Setting the stream to 0 after each call to DoBlasInternalImpl sounds less generic to me because we can potentially do more cleanup tasks in **NotifyStreamDestroyed** function (like for example, resetting user workspace bound to this stream?)

beckerhe: > Setting the stream to 0 after each call to DoBlasInternalImpl sounds less generic to me because we can potentially do more cleanup tasks in NotifyStreamDestroyed function (like for example, resetting user workspace bound to this stream?)

It's true that the listener approach is more generic, but it's also way more complicated. Therefore if the "Setting the stream to 0 after each call to DoBlasInternalImpl" has no serious downside, then I would highly prefer this approach over the other.

That of course doesn't mean that we can't introduce the listener approach in the future if there is a functionality that needs it - together with some proper testing of this mechanism.

xla/service/gpu/autotuning/gemm_algorithm_picker.cc
```diff
@@ -100,6 +100,8 @@ class GemmAutotuner {
   explicit GemmAutotuner(const AutotuneConfig& autotune_config)
       : autotune_config_(autotune_config) {}
 
+  const AutotuneConfig& config() const { return autotune_config_; }
+
```
beckerhe, line 104: With the recent change, the changes to `gemm_algorithm_picker.cc` are not needed anymore, are they?

xla/service/gpu/autotuning/gemm_algorithm_picker.cc
beckerhe: Sounds good. Thank you for making the change.

xla/service/gpu/autotuning/gemm_algorithm_picker.cc
pemeliya: yes I also did some small refactoring there: basically I moved the creation of **GemmAutotuner** object to the top function, so that it won't create a new stream for each new custom-call instruction found in the HLO module. And also made the code a bit less verbose

beckerhe (approved):
Thank you!

pemeliya:
Hello, I wonder if there anything still needs to be done on this PR ? Is it pending an internal review ?

akuegel:
> Hello, I wonder if there anything still needs to be done on this PR ? Is it pending an internal review ?

It was blocked because there is no test change. Is it possible to test this?

dimitar-asenov (changes requested):
Please add a test that verifies the changes.

pemeliya:
> Please add a test that verifies the changes.

Yep, I will think about this: basically the changes to GemmAlgorithmPicker are NFC so they shall already be covered by gemm_algorithm_picker_test. But for reset stream, I will figure out some gemm_rewriter subtest

pemeliya:
In order to test it I had to add a test method **IsMainStreamSet** to the blas interface.. Basically what it does, is returns true if the stream currently used by the blas API is stream 0 (main). I cannot figure out anything more elegant here.

dimitar-asenov (approved):

dimitar-asenov:
This looks fine to me, but I'd like @beckerhe to weigh in on the blas interface change.

xla/stream_executor/blas.h
```diff
@@ -222,6 +222,10 @@ class BlasSupport {
 
   virtual gpu::BlasLt *GetBlasLt() = 0;
 
+  // For tests only: sets *is_main_stream to true if the underlying Blas library 
+  // has stream 0 set as its current stream.
+  virtual bool IsMainStreamSet(bool *is_main_stream) = 0;
```
beckerhe, line 227: Can you make this return a `absl::StatusOr<bool>` instead of using an out-parameter?

beckerhe: And can you make this function `const`? (This probably requires marking the mutex in the implementations as `mutable`.

beckerhe (commented):
Soory for the delayed response. I added a minor comment. Thank you for adding a test.

xla/stream_executor/blas.h
pemeliya: yep, I have changed the interface accordingly: initially I just wanted to keep it similar to all other BlasSupport functions..

beckerhe (approved):
Thank you!

PR #16527, opened 2024-08-27
Enable the use of ACL operations in an mkl_aarch64 build

Rohanjames1997:
1. Enables the `xla_cpu_use_acl` option when built with config: `mkl_aarch64` 
Without this change, the following line evaluates to false everytime
https://github.com/openxla/xla/blob/9d8668f0eb7773905d4047692f481e277dd5ba09/xla/service/cpu/dot_op_emitter.cc#L770

2. This PR also enables a VLOG for a successful ACL matmul operation.
Tested by using `TF_CPP_MAX_VLOG_LEVEL=2 TF_CPP_MIN_LOG_LEVEL=0`


Rohanjames1997:
Closing in favor of #16438

PR #16559, opened 2024-08-28
[ROCm] Fix transpose related tests which breaks due to 6fabf59

hsharsha:
Failure is due to https://github.com/openxla/xla/pull/16336

### DIFF
--- a/xla/service/gpu/tests/gpu_fusion_test.cc
+++ b/xla/service/gpu/tests/gpu_fusion_test.cc
@@ -127,7 +127,7 @@ ENTRY main {
 // CHECK-NEXT:   [[s_1_1:%[^ ]+]] = f32[16,32]{1,0} sqrt([[param_0_1_0]])
 // CHECK-NEXT:   ROOT [[c_1_2:%[^ ]+]] = f32[32,16]{1,0} transpose([[s_1_1]]), dimensions={1,0}
 // CHECK-NEXT: }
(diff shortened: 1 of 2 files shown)


mooskagh:
How exactly is `CompileAndVerifyIr` failing? Maybe it makes sense to replace the check with something else rather than removing it?

hsharsha:
> How exactly is `CompileAndVerifyIr` failing? Maybe it makes sense to replace the check with something else rather than removing it?

It is not getting this string `call void BARRIER()`
It was removed in older tests at https://github.com/openxla/xla/pull/16336/files#diff-8874213f2bac0a0c682b006a9990a5a49aff4eecff481f074782692633e55c70L218-L224. Hence I removed them

akuegel (changes requested):
I think this is the wrong fix. You only see those differences because ROCM doesn't have the new MLIR emitters enabled by default. If we change the test expectations, it will fail on Nvidia GPUs where it is enabled by default.
Maybe disable the tests for now on ROCM, until the mlir emitters are enabled there as well?

hsharsha:
Test works after [fbd00d9](https://github.com/openxla/xla/commit/fbd00d9b94ba9fd155b5fca572122c55f691424f)

akuegel:
Thanks for the confirmation, happy that they work again! And sorry for temporarily breaking them.

PR #16696, opened 2024-09-01
Various macOS QOL enchancements

steeve:
This PR adds various small quality of life improvements to macOS builds:
- drop the `.so` suffix for PjRt plugin targets (`.dylib` on macOS)
- add compatibility with Apple Command Line Tools (no need for Xcode anymore)
- only export the `GetPjrtApi` symbol on macOS
- leverage bazel's `cc_binary.additional_linker_inputs` instead of using `deps`

It is probable the `.so` change my break some other builds, but I couldn't find any use in the XLA repo to patch ?

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -698,7 +698,7 @@ build:release_macos_x86 --config=avx_linux
 build:release_macos_x86 --cpu=darwin
 # Target Catalina as the minimum compatible OS version
 build:release_macos_x86 --macos_minimum_os=10.15
-build:release_macos_x86 --action_env MACOSX_DEPLOYMENT_TARGET=10.15
+build:release_macos_x86 --macos_sdk_version=10.15
 
 # Build configs for macOS Arm64
 build:release_macos_arm64 --config=release_macos_base
@@ -706,7 +706,7 @@ build:release_macos_arm64 --cpu=darwin_arm64
 build:release_macos_arm64 --define=tensorflow_mkldnn_contraction_kernel=0
 # Target Moneterey as the minimum compatible OS version
 build:release_macos_arm64 --macos_minimum_os=12.0
-build:release_macos_arm64 --action_env MACOSX_DEPLOYMENT_TARGET=12.0
+build:release_macos_arm64 --macos_sdk_version=12.0
 
 # Base test configs for macOS
 test:release_macos_base --verbose_failures=true --local_test_jobs=HOST_CPUS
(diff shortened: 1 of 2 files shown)


ddunl (approved):
LGTM - Unfortunately we basically have no ability to debug x86 Macos interactively, so I may need your help if I run the nightlies on this change and it comes back red. Thanks for the PR!!

ddunl:
Ok sorry - this PR slipped my mind. I think I need to patch this to account for how TF/JAX use the CPU/GPU plugins. I'll work on this more this week. Sorry again for the delay and that you for the PR!

dimitar-asenov:
@ddunl Did you have a chance to look this PR internally?

ddunl:
I believe this needs JAX changes (possibly TF as well), but this slipped through the cracks for me, I'll try to merge this later this week or early next.

ddunl:
Sorry about this, I've dropped the ball here. I think what I'll do soon is try to split this change up and merge them one after another. We mostly test MacOS on a continuous/nightly basis and I think these changes could affect TF and JAX as well, so I want to be careful. Sorry again for the delay, I have taken an unacceptably long time to take care of this PR.

steeve:
I'm still marveling at the pace of the project, you're fine

ddunl:
Ok, I've patched this in, I'm going to start merging this split into a few parts to try to avoid the cross-repo issues and ideally prevent a rollback. Starting with the .bazelrc bits which now live in tensorflow.bazelrc

dimitar-asenov:
@ddunl Could you please provide an update on this PR?

ddunl:
Yes, I need to ping for review on the second part of this PR and then I'll submit the final part. I'm most concerned about the last part wrt to MacOS builds because I need to patch it changes to JAX and possibly TensorFlow.

dimitar-asenov:
@ddunl Could you please provide an update on this? If you're no longer working on this PR, could you suggest someone else that could help merge it?

ddunl:
Ok - sorry again for the large delay, https://github.com/openxla/xla/pull/28498 contains the last of the changes. Unfortunately the rename off of .so doesn't work, there's certainly some way to do it but all of them are a bit annoying - we probably have to make a pjrt plugin bazel macro that makes the correct target name based on the OS and then aliases that to a generically named target. So my last PR just updates the `install_name`.

I think the constraint is just that the plugin ends up having the `.so` suffix. I tried using `-Wl,soname` but that still had tests failing internally, and using a macro to change the name of the target also doesn't work because bazel doesn't like that the name is "configurable" due to the OS config settings required.

Sorry again for the delay and thank you for the contribution!

PR #16734, opened 2024-09-03
[NV] Use FP8 conversion intrinsics, when available

sergey-kozub:
PTX "cvt" instruction supports converting to/from FP8 types. The NV hardware supports E4M3FN and E5M2 types.
This PR updates the MLIR emitter to use this instruction instead of emitting a long sequence of operations (this matters in compute-bound FP8 kernels).

The NVVM intrinsic allows converting two FP8 values with a single instruction, but as the emitter is elementwise, only one of the inputs is used. This is wasteful, but still much faster than emitting the sequence of instructions.

Before ptx 7.8 (cuda 11.8), the instruction is not supported. Starting with ptx 8.1 (cuda 12.1), the instruction is supported for sm89+. Between those versions, the instruction is supported for sm90+, thas is, if trying to compile on Ada (sm89) with cuda version < 12.1, the ptxas will complain..

Reference:
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cvt (see "PTX ISA Notes" and "Target ISA Notes").

### DIFF
--- a/xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
+++ b/xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
@@ -307,7 +307,7 @@ MlirFusionEmitterBase::CreateLLVMModule(
   mlir::PassManager pm(&mlir_context);
   AddXlaGpuOpsOptimizationPasses(pm);
   AddLoopTransformationPasses(pm);
-  AddLoweringPasses(pm, device);
+  AddLoweringPasses(pm, device, hlo_module->config().debug_options());
   auto pipeline_status = RunPassPipeline(module.get(), pm, trace.get());
   if (trace) {
     DumpPerModuleProtobufToFile(
@@ -566,7 +566,8 @@ void AddLoopTransformationPasses(mlir::OpPassManager& pm) {
 }
 
 void AddLoweringPasses(mlir::OpPassManager& pm,
-                       const se::DeviceDescription& device) {
+                       const se::DeviceDescription& device,
+                       const DebugOptions& debug_options) {
   bool is_amd = std::holds_alternative<se::RocmComputeCapability>(
       device.gpu_compute_capability());
   pm.addNestedPass<FuncOp>(CreateConvertPureCallOpsPass());
@@ -590,6 +591,14 @@ void AddLoweringPasses(mlir::OpPassManager& pm,
   pm.addPass(mlir::createLoopInvariantCodeMotionPass());
   pm.addPass(mlir::createSymbolDCEPass());
   pm.addPass(mlir::createCSEPass());
+
+  // This pass has to run before `ExpandFloatOpsPass`.
+  auto maybe_convert_fp8 = CreateConvertFloatNvidiaPass(
+      device.cuda_compute_capability(), debug_options.xla_gpu_cuda_data_dir());
+  if (maybe_convert_fp8.has_value()) {
+    pm.addPass(std::move(*maybe_convert_fp8));
+  }
+
   pm.addPass(CreateExpandFloatOpsPass());
   pm.addPass(mlir::createLowerAffinePass());
   pm.addPass(mlir::createConvertSCFToCFPass());
--- a/xla/stream_executor/cuda/cuda_asm_compiler.cc
+++ b/xla/stream_executor/cuda/cuda_asm_compiler.cc
@@ -305,9 +305,11 @@ static absl::StatusOr<std::string> FindPtxAsExecutable(
 
 absl::StatusOr<SemanticVersion> GetAsmCompilerVersion(
     std::string_view preferred_cuda_dir) {
-  TF_ASSIGN_OR_RETURN(std::string ptxas_path,
-                      FindPtxAsExecutable(preferred_cuda_dir));
-  return GetToolVersion(ptxas_path);
+  return ::xla::GetOnce([&]() -> absl::StatusOr<SemanticVersion> {
+    TF_ASSIGN_OR_RETURN(std::string ptxas_path,
+                        FindPtxAsExecutable(preferred_cuda_dir));
+    return GetToolVersion(ptxas_path);
+  });
 }
 
 absl::StatusOr<std::vector<uint8_t>> CompileGpuAsmUsingPtxAs(
--- a/xla/service/gpu/fusions/mlir/mlir_fusion_emitter.h
+++ b/xla/service/gpu/fusions/mlir/mlir_fusion_emitter.h
@@ -122,7 +122,8 @@ void AddLoopTransformationPasses(mlir::OpPassManager& pm);
 
 // Adds passes that lower transformed loops to LLVM.
 void AddLoweringPasses(mlir::OpPassManager& pm,
-                       const se::DeviceDescription& device);
+                       const se::DeviceDescription& device,
+                       const DebugOptions& debug_options);
 
 }  // namespace gpu
 }  // namespace xla
--- a/xla/service/gpu/fusions/transforms/BUILD
+++ b/xla/service/gpu/fusions/transforms/BUILD
@@ -1,4 +1,5 @@
 load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library")
+load("@tsl//tsl/platform/default:cuda_build_defs.bzl", "if_cuda_is_configured")
 
 package(
     # copybara:uncomment default_applicable_licenses = ["//tensorflow:license"],
@@ -48,7 +49,9 @@ cc_library(
         "simplify_arith.cc",
         "unswitch_loops.cc",
         "vectorize_loads_stores.cc",
-    ],
+    ] + if_cuda_is_configured([
+        "convert_float_nvidia.cc",
+    ], no_cuda=["cuda_stub.cc"]),
     hdrs = ["passes.h"],
     deps = [
         ":passes_inc_gen",
@@ -104,5 +107,8 @@ cc_library(
         "@llvm-project//mlir:VectorDialect",
         "@llvm-project//mlir:VectorToLLVM",
         "@llvm-project//mlir:VectorTransforms",
-    ],
+    ] + if_cuda_is_configured([
+        "//xla/stream_executor:semantic_version",
+        "//xla/stream_executor/cuda:cuda_asm_compiler",
+    ]),
 )
--- a/xla/service/gpu/fusions/transforms/convert_float_nvidia.cc
+++ b/xla/service/gpu/fusions/transforms/convert_float_nvidia.cc
@@ -1,0 +1,258 @@
+/* Copyright 2024 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include <cassert>
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+#include <utility>
+
+#include "llvm/ADT/APFloat.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/IR/ImplicitLocOpBuilder.h"
+#include "mlir/IR/Value.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "tsl/platform/statusor.h"
+#include "xla/service/gpu/fusions/transforms/passes.h"
+#include "xla/stream_executor/cuda/cuda_asm_compiler.h"
+#include "xla/stream_executor/device_description.h"
+#include "xla/stream_executor/semantic_version.h"
+
+namespace xla {
+namespace gpu {
+
+namespace ma = ::mlir::arith;
+namespace ml = ::mlir::LLVM;
+using mlir::Value;
+
+#define GEN_PASS_DEF_CONVERTFLOATNVIDIAPASS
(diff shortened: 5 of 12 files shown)


jreiffers:
Thanks a lot! We either need to do this in LLVM (probably best?) or check the device we're compiling for, since AMD is migrating to this code path as well.

jreiffers:
Oh, reading is hard. I missed that you already had a guard. Looking again.

xla/service/gpu/fusions/transforms/expand_float_ops.cc
```diff
@@ -516,9 +630,42 @@ struct RewriteExtFPattern : public mlir::OpRewritePattern<ma::ExtFOp> {
...
+                               : EmitFloatConversion(src, dst_ty, b));
     return mlir::success();
   }
+
+  Value EmitExtFromF8Intrinsic(Value value, mlir::FloatType to_ty,
+                               mlir::ImplicitLocOpBuilder& b) const {
+    // Extend the smaller type to the FP16 type using the intrinsic, and then
+    // to the destination type. In the case of BF16 go through the intermediate
+    // FP32 type (as there's no F2F op for f16->bf16).
+    Value input = b.create<ml::ZExtOp>(
```
jreiffers, line 646: Let's add an assertion here that value.getType is f8e4m3fn or f8e5m2.

jreiffers (approved):

sergey-kozub:
It looks like there's an issue introduced by the rebase to main. Looking into this now.

xla/service/gpu/fusions/transforms/expand_float_ops.cc
sergey-kozub: Added.

jreiffers (approved):

sergey-kozub:
Fixed the issue by updating "check-prefixes" in the lit test header.

jreiffers (approved):

xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
```diff
@@ -590,8 +593,22 @@ void AddLoweringPasses(mlir::OpPassManager& pm,
   pm.addPass(mlir::createLoopInvariantCodeMotionPass());
   pm.addPass(mlir::createSymbolDCEPass());
   pm.addPass(mlir::createCSEPass());
+
+  auto ptx_version =
+      se::GetAsmCompilerVersion(debug_options.xla_gpu_cuda_data_dir());
   pm.addPass(CreateExpandFloatOpsPass(
-      !device.cuda_compute_capability().IsAtLeastAmpere()));
+      /*pre_ampere=*/!device.cuda_compute_capability().IsAtLeastAmpere(),
+      /*has_f8_cvt=*/!ptx_version.ok()
```
akuegel, line 601: Could you please extract this logic into a function with several if conditions? I think that would make it easier to read.

jreiffers: Yeah, I think that's probably better. We're running into a bunch of RoCM build failures, having a separate pass might make that more manageable without littering ifdefs all over the place.

sergey-kozub: Extracting into a function with `if`s is clear to me, but adding a separate pass is not - could you elaborate?

jreiffers: I was imagining an Nvidia-specific pass that runs before this one, replacing all the conversions that we have efficient intrinsics for, so this pass only cleans up the rest afterwards. Then we can just add that one to the pipeline if we're compiling for Nvidia. I'm worried if we add all these switches in this pass, it will get a bit difficult to follow.

WDYT?

sergey-kozub: Makes sense, will update this PR soon.

sergey-kozub: Extracted the NVidia-specific logic into a separate (optional) pass. PTAL

xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
```diff
@@ -590,8 +593,21 @@ void AddLoweringPasses(mlir::OpPassManager& pm,
   pm.addPass(mlir::createLoopInvariantCodeMotionPass());
   pm.addPass(mlir::createSymbolDCEPass());
   pm.addPass(mlir::createCSEPass());
-  pm.addPass(CreateExpandFloatOpsPass(
-      !device.cuda_compute_capability().IsAtLeastAmpere()));
+
+  auto ptx_version =
```
jreiffers, line 597: We need to guard this call with an ifdef and only add the include if we're compiling for Nvidia. See MinThreadsXRowReduction in xla/service/gpu/reduction_utils.cc for an example. That function also says that this function is expensive, so we will probably want to have some caching here as well (or maybe unify that somehow, but that doesn't need to be in this PR).

xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
jreiffers: Awesome, thank you so much! Looks great.

xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
sergey-kozub: A few lines above in the same file (line 578) we access `device.cuda_compute_capability()` without an `#ifdef`
What is the downside of this approach, compared to using an `#ifdef GOOGLE_CUDA` ? (btw, updated the PR)

jreiffers: `cuda_compute_capability` exists in both builds, but `cuda_asm_compiler` isn't built in a ROCm build, so it needs to be guarded by an `ifdef` in the cc file and an `if_cuda_is_configured` in the BUILD file. We could definitely centralize these guards and the caching, but maybe not in this PR?

sergey-kozub: Clear, thanks. Updated the PR.

jreiffers: If the comment in reduction_utils is true, we still need caching here, or maybe better directly in se::GetAsmCompilerVersion.

sergey-kozub: After reading the code in reduction_utils, I still cannot figure out why atomics are needed there.

Anyway, added a simple utility for getting a value once during the execution (haven't found a similar one in the XLA codebase) - please suggest a namespace where I should put it in (it doesn't really belong to stream_executor).

jreiffers: Looks good in principle, but I think if we want to use this, we have to do it directly in `GetAsmCompilerVersion`, since otherwise the value won't be cached across multiple invocations in different code locations (we don't want to evaluate it once from `reduction_utils` and then again from here).

As to where to put it, I can't really find a really good place either. How about xla/util.h? That contains a bunch of semi-related utilities, so at least we're not making it much worse.

sergey-kozub: Moved the helper to `util.h` and updated `GetAsmCompilerVersion`.

We can also update a similar method `GetNvLinkVersion` and remove the redundancy in `reduction_utils` (in a separate PR).

sergey-kozub:
Note that `clang-format` failed in the CI on a line unaffected by this PR, but running it locally (18.1.3) yields no error.

jreiffers (approved):

xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
```diff
@@ -590,8 +593,25 @@ void AddLoweringPasses(mlir::OpPassManager& pm,
   pm.addPass(mlir::createLoopInvariantCodeMotionPass());
   pm.addPass(mlir::createSymbolDCEPass());
   pm.addPass(mlir::createCSEPass());
-  pm.addPass(CreateExpandFloatOpsPass(
-      !device.cuda_compute_capability().IsAtLeastAmpere()));
+
+#ifdef GOOGLE_CUDA
+  // The call to `GetAsmCompilerVersion` is expensive, but the result never
```
akuegel, line 598: Can this check be moved to the pass?
Currently, we get a build breakage due to the header include of xla/stream_executor/cuda/cuda_asm_compiler.h
This could be fixed by yet another #ifdef, but we try to avoid those. So if the check is moved to a pass that is compiled only for cuda, the include can be used unconditionally.

sergey-kozub: Do we have an example of passes that are compiled for a specific platform only?

Will I need to create a separate "passes_nvidia.td" file (along with the header)?

akuegel: Unfortunately no examples yet, and thinking a bit more about this, I can understand the problem and why you ask about separate passes_nvidia.td file. If we have a pass that should be compiled on a specific platform only, we need to create a stub implementation if it is compiled on a different platform. So at least on the BUILD side I cannot see a way to avoid a if_cuda_is_configured guard, but we can avoid the #ifdef in the code.
So essentially my suggestion would be to create a convert_float_nvidia_stub.cc file that returns a pass that does nothing, and use the cuda_asm_compiler.h header only in the "real" convert_float_nvidia.cc file. Then, we need neither a #ifdef for the header include, nor for adding the pass to the pipeline.

pifon2a (approved):

xla/service/gpu/fusions/mlir/mlir_fusion_emitter.cc
sergey-kozub: Updated the code, looks cleaner. PTAL

xla/stream_executor/cuda/cuda_asm_compiler.cc
```diff
@@ -305,9 +305,11 @@ static absl::StatusOr<std::string> FindPtxAsExecutable(
 
 absl::StatusOr<SemanticVersion> GetAsmCompilerVersion(
     std::string_view preferred_cuda_dir) {
-  TF_ASSIGN_OR_RETURN(std::string ptxas_path,
-                      FindPtxAsExecutable(preferred_cuda_dir));
-  return GetToolVersion(ptxas_path);
+  return ::xla::GetOnce([&]() -> absl::StatusOr<SemanticVersion> {
```
akuegel, line 308: In the internal review, I got a comment from @beckerhe:

This kind of caching doesn't take different values of preferred_cuda_dir into account. It might happen to work right now, but it's hack nonetheless.

I have cloned your PR and have reverted the changes in this file, and the util.h change it is using, as they are unrelated to the other changes anyway, and will get that uncontroversial part landed. Can you please create a separate PR for these other changes, and get it reviewed by Henning?

akuegel:
@sergey-kozub This had to be rolled back because it broke the OSS GPU Kokoro. The failure is:

mlir_fusions_opt: error while loading shared libraries: libcudnn_engines_precompiled.so.9: cannot open shared object file: No such file or directory

I did quite a few fixes to your PR before I was able to merge it, so when you roll forward, I would suggest basing your new PR on what was merged. For example the stub file needs to contain an implementation for the constructor, also usually we should have a constructor defined in the .td file.

akuegel:
I am trying to add //xla/stream_executor:cuda_platform as dependency (guarded with if_cuda_is_configured) to mlir_fusion_opt, maybe that will help.

xla/stream_executor/cuda/cuda_asm_compiler.cc
sergey-kozub: Created https://github.com/openxla/xla/pull/16734

sergey-kozub:
How do I reproduce the failure?

akuegel:
> How do I reproduce the failure?

That is a good question. Internally I also cannot reproduce it, so possibly it only fails in the hermetic cuda environment. The thing I mentioned above with adding the dependency didn't help (I tried it because I saw that gpu_hlo_opt has this dependency).

akuegel:
If it is indeed related to hermetic cuda, then maybe this may help to reproduce: --repo_env=HERMETIC_CUDA_VERSION=12.3.2 --repo_env=HERMETIC_CUDNN_VERSION=9.3.0

sergey-kozub:
> If it is indeed related to hermetic cuda, then maybe this may help to reproduce: --repo_env=HERMETIC_CUDA_VERSION=12.3.2 --repo_env=HERMETIC_CUDNN_VERSION=9.3.0

I believe I'm running this bazel config by default (and it doesn't fail).

akuegel:
To be honest, I don't understand the failure. I also tried adding a dependency to cudnn library (https://github.com/openxla/xla/commit/7d7e39107c304317d730dc84a88e0a698ebbab9c#diff-663bc927847985483d6ce4cc1ca6990c3665074243c29351317a556639f5dde0 ) but that also didn't help. As far as I can tell, the error indicates a missing dependency, but which one, I am not sure.

akuegel:
So we have started a discussion internally how to best solve this (adding a pass depending on the PTX version). There is some agreement that my suggestion to move the call to the pass target was wrong, and most likely if we didn't have the stream_executor dependencies in the pass target, it would also not fail in that weird way. But @beckerhe also said that using GetAsmCompilerVersion to query the ptxas version is wrong, because that doesn't take into account whether we even use ptxas for compilation.

Quote:

```
To determine the version in the first place the actual compilation method needs to be taken into account (ptxas, driver, library, etc). Currently this is determined during the compilation process which is not great. We should do that when constructing NVPTXCompiler.
```

akuegel:
Also another suggestion from @beckerhe

I think the best solution right now is to use DeviceDescription::runtime_version() which gives you the version of the CUDA toolkit. From that you can infer what the PTX lowering support no matter which method is used.

sergey-kozub:
> I think the best solution right now is to use DeviceDescription::runtime_version() which gives you the version of the CUDA toolkit. From that you can infer what the PTX lowering support no matter which method is used.

Thanks! I'll create another PR and use the method you propose.

akuegel:
Thanks, and sorry for all that back and forth. Unfortunately we don't have something similar yet, so you are hitting all the obstacles right now.

PR #16754, opened 2024-09-03
Disable VerifyInstructionNameUnchanged() for GPUs due to crashes

zhenying-liu:
After hlo verifier was enabled to check instruction name changes, we saw applications failed.
We disable this option for GPUs for now until all the fixes are in place.


zhenying-liu:
This is an alternative for https://github.com/openxla/xla/pull/16236, if we can't fix the scheduling names after the hlo rematerialization pass.

akuegel (approved):
Thank you, I think disabling is the best solution for now.

Issue #14600, opened 2024-07-08
Channel ids for collectives are not unique when custom sharding is inlined in XLA


Tixxx:
We are seeing a compilation crash with a custom partitioning defined by the user. I'm attaching the repro script and instructions to repro.  The error happens when running with TransformerEngine in our jax toolbox docker container, although I'd think this would happen for any inlined custom sharding.
My repro needs an 8-gpu machine to run. 
here's the smalle repro script(https://github.com/Tixxx/fileshare/blob/main/scan_unbound_nvbug_with_te_refactor_v2.py).
Instructions:
docker run --gpus all -it ghcr.io/nvidia/jax:pax-2024-06-18
python scan_unbound_nvbug_with_te_refactor_v2.py

The error is
"INTERNAL: RET_CHECK failure (external/xla/xla/service/hlo_verifier.cc:2494) first->opcode() == instr->opcode() channel 1 is used for different types of channel instructions"

Apparently, spmd paritioner assigns the same channel id to 2 different collectives.
When the spmd partitioner tries to inline the custom partitioning call, it assigns channel ids to collectives in that computation incrementally, but when it tries to create collectives for sharded instruction in the main computation, the state of the increment is lost and the channels starts from the original value again. The state is passed around in numerous places when creating collectives in the partitioner. I haven't been able to pin point where the state is lost.
But maybe we can do a post processing of the partitioned module to scan through collectives, detect duplicate channel ids and re-assign them with unique ones.

ptoulme-aws:
i have noticed this also when unrolling while loops with custom-calls inside of them. My solution was to add a channel id legalizer pass that enforces unique channel id for each collective

Tixxx:
> i have noticed this also when unrolling while loops with custom-calls inside of them. My solution was to add a channel id legalizer pass that enforces unique channel id for each collective

Thanks. I was thinking along the same line to have an uniquifyer to post-process the graph.

nouiz:
@ptoulme-aws Any hope of upstreaming or making publicly available your fix?

ptoulme-aws:
I will PR the pass tonight.
Best,
Patrick Toulme

On Jul 16, 2024, at 5:38 PM, Frédéric Bastien ***@***.***> wrote:

﻿

@ptoulme-aws<https://github.com/ptoulme-aws> Any hope of upstreaming or making publicly available your fix?

ptoulme-aws:
This PR should unblock you - https://github.com/openxla/xla/pull/15002

[...]

PR #16779, opened 2024-09-04
[NVIDIA] Added an option in hlo verifier options to skip checking duplicate channel ids

Tixxx:
This is to address this issue(https://github.com/openxla/xla/issues/14600).
In gpu pipeline, the uniqueness of channel id is not used anywhere, only the presence of it is used for determining replica/partition properties and for device assignment. This pr introduces an option in hlo verifier to skip checking for duplicate channel ids.
The duplication check is disabled in all gpu pipelines.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -291,7 +291,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_executable_warn_stuck_timeout_seconds(10);
   opts.set_xla_gpu_executable_terminate_timeout_seconds(30);
   opts.set_xla_gpu_experimental_disable_binary_libraries(false);
-  opts.set_xla_experimental_ignore_channel_id(true);
+  opts.set_xla_experimental_ignore_channel_id(false);
   return opts;
 }
 
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -291,6 +291,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_executable_warn_stuck_timeout_seconds(10);
   opts.set_xla_gpu_executable_terminate_timeout_seconds(30);
   opts.set_xla_gpu_experimental_disable_binary_libraries(false);
+  opts.set_xla_experimental_ignore_channel_id(true);
   return opts;
 }
 
@@ -1944,6 +1945,11 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       debug_options->xla_gpu_experimental_disable_binary_libraries(),
       "Disable XLA GPU passes that depend on non-open source binary "
       "libraries"));
+  flag_list->push_back(tsl::Flag(
+      "xla_experimental_ignore_channel_id",
+      bool_setter_for(&DebugOptions::set_xla_experimental_ignore_channel_id),
+      debug_options->xla_experimental_ignore_channel_id(),
+      "Experimental: ignore channel ids for collective operations."));
 }  // NOLINT(readability/fn_size)
 
 // Allocates flag_values and flag_objects; this function must not be called more
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -474,9 +474,10 @@ GpuCompiler::GpuCompiler(se::Platform::Id platform_id,
 
 namespace {
 // Adds the HloVerifier for GPU to the given pipeline.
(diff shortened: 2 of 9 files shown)


xla/service/hlo_verifier.h
```diff
@@ -149,6 +149,9 @@ struct HloVerifierOpts {
   // cloned (".clone" suffix) or rematted (".remat");
   bool verify_instruction_name_unchanged = false;
 
+  // Check if channel instructions all have unique channel ids.
+  bool verify_duplicate_channel_ids = true;
```
frgossen, line 153: Verify unique channel IDs?

frgossen (changes requested):

xla/service/hlo_verifier.h
Tixxx: good point, i changed it.

reedwm:
I'll let Frederik review since he's more familiar with this than I am.

But I'm unsure why the issue is occurring. In the original issue, you stated

> When the spmd partitioner tries to inline the custom partitioning call, it assigns channel ids to collectives in that computation incrementally, but when it tries to create collectives for sharded instruction in the main computation, the state of the increment is lost and the channels starts from the original value again.

Why is "the state of the increment lost"? Shouldn't the spmd partitioner keep track of the latest channel id throughout the entire run of the spmd partitioner?

Also, is it possible to add an e2d test that doesn't directly assign the same channel ID to two collectives, but instead reproduces the case where the SPMD partitioner automatically assigns the same channel ID to two collectives?

Tixxx:
> I'll let Frederik review since he's more familiar with this than I am.
> 
> But I'm unsure why the issue is occurring. In the original issue, you stated
> 
> > When the spmd partitioner tries to inline the custom partitioning call, it assigns channel ids to collectives in that computation incrementally, but when it tries to create collectives for sharded instruction in the main computation, the state of the increment is lost and the channels starts from the original value again.
> 
> Why is "the state of the increment lost"? Shouldn't the spmd partitioner keep track of the latest channel id throughout the entire run of the spmd partitioner?
> 
> Also, is it possible to add an e2d test that doesn't directly assign the same channel ID to two collectives, but instead reproduces the case where the SPMD partitioner automatically assigns the same channel ID to two collectives?

Yes, the spmd partitioner should keep track of the incremental state of channel id, there's a bug in the spmd partitioner which still needs to be fixed. However, the channel id is irrelevant to gpu pipeline so even there are duplicated ids, the verifier shouldn't fail when compiling for gpu. This pr is addressing the verifier part.
I think the e2e test you suggest would fit better for the fix of the lost channel id state, for the verifier change, we just need to make sure it wont fail compilation when duplicate channel ids are present.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -471,6 +471,7 @@ namespace {
 // Adds the HloVerifier for GPU to the given pipeline.
 void AddHloVerifier(HloPassPipeline* pipeline, HloVerifierOpts&& opts = {},
                     bool debug_only = false) {
+  opts.verify_unique_channel_ids = false;
```
frgossen, line 474: Since this is part of an attempt to remove the channel ID, can we base this on an experimental flag? Maybe `experimental_ignore_channel_ids`

frgossen (changes requested):

xla/service/gpu/gpu_compiler.cc
Tixxx: are you thinking to use the same flag to control when to ignore channel id in other places when we try to remove it? if so,  adding it makes sense.

Tixxx: added the debug option

xla/debug_options_flags.cc
```diff
@@ -287,6 +287,8 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_cudnn_gemm_max_plans(5);
 
   opts.set_xla_gpu_enable_triton_gemm_int4(false);
+
+  opts.set_xla_experimental_ignore_channel_id(true);
```
frgossen, line 291: If it's experimental, the default should be false, no?

frgossen (changes requested):

xla/debug_options_flags.cc
Tixxx: I thought we'd want to ignore channel ids by default. If that's the case, defaulting it to true would make more sense. Or we just dont call it experimental at all.

Tixxx: ok i changed the default to false for now.

Tixxx:
@frgossen  can you take another look at this pr?

frgossen (approved):

Issue #16569, opened 2024-08-28
Hard error with PGLE Accruracy Checker


shraiysh:
In a [recent commit](https://github.com/openxla/xla/commit/737a7da3c5405583dc95773ac0bb11b1349fc9ea), a hard error was introduced in PGLE when instructions are missing from the profile. The hard error has the following problems:

1. If there is a conditional operation (if-else) in HLO, and it always jumps to the same branch, then the other computation never gets executed - hence will not have its costs in the profile. If we then use this profile for PGLE, it will crash.
2. There are some modules that only run during initialization. If the profiler runs after a few training steps, then these modules do not run for the profiler. If we then use this profile for PGLE, it will crash.

Possible solutions:

1. Hard error behind a flag like `--strict-pgle`. The default behavior should not have a hard error with missing instructions, instead it should use a default cost for them.
2. Discuss a testing methodology that should be able to test the effect of PGLE on HLO schedule so we do not have to have this hard error at all.

shraiysh:
@golechwierowicz

abhinavgoel95:
cc @Tixxx for viz.

golechwierowicz:
Hey,

As a mitigation would disabling the pass work for you? It is implemented as such: https://github.com/tensorflow/tensorflow/pull/73606/files.

golechwierowicz:
Solution 1 is implemented in the compiler and it issues a warning if missing instructions are found.

PR #16804, opened 2024-09-04
Add flag `xla_gpu_enable_pgle_accuracy_checker` to enable accuracy checker

shraiysh:
To enable the strict pgle accuracy checker, set `--xla_gpu_enable_pgle_accuracy_checker` in XLA_FLAGS

### DIFF
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -487,7 +487,11 @@ absl::StatusOr<ScheduleMetadata> ScheduleGpuModule(
         std::move(aggregator));
     LOG(INFO) << "Found profile, using profile guided latency estimator";
     VLOG(1) << "Profile:\n" << profile->DebugString();
-    pipeline.AddPass<PGLEAccuracyChecker>(*pg_latency_estimator);
+    if (module->config()
+            .debug_options()
+            .xla_gpu_enable_pgle_accuracy_checker()) {
+      pipeline.AddPass<PGLEAccuracyChecker>(*pg_latency_estimator);
+    }
     latency_estimator = std::move(pg_latency_estimator);
   } else if (enable_analytical_latency_estimator) {
     latency_estimator = std::make_unique<AnalyticalLatencyEstimator>(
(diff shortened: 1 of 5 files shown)


frgossen (approved):
Thanks!

shraiysh:
Are there any problems with this PR? It hasn't landed yet.

shraiysh:
cc: @xla-rotation

golechwierowicz:
Hey, can we actually implement the solution 1 mentioned in https://github.com/openxla/xla/issues/16569#issuecomment-2336817021?

A separate flag, disabled by default and passed through HloModule? This should be approx ~10 loc more but this would be a proper solution.

shraiysh:
>Hey, can we actually implement the solution 1 mentioned in https://github.com/openxla/xla/issues/16569#issuecomment-2336817021?

In that case, we do not need an additional flag. We could simply use `--xla_disable_hlo_passes=pgle-accuracy-checker`. I can close this PR while we discuss long term fix on the issue.

shraiysh:
This is a blocker for JAX release. While we work on a long term fix, can we merge this PR so JAX can be released with default flags?

xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -431,7 +431,7 @@ static int64_t GetSchedulerMemoryLimit(
 
 absl::StatusOr<ScheduleMetadata> ScheduleGpuModule(
     HloModule* module, int64_t pointer_size,
-    const se::DeviceDescription& gpu_device_info) {
+    const se::DeviceDescription& gpu_device_info, bool run_accuracy_checker) {
```
cheshire, line 434: Instead of a bool, could we make this a flag in xla.proto , have it the correct default, and consult module options?

cheshire (changes requested):

xla/service/gpu/gpu_hlo_schedule.cc
shraiysh: Done

golechwierowicz (approved):

cheshire (approved):

PR #16841, opened 2024-09-05
Delete FP8 Scaling Factors in GEMM Rewriter

philipphack:
Removes the scaling factors of C and D (matrix bias and result) from FP8 Custom Calls created in the GEMM rewriter when their data types are not FP8. See https://github.com/openxla/xla/pull/15795.

### DIFF
--- a/xla/service/gpu/transforms/gemm_rewriter.cc
+++ b/xla/service/gpu/transforms/gemm_rewriter.cc
@@ -1083,12 +1083,18 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
 
     // cuBLASLt FP8 GEMM kernels require the scaling factors to be in F32
     // format. Set the factors to one when no scaling factors were captured.
-    Literal one_literal = LiteralUtil::One(F32);
-    HloInstruction *one = instr->AddInstruction(
-        HloInstruction::CreateConstant(one_literal.Clone()));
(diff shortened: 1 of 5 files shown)


xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -1081,12 +1081,14 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
-    Literal one_literal = LiteralUtil::One(F32);
-    HloInstruction *one = instr->AddInstruction(
-        HloInstruction::CreateConstant(one_literal.Clone()));
     std::array<bool, 2> mult_scale{a.mult_scale, b.mult_scale};
     std::array<HloInstruction *, 2> scales{a.scale, b.scale}, inv_scales,
         scales_f32;
+    HloInstruction* one;
+    if (!scales[0] || !scales[1] || !mult_scale[0] || !mult_scale[1]) {
+      one = instr->AddInstruction(
+          HloInstruction::CreateConstant(LiteralUtil::One(F32)));
```
reedwm, line 1090: Why not unconditionally create this, as was done before? Even before, it's possible this was unused. I worry this may be accidentally used without being initialized in the future.

reedwm (changes requested):

reedwm (approved):

PR #16849, opened 2024-09-05
[XLA:GPU] Fix dropout state in cuDNN FMHA

Cjkkkk:
* Fix issue that dropout state is not set after switching to cuDNN thunk in commit: https://github.com/openxla/xla/commit/0aa816e0e4c5c9544237da664eee6ac8e13aabb5.
* Maintain a separate offset copy for each device as thunk can be executed by multiple threads concurrently.
* Add an unit test for the fmha dropout lowering. As it is hard to implement similar philox rng in HLOs, it only guarantees the the fmha custom call with dropout can be lowered correctly.

### DIFF
--- a/xla/service/gpu/runtime/cudnn_thunk.cc
+++ b/xla/service/gpu/runtime/cudnn_thunk.cc
@@ -23,21 +23,23 @@ limitations under the License.
 #include "absl/base/call_once.h"
 #include "absl/status/status.h"
 #include "absl/types/span.h"
+#include "tsl/platform/errors.h"
 #include "xla/service/buffer_assignment.h"
 #include "xla/service/gpu/kernel_arguments.h"
 #include "xla/service/gpu/runtime/thunk.h"
+#include "xla/stream_executor/cuda/cuda_dnn.h"
 #include "xla/stream_executor/device_memory.h"
-#include "xla/stream_executor/dnn.h"
-#include "tsl/platform/errors.h"
 
 namespace xla {
 namespace gpu {
 
 CuDnnThunk::CuDnnThunk(std::string fingerprint, ThunkInfo thunk_info,
-                       absl::Span<const KernelArgument> kernel_arguments)
+                       absl::Span<const KernelArgument> kernel_arguments,
+                       std::optional<int64_t> sdpa_dropout_seed)
     : Thunk(Kind::kCuDnn, std::move(thunk_info)),
       fingerprint_(std::move(fingerprint)),
-      graph_(std::make_shared<se::dnn::LazyDnnGraph>(nullptr)) {
+      graph_(std::make_shared<se::dnn::LazyDnnGraph>(nullptr)),
+      sdpa_dropout_seed_(sdpa_dropout_seed) {
   args_.reserve(kernel_arguments.size());
   for (const KernelArgument& kernel_argument : kernel_arguments) {
     args_.push_back(kernel_argument.slice());
@@ -52,6 +54,11 @@ absl::Status CuDnnThunk::Initialize(const InitializeParams& params) {
     std::string().swap(fingerprint_);
     if (result.ok()) {
       graph_->swap(*result);
+      if (sdpa_dropout_seed_.has_value()) {
+        dynamic_cast<stream_executor::gpu::CudnnGraph*>(graph_->get())
+            ->InitDropoutState(params.local_device_count, *sdpa_dropout_seed_,
+                               16);
+      }
     }
     ret = result.status();
   });
@@ -68,7 +75,8 @@ absl::Status CuDnnThunk::ExecuteOnStream(const ExecuteParams& params) {
     buffer_args.push_back(params.buffer_allocations->GetDeviceAddress(arg));
   }
   return graph_->get()->Execute(*params.stream,
-                                absl::Span<se::DeviceMemoryBase>(buffer_args));
+                                absl::Span<se::DeviceMemoryBase>(buffer_args),
+                                params.collective_params->local_device_ordinal);
 }
 
 }  // namespace gpu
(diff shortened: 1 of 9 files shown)


Cjkkkk:
@akuegel Hi Adrian, could you take a look at this PR?

akuegel (commented):
Is it possible to add a test that would fail without this fix?

Cjkkkk:
> Is it possible to add a test that would fail without this fix?

@akuegel  Sorry for the late reply, we had free days. The unit test added will fail without this fix. 
```
[ RUN      ] FlashAttentionBMMScaleSoftmaxDropoutBMM.Flash_Attention_Training_BMM1_Softmax_Dropout_BMM2              
unknown file: Failure                                                                                                
C++ exception with description "Bad StatusOr access: INTERNAL: Uid 9 does not exist in variant pack.                 
in xla/stream_executor/cuda/cuda_dnn.cc(8292): 'graph_.execute( dnn_support.cudnn_ ->GetHandle(ExtractGpuExecutor(str
eam.parent()), &stream) .handle(), tensor_to_ptr_map, workspace.opaque())' " thrown in the test body.
```

Two buffers won't be passed to cudnn without the fix, thus result in a cudnn error.

akuegel:
> > Is it possible to add a test that would fail without this fix?
> 
> @akuegel Sorry for the late reply, we had free days. The unit test added will fail without this fix.
> 
> ```
> [ RUN      ] FlashAttentionBMMScaleSoftmaxDropoutBMM.Flash_Attention_Training_BMM1_Softmax_Dropout_BMM2              
> unknown file: Failure                                                                                                
> C++ exception with description "Bad StatusOr access: INTERNAL: Uid 9 does not exist in variant pack.                 
> in xla/stream_executor/cuda/cuda_dnn.cc(8292): 'graph_.execute( dnn_support.cudnn_ ->GetHandle(ExtractGpuExecutor(str
> eam.parent()), &stream) .handle(), tensor_to_ptr_map, workspace.opaque())' " thrown in the test body.
> ```
> 
> Two buffers won't be passed to cudnn without the fix, thus result in a cudnn error.

Ah, so sorry, I somehow only looked at the changes from the last commit to your branch, and forgot to change it to "changes from all commits". Yes, I see the test now.

xla/service/gpu/runtime/cudnn_thunk.cc
```diff
@@ -52,6 +58,13 @@ absl::Status CuDnnThunk::Initialize(const InitializeParams& params) {
     std::string().swap(fingerprint_);
     if (result.ok()) {
       graph_->swap(*result);
+#if GOOGLE_CUDA
+      if (sdpa_dropout_seed_.has_value()) {
+        dynamic_cast<stream_executor::gpu::CudnnGraph*>(graph_->get())
+            ->InitDropoutState(params.local_device_count, *sdpa_dropout_seed_,
```
akuegel, line 64: @beckerhe Do you have an idea how we can avoid the #ifdef here? So I think in your document, one of your suggestions is to move the logic to StreamExecutor, so in this case possibly create a InitDropoutState method for DnnGraph. As far as I can tell, DnnGraph currently is only implemented by cuda backend, so to avoid that we need to cast to CudnnGraph would require specifying some default implementations for DnnGraph that return Unimplemented by default.

beckerhe: 2 options:

1. If `CudnnThunk` was only part of the CUDA build, then the ifdefs are just not needed. They could be removed and the target could be tagged `no_rocm` and `gpu` in the `BUILD` file to exclude it from compilation in non-GPU non-CUDA builds. Unfortunately I believe `CudnnThunk` is also needed for the ROCm build despite its name. Also this wouldn't avoid the `dynamic_cast` which I would consider a code smell.

2. The better alternative is probably to make `InitDropoutState` a pure virtual function in `DnnGraph` and then override it in `CudnnGraph` (and other implementations). This way it could always be called here no matter if it's a CUDA build or not.

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1349,6 +1349,67 @@ class FlashAttentionBMMScalePaddingMaskSoftmaxBMMF8
   }
 };
 
+class FlashAttentionBMMScaleSoftmaxDropoutBMM
+    : public MultiHeadedAttentionTest {
+ protected:
+  const std::string  // NOLINT
+  GetModuleFlash_Attention_Training_BMM1_Softmax_Dropout_BMM2_HloString_BF16() {  // NOLINT
+    const std::string hlo_text = R"(
```
beckerhe, line 1357: Why the NOLINT and why the `const std::string`? This has no real benefit here?

I recommend making this a constant instead of a function:

`constexpr absl::string_view kModuleFlashAttentionTrainingBMM1SoftmaxDropoutBMM2HloStringBF16 = R"(...)";`

xla/service/gpu/runtime/cudnn_thunk.cc
Cjkkkk: I removed the #ifdef with option 2.

xla/service/gpu/tests/gpu_fused_mha_test.cc
Cjkkkk: Fixed.

akuegel (approved):
Thank you!

Cjkkkk:
@akuegel I think there is an error in `import/copybara`. Could you share some infos on this?

akuegel:
Not sure about the error, I don't know how I can see it, but I will try with just approving again, hopefully it will work this time.

akuegel (approved):

PR #16882, opened 2024-09-06
Symlink hermetic cuda headers to permit clang cuda version detection

### LINKED ISSUE #16877: Clang cannot detect hermetic cuda version
When using the hermetic CUDA toolchain with Clang, the CUDA version is not actually detected by Clang and defaults to the latest known toolkit version. Clang searches for `external/cuda_nvcc/include/cuda.h`, and when it is not found, the version detection logic switches to `CudaVersion::NEW`.

For instance, in clang 18.1.8, CudaVersion::NEW is 12.3. When we set `HERMETIC_CUDA_VERSION="12.2.0"`, this mismatch and failed detection causes clang to use isa 8.3 which is incompatible with cuda 12.2.0. 

The easiest fix is probably to copy `external/cuda_cudart/include/cuda.h` into `external/cuda_nvcc/include/cuda.h`.

tchatow:
Fixes #16877

### DIFF
--- a/third_party/tsl/third_party/gpus/cuda/hermetic/cuda_redist_init_repositories.bzl
+++ b/third_party/tsl/third_party/gpus/cuda/hermetic/cuda_redist_init_repositories.bzl
@@ -219,6 +219,10 @@ def _create_libcuda_symlinks(
         repository_ctx.symlink(nvidia_driver_path, "lib/libcuda.so.1")
         repository_ctx.symlink("lib/libcuda.so.1", "lib/libcuda.so")
 
+def _create_cuda_header_symlinks(repository_ctx):
+    if repository_ctx.name == "cuda_nvcc":
+        repository_ctx.symlink("../cuda_cudart/include/cuda.h", "include/cuda.h")
+
 def use_local_path(repository_ctx, local_path, dirs):
     # buildifier: disable=function-docstring-args
     """Creates repository using local redistribution paths."""
@@ -339,6 +343,7 @@ def _use_downloaded_cuda_redistribution(repository_ctx):
         repository_ctx,
         lib_name_to_version_dict,
     )
+    _create_cuda_header_symlinks(repository_ctx)
     repository_ctx.file("version.txt", major_version)
 
 def _cuda_repo_impl(repository_ctx):
--- a/.bazelrc
+++ b/.bazelrc
@@ -242,6 +242,8 @@ build:cuda_clang --copt=-Qunused-arguments
 # major release. Example: sm_80 kernels can run on sm_89 GPUs but
 # not on sm_90 GPUs. compute_80 kernels though can also run on sm_90 GPUs.
 build:cuda_clang --repo_env=HERMETIC_CUDA_COMPUTE_CAPABILITIES="sm_60,sm_70,sm_80,sm_89,compute_90"
+# Permit newer CUDA versions than Clang is aware of
+build:cuda_clang --copt="-Wno-unknown-cuda-version"
 # Set lld as the linker.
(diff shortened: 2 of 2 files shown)


beckerhe:
Looks good to me, but I would like @ybaturina to have a look.

third_party/tsl/third_party/gpus/cuda/hermetic/cuda_redist_init_repositories.bzl
```diff
@@ -218,6 +218,10 @@ def _create_libcuda_symlinks(
         repository_ctx.symlink(nvidia_driver_path, "lib/libcuda.so.1")
         repository_ctx.symlink("lib/libcuda.so.1", "lib/libcuda.so")
 
+def _create_cuda_header_symlinks(repository_ctx):
+    if repository_ctx.name == "cuda_nvcc":
+        repository_ctx.symlink("../cuda_cudart/include/cuda.h", "include/cuda.h")
```
ybaturina, line 223: Would you explain the algorithm here please, e.g. how does it work?
When does Clang use this symlink?

tchatow: Clang uses `cuda.h` in a location relative to nvcc to determine the cuda version. If it is missing, Clang assumes the latest known Cuda version which is usually incorrect. The hermetic toolchain breaks this by downloading nvcc and cudart to separate directories.

See: https://github.com/llvm/llvm-project/blob/b592917eec407a01ef07334af5aafe6c8922c80a/clang/lib/Driver/ToolChains/Cuda.cpp#L212

ybaturina: Thank you. Are you planning to use this change for CLANG builds (not CLANG+NVCC)?

tchatow: Yes, this is for CLANG builds.

tchatow: I think this change merely exposes an already broken build configuration. In your case you should set `-Wno-unknown-cuda-version` to get the same behavior as today. Defaulting to `CudaVersion::NEW` completely ignores the hermetic nvcc version and causes isa generation incompatible with ptxas.

ybaturina: In this case may I ask you to change .bazelrc file and add the following to the `build:cuda_clang` config:

`build:cuda_clang --copt="-Wno-unknown-cuda-version"`

And then I'll be able to approve it.

thomasjoerg:
ybaturina is OOO. 
@beckerhe do you have enough context to take over?

beckerhe:
> ybaturina is OOO. @beckerhe do you have enough context to take over?

I believe Yulia's ask from the comment chain above still stands. Once that's resolved it can be merged.

beckerhe:
Thank you. I think this is good to go but would like to trigger a new run of the CI. Unfortunately the PR needs to be synced to HEAD due to a breakage on main that has been fixed by now.

I don't have the permission to sync the PR. @thomasjoerg or @dimitar-asenov, can you sync the PR and approve a run of the CI?

beckerhe (approved):

tchatow:
Rebased

dimitar-asenov:
Thanks, Sorry for the delay on submitting this. I hope to have it in today.

dimitar-asenov (approved):

PR #16901, opened 2024-09-07
[XLA:GPU] Fix default device mesh for auto sharding

i-Pear:
When the user does not specify the number of GPUs for auto sharding, XLA defaults to using all available GPUs.

The current implementation uses the number of cores (SMs) on the GPU as the default shard count. For example, on an A100, the sharding algorithm will try to shard into 108 devices, which can be confusing for users.

This patch changes the shard count to the number of cards, which has been tested to work correctly on an 8-card A100 machine.


bixia1 (approved):

dimitar-asenov (approved):

cheshire:
Is it possible to test this?

dimitar-asenov (changes requested):
Please add a test for the new change.

PR #16913, opened 2024-09-08
[PJRT:GPU] Enable creating topology without a GPU device

jaro-sevcik:
Currently PJRT_TopologyDescription_Create always creates topology from the local client. This requires having a local GPU device.

This patch allows explicitly specifying topology shape and device config in PJRT_TopologyDescription_Create call, without querying local client. This enables deviceless compilation.

### DIFF
--- a/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
+++ b/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
@@ -170,19 +170,23 @@ PJRT_Error* PJRT_ExecuteContext_Create(PJRT_ExecuteContext_Create_Args* args) {
 namespace {
 
 struct TargetConfigAndDevices {
-  tensorflow::se::GpuTargetConfigProto target_config_proto;
+  stream_executor::GpuTargetConfigProto target_config_proto;
   std::vector<int> device_ids;
 };
 
+// Parses the 'target_config' entry in 'options'. The option is
+// parsed as GpuTargetConfigProto. If there is no 'target_config' in
+// 'options', the function falls back to creating a local client,
+// returning the local client's target config.
 absl::StatusOr<TargetConfigAndDevices> GetTargetConfigFromOptions(
-    absl::flat_hash_map<std::string, xla::PjRtValueType>& options) {
-  tensorflow::se::GpuTargetConfigProto target_config_proto;
+    const absl::flat_hash_map<std::string, xla::PjRtValueType>& options) {
+  stream_executor::GpuTargetConfigProto target_config_proto;
   std::vector<int> device_ids;
   std::string target_config_proto_string;
   const auto& debug_options = xla::GetDebugOptionsFromFlags();
 
-  auto target_config_it = options.find("target_config");
-  if (target_config_it != options.end()) {
+  if (auto target_config_it = options.find("target_config");
+      target_config_it != options.end()) {
     target_config_proto_string =
         std::get<std::string>(target_config_it->second);
     if (!tsl::protobuf::TextFormat::ParseFromString(target_config_proto_string,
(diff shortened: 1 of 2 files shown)


jaro-sevcik:
Here is an example of how this can be used from JAX: https://gist.github.com/jaro-sevcik/3495718bb04c6096c0f998fc29220c2b (it requires a one-line tweak in JAX, principled solution to be discussed with the JAX team).

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -167,12 +167,100 @@ PJRT_Error* PJRT_ExecuteContext_Create(PJRT_ExecuteContext_Create_Args* args) {
   return nullptr;
 }
 
+namespace {
+
+absl::StatusOr<std::unique_ptr<xla::StreamExecutorGpuTopologyDescription>>
+TryParseTopology(
+    xla::PjRtPlatformId platform_id, std::string platform_name,
+    absl::flat_hash_map<std::string, xla::PjRtValueType>& options) {
+  auto topology_it = options.find("topology");
```
cheshire, line 176: Who puts it there in the first place? The test just manually adds it to `options`, maybe we need a more e2e-like test?

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -167,12 +167,100 @@ PJRT_Error* PJRT_ExecuteContext_Create(PJRT_ExecuteContext_Create_Args* args) {
...
+      !absl::SimpleAtoi(topology_components[2], &num_devices_per_host)) {
+    return absl::InternalError(
+        "topology must be of shape "
+        "\"<num-slices>x<num-hosts-per-slice>x<num-devices-per-host>\"");
+  }
+
+  std::string target_config_proto_string;
+  const auto& debug_options = xla::GetDebugOptionsFromFlags();
+
+  auto target_config_it = options.find("target_config");
```
cheshire, line 200: TBH this part never made sense to me, if we already have a debug flag/a separate option to pass through debug_config, do we need to put it into `options` as well?

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -167,12 +167,100 @@ PJRT_Error* PJRT_ExecuteContext_Create(PJRT_ExecuteContext_Create_Args* args) {
...
+  tensorflow::se::GpuTargetConfigProto target_config_proto;
+  if (!tsl::protobuf::TextFormat::ParseFromString(target_config_proto_string,
+                                                  &target_config_proto)) {
+    return absl::FailedPreconditionError(
+        "Failed to parse GpuTargetConfigProto");
+  }
+
+  int num_devices = num_slices * num_hosts_per_slice * num_devices_per_host;
+  std::vector<int> device_ids;
+  device_ids.reserve(num_devices);
```
cheshire, line 220: Super-nit: absl::c_iota?

cheshire (changes requested):

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
jaro-sevcik: Currently, the options come from the kwargs of the [`jax.experimental.topologies.get_topology_desc`](https://github.com/google/jax/blob/baf9cc70bd4c9ad544615e54064fc209e1263fb9/jax/experimental/topologies.py#L49) call. See the [example in the gist](https://gist.github.com/jaro-sevcik/3495718bb04c6096c0f998fc29220c2b#file-aot-save-py-L54). So the idea is that the user will put it there.

By E2E example, do you mean creating the topology, then creating [compile-only-client](https://github.com/openxla/xla/blob/fe141265482a26d25318a5720d49963f9b45113c/xla/python/py_compile_only_client.cc#L352) and compiling with that client? Do we have a GPU-only XLA python test suite where I could put this?

Note that the compile-only part is already [somewhat tested in JAX](https://github.com/google/jax/blob/baf9cc70bd4c9ad544615e54064fc209e1263fb9/tests/aot_test.py#L103), this PR only makes it so that we do not require GPU device to run. (Unfortunately, I believe that the unit test does not really verify we can create the topology without a device.)

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
jaro-sevcik: So I am not a fan of the debug flag solution. My reasons:

* The name **debug** flag suggests it's for debugging. `target_config` is not for debugging.
* Global flag does not let you create a compile server that can compile for different GPUs easily.
* The `xla_gpu_target_config_filename` flag forces you to create a file for `target_config` and to use the filesystem. This is not always convenient.
* Something as important as target_config should be an explicit call parameter rather than a global flag.

Am I missing something?

cheshire (approved):
OK you're right, this makes sense!

reedwm:
@cheshire can you re-review and re-approve? The PR has been modified since your last approval.

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -167,34 +167,101 @@ PJRT_Error* PJRT_ExecuteContext_Create(PJRT_ExecuteContext_Create_Args* args) {
...
+  if (target_config_it != options.end()) {
+    target_config_proto_string =
+        std::get<std::string>(target_config_it->second);
+    if (!tsl::protobuf::TextFormat::ParseFromString(target_config_proto_string,
+                                                    &target_config_proto)) {
+      return absl::FailedPreconditionError(
+          "Failed to parse GpuTargetConfigProto "
+          "from the 'target_config' parameter.");
+    }
+  } else if (!debug_options.xla_gpu_target_config_filename().empty()) {
```
cheshire, line 194: TBH reading XLA_FLAGS here feels like layering violation, should we be using PjRT options instead? IMO we should use one or another, not both

cheshire (changes requested):

jaro-sevcik (commented):
@cheshire  Could you take another look?

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -167,33 +167,93 @@ PJRT_Error* PJRT_ExecuteContext_Create(PJRT_ExecuteContext_Create_Args* args) {
...
+};
+
+// Parses the 'target_config' entry in 'options'. The option is
+// parsed as GpuTargetConfigProto. If there is no 'target_config' in
+// 'options', the function falls back to creating a local client,
+// returning the local client's target config.
+absl::StatusOr<TargetConfigAndDevices> GetTargetConfigFromOptions(
+    const absl::flat_hash_map<std::string, xla::PjRtValueType>& options) {
+  stream_executor::GpuTargetConfigProto target_config_proto;
+  std::vector<int> device_ids;
```
cheshire, line 184: Super-nit: I wouldn't have this variable here at all, return `{}` under the branch, and then just create it in the last branch.

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -167,33 +167,93 @@ PJRT_Error* PJRT_ExecuteContext_Create(PJRT_ExecuteContext_Create_Args* args) {
...
+    absl::flat_hash_map<std::string, xla::PjRtValueType>& options) {
+  auto topology_it = options.find("topology");
+  if (topology_it == options.end()) {
+    return std::nullopt;
+  }
+
+  std::string topology_string = std::get<std::string>(topology_it->second);
+
+  TopologySizes sizes;
+  std::vector<std::string> topology_components =
```
cheshire, line 236: For simplicity I'd make that a static constructor method on TopologySizes::FromString, but I guess not blocking

cheshire (approved):

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
jaro-sevcik: Done. With the TopologySizes::FromString method, GetTopologyFromOptions got small, so I inlined it into PJRT_GpuDeviceTopology_Create.

jaro-sevcik (commented):
Addressed the comments.

jaro-sevcik:
@cheshire Could you take one more look?

cheshire (approved):

beckerhe (approved):

PR #16921, opened 2024-09-09
[PJRT:GPU] Treat GPU collective memory space as device memory space

zhenying-liu:
This is a regression fix when using --xla_gpu_enable_nccl_user_buffers=true.
Return device memory space when collective memory space is used as an output on GPU.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
@@ -1236,7 +1236,7 @@ TEST(StreamExecutorGpuClientTest,
 
   auto device = client->addressable_devices()[0];
   TF_EXPECT_OK(device->default_memory_space());
-  TF_ASSIGN_OR_RETURN(
+  TF_ASSERT_OK_AND_ASSIGN(
       auto input, client->BufferFromHostBuffer(
                       data.data(), shape.element_type(), shape.dimensions(),
                       /*byte_strides=*/std::nullopt,
--- a/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
@@ -1131,6 +1131,24 @@ constexpr char const* kD2HProgramTupleOutput = R"(
   }
 )";
 
+constexpr char const* kCollectiveMemorySpaceOutput = R"(
+
+  HloModule jit__psum, entry_computation_layout={(s32[1,4]{1,0})->s32[4]{0}}
+
+  region_0.3 {
+    Arg_0.0 = s32[] parameter(0)
+    Arg_1.0 = s32[] parameter(1)
+    ROOT add.0 = s32[] add(Arg_0.0, Arg_1.0)
+  }
+
+  ENTRY main.10_spmd {
+    param = s32[1,4]{1,0} parameter(0)
+    reshape = s32[4]{0} reshape(param)
+    ROOT all-reduce = s32[4]{0} all-reduce(reshape), channel_id=1, to_apply=region_0.3
+  }
+
+)";
+
 }  // namespace
 
 TEST(StreamExecutorGpuClientTest, ExecutePinnedHostOutputTest) {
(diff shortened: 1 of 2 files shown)


NaiyerRizz:
Hi @zhenying-liu 
Seems like there are some conflicts in the branch, Can you please resolve those.
Thank you

zhenying-liu:
> Hi @zhenying-liu Seems like there are some conflicts in the branch, Can you please resolve those. Thank you

Thanks. Resolved conflicts.

hawkinsp (approved):

cheshire:
Any chance we could separate out XLA and PJRT changes? 

I'm somewhat new to that code, but hardcoding memory spaces in layout.h seems strange, is that the usual approach?

cheshire:
Is it possible to describe in general terms, not in specific fixes, what the change is doing? Is it fixing a crash? Is it performance improvement? If it's a bullet list of five things, maybe it should be 5 PRs?

zhenying-liu:
> Any chance we could separate out XLA and PJRT changes?
> 
> I'm somewhat new to that code, but hardcoding memory spaces in layout.h seems strange, is that the usual approach?

It's defined in XLA. It's harder to separate XLA and PJRT. But I can try my best to split.
I think defining memory spaces in layout.h is the best way that I can think of. Otherwise, defining it in different files could cause the conflicts. Currently both fast memory space and collect memory space has the value 1.

zhenying-liu:
> Is it possible to describe in general terms, not in specific fixes, what the change is doing? Is it fixing a crash? Is it performance improvement? If it's a bullet list of five things, maybe it should be 5 PRs?

There is a crash when using NCCL user buffer: --xla_gpu_enable_nccl_user_buffers=true. The error message is: "INVALID_ARGUMENT: Unexpected memory space 1 in output layout". It happened in pjrt_stream_executor_client.cc:MemoryKindFromSimpleShape().  
Jax doesn't support any memory kind other than device and pinned_host. In this case, the output memory kind is neither device nor pinned_host. Our strategy is to treat CollectiveMemorySpace as device memory space to please Jax. But inside XLA, we keep using collective memory space to call nccl memory allocator to handle allocation/deallocation. The existing memory space color for collective memory space is 1, conflicting with Layout::kGenericFastMemorySpace. So we define a constant kCollectiveMemorySpace in layout.h with a unique value 6, and use this constant whenever it is needed.

nouiz:
> > Is it possible to describe in general terms, not in specific fixes, what the change is doing? Is it fixing a crash? Is it performance improvement? If it's a bullet list of five things, maybe it should be 5 PRs?
> 
> There is a crash when using NCCL user buffer: --xla_gpu_enable_nccl_user_buffers=true. The error message is: "INVALID_ARGUMENT: Unexpected memory space 1 in output layout". It happened in pjrt_stream_executor_client.cc:MemoryKindFromSimpleShape(). Jax doesn't support any memory kind other than device and pinned_host. In this case, the output memory kind is neither device nor pinned_host. Our strategy is to treat CollectiveMemorySpace as device memory space to please Jax. But inside XLA, we keep using collective memory space to call nccl memory allocator to handle allocation/deallocation. The existing memory space color for collective memory space is 1, conflicting with Layout::kGenericFastMemorySpace. So we define a constant kCollectiveMemorySpace in layout.h with a unique value 6, and use this constant whenever it is needed.

This is a regression fix when using --xla_gpu_enable_nccl_user_buffers=true
Jane, can you update the description to tell that?

nouiz:
> Is it possible to describe in general terms, not in specific fixes, what the change is doing? Is it fixing a crash? Is it performance improvement? If it's a bullet list of five things, maybe it should be 5 PRs?

This is a regression fix. Google JAX team want to make a releases, so merging this soon would be great.
@hawkinsp approved the changes, so all PJRT part should be good.
Splitting into multiple from outside would take extra times and this compete with doing a release of JAX very soon. So I let you @cheshire and @hawkinsp discuss that. Having regression in public version isn't great.

xla/layout.h
```diff
@@ -404,6 +404,7 @@ class Layout {
   static constexpr int64_t kDefaultMemorySpace = 0;
   static constexpr int64_t kGenericFastMemorySpace = 1;
   static constexpr int64_t kHostMemorySpace = 5;
+  static constexpr int64_t kCollectiveMemorySpace = 6;
```
cheshire, line 407: Does this enum have to live in `layout.h` ? Maybe we could move it to a GPU-specific file? After all , GPU has to not collide only with GPU. Also having to recompiling XLA on any changes to this enum does not sound great.

cheshire (changes requested):

xla/layout.h
jaro-sevcik: At the same time, it would be good if the constants would not collide, so I would prefer to have all the constants in one place. How about creating a new header file just for the memory space constants?

jaro-sevcik: @cheshire Conceptually what you suggest sounds consistent. I would be concerned that someone takes a layout from an array, tries to use for some other device and gets surprising results without any warning, but perhaps that's already not possible because one cannot really sneak in memory spaces in layout.

In practice, this PR is holding up a JAX release, so we should decide what we want soon. Are you proposing to undo the move/rename of the kCollectiveMemorySpaceColor constant and change of its value to 6 (without worrying about the collective space id 1 colliding with kGenericFastMemorySpace) and just keep the changes to xla/pjrt/pjrt_stream_executor_client.cc and to xla/pjrt/gpu/se_gpu_pjrt_client_test.cc?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1197,6 +1215,50 @@ TEST(StreamExecutorGpuClientTest, ExecutablePinnedHostOutputMemoryKindTest) {
...
+                      /*on_done_with_host_buffer=*/nullptr, device));
+  EXPECT_EQ(input->memory_space()->kind(), "device");
+
+  TF_ASSERT_OK_AND_ASSIGN(auto memory_kinds,
+                          executable->GetOutputMemoryKinds());
+  EXPECT_EQ(memory_kinds.size(), 1);
+  EXPECT_EQ(memory_kinds[0].size(), 1);
+  EXPECT_EQ(memory_kinds[0][0], "device");
+
+  TF_ASSERT_OK_AND_ASSIGN(
```
cheshire, line 1253: What happens if we execute it without the changes in this PR?

zhenying-liu: It will have the assertion failure in pjrt_stream_execution_client.cc:MemoryKindFromSimpleShape():
`INVALID_ARGUMENT: Unexpected memory space 1 in output layout`

xla/layout.h
zhenying-liu: @cheshire I have the smallest changes to xla/pjrt/pjrt_stream_executor_client.cc and xla/pjrt/gpu/se_gpu_pjrt_client_test.cc ready. With that one, I have to use kGenericFastMemorySpace for collective space id 1. If so, we can fix the regression now and do the clean up the definition later.

zhenying-liu: @cheshire Please take another look at the code review as I changed the code as mentioned above.

nouiz:
Note, this PR was updated with the minimal fix to the regression.
The refactoring/code clean-up can be done after the JAX releases.
@cheshire @hawkinsp

cheshire (approved):

nouiz:
This approved PR is failing in Google internal import.
Who can look at it? @xla-rotation

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1197,6 +1215,50 @@ TEST(StreamExecutorGpuClientTest, ExecutablePinnedHostOutputMemoryKindTest) {
...
+      CompileExecutable(kCollectiveMemorySpaceOutput, *client, options));
+  std::vector<int32_t> data{1, 2, 3, 4};
+  // Build the input shape with the correct memory space set.
+  Shape shape = ShapeUtil::MakeShapeWithDenseLayout(S32, {1, 4},
+                                                    /*major_to_minor=*/{1, 0});
+  shape.mutable_layout()->set_memory_space(Layout::kDefaultMemorySpace);
+
+  auto device = client->addressable_devices()[0];
+  TF_EXPECT_OK(device->default_memory_space());
+  TF_ASSIGN_OR_RETURN(
```
akuegel, line 1239: void function 'TestBody' should not return a value [-Wreturn-mismatch]
 1239 |   TF_ASSIGN_OR_RETURN(

akuegel (commented):
Merging this PR is blocked on a compile error. I wonder whether this actually compiles on your end?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
zhenying-liu: Thanks for pointing this out. Sorry for the mistake. Fixed.

akuegel (approved):

PR #16950, opened 2024-09-09
Clarify that sm_90a should only be used on Hopper

dimvar:
Same as the comment in xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.cc, from the original change:
https://github.com/openxla/xla/commit/754b68363d48215baf90f32288d411406761af9f


NaiyerRizz (approved):

cheshire:
If the change is marked as "NFC" it's simpler to review :P

cheshire (approved):

PR #16954, opened 2024-09-09
[ds-fusion] Avoid dynamic-slice-fusion when reduce-scatter has tuple output

shraiysh:
Tuple outputs are not handled while emitting thunks for dynamic-slice fusion. This patch avoids fusing them altogether while the support for multiple outputs is implemented.


akuegel (approved):

shraiysh:
Are there any internal failures with this PR?

PR #16959, opened 2024-09-09
[cuda] Cache the results of slow cuda_asm_compiler functions

sergey-kozub:
Following the comments in https://github.com/openxla/xla/pull/16734

Implementation of `GetAsmCompilerVersion` and `GetNvLinkVersion` methods is rather slow (they search for an executable and create a process), making it impractical to repeatedly call them during the compilation without caching.

The previous approach implemented in `reduction_utils.cc` is not quite correct, as the cached value doesn't depend on the HloModule options (as Adrian pointed out in PR#16374 it's a hack).

This PR extracts the common code path to `util.h` as a template function, and updates the call sites. Note that there's a separate cache per each init function.

### DIFF
--- a/xla/service/gpu/reduction_utils.cc
+++ b/xla/service/gpu/reduction_utils.cc
@@ -17,7 +17,6 @@ limitations under the License.
 
 #include <algorithm>
 #include <array>
-#include <atomic>
 #include <cstdint>
 #include <ostream>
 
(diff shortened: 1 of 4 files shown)


sergey-kozub:
Same as before, clang-format CI complains about unrelated code in `util.h` (and I cannot reproduce the failure locally).

beckerhe:
Hey Sergey,
regarding the slowness:

> Implementation of GetAsmCompilerVersion and GetNvLinkVersion methods is rather slow (they search for an executable and create a process), making it impractical to repeatedly call them during the compilation without caching.

I don't see that. `GetToolVersion` is already cached and it's the only function that actually does IO. The other part is `FindCudaExecutable` which basically constructs a vector of a handful of potential paths and then iterates through that list.

Is that really so slow that it justifies some added complexity - in particular since this is only called a few times?

sergey-kozub:
> Is that really so slow that it justifies some added complexity - in particular since this is only called a few times?

This was suggested in the review by Johannes in https://github.com/openxla/xla/pull/16734#discussion_r1746568049

I'm totally fine with dropping this PR.

sergey-kozub:
@jreiffers

jreiffers:
Well, I qualified my suggestion with `If the comment in reduction_utils is true...` :-). Looks like Henning thinks it isn't.

sergey-kozub:
> Well, I qualified my suggestion with `If the comment in reduction_utils is true...` :-). Looks like Henning thinks it isn't.

Nothing to do here then. Marked as draft.

PR #17013, opened 2024-09-10
[GPU] Gracefully handle ptxas insufficient registers error.

sergachev:


xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
```diff
@@ -833,24 +833,32 @@ GemmFusionAutotunerImpl::CompileAll(AutotunerCompileUtil& compile_util,
...
+              if (has_executable.ok()) {
+                log(has_executable.value());
+              } else {
+                log(/*success=*/false);
+                LOG(WARNING)
+                    << "Failure occured when compiling fusion "
+                    << fusion->name() << " with config '" << ToString(config)
+                    << "'\nFused HLO computation:\n"
+                    << fusion->fused_instructions_computation()->ToString();
+                if (!debug_options_.xla_gpu_exhaustive_tiling_search()) {
```
cheshire, line 856: Wait are you saying we still CHECK-fail otherwise? I definitely have seen cases where we fail to compile some of the tilings, but the execution continues. Or is it another check?

Forking the logic like that indeed does not seem ideal. (neither is crash-checking, i guess it could update a shared status with a mutex)

xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc
```diff
@@ -883,6 +883,24 @@ ENTRY e {
       [](const TritonGemmConfig& config) { return config.block_k > 16; }));
 }
 
+TEST_F(GemmFusionAutotunerExhaustiveTest, FailingConfigsDoNotAbortCompilation) {
+  CheckTritonAutotuning(R"(
```
cheshire, line 887: Does this currently crash?

sergachev: Yes.

xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
sergachev: > Wait are you saying we still CHECK-fail otherwise?

Yes.

xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc
cheshire: Could you provide the crash output?

sergachev:
```
[ RUN      ] GemmFusionAutotunerExhaustiveTest.FailingConfigsDoNotAbortCompilation
2024-09-10 16:07:22.894095: F xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:846] Non-OK-status: has_executable.status()
Status: INTERNAL: ptxas exited with non-zero error code 65280, output: ptxas fatal   : (C7602) Insufficient registers (128) to compile instruction at line 320 in function c_1. Try to compile with register target of 154 or higher.
ptxas fatal   : Ptx assembly aborted due to errors
: If the error message indicates that a file could not be written, please verify that sufficient filesystem space is provided.Failure occured when compiling fusion c.1 with config '{block_m:128,block_n:512,block_k:32,split_k:1,num_stages:1,num_warps:16,num_ctas:1}'
Fused HLO computation:
%f (a: f16[2048,2048], b: f16[2048,2048]) -> f16[2048,2048] {
  %a = f16[2048,2048]{1,0} parameter(0)
  %b = f16[2048,2048]{1,0} parameter(1)
  ROOT %c = f16[2048,2048]{1,0} dot(f16[2048,2048]{1,0} %a, f16[2048,2048]{1,0} %b), lhs_contracting_dims={1}, rhs_contracting_dims={0}
}
[5 / 6] Testing //xla/service/gpu/autotuning:gemm_fusion_autotuner_test_gpu_h100; 24s local
FAIL: //xla/service/gpu/autotuning:gemm_fusion_autotuner_test_gpu_h100 (see /.../test.log)
Target //xla/service/gpu/autotuning:gemm_fusion_autotuner_test_gpu_h100 up-to-date:
  bazel-bin/xla/service/gpu/autotuning/gemm_fusion_autotuner_test_gpu_h100
INFO: Elapsed time: 35.539s, Critical Path: 34.79s
INFO: 6 processes: 1 internal, 5 local.
INFO: Build completed, 1 test FAILED, 6 total actions
//xla/service/gpu/autotuning:gemm_fusion_autotuner_test_gpu_h100         FAILED in 24.3s
 ```

File: xla/service/gpu/autotuning/gemm_fusion_autotuner.cc (hunk 1 above)
Comment (cheshire) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:856]: I'm still a bit hesitant about adding a conditional there, seems weird to only not-crash in exhaustive mode, what stops the non-exhaustive mode from crashing. I'm really confused though how I was seeing those as log messages and not as crashes.

File: xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc (hunk 2 above)
Comment (cheshire) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc:887]: Yeah exactly, those are the ones I was seeing.

File: xla/service/gpu/autotuning/gemm_fusion_autotuner.cc (hunk 1 above)
Comment (sergachev) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:856]: > what stops the non-exhaustive mode from crashing

Probably that all the fixed configs used in the non-exhaustive mode are rather sane.

Comment (sergachev) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:856]: > I'm still a bit hesitant

How do you suggest to fix the problem then?

Comment (cheshire) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:856]: @vwbaker suggests that there's already some magic to skip such tilings, does it pass if you update the regex here to include your error? https://github.com/openxla/xla/blob/main/xla/stream_executor/cuda/cuda_asm_compiler.cc#L385

Comment (sergachev) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:856]: yes, that works

Comment (cheshire) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:856]: do you want to change the PR to do that? You probably have newer/internal ptxas?

Comment (sergachev) [reply xla/service/gpu/autotuning/gemm_fusion_autotuner.cc:856]: > do you want to change the PR to do that? 

OK

> You probably have newer/internal ptxas?

Nothing special - Cuda compilation tools, release 12.5, V12.5.40, Build cuda_12.5.r12.5/compiler.34177558_0

File: xla/stream_executor/cuda/cuda_asm_compiler.cc [hunk 3]
```diff
@@ -381,10 +381,14 @@ absl::StatusOr<std::vector<uint8_t>> CompileGpuAsmUsingPtxAs(
...
     }
-    if (absl::StrContains(stderr_output, "ptxas fatal") &&
-        absl::StrContains(stderr_output, "Register allocation failed")) {
-      LOG(INFO) << stderr_output;
-      return absl::ResourceExhaustedError("Register allocation failed");
+    if (absl::StrContains(stderr_output, "ptxas fatal")) {
+      if (absl::StrContains(stderr_output, "Register allocation failed")) {
+        LOG(INFO) << stderr_output;
+        return absl::ResourceExhaustedError("Register allocation failed");
+      } else if (absl::StrContains(stderr_output, "Insufficient registers")) {
```
cheshire, line 388: Super-nit: maybe group them into one and just give it full stderr_output?

cheshire (changes requested):

cheshire (approved):

reedwm:
I'm seeing the test time out after 15 minutes. The test normally passes in less than 3 minutes so the newly added test case takes at least 12 minutes. The test outputs thousands of messages like

```
E0912 17:10:32.285204    7528 slow_operation_alarm.cc:65] 
********************************
[Compiling module rr] Slow compile? XLA was built without compiler optimizations, which can be slow. Try rebuilding with -c opt.
********************************
E0912 17:10:32.285276    7528 slow_operation_alarm.cc:65] 
********************************
[Compiling module rr] Slow compile? XLA was built without compiler optimizations, which can be slow. Try rebuilding with -c opt.
********************************
```

Despite these messages, the test is built with `-c opt`, but we also pass `--copt=-UNDEBUG` which probably causes the messages to appear. @sergachev, do you know why the test takes so long and if this can be fixed?

sergachev:
The test with exhaustive autotuning I added was compiling hundreds of kernels. It was taking only ~15 seconds on Hopper with ~200 CPU threads but can be slower elsewhere. I removed the test because the new behavior is also captured by the change to DoNotRunAutotuningKernelSpillingRegisters.

cheshire (approved):

xla/stream_executor/cuda/ptx_compiler_support.h
```diff
@@ -16,10 +16,15 @@ limitations under the License.
...
 
+#include "absl/strings/string_view.h"
+
 namespace stream_executor {
 // Returns true if XLA was built with libnvptxcompiler support. Otherwise false
 // is returned.
 bool IsLibNvPtxCompilerSupported();
+
+// Checks whether ptxas log contains errors related to register allocation.
+bool IsPtxRegisterAllocationError(absl::string_view);
```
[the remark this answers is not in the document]
sergachev: The file isn't specific to libnvptxcompiler either, it's called ptx_compiler_support. I can add another pair of files and a target, but does every single-line function really need a separate target?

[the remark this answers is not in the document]
beckerhe: This file is specific to libnvptxcompiler, it's just the name which is unfortunately not aligned.

The problem with this particular target is that it's supposed to be included in builds that don't support CUDA. It's a single function that returns a bool - no additional and dependencies. By putting this functions into this file we are adding more dependencies to the target and are opening the door for adding more and more functions and dependencies into a target which should be as lean as possible.

But the better argument anyway is that it just doesn't belong here from a code structure perspective.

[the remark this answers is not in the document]
sergachev: https://github.com/openxla/xla/pull/17412

PR #17021, opened 2024-09-10
[NV] Use FP8 conversion intrinsics, when available

sergey-kozub:
The previous #16734 was rolled back. This PR addresses some comments from it.

PTX "cvt" instruction supports converting to/from FP8 types. The NV hardware supports E4M3FN and E5M2 types. This PR updates the MLIR emitter to use this instruction instead of emitting a long sequence of operations (this matters in compute-bound FP8 kernels).

The NVVM intrinsic allows converting two FP8 values with a single instruction, but as the emitter is elementwise, only one of the inputs is used. This is wasteful, but still much faster than emitting the sequence of instructions.

Before ptx 7.8 (cuda 11.8), the instruction is not supported. Starting with ptx 8.1 (cuda 12.1), the instruction is supported for sm89+. Between those versions, the instruction is supported for sm90+, thas is, if trying to compile on Ada (sm89) with cuda version < 12.1, the ptxas will complain..

Reference:
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cvt (see "PTX ISA Notes" and "Target ISA Notes").

### DIFF
--- a/xla/service/gpu/fusions/transforms/BUILD
+++ b/xla/service/gpu/fusions/transforms/BUILD
@@ -33,6 +33,7 @@ gentbl_cc_library(
 cc_library(
     name = "passes",
     srcs = [
+        "convert_float_nvidia.cc",
         "convert_xla_gpu_pure_call_ops.cc",
         "erase_dead_functions.cc",
         "expand_float_ops.cc",
@@ -60,6 +61,7 @@ cc_library(
         "//xla/service/gpu:ir_emission_utils",
         "//xla/service/gpu/fusions/ir:xla_gpu",
         "//xla/service/gpu/fusions/mlir:elemental_hlo_to_mlir",
+        "//xla/service/gpu/llvm_gpu_backend:llvm_gpu_backend",
         "//xla/service/gpu/model:indexing_analysis",
         "//xla/stream_executor:device_description",
         "@com_google_absl//absl/algorithm:container",
--- a/xla/service/gpu/fusions/transforms/convert_float_nvidia.cc
+++ b/xla/service/gpu/fusions/transforms/convert_float_nvidia.cc
@@ -1,0 +1,268 @@
+/* Copyright 2024 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


xla/service/gpu/fusions/transforms/convert_float_nvidia.cc
```diff
@@ -0,0 +1,269 @@
...
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/ImplicitLocOpBuilder.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/Value.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Support/LLVM.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "xla/service/gpu/fusions/transforms/passes.h"
+#include "xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.h"
+#include "xla/stream_executor/cuda/cuda_asm_compiler.h"
```
beckerhe, line 36: This include is now not needed anymore, right?

xla/service/gpu/fusions/transforms/BUILD
```diff
@@ -48,7 +49,12 @@ cc_library(
         "simplify_arith.cc",
         "unswitch_loops.cc",
         "vectorize_loads_stores.cc",
-    ],
+    ] + if_cuda_is_configured(
+        [
+            "convert_float_nvidia.cc",
+        ],
+        no_cuda = ["cuda_stub.cc"],
+    ),
```
beckerhe, line 57: Since the .cc file now does not depend on CUDA-only targets anymore, it can be a normal src file, right?

Also the stub is not needed anymore, right?

xla/service/gpu/llvm_gpu_backend/gpu_backend_lib.h
```diff
@@ -66,8 +66,8 @@ absl::StatusOr<std::string> CompileToPtx(
     std::function<void(llvm::TargetMachine*)> configure_target = nullptr);
 
 // Determine PTX version from CUDA version.
-using Version = std::pair<int, int>;
-Version DetermineHighestSupportedPtxVersionFromCudaVersion(
+stream_executor::SemanticVersion
+DetermineHighestSupportedPtxVersionFromCudaVersion(
```
beckerhe, line 70: Thank you for this update!

xla/service/gpu/fusions/transforms/BUILD
sergey-kozub: Indeed, that was the whole point. Done.

xla/service/gpu/fusions/transforms/convert_float_nvidia.cc
sergey-kozub: Removed.

beckerhe (approved):
Awesome! LGTM!

PR #17133, opened 2024-09-13
Dedup LiteralComparisonTests

apivovarov:
Deduplicate LiteralComparisonTests using TYPED_TEST

### DIFF
--- a/xla/literal_comparison_test.cc
+++ b/xla/literal_comparison_test.cc
@@ -26,173 +26,67 @@ limitations under the License.
 namespace xla {
 namespace {
 
-TEST(LiteralComparisonTest, F8E4M3FNCompareNear_Equal) {
-  auto actual =
-      LiteralUtil::CreateR0<tsl::float8_e4m3fn>(tsl::float8_e4m3fn(8.0));
-  auto expected =
-      LiteralUtil::CreateR0<tsl::float8_e4m3fn>(tsl::float8_e4m3fn(8.0));
+template <typename T>
+class LiteralComparisonTest : public ::testing::Test {};
+
+using TestedTypes = ::testing::Types<tsl::float8_e4m3fn,
+                                     tsl::float8_e4m3b11fnuz, tsl::float8_e5m2>;
+TYPED_TEST_SUITE(LiteralComparisonTest, TestedTypes);
+
+TYPED_TEST(LiteralComparisonTest, CompareNear_Equal) {
+  auto actual = LiteralUtil::CreateR0<TypeParam>(TypeParam(8.0));
+  auto expected = LiteralUtil::CreateR0<TypeParam>(TypeParam(8.0));
   TF_EXPECT_OK(literal_comparison::Near(actual, expected, ErrorSpec(0.0, 0.0),
                                         /*detailed_message=*/false,
                                         /*miscompare_callback=*/nullptr));
 }
 
-TEST(LiteralComparisonTest, F8E4M3FNCompareNear_NotEqual_1ulp) {
-  auto actual =
-      LiteralUtil::CreateR0<tsl::float8_e4m3fn>(tsl::float8_e4m3fn(8.0));
-  auto expected =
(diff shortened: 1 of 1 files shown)


xla/literal_comparison_test.cc
```diff
@@ -25,173 +25,115 @@ limitations under the License.
...
+struct TypeTraits {
+  static constexpr float expected_1ulp = 9.0;
+  static constexpr int within_n_values_1ulp = 1;
+  static constexpr float expected_4ulp = 12.0;
+  static constexpr float expected_using_float_4ulp = 12.1;
+};
+
+template <>
+struct TypeTraits<tsl::float8_e4m3fn> : public TypeTraits<void> {
+  static constexpr PrimitiveType primitive_type = PrimitiveType::F8E4M3FN;
```
reedwm, line 38: Instead of having this field, you can just have `GetPrimitiveType` below call [`NativeToPrimitiveType`](https://github.com/openxla/xla/blob/536ba0b7d74f6637a7a772471a99ecf4f578aef2/xla/primitive_util.h#L78)

xla/literal_comparison_test.cc
```diff
@@ -25,173 +25,115 @@ limitations under the License.
...
+template <>
+struct TypeTraits<tsl::float8_e4m3b11fnuz> : public TypeTraits<void> {
+  static constexpr PrimitiveType primitive_type = PrimitiveType::F8E4M3B11FNUZ;
+};
+
+template <>
+struct TypeTraits<tsl::float8_e5m2> : public TypeTraits<void> {
+  static constexpr PrimitiveType primitive_type = PrimitiveType::F8E5M2;
+  static constexpr float expected_1ulp = 10.0;
+  static constexpr int within_n_values_1ulp = 4;
```
reedwm, line 50: I'm not sure what this field is. The `NotEqual_1ulp` tests always passed `within_n_values=1` before this PR.

xla/literal_comparison_test.cc
```diff
@@ -25,173 +25,115 @@ limitations under the License.
...
-  auto actual =
-      LiteralUtil::CreateR0<tsl::float8_e4m3fn>(tsl::float8_e4m3fn(8.0));
-  auto expected =
-      LiteralUtil::CreateR0<tsl::float8_e4m3fn>(tsl::float8_e4m3fn(8.0));
+template <typename T>
+struct TypeTraits {
+  static constexpr float expected_1ulp = 9.0;
+  static constexpr int within_n_values_1ulp = 1;
+  static constexpr float expected_4ulp = 12.0;
+  static constexpr float expected_using_float_4ulp = 12.1;
```
reedwm, line 33: It's unclear what these values are without reading the tests. Instead, I think we should probably define this values inline with the tests, instead of having type traits. For example, we could have CompareNear_NotEqual_1ulp start with

```c++
TYPED_TEST(LiteralComparisonTest, CompareNear_NotEqual_1ulp) {
  PrimitiveType type = GetPrimitiveType();
  TypeParam 1ulp_from_8 = type == F8E5M2 ? TypeParam{10} : TypeParam{9};
  auto actual = LiteralUtil::CreateR0<TypeParam>(TypeParam(8.0));
  auto expected = LiteralUtil::CreateR0<TypeParam>(1ulp_from_8);
  ...
```

reedwm (changes requested):

xla/literal_comparison_test.cc
apivovarov: fixed

xla/literal_comparison_test.cc
apivovarov: agree. fixed.

xla/literal_comparison_test.cc
apivovarov: removed TypeTraits. Inlined value selection looks much better.

reedwm (approved):
Thanks! This test is a lot cleaner and simpler now that it doesn't duplicate each test for each FP8 dtype!

PR #17142, opened 2024-09-13
[ROCm] Disable gemm triton fusions for ROCm

zoranjovanovic-ns:
Until autotuner is functional, avoid preformance drop.

### DIFF
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -85,6 +85,13 @@ class GpuCompilerTest : public HloTestBase {
     return tensorflow::down_cast<GpuCompiler*>(compiler)
         ->RunPostSchedulingPipelines(module, 4 * 1024 * 1024, gpu_device_info);
   }
+
+  const stream_executor::GpuComputeCapability& GpuComputeComp() {
+    return backend()
+        .default_stream_executor()
+        ->GetDeviceDescription()
+        .gpu_compute_capability();
+  }
 };
 
 TEST_F(GpuCompilerTest, CompiledProgramsCount) {
@@ -970,6 +977,10 @@ using GpuCompilerPassTest = GpuCompilerTest;
 
 TEST_F(GpuCompilerPassTest,
        GpuCompilerRunsTritonGemmRewriterByDefaultFromAmpere) {
+  if (std::holds_alternative<se::RocmComputeCapability>(GpuComputeComp())) {
+    GTEST_SKIP() << "TritonGemmRewriter disabled for ROCm until autotuner "
+                 << "is included.";
+  }
   auto cc = backend()
                 .default_stream_executor()
                 ->GetDeviceDescription()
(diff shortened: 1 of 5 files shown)


xla/debug_options_flags.cc
```diff
@@ -189,7 +189,12 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_partitioning_algorithm(
       DebugOptions::PARTITIONING_ALGORITHM_NOOP);
 
+#ifdef TENSORFLOW_USE_ROCM
```
dimitar-asenov, line 192: We are actively removing `#ifdef` from XLA's code-base and also avoid new ones wherever possible.

Instead of setting this here, please adjust the code where the flag is used:
https://github.com/openxla/xla/blob/main/xla/service/gpu/gpu_compiler.cc#L1455-L1458

xla/service/gpu/fusions/triton/BUILD
dimitar-asenov, line 1: Most of the changes in this file are not related to GEMMs, but Triton fusions in general (e.g. kernels with reductions, transpositions, elementwise ops, reshapes). Is this your intention? If so, please adjust the PR description to say that you disable the generic triton emitter tests for ROCm and explain why.

If at all possible, I would advise to leave this file as is. Perhaps my other suggested change will enable this.

dimitar-asenov (changes requested):

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -970,6 +977,10 @@ using GpuCompilerPassTest = GpuCompilerTest;
 
 TEST_F(GpuCompilerPassTest,
        GpuCompilerRunsTritonGemmRewriterByDefaultFromAmpere) {
+  if (std::holds_alternative<se::RocmComputeCapability>(GpuComputeComp())) {
```
dimitar-asenov, line 980: I'm confused about this code. A few lines below there is an expectation that the triton gemm rewriter has run on rocm.

Can you please do two things:
- Explain why the rewriter is no longer needed. Did it ever run? What changed?
- Remove the `is_rocm` expecation below (lines 989-995).

dimitar-asenov (changes requested):

xla/service/gpu/gpu_compiler_test.cc
zoranjovanovic-ns: @dimitar-asenov 
> * Explain why the rewriter is no longer needed. Did it ever run? What changed?

The reason why we want, temporarily, to remove rewriter is that we have noticed significant performance drop when it is executed.
Performance drop is most probably related to the fact that, currently, triton gemm autotuner is not executed for rocm, 

Idea with this PR is to, temporarily, switch off triton gemm rewriter until implementation of rocm autotuner is finished, after that perform benchmarking, performance analysis and then switch it on again.

dimitar-asenov (approved):

PR #17173, opened 2024-09-13
[NVIDIA GPU] Optimize collective matmul loops when contracting dim is sharded

Tixxx:
When contracting dim is sharded, both LHS and RHS will be sharded and the partial dots' results will be accumulated. However the accumulation is done 1 step at a time which leads to a number of individual add kernels, these add kernels will impact overlap between gemms due to:
1. they each occupy SMs which prevent gemm overlap
2. they might be fused by gemm_rewriter which introduces data dependency between gemms

This pr optimizes it to get rid of partial accumulations, instead all partials will be concatenated into a contiguous buffer and a global reduction is applied on the buffer.

### DIFF
--- a/xla/service/gpu/transforms/windowed_einsum_handler.cc
+++ b/xla/service/gpu/transforms/windowed_einsum_handler.cc
@@ -467,6 +467,24 @@ bool ShouldAddToChain(const HloInstruction* inst) {
       return false;
   }
 }
+
+HloComputation* MakeSumComputation(PrimitiveType type, HloModule* module) {
+  HloComputation::Builder sum_b("add");
+  auto x = sum_b.AddInstruction(HloInstruction::CreateParameter(
+      /*parameter_number=*/0, ShapeUtil::MakeShape(type, {}), "x"));
+  auto y = sum_b.AddInstruction(HloInstruction::CreateParameter(
+      /*parameter_number=*/1, ShapeUtil::MakeShape(type, {}), "y"));
+  if (type == PRED) {
+    sum_b.AddInstruction(HloInstruction::CreateBinary(
+        ShapeUtil::MakeShape(type, {}), HloOpcode::kOr, x, y));
+  } else {
+    sum_b.AddInstruction(HloInstruction::CreateBinary(
+        ShapeUtil::MakeShape(type, {}), HloOpcode::kAdd, x, y));
+  }
+  HloComputation* reduction = module->AddEmbeddedComputation(sum_b.Build());
+  return reduction;
+}
+
 absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
   HloComputation* while_body = loop->while_body();
   // This is to set force delay for the first collective permute so it can
@@ -477,6 +495,7 @@ absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
           WindowedEinsumHandler::kWindowedEinsumRsLoopName) == 0
           ? 2
           : 0;
+  std::vector<HloInstruction*> partial_accumulations;
   for (HloInstruction* inst : while_body->MakeInstructionPostOrder()) {
     HloInstruction* matched_cp;
     if (Match(inst,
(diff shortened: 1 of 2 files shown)


xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -492,6 +511,125 @@ absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
...
+  //
+  // transformed into:
+  // shape(x,y) cp         shape(x,y) dot0    shape(x,y) dot1
+  //    \                     /                   /
+  //     \                   /                   /
+  //     shape(n,x,y) concatenate on first axis, n is the number of partitions
+  //                        |
+  //             shape(n,x,y) loop output
+  //                        |
+  //             shape(x,y) reduction on first axis
```
reedwm, line 547: The diagram doesn't match the test case. In the test case, you do:

accumulator += dot(lhs, slice(rhs))

The diagram shows `cp` as an input to `add0`, but this doesn't happen in the test case. Yet the rewrite still occurs in the test case. Can you make the diagram a bit more general? All that is necessary for the rewrite to happen is a dot output being added to an accumulator.

Also, in what cases is a collective-permute directly fed into an add? I thought an add always has a local accumulator as the LHS, but I am not aware of very many cases.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -492,6 +511,125 @@ absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
       TF_RETURN_IF_ERROR(UpdateDotAndConsumerConfig(inst, stream_id));
       ++stream_id;
     }
+    // If dot's result is accumulated, this means we found a loop with
+    // contracting dim sharded.
+    HloInstruction* partial_dot;
+    if (Match(inst, m::AddAnyOrder(m::Op(),
+                                   m::Dot(&partial_dot, m::Op(), m::Op())))) {
+      partial_accumulations.push_back(partial_dot);
+    }
```
reedwm, line 520: Move all the added logic into a dedicated function, which turns many adds inside the unrolled while-loop to a reduce outside the unrolled while-loop. A possible name could be MoveReductionOutsideLoop.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -492,6 +511,125 @@ absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
...
+          WindowedEinsumHandler::kWindowedEinsumAgLoopName) !=
+          std::string::npos) {
+    std::vector<HloInstruction*> partials_to_concat;
+
+    // We reshape it to a N+1 dimensioned tensor with left-most dim being 1.
+    Shape shape = partial_accumulations[0]->shape();
+    shape = ShapeUtil::PrependMajorDimension(1, shape);
+
+    for (auto& inst : partial_accumulations) {
+      CHECK(inst->user_count() == 1);
```
reedwm, line 562: I would not assume the user count is exactly 1. I think this is currently the case for SPMD partitioner, but we should not assume SPMD partitioner will always use a dot at most once. In general, we shouldn't assume too much about the precise output of SPMD partitioner since the output can change in the future.

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -492,6 +511,125 @@ absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
...
+    // it can be fused or overlap with other instructions in the main
+    // computation.
+    Literal zero_literal =
+        LiteralUtil::Zero(windowed_lhs->shape().element_type());
+    HloInstruction* zero = comp->AddInstruction(
+        HloInstruction::CreateConstant(std::move(zero_literal)));
+    Shape zero_bcast_shape = ShapeUtil::ChangeElementType(
+        concat_shape, windowed_lhs->shape().element_type());
+    HloInstruction* zero_bcast = MakeBroadcastHlo(zero, {}, zero_bcast_shape);
+    loop->mutable_operand(0)->AppendOperand(zero_bcast);
```
reedwm, line 588: Check there is only one user of `loop->mutable_operand(0)` and bail if there isn't, as is done [here](https://github.com/openxla/xla/blob/c04ecd9240b648c3a2d9323fc99b8392ccb611ca/xla/service/gpu/transforms/windowed_einsum_handler.cc#L68).

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -492,6 +511,125 @@ absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
...
+
+    // Update the parameter tuples of while's body and condition
+    // computations.
+    for (HloComputation* while_comp : {while_body, loop->while_condition()}) {
+      while_comp->ReplaceParameter(
+          0, HloInstruction::CreateParameter(
+                 0, loop->mutable_operand(0)->shape(),
+                 while_comp->parameter_instruction(0)->name()));
+    }
+    HloInstruction* root = while_body->root_instruction();
```
reedwm, line 600: Instead of creating a new root instruction can you just mute the original root instruction? Similar to how you mutated the operand on the line `loop->mutable_operand(0)->AppendOperand(zero_bcast);`?

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -492,6 +511,125 @@ absl::Status PostProcessUnrolledLoop(HloInstruction* loop, int64_t stream_id) {
...
+        root, new_output_tuple));
+
+    // Update the shape of the while loop instruction.
+    *loop->mutable_shape() = loop->operand(0)->shape();
+
+    // The final reduction
+    HloInstruction* concat_result_gte =
+        comp->AddInstruction(HloInstruction::CreateGetTupleElement(
+            loop, (loop->operand(0)->shape().tuple_shapes_size() - 1)));
+    HloInstruction* reduced_result =
```
reedwm, line 616: It might be simpler to add the reduction in the while-loop, instead of outside of it. That way, you don't have to add a new tuple element to the input and output of the while-loop, but instead can replace the accumulated result (with tuple index 2) with the new reduced result.

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
```diff
@@ -1064,7 +1064,7 @@ ENTRY main {
 TEST_F(WindowedEinsumHandlerTest,
        AgLoopsMultipleConsumersAreChainedWithShardedContratingDim) {
   constexpr absl::string_view kHloString = R"(
-HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(bf16[16,2048,512]{2,1,0}, bf16[4096,6288]{1,0}, bf16[16,2048,6288]{2,1,0})->bf16[4096,6288]{1,0}}, num_partitions=8
+HloModule pjit__unnamed_wrapped_function_, entry_computation_layout={(bf16[16,2048,512]{2,1,0}, bf16[4096,6288]{1,0}, bf16[16,2048,6288]{2,1,0})->(bf16[16,2048,6288]{2,1,0}, bf16[4096,6288]{1,0})}, num_partitions=8
```
reedwm, line 1067: I think you should also add a new test case dedicated to testing this change. In the new case, you can have the pre-optimized HLO have an add take as input a collective-permute, as is shown in the current version of the diagram you added in windowed_einsum_handler.cc.

Also, I don't know in what cases a CP is the input of an add. Can you describe what the pre-optimized HLO looks like for this to happen?

xla/service/gpu/transforms/windowed_einsum_handler.cc
```diff
@@ -467,6 +467,24 @@ bool ShouldAddToChain(const HloInstruction* inst) {
...
+
+HloComputation* MakeSumComputation(PrimitiveType type, HloModule* module) {
+  HloComputation::Builder sum_b("add");
+  auto x = sum_b.AddInstruction(HloInstruction::CreateParameter(
+      /*parameter_number=*/0, ShapeUtil::MakeShape(type, {}), "x"));
+  auto y = sum_b.AddInstruction(HloInstruction::CreateParameter(
+      /*parameter_number=*/1, ShapeUtil::MakeShape(type, {}), "y"));
+  if (type == PRED) {
+    sum_b.AddInstruction(HloInstruction::CreateBinary(
+        ShapeUtil::MakeShape(type, {}), HloOpcode::kOr, x, y));
```
reedwm, line 479: Is summing a group of booleans considered an "Or" operation? In any case, I don't think we'll run this codepath, since we only match Add instructions, not Or instruction, below.

reedwm (changes requested):

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: Yea since it's only 1 bit, adding bunch of 1s will make it cap at 1 so it's similar to an "or". Anyway i think your point about this being irrelevant is valid. I removed it.

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: Sounds good, moved

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: This is my mistake, i just put cp as an arbitrary op to illustrate the flow. I will make it more precise.

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: Yea agreed, will remove the check

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: makes sense, i will do the check at the beginning of the new function.

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: For some reason when i tried just mutating the instruction directly, it didnt change the call signature of the while computation so the verifier failed. I think that might work for a single instruction but not for root instruction of a comp. Replacing the instruction is the safest way it seems.

xla/service/gpu/transforms/windowed_einsum_handler.cc
Tixxx: The reason for the reduction to be moved outside is to allow it to overlap with some other collectives or fused with other kernels. Due to data dependency, the reduction is always at the end of the loop and since the scheduler cannot schedule across computations, there's a good amount of compute space to hide some other collectives.

xla/service/gpu/transforms/windowed_einsum_handler_test.cc
Tixxx: I think the CP as an input is an random error in the diagram, I just put an arbitrary op there to illustrate the flow. But in reality, we will only have partially accumulated results as the input to the add. I will change the diagram to reflect that. This test should exercise the correct code path.

reedwm (approved):

PR #17205, opened 2024-09-16
[ROCM] fixing build-brake: noexcept

pemeliya:
The member variable of **ExecutionOutput** with type **ScopedShapeBuffer** has a move-ctor and move operator= without noexcept keyword which causes the following compile error on ROCM side:

‘xla::ExecutionOutput::ExecutionOutput(xla::ExecutionOutput&&) noexcept’ is implicitly deleted because its exception-specification does not match the implicit exception-specification ‘’

@xla-rotation: would you please have a look ?


dimitar-asenov (approved):

thomasjoerg:
Hi @dimitar-asenov , looks like the PR has been changed since your approval. Please take another look. Thank you!

dimitar-asenov (approved):

dimitar-asenov:
It hasn't been changed, but it seems to be stuck in Copybara. I'm on rotation this week, so I'll take care of merging it.

PR #17257, opened 2024-09-17
[ds-fusion] Fix the command buffer scheduling pass return value

### LINKED ISSUE #17216: Command buffer scheduling pass returns true even if there are no changes
It does not check if the collected command sequences are empty [here](https://github.com/openxla/xla/blob/dc5d8b0274d8698754a76bce7e4f89b25c0f68e3/xla/service/gpu/transforms/command_buffer_scheduling.cc#L791-L794), but instead [returns true](https://github.com/openxla/xla/blob/dc5d8b0274d8698754a76bce7e4f89b25c0f68e3/xla/service/gpu/transforms/command_buffer_scheduling.cc#L813) every time it is run.

shraiysh:
Fixes #17216. Returns true when either parameters are moved, or command buffer is created.


ezhulenev (approved):

shraiysh:
The clang format failure was unrelated but I fixed it in this PR itself. It is just a change of spaces.

ezhulenev (approved):

PR #17259, opened 2024-09-17
Adding Strictness level to PGLE accuracy checker.

shraiysh:
Values for the new flag: `xla_gpu_pgle_accuracy_checker: {OFF, WARN, ERROR}`

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -288,7 +288,11 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
 
   opts.set_xla_gpu_cudnn_gemm_max_plans(5);
 
+  // TODO: remove this as it is replaced by xla_gpu_pgle_accuracy_checker.
   opts.set_xla_gpu_enable_pgle_accuracy_checker(false);
+
+  opts.set_xla_gpu_pgle_accuracy_checker(
+      DebugOptions::PGLE_STRICTNESS_LEVEL_WARN);
 
   opts.set_xla_gpu_executable_warn_stuck_timeout_seconds(10);
   opts.set_xla_gpu_executable_terminate_timeout_seconds(30);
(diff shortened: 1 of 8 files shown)


golechwierowicz (commented):
Thanks for the PR!

Have we established the plan on how this is supposed to integrate with JAX?  For one, I don't think this is how auto PGLE does the integration.

https://github.com/google/jax/pull/20475/files
https://github.com/openxla/xla/pull/11024/files

jprabhas:
How about just using a string for the flag xla_gpu_pgle_accuracy_checker that can take values {'OFF' , 'WARN', 'ERROR'}

golechwierowicz:
Yes, enum also sounds good to me. I would also assume this would resemble what is in JAX a little bit closer.

xla/debug_options_flags.cc
```diff
@@ -287,7 +287,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
 
   opts.set_xla_gpu_enable_triton_gemm_int4(false);
 
-  opts.set_xla_gpu_enable_pgle_accuracy_checker(false);
+  opts.set_xla_gpu_pgle_accuracy_checker(DebugOptions_PGLEStrictnessLevel_WARN);
```
golechwierowicz, line 290: Can you use `DebugOptions::PGLEStrictnessLevel::WARN` (here and everywhere else)?

shraiysh: That is not a valid value. The generated enum looks like:
```
enum DebugOptions_PGLEStrictnessLevel : int {
  DebugOptions_PGLEStrictnessLevel_OFF = 0,
  DebugOptions_PGLEStrictnessLevel_WARN = 1,
  DebugOptions_PGLEStrictnessLevel_ERROR = 2,
  DebugOptions_PGLEStrictnessLevel_DebugOptions_PGLEStrictnessLevel_INT_MIN_SENTINEL_DO_NOT_USE_ = std::numeric_limits<int32_t>::min(),
  DebugOptions_PGLEStrictnessLevel_DebugOptions_PGLEStrictnessLevel_INT_MAX_SENTINEL_DO_NOT_USE_ = std::numeric_limits<int32_t>::max()
};
```
Should I change it to `DebugOptions::PGLEStrictnessLevel::DebugOptions_PGLEStrictnessLevel_WARN`?

golechwierowicz: If you add a proper prefix to the enum values then you could potentially use `DebugOptions::PGLE_STRICTNESS_LEVEL_OFF`. Check https://github.com/search?q=repo%3Aopenxla%2Fxla%20xla_partitioning_algorithm&type=code for an example.

shraiysh: Done

golechwierowicz (approved):

shraiysh:
@golechwierowicz I have taken care of the build failures. Can you take a look?

golechwierowicz (approved):

shraiysh:
Fixed the merge conflict with `gpu_hlo_schedule_test.cc`.

golechwierowicz (approved):

shraiysh:
Is there an internal failure or error blocking this PR?

golechwierowicz:
I looked into it and it complains that we change the type of the existing field in the proto in the incompatible way. Can you create a new field and operate on this? I will make a follow up change which cleans up the old field.

shraiysh:
> I looked into it and it complains that we change the type of the existing field in the proto in the incompatible way. Can you create a new field and operate on this? I will make a follow up change which cleans up the old field.

Done. Thanks for pointing it out.

golechwierowicz (approved):

golechwierowicz (approved):

PR #17279, opened 2024-09-17
Use switch case without default in LiteralToValue

apivovarov:
### Recommendation:
Missing switch cases are a compiler error, so having a switch without a default case is preferable when possible

### Change:
in `xla/mlir/tools/mlir_replay/public/execution_trace_utils.cc :: LiteralToValue(Literal) -> InterpreterValue`
This PR removes IsTuple(), IsToken(), IsArray(), IsF8Type() and refactors the code to use single switch case block.


reedwm (commented):
Thank you for the PR! However, I prefer the way you did this in https://github.com/openxla/xla/pull/17170. It is nice having the compiler verify you covered all types, but it also has the issue where if new types are added, they have to be added to this switch statement. In general when adding a new type, you have to modify very many places (as you are doing in https://github.com/openxla/xla/pull/16585), so we should try reducing the number of places where each type needs to be listed separately. So I'll close this PR.

PR #17281, opened 2024-09-17
Type Conversions in Layer Norm Fusion

philipphack:
Enables the fusion of layer norm graphs with type conversions of input, scale and bias.


reedwm (changes requested):

reedwm:
This looks good to me. As we discussed offline, you want to do some more testing on models first, so I'll wait until you do that before approving, otherwise this may get prematurely merged.

reedwm (approved):

reedwm:
Approving as you confirmed the testing did not find issues.

Issue #17324, opened 2024-09-18
ElementalIrEmitterExecutionTest.ConvertFloatsToF8E4FN failed -0.0586 vs -0.0625


apivovarov:
I tried to add new test `ConvertFloatsToF8E4FN` based on `ConvertFloatsToF8E4FNUZ` code ([xla/service/elemental_ir_emitter_test.cc](https://github.com/openxla/xla/blob/main/xla/service/elemental_ir_emitter_test.cc#L399))

New test failed on CPU (x86_64)
```c
XLA_TEST_F(ElementalIrEmitterExecutionTest, ConvertFloatsToF8E4FN) {
  RunTypeConversionTest(R"(
    HloModule convertToF8E4FN
    ENTRY ConvertToF8E4FN
        (f16_ f16[], f32_ f32[], f64_ f64[], bf16_ bf16[]) -> (f8e4m3fn[], f8e4m3fn[], f8e4m3fn[], f8e4m3fn[]) {
      f16_ = f16[] parameter(0)
      f32_ = f32[] parameter(1)
      f64_ = f64[] parameter(2)
      bf16_ = bf16[] parameter(3)
      converted_f16 = f8e4m3fn[] convert(f16[] f16_)
      converted_f32 = f8e4m3fn[] convert(f32[] f32_)
      converted_f64 = f8e4m3fn[] convert(f64[] f64_)
      converted_bf16 = f8e4m3fn[] convert(bf16[] bf16_)
      ROOT tuple = (f8e4m3fn[], f8e4m3fn[], f8e4m3fn[], f8e4m3fn[]) tuple(
          converted_f16, converted_f32, converted_f64, converted_bf16)
    }
  )");
}
```
Error:
```bash
xla/service/elemental_ir_emitter_test.cc:64: Failure
Value of: RunAndCompare(std::move(module), ErrorSpec{(0.)})
  Actual: false (
Mismatches in shape (f8e4m3fn[], f8e4m3fn[], f8e4m3fn[], f8e4m3fn[]) (4 elements):
Array at shape index {1},
Mismatch count 1 (100.0000%) in shape f8e4m3fn[] (1 elements), abs bound 0, rel bound 0
Top relative error mismatches:
  actual -0.0625, expected -0.0586, index {}, rel error   0.0667, abs error  0.00391


Expected literal:
(
f8e4m3fn[] -0.102,
f8e4m3fn[] -0.0586,
f8e4m3fn[] 0.0391,
[...]

Issue #17323, opened 2024-09-18
ElementalIrEmitterExecutionTest.IotaF8E4M3 - Invalid LLVM IR


apivovarov:
I tried to add new test `IotaF8E4M3FN` based on `IotaF8E4FNUZ` code ([xla/service/elemental_ir_emitter_test.cc](https://github.com/openxla/xla/blob/main/xla/service/elemental_ir_emitter_test.cc#L540))

New test failed
```c
XLA_TEST_F(ElementalIrEmitterExecutionTest, IotaF8E4M3FN) {
  constexpr char hlo_text[] = R"(
  HloModule IotaF8E4M3FN
  ENTRY main {
    ROOT iota_ = f8e4m3fn[4] iota(), iota_dimension=0
  }
  )";

  RunTest(hlo_text, {});
}
```
Error:
```bash
[ RUN      ] ElementalIrEmitterExecutionTest.IotaF8E4FN
2024-09-18 20:44:24.550357: W ./xla/service/compiler.h:213] Ignoring the buffer assignment proto provided.
2024-09-18 20:44:24.552712: E xla/status_macros.cc:56] INTERNAL: RET_CHECK failure (xla/service/cpu/cpu_compiler.cc:954) !llvm::verifyModule(llvm_module, &err_stream) Invalid LLVM IR before optimizations:
UIToFP result must be FP or FP vector
  %5 = uitofp i64 %4 to i8

This probably indicates a bug in the HLO -> LLVM IR lowering. Rerun with --xla_dump_to to get the IR.
*** Begin stack trace ***
  tsl::CurrentStackTrace[abi:cxx11]()

  xla::status_macros::MakeErrorStream::Impl::GetStatus()

  xla::cpu::CpuCompiler::CompileLegacyCpuExecutable(std::unique_ptr<xla::HloModule, std::default_delete<xla::HloModule> >)
  xla::cpu::CpuCompiler::RunBackend(std::unique_ptr<xla::HloModule, std::default_delete<xla::HloModule> >, stream_executor::StreamExecutor*, xla::Compiler::CompileOptions const&)
  xla::HloRunner::CreateExecutableWithBufferAssignment(std::unique_ptr<xla::HloModule, std::default_delete<xla::HloModule> >, xla::BufferAssignmentProto const*, bool)
[...]

PR #17383, opened 2024-09-19
Parameterize elemental_ir_emitter_test.cc float tests

apivovarov:
`elemental_ir_emitter_test.cc` contains multiple similar tests for  `bf16`, `f8e4m3fnuz` and , `f8e5m2fnuz`.

Changes:
- Parameterize the float tests in elemental_ir_emitter_test.cc.
- Add additional types to the list of tested types - `f8e5m2`, `f8e4m3fn`, `f8e4m3b11fnuz`.

Some tests failed for newly added types. Temporary use `GTEST_SKIP` for such cases:

Related issues:
- https://github.com/openxla/xla/issues/17323
- https://github.com/openxla/xla/issues/17324

### DIFF
--- a/xla/service/elemental_ir_emitter_test.cc
+++ b/xla/service/elemental_ir_emitter_test.cc
@@ -22,6 +22,7 @@ limitations under the License.
 #include <utility>
 
 #include "absl/strings/string_view.h"
+#include "absl/strings/str_replace.h"
 #include "absl/types/span.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/LLVMContext.h"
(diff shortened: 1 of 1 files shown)


xla/service/elemental_ir_emitter_test.cc
```diff
@@ -77,6 +78,23 @@ class ElementalIrEmitterExecutionTestWithoutFastMinMax
...
+        primitive_util::NativeToPrimitiveType<T>());
+  }
+};
+
+using FloatTypes =
+    ::testing::Types<bfloat16, tsl::float8_e4m3fn, tsl::float8_e4m3fnuz,
+                     tsl::float8_e4m3b11fnuz, tsl::float8_e5m2,
+                     tsl::float8_e5m2fnuz>;
+
+TYPED_TEST_SUITE(ElementalIrEmitterExecutionTypedTest, FloatTypes);
```
reedwm, line 96: Since we only use the type name, not the actual C++ type, this can be a value-parameterized test (with TEST_P) instead of a type-parameterized test (with TYPED_TEST)

reedwm (changes requested):

xla/service/elemental_ir_emitter_test.cc
apivovarov: I initially attempted that approach, but I realized the `CompareFloat` test requires `TypeParam` (used in `LiteralUtil::CreateR1<TypeParam>`) to execute `RunTest` with a Span of Literals of the specified type. Therefore, I refactored the original implementation using `TYPED_TEST_SUITE`.

reedwm (approved):

PR #17422, opened 2024-09-20
[ffi] Support handler bundles in GPU plugin extension

gspschmid:
Currently only the execute handler is supported. This PR allows all of them to be set, bringing the GPU plugin extensions in sync with https://github.com/openxla/xla/blob/b9fcb2422f5e38ed8aecaec6b604d2fc86755c4d/xla/python/xla_compiler.cc#L306-L308 .

See also the JAX part of this change: https://github.com/jax-ml/jax/pull/23806

### DIFF
--- a/xla/pjrt/c/pjrt_c_api_gpu_extension.h
+++ b/xla/pjrt/c/pjrt_c_api_gpu_extension.h
@@ -24,17 +24,20 @@ limitations under the License.
 extern "C" {
 #endif
 
-#define PJRT_API_GPU_EXTENSION_VERSION 1
+#define PJRT_API_GPU_EXTENSION_VERSION 2
 
 struct PJRT_Gpu_Register_Custom_Call_Args {
   size_t struct_size;
   const char* function_name;
   size_t function_name_size;
   int api_version;  // 0 for an untyped call, 1 -- for typed
-  void* custom_call_function;
+  void* handler_instantiate;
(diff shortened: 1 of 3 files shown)


gspschmid:
cc @ezhulenev

gspschmid:
I'm not entirely sure what the versioning scheme for these extensions entails, but it seems that the renaming of `custom_call_function` -> `handler_execute` would require extension implementers to be aware. If need be we can also _not_ rename that particular field.

(The corresponding JAX PR that correctly deals with `PJRT_API_GPU_EXTENSION_VERSION >= 2` will have to be merged before this one.)

ezhulenev (approved):

gspschmid:
I think the GPU plugins will simply fail to build, because they're going to refer to a field that doesn't exist.

gspschmid:
@ezhulenev I put up the JAX PR I had in mind: https://github.com/jax-ml/jax/pull/23806

gspschmid:
@ezhulenev With https://github.com/jax-ml/jax/pull/23806 merged we could also try to re-land this one.

ezhulenev (approved):

ezhulenev:
I has bunch of internal failures, I'll try to patch and fix it (or maybe someone else who gets to it first)

gspschmid:
Thank you!

PR #17430, opened 2024-09-20
[ROCm] Use unique_ptr for TupleHandle in pjrt_se_client

hsharsha:
On older gcc (specifically in Ubuntu-20.04) and absl copy operator fails to copy created TupleHandle into an existing variable. Instead use unique_ptr to accompalish the same

### DIFF
--- a/xla/pjrt/pjrt_stream_executor_client.cc
+++ b/xla/pjrt/pjrt_stream_executor_client.cc
@@ -2202,7 +2202,7 @@ absl::Status CheckCompatibleShapes(bool strict_shape_checking,
 }
 
 // Makes a tuple from the arguments to an execution.
-absl::StatusOr<TupleHandle> MakeTupleHelper(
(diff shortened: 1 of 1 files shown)


xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -2202,7 +2202,7 @@ absl::Status CheckCompatibleShapes(bool strict_shape_checking,
 }
 
 // Makes a tuple from the arguments to an execution.
-absl::StatusOr<TupleHandle> MakeTupleHelper(
+absl::StatusOr<std::shared_ptr<TupleHandle>> MakeTupleHelper(
```
klucke, line 2205: Can you make this a std::unique_ptr instead?  A std::shared_ptr implies that you're going to give this off to a different thread -- which is not what you're intending here.

klucke (changes requested):

klucke (approved):

klucke (approved):

PR #17437, opened 2024-09-20
Check all F8 dtype combinations in //xla/tests:convert_test

sergey-kozub:
This PR parametrizes float8 conversion tests to make sure all combinations are covered.
All 5 supported float8 types' tests now follow the same pattern in the test file.

Previously, the following tests were missing:
- Conversion from F8e4m3b11fn to PRED;
- Denormal and negative zero tests from F16 to F8e5m2;
- `ConvertF32F8e5m2Roundtrip` was missing (test conversion from F32 to F8e5m2);
- The complete set of exhaustive tests (vs other types) was only present for FNUZ types;
- ...probably more

Also merged the "Exhaustive4" and "Exhaustive5" suffixed tests that followed a similar pattern.
Removed unnecessary saving/restoring of the debug options.

### DIFF
--- a/xla/tests/convert_test.cc
+++ b/xla/tests/convert_test.cc
@@ -54,10 +54,19 @@ class ConvertTestT : public ConvertTest {
   using ConvertTest::ConvertTest;
 };
 using FloatingPointTypeList =
-    ::testing::Types<tsl::float8_e5m2, tsl::float8_e4m3fn, tsl::float8_e5m2fnuz,
-                     tsl::float8_e4m3fnuz, Eigen::half, bfloat16, float,
+    ::testing::Types<tsl::float8_e4m3fn, tsl::float8_e4m3b11fnuz,
+                     tsl::float8_e4m3fnuz, tsl::float8_e5m2,
+                     tsl::float8_e5m2fnuz, Eigen::half, bfloat16, float,
                      double>;
 TYPED_TEST_SUITE(ConvertTestT, FloatingPointTypeList);
+
+template <typename T>
+class ConvertTestF16 : public ConvertTest {
+ public:
+  using ConvertTest::ConvertTest;
+};
+using F16TypeList = ::testing::Types<Eigen::half, bfloat16>;
+TYPED_TEST_SUITE(ConvertTestF16, F16TypeList);
 
 TEST_F(ConvertTest, ConvertR1S32ToR1S32) {
   XlaBuilder builder(TestName());
@@ -729,8 +738,21 @@ XLA_TEST_F(ConvertTest, ConvertF32BF16) {
   }
 }
 
+XLA_TYPED_TEST(ConvertTestT, ConvertFPToPred) {
+  XlaBuilder builder(this->TestName());
+  using FP = TypeParam;
+
+  auto a = ConstantR1<FP>(&builder, {FP{0.0}, FP{0.25}, FP{2.0}, FP{-0.0}});
+  ConvertElementType(a, PRED);
+
+  std::array<bool, 4> expected = {false, true, true, false};
+  this->template ComputeAndCompareR1<bool>(&builder, expected, {});
+}
+
+// ----- F8E5M2
+
 XLA_TEST_F(ConvertTest, ConvertF16F8e5m2Roundtrip) {
-  // Convert from FP16 to FP8, then back to FP16
+  // Convert from FP16 to FP8, then back to FP16.
   XlaBuilder builder(TestName());
   float nan = std::numeric_limits<float>::quiet_NaN();
   float inf = std::numeric_limits<float>::infinity();
@@ -741,6 +763,7 @@ XLA_TEST_F(ConvertTest, ConvertF16F8e5m2Roundtrip) {
   } test_cases[] = {
       // clang-format off
       {0.0, 0.0},
+      {-0.0, 0.0},
       {1.0, 1.0},
       {-1.0, -1.0},
       {nan, nan},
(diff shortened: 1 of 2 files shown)


jreiffers:
Nice, thanks!

xla/tests/convert_test.cc
```diff
@@ -910,8 +948,8 @@ XLA_TEST_F(ConvertTest, ConvertF16F8e4m3fnRoundtrip) {
       {0x1.0p-8, 0x1.0p-8},    // Denormal without rounding
       {0x1.4p-8, 0x1.0p-8},    // Round-to-even down
       {0x1.Cp-8, 0x1.0p-7},    // Round-to-even up
-      {0x1.5p-7, 0x1.4p-7},    // Round-to-nearest down
-      {0x1.3p-7, 0x1.4p-7},    // Round-to-nearest up
+      {0x1.3p-8, 0x1.0p-8},    // Round-to-nearest down
```
jreiffers, line 951: What happened here?

jreiffers (approved):

xla/tests/convert_test.cc
sergey-kozub: I checked those values by hand, and found it's easier to reason about values with the same exponent.
I believe nothing really changed here, the test is essentially the same.

akuegel:
It looks like some of the new tests fail on CPU. @sergey-kozub can you please double-check and disable those?

sergey-kozub:
> It looks like some of the new tests fail on CPU. @sergey-kozub can you please double-check and disable those?

Looking into this now.

sergey-kozub:
> It looks like some of the new tests fail on CPU.

Adding missing tests actually uncovered conversion corner cases:

1) F16 -> F8e4m3b11fnuz - this is fixed by PR#17487 (this PR should be submitted after the fix)

2) F32 -> F8e5m2/F8e4m3fn/F32F8e4m3b11fnuz
This conversion is implemented in `elemental_ir_emitter.cc` using specialized methods like `EmitF16ToF8e4m3fn`, i.e. there's no generic method (like the GPU emitter has). The main issue is that F32 is converted to F16 first, then to F8 - this results in double rounding, so the tests that verify threshold F32 values (like "Largest number that doesn't overflow") fail.

Fixing the latter issue is not trivial, so I just marked the affected tests as `DISABLED_ON_CPU` (as they expectedly fail on CPU due to a known issue) in this PR.

akuegel (approved):

sergey-kozub:
Added the fix from https://github.com/openxla/xla/pull/17487 to this PR.

xla/tests/convert_test.cc
```diff
@@ -752,8 +775,18 @@ XLA_TEST_F(ConvertTest, ConvertF16F8e5m2Roundtrip) {
       {0x1.DFCp15, 0x1.Cp15},  // Largest number that doesn't overflow
       {0x1.Ep15, inf},         // Smallest number that overflows
       {0x1p16, inf},           // Overflow
-      {0x1p-14, 0x1p-14},      // Smallest normal
-      {0x1.8p-15, 0x1.8p-15},  // Denormal
+      {0x1p-14, 0x1p-14},      // Smallest F8 normal
+      {0x1.Cp-15, 0x1p-14},    // Smallest number rounding up to normal
+
+      // Denormal tests
+      {0x1.0p-15, 0x1.0p-15},   // Denormal without rounding
```
apivovarov, line 782: better to use denormal notation (`0.`) for Denormal numbers: `0x1.0p-15`  --> `0x0.8p-14`

xla/tests/convert_test.cc
```diff
@@ -752,8 +775,18 @@ XLA_TEST_F(ConvertTest, ConvertF16F8e5m2Roundtrip) {
...
+      {0x1.4p-15, 0x1.0p-15},   // Round-to-even down
+      {0x1.Cp-15, 0x1.0p-14},   // Round-to-even up
+      {0x1.3p-15, 0x1.0p-15},   // Round-to-nearest down
+      {0x1.5p-15, 0x1.8p-15},   // Round-to-nearest up
+      {0x1p-17, 0},             // Largest number that underflows
+      {0x1.04p-17, 0x1p-16},    // Smallest number that doesn't underflow
```
apivovarov, line 788: I think you wanted to pass the following input for "Smallest number that doesn't underflow"
```
Input: 0.00000.0010 0000 01
Expected: 0.00000.0100 0000 00   # min subnormal
```
If you use `0x0.` denorm notation the you can easily convert the binary numbers above to hex
```
Input: 0.00000.0010 0000 01
mantisa 0010 0000 0100 = 0x0.204
exponent 00000 =  p-14 (denorm e5m2)
0x0.204p-14

Expected: 0.00000.0100 0000 00   # min subnormal
mantisa 0100 0000 0000 = 0x0.400
exponent 00000 =  p-14 (denorm e5m2)
0x0.4p-14
```

The correct Smallest number that doesn't underflow is `0x0.204p-14` which is smaller than `0x1.04p-17`
```
## Fixed test
{0x0.204p-14, 0x0.4p-14}, // Smallest number that doesn't underflow
```

xla/tests/convert_test.cc
```diff
@@ -752,8 +775,18 @@ XLA_TEST_F(ConvertTest, ConvertF16F8e5m2Roundtrip) {
...
+      {0x1.BFp-15, 0x1.8p-15},  // Largest number that rounds to denormal
```
apivovarov, line 789: Same logic as above
```
Input:     0.00000.1101 1111 1100 = 0x0.DFCp-14
Expected:  0.00000.1100 0000 0000 = 0x0.C00p-14   ## max subnormal
```
Fixed test
```
{0x0.DFCp-14, 0x0.Cp-14}, // Largest number that rounds to denormal
```
`0x0.DFCp-14` is larger than `0x1.BFp-15`

sergey-kozub:
Thanks, created https://github.com/openxla/xla/pull/17527 to address your comments.

Issue #17276, opened 2024-09-17
Tranposing to different layout permutations results in different numerics 


elfiegg:
Hello, we stumbled upon a numerical issue for below modules while training fp8 quantizated models. 
```
ENTRY main {
    %p0 = f8e4m3fn[12288,4096]{0,1} parameter(0)
    %b = f8e4m3fn[4096,12288]{1,0} bitcast(%p0)
    %transpose = f8e4m3fn[12288,4096]{1,0} transpose(%b), dimensions={1,0}
    %p1 = f8e4m3fn[4096,16384]{0,1} parameter(1)
    %p2 = f32[] parameter(2)
    %constant_1 = f32[] constant(1)
    ROOT %cublas-gemm.1.0 = (bf16[12288,16384]{1,0}, s8[33554432]{0}) custom-call(f8e4m3fn[12288,4096]{1,0} %transpose,  f8e4m3fn[4096,16384]{0,1} %p1,f32[] %p2, f32[]%constant_1, f32[]%constant_1, f32[]%constant_1), custom_call_target="__cublas$lt$matmul$f8", backend_config={"operation_queue_id":"0","wait_on_operation_queues":[],"gemm_backend_config":{"alpha_real":1,"alpha_imag":0,"beta":0,"dot_dimension_numbers":{"lhs_contracting_dimensions":["1"],"rhs_contracting_dimensions":["0"],"lhs_batch_dimensions":[],"rhs_batch_dimensions":[]},"precision_config":{"operand_precision":["DEFAULT","DEFAULT"],"algorithm":"ALG_UNSET"},"epilogue":"DEFAULT","damax_output":false,"selected_algorithm":"6","lhs_stride":"50331648","rhs_stride":"67108864","grad_x":false,"grad_y":false},"force_earliest_schedule":false}
  }
```

```
 ENTRY main {
    %p0 = f8e4m3fn[12288,4096]{0,1} parameter(0)
    %transpose = f8e4m3fn[4096,12288]{0,1} transpose(%p0), dimensions={1,0}
    %p1 = f8e4m3fn[4096,16384]{0,1} parameter(1)
    %p2 = f32[] parameter(2)
    %constant_1 = f32[] constant(1)
[...]

PR #17440, opened 2024-09-20
Inject desired pattern for handling Transpose for fp8 gemm rewrite

wenscarl:
Related to https://github.com/openxla/xla/issues/17276 and https://github.com/openxla/xla/pull/16975. 
This PR updates the GemmRewriter to handle the transpose of non-descending layouts directly, eliminating the need for the layout_normalization pass to correct this error-prone pattern post-rewrite. The desired transformation is now injected into GemmRewriter, ensuring the problematic layout is handled internally. This PR transforms the following error-prone pattern, where the transpose of a non-descending layout is the issue:
```
a = f8e4m3fn[x,y]{0,1} xxx
transpose.0 = f8e4m3fn[y,x]{0,1} transpose(a), dimensions=(1,0)
custom-call(a,...)
```
to 
```
a = f8e4m3fn[x,y]{0,1} xxx
bt = f8e4m3fn[y,x]{1,0} bitcast(a)
transpose.1 = f8e4m3fn[x,y]{1,0} transpose(bt), dimensions=(1,0)
bt.1= f8e4m3fn[y,x]{0,1} bitcast(transpose.1)
custom-call(bt.1,...)
```


elfiegg:
Also cc @akuegel

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -361,28 +362,67 @@ std::optional<MatchedFp8Param> MatchFp8Param(HloInstruction *instr) {
 // dimension. There must be only one contracting and only one non-contracting
 // dimension. Keeps the layout the same.
 HloInstruction *TransposeMatrix(HloInstruction *instr, int64_t contracting_dim,
-                                absl::Span<const int64_t> batch_dims) {
+                                absl::Span<const int64_t> batch_dims,
+                                bool col_maj = false) {
   // Identify the dimensional order which describes a transpose of the
   // contracting and non-contracting dimensions of the GEMM.
   std::vector<int64_t> permutation(instr->shape().dimensions_size(), -1);
-  // Discard the batch dimensions.
```
akuegel, line 368: I believe if you move this down below the "Identify the non-contracting dimension" block, it will break the logic.
See the check

```
 if (permutation[i] == -1 && contracting_dim != i) {
      non_contracting_dim = i;
    }
```

We rely on the permutation vector to contain values >= 0 for the batch dimensions, otherwise we might detect the first batch dimension as non_contracting_dim.

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -361,28 +362,67 @@ std::optional<MatchedFp8Param> MatchFp8Param(HloInstruction *instr) {
...
+    permutation[contracting_dim] = non_contracting_dim;
+
+    Shape new_shape = ShapeUtil::PermuteDimensions(permutation, instr->shape());
+    *new_shape.mutable_layout() = instr->shape().layout();
+
+    return instr->AddInstruction(
+        HloInstruction::CreateTranspose(new_shape, instr, permutation));
+  }
+
+  Shape normalized_input_shape =
```
akuegel, line 393: By going to the normalized input shape, we essentially apply the permutation InversePermutation(Reverse(instr->shape().layout().minor_to_major())) to the shape.

I see that the loop below computes new_contracting_dim and new_non_contracting_dim correctly, but it does not apply any permutation to the batch dimensions. Is this ok, as we don't care about the order of the batch dimensions? Or is it a bug?

xla/service/gpu/transforms/gemm_rewriter.cc
wenscarl: You're right, that is my overlook. Fixed.

xla/service/gpu/transforms/gemm_rewriter.cc
wenscarl: Yes, it is intended that the batch dims shouldn't be permuted. E.g. a is the input tensor with non-normalized shape. where batch_dims = [1,2], non_contracting_dim=[0] and contracting_dim=[3] , d is output.
```
a = [2,3,4,5]{0,2,3,1}
a0 = [4,3,5,2]{3,2,1,0}bitcast(a)
c = [4,3,2,5]{3,2,1,0}transpose(a0),dimensions=(0,1,3,2)
d =[5,3,4,2]{0,2,3,1}bitcast(c)
```
since the transpose is applied on a0 which has descending layout already, the it's ok to leave batch dims alone because they are also in descending order.

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -376,13 +378,52 @@ HloInstruction *TransposeMatrix(HloInstruction *instr, int64_t contracting_dim,
...
+    } else {
+      // Discard the batch dimensions.
+      permutation[i] = i;
+    }
+  }
+
+  permutation[new_non_contracting_dim] = new_contracting_dim;
+  permutation[new_contracting_dim] = new_non_contracting_dim;
+
+  Shape transpose_shape =
```
akuegel, line 414: I think computing permutation can be simplified quite a bit. You can compute `inv_perm` first (like below). Then

```
new_contracting_dim = inv_perm[contracting_dim];
new_non_contracting_dim = inv_perm[non_contracting_dim];
permutation = absl::c_iota(permutation, 0);
std::swap(permutation[new_contracting_dim], permutation[new_non_contracting_dim]);
```

akuegel (commented):

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -376,13 +378,52 @@ HloInstruction *TransposeMatrix(HloInstruction *instr, int64_t contracting_dim,
...
+  if (!col_maj) {
+    permutation[non_contracting_dim] = contracting_dim;
+    permutation[contracting_dim] = non_contracting_dim;
 
-  Shape new_shape = ShapeUtil::PermuteDimensions(permutation, instr->shape());
-  *new_shape.mutable_layout() = instr->shape().layout();
-  return instr->AddInstruction(
-      HloInstruction::CreateTranspose(new_shape, instr, permutation));
+    Shape new_shape = ShapeUtil::PermuteDimensions(permutation, instr->shape());
+    *new_shape.mutable_layout() = instr->shape().layout();
```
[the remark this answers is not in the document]
wenscarl: You're right. The purpose of this change is to insert `bitcast-transpose-bitcast` to the lhs if it's column major.  if the original operand is row-major already, `transpose` shouldn't be a issue.

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -361,7 +362,8 @@ std::optional<MatchedFp8Param> MatchFp8Param(HloInstruction *instr) {
 // dimension. There must be only one contracting and only one non-contracting
 // dimension. Keeps the layout the same.
 HloInstruction *TransposeMatrix(HloInstruction *instr, int64_t contracting_dim,
-                                absl::Span<const int64_t> batch_dims) {
+                                absl::Span<const int64_t> batch_dims,
+                                bool col_maj = false) {
```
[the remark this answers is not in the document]
wenscarl: It was used to handle col-maj lhs.

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -1223,8 +1264,12 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
       } else {
         dim_nums->set_lhs_contracting_dimensions(0, num_batch_dims);
       }
+
       a.fp8_input =
           TransposeMatrix(a.fp8_input, a_contracting_dims[0], a_batch_dims);
+
+      a.fp8_input = RegulateColMajorTransposeMatrixF8(
```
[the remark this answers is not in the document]
wenscarl: That's a typo, removed.

xla/service/gpu/transforms/gemm_rewriter_test.cc
```diff
@@ -5032,6 +5032,60 @@ TEST_P(ParameterizedFp8GemmRewriteTest, UnscaledABUnscaledDMatrixBiasF8) {
       )");
 }
 
+TEST_P(ParameterizedFp8GemmRewriteTest, ScaledABUnscaledDColMajorLhsF8) {
+  const char* hlo_text = R"(
```
akuegel, line 5036: Can you change this test to involve batch dimensions as well? Just so that we can see that those can be handled as well.

wenscarl: Updated.

xla/service/gpu/transforms/gemm_rewriter.cc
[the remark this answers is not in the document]
akuegel: It seems this comment is not addressed yet.

akuegel:
I don't know how to mark a conversation again as unresolved, but for visibility I am copying my answer to a resolved conversation here as well:

I am not sure whether I understand your argument correctly, but to me it seems you assume that RowMajor means that the XLA layout of the operand will be the default layout? Why can we assume that? And even if it turns out that we don't need bitcast-transpose-bitcast, because the operand already has default layout, AlgebraicSimplifier would simplify the bitcasts away. So my suggestion is to always generate the bitcast - transpose - bitcast sequence, and let AlgebraicSimplifier clean it up later. Alternatively, don't distinguish between whether the call was for the RowMajor->ColumnMajor swap or for the ColumnMajor -> RowMajor swap, and instead check whether the operand has default layout or not (and in case of default layout, use the if (!col_maj) code path. That would be cleaner, as we don't need the col_maj parameter, and we are checking exactly the precondition for whether we need bitcast-transpose-bitcast or a single transpose will be enough.

xla/service/gpu/transforms/gemm_rewriter.cc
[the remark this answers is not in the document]
wenscarl: Thank you for the reminder. I've opted for the alternative approach suggested to ensure that other FP8-related rewriter tests remain unaffected

akuegel (approved):

xla/service/gpu/transforms/gemm_rewriter_test.cc
```diff
@@ -5032,6 +5032,58 @@ TEST_P(ParameterizedFp8GemmRewriteTest, UnscaledABUnscaledDMatrixBiasF8) {
       )");
 }
 
+TEST_P(ParameterizedFp8GemmRewriteTest, ScaledABUnscaledDColMajorLhsF8) {
```
akuegel, line 5035: The test is failing. Can you rebase and check whether you can reproducer the failure?

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -48,6 +48,7 @@ limitations under the License.
 #include "xla/layout.h"
 #include "xla/literal.h"
 #include "xla/literal_util.h"
+#include "xla/permutation_util.h"
```
akuegel, line 51: You need to update the BUILD file and add a dependency to the target that has this header in its hdrs section.

akuegel (commented):
Internal testing revealed some issues. Please see my comments.

xla/service/gpu/transforms/gemm_rewriter.cc
akuegel: Also missing:

#include "xla/layout_util.h"

xla/service/gpu/transforms/gemm_rewriter_test.cc
wenscarl: Yes. The failure is reproducible after rebasing. The cause is that the original shape `f8e4m3fn[2,16,32]{1,0,2}` is transformed into `f8e4m3fn[32,2,16]{2,1,0}` before entering `TransposeMatrrix` in gemm_rewriter by pass `hlo_normalization.after_transpose-folding`. This makes the test case not valid. Updated.

akuegel (approved):

xla/service/gpu/transforms/gemm_rewriter.cc
```diff
@@ -362,27 +364,61 @@ std::optional<MatchedFp8Param> MatchFp8Param(HloInstruction *instr) {
...
+    *new_shape.mutable_layout() = input_shape.layout();
+
+    return instr->AddInstruction(
+        HloInstruction::CreateTranspose(new_shape, instr, permutation));
+  }
+
+  Shape normalized_input_shape =
+      ShapeUtil::MakeShapeWithDescendingLayoutAndSamePhysicalLayout(
+          input_shape);
+  auto a0 = MakeBitcastHlo(instr, normalized_input_shape);
```
philipphack, line 398: Would it be conceptually simpler to insert a copy before the transpose to change the layout of the input? (This assumes that the copy -> transpose sequence is optimized by another pass which I haven't verified.)

Also, can we pick a more descriptive variable name here?

akuegel: This pass runs after layout normalization which turns copies into bitcast + transpose, so it should not produce any Copy ops that change the layout (otherwise we would have to run layout normalization again).

PR #17444, opened 2024-09-20
[XLA:CPU] Increase minimum alignment of buffers for CPU

kanvi-nervana:
Change minimum alignment of buffers passed to XLA:CPU from 16 to 64. This improves performance for many workloads including ViT, Bert, gemma by upto 30%

### DIFF
--- a/xla/cpu_function_runtime.h
+++ b/xla/cpu_function_runtime.h
@@ -179,7 +179,8 @@ class BufferInfo {
 inline constexpr size_t Align() { return 64; }
 
 // The minimum alignment of buffers passed to XLA:CPU.
-inline constexpr size_t MinAlign() { return 16; }
+// Aligning to 64 byte cacheline.
+inline constexpr size_t MinAlign() { return 64; }
 
 // When declaring variables that will be passed to an XLA instance as input via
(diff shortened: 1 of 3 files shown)


penpornk (commented):
Thank you for your PR! 
@ezhulenev and @Adam-Banas have been looking into bumping this too, but there were internal code (and failures) that made it challenging to bump to 64B. For the first step, we are trying at least bump it to match Eigen's requirement in https://github.com/openxla/xla/pull/16505 then work from there. The PR should be able to land soon. Maybe close this PR for now?

mdfaijul:
@penpornk Can we make the alignment based on CPU features?

penpornk:
@mdfaijul I think we will probably still need to keep the minimum alignment in sync with Eigen's since we feed data in and out of it a lot, i.e., if we use 64B for some architecture, Eigen needs to use 64B for that architecture too.

kanvi-nervana:
Closing this PR for now, will reopen if needed.

snadampal:
Hi @kanvi-nervana , do you have the scripts to show that 30% perf improvement ? I'm seeing perf drop for bart-large model, so, would like to check with your scripts.

PR #17453, opened 2024-09-20
Reorder Collective Optimization Passes

philipphack:
Moves the collective quantizer pass ahead of the collective pipeliner to preserve FP8 quantization and dequantization patterns preceded or followed by collectives without running the collective pipeliner post layout assignment. See #12866 and #15292.

### DIFF
--- a/xla/tests/collective_ops_e2e_test.cc
+++ b/xla/tests/collective_ops_e2e_test.cc
@@ -999,10 +999,9 @@ ENTRY main.9_spmd {
   CollectiveOpsCompareWindowedNonWindowed(kModuleReplicatedStr);
 }
 
-TEST_F(CollectiveOpsTestE2E, PostLayoutCollectivePipeliner) {
-  // We need fp8 support to test the post-layout collective pipeliner. This will
-  // preserve the desired fp8 patterns and so the gemm rewriter can correctly
-  // recognize them and rewrite to custom fp8 gemm calls.
+TEST_F(CollectiveOpsTestE2E, CollectivePipelinerF8) {
+  // Verify that FP8 patterns are preserved when collectives are pipelined so
+  // the GEMM rewriter can create FP8 matmuls.
   if (!HasFp8Support()) {
     GTEST_SKIP() << "Test requires a post-Ada GPU.";
   }
@@ -1075,7 +1074,6 @@ ENTRY entry {
   HloModuleConfig config =
       GetModuleConfigForTest(/*replica_count=*/kNumReplicas);
   auto opts = GetDebugOptionsForTest();
-  opts.set_xla_gpu_run_post_layout_collective_pipeliner(true);
   opts.set_xla_gpu_enable_pipelined_collectives(true);
   opts.set_xla_gpu_enable_triton_gemm(false);
   CollectiveOpsVerifyF8Matmul(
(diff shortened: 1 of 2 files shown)


reedwm (changes requested):
IIUC from reading https://github.com/openxla/xla/pull/12866, the motivation is as follows:

You have the pattern

```
while{Q->DQ->AG->dot}
```

Then, collective_pipeliner apparently moves things before AG outside the while-loop (why?), turning it into

```
Q->DQ->AG->while{dot}
```

This is bad because the DQ needs to be close to the dot so gemm_rewriter can match it. gemm_rewriter can handle an AG between the DQ and the dot, but cannot handle the DQ being outside a while-loop and the dot being inside it (and similarly Triton will not do well here).

https://github.com/openxla/xla/pull/12866 fixed this by moving collective_pipeliner after gemm_rewriter. But this was effectively disabled by https://github.com/openxla/xla/pull/15292 (why?) and so we need to fix this another way.

This PR fixes this by running collective_quantizer before collective_pipeliner. collective_quantizer turns the pattern into:

```
while{Q->AG->DQ->dot}
```

Then, collective_pipeliner will turn this into:

```
Q->AG->while{DQ->dot}
```

This pattern can be handled by gemm_rewriter just fine.

Can you confirm whether I'm correct, and can you update the PR description to describe the motivation in more detail?

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1075,7 +1074,6 @@ ENTRY entry {
   HloModuleConfig config =
       GetModuleConfigForTest(/*replica_count=*/kNumReplicas);
   auto opts = GetDebugOptionsForTest();
-  opts.set_xla_gpu_run_post_layout_collective_pipeliner(true);
```
nouiz, line 1078: Should you update the test to run in both case?

reedwm: Should we just remove the `xla_gpu_run_post_layout_collective_pipeliner` option? It has no purpose after this PR is submitted, right?

philipphack: The flag is obsolete and will be removed after this been merged.

philipphack:
> Can you confirm whether I'm correct

That's a generally accurate description.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -999,73 +999,57 @@ ENTRY main.9_spmd {
...
-  // We need fp8 support to test the post-layout collective pipeliner. This will
-  // preserve the desired fp8 patterns and so the gemm rewriter can correctly
-  // recognize them and rewrite to custom fp8 gemm calls.
+TEST_F(CollectiveOpsTestE2E, CollectivePipelinerF8) {
+  // Verify that FP8 patterns are preserved when collectives are pipelined so
+  // the GEMM rewriter can create FP8 matmuls.
   if (!HasFp8Support()) {
-    GTEST_SKIP() << "Test requires a post-Ada GPU.";
+    GTEST_SKIP() << "Test requires Hopper or newer architecture.";
   }
```
reedwm, line 1007: The test requires Ada or newer, not Hopper or newer.

reedwm (changes requested):

xla/tests/collective_ops_e2e_test.cc
philipphack: Hopper predates Ada.

reedwm (approved):

PR #17476, opened 2024-09-22
Fix chlo_legalize_to_mhlo.mlir.test by using CHECK-DAG

apivovarov:
To handle situations where `mhlo.multiply` and `mhlo.subtract` can run in any order, we need to make the test more flexible by using `CHECK-DAG` instead of `CHECK`.

### Testing
```bash
bazel test //xla/mlir_hlo/tests:Dialect/chlo/chlo_legalize_to_mhlo.mlir.test
```


akuegel (approved):
Thank you!

PR #17487, opened 2024-09-23
Fix F16->F8e4m3b11fnuz conversion on CPU

sergey-kozub:
Adding missing test combinations in [PR#17437](https://github.com/openxla/xla/pull/17437) uncovered an issue converting F16 small negative values to F8e4m3b11fnuz type (that doesn't have negative zeros) - it results in `-nan` instead of `0`. This PR fixes this conversion corner case.


akuegel (commented):
Thank you very much for the fix :)
According to new policy, it is easier to get this merged if you include it in the other PR that has the tests that expose the issue. Can you please add it there?

sergey-kozub:
Merged this PR into https://github.com/openxla/xla/pull/17437 which has the tests.

PR #17493, opened 2024-09-23
[XLA:GPU] Sort groups in NCCL clique keys

jaro-sevcik:
It turns out that in some cases, we get two calls to GetNcclCliqueKey that only differ by the order of the groups (for instance, one call with groups={{0,1},{2, 3}}, another call with groups={{2,3},{0,1}}. This leads to unnecessary creation of equivalent NCCL communicators.

This patch sorts the groups in NCCL clique key so that we create the same key for the same set of groups (up to permutation).

### DIFF
--- a/xla/service/gpu/runtime/nccl_clique_key.cc
+++ b/xla/service/gpu/runtime/nccl_clique_key.cc
@@ -37,6 +37,18 @@ namespace xla::gpu {
 // NcclCliqueKey
 //===----------------------------------------------------------------------===//
 
+namespace {
+
+// Compare by the first element.
+bool CompareGroups(const std::vector<GlobalDeviceId>& lhs,
+                   const std::vector<GlobalDeviceId>& rhs) {
+  if (rhs.empty()) return false;
+  if (lhs.empty()) return true;
+  return lhs[0] < rhs[0];
+}
+
+}  // namespace
+
 NcclCliqueKey::NcclCliqueKey(
     std::vector<GlobalDeviceId> devices, NcclStreamId stream_id,
     AsyncStreamKind stream_kind,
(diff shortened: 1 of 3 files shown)


jaro-sevcik:
@ezhulenev Could you take a look, please?

I am told the same thing happens for devices, I will submit a separate PR for that once we agree on how to handle the equality here.

xla/service/gpu/runtime/nccl_clique_key.cc
```diff
@@ -37,14 +37,32 @@ namespace xla::gpu {
...
     std::vector<GlobalDeviceId> devices, NcclStreamId stream_id,
     AsyncStreamKind stream_kind,
     std::vector<std::vector<GlobalDeviceId>> participant_groups)
     : devices_(std::move(devices)),
       stream_id_(stream_id),
       stream_kind_(stream_kind),
-      participant_groups_(std::move(participant_groups)) {}
+      participant_groups_(std::move(participant_groups)) {
+  for (std::vector<GlobalDeviceId>& group : participant_groups_) {
+    std::sort(group.begin(), group.end());
```
ezhulenev, line 61: prefer `absl::c_sort`

xla/service/gpu/runtime/nccl_clique_key.cc
```diff
@@ -37,14 +37,32 @@ namespace xla::gpu {
 // NcclCliqueKey
 //===----------------------------------------------------------------------===//
 
+namespace {
+
+// Compare by the first element.
+bool CompareGroups(const std::vector<GlobalDeviceId>& lhs,
+                   const std::vector<GlobalDeviceId>& rhs) {
+  if (rhs.empty()) return false;
```
ezhulenev, line 45: I think lambda should be more readable, otherwise it's a bit hard to get why sorting by `[0]` works. Also can the group be empty? I think this should never happen.

ezhulenev (approved):

PR #17527, opened 2024-09-24
Improve readability for denormal numbers in FP8 conversion tests

sergey-kozub:
As suggested in post-submit comments in https://github.com/openxla/xla/pull/17437

Limit the exponent of the test literals to the minimum representable value.

### DIFF
--- a/xla/tests/convert_test.cc
+++ b/xla/tests/convert_test.cc
@@ -779,14 +779,14 @@ XLA_TEST_F(ConvertTest, ConvertF16F8e5m2Roundtrip) {
       {0x1.Cp-15, 0x1p-14},    // Smallest number rounding up to normal
 
       // Denormal tests
-      {0x1.0p-15, 0x1.0p-15},   // Denormal without rounding
-      {0x1.4p-15, 0x1.0p-15},   // Round-to-even down
-      {0x1.Cp-15, 0x1.0p-14},   // Round-to-even up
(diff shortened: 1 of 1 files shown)


jreiffers (approved):

xla/tests/convert_test.cc
```diff
@@ -831,14 +831,14 @@ XLA_TEST_F(ConvertTest, DISABLED_ON_CPU(ConvertF32F8e5m2Roundtrip)) {
       {0x1.Cp-15, 0x1p-14},       // Smallest number rounding up to normal
 
       // Denormal tests
-      {0x1.0p-15, 0x1.0p-15},       // Denormal without rounding
-      {0x1.4p-15, 0x1.0p-15},       // Round-to-even down
+      {0x1.0p-15, 0x0.8p-14},       // Denormal without rounding
+      {0x1.4p-15, 0x0.8p-14},       // Round-to-even down
```
apivovarov, line 835: Shouldn't it be similar to the F16 test above?
``` 
     {0x0.8p-14, 0x0.8p-14},    // Denormal without rounding
      {0x0.Ap-14, 0x0.8p-14},    // Round-to-even down
```
etc

sergey-kozub: In this PR, I limited the exponents of the constants to the minimum value a type supports, as you suggested in the other PR.
Essentially, `0x1.0p-15` is equal to `0x0.8p-14`. But for F16, I'm using "p-14", as it's the minimum exponent for the type.

PR #17625, opened 2024-09-26
[GPU] Optimize zero-clamping of index operands known to be non-negative.

sergachev:
This is a minor optimization and fixes
LaxBackedScipyTests.testSphHarmAccuracy from JAX lax_scipy_test on H100 with CUDA 12.5.0 which has incorrect handling of max(min(abs(x), y), z) in ptxas.

### DIFF
--- a/xla/service/gpu/fusions/transforms/simplify_arith.cc
+++ b/xla/service/gpu/fusions/transforms/simplify_arith.cc
@@ -24,6 +24,7 @@ limitations under the License.
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/IR/Builders.h"
+#include "mlir/IR/Matchers.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/IR/Value.h"
 #include "mlir/Pass/Pass.h"
@@ -228,6 +229,24 @@ struct RewriteTruncExtShuffle : public OpRewritePattern<mlir::gpu::ShuffleOp> {
   }
 };
 
+static std::optional<Interval> GetSelectRange(mlir::Operation* sel) {
+  // Match |x| implemented as (x >= 0) ? x : (0 - x).
+  auto x = mlir::matchers::m_Val(sel->getOperand(1));
+  if (!sel->getOperand(1).getType().isInteger(32) ||
+      !mlir::matchPattern(
+          sel, mlir::m_Op<mlir::arith::SelectOp>(
+                   mlir::m_Op<mlir::arith::CmpIOp>(x, mlir::m_Zero()), x,
+                   mlir::m_Op<mlir::arith::SubIOp>(mlir::m_Zero(), x)))) {
+    return std::nullopt;
+  }
+  if (sel->getOperand(0).getDefiningOp<mlir::arith::CmpIOp>().getPredicate() !=
+      CmpIPredicate::sge) {
+    return std::nullopt;
+  }
+  // Annotate |x| as >= 0.
+  return std::make_optional<Interval>({0, std::numeric_limits<int32_t>::max()});
+}
+
 void AnnotateRanges(mlir::func::FuncOp func) {
   func->walk([](mlir::Operation* op) {
     if (op->getNumResults() != 1) {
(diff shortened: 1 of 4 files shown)


xla/service/gpu/fusions/transforms/simplify_arith.cc
```diff
@@ -228,6 +229,24 @@ struct RewriteTruncExtShuffle : public OpRewritePattern<mlir::gpu::ShuffleOp> {
   }
 };
 
+static std::optional<Interval> GetSelectRange(mlir::Operation* sel) {
+  // Match |x| implemented as (x >= 0) ? x : (0 - x).
+  auto x = mlir::matchers::m_Val(sel->getOperand(1));
+  if (!sel->getOperand(1).getType().isInteger(32) ||
```
akuegel, line 235: I guess we could also handle 64 bit int as well? We would just need to use std::numeric_limits<int64_t>::max() as upper bound, otherwise it should be the same. I think it is somewhat likely that s64 is used for indexing as well.

jreiffers: I'd just go with isSignlessInteger and use `static_cast<int64_t>(1ull << (ty.getIntOrFloatBitWidth() - 1)) - 1` (the ull to prevent UB).

xla/service/gpu/fusions/transforms/simplify_arith.cc
```diff
@@ -228,6 +229,24 @@ struct RewriteTruncExtShuffle : public OpRewritePattern<mlir::gpu::ShuffleOp> {
...
+          sel, mlir::m_Op<mlir::arith::SelectOp>(
+                   mlir::m_Op<mlir::arith::CmpIOp>(x, mlir::m_Zero()), x,
+                   mlir::m_Op<mlir::arith::SubIOp>(mlir::m_Zero(), x)))) {
+    return std::nullopt;
+  }
+  if (sel->getOperand(0).getDefiningOp<mlir::arith::CmpIOp>().getPredicate() !=
+      CmpIPredicate::sge) {
+    return std::nullopt;
+  }
+  // Annotate |x| as >= 0.
```
jreiffers, line 246: We should probably refine the range we return here with the range of `x`. Also, in general, let's get the range of both operands and return the union.

sergachev: Included the range of x. Didn't understand about "both operands": x is the only input to the expression matched here.

akuegel (approved):

xla/service/gpu/fusions/transforms/simplify_arith.cc
jreiffers: Sorry, I meant when the pattern doesn't match:

select(pred, x, y) -> union(getrange(x), getrange(y))

in the pattern, we should get the abs of range of x and then intersect with what you compute. I think something like this should be right:

```
Interval result ...
if (x_range.has_value()) {
  auto positive_range = x_range.max(*x_range, {0, 0});
  auto abs_negative_range = -x_range.min(*y_range, {0, 0});
  auto abs_x_range = positive_range.Union(abs_negative_range);
  return result.Intersect(abs_x_range);
} 
return result;
```

sergachev: I see. I think my change is still valid (it annotates only the exact |x| pattern) and you can add your more general logic for select on top.

jreiffers: Taking the union of x's range and the nonnegative integers is not correct, no. But feel free to just remove that part again.

sergachev: Indeed. range(|x|) = {0, x.upper}, right? Then should it just be
```
  if (x_range.has_value()) {
    return result.Intersect(*x_range);
  }
```
?

sergachev: Indeed!

sergachev: I added a couple more tests with x's range.

jreiffers: Thanks!

jreiffers (approved):

PR #17704, opened 2024-09-27
[jax.distributed] Allow enabling grpc channel compression

gspschmid:
Allows passing an additional boolean argument `use_compression` via `xla_extension.get_distributed_runtime_client(...)` that controls whether compression is enabled on the gRPC channels created for each distributed runtime client.

Motivation: XLA sends O(mesh) [device topologies](https://github.com/openxla/xla/blob/9fb4f21c3542c10b6a5bd98144801bbeec10b489/xla/pjrt/distributed/protocol.proto#L84) through its centralized coordination service and we have reason to believe that this becomes a bottleneck at large scale. Compression of the underlying gRPC communication is currently implicitly disabled, and might give us a cheap avenue to scale a bit further with the centralized KV store design.

One small note: I refrained from adding `use_compression` to `DistributedRuntimeClient::Options` because the new flag is only relevant during channel creation in `distributed.cc`, but not within `DistributedRuntimeClient`. If we added `use_compression` to Options then the `GetDistributedRuntimeClient(channel, options)` defined in `client.cc` would seem to allow controlling compression, but it's really ignored. Let me know if you'd rather go that way.

Corresponding JAX PR: https://github.com/jax-ml/jax/pull/23969

### DIFF
--- a/xla/pjrt/distributed/distributed.cc
+++ b/xla/pjrt/distributed/distributed.cc
@@ -38,10 +38,22 @@ GetDistributedRuntimeService(std::string address,
 }
 
 std::shared_ptr<DistributedRuntimeClient> GetDistributedRuntimeClient(
-    std::string address, const DistributedRuntimeClient::Options& options) {
-  std::shared_ptr<grpc::Channel> channel = grpc::CreateChannel(
-      address, tsl::GetClientCredentials(kVerifySecureCredentials));
+    std::string address, const DistributedRuntimeClient::Options& options,
+    bool use_compression) {
+  auto channel = GetDistributedRuntimeClientChannel(
+      address, tsl::GetClientCredentials(kVerifySecureCredentials),
+      use_compression);
   return GetDistributedRuntimeClient(channel, options);
+}
+
+std::shared_ptr<::grpc::Channel> GetDistributedRuntimeClientChannel(
+    std::string address, std::shared_ptr<::grpc::ChannelCredentials> creds,
+    bool use_compression) {
+  grpc::ChannelArguments args;
+  if (use_compression) {
+    args.SetCompressionAlgorithm(GRPC_COMPRESS_GZIP);
+  }
+  return ::grpc::CreateCustomChannel(address, creds, args);
 }
 
 }  // namespace xla
--- a/xla/python/xla.cc
+++ b/xla/python/xla.cc
@@ -36,16 +36,16 @@ limitations under the License.
 #include "llvm/Support/Casting.h"
 #include "nanobind/nanobind.h"
 #include "nanobind/nb_defs.h"
-#include "nanobind/stl/function.h"  // IWYU pragma: keep
-#include "nanobind/stl/optional.h"  // IWYU pragma: keep
-#include "nanobind/stl/pair.h"  // IWYU pragma: keep
-#include "nanobind/stl/set.h"  // IWYU pragma: keep
-#include "nanobind/stl/shared_ptr.h"  // IWYU pragma: keep
-#include "nanobind/stl/string.h"  // IWYU pragma: keep
+#include "nanobind/stl/function.h"     // IWYU pragma: keep
+#include "nanobind/stl/optional.h"     // IWYU pragma: keep
+#include "nanobind/stl/pair.h"         // IWYU pragma: keep
+#include "nanobind/stl/set.h"          // IWYU pragma: keep
+#include "nanobind/stl/shared_ptr.h"   // IWYU pragma: keep
+#include "nanobind/stl/string.h"       // IWYU pragma: keep
 #include "nanobind/stl/string_view.h"  // IWYU pragma: keep
-#include "nanobind/stl/unique_ptr.h"  // IWYU pragma: keep
-#include "nanobind/stl/variant.h"  // IWYU pragma: keep
-#include "nanobind/stl/vector.h"  // IWYU pragma: keep
+#include "nanobind/stl/unique_ptr.h"   // IWYU pragma: keep
+#include "nanobind/stl/variant.h"      // IWYU pragma: keep
+#include "nanobind/stl/vector.h"       // IWYU pragma: keep
 #include "xla/pjrt/c/pjrt_c_api.h"
 #include "xla/pjrt/distributed/client.h"
 #include "xla/pjrt/distributed/distributed.h"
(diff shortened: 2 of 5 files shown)


xla/pjrt/distributed/distributed.cc
```diff
@@ -38,9 +38,14 @@ GetDistributedRuntimeService(std::string address,
 }
 
 std::shared_ptr<DistributedRuntimeClient> GetDistributedRuntimeClient(
-    std::string address, const DistributedRuntimeClient::Options& options) {
-  std::shared_ptr<grpc::Channel> channel = grpc::CreateChannel(
-      address, tsl::GetClientCredentials(kVerifySecureCredentials));
+    std::string address, bool use_compression,
```
cheshire, line 41: Super-nit: we already have an options struct, maybe put `use_compression` here?

xla/python/xla.cc
```diff
@@ -779,8 +779,13 @@ NB_MODULE(xla_extension, m_nb) {
          std::optional<std::function<void(absl::Status,
                                           bool coordinator_reported_failure)>>
              missed_heartbeat_callback,
-         std::optional<bool> shutdown_on_destruction)
+         std::optional<bool> shutdown_on_destruction,
+         std::optional<bool> use_compression)
           -> std::shared_ptr<DistributedRuntimeClient> {
+        bool compression = false;
```
cheshire, line 785: Nit: `bool compression = use_compression.value_or(false)` ?

cheshire (changes requested):
How would we test this from XLA?

xla/pjrt/distributed/distributed.cc
gspschmid: See the comment on that in the PR description! (But do let me know if you feel strongly about it.)

gspschmid:
> How would we test this from XLA?
Not sure how the compression level would be observable at the level of functional tests. Or are you asking to inspect whether the channel was correctly created with `SetCompressionAlgorithm(...)` args?

So far I manually convinced myself by running with gRPC tracing (see the description of the companion JAX PR: https://github.com/jax-ml/jax/pull/23969)

cheshire:
> Not sure how the compression level would be observable at the level of functional tests

I guess make sure we at least exercise that codepath and it doesn't crash? For whether it's observable - IDK, maybe it's possible to inspect byte size of traffic somehow?

gspschmid:
> I guess make sure we at least exercise that codepath and it doesn't crash?

Added a basic smoke test.

> For whether it's observable - IDK, maybe it's possible to inspect byte size of traffic somehow?

I did a cursory search before, but couldn't find any public gRPC (C++) interfaces that would expose such stats. Happy to extend the test if you are aware of any!

cheshire (approved):

gspschmid:
@cheshire Ran `clang-format` on all touched files. Besides `client_server_test.cc`, `xla/python/xla.cc` also seemed to require formatting, but in a place I hadn't touched (https://github.com/openxla/xla/actions/runs/11071790274/job/30790008568?pr=17704#step:4:417).

gspschmid:
The copybara'd PR (https://github.com/openxla/xla/pull/17771) seems to have failed on `XLA Linux ARM64 CPU` but I can't access the internal build output (https://storage.cloud.google.com/xla-public/prod/tensorflow/xla/linux/arm64/build_cpu/33350/20240930-114900/index.html). Anything I can fix?

cheshire:
Nope, that's not the error, seems like the function removed in this PR had other users?

```
error: no matching function for call to 'GetDistributedRuntimeClient'
 2160 |       xla::GetDistributedRuntimeClient(service_address, client_options);
```

cheshire:
> So could this be in some internal code? Should only require inserting

maybe make it default option then? We have other internal PjRT callers.

gspschmid:
@cheshire Updated to make `use_compression=false` by default. Please give this another go.

cheshire (approved):

gspschmid:
@cheshire Sorry to bug you about this again -- did something else break internally?

gspschmid:
@akuegel shared the failure in `client_server_test` the internal build failed with:
```
*** SIGTERM received by PID 6078 (TID 6078) on cpu 139 from PID 1; stack trace: ***
PC: @     0x7fb99099f8ec  (unknown)  AbslInternalPerThreadSemWait
    @     0x7fb9864fd4a1       1888  FailureSignalHandler()
    @     0x7fbc4f328e80  (unknown)  (unknown)
    @     0x7fb99099f8ec         64  AbslInternalPerThreadSemWait
    @     0x7fb967e8f3b1         64  absl::Mutex::Block()
    @     0x7fb967e8f7f1        128  absl::Mutex::LockSlowWithDeadline()
    @     0x7fb967e91d3c         64  absl::Notification::WaitForNotification()
    @     0x7fbcf81c2afe        384  tsl::(anonymous namespace)::CoordinationServiceAgentImpl::Connect()
    @     0x7fbdbccdb40b         96  xla::DistributedRuntimeCoordinationServiceClient::Connect()
    @     0x7fbdbd5079c3        416  xla::(anonymous namespace)::ClientServerTest_UseCompression_Test::TestBody()
    @     0x7fbda3101779         48  testing::Test::Run()
    @     0x7fbda31027b9         64  testing::TestInfo::Run()
    @     0x7fbda31038f3         80  testing::TestSuite::Run()
    @     0x7fbda31162df        192  testing::internal::UnitTestImpl::RunAllTests()
    @     0x7fbda3115358         64  testing::UnitTest::Run()
    @     0x7fbda65b8bd8         64  main
    @     0x7fbb938e13d4        192  __libc_start_main
```

It's not entirely clear why it fails, but I wonder whether it's related to the Google-internal build passing different credentials and subsequently timing out here (https://github.com/openxla/xla/pull/17704/files#diff-097d38d9e18b20e9add655da9fb53c61fb9fd361f24f1955bcd68c4fe5190175R48) during the connection attempt. I don't think it's related to other tests in `client_server_test` going through the `GetClient(...)` helper, but perhaps.

In any case, I adapted the PR in a way that should account for either of those two options: channel creation has been factored into a separate function `GetDistributedRuntimeClientChannel(address, creds, use_compression=false)` that is used both by `GetDistributedRuntimeClient(address, client_options)` and (some of) `client_server_test`. During tests we use it to create the channel and explicitly provide insecure credentials, then pass it to the existing `GetClient` helper.

xla/python/xla.cc
```diff
@@ -36,16 +36,16 @@ limitations under the License.
 #include "llvm/Support/Casting.h"
 #include "nanobind/nanobind.h"
 #include "nanobind/nb_defs.h"
-#include "nanobind/stl/function.h"  // IWYU pragma: keep
```
cheshire, line 39: Could you avoid reformatting those? Creates quite a bit of churn/increases the likelihood of PR not landing.

cheshire (changes requested):

xla/python/xla.cc
gspschmid: The PR touches that file. An earlier build contained errors complaining about said file's formatting, including that unrelated region (https://github.com/openxla/xla/pull/17704#issuecomment-2382693881). Do you think if I reverted those formatting changes the CI would be smart enough to ignore formatting violations in the unrelated region?

cheshire (approved):

PR #17719, opened 2024-09-27
Allow compare/select on int4 data

sergey-kozub:
A simple HLO with kCompare and kSelect opcodes doesn't work for int4 types.
This PR fixes the issue.

### DIFF
--- a/xla/service/layout_normalization_test.cc
+++ b/xla/service/layout_normalization_test.cc
@@ -922,5 +922,21 @@ ENTRY main.17 {
       });
 }
 
+TEST_F(LayoutNormalizationTest, CompareInt4) {
+  const char* hlo = R"(
+HloModule module
+
+ENTRY main {
+  a = s4[10]{0:E(4)} parameter(0)
+  b = s4[10]{0:E(4)} parameter(1)
+  ROOT out = compare(a, b), direction=EQ
+}
+)";
+
+  CheckLayoutNormalization(hlo, R"(
+// CHECK: pred[10]{0} compare({{.*}})
+)");
+}
+
 }  // namespace
 }  // namespace xla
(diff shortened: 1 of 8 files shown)


akuegel (approved):
Thank you!

xla/service/layout_normalization_test.cc
```diff
@@ -922,5 +922,21 @@ ENTRY main.17 {
...
+
+ENTRY main {
+  a = s4[10]{0:E(4)} parameter(0)
+  b = s4[10]{0:E(4)} parameter(1)
+  ROOT out = compare(a, b), direction=EQ
+}
+)";
+
+  CheckLayoutNormalization(hlo, R"(
+// CHECK: pred[10]{0} compare({{.*}})
```
akuegel, line 937: It seems this CHECK line is wrong. We get this output:

%compare = pred[10]{0:E(4)} compare(%bitcast, %bitcast.2), direction=EQ

And as far as I understand, this is also the intended output.

sergey-kozub: Without the changes in this PR, it'd produce "pred[10]{0:E(4)}", which is wrong (the PRED type cannot have element_size_in_bits layout attribute set).

sergey-kozub: Are you running the test with the PR changes applied?

akuegel: Or if it is not the intended output, then I guess shape_inference.cc needs another adjustment, as MakeCompareHlo will use it to derive the output layout.

sergey-kozub: It looks I split my commit incorrectly, and some changes are missing. Fixing now.

sergey-kozub: Added the missing change. Sorry for the inconvenience.

akuegel (approved):

PR #17749, opened 2024-09-30
[NVIDIA] Added a flag to control the total number of collective resource in LHS

Tixxx:
Added a flag to control the total number of collective resource in LHS, the goal is to allow the scheduler to overlap multiple collectives under a large compute:
```
  With xla_gpu_scheduler_parallel_collective_resource = 1:
     coll.1-start = collective(input)
     coll.1-done = collective(coll.1-start)
     coll.2-start = collective(input2)
     large_compute = compute(op)  // because only 1 collective is allowed in-flight, only schedule 1 start-done pair here
     coll.2-done = collective(coll.2-start)
   With xla_gpu_scheduler_parallel_collective_resource = 2:
     coll.1-start = collective(input)
     coll.2-start = collective(input2)
     large_compute = compute(op)  // 2 collectives can be schedule in-flight at once now, much better schedule
     coll.1-done = collective(coll.1-start)
     coll.2-done = collective(coll.2-start)

```

Note that since we only have 1 collective stream, collectives will be serialized anyway, but with more resource, some of them can start much earlier.

### DIFF
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -463,6 +463,16 @@ absl::StatusOr<ScheduleMetadata> ScheduleGpuModule(
       module->config()
           .debug_options()
           .xla_gpu_experimental_parallel_collective_overlap_limit());
+  CHECK((config.collective_broadcast_overlap_limit <=
+         config.parallel_collective_overlap_limit) &&
+        (config.all_to_all_overlap_limit <=
+         config.parallel_collective_overlap_limit) &&
+        (config.all_gather_overlap_limit <=
+         config.parallel_collective_overlap_limit) &&
+        (config.all_reduce_overlap_limit <=
+         config.parallel_collective_overlap_limit) &&
+        (config.reduce_scatter_overlap_limit <=
+         config.parallel_collective_overlap_limit));
   auto gpu_latency_estimator =
       std::make_unique<GpuLatencyEstimator>(pointer_size);
 
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -259,7 +259,7 @@ SchedulerConfig GetSchedulerConfig(int64_t memory_limit,
   config.aggressive_scheduling_policies = true;
   config.schedule_send_recvs = true;
   config.memory_limit = memory_limit;
-  config.global_collective_scheduling_resource = collective_resource;
+  config.parallel_collective_overlap_limit = collective_resource;
   return config;
 }
 
@@ -459,9 +459,10 @@ absl::StatusOr<ScheduleMetadata> ScheduleGpuModule(
   }
 
   SchedulerConfig config = GetSchedulerConfig(
-      memory_limit, module->config()
-                        .debug_options()
-                        .xla_gpu_scheduler_parallel_collective_resource());
+      memory_limit,
+      module->config()
+          .debug_options()
+          .xla_gpu_experimental_parallel_collective_overlap_limit());
   auto gpu_latency_estimator =
       std::make_unique<GpuLatencyEstimator>(pointer_size);
 
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -249,7 +249,8 @@ HloInstructionSequence PostprocessorToScheduleSyncCollectives(
 
 // Latency hiding scheduler support.
 
-SchedulerConfig GetSchedulerConfig(int64_t memory_limit) {
+SchedulerConfig GetSchedulerConfig(int64_t memory_limit,
+                                   int64_t collective_resource) {
   SchedulerConfig config;
   config.all_reduce_overlap_limit = 1;
   config.collective_broadcast_overlap_limit = 1;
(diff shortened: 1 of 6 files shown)


xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -258,6 +259,7 @@ SchedulerConfig GetSchedulerConfig(int64_t memory_limit) {
   config.aggressive_scheduling_policies = true;
   config.schedule_send_recvs = true;
   config.memory_limit = memory_limit;
+  config.global_collective_scheduling_resource = collective_resource;
```
golechwierowicz, line 262: How will this work with other configurations options above like `all_reduce_overlap_limit`?

xla/service/latency_hiding_scheduler.h
```diff
@@ -139,6 +139,7 @@ struct SchedulerConfig {
   bool enable_selective_resources = false;
   int64_t max_hops_to_closest_selective_overlap = 0;
   int64_t rerun = 0;
+  int64_t global_collective_scheduling_resource = 1;
```
golechwierowicz, line 142: Maybe change to `parallel_collective_overlap_limit`?

golechwierowicz (changes requested):
I have an additional high level question. Where are you going to implement the logic which avoids NCCL deadlocks?

Tixxx:
> I have an additional high level question. Where are you going to implement the logic which avoids NCCL deadlocks?

We will provide additional logic to override the IsResourceConstrained function when the global collective resource is >1. The function is used for determining whether a collective should be scheduled based on the currently in-flight collectives. The additional logic will determine whether multiple collectives can be in parallel based on these:

1. if total resource == 2, we will allow collectives to be scheduled in parallel if their replica groups have at most 1 overlapping rank, ie {0,1,2} and {2,3,4} - okay; {0,1,2} and {1,2,3} - NOT okay
2. if total resource > 2, we will only allow collectives to be scheduled in parallel if all of their replica groups are completely non-overlapping, ie for resource=3, {0,1,2}, {3,4,5} and {5,6,7} - okay; {0,1,2}, {2,3,4}, {0,4,5} - NOT okay

Also a heads-up, we will be sending another pr with the above change soon.

xla/xla.proto
```diff
@@ -1002,7 +1002,21 @@ message DebugOptions {
...
+  //   coll.1-start = collective(input)
+  //   coll.1-done = collective(coll.1-start)
+  //   coll.2-start = collective(input2)
+  //   coll.2-done = collective(coll.2-start)
+  // With xla_gpu_scheduler_parallel_collective_resource = 2:
+  //   coll.1-start = collective(input)
+  //   coll.2-start = collective(input2)
+  //   coll.1-done = collective(coll.1-start)
+  //   coll.2-done = collective(coll.2-start)
+  int32 xla_gpu_scheduler_parallel_collective_resource = 335;
```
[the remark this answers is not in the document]
Tixxx: all good points, the name makes sense, i will change it.

xla/service/gpu/gpu_hlo_schedule.cc
Tixxx: the other overlap limits are not used anywhere so im not sure how this will interact with the other ones. They are probably used in other backends. IIRC the other limits controls individual collective scheduling resource but not the mix-and-match resource limit.

xla/service/latency_hiding_scheduler.h
Tixxx: yea sounds good

xla/service/gpu/gpu_hlo_schedule.cc
golechwierowicz: Thanks! Would it make sense to CHECK fail if there are different (>1) values provided for these fields AND global_collective_scheduling_resource > 1?

golechwierowicz (approved):

derdrdirk:
Hi @Tixxx,
could you add a unit test for your changes?

Tixxx:
> Hi @Tixxx, could you add a unit test for your changes?

Thanks @derdrdirk , added a unit test for it.

xla/service/gpu/gpu_hlo_schedule.cc
Tixxx: I think having a check makes sense, though we should still allow these individual values to be set > 1 together with global resource, but each individual value cannot exceed the global resource.

derdrdirk (approved):

derdrdirk:
Thank you :)

felixwqp:
> > I have an additional high level question. Where are you going to implement the logic which avoids NCCL deadlocks?
> 
> We will provide additional logic to override the IsResourceConstrained function when the global collective resource is >1. The function is used for determining whether a collective should be scheduled based on the currently in-flight collectives. The additional logic will determine whether multiple collectives can be in parallel based on these:
> 
> 1. if total resource == 2, we will allow collectives to be scheduled in parallel if their replica groups have at most 1 overlapping rank, ie {0,1,2} and {2,3,4} - okay; {0,1,2} and {1,2,3} - NOT okay
> 2. if total resource > 2, we will only allow collectives to be scheduled in parallel if all of their replica groups are completely non-overlapping, ie for resource=3, {0,1,2}, {3,4,5} and {5,6,7} - okay; {0,1,2}, {2,3,4}, {0,4,5} - NOT okay
> 
> Also a heads-up, we will be sending another pr with the above change soon.


can you share the rationale behind` if total resource == 2, we will allow collectives to be scheduled in parallel if their replica groups have at most 1 overlapping rank`

Can I assume NCCL will not deadlock for this scenario?

PR #17754, opened 2024-09-30
[NFC] Extract common builder code into hlo_creation_utils

sergey-kozub:
Added `XlaComputationToHloComputation` helper function to `hlo_creation_utils` and replace the uses.

### DIFF
--- a/xla/service/hlo_creation_utils.h
+++ b/xla/service/hlo_creation_utils.h
@@ -22,6 +22,7 @@ limitations under the License.
 #include <vector>
 
 #include "absl/types/span.h"
+#include "xla/hlo/builder/xla_computation.h"
 #include "xla/hlo/ir/hlo_computation.h"
 #include "xla/hlo/ir/hlo_instruction.h"
 #include "xla/literal_util.h"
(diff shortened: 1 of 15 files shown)


akuegel (approved):

sergey-kozub:
Which copybara checks have failed? Is this something I can fix in OSS?

xla/service/hlo_creation_utils.h
```diff
@@ -22,6 +22,7 @@ limitations under the License.
 #include <vector>
 
 #include "absl/types/span.h"
+#include "xla/hlo/builder/xla_computation.h"
```
akuegel, line 25: You need to add a corresponding BUILD dependency to the hlo_creation_utils target.

sergey-kozub: Fixed by replacing the include with "xla/client/xla_computation.h"
I guess we should remove this alias at some point.

akuegel (approved):

xla/service/hlo_creation_utils.h
sergey-kozub: Updated, there was just one file; the rest just depend on ":hlo_creation_utils".

akuegel (approved):

PR #17809, opened 2024-10-01
Added free-threading support to WeakrefLRUCache

vfdev-5:
Decsription:
- Added free-threading support to WeakrefLRUCache
- Added another multithreaded test

### DIFF
--- a/xla/python/weakref_lru_cache.cc
+++ b/xla/python/weakref_lru_cache.cc
@@ -215,6 +215,12 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
           if (cache == nullptr) {
             return;
           }
+          // Set up PyCriticalSection for cache python associated object;
+          auto py_cache = nb::find(cache);
+          // This should never happen as python cache should always be found
+          CHECK(py_cache.ptr() != NULL);
+          nb::ft_object_guard lock(py_cache);
+
           // The object the reference referred to is now in the process of being
           // destroyed, so we cannot refer to its contents. Python weakref
           // objects compare based on identity if the object they refer to is
@@ -367,10 +373,10 @@ void BuildWeakrefLRUCacheAPI(nb::module_& m) {
       nb::class_<WeakrefLRUCache>(m, "WeakrefLRUCache",
                                   nb::is_weak_referenceable(),
                                   nb::type_slots(WeakrefLRUCache::slots_))
-          .def("__call__", &WeakrefLRUCache::Call)
-          .def("cache_keys", &WeakrefLRUCache::GetKeys)
-          .def("cache_info", &WeakrefLRUCache::GetCacheInfo)
-          .def("cache_clear", &WeakrefLRUCache::Clear);
+          .def("__call__", &WeakrefLRUCache::Call, nb::lock_self())
+          .def("cache_keys", &WeakrefLRUCache::GetKeys, nb::lock_self())
+          .def("cache_info", &WeakrefLRUCache::GetCacheInfo, nb::lock_self())
+          .def("cache_clear", &WeakrefLRUCache::Clear, nb::lock_self());
   nb::class_<WeakrefLRUCache::CacheInfo>(weakref_lru_cache,
                                          "WeakrefLRUCacheInfo")
       .def_ro("hits", &WeakrefLRUCache::CacheInfo::hits)
--- a/xla/python/weakref_lru_cache_test.py
+++ b/xla/python/weakref_lru_cache_test.py
@@ -76,6 +76,37 @@ class WeakrefLRUCacheTest(absltest.TestCase):
       cache(wrkey, GilReleasingCacheKey())
     t.join()
 
+  def testAnotherMultiThreaded(self):
+    num_workers = 5
+    barrier = threading.Barrier(num_workers)
+    cache = xla_client.weakref_lru_cache(lambda: None, lambda x, y: y, 2048)
+
+    class WRKey:
+      pass
+
+    def WorkerAddToCache():
+        barrier.wait()
+        wrkey = WRKey()
+        for i in range(10):
+            cache(wrkey, i)
+
+    def WorkerCleanCache():
+        barrier.wait()
+        for _ in range(10):
+            cache.cache_clear()
+
+    workers = [
+        threading.Thread(target=WorkerAddToCache) for _ in range(num_workers - 1)
+    ] + [
+        threading.Thread(target=WorkerCleanCache)
+    ]
+
+    for t in workers:
+        t.start()
+
+    for t in workers:
+        t.join()
+
   def testKwargsDictOrder(self):
     miss_id = 0
 


xla/python/weakref_lru_cache.cc
```diff
@@ -150,6 +150,11 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
       : cache_context_fn_(cache_context_fn), fn_(fn), lru_list_(maxsize) {}
 
   std::shared_ptr<Cache> GetCache(WeakrefCacheKey key) {
+#ifdef NB_FREE_THREADED
+    mu_.Lock();
```
hawkinsp, line 154: In general, you should use:

```
absl::MutexLock lock(&mu_);
```
rather than using `absl::Cleanup`.

(What you wrote works, but there's a specialized RAII lock holder widget for this exact purpose.)

xla/python/weakref_lru_cache.cc
```diff
@@ -184,14 +194,26 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
...
-    nb::object context = cache_context_fn_();
-    std::shared_ptr<Cache> cache_ptr = GetCache(WeakrefCacheKey{
-        weakref_key, static_cast<size_t>(nb::hash(weakref_key))});
+#ifdef NB_FREE_THREADED
+    mu_.Lock();
+#endif
+    auto wrcachekey = WeakrefCacheKey{
+        weakref_key, static_cast<size_t>(nb::hash(weakref_key))};
+#ifdef NB_FREE_THREADED
+    mu_.Unlock();
```
hawkinsp, line 203: Rather than locally unlocking around the call to `GetCache`, I'd just have `GetCache()` assume the lock is held? You hold the lock for the entirety of `GetCache` anyway.

Additionally, you can compute wrcachekey without holding the lock.

xla/python/weakref_lru_cache.cc
```diff
@@ -184,14 +194,26 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
...
+#endif
+    std::shared_ptr<Cache> cache_ptr = GetCache(wrcachekey);
+#ifdef NB_FREE_THREADED
+    mu_.Lock();
+    absl::Cleanup unlock = [this]() ABSL_UNLOCK_FUNCTION(mu_) { mu_.Unlock(); };
+#endif
     Cache& cache = *cache_ptr;
+    nb::object context = cache_context_fn_();
```
hawkinsp, line 211: Do work that doesn't need the lock outside the lock scope, wherever possible. Here `cache_context_fn` does not need the lock.

xla/python/weakref_lru_cache.cc
hawkinsp: I also wonder if we could build a widget:

class MutexLockIfFreeThreaded {

};

that expands to a MutexLock if freethreading is enabled and nothing otherwise, to save all the ifdefs?

xla/python/weakref_lru_cache.cc
```diff
@@ -184,14 +194,26 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
...
     ++total_queries_;
 
     bool inserted = false;
     std::shared_ptr<CacheEntry> entry;
+#ifndef NB_FREE_THREADED
```
hawkinsp, line 216: I think you can make this an else branch to the free threaded case above.

hawkinsp (changes requested):

xla/python/weakref_lru_cache.cc
```diff
@@ -174,6 +174,9 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
       : cache_context_fn_(cache_context_fn), fn_(fn), lru_list_(maxsize) {}
 
   std::shared_ptr<Cache> GetCache(WeakrefCacheKey key) {
+#ifdef NB_FREE_THREADED
```
hawkinsp, line 177: Even if we keep using this locking, wouldn't it be simpler to simply move the call to `GetCache` of `mu_` after the lock acquisition in `Call`? No need to repeatedly acquire and release the lock.

xla/python/weakref_lru_cache.cc
```diff
@@ -240,10 +247,11 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
         inserted = true;
```
hawkinsp, line 247: I'd just move the increment of `misses` somewhere here while you already hold `mu_`.

hawkinsp (changes requested):

xla/python/weakref_lru_cache.cc
```diff
@@ -301,6 +300,15 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
     deferred_deletes.clear();
   }
 
+protected:
```
hawkinsp, line 303: `private:`

We don't expect this class to be subclassed, so I think just say `private`.

xla/python/weakref_lru_cache.cc
```diff
@@ -240,10 +236,13 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
         inserted = true;
         return std::make_shared<CacheEntry>();
       });
+      ++total_queries_;
+      if (inserted) {
```
hawkinsp, line 240: You could make this

```
if (inserted || !entry->completed.HasBeenNotified())
```

which slightly overcounts (if the entry is in the process of being populated) but I think seems better to me.

xla/python/weakref_lru_cache.cc
```diff
@@ -216,21 +211,22 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
...
-    std::shared_ptr<Cache> cache_ptr = GetCache(wrcache_key);
-    Cache& cache = *cache_ptr;
-    ++total_queries_;
 
     bool inserted = false;
     std::shared_ptr<CacheEntry> entry;
     {
       // Because the gil can be released during cache insertion, this forces
       // the lock order to be mu_ then gil so we must release the gil first.
+      // In free-threading  this operation also temporarily releases all
```
hawkinsp, line 220: This may be unnecessary, but let's leave it in for now.

xla/python/weakref_lru_cache.cc
```diff
@@ -318,8 +326,8 @@ void BuildWeakrefLRUCacheAPI(nb::module_& m) {
                                   nb::is_weak_referenceable())
           .def("__call__", &WeakrefLRUCache::Call)
           .def("cache_keys", &WeakrefLRUCache::GetKeys)
-          .def("cache_info", &WeakrefLRUCache::GetCacheInfo)
-          .def("cache_clear", &WeakrefLRUCache::Clear);
+          .def("cache_info", &WeakrefLRUCache::GetCacheInfo, nb::lock_self())
```
hawkinsp, line 329: Remove these. We're not usefully protecting anything with the self lock.

xla/python/weakref_lru_cache.cc
```diff
@@ -279,7 +279,6 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
         results.push_back(std::move(result));
       }
     }
-    mu_.Unlock();
     return results;
   }
   CacheInfo GetCacheInfo() const {
```
hawkinsp, line 284: You need to lock mu_ around the data structure accesses in GetCacheInfo() and Clear(), if mu_ is the lock that protects the data structure.

You should release the gil before acquiring that mutex because of lock ordering.

hawkinsp (changes requested):

xla/python/weakref_lru_cache.cc
```diff
@@ -279,10 +286,12 @@ class WeakrefLRUCache : public std::enable_shared_from_this<WeakrefLRUCache> {
...
+#ifdef NB_FREE_THREADED
+    absl::MutexLock lock(&mu_);
```
vfdev-5, line 293: There are no callbacks to python in this method, so I think it is safe not releasing GIL before locking the mutex

dimitar-asenov:
@hawkinsp Could you please take a look at the updated PR and comments?

hawkinsp:
I spent a fair amount of time thinking about this PR today.

I came to the following conclusion:
a) we should revert everything except for the changes to the test
b) add an `nb::lock_self()` guard on the bindings around the methods:
```
          .def("__call__", &WeakrefLRUCache::Call, nb::lock_self())
          .def("cache_keys", &WeakrefLRUCache::GetKeys, nb::lock_self())
          .def("cache_info", &WeakrefLRUCache::GetCacheInfo, nb::lock_self())
          .def("cache_clear", &WeakrefLRUCache::Clear, nb::lock_self());
```

Can you give that a try?

The locking in this class is actually really complicated, but I think I've figured out how it is supposed to work, and the key thing is that using a Python critical section instead of the GIL should more or less replicate how it used to work with the GIL.

hawkinsp (approved):

hawkinsp:
The clang-format errors are not from this change, so I think we can ignore them.

xla/python/weakref_lru_cache_test.py
```diff
@@ -76,6 +76,37 @@ def Body():
...
 
+  def testAnotherMultiThreaded(self):
+    num_workers = 5
+    barrier = threading.Barrier(num_workers)
+    cache = xla_client.weakref_lru_cache(lambda: None, lambda x, y: y, 2048)
+
+    class WRKey:
+      pass
+
+    def worker_add_to_cache():
```
penpornk, line 87: Please use Camel case for function names, e.g., WorkerAddToCache.
```suggestion
    def WorkerAddToCache():
```

xla/python/weakref_lru_cache_test.py
```diff
@@ -76,6 +76,37 @@ def Body():
...
+        barrier.wait()
+        wrkey = WRKey()
+        for i in range(10):
+            cache(wrkey, i)
+
+    def worker_clean_cache():
```
penpornk, line 93: Same here. Please use Camel case for function names.
```suggestion
    def WorkerCleanCache():
```

xla/python/weakref_lru_cache_test.py
```diff
@@ -76,6 +76,37 @@ def Body():
...
+        barrier.wait()
+        for i in range(10):
```
penpornk, line 95: Variable `i` is unused. Please use `_` instead.

penpornk (changes requested):

penpornk (approved):
Thank you!

xla/python/weakref_lru_cache_test.py
```diff
@@ -76,6 +76,37 @@ def Body():
...
+    barrier = threading.Barrier(num_workers)
+    cache = xla_client.weakref_lru_cache(lambda: None, lambda x, y: y, 2048)
+
+    class WRKey:
+      pass
+
+    def WorkerAddToCache():
+        barrier.wait()
+        wrkey = WRKey()
+        for _ in range(10):
```
penpornk, line 90: This loop does use `i` (unlike the other loop). Could you please add `i` back?

vfdev-5: Sorry, I should have checked better this before updating.

penpornk (approved):
Thank you for the fix!

PR #17819, opened 2024-10-01
Added use_enabled_free_threading flag to build nanobind with NB_FREE_THREADED=1

vfdev-5:
Description:
- Added use_enabled_free_threading flag to build nanobind with NB_FREE_THREADED=1 
- Set nanobind version to tag v2.2.0: https://github.com/wjakob/nanobind/releases/tag/v2.2.0

Build command:
```bash
./configure.py --backend=CPU

bazel build --test_output=all --spawn_strategy=sandboxed --define=use_enabled_free_threading=true //xla/...
```

### DIFF
--- a/third_party/nanobind/nanobind.BUILD
+++ b/third_party/nanobind/nanobind.BUILD
@@ -1,6 +1,19 @@
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 licenses(["notice"])
 
 package(default_visibility = ["//visibility:public"])
+
+bool_flag(
+    name = "enabled_free_threading",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "use_enabled_free_threading",
+    flag_values = {
+        ":enabled_free_threading": "True",
+    },
+)
 
 cc_library(
     name = "nanobind",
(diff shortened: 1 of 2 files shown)


hawkinsp (approved):

hawkinsp (approved):

third_party/nanobind/nanobind.BUILD
```diff
@@ -1,7 +1,20 @@
...
 
+bool_flag(
+    name = "enabled_free_threading",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "use_enabled_free_threading",
+    flag_values = {
+        ":enabled_free_threading": "False",
```
hawkinsp, line 14: The sense of this condition is negated, it should be "True" I think instead.

thomasjoerg:
@hawkinsp The internal CI detected a build breakage for JAX Linux CPU:
`error: "Free-threaded extensions require a free-threaded version of Python"`

Does this need to be addressed or is it benign?

third_party/nanobind/nanobind.BUILD
vfdev-5: We can rename this flag to make it transparent. The idea is to set True for free-threading enabled cpython and by default to have GIL <-> False

vfdev-5:
> Does this need to be addressed or is it benign?

I think we should address it in this PR and by default have the flag using GIL enabled mode

thomasjoerg:
> > Does this need to be addressed or is it benign?
> 
> I think we should address it in this PR and by default have the flag using GIL enabled mode

Thank you. I'll make another attempt at merging this PR once the flag has been added.

hawkinsp (approved):

PR #17838, opened 2024-10-01
Add .pylintrc Configuration File

apivovarov:
This PR introduces a `.pylintrc` file to the XLA project.

XLA, which is used in TensorFlow, relies on TensorFlow's CI builds to automatically test its code, including Python components like xla_client. Recently, issues were encountered with TensorFlow's CI reporting errors such as pylint C0301: Line too long in xla_client_test.py.

TF CI links:
- https://github.com/tensorflow/tensorflow/commit/dd2c1e9d94cfe5c471b5788917a2f74140d9b144
- https://github.com/tensorflow/tensorflow/actions/runs/11130353052/job/30929504884

I found that XLA lacked a dedicated Pylint configuration (e.g., `.pylintrc` or `pyproject.toml`), which made it harder to automatically receive proper warnings and errors, particularly when working in editors like VSCode.

To address this, I have added a `.pylintrc` file to the XLA project. This file is based on TensorFlow's configuration, with modifications made to remove or fix unsupported configuration lines.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,277 @@
+[MASTER]
+
+# Add files or directories to the denylist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# List of plugins (as comma separated values of python modules names) to load,
+# usually to register additional checkers.
+load-plugins=pylint.extensions.docparams
+accept-no-param-doc=no
+
+[MESSAGES CONTROL]
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once).You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use"--disable=all --enable=classes
+# --disable=W"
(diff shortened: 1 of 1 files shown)


reedwm:
I don't know much about pylint. @ddunl can you review this?

ddunl:
I'll have to investigate what we do internally, an explicit goal I have for XLA is that the CI maps directly to what we do internally as often as possible (this is a work in progress, and sometimes is not possible). I'll have to see what we do internally. It's not totally clear to me that what TF does here is correct, because I don't think that on e.g. a PR to TF, that this pylint config is enforced in any meaningful way. I'll look into this and get back to you

apivovarov:
JAX utilizes the [pyproject.toml](https://github.com/jax-ml/jax/blob/main/pyproject.toml) file, which includes the [tool.ruff.lint] section and other Ruff configurations.

However, since the XLA development workflow is closely integrated with TensorFlow CI, we need to adhere to TensorFlow's Pylint rules. This approach helps us avoid spending valuable time resolving XLA PR merge issues reported by TensorFlow CI.

ddunl:
TF and JAX behave differently in this regard, TF may have that pylint job on PRs, but internally the status of the pylint on the PR is irrelevant - only the internal linter checks are respected. JAX does what I would prefer XLA do which is check both. I'm not exactly sure the mechanism by which JAX keeps the internal config in sync with the external one. One starting point I'd be interested in is to use https://google.github.io/styleguide/pylintrc and see if that currently agrees with all code we have checked in

apivovarov:
[TF pylintrc](https://gist.github.com/apivovarov/c60b957af4656478bc417423c1843fea) reports about 80 issues in xla/python

[Styleguide pylintrc](https://gist.github.com/apivovarov/0b26c8c318da8ad81b96fe4eca533feb) reports about 1000 issues

PR #17890, opened 2024-10-03
[ROCM] Add nanoo fp8 support in type traits

ScXfjiang:
We are trying to add NANOO FP8 support in TensorFlow. To achieve this, a small modification in the `type_traits.h` file on the XLA side is required, where the NANOO FP8 should also be classified as a simple type. This change will be utilized in tensor.cc within TensorFlow. You can find the relevant code here: [TensorFlow tensor.cc](https://github.com/tensorflow/tensorflow/blob/ba7f93b8dc32b64cc0bb5ddb44e75bd344ae67d0/tensorflow/core/framework/tensor.cc#L175).

### DIFF
--- a/xla/tsl/framework/type_traits.h
+++ b/xla/tsl/framework/type_traits.h
@@ -73,8 +73,10 @@ struct is_simple_type {
       std::is_same<T, float8_e3m4>::value ||
       std::is_same<T, float8_e4m3>::value ||
       std::is_same<T, float8_e4m3fn>::value ||
+      std::is_same<T, float8_e4m3fnuz>::value ||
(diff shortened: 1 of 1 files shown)


ScXfjiang:
FYI: For the context on NANOO FP8, please refer to this comment:
 https://github.com/google/flax/pull/3993#issue-2350000228

ScXfjiang:
Hi @majnemer could you take a look at this PR?

i-chaochen:
Hi @xla-rotation could you check this PR please?

ezhulenev (approved):

PR #17980, opened 2024-10-07
[ROCm] Fix pjrt_c_api_gpu_test for ROCm

hsharsha:


xla/pjrt/c/pjrt_c_api_gpu_test.cc
```diff
@@ -507,9 +507,13 @@ TEST(PJRTGpuDeviceTopologyTest, CreateExplicitGpuTopologyAndTargetConfig) {
       reinterpret_cast<const PJRT_TopologyDescription*>(args.topology);
   ASSERT_NE(pjrt_topology, nullptr);
 
+#ifdef TENSORFLOW_USE_ROCM
```
dimitar-asenov, line 510: Please do not use macros in new XLA code, we're actively trying to remove them.

Use a run-time call to determine the platform. Not sure if there is a better approach here, but in the worst case you can check that platform_id is either xla::RocmId() or xla::CudaId() and check that the platform name is consistent.

dimitar-asenov: Another comment: On line 492, this creates an explicit target_config that uses `platform_name: CUDA` (line 479). Is this supposed to work with ROCm? That doesn't seem right. If yes, then could you please write a comment right before line 490, to explain why this is OK?

dimitar-asenov (changes requested):

jcaraban:
@dimitar-asenov how about replacing the `#ifdefs ROCM` for `AnyOf(cuda,rocm)`? At least this fixes the broken test (which is the immediate issue) w/o extra macros. The remaining macros require a deeper reorganization of the test, I believe.

https://github.com/openxla/xla/blob/3e3a5fd68f1e2e7b2d2b6baa4ce3fdd982737a79/xla/pjrt/c/pjrt_c_api_gpu_test.cc#L63-L71
https://github.com/openxla/xla/blob/3e3a5fd68f1e2e7b2d2b6baa4ce3fdd982737a79/xla/pjrt/c/pjrt_c_api_gpu_test.cc#L274-L284

xla/pjrt/c/pjrt_c_api_gpu_test.cc
```diff
@@ -437,13 +435,10 @@ TEST(PJRTGpuDeviceTopologyTest, CreateGpuTopology) {
...
-#ifdef TENSORFLOW_USE_ROCM
-  EXPECT_EQ(pjrt_topology->topology->platform_id(), xla::RocmId());
-  EXPECT_EQ(pjrt_topology->topology->platform_name(), xla::RocmName());
-#else
-  EXPECT_EQ(pjrt_topology->topology->platform_id(), xla::CudaId());
-  EXPECT_EQ(pjrt_topology->topology->platform_name(), xla::CudaName());
-#endif
+  EXPECT_THAT(pjrt_topology->topology->platform_id(),
+              testing::AnyOf(xla::CudaId(), xla::RocmId()));
+  EXPECT_THAT(pjrt_topology->topology->platform_name(),
```
dimitar-asenov, line 440: This is fine so far, but it would be good to test that platform_id and name are consistent (both Cuda, or both Rocm)

dimitar-asenov (changes requested):

dimitar-asenov:
> @dimitar-asenov how about replacing the `#ifdefs ROCM` for `AnyOf(cuda,rocm)`? At least this fixes the broken test (which is the immediate issue) w/o extra macros. The remaining macros require a deeper reorganization of the test, I believe.
> 
> https://github.com/openxla/xla/blob/3e3a5fd68f1e2e7b2d2b6baa4ce3fdd982737a79/xla/pjrt/c/pjrt_c_api_gpu_test.cc#L63-L71
> 
> 
> https://github.com/openxla/xla/blob/3e3a5fd68f1e2e7b2d2b6baa4ce3fdd982737a79/xla/pjrt/c/pjrt_c_api_gpu_test.cc#L274-L284

That's fine, but we should also test for consistency (see my other comment). That wouldn't require any macros.

xla/pjrt/c/pjrt_c_api_gpu_test.cc
```diff
@@ -437,13 +435,10 @@ TEST(PJRTGpuDeviceTopologyTest, CreateGpuTopology) {
...
   ASSERT_NE(pjrt_topology, nullptr);
 
-#ifdef TENSORFLOW_USE_ROCM
-  EXPECT_EQ(pjrt_topology->topology->platform_id(), xla::RocmId());
-  EXPECT_EQ(pjrt_topology->topology->platform_name(), xla::RocmName());
-#else
-  EXPECT_EQ(pjrt_topology->topology->platform_id(), xla::CudaId());
-  EXPECT_EQ(pjrt_topology->topology->platform_name(), xla::CudaName());
-#endif
+  EXPECT_TRUE((pjrt_topology->topology->platform_id() == xla::CudaId() &&
```
jcaraban, line 438: @dimitar-asenov now testing for `(CudaID and CudaName) or (RocmID and RocmName)`

dimitar-asenov (approved):

xla/pjrt/c/pjrt_c_api_gpu_test.cc
dimitar-asenov: Thanks. Looks good.

PR #17999, opened 2024-10-07
[NVIDIA GPU] Added a predicate function to dot merger to determine eligibility

Tixxx:
Expose a predicate function interface through dot merger to pass in backend-specific compatibility check.
This is used in gpu compiler to avoid merging dots that are assigned with 2 different non-default stream ids as they will be optimized separately.

### DIFF
--- a/xla/service/dot_merger.h
+++ b/xla/service/dot_merger.h
@@ -56,8 +56,12 @@ namespace xla {
 // operands to be concatenated.
 class DotMerger : public HloModulePass {
  public:
-  explicit DotMerger(int64_t max_size_to_merge)
-      : max_size_to_merge_(max_size_to_merge) {}
+  explicit DotMerger(
+      int64_t max_size_to_merge,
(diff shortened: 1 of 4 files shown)


xla/service/dot_merger.h
```diff
@@ -67,6 +71,9 @@ class DotMerger : public HloModulePass {
 
  private:
   int64_t max_size_to_merge_;
+  // Predicate function for backend-specific compatibility check.
+  std::function<bool(const HloInstruction* a, const HloInstruction* b)>
+      is_compatible_;
```
reedwm, line 76: I would call this `can_merge_`. And call the arguments `dot_a` and `dot_b` to make it clear these are dot instructions to potentially merge.

reedwm (changes requested):

xla/service/dot_merger.h
Tixxx: sounds good, changed the names

reedwm (approved):

PR #18065, opened 2024-10-08
Pass Ordering Test for GPU Compiler

philipphack:
Adds a class for testing the order of passes in the GPU compiler. The names of the passes expected to run first and last can be described by regular expressions.

Also adds a test for verifying the order of the collective quantizer and collective pipeliner passes.

### DIFF
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -1058,8 +1058,6 @@ ENTRY main {
 
 class PassOrderTest : public GpuCompilerTest {
  public:
-  PassOrderTest() { SetModuleConfig(GetModuleConfigForTest()); }
-
   void SetDebugOptions(const DebugOptions& options) {
     HloModuleConfig config = GetModuleConfigForTest();
     config.set_debug_options(options);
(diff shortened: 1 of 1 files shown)


xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1056,67 +1056,74 @@ ENTRY main {
...
-  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
-                          ParseAndReturnVerifiedModule(constant_module));
-  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> optimized_module,
-                          GetOptimizedModule(std::move(module)));
+  // Fails if any of the passes with names matching the regular expression
+  // first_regex run after any of the passes matching last_regex or if
+  // first_regex or last_regex matches none of the executed passes.
+  void VerifyPassOrder(absl::string_view first_regex,
+                       absl::string_view last_regex) {
+    int first_last_run = -1;
```
reedwm, line 1086: Change this to `latest_run_of_first`. Or maybe `first_latest_run`. And similarly for `last_first_run`.

reedwm (changes requested):

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1056,67 +1056,75 @@ ENTRY main {
...
+    config.set_debug_options(options);
+    SetModuleConfig(config);
   }
 
-  return passes;
-}
-
-TEST_F(GpuCompilerPassTest, PassesAreRunInCorrectOrder) {
-  constexpr absl::string_view constant_module = R"(
+  void SetModuleConfig(const HloModuleConfig& config) {
```
bchetioui, line 1067: Do we need this additional indirection? (The name is also not intuitive for what it does.)

It seems like we could get away just fine with `SetDebugOptions`, and have a `private` "`CompileModule`" that would generate the module if it doesn't exist, and is called in `VerifyPassOrder`.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1056,67 +1056,75 @@ ENTRY main {
...
 
-  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
-                          ParseAndReturnVerifiedModule(constant_module));
-  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> optimized_module,
-                          GetOptimizedModule(std::move(module)));
+  // Fails if any of the passes with names matching the regular expression
+  // first_regex run after any of the passes matching last_regex or if none of
+  // the executed passes matches first_regex or last_regex.
+  void VerifyPassOrder(absl::string_view first_regex,
+                       absl::string_view last_regex) {
```
bchetioui, line 1083: I'm having a bit of a hard time following the terminology between `first_latest` and `last_earliest`.

Could we stick to the previous terminology? (I.e. `first` -> `before`, `earliest` -> `first`, ...)
We could even make it `before_pass_regex` to enhance clarity further.

bchetioui (changes requested):

bchetioui (approved):
Thanks for the enhancement!

PR #18117, opened 2024-10-09
Use StableHLO WEEK_4 in GetDefaultStablehloVersion

apivovarov:
This is needed to test new Stable HLO types (f8e4m3 and f8e3m8) in JAX on pjrt devices such as GPU.

WEEK_12 currently returns 1.3.0 - too old
f8e4m3 and f8e3m8 dtypes were implemented in Stable HLO 1.7.0 (released on Sep 4, 2024)
WEEK_4 currently returns 1.7.0
Using WEEK_4 in pjrt GetDefaultStablehloVersion() should solve the issue with JAX GPU tests

More details:

- `PjRtCApiCompiler::Compile()` uses `xla::GetDefaultStablehloVersion()` - xla/pjrt/pjrt_c_api_client.cc:2317
- `GetDefaultStablehloVersion` uses mlir::vhlo::Version::CompatibilityRequirement::**WEEK_12** - xla/pjrt/mlir_to_hlo.cc:257
**WEEK_12** returns StableHLO version **1.3.0** instead of 1.7.0

Related JAX PR: https://github.com/jax-ml/jax/pull/23585

### DIFF
--- a/xla/pjrt/mlir_to_hlo.cc
+++ b/xla/pjrt/mlir_to_hlo.cc
@@ -245,16 +245,16 @@ absl::Status UpgradeVersionedStablehlo(mlir::ModuleOp mlir_module) {
 }
 
 std::string GetDefaultStablehloVersion(std::optional<int64_t> plugin_version) {
-  // TODO: (b/370803410) Use WEEK_12 in PJRT, some plugins were not up to date,
+  // TODO: (b/370803410) Use WEEK_4 in PJRT, some plugins were not up to date,
   // so temporarily using 1.0.0 to allow them time for a new release.
-  // PJRT v54 released Jun 10, so most plugins should use WEEK_12 by default.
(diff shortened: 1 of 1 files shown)


GleasonK:
The default version needs to be 12w, since the existing contract with plugins is that they only need to release quarterly. Bumping to 4w would cause issues.

That said, I believe this shouldn't be an issue now that plugins are communicating their StableHLO version to the framework:
- https://github.com/openxla/xla/commit/84ad5fab88e3979b9c43ce93089e0ef537d14b88

We had to partially roll back the change but in the meantime we're pinning v1.7.0 for new plugins:
- https://github.com/openxla/xla/commit/2f99455cdf99e844ddad17de9f4714997023d243
- https://github.com/openxla/xla/commit/7485cf66159a06dbeb125e0108b28867e60f7e13
 
In ~3w plugins will communicate their exact versions.

GleasonK (changes requested):
adding blocking review, see other comment

PR #18143, opened 2024-10-10
Add math tests for FP8 types

sergey-kozub:
This PR also fixes the issues in `IsPosInf`, `IsNegInf`, `IsNegZero` implementations discovered by the tests.

### DIFF
--- a/xla/hlo/builder/lib/math.cc
+++ b/xla/hlo/builder/lib/math.cc
@@ -123,6 +123,10 @@ static absl::Status EnsureOperandIsRealFp(absl::string_view op_name,
   return absl::OkStatus();
 }
 
+XlaOp PredFalse(XlaBuilder* builder, const Shape& shape) {
+  return Broadcast(ConstantR0<bool>(builder, false), shape.dimensions());
+}
+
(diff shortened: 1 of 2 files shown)


xla/hlo/builder/lib/math.cc
```diff
@@ -123,14 +124,21 @@ static absl::Status EnsureOperandIsRealFp(absl::string_view op_name,
   return absl::OkStatus();
 }
 
+XlaOp PredFalse(XlaBuilder* builder, const Shape& shape) {
+  Shape pred_shape = ShapeUtil::ChangeElementType(shape, PRED);
+  return ConstantLiteral(builder, Literal::CreateFromShape(pred_shape));
```
akuegel, line 129: Maybe instead use:

```
return Broadcast(ConstantR0WithType(builder, PRED, false));
```

Using broadcast of a scalar constant is preferred, and making the init value of the constant literal explicit is easier to understand.

sergey-kozub: Done, thanks.

akuegel (approved):

PR #17507, opened 2024-09-23
[ROCm] Fix build break due to 1c21b0bba

hsharsha:

PR #18145, opened 2024-10-10
[ROCm] Fix build issue caused by unnecessary hip_runtime.h include

mmakevic-amd:
Issue was introduced here: https://github.com/openxla/xla/commit/7e2ff2b1405155d7d6db6bdbca14b09165c122db

### DIFF
--- a/xla/stream_executor/rocm/BUILD
+++ b/xla/stream_executor/rocm/BUILD
@@ -246,6 +246,7 @@ cc_library(
     ]),
     visibility = ["//visibility:public"],
     deps = [
+        ":rocm_driver",
         "//xla/stream_executor:launch_dim",
         "//xla/stream_executor/gpu:gpu_driver_header",
         "//xla/stream_executor/gpu:gpu_executor_header",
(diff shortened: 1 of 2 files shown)


beckerhe:
Hey, this seems to be fine on the CI?

Can you share the exact error that you're seeing?

I'm a bit reluctant to approve that change because I don't think the include is unecessary. The header references `hipFunction_t` therefore it should include the API header declaring that type.

beckerhe:
Is this rocm_driver_wrapper.h related?

I've just submitted https://github.com/openxla/xla/commit/26c6f1fe46dc34a41285fe95da49093834c9d985

I'm wondering if this also fixes your issue?

mmakevic-amd:
Yes defining `__HIP_DISABLE_CPP_FUNCTIONS__` before inclusion of `hip_runtime.h` is also alright.  I opted for using the wrapper to stay consistent with recent build fixes I found (e.g. https://github.com/openxla/xla/pull/17507)

beckerhe:
> Yes defining `__HIP_DISABLE_CPP_FUNCTIONS__` before inclusion of `hip_runtime.h` is also alright. I opted for using the wrapper to stay consistent with recent build fixes I found (e.g. #17507)

Gotcha.

Since we try to follow the include-what-you-use scheme I would like not to remove the include in this case.

Since this is fixed due to my other change as well, can we close this or do you still see a strong reason to merge this?

mmakevic-amd:
> Since we try to follow the include-what-you-use scheme I would like not to remove the include in this case.
> Since this is fixed due to my other change as well, can we close this or do you still see a strong reason to merge this?

Sure thing, I'll close it since it is not necessary anymore

PR #18170, opened 2024-10-10
Cleanup hlo_extractor and hlo_bisect dependecies

shraiysh:
//xla/tools:hlo_extractor and //xla/tools/hlo_bisect:hlo_bisect_state depend on //xla/tests:test_utils. This dependency should not exist. Moving the relevant functions to //xla:literal_util.

### DIFF
--- a/xla/tools/hlo_extractor.cc
+++ b/xla/tools/hlo_extractor.cc
@@ -42,7 +42,6 @@ limitations under the License.
 #include "xla/service/hlo_verifier.h"
 #include "xla/shape.h"
 #include "xla/shape_util.h"
-#include "xla/tests/test_utils.h"
 #include "tsl/platform/status.h"
 
 namespace xla {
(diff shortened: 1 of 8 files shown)


xla/tools/hlo_extractor.cc
```diff
@@ -42,7 +42,6 @@ limitations under the License.
 #include "xla/service/hlo_verifier.h"
 #include "xla/shape.h"
 #include "xla/shape_util.h"
-#include "xla/tests/test_utils.h"
```
akuegel, line 45: Please add xla/literal_util.h include instead. Also the corresponding BUILD dependency.

akuegel (approved):
Thanks

xla/tools/hlo_extractor.cc
shraiysh: That is already included in line number 38 and the dependency also exists already.

shraiysh: It is already added in line 39. And the dependency also exists.

akuegel: Ah, sorry, I should have expanded the view to see it.

akuegel (approved):

PR #18213, opened 2024-10-11
[NFC] Deduplicate default ShapeSizeBytesFunction and kPointerSize.

sergachev:


xla/service/gpu/model/gpu_indexing_performance_model_test.cc
```diff
@@ -713,10 +706,7 @@ class FlopsPerElementTest : public GpuIndexingPerformanceModelTest {
                             ParseAndReturnVerifiedModule(hlo_module_string));
 
     GpuHloCostAnalysis cost_analysis(
-        GpuHloCostAnalysis::Options{ShapeSizeBytesFunction(),
-                                    /*per_second_rates=*/{},
-                                    /*min_latencies_seconds=*/{},
-                                    /*count_multiple_input_accesses=*/true},
+        GpuHloCostAnalysis::Options{.count_multiple_input_accesses = true},
```
bchetioui, line 709: I'm surprised this builds in OSS---this is a C++20 feature, right? I expect that this might break things as we pull it in, afaik we're not set up for C++20 in XLA yet.

I'll stamp it and try to import it, but I think you may need to update this to use the previous style (and same elsewhere).

bchetioui (approved):
Thanks for the cleanup Ilia, let me see if that merges correctly!

xla/service/gpu/model/gpu_indexing_performance_model_test.cc
sergachev: Seems to work already: https://github.com/openxla/xla/blob/main/xla/service/algebraic_simplifier_test.cc#L6013-L6020

PR #18254, opened 2024-10-12
[XLA:GPU] Simplifies All-Reduce if they can be simplified or are degenerated.

yliu120:
This PR adds back the AllReduceSimplifier pass to the GPU compiler as in some cases we see JAX programs will add some degenerated psum to the HLOs,

Code pointer in JAX:
1) https://cs.opensource.google/jax/jax/+/main:jax/experimental/shard_map.py;drc=875f44c63a6c0f082ef3b228559e86e1d0a398d1;l=1674
2) pbroadcast to psum conversions. 

The compiler should generally remove these unnecessary ARs so that it unblocks many optimizations like collective pipeliner. It also emits unncessary runtime thunks as well.

### DIFF
--- a/xla/service/gpu/tests/gpu_compiler_e2e_test.cc
+++ b/xla/service/gpu/tests/gpu_compiler_e2e_test.cc
@@ -1,4 +1,4 @@
-/* Copyright 2023 The OpenXLA Authors.
+/* Copyright 2024 The OpenXLA Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -14,6 +14,8 @@ limitations under the License.
 ==============================================================================*/
 
 #include "absl/strings/string_view.h"
+#include "xla/service/pattern_matcher.h"
+#include "xla/service/pattern_matcher_gmock.h"
 #include "xla/tests/hlo_test_base.h"
 #include "xla/tsl/lib/core/status_test_util.h"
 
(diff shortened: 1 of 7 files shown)


golechwierowicz:
I don't know the context. When was this pass removed?

yliu120:
> I don't know the context. When was this pass removed?

I don't know as well... I think it should be there. At least it doesn't hurt anything.

yliu120:
I was just fixing this issue so I don't check about the historical context.

golechwierowicz:
Can you paste in a simple runnable JAX program which demonstrates the speedup? What speedups do you observe on such example and on which hardware?

ezhulenev (approved):

ezhulenev:
It breaks bunch of internal tests, will try to land it tomorrow

yliu120:
> It breaks bunch of internal tests, will try to land it tomorrow

OK. Thanks. I cannot see what is broken.. I guess something internally are not updated with this change.

ezhulenev:
It's the constructor that got deleted

yliu120:
> It's the constructor that got deleted

Yes. I removed that constructor because it does nothing. The field there is dead code and just confusing.

xla/service/gpu/tests/gpu_compiler_e2e_test.cc
golechwierowicz, line 1: There's already a test file which captures gpu e2e pipeline and with collective ops testing in mind.

Can you move the test there: https://github.com/openxla/xla/blob/b527ed56098b3b1a7df99102c7e7c09d5f51c237/xla/tests/collective_ops_e2e_test.cc?

golechwierowicz: Synced internally with @ezhulenev. We will rewrite the test internally.

yliu120: Thank you both.

golechwierowicz (approved):

golechwierowicz:
This is in https://github.com/openxla/xla/commit/23a255e065405ca12cde2824c09bf1fc7e4bfb0c.

PR #18287, opened 2024-10-14
Pass the compile options for deserialization via PJRT C API

### LINKED ISSUE #18286: Deserialization of executables fails on non-zero ranks when deserializing single-device executable
If we deserialize executables via the wrapper C API client, [the compile options are ignored](https://github.com/openxla/xla/blob/b4abe20e7301e98abf641a7d65e0226c4247242b/xla/pjrt/pjrt_c_api_client.cc#L408). In practice, this means that JAX compilation cache fails when deserializing executables for rank zero on non-zero ranks.

JAX repro:

```python
import jax
import jax.numpy as jnp
import logging
import argparse
import socket

parser = argparse.ArgumentParser(
                    prog='mock-test',
                    description='Tests mocking',
                    epilog='...')
parser.add_argument('-r', '--rank', type=int)
args = parser.parse_args()

if args.rank == 1:
  logging.basicConfig(format='%(asctime)s %(message)s')
  logging.getLogger("jax._src.compiler").setLevel(logging.DEBUG)

jax.config.update("jax_compilation_cache_dir", "/tmp/compilation_cache")
jax.config.update("jax_persistent_cache_min_compile_time_secs", 0)

jax.distributed.initialize(socket.gethostname() + ":1234", 2, args.rank, local_device_ids = [args.rank])

f = jax.jit(lambda x: x)
jax.block_until_ready(f(jnp.zeros((1,))))
```

Run `(python test.py -r0 &) && (python test.py -r1)` twice on a machine with 2+ GPUs.

The output (from rank 1) then contains the following error message:
```
/opt/jax/jax/_src/compiler.py:691: UserWarning: Error reading persistent compilation cache entry for 'jit__lambda_':
XlaRuntimeError: INVALID_ARGUMENT: Device assignment (Computations: 1 Replicas: 1
Computation 0: 0
) does not have any local devices.
  warnings.warn(
2024-10-14 09:16:30,222 PERSISTENT COMPILATION CACHE MISS for 'jit__lambda_' with key 'jit__lambda_-502ff86f0064419e429f73e9641f94cc3ab91a275910dec17b3ba6186556a297'
```

jaro-sevcik:
This enables JAX to supply different device assignment when deserializing single-device executables from compilation cache.

Fixes https://github.com/openxla/xla/issues/18286.

### DIFF
--- a/xla/pjrt/c/pjrt_c_api.h
+++ b/xla/pjrt/c/pjrt_c_api.h
@@ -1580,8 +1580,8 @@ struct PJRT_Executable_DeserializeAndLoad_Args {
   // Serialized CompileOptionsProto or null (to use the options
   // from the serialized executable).
   // (https://github.com/openxla/xla/blob/main/xla/pjrt/compile_options.proto)
-  const char* compile_options;
-  size_t compile_options_size;
+  const char* overridden_serialized_compile_options;
+  size_t overridden_serialized_compile_options_size;
 };
 PJRT_DEFINE_STRUCT_TRAITS(PJRT_Executable_DeserializeAndLoad_Args,
                           loaded_executable);
--- a/xla/pjrt/c/pjrt_c_api.h
+++ b/xla/pjrt/c/pjrt_c_api.h
@@ -1576,6 +1576,8 @@ struct PJRT_Executable_DeserializeAndLoad_Args {
   PJRT_Client* client;
   const char* serialized_executable;
   size_t serialized_executable_size;
+  const char* compile_options;
+  size_t compile_options_size;
   PJRT_LoadedExecutable* loaded_executable;  // out
 };
 PJRT_DEFINE_STRUCT_TRAITS(PJRT_Executable_DeserializeAndLoad_Args,
--- a/xla/pjrt/c/pjrt_c_api.h
+++ b/xla/pjrt/c/pjrt_c_api.h
@@ -1576,9 +1576,12 @@ struct PJRT_Executable_DeserializeAndLoad_Args {
   PJRT_Client* client;
   const char* serialized_executable;
   size_t serialized_executable_size;
+  PJRT_LoadedExecutable* loaded_executable;  // out
+  // Serialized CompileOptionsProto or null (to use the options
+  // from the serialized executable).
+  // (https://github.com/openxla/xla/blob/main/xla/pjrt/compile_options.proto)
   const char* compile_options;
   size_t compile_options_size;
-  PJRT_LoadedExecutable* loaded_executable;  // out
 };
 PJRT_DEFINE_STRUCT_TRAITS(PJRT_Executable_DeserializeAndLoad_Args,
                           loaded_executable);
--- a/xla/pjrt/c/pjrt_c_api.h
+++ b/xla/pjrt/c/pjrt_c_api.h
@@ -79,7 +79,7 @@ PJRT_DEFINE_STRUCT_TRAITS(PJRT_Extension_Base, next);
 // Changes include:
 // * Adding a new field to the PJRT_Api or argument structs
 // * Renaming a method or argument (doesn't affect ABI)
-#define PJRT_API_MINOR 55
+#define PJRT_API_MINOR 56
 
 // The plugin should set the major_version and minor_version of
 // PJRT_Api.pjrt_api_version to be the `PJRT_API_MAJOR` and `PJRT_API_MINOR` in
--- a/xla/pjrt/c/pjrt_c_api_wrapper_impl.cc
+++ b/xla/pjrt/c/pjrt_c_api_wrapper_impl.cc
@@ -1570,9 +1570,17 @@ PJRT_Error* PJRT_Executable_DeserializeAndLoad(
   absl::string_view serialized(args->serialized_executable,
                                args->serialized_executable_size);
 
-  PJRT_ASSIGN_OR_RETURN(std::unique_ptr<xla::PjRtLoadedExecutable> executable,
-                        args->client->client->DeserializeExecutable(
-                            serialized, /*options=*/std::nullopt));
+  std::optional<xla::CompileOptions> options;
+
+  if (args->compile_options && args->compile_options_size > 0) {
+    PJRT_ASSIGN_OR_RETURN(
+        options, ParseCompileOptions(absl::string_view(
+                     args->compile_options, args->compile_options_size)));
+  }
+
+  PJRT_ASSIGN_OR_RETURN(
+      std::unique_ptr<xla::PjRtLoadedExecutable> executable,
+      args->client->client->DeserializeExecutable(serialized, options));
 
   args->loaded_executable =
       new PJRT_LoadedExecutable(std::move(executable), args->client);
--- a/xla/pjrt/c/pjrt_c_api_wrapper_impl.cc
+++ b/xla/pjrt/c/pjrt_c_api_wrapper_impl.cc
@@ -1570,17 +1570,20 @@ PJRT_Error* PJRT_Executable_DeserializeAndLoad(
   absl::string_view serialized(args->serialized_executable,
                                args->serialized_executable_size);
 
-  std::optional<xla::CompileOptions> options;
+  std::optional<xla::CompileOptions> overriden_options;
 
-  if (args->compile_options && args->compile_options_size > 0) {
+  if (args->overridden_serialized_compile_options &&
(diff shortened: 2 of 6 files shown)


xla/pjrt/c/pjrt_c_api.h
```diff
@@ -1576,6 +1576,8 @@ struct PJRT_Executable_DeserializeAndLoad_Args {
   PJRT_Client* client;
   const char* serialized_executable;
   size_t serialized_executable_size;
+  const char* compile_options;
```
changm, line 1579: Is this the same compile options proto as in Compile()? 

https://github.com/openxla/xla/blob/b46d80b8b651ca230d5b797a128dbd3dbe30de5c/xla/pjrt/c/pjrt_c_api.h#L2106

If so, mind adding a comment explicitly stating it's the same?

xla/pjrt/c/pjrt_c_api.h
```diff
@@ -1576,6 +1576,8 @@ struct PJRT_Executable_DeserializeAndLoad_Args {
...
+  size_t compile_options_size;
```
changm, line 1580: Also, due to ABI compatibility, I think these two arguments have to go at the end of the struct. This part is fuzzy to me due to the last arg here being the output.

@skye - Can you confirm here please?

xla/pjrt/c/pjrt_c_api_wrapper_impl.cc
```diff
@@ -1570,9 +1570,17 @@ PJRT_Error* PJRT_Executable_DeserializeAndLoad(
   absl::string_view serialized(args->serialized_executable,
                                args->serialized_executable_size);
 
-  PJRT_ASSIGN_OR_RETURN(std::unique_ptr<xla::PjRtLoadedExecutable> executable,
-                        args->client->client->DeserializeExecutable(
-                            serialized, /*options=*/std::nullopt));
+  std::optional<xla::CompileOptions> options;
```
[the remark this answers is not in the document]
jaro-sevcik: I believe 'nullopt' means that the options will be taken from the serialized executable (i.e., exactly the options that were used for compilation).

xla/pjrt/c/pjrt_c_api.h
jaro-sevcik: Done.

xla/pjrt/c/pjrt_c_api.h
jaro-sevcik: Reordered.

jaro-sevcik:
@changm Addressed the comments. Could you take another look, please?

xla/pjrt/c/pjrt_c_api_wrapper_impl.cc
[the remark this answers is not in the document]
changm: Ahh let me try again :)

I don't think you need to make this optional in line 1573 and check for null in line 1575. Just always declare an instance of options and ParseCompileOptions. ParseCompileOptions should handle a nullptr just fine.

Does that work? Thanks!

changm (approved):
LGTM assuming last comment that you can fix up please :). Thank you!

xla/pjrt/c/pjrt_c_api_wrapper_impl.cc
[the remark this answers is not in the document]
jaro-sevcik: `ParseCompiledOptions` does not seem to handle `nullptr`. It does not even return optional, it just returns parsed options or a (generic) error on parse failure. Are you suggesting to turn parse failures into default compile options?

The `nullopt` value is semantically different from default compile options, so we should not mix them up.

Does this make sense or am I missing something?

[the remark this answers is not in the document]
jaro-sevcik: To answer your original question ("Any particular reason it's optional here but not during Compile()?"):

The `optional` here allows to optionally override the compile options that are already serialized in the `serialized` argument.

The `nullopt` options for the `DeserializeExecutable` method means that the runtime should use the serialized compile options from the `serialized` argument to set some of the properties of the in-memory executable object (such as device assignment).

The non-nullopt options for the `DeserializeExecutable` method override the serialized compile options and when setting the properties of the in-memory executable object (device assignment).

See here: https://github.com/openxla/xla/blob/eef7ee50d0980848436f0b4f402cec8c5bf86f21/xla/pjrt/pjrt_stream_executor_client.cc#L3667

[ The options for the `Compile` method do not optionally override anything, they are always used to set the properties of the executable object - there is no need for the `nullopt` value. ]

jaro-sevcik:
I am still unsure what you had in mind in your last comment. I have added some additional explanation on the comment.

xla/pjrt/c/pjrt_c_api.h
```diff
@@ -1577,6 +1577,11 @@ struct PJRT_Executable_DeserializeAndLoad_Args {
   const char* serialized_executable;
   size_t serialized_executable_size;
   PJRT_LoadedExecutable* loaded_executable;  // out
+  // Serialized CompileOptionsProto or null (to use the options
+  // from the serialized executable).
+  // (https://github.com/openxla/xla/blob/main/xla/pjrt/compile_options.proto)
+  const char* compile_options;
```
changm, line 1583: nit: Rename this to overridden_serialized_compile_options here and everywhere else

xla/pjrt/pjrt_c_api_client_test.cc
```diff
@@ -163,5 +163,41 @@ TEST(PjRtClientTest, CreateViewAndCopyToDeviceAsyncExternalCpuOnly) {
...
+  TF_ASSERT_OK_AND_ASSIGN(std::string serialized_executable,
+                          executable->SerializeExecutable());
+
+  // Deserialize the executable for device 1.
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto deserialized_executable,
+      client->DeserializeExecutable(serialized_executable,
+                                    compile_options_for_device(1)));
+
+  // Check that the executable was indeed serialized for device 1.
```
changm, line 197: nit: "Check that the executable's compile options were overrideen with device id 1"

xla/pjrt/c/pjrt_c_api.h
jaro-sevcik: Done.

xla/pjrt/pjrt_c_api_client_test.cc
jaro-sevcik: Done.

jaro-sevcik:
Addressed the comments, could you take another look?

changm (approved):

changm:
Thanks LGTM!

changm:
Not sure it's too late, but I also think you have to bump up the minor version number here since you changed the API - https://github.com/openxla/xla/blob/9f24c04c94c3c65c70ed9eec6275cc8c6a963eb6/xla/pjrt/c/pjrt_c_api.h#L82

changm (approved):

nouiz:
Any idea what is blocking the PR as it is approved? We don't have visibility on the issue.

changm:
Hmm sorry for the annoying things. Looks like you have to update the CHANGELOG here - [xla/pjrt/c/CHANGELOG.md](https://github.com/openxla/xla/blob/975f007cfc6045ec692d74ef2b5b06d3384abd2a/xla/pjrt/c/CHANGELOG.md#L4)

changm:
I think this PR still needs to be updated to add the new version in changelog file. After that @golechwierowicz should be shepherding this through.

changm (approved):

golechwierowicz (approved):

changm (approved):

changm:
Hmm, I think everything went through but not sure why this PR still is open. @golechwierowicz Any idea or whose a reviewer with write access?

golechwierowicz:
It's in :)

loislo:
it seems the pr was the reason for a failure in one of the prod pipeline. It was rolled back.

jaro-sevcik:
> it seems the pr was the reason for a failure in one of the prod pipeline. It was rolled back.

@loislo Could you share more details on what exactly has failed?

Also, could you please reopen https://github.com/openxla/xla/issues/18286? Since you reverted the fix, JAX compilation caching is still broken on multi-node configs.

loislo:
done.

loislo:
the deserialization process was failing in some cases.
The observed behavior was:
the size of the override was nonzero and big.
the proto parsing was failing.

PR #18331, opened 2024-10-15
[XLA:CPU] upgrading onednn version to 3.6

ashiqimranintel:
This PR upgrades oneDNN version from v3.5 to v3.6, this PR has been tested on several models across different platforms including cascade-lake, sapphire-rapids, and granite-rapids

Several bug fixes have been resolved in this version. Details can be found here https://github.com/oneapi-src/oneDNN/releases


third_party/tsl/third_party/mkl_dnn/mkldnn_v1.BUILD
```diff
@@ -6,12 +6,15 @@ exports_files(["LICENSE"])
 
 _CMAKE_COMMON_LIST = {
     "#cmakedefine DNNL_GPU_RUNTIME DNNL_RUNTIME_${DNNL_GPU_RUNTIME}": "#define DNNL_GPU_RUNTIME DNNL_RUNTIME_NONE",
+    "#cmakedefine DNNL_GPU_VENDOR DNNL_VENDOR_${DNNL_GPU_VENDOR}": "#define DNNL_VENDOR_INTEL",
```
penpornk, line 9: Intel GPU changes are in the PJRT plug-in. Do we need this configuration here?

penpornk (commented):
Thank you for the PR!

third_party/tsl/third_party/mkl_dnn/mkldnn_v1.BUILD
```diff
@@ -186,6 +199,7 @@ cc_library(
             "src/cpu/aarch64/**",
             "src/cpu/rv64/**",
             "src/cpu/x64/gemm/**/*_kern_autogen.cpp",
+            "src/cpu/sycl/**",
```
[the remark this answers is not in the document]
dzarukin: Hi @penpornk, oneDNN split SYCL and OCL backend details over folders and CPU got its part here. It contains handling sycl objects abstractions in a stream and an engine objects as well as submission of primitives into a queue. It doesn't affect any implementation codes that used to be there before. If the build is not using DNNL_CPU_BACKEND=SYCL (which is not the case the last time I checked), there's no sense in adding those files, that's why they got excluded in here.

Once the team would like to start building CPU backend for SYCL, those would be needed and extra dependency would be added.
Thanks for checking.

third_party/tsl/third_party/mkl_dnn/mkldnn_v1.BUILD
penpornk: I mean, we can set it to `DNNL_VENDOR_NONE`, right?

third_party/tsl/third_party/mkl_dnn/mkldnn_v1.BUILD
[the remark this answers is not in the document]
penpornk: Thank you for the quick reply, @dzarukin!
The `exclude = [` part was hidden and I thought this was in `src = [` (include) list. My bad! 
I don't think we need SYCL sources as well, that's why I asked whether we really used it.

third_party/tsl/third_party/mkl_dnn/mkldnn_v1.BUILD
dzarukin: That should work as well, at least I don't anticipate any issues why it shouldn't.

ashiqimranintel: Yes, you can set it to DNNL_VENDOR_NONE, do you want me to change it?

penpornk: Yes, please. Thank you! :)

penpornk (changes requested):

ashiqimranintel:
@penpornk , it's done. please review it again. Thanks

ashiqimranintel:
@penpornk , FYI, https://github.com/tensorflow/tensorflow/pull/77927 PR is approved on the TF side.

penpornk (approved):
Thank you again for the PR!

> @penpornk , FYI, [tensorflow/tensorflow#77927](https://github.com/tensorflow/tensorflow/pull/77927) PR is approved on the TF side.

I have requested the `DNNL_VENDOR_NONE` change there too, although both this and that PR will go to the same source of truth anyway.

penpornk:
The upgrade caused TF CI failures, e.g., [Linux x86 CPU CI](https://btx.cloud.google.com/invocations/43a4dd35-900b-4a65-b751-35bd27d8013a/targets). I have also commented on https://github.com/tensorflow/tensorflow/pull/77927.

penpornk:
Since this overlaps with https://github.com/tensorflow/tensorflow/pull/77927 and we are resolving TF test failures, I think we can just proceed on that PR alone. The changes there will also affect XLA when the TF PR is merged. I'll close this PR.

PR #18389, opened 2024-10-16
[PJRT:GPU] Fix pinned memory space ids

jaro-sevcik:
The memory space ids should be unique.

Duplicated memory space ids cause assertion failures if we try to use GPU client via IFRT proxy server. See https://github.com/openxla/xla/blob/3f6598f7d9d253b364a3316c8a532f7addebc207/xla/python/ifrt_proxy/server/ifrt_backend.cc#L314


cheshire:
Could you expand PR description to note what bugs the current behavior is causing?

cheshire:
Sorry I've meant commit description, so we have it in history.

jaro-sevcik:
> Sorry I've meant commit description, so we have it in history.

The first comment is the commit description, no? (I have updated that one.)

cheshire (approved):

PR #18404, opened 2024-10-16
[XLA:Python] Release GIL around CopyToHostBuffer in CopyToHostAsync

kevinji:
`CopyToHostBuffer` can cause GIL contention, so release the GIL before running the function. `value_.mutable_data()` points to the data buffer of the Numpy array, so I believe it is safe to drop the GIL here.

The GIL used to be dropped in `CopyToHostAsync` but was removed in #10017.


kevinji:
@hawkinsp anything I can do to help with reviewing this PR?

hawkinsp (approved):

PR #18407, opened 2024-10-16
Fix xla-mlir failures on Windows

mraunak:
This PR aims to enable the XLA/mlir/tool test cases on the Windows Platform.

Error:
//xla/mlir/tools/mlir_bisect/... tests were failing on the Windows platform with the errors shown below:

Errors
Error 1.Error with llvm::seq
no matching function for call to 'seq'
for (auto i : llvm::seq(0ul, sizeof...(T))) {
Solution: change to llvm::seq(0, sizeof...(T))
By explicitly specifying the type (unsigned long) in llvm::seq, the compiler now clearly understands the type of the sequence.

Error 2. Missing dlfcn.h:
Location: xla/mlir/tools/mlir_interpreter/dialects/func.cc
fatal error: 'dlfcn.h' file not found
Solution: include 'windows.h' for Windows platform
Error 3.
Use of Undeclared Identifiers sym and RTLD_DEFAULT:
Location: xla/mlir/tools/mlir_interpreter/dialects/func.cc
use of undeclared identifier 'sym'
sym = dlsym(RTLD_DEFAULT, callee.getSymName().str().c_str());
^
use of undeclared identifier 'RTLD_DEFAULT'
Solution:
On Windows, the approach to obtaining a symbol's address differs from Unix-based systems.
GetModuleHandle function retrieves a handle to the specified module (DLL) that is loaded in the address space of the calling process. This handle is necessary to access the module's symbols.
GetProcAddress function locates the address of an exported function or variable by name.

### DIFF
--- a/xla/mlir/tools/mlir_interpreter/dialects/BUILD
+++ b/xla/mlir/tools/mlir_interpreter/dialects/BUILD
@@ -22,8 +22,8 @@ cc_library(
         ],
         exclude = ["util.cc"],
     ),
+    hdrs= ["symbol_finder.h"],
     deps = [
-        ":find_symbol",
         ":dialect_utils",
         "//xla/mlir/tools/mlir_interpreter/framework",
         "//xla/mlir_hlo",
@@ -65,15 +65,15 @@ cc_library(
 )
 
 cc_library(
-  name = "find_symbol_posix",
-  srcs = ["symbol_finder_windows.cc"],
+  name = "find_symbol_linux",
+  srcs = ["symbol_finder_linux.cc"],
   deps = ["@com_google_absl//absl/status:statusor"],
   tags = ["manual"]
 )
 
 cc_library(
   name = "find_symbol_windows",
-  srcs = ["symbol_finder_linux.cc"],
+  srcs = ["symbol_finder_windows.cc"],
   deps = ["@com_google_absl//absl/status:statusor"],
   tags = ["manual"]
 )
@@ -81,6 +81,6 @@ cc_library(
 cc_library(
   name = "find_symbol",
   hdrs = ["symbol_finder.h"],
-  deps = if_windows([":find_symbol_windows"], [":find_symbol_posix"])
+  deps = if_windows([":find_symbol_windows"], [":find_symbol_linux"])
 )
 
--- a/xla/mlir/tools/mlir_interpreter/dialects/BUILD
+++ b/xla/mlir/tools/mlir_interpreter/dialects/BUILD
@@ -21,6 +21,7 @@ cc_library(
         ],
         exclude = ["util.cc"],
     ),
+    hdrs=["symbol_finder.h"],
     deps = [
         ":dialect_utils",
         "//xla/mlir/tools/mlir_interpreter/framework",
--- a/xla/mlir/tools/mlir_interpreter/dialects/BUILD
+++ b/xla/mlir/tools/mlir_interpreter/dialects/BUILD
@@ -76,31 +76,25 @@ cc_library(
 )
 
 cc_library(
-    name = "symbol_finder_header",
-    hdrs = ["symbol_finder.h"],
-)
-
-cc_library(
     name = "symbol_finder_linux",
-    srcs = ["symbol_finder_linux.cc"],
+    srcs = ["symbol_finder.h","symbol_finder_linux.cc"],
     deps = [
-        "@com_google_absl//absl/status:statusor",
-        ":symbol_finder_header",
+        "@com_google_absl//absl/status:statusor"
     ],
     tags = ["manual"],
 )
 
 cc_library(
     name = "symbol_finder_windows",
-    srcs = ["symbol_finder_windows.cc"],
+    srcs = ["symbol_finder.h","symbol_finder_windows.cc"],
     deps = [
         "@com_google_absl//absl/status:statusor",
-        ":symbol_finder_header",
     ],
     tags = ["manual"],
 )
 
 cc_library(
     name = "find_symbol",
+    hdrs = ["symbol_finder.h"],
     deps = if_windows([":symbol_finder_windows"], [":symbol_finder_linux"]),
 )
--- a/xla/mlir/tools/mlir_interpreter/dialects/BUILD
+++ b/xla/mlir/tools/mlir_interpreter/dialects/BUILD
@@ -1,4 +1,5 @@
 load("@tsl//tsl/platform:rules_cc.bzl", "cc_library")
+load("//xla/tsl:tsl.bzl", "if_windows")
 
 package(
     # copybara:uncomment default_applicable_licenses = ["//tensorflow:license"],
@@ -21,8 +22,8 @@ cc_library(
         ],
         exclude = ["util.cc"],
     ),
-    hdrs=["symbol_finder.h"],
     deps = [
+        ":find_symbol",
         ":dialect_utils",
         "//xla/mlir/tools/mlir_interpreter/framework",
         "//xla/mlir_hlo",
@@ -62,3 +63,24 @@ cc_library(
         "@llvm-project//mlir:ViewLikeInterface",
     ],
 )
+
+cc_library(
+  name = "find_symbol_posix",
+  srcs = ["symbol_finder_windows.cc"],
+  deps = ["@com_google_absl//absl/status:statusor"],
+  tags = ["manual"]
+)
+
+cc_library(
+  name = "find_symbol_windows",
+  srcs = ["symbol_finder_linux.cc"],
+  deps = ["@com_google_absl//absl/status:statusor"],
+  tags = ["manual"]
+)
+
+cc_library(
+  name = "find_symbol",
+  hdrs = ["symbol_finder.h"],
+  deps = if_windows([":find_symbol_windows"], [":find_symbol_posix"])
+)
+
--- a/xla/mlir/tools/mlir_interpreter/dialects/BUILD
+++ b/xla/mlir/tools/mlir_interpreter/dialects/BUILD
@@ -64,7 +64,6 @@ cc_library(
     ],
 )
 
-# New target for the header file
 cc_library(
     name = "symbol_finder_header",
     hdrs = ["symbol_finder.h"],
@@ -75,7 +74,7 @@ cc_library(
     srcs = ["symbol_finder_linux.cc"],
     deps = [
         "@com_google_absl//absl/status:statusor",
-        ":symbol_finder_header",  # Depend on the header-only target
+        ":symbol_finder_header", 
     ],
     tags = ["manual"]
 )
@@ -85,7 +84,7 @@ cc_library(
     srcs = ["symbol_finder_windows.cc"],
     deps = [
         "@com_google_absl//absl/status:statusor",
-        ":symbol_finder_header",  # Depend on the header-only target
+        ":symbol_finder_header", 
     ],
     tags = ["manual"]
 )
@@ -94,4 +93,3 @@ cc_library(
     name = "find_symbol",
     deps = if_windows([":symbol_finder_windows"], [":symbol_finder_linux"])
 )
-
--- a/xla/mlir/tools/mlir_interpreter/dialects/BUILD
+++ b/xla/mlir/tools/mlir_interpreter/dialects/BUILD
@@ -74,9 +74,9 @@ cc_library(
     srcs = ["symbol_finder_linux.cc"],
     deps = [
         "@com_google_absl//absl/status:statusor",
-        ":symbol_finder_header", 
+        ":symbol_finder_header",
     ],
-    tags = ["manual"]
+    tags = ["manual"],
 )
 
 cc_library(
@@ -84,12 +84,12 @@ cc_library(
     srcs = ["symbol_finder_windows.cc"],
     deps = [
         "@com_google_absl//absl/status:statusor",
-        ":symbol_finder_header", 
+        ":symbol_finder_header",
     ],
-    tags = ["manual"]
+    tags = ["manual"],
 )
 
 cc_library(
     name = "find_symbol",
-    deps = if_windows([":symbol_finder_windows"], [":symbol_finder_linux"])
+    deps = if_windows([":symbol_finder_windows"], [":symbol_finder_linux"]),
 )
--- a/xla/mlir/tools/mlir_interpreter/dialects/BUILD
+++ b/xla/mlir/tools/mlir_interpreter/dialects/BUILD
@@ -22,8 +22,8 @@ cc_library(
         ],
         exclude = ["util.cc"],
     ),
-    hdrs= ["symbol_finder.h"],
     deps = [
+        ":find_symbol",
         ":dialect_utils",
         "//xla/mlir/tools/mlir_interpreter/framework",
         "//xla/mlir_hlo",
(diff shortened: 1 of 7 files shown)


xla/mlir/tools/mlir_interpreter/dialects/comparators.h
```diff
@@ -43,14 +44,21 @@ struct FloatCompare : CwiseAll {
...
+  // Handling complex types
   template <typename T>
   static bool isnan(std::complex<T> a) {
     return std::isnan(std::real(a)) || std::isnan(std::imag(a));
   }
+  // Template for isnan (Floating numbers)
+  template <typename T>
+  static bool isnan(std::is_floating_point<T> a) {
+    return std::isnan(a);
+  }
```
beckerhe, line 61: What motivates this change?

beckerhe, line 61: I don't see how this should work. `std::is_floating_point<T>` is not a placeholder for any floating point type.

(https://gcc.godbolt.org/z/Ej86a8jKK)

xla/mlir/tools/mlir_interpreter/dialects/func.cc
```diff
@@ -95,12 +99,31 @@ llvm::SmallVector<InterpreterValue> Call(MutableArrayRef<InterpreterValue> args,
...
+          state.AddFailure("Callee not found");
+          return {};
+      }
+  }
+  else {
+       state.AddFailure("Failed to get module handle");
+       return {};
+  }
+  #else
+  sym = dlsym(RTLD_DEFAULT, callee.getSymName().str().c_str());
```
beckerhe, line 121: Can you refactor this logic into a separate function? Maybe something like `absl::StatusOr<void*> FindSymbolInProcess(const std::string& symbol_name)` and then split the logic into a separate files for Windows and Posix?

I imagine build rules like this:

```
cc_library(
  name = "find_symbol_posix",
  srcs = ["find_symbol_posix.cc"],
  deps = ["//third_party/absl/status:statusor"]
  tags = ["manual"]
)

cc_library(
  name = "find_symbol_windows",
  srcs = ["find_symbol_windows.cc"],
  deps = ["//third_party/absl/status:statusor"]
  tags = ["manual"]
)

cc_library(
  name = "find_symbol",
  hdrs = ["find_symbol.h"],
  deps = if_windows([":find_symbol_windows"], [":find_symbol_posix"])
)
```

beckerhe (commented):
Thank you for your contribution. I've left some comments in the code.

xla/mlir/tools/mlir_interpreter/dialects/comparators.h
mraunak: Hi @beckerhe xla-mlir tests(//xla/mlir/tools/mlir_interpreter/dialects/tests/...) are failing on the Windows with the error below. The detailed error log is atta
[xla-mlir-failure.txt](https://github.com/user-attachments/files/17455936/xla-mlir-failure.txt)
ched
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\corecrt_math.h(419,16): error: call to 'fpclassify' is ambiguous
  419 |         return fpclassify(_X) == FP_NAN;
      |                ^~~~~~~~~~
.\xla/mlir/tools/mlir_interpreter/dialects/comparators.h(48,17): note: in instantiation of function template specialization 'isnan<signed char>' requested here
   48 |     return std::isnan(a);
      |                 ^
.\xla/mlir/tools/mlir_interpreter/dialects/comparators.h(34,9): note: in instantiation of function template specialization 'mlir::interpreter::FloatCompare<0, true, false>::isnan<signed char>' requested here
   34 |     if (isnan(a) || isnan(b)) return nan_result;
      |         ^
.\xla/mlir/tools/mlir_interpreter/framework/interpreter_value_util.h(114,19): note: in instantiation of function template specialization 'mlir::interpreter::FloatCompare<0, true, false>::Apply<signed char>' requested here
  114 |       return {Fn::Apply(t, std::get<T>(rhs.storage))};
      |                   ^

beckerhe: Thank you for the full error message. I believe you're hitting a bug in MSVC.

According to the release notes (https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes) it has been fixed in version 17.11 of Visual Studio 2022. (`Missing integral types overload for cmath functions` in top bug fixes)

Is it feasible for you to update MSVC?

If not we can certainly change the code, but I'm reluctant since the code is actually correct in its current form.

mraunak: Hi @beckerhe, yes, I tested on my end, 17.11 works well. Thank you. I will make the suggested changes asap.

mraunak:
Hi @beckerhe, please review the PR and let me know if any further change is required.

beckerhe:
> Hi @beckerhe, please review the PR and let me know if any further change is required.

Looks good in general, but you're still missing the build rules. (See my comment https://github.com/openxla/xla/pull/18407#discussion_r1805956206_ from above for a template).

mraunak:
Hi @beckerhe , I am made changes as per the template, please review

xla/mlir/tools/mlir_interpreter/dialects/BUILD
```diff
@@ -61,3 +63,24 @@ cc_library(
         "@llvm-project//mlir:ViewLikeInterface",
     ],
 )
+
+cc_library(
+  name = "find_symbol_posix",
```
beckerhe, line 68: Can you adjust the names of the targets so that it matches the file names?

Also the windows and linux target will also need to have the header file listed in the `srcs` (!) attribute.

xla/mlir/tools/mlir_interpreter/dialects/func.cc
```diff
@@ -13,7 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-#include <dlfcn.h>
+#include "symbol_finder.h"
```
beckerhe, line 16: This include needs to be relative to the repository root, meaning it should start with `xla/mlir/tools/...`

xla/mlir/tools/mlir_interpreter/dialects/func.cc
```diff
@@ -95,12 +96,14 @@ llvm::SmallVector<InterpreterValue> Call(MutableArrayRef<InterpreterValue> args,
...
-    state.AddFailure("callee not found");
+  
+  // Use the new FindSymbolInProcess function
+  absl::StatusOr<void*> symbol_status = FindSymbolInProcess(callee.getSymName().str());
+  if (!symbol_status.ok()) {
+     state.AddFailure(symbol_status.status().message());
     return {};
-  }
+    }
+    void* sym = *symbol_status;  // Retrieve the symbol pointer
```
beckerhe, line 106: This looks not formatted. Can you run clang-format?

beckerhe (commented):
Hey, I added some comments.

On a related note: I don't think it compiles in the current state. Would be great if you could fix the compile errors first. The compiler is a lot better in giving you timely feedback than I am.

mraunak:
Hi @beckerhe, I will fix it asap, I'm surprised it passed on both Linux and Windows on my end.

beckerhe:
> Hi @beckerhe, I will fix it asap, I'm surprised it passed on both Linux and Windows on my end.

Oh, that's interesting. Maybe this is because Bazel is not doing sandboxing by default. Anyway, thanks for checking. It would still be great if you made the changes I mentioned in the comments, because it wouldn't compile in our internal CI.

mraunak:
I have made the changes, please review, I am not sure why AMD ROCm is failing, the tests failing in the above log is not related to my changes

xla/mlir/tools/mlir_interpreter/dialects/BUILD
```diff
@@ -21,6 +22,7 @@ cc_library(
         ],
         exclude = ["util.cc"],
     ),
+    hdrs= ["symbol_finder.h"],
```
beckerhe, line 25: This shouldn't be here. The `dialects` target should rather depend on `:find_symbol`

xla/mlir/tools/mlir_interpreter/dialects/BUILD
```diff
@@ -61,3 +63,24 @@ cc_library(
         "@llvm-project//mlir:ViewLikeInterface",
     ],
 )
+
+cc_library(
+  name = "find_symbol_linux",
+  srcs = ["symbol_finder_linux.cc"],
```
beckerhe, line 69: Can you still align the names of the target and the .cc file? For example by renaming the target to `symbol_finder_linux`?

xla/mlir/tools/mlir_interpreter/dialects/BUILD
```diff
@@ -61,3 +63,24 @@ cc_library(
...
+  deps = ["@com_google_absl//absl/status:statusor"],
+  tags = ["manual"]
+)
+
+cc_library(
+  name = "find_symbol_windows",
```
beckerhe, line 75: Same here. Can you rename the target to `symbol_finder_windows`?

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc
```diff
@@ -0,0 +1,19 @@
+/* This code defines a POSIX-compatible function, FindSymbolInProcess, 
+   which uses dlsym to locate a function or variable symbol (symbol_name) in the current process;
+   it returns the symbol's address if found or an error if not */
+   
+#ifndef _WIN32
```
beckerhe, line 5: Can you remove this `#ifndef _WIN32`? The whole file is Linux-only code and this is not needed.

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_windows.cc
```diff
@@ -0,0 +1,23 @@
+/* Implemented Function FindSymbolInProcess locate a function or variable
+ symbol in the current process's address space */
+
+#ifdef _WIN32
```
beckerhe, line 4: Same here. Can you remove the ifdef?

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc
mraunak: if I remove ifndef _WIN32, I get the following errors
xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc(7,10): fatal error: 'dlfcn.h' file not found
    7 | #include <dlfcn.h>
      |          ^~~~~~~~~
      
      xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc(14,21): error: use of undeclared identifier 'RTLD_DEFAULT'
   14 |   void* sym = dlsym(RTLD_DEFAULT, symbol_name.c_str());
      |                     ^

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_windows.cc
mraunak: Fails on Linux on my end with an error
xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_windows.cc:5:10: fatal error: windows.h: No such file or directory
    5 | #include <windows.h>
      |          ^~~~~~~~~~~
compilation terminated.

xla/mlir/tools/mlir_interpreter/dialects/BUILD
```diff
@@ -61,3 +63,33 @@ cc_library(
         "@llvm-project//mlir:ViewLikeInterface",
     ],
 )
+
+cc_library(
+    name = "symbol_finder_header",
+    hdrs = ["symbol_finder.h"],
+)
```
beckerhe, line 70: We try not to have these header targets that don't work without an implementation.

Can you do the following instead:

1. Remove this target
2. Add the header file name to all 3(!) cc_library targets below.
3. For `symbol_finder_linux` you add it into `srcs` instead of `hdrs`
4. For `symbol_finder_windows` you do the same - you add it into `srcs` instead of `hdrs`
5. In `find_symbol` you should add just as a normal headers in `hdrs`.

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc
beckerhe: What is your exact bazel test command?

This is supposed to work the following way:

1. Both `symbol_finder_linux` and `symbol_finder_windows` are tagged `manual`. That means they are not getting built by a wildcard build like `bazel build //xla/...`.
2. They are only getting built when you either specify the target explicitly on the command line (i.e. `bazel build //xla/mlir/tools/mlir_interpreter/dialects:symbol_finder_windows`) or when some other target depends on it.
3. `:symbol_finder` should always depend on the right implementation target, so there should be no error coming from there.

beckerhe (commented):
Sorry for the delay. I was out sick the last few days.

mraunak:
Command used:
bazel ^
    --output_user_root=d:\ ^
    --windows_enable_symlinks ^
     test ^
    --nodistinct_host_configuration ^
    --enable_runfiles ^
    --dynamic_mode=off ^
    -k ^
    --config=xla ^
    --config=nonccl ^
    --config=short_logs ^
    --build_tag_filters=-oss_excluded,-no_windows,-oss_serial,-gpu,-no_oss,-requires-gpu-nvidia,-requires-gpu-amd^
    --test_tag_filters=-oss_excluded,-oss_serial,-no_oss,-no_windows,-gpu,-requires-gpu-nvidia,-requires-gpu-amd ^
    --announce_rc ^
    --repo_env=TF_PYTHON_VERSION=3.11 ^
    --config=win_clang_xla ^
    --build_tests_only ^
    --config=monolithic ^
    --test_timeout="300,450,1200,3600" ^
    --copt=/d2ReducedOptimizeHugeFunctions ^
    --host_copt=/d2ReducedOptimizeHugeFunctions ^
    -- //xla/mlir/tools/mlir_interpreter/dialects/tests/...

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc
beckerhe: > If I remove ifndef _WIN32, I get the following errors while running the bazel test on Windows
> 
> xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc(7,10): fatal error: 'dlfcn.h' file not found 7 | #include <dlfcn.h> | ^~~~~~~~~
> 
> ```
>   xla/mlir/tools/mlir_interpreter/dialects/symbol_finder_linux.cc(14,21): error: use of undeclared identifier 'RTLD_DEFAULT'
> ```
> 
> 14 | void* sym = dlsym(RTLD_DEFAULT, symbol_name.c_str()); | ^

Oh I see why this is not working. There is a `glob(*)` in the `:dialects` target.

Can you remove the glob there and list all the files explicitly? Then you should be able to remove the ifdefs.

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder.h
```diff
@@ -0,0 +1,11 @@
+/* This header file declares the FindSymbolInProcess function, which takes a symbol name as 
+input and returns either the symbol's address or an error status, encapsulated in absl::StatusOr<void*> */
+
+#ifndef SYMBOL_FINDER_H_
+#define SYMBOL_FINDER_H_
+
+#include <string>
+#include "absl/status/statusor.h"
+
+absl::StatusOr<void*> FindSymbolInProcess(const std::string& symbol_name);
```
beckerhe, line 10: Can you put this function into the the `mlir::interpreter` namespace?

xla/mlir/tools/mlir_interpreter/dialects/symbol_finder.h
```diff
@@ -0,0 +1,11 @@
+/* This header file declares the FindSymbolInProcess function, which takes a symbol name as 
+input and returns either the symbol's address or an error status, encapsulated in absl::StatusOr<void*> */
+
+#ifndef SYMBOL_FINDER_H_
```
beckerhe, line 4: Can you follow our style for header guards here? I think this should be name `XLA_MLIR_TOOLS_MLIR_INTERPRETER_DIALECTS_SYMBOL_FINDER_H_`

mraunak:
Hi @beckerhe, please review the changes.

xla/mlir/tools/mlir_interpreter/dialects/BUILD
beckerhe: Hey, I think this is still missing. If it is unclear what to do , let me know.

mraunak: Sorry, I missed it. Please review the changes made

beckerhe (approved):
Cool. Thanks. Looks good to me.

PR #18418, opened 2024-10-17
[Draft] [ROCm] removing #ifdef CUDA / ROCM macros where possible

jcaraban:
Hi @xla-rotation, I'm new to XLA and I intend to work on the ROCm side. I thought I could remove some `#ifdef USE_ROCM` macros as a way to get familiar with the repo. Before I go ahead and make a mess, do you have any general advice or preferences on how to achieve this?

I guess I can start with the tests, and try to replace the compile-time `#if #else` with runtime checks? As an example, see the use of `PlatformUtil::CanonicalPlatformName` below. Internally it still uses macros, but at least those are more localized to a single place now.


akuegel:
Thanks, this is definitely a good idea for a starter project. Small advice: please prefer creating several small PRs instead of one big PR, it is usually easier to review and merge small PRs (less potential for merge conflicts).

jcaraban:
Closing, as priorities shifted and I never had the time, unfortunately.

PR #18435, opened 2024-10-17
Shape inference: check that atan2 inputs are float or complex.

sergachev:


bchetioui (approved):

sergachev:
Added a fix for triton_support_test.

xla/service/gpu/fusions/triton/triton_support_test.cc
```diff
@@ -737,7 +745,10 @@ using ReductionComputationTest =
 // computation and in regular HLO. See triton_support.cc for more details.
 TEST_P(ReductionComputationTest, DifferentBinaryOps) {
   auto [data_type, opcode, cc] = GetParam();
-  bool dtype_is_complex = data_type == C64 || data_type == C128;
+  if (opcode == HloOpcode::kAtan2 &&
```
bchetioui, line 748: Same comment as in the `cbrt` PR, this check should be done in `DoesOpSupportType` (and same for the above filter)!

bchetioui (changes requested):

PR #18438, opened 2024-10-17
Shape inference: mark complex cbrt unsupported.

sergachev:


bchetioui (approved):

sergachev:
Added a fix for triton_support_test.

xla/service/gpu/fusions/triton/triton_support_test.cc
```diff
@@ -271,6 +271,9 @@ using UnaryElementwiseTest = TritonSupportTestWithTypeAndOpcodeAndDeviceParam;
 
 TEST_P(UnaryElementwiseTest, IsTritonSupportedUnaryElementwise) {
   auto [data_type, opcode, cc] = GetParam();
+  if (opcode == HloOpcode::kCbrt && primitive_util::IsComplexType(data_type)) {
```
bchetioui, line 274: That's actually not the best fix here! We have a filter above for HLOs that are not supported in general, such that we don't need to modify the test bodies---see `DoesOpSupportType` above. Could you please move this filter there instead?

bchetioui (changes requested):

xla/service/gpu/fusions/triton/triton_support_test.cc
sergachev: Done

PR #18465, opened 2024-10-18
Partially Replicated Instructions in Replication Analysis

philipphack:
Extends the HLO replication analysis to identify partially replicated instructions when the group mode of the collective is flattened ID.

Also enables the exchange of collectives and quantizations or dequantizations with partially replicated scaling factors in the collective quantizer pass.

### DIFF
--- a/xla/hlo/analysis/hlo_replication_analysis.cc
+++ b/xla/hlo/analysis/hlo_replication_analysis.cc
@@ -94,13 +94,20 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
         global_id = Cast<HloAllGatherInstruction>(hlo)->use_global_device_ids();
       }
       if (global_id) {
+        // TODO(philipphack): The following is incorrect if partitions are
+        // replicated differently on replicas, or if replicas are replicated
+        // differently on partitions.
         bool replicated_across_partitions = true;
         bool replicated_across_replicas = true;
         const int64_t num_partitions =
             hlo->GetModule()->config().num_partitions();
         absl::flat_hash_set<int64_t> visited_partitions;
         absl::flat_hash_set<int64_t> visited_replicas;
+        std::vector<int64_t> device_set;
+        std::vector<absl::Span<const int64_t>> device_sets;
+        std::vector<std::vector<int64_t>> device_sets_storage;
         for (const auto& group : hlo->replica_groups()) {
+          device_set.clear();
           visited_partitions.clear();
           visited_replicas.clear();
           visited_replicas.reserve(group.replica_ids().size());
@@ -110,16 +117,25 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
             int64_t pid = id % num_partitions;
             visited_partitions.insert(pid);
             visited_replicas.insert(rid);
+            if (support_partial_replication) {
+              device_set.push_back(cross_partition_spmd ? pid : rid);
+            }
           }
           replicated_across_partitions &=
               visited_partitions.size() == num_partitions;
           replicated_across_replicas &=
               visited_replicas.size() ==
               hlo->GetModule()->config().replica_count();
+          if (support_partial_replication) {
+            device_sets_storage.push_back(device_set);
+            device_sets.push_back(device_sets_storage.back());
+          }
         }
         if ((cross_partition_spmd && replicated_across_partitions) ||
             (!cross_partition_spmd && replicated_across_replicas)) {
           return HloReplication::ReplicatedOnAllDevices();
+        } else if (support_partial_replication) {
+          return HloReplication::PartiallyReplicated(device_sets);
         } else {
           return HloReplication::UniqueOnAllDevices();
         }
--- a/xla/hlo/analysis/hlo_replication_analysis_test.cc
+++ b/xla/hlo/analysis/hlo_replication_analysis_test.cc
@@ -673,6 +673,122 @@ ENTRY entry {
       FindInstruction(module.get(), "dynamic-slice"), {}, replica_groups_2));
 }
 
+TEST_F(HloReplicationAnalysisTest,
+       PartiallyReplicatedAllGatherFlattenedIDReplicaAnalysis) {
+  const std::string module_str = R"(
+HloModule PartiallyReplicatedAllGatherFlattenedIDReplicaAnalysis
+
+ENTRY entry {
+  param = s32[2] parameter(0)
+  all-gather0 = s32[4] all-gather(param), dimensions={0}, replica_groups={{0,2},{4,6},{1,3},{5,7}}, channel_id=1, use_global_device_ids=true
+  all-gather1 = s32[4] all-gather(param), dimensions={0}, replica_groups={{0,4},{2,6},{1,5},{3,7}}, channel_id=2, use_global_device_ids=true
+  ROOT tuple = (s32[4], s32[4]) tuple(all-gather0, all-gather1)
+}
+)";
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto module_replica_analysis,
+      ParseAndReturnVerifiedModule(module_str, /*replica_count=*/4,
+                                   /*num_partitions=*/2));
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<HloReplicationAnalysis> replica_analysis,
+      HloReplicationAnalysis::RunWithPartialReplication(
+          module_replica_analysis.get(),
+          /*cross_partition_spmd=*/false));
+
+  std::array<ReplicaGroup, 2> replica_groups0;
+  replica_groups0[0].add_replica_ids(0);
+  replica_groups0[0].add_replica_ids(1);
+  replica_groups0[1].add_replica_ids(2);
+  replica_groups0[1].add_replica_ids(3);
+
+  std::array<ReplicaGroup, 2> replica_groups1;
+  replica_groups1[0].add_replica_ids(0);
+  replica_groups1[0].add_replica_ids(2);
+  replica_groups1[1].add_replica_ids(1);
+  replica_groups1[1].add_replica_ids(3);
+
+  EXPECT_FALSE(replica_analysis->HloInstructionIsReplicatedAt(
+      FindInstruction(module_replica_analysis.get(), "all-gather0"), {}));
+
+  EXPECT_TRUE(replica_analysis->HloInstructionIsReplicatedAt(
+      FindInstruction(module_replica_analysis.get(), "all-gather0"), {},
+      replica_groups0));
+
+  EXPECT_FALSE(replica_analysis->HloInstructionIsReplicatedAt(
+      FindInstruction(module_replica_analysis.get(), "all-gather0"), {},
+      replica_groups1));
+
+  EXPECT_FALSE(replica_analysis->HloInstructionIsReplicatedAt(
+      FindInstruction(module_replica_analysis.get(), "all-gather1"), {}));
+
+  EXPECT_TRUE(replica_analysis->HloInstructionIsReplicatedAt(
+      FindInstruction(module_replica_analysis.get(), "all-gather1"), {},
+      replica_groups1));
+
+  EXPECT_FALSE(replica_analysis->HloInstructionIsReplicatedAt(
+      FindInstruction(module_replica_analysis.get(), "all-gather1"), {},
+      replica_groups0));
+}
+
+TEST_F(HloReplicationAnalysisTest,
+       PartiallyReplicatedAllGatherFlattenedIDPartitionAnalysis) {
+  const std::string module_str = R"(
+HloModule PartiallyReplicatedAllGatherFlattenedIDPartitionAnalysis
+
+ENTRY entry {
+  param = s32[2] parameter(0)
+  all-gather0 = s32[4] all-gather(param), dimensions={0}, replica_groups={{0,1},{2,3},{4,5},{6,7}}, channel_id=1, use_global_device_ids=true
+  all-gather1 = s32[4] all-gather(param), dimensions={0}, replica_groups={{0,2},{1,3},{4,6},{5,7}}, channel_id=2, use_global_device_ids=true
+  ROOT tuple = (s32[4], s32[4]) tuple(all-gather0, all-gather1)
+}
+)";
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto module_partition_analysis,
+      ParseAndReturnVerifiedModule(module_str, /*replica_count=*/2,
+                                   /*num_partitions=*/4));
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<HloReplicationAnalysis> partition_analysis,
+      HloReplicationAnalysis::RunWithPartialReplication(
+          module_partition_analysis.get(),
+          /*cross_partition_spmd=*/true));
+
+  std::array<ReplicaGroup, 2> replica_groups0;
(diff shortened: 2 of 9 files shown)


reedwm:
CC @yliu120

xla/hlo/analysis/hlo_replication_analysis.cc
```diff
@@ -120,6 +124,8 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
         if ((cross_partition_spmd && replicated_across_partitions) ||
             (!cross_partition_spmd && replicated_across_replicas)) {
           return HloReplication::ReplicatedOnAllDevices();
+        } else if (support_partial_replication) {
+          return HloReplication::PartiallyReplicated(device_sets);
```
reedwm, line 128: The device IDs in device_sets should be partition IDs if cross_partition_spmd=true or replica IDs if cross_partition_spmd=false. But you're adding flattened IDs here. 

Convert to replica or partition IDs, by dividing or modding by num_partitions. This isn't strictly correct: You should also ensure that if cross_partition_spmd=true, all replicas have the particular partial replication, and simiarly if cross_partition_spmd=false, all partitions have the particular partial replication. But we already have a similar bug in the original code where we might incorrectly return ReplicatedOnAllDevices since we don't check individual replicas/partitions. Feel free to fix this here and the original code, or not. If not, I will fix this up myself later.

It might be useful to support flattened IDs here, but you'll have to provide an API to this class to allow the user to query either flattened IDs or partitions/replica IDs. I think we initially should not add an API, and add it later if necessary.

xla/hlo/analysis/hlo_replication_analysis_test.cc
```diff
@@ -673,6 +673,50 @@ ENTRY entry {
...
+)";
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto module, ParseAndReturnVerifiedModule(module_str, /*replica_count=*/1,
+                                                /*num_partitions=*/4));
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<HloReplicationAnalysis> replica_analysis,
+      HloReplicationAnalysis::RunWithPartialReplication(
+          module.get(),
+          /*cross_partition_spmd=*/true));
```
reedwm, line 695: Test cross_partition_spmd=false as well.

xla/hlo/transforms/collectives/collective_quantizer_test.cc
```diff
@@ -279,6 +279,121 @@ TEST_F(CollectiveQuantizerTest, AllGatherQuantizeNonReplicatedScale) {
...
+  EXPECT_THAT(module->entry_computation()->root_instruction(),
+              op::AllToAll(op::Convert(op::Clamp(
+                  op::Broadcast(), op::Divide(op::Parameter(), op::Broadcast()),
+                  op::Broadcast()))));
+  HloInstruction* all_to_all = module->entry_computation()->root_instruction();
+  EXPECT_THAT(all_to_all->shape().element_type(), F8E4M3FN);
+}
+
+TEST_F(CollectiveQuantizerTest,
+       AllGatherQuantizePartialReplicationGroupMismatch) {
```
reedwm, line 363: Add a comment describing why it fails. E.g. "Test case where the rewrite is skipped because the scale is not replicated in the same way as all-gather".

And same for DequantizeAllGatherPartialReplicationGroupMismatch.

reedwm (changes requested):
I think it's worth adding an e2e test as well.

yliu120:
> CC @yliu120

Thank you.

xla/hlo/transforms/collectives/collective_quantizer.cc
```diff
@@ -97,22 +97,26 @@ HloInstruction* ApplyUnaries(HloInstruction* instr,
...
-  TF_ASSIGN_OR_RETURN(
-      auto replication_analysis,
-      HloReplicationAnalysis::Run(module,
-                                  /*cross_partition_spmd=*/true));
-  return replication_analysis->HloInstructionIsReplicatedAt(instr, {});
+  TF_ASSIGN_OR_RETURN(auto replication_analysis,
+                      HloReplicationAnalysis::RunWithPartialReplication(
+                          module,
+                          /*cross_partition_spmd=*/true));
+  return replication_analysis->HloInstructionIsReplicatedAt(instr, {},
```
yliu120, line 118: This is not very efficient.
You can run the replication analysis once as you don't remove instructions during rewrite, you can query the same analysis result.

xla/hlo/transforms/collectives/collective_quantizer.cc
```diff
@@ -307,29 +315,32 @@ absl::Status MatchDequantization(HloInstruction* instr, bool* changed) {
...
+  VLOG(5) << "Collective " << instr->ToString() << " has been replaced with "
+          << new_collective->ToString();
 
-  return absl::OkStatus();
+  return true;
 }
 
-absl::Status MatchQuantization(HloInstruction* instr, bool* changed) {
+absl::StatusOr<bool> MatchQuantization(HloInstruction* instr) {
   std::optional<ConversionSubgraph> subgraph;
```
yliu120, line 325: You 'd better add some VLOG() to easily tell users/developers why something was labeled as FP8 type but not get quantized or dequantized.

You can copy over the same logic in my PR: https://github.com/openxla/xla/pull/18330

Otherwise when things go wrong, I need to go ahead to modify the binary for adding some necessary VLOG to understand why F8 conversion not happening.

xla/hlo/transforms/collectives/collective_quantizer_test.cc
```diff
@@ -449,5 +564,105 @@ TEST_F(CollectiveQuantizerTest, DequantizeAllGatherUnary) {
...
+  )";
+
+  HloModuleConfig config;
+  config.set_num_partitions(8);
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnVerifiedModule(hlo_string, config));
+  TF_ASSERT_OK_AND_ASSIGN(bool changed, RunCollectiveQuantizer(module.get()));
+  EXPECT_FALSE(changed);
+}
+
```
yliu120, line 666: It is worth testing the case where the scale is allreduced in another computation.

xla/hlo/analysis/hlo_replication_analysis_test.cc
```diff
@@ -673,6 +673,69 @@ ENTRY entry {
...
+          module.get(),
+          /*cross_partition_spmd=*/false));
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<HloReplicationAnalysis> partition_analysis,
+      HloReplicationAnalysis::RunWithPartialReplication(
+          module.get(),
+          /*cross_partition_spmd=*/true));
+
+  std::array<ReplicaGroup, 2> replica_groups0;
+  replica_groups0[0].add_replica_ids(0);
```
reedwm, line 703: The `HloInstructionIsReplicatedAt` method is not designed to have duplicated replica IDs. I think the intent of this method is that, assuming cross_partition_spmd=true, the function returns true if and only if for each replica, the instruction output is replicated across the partitions inside the replica. E.g. if there are 2 replicas and 4 partitions and you pass in `replica_groups={{0,2},{1,3}}`, then the function returns true if and only if devices with flattened_id 0 and 2 have the same value, 1 and 3 have the same value, 4 and 6 have the same value, and 5 and 7 have the same value. (Remember that flattened_Id = replia_id * num_partitions + partition_id).

I think it's hard to have a meaningful test with only 2 partitions if cross_partition_spmd=false, and with only 2 replicas if cross_partition_spmd=true.

xla/hlo/transforms/collectives/collective_quantizer.cc
```diff
@@ -36,13 +37,29 @@ struct ConversionSubgraph {
...
   // in top-down order (operand-to-user).
   std::vector<HloInstruction*> unaries;
+
+  std::string ToString() const {
+    auto short_string = [](const HloInstruction* instr) -> std::string {
+      return instr ? instr->ToShortString() : "null";
+    };
+
+    std::ostringstream oss;
+    oss << " conversion: " << short_string(convert)
```
reedwm, line 47: Don't start with a space. And change users of this method to take into account there isn't a space.

xla/service/collective_ops_utils.cc
```diff
@@ -133,6 +133,31 @@ absl::StatusOr<std::vector<int>> GetParticipatingIDs(
                           group->replica_ids().end());
 }
 
+// Returns the group formation mode of instr, assuming that instr is, or is
+// dervied from, an HloAllGatherInstruction, HloAllReduceInstructionBase,
+// HloCollectiveInstruction or HloCollectivePermuteInstruction.
+absl::StatusOr<CollectiveOpGroupMode> GetCollectiveOpGroupMode(
+    HloInstruction* instr) {
+  if (HloAllGatherInstruction::ClassOf(instr)) {
```
reedwm, line 141: It's slightly more concise if you write this as:

```c++
if (auto collective = DynCast<HloAllGatherInstruction>(instr)) {
    return GetCollectiveOpGroupMode(collective->channel_id().has_value(),
                                    collective->use_global_device_ids());
}
```

xla/service/collective_ops_utils.cc
```diff
@@ -133,6 +133,31 @@ absl::StatusOr<std::vector<int>> GetParticipatingIDs(
                           group->replica_ids().end());
 }
 
+// Returns the group formation mode of instr, assuming that instr is, or is
+// dervied from, an HloAllGatherInstruction, HloAllReduceInstructionBase,
+// HloCollectiveInstruction or HloCollectivePermuteInstruction.
+absl::StatusOr<CollectiveOpGroupMode> GetCollectiveOpGroupMode(
```
reedwm, line 139: Add a test for this function.

xla/service/collective_ops_utils.cc
```diff
@@ -133,6 +133,31 @@ absl::StatusOr<std::vector<int>> GetParticipatingIDs(
                           group->replica_ids().end());
 }
 
+// Returns the group formation mode of instr, assuming that instr is, or is
+// dervied from, an HloAllGatherInstruction, HloAllReduceInstructionBase,
+// HloCollectiveInstruction or HloCollectivePermuteInstruction.
```
reedwm, line 138: No need to duplicate comment from header. Also no need to explicitly call out all-gather and all-reduce since they already derive from HloCollectiveInstruction. But maybe it's better to list the HLO collective names instead of classes, e.g. "all-gather, all-reduce", in which case you should list them all.

xla/service/collective_ops_utils.cc
```diff
@@ -133,6 +133,31 @@ absl::StatusOr<std::vector<int>> GetParticipatingIDs(
...
+    HloInstruction* instr) {
+  if (HloAllGatherInstruction::ClassOf(instr)) {
+    auto collective = Cast<HloAllGatherInstruction>(instr);
+    return GetCollectiveOpGroupMode(collective->channel_id().has_value(),
+                                    collective->use_global_device_ids());
+  } else if (HloAllReduceInstructionBase::ClassOf(instr)) {
+    auto collective = Cast<HloAllReduceInstructionBase>(instr);
+    return GetCollectiveOpGroupMode(collective->channel_id().has_value(),
+                                    collective->use_global_device_ids());
+  } else if (HloCollectiveInstruction::ClassOf(instr)) {
```
reedwm, line 149: I'm worried if a new subclass is added with use_global_device_ids, we'll forget to update this. I would either handle each subclass of this separately, or check that it's a hardcoded list of known subclasses (all-to-all, ragged-all-to-all, collective-broadcast)

reedwm (changes requested):

xla/hlo/analysis/hlo_replication_analysis.cc
```diff
@@ -100,7 +100,11 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
             hlo->GetModule()->config().num_partitions();
         absl::flat_hash_set<int64_t> visited_partitions;
         absl::flat_hash_set<int64_t> visited_replicas;
+        std::vector<int64_t> device_set;
+        std::vector<absl::Span<const int64_t>> device_sets;
+        std::vector<std::vector<int64_t>> device_sets_storage;
         for (const auto& group : hlo->replica_groups()) {
+          device_set.clear();
```
reedwm, line 107: Since you clear this at the start of the for-loop, any mutations you make are ignored except for the last iteration.

Because the existing logic is already incorrect even without this PR, it's somewhat hard to review this, as I'm undecided whether bugs are just an extension of the original bug and therefore perhaps acceptable to merge, or whether things like this are considered a new bug and therefore should be fixed in this PR. Instead of fixing this in a separate PR after this PR is merged, I think it makes sense to fix it before this PR. Once your fix is merged, you can merge master into this PR and I can re-review.

xla/hlo/analysis/hlo_replication_analysis_test.cc
```diff
@@ -673,6 +673,81 @@ ENTRY entry {
       FindInstruction(module.get(), "dynamic-slice"), {}, replica_groups_2));
 }
 
+TEST_F(HloReplicationAnalysisTest, PartiallyReplicatedAllGatherFlattenedID) {
+  const std::string module_str = R"(
+HloModule PartiallyReplicatedAllGatherFlattenedID
+
+ENTRY entry {
+  param = s32[2] parameter(0)
+  all-gather0 = s32[4] all-gather(param), dimensions={0}, replica_groups={{0,2},{4,6},{1,3},{5,7}}, channel_id=1, use_global_device_ids=true
```
reedwm, line 682: Remembering what these all-gather replica groups is hard as I scroll down to the rest of the test. I kept having to scroll back up again.

Instead, group all the relevant parts of each instruction being tested. Perhaps by having a lambda taking in the `replica_groups` of the instruction, the `replica_groups` pass to `HloInstructionIsReplicatedAt`, and the value of cross_partition_spmd, and returning whether HloInstructionIsReplicatedAt returned true or not. Then all the information needed to determine whether it should be replicated or not is only a single line.

xla/hlo/analysis/hlo_replication_analysis_test.cc
```diff
@@ -679,61 +679,73 @@ HloModule PartiallyReplicatedAllGatherFlattenedID
...
   EXPECT_TRUE(partition_analysis->HloInstructionIsReplicatedAt(
-      FindInstruction(module.get(), "all-gather0"), {}, replica_groups1));
+      FindInstruction(module_partition_analysis.get(), "all-gather2"), {},
+      replica_groups0));
 
   EXPECT_FALSE(partition_analysis->HloInstructionIsReplicatedAt(
-      FindInstruction(module.get(), "all-gather1"), {}));
+      FindInstruction(module_partition_analysis.get(), "all-gather3"), {}));
 
   EXPECT_TRUE(partition_analysis->HloInstructionIsReplicatedAt(
```
reedwm, line 746: It's worth having some tests where different replicas have different partition_ids in replica_groups. E.g., doing the following

num_partitions = 5
num_replicas = 2
cross_partition_spmd = true
all gather replica_groups = {{0,1,2,5,6},{3,4,7,8,9}}
replica_groups passed to HloInstructionIsReplicatedAt = {{0,1},{2,3,4}}

reedwm (changes requested):

reedwm (approved):
The logic is slightly incorrect in hlo_replication_analysis, but the logic was still incorrect even before this change and you've agreed to fix it, so approving.

loislo:
the cl cannot be built

PR #18488, opened 2024-10-18
compute_library: set openmp runtime off by default

snadampal:
The official wheel binary is using Eigen thread pool runtime and there is no dependency on OpenMP. Having openmp ON by default in compute library bazel config causing an unnecessary dependency on libgomp/libomp for aarch64 linux wheel. This PR is to remove the dependency. There is no impact on the performance because openmp runtime is not being used anyway.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -214,6 +214,7 @@ build:mkl_aarch64 -c opt
 # Config setting to build oneDNN with Compute Library for the Arm Architecture (ACL).
 # with Eigen threadpool support
 build:mkl_aarch64_threadpool --define=build_with_mkl_aarch64=true
+build:mkl_aarch64_threadpool --@compute_library//:openmp=false
 build:mkl_aarch64_threadpool -c opt
 
 # CUDA: This config refers to building CUDA op kernels with nvcc.


ddunl:
Since it's a `bool_flag` could we instead set this in the bazelrc?

snadampal:
Hi @ddunl , actually there is already a bazelrc flag to configure `openmp` vs `threadpool` runtime though out (oneDNN and ACL). For the official wheel building, we are using the `threadpool` config.
https://github.com/openxla/xla/blob/main/.bazelrc#L210 -->this is the openmp config but not used
https://github.com/openxla/xla/blob/main/.bazelrc#L216 --> this is the threadpool config

That's the reason why I wanted to keep the ACL default openmp setting to `false` so that there won't be any unwanted `libgomp or libomp` dependency for the wheels.

ddunl:
I agree with the spirit of the change, but I just mean that under those configs in the bazelrc, couldn't we add `--@compute_library//:openmp=true` (possibly botching the exact syntax there) to achieve the same effect rather than patching the BUILD file?

snadampal:
Hi @ddunl , I have updated the PR, it looks much neat now, thanks for your suggestion!

ddunl:
LGTM, but it seems like we should also add it to the `mkl_threadpool` config (the non-aarch64 one)

snadampal:
the compute library and hence this issue is applicable to only aarch64 platforms.

ddunl:
Oh I see - my bad. Thanks for the PR! I'll try to get it merged today/tomorrow.

ddunl (approved):

ddunl:
Oh I see this change is also in https://github.com/tensorflow/tensorflow/pull/76831, we actually currently mirror TF's bazelrc so I'll try to get that one merged internally as it's already been imported. Once that happens we can close this PR. There are some Windows errors but it seems like it's a flake (?).

ddunl:
Ok! With https://github.com/tensorflow/tensorflow/pull/76831 merged I'll close this. Thanks!!

snadampal:
thanks, @ddunl !

PR #18558, opened 2024-10-21
[GPU][NFC] Add HLO op profiler build test.

sergachev:
hlo_op_profiler_run.cc wasn't compiled regularly and got build failures from time to time.


dimitar-asenov (approved):

xla/service/gpu/model/BUILD
```diff
@@ -967,6 +968,13 @@ cc_library(
     ]
 ]
 
+build_test(
```
akuegel, line 971: In fact, the build test is showing that there is currently a breakage when layering check is enabled. There is no dependency for the "third_party/tensorflow/tsl/platform/protobuf.h" include.

sergachev: fixed

akuegel (approved):

PR #18562, opened 2024-10-21
Compute cost tool: add an option to print costs of each instruction.

sergachev:


xla/tools/compute_cost.cc
```diff
@@ -72,14 +77,39 @@ int main(int argc, char** argv) {
...
+      *xla::LoadModuleFromFile(input, format, {});
+
+  TF_CHECK_OK(
+      module->entry_computation()->root_instruction()->Accept(&*analysis));
 
-  std::cout << std::setw(5) << std::setprecision(4)
-            << analysis->flop_count() / (1e9) << " GFLOPS. "
-            << analysis->bytes_accessed() / (1e6) << " MB." << std::endl;
+  if (all) {
+    absl::flat_hash_map<std::string, std::string> fingerprint_to_name;
```
dimitar-asenov, line 87: Please put the code from this branch in a separate function.

xla/tools/compute_cost.cc
```diff
@@ -43,18 +44,22 @@ The input file can be obtained from XProf graph viewer by clicking
...
   bool gpu = false;
+  bool all = false;
   std::vector<tsl::Flag> flag_list = {
       tsl::Flag("input", &input, "input file"),
       tsl::Flag("format", &format, "hlo|pb|pbtxt"),
       tsl::Flag("gpu", &gpu,
-                "Use GPU flavor of cost analysis instead of the generic one")};
+                "Use GPU flavor of cost analysis instead of the generic one"),
+      tsl::Flag("all", &all,
+                "Print costs and deduplicated name of each instruction, not "
```
dimitar-asenov, line 61: This suggests that you print two things:
- the cost of each instruciton
- the total costs.

However, the code below prints either one or the other. Please make the description and code consistent. I would prefer a change to the code where you print the individual costs for each instruction and then also print the totals at the end.

dimitar-asenov (changes requested):

xla/tools/compute_cost.cc
sergachev: done

xla/tools/compute_cost.cc
sergachev: done

dimitar-asenov (approved):

PR #18568, opened 2024-10-21
[NVIDIA] Swap the collective pipeliner of AR and AG passes

kaixih:
This PR provides a workaround by swapping the collective pipelining of AllReduce (AR) and AllGather (AG). We originally encountered a divergence issue when both AR and AG were enabled simultaneously. After swapping their order, the issue disappeared. In theory, these two passes should be independent, meaning either order should generate a mathematically equivalent graph. For now, we are swapping the order as a temporary fix for the issue.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -874,21 +874,6 @@ absl::Status RunCollectiveOptimizationPasses(
   collectives_pipeline.AddPass<HloDCE>();
 
   if (debug_options.xla_gpu_enable_pipelined_collectives() ||
-      debug_options.xla_gpu_enable_pipelined_all_reduce()) {
-    CollectivePipeliner::Config config{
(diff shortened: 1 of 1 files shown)


golechwierowicz:
I'm not sure we should merge it given we do not know the root cause yet and that pipeliners are off by default.

Personally I won't get into debugging this week.

jprabhas:
My understand

> I'm not sure we should merge it given we do not know the root cause yet and that pipeliners are off by default.
> 
> Personally I won't get into debugging this week.

+1. It does not seem necessary to merge a workaround just to suppress the issue here.

yliu120:
I agree with Grzegorz because it doesn't seem this transformation is the root cause.
Must be some other things. I can take a look later as we are also blocked by this numerical bug

kaixih:
@yliu120 Can you take a look at the #19363 which has root caused the issue and fixed it?

PR #18595, opened 2024-10-22
Verify device compatibility for execution early

jaro-sevcik:
Let us verify that all the devices to run a PJRT executable are compatible with the device that was used to build the executable. Currently we only verify the compatibility on each device separately, just before running the executable on the device.

This fixes a problem where running an executable with a collective on two different devices crashes on NCCL timeout. When only one of the device will not pass the compatibility check while the other device (the one passes the check) runs, the running device will hang on the NCCL collective (since the incompatible device program has not been executed).

### DIFF
--- a/xla/pjrt/pjrt_stream_executor_client.cc
+++ b/xla/pjrt/pjrt_stream_executor_client.cc
@@ -3148,6 +3148,24 @@ PjRtStreamExecutorLoadedExecutable::ExecuteHelper(
   return Result({/*future=*/std::move(future), /*buffers=*/std::move(outputs)});
 }
 
+absl::Status PjRtStreamExecutorLoadedExecutable::VerifyCompatibleDevices()
+    const {
+  const int num_addressable_devices = addressable_devices_.size();
+  for (int i = 0; i < num_addressable_devices; ++i) {
+    PjRtDevice* device = addressable_devices_[i];
+    const int device_ordinal =
+        tensorflow::down_cast<PjRtStreamExecutorDevice*>(device)
+            ->local_device_state()
+            ->local_device_id()
+            .value();
+    const int partition = addressable_device_logical_ids_[i].partition;
+    const int executable_idx = executables_.size() > 1 ? partition : 0;
+    TF_RETURN_IF_ERROR(executables_[executable_idx]->VerifyRunDeviceCompatible(
+        device_ordinal));
+  }
+  return absl::OkStatus();
+}
+
 absl::StatusOr<std::vector<std::vector<std::unique_ptr<PjRtBuffer>>>>
 PjRtStreamExecutorLoadedExecutable::Execute(
     absl::Span<const std::vector<PjRtBuffer*>> argument_handles,
@@ -3172,6 +3190,7 @@ PjRtStreamExecutorLoadedExecutable::Execute(
         num_partitions());
   }
 
+  TF_RETURN_IF_ERROR(VerifyCompatibleDevices());
   VLOG(1) << "Executing computation " << name()
           << "; num_replicas=" << num_replicas()
           << " num_partitions=" << num_partitions()
--- a/xla/client/local_client.cc
+++ b/xla/client/local_client.cc
@@ -126,22 +126,7 @@ absl::Status LocalExecutable::ValidateExecutionOptions(
                              ? run_options.stream()->parent()->device_ordinal()
                              : backend_->default_device_ordinal();
   }
-  TF_ASSIGN_OR_RETURN(bool devices_equivalent,
-                      backend_->devices_equivalent(
-                          run_device_ordinal, build_options_.device_ordinal()));
-  if (!devices_equivalent) {
-    TF_ASSIGN_OR_RETURN(se::StreamExecutor * run_executor,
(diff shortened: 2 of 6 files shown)


xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3148,6 +3148,24 @@ PjRtStreamExecutorLoadedExecutable::ExecuteHelper(
   return Result({/*future=*/std::move(future), /*buffers=*/std::move(outputs)});
 }
 
+absl::Status PjRtStreamExecutorLoadedExecutable::VerifyCompatibleDevices()
+    const {
+  const int num_addressable_devices = addressable_devices_.size();
+  for (int i = 0; i < num_addressable_devices; ++i) {
```
klucke, line 3154: Shouldn't you be iterating over both the addressable_devices_ AND the executables_?  (i.e. shouldn't you ensure that all the executables are compatible with all addressable devices?)

jaro-sevcik: From what I see in the code, for each device, there is only one executable that will be run on the device. See L3162 below (mirroring the logic from https://github.com/openxla/xla/blob/6ec8b9e4ffaf2e02a3d7dfb6ec30ed6c81365d42/xla/pjrt/pjrt_stream_executor_client.cc#L3075).

So we just check that the device is compatible with "its" executable.

klucke: Then shouldn't you just iterate over each executable then instead?  Then you don't have to make the check for executables_.size when selecting which device to see if it's compatible.

jaro-sevcik: > Then shouldn't you just iterate over each executable then instead?

I am puzzled how this would be done. How do I then know which devices should the executable be checked against for compatibility?

When I want to "check that each device is compatible with 'its' executable", the natural way is looping over all devices, finding the local executable that will be run on that device and making sure that the executable is compatible with that device. No? This is exactly what the code above is doing...  

Or, to put it differently, my patch just tries mirrors how the Execute method verifies compatibility, we just do the compatibility checking upfront, before anything gets scheduled/executed. I am really struggling to understand why we need to check this differently from how XLA executables actually get executed.

Here is where all the components of the check come from, all of them are inside the Execute method's (async) call graph:

Here is the source code of the loop (from the Execute method):
https://github.com/openxla/xla/blob/6ec8b9e4ffaf2e02a3d7dfb6ec30ed6c81365d42/xla/pjrt/pjrt_stream_executor_client.cc#L3193

Here is how we get the partition (also from the Execute method):
https://github.com/openxla/xla/blob/6ec8b9e4ffaf2e02a3d7dfb6ec30ed6c81365d42/xla/pjrt/pjrt_stream_executor_client.cc#L3195

Here is how we get the device ordinal (from ExecuteHelper, called by Execute):
https://github.com/openxla/xla/blob/6ec8b9e4ffaf2e02a3d7dfb6ec30ed6c81365d42/xla/pjrt/pjrt_stream_executor_client.cc#L3065

Here is how the executable gets determined (also from ExecuteHelper):
https://github.com/openxla/xla/blob/6ec8b9e4ffaf2e02a3d7dfb6ec30ed6c81365d42/xla/pjrt/pjrt_stream_executor_client.cc#L3075

Here is where the current check is done (just before executing on device, call path is Execute->ExecuteHelper->LocalClient::RunAsync->RunHelper->ValidateExecutionOptions, with a bit of scheduling in between):
https://github.com/openxla/xla/blob/450b61ff3d9cb50961fdaa811dbd24a1269f5677/xla/client/local_client.cc#L129

> Then you don't have to make the check for executables_.size when selecting which device to see if it's compatible.

The check for executables_.size is not selecting a device, it is selecting an executable.

According to the comment below, the size check is there because SPMD sharding does not bother to create an executable per partition, so if there is just a single executable, we will use it for all the partitions/devices.
https://github.com/openxla/xla/blob/6ec8b9e4ffaf2e02a3d7dfb6ec30ed6c81365d42/xla/pjrt/pjrt_stream_executor_client.cc#L3074

I agree that it might be nice to explicitly check that is `executables_.size() > 1` then `partition < executables_.size()`. Incidentally, something even stronger is checked here:
https://github.com/openxla/xla/blob/6ec8b9e4ffaf2e02a3d7dfb6ec30ed6c81365d42/xla/pjrt/pjrt_stream_executor_client.cc#L2421

klucke (approved):

golechwierowicz:
Is it possible to write a test for this change?

jaro-sevcik:
> Is it possible to write a test for this change?

Not easily, as far as I can see. I do not see a way to create two devices with different names without having two different GPUs in a machine.

I tried to subclass HostPlatform and create a client for that, but that did not work (bunch of assertions fail).

Perhaps one way would be to change the host platform to allow creating devices with supplied names. This would require some code in HostPlatform that would be there just for the test. Does that sound reasonable?

golechwierowicz:
@klucke WDYT? or should we push it through internally without a test?

klucke:
> @klucke WDYT? or should we push it through internally without a test?

Let's just push through without a test for now.

PR #18616, opened 2024-10-22
[XLA:CPU][oneDNN] Refactor code that fuses Add operation with oneDNN primitives

akhilgoe:
This PR refactors the code that fuses add operation to matmul / convolution primitives. It removes usage of macros and separate templatized handlers for matmul and convolution cases.


xla/service/cpu/onednn_contraction_rewriter.h
```diff
@@ -61,18 +63,26 @@ class OneDnnContractionRewriter : public HloModulePass {
...
-          BackendConfig::BackendConfigOneofCase::kOnednnMatmulConfig>( \
-          contraction, __VA_ARGS__);                                   \
-    default:                                                           \
-      return internal_callee<                                          \
-          BackendConfig::BackendConfigOneofCase::kOnednnConvConfig>(   \
-          contraction, __VA_ARGS__);                                   \
+using ContractionVariant = std::variant<PrimitiveTrait<kOnednnConvConfig>,
+                                        PrimitiveTrait<kOnednnMatmulConfig>>;
+using FusionsConfigPointer = xla::cpu::OneDnnFusionConfig*;
+using OptimizationsConfigPointer = xla::cpu::OneDnnOptimizationConfig*;
```
penpornk, line 69: Per Google's [C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Aliases), we avoid making aliases in header if their purpose is only for convenience. We only have aliases in header when it's intended to be part of the API (e.g., for users to use).

I think we can keep `ContractionVariant`, but maybe rename it to `OneDnnContractionVariant` (or put it in a `onednn` namespace) to make it clear this is specific to oneDNN.

`FusionsConfigPointer` and `OptimizationsConfigPointer` seem trivial and shouldn't be declared here.

(Aliases are fine in .cc files.)

xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -612,141 +644,125 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
...
+      auto fusions_config = GetFusionsConfig(&backend_config);
+      auto optimization_config = GetOptimizationsConfig(&backend_config);
+      // TODO(intel-tf): Remove this restriction once oneDNN has an optimized
+      // implementation for broadcasted add across all dimensions.
+      OneDnnFusionConfig_FusionKind kind = OneDnnFusionConfig::UNDEFINED;
+      kind =
+          (addend->shape().rank() == 1)
+              ? (fusions_config->ops().empty() ? OneDnnFusionConfig::BIAS
+                                               : OneDnnFusionConfig::UNDEFINED)
+              : OneDnnFusionConfig::BINARY_ADD;
```
penpornk, line 718: Nit: Lines 715-718 can be put directly in the initialization in line 713.

xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -612,141 +644,125 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
...
+      auto backend_config = custom_call->backend_config<BackendConfig>();
+      auto fusions_config = GetFusionsConfig(&backend_config);
+      auto optimization_config = GetOptimizationsConfig(&backend_config);
+      // TODO(intel-tf): Remove this restriction once oneDNN has an optimized
+      // implementation for broadcasted add across all dimensions.
+      OneDnnFusionConfig_FusionKind kind = OneDnnFusionConfig::UNDEFINED;
+      kind =
+          (addend->shape().rank() == 1)
+              ? (fusions_config->ops().empty() ? OneDnnFusionConfig::BIAS
+                                               : OneDnnFusionConfig::UNDEFINED)
```
penpornk, line 717: Does this mean the code doesn't support having 1D addend (bias) and fused ops at the same time? Please document this as a comment in the code.

penpornk (changes requested):

xla/service/cpu/onednn_contraction_rewriter.cc
akhilgoe: This should ensure that 1D addends are fused only when they are the first post-op operation. So the following should be acceptable:

1. Bias (1D) + <0 or more non-add post-op>
2. Bias (1D) + <0 or more non-add post-op> + Add (non-1D) + <0 or more non-add post-op>  

Following is not allowed:

1. Bias (1D) + <0 or more non-add post-op> + Add (1D) + <0 or more non-add post-op> 

This condition was added a few months ago because at the time oneDNN had optimized implementations for broadcasted add operations across certain dimensions only. As a result, some cases defaulted to the ref implementation, which significantly impacted performance.
We can re-evaluate this restriction with the latest oneDNN release and/or relax this a bit by blocking only those 1D cases where broadcasting occurs along some specific dimensions.

xla/service/cpu/onednn_contraction_rewriter.cc
akhilgoe: Done

xla/service/cpu/onednn_contraction_rewriter.h
akhilgoe: Done

akhilgoe:
Hi @penpornk, thanks for the review! I have added a commit to address your comments. Please take a look whenever possible. Thanks!

penpornk (approved):
Thank you for the changes and I'm very sorry for the delay!

PR #18623, opened 2024-10-22
Fix while loop analysis

shraiysh:
This PR reverts the unsafe changes to while loop analysis in PR #15417. If the loop induction variable is passed to a custom-call and the output of that custom-call is used as the loop induction variable, we cannot safely assume that the custom-call returns the loop induction variable unchanged. This PR removes this unsafe relaxation.

### DIFF
--- a/xla/hlo/analysis/while_loop_analysis.cc
+++ b/xla/hlo/analysis/while_loop_analysis.cc
@@ -83,9 +83,7 @@ static optional<int64_t> GetGTEOperandIndex(const HloInstruction* instr,
     }
 
     if (!Match(possibly_gte_operand,
-               m::GetTupleElement(m::Op().Is(gte_operand))) &&
-        !Match(possibly_gte_operand,
-               m::GetTupleElement(m::CustomCall(m::Op().Is(gte_operand))))) {
+               m::GetTupleElement(m::Op().Is(gte_operand)))) {
       return nullopt;
     }
 
@@ -282,7 +280,7 @@ optional<int64_t> GetLoopInductionVarTupleIdx(const HloInstruction* while_op) {
     return nullopt;
   }
 
-  // The while_body computation should have one of the following forms:
+  // The while_body computation should have the form:
   //
   // Form 1:
   //   while_body_inc =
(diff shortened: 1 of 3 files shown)


ptoulme-aws:
@shraiysh before we revert, I have one more idea. Can we keep the PR but change while loop analysis to check if the custom-call has in backend_config something like "Pass_through_loop_var=True" indicating it passes loop var unchanged??

shraiysh:
>"Pass_through_loop_var=True"

Adding this flag specifically for this usage doesn't seem like a compelling case to add a new attribute to an operation -- and if we do, it would be have to be a more complicated attribute, because we would (at least) also have to pass the index of the operand we are returning unchanged (consider a custom-call that takes 2 integers as argument).

I would however not be strongly opposed to it, if the reviewers @bchetioui @fhoushmand think that it is a good idea.

I would still suggest that we should land this PR so that while loop analysis becomes safe upstream and then we should discuss the solution while trying to land the feature safely again.

ptoulme-aws:
> > "Pass_through_loop_var=True"
> 
> Adding this flag specifically for this usage doesn't seem like a compelling case to add a new attribute or config to an operation -- and if we do, it would be have to be a more complicated attribute, because we would (at least) also have to pass the index of the operand we are returning unchanged (consider a custom-call that takes 2 integers as argument).
> 
> I would however not be strongly opposed to it, if the reviewers @bchetioui @fhoushmand think that it is a good idea.
> 
> I would still suggest that we should land this PR so that while loop analysis becomes safe upstream and then we should discuss the solution while trying to land the feature safely again.

Not an op attribute. A boolean in the backend_config of the custom-call saying its safe to unroll meaning any while loop value is pass through. This means the caller, such as Neuron must put in their custom-calls (unroll_safe=True) meaning while loop analysis allows unrolling. This would be a 2 line PR in while loop analysis. Just check that unroll_safe=True

bchetioui:
> Not an op attribute. A boolean in the backend_config of the custom-call saying its safe to unroll meaning any while loop value is pass through. This means the caller, such as Neuron must put in their custom-calls (unroll_safe=True) meaning while loop analysis allows unrolling. This would be a 2 line PR in while loop analysis. Just check that unroll_safe=True

This is the moral equivalent of an op attribute, though. Even though it's easier to define this as an annotation on a `custom-call`, we're ultimately saying "here is a new kind of `custom-call` that introduces new semantics". Although it seems like these added semantics would be fairly well scoped, modifying the IR in order to facilitate working around the existing semantics does not seem reasonable absent a clear use case where it's not better to find a less invasive, workable solution.

Here, my understanding is that you work with internal `custom-call`s that return a loop induction variable, but actually promise to be a complete no-op with regards to the loop induction variable. I agree with @shraiysh that this does not seem like a compelling reason to extend the IR.

It's really not clear to me why we need to return that loop induction variable in the first place. Perhaps the code that produce this `custom-call` could just produce a different `custom-call` instead that does not return a loop induction variable, if it already knows that it is unnecessary?

Either way, it seems like we shouldn't put the burden of working around this downstream design choice on the IR, nor on this pass. It seems like something that should be solved downstream. Let's revert this for now.

bchetioui (approved):
Thanks!

akuegel:
I completely agree with @bchetioui . As I said in a comment on the original PR, "In fact, if the custom call returns the induction variable unchanged, why make it an output of the custom call in the first place?"

PR #18765, opened 2024-10-25
[ds-fusion]Add support for fusion operations in while loop analysis.

shraiysh:
After optimizations, while loop analysis was unable to handle the fused
operations, primarily for `GetLoopInductionVariableIndex` and
`ComputeWhileLoopTripCount`. This patch handles
`GetloopInductionVariableIdx`.

  * Enhanced `GetGTEOperandIndex` to `GetUniqueGTEDependenceIndex`

### DIFF
--- a/xla/hlo/analysis/while_loop_analysis.cc
+++ b/xla/hlo/analysis/while_loop_analysis.cc
@@ -18,7 +18,10 @@ limitations under the License.
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
+#include <memory>
+#include <optional>
 
+#include "absl/algorithm/container.h"
 #include "absl/base/casts.h"
 #include "absl/container/flat_hash_map.h"
 #include "xla/comparison_util.h"
@@ -27,10 +30,12 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_module.h"
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/hlo/ir/hlo_reachability.h"
+#include "xla/hlo/utils/hlo_query.h"
 #include "xla/literal.h"
 #include "xla/literal_util.h"
 #include "xla/service/pattern_matcher.h"
 #include "xla/shape_util.h"
+#include "xla/tools/hlo_extractor.h"
 
 namespace xla {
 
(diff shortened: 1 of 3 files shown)


akuegel (approved):

shraiysh:
Rebased this to main. Requesting review.

loislo:
one of our tests fails with this cl. Most likely it fails with OOM

loislo:
the test dies with the timeout. 
image

also the analysis allocates a lot of memory
image

xla/hlo/analysis/while_loop_analysis.cc
```diff
@@ -27,77 +27,78 @@ limitations under the License.
...
+// For example, in the following HLO, this function will return `1`:
+//   in = (s32[], s32[], s32[]) tuple(a,b,c)
+//   gte.1 = get-tuple-element(in), index=1
+//   out = fusion(gte.1), ...
+std::optional<int64_t> GetUniqueGTEDependenceIndex(const HloInstruction* out,
+                                                   const HloInstruction* in) {
+  if (out->parent() != in->parent() || !in->shape().IsTuple()) {
+    return std::nullopt;
+  }
+  std::unique_ptr<HloModule> extracted = ExtractModule(
```
fhoushmand, line 75: Can you add a comment explaining what is this doing?

shraiysh: Added it. And I removed the HloEvaluator changes from this PR. Can you check the test that was timing out above? Does it still fail?

shraiysh:
Updated the PR to remove the changes in brute force which was causing OOM issues and rebased it to main.

xla/hlo/analysis/while_loop_analysis.cc
```diff
@@ -21,83 +21,89 @@ limitations under the License.
...
+  }
+  HloInstruction* param = entry->parameter_instruction(0);
+  if (param->user_count() == 0) return nullopt;
+  if (llvm::any_of(param->users(), [](const HloInstruction* inst) -> bool {
+        return inst->opcode() != HloOpcode::kGetTupleElement;
+      })) {
+    return std::nullopt;
+  }
+  int64_t candidate_index = param->users()[0]->tuple_index();
+  if (llvm::any_of(param->users(),
```
fhoushmand, line 100: why are you using llvm library? use absl::c_any_of instead

xla/hlo/analysis/while_loop_analysis.cc
fhoushmand: fix the comment please. "... we can uniquely determine which ..."

fhoushmand: I'm still not following what is being done in this function. Can you add comment for each part please?
Also, what happens if you have multiple gtes but with the same input and index?

fhoushmand (changes requested):

xla/hlo/analysis/while_loop_analysis.cc
shraiysh: Thanks! Changed it.

xla/hlo/analysis/while_loop_analysis.cc
shraiysh: > fix the comment please. "... we can uniquely determine which ..."

Thanks! Fixed it.

> I'm still not following what is being done in this function. Can you add comment for each part please?

I added some more comments.

> Also, what happens if you have multiple gtes but with the same input and index?

That is handled, as the last condition in this function. It allows for multiple gtes, but ensures that we return a value for index only if all of them have the same index.

shraiysh:
Gentle ping for review!

fhoushmand (approved):

fhoushmand:
> Gentle ping for review!

Can you fix these please?

no header providing "std::optional" is directly included
no header providing "std::unique_ptr" is directly included
no header providing "absl::c_any_of" is directly included

fhoushmand (approved):

akuegel:
@shraiysh This PR was rolled back because it increased compile times significantly and led to test timeouts. I suppose the simulation in while loop analysis is at fault, we probably need some heuristic to guard against doing too much computation there.

PR #18785, opened 2024-10-28
[nfc] Cleanup build dependencies for collective transforms.

shraiysh:
Create a package for the internal folder with collective optimizations so the source files are in sync with their build files.


akuegel (commented):
Personally I agree with this change, but given that so far all targets are defined in the transforms BUILD file, it seems this might have been intentional. I think it would be better to get this reviewed by someone working on this refactoring.

akuegel:
@sdasgup3 seems to be out of office, so adding Mason as reviewer who might know more about what the plan is.

changm:
This seems fine with me but redirecting to @GleasonK who is closer to the plan.

GleasonK (approved):

GleasonK:
No objections but also not entirely sure of the benefits. Would make sense if we wanted to limit visibility of the collectives package, but otherwise unclear why one choice is better than the other. Regardless LGTM.

shraiysh:
> No objections but also not entirely sure of the benefits. Would make sense if we wanted to limit visibility of the collectives package, but otherwise unclear why one choice is better than the other. Regardless LGTM.

The reason for this cleanup is mainly convenience while working on transforms :) . I was working on a pass and I usually just run `bazel test xla/hlo/transforms/collectives/all_gather_combiner`, relying on autocomplete to help me quickly find the target. The autocomplete doesn't work in this case, and so I thought I might just clean this up. Also, the build file got pretty huge (over 2k lines). While doing the same thing for expanders and simplifiers, I also found a couple transforms that do not have tests associated with them too. I can send out those patches too, but wanted to start with a smaller patch to see if this would be accepted.

akuegel:
> > No objections but also not entirely sure of the benefits. Would make sense if we wanted to limit visibility of the collectives package, but otherwise unclear why one choice is better than the other. Regardless LGTM.
> 
> The reason for this cleanup is mainly convenience while working on transforms :) . I was working on a pass and I usually just run `bazel test xla/hlo/transforms/collectives/all_gather_combiner`, relying on autocomplete to help me quickly find the target. The autocomplete doesn't work in this case, and so I thought I might just clean this up. Also, the build file got pretty huge (over 2k lines). While doing the same thing for expanders and simplifiers, I also found a couple transforms that do not have tests associated with them too. I can send out those patches too, but wanted to start with a smaller patch to see if this would be accepted.

Another advantage I want to point out is that there is less risk of merge conflicts, if we have a separate BUILD file for these targets. So it might be worth it to do the same kind of cleanup for the other passes that are in subdirectories.

shraiysh:
Thanks! By the way, this patch still needs an approval to merge (I had to rebase it because of a merge conflict).

GleasonK (approved):

shraiysh:
@GleasonK Is there an internal failure with this PR?

loislo:
> @GleasonK Is there an internal failure with this PR?

we have a number of internal targets that depend on the targets that you moved. As a result the pr cannot be submitted directly by copybara and need a number of manual fixed.

PR #18794, opened 2024-10-28
[GPU] Run horizontal loop fusion not only in the entry computation.

sergachev:


xla/service/gpu/transforms/horizontal_loop_fusion.cc
```diff
@@ -738,11 +737,22 @@ absl::StatusOr<bool> HorizontalLoopFusion::Run(
...
-  TF_ASSIGN_OR_RETURN(bool changed,
-                      RunOnComputation(module->entry_computation()));
+  bool any_changed = false;
+  std::unique_ptr<CallGraph> call_graph = CallGraph::Build(module);
+  for (HloComputation* computation :
+       module->MakeNonfusionComputations(execution_threads)) {
+    if (computation->IsEntryComputation() ||
+        call_graph->GetComputationCallers(computation).size() == 1 &&
+            call_graph->GetComputationCallers(computation)[0]->opcode() ==
+                HloOpcode::kWhile &&
```
olegshyshkov, line 747: Interesting, I never noticed that it only runs on the entry computation. 

Any reason why we want this inside while body, but not other fusible computations? For example, conditional body? For all fusible computations we have this helper: 
https://github.com/openxla/xla/blob/06e520756a5d9288d8b689fe387acab810bc3de8/xla/service/gpu/gpu_fusible.h#L81

sergachev: Thanks! I was looking for such a helper. While body just typically contains almost everything that is not in the entry computation.

olegshyshkov (approved):
Thanks!

akuegel:
It seems a Jax test starts failing with this change:

[jax/tests:experimental_rnn_test](https://github.com/jax-ml/jax/blob/main/tests/experimental_rnn_test.py)

AssertionError: 
Not equal to tolerance rtol=1e-05, atol=1e-05

Mismatched elements: 3283 / 3312 (99.1%)
Max absolute difference among violations: 0.52820003
Max relative difference among violations: 284.8273
 ACTUAL: array([ 0.000563, -0.002013, -0.001009, ...,  0.016082, -0.41868 ,
       -0.12589 ], dtype=float32)
 DESIRED: array([ 3.277339e-04, -1.822406e-03, -9.853034e-04, ...,  5.046067e-02,
       -3.675035e-01, -1.222960e-01], dtype=float32)

Can you reproduce this?

sergachev:
I tried to run it on H100 and it has many failures with unmodified XLA. Which GPU do you see the failure on and in which test case?

akuegel:
> I tried to run it on H100 and it has many failures with unmodified XLA. Which GPU do you see the failure on and in which test case?

It is failing on H100, and these two tests are failing:

RnnTest.test_lstm33 (batch_size=4, seq_len=4, input_size=1, hidden_size=6, num_layers=4, bidirectional=True)
RnnTest.test_lstm28 (batch_size=4, seq_len=4, input_size=2, hidden_size=6, num_layers=4, bidirectional=True)

Our test infra says potentially newly failing, but when I look at the history of test results it seems to be consistently passing before.

sergachev:
Thanks!

Something is strange with this test: it only runs RnnTest.test_lstm0...RnnTest.test_lstm9; test_lstm28 and test_lstm33 are not even generated. I reduced the test parameter space and reproduced the failures with cuDNN 8.9.

It doesn't fail anymore after simply rebasing though. Maybe something else got fixed in the meanwhile.

loislo:
experimental_rnn_test in jax/tests is failing. Very big difference in numbers

loislo:
https://github.com/jax-ml/jax/blob/main/tests/BUILD
experimental_rnn_test

sergachev:
@mooskagh @loislo I didn't get any failures on H100 in //tests:experimental_rnn_test_gpu with the PR rolled forward at head. Which hardware does it fail on for you? Which of the 128 test cases fail?

loislo:
+Alexander Lyashuk ***@***.***>

Ilya,
OpenXLA Team

PR #18801, opened 2024-10-29
[GPU] Horizontal loop fusion: remove layout check.

sergachev:
The check as it's written doesn't prevent uncoalesced access as it intends but does prevent some reasonable fusions.

### DIFF
--- a/xla/service/gpu/transforms/horizontal_loop_fusion.cc
+++ b/xla/service/gpu/transforms/horizontal_loop_fusion.cc
@@ -232,27 +232,6 @@ bool IsProfitableFusionCandidate(const HloInstruction& instr,
   return true;
 }
 
-// Returns whether `fusion_instr` has only row-major layouts.
-// The horizontal fusion excludes computations with non-row-major layouts,
-// because fusing computations with different layouts can result in uncoalesced
-// memory accesses and cause great performance overhead.
-bool HasOnlyRowMajorLayout(const HloInstruction& instr) {
-  if (instr.opcode() != HloOpcode::kFusion) {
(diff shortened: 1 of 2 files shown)


olegshyshkov (approved):

xla/service/gpu/transforms/horizontal_loop_fusion.cc
```diff
@@ -232,27 +232,6 @@ bool IsProfitableFusionCandidate(const HloInstruction& instr,
   return true;
 }
 
-// Returns whether `fusion_instr` has only row-major layouts.
```
jprabhas, line 235: How does codegen handle it if we do end up with different layouts in the same fusion kernels, atleast some of the reads/writes will be uncoalesced?

sergachev: The way the check is written it doesn't actually check what would be the iteration order over physical tensors and thus is far from reasoning about coalescence. A fusion can be perfectly coalesced but rejected by this check like in the test I added or can be uncoalesced with slices / reductions / transposes but would pass through the check because it has default layouts.

Independently of that, codegen should work with any combinations of layouts, it's elementwise and doesn't enforce / rely on coalescence.

sergachev:
Rebased to resolve a conflict.

olegshyshkov (approved):

loislo:
xla/tests:client_test_gpu is failing

sergachev:
> xla/tests:client_test_gpu is failing

Thanks, fixed another assumption about layouts and added another test.

sergachev:
Rebased.

akuegel (approved):

PR #18837, opened 2024-10-29
Enable HLO elementwise ops for int4 types

sergey-kozub:
Allow more HLO ops on sub-byte types: all elementwise ops and opt-barrier.

### DIFF
--- a/xla/service/cpu_gpu_shape_verifier.cc
+++ b/xla/service/cpu_gpu_shape_verifier.cc
@@ -75,6 +75,7 @@ absl::Status VerifyS4U4Usage(HloInstruction* instruction) {
     case HloOpcode::kDynamicUpdateSlice:
     case HloOpcode::kFusion:
     case HloOpcode::kGetTupleElement:
+    case HloOpcode::kOptimizationBarrier:
     case HloOpcode::kParameter:
     case HloOpcode::kSlice:
     case HloOpcode::kTuple:
@@ -86,7 +87,9 @@ absl::Status VerifyS4U4Usage(HloInstruction* instruction) {
       }
       ABSL_FALLTHROUGH_INTENDED;
     default:
-      return verify_subshape(instruction);
+      if (!instruction->IsElementwise()) {
+        return verify_subshape(instruction);
+      }
   }
 
   return absl::OkStatus();
--- a/xla/service/gpu/fusions/triton/triton_fusion_emitter_device_legacy_test.cc
+++ b/xla/service/gpu/fusions/triton/triton_fusion_emitter_device_legacy_test.cc
@@ -162,7 +162,7 @@ TEST_F(TritonGemmTest, RejectDotInt4HLO) {
               StatusIs(tsl::error::INVALID_ARGUMENT));
 }
 
-TEST_F(TritonGemmTest, RejectInt4NegatePlusConvertHLO) {
+TEST_F(TritonGemmTest, Int4NegatePlusConvertHLO) {
   constexpr std::string_view kHloText = R"(
     HloModule t
 
@@ -178,8 +178,8 @@ TEST_F(TritonGemmTest, RejectInt4NegatePlusConvertHLO) {
           rhs_batch_dims={0}
     }
   )";
-  EXPECT_THAT(GetOptimizedModule(kHloText).status(),
-              StatusIs(tsl::error::INVALID_ARGUMENT));
+  EXPECT_TRUE(RunAndCompareNoHloPasses(
+      kHloText, ErrorSpec{/*aabs=*/1e-3, /*arel=*/1e-3}));
 }
 
 TEST_F(TritonGemmTest, RejectTritonFusionForInt4WithMinorBatchDim) {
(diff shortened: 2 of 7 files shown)


xla/service/cpu_gpu_shape_verifier.cc
```diff
@@ -75,7 +76,9 @@ absl::Status VerifyS4U4Usage(HloInstruction* instruction) {
     case HloOpcode::kDynamicUpdateSlice:
     case HloOpcode::kFusion:
     case HloOpcode::kGetTupleElement:
+    case HloOpcode::kOptimizationBarrier:
     case HloOpcode::kParameter:
+    case HloOpcode::kSelect:
```
bchetioui, line 81: Why is `select` here? It's an elementwise, right?

bchetioui (changes requested):
Do we know that we want to support all elementwise ops on `int4`? What's the support plan here---I imagine that codegen will currently fail for all these ops, right?

xla/service/cpu_gpu_shape_verifier.cc
sergey-kozub: Indeed, same as kBitcastConvert. Removed both.

sergey-kozub:
> What's the support plan here---I imagine that codegen will currently fail for all these ops, right?

Quite the opposite - this all works today (after this PR applied), I see no reason to keep this disabled.

bchetioui:
> > What's the support plan here---I imagine that codegen will currently fail for all these ops, right?
> 
> Quite the opposite - this all works today (after this PR applied), I see no reason to keep this disabled.

Awesome! Are there codegen tests already enabled?

sergey-kozub:
> Are there codegen tests already enabled?

I cannot find any tests, even though there is int4 specific code in e.g. "lower_tensors.cc"
I'll add a few basic tests to this PR that compare vs the interpreter.

sergey-kozub:
Added a test, PTAL.

xla/tests/int4_test.cc
```diff
@@ -157,6 +157,23 @@ XLA_TEST_F(HloTestBase, HorizontalLoopFusion) {
   EXPECT_TRUE(RunAndCompare(hlo_text, std::nullopt));
 }
 
+XLA_TEST_F(HloTestBase, ElementwiseOps) {
```
bchetioui, line 160: Could we make an exhaustive test that tests each and every unary, binary, and ternary elementwise op?

We could have three parametrized test suites.

bchetioui (changes requested):
PTAL at the comment :)

xla/tests/int4_test.cc
sergey-kozub: Added exhaustive tests for ops that support integer arguments.

xla/tests/int4_test.cc
```diff
@@ -186,5 +190,161 @@ XLA_TEST_F(HloTestBaseWithAlgsimpDisabled, TwoDots) {
 )";
   EXPECT_TRUE(RunAndCompare(hlo_text, std::nullopt));
 }
+
+class ElementwiseTest : public HloTestBase,
+                        public ::testing::WithParamInterface<
+                            std::tuple<HloOpcode, PrimitiveType>> {
+ public:
+  static std::vector<HloOpcode> GetOpcodesWithArity(int arity) {
```
bchetioui, line 198: `GetElementwiseOpcodesWithIntSupportWithArity`?

The name of the function right now is not telling the whole story.

xla/service/gpu/fusions/triton/triton_fusion_emitter_device_legacy_test.cc
```diff
@@ -162,26 +162,6 @@ TEST_F(TritonGemmTest, RejectDotInt4HLO) {
               StatusIs(tsl::error::INVALID_ARGUMENT));
 }
 
-TEST_F(TritonGemmTest, RejectInt4NegatePlusConvertHLO) {
```
bchetioui, line 165: What's the expectation here, that it's fused?

If so, should we check correctness of the lowering?

bchetioui (changes requested):
Thanks Sergey, just a few more things related to testing.

xla/tests/int4_test.cc
sergey-kozub: Updated the name as you suggest.

Note that the function name usually isn't telling the whole story, the function body does.

xla/service/gpu/fusions/triton/triton_fusion_emitter_device_legacy_test.cc
sergey-kozub: Restored the tests, removed "Reject" from the test name.

xla/tests/int4_test.cc
bchetioui: The function body shouldn't introduce behaviours that are not clear from the name + the docstring. In this case, the behaviour was quite different from what we could expect from just looking at the prototype---that is an antipattern.

bchetioui (approved):
Thank you Sergey!

PR #18883, opened 2024-10-31
Support S32 operand0 in ConditionalToSelect pass

apivovarov:
HLO Conditional operation has two forms:

Form1: Conditional(pred, true_operand, true_computation, false_operand, false_computation)
- binary selections

Form2: Conditional(branch_index, branch_computations, branch_operands)
- branch_index - tells what branch and operand to use
- branch_computations - list (can be larger than 2)
- branch_operands - list (can be larger than 2)

HLO operations semantics: https://openxla.org/xla/operation_semantics#conditional

The ConditionalToSelect pass currently lacks support for Form2 and incorrectly assumes operand0 is always of type PRED and the op has true_computation attr, leading to XLA crashes. This PR adds support for cases where operand0 is of type S32 and the op has branch_computations list.

Related issue: 
- https://github.com/jax-ml/jax/issues/12615

### DIFF
--- a/xla/service/conditional_to_select.cc
+++ b/xla/service/conditional_to_select.cc
@@ -30,10 +30,46 @@ limitations under the License.
 namespace xla {
 
 static absl::StatusOr<bool> DoConditionalToSelect(HloInstruction* conditional) {
+  if (conditional->shape().IsTuple()) {
+    VLOG(1) << "Not transforming tuples to 'select'";
+    return false;
+  }
+  bool is_form2 = false;
+  HloInstruction* true_operand;
+  HloInstruction* false_operand;
+  HloComputation* true_computation;
+  HloComputation* false_computation;
+
+  if (ShapeUtil::IsScalarWithElementType(conditional->operand(0)->shape(),
+                                         PrimitiveType::PRED)) {
+    // Form 1
+    // Conditional(pred, true_oprnd, true_comp, false_oprnd, false_comp)
+    true_operand = conditional->mutable_operand(1);
+    false_operand = conditional->mutable_operand(2);
+    true_computation = conditional->true_computation();
+    false_computation = conditional->false_computation();
+  } else if (ShapeUtil::IsScalarWithElementType(
+                 conditional->operand(0)->shape(), PrimitiveType::S32)) {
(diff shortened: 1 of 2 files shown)


xla/service/conditional_to_select.cc
```diff
@@ -30,29 +30,67 @@ limitations under the License.
...
+          conditional->shape(), {true_operand}, true_computation));
   conditional->SetupDerivedInstruction(if_call_op);
   HloInstruction* else_call_op =
       computation->AddInstruction(HloInstruction::CreateCall(
-          conditional->shape(), {conditional->mutable_operand(2)},
-          conditional->false_computation()));
+          conditional->shape(), {false_operand}, false_computation));
   conditional->SetupDerivedInstruction(else_call_op);
-  HloInstruction* condition = conditional->mutable_operand(0);
   if (else_call_op->shape().IsTuple()) {
```
akuegel, line 94: It looks to me that this can be checked earlier (by checking whether conditional->shape() is a tuple).

xla/service/conditional_to_select.cc
```diff
@@ -30,29 +30,67 @@ limitations under the License.
...
-  if (conditional->true_computation()->HasSideEffect() ||
-      conditional->false_computation()->HasSideEffect()) {
+  if (true_computation->HasSideEffect() || false_computation->HasSideEffect()) {
     VLOG(1) << "Not transforming conditional; branches have side effects:"
             << conditional->ToString();
     return false;
   }
 
   auto computation = conditional->parent();
+  if (is_form2) {
```
akuegel, line 78: I would prefer if this is moved closer to where condition variable is used. So essentially after original line 55 (and not deleting that line).

akuegel (commented):
Thanks for the fix. Looks good, just minor points.

xla/service/conditional_to_select.cc
apivovarov: fixed

xla/service/conditional_to_select.cc
apivovarov: moved it on top

akuegel (approved):

PR #18926, opened 2024-10-31
Add gpu_b100 gpu backend

hmonishN:
Added gpu_b100 gpu backend for Nvidia blackwell gpu. 
Added gpu_b100 to the backends field of test targets currently tagged with gpu_a100 and/or gpu_h100 to execute them on blackwell gpus in Nvidia CI. 
We may later remove some gpu_b100 tags if we realize the test targets should not run on Blackwell.


ddunl (approved):
LGTM, I can't tell off the top of my head whether we will need to patch this, but if so I'll handle it this week. Thanks!

ddunl:
Actually one thing I missed - can you update `build_tools/lint/tags.py` with the new tags?

hmonishN:
Added the documentation

ddunl:
Ok sorry one more thing! I think for now we need to wrap the usage of the gpu_b100 tag in `if_oss` because internal CI gets confused because it doesn't yet know about `requires-sm100-only` and we can remove them once internal CI catches up.

ddunl (approved):

ddunl:
sorry, last thing I think: https://github.com/openxla/xla/blob/main/build_tools/ci/build.py#L237 need to add 100 here I think to keep the GPU build working

ddunl (approved):
Thanks!! Sorry for the back and forth.

hmonishN:
No worries.

hmonishN:
Hi David, 
Is it good to merge?

ddunl:
Ah I forgot to check on CI again yesterday, now just need to update the goldens for the build script. it should be ```KOKORO_JOB_NAME=GOLDENS PYTHONDONTWRITEBYTECODE=1 python3 build.py > golden_commands.txt``` inside build_tools/ci

ddunl (approved):

PR #18959, opened 2024-11-01
[gpu] Fix the pipelined p2p bug

shraiysh:
The PipelinedP2PRewriter had two issues:

 1. For computations that did not have any collectives - the ProcessComputation function did not set the flag `collective_in_computation[computation]` to false. This caused crash on operations like `sort` with a no-collective computation.
 2. For fusion operations, the p2p rewriter assumed that there would be no collective emitted. In case of dynamic-slice-fusion, it may invoke a collective. So, changed this to handle it the same way for both fused and non-fused computations.

### DIFF
--- a/xla/service/gpu/transforms/pipelined_p2p_rewriter.cc
+++ b/xla/service/gpu/transforms/pipelined_p2p_rewriter.cc
@@ -667,6 +667,13 @@ absl::StatusOr<bool> ProcessComputation(
                                      instruction_sequence, hlo));
     changed = true;
   }
+  // After processing all the operations, if nothing set the flag
+  // collective_in_computation[computation] to true, then there must be no
+  // collective in this computation.
+  if (collective_in_computation.find(computation) ==
+      collective_in_computation.end()) {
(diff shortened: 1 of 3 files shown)


akuegel (approved):

xla/service/gpu/transforms/pipelined_p2p_rewriter.cc
```diff
@@ -667,6 +667,13 @@ absl::StatusOr<bool> ProcessComputation(
                                      instruction_sequence, hlo));
     changed = true;
   }
+  // After processing all the operations, if nothing set the flag
+  // collective_in_computation[computation] to true, then there must be no
+  // collective in this computation.
+  if (collective_in_computation.find(computation) ==
+      collective_in_computation.end()) {
+    collective_in_computation[computation] = false;
+  }
```
sanjoy, line 676: Nit: this could just be `collective_in_computation.insert({computation, false});`.  `insert` is a no-op if the key already exists in the `flat_hash_map`.

(If you do make this change please also add a comment that explaining this.)

PR #19038, opened 2024-11-05
Add ReducePrecision handler to algebraic_simplifier

apivovarov:
```
ReducePrecision(operand, mantissa_bits, exponent_bits)

Arguments        Type     Semantics
---------------------------------------------------------------------------
operand          XlaOp    array of floating-point type T.
exponent_bits    int32    number of exponent bits in lower-precision format
mantissa_bits    int32    number of mantissa bits in lower-precision format
```

If the target exponent bits and mantissa bits are equal to or greater than those of the operation's data type, the operation is a no-op and can be removed.

https://openxla.org/xla/operation_semantics#reduceprecision

_The number of exponent or mantissa bits may exceed the corresponding value for type T; the corresponding portion of the conversion is then simply a no-op._

### DIFF
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
@@ -5353,6 +5353,25 @@ absl::Status AlgebraicSimplifierVisitor::HandleConvert(
   return TryRemoveUpcastAndDowncastSurroundingBinaryOp(convert);
 }
 
+absl::Status AlgebraicSimplifierVisitor::HandleReducePrecision(
+    HloInstruction* hlo) {
+  HloReducePrecisionInstruction* reduce_precision =
+      Cast<HloReducePrecisionInstruction>(hlo);
+  int tgt_exponent_bits = reduce_precision->exponent_bits();
+  int tgt_mantissa_bits = reduce_precision->mantissa_bits();
+  PrimitiveType dtype = hlo->shape().element_type();
+  int dtype_exponent_bits = primitive_util::ExponentWidth(dtype);
+  int dtype_mantissa_bits = primitive_util::SignificandWidth(dtype) - 1;
+  if (tgt_exponent_bits >= dtype_exponent_bits &&
+      tgt_mantissa_bits >= dtype_mantissa_bits) {
+    // If the target exponent bits and mantissa bits are equal to or greater
+    // than those of the operation's data type, the operation is a no-op and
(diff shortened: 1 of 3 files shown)


akuegel (commented):
I already tried that myself in https://github.com/openxla/xla/pull/18338

Doesn't work, Jax is relying on ReducePrecisionOps to work as a blocker for downcast/upcast optimization, enabled by default via xla_allow_excess_precision flag.
But on GPU we make sure that those ops are actually not emitted, if they are the default.

apivovarov:
What you think if I add this to mhlo to https://github.com/openxla/xla/blob/main/xla/mlir_hlo/mhlo/IR/hlo_ops.cc ?
```
OpFoldResult ReducePrecisionOp::fold(FoldAdaptor) {
  ...
  if (tgt_exponent_bits >= dtype_exponent_bits &&
      tgt_mantissa_bits >= dtype_mantissa_bits) {
    return getOperand();
  }
  return {};
}
```

Smth similar to CopyOp
```
// CopyOp

OpFoldResult CopyOp::fold(FoldAdaptor) { return getOperand(); }
```

akuegel:
> What you think if I add this to mhlo to https://github.com/openxla/xla/blob/main/xla/mlir_hlo/mhlo/IR/hlo_ops.cc ?
> 
> ```
> OpFoldResult ReducePrecisionOp::fold(FoldAdaptor) {
>   ...
>   if (tgt_exponent_bits >= dtype_exponent_bits &&
>       tgt_mantissa_bits >= dtype_mantissa_bits) {
>     return getOperand();
>   }
>   return {};
> }
> ```
> 
> Smth similar to CopyOp
> 
> ```
> //===----------------------------------------------------------------------===//
> // CopyOp
> //===----------------------------------------------------------------------===//
> 
> OpFoldResult CopyOp::fold(FoldAdaptor) { return getOperand(); }
> ```

I also already tried it, but Jax is using MHLO when converting from stablehlo to HLO. So it would be folded too early. See also  https://github.com/openxla/xla/blob/685418e27732afd95b2167be702edbecc4411bb2/xla/mlir_hlo/mhlo/transforms/map_mhlo_to_scalar_op.h#L495

akuegel:
So as long as JAX is relying on the workaround (which is required as long as the flag xla_enable_excess_precision is set to true by default), I fear we cannot do any of the changes you are suggesting. Whether the flag can be turned off by default is another question. Therefore closing this PR.

PR #19067, opened 2024-11-05
[XLA:CPU][oneDNN] Move simplification pass before oneDNN pass

mahmoud-abuzaina:
This PR moves the simplification pass before oneDNN rewriter pass which simplifies the pattern matching for quantization support by getting rid of redundant copy ops.


xla/service/cpu/cpu_compiler.cc
```diff
@@ -795,6 +795,30 @@ absl::Status CpuCompiler::RunHloPassesAfterLayoutAssn(
 
   pipeline.AddPass<ReshapeDecomposer>();
 
+  // The LayoutAssignment pass may leave behind kCopy instructions which are
+  // duplicate or NOPs, so remove them with algebraic simplification and CSE.
+  // Run this to a fixed point.
+  [&pipeline = pipeline.AddPass<HloPassFix<HloPassPipeline>>(
```
dimitar-asenov, line 801: Why do we need this lambda here? Can't we just introduce a nested scope, like is done e.g. on line 783 above?

dimitar-asenov (changes requested):

xla/service/cpu/cpu_compiler.cc
mahmoud-abuzaina: I don't know if there was a reason to use lambda, I just moved existing code. I will change it to nested scope and see it works fine.

penpornk: It used to be a nested scope before and was changed to lambda capture to fix a bug:
https://github.com/openxla/xla/commit/1bb0c2e87858e4446c1caf2fb19ad56077780e00

Let's keep the lambda for now.

penpornk (changes requested):
Could you please add a unit test (in onednn_ test file) detecting the simplified pattern matching?

mahmoud-abuzaina:
Thank you for reviewing the PR! I have added a unit test.

dimitar-asenov (approved):

penpornk (approved):
Thank you very much!

sdasgup3:
The approvals looks stale and may that is way workflows are waiting.  
@penpornk Can you please approve this again?

dimitar-asenov (approved):

mahmoud-abuzaina:
Hi @penpornk, is there an update on the reproducer or more details on the failure?

mahmoud-abuzaina:
Hi @penpornk, is there an update about this PR?

ezhulenev:
can you please rebase this PR

mahmoud-abuzaina:
I have rebased it.

ezhulenev (approved):

dimitar-asenov (approved):

leroy182 (approved):

dimitar-asenov (approved):

dimitar-asenov:
@penpornk Could you please look into merging this PR? I believe you had some internal changes that you needed to do first so that this one merges cleanly.

penpornk:
@dimitar-asenov Will do. IIRC, I had landed that prelim change, but there were still some failures that needed to be investigated.

@mahmoud-abuzaina Sorry for the long delay! Could you please confirm that this PR is still needed by https://github.com/openxla/xla/pull/32959 ?

penpornk:
Chatted with the team on sync call. This PR is still needed. But we'll close it for now since it will need internal changes from me.

PR #18825, opened 2024-10-29
[GPU] GEMM fusions: let fusing effective parameters and their broadcasts in the epilogues.

sergachev:

PR #19112, opened 2024-11-06
[GPU] GEMM fusion: support more broadcasts.

sergachev:
Support broadcasts involving 1-sized fragments of dimensions like [1,n] -> broadcast -> [1,m,n].

### DIFF
--- a/xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
+++ b/xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
@@ -638,10 +638,13 @@ absl::StatusOr<Value> EmitBroadcast(ImplicitLocOpBuilder& b,
   Value expanded_input = tensor_input;
   int dim_idx = 0;
   for (const DimProperties& dim : side.tiled_dims) {
-    if (auto* spec = analysis->IterSpec(side.scope, &broadcast, dim.index);
-        spec != nullptr && spec->at(0).stride > 0) {
-      if (analysis->IterSpec(side.scope, broadcast.operand(0), dim.index) ==
-          nullptr) {
+    const auto* output_spec =
+        analysis->IterSpec(side.scope, &broadcast, dim.index);
+    if (output_spec != nullptr && output_spec->at(0).stride > 0) {
+      const auto* input_spec =
+          analysis->IterSpec(side.scope, broadcast.operand(0), dim.index);
+      if (input_spec == nullptr ||
+          output_spec->at(0).count > input_spec->at(0).count) {
         // Broadcasted dimension.
         expanded_input = b.create<mt::ExpandDimsOp>(expanded_input, dim_idx);
       }
@@ -1090,8 +1093,28 @@ class MatMulEmitterHelper {
     } else if (scope == TritonFusionAnalysis::Scope::RHS) {
       return dims_.rhs_noncontracting_dim_idx;
     } else {
-      CHECK(false) << "This shouldn't be called for the output scope.";
+      CHECK(false) << "This shouldn't be called for the other scopes.";
     }
+  }
+
+  bool IsNonTrivialTiledDimension(TritonFusionAnalysis::Scope scope,
+                                  int64_t dim_index) {
+    switch (scope) {
+      case TritonFusionAnalysis::Scope::LHS:
+        return (dim_index == dims_.lhs_noncontracting_dim_idx && dims_.m > 1) ||
+               (dim_index == dims_.lhs_contracting_dim_idx && dims_.k > 1);
+      case TritonFusionAnalysis::Scope::RHS:
+        return (dim_index == dims_.rhs_noncontracting_dim_idx && dims_.n > 1) ||
+               (dim_index == dims_.rhs_contracting_dim_idx && dims_.k > 1);
+      case TritonFusionAnalysis::Scope::OUTPUT:
+        return (dim_index == dims_.out_lhs_noncontracting_dim_idx &&
+                dims_.m > 1) ||
+               (dim_index == dims_.out_rhs_noncontracting_dim_idx &&
+                dims_.n > 1);
+      default:
+        break;
+    }
+    return false;
   }
 
   // Return the batch stride of the HLO passed as a parameter. If the
@@ -1207,8 +1230,18 @@ class MatMulEmitterHelper {
     }
 
     auto add_dim = [&](const DimProperties& properties) -> absl::Status {
-      if (analysis_.IterSpec(side.scope, hlo, properties.index) == nullptr) {
-        return absl::OkStatus();
+      {
+        const TensorIterationSpec::DimIterationSpec* spec =
+            analysis_.IterSpec(side.scope, hlo, properties.index);
+        if (spec == nullptr ||
+            (IsNonTrivialTiledDimension(side.scope, properties.index) &&
+             spec->size() == 1 && spec->at(0).count == 1)) {
+          // If a non-trivial tiled dimension has a trivial iteration spec at
+          // the parameter, it's being broadcasted. Skip it in the tensor
+          // pointer to prevent it from being padded to the tile size on load
+          // instead of being broadcasted.
+          return absl::OkStatus();
+        }
       }
       Value pid_offset =
           (properties.pid == nullptr)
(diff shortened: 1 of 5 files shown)


sergachev:
This should unblock https://github.com/openxla/xla/pull/18825.

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
```diff
@@ -637,10 +637,13 @@ absl::StatusOr<Value> EmitBroadcast(ImplicitLocOpBuilder& b,
...
-        spec != nullptr && spec->at(0).stride > 0) {
-      if (analysis->IterSpec(side.scope, broadcast.operand(0), dim.index) ==
-          nullptr) {
+    const auto* output_spec =
+        analysis->IterSpec(side.scope, &broadcast, dim.index);
+    if (output_spec != nullptr && output_spec->at(0).stride > 0) {
+      const auto* input_spec =
+          analysis->IterSpec(side.scope, broadcast.operand(0), dim.index);
+      if (input_spec == nullptr ||
+          output_spec->at(0).count > input_spec->at(0).count) {
```
bchetioui, line 646: Could you clarify when that is true, both in a comment and in the CL message?

It's not quite clear what "more broadcasts" means here. Can you qualify the "more"?

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
```diff
@@ -1206,8 +1229,18 @@ class MatMulEmitterHelper {
...
     auto add_dim = [&](const DimProperties& properties) -> absl::Status {
-      if (analysis_.IterSpec(side.scope, hlo, properties.index) == nullptr) {
-        return absl::OkStatus();
+      {
+        const TensorIterationSpec::DimIterationSpec* spec =
+            analysis_.IterSpec(side.scope, hlo, properties.index);
+        if (spec == nullptr ||
+            (IsNonTrivialTiledDimension(side.scope, properties.index) &&
+             spec->size() == 1 && spec->at(0).count == 1)) {
+          // If a non-trivial tiled dimension has a trivial iteration spec at
```
bchetioui, line 1238: Can you define formally what a trivial iteration spec is? Is it a spec with size 1 and count 1?

Maybe make a helper `IsTrivialIterationSpec` that documents the expectations & meaning?

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
```diff
@@ -1089,8 +1092,28 @@ class MatMulEmitterHelper {
     } else if (scope == TritonFusionAnalysis::Scope::RHS) {
       return dims_.rhs_noncontracting_dim_idx;
     } else {
-      CHECK(false) << "This shouldn't be called for the output scope.";
+      CHECK(false) << "This shouldn't be called for the other scopes.";
+    }
+  }
+
+  bool IsNonTrivialTiledDimension(TritonFusionAnalysis::Scope scope,
```
bchetioui, line 1099: nit: it would be slightly nicer to flip the check around to avoid the negation (i.e. make this `IsTrivialTiledDimension`)

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
```diff
@@ -1400,7 +1433,7 @@ class MatMulEmitterHelper {
     if (dims_.out_split_k_dim_idx.has_value()) {
       const TensorIterationSpec::DimIterationSpec* spec = analysis_.IterSpec(
           TritonFusionAnalysis::Scope::OUTPUT, hlo, *dims_.out_split_k_dim_idx);
-      if (spec != nullptr) {
+      if (spec != nullptr && spec->at(0).count > 1) {
```
bchetioui, line 1436: What changes in the `split-k` case?

xla/service/gpu/triton_tiling_propagation.cc
```diff
@@ -786,7 +786,9 @@ DimOrderMapOrError GetPropagatedDimOrdersForDimAlteringOp(
       for (const Fragment* subdim : dst_logical[dim_idx]) {
         dst_fragments_order.push_back(*subdim);
         src_to_dst[subdim] = dst_fragments_order.size() - 1;
-        dim_numbers_present_in_dst.insert(subdim->dst_dim_number());
+        if (subdim->full_count() > 1) {
```
bchetioui, line 789: Why is this correct? Does the dimension not appear in `dst`, or is it just trivial? (`dim_numbers_present_in_dst` seems to indicate that this condition only makes sense if the former is true; if so, please document why. If not, then we need to think about how to change the variable names and/or add comments to make the expectation clear.)

xla/service/gpu/triton_fusion_analysis_test.cc
```diff
@@ -582,6 +582,42 @@ ENTRY e {
                                     /*subfragments=*/ElementsAre(4))));
 }
 
+TEST_F(TritonDotAnalysisTest, BroadcastFromTriviallySizedDimensionIsSupported) {
```
bchetioui, line 585: Do we have negative tests for the broadcasts that should be filtered out? It's a bit hard to gauge whether the code is as tight as it should be.

bchetioui (changes requested):

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
sergachev: I don't think so because out of 4 combinations (trivial/non-trivial x tiled/non-tiled) we are only interested in this one: tiled and non-trivial.

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
sergachev: I'll reformulate to say that the dimension has only one element to avoid new terms then.

xla/service/gpu/triton_tiling_propagation.cc
sergachev: > Does the dimension not appear in dst, or is it just trivial?

It's trivial. Updated for clarity.

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
sergachev: With split-K in the epilogue of the fusion there are input tensors which do not have the split-K-batch dimension and output tensors which do. Only the output ones should be iterated over along this dimension. EmitTensorPointer() is used for both. Before this change the split-K-batch dimension would never appear at the input tensor iteration specs, so it was enough to check only for its presence. Now it can appear, with a trivial spec (count=1), because fused broadcasts with trivial dimensions are now supported.

sergachev: The added test BroadcastsOfTriviallySizedNonContractingDimensionsAreSupported will fail without this change, demonstrating the problem.

xla/service/gpu/fusions/triton/triton_fusion_emitter_legacy_matmul.cc
sergachev: Done.

xla/service/gpu/triton_fusion_analysis_test.cc
sergachev: Added BroadcastWithinDimensionIsNotSupported.

bchetioui (approved):

reedwm:
This is failing internally. It fails with `-c opt --copt=-UNDEBUG` but not just `-c opt`, so I think you'll be able to reproduce by adding `--copt=-UNDEBUG` to the bazel test command when testing `triton_fusion_emitter_device_legacy_test.cc`.

The error is

```
[ RUN      ] TritonGemmTest.BroadcastsOfTriviallySizedContractingDimensionsAreSupported
F1112 19:18:49.659305    5238 logging.cc:62] assert.h assertion failed at third_party/triton/lib/Dialect/Triton/Transforms/RewriteTensorPointer.cpp:259 in Operation *RewriteTensorPointerPass::rewriteAdvanceOp(OpBuilder &, triton::AdvanceOp, std::stack<Operation *> &): info.length() == op.getOffsets().size()
```

sergachev:
> This is failing internally. It fails with `-c opt --copt=-UNDEBUG` but not just `-c opt`, so I think you'll be able to reproduce by adding `--copt=-UNDEBUG` to the bazel test command when testing `triton_fusion_emitter_device_legacy_test.cc`.

Thanks! Fixed.

bchetioui (approved):

PR #19160, opened 2024-11-07
[ds-fusion]Add support for fusion operations in GetLoopInductionVariableIndex

shraiysh:
After optimizations, while loop analysis was unable to handle the fused operations, primarily for GetLoopInductionVariableIndex and ComputeWhileLoopTripCount. This patch handles only GetloopInductionVariableIdx.

If the GetGTEOperandIndex function fails to find a precise GTE operand, we try to deduce the unique dependence by using the HloExtractor.

### DIFF
--- a/xla/hlo/analysis/BUILD
+++ b/xla/hlo/analysis/BUILD
@@ -145,10 +145,12 @@ cc_library(
         "//xla/hlo/utils:hlo_query",
         "//xla/service:collective_ops_utils",
         "//xla/service:pattern_matcher",
+        "//xla/tools:hlo_extractor",
         "@com_google_absl//absl/base",
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/log",
         "@com_google_absl//absl/log:check",
+        "@com_google_absl//absl/algorithm:container",
     ],
 )
 
--- a/xla/hlo/analysis/while_loop_analysis.cc
+++ b/xla/hlo/analysis/while_loop_analysis.cc
@@ -18,7 +18,10 @@ limitations under the License.
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
+#include <memory>
+#include <optional>
 
+#include "absl/algorithm/container.h"
 #include "absl/base/casts.h"
 #include "absl/container/flat_hash_map.h"
 #include "absl/log/check.h"
@@ -35,6 +38,7 @@ limitations under the License.
 #include "xla/service/collective_ops_utils.h"
 #include "xla/service/pattern_matcher.h"
 #include "xla/shape_util.h"
+#include "xla/tools/hlo_extractor.h"
 
 namespace xla {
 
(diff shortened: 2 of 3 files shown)


shraiysh:
This is another attempt at #18765. There is a fast path now, that relies on exact pattern matching. If the pattern matching is successful, then the slower version of extraction is not used. This should help with the timeouts.

akuegel (changes requested):
Maybe the fast path would already help for those timeouts we have seen, but I think you are still not guarding about a while loop conditional that is conceptually like this:

Shape s32[10000000]
while loop body does divide by 2
while loop condition does a reduction and checks whether the sum of values in the tensor is 0.
This will be a quite expensive conditional computation that we probably don't want to evaluate during while loop analysis. My suggested fix is to never allow calling the Evaluator if there is any non-scalar shape involved. Can you please add a guard like that?

shraiysh:
> Maybe the fast path would already help for those timeouts we have seen, but I think you are still not guarding about a while loop conditional that is conceptually like this:
> 
> Shape s32[10000000] while loop body does divide by 2 while loop condition does a reduction and checks whether the sum of values in the tensor is 0. This will be a quite expensive conditional computation that we probably don't want to evaluate during while loop analysis. My suggested fix is to never allow calling the Evaluator if there is any non-scalar shape involved. Can you please add a guard like that?

I am not making any changes to the evaluation workflow for the computation of while loop -- that is done in the while loop trip count functions, that I have not edited (except making them stricter with the pattern they were already expecting). This patch does not actually execute the HLO instructions. This only deals with extracting the HLO and deciding whether the only dependence is on the induction variable or not. I thought I will do the changes for evaluation in a separate patch so that this lands in a clean fashion.

akuegel:
> I am not making any changes to the evaluation workflow for the computation of while loop -- that is done in the while loop trip count functions, that I have not edited (except making them stricter with the pattern they were already expecting). This patch does not actually execute the HLO instructions. This only deals with extracting the HLO and deciding whether the only dependence is on the induction variable or not. I thought I will do the changes for evaluation in a separate patch so that this lands in a clean fashion.

I think your change makes it much more likely that we will evaluate the condition computation. It is quite likely that we didn't need any guards so far because not supporting fusion instructions already acted as a guard. I agree that maybe it would be better to add such a guard in a separate PR, but that one would have to land first before we can land this one.

shraiysh:
> I think your change makes it much more likely that we will evaluate the condition computation. It is quite likely that we didn't need any guards so far because not supporting fusion instructions already acted as a guard. I agree that maybe it would be better to add such a guard in a separate PR, but that one would have to land first before we can land this one.

Added it in #19194.

akuegel:
> > I think your change makes it much more likely that we will evaluate the condition computation. It is quite likely that we didn't need any guards so far because not supporting fusion instructions already acted as a guard. I agree that maybe it would be better to add such a guard in a separate PR, but that one would have to land first before we can land this one.
> 
> Added it in #19194.

Thank you, I will review it on Monday, today it is already end of work time for me.

fhoushmand (approved):

akuegel:
I verified that this PR now does not trigger the timeouts anymore. Also, I found out why we were hitting the timeouts before. HloExtractor has a performance bug in FinishVisit() method, I am fixing that now and that alone would have already fixed the issue. But still good that we have the other guard, too.

shraiysh:
> I verified that this PR now does not trigger the timeouts anymore. Also, I found out why we were hitting the timeouts before. HloExtractor has a performance bug in FinishVisit() method, I am fixing that now and that alone would have already fixed the issue. But still good that we have the other guard, too.

Oh, that's good to know. I suspected that there could be issues with HloExtractor. Thanks for looking into it :)

akuegel:
> Do we know why this was reverted @akuegel @fhoushmand ?

This time it was not due to timeouts, but due to a verifier error. It points to a shape mismatch, so probably a missing check somewhere. I will try to extract a reproducer.

akuegel:
> > Do we know why this was reverted @akuegel @fhoushmand ?
> 
> This time it was not due to timeouts, but due to a verifier error. It points to a shape mismatch, so probably a missing check somewhere. I will try to extract a reproducer.

So far I narrowed it down to the WhileLoopFusibleSinking pass. When I disable that pass, the problem goes away, so I guess this pass has some assumption that only holds for trivial pattern matching. Will now look at the code of the pass if I see something.

akuegel:
Rolled forward in https://github.com/openxla/xla/commit/f25e9a00b48a083ede7f2075f16ca50aebf280a0

PR #19161, opened 2024-11-07
Asymmetrically Replicated Instructions in Replication Analysis

philipphack:
Extends the HLO replication analysis to handle asymmetrically replicated instructions with replica groups covering multiple partitions and replicas.

### DIFF
--- a/xla/hlo/analysis/hlo_replication_analysis.cc
+++ b/xla/hlo/analysis/hlo_replication_analysis.cc
@@ -37,6 +37,81 @@ limitations under the License.
 #include "xla/xla_data.pb.h"
 
 namespace xla {
+namespace {
+// When cross_partition_spmd is true, returns the partition IDs of all
+// replica groups in which a given replica participates. Specfically, the k-th
+// element of the outermost vector in the returned data structure holds the
+// partition IDs converted from the global IDs in a collective's
+// replica_groups field for replica k.
+//
+// When cross_partition_spmd is false, returns the replica IDs of all
+// replica groups in which a given partition participates. Specfically, the k-th
+// element of the outermost vector in the returned data structure holds the
+// replica IDs converted from the global IDs in a collective's replica_groups
+// field for partition k.
+std::vector<std::vector<std::vector<int64_t>>> GroupsForReplicas(
+    absl::Span<const ReplicaGroup> groups, int64_t num_partitions,
+    int64_t replica_count, bool cross_partition_spmd) {
+  std::vector<std::vector<std::vector<int64_t>>> groups_for_replicas(
+      cross_partition_spmd ? replica_count : num_partitions);
+  int64_t num_replicas = cross_partition_spmd ? replica_count : num_partitions;
+  for (const ReplicaGroup& group : groups) {
+    for (int64_t idx = 0; idx < num_replicas; ++idx) {
+      std::vector<int64_t> ids;
+      for (int64_t id : group.replica_ids()) {
+        int64_t rid = id / num_partitions;
+        int64_t pid = id % num_partitions;
+        if ((cross_partition_spmd ? rid : pid) == idx) {
+          ids.push_back(cross_partition_spmd ? pid : rid);
+        }
+      }
+      if (!ids.empty()) {
+        groups_for_replicas[idx].emplace_back(std::move(ids));
+      }
+    }
+  }
+  return groups_for_replicas;
+}
+
+// Returns a vector of vectors such that two ints appear in the same inner
+// vector iff they appear in the same inner vector of both groups0 and groups1.
+//
+// Example:
+//
+//  groups0 = {{0,1,2,3},{4,5,6,7}}
+//  groups1 = {{0,3,4,5},{1,2,6,7}}
+//
+//  MergeGroups(groups0, groups1) returns {{0,3},{1,2},{4,5},{6,7}}.
+std::vector<std::vector<int64_t>> MergeGroups(
+    absl::Span<const std::vector<int64_t>> groups0,
+    absl::Span<const std::vector<int64_t>> groups1) {
+  auto groups_idx_for_target = [](absl::Span<const std::vector<int64_t>> groups,
+                                  int64_t target) -> int64_t {
+    for (int groups_idx = 0; groups_idx < groups.size(); ++groups_idx) {
+      for (int64_t element : groups[groups_idx]) {
+        if (element == target) {
+          return groups_idx;
+        }
+      }
+    }
+    LOG(FATAL) << "Unable to find target.";
+  };
+
+  std::vector<std::vector<int64_t>> merged_groups;
+  absl::flat_hash_map<std::pair<int64_t, int64_t>, std::vector<int64_t>>
+      groups_idxs_to_elements;
+  for (int groups0_idx = 0; groups0_idx < groups0.size(); ++groups0_idx) {
+    for (int64_t element : groups0[groups0_idx]) {
+      int64_t groups1_idx = groups_idx_for_target(groups1, element);
+      groups_idxs_to_elements[{groups0_idx, groups1_idx}].push_back(element);
+    }
+  }
+  for (const auto& [pair, group] : groups_idxs_to_elements) {
+    merged_groups.emplace_back(std::move(group));
+  }
+  return merged_groups;
+}
+}  // namespace
 
 // Determines whether an HLO instruction is replicated at index based on current
 // knowledge in hlo_replication. When cross_partition_spmd is true, the
@@ -50,10 +125,6 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
     const absl::flat_hash_map<const HloInstruction*, ShapeTree<HloReplication>>&
         hlo_replication,
     bool support_partial_replication) {
-  struct IDs {
-    int64_t pid;
-    int64_t rid;
-  };
   const auto merge_operand_replication = [&hlo_replication](
                                              const HloInstruction* inst) {
     HloReplication replication = HloReplication::ReplicatedOnAllDevices();
@@ -85,7 +156,10 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
         return HloReplication::ReplicatedOnAllDevices();
       }
       if (support_partial_replication) {
-        std::vector<std::vector<std::vector<int64_t>>> groups_for_replicas(1);
+        const int64_t num_partitions =
+            hlo->GetModule()->config().num_partitions();
+        std::vector<std::vector<std::vector<int64_t>>> groups_for_replicas(
+            num_partitions);
         for (const ReplicaGroup& replica_group : hlo->replica_groups()) {
           std::vector<int64_t> group_for_replica;
           for (auto id : replica_group.replica_ids()) {
@@ -93,6 +167,8 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
           }
           groups_for_replicas[0].push_back(group_for_replica);
         }
+        std::fill(groups_for_replicas.begin() + 1, groups_for_replicas.end(),
+                  groups_for_replicas.front());
         return HloReplication::PartiallyReplicated(groups_for_replicas);
       } else {
         return HloReplication::UniqueOnAllDevices();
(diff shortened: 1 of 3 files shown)


xla/hlo/analysis/hlo_replication_analysis.cc
```diff
@@ -94,45 +105,92 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
...
           }
-          replicated_across_partitions &=
-              visited_partitions.size() == num_partitions;
-          replicated_across_replicas &=
-              visited_replicas.size() ==
-              hlo->GetModule()->config().replica_count();
           if (support_partial_replication) {
-            device_sets_storage.push_back(device_set);
-            device_sets.push_back(device_sets_storage.back());
+            device_sets.push_back(std::move(device_set));
```
reedwm, line 135: This almost works, but it will potentially fail in the case where cross_partition_spmd=true if a device_sets contains different replicas. For example, I think the following case fails by having HloInstructionIsReplicatedAt incorrectly return true.

num_partitions = 2
num_replicas = 2
cross_partition_spmd = true
all gather replica_groups = {{0,3}{1,2}}
replica_groups passed to HloInstructionIsReplicatedAt = {{0,1}}

I haven't personally tried this case, but the issue is that device_sets will be {{0,1},{0,1}}. It will seem replicated since the two sets each have both partitions: 0 and 1. But they are on different replicas, and so are not truly replicated.

---

I'll describe the algorithm I think you should implement. @pkanwar23 helped me come up with it. I will use this example when describing it:

num_partitions = 6
num_replicas = 2
cross_partition_spmd = true
all gather replica_groups = {{0,1,4,5,6,10},{2,3,7,8,9,11}

The algorithm is as follows. Let `M = num_partitions, N = num_replicas`. In DetermineHloInstructionIsReplicated, create an array `int id_to_group[M][N]`, such that `id_to_group[m][n]` maps device `(m, n)` to the replica_group that device appears in. For example, the array in the example above would be (I use A to refer to replica_group 0 and B to refer to replica_group 1):

| | replica 0  | replica 1 |
| --- | ------------- | ------------- |
| **partition 0** | A  | A  |
| **partition 1** | A  | B  |
| **partition 2** | B  | B  |
| **partition 3** | B  | B  |
| **partition 4** | A  | A  |
| **partition 5** | A  | B  |

Store `id_to_group`  inside PartiallyReplicated and return it from `DetermineHloInstructionIsReplicated`

Within `IsReplicatedWithinSubgroup`, we are given a `replica_groups` parameter. We must use this and the `id_to_group` table to determine if for each replica, the partitions are replicated according to `replica_groups`. I'll describe how with a concrete example. Suppose the `replica_groups` parameter is `{0,4},{1,5},{2,3}`. We process each of the three groups separately, returning false if any group is non-replicated. For `{0,4}`, we check that rows 0 and 4 in the table are identical. In this case, they are both `(A, A)`, so we do not return false. Similarly, rows 1 and 5 are the same (both `(A, B)`), and rows 2 and 3 are the same (both `(B, B)`). So all groups are replicated, and we return true.

Now suppose `replica_groups` is instead `{0,4,5},{1,2,3}`. In this case, rows 4 and 5 within the first group are not the same , so we return false.

xla/hlo/analysis/hlo_replication_analysis.cc
```diff
@@ -39,14 +39,21 @@ limitations under the License.
 namespace xla {
 
 // Determines whether an HLO instruction is replicated at index based on current
-// knowledge in hlo_replication.
+// knowledge in hlo_replication. When cross_partition_spmd is true, the
+// instruction must be replicated across all partitions on each replica.
+// Similarly, when cross_partition_spmd is false, the instruction must be
+// replicated across all replicas on ecah partition.
```
reedwm, line 45: Typo: "ecah"

xla/hlo/analysis/hlo_replication_analysis_test.cc
```diff
@@ -789,6 +795,177 @@ ENTRY entry {
...
+          module.get(),
+          /*cross_partition_spmd=*/true));
+
+  std::array<ReplicaGroup, 3> replica_groups0;
+  replica_groups0[0].add_replica_ids(0);
+  replica_groups0[0].add_replica_ids(1);
+  replica_groups0[1].add_replica_ids(2);
+  replica_groups0[1].add_replica_ids(3);
+  replica_groups0[2].add_replica_ids(4);
+  replica_groups0[2].add_replica_ids(5);
```
reedwm, line 825: Create a helper function making it easy to create a vector of ReplicaGroups on one line so you can do

```C++
std::vector<ReplicaGroup> replica_groups0 = CreateReplicaGroups({{0,1},{2,3},{4,5}});
```

xla/hlo/analysis/hlo_replication_analysis_test.cc
```diff
@@ -789,6 +795,177 @@ ENTRY entry {
...
+
+TEST_F(
+    HloReplicationAnalysisTest,
+    PartiallyReplicatedAllGatherFlattenedIDReplicationAnalysisAsymmetricGroups) {
+  const std::string module_str = R"(
+HloModule GlobalIdAllGather
+
+ENTRY entry {
+  param = f32[1] parameter(0)
+  ROOT all_gather = f32[6] all-gather(param), replica_groups={{0,1,2,3,4,6},{5,7,8,9,10,11}}, dimensions={0}, use_global_device_ids=true, channel_id=1
```
reedwm, line 849: The added functions are all almost identical. To avoid verbosity, I would have have a helper function `CollectiveIsReplicated` so you could do, e.g.

```c++
EXPECT_FALSE(CollectiveIsReplicated(
    /*collective_replica_groups=*/{{0,1,2,3,4,6},{5,7,8,9,10,11}}, 
    /*replica_count=*/6, 
    /*num_partitions=*/2, 
    /*cross_partition_spmd=*/false, 
    /*analysis_replica_groups=*/{{0,1},{2,3},{4,5}}));
```

reedwm (changes requested):

xla/hlo/analysis/hlo_replication_analysis.cc
philipphack: Thanks for the suggestion! As discussed, I made a small change to the existing implementation which should cover the case you mentioned.

reedwm (changes requested):

xla/hlo/analysis/hlo_replication_analysis.cc
```diff
@@ -37,6 +37,81 @@ limitations under the License.
...
+      for (int64_t element : groups[groups_idx]) {
+        if (element == target) {
+          return groups_idx;
+        }
+      }
+    }
+    LOG(FATAL) << "Unable to find target.";
+  };
+
+  std::vector<std::vector<int64_t>> merged_groups;
```
reedwm, line 100: Move this right before the for-loop where it's used

xla/hlo/analysis/hlo_replication_analysis.cc
```diff
@@ -37,6 +37,81 @@ limitations under the License.
...
+// replica groups in which a given partition participates. Specfically, the k-th
+// element of the outermost vector in the returned data structure holds the
+// replica IDs converted from the global IDs in a collective's replica_groups
+// field for partition k.
+std::vector<std::vector<std::vector<int64_t>>> GroupsForReplicas(
+    absl::Span<const ReplicaGroup> groups, int64_t num_partitions,
+    int64_t replica_count, bool cross_partition_spmd) {
+  std::vector<std::vector<std::vector<int64_t>>> groups_for_replicas(
+      cross_partition_spmd ? replica_count : num_partitions);
+  int64_t num_replicas = cross_partition_spmd ? replica_count : num_partitions;
```
reedwm, line 57: Swap these two lines, and use `num_replicas` in the definition of `groups_for_replicas`

xla/hlo/analysis/hlo_replication_analysis.cc
```diff
@@ -78,11 +156,20 @@ HloReplicationAnalysis::DetermineHloInstructionIsReplicated(
         return HloReplication::ReplicatedOnAllDevices();
       }
       if (support_partial_replication) {
-        std::vector<absl::Span<const int64_t>> device_sets;
+        const int64_t num_partitions =
+            hlo->GetModule()->config().num_partitions();
+        std::vector<std::vector<std::vector<int64_t>>> groups_for_replicas(
+            num_partitions);
```
reedwm, line 162: I think it's slightly more elegant to create a doubly-nested vector, std::vector<std::vector<int64_t>>, fill it out, then call `std::fill` on it. You can name it `groups`

reedwm (changes requested):

reedwm (approved):

golechwierowicz:
@philipphack, looks like it breaks `WhileLoopAllReduceCodeMotionTest.DynamicSliceAllReduceDynamicUpdateSliceAccumulate`.

reedwm (approved):

reedwm:
As we discussed offline -- unfortunately this causes an internal model to have a regression in compile time, on 8192 devices. It goes from taking 2 minutes to 19 minutes to compile. One issue is the `GroupsForReplicas` has an O(N^2) algorithm. By repalcing the function with the following, `GroupsForReplicas` goes from about .25 seconds a call to .0005 seconds a call.

```c++
std::vector<std::vector<std::vector<int64_t>>> GroupsForReplicas(
    absl::Span<const ReplicaGroup> groups, int64_t num_partitions,
    int64_t replica_count, bool cross_partition_spmd) {
  LOG(INFO) << "RRR Starting GroupsForReplicas, num_partitions=" << num_partitions << ", replica_count=" << replica_count << ", cross_partition_spmd=" << cross_partition_spmd;
  int64_t num_replicas = cross_partition_spmd ? replica_count : num_partitions;
  std::vector<std::vector<std::vector<int64_t>>> groups_for_replicas(
      num_replicas);
  for (const ReplicaGroup& group : groups) {
    for (int idx = 0; idx < num_replicas; ++idx) {
      if (groups_for_replicas[idx].empty() || !groups_for_replicas[idx].back().empty()) {
        groups_for_replicas[idx].push_back({});
      }
    }
    for (int64_t id : group.replica_ids()) {
      int64_t rid = id / num_partitions;
      int64_t pid = id % num_partitions;
      int64_t outer_id = cross_partition_spmd ? rid : pid;
      int64_t inner_id = cross_partition_spmd ? pid : rid;
      groups_for_replicas[outer_id].back().push_back(inner_id);
    }
  }
  for (int idx = 0; idx < num_replicas; ++idx) {
    if (!groups_for_replicas[idx].empty() && groups_for_replicas[idx].back().empty()) {
      groups_for_replicas[idx].pop_back();
    }
  }

  LOG(INFO) << "RRR Finished GroupsForReplicas";
  return groups_for_replicas;
}
```

But, the HLO still takes 13 minutes to compile. I'll either get you some HLO to reproduce, or keep debugging later.

reedwm:
As further discussed offline, we can improve performance further by replacing functions `HloReplication::Equal` and `MergeGroups` with the following:

```c++
bool HloReplicationAnalysis::HloReplication::Equal(
    const HloReplication& other) const {
  if (state_ != other.state_) {
    return false;
  }
  // LOG(INFO) << "RRR Starting HloReplication::Equal, groups_for_replicas_.size()=" << groups_for_replicas_.size();
  // LOG(INFO) << "RRR Starting HloReplication::Equal";
  CHECK_EQ(groups_for_replicas_.size(), other.groups_for_replicas_.size());
  for (int k = 0; k < groups_for_replicas_.size(); ++k) {
    // LOG(INFO) << "RRR Starting HloReplication::Equal for replica " << k << " with " << groups_for_replicas_[k].size() << " groups";
    std::set<std::vector<int64_t>> groups_for_replica_set(
        groups_for_replicas_[k].begin(), groups_for_replicas_[k].end());
    std::set<std::vector<int64_t>> other_groups_for_replica_set(
        other.groups_for_replicas_[k].begin(),
        other.groups_for_replicas_[k].end());
    if (groups_for_replica_set != other_groups_for_replica_set) {
      // LOG(INFO) << "RRR Finished HloReplication::Equal";
      return false;
    }
  }
  // LOG(INFO) << "RRR Finished HloReplication::Equal";
  return true;
}
```

```c++
std::vector<std::vector<int64_t>> MergeGroups(
    absl::Span<const std::vector<int64_t>> groups0,
    absl::Span<const std::vector<int64_t>> groups1) {
  // LOG(INFO) << "RRR Starting MergeGroups, groups0.size=" << groups0.size() << ", groups1.size=" << groups1.size();
  std::unordered_map<int64_t, int64_t> groups1_elem_to_groups_idx;
  for (int groups1_idx = 0; groups1_idx < groups1.size(); ++groups1_idx) {
    for (int64_t element : groups1[groups1_idx]) {
      groups1_elem_to_groups_idx[element] = groups1_idx;
    }
  }
  auto groups_idx_for_target = [&groups1_elem_to_groups_idx](absl::Span<const std::vector<int64_t>> groups,
                                  int64_t target) -> int64_t {
    CHECK(groups1_elem_to_groups_idx.count(target));
    return groups1_elem_to_groups_idx[target];
    LOG(FATAL) << "Unable to find target.";
  };

  absl::flat_hash_map<std::pair<int64_t, int64_t>, std::vector<int64_t>>
      groups_idxs_to_elements;
  for (int groups0_idx = 0; groups0_idx < groups0.size(); ++groups0_idx) {
    for (int64_t element : groups0[groups0_idx]) {
      int64_t groups1_idx = groups_idx_for_target(groups1, element);
      groups_idxs_to_elements[{groups0_idx, groups1_idx}].push_back(element);
    }
  }
  std::vector<std::vector<int64_t>> merged_groups;
  for (const auto& [pair, group] : groups_idxs_to_elements) {
    merged_groups.emplace_back(std::move(group));
  }
  // LOG(INFO) << "RRR Finished MergeGroups";
  return merged_groups;
}
```

This new functions improve the big-O complexity to O(N) from O(N^2), where N is the number of devices (8192 in the internal model). But even with these new function, compilation take 2 minutes 30 seconds to compile, vs 1 minute 58 seconds without this PR.

The profile for the clean runs show HloReplicationAnalysis taking about 1.76% of the time, with about half of that in Merge. The PR run show it taking 25% of the time, with slightly more than half of the time in Merge and Equal. Unfortunately, I think we'll need much more optimization work. This PR fixes existing correctness issues, so its important to get it in, but I think this is doable without sacrificing compile-time performance.

I'll see if I can get a small example to reproduce this issue, which will make it easier to debug and additionally allow you to debug.

reedwm (approved):

reedwm:
I'm currently working on merging this. As discussed, I'm changing the data structure in `HloReplication` from `std::vector<absl::flat_hash_set<std::vector<int64_t>>>` to `std::vector<std::vector<int64_t>>` when merging. The pending commit with these changes is https://github.com/openxla/xla/pull/20347/commits/e936938439dd81a4c4437223278bd33635be7b8c. The new data structure stores, in `device_set_root_per_replica_[i][j]`, the first partition ID in the group that partition `j` is part of, within replica `i`. The algorithms in the implementation have been changed accordingly. This fixes the compile-time regression in the internal model I mentioned above.

I think the new data structure is relatively simple and reasonable, but let me know if you disagree.

There's also an issue in `collective_quantizer_test.cc` in that it didn't specify the number of partitions, causing it to default to 1, which caused a crashed in HLO replication analysis since it assumed partition IDs are never out-of-bounds. I'll fix collective_quantizer_test.cc in the merge commit as well.

PR #19194, opened 2024-11-08
Add guard for scalar operations when brute-forcing while loop analysis

shraiysh:
If `MatchTrivialLoopTripCount`, fails we brute force the while loop with `ComputeWhileLoopTripCount`. In this patch we add a guard to ensure that we only brute-force the while loop when the operation is a scalar operation or a collection of scalar operations (via call, or fusion).

### DIFF
--- a/xla/hlo/analysis/while_loop_analysis.cc
+++ b/xla/hlo/analysis/while_loop_analysis.cc
@@ -29,8 +29,10 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_module.h"
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/hlo/ir/hlo_reachability.h"
+#include "xla/hlo/utils/hlo_query.h"
 #include "xla/literal.h"
 #include "xla/literal_util.h"
+#include "xla/service/collective_ops_utils.h"
 #include "xla/service/pattern_matcher.h"
 #include "xla/shape_util.h"
 
@@ -42,18 +44,18 @@ namespace m = match;
 
 // Finds and returns the non-constant operand in instr.
 //
-// CHECK-fails if instr doesn't have exactly one unique non-constant operand.
+// If the instruction doesn't have exactly one unique non-constant operand,
+// nullptr is returned.
 static const HloInstruction* NonConstantOperand(const HloInstruction* instr) {
   const HloInstruction* result = nullptr;
   for (const HloInstruction* operand : instr->operands()) {
     if (!operand->IsConstant()) {
-      if (result != nullptr) {
-        CHECK_EQ(result, operand);
+      if (result != nullptr && result != operand) {
+        return nullptr;
       }
       result = operand;
     }
   }
-  CHECK_NE(result, nullptr);
   return result;
 }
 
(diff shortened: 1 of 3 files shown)


shraiysh:
This should unblock #19160.

xla/hlo/analysis/while_loop_analysis.cc
```diff
@@ -47,13 +49,12 @@ static const HloInstruction* NonConstantOperand(const HloInstruction* instr) {
   const HloInstruction* result = nullptr;
   for (const HloInstruction* operand : instr->operands()) {
     if (!operand->IsConstant()) {
-      if (result != nullptr) {
-        CHECK_EQ(result, operand);
```
akuegel, line 51: Please also update the comment in line 47

xla/hlo/analysis/while_loop_analysis.cc
```diff
@@ -354,6 +355,29 @@ optional<int64_t> CheckedSubtract(int64_t a, int64_t b) {
...
+  switch (op->opcode()) {
+    case HloOpcode::kSend:
+    case HloOpcode::kSendDone:
+    case HloOpcode::kRecv:
+    case HloOpcode::kRecvDone:
+    case HloOpcode::kCustomCall:
+      return false;
+    default:
+      break;
+  }
```
akuegel, line 372: Can you please also check whether op->shape() is a scalar (ShapeUtil::IsEffectiveScalar())? If each op in the while condition computation has a scalar shape, we can be relatively sure that it can be evaluated quickly. Only exception: parameters of the computation may have larger shapes.

xla/hlo/analysis/while_loop_analysis_test.cc
```diff
@@ -425,5 +425,159 @@ TEST_F(WhileLoopAnalysisTest, AIVNoChain) {
...
+      ROOT while = (s32[], s32[]) while(tuple), body=body, condition=condition
+    }
+  )";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo));
+  const HloInstruction* while_op =
+      module->entry_computation()->root_instruction();
+  EXPECT_EQ(ComputeWhileLoopTripCount(while_op), std::nullopt);
+}
+
```
akuegel, line 540: Can you please add a test case showing what kind of things I want to be guarded against?

E.g. a condition computation like this:

```
or {
  x = pred[] parameter(0)
  y = pred[] parameter(1)
  ROOT res = pred[] or(x, y)
}
cond {
  param.0 = (pred[100000000]) parameter(0)
  gte = pred[100000000] get-tuple-element(param.0), index=0
  false = pred[] constant(false)
  ROOT any = pred[] reduce(gte, false), dimensions={0}, to_apply=or
}
```
(needs some tweaking to the parameter tuple to have extra entries and some matching body computation, but I hope you get the idea - for example that indvar tuple entry could be the result of a compare op in the body computation).

akuegel (changes requested):

xla/hlo/analysis/while_loop_analysis.cc
shraiysh: Done

xla/hlo/analysis/while_loop_analysis_test.cc
shraiysh: Added the test.

xla/hlo/analysis/while_loop_analysis.cc
shraiysh: Done.

akuegel (approved):
Thank you!

PR #19237, opened 2024-11-11
[GPU] Fix passing of key-value store handle from client to compiler.

sergachev:


PatriosTheGreat (approved):
Thanks for this fix!

Just our of curiosity, which PJRT client you are using that goes directly to PjRtStreamExecutorClient::Compile?

I recently added the same fix for JAX to PyClient::CompileIfrtProgram: https://github.com/openxla/xla/commit/f548642d58427ee18ee6c2ac812a4e3a3a24582d

I'm curious if we need to keep only one of these fixes.

sergachev:
> Just our of curiosity, which PJRT client you are using that goes directly to PjRtStreamExecutorClient::Compile?

JAX uses this in the reproducer you provided for example.

PatriosTheGreat (approved):

sergachev:
While this isn't merged, I realized, there is a better way to fix this: setting the key-value store in CompileInternal() is better because all paths merge at it. Added a 2nd test case for that.

PatriosTheGreat (approved):

sergachev:
Rebased and resolved a conflict.

PatriosTheGreat (approved):

reedwm:
I'm seeing internal test failures for tests `//xla/pjrt/gpu:se_gpu_pjrt_client_test` and `//xla/pjrt:pjrt_stream_executor_client_test`, but they don't fail externally. I'm also seeing a third internal-only test fail.

The issue with `pjrt_stream_executor_client_test` and the internal test seems to be calling `key_value_store()` in `pjrt_stream_executor_client.cc` without checking there is a value in the std::optional. I see the internal test using the [InterpreterDevice](https://github.com/openxla/xla/blob/main/xla/pjrt/interpreter_device.cc) which does not override `key_value_store`. Not sure about `pjrt_stream_executor_client_test` though, and why it passes externally. In any case, can you fix the case where `key_value_store()` is nullopt?

`se_gpu_pjrt_client_test` is failing internally with the error

```
INTERNAL: RET_CHECK failure (third_party/tensorflow/compiler/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc:1886) absl::StrContains( executable->GetHloModules()->front()->ToString(), "triton_gemm")
```

Any ideas what the issue is? If not, I can investigate internally.

sergachev:
Thanks! I fixed pjrt_stream_executor_client_test.

se_gpu_pjrt_client_test - I couldn't guess yet; added printing of HLO when it fails at that line. Could you paste the output please?

reedwm:
> Thanks! I fixed pjrt_stream_executor_client_test.
> 
> se_gpu_pjrt_client_test - I couldn't guess yet; added printing of HLO when it fails at that line. Could you paste the output please?

Sure, with an added log at the end of `ShardedAutotuningWorksTestBody` that prints the compiled optimized HLO module, the logs are https://gist.github.com/reedwm/57d6e21375b30c3d3666a02c5710d91e.

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1786,5 +1789,130 @@ TEST(StreamExecutorGpuClientTest, AutoLayoutIsSupported) {
...
+  }
+
+  xla::DistributedRuntimeClient::Options distributed_options;
+  distributed_options.node_id = node_id;
+  distributed_options.init_timeout = absl::Seconds(120);
+  auto distributed_client =
+      GetDistributedRuntimeClient("127.0.0.1:12345", distributed_options);
+  TF_QCHECK_OK(distributed_client->Connect());
+  GpuClientOptions options;
+  options.node_id = node_id;
```
PatriosTheGreat, line 1842: Could you please also add here: `options.allowed_devices = {node_id};`

Otherwise each process thinks that it works with both local GPUs ignoring CUDA_VISIBLE_DEVICES above, which leads to the test failure.

PatriosTheGreat (changes requested):

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
sergachev: done

sergachev:
> > Thanks! I fixed pjrt_stream_executor_client_test.
> > se_gpu_pjrt_client_test - I couldn't guess yet; added printing of HLO when it fails at that line. Could you paste the output please?
> 
> Sure, with an added log at the end of `ShardedAutotuningWorksTestBody` that prints the compiled optimized HLO module, the logs are https://gist.github.com/reedwm/57d6e21375b30c3d3666a02c5710d91e.

Thanks! I guess it ran on a pre-Ampere GPU. Made the test skip them.

PatriosTheGreat (approved):

PatriosTheGreat (approved):

PR #19336, opened 2024-11-14
[ROCm] Fix rocm_executor_test

mmakevic-amd:
The issue was introduced here https://github.com/openxla/xla/commit/7c468b089a5fb467f7ea9104d3bf6218ffb37d5d
Error log:
```
[ RUN      ] RocmExecutorTest.GetRocmKernel
xla/stream_executor/rocm/rocm_executor_test.cc:70: Failure
Value of: rocm_executor->GetRocmKernel(kernel.get())
Expected: is OK and has a value that is equal to 0x55c94b2c7730
Actual: 16-byte object <01-28 2F-4B C9-55 00-00 00-00 00-00 00-00 00-00>, which has status NOT_FOUND: Kernel not loaded in this executor.
[  FAILED  ] RocmExecutorTest.GetRocmKernel (5 ms)
```


beckerhe (approved):
Nice. Thank you!

PR #16953, opened 2024-09-09
Revert "[XLA:GPU] Enable --xla_gpu_enable_pipelined_{all_gather,all_r…

kaixih:
A recent commit enabled the pipelined collectives. However, we saw some divergence issue after the change (Will share more info on this). This PR reverts those changes.

PR #19363, opened 2024-11-14
Loop Counter Increment in Collective Pipeliner

philipphack:
Sets the loop iteration counter increment in the backward transformation of the collective pipeliner pass to account for cases with non-zero initial value of the loop iteration counter. See #16953 and #18568.

### DIFF
--- a/xla/service/collective_pipeliner_test.cc
+++ b/xla/service/collective_pipeliner_test.cc
@@ -1349,6 +1349,86 @@ ENTRY entry {
 }
 
 TEST_F(CollectivePipelinerTest,
+       TransformIncrementIndexByOneStartFromOneBackwards) {
+  constexpr absl::string_view hlo_string = R"(
+HloModule module
+
+add {
+  lhs = bf16[] parameter(0)
+  rhs = bf16[] parameter(1)
+  ROOT add = bf16[] add(lhs, rhs)
+}
+
+while_cond {
+  param = (s32[], bf16[3,8,128], bf16[3,1,2,128]) parameter(0)
+  loop_index = s32[] get-tuple-element(param), index=0
+  c4 = s32[] constant(4)
+  ROOT cmp = pred[] compare(loop_index, c4), direction=LT
+}
+
+while_body {
+  param = (s32[], bf16[3,8,128], bf16[3,1,2,128]) parameter(0)
+  loop_index = s32[] get-tuple-element(param), index=0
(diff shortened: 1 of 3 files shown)


yliu120:
Thanks. I guess this PR fixes the CP bug.

But there is another bug that causes corrupted numerics when the stack layer is unrolled. @hawkinsp

abhinavgoel95:
@yliu120 what is the other bug? Is it related to this PR?

yliu120:
> @yliu120 what is the other bug? Is it related to this PR?

Unrelated. This PR fixed the issue in the CP pass and please go ahead!

xla/service/collective_pipeliner_test.cc
```diff
@@ -1348,6 +1348,86 @@ ENTRY entry {
...
+                           CollectivePipeliner::PipeliningDirection::kBackward,
+                           IsAllGather)
+                  .value());
+  XLA_VLOG_LINES(1, module->ToString());
+  const int64_t while_count = absl::c_count_if(
+      module->entry_computation()->instructions(),
+      [](const HloInstruction* instruction) {
+        return HloPredicateIsOp<HloOpcode::kWhile>(instruction);
+      });
+  EXPECT_EQ(while_count, 1);
```
golechwierowicz, line 1409: Can you rewrite this part to filecheck? All you want to check is the increment so it should not be long, I think.

golechwierowicz (commented):
Is it possible to reproduce the numerical issue in `collective_ops_e2e_test.cc` and write a test there to make sure it does not regress?

hawkinsp:
@philipphack Any chance you can address Greg's comments soon? I'd like to get this PR into the next release of JAX (tomorrow, hopefully).

xla/service/collective_pipeliner_test.cc
philipphack: I removed some of the verifications duplicated in other tests but didn't switch to FileCheck since e.g. TransformIncrementIndexByOneBackwards doesn't use it either. Let me know if you want to use it anyway.

golechwierowicz (approved):
Thanks for adding e2e test!

xla/service/collective_pipeliner_test.cc
golechwierowicz: Without so many assertions I think in practice it's now less mandated :)

yliu120:
Thanks everyone for the bug fix.

PR #19372, opened 2024-11-14
[GPU] Consider small kInput fusions with concatenations in the horizontal loop fusion pass.

sergachev:


akuegel:
Do we know this is always strictly better? Do you have a HloModule with benchmark results?
The input fusions with concatenate root use the dedicated concat emitter which can be quite a bit faster than emitting the concat as part of a loop fusion.

sergachev:
I realized, the benchmark can be made better by 1) adding some elementwise ops so that the special path for copy is not used 2) testing separately the two paths of the horizontal loop fusion pass, for same and different output shapes, as they have different thresholds. With the same shapes, because of the higher threshold, the fusions can be more significant and the effect of this change is still positive, but less important. So probably, to be safe, I should restrict this change to only small fusions.

Different shapes:
```
       name  duration
0  f1_input      2720
1  f2_input      2400
2   f1_loop      2560
3   f2_loop      2368
4     fused      4096
```


Same shapes:
```
       name  duration
0  f1_input     85319
1  f2_input     93799
2   f1_loop     85414
3   f2_loop     94280
4     fused    174286
```

[different_fused.txt](https://github.com/user-attachments/files/17775161/different_fused.txt)
[different_input.txt](https://github.com/user-attachments/files/17775162/different_input.txt)
[different_loop.txt](https://github.com/user-attachments/files/17775163/different_loop.txt)
[same_fused.txt](https://github.com/user-attachments/files/17775164/same_fused.txt)
[same_input.txt](https://github.com/user-attachments/files/17775165/same_input.txt)
[same_loop.txt](https://github.com/user-attachments/files/17775167/same_loop.txt)

sergachev:
I updated the change, please take another look.

akuegel (approved):
Thanks for the explanation, for small fusions I can understand that it will be better to use horizontal fusion.

PR #19430, opened 2024-11-18
Pass process index and count from PJRT client.

sergachev:
Process index and count are used in sharded autotuning in the GPU compiler. To make this work also in MPMD programs they have to be set at the higher level which knows which of the connected processes are participating in compilation of an HLO module.

I didn't find an easy way to run tests from xla/python in OSS so I'll add a test for this to JAX.


PatriosTheGreat (approved):

sergachev:
Actually it seems like we found a better way to calculate process index and count using device assignment + global GPU device IDs already available in XLA; let's put this PR on pause.

PR #19432, opened 2024-11-18
[GPU] Sharded autotuning: exchange only results of the latest compilation.

sergachev:
Exchanging the cached results of previous compilation of other modules isn't necessary and leads to conflicts.


PatriosTheGreat (approved):
Thanks for the fix!

PR #19444, opened 2024-11-18
[ds-fusion][gpu] Add support to compute offset at runtime on host for dynamic slice fusion

shraiysh:
The offset is stored as a small HloModule, along with induction variable init and update modules. These modules are stored in the dynamic slice thunk and they are "evaluated" at runtime to get the value of the offset.


xla/service/gpu/runtime/dynamic_slice_thunk.h
```diff
@@ -56,7 +56,10 @@ class DynamicSliceThunk : public Thunk {
       std::vector<std::optional<std::vector<Offset>>> offsets,
       std::vector<std::optional<Shape>> orig_shapes,
       std::vector<std::optional<Shape>> sliced_shapes,
-      std::vector<std::optional<uint64_t>> offset_byte_sizes);
+      std::vector<std::optional<uint64_t>> offset_byte_sizes,
+      std::unique_ptr<HloModule> while_init,
+      std::unique_ptr<HloModule> while_update,
+      std::vector<std::unique_ptr<HloModule>> fake_offset_modules);
```
bchetioui, line 62: What's the point of `fake_offset_modules`? It doesn't seem used, apart from branching on `fake_offset_modules_.empty()`?

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+    VLOG(2) << "Expected init to be a tuple, found: " << init_tuple->ToString();
+    return nullptr;
+  }
+  const HloInstruction* init = init_tuple->operand(*indvar_idx);
+  std::unique_ptr<HloModule> init_module = ExtractModule(init);
+  if (init_module == nullptr) {
+    VLOG(2) << "Unable to get init module for init: " << init->ToString();
+    return nullptr;
+  }
+  if (!IsValid(init_module.get(), nullptr)) {
```
bchetioui, line 293: Is it really worth factoring `IsValid`? It's pretty small, and the semantics are unclear.

Seems like you're calling it exactly twice, once with a set `arg_shape` and one with a `nullptr`. How about factoring only the part about checking for `ReplicaId` and `PartitionId`? (Although it's still not clear why you need to check those.)

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+    const HloInstruction* offset_value) {
+  HloInstruction* while_op = GetWhileOp(offset_value);
+  if (while_op == nullptr) {
+    VLOG(2) << "Unable to find surrounding while op"
+            << offset_value->ToString();
+    return nullptr;
+  }
+  const HloComputation* while_body = while_op->while_body();
+  std::optional<int64_t> indvar_idx = GetLoopInductionVarTupleIdx(while_op);
+  if (indvar_idx == std::nullopt) {
```
bchetioui, line 237: `!indvar_idx.has_value()`?

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+  }
+  HloInstruction* indvar = hlo_query::GetUniqueGteInstruction(
+      while_body->parameter_instruction(0), *indvar_idx);
+  if (indvar == nullptr) {
+    VLOG(2) << "Unable to find induction variable(idx=" << *indvar_idx
+            << ") in while body: " << while_body->ToString();
+    return nullptr;
+  }
+  std::unique_ptr<HloModule> offset_module = ExtractModule(
+      offset_value, -1,
```
bchetioui, line 255: `/*height=*/-1, ... /*cross_computation=*/true, /*inline_calls_and_fusions=*/true, /*run_verifier=*/true`

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
     const BufferAssignment& buffer_assignment,
     const HloInstruction& fusion_instr,
     absl::Span<HloInstruction*> slice_instrs,
     std::vector<std::optional<std::vector<DynamicSliceThunk::Offset>>>& offsets,
     std::vector<std::optional<Shape>>& orig_shapes,
     std::vector<std::optional<Shape>>& sliced_shapes,
-    std::vector<std::optional<uint64_t>>& offset_byte_sizes, unsigned arg_idx) {
+    std::vector<std::optional<uint64_t>>& offset_byte_sizes, unsigned arg_idx,
+    bool can_compute_indvar_on_host,
+    std::vector<std::unique_ptr<HloModule>>& fake_modules) {
```
bchetioui, line 332: Also not clear on the `fake_modules` here---what do we use them for?

xla/service/gpu/runtime/dynamic_slice_thunk.h
```diff
@@ -56,7 +56,10 @@ class DynamicSliceThunk : public Thunk {
       std::vector<std::optional<std::vector<Offset>>> offsets,
       std::vector<std::optional<Shape>> orig_shapes,
       std::vector<std::optional<Shape>> sliced_shapes,
-      std::vector<std::optional<uint64_t>> offset_byte_sizes);
+      std::vector<std::optional<uint64_t>> offset_byte_sizes,
+      std::unique_ptr<HloModule> while_init,
```
bchetioui, line 60: The general approach should be documented as part of this thunk's documentation.

xla/service/gpu/runtime/dynamic_slice_thunk.h
```diff
@@ -56,7 +56,10 @@ class DynamicSliceThunk : public Thunk {
...
+      std::unique_ptr<HloModule> while_update,
```
bchetioui, line 61: When looking at the thunk, does it make sense to talk about `while_{init,update}`? It seems like the fact that they come from a while loop is irrelevant at this point (?)

xla/service/gpu/runtime/dynamic_slice_thunk.h
```diff
@@ -129,6 +132,16 @@ class DynamicSliceThunk : public Thunk {
...
   std::vector<int64_t> offsets_allocs_base_;
+
+  std::unique_ptr<HloModule> while_init_, while_update_;
+  std::vector<std::unique_ptr<HloModule>> fake_offset_modules_;
+
+  // We need a map of replica-id -> induction variable. This is because all the
+  // replicas share the same executable thunk objects. For the same reason we
+  // need a mutex to access this, because flat_hash_map is not thread safe for
+  // mutliple replicas to access concurrently.
+  absl::flat_hash_map<int, Literal> indvar_;
```
bchetioui, line 143: nit proposal:

```
using ReplicaId = int;
absl::flat_hash_map<ReplicaId, Literal> indvar_;
```

xla/service/gpu/runtime/dynamic_slice_thunk.h
```diff
@@ -129,6 +132,16 @@ class DynamicSliceThunk : public Thunk {
...
   // A mapping from argument index to the base offset in the `offsets_allocs_`.
   std::vector<int64_t> offsets_allocs_base_;
+
+  std::unique_ptr<HloModule> while_init_, while_update_;
+  std::vector<std::unique_ptr<HloModule>> fake_offset_modules_;
+
+  // We need a map of replica-id -> induction variable. This is because all the
+  // replicas share the same executable thunk objects. For the same reason we
+  // need a mutex to access this, because flat_hash_map is not thread safe for
+  // mutliple replicas to access concurrently.
```
bchetioui, line 142: multiple

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -180,7 +217,28 @@ absl::Status DynamicSliceThunk::ExecuteOnStream(const ExecuteParams& params) {
         VLOG(2) << "  - arg " << argument_idx << "[" << offset_idx
                 << "]: constant offset = " << *const_offset;
         offset_value(argument_idx, offset_idx) = *const_offset;
-
+      } else if (HloModule** offset_module = std::get_if<HloModule*>(&offset)) {
+        VLOG(0) << "Offset module: " << (*offset_module)->ToString();
```
bchetioui, line 221: `VLOG(2)`?

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -180,7 +217,28 @@ absl::Status DynamicSliceThunk::ExecuteOnStream(const ExecuteParams& params) {
...
+                            GetReplicaId(params.collective_params));
+
+        // Access indvar behind a lock.
+        indvar_mutex.Lock();
+        Literal indvar = indvar_[replica_id].Clone();
+        indvar_mutex.Unlock();
+
+        TF_ASSIGN_OR_RETURN(
+            Literal offset,
+            HloEvaluator().Evaluate(**offset_module, {indvar.Clone()}));
```
bchetioui, line 232: Why do we need to `Clone()` here? Seems like we already do it above and we don't reuse `indvar`.

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -180,7 +217,28 @@ absl::Status DynamicSliceThunk::ExecuteOnStream(const ExecuteParams& params) {
...
                 << "]: constant offset = " << *const_offset;
         offset_value(argument_idx, offset_idx) = *const_offset;
-
+      } else if (HloModule** offset_module = std::get_if<HloModule*>(&offset)) {
+        VLOG(0) << "Offset module: " << (*offset_module)->ToString();
+        TF_ASSIGN_OR_RETURN(auto replica_id,
+                            GetReplicaId(params.collective_params));
+
+        // Access indvar behind a lock.
+        indvar_mutex.Lock();
```
bchetioui, line 226: How about using the same style in both places where you lock the mutex? E.g. using a nested block in both places.

bchetioui (changes requested):

xla/service/gpu/runtime/dynamic_slice_thunk.h
shraiysh: The fake_offset_modules here are just for easier management of ownership. This allows us to store the pointer to these modules in the [std::variant offset vector here](https://github.com/shraiysh/xla/blob/c1193c0f612f3efea1e624f423d89216f2cfa16a/xla/service/gpu/runtime/dynamic_slice_thunk.h#L50) instead of the `std::unique_ptr`. Let me see if I can store the unique_ptr directly there, but I remember the compiler complaining about calls to deleted constructor with [std::vector used here](https://github.com/shraiysh/xla/blob/c1193c0f612f3efea1e624f423d89216f2cfa16a/xla/service/gpu/runtime/dynamic_slice_thunk.h#L56).

shraiysh:
Thank you for the detailed review. I will address these comments.

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
   return absl::InternalError("WTF");
 }
 
+// Gets the innermost while op surrounding the operation `op`.
+HloInstruction* GetWhileOp(const HloInstruction* op) {
```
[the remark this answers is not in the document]
shraiysh: Done

xla/service/gpu/runtime/for_all_thunks_test.cc
```diff
@@ -65,8 +65,9 @@ TEST(ForAllThunksTest, DynamicSliceThunk) {
   auto thunk_sequence = std::make_unique<ThunkSequence>();
   thunk_sequence->push_back(std::move(thunk));
 
-  DynamicSliceThunk dynamic_slice_thunk(
-      Thunk::ThunkInfo(), std::move(thunk_sequence), {}, {}, {}, {}, {}, {});
+  DynamicSliceThunk dynamic_slice_thunk(Thunk::ThunkInfo(),
+                                        std::move(thunk_sequence), {}, {}, {},
```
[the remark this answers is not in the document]
shraiysh: Done.

xla/service/gpu/runtime/dynamic_slice_thunk_test.cc
```diff
@@ -148,7 +148,8 @@ TEST(DynamicSliceThunkTest, SlicedGemm) {
        std::nullopt, std::nullopt},
       {ShapeUtil::MakeShape(PrimitiveType::F32, {1, 3}), std::nullopt,
        std::nullopt, std::nullopt},
-      {sizeof(int64_t), std::nullopt, std::nullopt, std::nullopt});
+      {sizeof(int64_t), std::nullopt, std::nullopt, std::nullopt}, nullptr,
```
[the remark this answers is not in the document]
shraiysh: Done.

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+    if (callers[0]->opcode() == HloOpcode::kWhile) return callers[0];
+    op = callers[0];
+  }
+  return nullptr;
+}
+
+// Checks the module object. This returns false if the module is a `nullptr`. If
+// the `arg_shape` is a `nullptr`, then the function returns true if the module
+// has no parameters. Otherwise, it returns true if the module has a single
+// parameter and the parameter shape matches the `arg_shape`.
```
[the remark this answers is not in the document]
shraiysh: Updated the comment and changed the function.

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
   return absl::InternalError("WTF");
 }
 
+// Gets the innermost while op surrounding the operation `op`.
+HloInstruction* GetWhileOp(const HloInstruction* op) {
+  std::unique_ptr<CallGraph> graph = CallGraph::Build(op->GetModule());
+  while (op) {
+    const HloComputation* parent = op->parent();
+    std::vector<HloInstruction*> callers = graph->GetComputationCallers(parent);
+    if (callers.size() != 1) return nullptr;
```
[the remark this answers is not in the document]
shraiysh: I think I have done this everywhere, please let me know if I missed something in the updated patch.

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+  return nullptr;
+}
+
+// Checks the module object. This returns false if the module is a `nullptr`. If
+// the `arg_shape` is a `nullptr`, then the function returns true if the module
+// has no parameters. Otherwise, it returns true if the module has a single
+// parameter and the parameter shape matches the `arg_shape`.
+bool IsValid(const HloModule* module, const Shape* arg_shape) {
+  if (module == nullptr) return false;
+  // TODO: support partition-id and replica-id.
```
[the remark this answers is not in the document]
shraiysh: Just wanted to do it in a separate patch so that the reviews are quicker (smaller patches).

xla/service/gpu/fusions/custom.cc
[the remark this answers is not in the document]
shraiysh: Done. Thanks!

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,151 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+  }
+  return nullptr;
+}
+
+// Checks the module object. This returns false if the module is a `nullptr`. If
+// the `arg_shape` is a `nullptr`, then the function returns true if the module
+// has no parameters. Otherwise, it returns true if the module has a single
+// parameter and the parameter shape matches the `arg_shape`.
+bool IsValid(const HloModule* module, const Shape* arg_shape) {
+  if (module == nullptr) return false;
```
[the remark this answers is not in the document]
shraiysh: Repurposed this function and am using references now. Thanks!

xla/service/gpu/fusions/custom.cc
shraiysh: You are right. Changed it to only check for partition-id and replica-id.

xla/service/gpu/runtime/dynamic_slice_thunk.cc
shraiysh: Thanks for pointing it out. Moving it instead of cloning it.

xla/service/gpu/runtime/dynamic_slice_thunk.h
shraiysh: This causes issues if I try to put the unique pointer directly inside the offsets: https://github.com/shraiysh/xla/blob/90af2896ab4992ff14a1cd2a75ce02e43f46c090/xla/service/gpu/runtime/dynamic_slice_thunk.cc#L70 
```
xla/service/gpu/runtime/dynamic_slice_thunk.cc:94:13: note: in instantiation of member function 'std::vector<xla::gpu::DynamicSliceThunk::SliceDef>::push_back' requested here
   94 |     slices_.push_back(SliceDef{
      |             ^
/usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/variant:1396:7: note: explicitly defaulted function was implicitly deleted here
 1396 |       variant(const variant& __rhs) = default;
```
Would any sort of `std::move` help here?

shraiysh:
Added @xla-rotation for review, because @bchetioui is OOO for this week.

xla/service/gpu/fusions/custom.cc
```diff
@@ -180,14 +185,166 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
   return absl::InternalError("WTF");
 }
 
+// Gets the innermost while op surrounding the operation `op` if it exists. This
+// also returns nullopt if there are multiple callers for the same while body.
+std::optional<HloInstruction*> GetWhileOp(const HloInstruction& op) {
+  std::unique_ptr<CallGraph> graph = CallGraph::Build(op.GetModule());
```
akuegel, line 191: Can this be moved out, so we build the call graph only once? I fear this may cause a performance problem if we process the whole module each time when GetWhileOp is called.

akuegel (commented):
I am also mostly OOO these days, but I took a quick look and have one comment.

shraiysh:
This PR has been split into multiple PRs. Closing it here.

PR #19451, opened 2024-11-18
Setting xla_gpu_multi_streamed_windowed_einsum to true by default

Tixxx:
We are trying to deprecate xla_gpu_multi_streamed_windowed_einsum  since we always have better perf with it enabled. This is the first pr to enable it by default to test for stability.

### DIFF
--- a/xla/service/gpu/transforms/stream_attribute_annotator.h
+++ b/xla/service/gpu/transforms/stream_attribute_annotator.h
@@ -47,8 +47,10 @@ namespace xla::gpu {
 class StreamAttributeAnnotator : public HloModulePass {
  public:
   explicit StreamAttributeAnnotator(
-      const se::DeviceDescription& device_description)
-      : device_description_(device_description) {}
+      const se::DeviceDescription& device_description,
+      bool annotate_copy_start = false)
+      : device_description_(device_description),
+        annotate_copy_start_(annotate_copy_start) {}
 
   absl::string_view name() const override {
     return "stream-attribute-annotator";
@@ -61,6 +63,7 @@ class StreamAttributeAnnotator : public HloModulePass {
 
  private:
   const se::DeviceDescription& device_description_;
+  bool annotate_copy_start_;
 };
 
 }  // namespace xla::gpu
--- a/xla/service/gpu/transforms/stream_attribute_annotator.h
+++ b/xla/service/gpu/transforms/stream_attribute_annotator.h
@@ -47,10 +47,8 @@ namespace xla::gpu {
 class StreamAttributeAnnotator : public HloModulePass {
  public:
   explicit StreamAttributeAnnotator(
-      const se::DeviceDescription& device_description,
-      bool annotate_copy_start = false)
-      : device_description_(device_description),
-        annotate_copy_start_(annotate_copy_start) {}
+      const se::DeviceDescription& device_description)
+      : device_description_(device_description) {}
 
   absl::string_view name() const override {
     return "stream-attribute-annotator";
@@ -63,7 +61,6 @@ class StreamAttributeAnnotator : public HloModulePass {
 
  private:
   const se::DeviceDescription& device_description_;
-  bool annotate_copy_start_;
 };
 
 }  // namespace xla::gpu
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -272,7 +272,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_bf16_3way_gemm(false);
   opts.set_xla_gpu_nccl_collective_max_nchannels(0);
   opts.set_xla_gpu_nccl_p2p_max_nchannels(0);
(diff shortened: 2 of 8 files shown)


reedwm (approved):

loislo:
I believe the implementation of the feature is good but the right procedure for a flag deprecation has a few steps:
1) make flag true by default for some time;
2) remove the implementation but keep the flag and mark it as deprecated for some time;
3) remove the flag.

Tixxx:
> I believe the implementation of the feature is good but the right procedure for a flag deprecation has a few steps:
> 
> 1. make flag true by default for some time;
> 2. remove the implementation but keep the flag and mark it as deprecated for some time;
> 3. remove the flag.
> 
> The way to deprecate: flag_list->push_back(tsl::Flag("xla_gpu_enable_triton_gemm_int4", noop_flag_setter, true, "[Deprecated, do not use]"));

Thanks @loislo , i can change this pr to just turn it on by default. How long is usually a good "some time" to evaluate? Can I make a followup pr to remove it in a week?

Tixxx:
Alright, I reverted the changes and turned the flag on by default. Please take another look. I will make another pr if no other issues are found.

reedwm (changes requested):
Can you change the PR title and description to indicate you're just changing the default?

Tixxx:
> Can you change the PR title and description to indicate you're just changing the default?

Done.

reedwm (approved):

dimitar-asenov:
This PR breaks these three tests from JAX:

```
ActivationOffloadingTest.test_remat_checkpoint_dots_with_no_batch_dims
DevicePutTest.test_output_streaming
DevicePutTest.test_parameter_and_output_streaming_with_scalar
```

From this file:
https://github.com/jax-ml/jax/blob/main/tests/memories_test.py

Could you please look into what would make these pass with the change? Is this an indication that we need to maintain this flag for longer?

Tixxx:
> This PR breaks these three tests from JAX:
> 
> ```
> ActivationOffloadingTest.test_remat_checkpoint_dots_with_no_batch_dims
> DevicePutTest.test_output_streaming
> DevicePutTest.test_parameter_and_output_streaming_with_scalar
> ```
> 
> From this file: https://github.com/jax-ml/jax/blob/main/tests/memories_test.py
> 
> Could you please look into what would make these pass with the change? Is this an indication that we need to maintain this flag for longer?

ok sure, let me dig into the failures. This flag guards a pass to convert instructions into async instructions, i suspect that some copy instructions might get affected.

Tixxx:
@dimitar-asenov  I fixed the test, there's one pass StreamAttributeAnnotator which used to be guarded by the flag that im deprecating. The pass is run twice, before and after remat. When the flag is turned on by default, it starts annotating copy start and changing DUS fusions before remat which is not correct. I changed the pass to have an option to skip copy and DUS conditionally.

xla/service/gpu/transforms/stream_attribute_annotator.h
```diff
@@ -47,8 +47,10 @@ namespace xla::gpu {
 class StreamAttributeAnnotator : public HloModulePass {
  public:
   explicit StreamAttributeAnnotator(
-      const se::DeviceDescription& device_description)
-      : device_description_(device_description) {}
+      const se::DeviceDescription& device_description,
+      bool annotate_copy_start = false)
```
dimitar-asenov, line 51: Please do not use a default value and initialize this explicitly everywhere. Also please explain in a code comment what exactly this parameter does and when we would like to set it to false and when to true.

I'm not entirely familiar with this pass, but a general question is: is there a way to infer this value from the HLO during the pass, as opposed to explicitly controlling the behavior via a constructor parameter?

dimitar-asenov (changes requested):

dimitar-asenov (approved):

Tixxx:
This is to address the revert in (https://github.com/openxla/xla/pull/20577). The customer is seeing MFU drop with the previous commit of this pr. It enabled a2a+gemm rewrite by default, when it's applied on smaller gemms, the overlap is not as expected. The new commit introduces a flag to turn the a2a+gemm rewrite off.

reedwm (approved):

reedwm:
Not sure whether we can merge the same PR twice, but let's try and see.

dimitar-asenov (changes requested):
Thank you for implementing the fix. Please split these changes into two separate PRs:
- Change a flag to true.
- Add a new flag.

That way, if we need to revert something, the blast radius will be smaller. It's also generally preferable to have smaller, more focused PRs.

Tixxx:
Thanks, i have peeled out the new changes and added them to this pr(https://github.com/openxla/xla/pull/20744). Can you please take a look?

dimitar-asenov:
> Thanks, i have peeled out the new changes and added them to this pr(#20744). Can you please take a look?

Thank, both PRs look good now (after you implement the tiny fix suggested on the other PR).

dimitar-asenov (approved):

PR #19462, opened 2024-11-19
Add ExplicitStreamAnnotationAsyncWrapper pass

chaserileyroberts:
This PR introduces a new pass `ExplicitStreamAnnotationAsyncWrapper`. This pass takes `kCall` instructions that are annotated with the frontend attribute `xla_gpu_experimental_stream_annotation`, and wraps the call with an async start-done pair.

Initially, I tried to integrate this with the existing `StreamAttributeAnnotator` and `StreamAttributeAsyncWrapper` passes, but much of that code is specifically just for the `windowedeinsum` logic, and users wont necessarily want both enabled at the same time. Thus, I found it cleaner to instead just make a new pass.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -200,6 +200,7 @@ limitations under the License.
 #include "xla/service/gpu/transforms/dot_operand_converter.h"
 #include "xla/service/gpu/transforms/double_buffer_loop_unrolling.h"
 #include "xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.h"
+#include "xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.h"
 #include "xla/service/gpu/transforms/fusion_wrapper.h"
 #include "xla/service/gpu/transforms/gemm_broadcast_folding_rewriter.h"
 #include "xla/service/gpu/transforms/gemm_fusion.h"
@@ -1216,7 +1217,11 @@ absl::Status RunPostFusionSimplificationPasses(
         gpu_target_config.device_description);
     pipeline.AddPass<StreamAttributeAsyncWrapper>();
   }
-
+  if (hlo_module->config()
+          .debug_options()
+          .xla_gpu_experimental_stream_annotation()) {
+    pipeline.AddPass<ExplicitStreamAnnotationAsyncWrapper>();
+  }
   return pipeline.Run(hlo_module).status();
 }
 
--- a/xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
+++ b/xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
@@ -1,0 +1,71 @@
+/* Copyright 2024 The OpenXLA Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.h"
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/strings/string_view.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/service/gpu/backend_configs.pb.h"
+#include "xla/service/gpu/runtime/thunk.h"
+#include "xla/side_effect_util.h"
+#include "xla/util.h"
+#include "xla/xla_data.pb.h"
+#include "tsl/platform/errors.h"
+#include "tsl/platform/logging.h"
+#include "tsl/platform/statusor.h"
+
+namespace xla::gpu {
+
+namespace {
+static absl::StatusOr<bool> AsynchronizeInstruction(HloInstruction* instr) {
+  if (instr->opcode() != HloOpcode::kCall ||
+      !instr->frontend_attributes().map().contains(kXlaStreamAnnotationAttr)) {
+    return false;
+  }
+  HloComputation* computation = instr->parent();
+  TF_ASSIGN_OR_RETURN(
+      HloInstruction * done,
+      computation->CreateAsyncInstructions(
+          instr, {},
+          ExplicitStreamAnnotationAsyncWrapper::kExplicitExecutionThread,
+          /*replace=*/true));
+  TF_ASSIGN_OR_RETURN(GpuBackendConfig gpu_config,
+                      done->backend_config<GpuBackendConfig>());
+  // Set the false delay of done op to be false so it can be scheduled
+  // far apart from start.
+  gpu_config.set_force_earliest_schedule(false);
+  TF_RETURN_IF_ERROR(done->set_backend_config(gpu_config));
+  VLOG(5) << "Created async instruction: " << done->ToString();
+  return true;
+}
+}  // namespace
+
+absl::StatusOr<bool> ExplicitStreamAnnotationAsyncWrapper::Run(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  bool changed = false;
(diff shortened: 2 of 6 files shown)


xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
```diff
@@ -0,0 +1,76 @@
...
+}  // namespace
+
+absl::StatusOr<bool> ExplicitStreamAnnotationAsyncWrapper::Run(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  XLA_VLOG_LINES(2, "ExplicitStreamAnnotationAsyncWrapper::Run(), before:\n" +
+                        module->ToString());
+  bool changed = false;
+  for (const HloComputation* comp :
+       module->MakeComputationSorted(execution_threads)) {
```
frgossen, line 65: Is the iteration order needed here?

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper_test.cc
```diff
@@ -0,0 +1,72 @@
...
+
+  auto debug_options = HloTestBase::GetDebugOptionsForTest();
+  debug_options.set_xla_gpu_experimental_stream_annotation(true);
+  auto module = ParseAndReturnVerifiedModule(hlo_string).value();
+  module->mutable_config().set_debug_options(debug_options);
+  ExplicitStreamAnnotationAsyncWrapper wrapper_pass;
+
+  TF_ASSERT_OK_AND_ASSIGN(bool mutated, wrapper_pass.Run(module.get()));
+  absl::StatusOr<bool> filecheck_result = RunFileCheck(module->ToString({}), R"(
+  //CHECK: %lhs.1 = f32[] constant(42)
```
frgossen, line 62: nit: space missing `// CHECK`

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper_test.cc
```diff
@@ -0,0 +1,72 @@
...
+#include "xla/tests/hlo_test_base.h"
+#include "xla/tsl/lib/core/status_test_util.h"
+#include "tsl/platform/statusor.h"
+
+namespace xla::gpu {
+namespace {
+
+using ExplicitStreamAnnotationAsyncWrapperTest = HloTestBase;
+
+TEST_F(ExplicitStreamAnnotationAsyncWrapperTest, AnnotatedOpIsWrapped) {
```
frgossen, line 39: It's great to have this small test case!

Can you also add a test case with an example that you actually care about? I assume this is for collective gemms?

frgossen (changes requested):
Can you provide more context? What is this needed for? Collective gemm? 

Can you add an execution test for this?

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
chaserileyroberts: No, I don't think it's required, I just didn't see an existing non-ordered API. Do you know which one I should use instead?

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper_test.cc
chaserileyroberts: Yes, the ultimate goal is to be able to have two partial gemms run in parallel. I'll add an example test case.

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper_test.cc
chaserileyroberts: Done.

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper_test.cc
```diff
@@ -0,0 +1,115 @@
...
+  // CHECK: %call-start = ((f32[]), f32[]) call-start(f32[] %lhs.1), async_execution_thread="explicit", to_apply=%sub, frontend_attributes={_xla_stream_annotation="1"}
+  // CHECK: ROOT %call-done = f32[] call-done(((f32[]), f32[]) %call-start), frontend_attributes={_xla_stream_annotation="1"}, backend_config={"operation_queue_id":"0","wait_on_operation_queues":[],"force_earliest_schedule":false}
+  )");
+  TF_ASSERT_OK(filecheck_result.status());
+  EXPECT_TRUE(*filecheck_result);
+
+  ASSERT_TRUE(mutated);
+}
+
+TEST_F(ExplicitStreamAnnotationAsyncWrapperTest, OverlappingGemms) {
```
frgossen, line 72: Thanks for adding this. Can you clean up the HLO a little so that it is readable?

frgossen (changes requested):

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper_test.cc
chaserileyroberts: I've done my best to simplify this test while also testing a "realistic" scenario.

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
frgossen: I think the typical thing people do is something like `HloComputation* comp : module->computations()`, which is all over the code base

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
```diff
@@ -0,0 +1,76 @@
...
+  TF_RETURN_IF_ERROR(done->set_backend_config(gpu_config));
+  VLOG(5) << "Created async instruction: " << done->ToString();
+  return true;
+}
+}  // namespace
+
+absl::StatusOr<bool> ExplicitStreamAnnotationAsyncWrapper::Run(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  XLA_VLOG_LINES(2, "ExplicitStreamAnnotationAsyncWrapper::Run(), before:\n" +
```
frgossen, line 61: Don't think you want to vlog "ExplicitStreamAnnotationAsyncWrapper::Run()" here, right? 
Also, instead of the '\n', how about two vlog statements?

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
```diff
@@ -0,0 +1,76 @@
...
+                        module->ToString());
+  bool changed = false;
+  for (const HloComputation* comp :
+       module->MakeComputationSorted(execution_threads)) {
+    for (HloInstruction* instr : comp->instructions()) {
+      TF_ASSIGN_OR_RETURN(bool result, AsynchronizeInstruction(instr));
+      changed |= result;
+    }
+  }
+  XLA_VLOG_LINES(2, "ExplicitStreamAnnotationAsyncWrapper::Run(), after:\n" +
```
frgossen, line 71: nit: same

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1216,7 +1217,11 @@ absl::Status RunPostFusionSimplificationPasses(
         gpu_target_config.device_description);
     pipeline.AddPass<StreamAttributeAsyncWrapper>();
   }
-
+  if (hlo_module->config()
```
frgossen, line 1220: Can you add an execution test for this? Will the runtime be able to deal with the async ops you create here?

frgossen (changes requested):

xla/service/gpu/gpu_compiler.cc
chaserileyroberts: I was planning to add the execution test to https://github.com/openxla/xla/pull/20155 once this PR is merged.

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
chaserileyroberts: I was just replicating what exists in other passes. I'm ok with removing the vlogs.

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
chaserileyroberts: done

chaserileyroberts:
I removed the vlogs and updated the loop to use `module->computations()`

xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
chaserileyroberts: Done.

frgossen (approved):
thanks!

PR #19528, opened 2024-11-20
[XLA:GPU] use separte command buffer cmd flag for conditional and loop

shawnwang18:
Observed in saxml workload that sharing the same command buffer cmd type (CONDITIONALS) for WHILE and CONDITIONAL command over kill the lowering opportunities.  

Many cases could allow CONDITIONAL instruction to lower into command buffer, while WHILE is not possible. 

This PR uses separate command buffer cmd type flag for CONDITIONAL and WHILE instructions when user specifies the type to lowering.

### DIFF
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -604,10 +604,11 @@ message DebugOptions {
     CUBLAS = 2;
     CUDNN = 3;
     COLLECTIVES = 4;
-    CONDITIONALS = 5;
-    CUSTOM_CALL = 6;
-    CUBLASLT = 7;
-    DYNAMIC_SLICE = 8;
+    CONDITIONAL = 5;
+    WHILE = 6;
+    CUSTOM_CALL = 7;
+    CUBLASLT = 8;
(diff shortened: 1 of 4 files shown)


xla/xla.proto
```diff
@@ -604,10 +604,11 @@ message DebugOptions {
...
     COLLECTIVES = 4;
-    CONDITIONALS = 5;
-    CUSTOM_CALL = 6;
-    CUBLASLT = 7;
-    DYNAMIC_SLICE = 8;
+    CONDITIONAL = 5;
+    WHILE = 6;
+    CUSTOM_CALL = 7;
+    CUBLASLT = 8;
+    DYNAMIC_SLICE = 9;
```
ezhulenev, line 611: let's call it DYNAMIC_SLICE_FUSION not to confuse with regular dynamic-slice operation

shawnwang18: Fixed

ezhulenev (approved):

PR #19571, opened 2024-11-20
PJRT: assign process index and count for compilation using device assignment.

sergachev:
Only a subset of processes may be participating in the compilation of a module.

### DIFF
--- a/xla/pjrt/pjrt_stream_executor_client.cc
+++ b/xla/pjrt/pjrt_stream_executor_client.cc
@@ -3481,6 +3481,8 @@ PjRtStreamExecutorClient::GetExecutableExtras(CompileOptions* options) {
   if (device_assignment != nullptr) {
     addressable_device_logical_ids.reserve(num_replicas * num_partitions);
     addressable_devices.reserve(num_replicas * num_partitions);
+    absl::flat_hash_set<int> all_process_indices;
+    std::optional<int> this_process_index;
     for (int replica = 0; replica < num_replicas; ++replica) {
       for (int partition = 0; partition < num_partitions; ++partition) {
         int64_t device_id = (*device_assignment)(replica, partition);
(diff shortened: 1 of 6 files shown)


xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3481,17 +3481,23 @@ PjRtStreamExecutorClient::GetExecutableExtras(CompileOptions* options) {
...
 
         TF_ASSIGN_OR_RETURN(PjRtDevice * device,
                             LookupDevice(global_device_id));
+        all_process_indices.insert(device->process_index());
         if (device->process_index() != process_index()) {
           VLOG(3) << "Non-local device: " << device_id;
           continue;
         }
+        if (!this_process_index.has_value()) {
+          this_process_index = all_process_indices.size() - 1;
```
PatriosTheGreat, line 3499: Not sure I understood this line correctly.
Doesn't this mean that all processes will assign this_process_index by the index of the first device?

PatriosTheGreat (changes requested):

xla/pjrt/pjrt_stream_executor_client.cc
sergachev: Each process iterates over all participating devices, looks at the process index corresponding to the device and compares it to its own process id. Using the hash set it counts how many other processes it has seen before it finds own one. In other words, this assigns process index for compilation based on first occurrence of this process in the {device: process} map.

PatriosTheGreat: Ah sorry, I didn't notice `if (device->process_index() != process_index()) {` line above, make sense.

Isn't this equal to just initializing `build_options.set_process_index(process_index())` ?

sergachev: No - for sharded autotuning to work we want to always re-enumerate participating processes for compilation to 0...count-1 disregarding their original indices because an arbitrary subset of them can be participating in a compilation.

PatriosTheGreat (approved):

golechwierowicz:
The following test fails: 

```
FunctionalHloRunnerTest.ShardedAutotuningWorks
```

with 

```
Expected equality of these values:
  child_status
    Which is: 256
  0
 node 0

```

on 2 H100.

sergachev:
Thanks! Fixed.

golechwierowicz (approved):

PR #19582, opened 2024-11-21
[ROCm] Fix launch dimension triplet for ROCm

hsharsha:
Owing to checks in https://github.com/openxla/xla/blob/main/xla/service/gpu/parallel_loop_emitter.cc#L169-L171 launch dimension can be of the form ((block.x, 1, 1), (thread.x, thread.y, 1)). And in ROCm it is expected that (block.x * thread.x) <= 0xFFFFFFFF

### DIFF
--- a/xla/service/gpu/launch_dimensions.cc
+++ b/xla/service/gpu/launch_dimensions.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include <algorithm>
 #include <cstdint>
 
+#include "xla/service/platform_util.h"
 #include "xla/shape.h"
 #include "xla/shape_util.h"
 #include "xla/stream_executor/device_description.h"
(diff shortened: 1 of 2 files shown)


hsharsha:
@olegshyshkov  This patch tries to address some of the violations introduced in https://github.com/openxla/xla/commit/edf18ce242f234fbd20d1fbf4e9c96dfa5be2847

olegshyshkov:
I would prefer that we don't revert https://github.com/openxla/xla/commit/edf18ce242f234fbd20d1fbf4e9c96dfa5be2847.

[parallel_loop_emitter.cc](https://github.com/openxla/xla/blob/main/xla/service/gpu/parallel_loop_emitter.cc#L169-L171) is a very old part of the emitter that is used only for a handful special instruction, so I wouldn't use it as a ground of truth. That change was aimed for Nvidia GPU, but I understand that ROCm has different requirement.

I think the solution here is to have a backend-specific logic for the check and to distribute blocks.

hsharsha:
> I would prefer that we don't revert https://github.com/openxla/xla/commit/edf18ce242f234fbd20d1fbf4e9c96dfa5be2847.
@olegshyshkov  I encountered a failure in jax maxtext model at parallel_loop_emitter that blocks.y > 1. The same would fail for nvidia as well. That was the reason for reverting.

olegshyshkov:
Interesting. Could you share an HLO snippet with the fusion that causes the failure?

hsharsha:
@olegshyshkov  Following snippet can cause the error
```
HloModule jit_train_step, entry_computation_layout={(s32[])->bf16[80,7,8192,8192]}

fused_broadcast.3 {
  constant_135_2 = bf16[] constant(0)
  ROOT broadcast.599.1 = bf16[80,7,8192,8192]{3,2,1,0} broadcast(constant_135_2), dimensions={}
}


ENTRY command_buffer {
  p = s32[] parameter(0)
  ROOT loop_broadcast_fusion.3 = bf16[80,7,8192,8192]{3,2,1,0} fusion(), kind=kLoop, calls=fused_broadcast.3
} // command_buffer
```

But I am not able to reproduce this on this branch. When I run run_hlo_module it gives llvm finger print error
```
INFO: Running command line: bazel-bin/xla/tools/run_hlo_module '--platform=GPU' '--reference_platform=default' '--xla_disable_all_hlo_passes=true' /path/to/trim.txt

 ** Running /home/hahavanu/xla/trim.txt**
Running HLO module with runner ROCM...
2024-11-27 17:26:31.539523: I xla/service/llvm_ir/llvm_command_line_options.cc:50] XLA (re)initializing LLVM with options fingerprint: 6139028976987762984
Segmentation fault (core dumped)
```
But can be reproduced with `run_hlo_module` utility on this branch https://github.com/ROCm/xla/tree/ci_rv_mt_fix_launch_dims

hsharsha:
Also On NVIDIA GPUs, you may have to increase the tensor size so as to exceed block_dim_limit().x in https://github.com/openxla/xla/blob/main/xla/service/gpu/launch_dimensions.cc#L45

hsharsha:
@olegshyshkov  were you able to reproduce the error?

olegshyshkov:
Sorry for the delayed reply!

> But I am not able to reproduce this on this branch. When I run run_hlo_module it gives llvm finger print error

This is not an LLVM finger print error. You're getting a `Segmentation fault (core dumped)`.

I'm not sure how to reproduce that, because the output of the HLO reproduce that you have is 75GB. We allow about 80% of GPU memory to be used for the buffers, so on H100 with 80GB we have around 64GB.

The way to go here would be to have backend-specific logic to distribute block_ids and threads along the dimensions.

hsharsha:
@olegshyshkov  I have updated this PR to accommodate platform specific changes for launch dims

i-chaochen:
please have a check @xla-rotation 

thanks!

xla/service/gpu/launch_dimensions.cc
```diff
@@ -35,18 +36,36 @@ LaunchDimensions CalculateLaunchDimensions(
...
+    int64_t num_blocks_total = CeilOfRatio(num_elements, threads_per_block);
+    int64_t num_blocks_y = CeilOfRatio<uint64_t>(
       num_blocks_total, gpu_device_info.block_dim_limit().x);
-  int64_t num_blocks_x = CeilOfRatio(num_blocks_total, num_blocks_y);
+    int64_t num_blocks_x = CeilOfRatio(num_blocks_total, num_blocks_y);
 
-  return LaunchDimensions(se::BlockDim(num_blocks_x, num_blocks_y, 1),
-                          se::ThreadDim(threads_per_block, 1, 1));
+    return LaunchDimensions(se::BlockDim(num_blocks_x, num_blocks_y, 1),
+                          se::ThreadDim(threads_per_block, 1, 1));\
```
olegshyshkov, line 67: Please remove the extra '\\' at the end of the line.

olegshyshkov (approved):

PR #19649, opened 2024-11-21
[ROCm] Implement hermetic rocm dependency

alekstheod:
This change has as a goal to introduce an external dependency to the rocm library and tools.

Building xla with the hermetic rocm is done by using these env variables:

--repo_env=OS=ubuntu_20.04
--repo_env=ROCM_VERSION=6.2.0

To use only hermetic libs define this flag:
--@local_config_rocm//rocm:use_rocm_hermetic_rpath=True
This flag will make rpaths and configs to look inside the sandbox
If flag is not set then default installation paths are used e.g /opt/rocm

One has to provie OS version and ROCm version to initialize a proper rocm repository.
If these flags are not set then default ROCm installation will be used to build XLA.

depends-on: https://github.com/openxla/xla/pull/19691

### DIFF
--- a/third_party/tsl/third_party/gpus/rocm/BUILD.tpl
+++ b/third_party/tsl/third_party/gpus/rocm/BUILD.tpl
@@ -72,7 +72,7 @@ cc_library(
     visibility = ["//visibility:public"],
     deps = [
         ":rocm_rpath",
-    ]
+    ],
 )
 
 cc_library(
@@ -92,6 +92,7 @@ cc_library(
         ":rocprofiler_register",
         ":rocsolver",
         ":roctracer",
+        ":rocsparse",
     ] + select_threshold(
         above_or_eq = [":hipfft"],
         below = [":rocfft"],
@@ -128,7 +129,6 @@ cc_library(
 
 cc_library(
     name = "hip",
-    data = glob(["%{rocm_root}/lib/**"]),
     visibility = ["//visibility:public"],
     deps = [
         ":rocm_hip",
@@ -140,7 +140,6 @@ cc_library(
     name = "rocm_hip",
     srcs = glob(["%{rocm_root}/lib/libamdhip*.so*"]),
     hdrs = glob(["%{rocm_root}/include/hip/**"]),
-    data = glob(["%{rocm_root}/lib/hip/**"]),
     include_prefix = "rocm",
     includes = [
         "%{rocm_root}/include",
@@ -178,7 +177,6 @@ cc_library(
 cc_library(
     name = "rocfft",
     srcs = glob(["%{rocm_root}/lib/librocfft*.so*"]),
-    data = glob(["%{rocm_root}/lib/librocfft*.so*"]),
     include_prefix = "rocm",
     includes = [
         "%{rocm_root}/include",
@@ -191,7 +189,6 @@ cc_library(
 cc_library(
     name = "hipfft",
     srcs = glob(["%{rocm_root}/lib/libhipfft*.so*"]),
-    data = glob(["%{rocm_root}/lib/libhipfft*.so*"]),
     include_prefix = "rocm",
     includes = [
         "%{rocm_root}/include",
@@ -204,7 +201,6 @@ cc_library(
     name = "hiprand",
     srcs = glob(["%{rocm_root}/lib/libhiprand*.so*"]),
     hdrs = glob(["%{rocm_root}/include/hiprand/**"]),
-    data = glob(["%{rocm_root}/lib/libhiprand*.so*"]),
     include_prefix = "rocm",
     includes = [
         "%{rocm_root}/include",
@@ -239,7 +235,6 @@ cc_library(
     name = "rccl",
     srcs = glob(["%{rocm_root}/lib/librccl*.so*"]),
     hdrs = glob(["%{rocm_root}/include/rccl/**"]),
-    data = glob(["%{rocm_root}/lib/librccl*.so*"]),
     include_prefix = "rocm",
     includes = [
         "%{rocm_root}/include",
@@ -310,7 +305,18 @@ cc_library(
     name = "rocsolver",
     srcs = glob(["%{rocm_root}/lib/librocsolver*.so*"]),
     hdrs = glob(["%{rocm_root}/include/rocsolver/**"]),
-    data = glob(["%{rocm_root}/lib/librocsolver*.so*"]),
+    include_prefix = "rocm",
+    includes = [
+        "%{rocm_root}/include/",
+    ],
+    strip_include_prefix = "%{rocm_root}",
+    visibility = ["//visibility:public"],
+    deps = [":rocm_config"],
+)
+
+cc_library(
+    name = "rocsparse",
+    srcs = glob(["%{rocm_root}/lib/librocsparse*.so*"]),
     include_prefix = "rocm",
     includes = [
         "%{rocm_root}/include/",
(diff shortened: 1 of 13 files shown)


beckerhe:
This is really exciting. Thank you. I won't be able to have a look before next week though.

third_party/tsl/third_party/gpus/rocm/BUILD.tpl
```diff
@@ -354,6 +355,21 @@ cc_library(
     deps = [":rocm_config"],
 )
 
+cc_library(
+    name = "libnuma",
+    srcs = glob(["rocm_dist/usr/lib/**/libnuma.so*"]),
+)
```
alekstheod, line 361: This has to be removed.

third_party/tsl/third_party/gpus/rocm/rocm_redist_ubuntu_20_04.bzl
```diff
@@ -0,0 +1,175 @@
+rocm_redist_ubuntu_20_04 = {
+    "6.2.0": {
+        "archives": [
+            struct(
+                url = "https://repo.radeon.com/rocm/apt/6.2/pool/main/c/comgr6.2.0/comgr6.2.0_2.8.0.60200-66~20.04_amd64.deb",
```
JehandadKhan, line 5: ⚓

third_party/tsl/third_party/gpus/rocm_configure.bzl
```diff
@@ -578,6 +568,53 @@ def _compute_rocm_extra_copts(repository_ctx, amdgpu_targets):
...
+
+def _remove_root_dir(path, root_dir):
+    if path.startswith(root_dir + "/"):
+        return path[len(root_dir) + 1:]
+    return path
+
+def _setup_rocm_distro_dir(repository_ctx):
+    """Sets up the rocm hermetic installation directory to be used in hermetic build"""
+    bash_bin = get_bash_bin(repository_ctx)
+    os = repository_ctx.os.environ.get(_OS)
```
ybaturina, line 604: can we identify Ubuntu version using `repository_os` data?
https://bazel.build/rules/lib/builtins/repository_os.html

third_party/tsl/third_party/gpus/rocm_configure.bzl
```diff
@@ -888,6 +874,8 @@ _ENVIRONS = [
     "TF_NEED_CUDA",  # Needed by the `if_gpu_is_configured` macro
     _ROCM_TOOLKIT_PATH,
     _TF_ROCM_AMDGPU_TARGETS,
+    _OS,
+    _ROCM_VERSION,
 ]
 
 remote_rocm_configure = repository_rule(
```
ybaturina, line 881: I think you'll need to remove it from here and from https://github.com/openxla/xla/blob/23c490876ad229ba6e481f1ab8e2a878a97d67ae/tools/toolchains/remote_config/rbe_config.bzl#L45.

The reason is that by implementing hermetic ROCM repo rules, you provide the ROCM data to Bazel cache on all machines, including RBE.

alekstheod: There are 2 modes in rocm build right now. The first one is like it was originally. We use the local rocm installation and just link the directory of it to our sandbox. The second one is the hermetic build where we bring the files from our deb repo. If we remove that remote_rocm_configure then we shall be sure that remote execution is using hermetic build. Shall it be done in this PR or better create a follow-up pr that would do exactly that thing?

third_party/tsl/third_party/gpus/rocm_configure.bzl
alekstheod: it doesn't look like we can. repository_os.name would contain 'linux' inside. There is no version field in os. Also I guess in a future we would like to do crosscompile so to build on ubuntu 22.04 but for ubuntu 24.04 or on ubuntu  for rhel and so on?

third_party/tsl/third_party/gpus/rocm/BUILD.tpl
```diff
@@ -1,8 +1,22 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@local_config_rocm//rocm:build_defs.bzl", "rocm_version_number", "select_threshold")
 
 licenses(["restricted"])  # MPL2, portions GPL v3, LGPL v3, BSD-like
 
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//visibility:private"])
+
+bool_flag(
```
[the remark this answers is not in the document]
alekstheod: in this implementation hermetic -> boolean flag mean use rpath from inside the sandbox. This is needed to execute the tests without rocm installation. I think this flag is a bit different with what is in cuda. Could you suggest a proper name that would fit the convention?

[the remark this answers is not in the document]
ybaturina: Got it. In this case suggest naming config_setting as noun and bool_flag - as a verb.

third_party/tsl/third_party/gpus/rocm_configure.bzl
ybaturina: In this case you can try `repository_ctx.execute(["uname", "-a"]).stdout`. It will print smth like this:

`Linux tf-temp-ubuntu-ybaturina-2024-11-25 5.15.0-1062-gcp #70~20.04.1-Ubuntu SMP Fri May 24 20:12:18 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`

I think that `OS` can be used for cross-compile build, but for the regular one the repository rule should be able to identify the OS name and version.

third_party/tsl/third_party/gpus/rocm_configure.bzl
ybaturina: Got it. Yes, I agree that it can be addressed in a separate PR.

beckerhe (approved):

beckerhe (approved):

beckerhe (approved):

alekstheod:
> This PR makes MacOS build fail with `ld: unknown option: -rpath=/opt/rocm/lib`.
> 
> Any ideas how to fix it?
Maybe we shall use -rpath,/opt/rocm/lib instead of -rpath=/opt/rocm/lib?
here https://github.com/openxla/xla/blob/f512c07368effd7ba6073c0fc07670297e9c4f6c/third_party/tsl/third_party/gpus/rocm/BUILD.tpl#L124

alekstheod:
> I've replaced `=` with `,` as it's indeed what internet suggests, let's see whether it fixes the build.
> 
> However, it would be nice not to include ROCm dependency in CPU builds at all.

I think some of the target dependency to rocm is not guarded by if_rocm_is_configured. So the dependency got leaked through.
There are 2 approaches to fix this. One is to find which exactly target has a dependency to rocm_rpath (through bazel query?) when compiled on macos and guard it with if_rocm_is_configured. The other approach implement the guard inside the BUILD.tpl file itself like so:
`

cc_library(
    name = "rocm_path_impl",
    linkopts = select({
        ":build_hermetic": [
            "-Wl,-rpath=%{rocm_toolkit_path}/lib",
        ],
        "//conditions:default": [
            "-Wl,-rpath=/opt/rocm/lib",
        ],
    }),
)

cc_library(
    name = "rocm_rpath",
    visibility = ["//visibility:public"],
    deps = if_rocm_is_configured([":rocm_path_impl"]),
)
`

PR #19655, opened 2024-11-21
[ROCm] Make MLIR Math dialect lowering more deterministic

draganmladjenovic:
First apply patterns from GpuToROCDLConversionPatterns then do the cleanup with MathToLLVMConversionPatterns

### DIFF
--- a/xla/service/gpu/fusions/transforms/lower_to_llvm.cc
+++ b/xla/service/gpu/fusions/transforms/lower_to_llvm.cc
@@ -68,8 +68,6 @@ class LowerToLLVMPass : public impl::LowerToLLVMPassBase<LowerToLLVMPass> {
     mlir::arith::populateArithExpandOpsPatterns(patterns);
     mlir::arith::populateArithToLLVMConversionPatterns(type_converter,
                                                        patterns);
-    mlir::populateMathToLLVMConversionPatterns(type_converter, patterns,
-                                               /* approximateLog1p */ false);
     if (!this->is_amd_gpu_) {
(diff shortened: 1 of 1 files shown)


akuegel (commented):
Do you have an example where it is currently not deterministic?
I remember that currently we handle the conflict between lowering to LLVM and lowering to GPU specific functions by declaring several LLVM ops as illegal. I agree this is a bit of a hack, but I thought it works?

draganmladjenovic:
Last time it was log1p/log/sqrt, but for time to time we had random crashes on other function changing between builds. If adding illegal ops fixes things I'm open going that way. Look by far cleaner. I need to check what happens when MathToLLVM tries to use unsupported op as part of lowering of a "complex" op.  What makes me uneasy I'm not quite clear if our bitcode implementation produces same results as the intrinsics, since unlike cuda our implementation is not a wrapper over intrinsic. So i chose to give priority to ROCDL patterns.

akuegel:
> Last time it was log1p/log/sqrt, but for time to time we had random crashes on other function changing between builds. If adding illegal ops fixes things I'm open going that way. Look by far cleaner. I need to check what happens when MathToLLVM tries to use unsupported op as part of lowering of a "complex" op. What makes me uneasy I'm not quite clear if our bitcode implementation produces same results as the intrinsics, since unlike cuda our implementation is not a wrapper over intrinsic. So i chose to give priority to ROCDL patterns.

Ok, thanks for the explanation.

akuegel (approved):
Normally we require a test, but I think this can be counted as non-functional change (at least on Nvidia backend).

PR #19660, opened 2024-11-21
[ROCm] switch rocm build to clang

alekstheod:
This PR switches the default rocm build to clang as the gcc config is broken at the moment.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -283,7 +283,6 @@ build:rocm --crosstool_top=@local_config_rocm//crosstool:toolchain
 build:rocm --define=using_rocm_hipcc=true
 build:rocm --define=tensorflow_mkldnn_contraction_kernel=0
 build:rocm --repo_env TF_NEED_ROCM=1
-build:rocm --action_env=CLANG_COMPILER_PATH="/usr/lib/llvm-18/bin/clang"
 build:rocm --action_env=TF_ROCM_CLANG="1"
 build:rocm --linkopt="-fuse-ld=lld"
 build:rocm --host_linkopt="-fuse-ld=lld"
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -283,6 +283,10 @@ build:rocm --crosstool_top=@local_config_rocm//crosstool:toolchain
 build:rocm --define=using_rocm_hipcc=true
 build:rocm --define=tensorflow_mkldnn_contraction_kernel=0
 build:rocm --repo_env TF_NEED_ROCM=1
+build:rocm --action_env=CLANG_COMPILER_PATH="/usr/lib/llvm-18/bin/clang"
```
beckerhe, line 286: This would break users that don't use Clang, a different version of clang, or just clang installed into a different path.

I assume the intention is to change it for the CI, correct? In this case I would recommend creating a separate config (similar to `cuda_clang_official` above) which inherits all settings from the `rocm` config and also sets these extra flags.

But that means you will also have to change the CI to actually use this new config.

alekstheod: You are right. It will break. I think we can remove the CLANG_COMPILER_PATH from this config.
This PR is a workaround as our gcc toolchain is broken at the moment. Our gcc build will resolve the symbolic links
and will break our hermetic build (see this PR: https://github.com/openxla/xla/pull/19649)

beckerhe: Okay cool. But this would still break GCC users, wouldn't it? Isn't there a way to do this without breaking GCC?

Also I don't think I understand the comment about resolving symbolic links, so maybe I'm just missing the point.

alekstheod: I adjusted the config based on your suggestion.
I created new PR fixing the gcc. It something that works and will unblock our hermetic build PR.
We maybe revisit this solution later and see how to move the compilation flags to the toolchain files instead
of keeping them in our script.
https://github.com/openxla/xla/pull/19691

beckerhe (approved):

PR #19699, opened 2024-11-22
Explicit stream annotation: Set ExecutionStreamId based on frontend attribute

chaserileyroberts:
This PR picks up the stream annotation frontend attribute on async methods and assigns the matching ExecutionStreamId.

This PR is another part of breaking up PR #17982.

### DIFF
--- a/xla/service/gpu/execution_stream_assignment.cc
+++ b/xla/service/gpu/execution_stream_assignment.cc
@@ -17,12 +17,16 @@ limitations under the License.
 
 #include <deque>
 #include <memory>
+#include <optional>
+#include <string>
 #include <utility>
 
 #include "absl/container/flat_hash_map.h"
 #include "absl/log/check.h"
+#include "absl/strings/numbers.h"
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
+#include "xla/side_effect_util.h"
 #include "xla/hlo/ir/hlo_casting_utils.h"
 #include "xla/hlo/ir/hlo_computation.h"
 #include "xla/hlo/ir/hlo_instruction.h"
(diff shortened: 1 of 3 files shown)


frgossen (changes requested):

xla/service/gpu/execution_stream_assignment_test.cc
```diff
@@ -240,6 +240,64 @@ TEST_F(ExecutionStreamAssignmentTest, UnreachableComputation) {
                 StatusIs(absl::StatusCode::kNotFound));
   }
 }
+TEST_F(ExecutionStreamAssignmentTest, CalledComputations) {
```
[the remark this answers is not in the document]
chaserileyroberts: I did my best to remove the noise from the HLO, but the cublass custom call has a lot of attributes that I wasn't sure I could safely remove/ignore.

frgossen (approved):
Thanks for breaking this into smaller PRs.

xla/service/gpu/execution_stream_assignment.cc
```diff
@@ -110,10 +117,20 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
...
-        CHECK_EQ(callsite.instruction()->opcode(), HloOpcode::kAsyncStart);
-        enqueue_called_computations(callsite, next_stream_id);
+        // Special logic for explicit stream annotations.
+        std::optional<ExecutionStreamId> optional_stream_id = std::nullopt;
+        auto instruction = callsite.instruction();
+        auto it = instruction->frontend_attributes().map().find(
+            kXlaStreamAnnotationAttr);
+        if (it != instruction->frontend_attributes().map().end()) {
+          int stream_id;
+          absl::SimpleAtoi(it->second, &stream_id);
```
mooskagh, line 127: Could you check the return value of this, so that the case where the frontend attribute is not a proper number, is handled?

mooskagh (changes requested):

xla/service/gpu/execution_stream_assignment.cc
chaserileyroberts: I added `CHECK(...)` around this method.

PR #19754, opened 2024-11-24
[ROCm] Enable gemm fusion autotuner.

zoranjovanovic-ns:


xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc
```diff
@@ -306,13 +323,26 @@ class GemmFusionAutotunerTest : public StatelessAutotunerTest {
     return debug_options;
   }
 
+  stream_executor::GpuComputeCapability CudaAmpereOrRocm() {
```
dimitar-asenov, line 326: This function can be somewhat confusing. I suggest you simplify the logic and directly inline it in the single place where it is used:

```
if (IsRocm()) {
return GetRocmComputeCapability();
} else {
...
}
```
I've another comment suggesting you introduce `IsRocm()`, and it's probably also good to have `GetRocmComputeCapability()`, similarly to how we have a function for cuda.

xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc
```diff
@@ -1260,6 +1316,10 @@ TEST_F(GemmFusionAutotunerTest,
 }
 
 TEST_F(GemmFusionAutotunerTest, RewritesGemmFusionToCustomKernelFusion) {
+  if (std::holds_alternative<se::RocmComputeCapability>(
```
dimitar-asenov, line 1319: This pattern appears quite often. Please make a peer to `GpuComputeComp` called `IsRocm` that does the check and use it everywhere instead.

xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc
dimitar-asenov: Once you introduce those two new function, please make a pass over the entire file and use them where appropriate.

xla/service/gpu/autotuning/BUILD
```diff
@@ -28,7 +32,9 @@ package_group(
 
 cc_library(
     name = "gemm_fusion_autotuner",
-    srcs = ["gemm_fusion_autotuner.cc"],
+    srcs = ["gemm_fusion_autotuner.cc"] +
+            if_cuda_is_configured(["gemm_fusion_autotuner_cuda.cc"]) +
+            if_rocm_is_configured(["gemm_fusion_autotuner_rocm.cc"]),
     hdrs = ["gemm_fusion_autotuner.h"],
     tags = [
         "cuda-only",
```
dimitar-asenov, line 40: You should remove this tag.

xla/service/gpu/autotuning/gemm_fusion_autotuner.h
```diff
@@ -198,6 +197,26 @@ class GemmFusionAutotunerImpl {
...
+  int block_k = 0;
+};
+
+absl::StatusOr<TileSizeLimit> GetLimits(const HloDotInstruction& dot);
+bool IsFusionKind(const HloInstruction& hlo, absl::string_view kind);
+
+// Minimum tile size.
+constexpr int kMinTileSize = 16;
+
+using BackendConfig = GemmFusionAutotunerImpl::BackendConfig;
```
dimitar-asenov, line 212: Please do not put `using` declarations in header files.

xla/service/gpu/autotuning/gemm_fusion_autotuner.h
```diff
@@ -198,6 +197,26 @@ class GemmFusionAutotunerImpl {
   std::vector<TritonGemmConfig> triton_configs_;
 };
 
+struct TileSizeLimit {
```
dimitar-asenov, line 200: All of the declarations below are currently directly in the `xla::gpu` namespace, but shouldn't be. The reason is that their name are too generic for top-level symbols.

Would moving these as members of `GemmFusionAutotuner` and/or `GemmFusionAutotunerImpl` also work?

xla/service/gpu/autotuning/gemm_fusion_autotuner_rocm.cc
```diff
@@ -0,0 +1,306 @@
...
+  TF_ASSIGN_OR_RETURN(std::vector<TritonGemmConfig> triton_configs,
+                      GenerateTritonConfigs(*dot));
+  for (TritonGemmConfig& config : triton_configs) {
+    configs.push_back(std::move(config));
+  }
+  return configs;
+}
+
+absl::StatusOr<std::vector<TritonGemmConfig>>
+GemmFusionAutotunerImpl::GenerateTritonConfigs(const HloDotInstruction& dot) {
```
dimitar-asenov, line 215: This method is too long, please break it down into multiple methods.

xla/service/gpu/autotuning/gemm_fusion_autotuner_cuda.cc
```diff
@@ -0,0 +1,383 @@
+/* Copyright 2024 The OpenXLA Authors.
```
dimitar-asenov, line 1: I looked at the diff between this file and the `_rocm` version. At the moment it is very small and I think we'll be much better off if we keep this a single file with a few places where the code branches. Also in general, I suspect that's desirable for the long-term.

Do you have a strong reason why splitting this code and duplicating a lot of it is desirable? If not, please revert this change.

xla/service/gpu/autotuning/gemm_fusion_autotuner.h
dimitar-asenov: Also look at my comment on the new `_cuda` file. Perhaps we can simply revert the change here as well, as suggested in that comment.

dimitar-asenov (changes requested):

xla/service/gpu/autotuning/gemm_fusion_autotuner_cuda.cc
zoranjovanovic-ns: The reason for creating two files (_cuda and _rocm versions) is that there are parts of the code that cannot be compiled as a part of cuda or rocm build (e.g. missing cuda header files in rocm build).
Keeping it in single file would require introduction of #if GOOGLE_CUDA (or similar) guards.
My understanding is that introduction of these new #ifdef is not acceptable,

xla/service/gpu/autotuning/gemm_fusion_autotuner_rocm.cc
zoranjovanovic-ns: Just copied existing corresponding Cuda method. If it necessary, I will break it to several smaller methods.

xla/service/gpu/autotuning/gemm_fusion_autotuner_cuda.cc
dimitar-asenov: You are correct, we should not introduce new `#ifdef` macros. However, my comment still mostly stands. We need a hybrid solution:

- Leave all common code in the current file.
- Isolate only parts that require Rocm/Cud headers into the new `_rocm`, `_cuda` files. If I look at the diff between the two, I believe that's just:
  - `const int64_t BLAS_GEMM_DEFAULT`
  - `GetCuDnnPlanCount` and more generally the CuDNN bit of code.
- Everything else could stay in one file. E.g. for `GetDefaultTritonConfigs()` the code can just add the common configs unconditionally, then check for cuda and add the remaining bits.

@beckerhe For having two different implementations of a common header (`rocm` and `cuda`), is the current solution OK (different `srcs` based on `if_rocm/cuda_configured`)? Is it preferable to have two entirely separate targets that are then conditionally added to `deps`, e.g. like we do [here](https://github.com/openxla/xla/blob/main/xla/mlir/tools/mlir_interpreter/dialects/BUILD#L73-L110)?

beckerhe: > Is it preferable to have two entirely separate targets that are then conditionally added to deps, e.g. like we do [here](https://github.com/openxla/xla/blob/main/xla/mlir/tools/mlir_interpreter/dialects/BUILD#L73-L110)?

Yes, that would be preferrable for a couple of reasons:

1. Both targets can have different dependencies and different build flags if needed
2. Dependency cleaners like `build_cleaner` only work if these are separate targets

zoranjovanovic-ns:
@dimitar-asenov Pushed new version

xla/service/gpu/autotuning/gemm_fusion_autotuner_rocm.cc
dimitar-asenov: Right, the original method is also too long. If you end up modifying the original (if there's just one left after the proposals) or if you fork it, please break it down. It's good to do minor clean ups as we touch old code.

xla/service/gpu/autotuning/BUILD
```diff
@@ -31,10 +31,11 @@ package_group(
 )
 
 cc_library(
-    name = "gemm_fusion_autotuner",
-    srcs = ["gemm_fusion_autotuner.cc"] +
-            if_cuda_is_configured(["gemm_fusion_autotuner_cuda.cc"]) +
-            if_rocm_is_configured(["gemm_fusion_autotuner_rocm.cc"]),
+    name = "gemm_fusion_autotuner_cuda",
+    srcs = [
+        "gemm_fusion_autotuner.cc",
```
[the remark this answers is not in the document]
beckerhe: Thank you. This looks good to me. (Only looked a the build rule things)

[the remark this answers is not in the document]
zoranjovanovic-ns: @dimitar-asenov Modified build rules.

xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
```diff
@@ -1267,33 +1241,45 @@ GemmFusionAutotunerImpl::GetExhaustiveTritonConfigs() const {
...
-      Config(16, 64, 256, 8, 1, 4),   Config(256, 256, 128, 1, 3, 8)};
-  if (GetComputeCapability().IsAtLeastHopper()) {
-    absl::c_copy(
-        std::vector<Config>{
-            Config(16, 32, 32, 8, 1, 2),
-            Config(16, 64, 128, 8, 1, 4),
-            Config(16, 64, 128, 16, 3, 4),
-        },
-        std::back_inserter(configs));
+  if(isRocm()) {
```
dimitar-asenov, line 1244: As far as I can see all the configs here apply to both Rocm and Cuda. Please remove the IsRocm codition and just add these undconditionally. Then below, you can have an `isCuda` and add only what's relevant there.

That makes it a bit more obvious what is common and what not.

xla/service/gpu/autotuning/gemm_fusion_autotuner.h
```diff
@@ -183,11 +185,21 @@ class GemmFusionAutotunerImpl {
...
-  se::CudaComputeCapability GetComputeCapability() const {
-    return std::get<se::CudaComputeCapability>(
-        config_.GetGpuComputeCapability());
+  se::GpuComputeCapability GetComputeCapability() const {
+    return config_.GetGpuComputeCapability();
+  }
+
+  bool isRocm() const {
+    return std::holds_alternative<se::RocmComputeCapability>(
+          GetComputeCapability()) ? true : false;
```
dimitar-asenov, line 194: Please remove `? true : false`, the result of `holds_alternative` is already what you want.

xla/service/gpu/autotuning/gemm_fusion_autotuner_rocm.cc
```diff
@@ -0,0 +1,127 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/service/gpu/autotuning/gemm_fusion_autotuner.h"
+#include "xla/stream_executor/rocm/rocm_blas.h"
+
+#include <algorithm>
```
dimitar-asenov, line 19: Please remove all unused headers (and corresponding deps in the `BUILD` file).

xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc
```diff
@@ -207,6 +211,26 @@ class StatelessAutotunerTest : public HloTestBase {
...
+        .rocm_compute_capability();
+  }
+
+  const stream_executor::GpuComputeCapability& GpuComputeComp() {
+    return backend().default_stream_executor()
+                    ->GetDeviceDescription().gpu_compute_capability();
+  }
+
+  bool isRocm() {
+    if (std::holds_alternative<se::RocmComputeCapability>(
```
dimitar-asenov, line 227: The entire body of this function should just be:
```
return std::holds_alternative<se::RocmComputeCapability>(
          GpuComputeComp());
```

dimitar-asenov (changes requested):

xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
zoranjovanovic-ns: I would like to separate parts of the code for adding configs for rocm and cuda.
I do expect adding/removing new configs for rocm and would like to avoid changing cuda related code when adding/removing configs for rocm.

dimitar-asenov: In that case, please implement this function separately in the `_cuda` and `_rocm` cc files, since the logic for the two cases will be completely disjoint.

dimitar-asenov (changes requested):

dimitar-asenov (approved):

dimitar-asenov:
Thank you for patiently addressing all of our comments!

PR #19834, opened 2024-11-26
[ds-fusion]Add support for async dynamic slice fusion

shraiysh:
This patch adds async handling to dynamic slice fusion when the hero operation is a collective operation. Currently, only reduce-scatter is supported as a hero operation in dynamic slice thunk, so this patch also follows the same.

Added a test with compute, to ensure that communication and compute overlap in the thunks emitted.


frgossen (commented):
Do you have a workload for which you see improvements? How can we verify and track this performance imporvement?

shraiysh:
> Do you have a workload for which you see improvements? How can we verify and track this performance imporvement?

I see performance improvements with async work + runtime calculation of offset work in llama 2 7b on paxml. I have not tested this for other models, but I expect better performance there too. I don't know if there is a way to verify this until the patches are done. I can share one branch with all the changes so you can reproduce the performance benefits. Does that work?

shraiysh:
I will try to run this on maxtext soon. This change along with #19444 gives ~4-5% improvement in llama 2 7b on paxml.

shraiysh:
Ping for review.

I tried running this on maxtext, and although this optimization is applicable there, there are some `convert` operations, `reduce-precision` operations and multiple uses of dynamic-slice. All of these can be handled (eliminated or peeled with simple passes), and I am working on writing passes for handling them. On PAXML, this optimization works as-is.

xla/service/gpu/transforms/command_buffer_scheduling.cc
```diff
@@ -97,6 +97,19 @@ static bool IsNoOp(const HloInstruction* hlo) {
 // done operation is not part of the same command buffer, we would change the
 // execution semantics and create additional synchronization point.
 
+// Returns true if the hlo instruction is async-start for dynamic-slice-fusion.
```
bchetioui, line 100: nit-ish: Where do you check that this is an `async-start`? Is `async-start` the only op that can have an `async_wrapped_instruction`? I might want to slightly edit the comment to say "Returns 'true' if `hlo` is an async wrapper around a dynamic-slice fusion."

xla/service/gpu/transforms/command_buffer_scheduling.cc
```diff
@@ -97,6 +97,19 @@ static bool IsNoOp(const HloInstruction* hlo) {
...
+static bool IsAsyncDynamicSliceFusion(const HloInstruction* hlo) {
+  HloInstruction* wrapped = hlo->async_wrapped_instruction();
+  absl::StatusOr<GpuBackendConfig> backend_config =
+      wrapped->backend_config<GpuBackendConfig>();
+  if (!backend_config.ok()) {
+    return false;
+  }
+  std::string name =
+      backend_config->fusion_backend_config().custom_fusion_config().name();
+  return name == "address_computation" || name == "dynamic_address_computation";
```
bchetioui, line 110: Same comment as above re. `constexpr` (and `absl::string_view` on the line above, if it can be used).

xla/service/gpu/transforms/command_buffer_scheduling_test.cc
```diff
@@ -1117,79 +1117,6 @@ TEST_F(CommandBufferSchedulingTest, AsyncAlltoAll) {
                             });
 }
 
-TEST_F(CommandBufferSchedulingTest, DynamicSliceFusionDynamicSlicing) {
```
bchetioui, line 1120: Can we turn this test into a useful one for the new behavior?

E.g., maybe we can check that we have command buffers in certain cases, now?

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3269,91 @@ TEST_F(DynamicSliceFusionTest,
...
+  ASSERT_NE(result_offset_0 && result_offset_1 && result_offset_2, false);
+  EXPECT_EQ(*result_offset_0, 1l);
+  EXPECT_EQ(*result_offset_1, 0l);
+  EXPECT_EQ(*result_offset_2, 0l);
+
+  // Finally, compare the results.
+  ErrorSpec error{1e-4, 1e-4};
+  EXPECT_TRUE(RunAndCompareTwoModulesReplicated(std::move(hlo_module_ref),
+                                                std::move(hlo_module_opt),
+                                                false, true, error));
```
bchetioui, line 3354: Use `/*param_name=*/` to prefix literal arguments

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3269,91 @@ TEST_F(DynamicSliceFusionTest,
...
+  hlo_module->mutable_config()
+      .mutable_debug_options()
+      .set_xla_gpu_enable_dynamic_slice_fusion(true);
+  TF_ASSERT_OK_AND_ASSIGN(hlo_module,
+                          GetOptimizedModule(std::move(hlo_module)));
+  TF_ASSERT_OK_AND_ASSIGN(hlo_module_ref,
+                          GetOptimizedModule(std::move(hlo_module_ref)));
+  std::unique_ptr<HloModule> hlo_module_opt = hlo_module->Clone();
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<Executable> exec,
+                          CreateExecutable(std::move(hlo_module), false));
```
bchetioui, line 3308: Use `/*param_name=*/` to prefix literal arguments

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3269,91 @@ TEST_F(DynamicSliceFusionTest,
       /*run_hlo_passes=*/false, /*use_threads=*/true, error));
 }
 
+TEST_F(DynamicSliceFusionTest, AsyncCollective) {
```
bchetioui, line 3272: Let's give a name that explains the behaviour that we're testing to this test.

Is it that "DynamicSliceFusionIsWrappedInsideAsyncCollective"?

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3269,91 @@ TEST_F(DynamicSliceFusionTest,
...
+    }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> hlo_module,
+                          ParseAndReturnVerifiedModule(hlo));
+  std::unique_ptr<HloModule> hlo_module_ref = hlo_module->Clone();
+  hlo_module->mutable_config()
+      .mutable_debug_options()
+      .set_xla_gpu_enable_dynamic_slice_fusion(true);
+  TF_ASSERT_OK_AND_ASSIGN(hlo_module,
+                          GetOptimizedModule(std::move(hlo_module)));
```
bchetioui, line 3303: Do we need to run the whole optimization pipeline? It seems like we should be able to write a unit test for the fusion emitter (checking that lowering an already-optimized async dynamic-slice-fusion produces the right thunks, and everything) and another test that does the optimization end-to-end and produces the right optimized HLO.

The point here is that we could get the same coverage but have much fewer moving parts to check the correctness of our lowering.

bchetioui (changes requested):

xla/service/gpu/transforms/command_buffer_scheduling_test.cc
shraiysh: I have not added a test for that, because that is not supported right now. This patch is getting bigger, I will add that support in a separate patch later. For now, I have ensured that command buffer doesn't consume dynamic slice fusion when there are dynamic addresses in it.

shraiysh:
Rebased and fixed the merge conflict. This is ready for review.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1624,6 +1624,66 @@ ROOT dot = f32[4,8,1024]{2,1,0} dot(p0, p1), lhs_batch_dims={1}, lhs_contracting
...
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(
+      bool filecheck_matched,
+      RunFileCheck(exec->module().ToString(HloPrintOptions{}
+                                               .set_print_operand_shape(false)
+                                               .set_print_metadata(false)),
+                   kExpected));
+  EXPECT_TRUE(filecheck_matched);
+
+  ErrorSpec error{1e-5, 1e-5};
```
bchetioui, line 1681: Can't we have `error{0, 0}`? There should be no issue of associativity with only two replicas, I think?

bchetioui (changes requested):
Sorry, I noticed that I had written comments for a large part of this PR a while back, but forgot to click "Submit review"!!

xla/hlo/parser/hlo_parser.cc
```diff
@@ -1467,18 +1467,37 @@ bool HloParserImpl::ParseInstructionRhs(HloComputation::Builder* builder,
   }
   if (metadata) {
     instruction->set_metadata(*metadata);
```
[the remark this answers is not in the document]
shraiysh: Input hlo:
```
ENTRY main {
  fusion = fusion-start(arg0, arg1, ...), calls=dynamic-slice-fusion, kind=kCustom, config=...{"name": "dynamic_address_computation"}...
}
```
Changes to:
```
async_wrapped {
  fusion = fusion(arg0, arg1, ...), calls=dynamic-slice-fusion, kind=kCustom
}
ENTRY main {
  async-start = async-start(arg0, arg1, ...), calls=async_wrapped, config=...{"name": "dynamic_address_computation"}...
  async-done = async-done(async-start)
}
```
The information about the backend_config is lost from the original fusion instruction and is instead pasted on the wrapping instruction. The syntax fusion-start is a short hand for the more descriptive representation in-memory where we have this wrapping computation too. If the fusion-start operation has an attribute, we should be able to request the value of that attribute from both the wrapped instruction and the wrapping instruction. This is especially important in this scenario for deciding whether the fusion is a dynamic_address_computation or a static address_computation. Without the propagation, this attribute is in the async-start operation and we do not see it anymore on the original fusion instruction, which doesn't seem accurate.

xla/service/gpu/transforms/command_buffer_scheduling_test.cc
```diff
@@ -1326,5 +1243,58 @@ TEST_F(CommandBufferSchedulingTest, ReturnTrueWhenOnlyParamMoved) {
   )");
 }
 
+TEST_F(CommandBufferSchedulingTest,
+       DynamicSliceFusionWithDynamicAddressesNotACommand) {
```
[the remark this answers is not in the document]
shraiysh: Yes, we intend to change it later.

xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3260,89 @@ TEST_F(DynamicSliceFusionTest,
...
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> hlo_module,
+                          ParseAndReturnVerifiedModule(hlo));
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<Executable> exec,
+      CreateExecutable(hlo_module->Clone(), /*run_hlo_passes=*/false));
+  GpuExecutable* gpu_exec = dynamic_cast<GpuExecutable*>(exec.get());
+  const ThunkSequence& thunks = gpu_exec->GetThunk().thunks();
+
+  // This is only needed to ensure that the next checks don't fail.
+  ASSERT_GE(thunks.size(), 6);
```
bchetioui, line 3308: Why `GE`? Don't we know exactly how many thunks we have?

xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3260,89 @@ TEST_F(DynamicSliceFusionTest,
...
+      dynamic_slice_thunk->embedded_thunk());
+  ASSERT_NE(embedded_thunk, nullptr);
+  ASSERT_EQ(embedded_thunk->thunks().size(), 1);
+  EXPECT_EQ(embedded_thunk->thunks()[0]->kind(),
+            Thunk::kNcclReduceScatterStart);
+
+  // Check that the offsets were propagated as constants, and not as device
+  // allocated buffers.
+  auto offsets = dynamic_slice_thunk->get_offsets();
+  auto arg_offset = offsets[0];
```
bchetioui, line 3332: Use a large matcher?

```cpp
EXPECT_THAT(offsets,
            ::testing::ElementsAre(::testing::ElementsAre(std::nullopt),
                                   ::testing::ElementsAre(Optional(1), Optional(0), Optional(0))));
```

xla/hlo/parser/hlo_parser_test.cc
```diff
@@ -5892,5 +5892,54 @@ ENTRY main {
...
+  EXPECT_EQ(wrapped_instr->statistics_viz().DebugString(),
+            "stat_index_to_visualize: 1\nstatistics {\n  stat_name: "
+            "\"stat-1\"\n  stat_val: 33\n}\nstatistics {\n  stat_name: "
+            "\"stat-2\"\n  stat_val: 44\n}\n");
+  EXPECT_EQ(OriginalValueToString(*wrapped_instr->original_value()),
+            "{\"v3\"}");
+  // Check the async-start and async-done instructions.
+  HloInstruction* async_done = m->entry_computation()->root_instruction();
+  HloInstruction* async_start = async_done->async_chain_start();
+  EXPECT_EQ(async_start->metadata().DebugString(),
```
bchetioui, line 5930: Could we just check that those correspond to what we've already checked on `wrapped_instr`?

(i.e. `EXPECT_EQ(async_start->property, wrapped_instr->property)`)

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1746,6 +1746,65 @@ ROOT dot = f32[4,8,1024]{2,1,0} dot(p0, p1), lhs_batch_dims={1}, lhs_contracting
...
+                          ParseAndReturnVerifiedModule(hlo));
+  std::unique_ptr<HloModule> m_ref = m->Clone();
+  m->mutable_config()
+      .mutable_debug_options()
+      .set_xla_gpu_enable_dynamic_slice_fusion(true);
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<Executable> exec,
+      CreateExecutable(m->Clone(), /*run_hlo_passes=*/true));
+  const char* kExpected = R"(
+    // CHECK:      dynamic-slice-fusion{{.+}} {
```
bchetioui, line 1783: This is a bit brittle in terms of names. Could we use a matcher to capture this name?

I also wonder if that whole thing would not look nicer using XLA's matchers?

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1746,6 +1746,65 @@ ROOT dot = f32[4,8,1024]{2,1,0} dot(p0, p1), lhs_batch_dims={1}, lhs_contracting
...
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m,
+                          ParseAndReturnVerifiedModule(hlo));
+  std::unique_ptr<HloModule> m_ref = m->Clone();
+  m->mutable_config()
+      .mutable_debug_options()
+      .set_xla_gpu_enable_dynamic_slice_fusion(true);
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<Executable> exec,
+      CreateExecutable(m->Clone(), /*run_hlo_passes=*/true));
```
bchetioui, line 1781: Why do we need to create an executable?

Can't we just `GetOptimizedModule`?

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1746,6 +1746,65 @@ ROOT dot = f32[4,8,1024]{2,1,0} dot(p0, p1), lhs_batch_dims={1}, lhs_contracting
...
+    // CHECK-NEXT:   %[[fusion_done:.+]] = {{.+}} fusion-done(%[[fusion_start]]), {{.+}}"name":"dynamic_address_computation"
+    // CHECK:        ROOT {{.+}} = {{.+}} tuple(%[[fusion_done]], %[[wrapped_dot]])
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(
+      bool filecheck_matched,
+      RunFileCheck(exec->module().ToString(HloPrintOptions{}
+                                               .set_print_operand_shape(false)
+                                               .set_print_metadata(false)),
+                   kExpected));
+  EXPECT_TRUE(filecheck_matched);
```
bchetioui, line 1801: Could just be `EXPECT_THAT(RunFileCheck..., ::testing::IsOkAndHolds(true))`

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1746,6 +1746,65 @@ ROOT dot = f32[4,8,1024]{2,1,0} dot(p0, p1), lhs_batch_dims={1}, lhs_contracting
...
+
+  RunAndCompareTwoModulesReplicated(std::move(m), std::move(m_ref),
```
bchetioui, line 1803: This seems beyond what the test claims it is testing. We can probably omit this part of the test?

xla/service/gpu/ir_emission_utils.cc
```diff
@@ -638,5 +639,40 @@ absl::StatusOr<std::string> GetProtoFingerprint(
...
+    return std::nullopt;
+  }
+  return fusion_backend_config.custom_fusion_config().name();
+}
+
+// Returns true if the given instruction is a custom fusion for dynamic slice
+// fusion. This is determined by checking the name of custom fusion config.
+bool IsDynamicSliceFusion(const HloInstruction* instr) {
+  std::optional<std::string> name = GetCustomFusionConfigName(instr);
+  if (!name.has_value()) {
```
bchetioui, line 670: This path is unnecessary, `optional` can directly check equality between an `optional` and a value.

xla/service/gpu/ir_emission_utils.h
```diff
@@ -88,6 +88,18 @@ inline constexpr absl::string_view kUncompilableFusion =
 
 inline constexpr absl::string_view kTopKCustomCallTarget = "__gpu$TopK";
 
+inline constexpr absl::string_view
```
bchetioui, line 91: Please add brief documentation on what these new constant represents (see above for `__triton` fusions for example).

xla/service/gpu/ir_emission_utils.h
```diff
@@ -88,6 +88,18 @@ inline constexpr absl::string_view kUncompilableFusion =
...
+    kDynamicSliceFusionWithStaticAddressComputationConfigName =
+        "address_computation";
+inline constexpr absl::string_view
+    kDynamicSliceFusionWithDynamicAddressComputationConfigName =
+        "dynamic_address_computation";
+
+std::optional<std::string> GetCustomFusionConfigName(
```
bchetioui, line 98: Exposed utils should have docstrings, and those should be in the `.h` file (it seems like both functions have docstrings in the `.cc` file for some reason).

xla/service/gpu/transforms/command_buffer_scheduling.cc
```diff
@@ -109,7 +109,17 @@ static bool IsAsyncStartCommand(const HloInstruction* hlo,
       return config.enabled_commands.contains(DebugOptions::CUBLAS);
     }
     if (hlo->async_wrapped_opcode() == HloOpcode::kFusion) {
-      return config.enabled_commands.contains(DebugOptions::FUSION);
+      // Currently only supporting static address computation in command buffer.
```
bchetioui, line 112: [We] currently only support static address computation[s] in command buffer[s].

xla/service/gpu/transforms/command_buffer_scheduling.cc
```diff
@@ -109,7 +109,17 @@ static bool IsAsyncStartCommand(const HloInstruction* hlo,
...
+
```
bchetioui, line 113: Remove unnecessary linefeed.

xla/service/gpu/transforms/command_buffer_scheduling.cc
```diff
@@ -136,7 +146,16 @@ static bool IsAsyncDoneCommand(const HloInstruction* hlo,
       return config.enabled_commands.contains(DebugOptions::CUBLAS);
     }
     if (hlo->async_wrapped_opcode() == HloOpcode::kFusion) {
-      return config.enabled_commands.contains(DebugOptions::FUSION);
+      // Currently only supporting static address computation in command buffer.
```
bchetioui, line 149: [We] currently only support static address computation[s] in command buffer[s].

xla/service/gpu/transforms/command_buffer_scheduling_test.cc
```diff
@@ -1206,26 +1133,33 @@ TEST_F(CommandBufferSchedulingTest, DynamicSliceFusionStaticSlicing) {
...
     input = s32[8,32]{1,0} reshape(slice)
-    ROOT rs = s32[4,32] reduce-scatter(input), channel_id=64, replica_groups={{0,1}}, use_global_device_ids=true, dimensions={0}, to_apply=add
+    rs = s32[4,32] reduce-scatter(input), channel_id=64, replica_groups={{0,1}}, use_global_device_ids=true, dimensions={0}, to_apply=add
+    dot = s32[128,128] dot(a,b), lhs_contracting_dims={1}, rhs_contracting_dims={0}
+    ROOT tuple = tuple(rs, dot)
   })";
 
-  TF_ASSERT_OK_AND_ASSIGN(auto m, GetOptimizedModule(hlo));
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m,
+                          ParseAndReturnVerifiedModule(hlo));
```
bchetioui, line 1148: nit: maybe we could create the config before calling `ParseAndReturnVerifiedModule` and pass it as an additional parameter?

xla/service/gpu/transforms/command_buffer_scheduling_test.cc
```diff
@@ -1249,32 +1184,14 @@ TEST_F(CommandBufferSchedulingTest, DynamicSliceFusionStaticSlicing) {
     ASSERT_NE(child->kind(), Thunk::kCommandBuffer);
   }
 
-  // FUSION off, COLLECTIVES on -> command buffer because static slices.
```
bchetioui, line 1252: This test has become a little unreadable now, I'm not sure I understand what's being tested. Some cases have disappeared, and others (e.g. `FUSION on, COLLECTIVES on` above) no longer match the comments.

Can we restructure this to follow a structured pattern?

If you want to set default options for the whole test, it's OK to subclass `CommandBufferSchedulingTest` to enable a few more debug options. That might declutter the test a little as well.

bchetioui (changes requested):

xla/service/gpu/gpu_compiler_test.cc
shraiysh: With just `GetOptimizedModule`, the HLO would not be scheduled. Here, I want to see that the gemm and collective overlap. That is why, we need to create the executable.

xla/service/gpu/gpu_compiler_test.cc
shraiysh: We would have the same issue here: with a matcher, we either have to list all the instructions to match all the ops generated in ENTRY computation, or we will lose the schedule information that is evident in the ordering of operations in HLO.

xla/service/gpu/gpu_compiler_test.cc
shraiysh: I added that to have an e2e test somewhere for the expected transformations. I have changed the name of this test to reflect that.

xla/service/gpu/ir_emission_utils.h
shraiysh: Thanks, I added them there by mistake.

shraiysh:
Ping for review.

xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3280,84 @@ TEST_F(DynamicSliceFusionTest,
...
+  // Check that the dynamic slice thunk only produces a start thunk, and not a
+  // done thunk.
+  DynamicSliceThunk* dynamic_slice_thunk =
+      dynamic_cast<DynamicSliceThunk*>(thunks[2].get());
+  ASSERT_NE(dynamic_slice_thunk, nullptr);
+  const SequentialThunk* embedded_thunk = dynamic_cast<const SequentialThunk*>(
+      dynamic_slice_thunk->embedded_thunk());
+  ASSERT_NE(embedded_thunk, nullptr);
+  ASSERT_EQ(embedded_thunk->thunks().size(), 1);
+  EXPECT_EQ(embedded_thunk->thunks()[0]->kind(),
```
bchetioui, line 3348: nit: this could be a `ThunkKindIs`. Could you make a small follow-up PR making that small fix?

Since there is not much to say here, I will try to land this one to avoid a longer back-and-forth.

bchetioui (approved):
Thanks @shraiysh, I think this looks good now!

golechwierowicz (commented):
In addition `CollectiveOpsTest.AsyncReduceScatter` fails.

```
F0204 02:18:51.811287    8673 statusor.cc:89] Attempting to fetch value instead of handling error INTERNAL: Could not convert JSON string to proto: invalid JSON in xla.gpu.GpuBackendConfig, near   1:12   (offset 11): no such field:   'is_sync'
*** Check failure stack trace: ***
    @     0x7f9846c77e24  absl::log_internal::LogMessage::SendToLog()
    @     0x7f9846c776b8  absl::log_internal::LogMessage::Flush()
```

xla/hlo/parser/hlo_parser_test.cc
```diff
@@ -5924,5 +5924,52 @@ ENTRY main {
...
+                statistics={visualizing_index=1,stat-1=33,stat-2=44}
+    ROOT add-done = s32[] add-done(add-start), origin={{"v3"}}
+  })";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m,
+                          ParseAndReturnVerifiedModule(hlo));
+  // Check the wrapped instruction.
+  HloInstruction* wrapped_instr =
+      m->entry_computation()->root_instruction()->async_wrapped_instruction();
+  EXPECT_EQ(wrapped_instr->metadata().DebugString(),
+            "op_type: \"add\"\nop_name: \"\\\"sample name\\n\"\nsource_file: "
```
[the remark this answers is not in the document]
shraiysh: The updated test should pass the CI. I am matching individual items now, and using debug string to compare against debug string. Let me know if you would like that changed too.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1768,6 +1768,65 @@ ROOT dot = f32[4,8,1024]{2,1,0} dot(p0, p1), lhs_batch_dims={1}, lhs_contracting
...
+    // CHECK:        ROOT {{.+}} = {{.+}} tuple(%[[fusion_done]], %[[wrapped_dot]])
+  )";
+  EXPECT_THAT(
+      RunFileCheck(exec->module().ToString(HloPrintOptions{}
+                                               .set_print_operand_shape(false)
+                                               .set_print_metadata(false)),
+                   kExpected),
+      ::tsl::testing::IsOkAndHolds(true));
+
+  RunAndCompareTwoModulesReplicated(std::move(m), std::move(m_ref),
```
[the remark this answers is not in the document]
shraiysh: Fixed this. Skipping running if number of devices is < 2.

shraiysh:
> In addition `CollectiveOpsTest.AsyncReduceScatter` fails.
> 
> ```
> F0204 02:18:51.811287    8673 statusor.cc:89] Attempting to fetch value instead of handling error INTERNAL: Could not convert JSON string to proto: invalid JSON in xla.gpu.GpuBackendConfig, near   1:12   (offset 11): no such field:   'is_sync'
> *** Check failure stack trace: ***
>     @     0x7f9846c77e24  absl::log_internal::LogMessage::SendToLog()
>     @     0x7f9846c776b8  absl::log_internal::LogMessage::Flush()
> ```

This is an unrelated bug, that was exposed because of the changes here. There are two very similar test cases: `CollectiveOpsTest.AsyncReduceScatter` and `CollectiveOpsTest.AsyncAllGather`. Even without my changes, both of these fail to find the attribute `is_async` from the backend config. The reason why the failure is not observed is because of the difference in thunk constructors ([AllReduce thunk](https://github.com/shraiysh/xla/blob/632793218db14a17d618b90fa750465b70bb3212/xla/backends/gpu/runtime/nccl_all_reduce_thunk.cc#L112) vs [ReduceScatter thunk](https://github.com/shraiysh/xla/blob/632793218db14a17d618b90fa750465b70bb3212/xla/backends/gpu/runtime/nccl_all_reduce_thunk.cc#L145-L147)). The call to `IsSyncCollective` while creating `AllReduceThunk` ensures that there the failure is caught ([here](https://github.com/shraiysh/xla/blob/632793218db14a17d618b90fa750465b70bb3212/xla/service/collective_ops_utils.cc#L803-L805)). This is not the right behaviour. The attribute should be visible for both the testcases. I will handle this accurately in a separate fix, but for now, I have used the same function `IsSyncCollective` while creating the ReduceScatter thunk. This should unblock this PR and the actual fix for the issue will be landed in a separate PR where it fixes other collectives too. Does it sound good @bchetioui @golechwierowicz ?

golechwierowicz (approved):

golechwierowicz (approved):

shraiysh:
Rebased this PR. Can you review (approve) this again @golechwierowicz ?

golechwierowicz (approved):

golechwierowicz:
There are still some conflicts, can you resolve them please?

shraiysh:
> There are still some conflicts, can you resolve them please?

Rebased again.

golechwierowicz:
I tried to submit it internally but it does not build. I saw that `CreateExecutable` creates `OpaqueExecutable` now and you lose access to `module()` accessor method.

shraiysh:
> I tried to submit it internally but it does not build. I saw that `CreateExecutable` creates `OpaqueExecutable` now and you lose access to `module()` accessor method.

Fixed it. Sorry about missing it!

xla/hlo/parser/hlo_parser.cc
```diff
@@ -1472,18 +1472,37 @@ bool HloParserImpl::ParseInstructionRhs(HloComputation::Builder* builder,
...
     instruction->set_original_value(*original_value);
+    if (instruction->IsAsynchronous()) {
+      instruction->async_wrapped_instruction()->set_original_value(
+          *original_value);
+    }
   }
   if (backend_config) {
-    instruction->set_raw_backend_config_string(std::move(*backend_config));
+    instruction->set_raw_backend_config_string(*backend_config);
+    if (instruction->IsAsynchronous()) {
```
golechwierowicz, line 1488: Unfortunately this exposed some bugs related to the internal custom shape verifier internally. Is it possible to get rid of just this line (this if + body) to get it merged?

PR #19847, opened 2024-11-26
Reduce the number of runs for AllReduce test

shraiysh:
The stress test `CollectiveOpsTest.AllReduce_AllCombinations` takes too long to run on a device with 8 GPUs because the number of combinations is 127. Instead of running it for all the GPUs, restricting it to 4 GPUs.


frgossen (approved):
Thanks!

PR #19869, opened 2024-11-27
[XLA] Wraps the FFI backend config opaque string with CustomCallBackendConfig

yliu120:


xla/pjrt/gpu/BUILD
```diff
@@ -153,7 +153,7 @@ xla_cc_test(
     srcs = if_gpu_is_configured(["se_gpu_pjrt_client_test.cc"]),
     tags = [
         "gpu",
-        "no_oss",
+        # "no_oss",
```
ezhulenev, line 156: Is it accidental?

yliu120: Sorry. Accident. I need to test so I comment it out. 
Unfortunately now I am falling into this category...

xla/pjrt/cpu/cpu_client_test.cc
```diff
@@ -446,6 +446,50 @@ TEST(TfrtCpuClientTest, ForwardUserDataToFfiHandler) {
...
+                         kMemsetFromAttr);
+
+TEST(StreamExecutorGpuClientTest, PassAttrToFfiHandler) {
+  static constexpr char const* kProgram = R"(
+    HloModule ffi_handler
+    ENTRY main {
+      ROOT %custom-call = f32[4] custom-call(),
+          custom_call_target="MemsetFromAttr",
+          api_version=API_VERSION_TYPED_FFI,
+          // Takes an encoded version of {"opaque": "{attr = 3.0 : f32}"}
```
ezhulenev, line 472: Why such an ugly encoded string? Because of bytes type? Isn’t it ascii string? I’m worried that it makes HLO unreadable for humans.

yliu120: Yes. That's your suggestion :)
Changing bytes to string will work for most cases. But as you said, if you want to embed a serialized proto in that field, that field becomes unreadable as well. 

The design decision would be just not putting encoded bytes in any proto.

ezhulenev: Ok, but we are all the way in the proto land, maybe we can to

```
oneof {
 bytes opaque = 1;
 string attributes = 2;
}
```

`opaque` for legacy custom calls and `attrs` for FFI

ezhulenev (approved):

xla/pjrt/cpu/cpu_client_test.cc
yliu120: Proto thing resolved offline.

ezhulenev (approved):

PR #19896, opened 2024-11-27
[ROCm] fix fp8 in buffercomparator

ScXfjiang:
NANOO FP8 data type is only supported by MI300 devices (gfx940, gfx941, and gfx942).

For the context on OCP FP8 and NANOO FP8, please refer to this comment: https://github.com/google/flax/pull/3993#issue-2350000228

### DIFF
--- a/xla/service/gpu/buffer_comparator.cu.cc
+++ b/xla/service/gpu/buffer_comparator.cu.cc
@@ -108,6 +108,7 @@ __global__ void xla_fp8_e4m3fnuz_comparison(__hip_fp8_storage_t* buffer_a,
                                             float rel_error_threshold,
                                             uint64_t buffer_length,
                                             int* mismatch_count) {
+#if (defined(__gfx940__) || defined(__gfx941__) || defined(__gfx942__))
   int idx = threadIdx.x + blockIdx.x * blockDim.x;
   if (idx >= buffer_length) return;
   __hip_fp8_e4m3_fnuz elem_a_fp8, elem_b_fp8;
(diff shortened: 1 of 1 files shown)


mooskagh:
I'm not sure I have enough context, but what currently happens for the code outside of those devices?  
Does it fail to compile, or just crashes? If crashes, maybe it makes sense to make it a compile-time check?

Also, I find it weird to replace the function contents with `abort()`, it feels (if we do have to use #ifdefs) putting ifdefs aroung the entire function, and then do corresponding checks at the calling site would be cleaner?

draganmladjenovic:
@mooskagh Calling code is on host and these kernels wont be launched on devices that don't support nanoo fp8 anyways. This is just to ensure that these kernels are compiled for targets that have native support for this specific fp8 format. Prerelase rocm headers would provided emulation for other targets, but this was removed hence the target define check in the change is needed for them to compile successfully.

mooskagh (approved):

majnemer:
FYI, I wrote up https://github.com/openxla/xla/pull/19948 which should fix this by properly handling emulation while also generalizing the scheme and reducing boilerplate.

dimitar-asenov:
As per the last comment, this PR should no longer be needed. I'm closing it. @ScXfjiang if you still think this PR should be merged, please reopen!

ScXfjiang:
> As per the last comment, this PR should no longer be needed. I'm closing it. @ScXfjiang if you still think this PR should be merged, please reopen!

Thanks @dimitar-asenov, please close it.

PR #20006, opened 2024-12-02
[XLA:GPU] Only allow horizontal loop fusion for default memory space

jaro-sevcik:
Horizontal loop fusion currently breaks weight offloading in JAX if it fuses a host-memory copy and device-memory copy because such a fusion results in a same-space buffer, triggering memory space assertions in JAX.

This PR avoids any horizontal loop fusions for host-memory (even though in practice, some fusions would work even in host space).

### DIFF
--- a/xla/service/gpu/transforms/horizontal_loop_fusion.cc
+++ b/xla/service/gpu/transforms/horizontal_loop_fusion.cc
@@ -70,6 +70,16 @@ PrimitiveType GetUniqueOutputTypeOfFusible(const HloInstruction& fusible) {
   return first_output_type;
 }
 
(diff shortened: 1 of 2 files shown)


akuegel (approved):

mooskagh:
It got stuck in the submission, I'll sort it out, sorry for the delay.

mooskagh:
Actually, it seems to have some merge conflicts.
Could you rebase the PR? I'll then merge.

nouiz:
I'm not the owner of this PR and he is OOTO.
If you can do it, it would be great. Otherwise, it will wait.

PR #20086, opened 2024-12-03
[NVIDIA GPU] Fix mem p2p init in collective permute thunk

Tixxx:
Move pointer initialization to the thunk init stage instead of runtime to get rid of the runtime blocking wait.
Add a device sync point using nccl allreduce before doing memcpy to make sure all gpus arrive at the same stage. Otherwise it's possible to have data corruptions when the receiving rank hasn't arrived at the memcpy.

### DIFF
--- a/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
+++ b/xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
@@ -161,11 +161,46 @@ absl::Status NcclCollectivePermuteStartThunk::Initialize(
   if (p2p_memcpy_enabled_) {
     TF_ASSIGN_OR_RETURN(const int64_t current_id,
                         GetCurrentId(params.collective_params, config_));
+    absl::MutexLock lock(&barrier_mutex_);
+    if (barrier_flags_.find(current_id) == barrier_flags_.end()) {
+      if (!params.stream->parent()->HostMemoryRegister(
+              &barrier_flags_[current_id], sizeof(uint8_t))) {
+        LOG(ERROR) << "Registering barrier flag failed.";
+      }
+    }
+
+    TF_ASSIGN_OR_RETURN(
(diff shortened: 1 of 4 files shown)


frgossen (approved):
Thanks for the fix!

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
```diff
@@ -161,13 +161,48 @@ absl::Status NcclCollectivePermuteStartThunk::Initialize(
...
+absl::Status NcclCollectivePermuteStartThunk::Cleanup(
+    const CleanupParams& params) {
+  TF_ASSIGN_OR_RETURN(const int64_t current_id,
+                      GetCurrentId(params.collective_params, config_));
+
+  absl::MutexLock lock(&barrier_mutex_);
+  if (!params.executor->HostMemoryUnregister(&barrier_flags_[current_id])) {
+    LOG(ERROR) << "Unregistering barrier flag failed.";
+  }
+}
```
thomasjoerg, line 204: Several TensorFlow tests are failing with:
`error: non-void function does not return a value in all control paths [-Werror,-Wreturn-type]`

thomasjoerg (changes requested):

xla/service/gpu/runtime/nccl_collective_permute_thunk.cc
Tixxx: added a returned status for the cleanup function

thomasjoerg (approved):

ddunl (approved):

reedwm:
I'm seeing this regress the Maxtext LLama 7B model with 4-way FSDP and 2-way TP when collective matmul is enabled, using the script you gave me @Tixxx.  I see `Tokens/s/device: 4125.938` before this PR and `Tokens/s/device: 3827.136` after it.

PR #20153, opened 2024-12-04
Delete Exchange of Collectives and Dequantization in GEMM Rewriter

philipphack:
Removes collectives from the set of ops that can be exchanged with dequantization in the GEMM rewriter.


reedwm (approved):
To expand on the motivation for any curious onlookers, the exchange of collectives and dequant is removed from GemmRewriter since [CollectiveQuantizer](https://github.com/openxla/xla/blob/main/xla/hlo/transforms/collectives/collective_quantizer.h) now does this

PR #20155, opened 2024-12-04
Add support for emitting a kCall from an async instruction

chaserileyroberts:
This is the last PR needed to enable the explicit stream annotation feature.

This PR enables the IR emitter for kCall subroutines that are wrapped in async-{start,done} pairs. For start, the instructions in the kCall are converted to a `SequentialThunk`. This thunk and the inner thunks are then annotated with their respective `execution_stream_ids`. (These ids come from #19699)

### DIFF
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -1295,18 +1295,19 @@ ENTRY entry {
 TEST_F(GpuCompilerTest, StreamAnnotationThunkTest) {
   const absl::string_view hlo_text = R"(
   HloModule composite
-%f.3 (Arg_0.4: f32[32,32], Arg_1.5: f32[32,32]) -> f32[32,32] {
-  %Arg_1.5 = f32[32,32]{1,0} parameter(1), metadata={op_name="jit(h)/jit(main)/pjit" scheduling_name="Arg_1.5"}
-  %Arg_0.4 = f32[32,32]{1,0} parameter(0), metadata={op_name="jit(h)/jit(main)/pjit" scheduling_name="Arg_0.4"}
-  %custom-call.1 = (f32[32,32]{1,0}, s8[8192]{0}) custom-call(f32[32,32]{1,0} %Arg_0.4, f32[32,32]{1,0} %Arg_1.5), custom_call_target="__cublas$gemm", metadata={op_name="jit(h)/jit(main)/jit(f)/dot_general" source_file="/mounted/scheduling-example/tiny.py" source_line=17 scheduling_name="custom-call.1"}, backend_config={"operation_queue_id":"0","wait_on_operation_queues":[],"gemm_backend_config":{"alpha_real":1,"alpha_imag":0,"beta":0,"dot_dimension_numbers":{"lhs_contracting_dimensions":["1"],"rhs_contracting_dimensions":["0"],"lhs_batch_dimensions":[],"rhs_batch_dimensions":[]},"precision_config":{"operand_precision":["DEFAULT","DEFAULT"],"algorithm":"ALG_UNSET"},"epilogue":"DEFAULT","damax_output":false,"lhs_stride":"1024","rhs_stride":"1024","grad_x":false,"grad_y":false},"force_earliest_schedule":false}
-  ROOT %get-tuple-element = f32[32,32]{1,0} get-tuple-element((f32[32,32]{1,0}, s8[8192]{0}) %custom-call.1), index=0, metadata={op_name="jit(h)/jit(main)/jit(f)/dot_general" source_file="/mounted/scheduling-example/tiny.py" source_line=17 scheduling_name="get-tuple-element"}
+gemm () -> f32[32,32] {
+  a = f32[32,32]{1,0} parameter(0)
+  b = f32[32,32]{1,0} parameter(1)
+  c = (f32[32,32]{1,0}, s8[8192]{0}) custom-call(b, a), custom_call_target="__cublas$gemm",
+    backend_config={"gemm_backend_config":{"dot_dimension_numbers":{"lhs_contracting_dimensions":["1"],"rhs_contracting_dimensions":["0"]}}}
+  ROOT %get-tuple-element = f32[32,32]{1,0} get-tuple-element(c), index=0
 }, execution_thread="explicit"
 
-ENTRY %main.8 (Arg_0.1.0: f32[32,32], Arg_1.2.0: f32[32,32]) -> f32[32,32] {
-  %Arg_1.2.0 = f32[32,32]{1,0} parameter(1), metadata={op_name="b" scheduling_name="Arg_1.2.0"}
-  %Arg_0.1.0 = f32[32,32]{1,0} parameter(0), metadata={op_name="a" scheduling_name="Arg_0.1.0"}
-  %call-start = ((f32[32,32]{1,0}, f32[32,32]{1,0}), f32[32,32]{1,0}) call-start(f32[32,32]{1,0} %Arg_0.1.0, f32[32,32]{1,0} %Arg_1.2.0), async_execution_thread="explicit", to_apply=%f.3, frontend_attributes={_xla_stream_annotation="1"}, metadata={scheduling_name="call-start"}
-  ROOT %call-done = f32[32,32]{1,0} call-done(((f32[32,32]{1,0}, f32[32,32]{1,0}), f32[32,32]{1,0}) %call-start), frontend_attributes={_xla_stream_annotation="1"}, metadata={scheduling_name="call-done"}, backend_config={"operation_queue_id":"0","wait_on_operation_queues":[],"force_earliest_schedule":false}
+ENTRY main () -> f32[32,32] {
+  a = f32[32,32]{1,0} parameter(1)
+  b = f32[32,32]{1,0} parameter(0)
+  %call-start = ((f32[32,32]{1,0}, f32[32,32]{1,0}), f32[32,32]{1,0}) call-start(b, a), to_apply=gemm, frontend_attributes={_xla_stream_annotation="1"}
+  ROOT %call-done = f32[32,32]{1,0} call-done(%call-start), frontend_attributes={_xla_stream_annotation="1"}
 })";
   auto module = ParseAndReturnVerifiedModule(hlo_text).value();
 
@@ -1324,6 +1325,10 @@ ENTRY %main.8 (Arg_0.1.0: f32[32,32], Arg_1.2.0: f32[32,32]) -> f32[32,32] {
   // We expect there to be only 3 thunks:
   // 1) wait 2) a sequential thunk and 3) another wait.
   EXPECT_EQ(gpu_exec->GetThunk().thunks().size(), 3);
+
+  // Check that the waits exist.
+  EXPECT_EQ(gpu_exec->GetThunk().thunks()[0]->kind(), Thunk::kWaitForStreams);
+  EXPECT_EQ(gpu_exec->GetThunk().thunks()[2]->kind(), Thunk::kWaitForStreams);
 
   // The second operation should be a custom call with a specified execution
   // stream id == 1
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -1292,22 +1292,36 @@ ENTRY entry {
   }
 }
 
+MATCHER_P(ThunkKindIs, kind, "") {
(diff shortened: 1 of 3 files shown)


chaserileyroberts:
I'm not sure what is the best way to test this. Perhaps I should just wait for #19462 and #19699 to be merged first?

bchetioui (changes requested):
I think adding an e2e test sounds good to me, but I'd also suggest adding a test that this generates a thunk with the correct stream information.

What do you think?

bchetioui:
I'm not sure there's a dedicated space atm. How about just creating a `ir_emitter_unnested_test.cc` file?

chaserileyroberts:
> How about just creating a ir_emitter_unnested_test.cc file?

Works for me.

bchetioui:
Hi @chaserileyroberts! What's the status of this change---do you still plan to land it?

chaserileyroberts:
Yes, I was waiting for #19462 and #19699 to land first to so I could make the e2e tests, which thankfully today they both just got merged. I'll update this asap.

mjsML:
Can we request the prioritization of this PR to land please and thank you.

bchetioui:
> Can we request the prioritization of this PR to land please and thank you.

If I read correctly, this is not yet updated/ready for review, right?

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1233,6 +1233,52 @@ ENTRY main {
           ->root_instruction()));
 }
 
+TEST_F(GpuCompilerTest, StreamAnnotationThunkTest) {
+  const absl::string_view hlo_text = R"(
+  HloModule composite
+%f.3 (Arg_0.4: f32[32,32], Arg_1.5: f32[32,32]) -> f32[32,32] {
```
bchetioui, line 1239: Could we prettify this test to enhance readability?

Here are a few recommendations that I think we should generally respect:

1. give meaningful names to variables as we would in real code;
2. omit leading `%`s which are not necessary and don't improve readability;
3. omit computation signatures;
4. remove all metadata, and only keep the strict minimum information necessary in backend configs (Is all of what is here necessary for the test to pass using `__cublas$gemm`? If so maybe we can use a custom call with less parameters to enhance readability.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1233,6 +1233,52 @@ ENTRY main {
...
+}, execution_thread="explicit"
+
+ENTRY %main.8 (Arg_0.1.0: f32[32,32], Arg_1.2.0: f32[32,32]) -> f32[32,32] {
+  %Arg_1.2.0 = f32[32,32]{1,0} parameter(1), metadata={op_name="b" scheduling_name="Arg_1.2.0"}
+  %Arg_0.1.0 = f32[32,32]{1,0} parameter(0), metadata={op_name="a" scheduling_name="Arg_0.1.0"}
+  %call-start = ((f32[32,32]{1,0}, f32[32,32]{1,0}), f32[32,32]{1,0}) call-start(f32[32,32]{1,0} %Arg_0.1.0, f32[32,32]{1,0} %Arg_1.2.0), async_execution_thread="explicit", to_apply=%f.3, frontend_attributes={_xla_stream_annotation="1"}, metadata={scheduling_name="call-start"}
+  ROOT %call-done = f32[32,32]{1,0} call-done(((f32[32,32]{1,0}, f32[32,32]{1,0}), f32[32,32]{1,0}) %call-start), frontend_attributes={_xla_stream_annotation="1"}, metadata={scheduling_name="call-done"}, backend_config={"operation_queue_id":"0","wait_on_operation_queues":[],"force_earliest_schedule":false}
+})";
+  auto module = ParseAndReturnVerifiedModule(hlo_text).value();
```
bchetioui, line 1252:
```
TF_ASSERT_OK_AND_ASSIGN(auto module, ...);
```

is the more usual pattern :)

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2647,6 +2648,37 @@ absl::Status IrEmitterUnnested::EmitHloInstruction(
               streams.destination_stream_id, streams.source_stream_id));
           return EmitFusion(Cast<HloFusionInstruction>(wrapped));
         }
+        case HloOpcode::kCall: {
+          const ExecutionStreamAssignment& stream_assignment =
+              ir_emitter_context_->execution_stream_assignment();
+          TF_ASSIGN_OR_RETURN(
+              auto stream, stream_assignment.GetSyncExecutionStreamId(wrapped));
+          auto ir_emitter = IrEmitterUnnested::Create(ir_emitter_context_);
+          DCHECK_EQ(wrapped->called_computations().size(), 1);
```
bchetioui, line 2657: Unless the check is expensive, XLA prefers using `CHECK`s to `DCHECK`s! Please change it!

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2647,6 +2648,37 @@ absl::Status IrEmitterUnnested::EmitHloInstruction(
               streams.destination_stream_id, streams.source_stream_id));
           return EmitFusion(Cast<HloFusionInstruction>(wrapped));
         }
+        case HloOpcode::kCall: {
+          const ExecutionStreamAssignment& stream_assignment =
```
bchetioui, line 2652: This is pretty long, let's use a helper function (`return Emit...`) instead to decrease this function's cognitive complexity!

bchetioui (changes requested):
Thanks! A few comments mostly around readability of the change & tests, but the core seems fine :)

xla/service/gpu/ir_emitter_unnested.cc
jreiffers: Why not TF_RET_CHECK?

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1233,6 +1233,52 @@ ENTRY main {
...
+                       {/*device_allocator=*/nullptr,
+                        /*thread_pool=*/nullptr,
+                        /*layout_canonicalization_callback=*/{},
+                        /*is_autotuning_compilation=*/false})
+          .value();
+  std::unique_ptr<GpuExecutable> gpu_exec(
+      static_cast<GpuExecutable*>(executable.release()));
+  // We expect there to be only 3 thunks:
+  // 1) wait 2) a sequential thunk and 3) another wait.
+  EXPECT_EQ(gpu_exec->GetThunk().thunks().size(), 3);
```
jreiffers, line 1268: Nit: `EXPECT_EQ(gpu_exec->GetThunk().thunks(), testing::SizeIs(3))` is typically preferred.

Alternatively, I think this test could be improved with a matcher, something like this:

```
MATCHER_P(ThunkKindIs, kind, "") {
  return ExplainMatchResult(Eq(kind), arg->kind(), result_listener);
}

EXPECT_THAT(gpu_exec->GetThunk().thunks(), ElementsAre(ThunkTypeIs(...), ThunkTypeIs(...), ThunkTypeIs(...)));
```

IMO the latter is more readable and has the advantage of testing all the thunk types, not just the middle one.

@bchetioui opinions?

xla/service/gpu/gpu_compiler_test.cc
chaserileyroberts: I did my best to reduce the complexity of the HLO. The gemm still has a lot of metadata, but I tried to keep it to the strict necessities for the test to pass.

xla/service/gpu/gpu_compiler_test.cc
chaserileyroberts: Fixed.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1233,6 +1233,52 @@ ENTRY main {
...
+
+  // The second operation should be a custom call with a specified execution
+  // stream id == 1
+  EXPECT_EQ(gpu_exec->GetThunk().thunks()[1]->kind(), Thunk::kSequential);
```
jreiffers, line 1276: This should be an assert, or the static cast down there should be a dynamic cast with a guard. Or both :-).

xla/service/gpu/ir_emitter_unnested.cc
chaserileyroberts: Fixed.

xla/service/gpu/ir_emitter_unnested.cc
chaserileyroberts: I broke out the code into `EmitAsyncComputation`.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1233,6 +1233,52 @@ ENTRY main {
...
+  EXPECT_EQ(gpu_exec->GetThunk().thunks()[1]->execution_stream_id(), 1);
+
+  // Within the sequential thunk, there should only be a single custom call
+  // thunk with the same execution stream id.
+  auto sequential_thunk =
+      static_cast<SequentialThunk*>(gpu_exec->GetThunk().thunks()[1].get());
```
jreiffers, line 1282: We'll want another SizeIs check here.

chaserileyroberts:
@bchetioui I believe I've addressed your comments.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2647,6 +2648,37 @@ absl::Status IrEmitterUnnested::EmitHloInstruction(
...
+              ir_emitter_context_->execution_stream_assignment();
+          TF_ASSIGN_OR_RETURN(
+              auto stream, stream_assignment.GetSyncExecutionStreamId(wrapped));
+          auto ir_emitter = IrEmitterUnnested::Create(ir_emitter_context_);
+          DCHECK_EQ(wrapped->called_computations().size(), 1);
+          auto computation = wrapped->called_computations().front();
+          TF_RETURN_IF_ERROR(ir_emitter->EmitHloComputation(computation));
+          std::unique_ptr<SequentialThunk> thunk_sequence =
+              ir_emitter->ConsumeThunkSequence();
+          for (auto& thunk : thunk_sequence->thunks()) {
```
jreiffers: Can you leave a comment explaining why we need to set this on both the sequential thunk and its contents? Just an API quirk?

xla/service/gpu/gpu_compiler_test.cc
jreiffers: How about something along the lines of this?

```
HloModule composite

async_call {
  p0 = f32[32,32] parameter(0)
  p1 = f32[32,32] parameter(1)
  gemm = (f32[32,32], s8[8192]) custom-call(p0, p1), custom_call_target="__cublas$gemm",
    backend_config={"operation_queue_id":"0","wait_on_operation_queues":[],
      "gemm_backend_config":{"alpha_real":1,"alpha_imag":0,"beta":0,
      "dot_dimension_numbers":
        {"lhs_contracting_dimensions":["1"],"rhs_contracting_dimensions":["0"]},
      "precision_config":{"operand_precision":["DEFAULT","DEFAULT"]},
      "lhs_stride":"1024","rhs_stride":"1024"}}
  ROOT get-tuple-element = f32[32,32] get-tuple-element(gemm), index=0
}, execution_thread="explicit"

ENTRY main {
  p0 = f32[32,32] parameter(0)
  p1 = f32[32,32] parameter(1)
  call-start = ((f32[32,32], f32[32,32]), f32[32,32]) call-start(p0, p1),
    async_execution_thread="explicit", to_apply=async_call, 
    frontend_attributes={_xla_stream_annotation="1"}
  ROOT call-done = f32[32,32]{1,0} call-done(call-start),
    frontend_attributes={_xla_stream_annotation="1"},
    backend_config={"operation_queue_id":"0"}
}
```

xla/service/gpu/gpu_compiler_test.cc
bchetioui: Good catch. I'd also be happiest with the matcher!

chaserileyroberts: Added

xla/service/gpu/gpu_compiler_test.cc
chaserileyroberts: I changed this to use the matcher pattern you suggested instead.

xla/service/gpu/gpu_compiler_test.cc
chaserileyroberts: Added.

jreiffers (approved):
Thanks!

bchetioui (approved):
Thanks @chaserileyroberts for thoughtfully addressing comments, and thanks @jreiffers for the review assist!

PR #20164, opened 2024-12-04
[ROCm] Fix tests breaking due to change in threads_per_warp

hsharsha:
This commit fixes the test break introduced in [5341798](https://github.com/openxla/xla/commit/53417984a9baf80ac9b677f13e628867298f8eae)

### DIFF
--- a/xla/stream_executor/stream_executor_common.h
+++ b/xla/stream_executor/stream_executor_common.h
@@ -51,6 +51,8 @@ class StreamExecutorCommon : public StreamExecutor {
 
   const Platform* GetPlatform() const override { return platform_; }
   const DeviceDescription& GetDeviceDescription() const override;
+  //To be used in tests only
+  void  SetThreadsPerWarp(const int64_t value) override;
   int64_t GetMemoryLimitBytes() const override { return memory_limit_bytes_; }
 
  private:
--- a/xla/service/gpu/fusions/triton/dot_algorithms_test.cc
+++ b/xla/service/gpu/fusions/triton/dot_algorithms_test.cc
@@ -133,6 +133,7 @@ class Triton6xBF16GemmTest : public AlgorithmTest {
 
  protected:
   void SetUp() override {
+    backend().default_stream_executor()->SetThreadsPerWarp(32);
     if (!SupportsBF16(GpuComputeComp())) {
       GTEST_SKIP() << "BF16 not supported.";
     }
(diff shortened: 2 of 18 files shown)


xla/stream_executor/stream_executor_common.h
```diff
@@ -51,6 +51,8 @@ class StreamExecutorCommon : public StreamExecutor {
 
   const Platform* GetPlatform() const override { return platform_; }
   const DeviceDescription& GetDeviceDescription() const override;
+  //To be used in tests only
+  void  SetThreadsPerWarp(const int64_t value) override;
```
klucke, line 55: Please don't add this method in the actual interface.  What is the actual bug that is happening here?  Is the number of threads_per_warp being incorrectly returned in the DeviceDescription for the underlying hardware?  The original PR was a change to use the correct threads_per_warp for the hardware being tested.

hsharsha: But most of the tests are written keeping 32 threads per warp. These tests need to be rewritten and again we may have to distinguish between AMD (64 threads/warp) and NVIDIA (I believe 32 threads/warp) hardware in the test, as they have different values. 

This is a make-shift patch until we fix the tests in a hardware agnostic way.

klucke: majnemer has authored https://github.com/openxla/xla/pull/20170 which fixes this.

hsharsha: https://github.com/openxla/xla/pull/20170 sets threads_per_warp on newer GPUs to 32 but our CI systems run on gfx910 which still get warp size of 64 and hence tests fails.
I am also not sure if we have to hardcode the values instead of relying on hipDevicGetAttributes api which would return correct warp size depending on if wavefrontsize64 is enabled or not.

klucke: I also see a TODO from https://github.com/openxla/xla/pull/10649 saying that threads_per_warp must be 32 in the triton pipeline or things don't work.  Maybe all this should just be dropped, and rocm_executor should just set threads_per_warp to be 32.

hsharsha: I believe that most of the tests are written with 32 threads/warp tiling in mind w.r.t memory allocation and the tensor layout to match that value. We may want to work on extending the tests to support 64 threads/warp as well. But hardcoding in rcom_executor to 32 may not be the right solution in the long run

klucke: I don't think these tests have any threads_per_warp baked into them.  Many of these tests (convolution_test.cc for certain) are backend-agnostic, and run on accelerators without any notion of threads_per_warp at all.

klucke: Since the test break is now fixed, I think this PR should be closed in favor of some future CL which fixes all the hardcoding of 32 threads_per_warp for AMD code.

hsharsha:
Investigation on tiling and indexing with 64 threads per warp is ongoing. Will submit a new PR with correct fix.

PR #20214, opened 2024-12-05
[ds-fusion]Evaluate simple offset values, if possible

shraiysh:
With while loop double buffering, a simple add operation gets added to the entry computation. If a dynamic slice fusion operation exists in this case, then we should be able to evaluate this. If the value is a constant, then no extra cost is incurred because of this change.

Also, fixed the type of Offset in DynamicSliceThunk from uint64_t to int64_t. During execution we deal with int64_t, and so, it only makes sense to store it in int64_t too.


xla/service/gpu/fusions/custom.cc
```diff
@@ -201,25 +199,24 @@ absl::Status CollectSliceInfo(
     const auto* param = Cast<HloParameterInstruction>(idx_op);
     const auto* offset_value = fusion_instr.operand(param->parameter_number());
 
-    if (auto* cst = DynCast<HloConstantInstruction>(offset_value)) {
+    VLOG(0) << "Offset value:" << offset_value->ToString();
```
bchetioui, line 202: This should be `VLOG(2)`, or maybe `VLOG(1)` (we typically use `VLOG(2)` for noisy information, and `VLOG(1)` for high-level pass information. `VLOG(0)` shouldn't be used.).

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3190,6 +3190,61 @@ TEST_F(DynamicSliceFusionTest, ReduceScatterDynamicSlice) {
                                                 false, true, error));
 }
 
+TEST_F(DynamicSliceFusionTest, OffsetComputation) {
```
bchetioui, line 3193: Can you make the test name explain clearly what behaviour is tested?

This seems to be about some `OffsetComputation` being supported. What exactly is that? `OffsetsThatCanBeEvaluatedSuccessfullyAreCorrectlyEmbeddedIntoThunks`?

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3190,6 +3190,61 @@ TEST_F(DynamicSliceFusionTest, ReduceScatterDynamicSlice) {
...
+      ROOT dus = s32[32,32] dynamic-update-slice(dest, rs, c0, add)
+    }
+  )";
+
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module_ref,
+                          ParseAndReturnVerifiedModule(hlo_ref));
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module_opt,
+                          ParseAndReturnVerifiedModule(hlo_opt));
+  ErrorSpec error{1e-3, 1e-3};
+  EXPECT_TRUE(RunAndCompareTwoModulesReplicated(
```
bchetioui, line 3244: How does this test that the computation of the offset was successful and happened when constructing the thunk? It seems like this tells us nothing of the sort---unless I'm missing something.

Perhaps a more precise test for the behaviour we're changing would simply check that the thunk is constructed with a simplified value. WDYT?

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3190,6 +3190,61 @@ TEST_F(DynamicSliceFusionTest, ReduceScatterDynamicSlice) {
...
+      std::move(module_ref), std::move(module_opt), false, true, error));
```
bchetioui, line 3245: Please use `/*arg_name=*/` prefixes when using literal values in function calls---otherwise it's hard to remember what things are supposed to be.

bchetioui (changes requested):

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
shraiysh: Yes, you are right. Fixed it.

bchetioui (approved):
This looks good @shraiysh, thanks for handling my comments!

thomasjoerg (changes requested):

shraiysh:
@thomasjoerg fixed the issue. Thanks for pointing it out.

thomasjoerg (approved):

PR #20216, opened 2024-12-05
[ds-fusion]Add pattern for offset as a function of loop iteration (ds fusion)

shraiysh:
Improving the pattern being recognized in dynamic slice fusion, to allow offset as a function of loop induction variable. This offset will later be calculated on the host at runtime (the host will keep track of the induction variable).


xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
```diff
@@ -152,70 +156,88 @@ bool IsAlignedSlice(const HloInstruction* slice) {
...
+      while_body->parameter_instruction(0), *tuple_idx);
+  if (!indvar) {
+    VLOG(5) << "Unable to find unique GTE for while induction variable idx: "
+            << *tuple_idx << ", while op: " << while_op->ToString();
     return false;
   }
+  const HloInstruction* update =
+      while_body->root_instruction()->operand(*tuple_idx);
+  std::unique_ptr<HloModule> update_module = ExtractModule(update);
+  std::unique_ptr<HloModule> value_module = ExtractModule(&value);
```
bchetioui, line 205: Instead of creating two new modules, we could just walk backwards from `update` and `value` all the way to the `while_body`'s tuple parameter? If we got there by any other path than through the right `GTE`, then we return `false`. Similarly if one of the instructions is a `replica-id` or a `partition-id`.

That'd be cheaper, and not very complex, I think. WDYT?

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
```diff
@@ -152,70 +156,88 @@ bool IsAlignedSlice(const HloInstruction* slice) {
...
+                     user->tuple_index() == *tuple_idx;
+            })) {
+      VLOG(5)
+          << "Not all users of the param are gte operations with tuple idx: "
+          << *tuple_idx << ", module: " << module->ToString();
+      return false;
+    }
 
-  return true;
+    // if any instruction is partition-id or replica-id then this is not valid.
```
bchetioui, line 226: Can you expand on why? My assumption is that this is because it inherently introduces additional, implicit parameters to the computation.

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter_test.cc
```diff
@@ -2143,4 +2143,53 @@ TEST_F(DynamicSliceFusionRewriterTest, ReduceScatterDynamicSlice) {
   RunAndFilecheckHloRewrite(hlo, DynamicSliceFusionRewriter("gpu"), expected);
 }
 
+TEST_F(DynamicSliceFusionRewriterTest, OffsetAsFunctionOfIndvar) {
```
bchetioui, line 2146: Can you specify what is the behaviour that you expect to test here in the test name?

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter_test.cc
```diff
@@ -2143,4 +2143,53 @@ TEST_F(DynamicSliceFusionRewriterTest, ReduceScatterDynamicSlice) {
...
+  const char* hlo = R"(
+    HloModule test, replica_count=2
```
bchetioui, line 2148: Seems like a lot of HLO---do we need all of it to exhibit the behaviour you intend to test?

bchetioui (changes requested):

bchetioui:
Additional request: could you expand your thinking a little more in the commit message? We'd like the commit message to make sense in a vacuum, and without having full context on what the file is doing :)

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
```diff
@@ -152,70 +156,88 @@ bool IsAlignedSlice(const HloInstruction* slice) {
...
+  if (!tuple_idx.has_value()) {
+    VLOG(5) << "Induction variable tuple index is nullopt";
     return false;
   }
-  // Check the update operation
-  const HloInstruction* updated_var =
-      offset.parent()->root_instruction()->operand(param_idx);
-  if (!Match(updated_var, m::Add(m::GetTupleElement(m::Parameter(0), param_idx),
-                                 m::ConstantScalar(1)))) {
+  if (while_body->num_parameters() != 1) {
```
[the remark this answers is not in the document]
shraiysh: Yes, thanks for pointing it out.

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
shraiysh: Yes, thanks for the suggestion. Implemented it that way.

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
shraiysh: Yes, you are correct. The host needs to know this (shouldn't be hard, but trying to support that will be the next step, after this offset work gets done.)

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter_test.cc
shraiysh: Done. Thanks.

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter_test.cc
shraiysh: Unfortunately, yes. I need the while loop, the collective, and a simple (yet random) seq of operations for making sure it is being fused as a general pattern.

shraiysh:
> Additional request: could you expand your thinking a little more in the commit message? We'd like the commit message to make sense in a vacuum, and without having full context on what the file is doing :)

I changed it. Let me know if you'd like me to add something.

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
```diff
@@ -152,70 +156,76 @@ bool IsAlignedSlice(const HloInstruction* slice) {
...
-                                m::Add(m::GetTupleElement(m::Parameter(0)),
-                                       m::ConstantScalar(trip_count)),
-                                m::GetTupleElement(m::Parameter())))) {
+// Returns true if the `consumer` only depends on the `producer` and no other
+// instructions. This is a recursive function checking all paths from the
+// `consumer` to the parameters of the computation and if there is any path
+// without `producer`, then it returns false.
+bool is_only_dependent_on(const HloInstruction* consumer,
+                          HloInstruction* producer,
+                          bool allow_replica_partition_id = false) {
```
bchetioui, line 165: With the new implementation, checking for `replica-id` and `partition-id` specifically seems unnecessary for the current PR.

The condition is essentially:

```cpp
if (consumer.operand_count() == 0) {
  return false;
}
```

How about we just simplify the API like this? We can implement special support later, if it's necessary.

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
```diff
@@ -152,70 +156,76 @@ bool IsAlignedSlice(const HloInstruction* slice) {
...
+  if (consumer == producer) return true;
```
bchetioui, line 166: As specified in a previous PR, we'd like to have braces around all controlled statements (even single-line). Please fix this here and below.

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
```diff
@@ -238,15 +248,17 @@ bool IsHandledConstantForDynamicSliceFusion(const HloInstruction& offset) {
...
 // constant or loop iteration offsets.
-bool HasConstantOrLoopIterationOffsets(
-    const HloDynamicIndexInstruction& instr) {
-  return llvm::all_of(instr.index_operands(), [](const HloInstruction* offset) {
-    return IsLoopIterationNumber(*offset) ||
-           IsHandledConstantForDynamicSliceFusion(*offset);
-  });
+bool HasConstantOrLoopIterationOffsets(const HloDynamicIndexInstruction& instr,
+                                       CallGraph* call_graph) {
+  return llvm::all_of(
```
bchetioui, line 253: Let's use `absl::c_all_of` here too (it is inconsistent to use two different methods to do this in the same file).

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter_test.cc
```diff
@@ -2143,4 +2143,56 @@ TEST_F(DynamicSliceFusionRewriterTest, ReduceScatterDynamicSlice) {
   RunAndFilecheckHloRewrite(hlo, DynamicSliceFusionRewriter("gpu"), expected);
 }
 
+TEST_F(DynamicSliceFusionRewriterTest, OffsetAsFunctionOfIndvarShouldFuse) {
```
bchetioui, line 2146: nit: Indvar -> InductionVariable

bchetioui (changes requested):
Looks much better @shraiysh! Just a few more minor requests.

Also another one: please don't squash when iterating on review comments---otherwise it is hard to see what changed :)

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
shraiysh: Done

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
shraiysh: Sorry about that. Missed it here. I think it is all covered now, but please let me know if you spot anywhere I missed it. (I will check if we can we get clang-format/clang-tidy to check for this?)

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
shraiysh: Done. Thanks.

shraiysh:
Requesting review again, addressed comments.

bchetioui (changes requested):
Thank you @shraiysh, this looks great! I left you one last small comment, please ping me once you've addressed it and I will approve.

(Unfortunately can't approve without resolving everything first, as this will trigger a merge request automatically.)

xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc
```diff
@@ -151,70 +151,72 @@ bool IsAlignedSlice(const HloInstruction* slice) {
...
+// `consumer` to the parameters of the computation and if there is any path
+// without `producer`, then it returns false.
+bool IsOnlyDependentOn(const HloInstruction* consumer,
+                       HloInstruction* producer) {
+  if (consumer == producer ||
+      HloPredicateIsOp<HloOpcode::kConstant>(consumer)) {
+    return true;
+  }
+  if (consumer->operand_count() == 0 ||
+      HloPredicateIsOp<HloOpcode::kParameter>(consumer)) {
```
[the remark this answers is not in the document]
shraiysh: Done.

bchetioui (approved):
Thanks!

PR #20313, opened 2024-12-09
[ds-fusion]Fix async wrapper to walk child computations

shraiysh:
Async wrapper should walk all the computations of instructions, except fusion instructions (especially while and condition instructions). This patch adds that, along with tests.

### DIFF
--- a/xla/service/gpu/transforms/async_wrapper.cc
+++ b/xla/service/gpu/transforms/async_wrapper.cc
@@ -65,9 +65,9 @@ absl::StatusOr<bool> AsyncWrapper::Run(
         continue;
       }
 
-      // Otherwise, follow any `calls` to discover other instructions that can
-      // potentially be made async.
-      if (HloPredicateIsOp<HloOpcode::kCall>(instruction)) {
+      // Otherwise, follow anything other than `fusion`s to discover other
+      // instructions that can potentially be made async.
+      if (!HloPredicateIsOp<HloOpcode::kFusion>(instruction)) {
         std::copy(instruction->called_computations().begin(),
                   instruction->called_computations().end(),
                   std::back_inserter(computations));
--- a/xla/service/gpu/transforms/async_wrapper.cc
+++ b/xla/service/gpu/transforms/async_wrapper.cc
@@ -67,7 +67,8 @@ absl::StatusOr<bool> AsyncWrapper::Run(
 
       // Otherwise, follow anything other than `fusion`s to discover other
       // instructions that can potentially be made async.
-      if (!HloPredicateIsOp<HloOpcode::kFusion>(instruction)) {
+      if (HloPredicateIsOp<HloOpcode::kCall, HloOpcode::kConditional,
+                           HloOpcode::kWhile>(instruction)) {
         std::copy(instruction->called_computations().begin(),
                   instruction->called_computations().end(),
                   std::back_inserter(computations));
--- a/xla/service/gpu/transforms/async_wrapper_test.cc
+++ b/xla/service/gpu/transforms/async_wrapper_test.cc
@@ -82,31 +82,15 @@ TEST_F(AsyncWrapperTest, BasicFusion) {
   EXPECT_TRUE(LiteralTestUtil::Equal(expected, result));
 }
 
-TEST_F(AsyncWrapperTest, BasicFusionWithinWhile) {
-  const char* hlo_text = R"(
+TEST_F(AsyncWrapperTest, OpWithinWhileShouldWrapInAsync) {
+  const char* hlo = R"(
   HloModule m
-
-  double1 {
-    p0 = f32[1] parameter(0)
-    ROOT add = f32[1] add(p0, p0)
-  }
-
-  double2 {
-    p0 = f32[1] parameter(0)
-    ROOT add = f32[1] add(p0, p0)
-  }
-
-  half1 {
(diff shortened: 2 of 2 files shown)


xla/service/gpu/transforms/async_wrapper.cc
```diff
@@ -65,9 +65,9 @@ absl::StatusOr<bool> AsyncWrapper::Run(
         continue;
       }
 
-      // Otherwise, follow any `calls` to discover other instructions that can
-      // potentially be made async.
-      if (HloPredicateIsOp<HloOpcode::kCall>(instruction)) {
+      // Otherwise, follow anything other than `fusion`s to discover other
+      // instructions that can potentially be made async.
+      if (!HloPredicateIsOp<HloOpcode::kFusion>(instruction)) {
```
bchetioui, line 70: Should we be defensive here, and instead enumerate the ops we expect to pass through here?

I'm guessing from your change that this is `kCall` and `kWhile`?

Wherever possible, allow-listing explicitly seems like a better pattern. That way, if we extend the set of HLO instructions later with something that has a computation but that we don't want to traverse, it won't break.

xla/service/gpu/transforms/async_wrapper_test.cc
```diff
@@ -81,5 +82,132 @@ TEST_F(AsyncWrapperTest, BasicFusion) {
   EXPECT_TRUE(LiteralTestUtil::Equal(expected, result));
 }
 
+TEST_F(AsyncWrapperTest, BasicFusionWithinWhile) {
```
bchetioui, line 85: As indicated in other PRs, let's explain what the behavior being tests is in the test name. (Same below.)

xla/service/gpu/transforms/async_wrapper_test.cc
```diff
@@ -81,5 +82,132 @@ TEST_F(AsyncWrapperTest, BasicFusion) {
...
+    tuple = (f32[1], s32[]) tuple(agg3, c0)
+    while = (f32[1], s32[]) while(tuple), body=body, condition=condition
+    ROOT done.1 = f32[1] get-tuple-element(while), index=0
+  })";
+
+  std::unique_ptr<VerifiedHloModule> module =
+      ParseAndReturnVerifiedModule(hlo_text).value();
+
+  AsyncWrapper wrapper(HloPredicateIsOp<HloOpcode::kFusion>);
+  EXPECT_THAT(wrapper.HloModulePass::Run(module.get()), IsOkAndHolds(true));
```
bchetioui, line 137: Just `wrapper.Run` should work? (Same below.)

xla/service/gpu/transforms/async_wrapper_test.cc
```diff
@@ -81,5 +82,132 @@ TEST_F(AsyncWrapperTest, BasicFusion) {
...
+  HloInstruction* while_op = hlo_query::FindInstruction(
+      module->entry_computation(), HloOpcode::kWhile);
+  ASSERT_NE(while_op, nullptr);
+  EXPECT_EQ(CountAsyncInstructions(while_op->while_body()), 4);
+
+  Literal argument = LiteralUtil::CreateR1<float>({1.0});
+  Literal expected = LiteralUtil::CreateR1<float>({128.0});
+
+  Literal result = ExecuteNoHloPasses(std::move(module), {&argument});
+  EXPECT_TRUE(LiteralTestUtil::Equal(expected, result));
```
bchetioui, line 148: Do we need to run this? Seems like we could get away with checking what the resulting HLO looks like, right?

xla/service/gpu/transforms/async_wrapper_test.cc
```diff
@@ -81,5 +82,132 @@ TEST_F(AsyncWrapperTest, BasicFusion) {
...
+}
+
+TEST_F(AsyncWrapperTest, BasicFusionWithinConditional) {
```
bchetioui, line 151: In this test and above, is it important that the instruction within the callable block is a fusion?

If not, I'd pick something more lightweight such that we can make the test code shorter and more to the point (showing that traversals work).

If it's important to also show that the computation of a fusion doesn't get walked, perhaps we can add a third test for that too.

bchetioui (changes requested):

xla/service/gpu/transforms/async_wrapper_test.cc
shraiysh: Done.

xla/service/gpu/transforms/async_wrapper.cc
shraiysh: Done.

xla/service/gpu/transforms/async_wrapper_test.cc
shraiysh: Yes, I changed it to that.

xla/service/gpu/transforms/async_wrapper_test.cc
shraiysh: Yes, I updated the test and removed the execution.

xla/service/gpu/transforms/async_wrapper_test.cc
shraiysh: Added a third test, and replaced fusion with a dummy custom-call.

bchetioui (approved):

PR #20332, opened 2024-12-09
[ds-fusion] Add runtime support for host calculation of offsets in ds fusion

shraiysh:
This patch adds the support for calculating offset on the host at runtime when the offset depends on the loop induction variable. This is done by extracting the offset computation, the induction variable initialization and the induction variable update as independent computations and they are evaluated on the host at runtime. This avoids device-to-host copy for this fusion in these cases.


xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3270,141 @@ TEST_F(DynamicSliceFusionTest,
...
+      dynamic_cast<DynamicSliceThunk*>(ds_thunk->get());
+  std::vector<std::optional<std::vector<DynamicSliceThunk::Offset>>> offsets =
+      ds_thunk_ptr->get_offsets();
+  // Atleast two offsets: one for the input, and one for the outputs.
+  ASSERT_GE(offsets.size(), 2);
+  ASSERT_TRUE(offsets[1].has_value());
+  std::vector<DynamicSliceThunk::Offset> output_offsets = *offsets[1];
+  ASSERT_EQ(output_offsets.size(), 2);
+  // The first value of offset must be an HloModule
+  HloModule** offset1 = std::get_if<HloModule*>(&output_offsets[0]);
```
bchetioui, line 3395: `offset1` seems a bit confusing since you check `output_offsets[1]` below. Maybe assign this to a variable `offset_0` instead? (Underscore added for general consistency.)

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3270,141 @@ TEST_F(DynamicSliceFusionTest,
...
+  ASSERT_NE(offset1, nullptr);
+  ASSERT_NE(*offset1, nullptr);
+  // The second offset must be a constant value
+  ASSERT_EQ(output_offsets[1], DynamicSliceThunk::Offset(0l));
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> unfused_module,
+                          ParseAndReturnVerifiedModule(hlo_unfused));
+  ErrorSpec error{1e-5, 1e-5};
```
bchetioui, line 3402: This is working with integers, why is the error spec not `0`?

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3270,141 @@ TEST_F(DynamicSliceFusionTest,
...
+  RunAndCompareTwoModulesReplicated(
+      std::move(fused_module), std::move(unfused_module),
+      /*run_hlo_passes=*/false, /*use_threads=*/true, error);
```
bchetioui, line 3405: How can we check that the offset computations are actually happening on the host and not on the device?

xla/service/gpu/fusions/custom.cc
```diff
@@ -589,6 +698,18 @@ absl::StatusOr<FusionEmissionResult> EmitCustomCall(
   std::vector<HloInstruction*> slice_instrs(num_args, nullptr);
   std::vector<std::optional<BufferAllocation::Slice>> arguments;
 
+  std::vector<std::unique_ptr<HloModule>> temp_modules;
```
bchetioui, line 701: See other comments regarding this variable name.

xla/service/gpu/fusions/custom.cc
```diff
@@ -847,6 +971,19 @@ absl::StatusOr<FusionEmissionResult> EmitCollective(
   std::vector<HloInstruction*> slice_instrs(2, nullptr);
   std::vector<std::optional<BufferAllocation::Slice>> arguments;
 
+  std::vector<std::unique_ptr<HloModule>> temp_modules;
```
bchetioui, line 974: See other comments regarding this variable name.

xla/service/gpu/fusions/custom.cc
```diff
@@ -408,6 +503,18 @@ absl::StatusOr<FusionEmissionResult> EmitGemm(
 
   std::vector<HloInstruction*> slice_instrs(4, nullptr);
 
+  std::vector<std::unique_ptr<HloModule>> temp_modules;
```
bchetioui, line 506: See other comments regarding this variable name.

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -47,14 +47,28 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+namespace {
+
+std::unique_ptr<Literal>& Indvar(DynamicSliceThunk* thunk) {
+  static thread_local absl::flat_hash_map<DynamicSliceThunk*,
+                                          std::unique_ptr<Literal>>
```
bchetioui, line 54: Why can't we just make `DynamicSliceThunk` hold a pointer to a literal, as opposed to having a `static thread_local` map?

This will outlive the existence of a thunk, and could potentially lead to some kind of UAF if a `DynamicSliceThunk` gets freed and another gets allocated in the same place.

Am I missing something?

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -47,14 +47,28 @@ limitations under the License.
...
 DynamicSliceThunk::DynamicSliceThunk(
     ThunkInfo thunk_info, std::unique_ptr<ThunkSequence> embedded_thunk,
     std::vector<std::optional<BufferAllocation::Slice>> arguments,
     std::vector<std::unique_ptr<BufferAllocation>> fake_allocations,
     std::vector<std::optional<std::vector<Offset>>> offsets,
     std::vector<std::optional<Shape>> orig_shapes,
     std::vector<std::optional<Shape>> sliced_shapes,
-    std::vector<std::optional<uint64_t>> offset_byte_sizes)
+    std::vector<std::optional<uint64_t>> offset_byte_sizes,
+    std::vector<std::unique_ptr<HloModule>> temp_modules,
```
bchetioui, line 69: See other comments regarding this variable name.

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -47,14 +47,28 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+namespace {
+
+std::unique_ptr<Literal>& Indvar(DynamicSliceThunk* thunk) {
```
bchetioui, line 52: This deserves documentation.

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -252,6 +290,12 @@ absl::Status DynamicSliceThunk::ExecuteOnStream(const ExecuteParams& params) {
   // Execute the underlying custom call thunk with the new buffers.
   TF_RETURN_IF_ERROR(embedded_thunk_->ExecuteOnStream(new_params));
 
+  if (indvar_update_ != nullptr) {
+    Indvar(this) = HloEvaluator()
+                       .Evaluate(*indvar_update_, {Indvar(this).get()})
+                       ->CloneToUnique();
```
bchetioui, line 296: Why do we need to clone to unique? Couldn't we just `move`?

xla/service/gpu/runtime/dynamic_slice_thunk.cc
```diff
@@ -47,14 +47,28 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+namespace {
+
+std::unique_ptr<Literal>& Indvar(DynamicSliceThunk* thunk) {
+  static thread_local absl::flat_hash_map<DynamicSliceThunk*,
```
bchetioui, line 53: Am I correct that this design only works if there is a single offset we can calculate for each `DynamicSliceThunk`?

Presumably, we could have several `offset_modules` to evaluate per thunk, right?

If there is an invariant that we always have a single `offset_module` per `DynamicSliceThunk`, then this needs to be enforced somewhere and documented.

If not, then we need a test with several offset modules, and we need to make sure that this works.

xla/service/gpu/fusions/fusions.h
```diff
@@ -54,18 +54,22 @@ class HloFusionInfo : public FusionInfo {
  public:
   HloFusionInfo(const HloFusionAnalysis& analysis,
                 const HloFusionInstruction* instr,
-                const BufferAssignment* buffer_assignment)
+                const BufferAssignment* buffer_assignment,
+                CallGraph& call_graph)
```
bchetioui, line 58: `const`? Seems like you store it as `const` anyway.

bchetioui (changes requested):

xla/hlo/analysis/while_loop_analysis.cc
```diff
@@ -177,7 +177,7 @@ static std::optional<int64_t> GetUniqueGTEDependenceIndex(
       [](const HloInstruction* inst) -> ReplaceType {
         return ReplaceType::kReplaceParam;
       },
-      /*cross_computation=*/false, /*inline_calls_and_fusions=*/false,
+      /*cross_computation=*/false, /*inline_calls_and_fusions=*/true,
```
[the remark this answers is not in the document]
shraiysh: Yes, it is needed. Let me split it in a separate PR with a test so that it can be independently landed. The effect should be minimal though, and it should not cause any issues. The inline pass is run after the extraction, which should be very fast because the extracted module is small.

xla/service/gpu/fusions/custom.cc
```diff
@@ -179,14 +181,102 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+    const HloComputation* parent = op_ptr->parent();
+    std::vector<HloInstruction*> callers =
+        call_graph.GetComputationCallers(parent);
+    if (callers.size() != 1) {
+      return std::nullopt;
+    }
+    if (callers[0]->opcode() == HloOpcode::kWhile) {
+      return callers[0];
+    }
+    op_ptr = callers[0];
```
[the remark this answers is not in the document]
shraiysh: Yes. That is the intent. Until we find the first while loop, we keep going to the parents.

xla/service/gpu/runtime/dynamic_slice_thunk.cc
shraiysh: I tried that, but the same object in the memory is used by multiple threads. I think these threads correspond to the replicas but I am not too sure. Meaning, when the first replica increments the value, all the other replicas also see the new value, and this leads to inconsistency. This is similar to the work here: https://github.com/shraiysh/xla/blob/e27a06fe71dbecc79c2ee636abd4bf23755bd7d6/xla/service/gpu/runtime/while_thunk.cc#L48. Is there a better solution to this problem?

xla/hlo/analysis/while_loop_analysis.cc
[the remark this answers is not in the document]
shraiysh: Moved it to #21375 (merged).

xla/service/gpu/fusions/custom.cc
```diff
@@ -179,14 +181,102 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+  }
+  return std::nullopt;
+}
+
+// Extracts the offset module as a function of induction variable. We do not
+// need rigrous checking here, because while fusing, we have ensured that the
+// offset is a function of the induction variable.
```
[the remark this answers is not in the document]
shraiysh: Added the checks.

xla/service/gpu/fusions/custom.cc
```diff
@@ -179,14 +181,102 @@ absl::StatusOr<BufferAllocation::Slice> GetOperandSlice(
...
+      [](const HloInstruction*) -> ReplaceType {
+        return ReplaceType::kReplaceParam;
+      },
+      /*cross_computation=*/false, /*inline_calls_and_fusions=*/true,
+      /*run_verifier=*/false);
+}
+
+// Extracts the while induction variable update as a function of the induction
+// variable. Here, the update must be a function of the induction variable,
+// because we ensured this while creating this fusion.
```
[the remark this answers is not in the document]
shraiysh: Added checks here too. Thanks for pointing this out.

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
```diff
@@ -3269,6 +3270,141 @@ TEST_F(DynamicSliceFusionTest,
...
+      absl::c_find_if(while_thunk_ptr->body_thunk_sequence()->thunks(),
+                      [](const std::unique_ptr<Thunk>& thunk) {
+                        return thunk->kind() == Thunk::kDynamicSlice;
+                      });
+  ASSERT_NE(ds_thunk, while_thunk_ptr->body_thunk_sequence()->thunks().end());
+  DynamicSliceThunk* ds_thunk_ptr =
+      dynamic_cast<DynamicSliceThunk*>(ds_thunk->get());
+  std::vector<std::optional<std::vector<DynamicSliceThunk::Offset>>> offsets =
+      ds_thunk_ptr->get_offsets();
+  // Atleast two offsets: one for the input, and one for the outputs.
```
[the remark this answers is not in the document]
shraiysh: Changed to assert_eq

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
shraiysh: Thanks for this, yes it should be 0.

xla/service/gpu/runtime/dynamic_slice_thunk.cc
shraiysh: No, this works even if there are multiple offsets. Added test for that. There is only one "induction variable" per thunk. Because the surrounding while loop has to be the same, and so the induction variable has to be the same.

shraiysh:
Thank you for the detailed review @bchetioui. As the changes for `dynamic_slice_thunk.(cc|h)` have been split to #20794, I will be addressing those comments there. I have tried to address the remaining comments. Please let me know if I have missed something. Opening this for review again (changes except the ones in `dynamic_slice_thunk`.)

xla/service/gpu/fusions/dynamic_slice_fusion_test.cc
shraiysh: Answered in a separate comment, where the check exists. (in this test)

xla/service/gpu/runtime/dynamic_slice_thunk.cc
bchetioui: > No, this works even if there are multiple offsets. Added test for that. There is only one "induction variable" per thunk. Because the surrounding while loop has to be the same, and so the induction variable has to be the same.

But nothing guarantees that a loop has a single induction variable, right? We could be updating `offsets` based on different induction variables. Or what is it that would enforce the invariant that the offset is unique?

shraiysh: > But nothing guarantees that a loop has a single induction variable, right?

There are checks for this. The dynamic slice fusion rewriter would not fuse this if there was no unique induction variable.

> what is it that would enforce the invariant that the offset is unique?

I am not sure I understand this, we are making sure that the offset is a function of the unique induction variable while fusing, and asserting this while lowering to avoid any correctness issues. Can you elaborate on this?

bchetioui: > There are checks for this. The dynamic slice fusion rewriter would not fuse this if there was no unique induction variable ([here](https://github.com/shraiysh/xla/blob/15d9a6804c87ca29c3161c4765bf7bcab309406f/xla/service/gpu/transforms/dynamic_slice_fusion_rewriter.cc#L196))

Yeah, the point is that this is not enforced at all in the thunk---so in effect you are assuming here that things behave a certain way in another part of the pipeline, and that this behaviour will never change.

This assumption about how things work needs to be clearly documented in the thunk as well.

> I am not sure I understand this, we are making sure that the offset is a function of the unique induction variable while fusing, and asserting this while lowering to avoid any correctness issues. Can you elaborate on this?

Basically my concern is that we rely on assumed invariants that are not documented nor guaranteed to be upheld by any static mechanism locally. I.e., when we try to refactor/update this code later, it'll be hard to ensure correctness.

shraiysh: > Yeah, the point is that this is not enforced at all in the thunk---so in effect you are assuming here that things behave a certain way in another part of the pipeline, and that this behaviour will never change.

This should be enforced by the code emitting the thunk. I have added checks for that [here](https://github.com/openxla/xla/blob/599fe492864e14b8811632cfa45c4f5bbd818e4e/xla/backends/gpu/codegen/custom.cc#L215). I don't know how we can enforce this within the thunk -- would it make sense for the thunk to accept a while operation, and check this itself, or would simply documenting this be enough? The fact that the thunk accepts only one `indvar_init` and `indvar_update` should force the user of this thunk to ensure that there is one induction variable, wouldn't it?

shraiysh:
Ping for review!

shraiysh:
Ping for review!

bchetioui (approved):

bchetioui (approved):

PR #20374, opened 2024-12-10
[XLA:GPU] Re-run the host-offload-legalize pass after CSE

### LINKED ISSUE #20373: Activation offloading with scan fails in XLA with same shape pinned-host/device scan outputs
If there is a while loop with one output on pinned host and other one on device with the same shape, a CSE pass will combine their buffers and this will break host offloader (on GPU).

The CSE pass was newly introduced as a [side effect](https://github.com/openxla/xla/blob/d5f50486b7b56c6076dcbaa6a06006165474ff95/xla/service/gpu/gpu_compiler.cc#L1605) of turning `--xla_gpu_experimental_enable_triton_softmax_priority_fusion` on by default (see https://github.com/openxla/xla/commit/eab45d5da2fab59de8c02678c2b7b9ae69d9fef8).

JAX repro (GPU):
```python
import jax
import jax.numpy as jnp

p = jax.sharding.SingleDeviceSharding(jax.devices()[0], memory_kind="pinned_host")

@jax.jit
def f():
  def g(_1, _2):
    return None, (jax.device_put(jnp.array(1.0), p), jnp.array(2.0))
  return jax.lax.scan(g, None, length = 4)[1]

print(f()[0].sharding)  # doesn't crash
```

Error message:
```
  File "repro.py", line 14, in <module>
    print(f()[0].sharding)  # doesn't crash
jaxlib.xla_extension.XlaRuntimeError: INTERNAL: during context [end-of-post-layout_assignment]: Expected instruction to have shape equal to (s32[], f32[4]{0:S(5)}, f32[4]{0:S(5)}), actual shape is (s32[], f32[4]{0:S(5)}, f32[4]{0}):
%tuple.4 = (s32[], f32[4]{0:S(5)}, f32[4]{0}) tuple(s32[] %constant.1, f32[4]{0:S(5)} %custom-call.3, f32[4]{0:S(5)} %custom-call.3), metadata={op_name="jit(f)/jit(main)/while" source_file="repro.py" source_line=12}
```

jaro-sevcik:
The CSE pass breaks an important invariant of host-offloader for scan -- each "host
buffer" must have an allocation sequence consisting of a zero constant
and a unique single-user broadcast, but CSE merges two broadcasts
if they have the same shape. This invariant is enforced by host-offload-legalize
pass, but a new CSE pass was recently introduced between
the host-offload-legalize and host-offloader passes (as a side effect of https://github.com/openxla/xla/commit/eab45d5da2fab59de8c02678c2b7b9ae69d9fef8).

This patch re-runs the host-offload-legalize after the offending CSE pass.

Fixes https://github.com/openxla/xla/issues/20373

### DIFF
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -1325,6 +1325,40 @@ class PassOrderTest : public GpuCompilerTest {
     CompileModule(config);
   }
 
+  // Fails if any of the passes matching `other_pass_regex` runs before
+  // the first occurence of the pass matching `first_pass_regex`.
+  void VerifyPassRunsBefore(absl::string_view first_pass_regex,
+                            absl::string_view other_pass_regex) {
+    if (!optimized_module_) {
+      CompileModule(GetModuleConfigForTest());
+    }
+    int first_pass_first_run = std::numeric_limits<int>::max();
+    int other_pass_first_run = std::numeric_limits<int>::max();
+    int run_index = 0;
+    for (const HloPassMetadata& pass_metadata :
+         optimized_module_->metadata()->proto().pass_metadata()) {
+      if (RE2::FullMatch(pass_metadata.pass_name(), first_pass_regex)) {
+        VLOG(2) << "Pass " << pass_metadata.pass_name()
+                << " matches first_pass_regex." << std::endl;
+        first_pass_first_run = std::min(first_pass_first_run, run_index);
+      }
+      if (RE2::FullMatch(pass_metadata.pass_name(), other_pass_regex)) {
+        VLOG(2) << "Pass " << pass_metadata.pass_name()
+                << " matches other_pass_regex." << std::endl;
+        other_pass_first_run = std::min(other_pass_first_run, run_index);
+      }
+      ++run_index;
+    }
+
+    EXPECT_NE(first_pass_first_run, std::numeric_limits<int>::max())
+        << "Did not run a pass matching " << first_pass_regex;
+    EXPECT_NE(other_pass_first_run, std::numeric_limits<int>::max())
+        << "Did not run a pass matching " << other_pass_regex;
+    EXPECT_LE(first_pass_first_run, other_pass_first_run)
+        << "A pass matching " << first_pass_regex
+        << " did not run before passes matching " << other_pass_regex;
+  }
+
   // Fails if any of the passes with names matching the regular expression
   // `first_pass_regex` run after any of the passes matching `last_pass_regex`
   // or if none of the executed passes matches `first_pass_regex` or
@@ -1405,6 +1439,15 @@ TEST_F(PassOrderTest, PassesAreRunInCorrectOrder) {
                   /*last_pass_regex=*/"priority-fusion");
   VerifyPassOrder(/*first_pass_regex=*/"layout-assignment",
                   /*last_pass_regex=*/"layout_normalization");
+}
+
+TEST_F(PassOrderTest, OffloadingPassesAreRunInCorrectOrder) {
+  VerifyPassRunsBefore(/*first_pass_regex=*/"host-offload-legalize",
+                       /*other_pass_regex=*/"layout_normalization");
+  auto pass_range =
+      VerifyPassOrder(/*first_pass_regex=*/"host-offload-legalize",
+                      /*last_pass_regex=*/"host-offloader");
+  VerifyNotRunInBetween(pass_range, /*pass_regex=*/"cse");
 }
 
 TEST_F(PassOrderTest, FusionBlockLevelRewriterRunsAfterAllFusionPasses) {
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -1405,8 +1405,6 @@ TEST_F(PassOrderTest, PassesAreRunInCorrectOrder) {
                   /*last_pass_regex=*/"priority-fusion");
   VerifyPassOrder(/*first_pass_regex=*/"layout-assignment",
                   /*last_pass_regex=*/"layout_normalization");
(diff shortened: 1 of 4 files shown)


xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1405,8 +1405,6 @@ TEST_F(PassOrderTest, PassesAreRunInCorrectOrder) {
                   /*last_pass_regex=*/"priority-fusion");
   VerifyPassOrder(/*first_pass_regex=*/"layout-assignment",
                   /*last_pass_regex=*/"layout_normalization");
-  VerifyPassOrder(/*first_pass_regex=*/"host-offload-legalize",
```
jaro-sevcik, line 1408: Unfotunately, the pass order test cannot quite express that we want to run host-offload-legalize just before each layout normalization, so I had to relax the test.

bchetioui: Why not? You could write a custom test that checks that each time `layout-normalization` is run, it is preceded by `host-offload-legalize`?

bchetioui (changes requested):

xla/service/gpu/gpu_compiler_test.cc
jaro-sevcik: Done.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1325,6 +1325,40 @@ class PassOrderTest : public GpuCompilerTest {
     CompileModule(config);
   }
 
+  // Fails if any of the passes matching `other_pass_regex` runs before
+  // the first occurence of the pass matching `first_pass_regex`.
```
bchetioui, line 1329: occurrence

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1325,6 +1325,40 @@ class PassOrderTest : public GpuCompilerTest {
...
+  void VerifyPassRunsBefore(absl::string_view first_pass_regex,
```
bchetioui, line 1330: `VerifyPassRunsAtLeastOnceBefore`?

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1405,8 +1439,15 @@ TEST_F(PassOrderTest, PassesAreRunInCorrectOrder) {
                   /*last_pass_regex=*/"priority-fusion");
   VerifyPassOrder(/*first_pass_regex=*/"layout-assignment",
                   /*last_pass_regex=*/"layout_normalization");
-  VerifyPassOrder(/*first_pass_regex=*/"host-offload-legalize",
-                  /*last_pass_regex=*/"layout_normalization");
+}
+
+TEST_F(PassOrderTest, OffloadingPassesAreRunInCorrectOrder) {
```
bchetioui, line 1444: Can you explain what the correct order is, at least in a comment?

Also, please reference the bug number that you're trying to work around in the test---since the check that we're not running CSE is testing for a big hack.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1405,8 +1439,15 @@ TEST_F(PassOrderTest, PassesAreRunInCorrectOrder) {
...
+  VerifyPassRunsBefore(/*first_pass_regex=*/"host-offload-legalize",
+                       /*other_pass_regex=*/"layout_normalization");
+  auto pass_range =
+      VerifyPassOrder(/*first_pass_regex=*/"host-offload-legalize",
```
bchetioui, line 1448: Can you comment on what the pass range being extract here is?

Also perhaps factor out a util `GetOrderedPassesRange` with a docstring explaining what it does? Calling `VerifyPassOrder` just to extract a `PassRange` seems odd.

bchetioui (changes requested):

xla/service/gpu/gpu_compiler_test.cc
jaro-sevcik: This is how it is used in other places in the file.

Also, the legalizer is supposed to prepare for running the offloader, so this constraint should make sense.

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1405,8 +1439,24 @@ TEST_F(PassOrderTest, PassesAreRunInCorrectOrder) {
...
+
+TEST_F(PassOrderTest, OffloadingPassesAreRunInCorrectOrder) {
+  // HostOffloadLegalize must run before LayoutNormalization to prevent
+  // the creation of invalid transpose/bitcast operations within
+  // host memory offloading segments.
+  VerifyPassRunsAtLeastOnceBefore(/*first_pass_regex=*/"host-offload-legalize",
+                                  /*other_pass_regex=*/"layout_normalization");
+
+  // CSE should not run between HostOffloadLegalize and HostOffloader
+  // because it could break the invariants esyablished
```
bchetioui: established

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1405,8 +1439,24 @@ TEST_F(PassOrderTest, PassesAreRunInCorrectOrder) {
...
+  // HostOffloadLegalize must run before LayoutNormalization to prevent
+  // the creation of invalid transpose/bitcast operations within
+  // host memory offloading segments.
+  VerifyPassRunsAtLeastOnceBefore(/*first_pass_regex=*/"host-offload-legalize",
+                                  /*other_pass_regex=*/"layout_normalization");
+
+  // CSE should not run between HostOffloadLegalize and HostOffloader
+  // because it could break the invariants established
+  // by the legalization, such as the buffer initialization broadcasts
+  // before loops having only single use
```
bchetioui, line 1454: only [a]

bchetioui (changes requested):

bchetioui (approved):
Thank you!

PR #20399, opened 2024-12-10
Elementwise Ops in Collective Pipeliner

philipphack:
Enables support for elementwise ops ahead of the dynamic-update-slice, such as in scaled FP8 GEMMs, in the collective pipeliner.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -880,7 +880,7 @@ absl::Status RunCollectiveOptimizationPasses(
         /*level_to_operate_on=*/0,
         /*max_pipelining_per_loop=*/INT64_MAX,
         /*last_run=*/true,
-        /*pipeline_use_tree=*/false,
+        /*pipeline_use_tree=*/true,
         /*process_different_sized_ops=*/true,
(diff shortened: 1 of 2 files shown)


frgossen (changes requested):
The change looks good but I'm not sure what the motivation is. Can you provide some benchmark numbers to show why this is beneficial? Is this useful for maxtext models?

philipphack:
@frgossen we found this in Paxml, but I think it could also be relevant elsewhere. Without this change, the scaling (i.e. multiplication) of the result of an FP8 dot prevents the pipelining of the collective. The graph in this case is dot -> all-reduce -> slice -> multiply -> dynamic-update-slice.

frgossen:
@philipphack , sorry for the delay. Do we have a concrete model to proof the performance gain here? Can you share the concrete PAXML config and the performance gains you observe?

hx89:
@frgossen I have tested on PAXML Grok proxy FP8 model and verified the RS are pipelined correctly after the fix, and the e2e perf gain is about 14% over the run before the fix. So I think this PR is good to merge.

frgossen (approved):
Thanks!

PR #20426, opened 2024-12-11
Layout assignment: Reset memory space in result layout

jaro-sevcik:
Layout assignment should not set any memory space on any of the instructions even if the entry computation layout has non-default memory space. At one place, the memory space was leaking (causing weight offloading crashes on real models), this patch addresses that.

Drive-by: Introduce a helper function for the copy-pasted implementations of resetting the memory space in a layout.

### DIFF
--- a/xla/service/layout_assignment.cc
+++ b/xla/service/layout_assignment.cc
@@ -657,6 +657,20 @@ absl::Status PropagateParameterLayoutToUsers(const HloInstruction* instruction,
   return absl::OkStatus();
 }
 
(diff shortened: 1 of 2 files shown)


SandSnip3r:
This feels a bit heavy-handed. 

> Layout assignment should not set any memory space

Is that true? Memory spaces are a part of layouts, so it seems like it's well within the rights of LayoutAssignment to set memory spaces.

What happens before this fix? Host memory space is also being set on a parameter?

SandSnip3r (approved):
Ok nevermind, it looks like we're doing this already but missed a few spots.

ddunl (approved):

PR #20428, opened 2024-12-11
[XLA:FFI] Fix C API

Corendos:
Some of the definition in `xla/ffi/api/c_api.h` were not valid C. I fixed them so that it's possible to import the header in a C project.


ezhulenev (approved):

steeve:
Perhaps best to do anonymous typedefs ?

Also, some structs are still not typedefed, such as `XLA_FFI_Attrs`

Wdyt @ezhulenev ?

ezhulenev:
Yes, let's fix all of them. I don't have a preference for `anonymous typedefs` or not, I think `anonymous` is fine

Corendos:
> Also, some structs are still not typedefed, such as `XLA_FFI_Attrs`

Doesn't need to, it's handled by `XLA_FFI_DEFINE_STRUCT_TRAITS` (it threw me off at the beginning too)

As for the anonymous struct `XLA_FFI_Extension_Base` is already valid and not anonymous so I did the same for consistency

ezhulenev (approved):

PR #20429, opened 2024-12-11
[PJRT] Expose `ExecutionContext` when executing a `LoadedExecutable`

Corendos:
Currently, there is no way to provide an `ExecutionContext` when launching an executable. This prevents using it in custom calls through FFI.

I exposed the field in the `PJRT_ExecuteOptions` struct and wired it in `xla::ExecuteOptions`


ezhulenev (approved):

xla/pjrt/c/pjrt_c_api.h
```diff
@@ -1364,6 +1364,7 @@ struct PJRT_ExecuteOptions {
   // during the call.
   const int64_t* non_donatable_input_indices;
   size_t num_non_donatable_input_indices;
+  PJRT_ExecuteContext* context;
 };
 PJRT_DEFINE_STRUCT_TRAITS(PJRT_ExecuteOptions, num_non_donatable_input_indices);
```
steeve, line 1369:
```suggestion
PJRT_DEFINE_STRUCT_TRAITS(PJRT_ExecuteOptions, context);
```

Need to update this too

steeve (changes requested):

ezhulenev (approved):

xla/pjrt/c/pjrt_c_api.h
```diff
@@ -79,7 +79,7 @@ PJRT_DEFINE_STRUCT_TRAITS(PJRT_Extension_Base, next);
 // Changes include:
 // * Adding a new field to the PJRT_Api or argument structs
 // * Renaming a method or argument (doesn't affect ABI)
-#define PJRT_API_MINOR 57
+#define PJRT_API_MINOR 58
```
thomasjoerg, line 82: Please add to the `xla/pjrt/c/CHANGELOG.md`

thomasjoerg (changes requested):

xla/pjrt/c/pjrt_c_api.h
Corendos: Addressed in https://github.com/openxla/xla/pull/20429/commits/5017a0ed720492ecf8d47cafd8d2b09d7222cbf8

thomasjoerg (approved):

penpornk (changes requested):
Could you please help add a unit test for this as well?

ezhulenev:
I'm not sure anything useful can be added as a test, but PR needs a rebase because I bumped pjrt version yesteday

Corendos:
> but PR needs a rebase because I bumped pjrt version yesteday

Addressed in [5b97f17](https://github.com/openxla/xla/pull/20429/commits/5b97f17549fc5ec0671d19f3fe86a598973f5b61)

penpornk (approved):
Thank you!

Corendos:
> //xla/pjrt:pjrt_c_api_client_test failed. Could you please help take a look? Here is the [XLA Linux CPU CI log](https://btx.cloud.google.com/invocations/7a84d36d-fd7d-4c88-8525-a700accf14a3/targets/tensorflow%2Fxla%2Flinux%2Fcpu%2Fbuild_cpu/log).

Indeed, thanks ! Addressed in [7128d16](https://github.com/openxla/xla/pull/20429/commits/7128d161701c2bf872b99d94b29ecc43a3d779d3)

penpornk (approved):

Corendos:
@penpornk It seems something is still wrong on the CI (if I read correctly), but I tested locally on a Linux x64 box and I couldn't reproduce the failing test (it was failing before my fix). Is there something I can do ?

penpornk (approved):
The internal tests just weren't rerun. I'll try re-approving to re-run the tests.

Corendos:
Awesome ! Thanks !

Corendos:
Ok, so I managed to dig a bit into this problem and it turned out to be quite a rabbit hole.

It seems that the issue is due to fields of `ExecuteOptions` not being correctly default initialized in the tests (`xla/pjrt/pjrt_c_api_client_test.cc`). Especially the `context` field now that we make use of it.

Now I think I uncover a small flaw in the C <-> C++ interaction here.

In `ExecuteOptions`, the `context` field is a non-owning pointer to an `ExecuteContext`.
`PJRT_ExecuteContext` is a wrapper around an owning `ExecuteContext` (through a `shared_ptr`).
In the tests, we need to convert the `ExecuteOptions` to a `PJRT_ExecuteOptions` and I think it's flawed because we can't convert the non-owning pointer to a `shared_ptr`, as it will become owned.

@ezhulenev Since you're the one that implemented the initial support, I'd like your opinion on this.
Also I see two potential solutions:
1) Switch to a non-owning `ExecuteContext` in `PJRT_ExecuteContext`
2) Switch to an owning `shared_ptr<ExecuteContext>` in `ExecuteOptions`

I personally favors the first option as it seems more consistent, but happy to discuss

Corendos:
Just a heads up, I implemented the version where I remove the `std::shared_ptr` [here](https://github.com/Corendos/xla/tree/corendos/pjrt-execution-context-2)

The tests that were failing are passing locally, but I'm not sure how to run the full test suite to see if I didn't break something else. Is there a bazel target that I can use ?

ezhulenev (approved):

ezhulenev:
PJRT_ExecuteContext is basically an RAII wrapper around ExecuteContext, and it's a user responsibility to destroy it, and then it automatically destroys wrapped C++ context. Unwrapping it and forwarding to XLA/PJRT seems like a right thing.

Corendos:
> PJRT_ExecuteContext is basically an RAII wrapper around ExecuteContext, and it's a user responsibility to destroy it, and then it automatically destroys wrapped C++ context. Unwrapping it and forwarding to XLA/PJRT seems like a right thing.

Agreed, but the thing is that, in the tests, we don't rely on a user-created `PJRT_ExecuteContext`. Instead we already have an `ExecuteContext` that we need to wrap inside a `PJRT_ExecuteContext`.

In https://github.com/openxla/xla/blob/b17a35bf7654b40ba656dd347326b95156d0656e/xla/pjrt/pjrt_c_api_client_test.cc#L96, we create a `ExecuteContext` that we feed to `execute`.

It then bubbles up in https://github.com/openxla/xla/blob/33e742fdb0344edf121a7ffecd3d4b6a2cfc37a6/xla/pjrt/pjrt_c_api_client.cc#L1711-L1806 where we need to create a `PJRT_ExecuteContext`. But we can't create a `std::shared_ptr<ExecuteContext>` from a raw `ExecuteContext` pointer. We also probably can't copy-construct the `ExecuteContext` because there is a `ffi::ExecutionContext` in the `ExecuteContext` and this one has a `flat_map` of `std::unique_ptr`.

If I want to implement this PR properly and expose the `ExecutionContext` to C++ users of XLA/PJRT, I need a solution to fix that 😁

ezhulenev:
Let’s make it an std::variant maybe? And have an internal api that constructs it from non owning pointer? I’ll see if I can find anything similar in our code base, FWIW in FFI we have these “internal APIs” to have direct access to internals

ezhulenev:
I think you can just go ahead with variant<ptr, shared_ptr>, it will be an implementation details that will not leak anywhere

Corendos:
Awesome, I'll do that and ping you afterwards 👍

Also, what's the bazel target I should use to run all the tests ?

ezhulenev:
I think it's bazel test //folder/..., just run all tests under pjrt, it should be enough

Corendos:
Ok, so it was a bit harder than I thought due to struct visibility.

I had to make `pjrt_c_api_client` depends on `//xla/pjrt/c:pjrt_c_api_wrapper_impl`. I'm not sure how it's supposed to be solved properly.

Maybe we should do like the other structures and have a `xla::PjRtExecuteContext` ?

Corendos:
Ok, I've been thinking about this and I think it all boils down to the fact that, contrary to the other `PJRT_*` struct, `PJRT_ExecuteOptions` is not manipulated using a pointer. So we can't just predeclare it and must have access to the struct definition.

One improvement that could be done to avoid being dependent on the `_impl.h` would be to extract the struct definition elsewhere. Apart from that I don't really know what else could be done. If you have idea @ezhulenev happy to discuss 😁

ezhulenev:
We certainly can't include impl.h headers anywhere :( I started looking at it, and got a bit lost in all our C APIs and structs, will try to come up with something today

ezhulenev:
I think I have a plan, will send PR later today

ezhulenev:
After a few attempts I think this is the one that will work: https://github.com/openxla/xla/pull/21317

Corendos:
Awesome, I'm looking forward to it being merged !

I guess I'll close this PR then

ezhulenev:
Submitted in https://github.com/openxla/xla/commit/7ce927ad69de159ccddd5af61385460ed1552219

Corendos:
Thanks a lot @ezhulenev !

PR #20503, opened 2024-12-13
fixed the typos in bfloat16_propagation_test.cc

kiransair:
updated typo in documents

### DIFF
--- a/xla/hlo/parser/hlo_parser_test.cc
+++ b/xla/hlo/parser/hlo_parser_test.cc
@@ -2892,7 +2892,7 @@ TEST_F(HloParserTest, Empty) {
 
 TEST_F(HloParserTest, Garbage) {
   const std::string original =
-      "HloModule thi$ str1ng makes# N0 sen$e @all!*&^%$";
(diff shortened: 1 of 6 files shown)


xla/hlo/parser/hlo_parser_test.cc
```diff
@@ -2892,7 +2892,7 @@ TEST_F(HloParserTest, Empty) {
 
 TEST_F(HloParserTest, Garbage) {
   const std::string original =
-      "HloModule thi$ str1ng makes# N0 sen$e @all!*&^%$";
+      "HloModule this str1ng makes# N0 sen$e @all!*&^%$";
```
penpornk, line 2895: This typo is intentional. Could you please change back?

penpornk (changes requested):

xla/hlo/parser/hlo_parser_test.cc
kiransair: changed as suggested

penpornk (approved):
Thank you for the change!

PR #20557, opened 2024-12-15
[ds-fusion] Add HandleReducePrecision to algebraic simplifier

shraiysh:
When the mantissa and exponent of the reduce-precision instruction are the same as the mantissa and exponent of the primitive type of the operand, then the reduce-precision operation is a no-op.

### DIFF
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
@@ -12698,7 +12698,7 @@ TEST_F(AlgebraicSimplifierTest,
   })";
   TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m,
                           ParseAndReturnVerifiedModule(hlo));
-
+  default_options_.set_enable_remove_no_op_reduce_precision(true);
   EXPECT_TRUE(AlgebraicSimplifier(default_options_).Run(m.get()).value());
   EXPECT_THAT(m->entry_computation()->root_instruction(),
               GmockMatch(m::Parameter()));
@@ -12715,6 +12715,7 @@ TEST_F(AlgebraicSimplifierTest,
   TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m,
                           ParseAndReturnVerifiedModule(hlo));
 
+  default_options_.set_enable_remove_no_op_reduce_precision(true);
   EXPECT_FALSE(AlgebraicSimplifier(default_options_).Run(m.get()).value());
 }
 
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
@@ -12688,8 +12688,9 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
   EXPECT_FALSE(simplifier.Run(module.get()).value());
 }
 
-TEST_F(AlgebraicSimplifierTest,
-       ReducePrecisionWithSamePrecisionAsOperandShouldBeRemoved) {
+TEST_F(
+    AlgebraicSimplifierTest,
+    ReducePrecisionWithSamePrecisionAsOperandShouldBeRemovedWhenRemoveNoOpReducePrecisionIsSet) {
   const char* hlo = R"(
   HloModule test
   ENTRY main {
@@ -12704,8 +12705,9 @@ TEST_F(AlgebraicSimplifierTest,
               GmockMatch(m::Parameter()));
 }
 
-TEST_F(AlgebraicSimplifierTest,
-       ReducePrecisionWithDifferentPrecisionFromOperandShouldNotBeModified) {
+TEST_F(
+    AlgebraicSimplifierTest,
+    ReducePrecisionWithDifferentPrecisionFromOperandShouldNotBeModifiedByDefault) {
   const char* hlo = R"(
   HloModule test
   ENTRY main {
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
@@ -12688,5 +12688,31 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
   EXPECT_FALSE(simplifier.Run(module.get()).value());
 }
 
+TEST_F(AlgebraicSimplifierTest,
+       ReducePrecisionWithSameDatatypeShouldBeRemoved) {
+  const char* hlo = R"(
+  HloModule test
+  ENTRY main {
+    param.0 = bf16[2,16,64]{2,1,0} parameter(0)
+    param.1 = bf16[32,64]{1,0} parameter(1)
+    reduce-precision.46 = bf16[32,64]{1,0} reduce-precision(param.1), exponent_bits=8, mantissa_bits=7
+    bitcast.7734 = bf16[2,16,64]{2,1,0} bitcast(reduce-precision.46)
+    multiply.1006 = bf16[2,16,64]{2,1,0} multiply(param.0, bitcast.7734)
+  })";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnVerifiedModule(hlo));
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      bool changed,
+      RunHloPass(AlgebraicSimplifier(default_options_), module.get()));
+  EXPECT_TRUE(changed);
+  EXPECT_EQ(module->entry_computation()
+                ->root_instruction()
+                ->operand(1)
+                ->operand(0)
+                ->opcode(),
+            HloOpcode::kParameter);
+}
+
 }  // namespace
 }  // namespace xla
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
@@ -7258,13 +7258,14 @@ TEST_F(AlgebraicSimplifierTest, SliceOfReduceWindowTwoReduceDims) {
   auto root = m->entry_computation()->root_instruction();
   EXPECT_THAT(
       root,
-      GmockMatch(m::Reshape(m::Reduce(m::Parameter(0), m::Constant())
-                                .WithShape(S32, {3})
-                                .WithPredicate([](const HloInstruction* instr) {
-                                  return instr->dimensions() ==
-                                         std::vector<int64_t>({1, 2});
-                                }))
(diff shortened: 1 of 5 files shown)


xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12688,5 +12688,31 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
   EXPECT_FALSE(simplifier.Run(module.get()).value());
 }
 
+TEST_F(AlgebraicSimplifierTest,
+       ReducePrecisionWithSameDatatypeShouldBeRemoved) {
```
bchetioui, line 12692: `ReducePrecisionWithSamePrecisionAsDataTypeShouldBeRemoved` (the name is not precise enough, reduce precision is always on the "same" data type for i/o :)

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12688,5 +12688,31 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
...
+    multiply.1006 = bf16[2,16,64]{2,1,0} multiply(param.0, bitcast.7734)
+  })";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnVerifiedModule(hlo));
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      bool changed,
+      RunHloPass(AlgebraicSimplifier(default_options_), module.get()));
+  EXPECT_TRUE(changed);
+  EXPECT_EQ(module->entry_computation()
```
bchetioui, line 12709: Maybe use gmock matchers to simplify the structure here? This is hard to read.

See the above `AlgebraicSimplifierTest.PathologicalComplexity` for an example of how to do that.

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12688,5 +12688,31 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
...
+    reduce-precision.46 = bf16[32,64]{1,0} reduce-precision(param.1), exponent_bits=8, mantissa_bits=7
+    bitcast.7734 = bf16[2,16,64]{2,1,0} bitcast(reduce-precision.46)
+    multiply.1006 = bf16[2,16,64]{2,1,0} multiply(param.0, bitcast.7734)
+  })";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnVerifiedModule(hlo));
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      bool changed,
+      RunHloPass(AlgebraicSimplifier(default_options_), module.get()));
```
bchetioui, line 12707: nit: this seems inconsistent with how the rest of the tests in the file calls the pass.

Check out `AlgebraicSimplifier.PathologicalComplexity`, for instance. You could just write:

```
EXPECT_TRUE(AlgebraicSimplifier(default_options_).Run(m.get()).value());
```

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12688,5 +12688,31 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
...
+TEST_F(AlgebraicSimplifierTest,
+       ReducePrecisionWithSameDatatypeShouldBeRemoved) {
+  const char* hlo = R"(
+  HloModule test
+  ENTRY main {
+    param.0 = bf16[2,16,64]{2,1,0} parameter(0)
+    param.1 = bf16[32,64]{1,0} parameter(1)
+    reduce-precision.46 = bf16[32,64]{1,0} reduce-precision(param.1), exponent_bits=8, mantissa_bits=7
+    bitcast.7734 = bf16[2,16,64]{2,1,0} bitcast(reduce-precision.46)
+    multiply.1006 = bf16[2,16,64]{2,1,0} multiply(param.0, bitcast.7734)
```
bchetioui, line 12700: nit: missing `ROOT` annotation.

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12688,5 +12688,31 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
...
 
+TEST_F(AlgebraicSimplifierTest,
+       ReducePrecisionWithSameDatatypeShouldBeRemoved) {
+  const char* hlo = R"(
+  HloModule test
+  ENTRY main {
+    param.0 = bf16[2,16,64]{2,1,0} parameter(0)
+    param.1 = bf16[32,64]{1,0} parameter(1)
+    reduce-precision.46 = bf16[32,64]{1,0} reduce-precision(param.1), exponent_bits=8, mantissa_bits=7
+    bitcast.7734 = bf16[2,16,64]{2,1,0} bitcast(reduce-precision.46)
```
bchetioui, line 12699: Do we need this many instructions to exercise the test? I would have assumed that

```
p0 = bf16[64] parameter(0)
ROOT reduce-precision = bf16[64] reduce-precision(p0), exponent_bits=8, mantissa_bits=7
```

Would have been sufficient?

bchetioui (changes requested):

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
shraiysh: Done for consistency. However, if the pass returns a failure status, then wouldn't `.value()` fail awkwardly?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1675,6 +1675,9 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
     pipeline.AddPass<SimplifyFPConversions>();
   }
 
+  simplifier_options.set_enable_remove_no_op_reduce_precision(true);
+  pipeline.AddPass<GpuAlgebraicSimplifier>(simplifier_options);
```
bchetioui, line 1679: It'd be nice if we could add a test that this specific pass with this specific option always happens after any possible `SimplifyFPConversions` (since this is the pass that will be called when using `xla_allow_excess_precision`).

We have pass order tests in `gpu_compiler_test.cc` which enforce this kind of invariants; the slightly difficult thing here is that this is a pass option, and can't be recovered directly from the pass metadata.

What I can suggest (which is a slight hack, but one that may be OK for now to ensure that we don't end up breaking this important pass ordering), is to make a sub-pipeline here that contains only this pass, and then write a test that checks that this particular, aptly named pipeline runs after any possible call to `SimplifyFPConversions`. Both the test and the pipeline should have comments explaining what is being enforced, and why it is important.

WDYT?

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12688,5 +12688,36 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
...
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m,
+                          ParseAndReturnVerifiedModule(hlo));
+  default_options_.set_enable_remove_no_op_reduce_precision(true);
+  EXPECT_TRUE(AlgebraicSimplifier(default_options_).Run(m.get()).value());
+  EXPECT_THAT(m->entry_computation()->root_instruction(),
+              GmockMatch(m::Parameter()));
+}
+
+TEST_F(AlgebraicSimplifierTest,
+       ReducePrecisionWithDifferentPrecisionFromOperandShouldNotBeModified) {
```
bchetioui, line 12708: Modified[ByDefault] maybe?

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12688,5 +12688,36 @@ TEST_F(AlgebraicSimplifierTest, TestNew123) {
   EXPECT_FALSE(simplifier.Run(module.get()).value());
 }
 
+TEST_F(AlgebraicSimplifierTest,
+       ReducePrecisionWithSamePrecisionAsOperandShouldBeRemoved) {
```
bchetioui, line 12692: Removed[WhenRemoveNoOpReducePrecisionIsSet]?

(Just concerned that the two test names contradict each other :))

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1675,6 +1675,17 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
     pipeline.AddPass<SimplifyFPConversions>();
   }
 
+  {
+    HloPassPipeline& remove_no_op_reduce_precision_pipeline =
```
bchetioui, line 1679: Thanks! As requested in my previous comment:

> Both the test and the pipeline should have comments explaining what is being enforced, and why it is important.

Could you also do that, please? (Please even link to the relevant JAX issue.)

bchetioui (changes requested):

bchetioui (approved):

bchetioui:
Thank you @shraiysh for addressing all the comments!

shraiysh:
Thanks for approving the PR @bchetioui, is there an internal failure here?

akuegel:
Sorry, JAX is relying on the workaround to insert no-op ReducePrecision calls. I have personally tried this kind of change before, doesn't work.

bchetioui:
@akuegel from the discussion we had on chat earlier in December with @shraiysh and the JAX folks, we think this may work with an option, provided that the option is put in far enough within the compilation pipeline.

akuegel:
So what is the goal of this change, if we change HLO late in the pipeline? We already handle ReducePrecision no-ops as no-ops during emission time. So from a performance point, this change seems not necessary?

bchetioui:
> So what is the goal of this change, if we change HLO late in the pipeline? We already handle ReducePrecision no-ops as no-ops during emission time. So from a performance point, this change seems not necessary?

A short cross-post of the chat that ensued to make sure that things are documented where they should be:

1. I think this is about simplifying analysis for `dynamic-slice` fusions (@shraiysh can confirm);
2. `reduce-precision` is not handled as a no-op in the Triton emitter, and introducing this kind of optimizations in codegen seems like the wrong place;
3. There is a concern that someone might later introduce optimizations relying on excess precision after this pass has run. To gate that, we could make sure to explicitly disable excess precision after running this pass.

I think that the optimization is reasonable, although the ability to use excess precision unfortunately makes this pretty awkward.

shraiysh:
> I think this is about simplifying analysis for dynamic-slice fusions (I believe---@shraiysh can confirm);

That is correct. We are seeing no-op reduce precision operations between a slice and hero operation of dynamic-slice fusion and because of that we cannot fuse it.

PR #20564, opened 2024-12-16
Remove hard-coded workspace size for SDPA FP8 tests

wenscarl:
For cuDNN custom-call tests, the required workspace size can vary depending on the GPU architecture. If the required workspace size is 0 but a device pointer is provided as part of a variant pack, a cuDNN failure may occur. Return actual size required even it's 0.

The 2 tests affected by the workspace size on blackwell are
FlashAttentionBMMScaleSoftmaxBMMF8.Flash_Attention_Inference_BMM1_NoMask_Softmax_BMM2_BNTH_F8 and 
FlashAttentionBMMScaleSoftmaxBMMF8.Flash_Attention_Inference_BMM1_NoMask_Softmax_BMM2_BNTH_F8.

On Blackwell, the required workspace size is 0 as oppose to 16 on Hopper. Removing hardcoded workspace size to have cuDNN compiler handle it automatically.

### DIFF
--- a/xla/service/gpu/tests/gpu_fused_mha_test.cc
+++ b/xla/service/gpu/tests/gpu_fused_mha_test.cc
@@ -1471,8 +1471,7 @@ XLA_TEST_F(FlashAttentionBMMScaleSoftmaxBMMF8,
   XlaBuilder builder(TestName());
   std::string ref_bnth = R"(
     custom-call.4.0 = (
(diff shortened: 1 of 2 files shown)


hawkinsp (approved):

hawkinsp:
I asked for a test, offline, but I'm happy to merge this with that follow up in a future PR given this currently causes wrong outputs.

PR #20581, opened 2024-12-16
Update cadence for cpu_benchmarks.yml

juliagmt-google:


juliagmt-google:
remove PR trigger logic in the workflow

PR #20739, opened 2024-12-19
Arch specific FP8 SDPA test

wenscarl:
The workspace size required by CuDNN are different on Hopper and Blackwell. To make the HLO string arch agnostic, pin the workspace size to 0.

### DIFF
--- a/xla/service/gpu/tests/gpu_fused_mha_test.cc
+++ b/xla/service/gpu/tests/gpu_fused_mha_test.cc
@@ -1545,7 +1545,8 @@ XLA_TEST_F(FlashAttentionBMMScaleSoftmaxBMMF8,
     custom-call.21.0 = (
         f8e4m3fn[4,4,16,16]{3,1,2,0},
         f32[1,1,1,1]{3,2,1,0},
-        f32[1,1,1,1]{3,2,1,0}
+        f32[1,1,1,1]{3,2,1,0},
+        u8[0]{0}
     ) custom-call(
         convert.18,
         convert.30,
@@ -1723,7 +1724,8 @@ XLA_TEST_F(FlashAttentionBMMScaleSoftmaxBMMF8,
     custom-call.21.0 = (
         f8e4m3fn[4,16,4,16]{3,2,1,0},
         f32[1,1,1,1]{3,2,1,0},
-        f32[1,1,1,1]{3,2,1,0}
+        f32[1,1,1,1]{3,2,1,0},
+        u8[0]{0}
     ) custom-call(
         convert.18,
         convert.30,


xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1545,7 +1545,8 @@ XLA_TEST_F(FlashAttentionBMMScaleSoftmaxBMMF8,
     custom-call.21.0 = (
         f8e4m3fn[4,4,16,16]{3,1,2,0},
         f32[1,1,1,1]{3,2,1,0},
-        f32[1,1,1,1]{3,2,1,0}
+        f32[1,1,1,1]{3,2,1,0},
+        u8[0]{0}
```
sergachev, line 1549: Isn't it zero by default when unspecified?

wenscarl: Yes. On Hopper, it has to be :
```
custom-call.21.0 = (
        f8e4m3fn[4,4,16,16]{3,1,2,0},
        f32[1,1,1,1]{3,2,1,0},
        f32[1,1,1,1]{3,2,1,0}
        f32[1,1,1,1]{3,2,1,0})
```
but on Blackwell, it can also be:
```
custom-call.21.0 = (
        f8e4m3fn[4,4,16,16]{3,1,2,0},
        f32[1,1,1,1]{3,2,1,0},
        f32[1,1,1,1]{3,2,1,0}
        f32[1,1,1,1]{3,2,1,0}
```
To be arch agnostic, use 0 for simplicity.

sergachev: Does this mean that the test currently doesn't work on Hopper?

wenscarl: On hopper, the test also passes with u8[16]{0}.

sergachev: That's clear, specifying the exact workspace size works. What I still don't understand - what fails currently, without this change?

wenscarl: Without this change, it doesn't fail on Hopper, because the workspace_size is indeed 16. But it will fail on Blackwell, since the workspace size is no-longer 16 but 0. Setting to 0 regardless of arch to have cudnn compiler decide automatically.

sergachev: But it already isn't hardcoded, you removed the hardcoded value 16 here: https://github.com/openxla/xla/commit/5ed9f2fe1fe421f53e747449d956c41bfeb60298. Currently it's unset, which, in my understanding, should work as well, as setting it to zero.

wenscarl: https://github.com/openxla/xla/commit/5ed9f2fe1fe421f53e747449d956c41bfeb60298 make it fail on Hopper at graph execution, [here](https://github.com/openxla/xla/blob/949698b29f3b0e77ba21a7019b77e49fe23541a8/xla/stream_executor/cuda/cuda_dnn.cc#L8449).

sergachev: OK, makes sense then, unspecified means not using a workspace at all, zero-sized workspace means use if necessary, query cuDNN for the size.

sergachev (approved):

akuegel (approved):

PR #20744, opened 2024-12-19
[NVIDIA GPU] Add a flag to control a2a collective matmul rewrite

Tixxx:
This is address the revert in https://github.com/openxla/xla/pull/19451 where customers see MFU when enabling collective matmul by default. The a2a collective matmul kicks in by default on some small gemms and lead to inefficient transformation.
Adding a flag to disable it by default since it's experimental.

### DIFF
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -1102,7 +1102,8 @@ message DebugOptions {
   bool xla_gpu_enable_scatter_determinism_expander = 345;
 
   // Enable windowed einsum(collective matmul) rewrite for all-to-all + gemm
-  // This feature is still experimental.
+  // This feature is still experimental and effective only
+  // xla_gpu_multi_streamed_windowed_einsum is set to true.
   bool xla_gpu_experimental_enable_alltoall_windowed_einsum = 360;
 
   // Next id: 361
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -1101,7 +1101,11 @@ message DebugOptions {
   // be deterministic, although with additional overhead.
   bool xla_gpu_enable_scatter_determinism_expander = 345;
 
-  // Next id: 360
+  // Enable windowed einsum(collective matmul) rewrite for all-to-all + gemm
+  // This feature is still experimental.
+  bool xla_gpu_experimental_enable_alltoall_windowed_einsum = 360;
+
+  // Next id: 361
 
   // Extra options to pass to the compilation backend (e.g. LLVM); specific
   // interpretation of these values is left to the backend.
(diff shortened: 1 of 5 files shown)


xla/xla.proto
```diff
@@ -1098,7 +1098,11 @@ message DebugOptions {
   // be deterministic, although with additional overhead.
   bool xla_gpu_enable_scatter_determinism_expander = 345;
 
-  // Next id: 359
+  // Enable windowed einsum(collective matmul) rewrite for all-to-all + gemm
+  // This feature is still experimental.
```
reedwm, line 1102: You should mention this only applies when the flag xla_gpu_multi_streamed_windowed_einsum is true, considering the flag (briefly) is still false by default.

reedwm (changes requested):

xla/xla.proto
Tixxx: ah good point. updated

reedwm (approved):

xla/xla.proto
```diff
@@ -1098,7 +1098,12 @@ message DebugOptions {
   // be deterministic, although with additional overhead.
   bool xla_gpu_enable_scatter_determinism_expander = 345;
 
-  // Next id: 359
+  // Enable windowed einsum(collective matmul) rewrite for all-to-all + gemm
+  // This feature is still experimental and effective only
+  // xla_gpu_multi_streamed_windowed_einsum is set to true.
+  bool xla_gpu_experimental_enable_alltoall_windowed_einsum = 359;
```
penpornk, line 1104: 359 is recently taken. Could you please help resolve conflict?

dimitar-asenov (approved):

xla/xla.proto
Tixxx: updated

reedwm (approved):

dimitar-asenov (approved):

reedwm (changes requested):
Can you modify the four tests in `windowed_einsum_handler_test.cc` with "a2a" in their name to set the new flag to true? Otherwise they fail.

Also sorry for missing this before, but can you add the new flag to debug_options_flags.cc? Otherwise I think it cannot be set as an env var.

Tixxx:
> Can you modify the four tests in `windowed_einsum_handler_test.cc` with "a2a" in their name to set the new flag to true? Otherwise they fail.
> 
> Also sorry for missing this before, but can you add the new flag to debug_options_flags.cc? Otherwise I think it cannot be set as an env var.

Thanks for catching these, I updated them.

Tixxx:
@reedwm  Could you take another look at this pr? it'd be good to get it merged so we can merge the other reverted one.

dimitar-asenov (approved):

PR #20808, opened 2024-12-22
[GSPMD] Partitions collective permute instructions in manual sharding group.

yliu120:
This is a small fix in GSPMD partitioning for partitioning collective permutes instructions added in manual sharding group.

In JAX, we can add `ppermute` instruction in shard_map. In cases where we have shard_map with auto axes specified, collective permuting an operand even with the same sharding will end up with an `all-gather` and then collective permute, which leads to inefficient collectives. The correct and efficient way is to partition the collective permute as an element-wise op.

The unit test added provides a repro. Also, the JAX unit test in https://github.com/jax-ml/jax/blob/fa9c7edf736516052df6eab22947bc627d0deca3/tests/shard_map_test.py#L2167 gives a real-world JAX example.

### DIFF
--- a/xla/service/spmd/spmd_partitioner.cc
+++ b/xla/service/spmd/spmd_partitioner.cc
@@ -2566,6 +2566,11 @@ absl::Status SpmdPartitioningVisitor::HandleElementwise(HloInstruction* hlo) {
   return absl::OkStatus();
 }
 
+absl::Status SpmdPartitioningVisitor::HandleCollectivePermute(
+    HloInstruction* hlo) {
+  return HandleElementwise(hlo);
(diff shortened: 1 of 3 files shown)


yliu120:
@mattjj @pschuh collaborating with Matt and Parker on the partial-auto shard_map support.

blakehechtman (approved):

blakehechtman (approved):

yliu120:
@blakehechtman Thanks for the review. Could you please take another look?

blakehechtman (approved):

hawkinsp (approved):

PR #20944, opened 2025-01-01
Update operand layout constraints for custom call instrs when cloning.

yliu120:


xla/hlo/ir/hlo_instructions.cc
```diff
@@ -3460,7 +3460,20 @@ HloCustomCallInstruction::CloneWithNewOperandsImpl(
       opaque(), api_version_);
   if (layout_constrained()) {
     cloned->layout_constrained_ = true;
-    cloned->operand_shapes_with_layout_ = operand_shapes_with_layout();
+    std::vector<Shape> new_operand_shapes_with_layout;
```
akuegel, line 3463: I am not convinced this logic should go here?
The method CloneWithNewOperandsImpl is meant to be passed the new operands with shapes (+potentially layout) attached to them. It is the responsibility of the caller to add the right layout to the new operands.

akuegel (changes requested):

xla/hlo/ir/hlo_instructions.cc
yliu120: I guess the judgement would be whether you want to create an invalid instruction when you do the cloning.

Leaving the atttributed unchecked during cloning will lead to an invalid HLO. Then you will basically depend on some verifier to tell you the HLO is invalid. The verifier run would be some blocks away from where the HLO was created then you will end up a debugging session to see where the invalid HLO was created.

Fixing here will automatically fix all other call sites. The layout could be either pinned by the users or some passes. And I really don't know how many call sites there would be.

akuegel: The problem I see is that custom call can be anything. You don't know whether it is actually a fix to copy the original layouts, the user could have specifically chosen a different layout. I am sympathetic with the idea to fix it just in one site, but I believe it has risks, and better to fix several call sites (if needed).

@majnemer what do you think?

yliu120: Even we fix at GSPMD, it doesn't help to know the info as well. For other call sites rather than SPMD partitioner, the shape is very unlikely to be altered then this code is basically just checking whether we inherits the layout constraints correctly.

yliu120: I attached a test in this PR.
Actually the "call site" you requested doesn't exist. It falls into the Clone() method in the DefaultAction() of the SpmdPartitionVisitor: https://cs.opensource.google/tensorflow/tensorflow/+/master:third_party/xla/xla/service/spmd/spmd_partitioner.cc;drc=11273af29a81a580964d9ba1fde1baef6dc04706;l=2498

> For those calls, I am saying it is the responsibility of the caller to make sure that the passed shape or passed
> operands already have the right layout constraints.

This is not true. I really don't understand why the caller can have more information here... The layouts are pinned from the JAX user end and XLA has no knowledge about the semantics of the layout. The best it can do is to "pass" the layouts to the new shape and see if it is valid. If it is not valid, the compilation stops and gives the user an error from the place where Clone() happens.

A better way to convey semantics would be having some enums. Let's say "MajorToMinor", "MinorToMajor", etc. So that when shape changes, XLA knows how to reconstruct the layouts according to these semantics. Unfortunately they are not there.

yliu120: @mattjj @hawkinsp I guess this is a blocker for having FFI work inside shard_map partial auto.

akuegel: Thanks for pointing to the call site. As you can see, it calls CloneWithNewOperands:

https://cs.opensource.google/tensorflow/tensorflow/+/master:third_party/xla/xla/service/spmd/spmd_partitioner.cc;drc=11273af29a81a580964d9ba1fde1baef6dc04706;l=2535

In this call, it should pass the new operands with the layout constraint. In my opinion it is the responsibility of spmd_partitioner to copy the layout constraints if needed.

akuegel: So here my attempt to explain why I think it would be better if this is handled in spmd_partitioner: In spmd_partitioner we have the context to know that the new operands should in fact be derived from the original operands (including preserving layout constraints if they exist). For a potential other call site of CloneWithNewOperands, we don't necessarily know that, so if you change it in CloneWithNewOperandsImpl, you decide for every caller of the method, even if they may not actually want to preserve the layout constraint. Maybe it is a theoretical problem, I don't know.

yliu120: Thanks for your explanation. 

I have been considering this option before sending this PR out. The issue is that this operand_constraints thing only exist for HloCustomCallInstruction. It would be very weird to either add some custom logic for CustomCallInstruction in SPMD partitioner's DefaultAction, or to add a HandleCustomCall() function in the visitor for this purpose.

If there are other instructions having this layout constraints, I guess it makes sense to have a util function and reference the util function at call sites. 

> so if you change it in CloneWithNewOperandsImpl, you decide for every caller of the method
Understand. But if you don't decide, you generate invalid HLO. At least you need to check the validity of the instruction while doing the clone.

akuegel: Ok, good point that this is specific to HloCustomCallInstruction. I think I understand it a bit better now, sorry that it took so long. I was probably also the wrong reviewer for this change as I am not familiar with this layout constraint. So the problem is that this operand_shapes_with_layout_ member of HloCustomCallInstruction has both the shapes and the layouts of the operands as constraint. The current logic in CloneWithNewOperandsImpl assumes that the operand shapes don't change, which is a problem because there are possible ways to call Clone (like with CloneWithNewOperands) where the shapes of the operands *can* be changed. Right now, there is no way to update the shapes of this operand_shapes_with_layout_ member. Maybe your approach is indeed the best solution, at least I can also see the downsides in moving it to spmd_partitioner.
I would still prefer if someone else more familiar with this logic would weigh in.
@blakehechtman maybe?

akuegel: So something that would still not be fixed with this change is if a caller of CloneWithNewOperands for example modifies the number of operands, or the order of the operands. I think it is a somewhat fundamential problem of the layout constraint method used here that it implicitly assumes that operand shapes don't change. Copying the shapes of the new operands may work in the spmd_partitioner case, but may be wrong in a different context and we should better give an error in a different context.

yliu120: As Peter suggested in another meeting, I linked the JAX example here:
https://github.com/jax-ml/jax/pull/25705

In JAX, there are two things (afaik) will lower custom calls with user pinned operands layout constraints, 1) FFI and 2) io_callback.

In the above PR, I provided an example of how io_callback not working inside the partial auto feature. This is because GSPMD will further partition the "operands" or "inputs" of the custom calls. And that will need updates on the constraints.

akuegel:
@xla-rotation Can you please find a different reviewer for this PR? I am probably not the best reviewer for this.

dimitar-asenov:
@blakehechtman Could you please take a look?

reedwm:
Thank you @yliu120 for the PR and the good discussion it generated! I've been told https://github.com/jax-ml/jax/pull/25887 fixes the same issue as this PR, so closing this PR.

PR #20948, opened 2025-01-01
[GPU] Fix NCCL with CUDA 12.6.3.

sergachev:


beckerhe (approved):

reedwm:
We cannot import PRs that only touch TSL.

@ddunl what do we do here? The [TSL repo](https://github.com/google/tsl/tree/main) states to to contribute to "the [upstream location](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tsl)" but gives a dead link. Is TSL still in TensorFlow?

johnnkp:
@sergachev Please try to create pull request for https://github.com/google/tsl/blob/main/third_party/gpus/cuda/hermetic/cuda_redist_versions.bzl instead.

ddunl:
We just have to patch this in manually. At some point this quarter I will have moved all of TSL into XLA and moved the `third_party` stuff from TF to XLA which will solve this problem. I will take a look at this on Monday or Tuesday to make the patch

dimitar-asenov:
@ddunl I'm creating the necessary internal patch now, so there should be no need for you to do anything.

PR #20994, opened 2025-01-03
[GPU] Allow horizontal fusion with shared operands via concatenation.

sergachev:


akuegel (approved):
Thank you, lets give it a try.

PR #21104, opened 2025-01-07
[NVIDIA GPU] Preserve backend config when folding transpose

Tixxx:
Transpose folding pass doesn't preserve backend config when creating the new dot with transpose folded. Changing the behavior to copy the old dot's config to the new dot.

### DIFF
--- a/xla/service/transpose_folding.cc
+++ b/xla/service/transpose_folding.cc
@@ -104,7 +104,7 @@ absl::Status FoldTransposeIntoDot(InstructionOperandsPair& pair) {
   HloInstruction* new_dot =
       dot->parent()->AddInstruction(HloInstruction::CreateDot(
           dot->shape(), lhs, rhs, new_dot_dims, dot->precision_config()));
-  new_dot->CopyBackendConfigFrom(dot);
+  dot->SetupDerivedInstruction(new_dot);
   return dot->parent()->ReplaceInstruction(dot, new_dot);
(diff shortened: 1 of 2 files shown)


xla/service/transpose_folding.cc
```diff
@@ -101,10 +101,11 @@ absl::Status FoldTransposeIntoDot(InstructionOperandsPair& pair) {
...
     }
   }
-
-  return dot->parent()->ReplaceWithNewInstruction(
-      dot, HloInstruction::CreateDot(dot->shape(), lhs, rhs, new_dot_dims,
-                                     dot->precision_config()));
+  HloInstruction* new_dot =
+      dot->parent()->AddInstruction(HloInstruction::CreateDot(
+          dot->shape(), lhs, rhs, new_dot_dims, dot->precision_config()));
+  new_dot->CopyBackendConfigFrom(dot);
```
reedwm, line 107: There's also metadata that can be copied. Instead of calling `CopyBackendConfigFrom`, I think calling `new_dot->SetupDerivedInstruction(dot)` is the best way to handle everything that should be copied.

reedwm (changes requested):

xla/service/transpose_folding.cc
Tixxx: updated

reedwm (approved):

PR #21192, opened 2025-01-09
[xla:cpu] Add XLA_VLOG_LINES to oneDNN rewriter passes

cfRod:
Enables logging when we set TF_CPP_MAX_VLOG_LEVEL and TF_CPP_MIN_LOG_LEVEL


cfRod:
@penpornk

dimitar-asenov (approved):

PR #21236, opened 2025-01-10
[ROCm] Fix build break due to nvtx_with_cuda_kernels_test

mmakevic-amd:
Issue introduced here https://github.com/openxla/xla/commit/e94e29fa61bbd1ce28e60b0f6b8875581b3eacd6

This test should be skipped for ROCm so `requires-gpu-nvidia` tag is necessary


mmakevic-amd:
It seems that the `cuda-only` tag was added in the meantime, this fixes the problem as well. I will close the PR

PR #21375, opened 2025-01-13
[ds-fusion] Get While loop analysis with copy fusion

shraiysh:
In later stages of optimization, there are instances of copy fusion on the parameter of the while body. With this, we need to allow inlining of fusions while getting the induction variable index, otherwise we cannot deduce the tuple index.

### DIFF
--- a/xla/tools/hlo_extractor.cc
+++ b/xla/tools/hlo_extractor.cc
@@ -329,10 +329,20 @@ absl::Status Inline(HloModule* module) {
   for (HloComputation* computation : module->computations()) {
     for (HloInstruction* instruction : computation->instructions()) {
       if (instruction->opcode() == HloOpcode::kFusion) {
-        TF_RETURN_IF_ERROR(computation->ReplaceWithNewInstruction(
-            instruction, HloInstruction::CreateCall(
-                             instruction->shape(), instruction->operands(),
-                             instruction->fused_instructions_computation())));
+        HloInstruction* new_instruction =
+            computation->AddInstruction(HloInstruction::CreateCall(
+                /*shape=*/instruction->shape(),
+                /*operands=*/instruction->operands(),
+                /*computation=*/
+                instruction->fused_instructions_computation()));
+        TF_RETURN_IF_ERROR(computation
+                               ->ReplaceInstruction(
+                                   /*old_instruction=*/instruction,
+                                   /*new_instruction=*/new_instruction,
+                                   /*preserve_sharding=*/false,
+                                   /*relay_control_dependency=*/true,
+                                   /*remove_unused_operands=*/true)
+                               .status());
       }
     }
   }
--- a/xla/hlo/analysis/while_loop_analysis.cc
+++ b/xla/hlo/analysis/while_loop_analysis.cc
@@ -177,7 +177,7 @@ static std::optional<int64_t> GetUniqueGTEDependenceIndex(
       [](const HloInstruction* inst) -> ReplaceType {
         return ReplaceType::kReplaceParam;
       },
-      /*cross_computation=*/false, /*inline_calls_and_fusions=*/false,
+      /*cross_computation=*/false, /*inline_calls_and_fusions=*/true,
       /*run_verifier=*/false);
   HloComputation* entry = extracted->entry_computation();
 
--- a/xla/hlo/analysis/while_loop_analysis_test.cc
+++ b/xla/hlo/analysis/while_loop_analysis_test.cc
@@ -953,5 +953,46 @@ TEST_F(WhileLoopAnalysisTest,
   EXPECT_EQ(trip_count, std::nullopt);
 }
 
+TEST_F(WhileLoopAnalysisTest, GetIndvarIndexShouldWorkWhenParamIsCopied) {
+  const char* hlo = R"(
+    HloModule test
+
+    fused_copy {
+      param.1 = (s32[],s32[]) parameter(0)
+      ROOT copy = (s32[], s32[]) copy(param.1)
+    }
+
+    body {
+      param.1 = (s32[], s32[]) parameter(0)
+      copy_fusion = (s32[], s32[]) fusion(param.1), kind=kInput, calls=fused_copy
+      iter.1 = s32[] get-tuple-element(copy_fusion), index=0
+      c.1 = s32[] constant(1)
+      add.1 = s32[] add(iter.1, c.1)
+      data.1 = s32[] get-tuple-element(copy_fusion), index=1
+      ROOT tuple = (s32[], s32[]) tuple(add.1, data.1)
+    }
+
+    condition {
+      param = (s32[], s32[]) parameter(0)
+      iter = s32[] get-tuple-element(param), index=0
+      c.10 = s32[] constant(10)
+      ROOT compare = pred[] compare(iter, c.10), direction=LT
+    }
+
+    ENTRY main {
+      c0 = s32[] constant(0)
+      data = s32[] parameter(0)
+      tuple = (s32[], s32[]) tuple(c0, data)
+      ROOT while = (s32[], s32[]) while(tuple), body=body, condition=condition
+    }
+  )";
+
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m,
+                          ParseAndReturnVerifiedModule(hlo));
+  HloInstruction* while_op = m->entry_computation()->root_instruction();
+  ASSERT_EQ(while_op->opcode(), HloOpcode::kWhile);
+  EXPECT_EQ(GetLoopInductionVarTupleIdx(while_op), 0);
+}
+
 }  // namespace
 }  // namespace xla


fhoushmand (approved):

fhoushmand:
Rolling back due to internal test breakage.

shraiysh:
Do we have a reduced testcase or the reason for this breakage? I can investigate and try to submit again.

dimitar-asenov:
I'm working on trying to get a minimal reproducer. In the mean time, here's then failing stack trace:

```
Check failed at xla/hlo/ir/hlo_computation.cc:1413: DCHECK(changed) 

xla/hlo/ir/hlo_computation.cc:1413 xla::HloComputation::ReplaceInstruction()
xla/hlo/ir/hlo_computation.cc:1381 xla::HloComputation::ReplaceWithNewInstruction()
xla/tools/hlo_extractor.cc:332 xla::(anonymous namespace)::Inline()
xla/tools/hlo_extractor.cc:369 xla::ExtractModule()
xla/hlo/analysis/while_loop_analysis.cc:173 xla::GetUniqueGTEDependenceIndex()
xla/hlo/analysis/while_loop_analysis.cc:432 xla::GetLoopInductionVarTupleIdx()
xla/hlo/analysis/while_loop_analysis.cc:487 xla::MatchTrivialLoopRange()
```

Enabling extra logging reveals that `CHECK` fails happens because the replacement fails with:

```
hlo_computation.cc:1432 Skipping replacement because old instruction has control dependencies
```

In this particular case the old instruction is a fusion and the new one is a call.

Will share a reproducer once I get a more minimal example.

shraiysh:
Thanks for looking into this. I think the fix should be dropping those control dependencies (or relaying them) in the extracted module -- it shouldn't matter which one is done. For now, I have added dropping for those. If you have a testcase, we can add that for this. I tried but could not build one.

dimitar-asenov:
I spent some time trying to get a minimal example. Below is what I have so far and I won't try to minimize it even more. It seems somewhat convoluted and removing anything from it (tuples, or one of the loops) causes it not to fail. Please look into this and either include it as a test, or include some more minimal version.

As is, this crashes with the same stack trace as above and due to the same reason `Skipping replacement because old instruction has control dependencies`:

```C++
  constexpr absl::string_view kHlo = R"(
cond {
  p = (s8[2], s8[]) parameter(0)
  i = s8[] get-tuple-element(p), index=1
  z = s8[] constant(0)
  ROOT r = pred[] compare(i, z), direction=LT
}

inner_body {
  p = (s8[2], s8[]) parameter(0)
  s = s8[] get-tuple-element(p), index=1

  sz = s8[] constant(0)
  z = s8[2] constant(0)
  add = s8[] add(s, sz), control-predecessors={z}
  ROOT r = (s8[2], s8[]) tuple(z, add)
}

outer_body {
  p = (s8[2], s8[]) parameter(0)
  s = s8[] get-tuple-element(p), index=1

  z = s8[2] constant(0)
  t = (s8[2], s8[]) tuple(z, s)
  w = (s8[2], s8[]) while(t), condition=cond, body=inner_body
  ss = s8[] get-tuple-element(w), index=1

  ROOT r = (s8[2], s8[])  tuple(z, ss)
}

ENTRY main {
  p = s8[2] parameter(0)
  z = s8[] constant(0)
  t = (s8[2], s8[]) tuple(p, z)
  ROOT w = (s8[2], s8[]) while(t), condition=cond, body=outer_body
}
)";

  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
                          ParseAndReturnVerifiedModule(kHlo));
  auto loop = hlo_query::GetFirstInstructionWithOpcode(
      *module->entry_computation(), HloOpcode::kWhile);
  auto index = GetLoopInductionVarTupleIdx(loop);
  ```

**Comment by shraiysh** [comment]
Thank you so much. This is small enough, I am taking a look. 🙇

**Comment by shraiysh** [comment]
This was a bug in the algebraic simplifier. Submitted a separate PR for that bug: #21437.

File: xla/tools/hlo_extractor.cc [hunk 1]
```diff
@@ -329,6 +329,7 @@ absl::Status Inline(HloModule* module) {
   for (HloComputation* computation : module->computations()) {
     for (HloInstruction* instruction : computation->instructions()) {
       if (instruction->opcode() == HloOpcode::kFusion) {
+        TF_RETURN_IF_ERROR(instruction->DropAllControlDeps());
```
dimitar-asenov, line 332: Why is it reasonable to drop all control deps here?

shraiysh: We had two choices here: drop the control dependencies or relay them. It doesn't really matter what we do, because these instructions are going to get inlined anyways. I have tried to relay those dependencies instead of dropping them here. Let me know if this causes the test to fail, I will switch it back to dropping them instead.

dimitar-asenov: Thanks.

I tried the new code, but it fails in the same way as it did before. So I guess we could go back to dropping these.

> It doesn't really matter what we do, because these instructions are going to get inlined anyways.

Could you please help me understand, why control-dependencies are irrelevant when inlining?

shraiysh: > Could you please help me understand, why control-dependencies are irrelevant when inlining?

Based on my understanding of the call inliner pass, the control dependencies on the `call` instruction are not relayed. The only [dependency handling](https://github.com/openxla/xla/blob/e38c3f4e9d60d8984ef355215044dd99e97ce29c/xla/service/call_inliner.cc#L80C36-L80C58) they have there is for the instructions within the called computation and not the caller itself. It also isn't obvious what would the dependency translate to while inlining (which is something that the call inliner must answer and does not concern hlo_extractor or while_loop_analysis). That is, if the call instruction depends on an operation `op`, then after inlining, do all the inlined operations depend on `op` (?) or do the root instructions depend on `op` (?) or do the parameter instructions depend on `op` (?). The call inliner just chooses to not relay this dependency and would instead fail to replace the call instruction [here](https://github.com/openxla/xla/blob/e38c3f4e9d60d8984ef355215044dd99e97ce29c/xla/service/call_inliner.cc#L101). Please let me know if I am incorrect here.

dimitar-asenov: Thanks for the detailed explanation. The example where we saw a failure looks like this:

```
  ...
  fusion_1 = fusion(...), calls=f1
  copy = copy(/*args unrelated to fusion_1*/), control-predecessors={fusion_1}
  ...
  fusion_2 = fusion(b, ...) calls=f2
  add = add(/*args unrelated to fusion_2*/), control-predecessors={fusion_2}
  ...
  ```
  
Not sure if it matters, but the above was the body of an inner loop (see https://github.com/openxla/xla/pull/21375#issuecomment-2591096389 above).
  
So in this case, assuming we're inlining `fusion_1` and `fusion_2`, I think adjusting the existing control dependencies to point to the newly inlined roots of those fusions makes sense. What do you think?

shraiysh: That's probably the right thing to do, but this changes the behavior of the `CallInliner` and not in the scope of the changes here. Ofcourse, relaying the dependencies is the absolute best thing to do, while replacing the fusion with call in hlo_extractor, but because the CallInliner does not care about the control dependencies at the moment (and I do not have access to the testcase to debug and fix this issue very well), it doesn't matter if we relay or drop the control dependencies here. I am open to suggestions about the next steps - it could be as simple as what you mentioned, but without the testcase, it would be hard to ascertain if that would fix the issue.

shraiysh: Additional clarification: the dropping of control dependencies here is happening on the extracted module. The original module stays as it is and is not modified.

shraiysh:
Ping for review.

xla/tools/hlo_extractor.cc
dimitar-asenov: I still feel uneasy about this. Here's why:
- In the specific example that you care about, `Extract` is called as an implementation detail of a pure method (`GetUniqueGTEDependenceIndex`). I agree that in that case, it probably doesn't matter what we do and dropping the dependencies is fine.
- However, `Extract` is a public API and it is used in other cases. We have no guarantee that in those cases it's ok to simply drop dependencies. Or is your argument that those cases currently fail anyway if they contain such dependencies, so we can't make things worse? On some level that might be true, but we might start silently allowing new ways of calling `Extract` that should fail, but don't.

So how about an extra level of defense:
- Add one more parameter to `Extract` called `drop_control_dependencies`.
- Explain what it does in the comment of `Extract`.
- Forward this bool to `Inline` and conditionally do the dropping based on whether it's used or not.
- Set it to true in the `GetUniqueGTEDependenceIndex` case.

At least then we'll be sure that nothing is broken.

dimitar-asenov (changes requested):

xla/tools/hlo_extractor.cc
dimitar-asenov: @fhoushmand Does the above make sense to you?

shraiysh: Yes, I can do this. Or, if we can quickly land the fix to the CallInliner (what you suggested above), then I can do that too in a separate PR - would that be okay?

dimitar-asenov: The fix above would be preferable, yes. It keeps the API clean. Landing as a separate PR would be great.

shraiysh: Sure, let me open up a PR for that (hopefully that doesn't cause other failures :laughing:)

shraiysh: Added a new PR for this change in call inliner here: https://github.com/openxla/xla/pull/21680.

shraiysh:
Updated this PR with relaying control dependencies. @dimitar-asenov can you check if this passes with the internal tests? If not, can you please give me a stack trace?

dimitar-asenov:
> Updated this PR with relaying control dependencies. @dimitar-asenov can you check if this passes with the internal tests? If not, can you please give me a stack trace?

The test which was failing before now passes. I don't yet know if something else broke, but that's a good start.

@fhoushmand Does the new change look good to you?

akuegel (approved):

akuegel:
@shraiysh FYI this triggered a bug in AlgebraicSimplifier, by now I have a fix ready, but we reverted first to unbreak the user as soon as possible. After my fix has landed we will attempt to reland.

dimitar-asenov:
The fix has landed and this PR was re-merged. Let's hope there are no more breakages.

shraiysh:
Thanks for helping land this @akuegel @dimitar-asenov! Let me know if I can help if there are future breakages.

dimitar-asenov:
Unfortunately, we'll have to roll this back again as it caused another preexisting issue with the Algebraic Simplifier to surface. We'll look into it next week, I don't have a reproducer at the moment.

shraiysh:
> Unfortunately, we'll have to roll this back again as it caused another preexisting issue with the Algebraic Simplifier to surface. We'll look into it next week, I don't have a reproducer at the moment.

Sure. No problem. Please let me know how I can help.

dimitar-asenov:
This change was re-merged. Let's hope that was the last issue.

PR #20975, opened 2025-01-02
[ROCm] Fix build break due to XNNPACK update and cuda profiler test

hsharsha:
ROCm build and test breaks due to
https://github.com/openxla/xla/pull/20542 and
https://github.com/openxla/xla/pull/20488

PR #21396, opened 2025-01-14
[ROCm] Fix build break due to XNNPACK update and add NCCL_MAX_NCHANNELS to multi gpu tests

mmakevic-amd:
`NCCL_MAX_NCHANNELS=1`  is necessary for collective ops tests to pass in CI.

As for XNNPACK problem, similar fix has already been done for single-gpu tests -> https://github.com/openxla/xla/pull/20975


build_tools/rocm/run_xla_multi_gpu.sh
```diff
@@ -72,6 +72,7 @@ TAGS_FILTER="${TAGS_FILTER},${UNSUPPORTED_GPU_TAGS// /,}"
 
 bazel \
     test \
+    --define xnn_enable_avxvnniint8=false --define xnn_enable_avx512fp16=false \
```
mooskagh, line 75: All other arguments are one per like, could you split this too?

mooskagh (approved):

PR #21430, opened 2025-01-14
When a profile is not found for the current device, default to the latest available instead of sm_86.

dimvar:
Often, when a profile is not available, it is for a new device. So, it makes sense to grab the latest available profile, instead of the somewhat arbitrary sm_86.

### DIFF
--- a/xla/service/gpu/model/hlo_op_profiles.cc
+++ b/xla/service/gpu/model/hlo_op_profiles.cc
@@ -21,22 +21,21 @@ limitations under the License.
 #include <variant>
 
 #include "absl/memory/memory.h"
+#include "absl/strings/numbers.h"
 #include "absl/strings/str_cat.h"
+#include "tsl/platform/logging.h"
+#include "tsl/platform/protobuf.h"
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/service/gpu/model/hlo_op_profile.pb.h"
 #include "xla/service/gpu/model/hlo_op_profiles_data.h"
 #include "xla/stream_executor/device_description.h"
-#include "tsl/platform/logging.h"
-#include "tsl/platform/protobuf.h"
 
 namespace xla {
 namespace gpu {
 
 /*static*/ const HloOpProfiles& HloOpProfiles::Singleton() {
   static const auto* hlo_op_profiles =
-      HloOpProfiles::Load(kDeviceHloOpProfiles,
-                          /*default_profile_name=*/"sm_86")
-          .release();
+      HloOpProfiles::Load(kDeviceHloOpProfiles).release();
   return *hlo_op_profiles;
 }
 
@@ -50,8 +49,7 @@ namespace gpu {
 }
 
 /*static*/ std::unique_ptr<HloOpProfiles> HloOpProfiles::Load(
-    absl::string_view profiles_text_proto,
-    absl::string_view default_profile_name) {
+    absl::string_view profiles_text_proto) {
   ProfilesNestedMap profiles_map;
   DeviceHloInstructionProfiles all_device_profiles;
   CHECK(tsl::protobuf::TextFormat::ParseFromString(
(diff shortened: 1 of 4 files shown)


dimvar:
Hi. Any update on this?

xla/service/gpu/model/hlo_op_profiles.cc
```diff
@@ -65,15 +63,34 @@ namespace gpu {
           op_code, element_type)] = entry.clock_cycles();
     }
   }
-  return absl::WrapUnique(
-      new HloOpProfiles(std::move(profiles_map), default_profile_name));
+  return absl::WrapUnique(new HloOpProfiles(std::move(profiles_map)));
+}
+
+const HloOpProfiles::HloOpProfile& HloOpProfiles::GetLatestProfile() const {
+  auto name_to_version = [](std::string profile_name) {
```
olegshyshkov, line 70: It would be nice to move the logic that finds the sm version of the latest profile to the Load function. Load will be called only once per program run in the Singleton function, while GetLatestProfile will be called every time we create GpuHloCostAnalysis.

olegshyshkov (changes requested):

xla/service/gpu/model/hlo_op_profiles.cc
dimvar: Good point. Done.

olegshyshkov (approved):
Nice! Thank you!

xla/service/gpu/model/hlo_op_profiles.h
```diff
@@ -50,22 +50,21 @@ class HloOpProfiles {
...
 
-  const HloOpProfile& GetDefaultProfile() const { return default_profile_; }
+  const HloOpProfile& GetLatestProfile() const { return latest_profile_; }
 
  private:
-  HloOpProfiles(ProfilesNestedMap profiles,
-                absl::string_view default_profile_name)
-      : profiles_(std::move(profiles)),
-        default_profile_(profiles_.at(default_profile_name)) {}
+  HloOpProfiles(ProfilesNestedMap profiles)
```
mooskagh, line 61: Could you mark this constructor `explicit`?  
(getting a linter warning for that)

xla/service/gpu/model/hlo_op_profiles.cc
```diff
@@ -21,22 +21,21 @@ limitations under the License.
 #include <variant>
 
 #include "absl/memory/memory.h"
+#include "absl/strings/numbers.h"
```
mooskagh, line 24: Could you also add:

```
#include <cstdint>

#include "absl/log/check.h"
#include "absl/log/log.h"
```

(linter warning, for `int32_t`, `CHECK()` and `VLOG()`)

xla/service/gpu/model/hlo_op_profiles.h
dimvar: done

xla/service/gpu/model/hlo_op_profiles.cc
dimvar: done

akuegel:
I see a new revision was pushed, commenting to get the changes imported.

mooskagh:
These tests are failing currently -- could you fix them?
```
//xla/service/gpu/model:gpu_hlo_cost_analysis_test
//xla/service/gpu/model:gpu_performance_model_test
```

It makes sense to somehow fix the device description for which the test is running, rather than just update the numbers, to make it not fail next time similar change is made.

olegshyshkov (changes requested):
> These tests are failing currently -- could you fix them?
> 
> ```
> //xla/service/gpu/model:gpu_hlo_cost_analysis_test
> //xla/service/gpu/model:gpu_performance_model_test
> ```
> 
> It makes sense to somehow fix the device description for which the test is running, rather than just update the numbers, to make it not fail next time similar change is made.

@dimvar Could you please update the tests that are failing?

dimvar:
Hi. Yes, I looked at this briefly last week but then got caught up with other stuff. I'll update the PR today or tomorrow.

dimvar:
Hi, I looked at the failures and I didn't see a fast way to fix them without just hardcoding the new numbers. It's a bit of a rabbit hole to fix them properly. There's a handful of other higher-priority things I'm looking at right now, so I think for now I'll just close the PR, and I may return to it in the future.

PR #21437, opened 2025-01-14
[ds-fusion] Fix algebraic simplifier error in debug mode.

shraiysh:
This error was observed while trying to land #21375 (which is needed for the ds-fusion work).

This error occurs when there is a constant operation that can be converted into a scalar broadcast, but some other operation is a successor for the constant operation (via control dependency). Such a dependency is not relayed and so the operation is not converted even after the
`ReplaceWithNewInstruction` function call. This causes a runtime error in debug mode testing. Fixing this by relaying this control dependency.

### DIFF
--- a/xla/hlo/ir/dfs_hlo_visitor_with_default.h
+++ b/xla/hlo/ir/dfs_hlo_visitor_with_default.h
@@ -352,16 +352,16 @@ class DfsHloRewriteVisitor : public DfsHloVisitorWithDefault {
   // Replaces the existing HLO instruction old_instruction, with
   // new_instruction, and marks the optimizer status as changed.
   // Returns the absl::Status representing the result of the replace operation.
-  absl::StatusOr<bool> ReplaceInstruction(
-      HloInstruction* old_instruction, HloInstruction* new_instruction,
-      bool preserve_sharding, bool relay_control_dependency = false) {
+  absl::StatusOr<bool> ReplaceInstruction(HloInstruction* old_instruction,
+                                          HloInstruction* new_instruction,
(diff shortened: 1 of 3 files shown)


shraiysh:
Clang format errors are unrelated to this patch. Please let me know if you would like for me to format that anyway. 

Also, the problem of control dependencies is a wider problem within the algebraic simplifier, i.e. control dependencies are usually not relayed and they could lead to such failures in debug mode (and missed opportunities for simplification in opt mode). I am open to other ideas for solving this instead of relaying the dependencies only in this case. I chose to handle it only here because I did not want to stall the other PR because of issues in the algebraic simplifier.

akuegel (approved):

dimitar-asenov:
I don't think this is the right solution. Maybe the approach is right, but is incomplete.

When I distilled the minimal failing example in https://github.com/openxla/xla/pull/21375, I did change instructions to simplify the HLO. In particular, the original HLO contains two `control-predecessors={...}` one appears on a `copy` instruction, the other on an `add` instruction. Both specify a dependency on a `fusion` (different fusions). This is unlike the minimal example where the dependency is on a constant.

The fix in this PR works for the minimal example, but not for the original failing HLO. We need a more general solution that works across the board.

dimitar-asenov:
OK, the latest change seems to work. I'm wondering if it may have some unforeseen side-effects. 

@Tongfei-Guo do you think this PR is reasonable?

shraiysh:
Ping for review!

dimitar-asenov:
We merged this, as Tongfei confirmed the change looks good.

PR #21511, opened 2025-01-16
Fix bitcast transposes in layout normalization.

### LINKED ISSUE #25759: XLA runtime error when taking grad + vmap + scan on GPU
### Description

I'm encountering a bug that only appears when using JAX on the GPU and when taking the gradient of a function that composes vmap and scan.  The code that I've pasted below seems to work if when the device, vmap or scan are changed to CPU mode or for loops.

```python
import jax
import jax.numpy as jnp
import jax.random as random
from jaxtyping import Array, Float

def recreate_bug():
  key = random.PRNGKey(0)
  dim = 6
  T = 3
  batch_size = 2

  # Each (2, 2, dim//2) matrix is a 2x2 block matrix where each block is a dim//2 x dim//2 diagonal matrix
  batched_matrix_elements = random.normal(key, (batch_size, T, 2, 2, dim//2))

  # Create a dummy global vector to initialize the scan with
  global_x = random.normal(key, (dim,))

  def objective(batched_matrix_elements: Float[Array, 'B T 2 2 dim//2']):
    """Compute the average of the objective function over the batch dimension"""

    def unbatched_objective(matrix_elements: Float[Array, 'T 2 2 dim//2']):
      """If matrix elements represent the elements of the 2x2 block matrices (A, B, C),
         then this function computes ... C @ B @ A @ global_x"""

      def scan_body(x: Float[Array, 'dim'], matrix_elements: Float[Array, '2 2 dim//2']):
        """This is just a matrix multiplication of the 2x2 block matrix with x"""
        x_reshaped = x.reshape((2, -1))
        y = jnp.einsum('i j a, j a -> i a', matrix_elements, x_reshaped).ravel()
        return y, ()

      # Code works if we use a for loop instead of a scan!
      xT, _ = jax.lax.scan(scan_body, global_x, matrix_elements)
      return xT.mean()

    # Code works if we don't vmap!
    return jax.vmap(unbatched_objective)(batched_matrix_elements).mean()

  grad = jax.grad(objective)(batched_matrix_elements)
  print('Worked!')

if __name__ == '__main__':
  jax.print_environment_info()
  recreate_bug()

```

I've been able to recreate the bug on an a100 gpu (see the system info below) and also a 2080ti.  Any help would be greatly appreciated!

### System info (python version, jaxlib version, accelerator, etc.)

```jax:    0.4.38
jaxlib: 0.4.38
numpy:  1.24.4
python: 3.10.15 (main, Oct  3 2024, 07:27:34) [GCC 11.2.0]
device info: NVIDIA A100-SXM4-80GB-1, 1 local devices"
process_count: 1
platform: uname_result(system='Linux', node='gpu021', release='6.8.0-48-generic', version='#48-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 27 14:04:52 UTC 2024', machine='x86_64')


$ nvidia-smi
Tue Jan  7 21:00:10 2025       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.127.05             Driver Version: 550.127.05     CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA A100-SXM4-80GB          On  |   00000000:4E:00.0 Off |                    0 |
| N/A   34C    P0             71W /  500W |     429MiB /  81920MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
                                                                                         
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A   4005205      C   python                                        420MiB |
+-----------------------------------------------------------------------------------------+

jreiffers:
Currently, transposes that are bitcasts are converted to a bitcast that does not satisfy the invariants of the layout normalization pass.

As far as I can tell, the special handling of bitcast transposes does nothing useful, so we can simply remove it.

While we're here, we can stop emitting identity transposes.

This fixes https://github.com/jax-ml/jax/issues/25759.

### DIFF
--- a/xla/service/layout_normalization.cc
+++ b/xla/service/layout_normalization.cc
@@ -515,15 +515,7 @@ class LayoutNormalizationVisitor : public DfsHloRewriteVisitor {
     return absl::OkStatus();
   }
 
-  // For bitcasting transposes, converts:
-  //
-  // A{I} -> bitcast[S]{L} -> transpose{L2}
-  //
-  // Into:
-  //
-  // A{I} -> bitcast{L2}
-  //
-  // For non-bitcasting ones, converts:
+  // Converts:
   //
   // A{I} -> bitcast[S0]{L} -> transpose[S]{L2}
   //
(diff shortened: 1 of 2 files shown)


bchetioui (approved):

akuegel:
Thanks for the fix. As far as I understand, the *intent* of the check for bitcast transposes was to avoid emitting identity transposes. It just did the wrong check, which you now have fixed :)

jreiffers:
Thanks for taking a look, Adrian. There was also a test that explicitly verified the presence of an identity transpose, so it's a bit strange. Hopefully nothing depended on that behavior :-).

PR #21638, opened 2025-01-20
Add the hlo verifier before host offloader to check host memory space

zhenying-liu:
Ensure No Instructions Have Host Memory Space S(5) Before Host Offloader

This change verifies that no instruction possesses host memory space S(5) prior to the host offloader pass. It addresses an issue where the HLO passes before the host offloader could inadvertently leak memory space annotations from the entry computation layout to the graph.

In PR https://github.com/openxla/xla/pull/20426, the layout assignment pass was corrected to prevent instructions from inheriting memory space S(5) from the entry computation layout. This commit further ensures that such annotations are not propagated, keeping host memory space not changed until the host offloader pass.

### DIFF
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -1593,6 +1593,12 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
   // also have unsorted update_window_dims.
   pipeline.AddPass<ScatterSimplifier>();
 
+  // Verify the host memory space before the host offloader pass
+  std::unique_ptr<TargetVerifierMetadata> verifier_metadata =
+      std::make_unique<CpuGpuVerifierMetadata>(
+          std::move(HloVerifierOpts{}.VerifyNoHostMemorySpace()));
+  pipeline.AddPass<HloVerifier>(std::move(verifier_metadata));
+
   pipeline.AddPass<HostOffloader>();
 
   TF_RETURN_IF_ERROR(
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -1596,7 +1596,7 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
   // Verify the host memory space before the host offloader pass
   std::unique_ptr<TargetVerifierMetadata> verifier_metadata =
       std::make_unique<CpuGpuVerifierMetadata>(
-          std::move(HloVerifierOpts{}.VerifyNoHostMemorySpace()));
+          HloVerifierOpts{}.VerifyNoHostMemorySpace());
   pipeline.AddPass<HloVerifier>(std::move(verifier_metadata));
 
   pipeline.AddPass<HostOffloader>();
--- a/xla/service/hlo_verifier.cc
+++ b/xla/service/hlo_verifier.cc
@@ -3107,19 +3107,20 @@ class InstructionVerifier : public DfsHloVisitorWithDefault {
   // memory space.
   static absl::Status VerifyNoHostMemorySpace(
       const HloInstruction* instruction) {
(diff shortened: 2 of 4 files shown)


xla/service/gpu/gpu_compiler.cc
```diff
@@ -1660,6 +1660,12 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
   // also have unsorted update_window_dims.
   pipeline.AddPass<ScatterSimplifier>();
 
+  // Verify the host memory space before the host offloader pass
+  std::unique_ptr<TargetVerifierMetadata> verifier_metadata =
+      std::make_unique<CpuGpuVerifierMetadata>(
+          std::move(HloVerifierOpts{}.VerifyNoHostMemorySpace()));
```
SandSnip3r, line 1666: The `std::move` here is redundant as the `HloVerifierOpts` object is already an r-value.

xla/service/hlo_verifier.cc
```diff
@@ -3077,6 +3080,24 @@ class InstructionVerifier : public DfsHloVisitorWithDefault {
     return absl::OkStatus();
   }
 
+  // Returns an error status if an instruction or any operand contains host
+  // memory space.
+  static absl::Status VerifyNoHostMemorySpace(
+      const HloInstruction* instruction) {
+    if (LayoutUtil::IsDenseArray(instruction->shape()) &&
```
SandSnip3r, line 3087: Wouldn't it be better here to use `ForEachSubshape`? Then you don't need `IsDenseArray`.

xla/service/hlo_verifier.cc
```diff
@@ -3077,6 +3080,24 @@ class InstructionVerifier : public DfsHloVisitorWithDefault {
...
+        instruction->shape().has_layout()) {
+      const Shape& result_shape = instruction->shape();
+      const Layout& result_layout = result_shape.layout();
+      if (result_layout.memory_space() == Layout::kHostMemorySpace) {
+        return absl::InvalidArgumentError(absl::StrCat(
```
SandSnip3r, line 3092: `InvalidArgumentError` does not best fit. `Internal` seems to be more used in this file. Since the error will be propagated up to the user, it is better that they see it as an "internal" failure, which it is, rather than an invalid input(argument) to XLA.

SandSnip3r (changes requested):

xla/service/gpu/gpu_compiler.cc
zhenying-liu: Agree. Removed this std::move.

xla/service/hlo_verifier.cc
zhenying-liu: There is an outer loop for Postprocess() to traverse each operand of an instruction. We only need to check the current instruction here, not its operands. So we don't need to introduce a loop like ForEachSubShape.

xla/service/hlo_verifier.cc
zhenying-liu: Thanks for pointing this out. Replaced InvalidArgumentError with Internal.

zhenying-liu:
@SandSnip3r Please have another look at the review. Thanks!

xla/service/hlo_verifier.cc
SandSnip3r: ForEachSubShape does not "traverse each operand". It will only iterate over the subshapes of the given shape. It's essentially a way to traverse the leaves of the tree of arbitrarily nested tuples. Your approach only checks if the root level is not a tuple.

zhenying-liu: Thanks for the explanation. Replaced with ForEachSubshapeWithStatus. Done.

zhenying-liu: @SandSnip3r Please take another look. Thanks.

SandSnip3r (approved):

PR #20395, opened 2024-12-10
[XLA:GPU] Add NVSHMEM library and initialization test

trevor-m:
This PR is the first step in integrating the [NVSHMEM library](https://developer.nvidia.com/nvshmem) into XLA. There are two use cases in XLA that we are targeting with NVSHMEM: 1. Host-side collectives which are much faster than NCCL at small message sizes and 2. Generating fused communication and compute kernels with Pallas and Mosaic GPU.

This PR adds the library dependency which follows the stub loading mechanism used by other NVIDIA libraries in XLA. We also add a singleton wrapper API around it, and a unit test which checks that NVSHMEM can be initialized successfully. The build rules download the nvshmem headers during XLA build and uses the system installed shared library at runtime.

cc @nvcastet @Tixxx

PR #21683, opened 2025-01-21
[XLA:GPU] NVSHMEM allocation

trevor-m:
Requires https://github.com/openxla/xla/pull/20395 which adds the NVSHMEM library dependency. 

This PR adds the following:
1. Nvshmem flag to enable nvshmem
2. Set nvshmem initialization issue when GPU PJRT client is created. The first time NVSHMEM is used, it will be initialized.
3. Uses the user buffer memory pool for nvshmem. If nvshmem is enabled, it will be allocated using `nvshmem_malloc`. This same memory can be used by user buffers if nccl user buffers is also enabled.
4. Update the `CollectiveColorer` so that mosaic_gpu custom calls use the nvshmem memory space.


xla/debug_options_flags.cc
```diff
@@ -159,6 +159,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_nccl_termination_timeout_seconds(-1);
   opts.set_xla_gpu_enable_shared_constants(true);
   opts.set_xla_gpu_enable_nccl_user_buffers(false);
+  opts.set_xla_gpu_enable_nvshmem(true);
```
frgossen, line 162: Can you rename the flag to `xla_gpu_experimental_enable_nvshmem` as we aim to enable this by default at some point

xla/debug_options_flags.cc
```diff
@@ -1505,6 +1506,11 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       "Enables NCCL User Buffer Registration. collective_memory_size in the "
       "allocator config must also be set to a non-zero value that is large "
       "enough to meet peak collective memory usage."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_enable_nvshmem",
```
frgossen, line 1510: Same

frgossen (commented):
Can you think of a meaningful way to add a test case here? 
I assume we can only do that when NVSHMEM is integrated as a collectives backend

frgossen (approved):

frgossen (changes requested):
Looks good, can you chnage the flag name?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1186,6 +1187,33 @@ absl::StatusOr<DeviceTopologyPair> BuildDistributedDevices(
   gpu_executable_run_options->set_gpu_global_device_ids(
       std::move(gpu_device_ids));
 #if GOOGLE_CUDA || TENSORFLOW_USE_ROCM
+  static const bool xla_gpu_enable_nvshmem =
+      xla::GetDebugOptionsFromFlags().xla_gpu_enable_nvshmem();
+  if (xla_gpu_enable_nvshmem) {
+    std::weak_ptr<KeyValueStoreInterface> kv_store_weak_ptr = kv_store;
+    auto store_get_fn =
+        [kv_store_weak_ptr](std::string_view key) -> absl::StatusOr<std::string> {
```
penpornk, line 1195: Nit: We prefer `absl::string_view`. Same for other places in this file.
```suggestion
        [kv_store_weak_ptr](absl::string_view key) -> absl::StatusOr<std::string> {
```

penpornk (changes requested):
+1 on adding a test. If this doesn't need a test, could you please provide the reason. Thank you!

trevor-m:
Thanks @frgossen and @penpornk for reviewing! I will add a test.

xla/service/gpu/gpu_memory_space_assignment.h
```diff
@@ -49,15 +45,37 @@ inline BufferAssigner::Colorer CollectiveColorer() {
...
     };
+  if (kSupportedOpcodes->contains(alias->instruction()->opcode()))
+    return true;
+  return (alias->instruction()->opcode() == HloOpcode::kAsyncStart ||
+          alias->instruction()->opcode() == HloOpcode::kAsyncDone) &&
+             kSupportedOpcodes->contains(
+                 alias->instruction()->async_wrapped_opcode());
+}
+
+inline bool IsNvshmemOp(const HloValue* alias) {
```
frgossen, line 56: Is this limited to mosaic? If so, maybe include mosaic in the name. If not, can you add a comment what is considered an nvshmem op?

frgossen (approved):
Thank you!

xla/service/gpu/gpu_memory_space_assignment.h
trevor-m: This code currently looks for mosaic gpu custom call ops which have nvshmem instructions - I think the current comment explains this well? IF not, I can add more details.
With @Tixxx's host-side nvshmem collectives which are coming soon, we will also have to extend this to cover collective ops which use the nvshmem backend.

xla/stream_executor/cuda/cuda_collectives.cc
```diff
@@ -32,9 +34,12 @@ namespace stream_executor::gpu {
 /* static */ absl::StatusOr<void*> CudaCollectives::CollectiveMemoryAllocate(
     StreamExecutor* executor, uint64_t bytes) {
   if (bytes == 0) return nullptr;
-
   std::unique_ptr<ActivateContext> activation = executor->Activate();
 
+  if (xla::GetDebugOptionsFromFlags().xla_gpu_experimental_enable_nvshmem()) {
```
PatriosTheGreat, line 39: This flag is a module-wise. This means that if this flag is used all collectives (both NCCL and NVSHMEM) will use NVSHMEM allocator. Is this correct? If so should could you please add a unit test on this?

PatriosTheGreat (changes requested):

xla/stream_executor/cuda/cuda_collectives.cc
nvcastet: yes It will use nvshmem_malloc for both NCCL and NVSHMEM use cases for registered buffer memory since they are compatible.
Is the new unit test in `se_gpu_pjrt_client_test.cc` not sufficient?

PatriosTheGreat: There are two problems here:

1. xla::GetDebugOptionsFromFlags() basically goes to XLA_FLAGS environment variable, so this check is not even module-wise it's a run-wise. This is harder to test and also setting ` options.executable_build_options.mutable_debug_options()->set_xla_gpu_experimental_enable_nvshmem(true);` like it's done on `se_gpu_pjrt_client_test` wouldn't have any affect here.
2. We might want to dynamically chose which collection to emit (nccl or nvshmem) and using a new memory allocation mechanism (nvshmem allocate instead of ncclMemAlloc) might be a source of a new issues.

I see two possible solutions here (but maybe @beckerhe @ezhulenev might know a better way):
1. We can add a separate memory space for an NVSHMEM memory. Instead of using kCollectiveMemorySpaceColor we can use kCollectiveNvshmemMemorySpaceColor and has another cuda_nvshmem_collectives class which will use nvshmem allocator. We would then need to create yet another BFC allocator in the `se_gpu_pjrt_client`
2. If we are sure that using nvshmem allocator for NCCL collectives is secure and we are ready to accept the fact that `xla_gpu_experimental_enable_nvshmem` will switch all collectives allocations to use nvshmem allocator then we need to cover this test case in the `xla/service/collective_ops_utils.cc` tests for all-reduces all-gathers and so on (add a separate test-case for this cases which will flip `xla_gpu_experimental_enable_nvshmem` flag and make sure that everything still works fine).

PatriosTheGreat (approved):

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1594,6 +1594,81 @@ TEST(StreamExecutorGpuClientTest,
...
+          api_version=API_VERSION_TYPED_FFI,
+          backend_config={"custom_call_backend_config": {"attributes": "{module = \"nvshmem\"}"}}
+    })";
+  // Nvshmem requires one gpu per process.
+  GpuClientOptions client_options;
+  client_options.node_id = 0;
+  client_options.allowed_devices = {0};
+  client_options.num_nodes = 1;
+  client_options.kv_store = std::make_shared<InMemoryKeyValueStore>();
+  TF_ASSERT_OK_AND_ASSIGN(auto client,
```
toli-y, line 1634: please use explicit types: https://google.github.io/styleguide/cppguide.html#Type_deduction

xla/stream_executor/cuda/cuda_collectives.cc
trevor-m: Hi @PatriosTheGreat, thanks for reviewing.

Memory allocated by nvshmem_allocate() can be used for NCCL user buffers. We think that sharing the memory space is better than adding a new space and allocator because it won't cause as much fragmentation. Managing two memory pools is already a challenge and adding a third would be even worse. 

I can add a test case for nccl buffers + nvshmem memory. It also sounds like I need to change how the `set_xla_gpu_experimental_enable_nvshmem` flag is checked - looks like it needs to be read from the moduel and plumbed down properly. Let me try this.

trevor-m: Hi @PatriosTheGreat I've added some unit tests to `xla/tests/collective_ops_e2e_test.cc` to test the different combinations of nccl user buffers and nvshmem allocation.

For the flag check, I had some thoughts. I think this setting has to be client-wise or run-wise rather than module-wise. Since the allocator is created by the client during initialization before any modules are being ran, we cannot read the flag from the modules. Would it then be sufficient to just change the tests to set XLA_FLAGS?

xla/xla.proto
```diff
@@ -458,6 +458,9 @@ message DebugOptions {
 
   bool xla_gpu_enable_pipelined_reduce_scatter = 231;
 
+  // Enable NVSHMEM.
+  bool xla_gpu_experimental_enable_nvshmem = 379;
```
PatriosTheGreat, line 462: Cloud you please add a comment here that this flag can be set only via XLA_FLAGS env variable and can't be set through HLO Config->ExecutionOptions

PatriosTheGreat (changes requested):

PatriosTheGreat (approved):

penpornk (approved):

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1248,6 +1249,15 @@ absl::StatusOr<DeviceTopologyPair> BuildDistributedDevices(
   }
   gpu_executable_run_options->set_gpu_global_device_ids(
       std::move(gpu_device_ids));
+#if GOOGLE_CUDA
+  static const bool xla_gpu_experimental_enable_nvshmem =
```
ezhulenev, line 1253: This should not be here:
1. No new GOOGLE_CUDA when it can be avoided
2. Do not make nvshmem public and include it in SE (or anywhere outside for gpu/collectives folder)

This should go under GpuCollectives API and be implemented in NcclCollectives and NvsmemCollectives. The whole point of this API is so that we don't leak nvshmem/nccl/anything-else all over XLA:GPU codebase.

xla/backends/gpu/collectives/nvshmem_collectives_test.cc
```diff
@@ -90,24 +112,118 @@ absl::Status InitializationTestBody(const int node_id, const int num_nodes) {
...
+  TF_QCHECK_OK(distributed_client->Connect());
+  GpuClientOptions client_options;
+  client_options.node_id = node_id;
+  client_options.allowed_devices = {node_id};
+  client_options.num_nodes = num_nodes;
+  client_options.kv_store =
+      GetDistributedKeyValueStore(distributed_client, /*key_prefix=*/"gpu:");
+  ;
+  TF_ASSIGN_OR_RETURN(std::unique_ptr<PjRtClient> client,
+                      GetStreamExecutorGpuClient(client_options));
```
PatriosTheGreat, line 160: This API should not be used by the new targets. Either us here `xla/pjrt/c/...` interfaces, or move this to se_gpu_pjrt_client_test or create another test based on `xla/tests/collective_ops_e2e_test.cc` like collective_ops_e2e_nvshmem_test.

PatriosTheGreat (changes requested):

xla/pjrt/gpu/se_gpu_pjrt_client.cc
trevor-m: This particular code is specific to nvshmem since it requires a global initialization, so there wouldn't be any equivalent abstraction on the nccl side.
However, I can hide it behind something like:
`CollectivesRegistry::Get("gpu", "nvshmem").SetEnvInfo(...)`
Would that work?

xla/backends/gpu/collectives/nvshmem_collectives_test.cc
trevor-m: Originally I had these tests based on the collectives_ops_e2e_test, but unfortunately that only seems to support single progress multiple gpu. For nvshmem, we need 1 process for GPU. 

I can move this to se_gpu_pjrt_client_test

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1676,6 +1676,88 @@ TEST(StreamExecutorGpuClientTest,
...
+                           .Attr<absl::string_view>("module"));
+
+XLA_FFI_REGISTER_HANDLER(ffi::GetXlaFfiApi(), "mosaic_gpu",
+                         PlatformUtil::CanonicalPlatformName("GPU").value(),
+                         kMockMosaicGpu);
+
+// Verify that the client can initialize NVSHMEM and that buffers used by
+// mosaic_gpu custom calls are assigned to the collective memory space.
+TEST(StreamExecutorGpuClientTest, NvshmemMemoryTest) {
+  tsl::setenv("XLA_FLAGS", "--xla_gpu_experimental_enable_nvshmem=true",
```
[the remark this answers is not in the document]
trevor-m: Do you mean to create a new file ex: `se_gpu_pjrt_client_nvshmem_test.cc` just for this test?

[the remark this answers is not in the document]
PatriosTheGreat: Yes, otherwise xla_flags_reset breaks other test-cases in this test

xla/backends/gpu/collectives/nvshmem_collectives_test.cc
PatriosTheGreat: Yes that would also work

trevor-m:
@PatriosTheGreat @ezhulenev I believe I've addressed the remaining comments. Could you please take a look?

frgossen:
@PatriosTheGreat is looking into importing the PR and addressing the issues

PatriosTheGreat:
I'm still working on onboarding this PR. One of code changes for this PR caused presubmits failure. Looking for the way to fix it. Hopefully will complete everything in the next few days.

trevor-m:
> I'm still working on onboarding this PR. One of code changes for this PR caused presubmits failure. Looking for the way to fix it. Hopefully will complete everything in the next few days.

Thank you @PatriosTheGreat let me know if there's anything I can do to help.

PatriosTheGreat:
I rolled back this PR since it broke several internal tests. Working of re-submitting this.

trevor-m:
@PatriosTheGreat Let us know if there's any way we can help

PatriosTheGreat:
This change was reland with https://github.com/openxla/xla/pull/25389

PR #21687, opened 2025-01-22
Fix executable creation in HloRunnerPjRt

mmakevic-amd:
The order of evaluating function arguments is unspecified ([ref](https://en.cppreference.com/w/cpp/language/eval_order)), so pjrt_executable would get released first, leading to seg faults.

This caused the following test failures:
- all_reduce_test
- broadcast_test
- gather_operation_test
- copy_test
- replicated_io_feed_test

### DIFF
--- a/xla/service/hlo_runner_pjrt.cc
+++ b/xla/service/hlo_runner_pjrt.cc
@@ -169,6 +169,13 @@ class PjRtWrappedExecutable : public Executable {
       : Executable(hlo_module),
         pjrt_loaded_executable_(pjrt_loaded_executable) {}
 
+  explicit PjRtWrappedExecutable(std::shared_ptr<HloModule> hlo_module,
+                                 std::unique_ptr<PjRtLoadedExecutable>& pjrt_loaded_executable)
+      : Executable(hlo_module)
+      {
+        pjrt_loaded_executable_ = std::move(pjrt_loaded_executable);
+      }
+
   absl::StatusOr<ExecutionOutput> ExecuteAsyncOnStream(
       const ServiceExecutableRunOptions* run_options,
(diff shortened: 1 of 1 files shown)


mmakevic-amd:
@xla-rotation can you take a look, please?

xla/service/hlo_runner_pjrt.cc
```diff
@@ -169,6 +169,13 @@ class PjRtWrappedExecutable : public Executable {
       : Executable(hlo_module),
         pjrt_loaded_executable_(pjrt_loaded_executable) {}
 
+  explicit PjRtWrappedExecutable(std::shared_ptr<HloModule> hlo_module,
+                                 std::unique_ptr<PjRtLoadedExecutable>& pjrt_loaded_executable)
```
reedwm, line 173: I would pass the unique_ptr by value, not reference, and set `pjrt_loaded_executable_` in the member initializer list.

```c++
  explicit PjRtWrappedExecutable(std::shared_ptr<HloModule> hlo_module,
                                 std::unique_ptr<PjRtLoadedExecutable> pjrt_loaded_executable)
      : Executable(hlo_module),
        pjrt_loaded_executable_(std::move(pjrt_loaded_executable)) {}
```

This makes it very clear from the signature the constructor takes ownership of the pointer. You'll have to use std::move below when passing `pjrt_executable` to the constructor.

Also delete the other constructor since it's no longer used.

reedwm (changes requested):

xla/service/hlo_runner_pjrt.cc
mmakevic-amd: Fixed

reedwm (approved):

PR #21822, opened 2025-01-24
[XLA:GPU] Add support for SM100a architecture (Blackwell)

sergey-kozub:
Created `ShouldUsePtxExtension` helper for the extension suffix (this will also be used for sm120, etc).

CUDA 12.8 was recently released, which supports PTX 8.7, but that is not supported by the integrated LLVM (support added in https://github.com/llvm/llvm-project/pull/124155), so leaving the association with PTX 8.6 - this doesn't raise warnings during compilation.

### DIFF
--- a/xla/stream_executor/cuda/ptx_compiler_helpers.h
+++ b/xla/stream_executor/cuda/ptx_compiler_helpers.h
@@ -43,6 +43,15 @@ absl::Status CreateErrorFromPTXASLog(absl::string_view log,
 void WarnIfBadPtxasVersion(absl::string_view method,
                            const CudaComputeCapability& cc,
                            SemanticVersion compiler_version);
+
+// Determine whether the PTX extension for a compute capability should be used.
+//
+// Returns true if the argument compute capability has PTX extensions that are
+// only valid for that compute capability. For example, "sm_90" only includes
+// features that are forward compatible, whereas "sm_90a" (the extension) also
+// includes Hopper-specific features, such as WGMMA. We want to use the latter.
+bool ShouldUsePtxExtension(const CudaComputeCapability& cc);
+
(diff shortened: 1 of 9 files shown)


xla/stream_executor/cuda/ptx_compiler_helpers.h
```diff
@@ -43,6 +43,10 @@ absl::Status CreateErrorFromPTXASLog(absl::string_view log,
 void WarnIfBadPtxasVersion(absl::string_view method,
                            const CudaComputeCapability& cc,
                            SemanticVersion compiler_version);
+
+// Determine whether the PTX extension for a compute capability should be used.
```
reedwm, line 47: It's a bit unclear what "PTX extension" means (does this feature have an official name?) I would add a sentence like "E.g., returns true for compute capability 9.0 since we compile with architecture sm_90a, not sm_90, for such GPUs"

reedwm (changes requested):

xla/stream_executor/cuda/ptx_compiler_helpers.h
sergey-kozub: Added a comment that hopefully explains it.

dimvar: The spec calls these "accelerated features", so we can use that name here.
https://docs.nvidia.com/cuda/parallel-thread-execution

reedwm: I'm fine with either the current terminology or "accelerated features"

nouiz:
@reedwm Can you review again?

reedwm (approved):

PR #21825, opened 2025-01-24
Exclude the usage of CPU memory from the GPU memory scheduler

zhenying-liu:
In the MaxText optimizer state offloading, we observed no memory savings when switching from f16 to f32. The root cause is that the GPU memory scheduler does not distinguish between CPU memory and GPU memory. This commit modifies the scheduler to exclude CPU memory.

### DIFF
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -714,14 +714,14 @@ absl::StatusOr<ScheduleMetadata> ScheduleGpuModule(
 absl::StatusOr<HloSchedule> ScheduleGpuModuleWithMemoryScheduler(
     const HloModule* module, int64_t pointer_size, int64_t* peak_memory_bytes) {
   BufferValue::SizeFunction size_func =
-      [pointer_size](const BufferValue& buffer) {
-        const Shape& shape = buffer.shape();
-        if (shape.has_layout() &&
-            shape.layout().memory_space() == Layout::kHostMemorySpace) {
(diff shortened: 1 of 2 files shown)


xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -671,7 +671,12 @@ absl::StatusOr<HloSchedule> ScheduleGpuModuleWithMemoryScheduler(
     const HloModule* module, int64_t pointer_size, int64_t* peak_memory_bytes) {
   BufferValue::SizeFunction size_func =
       [pointer_size](const BufferValue& buffer) {
-        return ShapeUtil::ByteSizeOf(buffer.shape(), pointer_size);
+        const Shape& shape = buffer.shape();
+        if (shape.has_layout() &&
+            shape.layout().memory_space() == Layout::kHostMemorySpace) {
+          return static_cast<int64_t>(0);
```
frgossen, line 677: Would annotating the closure work here? `-> int64_t`

frgossen (approved):

xla/service/gpu/gpu_hlo_schedule.cc
frgossen: Yes, I think that makes sense. It's a minor thing, I know..

frgossen: Let me know when I should start the import

zhenying-liu: Great. Thanks!

frgossen (approved):

PR #21885, opened 2025-01-27
Introduce env var to filter trace entries in Python profiling

petergtz:
**The problem:** As described in https://github.com/jax-ml/jax/issues/21295, by default, the profiler will stop collecting trace events when there's more than 1 million events. This can be overridden by `TF_PROFILER_TRACE_VIEWER_MAX_EVENTS`, but effectively this will quickly result in trace files of multiple GBs in size. Perfetto handles them very slowly or can't handle them at all.

**Proposed solution:** This commit introduces `XLA_PYTHON_TRACER_MIN_EVENT_DURATION_NS`. Setting it e.g. to 1000000 will ignore all Python-level entries that are shorter than 1ms. This reduces the number of Python-level entries significantly, so that trace file sizes can be kept at manageable sizes.

### DIFF
--- a/xla/backends/profiler/cpu/python_tracer_factory.cc
+++ b/xla/backends/profiler/cpu/python_tracer_factory.cc
@@ -13,6 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 #include <memory>
+#include <string>
+#include <stdlib.h>
 
 #include "xla/backends/profiler/cpu/python_tracer.h"
 #include "tsl/profiler/lib/profiler_factory.h"
@@ -28,6 +30,14 @@ std::unique_ptr<tsl::profiler::ProfilerInterface> CreatePythonTracer(
   PythonTracerOptions options;
   options.enable_trace_python_function = profile_options.python_tracer_level();
   options.enable_python_traceme = profile_options.host_tracer_level();
+
+  char* min_entry_duration_ns = getenv("XLA_PYTHON_TRACER_MIN_EVENT_DURATION_NS");
+  if (min_entry_duration_ns != nullptr) {
+    options.min_entry_duration_ns = std::stoull(min_entry_duration_ns, nullptr, 10);
+  } else {
+    options.min_entry_duration_ns = 0;
+  }
+
   return CreatePythonTracer(options);
 }
 
--- a/xla/backends/profiler/cpu/python_tracer.cc
+++ b/xla/backends/profiler/cpu/python_tracer.cc
@@ -94,6 +94,7 @@ std::unique_ptr<tsl::profiler::ProfilerInterface> CreatePythonTracer(
       options.enable_trace_python_function;
   pyhooks_options.enable_python_traceme = options.enable_python_traceme;
   pyhooks_options.end_to_end_mode = options.end_to_end_mode;
+  pyhooks_options.min_entry_duration_ns = options.min_entry_duration_ns;
   return std::make_unique<PythonTracer>(pyhooks_options);
 }
 
--- a/xla/backends/profiler/cpu/python_tracer.h
+++ b/xla/backends/profiler/cpu/python_tracer.h
@@ -15,9 +15,11 @@ limitations under the License.
 #ifndef XLA_BACKENDS_PROFILER_CPU_PYTHON_TRACER_H_
 #define XLA_BACKENDS_PROFILER_CPU_PYTHON_TRACER_H_
 
+#include <cstdint>
 #include <memory>
 
 #include "tsl/profiler/lib/profiler_interface.h"
+#include "tsl/platform/types.h"
 
 namespace xla {
 namespace profiler {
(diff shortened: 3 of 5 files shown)


xla/backends/profiler/cpu/python_tracer_factory.cc
```diff
@@ -28,6 +30,14 @@ std::unique_ptr<tsl::profiler::ProfilerInterface> CreatePythonTracer(
   PythonTracerOptions options;
   options.enable_trace_python_function = profile_options.python_tracer_level();
```
rengoog, line 31: just to understand - would adding a new variable in profiler_options work instead of setting the env variable here?

rengoog (commented):
Hi, 

Thanks for this PR, just a minor nit about using getenv vs adding to profiler_options.

Hi @cliveverghese could you ptal @ this as well?

cliveverghese:
> The problem: As described in https://github.com/jax-ml/jax/issues/21295, by default, the profiler will stop collecting trace events when there's more than 1 million events. This can be overridden by TF_PROFILER_TRACE_VIEWER_MAX_EVENTS, but effectively this will quickly result in trace files of multiple GBs in size. Perfetto handles them very slowly or can't handle them at all.

Please prefer to use trace viewer as part of tensorboard-profile-plugin. It does not have the truncation issues and can handle larger traces better. 

```
pip install tensorboard tensorboard-plugin-profile

tensorboard --port 6006 --logdir <logdir>
```

petergtz:
@cliveverghese Tensorboard is not a very practical solution in many environments. We often run trainings as part of test-automation and record traces for performance reasons. In those scenarios we prefer directly recording traces instead of the more complex setup with tensorboard.

petergtz:
@cliveverghese Can you help me understand: if I want to make test runs as part of an automation pipeline, record profiles of the test runs, and store them in some storage for some potential later analysis (e.g. for troubleshooting a training regression). How would I use the [ProfileData](https://github.com/openxla/xla/blob/9a023e7a3712fc7b4c197e974af800b2a7c5df66/xla/python/profiler/profile_data.pyi) API here? I haven't found any documentation for it.

xla/backends/profiler/cpu/python_tracer_factory.cc
petergtz: @rengoog  Yes, I think so. Two reasons I didn't do it:
1. I didn't realize until just now that `tensorflow::ProfileOptions` is actually defined in the xla package. I thought it's imported from tensorflow.
2. I didn't find how to actually set them. If you could point me to the right place, I'd be happy to make that change.

petergtz:
@cliveverghese Please also note that I do think it has quite some value to exclude short events below a threshold. It makes the work with a trace with JAX complexity much simpler conceptually, because – just as a simple example – the call stack contains only relevant events from a performance perspective (I would also be surprised if a 10GB file can be handled swiftly by tensorboard. It's **a lot** of objects in a UI.)

cliveverghese:
1. With regards to tensorboard Trace Viewer handling larger traces : Trace Viewer in tensorboard does not push all the events to the frontend, Instead it filters based on the viewport and zoom level. This allows it to handle profiles with a large number of events. 
2. When profiling completes, We write a *.xplane.pb file to the logdir provided by the user, Based on the use cases needed, This can either be accessed within the test, or later. This also contains all the events collected during the profiling duration, and are more compressed than the trace.json. To access this file, ProfileData accessor can be used.

xla/backends/profiler/cpu/python_tracer_factory.cc
rengoog: ah nevermind - PTAL @ @cliveverghese  's comments. As you pointed out, it may not be straightforward to plumb profiler_options .

dimitar-asenov:
@ezhulenev Could you please review this PR?

cliveverghese:
Hi, 

https://github.com/openxla/xla/pull/21885#issuecomment-2683278866 provides appropriate ways to access the profiles both programmatically or via UI.   

Using environment variables to configure the backends limits the flexibility of the system, such as the user needs not configure this before the workload is started, and on demand profiling (programmatic and remote capture) cannot update this once it's set. 

Please prefer methods as mentioned in the above comment.

cliveverghese:
Hi, 

The preferred approach to use the trace events would be to use the xplane.pb file that is generated instead of the perfetto file. This does not require tensorboard. 

However, If you still need this options, Please update [xla/tsl/profiler/convert/xplane_to_trace_events.cc ](https://github.com/openxla/xla/blob/cc4fbd8dd81274a9bbed8c34309944c30f5678f4/xla/tsl/profiler/convert/xplane_to_trace_events.cc#L4) instead of the python tracer, to skip the events in the perfetto output based on the environment variable.

petergtz:
@cliveverghese I will go with your suggestion to use xplane.pb file. It wasn't clear to me before that this file is equivalent to the json file. Since I don't see any downside to that right now, I will close this PR (would re-open if necessary).

cliveverghese:
Thank you @petergtz .

PR #21901, opened 2025-01-27
Ci add rocm6.1 deps for ubuntu 20.04

alekstheod:
Add rocm 6.1.0 dependency for ubuntu 20.04

### DIFF
--- a/third_party/tsl/third_party/gpus/rocm/BUILD.tpl
+++ b/third_party/tsl/third_party/gpus/rocm/BUILD.tpl
@@ -57,21 +57,25 @@ cc_library(
 )
 
 cc_library(
+    name = "rocm_headers_includes",
+    hdrs = glob([
+        "%{rocm_root}/include/**",
+    ]),
+    strip_include_prefix = "%{rocm_root}/include",
+)
+
+cc_library(
     name = "rocm_headers",
     hdrs = glob([
         "%{rocm_root}/include/**",
         "%{rocm_root}/lib/llvm/lib/**/*.h",
     ]),
     include_prefix = "rocm",
-    includes = [
-        "%{rocm_root}/include",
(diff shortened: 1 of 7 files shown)


third_party/tsl/third_party/gpus/rocm/BUILD.tpl
```diff
@@ -56,22 +56,26 @@ cc_library(
     }),
 )
 
+cc_library(
+    name = "rocm_headers_includes",
+    hdrs = glob([
+        "%{rocm_root}/include/**",
+    ]),
+    strip_include_prefix = "%{rocm_root}/include",
+)
```
beckerhe, line 65: Why this new target? Looks like you could just change `rocm_headers`, couldn't you?

alekstheod: You can't we actually need both, both include paths are used:
#include "include/{whateverlib}/whatever.h"
#include "whateverlib/whatever.h"
So hence a new target which will provided as a virtual include.
I think a clean solution would be to modify the include directives everywhere so they use "whateverlib/whatever.h" and strip out the include prefix.

beckerhe: Is it because ROCm 6.1.0 uses the new include paths?

because if both patterns are used in XLA then we should just fix it in XLA.

Otherwise it's probably fine as a temporary measure.

alekstheod: This seems to be a right approach to add additional include dirs. If we are adding them via includes we get bazel complaint that include file is not registered in the deps. So that way we always use virtual includes and everything is done through the deps declaration.

beckerhe (approved):

PR #21913, opened 2025-01-27
Add xla_experimental_ignore_channel_id flag to CSE pass

sfvaroglu:
When sequence parallelism is enabled, we are seeing duplicated all-gather ops in some models. 

Currently, CSE removes duplicate collectives if no `channel_id` is set. However, it considers them as unique ops if `channel_id` is set and does not replace them. 

To eliminate duplicates and improve performance, this PR:
- adds logic to use `IdenticalIgnoringChannelIdValues` to check op equality in CSE
- checks `xla_experimental_ignore_channel_id` flag along with HasSideEffects() in CSE

### DIFF
--- a/xla/service/BUILD
+++ b/xla/service/BUILD
@@ -4160,6 +4160,7 @@ cc_library(
         "//xla:shape_util",
         "//xla/hlo/ir:hlo",
         "//xla/hlo/pass:hlo_pass",
+        "//xla/service:collective_ops_utils",
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/container:flat_hash_set",
         "@com_google_absl//absl/status:statusor",
--- a/xla/service/hlo_cse.cc
+++ b/xla/service/hlo_cse.cc
@@ -29,6 +29,7 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_instructions.h"
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/literal.h"
+#include "xla/service/collective_ops_utils.h"
 #include "xla/service/hlo_domain_map.h"
 #include "xla/shape_util.h"
 #include "tsl/platform/errors.h"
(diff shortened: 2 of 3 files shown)


sfvaroglu:
cc @frgossen @Tixxx

akuegel:
I am not familiar with this and defer to @frgossen for review.
Just a general comment: is it really needed to make a flag for this? If we already know of a case where this change would cause issues, then maybe the general idea is wrong that we can deduplicate ops with different channel_id flags?
I quickly checked where `IdenticalIgnoringChannelIdValues()` is used, and it is used in the pass ScheduleAwareCollectiveOpsCSE: https://github.com/openxla/xla/blob/main/xla/service/spmd/schedule_aware_collective_ops_cse.h
Maybe a better fix would be to add this pass to the GPU pipeline if possible?

frgossen:
Can we try landing this w/o a flag to see if/where the uniqueness of the channel ID is necessary?

sfvaroglu:
> Can we try landing this w/o a flag to see if/where the uniqueness of the channel ID is necessary?

We would like to enable this flag by default (see https://github.com/openxla/xla/pull/22041). We could land that first and then re-run CI for this PR. How does that sound?

akuegel:
@sfvaroglu What about using the pre-existing ScheduleAwareCollectiveOpsCSE pass, did you try whether it does what you want?

sfvaroglu:
> @sfvaroglu What about using the pre-existing ScheduleAwareCollectiveOpsCSE pass, did you try whether it does what you want?

@akuegel Is this enabled by default? If so, it didn't help with our case.

akuegel:
> > @sfvaroglu What about using the pre-existing ScheduleAwareCollectiveOpsCSE pass, did you try whether it does what you want?
> 
> @akuegel Is this enabled by default? If so, it didn't help with our case.

No, it is not enabled. As I said in my previous comment, my suggestion is to add it to the GPU pipeline, currently we don't use this pass. But it looks to me that it could do what you want. Please try it out.

sfvaroglu:
> > > @sfvaroglu What about using the pre-existing ScheduleAwareCollectiveOpsCSE pass, did you try whether it does what you want?
> > 
> > 
> > @akuegel Is this enabled by default? If so, it didn't help with our case.
> 
> No, it is not enabled. As I said in my previous comment, my suggestion is to add it to the GPU pipeline, currently we don't use this pass. But it looks to me that it could do what you want. Please try it out.

Yes, it helps. I'll close this PR and create a new one to add the pass to the GPU pipeline. Thanks for the suggestion.

PR #21966, opened 2025-01-28
[GPU] Stop collective combiners from creating scheduling footguns

yliu120:
The collective combiners in GPU works incorrectly in many points:

1. When user sets the size flag, the old logic combining the pipelined collectives stops working and starts to create footguns like merging pipelined collectives with non-pipelined collectives, which totally blows up the point of pipelining collectives.
2. Don't run two pass. just pass the is_pipelined flag to the combined key. running two passes is buggy because remnants coming from the first pass will start to be merged with non-pipelined collectives in the second pass.
3. Always respect user's flag for both pipelined and non-pipelined cases if it doesn't blow up memory.
4. Don't drop metadata and backend_configs even we combine instructions. We need to fix the interface for that in long term.

### DIFF
--- a/xla/service/gpu/all_gather_combiner.cc
+++ b/xla/service/gpu/all_gather_combiner.cc
@@ -72,10 +72,10 @@ absl::StatusOr<bool> GpuAllGatherCombiner::Run(
   // Always respects the threshold users set if it doesn't increase
   // memory pressure.
   int64_t previous_combiner_threshold = combine_threshold_in_bytes_;
-  combine_threshold_in_bytes_ = std::min(
-      ComputeSuggestedCombinerThreshold(
-          *module, device_info_, HloOpcode::kReduceScatter, pointer_size_),
-      previous_combiner_threshold);
+  combine_threshold_in_bytes_ =
+      std::min(ComputeSuggestedCombinerThreshold(
+                   *module, device_info_, HloOpcode::kAllGather, pointer_size_),
+               previous_combiner_threshold);
   TF_ASSIGN_OR_RETURN(
       bool combined_pipelined_instructions,
       RunWithKeyCombiner(module, execution_threads, PipelinedCombinerKey));
--- a/xla/service/gpu/all_gather_combiner.cc
+++ b/xla/service/gpu/all_gather_combiner.cc
@@ -41,9 +41,19 @@ std::optional<AllGatherCombiner::GroupKey> PipelinedCombinerKey(
   if (backend_config.ok()) {
     is_pipelined = backend_config->collective_backend_config().is_pipelined();
   }
-  return AllGatherCombiner::CombineKey(
-      instruction, domain_map, combine_by_dim, combine_different_dtypes,
-      is_pipelined ? "pipelined" : "unpipelined");
+  auto key = AllGatherCombiner::CombineKey(
+      instruction, domain_map, combine_by_dim, combine_different_dtypes);
+  if (!key.has_value()) {
+    return std::nullopt;
+  }
+  return AllGatherCombiner::GroupKey{
+      std::get<0>(key.value()),
+      std::get<1>(key.value()),
+      std::get<2>(key.value()),
+      std::get<3>(key.value()),
+      std::get<4>(key.value()),
+      std::move(std::get<5>(key.value())),
+      is_pipelined ? "pipelined" : "non-pipelined"};
 }
 
 }  // namespace
--- a/xla/service/gpu/all_gather_combiner.cc
+++ b/xla/service/gpu/all_gather_combiner.cc
@@ -52,11 +52,6 @@ std::optional<AllGatherCombiner::GroupKey> PipelinedCombinerKey(
 absl::StatusOr<bool> GpuAllGatherCombiner::Run(
     HloModule* module,
     const absl::flat_hash_set<absl::string_view>& execution_threads) {
-  // Combiner threshold is specified. Running parent pass code.
-  if (combine_threshold_in_bytes_ != default_combine_threshold_in_bytes_) {
-    return AllGatherCombiner::Run(module, execution_threads);
-  }
-
   // If there are no pipelined instructions in the IR, the optimizations below
   // do not kick in anyway.
   // Exit early so we do not perform expensive scheduling dry run below.
@@ -65,9 +60,13 @@ absl::StatusOr<bool> GpuAllGatherCombiner::Run(
   }
 
   // Combine as much as possible for pipelined collectives.
-  int previous_combiner_threshold = combine_threshold_in_bytes_;
-  combine_threshold_in_bytes_ = ComputeSuggestedCombinerThreshold(
-      *module, device_info_, HloOpcode::kAllGather, pointer_size_);
+  // Always respects the threshold users set if it doesn't increase
+  // memory pressure.
+  int64_t previous_combiner_threshold = combine_threshold_in_bytes_;
+  combine_threshold_in_bytes_ = std::min(
+      ComputeSuggestedCombinerThreshold(
+          *module, device_info_, HloOpcode::kReduceScatter, pointer_size_),
+      previous_combiner_threshold);
   TF_ASSIGN_OR_RETURN(
       bool combined_pipelined_instructions,
       RunWithKeyCombiner(module, execution_threads, PipelinedCombinerKey));
--- a/xla/service/gpu/all_gather_combiner.cc
+++ b/xla/service/gpu/all_gather_combiner.cc
@@ -36,15 +36,14 @@ namespace {
 std::optional<AllGatherCombiner::GroupKey> PipelinedCombinerKey(
     const HloInstruction* instruction, const HloDomainMap& domain_map,
     bool combine_by_dim, bool combine_different_dtypes) {
+  bool is_pipelined = false;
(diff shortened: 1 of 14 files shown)


golechwierowicz:
Hey, thanks for the diff!
Current implementation is WAI, but I'm willing to change the semantics if you empirically find it more useable.

Re 1, 2: From what I get from this diff, you explicitly want to disallow combining pipelined collectives with non-pipelined collectives. Do you have a mini HLO example of why is this is a problem?

Re 3: With your diff if a user provides a value which is higher than available peak memory then available peak mem will be chosen. Is this what you want? I mean, user explicitly provided some value but the combiner will combine more (and likely crash).

Re 4: This should be fixed already with https://github.com/openxla/xla/pull/21718/files.

xla/service/gpu/all_gather_combiner.cc
```diff
@@ -65,19 +69,17 @@ absl::StatusOr<bool> GpuAllGatherCombiner::Run(
...
   // Combine as much as possible for pipelined collectives.
-  int previous_combiner_threshold = combine_threshold_in_bytes_;
-  combine_threshold_in_bytes_ = ComputeSuggestedCombinerThreshold(
-      *module, device_info_, HloOpcode::kAllGather, pointer_size_);
+  // Always respects the threshold users set if it doesn't increase
+  // memory pressure.
+  int64_t previous_combiner_threshold = combine_threshold_in_bytes_;
+  combine_threshold_in_bytes_ = std::min(
+      ComputeSuggestedCombinerThreshold(
+          *module, device_info_, HloOpcode::kReduceScatter, pointer_size_),
```
golechwierowicz, line 77: s/kReduceScatter/kAllGather.

xla/service/gpu/all_reduce_combiner.cc
```diff
@@ -63,19 +61,18 @@ absl::StatusOr<bool> GpuAllReduceCombiner::Run(
...
   // Combine as much as possible for pipelined collectives.
-  int previous_combiner_threshold = combine_threshold_in_bytes_;
-  combine_threshold_in_bytes_ = ComputeSuggestedCombinerThreshold(
-      *module, device_info_, HloOpcode::kAllReduce, pointer_size_);
+  // Always respects the threshold users set if it doesn't increase
+  // memory pressure.
+  int64_t previous_combiner_threshold = combine_threshold_in_bytes_;
+  combine_threshold_in_bytes_ = std::min(
+      ComputeSuggestedCombinerThreshold(
+          *module, device_info_, HloOpcode::kReduceScatter, pointer_size_),
```
golechwierowicz, line 69: s/kReduceScatter/kAllReduce.

yliu120:
> Hey, thanks for the diff! Current implementation at head is WAI, but I'm willing to change the semantics if you empirically find it more useable. The core idea of the current pipelined/non-pipelined logic is that if the user did not provide any specified value then we will combine pipelined collectives up to a peak mem, and non-pipelined we'll just bucket up to a default threshold. If the user provided the combiner value then we will always respect it, regardless of pipelined/non-pipelined distinction.
> 
> Re 1, 2: From what I get from this diff, you explicitly want to disallow combining pipelined collectives with non-pipelined collectives. Do you have a mini HLO example of why is this is a problem?
> 
> Re 3: With your diff if a user provides a value which is higher than available peak memory then available peak mem will be chosen. Is this what you want? I mean, user explicitly provided some value but the combiner will combine less.
> 
> Re 4: This should be fixed already with https://github.com/openxla/xla/pull/21718/files.
> 
> @allanrenucci also pointed out that this PR makes combiners no longer combine pipelined collectives up to a peak available memory but to min(user_specified_threshold, peak_available_mem). Is this change on your side intentional?

Thanks for the review.

Yes, we do want to treat pipelined collectives different as non-pipelined. There is no need to have any HLO to explain.
The collectives are pipelined because we want to *free* them from dependencies inside a scan layer, which means that after pipelining, they have the entire freedom of being schedule with almost any compute in the previous/after iteration. This is how we achieve maximal overlaps.

Having the combined with `non-pipelined` collectives is basically just locking them again with data dependencies which makes scheduling restricted. Having a release-and-then-lock is a bit awkward.

For memory thing, we don't notice any memory pressure actually from a practical perspective, actually I don't worry too much about that part and just simply respect what the system might tell. If the scheduler thinks the value users provide will OOM, we just respect that.

https://github.com/openxla/xla/pull/21718/files doesn't pass backend_config and for metadata is not totally correct. I am wondering whether all combined collectives come from the same source line.

golechwierowicz:
Thanks for the explanation. I agree combining pipelined and non-pipelined collectives restricts instruction set which can be potentially overlapped. We just did not see this issue in practice, that's why given your experience I'm very happy to change the semantics :)

Given pipelined collectives can be essentially overlapped with almost any compute inside the while loop body computation it makes sense to combine them maximally to fully utilize networking bandwidth, right? Can you add a piece of logic which does that in the absence of user provided combiner threshold? I'll LGTM after that.

yliu120:
> Thanks for the explanation. I agree combining pipelined and non-pipelined collectives restricts instruction set which can be potentially overlapped. We just did not see this issue in practice, that's why given your experience I'm very happy to change the semantics :)
> 
> Given pipelined collectives can be essentially overlapped with almost any compute inside the while loop body computation it makes sense to combine them maximally to fully utilize networking bandwidth, right? Can you add a piece of logic which does that in the absence of user provided combiner threshold? I'll LGTM after that.

Thanks. After I am freed up by some other things, I will fix the thing you mentioned, which makes lots of sense to me.

yliu120:
> Thanks for the explanation. I agree combining pipelined and non-pipelined collectives restricts instruction set which can be potentially overlapped. We just did not see this issue in practice, that's why given your experience I'm very happy to change the semantics :)
> 
> Given pipelined collectives can be essentially overlapped with almost any compute inside the while loop body computation it makes sense to combine them maximally to fully utilize networking bandwidth, right? Can you add a piece of logic which does that in the absence of user provided combiner threshold? I'll LGTM after that.

After a bit rethinking, "combining pipelined collectives maximally" will not lead to the optimal performance. Let's say you are combining 6 all-gathers and it leads to a big one. In a lot of practical cases, we probably won't have a big enough chunks of computes to overlap it. Breaking it into pieces (2 all-gathers instead of 1) might yield good performance in practice. Usually I need to tune the param to balance between `network bandwidth` and `possibility of overlaps`.

I guess in the current impl, for both pipelined and non-pipelined cases, we respect the default size if users don't set it. In essence, there should be two flags to control these behaviors.

yliu120:
> > Thanks for the explanation. I agree combining pipelined and non-pipelined collectives restricts instruction set which can be potentially overlapped. We just did not see this issue in practice, that's why given your experience I'm very happy to change the semantics :)
> > Given pipelined collectives can be essentially overlapped with almost any compute inside the while loop body computation it makes sense to combine them maximally to fully utilize networking bandwidth, right? Can you add a piece of logic which does that in the absence of user provided combiner threshold? I'll LGTM after that.
> 
> After a bit rethinking, "combining pipelined collectives maximally" will not lead to the optimal performance. Let's say you are combining 6 all-gathers and it leads to a big one. In a lot of practical cases, we probably won't have a big enough chunks of computes to overlap it. Breaking it into pieces (2 all-gathers instead of 1) might yield good performance in practice. Usually I need to tune the param to balance between `network bandwidth` and `possibility of overlaps`.
> 
> I guess in the current impl, for both pipelined and non-pipelined cases, we respect the default size if users don't set it. In essence, there should be two flags to control these behaviors.

Let me provide some simple illustration.

-- compute -- compute -- barrier-1 -- compute -- barrier 2 --
--------------combined-all-gathers------ (elapsed time longer than any of the chunks)

Then it is better to do this,
-- compute -- compute -- barrier-1 -- compute -- barrier 2 --
---combined-all-gathers1-barrier-1 -- all-gather-2-- barrier-2 --

yliu120:
Let me experiment more and I will get a sense.
I feel like most all gather/ reduce scatter are pipelined all gather reduce scatters from FSDP.
Then in that case, we can always set the flag to larger values to achieve the same effect as you mentioned.

golechwierowicz:
> Let me experiment more and I will get a sense. I feel like most all gather/ reduce scatter are pipelined all gather reduce scatters from FSDP. Then in that case, we can always set the flag to larger values to achieve the same effect as you mentioned.

Yes, but the point is at Google we are spending a lot of effort for compiler to have good performance by default w/o any flag tuning. For pipelined collectives, assuming we pipelined trivial formatting ops (bitcasts, get tuple elements, light compute, etc.) then every such collective can be combined to maximize network bandwidth.

The way I think of this is the following. Every collective has a set S of compute which can be overlapped with it. Combining collectives produces a new collective with a new S' which is produced by intersecting every combined S. In case of pipelined collectives S' = S. So you will not lose any compute-communication overlap => it is safe to combine everything.

dimitar-asenov:
@yliu120 If you still intend to merge this PR, please provide an update. Otherwise please close the PR. It will be automatically closed after about one month of inactivity.

golechwierowicz:
@yliu120, your use case should be solved in https://github.com/tensorflow/tensorflow/pull/93846/files.

dimitar-asenov:
Closing this due to inactivity. Please feel free to reopen in case https://github.com/tensorflow/tensorflow/pull/93846 does not solve the issue.

PR #21998, opened 2025-01-29
Add kCall to GetFusibleComputations

chaserileyroberts:
Previously, we had issues with the stream annotation being applied to non-Gemm operations, and having the compilation fail at IR emitter stage. Operations that normally should've been fused weren't getting fused inside the async computation. This PR is intended to fix this issue by including the call computations when searching for fusible computations.

This was suggested by @jreiffers

### DIFF
--- a/xla/service/gpu/gpu_fusible.cc
+++ b/xla/service/gpu/gpu_fusible.cc
@@ -44,6 +44,7 @@ limitations under the License.
 #include "xla/service/instruction_fusion.h"
 #include "xla/shape.h"
 #include "xla/shape_util.h"
+#include "xla/side_effect_util.h"
 #include "xla/stream_executor/device_description.h"
 #include "xla/util.h"
 
@@ -847,6 +848,11 @@ bool MayPreventVectorization(const HloFusionAdaptor& fusion) {
   });
 }
 
+bool IsStreamAnnotatedComputation(const HloInstruction* instr) {
+  return instr->opcode() == HloOpcode::kCall &&
+         instr->frontend_attributes().map().contains(kXlaStreamAnnotationAttr);
+}
+
 std::vector<HloComputation*> GetFusibleComputations(
     const HloModule& module,
     const absl::flat_hash_set<absl::string_view>& execution_threads) {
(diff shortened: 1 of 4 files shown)


jreiffers:
This affects priority fusion, so we'll probably want a test in [priority_fusion_test.cc](https://github.com/openxla/xla/blob/8e6b84bd729a1d079ed5035f1ba8afc9d205587e/xla/service/gpu/transforms/priority_fusion_test.cc) as well.

xla/service/gpu/gpu_fusible.cc
```diff
@@ -858,6 +858,7 @@ std::vector<HloComputation*> GetFusibleComputations(
       // flow. See also fusion_wrapper.cc, which does the same.
       if (HloInstruction::MightHaveCalledComputations(instr->opcode()) &&
           instr->opcode() != HloOpcode::kWhile &&
+          instr->opcode() != HloOpcode::kCall &&
```
jreiffers, line 861: I'm not sure we want to fuse inside all calls. We might want to just enable the async ones you care about for now.

chaserileyroberts: I broke out this check, now it only applies to calls with stream annotations.

jreiffers (approved):

olegshyshkov:
This change LGTM. 

One note about the original motivation:

> and having the compilation fail at IR emitter stage. Operations that normally should've been fused weren't getting fused inside the async computation.

Not running PriorityFusion on a computation shouldn't fail in IR emitter, it would only result in degraded performance. You probably also need to do the same change (check annotated kCall computations) in [fusion_wrapper.cc](https://github.com/openxla/xla/blob/120ab113d6f65c440f1bf4ae1bb9e786ff0d88e4/xla/service/gpu/transforms/fusion_wrapper.cc#L42). The perfect solution would be to refactor `fusion_wrapper` to use `GetFusibleComputations`.

olegshyshkov (approved):

jreiffers:
> This change LGTM.
> 
> One note about the original motivation:
> 
> > and having the compilation fail at IR emitter stage. Operations that normally should've been fused weren't getting fused inside the async computation.
> 
> Not running PriorityFusion on a computation shouldn't fail in IR emitter, it would only result in degraded performance. You probably also need to do the same change (check annotated kCall computations) in [fusion_wrapper.cc](https://github.com/openxla/xla/blob/120ab113d6f65c440f1bf4ae1bb9e786ff0d88e4/xla/service/gpu/transforms/fusion_wrapper.cc#L42). The perfect solution would be to refactor `fusion_wrapper` to use `GetFusibleComputations`.

Oh, great point, I totally missed that. We'll send a followup.

chaserileyroberts:
> The perfect solution would be to refactor fusion_wrapper to use GetFusibleComputations.

I initially tried doing this, but there is a test in fusion_wrapper_test.cc that is causing issues, `WhileInFusion`. This tests that while ops within fusions do not have their computations wrapped. But, `GetFusibleComputations` currently always returns these computations, so this test ends up failing. :/

PR #22075, opened 2025-01-30
[ROCm] Support collective memory allocator for rocm

alekstheod:
This pr is adding support for a collective memory allocator in rocm build as it was implemented already for cuda.


beckerhe (commented):
Hey, a few general things. We are actively trying to get rid of all the #ifdefs in the code base. That means we can't add the support as you implemented it. Rocm and Cuda need to be separate implementations. If they share some common code, then you could introduce a base class `Collectives`, but we won't get around having `CudaCollectives` and `RocmCollectives` as separate classes in separate files.

alekstheod:
> Hey, a few general things. We are actively trying to get rid of all the #ifdefs in the code base. That means we can't add the support as you implemented it. Rocm and Cuda need to be separate implementations. If they share some common code, then you could introduce a base class `Collectives`, but we won't get around having `CudaCollectives` and `RocmCollectives` as separate classes in separate files.

Ok. What about the place I put that common code? Shall it be in gpu subdirectory or steram_executor is fine?

beckerhe:
> > Hey, a few general things. We are actively trying to get rid of all the #ifdefs in the code base. That means we can't add the support as you implemented it. Rocm and Cuda need to be separate implementations. If they share some common code, then you could introduce a base class `Collectives`, but we won't get around having `CudaCollectives` and `RocmCollectives` as separate classes in separate files.
> 
> Ok. What about the place I put that common code? Shall it be in gpu subdirectory or steram_executor is fine?

gpu subdirectory is better

alekstheod:
Abandon this PR in a favor of simpler solution.

PR #22091, opened 2025-01-30
[ROCM] Try targeting cuBLAS if it's not profitable to fuse dot

amd-jianli12:
TritonTest.NonstandardLayoutWithManyNonContractingDims
TritonTest.NonstandardLayoutWithManyNonContractingDimsReversedLayout

If it's not profitable to fuse dot and cuBLAS is not requiring extra padding then targeting custom call would be optimal. Align this check for both ROCm and CUDA here.
 
@xla-rotation would you please have a look?


mooskagh (approved):

golechwierowicz:
Can you add a test?

amd-jianli12:
> Can you add a test?

Added a test to check GemmFusion pass is not fusing nonprofitable dot. Please kindly have a look. @golechwierowicz

mooskagh (approved):

PR #22101, opened 2025-01-30
Define SO_ZEROCOPY, SO_EE_ORIGIN_ZEROCOPY when missing (e.g. in conda sysroot)

### LINKED ISSUE #22083: Failure in building on Ubuntu 20.04: undeclared SO_ZEROCOPY
Building JAX on Ubuntu 20.04.6 LTS (Linux 5.4.0-171-generic) leads to the following compilation error:
```
external/xla/xla/python/transfer/socket-server.cc:366:44: error: use of undeclared identifier 'SO_ZEROCOPY'
  366 |   CHECK_GE(setsockopt(send_fd, SOL_SOCKET, SO_ZEROCOPY, &value, sizeof(value)),
```

`SO_ZEROCOPY` dependency to XLA was introduced in https://github.com/openxla/xla/pull/21496.

As a temporary workaround, applying the following patch to XLA fixes the build failure:

<details>

```
diff --git a/xla/python/transfer/event_loop.cc b/xla/python/transfer/event_loop.cc
index 029d9c0b8c..9832b1cb29 100644
--- a/xla/python/transfer/event_loop.cc
+++ b/xla/python/transfer/event_loop.cc
@@ -40,6 +40,10 @@ limitations under the License.
 #include "tsl/platform/env.h"
 
+#ifndef SO_ZEROCOPY
+#define SO_ZEROCOPY    60
+#endif
+
 namespace aux {
 
 class PollEventLoopImpl : public PollEventLoop {
diff --git a/xla/python/transfer/socket-server.cc b/xla/python/transfer/socket-server.cc
index 69c8c027df..f8c51ea8b8 100644
--- a/xla/python/transfer/socket-server.cc
+++ b/xla/python/transfer/socket-server.cc
@@ -35,6 +35,10 @@ limitations under the License.
 #include "xla/python/transfer/transfer_socket.pb.h"
 #include "xla/tsl/concurrency/ref_count.h"
 
+#ifndef SO_ZEROCOPY
+#define SO_ZEROCOPY    60
+#endif
+
 namespace aux {
 
 class SocketServer::SocketNetworkState : public PollEventLoop::Handler {
diff --git a/xla/python/transfer/socket_bulk_transport.cc b/xla/python/transfer/socket_bulk_transport.cc
index 7b61c81c0e..7d41223426 100644
--- a/xla/python/transfer/socket_bulk_transport.cc
+++ b/xla/python/transfer/socket_bulk_transport.cc
@@ -42,6 +42,15 @@ limitations under the License.
 #include "xla/python/transfer/streaming.h"
 #include "xla/tsl/platform/env.h"
 
+#ifndef SO_ZEROCOPY
+#define SO_ZEROCOPY    60
+#endif
+
+#ifndef SO_EE_ORIGIN_ZEROCOPY
+#define SO_EE_ORIGIN_ZEROCOPY  5
+#endif
+
+
 namespace aux {
 
 ZeroCopySendAckTable::ZeroCopySendAckTable() {
```

</details>

zero-copy networking support was introduced in [2017](https://lwn.net/Articles/726917/) but the ubuntu linux kernel 5.4.0 dates from 2020, so the reason for undeclared `SO_ZEROCOPY` is not clear.

pearu:
As in the title.

Conda sysroot includes headers from an old version of linux kernel that does not define SO_ZEROCOPY nor SO_EE_ORIGIN_ZEROCOPY but are required by the feature implemented in https://github.com/openxla/xla/pull/21496 .

Fixes https://github.com/openxla/xla/issues/22083

### DIFF
--- a/xla/python/transfer/event_loop.h
+++ b/xla/python/transfer/event_loop.h
@@ -24,6 +24,17 @@ limitations under the License.
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
 
+// socket.h in conda sysroot include directory does not define
+// SO_ZEROCOPY and SO_EE_ORIGIN_ZEROCOPY that were introduced in a
+// newer kernel version compared to one used by the conda sysroot, see
+// openxla/xla#22083.
+#ifndef SO_ZEROCOPY
+#define SO_ZEROCOPY 60
+#endif
+#ifndef SO_EE_ORIGIN_ZEROCOPY
+#define SO_EE_ORIGIN_ZEROCOPY 5
+#endif
+
 namespace aux {
 
 // Basic event loop using poll().


hawkinsp (approved):

pearu:
AMD ROCm failure
```
17:34:23  LLVM ERROR: Unsupported DotOp found when converting TritonGPU to LLVM.
17:34:23  /root/.cache/bazel/_bazel_root/217377b0e928b171b843eb11ea7bc36e/execroot/xla/bazel-out/k8-opt/bin/xla/service/gpu/gpu_compiler_test_gpu_amd_any.runfiles/xla/tools/ci_build/gpu_build/parallel_gpu_execute: line 74:  1764 Aborted                 (core dumped) "$TEST_BINARY" $@
```
is not related to changes in this PR, AFAIK.

penpornk (commented):
Do you think we could add some test case for this? E.g., would it make sense to check whether `SO_ZEROCOPY` and `SO_EE_ORIGIN_ZEROCOPY` have values in some range that is consistent across all platforms, etc?

pearu:
> Do you think we could add some test case for this? E.g., would it make sense to check whether `SO_ZEROCOPY` and `SO_EE_ORIGIN_ZEROCOPY` have values in some range that is consistent across all platforms, etc?

Testing the values of these defines within XLA may not make sense as the values are defined by Linux kernel (sys/socket.h) that in this particular case is never included by the XLA build process because conda sysroot (which includes older Linux kernel header files) takes a preference over system includes. In principle, there is a chance that the values may be re-defined between different kernel versions but testing that would be out-of-scope for XLA.

To properly test this PR, one would need (i) a CI instance that uses conda environment for building/testing XLA and (ii) a test using the feature implemented in https://github.com/openxla/xla/pull/21496.

I assume https://github.com/openxla/xla/pull/21496 implements the necessary tests for (ii), so we just lack (i) to see if this PR is sensible across platforms. I confirm it works locally in the sense that JAX build from conda environment is successful and functional.

Notice that there exists also another way to fix the underlying issue and it would be similar to one used in https://github.com/NVIDIA/nccl/commit/7f2b337e703d73ed369937c9996e1f3d5f664ad0, that is, when `SO_ZEROCOPY` is undefined then disable the feature of https://github.com/openxla/xla/pull/21496. However, this may cause unnecessary inefficiency as it disables using kernel feature that most likely is available and functional.

For non-conda environments, `SO_ZEROCOPY` is always defined for all relevant platforms  as otherwise CI tests would never have passed on https://github.com/openxla/xla/pull/21496.

hawkinsp:
> Do you think we could add some test case for this? E.g., would it make sense to check whether `SO_ZEROCOPY` and `SO_EE_ORIGIN_ZEROCOPY` have values in some range that is consistent across all platforms, etc?

Yeah I don't think it's possible to sensibly test this. But it will be stable because it's part of the linux system call ABI.

penpornk:
Thank you for the clarifications!

PR #22117, opened 2025-01-30
[XLA:GPU] Fix FMHA unit tests to not use FMHA rewriter

Cjkkkk:
* use FMHA custom call directly instead of calling rewriter.


bchetioui (approved):
Thank you @Cjkkkk!

PR #22123, opened 2025-01-31
Add ScheduleAwareCollectiveOpsCSE to the GPU pipeline

sfvaroglu:


frgossen (changes requested):
Thanks for looking into this, Sevin! 
Can you provide some insights into performance implications? What workload can we see improvements for?

sfvaroglu:
> Thanks for looking into this, Sevin! Can you provide some insights into performance implications? What workload can we see improvements for?

Thanks for the review, Frederik! Seeing ~3% improvement in GPT MOE 500B model with sequence parallelism enabled.

akuegel (commented):
I think one based on the hlo_opt testing tool that checks the HLO after optimizations would work.
Example test with --stage=hlo

https://github.com/openxla/xla/blob/main/xla/service/gpu/tests/offload_scan_output.hlo

sfvaroglu:
> Can you please also add a test? I think one based on the hlo_opt testing tool that checks the HLO after optimizations would work. Example test with --stage=hlo
> 
> https://github.com/openxla/xla/blob/main/xla/service/gpu/tests/offload_scan_output.hlo

Done

xla/tools/hlo_opt/tests/gpu_hlo_collective_cse.hlo
```diff
@@ -0,0 +1,14 @@
...
+HloModule m
+
+// CHECK: %custom-call.1 = s32[32]{0} custom-call(s32[32]{0} %all-gather.1), custom_call_target="my_custom_call1"
+
+ENTRY e {
+  param0 = s32[4] parameter(0)
+  all-gather.1 = s32[32] all-gather(param0), dimensions={0}, replica_groups=[1,8]<=[8], channel_id=1
+  all-gather.2 = s32[32] all-gather(param0), dimensions={0}, replica_groups=[1,8]<=[8], channel_id=2
+  all-gather.3 = s32[32] all-gather(param0), dimensions={0}, replica_groups=[1,8]<=[8], channel_id=3
+  custom-call.1 = s32[32] custom-call(all-gather.3), custom_call_target="my_custom_call1"
```
akuegel, line 12: Nit: I think the test can be a bit improved. I would prefer if we don't rely on which of the "duplicate" HLOs "wins". Maybe use `// CHECK-COUNT-1: all-gather(`
which should make sure that we are left with just one all-gather op.
Also, I guess the user doesn't have to be a custom call? And I think it would be somewhat more realistic if all the all-gather ops have a user. So maybe something like

```
  all-gather.1 = s32[32] all-gather(param0), dimensions={0}, replica_groups=[1,8]<=[8], channel_id=1
  all-gather.2 = s32[32] all-gather(param0), dimensions={0}, replica_groups=[1,8]<=[8], channel_id=2
  all-gather.3 = s32[32] all-gather(param0), dimensions={0}, replica_groups=[1,8]<=[8], channel_id=3
  add = s32[32] add(all-gather.1, all-gather.2)
  ROOT add.1 = s32[32] add(add, all-gather.3)
```

akuegel (approved):
Thank you!

akuegel:
This was rolled back because it caused out of memory on an internal benchmark. This is somewhat an expected possible outcome, as CSE can increase memory requirements. We may need to introduce a flag to make the distance threshold configurable.

PR #22141, opened 2025-01-31
Add "block-scaled-dot" HLO op

sergey-kozub:
This PR adds a new "block-scaled-dot" HLO op, which allows representing a general dot op where one or both inputs have an accompanying scaling factor tensor.

The support for block scaled dot with [MXFP8](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf) inputs is available in Blackwell hardware, and is implemented in [cuDNN](https://github.com/NVIDIA/cudnn-frontend/blob/main/include/cudnn_frontend/node/block_scale_dequantize.h) (requires CUDA 12.8, cuDNN 9.7) and in [Triton](https://github.com/triton-lang/triton/blob/b3dcc32f387d1d54ccd6cbbbc087296c0539e703/include/triton/Dialect/Triton/IR/TritonOps.td#L671) (since 2025-01-28). SOTA models (like [DeepSeekV3](https://arxiv.org/html/2412.19437v1#S3.SS3.SSS2.Px1)) use block scaling. In XLA, there's an equivalent [custom call](https://github.com/openxla/xla/blob/5116f3abf896242a3860707eb8b1cd5cc52dc8f4/xla/service/gpu/transforms/block_scaling_rewriter.h#L44), the intent is to replace it with a first-class HLO op.

This PR only adds the new HLO op (similar to [ragged-dot](https://github.com/openxla/xla/commit/c65b83d1f75b28b07619ea092366801ce1c1a700)), the actual support for lowering this op (via cuDNN and Trtion) will follow. The existing [BlockScalingRewriter](https://github.com/openxla/xla/blob/e816ee557c3025663e0514566423f0eeb77ffb35/xla/service/gpu/transforms/block_scaling_rewriter.h#L65) pass may be used to rewrite this op to an equivalent HLO graph, until the lowering support (to e.g. Triton) is available.

## Syntax

The op has 3 (if one input is quantized) or 4 (if both inputs are quantized) operands. Possible variations:

1. block-scaled-dot(%lhs, %rhs, %lhs_scale), lhs_block_size=L
2. block-scaled-dot(%lhs, %rhs, %rhs_scale), rhs_block_size=R
3. block-scaled-dot(%lhs, %rhs, %lhs_scale, %rhs_scale), lhs_block_size=L, rhs_block_size=R _(block sizes are different)_
4. block-scaled-dot(%lhs, %rhs, %lhs_scale, %rhs_scale), block_size=LR _(block sizes are the same)_

The op has the attributes for describing the dot dimension numbers, same as the general dot op attributes: `lhs_contracting_dims` and `rhs_contracting_dims` are required, `lhs_batch_dims` and `rhs_batch_dims` are optional, and the non-contracting dimensions are implicit.

The op also has an optional attribute, `dequantize_type`. If set, it defines the intermediary dequantization type (e.g. `bf16`), otherwise the dequantization type is implied from the op resulting shape. The user may want to set this explicitly, as this will affect the numerics.

## Semantics

Conceptually, the "block-scaled-dot" op may be substituted with "dequantize -> dot" graph, where "dequantize" is "broadcast -> reshape -> multiply" sequence for each block scaled input (lhs, rhs, or both).

If LHS is quantized (i.e. both %lhs and %lhs_scale operands are present), the operand pair shapes (input/scale) must be the same, except for one dimension `d` (called the block scaled dimension), which has to satisfy "lhs.shape[d] == lhs_scale.shape[d] * lhs_block_size" invariant. Similarly for RHS.

Block scaled LHS/RHS examples:
- ```(f8e5m2[10,128], f8e8m0fnu[10,4])``` for MXFP8 (block size 32);
- ```(f4e2m1fn[12,128], f4e4m3fn[12,8])``` for NVFP4 (block size 16).

HLO op examples:
- Simple MXFP8:
```f32[16,32] block-scaled-dot(f8e5m2[16,128] %lhs, f8e5m2[32,128] %rhs, f8e8m0fnu[16,4] %lhs_scale, f8e8m0fnu[32,4] %rhs_scale), lhs_contracting_dims={1}, rhs_contracting_dims={1}, block_size=32```
- Custom configuration:
```f32[16,4,2,8,32] block-scaled-dot(f8e4m3fn[4,16,256,2,8] %lhs, f8e5m2[4,32,16,256] %rhs, bf16[4,16,32,2,8] %lhs_scale, bf16[4,32,16,16] %rhs_scale), lhs_batch_dims={1,0}, rhs_batch_dims={2,0}, lhs_contracting_dims={2}, rhs_contracting_dims={3}, lhs_block_size=8, rhs_block_size=16, dequantize_type=f32```

### DIFF
--- a/xla/hlo/builder/xla_builder.cc
+++ b/xla/hlo/builder/xla_builder.cc
@@ -2104,6 +2104,35 @@ XlaOp XlaBuilder::RaggedDot(
   });
 }
 
+XlaOp XlaBuilder::BlockScaledDot(
+    XlaOp lhs, XlaOp rhs, XlaOp lhs_scale, XlaOp rhs_scale,
+    const BlockScaledDotConfig& block_scaled_config,
+    std::optional<PrimitiveType> preferred_element_type) {
+  return ReportErrorOrReturn([&]() -> absl::StatusOr<XlaOp> {
+    TF_ASSIGN_OR_RETURN(const Shape* lhs_shape, GetShapePtr(lhs));
+    TF_ASSIGN_OR_RETURN(const Shape* rhs_shape, GetShapePtr(rhs));
+    TF_ASSIGN_OR_RETURN(
+        Shape block_scaled_dot_shape,
+        ShapeInference::InferDotOpShape(
+            *lhs_shape, *rhs_shape, block_scaled_config.dot_dimension_numbers(),
+            preferred_element_type));
+
+    std::vector<XlaOp> operands{lhs, rhs};
+    if (lhs_scale.valid()) {
+      operands.push_back(lhs_scale);
+    }
+    if (rhs_scale.valid()) {
+      operands.push_back(rhs_scale);
+    }
+
+    HloInstructionProto instr;
+    *instr.mutable_shape() = block_scaled_dot_shape.ToProto();
+    *instr.mutable_block_scaled_dot_config() = block_scaled_config;
+    return AddInstruction(std::move(instr), HloOpcode::kBlockScaledDot,
+                          operands);
+  });
+}
+
 absl::Status XlaBuilder::VerifyConvolution(
     const Shape& lhs_shape, const Shape& rhs_shape,
     const ConvolutionDimensionNumbers& dimension_numbers) const {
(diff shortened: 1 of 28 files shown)


mooskagh:
Is there any document describing adding a new op?  
I feel it should be a larger discussion than just reviewing the code changes.

sergey-kozub:
> Is there any document describing adding a new op? I feel it should be a larger discussion than just reviewing the code changes.

Created a discussion topic: https://github.com/openxla/xla/discussions/22535
BTW, could you point me to the discussion of the recently introduced "ragged-dot" HLO op?

mooskagh:
The ragged dot proposal was circulated internally and currently does not have an open-sourceable version available -- I'll try to get it published.

As for the `block-scaled-dot` proposal, we are currently exploring the possibility to add external design docs into the internal database, so that they are more discoverable. Would it be possible to write a Google Docs document with a proposal?

sergey-kozub:
> Would it be possible to write a Google Docs document with a proposal?

Will work on this, setting the PR to Draft in the meanwhile.

PR #22164, opened 2025-01-31
[GPU] Fix sharded autotuning compatibility with result caching.

sergachev:
The algorithm before this change was:
 - collect unique non-cached fusions
 - shard them
 - autotune the shard
 - publish the shard
 - read the other shards
 - merge them into the local cache

This did not work when the ranks observed the cache(s) in different states - sharding requires the set of autotuned fusions to be exactly the same.

The new algorithm is:
 - collect unique fusions ignoring the caches
 - shard them
 - skip fusions present in the local caches of the rank
 - autotune the remainder
 - publish cached + new autotuned results comprising the shard
 - read the other shards
 - merge them into the local cache overwriting on conflicts

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
@@ -38,6 +38,7 @@ limitations under the License.
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
 #include "absl/types/span.h"
+#include "gtest/gtest.h"
 #include "mlir/IR/MLIRContext.h"
 #include "xla/ffi/ffi.h"
 #include "xla/ffi/ffi_api.h"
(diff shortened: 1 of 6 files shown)


PatriosTheGreat (approved):

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -38,6 +38,7 @@ limitations under the License.
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
 #include "absl/types/span.h"
+#include "gtest/gtest.h"
```
PatriosTheGreat, line 41: There is already '#include <gtest/gtest.h>' above. Is this include necessary?

sergachev: No! Fixed.

PR #22256, opened 2025-02-03
[XLA:GPU] Enable cuDNN kernel for block scaled dot on Blackwell

sergey-kozub:
Update the `BlockScalingRewriter` pass to allow lowering the "__op$block_scaled_dot" to a cuDNN graph (instead of HLO graph) - this lowers to a block scaled dot kernel via cuDNN frontend (since v1.10).

The only format supported currently is MXFP8 (both dot inputs are quantized to E4M3FN/E5M2 tensors with E8M0 scales using block size 32).
MXFP8 docs: https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf

The cuDNN kernel has some requirements for the input shapes, so some padding may be needed. Also, the scale tensor must be swizzled (transposed in a specific manner). The tests are added that verify both the pass transformations and the runtime correctness.

The pass is also enabled in the same PR.

### DIFF
--- a/xla/service/gpu/transforms/block_scaling_rewriter.cc
+++ b/xla/service/gpu/transforms/block_scaling_rewriter.cc
@@ -23,6 +23,7 @@ limitations under the License.
 #include "absl/log/log.h"
 #include "absl/status/statusor.h"
 #include "absl/types/span.h"
+#include "xla/hlo/builder/lib/constants.h"
 #include "xla/hlo/builder/xla_builder.h"
 #include "xla/hlo/builder/xla_computation.h"
 #include "xla/hlo/ir/hlo_computation.h"
@@ -30,6 +31,7 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/literal.h"
 #include "xla/primitive_util.h"
+#include "xla/service/gpu/cublas_cudnn.h"
 #include "xla/service/hlo_creation_utils.h"
 #include "xla/service/shape_inference.h"
 #include "xla/shape.h"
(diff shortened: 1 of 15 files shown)


xla/service/gpu/transforms/block_scaling_rewriter.cc
```diff
@@ -194,39 +196,201 @@ absl::StatusOr<HloInstruction*> ExpandDequantizeCustomCall(
...
+  // Swizzle scales to match cuDNN kernel.
+  TF_ASSIGN_OR_RETURN(Shape scale_valid_shape, builder.GetShape(scale_op));
+  int64_t scale_rows = scale_valid_shape.dimensions(1);
+  int64_t scale_cols = scale_valid_shape.dimensions(2);
+  scale_op =
+      Reshape(scale_op, {batch_size, scale_rows / kInputNonContractingTileSize,
+                         4, 32, scale_cols / kScaleContractingTileSize,
+                         kScaleContractingTileSize});
+  scale_op = Transpose(scale_op, {0, 1, 4, 3, 2, 5});
+  scale_op = Reshape(scale_op, scale_valid_shape.dimensions());
```
akuegel, line 307: Is this reshape actually needed? I find it a bit confusing to change the shape back to the before transpose shape. I would expect that the shape should be what the custom call expects. Looking further at the usage of scale_op, I think it is only used to pass it to the custom call.

sergey-kozub: If we remove this one line here, we'd need to add this reshape in cuda_dnn.cc (more than one line), as cuDNN frontend expects a rank-3 tensor for scales. It's just convenient to do it here.

akuegel: Interesting, but does it actually expect this kind of shape? With this reshape, we are logically grouping some of scale_rows / kIputNonContractingTileSize together with scale_cols / kScaleContractingTileSize as dimension 1 (possibly some more of the other dims), and the rest into dimension 2. I find this confusing.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
```diff
@@ -393,6 +394,46 @@ absl::StatusOr<se::gpu::CudnnGraph> HloCustomCallToCuDnnGraph(
             bmm2_grad_gemm1_lhs, bmm2_grad_gemm2_rhs, d_output, d_bmm1_lhs,
             d_bmm1_rhs, d_bmm2_rhs, config.fmha_scale(), dnn_mask_type));
     return graph;
+  } else {
```
mooskagh, line 397: I know it was like that before, but this is a long function with long `if () {..} else if () {..} ...` chain.  
Would it be possible to put the body of every condition (`{..}`) into a separate function?

sergey-kozub: Done.

xla/service/gpu/transforms/block_scaling_rewriter.cc
```diff
@@ -194,39 +196,200 @@ absl::StatusOr<HloInstruction*> ExpandDequantizeCustomCall(
...
+  MXFP8_E4M3FN,
+  // Input: E5M2, scale: E8M0, block size: 32.
+  MXFP8_E5M2,
+};
+
+CompositeType GetCompositeType(const Shape& input_shape,
+                               const Shape& scale_shape) {
+  // Determine the block size from shapes.
+  auto maybe_block_size = GetBlockSize(input_shape, scale_shape);
+  int block_size = maybe_block_size.ok() ? *maybe_block_size : 0;
```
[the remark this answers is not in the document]
sergey-kozub: Thanks.

xla/service/gpu/transforms/block_scaling_rewriter.cc
sergey-kozub: It is totally confusing, I agree. The kernel requires the transpose to align with TMEM layout (https://docs.nvidia.com/cuda/parallel-thread-execution/#tcgen05-mma-scale-factor-a-layout-1x). I guess this could be avoided if the kernel did this transpose in shared memory (but it doesn't).

akuegel (approved):

mooskagh (approved):

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -5523,6 +5529,91 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionBackwardF8OperationGraph(
 #endif
 }
 
+absl::StatusOr<CudnnGraph> GetCudnnBlockScaledDotOperationGraph(
+    dnn::DnnSupport& dnn_support, const dnn::TensorDescriptor& lhs_data,
+    const dnn::TensorDescriptor& lhs_scale,
+    const dnn::TensorDescriptor& rhs_data,
+    const dnn::TensorDescriptor& rhs_scale, dnn::DataType result_type,
+    int block_size) {
+#if CUDNN_VERSION >= 90700
```
golechwierowicz, line 5538: Is there any plan to remove these #ifdef macros in the future?

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -5523,6 +5529,91 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionBackwardF8OperationGraph(
...
+  auto get_tensor_attr = [&](const dnn::TensorDescriptor& desc,
+                             bool is_rhs) -> absl::StatusOr<Tensor_attributes> {
+    TF_ASSIGN_OR_RETURN(std::vector<int64_t> dimensions,
+                        desc.GetPhysicalDimensionsMajorToMinor());
+    std::vector<int64_t> strides = desc.GetPhysicalStridesMajorToMinor();
+    if (dimensions.size() == 2) {
+      dimensions.insert(dimensions.begin(), 1);
+      strides.insert(strides.begin(), dimensions[1] * dimensions[2]);
+    }
+    CHECK(dimensions.size() == 3);
```
golechwierowicz, line 5565: Internally linter complains about this CHECK. Can you change it to CHECK_EQ?

sergey-kozub: Updated.

xla/stream_executor/cuda/cuda_dnn.cc
sergey-kozub: Not that I'm aware of.

mooskagh (approved):

xla/service/gpu/transforms/block_scaling_rewriter.cc
```diff
@@ -194,39 +196,199 @@ absl::StatusOr<HloInstruction*> ExpandDequantizeCustomCall(
...
+    int64_t padded_contracting =
+        RoundUpTo(scale_contracting, kScaleContractingTileSize);
+
+    // Build padding configs.
+    PaddingConfig input_padding_config = MakeNoPaddingConfig(/*rank=*/3);
+    if (size_noncontracting != padded_noncontracting) {
+      input_padding_config.mutable_dimensions(1)->set_edge_padding_high(
+          padded_noncontracting - size_noncontracting);
+    }
+    PaddingConfig scale_padding_config = input_padding_config;
```
mooskagh, line 281: Why is padding for the input the same as scale?

Suppose we have input dimension size 253 (so padding is 3 to pad to multiple of 4), and scaled block size is 64 (so the scales dimension is 4) -- we don't want to pad 3 to the scales dimension, do we?

mooskagh: Ah, just realized it's for non-contracting dimension, which is the same.

PR #22261, opened 2025-02-03
Change the default cuda compiler from nvcc to clang in configure.py

shraiysh:
Changing the default compiler from `configure.py` to use `clang` instead of `nvcc`.

### DIFF
--- a/build_tools/configure/configure.py
+++ b/build_tools/configure/configure.py
@@ -432,7 +432,7 @@ def _parse_args():
       "--cuda_compiler",
       type=CudaCompiler.from_str,
       choices=list(CudaCompiler),
-      default="nvcc",
+      default="clang",
   )
   parser.add_argument(
       "--rocm_compiler",


ddunl:
I'm not opposed to this change in principle, but afaik we still test with nvcc on CI right? I think I need more context to understand the motivation here cc: @beckerhe

shraiysh:
> I'm not opposed to this change in principle, but afaik we still test with nvcc on CI right? I think I need more context to understand the motivation here cc: @beckerhe

Please correct me if I am misinformed here, but Google internally and JAX OSS builds use `clang` as the cuda compiler. This is mainly to be consistent with the main user (JAX).

shraiysh:
Ping for review!

shraiysh:
Ping for review.

ddunl:
Internally it's true that we use clang, but it seems to me JAX uses nvcc by default still - https://github.com/jax-ml/jax/blob/main/build/build.py#L227

nouiz:
@ddunl 
> Internally it's true that we use clang, but it seems to me JAX uses nvcc by default still - https://github.com/jax-ml/jax/blob/main/build/build.py#L227

JAX was using nvcc, but it was too much work for them to maintain that path. So they moved to clang by default. Then we tried to support it, and also decided to moved to clang for our internal CI. So right now, there isn't anyone looking frequently at the nvcc build.
So mostly, this path is not tested. The default should be a tested path, not an untested one that will crash with very high-probability.

For the JAX line, I suppose JAX updated its default, but probably missed some changes.
But in all cases, XLA default should be a test one, so I don't think JAX state should delay merging this PR.

ddunl:
> @ddunl
> 
> > Internally it's true that we use clang, but it seems to me JAX uses nvcc by default still - https://github.com/jax-ml/jax/blob/main/build/build.py#L227
> 
> JAX was using nvcc, but it was too much work for them to maintain that path. So they moved to clang by default. Then we tried to support it, and also decided to moved to clang for our internal CI. So right now, there isn't anyone looking frequently at the nvcc build. So mostly, this path is not tested. The default should be a tested path, not an untested one that will crash with very high-probability.
> 
> For the JAX line, I suppose JAX updated its default, but probably missed some changes. But in all cases, XLA default should be a test one, so I don't think JAX state should delay merging this PR.

If there's a link to something that indicates JAX has switched that would be compelling. From what I can gather about how they are testing on CI at least they are still using nvcc there as well: https://github.com/jax-ml/jax/blob/main/.bazelrc#L214 and grepping for "with_clang" shows only the definition of the config but no users. What am I missing?

beckerhe:
The reason to make NVCC the default was the fact that it takes a long time for a new GPU architecture to make it into a clang release and for this clang release to make it into the usual Linux distributions.

For example the current Ubuntu LTS version (24.04) comes with Clang 18 which has no support for Blackwell. For XLA alone this support is probably not that important since we only have very few CUDA-C++ kernels and none of them use features of the newer GPU architectures (Note that this doesn't affect XLA's capability to generate code for Blackwell). So we would solely rely on PTX backwards compatibility. TF might be a different story (not sure how much performance might be gained when its kernels get compiled explicitly for Blackwell instead of compiling PTX for Hopper).

I thought our CI should also be using NVCC. If that's not the case we might wanna change that our change the strategy.

PR #22306, opened 2025-02-04
[XLA:CPU] Run algebraic simplifier after oneDNNContraction rewriter

kanvi-nervana:
Run AlgebraicSimplifier after OneDNNContractionRewriter and before InstructionFusion pass to remove redundant hlo operations, for example sequence of bitcast pattern. Such redundancy could be a result from oneDNN related HLO passes. When run after InstructionFusion such pattern is fused and hence cannot be optimized.


penpornk (approved):
Thank you for the PR!

akuegel:
This got blocked during merging due to a check that there needs to be a test for non-functional changes. Can you please add one? I guess it could be a test based on the hlo-opt tool with which you can check the HLO after the optimization.

Here is an example of a test with hlo-opt:

https://github.com/openxla/xla/blob/main/xla/tools/hlo_opt/tests/cpu_hlo.hlo

kanvi-nervana:
> This got blocked during merging due to a check that there needs to be a test for non-functional changes. Can you please add one? I guess it could be a test based on the hlo-opt tool with which you can check the HLO after the optimization.
> 
> Here is an example of a test with hlo-opt:
> 
> https://github.com/openxla/xla/blob/main/xla/tools/hlo_opt/tests/cpu_hlo.hlo

I have added a simple test. Thanks!

xla/tools/hlo_opt/tests/run_algsimp_pass.hlo
```diff
@@ -0,0 +1,15 @@
+// NOTE: Assertions have been autogenerated by hlo/tools/generate_hlo_test_checks.py
```
akuegel, line 1: This is missing a RUN: line. Please check the other test as reference how it needs to look like.

akuegel (changes requested):

xla/tools/hlo_opt/tests/run_algsimp_pass.hlo
```diff
@@ -0,0 +1,15 @@
...
+
+// CHECK-LABEL: ENTRY %main
+// CHECK-NEXT:  %[[arg0:[^ ]+]] = f32[768]{0} parameter(0)
+// CHECK-NEXT:  ROOT %[[bitcast1:[^ ]+]] = f32[1,12,64]{2,1,0} bitcast(f32[768]{0} %[[arg0]])
+
+HloModule test
+
+ENTRY main {
+  arg0 = f32[768] parameter(0)
+  bitcast = f32[1,768] bitcast(arg0)
```
akuegel, line 13: I don't think this is what we want to be testing. The idea is that we want to test that the CPU pipeline will have optimized HLO as you expect if you run the AlgebraicSimplifier after OneDNNContraction rewriter. Without this run, the test should fail. Essentially we want to have test coverage for your change.

kanvi-nervana: Would it be clear if the input hlo to the test comes from the output of oneDNNContractionRewriter or do you expect to run the entire cpu-pipeline including oneDNNContractionRewriter as part of the test?

kanvi-nervana: This PR is trying to optimize the HLO that results because of another optimization that is upstreamed but not yet merged to OneDNNContractionRewriter, so running the entire pipeline might not work. I can however, change this test to have the input that's the output of the OneDNNContractioRewriter. Here is the pattern that is observed
`custom-call.269.clone = (f32[1,12,197,197]{3,2,1,0}, u8[68352]{0}) custom-call(bitcast.3, bitcast.6, bitcast.447), custom_call_target="__onednn$matmul", metadata={...}, backend_config={"outer_dimension_partitions":[],"onednn_matmul_config":{"transpose_a":false,"transpose_b":true,"fusions":{"ops":["LINEAR","BINARY_ADD"],"alpha_typecast":1040187392},"optimization_config":{"weights_prepacked":false,"user_scratchpad":true,"bias_broadcast":true},"lhs":{"dims":"5","tensor":{"buffer":"11","dimensions":["1","3","2","4"]}},"rhs":{"dims":"6","tensor":{"buffer":"13","dimensions":["1","3","2","4"]}}}}
  get-tuple-element.3 = f32[1,12,197,197]{3,2,1,0} get-tuple-element(custom-call.269.clone), index=0, metadata={...}
  bitcast.448 = f32[12,197,197]{2,1,0} bitcast(get-tuple-element.3), metadata={...}
  bitcast.8 = f32[1,12,197,197]{3,2,1,0} bitcast(bitcast.448), metadata={...}
  custom-call.1 = f32[1,12,197,197]{3,2,1,0} custom-call(bitcast.8), custom_call_target="__onednn$softmax", metadata={...}, backend_config={"outer_dimension_partitions":[],"onednn_softmax_config":{"softmax_axis":3}}
  bitcast.9 = f32[12,197,197]{2,1,0} bitcast(custom-call.1), metadata={...}
  bitcast.449 = f32[1,12,197,197]{3,2,1,0} bitcast(bitcast.9), metadata={...}`

kanvi-nervana: Any update on this?

akuegel: Did the other optimization get merged already? Then you can write the end2end test. If merging that other optimization is blocked on this, maybe combine it into one PR?
In any case, you are changing the CPU pipeline, and ideally we want to have a test that covers this change. We don't need to test what algebraic simplifier does on some specific input, as AlgebraicSimplifier is not being changed here.

kanvi-nervana: No, the other optimization is not merged yet. I will see if we can combine into one PR or merge this separately. Thanks for your reply.

kanvi-nervana:
Closing this PR as it will be added with the dependent PR. Thanks!!

PR #22368, opened 2025-02-05
[XLA:GPU] Replace "gpu_b100" with "gpu_b200" test backend name

sergey-kozub:
B200 is the name of the released chip.
https://www.nvidia.com/en-us/data-center/dgx-b200/


ddunl (approved):
LGTM, this will require a small patch on our end that I will try to get to today

PR #22376, opened 2025-02-05
Fix erroneous HLO in testcases related to collective backend config.

shraiysh:
Collective backend config is expected to be under a key `"collective_backend_config"` under `backend_config`. In many testcases, this was not the case. This was not flagged by the testcases because the function `IsGPUSyncCollective` supressed any errors in the collective config and assumed `is_sync` to be `false` in those cases (which is generally the case we would like to test with collectives). Hence this went unnoticed so far. We have also stopped supressing the error in that function now.

### DIFF
--- a/xla/service/gpu/transforms/collectives/collective_ops_utils.cc
+++ b/xla/service/gpu/transforms/collectives/collective_ops_utils.cc
@@ -16,17 +16,17 @@ limitations under the License.
 #include "xla/service/gpu/transforms/collectives/collective_ops_utils.h"
 
 #include "xla/hlo/ir/hlo_instruction.h"
-#include "xla/service/collective_ops_utils.h"
 #include "xla/service/gpu/backend_configs.pb.h"
 
 namespace xla {
 namespace gpu {
 
 bool IsGPUSyncCollective(const HloInstruction& instr) {
-  if (!IsCollective(&instr)) return false;
-  GpuBackendConfig backend_config =
-      instr.backend_config<GpuBackendConfig>().value();
-  return backend_config.collective_backend_config().is_sync();
+  auto backend_config = instr.backend_config<GpuBackendConfig>();
+  if (!backend_config.ok()) {
+    return false;
+  }
+  return backend_config->collective_backend_config().is_sync();
 }
 
 }  // namespace gpu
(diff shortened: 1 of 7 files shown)


xla/service/gpu/transforms/collectives/collective_ops_utils.cc
```diff
@@ -22,11 +22,8 @@ namespace xla {
 namespace gpu {
 
 bool IsGPUSyncCollective(const HloInstruction& instr) {
-  auto backend_config = instr.backend_config<GpuBackendConfig>();
-  if (!backend_config.ok()) {
-    return false;
-  }
-  return backend_config->collective_backend_config().is_sync();
+  auto backend_config = instr.backend_config<GpuBackendConfig>().value();
```
Tixxx, line 25: I think it's still worthwhile to check if the backend_config is okay. Some instructions will carry some backend config that's not serializable to GpuBackendConfig proto. Will those instructions fail here?

shraiysh: I would expect this function to be only called from gpu compiler pipeline because it is a part of gpu collective ops utils (which should serialize to gpu backend config then).

Tixxx: That's actually not guaranteed. There are passes in gpu compiler pipeline that designed for all backends, some of those passes will expect non-gpu backend config. For example, the rng will be lowered to a threefry customcall and it carries a backend config that's not serializable to gpu backend config.

shraiysh: That's reasonable. How do we ensure such an error does not occur then? These two seem to be conflicting goals.
1) Allow any backend config, even if it is not serializable to GpuBackendConfig
2) Crash if the backend config is not how it is expected to be (in this case, we expect the field "collective_backend_config" but that does not exist and we are silently ignoring that).

So do we want to fix the testcases and keep ignoring if the backend config is not serializable to GpuBackendConfig? (i.e. keep the testcase changes, but revert the changes to the function)

Tixxx: I think 2 is a bit harsh, xla doesnt standardize on the format of backend config so anything could come in with a weird backend config and xla would crash. Maybe we should limit on what instruction we check on, how about first checking if it's a collective, if so and it has backend config attached, then we can fail if its backend config doesnt conform with the GPUBackendConfig format

shraiysh: That could work as a middle ground. I will update this to do that.

Tixxx (approved):

golechwierowicz (approved):

golechwierowicz:
Can you rebase? There's a fix at HEAD for some internal failures.

shraiysh:
> Can you rebase? There's a fix at HEAD for some internal failures.

Done.

Issue #20803, opened 2024-12-21
cuDNN, cuFFT, and cuBLAS Plugin Registration Errors


johnnkp:
This issue is copy from https://github.com/tensorflow/tensorflow/issues/62075#issuecomment-2535391485.

Tensorflow CUDA users have occurred the below errors since 2023 and unable to use GPU acceleration:
```
2023-10-09 13:36:23.355516: E tensorflow/compiler/xla/stream_executor/cuda/cuda_dnn.cc:9342] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2023-10-09 13:36:23.355674: E tensorflow/compiler/xla/stream_executor/cuda/cuda_fft.cc:609] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2023-10-09 13:36:23.355933: E tensorflow/compiler/xla/stream_executor/cuda/cuda_blas.cc:1518] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
```

A possible reason is missing checking before register factory. In `xla\stream_executor\rocm\rocm_*.cc`, plugin registration function will call `PluginRegistry::Instance()->HasFactory()` before `PluginRegistry::Instance()->RegisterFactory<>()`:
```
auto rocBlasAlreadyRegistered = PluginRegistry::Instance()->HasFactory(
      rocm::kROCmPlatformId, PluginKind::kBlas);

  if (!rocBlasAlreadyRegistered) {
    absl::Status status =
        PluginRegistry::Instance()
            ->RegisterFactory<PluginRegistry::BlasFactory>(...);

    if (!status.ok()) {
      LOG(ERROR) << "Unable to register rocBLAS factory: " << status.message();
    }
  }
```

Currently, initialize_cublas(), initialize_cudnn() and initialize_cufft() in `xla\stream_executor\cuda\cuda_*.cc` directly register factory without checking existing factory. XLA team please add the checking for Tensorflow contributors to merge, such that CUDA users can install nightly build and test the fix soon.

Update (3 Mar): PR 22391 approved and waiting merge to both XLA and Tensorflow.
```[tasklist]
### Tasks
- [x] cuda_blas.cc: Reduce log level to info in initialize_cublas()
- [x] cuda_dnn.cc: Reduce log level to info in initialize_cudnn()
- [x] cuda_fft.cc: Reduce log level to info in initialize_cufft()
```

[...]

PR #22391, opened 2025-02-06
[XLA:GPU] Reduce CUDA plugin registration messages to info

johnnkp:
This pull request is to reduce the CUDA plugin registration log level to avoid user confusing confusion in https://github.com/openxla/xla/issues/20803. Actually, no functionality errors happened with CUDA plugin registration.

### DIFF
--- a/xla/stream_executor/cuda/cuda_blas.cc
+++ b/xla/stream_executor/cuda/cuda_blas.cc
@@ -1404,7 +1404,7 @@ void initialize_cublas() {
           });
 
   if (!status.ok()) {
-    LOG(ERROR) << "Unable to register cuBLAS factory: " << status.message();
+    LOG(INFO) << "Unable to register cuBLAS factory: " << status.message();
   }
 }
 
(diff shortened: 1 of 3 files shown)


johnnkp:
@thomasjoerg @beckerhe Is there any update on code review?

akuegel:
@ezhulenev maybe you can take a look while @beckerhe is still out?

beckerhe:
Hey, sorry for the late response.

One question: The issue reported in https://github.com/openxla/xla/issues/20803 is not affecting the functionality. This is purely to avoid the log message, right?

I'm hesitant because checking the factory is just a band-aid. The actual problem is that the CUDA plugins get initialized twice which is some build/linker problem.

johnnkp:
I don't know where is the root of calling initialize function more than once. However, we should at least reduce the log level to avoid confusing users.

beckerhe:
> I don't know where is the root of calling initialize function more than once. However, we should at least reduce the log level to avoid confusing users.

Yes. Agreed. Thanks for making the change.

beckerhe:
Would you mind also updating the PR description?

johnnkp:
PR description also updated.

beckerhe (approved):
Thank you!

PR #22437, opened 2025-02-06
Added frontend attribute handling to explicit_stream_annotation_async_wrapper 

chaserileyroberts:
This is a small change that ensures the frontend attributes are correctly passed to both the `async-start` and `async-done` created pairs. This also clears the scheduling attributes that are directly on the call operation and inner ops.

The specific goal of this change is to have stable support combining the scheduling group ids with stream annotation in JAX. 

```python
with set_xla_metadata(_scheduling_group_id=1):
   result = compute_on("gpu_stream:1")(jitted_func)(...)
```

Currently, the issue stems from the `set_xla_metadata` context manager, which will apply the frontend attribute to all operations, including the ones within our `jitted_func`. When the same scheduling annotations is found in two `HloComputation`s, an error is raised in `LegalizeSchedulingAnnotations`. This is intended to avoid hitting this check, and cleaning up the annotations on the wrapped streamed computation.

### DIFF
--- a/xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
+++ b/xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
@@ -39,12 +39,18 @@ static absl::StatusOr<bool> AsynchronizeInstruction(HloInstruction* instr) {
     return false;
   }
   HloComputation* computation = instr->parent();
+  auto original_attributes = instr->frontend_attributes();
+  // Clear the op's frontend attributes.
+  instr->set_frontend_attributes(FrontendAttributes());
   TF_ASSIGN_OR_RETURN(
       HloInstruction * done,
       computation->CreateAsyncInstructions(
           instr, {},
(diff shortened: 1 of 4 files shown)


xla/service/gpu/transforms/explicit_stream_annotation_async_wrapper.cc
```diff
@@ -39,12 +39,18 @@ static absl::StatusOr<bool> AsynchronizeInstruction(HloInstruction* instr) {
     return false;
   }
   HloComputation* computation = instr->parent();
+  auto original_attributes = instr->frontend_attributes();
+  // Clear the op's frontend attributes.
```
jreiffers, line 43: Let's add a couple of lines of explanation why we're doing this. It's not obvious for the casual reader.

chaserileyroberts:
This has been updated and is ready for final review

jreiffers (approved):

akuegel (approved):

PR #22440, opened 2025-02-06
DRAFT: Add scheduling done group id

chaserileyroberts:
This is a simple draft patch of how we could implement a `_scheduling_done_group_id`, which would allow done operations to be scheduled in groups separate from their matching start op.

### DIFF
--- a/xla/service/latency_hiding_scheduler.h
+++ b/xla/service/latency_hiding_scheduler.h
@@ -359,8 +359,20 @@ class AnnotationTracker {
   std::vector<int64_t> GetAnnotations(const HloComputation* comp) const {
     return comp_annotation_map_.at(comp);
   }
+
+  bool IsAsyncDoneType(HloOpcode opcode) const {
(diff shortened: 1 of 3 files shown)


xla/service/latency_hiding_scheduler.h
toli-y, line 1: Is it possible to add a unit test?

vvvm23:
would be awesome to have this feature, it would unblock a lot of use cases for me

toli-y:
> would be awesome to have this feature, it would unblock a lot of use cases for me

@vvvm23  Can you elaborate the usecases? Best if you can provide some JAX snippets and describe how you would want to them to be scheduled.

vvvm23:
afaik not specifying when a scheduling group is "done" still results in some exposed comms (between groups) when overlapping compute and comms. Having this ability can remove the last of the exposed comms.

I think @chaserileyroberts can comment further on this.

PR #22450, opened 2025-02-07
[NVIDIA GPU] [XLA_GPU_MS_COLLECTIVE] Support round-robin runtime stream assignment

terryysun:
With https://github.com/openxla/xla/pull/19026, LHS can overlap appropriate async collectives without deadlock. This PR adds support at runtime where we leverage the overlapping schedule produced by LHS and perform a round-robin stream assignment for collectives.


xla/backends/gpu/runtime/nccl_send_thunk.cc
```diff
@@ -119,7 +120,7 @@ absl::Status NcclSendThunk::RunNcclCollective(const ExecuteParams& params,
     if (config_.validation_kind ==
         NcclP2PConfig::ValidationKind::kConditional) {
       se::StreamExecutor* executor = params.stream->parent();
-      TF_ASSIGN_OR_RETURN(int64_t* counter,
+      TF_ASSIGN_OR_RETURN(int64_t * counter,
```
frgossen, line 123: nit: formatting

xla/service/gpu/execution_stream_assignment.cc
```diff
@@ -45,7 +60,11 @@ ExecutionStreamAssignment::ExecutionStreamAssignment(
   // on the entrypoint computation will be assigned `ExecutionStreamId(0)`, and
   // each invocation of `async-start` will result in the target computation
   // being assigned a new `ExecutionStreamId`.
-  ExecutionStreamId next_stream_id = ExecutionStreamId(1);
+  ExecutionStreamId next_compute_stream_id = ExecutionStreamId(1);
+  ExecutionStreamId next_collective_stream_id =
```
frgossen, line 64: Why is the tyoe of the next collective stream id `ExecutionStreamId`?

xla/service/gpu/gpu_latency_hiding_scheduler.cc
```diff
@@ -126,19 +126,43 @@ bool IsAsyncPair(const HloInstruction& from, const HloInstruction& target) {
   return IsGpuAsyncStart(from) && IsGpuAsyncDone(target);
 }
 
+// Util function for getting replica groups from different data structures.
+static std::vector<std::vector<int64_t>> get_replica_groups(
```
frgossen, line 130: I think there is something like this in the collective ops utils. Can you mve it there and test it separately?

xla/service/gpu/gpu_latency_hiding_scheduler.cc
```diff
@@ -126,19 +126,43 @@ bool IsAsyncPair(const HloInstruction& from, const HloInstruction& target) {
...
+    const HloInstruction* instruction) {
+  std::vector<std::vector<int64_t>> replica_groups;
+  if (instruction->opcode() == HloOpcode::kCollectivePermuteStart) {
+    absl::c_transform(instruction->source_target_pairs(),
+                      std::back_inserter(replica_groups),
+                      [](const std::pair<int64_t, int64_t>& pair) {
+                        std::vector<int64_t> ids({pair.first, pair.second});
+                        return ids;
+                      });
+  } else {
```
frgossen, line 140: Does this work for all collectives except collective-permute start? Can we test that?

frgossen (changes requested):
Thank you for working on this. Multiple stream for collectives will be great to have! 

This is a big PR and I think at least the stream assignment, the runtime integration, and some util functions could be three separate PRs. 

One thing I am wondering is if we could do this in the latency-hiding scheduler. The scheduler models resources and we could have one per collective stream. That way the scheduler would perform the stream assignment and compose well with it. I'm not saying that would be better but it might be worth thinking about since you seem to run into issues with the LHS (implied by some comments).

dimitar-asenov:
@terryysun Could you please provide an update on this PR?

terryysun:
> @terryysun Could you please provide an update on this PR?

Hi @dimitar-asenov, we are breaking this PR to 3 steps so that it's easier to review and merge. The first step https://github.com/openxla/xla/pull/23522 is under review. I was thinking of keeping this PR as a reference for breakdown PRs and closing it eventually.

dimitar-asenov:
Thanks for clarifying @terryysun. If you no longer intend to submit this PR, could you please mark it as `draft`? That more accurately reflects its state and it also changes how we see it in our internal dashboards.

PR #22556, opened 2025-02-11
Verify GPU Peak Memory Usage for Latency Hiding Scheduler Pass

zhenying-liu:
When working with llama2-7b from MaxText, we encountered an out-of-memory error when enabling optimizer state offloading and latency hiding scheduling. However, it worked fine without latency hiding scheduling. This change list:
- Stores the GPU peak memory usage in the metadata of the latency hiding scheduler pass.
- Adds unit tests to verify the GPU peak memory usage.

### DIFF
--- a/xla/service/latency_hiding_scheduler.cc
+++ b/xla/service/latency_hiding_scheduler.cc
@@ -2760,7 +2760,8 @@ absl::StatusOr<bool> LatencyHidingScheduler::Run(
   LOG(INFO) << "LatencyHidingScheduler current memory usage: "
             << scheduler_core_->GetMemoryPeak()
             << " bytes. Current limit: " << scheduler_core_->GetMemoryLimit();
-  SetKVMetric(module, "peak_memory_usage", scheduler_core_->GetMemoryPeak());
+  SetKVMetric(module, std::string(kPeakMemoryUsage),
+              scheduler_core_->GetMemoryPeak());
   for (HloComputation* computation : computations_to_schedule) {
     VLOG(1) << "Statistics before scheduling:";
     LogScheduleStatistics(computation);
--- a/xla/service/latency_hiding_scheduler.cc
+++ b/xla/service/latency_hiding_scheduler.cc
@@ -2760,6 +2760,7 @@ absl::StatusOr<bool> LatencyHidingScheduler::Run(
   LOG(INFO) << "LatencyHidingScheduler current memory usage: "
             << scheduler_core_->GetMemoryPeak()
             << " bytes. Current limit: " << scheduler_core_->GetMemoryLimit();
+  SetKVMetric(module, "peak_memory_usage", scheduler_core_->GetMemoryPeak());
   for (HloComputation* computation : computations_to_schedule) {
     VLOG(1) << "Statistics before scheduling:";
     LogScheduleStatistics(computation);
--- a/xla/service/latency_hiding_scheduler.h
+++ b/xla/service/latency_hiding_scheduler.h
@@ -1164,6 +1164,8 @@ class LatencyHidingScheduler : public HloModulePass {
   absl::flat_hash_set<HloComputation*> computations_to_schedule_;
 };
 
+static constexpr absl::string_view kPeakMemoryUsage = "peak_memory_usage";
+
 }  // namespace xla
 
 #endif  // XLA_SERVICE_LATENCY_HIDING_SCHEDULER_H_
--- a/xla/service/latency_hiding_scheduler_test.cc
+++ b/xla/service/latency_hiding_scheduler_test.cc
@@ -35,8 +35,10 @@ limitations under the License.
 #include "absl/strings/string_view.h"
 #include "absl/types/span.h"
 #include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_module_metadata.h"
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/hlo/ir/hlo_schedule.h"
+#include "xla/hlo/pass/hlo_pass_pipeline.h"
 #include "xla/hlo/transforms/collectives/async_collective_creator.h"
 #include "xla/service/hlo_cost_analysis.h"
 #include "xla/service/legalize_scheduling_annotations.h"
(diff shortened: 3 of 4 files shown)


xla/service/latency_hiding_scheduler_test.cc
```diff
@@ -3074,6 +3088,51 @@ ENTRY AddR2 {
...
+        }
+      }
+    }
+  }
+  return 0;
+}
+
+TEST_F(LatencyHidingSchedulerTest, GetGPUPeakMemoryUsage) {
+  // Retrieve GPU peak memory usage and verify it excludes host memory
+  absl::string_view kPeakMemoryUsageWithCpuOffload = R"(
```
allanrenucci, line 3110: For consistency with the other tests:
```suggestion
  absl::string_view hlo_string = R"(
```

xla/service/latency_hiding_scheduler_test.cc
```diff
@@ -3074,6 +3088,51 @@ ENTRY AddR2 {
   XLA_VLOG_LINES(1, hlo_module->ToString());
 }
 
+// Function to retrieve the value for a specific key from HloPassMetadata
+// for a given pass_name. Returns 0 if the key or pass_name is not found.
+int64_t GetKeyValueFromHloPassMetadata(const HloModuleMetadataProto& metadata,
+                                       const std::string& pass_name,
+                                       const std::string& target_key) {
```
allanrenucci, line 3095: Prefer absl::string_view:
```suggestion
int64_t GetKeyValueFromHloPassMetadata(const HloModuleMetadataProto& metadata,
                                       absl::string_view pass_name,
                                       absl::string_view target_key) {
```

xla/service/latency_hiding_scheduler.cc
```diff
@@ -2817,6 +2817,7 @@ absl::StatusOr<bool> LatencyHidingScheduler::Run(
   LOG(INFO) << "LatencyHidingScheduler current memory usage: "
             << scheduler_core_->GetMemoryPeak()
             << " bytes. Current limit: " << scheduler_core_->GetMemoryLimit();
+  SetKVMetric(module, "peak_memory_usage", scheduler_core_->GetMemoryPeak());
```
allanrenucci, line 2820: This should probably be a constant defined in the header file and used in the test.

xla/service/latency_hiding_scheduler_test.cc
```diff
@@ -3074,6 +3088,51 @@ ENTRY AddR2 {
...
+                          ParseHloText(kPeakMemoryUsageWithCpuOffload));
+  EXPECT_TRUE(hlo_module->has_entry_computation());
+  auto sched_config = GetDefaultSchedConfig();
+  TF_EXPECT_OK(RunScheduler(hlo_module.get(), sched_config,
+                            std::make_unique<TestLatencyEstimator>(), nullptr));
+  const HloModuleMetadataProto& metadata = hlo_module->metadata()->proto();
+  std::string pass_name = "latency-hiding-scheduler";
+  std::string target_key = "peak_memory_usage";
+  int64_t peak_memory_bytes =
+      GetKeyValueFromHloPassMetadata(metadata, pass_name, target_key);
```
allanrenucci, line 3132:
```suggestion
  std::string target_key = "peak_memory_usage";
  int64_t peak_memory_bytes =
      GetKeyValueFromHloPassMetadata(metadata, LatencyHidingScheduler::kName, target_key);
```

allanrenucci (changes requested):

golechwierowicz:
Why do you need this change if you can just log the peak memory?

jreiffers:
@golechwierowicz The test is currently using an incorrect byte size function (and so is gpu_compiler, we'll send a separate PR to fix that, but it's more complicated). The metadata lets us verify the logic and the value. Do you have a better idea for how to test the peak memory usage value?

xla/service/latency_hiding_scheduler_test.cc
zhenying-liu: Done.

xla/service/latency_hiding_scheduler.cc
zhenying-liu: Yes, defined a constant.

xla/service/latency_hiding_scheduler_test.cc
zhenying-liu: Done.

xla/service/latency_hiding_scheduler.h
```diff
@@ -1164,6 +1164,8 @@ class LatencyHidingScheduler : public HloModulePass {
   absl::flat_hash_set<HloComputation*> computations_to_schedule_;
 };
 
+static constexpr absl::string_view kPeakMemoryUsage = "peak_memory_usage";
```
golechwierowicz, line 1167: This pollutes the entire xla namespace. Since it's just for verification & debugging I would either move it somewhere else (within LHS) or keep raw str approach.

xla/service/latency_hiding_scheduler.cc
```diff
@@ -2760,6 +2760,8 @@ absl::StatusOr<bool> LatencyHidingScheduler::Run(
   LOG(INFO) << "LatencyHidingScheduler current memory usage: "
             << scheduler_core_->GetMemoryPeak()
             << " bytes. Current limit: " << scheduler_core_->GetMemoryLimit();
+  SetKVMetric(module, std::string(kPeakMemoryUsage),
```
golechwierowicz, line 2763: If you're using HloPassPipeline primitives can you explicitly use the `hlo_pass_pipeline` include then?

xla/service/latency_hiding_scheduler_test.cc
```diff
@@ -151,19 +153,30 @@ absl::StatusOr<bool> RunScheduler(
...
-                      AsyncCollectiveCreator(std::move(config)).Run(module));
-  TF_ASSIGN_OR_RETURN(value, LegalizeSchedulingAnnotations(
-                                 LegalizeSchedulingAnnotations::Config())
-                                 .Run(module));
+  HloPassPipeline pipeline("latency_hiding_scheduler_pipeline");
+  pipeline.AddPass<AsyncCollectiveCreator>(std::move(config));
+  pipeline.AddPass<LegalizeSchedulingAnnotations>(
+      LegalizeSchedulingAnnotations::Config());
+
   HloCostAnalysis::ShapeSizeFunction shape_size_bytes =
```
golechwierowicz, line 161: Change seems GPU specific not XLA specific. Can't you write a test in `gpu_hlo_schedule_test.cc`? There you can manipulate the shape function to your liking. Also the subsequent change to gpu_pipeline would fit more nicely.

golechwierowicz (changes requested):

golechwierowicz:
Thanks @jreiffers. It was not clear to me from PR description alone what you're trying to achieve.

xla/service/latency_hiding_scheduler_test.cc
zhenying-liu: Thanks for the advice.

I agree that we need to use gpu_hlo_schedule to schedule first, and then run LHS with the proper shape function. Our goal is to set a memory limit for a unit test that LHS can schedule within that memory limit during the first rerun. That way, we can reproduce and fix the OOM issue when LHS is enabled.

So I'm going to close this PR for now.

PR #22575, opened 2025-02-11
[XLA:GPU] Fix triton sparse dot lowering on Blackwell

sergey-kozub:
Sparse dot is supported for MMA v2 and v3 only, and sm100/sm120 should use MMA v2 (v3 is Hopper-only).

### DIFF
--- a/xla/backends/gpu/codegen/triton/transforms/sparse_passes.cc
+++ b/xla/backends/gpu/codegen/triton/transforms/sparse_passes.cc
@@ -273,7 +273,9 @@ class SparseBlockedToMMA : public RewritePattern {
     assert(compute_capability_ >= 80 &&
            "SparseDot is only supported on Ampere or higher");
     bool allow_v3 = !triton::tools::getBoolEnv("DISABLE_MMA_V3");
-    int version_major = compute_capability_ >= 90 && allow_v3 ? 3 : 2;
+    // Sparse dot is supported for MMA v2 and v3 only, and sm100/sm120 should
+    // use MMA v2 (v3 is Hopper-only).
+    int triton_mma_version = compute_capability_ == 90 && allow_v3 ? 3 : 2;
 
     // get MMA encoding and new return type given the number of warps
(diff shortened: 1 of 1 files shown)


mooskagh (approved):

xla/backends/gpu/codegen/triton/xla_triton_sparse_passes.cc
```diff
@@ -273,7 +273,7 @@ class SparseBlockedToMMA : public RewritePattern {
     assert(compute_capability_ >= 80 &&
            "SparseDot is only supported on Ampere or higher");
     bool allow_v3 = !triton::tools::getBoolEnv("DISABLE_MMA_V3");
-    int version_major = compute_capability_ >= 90 && allow_v3 ? 3 : 2;
+    int triton_mma_version = compute_capability_ == 90 && allow_v3 ? 3 : 2;
```
thomasjoerg, line 276: Please add your PR description as code comment here: 
> Sparse dot is supported for MMA v2 and v3 only, and sm100/sm120 should use MMA v2 (v3 is Hopper-only).

Is this the desired end state or a temporary workaround until Blackwell support is added?

thomasjoerg, line 276: Is it possible to test this logic with reasonable effort?

thomasjoerg (changes requested):

xla/backends/gpu/codegen/triton/xla_triton_sparse_passes.cc
sergey-kozub: Added the comment.
Blackwell support is not planned, but certainly possible (the hardware supports 2:4 sparsity). If someone needs the feature, we can consider doing this.

sergey-kozub: Some JAX test already covers this code (that's how we learned about the regression).

thomasjoerg (approved):

PR #22593, opened 2025-02-11
[XLA:GPU] Fix triton dot op on sm120 (RTX50xx)

sergey-kozub:
Triton doesn't currently support sm120 GPUs - adding a patch to fix that, the upstream support should be available soon.

Converting "12.0" arch into "10.0" is not correct, as they're not compatible - removing this.


gflegar (approved):
LGTM

PR #22680, opened 2025-02-13
Changing the default value of the flag xla_dump_hlo_as_long_text

shraiysh:
This stems from the difference in HLO dumps from tools like hlo-opt, multihost_hlo_runner, and hlo_runner_main. Although the options related to printing HLO are tied to the HLO via DebugOptions, these tools have different behaviors because the tool hlo-opt uses `ToString()` function from `HloModule` while `functional_hlo_runner` uses a separate dumping utility from `xla/service/dump.cc`. To standardize this behavior, we first are making the default behavior uniform by setting the default value of long text HLO dumps to be true. This ensures that the HLO dumps will be functional by default (for example, backend_config will be printed).

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -76,7 +76,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_dump_include_timestamp(false);
   opts.set_xla_dump_max_hlo_modules(-1);
   opts.set_xla_dump_module_metadata(false);
-  opts.set_xla_dump_hlo_as_long_text(false);
+  opts.set_xla_dump_hlo_as_long_text(true);
   opts.set_xla_dump_large_constants(false);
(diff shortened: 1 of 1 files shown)


shraiysh:
I am expecting to see some failures in the CI (I will edit the testcases to fix them, if any), however I am inviting comments if anyone is concerned about this change and it affects them adversely.

akuegel (approved):
Thanks. Let's give it a try and see whether any tests need to be updated.

shraiysh:
> Thanks. Let's give it a try and see whether any tests need to be updated.

Copybara seems to have failed with this. If the tests are in OSS, can you point me to them so I can fix them?

akuegel:
> > Thanks. Let's give it a try and see whether any tests need to be updated.
> 
> Copybara seems to have failed with this. If the tests are in OSS, can you point me to them so I can fix them?

It was a unrelated breakage that has been resolved now. No tests had to be adapted, it could be merged without additional changes.

PR #22717, opened 2025-02-14
Fix kernel_thunk_test compilation error

apivovarov:
### Description

GCC 13 compilation error (Ubuntu 24.04)
```
Error:
kernel_thunk_test.cc:101:26: error: converting to 'std::optional<absl::lts_20230802::flat_hash_set<long int> >' from initializer list would use explicit constructor 'constexpr std::optional<_Tp>::optional(std::in_place_t, _Args&& ...) [with _Args = {}; typename std::enable_if<__and_v<std::is_constructible<_Tp, _Args ...> >, bool>::type <anonymous> = false; _Tp = absl::lts_20230802::flat_hash_set<long int>]'
  101 |       KernelThunk::Create({"add_f32"}, {slice}, {slice}, "add_f32",
      |       ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  102 |                           se::ThreadDim(4), /*invariant_arguments=*/{{}}));
      |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
The compilation error occurs because {} (invariant_arguments) is being used to initialize `std::optional<absl::flat_hash_set<int64_t>>`, but `std::optional` has an explicit constructor that takes an `in_place_t` argument, which prevents `{}` from being implicitly converted.

I think we can use `std::nullopt` here instead of `{{}}`

### Testing
```
//xla/backends/cpu/runtime:kernel_thunk_test                             PASSED in 0.0s
```

### DIFF
--- a/xla/backends/cpu/runtime/kernel_thunk_test.cc
+++ b/xla/backends/cpu/runtime/kernel_thunk_test.cc
@@ -97,9 +97,9 @@ TEST(KernelThunkTest, AddF32Inline) {
   BufferAllocation::Slice slice = CreateBufferAllocationSlice(alloc);
 
   TF_ASSERT_OK_AND_ASSIGN(
-      auto thunk,
-      KernelThunk::Create({"add_f32"}, {slice}, {slice}, "add_f32",
-                          se::ThreadDim(4), /*invariant_arguments=*/{{}}));
+      auto thunk, KernelThunk::Create({"add_f32"}, {slice}, {slice}, "add_f32",
+                                      se::ThreadDim(4),
+                                      /*invariant_arguments=*/std::nullopt));
 
   AddF32HostKernel host_kernels;
   Thunk::ExecuteParams params = {&host_kernels, &allocations};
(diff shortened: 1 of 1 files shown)


allanrenucci:
> I think we can use std::nullopt here instead of {{}}

IIUC using std::nullopt will disable some checks: https://github.com/openxla/xla/blob/6da089db33ff5cb1974846d8496e3e7b5e14eb40/xla/backends/cpu/runtime/kernel_thunk.cc#L208-L210

I am not sure why `invariant_arguments_` was made optional in #20993. I will defer to @WillFroom as the author of the change.

WillFroom:
> I am not sure why `invariant_arguments_` was made optional in #20993.

Ah I made that change to ease integration, I realised recently this was a mistake (for other reasons), I'll create a change to re-add them today and remove them being optional

WillFroom:
#22812 reverts the arguments being optional

allanrenucci:
Closing in favor of #22812. I believe it should also address the compilation error.

PR #22788, opened 2025-02-17
Implement GetCompiledMemoryStats for GPU AOT executables

jaro-sevcik:
This implements `GetCompiledMemoryStats` for ahead-of-time compiled executables. With this patch,
one can estimate memory consumption of a JAX function even without access to a GPU.

Unfortunately, the patch duplicates code between unloaded and loaded GPU executables
and between `GpuThunkAotCompilationResult::GetBufferAssignment()`  and `Compiler::BufferSizeBytesFunction()`+`GpuCompiler::ShapeSizeBytesFunction()`.
This could be perhaps improved by exposing the relevant compiler code as static methods,
but that does not seem worth the extra complexity.

The patch also threads `pointer_size` from `GpuCompiler` to `GpuThunkAotCompilationResult` so that we can
get buffer allocation sizes without direct access to the compiler. Another option would be embedding
`pointer_size` within `CompilationResultProto`.

Also note that this still does not set `generated_code_size_in_bytes` correctly - that would require
duplicating some code from `GpuExecutable::SizeOfGeneratedCodeInBytes()`.

### DIFF
--- a/xla/service/compiler.h
+++ b/xla/service/compiler.h
@@ -77,6 +77,11 @@ class AotCompilationResult {
     return Unimplemented("LoadExecutable unimplemented.");
   }
 
+  virtual absl::StatusOr<std::unique_ptr<BufferAssignment>> buffer_assignment()
+      const {
+    return Unimplemented("buffer_assignment unimplemented.");
+  }
+
   // Returns the optimized HLO module if one was computed and the implementation
   // supports it.
   virtual const HloModule* optimized_module() const = 0;
--- a/xla/pjrt/BUILD
+++ b/xla/pjrt/BUILD
@@ -550,6 +550,7 @@ cc_library(
         "//xla/hlo/ir:hlo",
         "//xla/service:compiler",
         "//xla/service:executable",
+        "//xla/service:hlo_proto_util",
         "//xla/tsl/platform:statusor",
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/log:check",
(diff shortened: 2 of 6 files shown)


xla/service/compiler.h
```diff
@@ -76,6 +76,11 @@ class AotCompilationResult {
     return Unimplemented("LoadExecutable unimplemented.");
   }
 
+  virtual absl::StatusOr<std::unique_ptr<BufferAssignment>>
+  GetBufferAssignment() const {
```
bchetioui, line 80: nit: maybe `buffer_assignment`, to match `optimized_module`?

bchetioui (changes requested):
Looks reasonable to me, but also would like @beckerhe's feedback.

beckerhe (approved):
LGTM! Thank you!

akuegel:
It seems this change breaks https://github.com/openxla/xla/blob/main/xla/pjrt/gpu/se_gpu_pjrt_compiler_aot_test.cc
@jaro-sevcik can you please check whether you can reproduce the failure, and fix it?

bchetioui:
Is there still a plan to land this PR? It's been on hold for a while now :)

xla/service/compiler.h
jaro-sevcik: Done.

jaro-sevcik:
@bchetioui I have fixed the failure, could you take another look?

bchetioui (approved):

PR #22800, opened 2025-02-18
Change the default value of print_operand_shape_ to false and print_large_constants_ to true.

shraiysh:
Operand shape in long hlo text adds redundant information, which shouldn't be required. Changing the default value to off.

The large constants were also printed earlier by default print options, and it is required for parsability and reproducibility. Turning this on by default. This is still controlled by debug option and the default value of that flag disables the large constants, and that behavior is not changed. Just the default print options change here.


akuegel:
> Hopefully, the tests will not require any changes here either :)

Actually they do require changes. A quick search for where hlo_opt tool is used should reveal most of the affected tests (but there could be others).

shraiysh:
@akuegel I also changed the default value of `dump_large_constants_`, because in the normal long form dump, I think they should be dumped for reproducibility. Does this sound good? If not, I can remove that change.

akuegel (approved):
Also changing print_large_constants sounds good to me.

shraiysh:
For TF tests, should I create a PR on that project?

akuegel:
> For TF tests, should I create a PR on that project?

I guess it is not possible to fix it "standalone" there. So I think this is something we have to do internally (we can change both Tensorflow and XLA repo at the same time). For compile_mlir_util tests, maybe something similar to https://github.com/tensorflow/tensorflow/pull/87613 can be done? If you could help with those, that would be great :)

shraiysh:
> > For TF tests, should I create a PR on that project?
> 
> I guess it is not possible to fix it "standalone" there. So I think this is something we have to do internally (we can change both Tensorflow and XLA repo at the same time). For compile_mlir_util tests, maybe something similar to [tensorflow/tensorflow#87613](https://github.com/tensorflow/tensorflow/pull/87613) can be done? If you could help with those, that would be great :)

Yes, I can do that. Working on it.

shraiysh:
This should help with the tensorflow tests: https://github.com/tensorflow/tensorflow/pull/87613

allanrenucci (approved):
Re-approving for internal testing

allanrenucci:
There are ~70 tests failing internally. I am taking a look. This may take some time.

allanrenucci:
Can you please fix `//xla/hlo/parser:hlo_parser_test`?

shraiysh:
> Can you please fix `//xla/hlo/parser:hlo_parser_test`?

Done. And Rebased.

akuegel (approved):

allanrenucci:
@shraiysh, sorry for the delay. I am OOO currently and will finish fixing internal tests once I am back. No need to rebase. Thanks!

PR #22827, opened 2025-02-18
Document the `input_format` flag for the `run_hlo_module` CLI

kasper0406:
I spent too much time figuring out how to run a StableHLO module. It would be helpful to have this information visible in the docs.

Today, if a StableHLO module is supplied, it errors the following way:
```
➜  xla git:(kn/docs-improvement) ✗ bazel run --spawn_strategy=sandboxed //xla/tools:run_hlo_module -- --platform=CPU /Volumes/git/xla-debug/error_module.mlir

 ** Running /Volumes/git/xla-debug/error_module.mlir**
INVALID_ARGUMENT: Invalid format from file extension: 'mlir'. Expected: hlo, txt, pb, or pbtxt
```

This had me confused for a while, thinking that StableHLO was not supported (enforced because I knew there is a special cli for `hlo-opt` in `//xla/mlir_hlo:mlir-hlo-opt` which works for StableHLO).

Ideally, the CLI would automatically figure out what the input type is, but this is not an easy change with the current structure of the code.
The new output of the CLI is as follows:
```
➜  xla git:(kn/docs-improvement) ✗ bazel run --spawn_strategy=sandboxed //xla/tools:run_hlo_module -- --platform=CPU /Volumes/git/xla-debug/error_module.mlir

 ** Running /Volumes/git/xla-debug/error_module.mlir**
INVALID_ARGUMENT: Invalid format from file extension: 'mlir'. Expected: hlo, txt, pb, or pbtxt
Did you forget to specify the `--input_format` flag in case of StableHLO?
```

I dont' like the way I check for the invalid argument status. If you have a better idea, let me know.

### DIFF
--- a/xla/tools/run_hlo_module_main.cc
+++ b/xla/tools/run_hlo_module_main.cc
@@ -297,6 +297,9 @@ int main(int argc, char** argv) {
       } else {
         failure_count++;
         std::cerr << result << "\n";
+        if (result.code() == absl::StatusCode::kInvalidArgument) {
+          std::cerr << "Did you forget to specify the `--input_format` flag?" << "\n";
+        }
       }
     }
 
--- a/xla/tools/run_hlo_module_main.cc
+++ b/xla/tools/run_hlo_module_main.cc
@@ -298,7 +298,7 @@ int main(int argc, char** argv) {
         failure_count++;
         std::cerr << result << "\n";
         if (result.code() == absl::StatusCode::kInvalidArgument) {
-          std::cerr << "Did you forget to specify the `--input_format` flag?" << "\n";
+          std::cerr << "Did you forget to specify the `--input_format` flag in case of StableHLO?" << "\n";
         }
       }
     }
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -23,15 +23,25 @@ with many other useful artifacts.
 
 The tool `run_hlo_module` operates on pre-optimization HLO, and by default
 bundles compilation, running and comparison with the reference interpreter
-implementation. For example, the usual invocation to run an input file
-`computation.hlo` on an NVIDIA GPU and to check it for correctness is:
+implementation. The tool can be used as follows:
 
-```
-$ run_hlo_module --platform=CUDA --reference_platform=Interpreter computation.hlo
+```bash
+$ run_hlo_module \
+    --input_format=[hlo|mhlo|pb|pbtxt|stablehlo] \
+    --platform=[CPU|CUDA|Interpreter] \
+    --reference_platform=[CPU|CUDA|Interpreter] \
+    path/to/[hlo|mhlo|pb|pbtxt|stablehlo]_module
 ```
 
-The `--input_format` flag can be used to specify the input type. For example, to
-run a StableHLO module specify `--input_format=stablehlo`.
+A typical usage employs the following flags:
+
+  * `--input_format=<string>` - The format (hlo, mhlo, pb, pbtxt, or
+stablehlo) of the input file.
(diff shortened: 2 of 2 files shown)


sdasgup3:
Thanks @kasper0406 for the issue!

Did you try the following switch which enable us to take stablehlo as input format. 

```
--input_format=[hlo|mhlo|pb|pbtxt|stablehlo]
```

kasper0406:
@sdasgup3 my proposal here is to make the `--input_format` flag easier to find.
I could not find the flags in the docs, and the CLI will not point you in the right direction, when you try to specify a StableHLO program to the CLI.

sdasgup3:
> @sdasgup3 my proposal here is to make the `--input_format` flag easier to find. I could not find the flags in the docs, and the CLI will not point you in the right direction, when you try to specify a StableHLO program to the CLI.

Acknowledged! I will update the docs and let you know. Thanks again!

xla/tools/run_hlo_module_main.cc
```diff
@@ -297,6 +297,9 @@ int main(int argc, char** argv) {
       } else {
         failure_count++;
         std::cerr << result << "\n";
+        if (result.code() == absl::StatusCode::kInvalidArgument) {
```
sdasgup3, line 300: IMO a better approach is to check at the begining if `opts.input_format` is one of `hlo`, `txt`, `pb`, `pbtxt`, `mhlo`, or `stablehlo`. That wasy we would never be in a situation where `result.ok()` is false because of unexpected file format.

kasper0406: I agree this would make the situation more clear. However, it will break any existing usages of the CLI with non-stablehlo/mhlo input: If the `input_format` flag is not explicitly set, the default now would be HLO, but the new behaviour would be an error. Is such a break allowed?
Maintaining the default value of HLO, would lead to the current situation.

An alternative is to try the StableHLO / MHLO conversions, and in case of errors keep falling through. I think this will cause debugging potential bad files to become annoying though.

sdasgup3: Hi @kasper0406 
I agree with your point. 

My proposal is to front load the population of `opts.input_format` using file extension from [here](https://github.com/openxla/xla/blob/fb6241ad510af945d3bf9f3ef8c7c27a3f1cc922/xla/tools/run_hlo_module.cc#L515) to somewhere [here](https://github.com/openxla/xla/blob/fb6241ad510af945d3bf9f3ef8c7c27a3f1cc922/xla/tools/run_hlo_module_main.cc#L203). 

And then check if the format belongs to the allowed values. If not, we error out early.

kasper0406: Ah cool, let me give that a go tomorrow!

abhigunj: @kasper0406  and @sdasgup3  

update -  The tool should automatically figure out what the input type is based on the input filename. I worked on this today and added this capability to the both `run_hlo_module` and `hlo-opt`. Now both the tools can accept MLIR module in text format (StableHLO or MHLO) and user don't need to pass additional `input_format` flag. 
I will submit the changes tomorrow. 

With that, @kasper0406  you don't need to spend more time on fixing this behavior :) Thank you for identifying this issue.

kasper0406: Ah, that's cool @abhigunj! I'll close out this one in favor of https://github.com/openxla/xla/pull/23213

kasper0406:
Closing out in favor of https://github.com/openxla/xla/pull/23213 which will automatically determine the input format.

PR #22782, opened 2025-02-17
[NFC] Move HLO-handling functions from runner to module util.

sergachev:

PR #22744, opened 2025-02-14
Add a tool to convert HLO files between different formats.

sergachev:

PR #22892, opened 2025-02-20
[NFC] Move more HLO-handling functions from runner to module util.

sergachev:


sergachev:
The version of the previous [PR](https://github.com/openxla/xla/pull/22782) that [landed](https://github.com/openxla/xla/commit/9bde2ca63bdc86b3805e573e69ffb4381e9eadb7) doesn't include LoadHloModuleAndArguments() and related code. I do want to move also this code out of the HLO runner to reuse it in another tool.

xla/tools/multihost_hlo_runner/functional_hlo_runner_test.cc
```diff
@@ -612,13 +612,13 @@ TEST_F(FunctionalHloRunnerTest, ReadHloUnoptimizedSnapshot) {
...
-          path_to_text_hlo, InputFormat::kUnoptimizedSnapshotProtoText));
+      LoadHloModuleAndArguments(path_to_text_hlo,
+                                InputFormat::kUnoptimizedSnapshotProtoText));
   // Read HloModuleAndArguments from binary dump.
   TF_ASSERT_OK_AND_ASSIGN(
       HloModuleAndArguments hlo_module_and_arguments_from_binary,
-      FunctionalHloRunner::LoadHloModuleAndArguments(
-          path_to_binary_hlo, InputFormat::kUnoptimizedSnapshotProtoBinary));
+      LoadHloModuleAndArguments(path_to_binary_hlo,
+                                InputFormat::kUnoptimizedSnapshotProtoBinary));
```
allanrenucci, line 621: We should move these tests into a new  hlo_module_util_test.cc file.

allanrenucci (changes requested):
Does the tool in #22744 need to support the `kSnapshotProtoBinary`, `kUnoptimizedSnapshotProtoBinary` and `kUnoptimizedSnapshotProtoText` formats? If not, maybe just call `ReadModuleFromHloTextFile`/`ReadModuleFromTextProtoFile`/`ReadModuleFromBinaryProtoFile` from the tool. Also, I see that the tool discards `HloModuleAndArguments::arguments` so I am not sure it makes sense to move `HloModuleAndArguments` to a shared library.

sergachev:
> Does the tool in #22744 need to support the `kSnapshotProtoBinary`, `kUnoptimizedSnapshotProtoBinary` and `kUnoptimizedSnapshotProtoText` formats? If not, maybe just call `ReadModuleFromHloTextFile`/`ReadModuleFromTextProtoFile`/`ReadModuleFromBinaryProtoFile` from the tool. Also, I see that the tool discards `HloModuleAndArguments::arguments` so I am not sure it makes sense to move `HloModuleAndArguments` to a shared library.

It doesn't have to, but, given that these dumps are easy to produce with --xla_dump_hlo_snapshots, why shouldn't they be supported too? Yes, arguments are not needed by this tool at the moment, but I don't see a problem in moving this code to make it accessible.

allanrenucci:
I think it is generally a good practice to only expose what is necessary as public API. `HloModuleAndArguments` is currently an implementation detail of the HLO runner and does not need to be part of `hlo_module_util`. If the tool needs to evolve and support arguments and snapshots, we can always revisit.

PR #22901, opened 2025-02-20
Remove transpose from reshape builder

shraiysh:
Reshape builder unnecessarily accepts two arguments:

 1. `dimensions` which permutation of the  layoyt of the shape.
 2. `new_sizes` which are the new dimensions of the shape.

This is confusing, and the we can simply remove the support for permutation of layout - as it is just a transpose and such a transpose can be easily constructed wherever required.

### DIFF
--- a/docs/operation_semantics.md
+++ b/docs/operation_semantics.md
@@ -2241,30 +2241,24 @@ and the [`Collapse`](#collapse) operation.
 
 Reshapes the dimensions of an array into a new configuration.
 
-<b> `Reshape(operand, new_sizes)` </b>
-<b> `Reshape(operand, dimensions, new_sizes)` </b>
+<b> `Reshape(operand, dimensions)` </b>
 
 Arguments    | Type           | Semantics
 ------------ | -------------- | ---------------------------------------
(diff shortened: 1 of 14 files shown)


shraiysh:
The tensorflow CI failure is because of no storage on device, and is not related to this PR.

akuegel:
If this is changed, the operation semantics should be changed as well:
https://openxla.org/xla/operation_semantics#reshape

In general I think it is probably a good idea to change this, as for stablehlo we already don't allow the implicit transpose:
https://github.com/openxla/stablehlo/blob/main/docs/spec.md#reshape

shraiysh:
>If this is changed, the operation semantics should be changed as well:
https://openxla.org/xla/operation_semantics#reshape

Can you tell me where the source for this is?

akuegel (approved):
Thank you!

shraiysh:
Are there internal failures with this PR? Let me know if there is anything I can do to help merge this.

akuegel:
There were internal adjustments needed. I did the last ones yesterday, but I need approval for them and unfortunately so far I didn't get it. Hopefully I can merge it either later today or tomorrow.

PR #22930, opened 2025-02-21
Initialize num_slices_ to 0 in Heap Simulator

apivovarov:
Ensure `num_slices_` class member is explicitly initialized to 0 in `SliceTimeAllPermutationIterator` and `SliceTimePreferredPermutationIterator` to prevent potential uninitialized variable issues.

### DIFF
--- a/xla/service/heap_simulator/heap_simulator.cc
+++ b/xla/service/heap_simulator/heap_simulator.cc
@@ -1387,7 +1387,7 @@ class SliceTimeAllPermutationIterator : public SliceTimePermutationIterator {
  private:
   SliceTimeAllPermutationIterator() = default;
 
-  int64_t num_slices_;
+  int64_t num_slices_ = 0;  // Initialize to 0
   bool done_ = true;
   std::vector<int64_t> permutation_;
 };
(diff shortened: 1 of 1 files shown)


reedwm:
I'm not familiar with the HeapSimulator, but skimming the code, it seems `num_slices_` is always assigned in the constructor.

sparc1998:
I don't know how to do an official review, but LGTM. As Reed pointed out, num_slices_ is already assigned in the constructor, so this should be a no-op.

apivovarov:
the default constructor `SliceTimeAllPermutationIterator()` is private.
`num_slices_` is initialized in the explicit public constructor.

`std::vector<int64_t> permutation_` is implicitly initialized in place as an empty vector (size 0). 
Since `permutation_` should always have a size equal to `num_slices_`, it makes sense to initialize `num_slices_` in place for consistency.

While explicitly setting a default value 0 for `num_slices_` is not strictly necessary, it improves safety and clarity, ensuring that it never holds an uninitialized value.

apivovarov:
Additional arguments:
- The default constructor is not deleted; it is marked as private.

- A private constructor can still be accessed within the class itself and by friend classes or friend functions

Strictly speaking, the default private constructor partially initializes class members:
- `permutation_` is implicitly initialized (as an empty vector).
- `done_` is initialized in-place (to true).
- `num_slices_` is not initialized at all.

So, ensuring `num_slices_` has a default 0 value can help prevent potential issues in future modifications.

reedwm:
Fair point. Since @sparc1998 LGTMed, I'll approve so this gets merged.

reedwm (approved):

PR #22950, opened 2025-02-21
Fix help text for `xla_cpu_enable_fast_min_max` and set `minmax_propagate_nan` on CPU

Zentrik:
It is my understanding that disabling `xla_cpu_enable_fast_min_max` causes NaNs to be always propagated and otherwise they might not be. For reference, see https://github.com/openxla/xla/blob/99a2ecdb2c9ec0faec6e5a762831908f484194db/xla/xla.proto#L165-L170 and the uses of this flag, e.g. at https://github.com/openxla/xla/blob/99a2ecdb2c9ec0faec6e5a762831908f484194db/xla/service/cpu/ir_emitter.cc#L1541-L1543 where the flag being enabled causes NaNs to not always be propagated.

Additionally, I set the `minmax_propagate_nan` on the CPU to match the `xla_cpu_enable_fast_min_max` flag.


xla/service/cpu/cpu_compiler.cc
```diff
@@ -834,9 +834,9 @@ absl::Status CpuCompiler::RunHloPassesAfterLayoutAssn(
     options.set_is_layout_sensitive(true);
     options.set_supports_non_canonical_dots(false);
     options.set_enable_dot_strength_reduction(false);
-    // TODO(b/209827141): XLA:CPU doesn't propagate NaN through min/max, but
-    // other platforms do, so it should be changed.
-    options.set_minmax_propagate_nan(false);
+    // "slow" minmax means we propagate nan.
+    options.set_minmax_propagate_nan(
+        !module->config.debug_options().xla_cpu_enable_fast_min_max());x());
```
Zentrik, line 839:
```suggestion
        !module->config.debug_options().xla_cpu_enable_fast_min_max());
```

dimitar-asenov:
@ezhulenev Could you please review this PR?

ezhulenev (approved):

reedwm:
@Zentrik can you resolve conflicts?

PR #22968, opened 2025-02-21
Add ScheduleAwareCollectiveOpsCSE to the GPU pipeline

sfvaroglu:
This PR adds ScheduleAwareCollectiveOpsCSE to the GPU pipeline before scheduling, as the pass is not schedule-aware. This is a second attempt, as the previous [PR](https://github.com/openxla/xla/pull/22123/) was reverted due to an out-of-memory error. This PR introduces a flag to control the distance threshold.


xla/debug_options_flags.cc
```diff
@@ -203,6 +203,8 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_experimental_pipeline_parallelism_opt_level(
       DebugOptions::PIPELINE_PARALLELISM_OPT_LEVEL_DISABLE);
 
+  opts.set_xla_gpu_experimental_collective_cse_distance_threshold(100);
```
reedwm, line 206: Let's set the default to zero for now. This allows you to experiment with the flag, while still ensuring it doesn't break any existing models.

I suspect it may not be feasible to set this to nonzero by default unless we make significant changes, like moving the pass to run post-scheduling and modifying the pass accordingly. But I'm not sure yet.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -980,6 +981,11 @@ absl::Status RunCollectiveOptimizationPasses(
 
   collectives_pipeline.AddPass<AllGatherBroadcastReorder>();
 
+  collectives_pipeline.AddPass<ScheduleAwareCollectiveOpsCSE>(
```
reedwm, line 984: Only call the pass if the distance threshold is non-zero.

xla/xla.proto
```diff
@@ -193,7 +193,7 @@ message DebugOptions {
   // When true, XLA:CPU all LLVM kernels entry points will be prefixed with the
   // module name. They will also be converted to C style names. This is required
   // for AOT compilation.
-  bool xla_cpu_generate_unique_c_style_kernel_entry_points = 372;
+  bool xla_cpu_generate_unique_c_style_kernel_entry_points = 374;
```
reedwm, line 196: Don't change numbers of existing flags, instead give the new flag a new number.

xla/tools/hlo_opt/compiled_opt_lib.cc
```diff
@@ -167,6 +168,7 @@ void CompiledOptProvider::RegisterSharedHardwareSpecificPasses() {
   RegisterPass<ScatterDeterminismExpander>();
   RegisterPass<ScatterSimplifier>();
   RegisterPass<ScatterSliceSimplifier>();
+  RegisterPass<ScheduleAwareCollectiveOpsCSE>(100, false);
```
reedwm, line 171: Since I suggested to set the default to zero, I think we should remove the pass from hlo_opt.

Unless we should check the DebugOptions here and get the xla_gpu_experimental_collective_cse_distance_threshold flag, maybe. @abhigunj, do you know if it's possible and recommended to check DebugOptions in hlo_opt?

reedwm (changes requested):

xla/tools/hlo_opt/compiled_opt_lib.cc
abhigunj: could you move the pass registration to gpu_opt.cc for now? 
gpu_opt.cc has access to DebugOptions.
`module.config().debug_options().xla_gpu_experimental_collective_cse_distance_threshold`. 

Let me know if that works.
compiled_opt_lib.cc can't accept DebugOptions right now, need some change.

xla/debug_options_flags.cc
sfvaroglu: Wouldn't defaulting to 0 disable the pass? Any concerns with setting it to a low value, perhaps 20?

reedwm: I think the pass should be disabled by default. I worry about regressions causing OOMs if not set to zero.

xla/tools/hlo_opt/gpu_opt.cc
```diff
@@ -145,6 +146,15 @@ class GpuOptProvider : public CompiledOptProvider {
     RegisterPass<gpu::ReductionDimensionGrouper>();
     RegisterPass<gpu::ReductionLayoutNormalizer>();
     RegisterPass<gpu::SanitizeConstantNames>();
+    if (module.config()
```
abhigunj, line 149: LGTM for hlo-opt change.

Nit: create a local var `debug_config`, we are using more than once and other passes will also need it. 
auto debug_config = module.config().debug_options();

reedwm (approved):

abhigunj (approved):

PR #22970, opened 2025-02-21
Fix bug in post order traversal of computation instructions

shraiysh:
While creating post order traversal, an instruction may have a user outside the computation. This is the case when we are constructing new instructions to store in replacements for cloning the computation later. This user should be ignored. Added test for the same.

Because of this, functions like `ToString()`, and
`GetUniqueGteInstruction()` encounter errors. They rely on post-order traversal to have all the instructions.

### DIFF
--- a/xla/hlo/ir/hlo_computation.cc
+++ b/xla/hlo/ir/hlo_computation.cc
@@ -743,7 +743,12 @@ std::vector<HloInstruction*> HloComputation::MakeInstructionPostOrder(
   dfs_stack_scratch.reserve(instruction_count());
 
   for (const auto& instruction : instructions()) {
-    if (instruction->users().empty()) {
+    // We don't consider users outside any computation as real users. This can
+    // happen when creating new instructions for replacement when cloning
+    // computations.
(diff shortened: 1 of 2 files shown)


shraiysh:
The JAX failures are unrelated.

xla/hlo/ir/hlo_computation.cc
```diff
@@ -743,7 +743,10 @@ std::vector<HloInstruction*> HloComputation::MakeInstructionPostOrder(
   dfs_stack_scratch.reserve(instruction_count());
 
   for (const auto& instruction : instructions()) {
-    if (instruction->users().empty()) {
+    if (instruction->users().empty() ||
```
akuegel, line 746: I would expect that the first condition `instruction->users().empty()` is now redundant with the second?
And I think it would be good to have a comment here, maybe something like this:
// We don't consider users outside any computation as real users. This can happen when creating new instructions for replacement when cloning computations.

Same below.

shraiysh: Yes, that seems accurate. Fixing this. Thanks.

akuegel (approved):

PR #22976, opened 2025-02-22
[gpu][convert-peeling] Add convert peeling

shraiysh:
This patch adds convert peeling optimization. If the while loop access a buffer, one slice at a time, and converts it inside the while body, then there is a scope for this optimization.

The conditions are:

 * The dynamic-slice operation must have exactly one variable index (lets call that index k).
 * The dynamic-slice operation must be a contiguous slice (i.e. for all indices i>k, we should have that dimension_of_buffer[i]=slice_size[i] and offset[i]=0.)
 * The dynamic-slice operation must cover the entire buffer. This means that
   * The variable index k must have the monotonic range [0, dimension[k]) and each of those values must be taken once (step=1).
   * For all indices i < k, we should have that dimension[i]=slice_size[i]=1 and offset[i]=0.
 * Obviously, the pattern: dynamic-slice followed by a convert operation (allowing no-op reshapes).
 * The buffer should be from the parameter, unmodified, and the buffer should go to the results, unmodified.
 * There should be exactly two uses of the buffer inside the while loop: convert operation and the root instruction.

If these conditions are met, then we can peel the convert outside the while loop:
 * We change the while loop body/condition signature to accept the input of destination type (from the convert operation).
 * We add the convert operation outside the while loop.
 * Any user of the previous results of while operation are adjusted.

### DIFF
--- a/xla/service/gpu/transforms/BUILD
+++ b/xla/service/gpu/transforms/BUILD
@@ -3416,3 +3416,33 @@ xla_cc_test(
         "@com_google_googletest//:gtest_main",
     ],
 )
+
+cc_library(
+    name = "while_loop_convert_peeler",
+    srcs = ["while_loop_convert_peeler.cc"],
+    hdrs = ["while_loop_convert_peeler.h"],
+    deps = [
+        "//xla/hlo/ir:hlo",
+        "//xla/hlo/pass:hlo_pass",
+        "//xla/hlo/analysis:while_loop_analysis",
+        "//xla/hlo/utils:hlo_query",
+        "//xla:literal_util",
+        "//xla/service:value_range",
+        "//xla/tests:test_utils",
+    ],
+)
+
+xla_cc_test(
+    name = "while_loop_convert_peeler_test",
+    srcs = ["while_loop_convert_peeler_test.cc"],
+    deps = [
+        ":while_loop_convert_peeler",
+        "@com_google_googletest//:gtest_main",
+        "//xla/tests:hlo_runner_agnostic_test_base",
+        "//xla/tests:test_utils",
+        "//xla/service:platform_util",
+        "//xla/service:hlo_runner",
+        "//xla/hlo/testlib:pattern_matcher_gmock",
+        "//xla/service:pattern_matcher",
+    ],
+)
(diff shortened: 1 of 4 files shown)


shraiysh:
This PR currently only adds support for dynamic-slice -> convert patterns. I will separately add support for dynamic-slice-fusion -> convert pattern for easier reviews, and smaller PRs.

This also does not have exhaustive testing especially for the cases where the transformation should not be applied. However the logic is written. I will keep adding relevant testcases, but I wanted to start the review process.

shraiysh:
@rosiezou Added testcases to cover everything. I know the patch has enlarged now, but majority of it (~60%) is just tests. Apologies for such a big patch, but unfortunately, I don't think I can make it smaller. Open to suggestions about it though.

rosiezou:
Thanks for the PR @shraiysh. Taking a look this week.

shraiysh:
Rosie, I think we do not need this pass. I will ping you about this on GChat, but I don't think this needs review. I will close this after discussing it with you.

shraiysh:
For context on why this is closing without merging:

We do not need that pass because of convert fusion. The module without this pass looks like the following:
```
Wrap Textwhile(buffer: f32) {
  ds = dynamic-slice(buffer)   // THIS IS IN CONVERT FUSION
  convert = f16 convert(ds)      // THIS IS IN CONVERT FUSION
  gemm = gemm(convert, ...)
}
```
Without this pass, the ds and convert get fused as a convert fusion, and the ds operation does not materialize. We do one read and write the converted value to a new buffer directly. This new buffer is then used for the gemm.

With this pass, we would have the convert peeled outside. It would again have one pass over the buffer to read the values, and write the converted values to a new buffer. The dynamic-slice(s) inside the loop will now be fused with gemms, and the gemm will read the values from the converted buffer.

In both these cases, the dynamic-slice is not materializing. The only difference is that, without this pass, we have `trip_count` number of convert kernel dispatches, each for `buffer_size/trip_count` size, but with this pass, we have only one convert kernel dispatch for `buffer_size` size. On the other hand, we now have increased memory pressure as both the original value and the converted value are alive for the entire duration of the while loop.

PR #23051, opened 2025-02-25
[GPU][NFC] Cleanup GPU compiler pipeline definition.

sergachev:
Rename the pipeline and update/remove misleading comments as it runs before scheduling and all the related passes thus well before IR emission.

### DIFF
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -1346,9 +1346,9 @@ cc_library(
 )
 
 cc_library(
-    name = "prepare_hlo_for_ir_emitting_pipeline",
-    srcs = ["prepare_hlo_for_ir_emitting_pipeline.cc"],
-    hdrs = ["prepare_hlo_for_ir_emitting_pipeline.h"],
+    name = "pre_scheduling_copy_insertion_pipeline",
+    srcs = ["pre_scheduling_copy_insertion_pipeline.cc"],
+    hdrs = ["pre_scheduling_copy_insertion_pipeline.h"],
     deps = [
         "//xla:xla_proto_cc",
         "//xla/hlo/analysis:hlo_dataflow_analysis",
@@ -1402,7 +1402,7 @@ cc_library(
         ":kernel_reuse_cache",
         ":matmul_utils",
         ":metrics",
-        ":prepare_hlo_for_ir_emitting_pipeline",
+        ":pre_scheduling_copy_insertion_pipeline",
         ":reduction_utils",
         ":runtime_intrinsics",
         ":stream_executor_util",
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -171,7 +171,7 @@ limitations under the License.
 #include "xla/service/gpu/model/gpu_cost_model_stats_collection.h"
 #include "xla/service/gpu/model/gpu_hlo_cost_analysis.h"
 #include "xla/service/gpu/model/sol_gpu_cost_model_stats_collection.h"
-#include "xla/service/gpu/prepare_hlo_for_ir_emitting_pipeline.h"
+#include "xla/service/gpu/pre_scheduling_copy_insertion_pipeline.h"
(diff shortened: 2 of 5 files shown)


xla/service/gpu/BUILD
```diff
@@ -1345,29 +1345,6 @@ cc_library(
     ],
 )
 
-cc_library(
```
bchetioui, line 1348: Any particular reason why you decided to re-inline the pipeline in `gpu_compiler.cc`? That file is pretty massive, and the pipeline is standalone. How about we just keep it in a separate file?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1274,11 +1277,69 @@ absl::Status GpuCompiler::RunCollectiveScheduleLinearizerPasses(
   return pipeline.Run(hlo_module).status();
 }
 
+absl::Status GpuCompiler::RunCopyInsertionAndFusionPipeline(
```
bchetioui, line 1280: This pipeline should still have a docstring explaining why it's necessary, and when it's expected to run.

bchetioui (commented):
I wonder if we're really not enforcing any important invariant here. I'm not 100% sure if the name change is actually an improvement.

Perhaps @akuegel knows---this was added a long time ago.

xla/service/gpu/BUILD
sergachev: I see no other uses of this pipeline anywhere. Now it takes 2 extra files and an empty method GpuCompiler::PrepareHloModuleForIrEmitting(). IMO moving it to the compiler class improves readability, but I wanted to hear your reasons to outline it.

xla/service/gpu/gpu_compiler.cc
sergachev: I actually don't see a reason to keep these passes as a separate pipeline.

xla/service/gpu/BUILD
bchetioui: In general, I think moving pipelines to dedicated files makes sense, simply because they're meant to delineate well-scoped, independent combinations of passes. Ideally, we should also have tests for them that ensure that they perform the adequate combinations of transformations---which is easier to do if they have their own file as well.

The main reason for outlining this particular one in the first place, though, was to prepare for changes that would involve calling this particular pipeline from within other (autotuning) passes, though. While it wasn't done in the end, keeping pipelines in `gpu_compiler.cc` also means that the functionality can not be reused in passes without creating circular dependencies.

akuegel:
> I wonder if we're really not enforcing any important invariant here. I'm not 100% sure if the name change is actually an improvement.
> 
> Perhaps @akuegel knows---this was added a long time ago.

Running CopyFusion pass is definitely needed for correctness and should run pretty late, shortly before scheduling. We run CopyFusion again after scheduling but only to potentially remove copies that are not needed after the schedule has been finalized.
So I am also not sure whether the new name is really better. But I guess the old name was not perfect either.

sergachev:
> Running CopyFusion pass is definitely needed for correctness and should run pretty late, shortly before scheduling. We run CopyFusion again after scheduling but only to potentially remove copies that are not needed after the schedule has been finalized.

CopyFusion or CopyInsertion? I thought only the latter is really needed for correctness.

> So I am also not sure whether the new name is really better. But I guess the old name was not perfect either.

Should we call it PreSchedulingCopyInsertion? Do you have better ideas?

akuegel:
> > Running CopyFusion pass is definitely needed for correctness and should run pretty late, shortly before scheduling. We run CopyFusion again after scheduling but only to potentially remove copies that are not needed after the schedule has been finalized.
> 
> CopyFusion or CopyInsertion? I thought only the latter is really needed for correctness.
> 
> > So I am also not sure whether the new name is really better. But I guess the old name was not perfect either.
> 
> Should we call it PreSchedulingCopyInsertion? Do you have better ideas?

Ah, sorry, of course I meant CopyInsertion :)
PreSchedulingCopyInsertion sounds good to me.

bchetioui (approved):

PR #23056, opened 2025-02-25
Remove std::move for trivially copyable types

apivovarov:
Changes:
- Removed unnecessary std::move calls for some trivially_copyable classes
- Literal::CopyFrom expects a reference (&), not an r-value reference (&&). Removed std::move on the first parameter.

### DIFF
--- a/xla/hlo/evaluator/hlo_evaluator.cc
+++ b/xla/hlo/evaluator/hlo_evaluator.cc
@@ -318,11 +318,11 @@ std::optional<ParamIndexAndValue> TryParsingInstructionAsParameterAndInteger(
   }
   std::optional<DynamicOrStaticInteger> integer_value =
       GetInstructionValueAsInteger(instruction, precomputed_analyses);
-  result.value = std::move(integer_value);
+  result.value = integer_value;
   if (!result.IsValid()) {
     return std::nullopt;
   }
-  return std::optional<ParamIndexAndValue>(std::move(result));
+  return result;
 }
 
 // Represents the while loop condition comparison.
@@ -377,8 +377,7 @@ std::optional<WhileCondComparisonOrNoOp> PatternMatchLoopCondComparison(
   if (!lhs.has_value() || !rhs.has_value()) {
     return std::nullopt;
   }
-  return WhileCondComparison{comparison->comparison_direction(),
-                             *std::move(lhs), *std::move(rhs)};
+  return WhileCondComparison{comparison->comparison_direction(), *lhs, *rhs};
 }
 // Finds the while loop condition comparison by matching the loop condition root
 // with known patterns.
@@ -1707,7 +1706,7 @@ absl::Status HloEvaluator::HandleTuple(const HloInstruction* tuple) {
     CHECK(new_result.IsDetermined(visitor_shape_index_));
     Literal literal;
     TF_RETURN_IF_ERROR(
-        literal.CopyFrom(std::move(new_result),
+        literal.CopyFrom(new_result,
                          /*dest_shape_index=*/visitor_shape_index_,
                          /*src_shape_index=*/visitor_shape_index_));
     SetEvaluatedLiteralFor(tuple, std::move(literal));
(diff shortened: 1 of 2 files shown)


xla/hlo/evaluator/hlo_evaluator.cc
```diff
@@ -318,11 +324,11 @@ std::optional<ParamIndexAndValue> TryParsingInstructionAsParameterAndInteger(
   }
   std::optional<DynamicOrStaticInteger> integer_value =
       GetInstructionValueAsInteger(instruction, precomputed_analyses);
-  result.value = std::move(integer_value);
+  result.value = integer_value;
```
apivovarov, line 327: removing std::move because DynamicOrStaticInteger is trivially copyable, its copy and move operations behave identically because it has no special resources

xla/hlo/evaluator/hlo_evaluator.cc
```diff
@@ -318,11 +324,11 @@ std::optional<ParamIndexAndValue> TryParsingInstructionAsParameterAndInteger(
...
   if (!result.IsValid()) {
     return std::nullopt;
   }
-  return std::optional<ParamIndexAndValue>(std::move(result));
+  return result;
```
apivovarov, line 331: - removing std::move because ParamIndexAndValue is trivially copyable.
- removing std::optional too because wrapping result with std::optional is done automatically.

xla/hlo/evaluator/hlo_evaluator.cc
```diff
@@ -377,8 +383,7 @@ std::optional<WhileCondComparisonOrNoOp> PatternMatchLoopCondComparison(
   if (!lhs.has_value() || !rhs.has_value()) {
     return std::nullopt;
   }
-  return WhileCondComparison{comparison->comparison_direction(),
-                             *std::move(lhs), *std::move(rhs)};
+  return WhileCondComparison{comparison->comparison_direction(), *lhs, *rhs};
```
apivovarov, line 386: lhs and rhs are ParamIndexAndValue which is trivially copyable - removing std::move

xla/hlo/evaluator/hlo_evaluator.cc
```diff
@@ -1696,7 +1701,7 @@ absl::Status HloEvaluator::HandleTuple(const HloInstruction* tuple) {
     CHECK(new_result.IsDetermined(visitor_shape_index_));
     Literal literal;
     TF_RETURN_IF_ERROR(
-        literal.CopyFrom(std::move(new_result),
+        literal.CopyFrom(new_result,
```
apivovarov, line 1704: Literal::CopyFrom expects a reference (&), not an r-value reference (&&). Removing std::move

xla/stream_executor/kernel.h
```diff
@@ -220,9 +223,7 @@ class Kernel {
       ThreadDim threads, size_t dynamic_shared_memory_bytes) const = 0;
 
   const KernelMetadata &metadata() const { return metadata_; }
-  void set_metadata(KernelMetadata metadata) {
-    metadata_ = std::move(metadata);
-  }
+  void set_metadata(KernelMetadata metadata) { metadata_ = metadata; }
```
apivovarov, line 226: removing std::move because KernelMetadata is trivially copyable.

xla/hlo/evaluator/hlo_evaluator.cc
```diff
@@ -255,6 +255,9 @@ struct DynamicOrStaticInteger {
   }
 };
 
+static_assert(std::is_trivially_copyable_v<DynamicOrStaticInteger>,
+              "DynamicOrStaticInteger is expected to be trivially copyable");
```
reedwm, line 259: I personally don't think it's worth static_asserting these types are trivially copyable. Without the context of this PR, it's non-obvious to someone reading the code why there is a static assert, and in the very-unlikely case someone does add a non-trivially-copyable field to these types, the only downside will be negligible lower performance since we don't std::move it.

reedwm (changes requested):

reedwm (approved):

PR #23254, opened 2025-02-28
googletest: Update to HEAD.

pizzud:
Doing so allows picking up new testonly library target and test case shuffling feature.


pizzud:
Good to know that this passes tests, but I think it's not happening until we're on bzlmod unless we want to try to pull in the Fuschia SDK. Sigh. I'll try again once we're there.

https://github.com/google/googletest/issues/4472 is relevant.

PR #23271, opened 2025-02-28
[XLA] Improve GPU memory limit handling and shape size calculation

zhenying-liu:
When running MaxText Llama2-7b with FP16 optimizer state offloading, an out of memory failure occurred. The root cause was the int64_t memory limit being incorrectly calculated and interpreted as a large uint64_t memory limit close to UINT64_MAX, leading to overly aggressive buffer allocation by the LHS.

This changelist addresses the out of memory failure by correctly handling the uint64_t memory limit and accurately calculating the device memory limit, excluding host memory space.

- Change memory limit type from int64_t to uint64_t to prevent negative values and better represent memory sizes.

- Add memory space filtering to exclude host memory when calculating input/output sizes that impact GPU memory usage. This ensures we only count buffers that actually reside in device memory.

- Replace direct GetSizeOfShape() calls with ShapeSizeBytesFunction() wrapper, which provides:
  * Consistent shape size calculation across the codebase
  * Optional memory space filtering capability
  * Proper handling of dynamic shapes and their metadata


frgossen (approved):
Thanks for fixing this!

PR #23314, opened 2025-03-03
[debug_options] Dump DebugOptions while compiling.

shraiysh:
This patch adds VLOG(2) for dumping debug options while compiling. This is a reliable way to retrieve the debug options based on the log.


shraiysh:
The failures in jobs are unrelated.

frgossen (approved):

shraiysh:
I think we can revert this @frgossen . I just realized that this is already being logged at level 1 in https://github.com/shraiysh/xla/blob/e7e24b97ae32374a3e46caf7f3dcf1118789741e/xla/service/gpu/gpu_compiler.cc#L518-L525

frgossen:
> I think we can revert this @frgossen . I just realized that this is already being logged at level 1 in https://github.com/shraiysh/xla/blob/e7e24b97ae32374a3e46caf7f3dcf1118789741e/xla/service/gpu/gpu_compiler.cc#L518-L525

Reverting it

PR #23347, opened 2025-03-04
[gpu] Allow explicitly setting slice_index in se_gpu_pjrt_client

gspschmid:
Allows overriding the slice index used by se_gpu_pjrt_client.

More explicit control over which slice a device ends up in is desirable:
- Various parts of the ecosystem equate slices with "devices communicating via fast interconnect". With the arrival of NVL72 we want devices managed by multiple hosts to form a single slice.
- For debugging purposes it can be useful to allow devices on the same host (managed in separate processes) to be treated as different slices. For example, [Orbax](https://github.com/google/orbax)'s local checkpointing presumes the existence of at least two slices, so overriding the boot id will allow us to test local checkpointing on a single host.

(Companion PR in JAX: https://github.com/jax-ml/jax/pull/26906)


hawkinsp (commented):
I'm commenting on this in chat also, but I'm wondering if we should find a more direct way to communicate the Nvlink topology to XLA. This feels a bit indirect.

gspschmid:
@ezhulenev @hawkinsp Updated to allow control over slice_index directly, rather than boot_id.

hawkinsp (approved):

PR #23355, opened 2025-03-04
[ROCm] Add rocm_base and rocm_gcc configs

mmakevic-amd:
Changed `rocm` config to `rocm_base`, and added `rocm_gcc`  so we can have gcc specific options. Since a lot of scripts still depend on `--config=rocm` I have temporarily left it in .bazelrc. It will be removed once I transition everything to use `rocm_gcc`


alekstheod (approved):
ltgm

ddunl (approved):

PR #23380, opened 2025-03-04
[ROCm] Add absolute paths to includes

charleshofer:
Includes will look for files in the `/opt/rocm-x.x.x/llvm/lib/clang/19/include` or `/opt/rocm-x.x.x/lib/llvm/lib/clang/19/include` directories, but these don't get added to Bazel's include directories. Configure Bazel to look in those directories.


beckerhe (approved):

charleshofer:
Going to look for a better solution an fixing the compiler path for this to make sure we're doing the right thing. This PR is a little hacky.

PR #23386, opened 2025-03-05
[debug_options] Printing all the fields in debug options

shraiysh:
This patch prints all the values of the debug options while dumping it under VLOG. This is specifically required for boolean fields which have the default value set to true in `xla/debug_options_flags.cc`. These values will not be printed in `DebugString()` if `XLA_FLAGS` overrides it to `false`.


shraiysh:
Ping for review

frgossen (approved):

PR #23400, opened 2025-03-05
Fix complex_unary_op_test_cpu on macOS Apple

### LINKED ISSUE #19824: //xla/tests:complex_unary_op_test_cpu fails on macOS Apple Silicon
This test fails //xla/tests:complex_unary_op_test_cpu on my Mac mini.

I hacked the test up a bit to simplify things.
```
[ RUN      ] ComplexUnaryOpTest.Log1pTest
2024-11-24 17:54:52.487071: I xla/tests/literal_test_util.cc:56] expected: c64[2] c64[2] {(inf, -2.3561945), (-0.5, -0.5)}
2024-11-24 17:54:52.487089: I xla/tests/literal_test_util.cc:58] actual:   c64[2] c64[2] {(inf, -2.3561945), (-0.5, -0)}
```

This is using the following input:
```c++
 { { -min, -min }, { -min, -min }, 0x1p+125f }
```

After a bit of further investigation, I was able to determine that the root cause is that the implementation of `atan2f` depends on subnormal support even if the inputs and outputs are not subnormal.

```c++
#include <cmath>
#include <cfloat>
#include <cstdio>
#include <fenv.h>

static void doit() {
  volatile const float a = -FLT_MIN;
  volatile const float b = 1.0f;
  volatile float c = atan2f(a, b);
  c *= 0x1p+125f;
  printf("%.19g\n", c);
}

int main (int argc, char *argv[]) {
  doit();
  fesetenv(FE_DFL_DISABLE_DENORMS_ENV);
  doit();
  return 0;
}
```

pearu:
As in the title.

Fixes https://github.com/openxla/xla/issues/19824

### DIFF
--- a/xla/tests/complex_unary_op_samples.h
+++ b/xla/tests/complex_unary_op_samples.h
@@ -48,7 +48,11 @@ struct Log1p {
       const T pi3_4 = 2.3561945f;
       const T zero = 0.0f;
       const T inf = std::numeric_limits<T>::infinity();
+#ifdef __aarch64__
+      const T min = std::nextafter(std::numeric_limits<T>::min(), T(1));
+#else
       const T min = std::numeric_limits<T>::min();
+#endif
       const T max = std::numeric_limits<T>::max();
       const TableType table{
           // clang-format off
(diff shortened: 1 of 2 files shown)


pearu:
@majnemer please review.

majnemer (approved):
LGTM

pearu:
@majnemer could give information about [feedback/copybara](http://cl/745956828) check failure?

PR #23412, opened 2025-03-05
Additional Activations for FP8 Graph Convolutions

philipphack:
Extends the functionality of graph-based FP8 convolutions to include ELU, ReLU6 and Leaky ReLU activations. Also allows instructions to have multiple users in the fused graph.

### DIFF
--- a/xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
+++ b/xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
@@ -66,6 +66,60 @@ namespace {
 
 namespace m = match;
 
+// Captures multiple HloInstruction pointers and verifies that their target
+// is identical.
+//
+// Example:
+// Pattern cos(x) / sin(x) with cos and sin intended to operate on the same
+// HloInstruction:
+//  UniqueHloInstruction x;
+//  bool m = Match(
+//      instr, m::Divide(m::Cos(m::Op().WithPredicate(x.CaptureOrVerifyFn())),
+//                       m::Sin(m::Op().WithPredicate(x.CaptureOrVerifyFn()))));
+// m is true and x.Instr() returns an HloInstruction pointer to the operand of
+// cosine and sine iff HloInstruction *instr points to a division of a cosine by
+// a sine that operate on the same instruction.
+//
+// TODO(philipphack): Move to shared header or integrate into pattern matcher.
+class UniqueHloInstruction {
+ public:
+  UniqueHloInstruction()
+      : is_set_(false),
+        instr_(nullptr),
+        capture_or_verify_([this](const HloInstruction* instr) -> bool {
+          return CaptureOrVerify(const_cast<HloInstruction*>(instr));
+        }) {}
+  HloInstruction* Instr() const { return instr_; }
+  void SetInstr(HloInstruction* instr) {
+    is_set_ = true;
+    instr_ = instr;
+  }
+
+  // Stores instr when invoked the first time. Otherwise, compares instr to the
+  // stored value and sets the stored value to nullptr if the comparison fails.
+  bool CaptureOrVerify(HloInstruction* instr) {
+    if (is_set_ && instr != instr_) {
+      instr_ = nullptr;
+    }
+    if (!is_set_) {
+      is_set_ = true;
+      instr_ = instr;
+    }
+    return instr_;
+  }
+
+  // Returns a std::function for capturing or verifying an instruction using
+  // WithPredicate.
+  std::function<bool(const HloInstruction*)> CaptureOrVerifyFn() const {
+    return capture_or_verify_;
+  }
+
+ private:
+  bool is_set_;
+  HloInstruction* instr_;
+  std::function<bool(const HloInstruction*)> capture_or_verify_;
+};
+
 bool IsConvCustomCall(const HloInstruction* instr) {
   return HloPredicateIsOp<HloOpcode::kCustomCall>(instr) &&
          (instr->custom_call_target() == kCudnnConvForwardCallTarget ||
@@ -361,33 +415,50 @@ absl::StatusOr<bool> FuseConvAlpha(HloComputation* comp) {
 
 // The format of the serialized graph describing a sequence of ops fused
 // into the cuDNN convolution Custom Call is
-// "UID:[output_type]conv();UID[output_type]:op_name(operand
-// UID);UID:[output_type]op_name(operand UID);..." with the convolution assumed
-// to be the first op in the graph. Operand UIDs identifying ops outside the
-// serialized graph are elided. Currently, multiplication and division by a
-// broadcast scalar, addition of a matrix bias, the application of a ReLU
-// activation and the calculation of the maximum of the absolute value are
-// supported.
+// "UID:[output_type]conv();UID:[output_type]op_name(operand
+// UIDs);UID:[output_type]op_name(operand UIDs);..." with the convolution
+// assumed to be the first op in the graph. Operand UIDs identifying ops outside
+// the serialized graph are elided. Currently, multiplication and division by a
+// broadcast scalar, addition, the application of various activations and the
+// calculation of the maximum of the absolute value are supported.
 class GraphString {
  public:
   GraphString() = default;
 
   bool AppendOp(std::string op_name, HloInstruction* op,
                 std::vector<HloInstruction*> operands = {}) {
-    std::optional<int64_t> operand_uid;
-    int num_operands_in_graph = 0;
+    return AppendOp(op_name, op, op->shape().element_type(), operands);
+  }
+
+  bool AppendOp(std::string op_name, HloInstruction* op,
+                PrimitiveType element_type,
+                std::vector<HloInstruction*> operands = {}) {
+    // Do not add the same op multiple times.
+    if (OpInGraph(op)) {
+      return false;
+    }
+
+    // Insert the op ahead of its first use as an operand of another op.
+    auto pos = std::find_if(
+        graph_.begin(), graph_.end(), [&op](OpDescriptor graph_op) -> bool {
+          return std::find(graph_op.operands.begin(), graph_op.operands.end(),
+                           op) != graph_op.operands.end();
+        });
+    pos = graph_.insert(pos, OpDescriptor{op, element_type, op_name, operands});
+
+    // If necessary, move the operands of the op already in the graph ahead of
+    // the op.
     for (HloInstruction* operand : operands) {
-      if (OpInGraph(operand->unique_id())) {
-        num_operands_in_graph++;
-        // Ops with more than one operand in the graph are not supported.
-        if (num_operands_in_graph > 1) {
-          return false;
-        }
-        operand_uid = operand->unique_id();
+      auto operand_pos = std::find_if(
+          pos, graph_.end(), [&operand](OpDescriptor graph_op) -> bool {
+            return operand == graph_op.instr;
+          });
+      if (operand_pos != graph_.end()) {
+        OpDescriptor operand_desc = *operand_pos;
+        graph_.erase(operand_pos);
+        pos = graph_.insert(pos, operand_desc);
       }
     }
-    graph_.emplace_back(OpDescriptor(
-        {op->unique_id(), op->shape().element_type(), op_name, operand_uid}));
     return true;
   }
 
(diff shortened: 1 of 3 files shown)


xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -488,83 +564,138 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
       continue;
     }
     // Inverse Scale
-    if (Match(user, m::Divide(&op, m::Op(&operand0),
+    if (Match(user, m::Divide(&op0, m::Op(&operand0),
                               m::Broadcast(m::Op(&operand1)))) &&
         ShapeUtil::IsScalar(operand1->shape())) {
-      if (graph_string.AppendOp("invscale", op, {operand0, operand1})) {
+      if (graph_string.AppendOp("invscale", op0, {operand0, operand1})) {
         operands.push_back(operand1);
```
reedwm, line 641: For Scale/InvScale, it's worth mentioning (and even DCHECKing (and even CHECKing)) that instr is operand0, not operand1. Since outputs of non-amax nodes in the graph are always non-scalar, and operand1 is scalar.

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -488,83 +564,138 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
+      if (graph_string.AppendOp("relu", op0, {operand0})) {
+        ++num_new_users;
         CaptureConvGraphRecursive(user, operands, aux_outputs, graph_string,
-                                  visited_instrs, final_instr);
+                                  visited_instrs, final_instr, num_endpoints);
       }
       continue;
     }
     //  Maximum of the absolute value (Amax) following ReLU (elided Abs) -- not
     //  a linear user
```
reedwm, line 659: The concept of "linear" users don't exist anymore. Instead can mention this is an aux output.

And same for the comment: "Maximum of the absolute value (Amax) -- not a linear user"

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -488,83 +564,138 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
-        num_nonlinear_users++;
+    if (Match(user, m::Reduce(&op0, m::Op(&operand0), m::Op())) &&
+        graph_string.OpInGraph(operand0, "relu") && AppliesMaxReduce(op0)) {
+      if (graph_string.AppendOp("amax", op0, {operand0})) {
+        aux_outputs.push_back(op0);
+        ++num_new_users;
+      }
+      continue;
+    }
+    // ReLU6, represented in the cuDNN graph as minimum(ReLU, 6)
```
reedwm, line 668: You should mention how its represented in the HLO graph as well. And same for ELU and Leaky Relu

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -576,33 +707,81 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
+                      .WithComparisonDirection(ComparisonDirection::kGt),
+                  m::Op().WithPredicate(unique_operand.CaptureOrVerifyFn()),
+                  m::MultiplyAnyOrder(
+                      &op1,
+                      m::Op().WithPredicate(unique_operand.CaptureOrVerifyFn()),
+                      m::Broadcast(m::ConstantEffectiveScalar(&operand1)))))) {
+        if (!graph_string.OpInGraph(op0) && !graph_string.OpInGraph(op1) &&
+            !graph_string.OpInGraph(op2)) {
+          graph_string.AppendOp("min", op0, op2->shape().element_type(),
+                                {unique_operand.Instr(), operand0});
```
reedwm, line 746: As discussed offline, this "hack" of having cudnn ops that don't correspond to HLO ops should be commented. And you need to check the number of users of certain intermediate leaky_relu ops for correctness reasons, as discussed offline.

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -576,33 +707,81 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
     operands = init_operands;
     aux_outputs = init_aux_outputs;
     final_instr = instr;
   }
+
+  // Reset the number of endpoints after rolling back the graph to a branching
+  // point that led to the addition of multiple endpoints.
+  if (num_endpoints - init_num_endpoints > 1) {
+    num_endpoints = init_num_endpoints + 1;
+  }
```
reedwm, line 784: I think these three if-statements can be reworked to make the behavior of this function more clear. I'll call three if-statements block-1, block-2, and block-3 respectively.

block-1 looks good, but I would slightly rephrase the comment to:

> 'instr' is an endpoint if it has users not in the cudnn graph, or if it has no users.

I believe the `num_new_users + num_existing_users < instr->user_count()` condition of block-2 is not necessary, and that you only need to check `if (num_endpoints > 1)`. If the first condition of block-2 is true, then block-1 will have executed, so there's already an endpoint. If there's exactly one endpoint, it means we found no other endpoints, and so resetting the graph will not make any difference. 

For block-3, I don't understand why the `num_endpoints = init_num_endpoints + 1;` isn't in block-2, and why block-3 is guarded by a separate if-statement. TBH, I'm having trouble understanding the condition `num_endpoints - init_num_endpoints > 1` and how it's relevant, but i cannot find a case where it does the wrong thing. Anyway, I think the block should be moved to block-2.

In total, I think the the three blocks should be replaced by the following:

```c++
  // 'instr' is an endpoint if it has users not in the cudnn graph, or if it has no users.
  if (num_new_users + num_existing_users < instr->user_count() ||
      instr->user_count() == 0) {
    final_instr = instr;
    ++num_endpoints;
  }

  // cuDNN only supports one endpoint, so reset the graph if we have multiple. Note this will add `instr` as an endpoint, which may still result in having multiple endpoints, in which case the recursive caller will also reset the graph.
  if (num_endpoints > 1) {
    // Reset the graph.
    graph_string = init_graph_string;
    operands = init_operands;
    aux_outputs = init_aux_outputs;
    final_instr = instr;
    // Resetting the graph removes any users of 'instr' from the graph, making instr an endpoint. So increment the number of endpoints.
    num_endpoints = init_num_endpoints + 1
  }
```

Does this new code pass the tests? I didn't actually run anything myself.

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4015,13 +4020,13 @@ class OpGraph {
...
-  absl::Status AddOp(int uid, std::optional<int> operand_uid, OpMode mode,
+  absl::Status AddOp(int uid, std::vector<int> operand_uids, OpMode mode,
                      TensorKind operand_kind, TensorKind result_kind,
                      dnn::DataType result_type) {
-    ops_.emplace_back(OpDescriptor({uid, operand_uid, mode, operand_kind,
+    ops_.emplace_back(OpDescriptor({uid, operand_uids, mode, operand_kind,
                                     result_kind, result_type, false, -1}));
-    // If it exists, the operand is virtual.
-    if (operand_uid.has_value()) {
+    // If existing, the operands are virtual.
```
reedwm, line 4028: The "existing" part of this comment is somewhat confusing. I would just state: `// The operands are virtual`

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4102,10 +4107,16 @@ GetGenericCudnnOperationGraph(
       std::string data_type_string = serialized_graph.substr(m + 1, n - m - 1);
       m = serialized_graph.find('(', pos);
       std::string op_string = serialized_graph.substr(n + 1, m - n - 1);
-      std::optional<int> operand;
-      std::string::size_type l = serialized_graph.find(')', m + 1);
-      if (l > m + 1) {
-        operand = std::stoi(serialized_graph.substr(m + 1, l - m - 1));
+      std::vector<int> operands;
```
reedwm, line 4110: Update the comment above starting with "The format of the serialized graph". You should specify there can be multiple operands.

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -576,33 +707,81 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
       std::optional<PrimitiveType> f8_type = IsSaturatingCastToF8(users_user);
```
reedwm, line 707: I think you should always check there is just one user of `user`. Otherwise you might miss an endpoint.

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -576,33 +707,81 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
       }
     }
   }
-  // Do not fuse into the cuDNN convolution Custom Call when there are more than
-  // one linear or nonlinear users, or when the number of users eligible for
-  // fusion is less than the total number of users.
-  if (num_linear_users > 1 || num_nonlinear_users > 1 ||
-      num_linear_users + num_nonlinear_users < instr->user_count()) {
+
+  // An endpoint is reached when the number of users eligibile for fusion is
```
reedwm, line 761: "eligibile" is not a word :)

xla/service/gpu/transforms/cudnn_fused_conv_rewriter_test.cc
```diff
@@ -1210,6 +1331,135 @@ TEST_F(CudnnFusedConvRewriterTest, TestConvScaledOutputMultipleUsersF8) {
...
+      R"(
+// CHECK: [[cudnn_conv_4_0:%[^ ]+]] = (f8e4m3fn[1,6,6,16]{3,2,1,0}, u8[{{.*}}]{0}) custom-call([[OPERAND0:%[^ ]+]], [[OPERAND1:%[^ ]+]], [[OPERAND2:%[^ ]+]], [[OPERAND3:%[^ ]+]], [[OPERAND4:%[^ ]+]]), window={size=3x3 pad=1_1x1_1}, dim_labels=b01f_o01i->b01f, custom_call_target="__cudnn$convForwardGraph"
+  )",
+      // serialized_graph
+      R"(
+// CHECK: "serialized_graph":"[[CONV_UID:[0-9]+]]:[f32]conv();[[SCALE0_UID:[0-9]+]]:[f32]scale([[CONV_UID]]);[[SCALE1_UID:[0-9]+]]:[f32]scale([[CONV_UID]]);[[ADD0_UID:[0-9]+]]:[f32]add([[SCALE0_UID]],[[SCALE1_UID]]);[[SCALE2_UID:[0-9]+]]:[f32]scale([[CONV_UID]]);[[ADD1_UID:[0-9]+]]:[f8e4m3fn]add([[ADD0_UID]],[[SCALE2_UID]]);"
+      )");
+}
+
+TEST_F(CudnnFusedConvRewriterTest,
```
reedwm, line 1416: You should have at least one test where the graph ends up being reset since there are multiple endpoints. There should be a test where the graph is partially reset but not fully reset (perhaps can be the same test)

reedwm (changes requested):

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -438,27 +384,34 @@ class GraphString {
...
-        OpDescriptor operand_desc = *operand_pos;
-        graph_.erase(operand_pos);
-        pos = graph_.insert(pos, operand_desc);
-      }
-    }
+    // If necessary, move the operands of the op already in the graph in front
+    // of the op. Recursively repeat for the operands' operands.
+    std::function<void(std::vector<HloInstruction*>,
+                       std::vector<OpDescriptor>::iterator)>
+        reorder_operands = [&](auto operands, auto pos) {
```
reedwm, line 399: This should be its own method, not a lambda

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -438,27 +384,34 @@ class GraphString {
...
+          for (HloInstruction* operand : operands) {
+            auto operand_pos = std::find_if(
+                pos, graph_.end(), [operand](OpDescriptor graph_op) -> bool {
```
reedwm, line 402: You look from `pos`, but the effective position of the instruction might change due to recursive calls to `reorder_operands`.

You should sort after the graph is built. You can call `MakeInstructionPostOrder` on the Hlo module, and for each, if its in the cudnn graph, put it in the list. This is short, simple, and only takes O(n) time

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -488,120 +527,217 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
+    if (Match(
+            user,
+            m::Clamp(
+                &op1,
+                m::Broadcast(&op0, m::ConstantEffectiveScalar(0)).WithOneUser(),
+                m::Op().Is(instr),
+                m::Broadcast(&operand0, m::ConstantEffectiveScalar(6))
+                    .WithOneUser()))) {
+      if (!graph_string.OpInGraph(op0) && !graph_string.OpInGraph(op1)) {
+        graph_string.AppendOp("relu", op0, {instr});
```
reedwm, line 640: You never addressed comment

> As discussed offline, this "hack" of having cudnn ops that don't correspond to HLO ops should be commented

Should have a comment stating its intentional the cudnn ops are different from the HLO ops, but that it's fine.

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -488,120 +527,217 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
+  // Since the cuDNN graph cannot have more than one endpoint, do not fuse
+  // into the cuDNN convolution Custom Call and roll back the graph when there
+  // are multiple endpoints. If the new graph still has more than one endpoint,
+  // the recursive caller will continue to roll back the graph.
+  if (num_endpoints > 1) {
     graph_string = init_graph_string;
     operands = init_operands;
     aux_outputs = init_aux_outputs;
     final_instr = instr;
+    // Reverting the graph removes any users of instr from the graph and makes
```
reedwm, line 738: Change "from the graph" to "from the cudnn graph"

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -438,27 +384,34 @@ class GraphString {
       return false;
```
reedwm, line 384: Did you address the comment:

> What if when building the cudnn graph, it has an output which is fed as input to another part of the graph

reedwm (changes requested):

reedwm (approved):

PR #23429, opened 2025-03-06
[ROCM] Fix vector bias add fusion into BLAS call

amd-jianli12:
If RHS operand of GEMM has no non-contracting dims, broadcast bias vector to guarantee its length will still match matrix D rows with HIPBLASLT_EPILOGUE_BIAS epilogue required by
(https://rocm.docs.amd.com/projects/hipBLASLt/en/latest/datatypes.html).

Consider IR sequence:
```
bf16[1024]{0} custom-call(bf16[1024,128]{1,0}, bf16[128]{0}), lhs_contracting_dims={1}, rhs_contracting_dims={0}  // gemm
bf16[1024]{0} broadcast(bf16[])  // bias vector
bf16[1024]{0} add(bf16[1024]{0}, bf16[1024]{0})  // add
```
which would be rewritten into cuBlasLt call with bias epilogue:
```
bf16[1024]{0} custom-call(bf16[1024,128]{1,0}, bf16[128]{0}, bf16[]), lhs_contracting_dims={1}, rhs_contracting_dims={0}
```
and this third operand of bf16[] led to bias vector buffer allocation with size of 1 x 2 bytes, however, cuBlasLt was expecting size of 1024 x 2 bytes that would match size of matrix D rows:
```
bf16[1024]{0} broadcast(bf16[])
bf16[1024]{0} custom-call(bf16[1024,128]{1,0}, bf16[128]{0}, bf16[1024]{0}), lhs_contracting_dims={1}, rhs_contracting_dims={0}
```

### DIFF
--- a/xla/service/gpu/transforms/cublas_gemm_rewriter_test.cc
+++ b/xla/service/gpu/transforms/cublas_gemm_rewriter_test.cc
@@ -3334,6 +3334,29 @@ ENTRY test {
 )");
 }
 
+TEST_F(CublasLtGemmRewriteTest, CublasLtFullyContractingRhsWithBias) {
+  const char* hlo_text = R"(
+HloModule test
+
+ENTRY test {
+  param_0 = bf16[10240,1024]{1,0} parameter(0)
+  param_1 = bf16[1024,1]{1,0} parameter(1)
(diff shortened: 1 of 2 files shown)


amd-jianli12:
Close this PR since alternative fix mentioned above is preferred.

PR #23459, opened 2025-03-06
[ds-fusion] Enable DS fusion by default

shraiysh:
This patch turns the default value of dynamic slice fusion to on.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -150,7 +150,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_enable_dumping(true);
 
   opts.set_xla_gpu_enable_custom_fusions(false);
-  opts.set_xla_gpu_enable_dynamic_slice_fusion(false);
+  opts.set_xla_gpu_enable_dynamic_slice_fusion(true);
   opts.set_xla_gpu_nccl_termination_timeout_seconds(-1);
   opts.set_xla_gpu_enable_shared_constants(true);
   opts.set_xla_gpu_enable_nccl_user_buffers(false);


bchetioui:
Hi @shraiysh, this is tagged as a draft. Is anything missing?

We'll have to run our benchmarks with this flag flip before we land.

shraiysh:
> Hi @shraiysh, this is tagged as a draft. Is anything missing?
> 
> We'll have to run our benchmarks with this flag flip before we land.

Some JAX tests are failing because loop induction variable isn't found. I have to work on a fix after getting the HLO.

shraiysh:
Depends on #24181

dimitar-asenov:
Hi @shraiysh, what is the state of this PR and DS in general? I see you closed this PR without merging it, is there still something that prevents us from enabling DS?

shraiysh:
Hi @dimitar-asenov, yes. TLDR is that DS fusion currently will fragment command buffer.

Long explanation is that DS Fusion requires host-compute at runtime to dispatch kernels without slicing. Consider the following example:

```
while_body {
  %before_ds_fusion_op = ...
  %ds_fusion = fusion(...), calls=%dynamic_address_computation ...
  %after_ds_fusion_op = ...
}

other_computation {
  %before_while_op = ...
  %while = while(...), body=while_body, ...
  %after_while_op = ...
}
```
[Assume that all the other operations can be pushed into command buffers.]

Without dynamic slice fusion: We could push everything into a command buffer, and all of `other_computation` could be a single command buffer.

What happens today: `%before_ds_fusion_op` and everything before that will be a part of a command buffer, `%after_ds_fusion_op` and everything after that will be a part of the command buffer. `%while_body` will have two command buffers and `%other_computation` cannot put `%while_op` inside its command buffer.

What we can do, but isn't supported today: While generating command buffer, we can emit `iter_count` iterations of `while_body` and ensure that `%before_ds_fusion_op`, `%ds_fusion` and `%after_ds_fusion_op` are inside a single command buffer. This would allow `%while_body` to be one command buffer, but `%before_while_op` and `%after_while_op` will be in a different command buffer still. Once we have this support (hopefully soon), we should discuss and decide whether it would be a good idea to enable this by default because command buffer would still be fragmented.

@shawnwang18 can you please check this explanation and correct me if I am missing something here?

dimitar-asenov:
Thank you for the detailed explanation, Shraiysh!

@ezhulenev Could you please read the comment above and weigh in? I'm not very familiar with command buffers and there's a question near the end about whether fragmenting the buffer would be OK.

shawnwang18:
The problem is how to support DUS in command buffer, @ezhulenev I had a discussion with @shraiysh , one possible solution is to unroll the WHILE loop if we want to lowering the DUS into command buffer.  And also it seems that through unrolling, we can also resolve the NCCL lowering issue (currently NCCL is not compatible with loop in cuda-graph).

PR #23494, opened 2025-03-07
[ROCm]  Use kernel cache per stream executor for redzone checker

hsharsha:
This is needed to load the kernel on each of the gpu device as rocm did not cache redzone_checker kernel per stream executor.

### DIFF
--- a/xla/stream_executor/gpu/BUILD
+++ b/xla/stream_executor/gpu/BUILD
@@ -398,6 +398,7 @@ gpu_kernel_library(
         "//xla/stream_executor:kernel",
         "//xla/stream_executor:stream_executor_h",
         "//xla/stream_executor:typed_kernel_factory",
+        "@com_google_absl//absl/container:node_hash_map",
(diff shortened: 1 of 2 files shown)


hsharsha:
@beckerhe  Can you please review this PR? It would be needed by TF with multi gpu model training

beckerhe (approved):
Sorry for the delay. LGTM

penpornk:
@hsharsha Quick question please, is this change covered by existing tests? If not, would it make sense to have a unit test for this?

hsharsha:
> @hsharsha Quick question please, is this change covered by existing tests? If not, would it make sense to have a unit test for this?

@penpornk  there are existing unit tests at https://github.com/openxla/xla/blob/main/xla/stream_executor/gpu/redzone_allocator_test.cc which covers functional part but they do not cover multi gpu scenario.

penpornk:
Thank you for the clarification!

PR #23508, opened 2025-03-07
[ROCm] Enable OCP FP8 for latest AMD archs

ScXfjiang:
We are moving from NANOO FP8 (i.e., F8E4M3FNUZ, F8E5M2FNUZ) to OCP FP8 (i.e., F8E4M3FN, F8E5M2) for new archs.

In this PR:

- We add support for OCP FP8 on ROCm. (`F8ConvertD` fusion isn't enabled now; we will submit another PR to support it soon.)
- We also find that, because a/b scales are always enabled in FP8 scenarios, it's necessary to pass these scaling factors to the autotuner (even though they are dummy pointers), ensuring it selects algorithms that handle a/b scales properly. We introduce another `is_fp8` field to `GemmBackendConfig` to support it.

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_thunk_test.cc
+++ b/xla/backends/gpu/runtime/command_buffer_thunk_test.cc
@@ -652,7 +652,7 @@ TEST(CommandBufferThunkTest, GemmCmd) {
       ShapeUtil::MakeShape(PrimitiveType::F32, {4, 3}), {}, {0},
       ShapeUtil::MakeShape(PrimitiveType::F32, {2, 3}), 1.0, 0.0, 0.0,
       PrecisionConfig::ALG_UNSET, std::nullopt,
-      se::blas::kDefaultComputePrecision, false, false,
+      se::blas::kDefaultComputePrecision, false, false, false,
       executor->GetDeviceDescription().gpu_compute_capability());
   ASSERT_TRUE(config.ok());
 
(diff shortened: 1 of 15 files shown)


ScXfjiang:
@xla-rotation Could you take a look at this PR? Many thanks in advance.

ScXfjiang:
Hi @mooskagh, could you take a look at this PR, please?

ScXfjiang:
> Hello and sorry for the delay in the review.
> 
> Could you elaborate why `is_fp8` is needed in the `GemmBackendConfig`? As far as I understand, the config for the fusion is autotuned after the fusion is built. Wouldn't it possible to check in the emitter that the Gemm is fp8?

Hi @mooskagh, thanks for the comment.

Yes, the gemm in the fusion is tuned and algorithm is fixed after the fusion is built. However, we find that, sometimes, the algorithm isn't selected correctly when we enable a/b_scale. The reason is that Autotuner (`GemmAlgorithmPicker`) doesn't handle a/b_scale, which is necessary info for `hipblasLtMatmulAlgoGetHeuristic()`. Without this info, the Autotuner may select algorithms that don't support a/b_scale. My solution is to add a `is_fp8` filed in `GemmBackendConfig` so that Autotuner will know when to select algorithms that support a/b_scale.

ScXfjiang:
Hi @mooskagh can you take a look at this? Many thanks!

reedwm:
Instead of adding an `is_fp8` attribute to GemmConfig, GemmBackendConfig, and MatmulDesc, can you directly check whether the inputs are FP8 in `BlasLt::MatmulPlan::GetAlgorithms` from the `MatrixLayout`s in `MatmulPlan`?

ScXfjiang:
Hi @reedwm @mooskagh, thanks for the feedback. I have removed `if_fp8` field from those configs. Now I'm checking the input data types directly in `GemmAlgorithmPicker`.

ScXfjiang:
Hi @reedwm @mooskagh could you take a look at the update, please? Many thanks in advance.

mooskagh (approved):

PR #23619, opened 2025-03-11
Fix the EmitSort checking after enabling NVLS and user buffer

serach24:
There is a reported bug from NVIDIA that running Midjourney model triggers XLA error after enabling NVLS and user buffer by setting NCCL_NVLS_ENABLE=1 and --xla_gpu_enable_nccl_user_buffers=true:
```
jaxlib.xla_extension.XlaRuntimeError: INTERNAL: RET_CHECK failure (external/xla/xla/service/gpu/ir_emitter_unnested.cc:1676) LayoutUtil::LayoutsInShapesEqual(keys_shape, sort->operand(i)->shape()).
```
This is because after enabling NVLS and user buffer, one of the operand of `sort` operation is from a different memory space (user buffer), and the previous `LayoutsInShapesEqual` check is too strong to pass as it also checks if operands are from the same memory space.

This MR makes the sort layout check weaker as operands do not have to be in the same memory space as long as they all on the device.

### DIFF
--- a/xla/service/gpu/ir_emitter_unnested.cc
+++ b/xla/service/gpu/ir_emitter_unnested.cc
@@ -1672,10 +1672,12 @@ absl::Status IrEmitterUnnested::EmitSort(const HloSortInstruction* sort) {
         sort->operand_count() > 1 ? ShapeIndex({i}) : ShapeIndex({});
     // We assume that the layout of all involved operands and outputs is the
     // same.
-    TF_RET_CHECK(LayoutUtil::LayoutsInShapesEqual(keys_shape,
-                                                  sort->operand(i)->shape()));
+    TF_RET_CHECK(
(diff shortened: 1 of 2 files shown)


dimitar-asenov (changes requested):
Thank you for this fix. Please add a test that fails without this change.

akuegel (commented):
Thank you. Can you please write a test for this as well?

akuegel (approved):

dimitar-asenov (approved):

dimitar-asenov:
@serach24 It seems that you have not yet agreed to the CLA. Could you please do that so we can merge this change? Thank you.

serach24:
> @serach24 It seems that you have not yet agreed to the CLA. Could you please do that so we can merge this change? Thank you.

Hi, I found that I need to manually trigger the CLA rescan for it to work, it should work now.

PR #23627, opened 2025-03-11
enable hipblaslt by default on ROCm

Ruturaj4:


beckerhe:
Hey, can you comment on why you're making this change?

@i-chaochen your thoughts?

beckerhe:
> > Hey, can you comment on why you're making this change?
> > @i-chaochen your thoughts?
> 
> Hi @beckerhe thanks for the asking. Yes, hipblaslt is getting more and more important on us and we would like to put hipblaslt as the default choice for ml/dl workloads.

Okay, fine with me.

@Ruturaj4 Have you tested the change? I only had a brief look and I don't think `debug_options_flag.cc` is compiled with the `TENSORFLOW_USE_ROCM` preprocessor definition set. That's why I'm curious whether this works.

dimitar-asenov:
@i-chaochen @Ruturaj4 I'm closing this due a long inactivity. If you still want to merge this PR, please reply to the last comment and reopen it.

PR #23650, opened 2025-03-12
Fix EvaluateWithSubstitutions when the evaluated thing is substituted.

jreiffers:
Currently, this results in an error, which is surprising. There seems no benefit to the current behavior, so let's just return the substitued value for this case.


jreiffers:
@akuegel this fixes (some of) the failures on https://github.com/openxla/xla/pull/23610.

akuegel (approved):

jreiffers:
The ARM failure looks like a flake to me. It doesn't look like the test exercises this functionality, and it passed on x86.

PR #23654, opened 2025-03-12
[ROCm] Enable mfma instructions by passing the correct arch name

### LINKED ISSUE #23574: [ROCm] Very slow Triton performance vs PyTorch for the same kernel
Hello folks,

When running a custom paged attention kernel, we get about ~10x performance hit on a Triton kernel when running on Triton/XLA than on Triton/PyTorch. We are using the kernel at https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/pa_decode.py#L309

We're not sure what can cause this, the grids are the same. We saw that ThreadPerWarp is hardcoded to 32 in XLA (should be 64), but I'm not sure that might explain it.

When trying this Pallas implementation at https://github.com/jax-ml/jax/blob/main/jax/experimental/pallas/ops/gpu/paged_attention.py in JAX, we get ~750us on H200 but 20ms (!) on MI300x.

We kind of exhausted the obvious stuff, and are starting to wonder this might be due to a problem of ROCm in XLA.

Thank you!

steeve:
This PR enables the Triton pipeline to emit `#triton_gpu.amd_mfma` annotations during the Triton to TritonGPU lowering.

This is done by the `TritonAMDGPUAccelerateMatmulPass`, which checks the GFX version to do that.

Correctly passing the `gfx_version` reduces our kernel runtime from **~5ms** to **~620us** on MI300X, matching the performance of the Python Triton Compiler used in Torch.

We expect this change to radiate quite a bit given that this pipeline is shared by the IR Fusion Emitter used widely across XLA if `tt.dot` ops are emitted.

Closes #23574

### DIFF
--- a/xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
+++ b/xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
@@ -90,7 +90,7 @@ absl::Status CreateTritonPipeline(mlir::OpPassManager* pm,
   pm->addPass(mt::gpu::createTritonGPUCoalesce());
   pm->addPass(mt::gpu::createTritonGPURemoveLayoutConversions());
   pm->addPass(mt::gpu::createTritonGPUOptimizeThreadLocality());
-  pm->addPass(mlir::createTritonAMDGPUAccelerateMatmulPass());
+  pm->addPass(mlir::createTritonAMDGPUAccelerateMatmulPass(cc.gfx_version()));
   pm->addPass(mt::gpu::createTritonGPURemoveLayoutConversions());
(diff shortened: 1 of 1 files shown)


i-chaochen (approved):
Thanks!

i-chaochen:
Hi @xla-rotation check this PR, please.

gflegar (approved):

i-chaochen:
> I assume this is what upstream Triton's pipeline already does, and we just didn't port it correctly to XLA?

Yes. But we're also thinking of preventing this hard coding things every time at upstream triton side as well. Otherwise, it's hard to follow all changes.

PR #23681, opened 2025-03-13
Add an API to get a computation's caller(s).

jreiffers:
This should replace CallGraph in most cases, and adds an alternative to the deprecated .*CallInstruction functions.


akuegel (approved):
Thank you!

PR #23764, opened 2025-03-14
Refactor hlo_text to hlo_file for Consistent Naming in Multihost HLO Runner

apivovarov:
FunctionalHloRunner::LoadHloModuleAndArguments takes an string_view parameter named `hlo_file`, representing the file path. The file can be in various formats, such as text, proto, or snapshot.

However, the higher-level API LoadAndRun previously used the variable name `hlo_text` for the same purpose, which could be misleading since it implies content rather than a file path.

This PR updates the variable name from `hlo_text` to `hlo_file` across the relevant functions for better clarity and consistency.

Additionally, hlo_runner_main.cc used the variable name `filename` - this has also been renamed to `hlo_file` to ensure consistent naming across all API levels


sergachev (approved):
LGTM but a Googler will need to approve and merge it.

mooskagh (approved):

PR #23790, opened 2025-03-17
Add explicit collectives grouping pass.

chaserileyroberts:
This PR adds an explicit frontend attribute to jitted JAX methods that will force the computation to be run in a single NCCL Group. 

The intention of this change is to enable "multi-directional communications" that better saturate NVLink systems. We reuse the existing NCCLGroupThunk logic during the IR emitter stage, so we only need to introduce the async wrapper and call inliner.

Here is an example of using this feature from JAX.

```python
import jax
from jax._src.xla_metadata import set_xla_metadata
from jax.experimental.shard_map import shard_map
from jax.sharding import Mesh, PartitionSpec as P
from functools import partial

num_devices = 4
mesh = Mesh(np.array(jax.devices()), ('i',))

# Unique source target pairs for our two ppermutes later.
perm_up = [(i, (i+1) % num_devices) for i in range(num_devices)]
perm_down = [(i, (i-1) % num_devices) for i in range(num_devices)]

# NCCL Group computations _must_ be jitted.
@jax.jit
def bidir_comms(a):
    b = jax.lax.ppermute(a, "i", perm_up)
    c = jax.lax.ppermute(a, "i", perm_down)
    return b, c

@jax.jit
@partial(shard_map, mesh=mesh, in_specs=P(None, 'i'), out_specs=P(None, 'i'))
def groups(a):
   # Running our jitted function in this context will force the use of a NCCL Group.
    with set_xla_metadata(_collectives_group="", inlineable="false"):
        b, c = bidir_comms(a)
    return b + c
```

This is nsys trace with the annotation.
Screenshot 2025-03-16 at 8 32 44 PM

Vs without the additional annotation
Screenshot 2025-03-16 at 8 31 56 PM

As you can see, there is only a single NCCL kernel in the annotated example.

### DIFF
--- a/xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.cc
+++ b/xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.cc
@@ -1,0 +1,81 @@
+/* Copyright 2024 The OpenXLA Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h"
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/log/log.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_clone_context.h"
+#include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_opcode.h"
+#include "xla/service/gpu/backend_configs.pb.h"
+#include "xla/side_effect_util.h"
+#include "xla/tsl/platform/errors.h"
+#include "xla/tsl/platform/statusor.h"
+#include "xla/util.h"
+#include "xla/xla_data.pb.h"
+
+namespace xla {
+namespace gpu {
+
+namespace {
+
+absl::StatusOr<bool> CreateNcclGroupAsyncPair(HloInstruction* instr) {
+  if (instr->opcode() != HloOpcode::kCall ||
+      !instr->frontend_attributes().map().contains(kNcclGroupAttr)) {
+    return false;
+  }
+  HloComputation* computation = instr->parent();
+  auto new_computation = instr->GetModule()->AddEmbeddedComputation(
+      instr->to_apply()->Clone("nccl_group"));
+  // Get the shapes for the original instruction.
+  std::vector<const Shape*> parameter_shapes(instr->operand_count());
+  for (int i = 0; i < instr->operand_count(); ++i) {
+    parameter_shapes[i] = &instr->operand(i)->shape();
+  }
+  std::vector<Shape> start_shapes = {
+      ShapeUtil::MakeTupleShapeWithPtrs(parameter_shapes), instr->shape()};
+  HloInstruction* async_start =
+      computation->AddInstruction(HloInstruction::CreateAsyncStart(
+          ShapeUtil::MakeTupleShape(start_shapes), instr->operands(),
+          new_computation, "explicit"));
+  HloInstruction* async_done = computation->AddInstruction(
+      HloInstruction::CreateAsyncDone(instr->shape(), async_start));
+  // Forward frontend attributes to both async instructions.
+  async_start->set_frontend_attributes(instr->frontend_attributes());
+  async_done->set_frontend_attributes(instr->frontend_attributes());
+  TF_RETURN_IF_ERROR(computation->ReplaceInstruction(instr, async_done));
+  return true;
+}
+}  // namespace
+
+absl::StatusOr<bool> ExplicitNcclGroupAsyncWrapper::Run(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  bool changed = false;
+  for (const HloComputation* comp : module->computations()) {
+    for (HloInstruction* instr : comp->instructions()) {
+      TF_ASSIGN_OR_RETURN(bool result, CreateNcclGroupAsyncPair(instr));
+      changed |= result;
+    }
+  }
+  return changed;
+}
+
+}  // namespace gpu
+}  // namespace xla
--- a/xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
+++ b/xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
@@ -1,0 +1,50 @@
+/* Copyright 2025 The OpenXLA Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef TENSORFLOW_COMPILER_XLA_SERVICE_GPU_TRANSFORMS_EXPLICIT_NCCL_GROUP_ASYNC_WRAPPER_H_
+#define TENSORFLOW_COMPILER_XLA_SERVICE_GPU_TRANSFORMS_EXPLICIT_NCCL_GROUP_ASYNC_WRAPPER_H_
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/hlo/pass/hlo_pass_interface.h"
+
+namespace xla {
+namespace gpu {
+
+// This pass will find the kCall instructions that
+// are annotated with explicit nccl groups in their frontend
+// attributes. It then will convert the kCall into an async
+// start-done pair with the same computation. This is then
+// picked up by the IR emitter stage, and the entire computation
+// will be launched in a single NCCL Group.
+class ExplicitNcclGroupAsyncWrapper : public HloModulePass {
+ public:
+  absl::string_view name() const override {
+    return "explicit-nccl-group-async-wrapper";
+  }
+
+  using HloPassInterface::Run;
+  absl::StatusOr<bool> Run(
+      HloModule* module,
+      const absl::flat_hash_set<absl::string_view>& execution_threads) override;
+};
+
+}  // namespace gpu
+}  // namespace xla
+
+#endif  // TENSORFLOW_COMPILER_XLA_SERVICE_GPU_TRANSFORMS_EXPLICIT_NCCL_GROUP_ASYNC_WRAPPER_H_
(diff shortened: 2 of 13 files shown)


xla/service/hlo_verifier.cc
```diff
@@ -3104,13 +3092,6 @@ absl::StatusOr<bool> HloVerifier::Run(
     for (auto* computation : module->computations(execution_threads)) {
       TF_RETURN_IF_ERROR(computation->Accept(shape_verifier.get()));
       TF_RETURN_IF_ERROR(computation->Accept(&instruction_verifier));
-      // Verify that async computations contain a single instruction or a
-      // collection of send/recv instructions. This is needed to represent NCCL
-      // groups on GPU.
-      if (computation->IsAsyncComputation() &&
-          !computation->OnlyContainsSendRecv()) {
```
rosiezou, line 3111: This part of HLO verifier is not GPU-specific. Instead of deleting the check, I suggest the following:
1. Keep `VerifyAsyncComputation`
2. Change the condition from `computation->IsAsyncComputation() && !computation->OnlyContainsSendRecv()` to the following
```
// in hlo_computation.h and hlo_computation.cc, add a check for caller type, something that looks like this
bool IncludedInAsyncNcclGroup(){
    std::unique_ptr<CallGraph> call_graph = CallGraph::Build(this->parent_);
    std::vector<HloInstruction*> callers = call_graph->GetComputationCallers(computation);
    for (HloInstruction* caller : callers) {
        if (caller->opcode() == HloOpcode::kCall && caller->get_frontend_attribute(kNcclGroupAttr) != nullptr){
            return true;
        }
    return false;
}

// In hlo_verifier.cc, line 3110
if (computation->IsAsyncComputation &&
    !computation->OnlyContainsSendRecv() &&
    !IncludedInAsyncNcclGroup()) {
    TF_RETURN_IF_ERROR(VerifyAsyncComputation(computation));
}
```
This way we don't break the parts of XLA that requires async-wrapped computations to only contain a single operation.

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.cc
```diff
@@ -0,0 +1,77 @@
...
+}
+}  // namespace
+
+absl::StatusOr<bool> ExplicitNcclGroupAsyncWrapper::Run(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  bool changed = false;
+  for (const HloComputation* comp : module->computations()) {
+    for (HloInstruction* instr : comp->instructions()) {
+      TF_ASSIGN_OR_RETURN(bool result, CreateNcclGroupAsyncPair(instr));
```
rosiezou, line 70: Looks like your feature supports creating more than one nccl group per program. If that is the case, please add corresponding test cases that contain multiple nccl groups.

xla/service/call_inliner.cc
```diff
@@ -176,6 +176,10 @@ bool InlineComposites(
              instruction->frontend_attributes().map().at("composite.name"));
 }
 
+bool InlineNcclGroups(HloInstruction* instruction) {
+  return !instruction->frontend_attributes().map().contains(kNcclGroupAttr);
```
rosiezou, line 180: I believe this function & the corresponding test case in `call_inliner_test.cc` are unnecessary since the runtime behavior in XLA:GPU is the same for inlined vs. un-inlined functions. Both are essentially inlined for buffer assignment purpose and resulting `ThunkSequence` is the same.

rosiezou (commented):
Left some comments. In summary:
1. Please clarify the behavior of the HLO pass and the expected dev ex for JAX users. In particular, please add examples for multiple NCCL groups if it is allowed, or clarify via inline comments and doc strings if you're limiting kCall nccl groups to a single group per program.
2. The way we test `ir_emitter_unnested.cc` is via execution tests such as `collective_ops_test.cc`. Please add similar execution tests for this feature. I have also left additional comments in the HLO pass regarding test coverage.

xla/service/hlo_verifier.cc
jreiffers: Creating a call graph should no longer be necessary here, computations now know their callers: https://github.com/openxla/xla/blob/7639178ebd11a1248a2fb3ac9138c1a14f0353aa/xla/hlo/ir/hlo_computation.h#L960

xla/service/call_inliner.cc
chaserileyroberts: We do end up needing this, but for a different reason than you state. 

The call operation is how we define the nccl grouping boundaries. If we don't include it in the inline exceptions, then the call will disappear very early in the compiler stack and we won't have a clear way to define our NCCL group during the async wrapper pass. Ultimately, the call operation does get deleted during the async wrapper pass, so it won't survive all the way to the thunk executions.

rosiezou: Got it. Thanks for the clarification!

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.cc
chaserileyroberts: This was a good catch. Previously, if you used the same computation twice for two different async nccl groups, then this [check](https://github.com/openxla/xla/blob/a31cce0bcfa13aedf6ca2ba398eea8781c014e1d/xla/hlo/ir/hlo_instructions.cc#L405) fails during the construction of the second async start instruction.

I modified the code to instead create a clone of the computation and added tests for that situation.

xla/service/hlo_verifier.cc
chaserileyroberts: I think just adding a check if the async op has the `_nccl_group` frontend attribute would be sufficient.

chaserileyroberts:
> Please clarify the behavior of the HLO pass and the expected dev ex for JAX users. In particular, please add examples for multiple NCCL groups if it is allowed, or clarify via inline comments and doc strings if you're limiting kCall nccl groups to a single group per program.

I added an example that uses multiple nccl groups from the same original computation and modified the program to make it work.

> The way we test ir_emitter_unnested.cc is via execution tests such as collective_ops_test.cc. Please add similar execution tests for this feature. I have also left additional comments in the HLO pass regarding test coverage.

There conveniently already exists `nccl_group_execution_test.cc` so I added a test there targeting the specific bidirectional communication pattern we want to eventually use.

xla/service/hlo_verifier.cc
rosiezou: Yea that works too. I'm cool with whatever implementation as long as you're not skipping the check completely.

chaserileyroberts: I modified the check and put back all of the tests I removed originally.

chaserileyroberts:
@rosiezou This is ready for another pass whenever you have time.

rosiezou:
Please fix the CI failures caused by `call_inliner_test`. Other than that, LGTM.

rosiezou (approved):

chaserileyroberts:
> Please fix the CI failures caused by call_inliner_test. Other than that, LGTM.

I actually ended up deleting everything related to the `call_inliner`. Instead, the frontend attribute `inlineable` does what we need. I added that to the example code.

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
```diff
@@ -0,0 +1,48 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef XLA_SERVICE_GPU_TRANSFORMS_NCCL_GROUP_ASYNC_WRAPPER_H_
+#define XLA_SERVICE_GPU_TRANSFORMS_NCCL_GROUP_ASYNC_WRAPPER_H_
```
rosiezou, line 17: @chaserileyroberts please use `TENSORFLOW_COMPILER_XLA_SERVICE_GPU_TRANSFORMS_EXPLICIT_NCCL_GROUP_ASYNC_WRAPPER_H_` in the header guard

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
```diff
@@ -0,0 +1,48 @@
...
+
+  using HloPassInterface::Run;
+  absl::StatusOr<bool> Run(
+      HloModule* module,
+      const absl::flat_hash_set<absl::string_view>& execution_threads) override;
+};
+
+}  // namespace xla::gpu
+
+#endif  // XLA_SERVICE_GPU_TRANSFORMS_NCCL_GROUP_ASYNC_WRAPPER_H_
```
rosiezou, line 48: same here please use `TENSORFLOW_COMPILER_XLA_SERVICE_GPU_TRANSFORMS_EXPLICIT_NCCL_GROUP_ASYNC_WRAPPER_H_` in the inline comment

rosiezou (changes requested):

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
```diff
@@ -0,0 +1,48 @@
...
+    return "explicit-nccl-group-async-wrapper";
+  }
+
+  using HloPassInterface::Run;
+  absl::StatusOr<bool> Run(
+      HloModule* module,
+      const absl::flat_hash_set<absl::string_view>& execution_threads) override;
+};
+
+}  // namespace xla::gpu
```
rosiezou, line 46: nit
```
}  // end namespace gpu
}  // end namespace xla
```

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
```diff
@@ -0,0 +1,48 @@
...
+#define XLA_SERVICE_GPU_TRANSFORMS_NCCL_GROUP_ASYNC_WRAPPER_H_
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/hlo/pass/hlo_pass_interface.h"
+
+namespace xla::gpu {
```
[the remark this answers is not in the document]
chaserileyroberts: Done, here and in the .cc file.

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
chaserileyroberts: done

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
chaserileyroberts: Done

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper.h
chaserileyroberts: Done

rosiezou (approved):

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper_test.cc
```diff
@@ -0,0 +1,106 @@
...
+#include "xla/service/gpu/backend_configs.pb.h"
+#include "xla/side_effect_util.h"
+#include "xla/tests/hlo_test_base.h"
+#include "xla/tsl/lib/core/status_test_util.h"
+#include "xla/tsl/platform/statusor.h"
+
+namespace xla::gpu {
+namespace {
+
+using ExplicitNcclGroupAsyncWrapperTest = HloTestBase;
```
ezhulenev, line 34: We removed Nccl from all type names in XLA:GPU and we use generic "collectives" + generic collectives APIs everywhere. Can you rename it to `ExplicitCollectiveGroupAsyncWrapper`.

chaserileyroberts: Sure, but this pass might be meaningless on non-Nvidia hardware. We also are still using `_nccl_group` as the frontend attribute, do you think that needs to be changed too?

frgossen (approved):

xla/service/gpu/transforms/explicit_nccl_group_async_wrapper_test.cc
ezhulenev: Yes, eventually most mentions of nccl should go away (when they have a generic replacement).

chaserileyroberts: I'll change it to `_collectives_group`

akuegel (approved):

PR #23793, opened 2025-03-17
Add stream dependency after calling memcpy p2p

Tixxx:
Currently we have stream dependency to make sure buffers are ready before calling memcpy p2p in collective permute thunks. Post-memcpy sync using stream dep is also needed to make sure all data has been fully written before receiving rank consumes it.


reedwm (changes requested):
Is it possible to write a test? E.g., a collective-permute followed by some compute that uses the output of collective-permute, that would fail if no synchronization was done after the collective-permute?

Tixxx:
> Is it possible to write a test? E.g., a collective-permute followed by some compute that uses the output of collective-permute, that would fail if no synchronization was done after the collective-permute?

I managed to reproduce a failure with a test on my machine. Hopefully it deterministically fails for other machines.

reedwm (approved):

PR #23799, opened 2025-03-17
Add support for emitting DUS as memcpy.

jreiffers:
This extends the dynamic memcpy support for dynamic-slice to dynamic-update-slice. The logic is largely the same (except this time, the destination offset is dynamic, not the source offset).

For now, only in-place DUSes are handled. We could handle out-of-place DUSes as well, but I don't think it's needed.

### DIFF
--- a/xla/backends/gpu/codegen/copy.cc
+++ b/xla/backends/gpu/codegen/copy.cc
@@ -106,6 +106,7 @@ absl::StatusOr<FusionEmissionResult> DynamicMemcpyFusion::Emit(
     source_operand_index = 1;
     copy_shape = &root.GetOperand(source_operand_index).shape();
   } else {
+    CHECK_EQ(root.opcode(), HloOpcode::kDynamicSlice);
     source_operand_index = 0;
     copy_shape = &root.shape();
   }
@@ -128,11 +129,15 @@ absl::StatusOr<FusionEmissionResult> DynamicMemcpyFusion::Emit(
 
 namespace {
 
+// Whether the offset in the given dimension of the slice operation is
+// guaranteed to be clamped to 0. This is the case if the slice size is the
+// same as the size of the dimension in the unsliced shape.
 bool IsZeroOffset(const HloInstruction* slice, int dim) {
   if (slice->opcode() == HloOpcode::kDynamicSlice) {
     return slice->dynamic_slice_sizes()[dim] ==
            slice->operand(0)->shape().dimensions(dim);
   }
+  CHECK_EQ(slice->opcode(), HloOpcode::kDynamicUpdateSlice);
   return slice->operand(1)->shape().dimensions(dim) ==
          slice->operand(0)->shape().dimensions(dim);
 }
@@ -158,7 +163,7 @@ std::vector<const HloInstruction*> GetCallStack(
   return stack;
 }
 
-int GetFirstOffset(const HloInstruction* slice) {
+int GetFirstOffsetOperandIndex(const HloInstruction* slice) {
   // dynamic-slice takes the full array, then the offsets.
   // dynamic-update-slice takes the full array, then the update slice, then the
   // offsets.
@@ -190,8 +195,8 @@ bool DynamicMemcpyFusion::IsCandidateFusion(
     return false;
   }
 
-  int first_offset = GetFirstOffset(root);
-  for (int i = 0; i < first_offset; ++i) {
+  int first_offset_index = GetFirstOffsetOperandIndex(root);
+  for (int i = 0; i < first_offset_index; ++i) {
     if (root->operand(i)->opcode() != HloOpcode::kParameter) {
       VLOG(5) << "Not a slice of a parameter.";
       return false;
@@ -200,7 +205,7 @@ bool DynamicMemcpyFusion::IsCandidateFusion(
 
   int rank = root->operand(0)->shape().rank();
   for (int i = 0; i < rank; ++i) {
-    auto* operand = root->operand(i + first_offset);
+    auto* operand = root->operand(i + first_offset_index);
     if (!IsZeroOffset(root, i) && operand->opcode() != HloOpcode::kConstant &&
         operand->opcode() != HloOpcode::kParameter) {
       VLOG(5) << "Dimension " << i << " is not a constant or a parameter.";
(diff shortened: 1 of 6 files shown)


xla/backends/gpu/codegen/copy.cc
```diff
@@ -83,28 +83,57 @@ absl::StatusOr<FusionEmissionResult> DynamicMemcpyFusion::Emit(
...
+                        buffer_assignment_->GetUniqueSlice(
+                            &root.GetOperand(0).instruction(), {}));
+    TF_ASSIGN_OR_RETURN(BufferAllocation::Slice dst,
+                        buffer_assignment_->GetUniqueSlice(&fusion, {}));
+    CHECK_EQ(input, dst);
+
+    source_operand_index = 1;
+    copy_shape = &root.GetOperand(source_operand_index).shape();
+  } else {
+    source_operand_index = 0;
```
bchetioui, line 109: nit: could we assert on what the opcode is expected to be here? It's not 100% clear locally, though I expect it's a DS? :)

xla/backends/gpu/codegen/copy.cc
```diff
@@ -83,28 +83,57 @@ absl::StatusOr<FusionEmissionResult> DynamicMemcpyFusion::Emit(
...
+      /*source_buffer=*/src_buffer,
+      /*destination_buffer=*/dst_buffer,
+      /*mem_size=*/ShapeUtil::ByteSizeOfElements(*copy_shape),
+      /*descriptor=*/descriptor_));
   return result;
 }
 
 namespace {
 
 bool IsZeroOffset(const HloInstruction* slice, int dim) {
```
bchetioui, line 131: Could we give a brief docstring for what we expect this to return for DS/DUS? (The code makes sense, I'm just not sure it has the behaviour I'd expect if I saw a call to `IsZeroOffset` in a near-vacuum.)

xla/backends/gpu/codegen/copy.cc
```diff
@@ -129,12 +158,22 @@ std::vector<const HloInstruction*> GetCallStack(
   return stack;
 }
 
+int GetFirstOffset(const HloInstruction* slice) {
```
bchetioui, line 161: nit: `GetFirstOffsetOperandIndex`? This is not returning the offset itself, right?

xla/service/gpu/tests/gpu_copy_test.cc
```diff
@@ -214,5 +214,99 @@ TEST_F(GpuCopyTest, DoNotUseMemcpyWithLayoutChange) {
...
+                          "known_init_step":{"init":"0","step":"1"},
+                          "known_induction_variable":{"tuple_index":"0"}}
+    })";
+
+TEST_F(GpuCopyTest, UseMemcpyForDynamicUpdateSlice) {
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<VerifiedHloModule> hlo_module,
+      ParseAndReturnVerifiedModule(kDynamicUpdateSliceModule));
+
+  CompileAndVerifyIr(std::move(hlo_module), "; CHECK-NOT: void @updated",
```
bchetioui, line 292: Wonder if there's anything better we can do than matching the name?

These tests could easily be passed by just generating different function names, which makes me think that they may not be super good at verifying specific invariants. Would it make sense to check emitted thunks instead?

xla/service/gpu/tests/gpu_copy_test.cc
```diff
@@ -214,5 +214,99 @@ TEST_F(GpuCopyTest, DoNotUseMemcpyWithLayoutChange) {
...
+                     /*match_optimized_ir=*/false,
+                     /*run_optimization_passes=*/false);
+  EXPECT_TRUE(RunAndCompareNoHloPasses(kDynamicUpdateSliceModule,
+                                       ErrorSpec{1e-5, 1e-5}));
```
bchetioui, line 296: nit: is there ever a case where `ErrorSpec{0, 0}` doesn't work here? It doesn't seem like there should be any associativity issue (?)

bchetioui (changes requested):
Thanks, looks mostly good! Just a few comments.

xla/service/gpu/tests/gpu_copy_test.cc
jreiffers: That's why we have a pair of tests. Maybe I don't understand your concern :-)

xla/service/gpu/tests/gpu_copy_test.cc
jreiffers: No, this was just copy&paste. We don't even have any floats here, so this is exactly the same.

xla/backends/gpu/codegen/copy.cc
jreiffers: Done.

xla/backends/gpu/codegen/copy.cc
jreiffers: Done.

xla/backends/gpu/codegen/copy.cc
jreiffers: Done.

jreiffers:
Thanks for the review! PTAL.

xla/service/gpu/tests/gpu_copy_test.cc
bchetioui: Should we just make it an exact match then?

xla/service/gpu/tests/gpu_copy_test.cc
bchetioui: My concern is that these tests only enforce that we pick two different lowering paths, but do not actually check if those lowering paths behave as we expect :) (which may be obvious now, but could very well change under our feet later).

Anyway, this is not a huge concern. Was just wondering if there was a more explicit test of "we are doing a `memcpy` here" that we could write easily---this is fine if not :)

bchetioui (changes requested):
Can we make the match exact to make it clear no numerical difference is allowed? Otherwise LGTM :)

EDIT: please ignore :)

bchetioui (approved):
Nevermind, sorry. The GitHub diff UI tripped me up, I see you made the change already. Thank you!

PR #21973, opened 2025-01-28
Add allreduce runtime support for nvshmem reduction on-stream api

Tixxx:
This pr adds runtime support for nvshmem reduction on-stream api.
It adds a new backend config to instruction the emitter to lower to nvshmem with corresponding collective thunks that call nvshmem reduction api.

PR #23829, opened 2025-03-17
Add CollectiveBackendAssigner pass for NVSHMEM vs. NCCL

sfvaroglu:
This PR adds `CollectiveBackendAssigner` pass, which assigns either NVSHMEM or NCCL as the backend for collective operations based on:
1. Communication pattern (intranode vs internode)
2. Message size (compared against threshold_in_bytes)

Also, there is some utility code taken from https://github.com/openxla/xla/pull/21973.


xla/service/gpu/gpu_compiler.cc
```diff
@@ -862,6 +865,13 @@ absl::Status RunCollectiveOptimizationPasses(
 
   HloPassPipeline collectives_pipeline("collective-optimizations");
   collectives_pipeline.AddPass<RaggedAllToAllCanonicalizer>();
+
+  if (debug_options.xla_gpu_experimental_enable_nvshmem() &&
+      NvshmemCollectives::Default()->device_count_per_process() == 1) {
+    collectives_pipeline.AddPass<CollectiveBackendAssigner>(
+        /*threshold_in_bytes*/ 16 * 1024 * 1024);
```
frgossen, line 872: Can you factor this out as a constant or make it a flag?

xla/service/gpu/transforms/collective_backend_assigner.h
```diff
@@ -0,0 +1,117 @@
...
+        instr.opcode() == HloOpcode::kAllReduceDone) {
+      return HasInternodeCommunication(instr.replica_groups(), num_processes);
+    }
+    if (instr.opcode() == HloOpcode::kCollectivePermute ||
+        instr.opcode() == HloOpcode::kCollectivePermuteStart ||
+        instr.opcode() == HloOpcode::kCollectivePermuteDone) {
+      return HasInternodeCommunication(instr.source_target_pairs(),
+                                       num_processes);
+    }
+    return false;
```
frgossen, line 87: Can we throw some error for the uncovered cases? For example, for a local reduce-scatter, this would be an unexpected return value

frgossen (changes requested):
Thanks for adding the pass! Adding Levon for awareness

xla/service/gpu/gpu_compiler.cc
jprabhas: Did we converge on 16MB being the right threshold for all collectives? Should this threshold be tunable by the user?

xla/service/gpu/transforms/collective_backend_assigner.h
```diff
@@ -0,0 +1,117 @@
...
+
+  absl::string_view name() const override {
+    return "collective-backend-assigner";
+  }
+
+  absl::StatusOr<bool> Run(
+      HloModule* module,
+      const absl::flat_hash_set<absl::string_view>& execution_threads) override;
+
+  static bool HasInternodeCommunication(
```
dimvar, line 41: Nit: for clarity it would be better to have only function headers in this file, and the implementation in the cc file.

xla/service/gpu/transforms/collective_backend_assigner.h
```diff
@@ -0,0 +1,117 @@
...
+#include "absl/container/flat_hash_set.h"
+#include "absl/status/statusor.h"
+#include "absl/log/log.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/hlo/pass/hlo_pass_interface.h"
+
+namespace xla {
+namespace gpu {
+
+class CollectiveBackendAssigner : public HloModulePass {
```
dimvar, line 28: Maybe add a comment here to briefly describe the purpose of the class.

sfvaroglu: done

sfvaroglu:
@frgossen @PatriosTheGreat PTAL

frgossen (approved):

xla/service/gpu/gpu_compiler.cc
```diff
@@ -918,6 +921,12 @@ absl::Status RunCollectiveOptimizationPasses(
 
   HloPassPipeline collectives_pipeline("collective-optimizations");
   collectives_pipeline.AddPass<RaggedAllToAllCanonicalizer>();
+
+  if (debug_options.xla_gpu_experimental_enable_nvshmem() &&
+      NvshmemCollectives::Default()->device_count_per_process() == 1) {
```
PatriosTheGreat, line 926: NVSHMEM collective might not be registered at this point. Compilation might be done by separate tool which doesn't have a dependency to any type of collectives (nccl or nvshmem). Here we either need to trust in sense that if xla_gpu_experimental_enable_nvshmem is enabled then single process per device constrains is meet, or

xla/service/gpu/transforms/collective_backend_assigner.cc
```diff
@@ -0,0 +1,132 @@
...
+    for (HloInstruction* instr : comp->instructions()) {
+      if (IsCollectiveOp(instr)) {
+        TF_ASSIGN_OR_RETURN(GpuBackendConfig gpu_config,
+                            instr->backend_config<GpuBackendConfig>());
+        CollectiveBackendConfig& backend_config =
+            *gpu_config.mutable_collective_backend_config();
+
+        auto backend =
+            !HasInternodeCommunication(
+                *instr, NvshmemCollectives::Default()->num_processes()) &&
```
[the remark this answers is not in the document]
sfvaroglu: Which field in module/target config do you recommend checking or do we need to add a new field?

Tixxx:
> How is this supposed to work for e.g. GB200? My understanding is that we can issue NVSHMEM collectives for any communication within NVLINK domain.

The current version of nvshmem in xla has some issues when running on GB200. Those are addressed in nvshmem 3.2. After we upgrade to 3.2 in xla, we can switch the default team from NODE to SHARED which will allow us to lower to nvshmem for any nvlink domain including B200 and GB200 nodes. For now we will only lower it to nvshmem for intra-node allreduces.

dimvar (commented):
LGTM from me. Some minor comments inline. I'll let the other reviewers provide the final approval.

xla/service/gpu/transforms/collective_backend_assigner.cc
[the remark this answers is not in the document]
PatriosTheGreat: You can use here something like xla/service/gpu/transforms/collectives/collective_ops_utils:CommunicationType == SingleHost.

xla/service/gpu/gpu_compiler.cc
sfvaroglu: removed

xla/service/gpu/transforms/collective_backend_assigner.cc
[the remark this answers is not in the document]
sfvaroglu: done

xla/backends/gpu/collectives/nvshmem_collectives.h
```diff
@@ -47,6 +47,9 @@ class NvshmemCollectives : public GpuCollectives {
                   size_t device_count_per_process,
                   std::weak_ptr<KeyValueStoreInterface> kv_store);
 
+  size_t device_count_per_process() const { return device_count_per_process_; }
+  size_t num_processes() const { return num_processes_; }
```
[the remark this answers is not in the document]
sfvaroglu: removed, no longer needed

xla/service/gpu/transforms/collective_backend_assigner.h
allanrenucci, line 1: Can we move these files into the gpu/transforms/collectives directory?

xla/service/gpu/gpu_compiler.cc
```diff
@@ -916,6 +917,11 @@ absl::Status RunCollectiveOptimizationPasses(
 
   HloPassPipeline collectives_pipeline("collective-optimizations");
   collectives_pipeline.AddPass<RaggedAllToAllCanonicalizer>();
+
+  if (debug_options.xla_gpu_experimental_enable_nvshmem()) {
+    collectives_pipeline.AddPass<CollectiveBackendAssigner>(gpu_version);
```
allanrenucci: Should we run this pass after collective combiners? Combiners will increase the size of the collectives.

xla/service/gpu/transforms/collectives/collective_ops_utils.h
```diff
@@ -34,10 +34,16 @@ enum class GPUCommunicationType {
...
 
+// Returns the type of communication pattern for a collective instruction.
 absl::StatusOr<GPUCommunicationType> CommunicationType(
     const HloCollectiveInstruction& instr,
     const se::GpuComputeCapability& gpu_version);
 
+// Returns the type of communication pattern for a channel instruction.
+absl::StatusOr<GPUCommunicationType> CommunicationType(
+    const HloChannelInstruction& instr,
+    const se::GpuComputeCapability& gpu_version);
```
allanrenucci: Why don't we extend the existing `CommunicationType` function instead of adding a new overload?

xla/service/gpu/transforms/collective_backend_assigner_test.cc
```diff
@@ -0,0 +1,266 @@
...
+  std::string usage = tsl::Flags::Usage(argv[0], flag_list);
+  tsl::Flags::Parse(&argc, argv, flag_list);
+  testing::InitGoogleTest(&argc, argv);
+  if (rank_id >= 0) {
+    return xla::gpu::CollectiveBackendAssignerTestBody(rank_id, num_ranks,
+                                                       hlo_module)
+        .raw_code();
+  }
+  return RUN_ALL_TESTS();
+}
```
allanrenucci: LINT error: missing new line

xla/service/gpu/transforms/collective_backend_assigner.h
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  absl::StatusOr<bool> Run(
+      HloModule* module,
+      const absl::flat_hash_set<absl::string_view>& execution_threads) override;
+
+  static bool IsCollectiveOp(const HloInstruction* instr);
+
+ private:
+  static int64_t GetShapeSize(const Shape& shape);
```
allanrenucci: Please move helper functions to the anonymous namespace and remove from header file.

xla/service/gpu/transforms/collective_backend_assigner_test.cc
```diff
@@ -0,0 +1,266 @@
...
+  xla::AppendDebugOptionsFlags(&flag_list);
+  std::string usage = tsl::Flags::Usage(argv[0], flag_list);
+  tsl::Flags::Parse(&argc, argv, flag_list);
+  testing::InitGoogleTest(&argc, argv);
+  if (rank_id >= 0) {
+    return xla::gpu::CollectiveBackendAssignerTestBody(rank_id, num_ranks,
+                                                       hlo_module)
+        .raw_code();
+  }
+  return RUN_ALL_TESTS();
```
allanrenucci: Why are we overriding the main function and running the entire compiler pipeline? Can we just test the new pass? I.e.

```c++
using CollectiveBackendAssignerTest = HloHardwareIndependentTestBase;

TEST_F(CollectiveBackendAssignerTest, ...) {
  absl::string_view hlo_text = ...
  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo_text));
  EXPECT_THAT(RunHloPass(
        CollectiveBackendAssigner(...), module.get()), IsOkAndHolds(true));
  // Validate modified HLO module
  ...
}
```

E.g. https://github.com/openxla/xla/blob/49901275672e96c604bdba657f937f632af8afea/xla/service/gpu/transforms/collectives/collective_combiner_annotator_test.cc#L47-L74

allanrenucci (changes requested):

xla/service/gpu/gpu_compiler.cc
sfvaroglu: Unless we do explicit coalescing of buffers, it shouldn't matter. I'll add this as a to-do to see if it helps in that case.

xla/service/gpu/transforms/collectives/collective_ops_utils.h
sfvaroglu: The code looks cleaner

xla/service/gpu/transforms/collective_backend_assigner.h
sfvaroglu: Done

xla/service/gpu/transforms/collective_backend_assigner_test.cc
sfvaroglu: We want to test the entire stack here (one process per device), similar to other nvshmem PRs.

xla/service/gpu/transforms/collectives/collective_backend_assigner.cc
```diff
@@ -0,0 +1,119 @@
...
+#include "xla/stream_executor/device_description.h"
+
+namespace xla {
+namespace gpu {
+
+bool CollectiveBackendAssigner::IsCollectiveOp(const HloInstruction* instr) {
+  return HloPredicateIsOp<HloOpcode::kAllReduce, HloOpcode::kAllReduceStart,
+                          HloOpcode::kCollectivePermute,
+                          HloOpcode::kCollectivePermuteStart>(instr);
+}
```
allanrenucci: Can this be moved to the anonymous namespace as well?

xla/service/gpu/transforms/collectives/collective_backend_assigner.cc
```diff
@@ -0,0 +1,119 @@
...
+
+namespace {
+
+int64_t GetShapeSize(const Shape& shape) {
+  int64_t size_in_bytes = 0;
+  if (shape.IsTuple()) {
+    for (int64_t i = 0; i < shape.tuple_shapes_size(); ++i) {
```
allanrenucci:
```suggestion
  if (shape.IsTuple()) {
    int64_t size_in_bytes = 0;
    for (int64_t i = 0; i < shape.tuple_shapes_size(); ++i) {
```

xla/service/gpu/gpu_compiler.cc
allanrenucci: Collective combiners currently drop the backend configs for combined instructions. Is it desirable to defaults to NCCL for combined instructions?

xla/service/gpu/transforms/collectives/collective_backend_assigner.cc
```diff
@@ -0,0 +1,119 @@
...
+      VLOG(1) << "CollectiveBackendAssigner: comm_type="
+              << static_cast<int>(comm_type)
+              << " shape_size=" << GetShapeSize(instr->shape())
+              << " threshold_in_bytes_=" << threshold_in_bytes_;
+      auto backend = comm_type == GPUCommunicationType::SINGLE_HOST &&
+                             GetShapeSize(instr->shape()) < threshold_in_bytes_
+                         ? CollectiveBackendConfig::NVSHMEM
+                         : CollectiveBackendConfig::DEFAULT;
+
+      backend_config.set_backend(backend);
```
allanrenucci: We don't need to modify the instruction if `backend == DEFAULT`. How about we rewrite as follow:

```c++
GPUCommunicationType comm_type = ...
bool use_nvshmem = comm_type == GPUCommunicationType::SINGLE_HOST &&
    GetShapeSize(instr->shape()) < threshold_in_bytes_;
if (!use_nvshmem) {
  continue;
}

GpuBackendConfig gpu_config = ...
gpu_config.mutable_collective_backend_config()->set_backend(CollectiveBackendConfig::NVSHMEM);
TF_RETURN_IF_ERROR(instr->set_backend_config(gpu_config));
changed = true;
```

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
```diff
@@ -137,6 +137,57 @@ absl::StatusOr<GPUCommunicationType> CommunicationType(
...
+    CommunicationMetadata comm{node_to_participant_count};
+    if (IsSingleHost(comm)) {
+      return GPUCommunicationType::SINGLE_HOST;
+    }
+    if (IsRailAligned(comm, num_devices_per_host)) {
+      return GPUCommunicationType::RAIL_ALIGNED;
+    }
+    if (IsNonRailAligned(comm, num_devices_per_host)) {
+      return GPUCommunicationType::NON_RAIL_ALIGNED;
+    }
```
allanrenucci: Let's not duplicate this logic. We can change `CommunicationContext` to accept `HloChannelInstruction`.

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
```diff
@@ -137,6 +137,57 @@ absl::StatusOr<GPUCommunicationType> CommunicationType(
   return GPUCommunicationType::UNDEFINED;
 }
 
+absl::StatusOr<GPUCommunicationType> CommunicationType(
```
allanrenucci: Do we need two `CommunicationType` overloads? `HloCollectiveInstruction` is a subtype of `HloChannelInstruction`. I believe we just need to keep the `HloChannelInstruction` overload.

xla/service/gpu/transforms/collective_backend_assigner_test.cc
allanrenucci: Please write small, self contained, preferably hardware independent unit tests for compiler passes.

We can have end to end tests for the entire stack but these should not replace small unit tests.

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
allanrenucci: Please add unit test(s) for `CommunicationType` in collective_ops_utils_test.cc

allanrenucci (changes requested):

xla/service/gpu/transforms/collectives/collective_backend_assigner.cc
sfvaroglu: done

xla/service/gpu/transforms/collectives/collective_backend_assigner.cc
sfvaroglu: done

xla/service/gpu/transforms/collectives/collective_backend_assigner.cc
sfvaroglu: done

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
sfvaroglu: done

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
sfvaroglu: done

sfvaroglu: added

sfvaroglu: looks like there are unit tests in main, so removing mine.

allanrenucci: Let's add a test that covers `HloCollectivePermuteInstruction` since this logic is new

allanrenucci (changes requested):
Please address the following comments:

* https://github.com/openxla/xla/pull/23829#discussion_r2126070264
* https://github.com/openxla/xla/pull/23829#discussion_r2126231511
* https://github.com/openxla/xla/pull/23829#discussion_r2127170662

xla/service/gpu/transforms/collective_backend_assigner_test.cc
sfvaroglu: Added

xla/service/gpu/gpu_compiler.cc
sfvaroglu: Right. NVSHMEM is expected to help with small messages, so it is unlikely to see improvement after the combiners. We would still need to launch collectives for each tensor individually, not the combined size.

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
sfvaroglu: done

sfvaroglu:
> Please address the following comments:
> 
> * [Add CollectiveBackendAssigner pass for NVSHMEM vs. NCCL #23829 (comment)](https://github.com/openxla/xla/pull/23829#discussion_r2126070264)
> * [Add CollectiveBackendAssigner pass for NVSHMEM vs. NCCL #23829 (comment)](https://github.com/openxla/xla/pull/23829#discussion_r2126231511)
> * [Add CollectiveBackendAssigner pass for NVSHMEM vs. NCCL #23829 (comment)](https://github.com/openxla/xla/pull/23829#discussion_r2127170662)

Addressed all. Please let me know if I missed anything. Thank you!

xla/service/gpu/transforms/collectives/collective_backend_assigner_test.cc
```diff
@@ -0,0 +1,334 @@
...
+ protected:
+  absl::StatusOr<bool> RunCollectiveBackendAssigner(HloModule* module) {
+    se::GpuComputeCapability gpu_version = se::CudaComputeCapability(8, 0);
+    return RunHloPass(
+        CollectiveBackendAssigner(gpu_version, /*num_devices_per_host=*/2),
+        module);
+  }
+
+  void VerifyNvshmemBackendConfig(const HloInstruction* instr) {
+    ASSERT_TRUE(instr->has_backend_config());
```
allanrenucci: The ASSERT macros are adjusting control flow by virtue of having a hidden return. It may lead to surprising behavior when used in helper functions (it returns form the helper function, does not abort the whole test). Prefer leaving assertions in the main test body.

Copied from internal style guide:

> Write an appropriate helper function to call within the expectation that holds the context, and signal success or failure with its return value, such as absl::Status or absl::StatusOr. Of course, such a function should be a pure function, since it's within a test.
>
> You can then write your test to EXPECT or ASSERT on the result of the helper function from the test body. This technique can be especially useful for converting data from one format to another when it can fail.

So here, let's define a helper function: 
```cpp
absl::StatusOr<CollectiveBackendConfig> CollectiveBackendConfig(const HloInstruction* instr) {
  TF_ASSIGN_OR_RETURN(GpuBackendConfig gpu_config, instr->backend_config<GpuBackendConfig>());
  return gpu_config.collective_backend_config().backend();
}
```
and then in the main test body:
```cpp
TEST(...) {
  ...
  EXPECT_THAT(CollectiveBackendConfig(inst), IsOkAndHolds(CollectiveBackendConfig::NVSHMEM));
}
```

xla/service/gpu/transforms/collectives/collective_backend_assigner_test.cc
```diff
@@ -0,0 +1,334 @@
...
+        source_target_pairs={{0,1},{1,0}}
+    }
+  )";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(kHloText));
+  EXPECT_THAT(RunCollectiveBackendAssigner(module.get()), IsOkAndHolds(false));
+}
+
+// End-to-end test that verifies the backend assigner works in a real execution
+// environment with multiple ranks.
```
allanrenucci: Thanks for adding small unit tests. These should be in their own file and have their own test target (regular `xla_cc_test`) distinct from the E2E tests. These can be run on any hardware (they don't require a GPU or a dependency on nvshmem) so grouping them with E2E tests defeats their purpose.

IMO, an E2E test here is not warranted. What are you testing that you can't cover with simple unit tests? If you want to test logic outside of `CollectiveBackendAssigner` then this is probably not the right place. Let's remove the E2E test.

allanrenucci (changes requested):

xla/service/gpu/transforms/collectives/collective_backend_assigner_test.cc
sfvaroglu: Updated

xla/service/gpu/transforms/collectives/collective_backend_assigner_test.cc
sfvaroglu: Removed. We were testing the e2e flow, 1 process per GPU that is needed for NVSHMEM.

allanrenucci (approved):
Thank you!

allanrenucci:
Please fix:
```
ERROR: xla/service/gpu/transforms/collectives/BUILD:51:11: in cc_library rule xla/service/gpu/transforms/collectives:collective_backend_assigner: Visibility error:
target 'xla/pjrt/gpu:gpu_topology' is not visible from
target 'xla/service/gpu/transforms/collectives:collective_backend_assigner'
```

xla/service/gpu/transforms/collectives/collective_backend_assigner.h
```diff
@@ -0,0 +1,67 @@
...
+#define XLA_SERVICE_GPU_TRANSFORMS_COLLECTIVES_COLLECTIVE_BACKEND_ASSIGNER_H_
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/status/statusor.h"
+#include "absl/log/log.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/hlo/pass/hlo_pass_interface.h"
+#include "xla/stream_executor/device_description.h"
+#include "xla/service/gpu/backend_configs.pb.h"
+#include "xla/pjrt/gpu/gpu_topology.h"
```
allanrenucci: I think you can remove this header and remove the dependency

sfvaroglu: Removed, thank you!

allanrenucci (approved):

PR #23838, opened 2025-03-17
[ds-fusion] Remove workspace checks in dynamic slice fusion

shraiysh:
The contents of the workspace are not guarenteed to be the same after a cublas gemm operation. We were comparing the contents of the workspace. This patch removes such checks and only looks for the value of the results.

### DIFF
--- a/xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
+++ b/xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
@@ -2213,6 +2213,99 @@ TEST_F(DynamicSliceFusionTest, CublasGemmDUS) {
       /*run_hlo_passes=*/false));
 }
 
+TEST_F(DynamicSliceFusionTest, CublasGemmDUSWithWorkspace) {
+  ErrorSpec error_spec{/*aabs=*/1e-3, /*arel=*/1e-3};
+
+  const char* hlo_ref = R"(
+  HloModule jit_slice
+  ENTRY %main.9 {
+    %p0 = f16[2,8,8]{2,1,0} parameter(0)
+    %p1 = f16[2,8,8]{2,1,0} parameter(1)
+    %p2 = f16[4,8,8]{2,1,0} parameter(2)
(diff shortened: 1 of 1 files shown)


xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
```diff
@@ -2197,105 +2213,6 @@ TEST_F(DynamicSliceFusionTest, CublasGemmDUS) {
       /*run_hlo_passes=*/false));
 }
 
-TEST_F(DynamicSliceFusionTest, CublasGemmDUSWithWorkspace) {
```
bchetioui, line 2200: Why do we delete this test? Can't we just make a similar rewrite as for the others? (Return the GTE at the root?)

bchetioui (changes requested):

xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
shraiysh: I thought it becomes very much like the previous testcase: `CublasGemmDUS`. Should I make it such that the fusion should return workspace (in tuple), but we should just gte the result from that at "entry" root?

bchetioui: I thought about it a bit, but I think it's still a bit subtly different: we do need to allocate an HBM buffer for the workspace---right? I think the coverage might still be useful.

I like the solution you suggest (`gte` at the entry root).

bchetioui (changes requested):
Forgot to request the change, sorry about that :)

Just make it a GTE as discussed maybe, and then we can land this!

xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
shraiysh: Done. Marking as "resolved".

bchetioui (approved):

PR #23861, opened 2025-03-18
[ROCm] Use MIOPEN_BETA_API for miopenSetTensorDescriptorV2

hsharsha:


hsharsha:
@beckerhe Can you please review this?

i-chaochen:
Hi @xla-rotation please take another look

beckerhe (approved):

reedwm:
Is it possible to reduce or remove the use of `#if`s and `#ifdef`s? We are trying to reduce the number of these, and have an internal check that prohibits changes from adding these, but I can disable the  check.

xla/stream_executor/rocm/rocm_dnn.cc
```diff
@@ -551,6 +556,12 @@ namespace wrap {
 // clang-format on
 #endif
 
+#ifdef MIOPEN_BETA_API
+#define MIOPEN_DNN_ROUTINE_EACH(__macro)                             \
+  __macro(miopenSetTensorDescriptorV2)                          \
```
reedwm, line 561: Also, I think this redefines the `MIOPEN_DNN_ROUTINE_EACH` macro. Did you mean to just add `__macro(miopenSetTensorDescriptorV2)` to the macro definition?

hsharsha:
@reedwm  and @beckerhe  I have addressed the review comments and tried to minimize use of #ifdefs.  In order to support ROCm 6.2 and lower, I would still need this one #define.

xla/stream_executor/rocm/BUILD
```diff
@@ -496,6 +497,7 @@ cc_library(
     name = "miopen_plugin",
     srcs = ["rocm_dnn.cc"],
     hdrs = ["rocm_dnn.h"],
+    defines = if_rocm_miopen_beta_api_supported(["MIOPEN_BETA_API=1"]),
```
beckerhe, line 500: This can be a `local_defines`, right?

beckerhe (approved):

xla/stream_executor/rocm/BUILD
hsharsha: @beckerhe made it local_defines

xla/stream_executor/rocm/rocm_dnn.cc
```diff
@@ -551,6 +551,10 @@ namespace wrap {
 // clang-format on
 #endif
 
+#ifdef MIOPEN_BETA_API
```
beckerhe, line 554: Hi @hsharsha,
internally the question came up, why we don't just check the ROCM version directly here using the `ROCM_VERSION_MAJOR`/`ROCM_VERSION_MINOR` macros from rocm_version.h?

beckerhe: This would avoid making changes to Bazel altogether

hsharsha: Initially I tried with TF_ROCM_VERSION macro but the comparison check was not successful for different rocm versions and failed to set MIOPEN_BETA_API macro which needs to be set before including miopen.h. So I resorted to making changes in bazel to support for different rocm versions.

chsigg: Could we `local_defines = ['MIOPEN_BETA_API=1']` unconditionally and then use the beta API based on the `TF_ROCM_VERSION` value?

hsharsha: @chsigg addressed your review comments. Now the change is localized to rocm_dnn.h and rocm_dnn.cc

hsharsha: @chsigg  Reworked as per your suggestion.

hsharsha:
@chsigg  Can you please review the latest changes. Thanks

chsigg (approved):

PR #23864, opened 2025-03-18
[XLA:GPU] Lower bitcast-convert as no-op

sergey-kozub:
HLO `bitcast-convert` op allows converting between types of different size (e.g. u4[2]->u8 or u8->u4[2]).

Currently, the `BitcastDtypesExpander` pass expands such conversions into a sequence of operations, but this could be avoided, as this is essentially a no-op. This PR removes that pass from the GPU pipeline.

XLA assumes little-endian memory layout on GPUs (We have some code like https://github.com/openxla/xla/blob/3bb765472122548cc227b8bd2990f00bd533f438/xla/codegen/emitters/transforms/lower_tensors.cc#L1084), so a simple bitcast works.

MLIR tensor dialect doesn't support bitcasting both the ranked tensor shape and the element type and the same time, so I'm emitting a "mhlo.bitcast_convert" op which is replaced by an "unrealized_conversion_cast" during lowering (and subsequently elided).


akuegel:
Thanks for this PR. We actually had an internal bug open for this, but my assessment was that we cannot always replace bitcast-convert with a bitcast, it depends on the chosen layout whether it actually is a bitcast or not. We can change layout assignment pass to propagate a layout that makes it a bitcast, but that only works in one direction.

Moreover, bitcasts inside fusions are not a no-op, they change indexing. I don't understand yet how this should work with just using UnrealizedConversionCast, but the main point is also that inside fusions, we could just as well use the pattern from the expansion for BitcastConvert. Only unfused bitcast-converts can be turned into a real no-op.
My idea to tackle this (until the bug was closed due to not high enough value):

We would add an AlgebraicSimplifier pattern and guard that with a flag, and only turn the flag on when we run AlgebraicSimplifier after the fusion passes. AlgebraicSimplifier doesn't touch fusion computations, so it would only replace the bitcast-convert ops outside fusions. Then we can run the BitcastDtypesExpander and expand the fused bitcast-converts inside fusions.

sergey-kozub:
Could you come up with an HLO example (with this PR applied) which would fail comparing to the CPU backend?

I understand that a non-default shape layout could make a bitcast-convert not a no-op, but I so far I assumed that the layout normalization pass takes care of this.

akuegel:
> Could you come up with an HLO example (with this PR applied) which would fail comparing to the CPU backend?
> 
> I understand that a non-default shape layout could make a bitcast-convert not a no-op, but I so far I assumed that the layout normalization pass takes care of this.

How should the layout normalization pass be able to handle it though? Layout normalization works by swapping logical dimensions in the shape together with physical dimensions, so it is still equivalent. But by doing that, don't we change semantics of BitcastConvert op? BitcastConvert op assumes that the most minor dimension is the one to be combined or splitted. If we move the most minor dimension in the shape, we change semantics of BitcastConvert.

Coming up with an example is maybe a bit more involved, as we need to find something where we would choose a non-default layout. But just from the theory side I am relatively sure that we cannot just assume that it will have default layout.

sergey-kozub:
Ok, I see. Marking as draft for now, I will think about this.

PR #23995, opened 2025-03-20
[ROCm] update for hlo verification fix for the change of instruction names during rematerialization 

cj401-amd:
@xla-rotation could you please kindly help review this PR? Thanks

During the rematerialization process, instruction names are uniquified using the algorithm in name_uniquer.cc (https://github.com/openxla/xla/blob/main/xla/service/name_uniquer.cc#L76). This appends an incremented number to the original name (e.g., adding "1" to a name without a number, or incrementing existing numbers like "1" to "2").
However, during final HLO verification, the system checks invariants and produces errors like:

```
jaxlib.xla_extension.XlaRuntimeError: FAILED_PRECONDITION: during context [hlo verifier]: Expected instruction name to remain the same. Was ***.1 is ***.2.
```
The code comments indicate "We do not enforce the invariant when the instruction has been cloned explicitly via .clone or .remat suffix," but this isn't sufficient. In this PR, we're broadening the exception to: "We do not enforce the invariant when the computation has been cloned."

### DIFF
--- a/xla/service/hlo_verifier.cc
+++ b/xla/service/hlo_verifier.cc
@@ -2186,6 +2186,10 @@ absl::Status VerifyInstructionNameUnchanged(const HloModule& module,
     return absl::OkStatus();
   }
   for (auto* comp : module.computations()) {
+    // We do not enforce the invariant when the computation has been cloned, is it too general?
+    if (absl::StrContains(comp->name(), ".clone")) {
+      continue;
+    }
     for (auto* inst : comp->instructions()) {
       if (inst->metadata().scheduling_name().empty()) {
         continue;
@@ -2917,6 +2921,12 @@ class InstructionVerifier : public DfsHloVisitorWithDefault {
   }
 
   absl::Status Preprocess(HloInstruction* instruction) override {
+    // skip processing of instructions that are cloned from remat
+    const HloComputation* compt = instruction->parent();
+    bool is_remat_cloned = absl::StrContains(instruction->name(), ".clone");
+    if (is_remat_cloned) {
+      return absl::OkStatus();
+    }
     auto [it, inserted] =
         instructions_by_name_.emplace(instruction->name(), instruction);
     TF_RET_CHECK(inserted) << "HLO has name that is not unique within module:\n"
--- a/xla/service/hlo_verifier_test.cc
+++ b/xla/service/hlo_verifier_test.cc
@@ -3794,5 +3794,160 @@ ENTRY main {
       HasSubstr("Instruction shouldn't have the layout of host memory space"));
 }
 
+TEST_F(HloVerifierTest, ClonedComputationNameBypass) {
+  const char* const hlo_string = R"(
+  HloModule test_module
+
+  computation {
+    p0 = f32[] parameter(0)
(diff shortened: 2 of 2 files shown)


xla/service/hlo_verifier.cc
```diff
@@ -2186,6 +2186,10 @@ absl::Status VerifyInstructionNameUnchanged(const HloModule& module,
     return absl::OkStatus();
   }
   for (auto* comp : module.computations()) {
+    // We do not enforce the invariant when the computation has been cloned, is it too general?
+    if (absl::StrContains(comp->name(), ".clone")) {
```
golechwierowicz, line 2190: IIRC this verification is on only when you run the XLA program with 

```
--xla_gpu_pgle_accuracy_checker='PGLE_STRICTNESS_LEVEL_ERROR'
```

Because in gpu_compiler pipeline remat happens post scheduling, it can be a case that computation gets rewritten and cloned and when a program is profiled we cannot match the original computation in the subsequent compilation. Since we run remat by default, this verification is also disabled by default.

Can you elaborate a little bit on the motivation of this change? What's your use case?

cj401-amd: The main motivation behind this change is to get pass HLO verification after remat which checks the invariants of the names of instructions and sub-instructions. 
In our experiments, we did not set up 
```
--xla_gpu_pgle_accuracy_checker='PGLE_STRICTNESS_LEVEL_ERROR'
```
However, some instructions along with the embedded sub-instructions change names during the process of remat, e.g., instruction names are uniquified using the algorithm in name_uniquer.cc (https://github.com/openxla/xla/blob/main/xla/service/name_uniquer.cc#L76). This appends an incremented number to the original name (e.g., adding "1" to a name without a number, or incrementing existing numbers like "1" to "2").
This leads to the failure of invariant checking. I believe there is a incompatability between the remat and invariant checking of HLO verifier. That's why we propose the above change, from which our experiments run smoothly.

golechwierowicz: I understand the failure of the invariant violated. I'm saying this behaviour is on only for the very specific use case, and is enabled only here: https://github.com/openxla/xla/blob/d505fef9c5eb6cc1bf282fdf62139783d7fe4ec5/xla/service/gpu/gpu_compiler.cc#L2741

Are you enabling the additional verification somewhere else in your codebase via `HloVerifierOpts{}.VerifyInstructionNameUnchanged`?

cj401-amd: Thanks for your prompt reply. I was not aware of `HloVerifierOpts{}.VerifyInstructionNameUnchanged` is enabled in the branch 0.4.31 https://github.com/openxla/xla/blob/d505fef9c5eb6cc1bf282fdf62139783d7fe4ec5/xla/service/gpu/gpu_compiler.cc#L2741.  
I wonder if it is enabled by default or how I can disable it. If it can be disabled by a flag, then the proposal is not relevant anymore.

golechwierowicz: It's disabled by default.

dimitar-asenov:
Closing this PR as it seems there's an alternative solution. If you would still like to merge this, please reopen the PR.

PR #24023, opened 2025-03-21
Remove HloComputation::CustomCallInstruction.

jreiffers:
Step 2/5 of removing instruction_type. I think kFusion should be last, since it's special (in that it's "sticky", you can't tell from instruction_callers whether it's a fusion computation).


akuegel (approved):
Thank you!

PR #24026, opened 2025-03-21
Remove call graph from copy fusion logic.

jreiffers:
Now that we have a call graph in HloComputation, we no longer need this.


jreiffers:
@akuegel Here's the cleanup I promised :-)

akuegel (approved):
Very nice :)

PR #24028, opened 2025-03-21
Fix some Shape::rank deprecation warnings.

jreiffers:
These headers are included from absolutely everywhere, and the warning spam can make it quite hard to see the actual problems.


allanrenucci (approved):

allanrenucci:
FYI I am marking these functions as candidate for inlining in #24105. Our automation should take care of updating deprecated references.

PR #24060, opened 2025-03-21
Exclude Circular Dependencies in FP8 Graph Convolutions

philipphack:
Excludes ops with external operands that can be reached from graph outputs of fused ops as well as ops with graph outputs that can reach external operands of fused ops from fusion into graph-based FP8 convolutions. Fusing these ops may lead to circular dependencies between fused and unfused instructions.

### DIFF
--- a/xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
+++ b/xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
@@ -765,11 +765,14 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
     ++num_endpoints;
   }
 
-  // Since the cuDNN graph cannot have more than one endpoint, do not fuse
-  // into the cuDNN convolution Custom Call and roll back the graph when there
-  // are multiple endpoints. If the resulting graph still has more than one
-  // endpoint, the recursive caller will continue to roll back the graph.
-  if (num_endpoints > 1) {
+  // Do not fuse into the cuDNN convolution Custom Call and roll back the graph
+  // when the number of users eligible for fusion is less than the total number
+  // of users of instr. Since the cuDNN graph cannot have more than one
+  // endpoint, also roll back the graph when there are multiple endpoints. If
+  // the resulting graph still has more than one endpoint, the recursive caller
+  // will continue to roll back the graph.
+  if (num_new_users + num_existing_users < instr->user_count() ||
+      num_endpoints > 1) {
     graph_string = std::move(init_graph_string);
     operands = init_operands;
     aux_outputs = init_aux_outputs;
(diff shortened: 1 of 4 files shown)


xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -568,18 +569,43 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
...
+  // External operands of ops eligible for fusion into the cuDNN graph must not
+  // be reachable from graph outputs of existing fused ops as this may create a
+  // circular dependency between fused and unfused instructions.
+  auto eligible_operand =
+      [reachability, &aux_outputs](const HloInstruction* operand) -> bool {
+    return std::none_of(aux_outputs.begin(), aux_outputs.end(),
+                        [reachability, operand](HloInstruction* aux_output) {
+                          return reachability->IsReachable(aux_output, operand);
+                        });
+  };
```
reedwm, line 582: Shouldn't you check normal outputs as well? E.g. what if you have the graph:

```
    ┌────┐  
    │conv│  
    ├────┤  
    │    │  
    │    │  
┌───▼┐   │  
│sine│   │  
└────┤   │  
     │   │  
     │   │  
     │   │  
     │   │  
     │   │  
     │   │  
     │   │  
    ┌▼───▼─┐
    │add   │
    └┬─────┘
     │      
    ┌▼───┐  
    │amax│  
    └────┘  
```

`conv`, `add`, and `amax` are in the cudnn graph. `conv` is an endpoint and `amax` is an aux output. But this graph is invalid since `sine` is both an input and uses an output of the graph. Yet you don't catch this, since you only check if there's a path from aux outputs, not normal endpoints.

Getting this right is a bit tricky, since you don't necessarily know the final endpoint yet. Maybe you should check reachability after the entire cudnn graph fusion is built, and give up on the fusion if there are reachability issues.

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -640,14 +670,16 @@ void CaptureConvGraphRecursive(HloInstruction* instr,
                 m::Broadcast(&op0, m::ConstantEffectiveScalar(0)).WithOneUser(),
                 m::Op().Is(instr),
                 m::Broadcast(&operand0, m::ConstantEffectiveScalar(6))
-                    .WithOneUser()))) {
+                    .WithOneUser())) &&
+        eligible_operand(operand0)) {
```
reedwm, line 674: operand0 is constant, so it cannot possibly have a dependency on an aux output. And similarly for leaky relu.

xla/service/gpu/transforms/cudnn_fused_conv_rewriter.cc
```diff
@@ -758,6 +792,9 @@ CaptureConvGraph(HloInstruction* instr, HloInstruction* convolution,
   GraphString graph_string;
   graph_string.AppendOp("conv", instr);
 
+  HloReachabilityMap* reachability =
+      HloReachabilityMap::Build(instr->parent()).release();
```
reedwm, line 796: Because you call release(), the pointer is never freed.

reedwm (changes requested):

reedwm (approved):

PR #24087, opened 2025-03-23
[GPU] Fix in-place updates in command buffers.

sergachev:


sergachev:
The GetUsers() traversal does not handle nested tuples correctly - for instance, in the added test the root tuple of the command buffer becomes a "user" of the DUS, which makes the fusion non-in-place.
I think it's OK to use normal HloInstruction traversal here when a fusion is already constructed and we don't need to traverse outside it instead of HloInstructionAdaptor but I'm open to better ideas how to fix this.

akuegel:
Thanks for the fix, but I wonder whether we should fix HloInstructionAdaptor instead? I need to double-check, but it could be that this is unintentional.

akuegel:
> Thanks for the fix, but I wonder whether we should fix HloInstructionAdaptor instead? I need to double-check, but it could be that this is unintentional.

Ok, I see that I tried to change it in the past to not return users outside of the FusionAdaptor parent, but apparently I did not succeed with this. There is even a test in hlo_traversal_test.cc that shows this issue. Not sure how I missed it. It seems currently it is not in a consistent state. Either we should return all users, or only those inside the FusionAdaptor. I think for the use cases we have, it seems better to only return those inside the FusionAdaptor, and that would also fix the problem you discovered (as far as I understand).

akuegel:
https://github.com/openxla/xla/pull/24088 should fix it. Please let me know in case it doesn't.

PR #24181, opened 2025-03-25
[ds-fusion] Fix the issue with JAX tests with no indvars

shraiysh:
JAX has some testcases with no induction variable but static slices inside the while loop. Such slices are fused with the hero operation, but while lowering they fail on an assert (which is not required).

Added a test to demonstrate this.

### DIFF
--- a/xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
+++ b/xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
@@ -3703,6 +3703,66 @@ TEST_F(DynamicSliceFusionTest,
       /*run_hlo_passes=*/false, /*use_threads=*/true, std::nullopt));
 }
 
+TEST_F(DynamicSliceFusionTest, WhileLoopSliceWithNoInductionVariable) {
+  const char* hlo = R"(
+  HloModule test, replica_count=2
+
+  add {
+    a = f32[] parameter(0)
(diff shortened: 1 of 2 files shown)


frgossen (approved):
Tnanks for the fix

xla/backends/gpu/codegen/dynamic_slice_fusion_test.cc
```diff
@@ -3703,6 +3703,66 @@ TEST_F(DynamicSliceFusionTest,
...
+      .set_xla_gpu_enable_dynamic_slice_fusion(false);
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> m_ref,
+                          GetOptimizedModule(m->Clone()));
+  m->mutable_config()
+      .mutable_debug_options()
+      .set_xla_gpu_enable_dynamic_slice_fusion(true);
+  TF_ASSERT_OK_AND_ASSIGN(m, GetOptimizedModule(std::move(m)));
+  // VLOG(0) << "Fused module: " << m->ToString();
+  ErrorSpec error_spec(1e-5, 1e-5);
+  RunAndCompareTwoModulesReplicated(std::move(m), std::move(m_ref),
```
akuegel, line 3762: This is ignoring the return value, and thus would ignore any errors. Can you please fix it?

akuegel (approved):

PR #24248, opened 2025-03-26
Move-to-memory-space custom calls use default layout

jaro-sevcik:
Using non-default layout for MoveToHost makes the compiler
insert a transpose operation on the host value if that value
flows into the root of the entry computation. Such transposes
cause the host offloader to emit a slow on-host transpose
(and a warning on the console). We see those warnings on 
maxtext llama2-7b with optimizer state offloading with fsdp=2.

This patch enforces default layout for on-host values so that
no transpose is necessary (as long as there is no override
of layout for host values in entry computation).

Note that such transposes cannot be sunk into the uses
by the offloading lagalizer because there is nowhere
to sink to - the value is returned from the computation.

### DIFF
--- a/xla/service/gpu/transforms/layout_assignment.cc
+++ b/xla/service/gpu/transforms/layout_assignment.cc
@@ -340,6 +340,15 @@ bool IsPackedInstruction(const HloInstruction* instruction) {
               instruction->operand(0)->shape().element_type()));
 }
 
+bool IsCustomCallToMemoryPlacement(const HloInstruction* hlo) {
+  if (hlo->opcode() != HloOpcode::kCustomCall) {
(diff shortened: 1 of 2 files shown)


mooskagh:
This change kind of reverts your earlier change https://github.com/openxla/xla/pull/10495 — just want to confirm that this is intended.

jaro-sevcik:
> This change kind of reverts your earlier change #10495 — just want to confirm that this is intended.

Yes, this is intended.

This change puts a stricter constraint in place. #10495 constrained input and output layouts to be the same, this PR insists that both the input and the output layouts are the default layout.

mooskagh (approved):

PR #24255, opened 2025-03-26
Fixed rules_python patch for python 3.14

vfdev-5:
Returning from finally clause raises SyntaxWarning in python 3.14

cc @hawkinsp

### DIFF
--- a/third_party/py/rules_python.patch
+++ b/third_party/py/rules_python.patch
@@ -72,15 +72,14 @@ index 774c24d1..91e59f9b 100644
  }
 
 diff --git a/python/private/python_bootstrap_template.txt b/python/private/python_bootstrap_template.txt
-index 0f9c90b3..6d1e2f61 100644
+index 0f9c90b3..567bdc88 100644
 --- a/python/private/python_bootstrap_template.txt
 +++ b/python/private/python_bootstrap_template.txt
-@@ -52,8 +52,16 @@ def GetWindowsPathWithUNCPrefix(path):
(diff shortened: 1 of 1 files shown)


third_party/py/rules_python.patch
```diff
@@ -72,25 +72,22 @@ index 774c24d1..91e59f9b 100644
...
 +  try:
 +    version = platform.win32_ver()[1]
 +  except (ValueError, KeyError):
 +    version = platform.win32_ver()[1]
-+  finally:
-+    if version and version >= '10.0.14393':
-+      return path
- 
++  if version and version >= '10.0.14393':
+     return path
```
allanrenucci, line 91: This code is confusing. Let's rewrite as:
```python
# The try-except block is needed to fix the flakiness of RBE tests
# on Windows 2022 using hermetic python 3.12.8.
try:
  if platform.win32_ver()[1] >= '10.0.14393':
    return path
except (ValueError, KeyError):
  pass
  
```

allanrenucci (changes requested):

third_party/py/rules_python.patch
vfdev-5: I agree that this code and the original patch code are not clear. I have an impression that the original code wanted to execute `version = platform.win32_ver()[1]` twice in case the first call raises an exception.

allanrenucci: I did not realise `platform.win32_ver()[1]` could fails non-deterministically. Then LGTM.

allanrenucci (approved):

PR #24269, opened 2025-03-27
Fix for fusion wrapper on async computations

chaserileyroberts:
Previously, in JAX if you do a simple stream annotated computation with just a single instruction, i.e.

```python
@compute_on('gpu_stream:1')
@jax.jit
def h(x, y):
  return x * y
```

Then the XLA compiler would fail with an error like

```
jaxlib.xla_extension.XlaRuntimeError: INTERNAL: Unsupported instruction opcode: multiply
```

This is because the fusion wrapper would not correctly look inside of async computations, leaving the instructions unfused. 

To fix this, we simply add `kCall` and `kAsyncStart` to the set of instructions that are treated recursively in the fusion wrapper.

### DIFF
--- a/xla/codegen/emitters/fusion_wrapper_base.cc
+++ b/xla/codegen/emitters/fusion_wrapper_base.cc
@@ -37,7 +37,8 @@ absl::StatusOr<bool> FusionWrapperBase::Run(
   std::function<absl::Status(HloInstruction*)> handle_instruction;
   handle_instruction = [&](HloInstruction* instruction) -> absl::Status {
     const HloOpcode opcode = instruction->opcode();
-    if (opcode == HloOpcode::kConditional || opcode == HloOpcode::kWhile) {
+    if (opcode == HloOpcode::kConditional || opcode == HloOpcode::kWhile ||
+        opcode == HloOpcode::kCall || opcode == HloOpcode::kAsyncStart) {
       for (auto* computation : instruction->called_computations()) {
         for (auto* inner_instruction :
              computation->MakeInstructionPostOrder()) {
--- a/xla/service/gpu/tests/BUILD
+++ b/xla/service/gpu/tests/BUILD
@@ -115,17 +115,17 @@ xla_test(
     ],
 )
 
-xla_cc_test(
+xla_test(
     name = "async_kernel_launch_test",
     srcs = ["async_kernel_launch_test.cc"],
     # "requires-net:external" tag allows uploading `xprof` results.
     tags = if_google(["requires-net:external"]) + tf_cuda_tests_tags(),
+    backends = ["gpu"],
     deps = [
         "//xla:debug_options_flags",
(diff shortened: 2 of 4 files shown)


cota (approved):

cota (changes requested):
After importing your PR, the added tests don't pass the verifier:

> FAILED_PRECONDITION: during context [Unknown]: Asynchronous computation async_wrapped expected to contain only the root and parameter instructions.

This is the computation:
```
%async_wrapped (async_param: f32[2], async_param.1: f32[2]) -> f32[2] {
  %async_param = f32[2]{0} parameter(0)
  %bitcast.31 = f32[2]{0} bitcast(%async_param)
  %bitcast.32 = f32[2]{0} bitcast(%bitcast.31)
  %async_param.1 = f32[2]{0} parameter(1)
  %bitcast.33 = f32[2]{0} bitcast(%async_param.1)
  %bitcast.34 = f32[2]{0} bitcast(%bitcast.33)
  ROOT %call = f32[2]{0} call(%bitcast.32, %bitcast.34), to_apply=%add_F32
}
```

Were you able to run the tests locally? I can't find them in the Github CI but I am not very familiar with it.

chaserileyroberts:
#WorkedOnMyMachine 

The bitcasts are really strange and ideally shouldn't be there. If we add `{0}` to each of the `f32[2]` to the type descriptions do you think that would fix it?

chaserileyroberts:
I added the additional `{0}` layout information to the dtypes, but still both versions passed on what I was able to test with (GB200s).

ezhulenev:
This is the test that doesn't pass:

```[xla/service/gpu/tests/async_kernel_launch_test.cc:98](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/service/gpu/tests/async_kernel_launch_test.cc?l=98&ws=tap-presubmit-server/344298036&snapshot=2)
Value of: RunAndCompare(hlo_text, ErrorSpec{1e-5, 1e-5})
  Actual: false (FAILED_PRECONDITION: during context [Unknown]: Asynchronous computation async_wrapped expected to contain only the root and parameter instructions.
=== Source Location Trace: ===```

chaserileyroberts:
I figured out why I wasn't getting the same error as the internal CI. The issue was that the test was under a `xla_cc_test` but not a normal `xla_test`, so even when I was compiling as a GPU binary, it was still using the CPU executor in the test. 

I changed the build and I switched to `RunAndCompareNoHloPasses`, now it seems to be working as expected.

cota (approved):

cota:
Thanks for following up. I have reapproved, which will re-trigger our internal testing. Will report back.

cota:
There were unrelated failures that are now solved. Let's try this once again.

cota (approved):

cota (approved):

PR #24293, opened 2025-03-27
Removes wrong and repetitive in IsGpuAsyncDone.

yliu120:


xla/service/gpu/gpu_latency_hiding_scheduler.cc
```diff
@@ -123,15 +122,12 @@ bool IsGpuAsyncStart(const HloInstruction& hlo) {
...
+         IsAsyncComputeStartOp(hlo) || hlo.opcode() == HloOpcode::kCopyStart;
 }
 
 // Marks async done operations to be scheduled as late as possible.
 bool IsGpuAsyncDone(const HloInstruction& hlo) {
-  return (hlo_query::IsAsyncCollectiveDoneOp(&hlo,
-                                             /*include_send_recv=*/true) &&
-          !IsGPUSyncCollective(*hlo.operand(0))) ||
-         IsAsyncComputeOp(hlo) || hlo.opcode() == HloOpcode::kCopyDone;
+  return hlo.operand_count() == 1 && IsGpuAsyncStart(*hlo.operand(0));
```
frgossen, line 130: I think it's safer to check the op type here. I'm not sure if `IsGpuAsyncStart` covers all the possible async starts. Some of these utility functions have missing cases.

frgossen (changes requested):
Can you add test cases for this?

PR #10685, opened 2024-03-19
[ROCM] GPU profiling code refactor - seperarted rocm event tracer and…

zahiqbal:
rocm profiler code refactored, Separated tracing and event collecting into rocm tracer (rocm_tracer.cc/h) and event collector (rocm_collector.cc/h).

PR #24329, opened 2025-03-28
[ROCm] Change logging level for unhandled operators

gabeweisz:
Previously, when profiling, the profiler would print thousands of messages for operators that the profiler did not handle unless all warning messages were dropped.

This could cause full buffers for stdout, which slowed down computation for at least one workload by 6-7x during profiling (~2s vs ~330ms).

This PR changes the default logging for the unhandled operators, which fixes the problem. Using VLOG(1) was suggested by @pemeliya

### DIFF
--- a/xla/backends/profiler/gpu/rocm_tracer.cc
+++ b/xla/backends/profiler/gpu/rocm_tracer.cc
@@ -1561,8 +1561,8 @@ absl::Status RocmTracer::DisableActivityTracing() {
   size_t threshold = 1;
   for (int i = 0; i < 6; i++, duration_ms *= 2, threshold *= 2) {
     if (GetPendingActivityRecordsCount() < threshold) break;
-    VLOG(3) << "Wait for pending activity records :" << " Pending count = "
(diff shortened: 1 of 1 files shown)


i-chaochen (approved):
Thanks! @xla-rotation since this profiler code is rocm only I think we can merge by ourself like this https://github.com/openxla/xla/pull/10685 before?

reedwm:
> Thanks! @xla-rotation since this profiler code is rocm only I think we can merge by ourself like this #10685 before?

@i-chaochen what do you mean? Don't we need to merge this into the XLA repo?

i-chaochen:
> > Thanks! @xla-rotation since this profiler code is rocm only I think we can merge by ourself like this #10685 before?
> 
> @i-chaochen what do you mean? Don't we need to merge this into the XLA repo?

Sorry if I made the confusion. I mean can we directly merge this code to openxla as it's rocm only?

dimitar-asenov (approved):

PR #24429, opened 2025-04-01
Fixed failing NCCL Group test on internal CI

chaserileyroberts:
Previously, this test was failing with 

```
INTERNAL: NCCL operation ncclRecv( recv_buffer.opaque(), ToNcclCount(dtype, count), nccl_dtype, source_rank->value(), comm_, se::gpu::AsGpuStreamValue(stream)) failed: invalid argument (run with NCCL_DEBUG=WARN for details). Last NCCL warning(error) log entry (may be unrelated) 'Recv : invalid root 3 (root should be in the 0..1 range)'.
```

This was because of a mix of the bad channel ids and this missing `replica_count` attribute.

### DIFF
--- a/xla/tests/nccl_group_execution_test.cc
+++ b/xla/tests/nccl_group_execution_test.cc
@@ -127,7 +127,7 @@ XLA_TEST_F(NcclGroupExecutionTest, NcclGroupSendRecvNoWhileLoop) {
 
 XLA_TEST_F(NcclGroupExecutionTest, BidirectionalCommunication) {
   const absl::string_view kModuleStr = R"(
-  HloModule module_main, entry_computation_layout={()->(u32[], u32[])}, replica_count=4
+  HloModule module_main, entry_computation_layout={()->(u32[], u32[])}, num_partitions=4
 
   bidirectional_ring {
     a = u32[] parameter(0)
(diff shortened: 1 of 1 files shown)


appujee (approved):

xla/tests/nccl_group_execution_test.cc
```diff
@@ -127,13 +127,13 @@ XLA_TEST_F(NcclGroupExecutionTest, NcclGroupSendRecvNoWhileLoop) {
 
 XLA_TEST_F(NcclGroupExecutionTest, BidirectionalCommunication) {
   const absl::string_view kModuleStr = R"(
-  HloModule module_main, entry_computation_layout={()->(u32[], u32[])}
+  HloModule module_main, entry_computation_layout={()->(u32[], u32[])}, replica_count=4
```
frgossen, line 130: Shouldnt this be numn_partitions?
I would exoect this to be with channel IDs in replica mode

frgossen (changes requested):

xla/tests/nccl_group_execution_test.cc
chaserileyroberts: I changed it to `num_parititions`, but I still needed to remove the channel ids to make the test pass.

rosiezou (approved):

rosiezou (approved):

xla/tests/nccl_group_execution_test.cc
frgossen: Had another look. You can either run this in replica mode or in partition mode. 

Replica mode: define replica_count, no chanel_id, no device assignement 
Partion mode: define num_partitions, define channel_id, define device assignement 

There is an example here: DeviceAssignment in xla/tests/collective_pipeline_parallelism_test.cc

PR #24491, opened 2025-04-02
[NVIDIA GPU] Collective-permute combiner ignores channel id when enabled

terryysun:
This PR pipes `xla_ignore_channel_id` flag to collective-permute combiner, allowing the combiner to ignore channel id difference when the flag is on.

### DIFF
--- a/xla/service/collective_permute_key.cc
+++ b/xla/service/collective_permute_key.cc
@@ -34,14 +34,7 @@ std::optional<CollectivePermuteKey> GetCollectivePermuteKey(
   }
 
   const auto* cp = Cast<HloCollectivePermuteInstruction>(instruction);
-  return CollectivePermuteKey{cp->source_target_pairs(),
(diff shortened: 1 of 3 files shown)


xla/service/collective_permute_key.cc
```diff
@@ -34,7 +34,14 @@ std::optional<CollectivePermuteKey> GetCollectivePermuteKey(
   }
 
   const auto* cp = Cast<HloCollectivePermuteInstruction>(instruction);
-  return CollectivePermuteKey{cp->source_target_pairs(), cp->channel_id()};
+  return CollectivePermuteKey{cp->source_target_pairs(),
```
frgossen, line 37: I'm wondering if this should always be ignored. Can we try that and see if we run into test failures?

frgossen (changes requested):

frgossen (approved):
Thanks!

PR #24533, opened 2025-04-02
Fix fp8 tests

Ruturaj4:
We observed failures in internal CI for FP8 types `F8E4M3FNUZ` and `F8E5M2FNUZ`. These types should not be implicitly upcast during dot operation operand conversion. This patch extends the existing check that prevents conversions between FP8 types by including these two additional variants.

@xla-rotation @draganmladjenovic can you please take a look?


mrodden (approved):
For additional context, this is fixing some fp8 related tests for the ROCm backend in JAX

reedwm (approved):

PR #24573, opened 2025-04-03
[ROCm] Fixed issue with builtin.module error.

zoranjovanovic-ns:


draganmladjenovic:
@xla-rotation PTAL

xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
```diff
@@ -90,41 +86,47 @@ absl::Status CreateTritonPipeline(mlir::OpPassManager* pm,
...
   pm->addPass(mt::gpu::createTritonGPUOptimizeThreadLocality());
-  pm->addPass(mlir::createTritonAMDGPUAccelerateMatmulPass(cc.gfx_version()));
+  // TODO ROCm Pass cc.gfx_version() after fixing issue with fmfa
+  pm->addPass(mlir::createTritonAMDGPUAccelerateMatmulPass());
   pm->addPass(mt::gpu::createTritonGPURemoveLayoutConversions());
   // TODO ROCm Check if we want to compare MI100 and greater
   pm->addPass(mlir::createTritonAMDGPUOptimizeEpiloguePass());
   pm->addPass(mt::gpu::createTritonGPUOptimizeDotOperands({true}));
-  pm->addPass(mlir::createTritonAMDGPUHoistLayoutConversionsPass());
+  pm->addNestedPass<mlir::triton::FuncOp>(
```
vwbaker, line 95: is there a reason for this? Trion inlines all the functions before this so there should be only one funcop in the module. I don't want to make unnecessary differences between xla's & triton's pipeline (in amd/backend/compiler.py).

vwbaker (approved):

xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
```diff
@@ -90,41 +86,47 @@ absl::Status CreateTritonPipeline(mlir::OpPassManager* pm,
...
+  if (cc.has_amd_matrix_core()) {
     pm->addPass(mt::gpu::createTritonGPUReorderInstructions());
   }
   if (false) {  // For upstream, this is enabled iff arch == gfx942.
     pm->addPass(mlir::createTritonAMDGPUBlockPingpongPass(num_stages));
   }
-  pm->addPass(mlir::createTritonAMDGPUCanonicalizePointersPass());
-  pm->addPass(mlir::createCanonicalizerPass());
-  pm->addPass(mlir::createTritonAMDGPUConvertToBufferOpsPass(arch_name));
+  if (false) {  // Not enabled by default.
```
bchetioui, line 125: (what's up with this `if (false)`?)

xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
zoranjovanovic-ns: With addPass there is number of following error when running unit tests:
loc("gemm_fusion___computation.clone"): error: 'builtin.module' op trying to schedule a pass on an unsupported operation

Same is in Triton:
https://github.com/openxla/triton/blob/6a9a7d8b76e51fe15bcc443b04deb5c31843c131/third_party/amd/python/triton_amd.cc#L65

xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
zoranjovanovic-ns: Yes.
At, the moment we are missing some run time flags. Those will be added sun, but we would like to avoid current test failures for now.
I was just following existing code pattern introduced when integrating latest Triton version.

PR #24689, opened 2025-04-04
Larger timeouts for first call rendezvous

giordano:
Probably not the best design, but these timeouts should really be configurable like [others are](https://github.com/openxla/xla/blob/ea7226c6def0ada11fd4635be19a288c7c66f874/xla/service/gpu/gpu_executable.cc#L429-L438) and this is a quick and dirty way to do it

### DIFF
--- a/xla/backends/gpu/runtime/collective_thunk.cc
+++ b/xla/backends/gpu/runtime/collective_thunk.cc
@@ -58,6 +58,10 @@ limitations under the License.
 #include "xla/tsl/platform/statusor.h"
 #include "xla/util.h"
 
+// TODO: These timeouts should really be configurable via debug options.
+extern "C" int XLA_FIRST_CALL_RENDEZVOUS_WARN = 20;
+extern "C" int XLA_FIRST_CALL_RENDEZVOUS_TERMINATE = 40;
+
 namespace xla::gpu {
 namespace {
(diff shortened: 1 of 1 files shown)


xla/backends/gpu/runtime/collective_thunk.cc
```diff
@@ -58,6 +58,10 @@ limitations under the License.
 #include "xla/tsl/platform/statusor.h"
 #include "xla/util.h"
 
+// TODO: These timeouts should really be configurable via debug options.
```
ezhulenev, line 61: Can you add proper options for that, extern "C" "options" are not used anywhere else in XLA and it will be a very bad example.

wsmoses: I think @giordano might need some help / pointers here (they're not familiar with the XLA codebase at all), if you can point to what would need to be updated

ezhulenev: It should be added to this section in the proto: https://github.com/openxla/xla/blob/main/xla/xla.proto#L231

And flag parsing initialized in: https://github.com/openxla/xla/blob/main/xla/debug_options_flags.cc

giordano:
Done in #25399

PR #24731, opened 2025-04-07
[NVIDIA GPU] Assign collective memory to both IO of collective instructions when user buffer is enabled

Tixxx:
Currently when user buffer is enabled, buffer assigner doesn't assign collective memory space to the inputs of collectives that are not in-place ops, ie all-gather or reduce-scatter, this prevents nccl from enabling NVLS + user buffer for these collectives.


frgossen (approved):
Thanks!

PR #24871, opened 2025-04-09
[NVIDIA GPU] Fix CollectiveColorer memory space propagation

terryysun:
`XLA_FLAGS="--xla_gpu_enable_nccl_user_buffers=1" pytest tests/memories_test.py` fails on latest main, memory space assignments need to be propagated in CollectiveColorer. 

This PR propagates the non-default memory space assignments in CollectiveColorer, fixing the failures above. Changes are guarded by StreamExecutorGpuClientTest.ExecutableCollectiveMemoryOutputMemoryKindTest, making sure it's not breaking NCCL user buffer.


frgossen (changes requested):
Can you add a test case?

terryysun:
> Can you add a test case?

The changes are guarded by `StreamExecutorGpuClientTest.ExecutableCollectiveMemoryOutputMemoryKindTest`. This is not a new feature but to fix a situation that we didn't notice before. It can be verified by `tests/memories_test.py` in JAX. We can probably have a JAX test for it once this fix is in.

frgossen (changes requested):
Unless impossible or extremely hard, I think we should add a test in XLA, too.

frgossen (approved):

PR #24889, opened 2025-04-09
[ROCM] Fix run_hlo_module build on ROCm

amd-jianli12:
`run_hlo_module` depends on `rocm_platform` that implements BlasLt. Linking would fail with following complaint without explicitly specifying dependency.

undefined reference to 'stream_executor::rocm::BlasLt::Init()'
undefined reference to 'vtable for stream_executor::rocm::BlasLt'

@xla-rotation Would you help have a look?


reedwm:
I think this has already been solved by https://github.com/openxla/xla/pull/24708, so closing this PR.

PR #24939, opened 2025-04-09
Dumping non-default debug options in a new file under xla_dump_to

shraiysh:
With this change, the non-default debug options are now dumped in the directory given by `--xla_dump_to` XLA_FLAGS. When dumping the non-default values, the default values are from
`DefaultDebugOptionsIgnoringFlags()`. The dump is in a protobuf parsable format. The file is emitted using the same format as other dumps with a `debug_options` suffix:
for example, `module_0001.test_module.debug_options`

`original debug options = DefaultDebugOptionsIgnoringFlags() + debug_options overwritten from file`

### DIFF
--- a/xla/service/dump.cc
+++ b/xla/service/dump.cc
@@ -753,12 +753,171 @@ std::vector<std::string> DumpHloModuleIfEnabled(const HloModule& module,
   return {};
 }
 
+std::string GetRepeatedValueAsString(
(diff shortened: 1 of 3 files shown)


reedwm (approved):
Thanks for this! This will be useful for reproducing JAX runs while keeping the same debug options

PR #25273, opened 2025-04-15
Move fusion-dispatch to RunBackend.

jreiffers:
The intention is that this pass runs after all passes that can create or modify fusions. However, this is currently not the case, since there are a couple of passes in RunBackend that violate this condition.

The tests didn't catch this because they only checked the passes in RunHloPasses, not the ones in RunBackend.

### DIFF
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -107,7 +107,11 @@ class GpuCompilerTest : public HloTestBase {
         ->RunPostSchedulingPipelines(module, 4 * 1024 * 1024, gpu_device_info);
   }
 
-  // Like GetOptimizedModule, but also runs the backend.
+  // Like GetOptimizedModule, but also runs the backend. This is important for
+  // tests that need to verify behavior of passes that run in RunBackend. The
+  // former function will only run the passes in RunHloPasses.
+  // This returns the module and the executable because the latter owns the
+  // former.
   absl::StatusOr<std::pair<const HloModule*, std::unique_ptr<OpaqueExecutable>>>
   GetOptimizedModuleForExecutable(absl::string_view hlo,
                                   const HloModuleConfig& config) {
@@ -1583,6 +1587,9 @@ class PassOrderTest : public GpuCompilerTest {
   }
 
  protected:
+  // Compiles a dummy module with the given configuration, running all passes,
+  // including the ones in RunBackend. This is important because otherwise, we
+  // might miss some passes when verifying pass order.
   void CompileModule(const HloModuleConfig& config) {
     constexpr absl::string_view constant_module = R"(
         ENTRY main {
@@ -1593,6 +1600,7 @@ class PassOrderTest : public GpuCompilerTest {
         GetOptimizedModuleForExecutable(constant_module, config));
   }
 
+  // Owns the optimized_module_ below.
   std::unique_ptr<OpaqueExecutable> compiled_executable_ = nullptr;
   const HloModule* optimized_module_ = nullptr;
 };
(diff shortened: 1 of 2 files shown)


xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -106,6 +106,22 @@ class GpuCompilerTest : public HloTestBase {
     return tensorflow::down_cast<GpuCompiler*>(compiler)
         ->RunPostSchedulingPipelines(module, 4 * 1024 * 1024, gpu_device_info);
   }
+
+  // Like GetOptimizedModule, but also runs the backend.
+  absl::StatusOr<std::pair<const HloModule*, std::unique_ptr<OpaqueExecutable>>>
+  GetOptimizedModuleForExecutable(absl::string_view hlo,
```
bchetioui, line 112: `GetOptimizedModuleAndExecutable`?

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1223,12 +1239,11 @@ ENTRY main {
 })";
 
   TF_ASSERT_OK_AND_ASSIGN(
-      std::unique_ptr<VerifiedHloModule> rewritable_transpose_module,
-      ParseAndReturnVerifiedModule(rewritable_transpose_string));
-
-  TF_ASSERT_OK_AND_ASSIGN(
-      std::unique_ptr<HloModule> rewritable_transpose_optimized_module,
-      GetOptimizedModule(std::move(rewritable_transpose_module)));
+      auto rewritable_transpose_module_and_executable,
```
bchetioui, line 1242: Seems like we never use `module_and_executable.second` on any of those---why do we need this change?

xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -1564,21 +1583,18 @@ class PassOrderTest : public GpuCompilerTest {
...
-        std::unique_ptr<VerifiedHloModule> module,
-        ParseAndReturnVerifiedModule(constant_module, config));
-    TF_ASSERT_OK_AND_ASSIGN(optimized_module_,
-                            GetOptimizedModule(std::move(module)));
+        std::tie(optimized_module_, compiled_executable_),
+        GetOptimizedModuleForExecutable(constant_module, config));
   }
 
-  std::unique_ptr<HloModule> optimized_module_;
+  std::unique_ptr<OpaqueExecutable> compiled_executable_ = nullptr;
```
bchetioui, line 1596: Where/when is this used? Is this just to `assert` that scheduling went well below?

bchetioui (changes requested):

xla/service/gpu/gpu_compiler_test.cc
jreiffers: Something needs to own the Executable.

xla/service/gpu/gpu_compiler_test.cc
jreiffers: Like above, something needs to own the Executable.

xla/service/gpu/gpu_compiler_test.cc
jreiffers: The idea is that this is the module for the executable (as opposed to what's usually called the OptimizedModule around here, which is not for the executable, but for some intermediate state). Happy to do the rename, but I don't think it's clearer.

bchetioui (approved):
Discussed offline, thanks for the improvement @jreiffers!

vwbaker:
Hey @jreiffers, I'm chasing down why the FusionDispatch is so low down in our stack and seeing if it could get hoisted back up - could you explain here why it's important for it to be below FusionWrapper and StreamAttributeAnnotator? (apparently "discussed offline" doesn't hold history quite as well :p) thanks!

PR #25275, opened 2025-04-15
Propagate loop metadata in double_buffer_loop_unrolling.

jreiffers:
Currently, this pass drops the `known_induction_variable` and `known_init_step` fields in `WhileLoopBackendConfig`. This change keeps them if they were present before and updates them with the updated value, just like `known_trip_count`. This is only done for double buffered loops, not for fully unrolled ones.


jreiffers:
@golechwierowicz Might make sense for you to look at this one :-)

Thanks!

golechwierowicz (approved):

PR #25298, opened 2025-04-15
[XLA:GPU] cudnn sdpa flex attention

Cjkkkk:
Support adding arbitrary pointwise operations between bmm1 and softmax to support new variants of cuDNN attention. For example: softcap.

An example of generated hlo for inserting soft_cap into cudnn sdpa:
```
HloModule jit__unnamed_wrapped_function_, is_scheduled=true, entry_computation_layout={(bf16[4,1024,4,64]{3,2,1,0}, bf16[4,1024,4,64]{3,2,1,0}, bf16[4,1024,4,64]{3,2,1,0}, bf16[4,4,1024,1024]{3,2,1,0}, bf16[4,4,1024,1024]{3,2,1,0})->bf16[4,1024,4,64]{3,2,1,0}}, allow_spmd_sharding_propagation_to_parameters={true,true,true,true,true}, allow_spmd_sharding_propagation_to_output={true}, frontend_attributes={fingerprint_before_lhs="bf3e2af9b31f0c7ef207cf2c64746153"}

%soft_cap.16 (Arg_0.17: bf16[4,4,1024,1024], Arg_1.18: bf16[4,4,1024,1024], Arg_2.19: bf16[4,4,1024,1024]) -> bf16[4,4,1024,1024] {
  %Arg_2.19 = bf16[4,4,1024,1024]{3,2,1,0} parameter(2)
  %Arg_1.18 = bf16[4,4,1024,1024]{3,2,1,0} parameter(1)
  %Arg_0.17 = bf16[4,4,1024,1024]{3,2,1,0} parameter(0)
  %divide.3.0 = bf16[4,4,1024,1024]{3,2,1,0} divide(%Arg_0.17, %Arg_1.18)
  %tanh.3.0 = bf16[4,4,1024,1024]{3,2,1,0} tanh(%divide.3.0)
  ROOT %multiply.3.0 = bf16[4,4,1024,1024]{3,2,1,0} multiply(%Arg_2.19, %tanh.3.0)
}

ENTRY %main.27 (Arg_0.1: bf16[4,1024,4,64], Arg_1.2: bf16[4,1024,4,64], Arg_2.3: bf16[4,1024,4,64], Arg_3.4: bf16[4,4,1024,1024], Arg_4.5: bf16[4,4,1024,1024]) -> bf16[4,1024,4,64] {
  %Arg_4.5 = bf16[4,4,1024,1024]{3,2,1,0} parameter(4)
  %Arg_3.4 = bf16[4,4,1024,1024]{3,2,1,0} parameter(3)
  %Arg_2.3 = bf16[4,1024,4,64]{3,2,1,0} parameter(2)
  %Arg_1.2 = bf16[4,1024,4,64]{3,2,1,0} parameter(1)
  %Arg_0.1 = bf16[4,1024,4,64]{3,2,1,0} parameter(0)
  %custom-call.23 = (bf16[4,4,1024,64]{3,1,2,0}, u8[256]{0}) custom-call(%Arg_0.1, %Arg_1.2, %Arg_2.3, %Arg_3.4, %Arg_4.5), custom_call_target="__cudnn$fmhaSoftmax", api_version=API_VERSION_STATUS_RETURNING, called_computations={%soft_cap.16}
  %get-tuple-element.24.0 = bf16[4,4,1024,64]{3,1,2,0} get-tuple-element(%custom-call.23), index=0
  ROOT %bitcast.21.0 = bf16[4,1024,4,64]{3,2,1,0} bitcast(%get-tuple-element.24.0)
}
```

Added an `ScoreModFunc` class to translate HLO computations into cudnn graphs.


Cjkkkk:
@akuegel Hi Adrian, could you help take a look at this? Thanks! I will address the CI failure shortly.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
```diff
@@ -175,15 +177,21 @@ absl::StatusOr<se::gpu::CudnnGraph> BuildGraphForCustomCallToForwardFMHA(
...
         TensorDescriptorFor(custom_call->operand(input_index++)->shape()));
   }
-  TF_RET_CHECK(input_index == custom_call->operand_count());
 
+  auto computations = custom_call->called_computations();
+  const HloComputation *score_mod_fwd_comp = nullptr;
+  if (computations.size() == 1) {
+    score_mod_fwd_comp = computations[0];
+    input_index += score_mod_fwd_comp->num_parameters() - 1;
+  }
```
akuegel, line 186: I guess computations.size() > 1 would be an error? Could you please add a TF_RET_CHECK for that?

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
```diff
@@ -345,6 +352,23 @@ absl::StatusOr<se::gpu::CudnnGraph> BuildGraphForCustomCallToBackwardFMHA(
                       GetDNNFmhaMaskKindFromCudnnFmhaMaskKind(cudnn_mask_type));
 
   const int sliding_window_length = config.sliding_window_length();
+  auto computations = custom_call->called_computations();
+  const HloComputation *score_mod_fwd_comp = nullptr;
+  const HloComputation *score_mod_bwd_comp = nullptr;
+  if (computations.size() == 1) {
```
akuegel, line 358: I guess computations.size() > 1 would be an error? Could you please add a TF_RET_CHECK for that?

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
```diff
@@ -345,6 +352,23 @@ absl::StatusOr<se::gpu::CudnnGraph> BuildGraphForCustomCallToBackwardFMHA(
...
+    score_mod_bwd_comp = computations[0];
+    auto softmax_aux = custom_call->mutable_operand(3);
+    HloInstruction *fwd_custom_call;
+    if (Match(softmax_aux,
+              m::GetTupleElement(m::CustomCall(&fwd_custom_call), 1))) {
+      score_mod_fwd_comp = fwd_custom_call->called_computations()[0];
```
akuegel, line 364: Somewhat easier to read due to less nesting:

if (!Match(...)) {
  return absl::InternalError(...)
}
score_mod_fwd_comp = fwd_custom_call->called_computations()[0];

Also, can we be sure that fwd_custom_call actually has exactly 1 computation?

xla/hlo/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc
```diff
@@ -2851,12 +2852,14 @@ LogicalResult ExportXlaOp(CustomCallOp op, OpLoweringContext ctx) {
     return op.emitOpError()
            << "cannot export with more than one called computations";
 
-  // Custom call can be exported either with called computation or with layout
-  // attributes. The XlaBuilder API does not allow both.
-  if (!op.getCalledComputations().empty() && op.getOperandLayouts() &&
-      op.getResultLayouts()) {
-    return op.emitOpError() << "cannot export if both called computation and "
-                               "layouts are specified";
+  if (op.getCallTargetName().str().find(kCudnnFMHA) == std::string::npos) {
```
akuegel, line 2855: Can you please add a test for this change?
When I read the code further down below, I do believe that it does not handle cases where a custom call has both a called computation and operand layout or result layout. You can see a check for called computations, and then an else if for operand layout and result layout.
So I believe that excluding a custom call with target kCudnnFMHA from the check here is not sufficient, you would also need to adapt the logic down below if you want to support having both called computation and result layout.
Also, is this something that needs to be done in this PR, or could it be separate?

Cjkkkk: Thanks for the feedback, I double checked and it is right that if computation is needed, `CustomCallWithComputation` will be called and if layout is needed, `CustomCallWithLayout` will be called. `CustomCallWithComputation`  always used `nullopt` when calls into `builder->CustomCall`. I think I can try modifying `CustomCallWithComputation`  to accept layout. Let me create a separate PR for this.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
Cjkkkk: I think at most 1 computation will be attached here because of this: https://github.com/openxla/xla/blob/74466a328188c70b5be64153aedb93516c5dd11b/xla/hlo/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc#L2560. Should we still check here?

Cjkkkk:
will be OOO until next Monday. Will follow up on the comments after.

xla/hlo/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc
Cjkkkk: I think we still need to add these changes to make sure it allow both layouts and computation?

akuegel: If the other change lands first, then we don't need this. I would prefer not to merge this PR if we do not actually support a custom call with computation and layout constraint. But maybe I misunderstood something? Why does this custom call have layout constraints, is it maybe not actually needed? If we merge the PR like this, we are essentially saying: ignore the layout constraints (as it would be handled by the code for custom calls with just a computation).

akuegel:
> I attempted to add a new `CustomCallWithComputationAndLayouts` in https://github.com/openxla/xla/compare/main...Cjkkkk:xla:custom_call_comp_with_layouts?expand=1 so we don't break anything else currently working in XLA. But the test failed with
> 
> ```
> xla/service/gpu/custom_call_test.cc:633: Failure
> Value of: _status_or_value45.status().ok()
>   Actual: false
> Expected: true
> INTERNAL: Called computation is not defined
> ```
> 
> Seems like computation is not propagated correctly. Anything I missed here?

I don't know whether it is the only problem, but in your change you forgot to remove the check earlier in the file that a custom call cannot have both layout constraints and a called computation. Since you are adding support for that case, this check is not needed anymore (and therefore, if your other PR lands first, this PR does not have to touch that file).

xla/hlo/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc
Cjkkkk: Yes, I think the other PR should go in first, I remember if no layout constraints are specified, then layouts may get altered by XLA and cuDNN has a specific requirement for the layout.

Cjkkkk:
yeah, that should be the removed, but the unit test fails for other reason as it directly calls into `CustomCallWithComputationAndLayouts` without going through that check. I will keep digging into this.

Cjkkkk:
Found the issue where proto to custom call creation ignores the computation when layout constraints is set. Opened a PR here: https://github.com/openxla/xla/pull/27276. Once it is merged, I will remove the checks in this PR.

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
```diff
@@ -39,6 +41,7 @@ limitations under the License.
 #include "xla/service/gpu/ir_emission_utils.h"
 #include "xla/service/gpu/stream_executor_util.h"
 #include "xla/service/gpu/transforms/block_scaling_rewriter.h"
+#include "xla/service/pattern_matcher.h"
```
akuegel, line 42: This requires a corresponding BUILD dependency

xla/hlo/translate/mhlo_to_hlo/mlir_hlo_to_hlo.cc
```diff
@@ -145,6 +145,7 @@ constexpr char kReductionDim[] = "reduction_dim";
 constexpr char kReductionInputSizeOverride[] = "reduction_input_size_override";
 constexpr char kReplicaGroups[] = "replica_groups";
 constexpr char kTopK[] = "top_k";
+constexpr char kCudnnFMHA[] = "__cudnn$fmha";
```
akuegel: I think this is a leftover from the previous change and likely is not needed anymore?

xla/service/gpu/transforms/cudnn_custom_call_compiler.cc
akuegel: Doesn't hurt to check here as well (could be that the HLO is created via some other path).

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4578,6 +4582,347 @@ void FixDimsForRaggedOffset(std::vector<int64_t>& dims, int max_reg_per_batch) {
...
+          case xla::Comparison::Direction::kGt:
+            return m::CMP_GT;
+          case xla::Comparison::Direction::kLe:
+            return m::CMP_LE;
+          case xla::Comparison::Direction::kLt:
+            return m::CMP_LT;
+        }
+        break;
+      case xla::HloOpcode::kConvert:
+        return m::IDENTITY;
```
akuegel: Just to double-check, does it mean convert is handled implicitly?

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4578,6 +4582,347 @@ void FixDimsForRaggedOffset(std::vector<int64_t>& dims, int max_reg_per_batch) {
...
+      case xla::HloOpcode::kAnd:
+        return m::LOGICAL_AND;
+      case xla::HloOpcode::kOr:
+        return m::LOGICAL_OR;
+      default:
+        return std::nullopt;
+    }
+  }
+
+  std::optional<cudnn_frontend::DataType_t> ToCudnnDataType(
```
akuegel: Looks like this function is not called?

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: It is called IDENTITY in cudnn frontend but it actually does casting if the input and output data type is different. An example here: https://github.com/NVIDIA/cudnn-frontend/blob/724f0ec8ce06027feada51f2d948cd3313e63720/samples/cpp/matmul/mixed_matmul.cpp#L67.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: Removed.

akuegel (approved):

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1387,6 +1387,110 @@ class FlashAttentionPagedAttention : public MultiHeadedAttentionTest {
...
+        se::dnn::VersionInfo(9, 7, 0)) {
+      GTEST_SKIP() << "Flash Attention requires cuDNN >= 9.7.0.";
+    }
+    // Extend cudnn sdpa soft capping using flex attention
+    std::string hlo_string =
+        GetModuleFlash_Attention_Flex_Attention_Soft_Capping_BF16();
+    // Reference implementation is pure xla sdpa soft capping.
+    std::string hlo_string_ref =
+        GetModuleFlash_Attention_Flex_Attention_Soft_Capping_Reference_BF16();
+    EXPECT_TRUE(RunAndCompareTwoModules(hlo_string, hlo_string_ref,
```
thomasjoerg, line 1488: This code makes sure the custom call computes the same result as the HLO above, right?
Should we test the code that rewrites the HLO into the custom call too?

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1387,6 +1387,110 @@ class FlashAttentionPagedAttention : public MultiHeadedAttentionTest {
   }
 };
 
+class FlashAttentionFlexAttention : public MultiHeadedAttentionTest {
+ protected:
+  const std::string
+  GetModuleFlash_Attention_Flex_Attention_Soft_Capping_BF16() {
+    const std::string hlo_text = R"(
```
thomasjoerg: Please make the test HLO more human readable:
* Remove default layouts
* Use readable names like `zero = f32[] constant(0)` (rather than `%constant.8`) or `add` rather than `%region_1.26`
*

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4578,6 +4582,322 @@ void FixDimsForRaggedOffset(std::vector<int64_t>& dims, int max_reg_per_batch) {
   dims[0] *= max_reg_per_batch;
 }
 
+class ScoreModFunc {
```
thomasjoerg: The following code looks complex enough to deserve some test coverage. Can you please add some?

thomasjoerg (changes requested):

xla/service/gpu/tests/gpu_fused_mha_test.cc
Cjkkkk: There is no rewrite happening in XLA for cudnn sdpa, the custom call is directly lowered from JAX api. Test is comparing the cudnn custom call against pure xla implementation.

thomasjoerg: Thanks for the explanation. I would argue that JAX is a better place for this test than XLA.

The idea is to ensure that cuDNN sdpa is equivalent to JAX's default HLO lowering, right? Let's add this test to JAX, so we cover the JAX HLO lowering too (which currently isn't tested). XLA codegen will be covered either way.

Cjkkkk:
Could someone share infos on the internal test failure here?

thomasjoerg:
> Could someone share infos on the internal test failure here?

This PR introduces a new `#if` macro, which is flagged by our internal CI. Context: We (XLA team) are currently trying to eliminate the use of macros where we can. @beckerhe can provide more details.

xla/stream_executor/cuda/cuda_dnn.cc
beckerhe: Can you put this class into its own file and add some tests? A class level comment would also be great.

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4578,6 +4582,322 @@ void FixDimsForRaggedOffset(std::vector<int64_t>& dims, int max_reg_per_batch) {
...
+  using Tensor_t = std::shared_ptr<cudnn_frontend::graph::Tensor_attributes>;
+  using Graph_t = std::shared_ptr<cudnn_frontend::graph::Graph>;
```
beckerhe: Nit: The Google style guide prefers not to have `_t` suffixes. Can you rename these aliases to `Tensor` and `Graph`?

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4578,6 +4582,322 @@ void FixDimsForRaggedOffset(std::vector<int64_t>& dims, int max_reg_per_batch) {
...
+    TF_RETURN_IF_ERROR(UpdateHloParameterToCudnnMap(graph, fwd_hlo_to_cudnn_,
+                                                    fwd_comp_, next_uid));
+    TF_RETURN_IF_ERROR(
+        UpdateHloConstantToCudnnMap(graph, fwd_hlo_to_cudnn_, fwd_comp_));
+    if (bwd_comp_ != nullptr) {
+      TF_RETURN_IF_ERROR(
+          UpdateHloConstantToCudnnMap(graph, bwd_hlo_to_cudnn_, bwd_comp_));
+    }
+    return absl::OkStatus();
+  }
```
beckerhe: Instead of having a separate constructor and an Initialize function, can you create a static factory function which either returns a `absl::StatusOr<ScoreModFunc>`?

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4578,6 +4582,322 @@ void FixDimsForRaggedOffset(std::vector<int64_t>& dims, int max_reg_per_batch) {
...
+  using Graph_t = std::shared_ptr<cudnn_frontend::graph::Graph>;
+
+ public:
+  ScoreModFunc(const xla::HloComputation* fwd_comp,
+               const xla::HloComputation* bwd_comp)
+      : fwd_comp_(fwd_comp), bwd_comp_(bwd_comp) {}
+
+  template <typename Functor>
+  absl::Status Initialize(cudnn_frontend::graph::Graph& graph,
+                          Functor* next_uid) {
```
beckerhe: Instead of having a template can you use `absl::FunctionRef<cudnn_frontend::graph::Tensor_attributes::uid_t()>` as the type? This makes it easier for users to understand what they can pass in.

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4590,7 +4910,8 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
     const std::optional<dnn::TensorDescriptor> page_table_v_descriptor,
     double scale, const bool use_dropout,
     const std::optional<double> dropout_rate, const dnn::FMHAMaskKind mask_type,
-    const int sliding_window_length, const int max_seg_per_batch) {
+    const int sliding_window_length, const int max_seg_per_batch,
+    const xla::HloComputation* fwd_comp) {
```
beckerhe: We're actively trying to reduce the usage of HLO in runtime components like stream_executor to make it easier to produce compatible compiled XLA programs. From what I understand this function is only getting called from the compiler during cuDNN graph construction, right? If not, can we avoid depending on HLO here?

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4772,6 +5093,14 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
   if (sliding_window_length > 0) {
     sdpa_options.set_sliding_window_length(sliding_window_length);
   }
+
+  if (fwd_comp != nullptr) {
+    auto score_mod = std::make_shared<ScoreModFunc>(fwd_comp, nullptr);
+    TF_RETURN_IF_ERROR(score_mod->Initialize(graph, &next_uid));
+    sdpa_options.set_score_mod(std::bind(&ScoreModFunc::Forward, score_mod,
+                                         std::placeholders::_1,
+                                         std::placeholders::_2));
```
beckerhe: Nit: Can you use a lambda expression instead? `std::bind` is known to have various issues.

beckerhe:
> > Could someone share infos on the internal test failure here?
> 
> This PR introduces a new `#if` macro, which is flagged by our internal CI. Context: We (XLA team) are currently trying to eliminate the use of macros where we can. @beckerhe can provide more details.

This particular use case is fine since it makes it possible to compile with older cuDNN versions.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: Added

Cjkkkk: Done.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: I removed the hlo part here by pass in scoreModFunc ptr here instead. But one problem is that I can't do a static factory function since I need to initialize it separately in graph creation to get correct uid order assigned.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: fixed with lambda expression.

Cjkkkk:
@beckerhe I addressed most of the comments except the static factory function which conflicts the removing hlo from cuda_dnn, see comment above. Could you take another look? Thanks!

xla/service/gpu/tests/gpu_fused_mha_test.cc
thomasjoerg: Can you please comment on how you resolved this comment thread? Thank you!

xla/service/gpu/transforms/cudnn_sdpa_score_mod.h
```diff
@@ -0,0 +1,78 @@
...
+using Graph = std::shared_ptr<cudnn_frontend::graph::Graph>;
+using Uid = absl::FunctionRef<int64_t()>;
+
+class ScoreModFunc {
+ public:
+  ScoreModFunc(const xla::HloComputation* fwd_comp,
+               const xla::HloComputation* bwd_comp);
+
+  std::optional<cudnn_frontend::PointwiseMode_t> GetElementwiseMode(
+      const xla::HloInstruction& instruction);
```
beckerhe: This function doesn't need to be on the class interface, right?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.h
```diff
@@ -0,0 +1,78 @@
...
+
+  cudnn_frontend::DataType_t GetComputeDataType(const xla::PrimitiveType type);
```
beckerhe: This function also doesn't need to be on the class interface, right?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.h
```diff
@@ -0,0 +1,78 @@
...
+
+  template <xla::PrimitiveType XlaT, typename T>
+  Tensor LiteralToCudnnTensor(const xla::HloInstruction* hlo,
+                              cudnn_frontend::graph::Graph& graph);
```
beckerhe: Also no need to declare this in the header.

xla/service/gpu/transforms/cudnn_sdpa_score_mod.h
```diff
@@ -0,0 +1,78 @@
...
+
+  absl::Status UpdateHloParameterToCudnnMap(
+      cudnn_frontend::graph::Graph& graph,
+      absl::flat_hash_map<const xla::HloInstruction*, Tensor>& hlo_to_cudnn,
+      const xla::HloComputation* computation, Uid next_uid);
+
+  absl::Status UpdateHloConstantToCudnnMap(
+      cudnn_frontend::graph::Graph& graph,
+      absl::flat_hash_map<const xla::HloInstruction*, Tensor>& hlo_to_cudnn,
+      const xla::HloComputation* computation);
```
beckerhe: Should these be private functions? It appears they don't get called from anywhere else.

xla/service/gpu/transforms/cudnn_sdpa_score_mod.h
```diff
@@ -0,0 +1,78 @@
...
+#include "absl/status/status.h"
+#include "third_party/cudnn_frontend/include/cudnn_frontend.h"
+#include "xla/hlo/ir/hlo_computation.h"
+
+namespace stream_executor {
+namespace gpu {
+
+using Tensor = std::shared_ptr<cudnn_frontend::graph::Tensor_attributes>;
+using Graph = std::shared_ptr<cudnn_frontend::graph::Graph>;
+using Uid = absl::FunctionRef<int64_t()>;
```
beckerhe: Nit: Can you call this type something like `UidGenerator` since it's not really a uid.

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
```diff
@@ -0,0 +1,366 @@
...
+      auto attn_score = computation->parameter_instruction(0);
+      hlo_to_cudnn[hlo] = graph->pointwise(hlo_to_cudnn[attn_score], attrs);
+      hlo_to_cudnn[hlo]->set_data_type(cudnn_frontend::DataType_t::INT32);
+    } else if (hlo->IsElementwise()) {
+      const auto compute_dtype =
+          GetComputeDataType(hlo->shape().element_type());
+      const auto mode = GetElementwiseMode(*hlo);
+      if (!mode.has_value()) {
+        LOG(FATAL) << "Unsupported elementwise operation: " << hlo->ToString()
+                   << "\n";
```
beckerhe: Is this really a broken invariance? This rather sounds like an error which we should handle with an error. How about making this function return an `absl::StatusOr` and return an error here.

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
```diff
@@ -0,0 +1,366 @@
...
+                       .set_mode(mode.value())
+                       .set_compute_data_type(compute_dtype)
+                       .set_name(std::string(hlo->name()));
+      if (hlo->operand_count() == 1) {
+        hlo_to_cudnn[hlo] = graph->pointwise(operand(0), attrs);
+      } else if (hlo->operand_count() == 2) {
+        if (!is_virtual(0) && !is_virtual(1)) {
+          LOG(FATAL)
+              << "cuDNN doesn't support both operands to pointwise op to "
+              << "be non virtual for now.";
```
beckerhe: Same here. What about returning an error?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
```diff
@@ -0,0 +1,366 @@
...
+        }
+        // make sure first operand is virtual
+        // remove this once cuDNN supports this
+        if (!is_virtual(0) && !HloOpcodeIsBinaryCommutative(hlo->opcode())) {
+          std::cerr << hlo->ToString() << "\n";
+          LOG(FATAL) << "first operand of cuDNN pointwise op is not virtual "
+                        "and op is not commutative.";
```
beckerhe: Also here. Would it make sense to return an error instead of crashing?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
```diff
@@ -0,0 +1,366 @@
...
+            o1 = graph->pointwise(o1, convert);
+            o0->set_data_type(cudnn_frontend::DataType_t::INT32);
+            o1->set_data_type(cudnn_frontend::DataType_t::INT32);
+          }
+        }
+        hlo_to_cudnn[hlo] = graph->pointwise(o0, o1, attrs);
+      } else if (hlo->operand_count() == 3) {
+        if (xla::HloPredicateIsNotOp<xla::HloOpcode::kSelect>(hlo)) {
+          LOG(FATAL) << "Unimplemented elementwise operation:"
+                     << hlo->ToString() << "\n";
```
beckerhe: Return error instead?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
```diff
@@ -0,0 +1,366 @@
...
+        }
+        // Operand order for select differs between HLO and cuDNN.
+        hlo_to_cudnn[hlo] =
+            graph->pointwise(operand(1), operand(2), operand(0), attrs);
+      } else {
+        LOG(FATAL) << "Unimplemented elementwise operation:" << hlo->ToString()
+                   << "\n";
```
beckerhe: Return error instead?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
```diff
@@ -0,0 +1,366 @@
...
+      }
+      hlo_to_cudnn[hlo]->set_data_type(compute_dtype);
+    } else {
+      LOG(FATAL) << "Unimplemented operation:" << hlo->ToString() << "\n";
```
beckerhe: Return error instead?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
```diff
@@ -0,0 +1,366 @@
...
+      // Pad dims and strides to rank 4
+      dims.push_back(1);
+      strides.push_back(1);
+    }
+    hlo_to_cudnn[parameter] = graph.tensor(
+        cudnn_frontend::graph::Tensor_attributes()
+            .set_dim(dims)
+            .set_stride(strides)
+            .set_data_type(ToCudnnFrontendDataType(desc.type()))
+            .set_name(absl::StrCat("score_mod_input_", std::to_string(i)))
```
beckerhe: Nit: No need to `std::to_string` since `absl::StrCat` can take integers as well.

xla/service/gpu/transforms/cudnn_sdpa_score_mod_test.cc
```diff
@@ -0,0 +1,135 @@
...
+  %fwd (Arg_0.15: f32[4,4,1024,1024]) -> f32[4,4,1024,1024] {
+    %Arg_0.15 = f32[4,4,1024,1024]{3,2,1,0} parameter(0)
+    %constant.1 = f32[] constant(3)
+    %broadcast.1 = f32[4,4,1024,1024]{3,2,1,0} broadcast(%constant.1), dimensions={}
+    ROOT %multiply.1 = f32[4,4,1024,1024]{3,2,1,0} multiply(%Arg_0.15, %broadcast.1)
+  }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(auto module,
+                          xla::ParseAndReturnUnverifiedModule(hlo));
+  auto comp = module->GetComputationWithName("fwd");
```
beckerhe: Can you use the real type here?

xla/service/gpu/transforms/cudnn_sdpa_score_mod_test.cc
```diff
@@ -0,0 +1,135 @@
...
+    %Arg_0.2 = f32[4,4,1024,1024]{3,2,1,0} parameter(0)
+    %constant_1_0 = f32[] constant(3)
+    %broadcast.1.0 = f32[4,4,1024,1024]{3,2,1,0} broadcast(%constant_1_0), dimensions={}
+    ROOT %multiply.1.0 = f32[4,4,1024,1024]{3,2,1,0} multiply(%Arg_0.2, %broadcast.1.0)
+  }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(auto module,
+                          xla::ParseAndReturnUnverifiedModule(hlo));
+  auto fwd_comp = module->GetComputationWithName("fwd");
+  auto bwd_comp = module->GetComputationWithName("bwd");
```
beckerhe: Can you use the real type here instead of auto?

xla/service/gpu/transforms/cudnn_sdpa_score_mod_test.cc
```diff
@@ -0,0 +1,135 @@
...
+    %Arg_0.15 = f32[4,4,1024,1024]{3,2,1,0} parameter(0)
+    %constant.1 = f32[] constant(3)
+    %broadcast.1 = f32[4,4,1024,1024]{3,2,1,0} broadcast(%constant.1), dimensions={}
+    ROOT %multiply.1 = f32[4,4,1024,1024]{3,2,1,0} multiply(%Arg_0.15, %broadcast.1)
+  }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(auto module,
+                          xla::ParseAndReturnUnverifiedModule(hlo));
+  auto comp = module->GetComputationWithName("fwd");
+  EXPECT_TRUE(comp != nullptr);
```
beckerhe: Can you make this an ASSERT_NE? Assert because there is no point in continuing if this is a nullptr and _NE to get better error message if it fails.

xla/service/gpu/transforms/cudnn_sdpa_score_mod_test.cc
```diff
@@ -0,0 +1,135 @@
...
+    %broadcast.1.0 = f32[4,4,1024,1024]{3,2,1,0} broadcast(%constant_1_0), dimensions={}
+    ROOT %multiply.1.0 = f32[4,4,1024,1024]{3,2,1,0} multiply(%Arg_0.2, %broadcast.1.0)
+  }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(auto module,
+                          xla::ParseAndReturnUnverifiedModule(hlo));
+  auto fwd_comp = module->GetComputationWithName("fwd");
+  auto bwd_comp = module->GetComputationWithName("bwd");
+  EXPECT_TRUE(fwd_comp != nullptr);
+  EXPECT_TRUE(bwd_comp != nullptr);
```
beckerhe: ASSERT_NE?

xla/service/gpu/transforms/cudnn_sdpa_score_mod_test.cc
```diff
@@ -0,0 +1,135 @@
...
+  Tensor score_grad =
+      graph->tensor(cudnn_frontend::graph::Tensor_attributes()
+                        .set_dim({4, 4, 1024, 1024})
+                        .set_stride({4194304, 1048576, 1024, 1})
+                        .set_data_type(cudnn_frontend::DataType_t::FLOAT)
+                        .set_uid(next_uid())
+                        .set_is_virtual(true));
+  auto fwd_output = score_mod->Forward(graph, attn_score);
+  auto bwd_output = score_mod->Backward(graph, score_grad);
+  auto json_string = graph->print();
```
beckerhe: Nit: Can you use the real type here instead of auto?

xla/service/gpu/transforms/cudnn_sdpa_score_mod_test.cc
```diff
@@ -0,0 +1,135 @@
...
+  EXPECT_THAT(score_mod->UpdateCudnnMap(*graph, next_uid), IsOk());
+  Tensor attn_score =
+      graph->tensor(cudnn_frontend::graph::Tensor_attributes()
+                        .set_dim({4, 4, 1024, 1024})
+                        .set_stride({4194304, 1048576, 1024, 1})
+                        .set_data_type(cudnn_frontend::DataType_t::FLOAT)
+                        .set_uid(next_uid())
+                        .set_is_virtual(true));
+  auto output = score_mod->Forward(graph, attn_score);
+  auto json_string = graph->print();
```
beckerhe: Nit: Can you use the real type here instead of auto?

xla/stream_executor/cuda/cuda_dnn.h
```diff
@@ -713,7 +714,7 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionOperationGraph(
     std::optional<dnn::TensorDescriptor> page_table_v_descriptor, double scale,
     bool use_dropout, std::optional<double> dropout_rate,
     dnn::FMHAMaskKind mask_type, int sliding_window_length,
-    int max_seg_per_batch);
+    int max_seg_per_batch, std::shared_ptr<ScoreModFunc> score_mod);
```
beckerhe: Does this need to be a shared_ptr? To me it looks like a non-owning raw pointer would be enough here.

xla/stream_executor/cuda/cuda_dnn.h
```diff
@@ -735,8 +736,10 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionBackwardOperationGraph(
     const std::optional<dnn::TensorDescriptor> bias_descriptor,
     const std::optional<dnn::TensorDescriptor> dbias_descriptor,
     std::optional<double> dropout_rate, std::optional<int64_t> seed,
-    double scale, bool use_dropout, bool use_bias, dnn::FMHAMaskKind mask_type,
-    bool force_deterministic, int sliding_window_length, int max_seg_per_batch);
+    double scale, bool use_dropout, bool use_bias,
+    const dnn::FMHAMaskKind mask_type, bool force_deterministic,
+    const int sliding_window_length, const int max_seg_per_batch,
+    std::shared_ptr<ScoreModFunc> score_mod);
```
beckerhe: Does this need to be a shared_ptr? To me it looks like a non-owning raw pointer would be enough here.

beckerhe (commented):
Thanks so much for the changes, Shanbin. This is a lot easier to review now.

Cjkkkk:
@beckerhe Thanks for the valuable feedback! I addressed all the comments, could you take another look? Thanks!

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
Cjkkkk: This `Compile` function will be a callback function called by cudnn frontend which doesn't handle the error so I think it is best to abort here?

Cjkkkk:
Hi, how can I fixed the issue: https://github.com/openxla/xla/actions/runs/15644443414/job/44079099779?pr=25298. Seems like it is building cudnn_sdpa_score_mod.cc on CPU even it is tagged as `cuda-only`.

beckerhe:
> Hi, how can I fixed the issue: https://github.com/openxla/xla/actions/runs/15644443414/job/44079099779?pr=25298. Seems like it is building cudnn_sdpa_score_mod.cc on CPU even it is tagged as `cuda-only`.

You will also need to add the `gpu` tag.

xla/service/gpu/tests/gpu_fused_mha_test.cc
beckerhe: Can you make this variable a `static constexpr absl::string_view kHloText = ...` and change the return type to `absl::string_view`?

xla/service/gpu/tests/gpu_fused_mha_test.cc
```diff
@@ -1387,6 +1387,110 @@ class FlashAttentionPagedAttention : public MultiHeadedAttentionTest {
...
+      %get-tuple-element.22 = bf16[4,4,1024,64]{3,1,2,0} get-tuple-element(%custom-call.21), index=0
+      ROOT %transpose.24 = bf16[4,1024,4,64]{3,2,1,0} transpose(%get-tuple-element.22), dimensions={0,2,1,3}
+    }
+  )";
+    return hlo_text;
+  }
+
+  const std::string
+  GetModuleFlash_Attention_Flex_Attention_Soft_Capping_Reference_BF16() {
+    const std::string hlo_text = R"(
```
beckerhe: Can you make this variable a `static constexpr absl::string_view kHloText = ...` and change the return type to `absl::string_view`?

xla/service/gpu/transforms/BUILD
```diff
@@ -1007,6 +1009,43 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "cudnn_sdpa_score_mod",
+    srcs = ["cudnn_sdpa_score_mod.cc"],
+    hdrs = ["cudnn_sdpa_score_mod.h"],
+    visibility = ["//visibility:public"],
+    tags = ["cuda-only"],
```
beckerhe: As mentioned in the other you will need to tag this as `gpu` as well.

xla/service/gpu/transforms/BUILD
```diff
@@ -1007,6 +1009,43 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "cudnn_sdpa_score_mod",
+    srcs = ["cudnn_sdpa_score_mod.cc"],
+    hdrs = ["cudnn_sdpa_score_mod.h"],
+    visibility = ["//visibility:public"],
```
beckerhe: Is this visibiliy tag needed? Giving this public visibility will make problems for us internally, so we will need to restrict this more. Do you know which target needs to depend on this target and doesn't have visibility through the default policy?

xla/service/gpu/transforms/cudnn_sdpa_score_mod.cc
beckerhe: Yeah. True. Makes sense.

beckerhe (commented):
Only minor comments. In general this looks good to me.

You should be able to make the CPU build not fail by adding the gpu tag.
I don't see from the top of my head why the GPU build is failing. Maybe a typo?

xla/service/gpu/transforms/BUILD
Cjkkkk: cuda_dnn.cc requires this and I think it is not visible by default

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -75,6 +76,7 @@ limitations under the License.
 #include "xla/tsl/protobuf/dnn.pb.h"
 #include "xla/tsl/util/env_var.h"
 #include "tsl/platform/tensor_float_32_utils.h"
+#include "xla/service/gpu/stream_executor_util.h"
```
beckerhe: Is this needed?

xla/stream_executor/cuda/BUILD
```diff
@@ -456,6 +457,7 @@ cc_library(
         "//xla/tsl/platform:statusor",
         "//xla/tsl/protobuf:dnn_proto_cc",
         "//xla/tsl/util:env_var",
+        "//xla/service/gpu:stream_executor_util",
```
beckerhe: Is this needed?

xla/service/gpu/transforms/BUILD
beckerhe: Ah I see. In that case it would be better if you placed this target (and the files) in `stream_executor/cuda/` so that we don't create an interdependency between the compiler parts (service/gpu) and the stream_executor parts.

beckerhe (approved):

Cjkkkk:
There was a merge conflict and I resolved it, could you take another look? @beckerhe Thanks!

akuegel (approved):
Reapproving after merge

Cjkkkk:
I think there is another internal test failure :( could someone share some details

beckerhe:
> I think there is another internal test failure :( could someone share some details

I think it just needed an extra approval. I retriggered the merge.

PR #25327, opened 2025-04-16
[XLA][CPU] Ensure correct scale is used when there are multiple linear fusion ops

gaurides:
When a pattern like MM + Mul + Tanh + Mul, with two Mul ops and each has it's unique multiplier, we need to ensure the multiplier scales are applied crrectly.
Without this change, the 2nd scale factor was overwriting the first scale factor. Changes in this PR fix the issue.

### DIFF
--- a/xla/service/cpu/onednn_config.proto
+++ b/xla/service/cpu/onednn_config.proto
@@ -55,7 +55,7 @@ message OneDnnFusionConfig {
   repeated FusionKind ops = 1;
   // To avoid protobuf failures for specific decimal values,
   // the original float value alpha is type-casted to int32.
-  int32 alpha_typecast = 2;
+  repeated int32 alpha_typecast = 2;
 }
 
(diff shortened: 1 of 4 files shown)


xla/service/cpu/onednn_config.proto
```diff
@@ -55,7 +55,7 @@ message OneDnnFusionConfig {
   repeated FusionKind ops = 1;
   // To avoid protobuf failures for specific decimal values,
   // the original float value alpha is type-casted to int32.
-  int32 alpha_typecast = 2;
+  repeated int32 alpha_typecast = 2;
```
penpornk, line 58: This will break protobuf backward compatibility, but it should be okay since we only use the proto in XLA's internal C++ implementation.

penpornk (approved):
Sorry for the delay and thank you for the PR!

PR #25610, opened 2025-04-23
[ROCM] Split dot_operation_test to 2 with triton and without

alekstheod:
Rocm nightly build is failing due to the dot_operation_test and triton.
Disabling triton makes this test to pass. It make sense to split the test to test
both scenarios with triton and without triton support.

### DIFF
--- a/xla/tests/BUILD
+++ b/xla/tests/BUILD
@@ -1060,6 +1060,39 @@ xla_test(
     ],
 )
 
+cc_library(
+    name = "dot_operation_test_lib",
+    testonly = True,
+    deps = [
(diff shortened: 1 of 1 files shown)


bchetioui:
This is a test shared across all backends, I don't think it makes sense to split it only to allow ROCm to pass. Can you say more about what is failing here?

alekstheod:
> This is a test shared across all backends, I don't think it makes sense to split it only to allow ROCm to pass. Can you say more about what is failing here?

Triton fix for rocm is supposed to be done in this PR:
https://github.com/openxla/xla/pull/23688

PR #25614, opened 2025-04-23
[ROCm] relax abs and rel error tolerance for fp16 and add another test for fp32

cj401-amd:
@xla-rotation , would you please take a look?

### DIFF
--- a/xla/service/gpu/transforms/dot_dimension_sorter_test.cc
+++ b/xla/service/gpu/transforms/dot_dimension_sorter_test.cc
@@ -78,7 +78,7 @@ ENTRY e {
 )");
 }
 
-TEST_F(WithoutDotDimensionSorterTest, DimOrderCanBeChanged) {
+TEST_F(WithoutDotDimensionSorterTest, DimOrderCanBeChangedF16) {
   const char* hlo_text_ref = R"(
 HloModule m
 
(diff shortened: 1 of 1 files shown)


xla/service/gpu/transforms/dot_dimension_sorter_test.cc
```diff
@@ -102,7 +102,35 @@ ENTRY e {
 )";
 
   EXPECT_TRUE(RunAndCompareTwoModules(hlo_text_ref, hlo_text_modified,
-                                      ErrorSpec{1e-5, 1e-3},
+                                      ErrorSpec{1e-3, 2e-3},
```
golechwierowicz, line 105: Is my understanding correct that it fails on ROCm and that's why you want to relax numerics? If that's the case it might be better to just branch the expectation on `is_built_with_rocm_`.

dimitar-asenov: @cj401-amd Could you please implement Greg's comment above? Then we can merge this PR.

cj401-amd: @dimitar-asenov , thanks for your message. the issue actually has gone and no need to PR for AMD side now.

dimitar-asenov:
> cj401-amd: the issue actually has gone and no need to PR for AMD side now.

Closing this PR, since it's no longer needed.

PR #25628, opened 2025-04-23
Fix dumping non-default debug options

shraiysh:
When the default value of a flag set in `xla/debug_options_flags.cc` is `true`, and the debug options set it to `false`, this option is not dumped. This patch fixes this issue.

### DIFF
--- a/xla/service/dump.cc
+++ b/xla/service/dump.cc
@@ -886,20 +886,10 @@ std::string GetNonDefaultDebugOptions(const DebugOptions& debug_options) {
       continue;
     }
 
-    // Check if this field differs from default
-    if (reflection->HasField(debug_options, field) &&
-        !reflection->HasField(default_options, field)) {
-      // Field exists in debug_options but not defaults
+    if (GetValueAsString(reflection, debug_options, field) !=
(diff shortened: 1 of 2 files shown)


xla/service/dump.cc
```diff
@@ -892,7 +892,7 @@ std::string GetNonDefaultDebugOptions(const DebugOptions& debug_options) {
       // Field exists in debug_options but not defaults
       absl::StrAppend(&non_default_options, field->name(), ": ",
                       GetValueAsString(reflection, debug_options, field), "\n");
-    } else if (reflection->HasField(debug_options, field)) {
+    } else {
```
reedwm, line 895: Do you need to check `reflection->HasField` above, when you check:

```
    if (reflection->HasField(debug_options, field) &&
        !reflection->HasField(default_options, field)) {
```

It presumably is possible to call GetValueAsString on a field that isn't explicitly set, so perhaps you don't have to call `reflection->HasField` at all.

shraiysh: That is a good point. Removed the check for `HasField`.

reedwm (approved):

PR #25635, opened 2025-04-23
Add parsing debug options support

shraiysh:
Depends on #25628 (merged)
Depends on #25691 (merged)


HamiAlyasin54 (approved):

xla/debug_options_flags.cc
```diff
@@ -2404,6 +2405,50 @@ void ParseDebugOptionFlagsFromEnv() {
...
+  absl::call_once(flags_init, &AllocateFlags, nullptr);
+  VLOG(2) << "Parsing flags from file: " << filename;
+  // Read the file content
+  std::string file_content;
+  absl::Status status = tsl::ReadFileToString(
+      tsl::Env::Default(), std::string(filename), &file_content);
+  if (!status.ok()) {
+    LOG(ERROR) << "Failed to read file: " << filename
+               << ", error: " << status.ToString();
+    return false;
```
reedwm, line 2418: Add a test for the case where this function returns false.

xla/debug_options_flags.cc
```diff
@@ -2404,6 +2405,50 @@ void ParseDebugOptionFlagsFromEnv() {
...
+  }
+
+  // Read from new_debug_options, and overwrite the flags in debug_options that
+  // are actually mentioned in file_contents.
+  std::vector<const google::protobuf::FieldDescriptor*> overwritten_fields;
+  int field_count = new_debug_options.GetDescriptor()->field_count();
+  for (int i = 0; i < field_count; i++) {
+    const google::protobuf::FieldDescriptor* field =
+        new_debug_options.GetDescriptor()->field(i);
+    if (absl::StrContains(file_content, field->name())) {
```
reedwm, line 2436: Is it possible to have a more robust check for the field? The file might have the name of a field in the value of another field, e.g. `--xla_dump_to=/tmp/dump_with_xla_enable_fast_math`

Asking Gemini, the [`Parser::WriteLocationsTo`](https://protobuf.dev/reference/cpp/api-docs/google.protobuf.text_format/#TextFormat.Parser.WriteLocationsTo.details) function seems promising. The returned ParseInfoTree I think has the info to determine if the field was explicitly specified.

Alternatively, each field is on a different line, right? Maybe you can just search for `"\n" + field->name()`. 

If this is a lot of work to implement, you don't have to do it yet, but add a TODO.

xla/debug_options_flags.h
```diff
@@ -41,6 +41,12 @@ void AppendDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
 // Parses the debug option flags from XLA_FLAGS environment variable.
 void ParseDebugOptionFlagsFromEnv();
 
+// Parse the debug options from debug_options file.
```
reedwm, line 44: Would be good to describe that this overrides any fields in the global variable returned from `GetDebugOptionsFromFlags` with fields explicitly specified in filename. E.g. something like:

"Given a string containing the textual form of a DebugOptions protobuf, parses it. The global variable containing the debug options returned from 'GetDebugOptionsFromFlags' is mutated by setting fields explicitly specified in the file."

Somewhat unrelatedly, probably worth moving GetNonDefaultDebugOptions to this file. It doesn't actually dump antyhing to disk, and is closely tied with this function. But that can be done in a separate PR

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
```diff
@@ -382,10 +407,24 @@ int main(int argc, char** argv) {
       tsl::Flag("profile_execution", &opts.profile_execution,
                 "If set, we will profile the execution and print the results."),
       tsl::Flag("xla_gpu_dump_xspace_to", &opts.xla_gpu_dump_xspace_to,
-                "A directory to dump xspace data for GPU profiling.")};
+                "A directory to dump xspace data for GPU profiling."),
+      // This option is not used during parsing, but it is added here for
+      // documentation, and for ensuring that the parser knows this should be
+      // ignored if present.
+      tsl::Flag("debug_options_file", &opts.debug_options_filename,
```
reedwm, line 414: Instead of adding a field `opts.debug_options_filename`, you can just create a local `bool unused_debug_options_filename;` variable and pass that here.

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
```diff
@@ -382,10 +407,24 @@ int main(int argc, char** argv) {
...
+                "A file containing debug options to be passed to the HLO "
+                "module. The file should contain a serialized DebugOptions "
+                "proto message. The order of precedence: command line flags > "
+                "XLA_FLAGS > debug_options_file > default flags.")};
 
   xla::AppendDebugOptionsFlags(&flag_list);
 
+  auto debugOptionsFilename = GetDebugOptionsFileName(argc, argv);
```
reedwm, line 422: It's hacky to parse argc and argv like this outside absl. An alternative might be to call `tsl::Flags::Parse` twice: once to look for the `--debug_options_file` flag, and again to parse everything else afterwards. But this admittedly is hacky as well, and I cannot think of a better way. 

You can keep this for now but I'll keep thinking if there's a better way.

reedwm (changes requested):
Thanks for the change! Will be nice to finally be able to specify the debug options from an HLO dump.

xla/debug_options_parsers_test.cc
```diff
@@ -126,6 +127,140 @@ TEST(FuelTest,
...
+  EXPECT_TRUE(debug_options.xla_dump_hlo_as_long_text());
+  EXPECT_FALSE(debug_options.xla_hlo_pass_fix_detect_cycles());
+
+  // default value (should not be in debug_options file).
+  debug_options.set_xla_dump_hlo_as_long_text(true);
+  // non-default value (should be in debug_options file).
+  debug_options.set_xla_hlo_pass_fix_detect_cycles(true);
+  std::string contents;
+  auto temp_file = WriteDebugOptionsToTempFile(debug_options, &contents);
+  ResetFlagValues();
```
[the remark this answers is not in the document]
shraiysh: That wouldn't be enough. The difference is that `ParseDebugOptionFlagsFromEnv()` only overwrites the flags that are explicitly mentioned and there could be leaking from the previous global state of flags. That is, after you parse some flags for one case, `ParseDebugOptionFlagsFromEnv()` wouldn't erase the values set by that parsing, but for testing we need to clear that state.

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
shraiysh: That wouldn't be printing the message with `--help`. Is there a way to add this message while also avoiding to add this as an option?

xla/debug_options_flags.cc
shraiysh: Done.

xla/debug_options_parsers_test.cc
```diff
@@ -126,6 +127,140 @@ TEST(FuelTest,
...
   EXPECT_TRUE(PassFuelIsSet("XYZ"));
 }
+
+std::string WriteDebugOptionsToTempFile(const DebugOptions& debug_options,
+                                        std::string* contents) {
+  std::string string_debug_options = GetNonDefaultDebugOptions(debug_options);
+  contents->append(string_debug_options);
+  std::string fname = tsl::testing::TmpDir() + "/debug_options.txt";
+  EXPECT_TRUE(
+      tsl::WriteStringToFile(tsl::Env::Default(), fname, *contents).ok());
```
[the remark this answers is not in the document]
shraiysh: Done.

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
shraiysh: Marking this as "resolved" for now.

shraiysh:
Thanks for the review @reedwm! I have tried to address all the comments. Can you take a look again?

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
reedwm: > That wouldn't be printing the message with `--help`. Is there a way to add this message while also avoiding to add this as an option?

I didn't mean to remove the `tsl::Flag`, just to remove the `debug_options_filename` field in `HloRunnerConfig`. E.g. instead of doing:

```c++
    ...
    tsl::Flag("debug_options_file", &opts.debug_options_filename, ...)};
```

Do:

```c++
bool unused_debug_options_filename;
    ...
    tsl::Flag("debug_options_file", &unused_debug_options_filename, ...)};
```

I think this will still print the message with `--help`. Anyway, not a big deal if you don't do this. Having a `debug_options_filename` field in `HloRunnerConfig` does make things more consistent, as then every flag has a corresponding field in `HloRunnerConfig`. So feel free to ignore this.

reedwm (changes requested):

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
shraiysh: Done. Sorry about the confusion - I don't know how I came to the conclusion that help won't be visible if we did this :D . Thanks for the detailed explanation.

xla/debug_options_parsers_test.cc
```diff
@@ -126,6 +127,194 @@ TEST(FuelTest,
   EXPECT_FALSE(PassFuelIsSet("PQR"));
   EXPECT_TRUE(PassFuelIsSet("XYZ"));
 }
+
+std::string WriteDebugOptionsToTempFile(const DebugOptions& debug_options,
+                                        std::string* contents) {
+  std::string string_debug_options = GetNonDefaultDebugOptions(debug_options);
+  contents->clear();
+  contents->append(string_debug_options);
```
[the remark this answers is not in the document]
shraiysh: Done.

reedwm (approved):

reedwm (approved):

reedwm (approved):

PR #25767, opened 2025-04-28
Improve warning message for high execution overhead with sharded arrays

zhenying-liu:
When performing computations on a JAX array that has sharding and  pinned_host memory kind, users may experience high execution overhead. Add guidance in the warning message to suggest using device_put() to transfer the array to the device for better performance.

### DIFF
--- a/xla/hlo/transforms/host_offloader.cc
+++ b/xla/hlo/transforms/host_offloader.cc
@@ -264,8 +264,8 @@ absl::StatusOr<bool> HostOffloader::WalkDownHostMemoryOffloadPaths(
           "Found an instruction (\"%s\") which does device compute in host "
           "memory space. Converting into host compute. This is likely to have "
           "a very slow execution time. If you're using JAX, use device_put() "
-          "on the inputs to move the computation to the "
-          "accelerator.",
+          "to move the inputs to the device so that computation happens on the "
+          "device.",
           instruction->name());
       host_offload_utils::SetHostComputeFrontendAttribute(*instruction);
     }
--- a/xla/hlo/transforms/host_offloader.cc
+++ b/xla/hlo/transforms/host_offloader.cc
@@ -263,7 +263,9 @@ absl::StatusOr<bool> HostOffloader::WalkDownHostMemoryOffloadPaths(
       LOG(WARNING) << absl::StreamFormat(
           "Found an instruction (\"%s\") which does device compute in host "
           "memory space. Converting into host compute. This is likely to have "
-          "a very high overhead.",
+          "a very slow execution time. If you're using JAX, use device_put() "
+          "on the inputs to move the computation to the "
+          "accelerator.",
           instruction->name());
       host_offload_utils::SetHostComputeFrontendAttribute(*instruction);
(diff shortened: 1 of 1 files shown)


zhenying-liu:
My JAX code example got such a warning so that we are motivated to add a bit more information for JAX user code. 
Below are HLOs before optimizations from that JAX code. Since host_offloader_test.cc is an xla_cc_test which can't be run on GPUs, I didn't include this as a negative unit test.
```
HloModule jit__lambda_, entry_computation_layout={(s32[8,2]{1,0:S(5)})->s32[8,2]{1,0}}

ENTRY %main.8 (Arg_0.1: s32[8,2]) -> s32[8,2] {
  %Arg_0.1 = s32[8,2]{1,0} parameter(0), sharding={maximal device=0}, metadata={op_name="x"}
  %constant.2 = s32[] constant(2)
  %broadcast.3 = s32[8,2]{1,0} broadcast(%constant.2), dimensions={}
  %add.4 = s32[8,2]{1,0} add(%Arg_0.1, %broadcast.3), metadata={op_name="jit(<lambda>)/jit(main)/add" source_file="testH2D.py" source_line=17}
  %custom-call.5 = s32[8,2]{1,0} custom-call(%add.4), custom_call_target="annotate_device_placement", custom_call_has_side_effect=true, frontend_attributes={_xla_buffer_placement="device"}
  %tuple.6 = (s32[8,2]{1,0}) tuple(%custom-call.5)
  ROOT %get-tuple-element.7 = s32[8,2]{1,0} get-tuple-element(%tuple.6), index=0, sharding={maximal device=0}
}
```

xla/hlo/transforms/host_offloader.cc
```diff
@@ -263,7 +263,9 @@ absl::StatusOr<bool> HostOffloader::WalkDownHostMemoryOffloadPaths(
       LOG(WARNING) << absl::StreamFormat(
           "Found an instruction (\"%s\") which does device compute in host "
           "memory space. Converting into host compute. This is likely to have "
-          "a very high overhead.",
+          "a very slow execution time. If you're using JAX, use device_put() "
+          "on the inputs to move the computation to the "
```
SandSnip3r, line 267: Can we make the wording slightly more accurate? `device_put` does not "move computation" anywhere. A more accurate alternative:
"use device_put() to move the inputs to the device so that computation happens on the device."

SandSnip3r (changes requested):

zhenying-liu:
The failed tests are not related the current commit as this commit only changes the warning message and clang format:
- xla/tests:matmul_test_gpu_amd_any
- xla/service:compiler_test_gpu_amd_any

SandSnip3r (approved):

SandSnip3r:
Why did so much formatting change? Are you using a different clangformat than us?

zhenying-liu:
> Why did so much formatting change? Are you using a different clangformat than us?

Because it failed Clang Format check otherwise. I'm using the same clang-format as you. Once the format was fixed, it could pass. Thanks for the code review!

PR #25769, opened 2025-04-28
[XLA:GPU] remove cuda graph emtpy node inside conditional node workaround

shawnwang18:
CUDA release r570/CUDA 12.8 has fixed the bug that cuda will hang when cuda-graph's conditional node has an empty body graph.  Remove the workaround in XLA, tests have been covered by GpuCommandBufferTest.ConditionalCaseEmptyGraph

### DIFF
--- a/xla/stream_executor/cuda/cuda_command_buffer.cc
+++ b/xla/stream_executor/cuda/cuda_command_buffer.cc
@@ -597,16 +597,19 @@ absl::StatusOr<size_t> CudaCommandBuffer::GetNodeCount() const {
 }
 
 absl::Status CudaCommandBuffer::PrepareFinalization() {
-  // TODO(b/362769658): Remove this workaround when cuda supports conditionals
-  // with empty graphs.
-  TF_ASSIGN_OR_RETURN(auto node_count, GetNodeCount());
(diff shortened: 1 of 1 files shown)


ezhulenev (approved):

xla/stream_executor/cuda/cuda_command_buffer.cc
```diff
@@ -597,16 +597,17 @@ absl::StatusOr<size_t> CudaCommandBuffer::GetNodeCount() const {
 }
 
 absl::Status CudaCommandBuffer::PrepareFinalization() {
-  // TODO(b/362769658): Remove this workaround when cuda supports conditionals
-  // with empty graphs.
+#if CUDA_VERSION < 12080
```
beckerhe, line 600: Shouldn't that be a runtime driver version check?

Something like this: 

```
if (parent_->GetDeviceDescription().driver_version() <
      SemanticVersion{12, 8, 0}) {
// ...
}
```

shawnwang18: Yes, runtime version check is better, I made a update for that.

beckerhe (approved):
Thanks!

PR #25796, opened 2025-04-28
[NVIDIA GPU] Add nvshmem to collective backend config and detect config setting in buffer colorer

Tixxx:
Split from this [pr](https://github.com/openxla/xla/pull/21973)
2: Add nvshmem to collective backend config and detect config setting in buffer colorer


PatriosTheGreat (approved):
Thanks!

PR #25867, opened 2025-04-29
Fix command buffer scheduling for async dynamic memcpy.

jreiffers:
Sync dynamic memcpy fusions correctly prevent the creation of commands, but async ones currently do not. For some reason, there's quite a bit of code duplication around this, so I took the opportunity to clean up some of that. There's still some duplication wrt. sync/async, but the behavior is actually slightly different for some fusions. I'll send a followup once I know whether that's intentional or not, for now just fix the bug.


akuegel (approved):
Thanks also for the refactoring, easier to read now.

PR #25914, opened 2025-04-30
[NVIDIA GPU] Add nvshmem communicator and runtime thunks

Tixxx:
Split from this [pr](https://github.com/openxla/xla/pull/21973)
3: Add nvshmem runtime thunks
4. communicator changes have been moved to this(https://github.com/openxla/xla/pull/26255)
5. Base nvshmem thunk has been moved to (https://github.com/openxla/xla/pull/26568).


Tixxx:
Hi @PatriosTheGreat , I added the ir lowering to this pr as well so it's easier to test e2e. Please let me know if this works.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2609,6 +2740,18 @@ std::optional<const HloInstruction*> GetCollectiveHeroForDynamicSliceFusion(
       [](const HloInstruction* instr) { return IsCollective(instr); });
 }
 
+inline bool IsNvshmemCollective(const HloInstruction* instr) {
```
PatriosTheGreat, line 2743: Is this inline needed?

PatriosTheGreat:
I see a deadlock when I tried to run this PR with 500 all-reduces. Strangely with 100 instructions everything works fine. I'm using the following HLO:

```
HloModule pmap__lambda_

region_add {
  lhs = f16[] parameter(0)
  rhs = f16[] parameter(1)
  ROOT ret = f16[] add(lhs, rhs)
}

ENTRY main.7 {
  
    param0 = f16[8]{0} parameter(0)
    out0 = f16[8]{0} all-reduce(param0), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
   ...
    param499 = f16[8]{0} parameter(499)
    out499 = f16[8]{0} all-reduce(param499), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
  ROOT result = tuple(
    out0,...,out499
  )
}
```

xla/backends/gpu/runtime/nvshmem_collective_thunk.cc
```diff
@@ -0,0 +1,424 @@
...
+//         or groups empty and num_replicas == 1 && num_partitions == 1.
+//   3. channel_id is set, use_global_device_ids = true (flattened-ids):
+//         degenerate if replica_groups are singleton (groups cannot be empty).
+//   4. no channel_id, no use_global_device_ids:
+//         identical to 1.
+//   5. channel_id is set, no use_global_device_ids:
+//         degenerate if replica_groups are singleton or group emty and
+//         num_partitions == 1 (since replica groups contain partition ids).
+//
+bool NvshmemCollectiveConfig::IsDegenerate(int64_t replica_count,
```
PatriosTheGreat, line 127: CollectiveThunk::IsDegenerate can be used instead in the invocation side. This method can be removed.

xla/backends/gpu/runtime/nvshmem_collective_thunk.cc
```diff
@@ -0,0 +1,424 @@
...
+      CHECK(!groups_empty)
+          << "replica groups cannot be empty if use_global_device_ids = true";
+      return all_groups_singleton;
+    default:
+      CHECK(0) << "Invalid collective op mode";
+      return false;
+  }
+}
+
+void NvshmemCollectiveConfig::SetCollectiveOpKindAndID(
```
PatriosTheGreat, line 157: These 3 methods can also be removed is CollectiveConfig is used instead

xla/backends/gpu/runtime/nvshmem_all_reduce_thunk.h
```diff
@@ -0,0 +1,93 @@
...
+#include "xla/backends/gpu/runtime/nvshmem_collective_thunk.h"
+#include "xla/core/collectives/communicator.h"
+#include "xla/hlo/ir/hlo_instructions.h"
+#include "xla/service/collective_ops_utils.h"
+#include "xla/stream_executor/stream.h"
+
+namespace xla {
+namespace gpu {
+
+struct NvshmemAllReduceConfig {
```
PatriosTheGreat, line 31: all_reduce_thunk::AllReduceConfig can be used instead

xla/backends/gpu/runtime/BUILD
```diff
@@ -1673,3 +1673,107 @@ xla_test(
...
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/status:statusor",
+        "@com_google_absl//absl/strings:str_format",
+        "@com_google_absl//absl/synchronization",
+        "@com_google_absl//absl/time",
+        "@llvm-project//mlir:IR",
+        "@tsl//tsl/platform:errors",
+        "@tsl//tsl/platform:statusor",
+    ] + if_cuda_is_configured([
+      "@nvshmem//:nvshmem_lib",
```
PatriosTheGreat, line 1777: Is this needed?

xla/backends/gpu/runtime/BUILD
```diff
@@ -1673,3 +1673,107 @@ xla_test(
...
```
PatriosTheGreat, line 1726: Is this needed?

xla/backends/gpu/runtime/nvshmem_all_reduce_thunk.cc
```diff
@@ -0,0 +1,142 @@
...
+  if (!MatchReductionComputation(inst->called_computations().front())
+           .has_value()) {
+    return absl::UnimplementedError("Unrecognized reduction computation");
+  }
+
+  return absl::OkStatus();
+}
+
+template <typename HloInstType>
+NvshmemAllReduceConfig GetNvshmemAllReduceConfigInst(HloInstType* inst) {
```
PatriosTheGreat, line 82: AllReduceThunk::GetAllReduceConfigInst and AllReduceThunk::GetGroupModeInst are identical, so this should be moved in a common area instead

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
```diff
@@ -0,0 +1,275 @@
...
+
+//===----------------------------------------------------------------------===//
+// NvshmemCollectiveThunk
+//===----------------------------------------------------------------------===//
+
+// Forward declare.
+class NvshmemCollectiveDoneThunk;
+
+// Thunk base class for NVSHMEM collective operations.
+class NvshmemCollectiveThunk : public Thunk {
```
PatriosTheGreat, line 115: This class ideally should be inherited from CollectiveThunk since there are many common snippets. I assume CollectiveThunk right now has a lot NCCL specific logic. This logic either need to be generalized or moved to another CollectiveThunk child which will be used as parent for NCCLThunks

PatriosTheGreat (changes requested):

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2234,6 +2236,135 @@ absl::Status IrEmitterUnnested::EmitCollectiveAsyncDone(
   return absl::OkStatus();
 }
 
+template <typename NvshmemThunkType, typename HloInstType>
```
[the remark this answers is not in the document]
Tixxx: yea that makes sense, updated

xla/service/gpu/ir_emitter_unnested.cc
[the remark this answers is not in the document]
Tixxx: no, removed

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
```diff
@@ -0,0 +1,275 @@
...
+ public:
+  NvshmemCollectiveThunk(Kind kind, ThunkInfo thunk_info, bool is_sync);
+
+  struct Buffer {
```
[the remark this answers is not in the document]
Tixxx: there used to be a sync flag in the draft pr but that's removed, will use collective thunk's buffer instead.

xla/core/collectives/communicator.h
```diff
@@ -36,6 +36,19 @@ limitations under the License.
 
 namespace xla {
 
+// This enum helps define the affinity that the communicator
+// spans.
+enum class CommAffinity {
```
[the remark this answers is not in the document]
Tixxx: ok i will remove that for now, but most HPC libs have simple enums to query for intra/inter node comms, it would be good to have a interface for that. I think there are already several places in gpu runtime where we use local devices to determine intra or inter node.

xla/backends/gpu/collectives/nvshmem_communicator.h
```diff
@@ -0,0 +1,127 @@
...
+
+  tsl::AsyncValueRef<Event> Recv(se::DeviceMemoryBase recv_buffer,
+                                 PrimitiveType dtype, size_t count, RankId peer,
+                                 const Executor& executor) {
+    return absl::UnimplementedError("Not implemented.");
+  };
+
+  std::string ToString() const final;
+
+  CommAffinity comm() const { return comm_; }
```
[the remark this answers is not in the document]
Tixxx: removed

xla/backends/gpu/collectives/nvshmem_communicator.cc
```diff
@@ -0,0 +1,326 @@
...
+      reduce, TYPENAME, TYPE, NVSHMEM_TEAM_WORLD, (TYPE*)source_ptr,         \
+      (TYPE*)dest_ptr, count, gpu_stream, reduction_kind);
+
+//==-----------------------------------------------------------------------===//
+// NVSHMEM Communicator
+//==-----------------------------------------------------------------------===//
+
+NvshmemCommunicator::NvshmemCommunicator(NvshmemCollectives* collectives,
+                                         CommAffinity comm)
+    : collectives_(collectives), comm_(comm) {
```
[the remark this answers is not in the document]
Tixxx: yea i needed a check to see if nvshmem is initialized or not.

xla/backends/gpu/runtime/BUILD
Tixxx: not anymore, forgot to remove this after refactoring

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
Tixxx: Yea i was thinking to do that at the beginning, there is some logic that needs to be refactored to have a single parent class for the 2 backends. I think it's better to do it in another pr to not block any subsequent nvshmem prs that we have. I added a TODO in nvshmem_collective_thunk for myself to clean it up.

xla/backends/gpu/runtime/nvshmem_all_reduce_thunk.cc
```diff
@@ -0,0 +1,129 @@
...
+  if (!MatchReductionComputation(inst->called_computations().front())
+           .has_value()) {
+    return absl::UnimplementedError("Unrecognized reduction computation");
+  }
+
+  return absl::OkStatus();
+}
+
+template <typename HloInstType>
+CollectiveOpGroupMode GetGroupModeInst(HloInstType* inst) {
```
PatriosTheGreat, line 82: AllReduce::GetGroupModeInst can be used instead

xla/backends/gpu/collectives/nvshmem_communicator.h
```diff
@@ -0,0 +1,125 @@
...
+    return absl::UnimplementedError("Not implemented.");
+  };
+
+  std::string ToString() const final;
+
+ private:
+  static absl::StatusOr<se::Stream*> ToStream(const Executor& executor);
+
+  NvshmemCollectives* collectives_;  // Parent NvshmemCollectives instance
+  nvshmemx_team_t comm_;             // Underlying NVSHMEM communicator
```
PatriosTheGreat, line 119: Is is field needed? Looks like it just initialized as `NVSHMEMX_TEAM_NODE`. Can this value be inlined in caller sides? Without this field this heads will not need to include nvshmem headers

xla/backends/gpu/collectives/nvshmem_communicator.cc
```diff
@@ -0,0 +1,321 @@
...
+#include "xla/util.h"
+#include "tsl/platform/casts.h"
+
+namespace xla::gpu {
+
+//==-----------------------------------------------------------------------===//
+// NVSHMEM Utility Functions
+//==-----------------------------------------------------------------------===//
+
+static size_t ToRealCount(PrimitiveType dtype, size_t count) {
```
PatriosTheGreat, line 55: Is static needed here?

xla/backends/gpu/collectives/nvshmem_communicator.cc
```diff
@@ -0,0 +1,321 @@
...
+  return primitive_util::IsComplexType(dtype) ? count * 2 : count;
+}
+
+static absl::StatusOr<std::string> ToNvshmemTeamString(nvshmem_team_t comm) {
```
PatriosTheGreat, line 59: Since currently only NVSHMEMX_TEAM_NODE is used, I assume this method can be inlined

PatriosTheGreat (changes requested):

xla/service/gpu/ir_emitter_unnested.h
```diff
@@ -157,6 +157,19 @@ class IrEmitterUnnested : public IrEmitter {
   absl::Status EmitCollectiveAsyncDone(Thunk::Kind kind,
                                        const HloInstruction* instr);
 
+  template <typename NvshmemAllReduceThunkType,
```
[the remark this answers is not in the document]
Tixxx: yea i use it to get the op name from nvshmem allreduce thunk, it used to be the same as allreduce thunk name but i had to add nvshmem specific key word to the name for debugging purposes.

xla/backends/gpu/runtime/nvshmem_collective_thunk.cc
```diff
@@ -0,0 +1,286 @@
...
+}
+
+}  // namespace
+
+NvshmemCollectiveThunk::NvshmemCollectiveThunk(Kind kind, ThunkInfo thunk_info,
+                                               bool is_sync)
+    : Thunk(kind, thunk_info),
+      async_events_(is_sync ? nullptr : new CollectiveThunk::AsyncEvents()) {}
+
+absl::StatusOr<GpuCliqueKey> GetGpuNvshmemCliqueKey(
```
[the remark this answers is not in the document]
Tixxx: yea the nvshmem_comm_id warning is also not applicable, that's a deprecated feature now. i will add it to the CollectiveThunk::GetGpuCliqueKey function

xla/backends/gpu/collectives/nvshmem_communicator.h
Tixxx: sure will remove it

xla/backends/gpu/runtime/nvshmem_all_reduce_thunk.h
```diff
@@ -0,0 +1,86 @@
...
+#include "xla/hlo/ir/hlo_instructions.h"
+#include "xla/service/collective_ops_utils.h"
+#include "xla/stream_executor/stream.h"
+
+namespace xla {
+namespace gpu {
+
+// Thunk that performs a nvshmem-based All-Reduce among CUDA
+// GPU-based replicas.
+class NvshmemAllReduceReduceScatterThunkBase : public NvshmemCollectiveThunk {
```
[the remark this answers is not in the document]
Tixxx: yup, they are mostly sharing a lot of data structures now, i think just some minor consolidation of member functions is needed.

Tixxx:
> HloModule pmap__lambda_
> 
> region_add {
>   lhs = f16[] parameter(0)
>   rhs = f16[] parameter(1)
>   ROOT ret = f16[] add(lhs, rhs)
> }
> 
> ENTRY main.7 {
>   
>     param0 = f16[8]{0} parameter(0)
>     out0 = f16[8]{0} all-reduce(param0), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
>    ...
>     param499 = f16[8]{0} parameter(499)
>     out499 = f16[8]{0} all-reduce(param499), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
>   ROOT result = tuple(
>     out0,...,out499
>   )
> }

Hi @PatriosTheGreat can you upload the full hlo somewhere also with the command that you launch the jobs, i assume it's through //xla/tools/multihost_hlo_runner:hlo_runner_main? I have trying to repro it but doesnt seem to see the deadlock behavior, the symmetric reg allocation takes a while but after that it runs okay.

PatriosTheGreat:
Hi @Tixxx 

I'm using the following script to run benchmarks via multihost_hlo_runner:

```
# Needs to be done once
./configure.py --backend=CUDA --nccl --cuda_compiler CLANG --clang_path=/usr/bin/clang-19

export ROOT_DIR=""
export BENCHMARK_DIR="$ROOT_DIR/nvshmem_benchmarks"
export RESULTS_DIR="$BENCHMARK_DIR/results"

hlo_file="$RESULTS_DIR/all_reduce_nvshmem_128.hlo"
xspace_file="$RESULTS_DIR/all_reduce_nvshmem_128.xplane"

for i in $(seq 0 7);
do
    if [[ "$i" == "0" ]]; then
        pass_xspace=$xspace_file
    else
        pass_xspace=""
    fi

    xla_flags="--xla_gpu_all_reduce_combine_threshold_bytes=1"

    if [[ "$type" == "nvshmem" ]]; then
        xla_flags="$xla_flags --xla_gpu_experimental_enable_nvshmem=true"
    fi

    NVSHMEM_BARRIER_TG_DISSEM_KVAL=8 CUDA_VISIBLE_DEVICES=$i XLA_FLAGS="$xla_flags" bazel run -c opt --config=cuda --copt="-Wno-deprecated-declarations" --copt="-Wno-invalid-offsetof" //xla/tools/multihost_hlo_runner:hlo_runner_main -- \
        --use_spmd_partitioning=true --num_replicas=8 --num_partitions=1 --device_type=gpu \
        --address=127.0.0.1:12345 --num_nodes=8 --task_id=$i --xla_gpu_dump_xspace_to=$pass_xspace \
        $hlo_file &
done

wait
```

An HLO I'm using is:

```
HloModule pmap__lambda_

region_add {
  lhs = f32[] parameter(0)
  rhs = f32[] parameter(1)
  ROOT ret = f32[] add(lhs, rhs)
}

ENTRY main.7 {
  param0 = f32[128]{0} parameter(0)
  out0 = f32[128]{0} all-reduce(param0), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
  param1 = f32[128]{0} parameter(1)
  out1 = f32[128]{0} all-reduce(param1), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
  ROOT result = tuple(out0, out1)
}
```

I'm compiling nvshmem from sources [using this PR](https://github.com/openxla/xla/pull/22019), but I think using dynamic library should also work.

First there are some compilation issues like:

```
In file included from xla/backends/gpu/collectives/nvshmem_collectives.cc:15:
./xla/backends/gpu/collectives/nvshmem_collectives.h:76:3: warning: 'CreateCommunicators' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
   76 |   CreateCommunicators(const CliqueKey& clique_key,
      |   ^
./xla/core/collectives/collectives.h:74:3: note: overridden virtual function is here
   74 |   CreateCommunicators(const CliqueKey& clique_key,
      |   ^
1 warning generated.
```

After fixing this issue I'm getting a deadlock during execution.

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2234,6 +2214,135 @@ absl::Status IrEmitterUnnested::EmitCollectiveAsyncDone(
...
+  // Find canonical async event.
+  CollectivesAsyncEvents& collectives_async_events =
+      GetCollectivesAsyncEvents();
+  auto async_events_it = collectives_async_events.find(start);
+  TF_RET_CHECK(async_events_it != collectives_async_events.end())
+      << "couldn't find async events for start operation";
+
+  // Can be null if no start thunk was created (e.g. if the start op is
+  // degenerate), in which case there's nothing to do here.
+  if (!async_events_it->second) return absl::OkStatus();
```
PatriosTheGreat, line 2337: Could you please make it multi-line if?
if () {
   return...
}

PatriosTheGreat (changes requested):

Tixxx:
> Hi @Tixxx
> 
> I'm using the following script to run benchmarks via multihost_hlo_runner:
> 
> ```
> # Needs to be done once
> ./configure.py --backend=CUDA --nccl --cuda_compiler CLANG --clang_path=/usr/bin/clang-19
> 
> export ROOT_DIR=""
> export BENCHMARK_DIR="$ROOT_DIR/nvshmem_benchmarks"
> export RESULTS_DIR="$BENCHMARK_DIR/results"
> 
> for i in $(seq 0 7);
> do
>     xla_flags="--xla_gpu_all_reduce_combine_threshold_bytes=1  --xla_gpu_experimental_enable_nvshmem=true"
>     NVSHMEM_BARRIER_TG_DISSEM_KVAL=8 CUDA_VISIBLE_DEVICES=$i XLA_FLAGS="$xla_flags" bazel run -c opt --config=cuda --copt="-Wno-deprecated-declarations" --copt="-Wno-invalid-offsetof" //xla/tools/multihost_hlo_runner:hlo_runner_main -- \
>         --use_spmd_partitioning=true --num_replicas=8 --num_partitions=1 --device_type=gpu \
>         --address=127.0.0.1:12345 --num_nodes=8 --task_id=$i --xla_gpu_dump_xspace_to=$pass_xspace \
>         $RESULTS_DIR/all_reduce_nvshmem_128.hlo &
> done
> 
> wait
> ```
> 
> An HLO I'm using is:
> 
> ```
> HloModule pmap__lambda_
> 
> region_add {
>   lhs = f32[] parameter(0)
>   rhs = f32[] parameter(1)
>   ROOT ret = f32[] add(lhs, rhs)
> }
> 
> ENTRY main.7 {
>   param0 = f32[128]{0} parameter(0)
>   out0 = f32[128]{0} all-reduce(param0), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
>   param1 = f32[128]{0} parameter(1)
>   out1 = f32[128]{0} all-reduce(param1), replica_groups={}, to_apply=region_add, backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
>   ROOT result = tuple(out0, out1)
> }
> ```
> 
> I'm compiling nvshmem from sources [using this PR](https://github.com/openxla/xla/pull/22019), but I think using dynamic library should also work.
> 
> First there are some compilation issues like:
> 
> ```
> In file included from xla/backends/gpu/collectives/nvshmem_collectives.cc:15:
> ./xla/backends/gpu/collectives/nvshmem_collectives.h:76:3: warning: 'CreateCommunicators' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
>    76 |   CreateCommunicators(const CliqueKey& clique_key,
>       |   ^
> ./xla/core/collectives/collectives.h:74:3: note: overridden virtual function is here
>    74 |   CreateCommunicators(const CliqueKey& clique_key,
>       |   ^
> 1 warning generated.
> ```
> 
> After fixing this issue I'm getting a deadlock during execution.

hi @PatriosTheGreat  just an update on this issue, I have seen it where if the copies are added to the nvshmem ar to copy them back to the default memory space before returning from the module, the deadlock doesnt happen anymore. There's some intricacies on when the synchronization is done that interferes with nvshmem's internal barrier. I will need to look into this to see if xla needs to add copies in this case for collective memory or other device memory spaces.
If the communicator api looks okay, i will split it out into a separate pr to not block other works. I will add the thunk later.

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -0,0 +1,279 @@
...
+    std::string stderr_str;
+    int child_status =
+        child[rank_id].Communicate(nullptr, &stdout_str, &stderr_str);
+    EXPECT_EQ(child_status, 0) << " rank " << rank_id << "\nstdout:\n"
+                               << stdout_str << "\nstderr:\n"
+                               << stderr_str;
+  }
+}
+
+absl::Status NvshmemCollectiveTestBody(int rank_id, int num_ranks,
```
PatriosTheGreat, line 147: Is this test-case passes on your machine? I somehow getting literal result 30 with NVSHMEM 3.25, not sure if this an issue of my machine configuration or the code

xla/pjrt/gpu/BUILD
```diff
@@ -564,3 +564,70 @@ cc_library(
     ],
     alwayslink = 1,
 )
+
+xla_test(
+    name = "nvshmem_gpu_collectives_test",
+    srcs = ["nvshmem_gpu_collectives_test.cc"],
```
PatriosTheGreat, line 570: This test and the `se_gpu_pjrt_client_nvshmem_test` should eventually move to `xla/tests` and be rewritten like `xla/tests/all_reduce_test.cc`.  We can do this in a separate PR, but please add a TODO for this target.

PatriosTheGreat (changes requested):

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -0,0 +1,279 @@
...
+  xla::CompileOptions options;
+  options.executable_build_options.mutable_debug_options()
+      ->set_xla_gpu_experimental_enable_nvshmem(true);
+  options.executable_build_options.set_run_backend_only(true);
+  options.executable_build_options.set_use_spmd_partitioning(false);
+  options.executable_build_options.set_num_replicas(num_ranks);
+  TF_ASSIGN_OR_RETURN(std::string data_type_str, GetDataTypeString(data_type));
+  const std::string kProgram =
+      absl::StrFormat(R"(
+    HloModule NvshmemAr
```
PatriosTheGreat, line 182: @olegshyshkov mentioned that all reduce semantic supports replica_groups. I checked the following HLO:

```
HloModule subgroups, entry_computation_layout={(f32[128,32]{0,1})->f32[128,32]{0,1}}, replica_count=4

add {
  lhs = f32[] parameter(0)
  rhs = f32[] parameter(1)
  ROOT add = f32[] add(lhs, rhs)
}

ENTRY AllReduceWithSubgroups {
  input = f32[128,32]{0,1} parameter(0)
  ROOT all-reduce = f32[128,32]{0,1} all-reduce(input), replica_groups={{0,1},{2,3}}, to_apply=add
}
```

It works with NCCL, but with 4 processes NVSHMEM it leads to the deadlock. Are you planning to support this semantic later or should it be added here?

PatriosTheGreat (changes requested):

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
Tixxx: yes, this would be supported with the team split logic. i had that in internal reviews but not in this first version. i will open an upstream pr for that separately.

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
Tixxx: ok so after i rebased and did a clean build from scratch, this test started hanging for me.. From the nvshmem traces, it looks like there's a semaphore in the allreduce call that doesn't get incremented because a write to the buffer never returned. i will debug this further

Tixxx: it looks like this is caused by compiling and running with different versions of nvshmem. when i compile with 3.2.5 header and run with runtime lib, the test passes. The hang i was seeing was also caused by that.

xla/pjrt/gpu/BUILD
Tixxx: done

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -0,0 +1,279 @@
...
+#include "xla/status_macros.h"
+#include "xla/stream_executor/device_memory.h"
+#include "xla/stream_executor/stream.h"
+#include "xla/tests/literal_test_util.h"
+#include "xla/tsl/lib/core/status_test_util.h"
+#include "xla/tsl/platform/subprocess.h"
+#include "xla/types.h"
+#include "xla/util.h"
+#include "xla/xla_data.pb.h"
+#include "tsl/platform/casts.h"
```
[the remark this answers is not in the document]
Tixxx: removed

xla/backends/gpu/runtime/thunk.h
```diff
@@ -492,13 +496,13 @@ class Thunk {
 
   // A helper function to get the `GpuCollectives*` pointer from the
   // CollectiveExecuteParams.
-  static absl::StatusOr<GpuCollectives* absl_nonnull> GetGpuCollectives(
```
PatriosTheGreat, line 508: No-op change

xla/backends/gpu/runtime/thunk.h
```diff
@@ -492,13 +496,13 @@ class Thunk {
...
+  static absl::StatusOr<GpuCollectives * absl_nonnull> GetGpuCollectives(
       CollectiveExecuteParams const& params);
 
   // A helper function to get the `GpuCollectives*` pointer from the
   // thunk parameters. Returns an error if collectives API is not provided.
   template <typename Params>
-  static absl::StatusOr<GpuCollectives* absl_nonnull> GetGpuCollectives(
+  static absl::StatusOr<GpuCollectives * absl_nonnull> GetGpuCollectives(
```
PatriosTheGreat, line 518: No-op change

xla/backends/gpu/runtime/thunk.cc
```diff
@@ -391,7 +395,7 @@ absl::StatusOr<ThunkProto> Thunk::ToProto() const {
   return proto;
 }
 
-absl::StatusOr<GpuCollectives* absl_nonnull> Thunk::GetGpuCollectives(
+absl::StatusOr<GpuCollectives * absl_nonnull> Thunk::GetGpuCollectives(
```
PatriosTheGreat, line 418: No-op change

xla/service/gpu/gpu_executable.cc
```diff
@@ -457,6 +464,18 @@ absl::Status MaybeSyncAndProfile(const ServiceExecutableRunOptions* run_options,
   return absl::OkStatus();
 }
 
+absl::Status BarrierAfterExecutable(const DebugOptions* debug_options,
```
PatriosTheGreat, line 468: This operation is an NVSHMEM specific. Can we try to move it down to nvshmem thunks?

For example we can add to thunk interface an additional method like "FinalizeExecution" which by default is no-op. A sequence_thunk can call this operation in a loop for all thunks. NVSHMEM thunks can override this operation and perform barrier once per call.

We might need to add some common nvshmem thunk class here to perform a single barrier.

PatriosTheGreat (changes requested):

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
PatriosTheGreat: Hmm I'm still getting wrong results internally. I tried to log execution and what I found is that both devices are using the same send/recv buffer pointers. Example:

```
Device 0:
Launch NVSHMEM AllReduce operation on device #0; send_buffer=0x10027c80600; recv_buffer=0x10027c80600; dtype=f32; count=1; reduction_kind=sum; comm=node; team=2;stream=0x55f822a623b0

Device 1:
Launch NVSHMEM AllReduce operation on device #1; send_buffer=0x10027c80600; recv_buffer=0x10027c80600; dtype=f32; count=1; reduction_kind=sum; comm=node; team=2;stream=0x55ac683760e0
```
Is this expected?

PatriosTheGreat: I tried to investigate a bit into this issue and looks like it happens only when nvshmem chooses to use `NVLS One Shot Allreduce (AR)` algorithm in include/non_abi/device/coll/reduce.cuh:1190. This algorithm was added in 3.25 IIUC and when I'm explicitly changing it to  `NVLS Two Shot Allreduce (RS + AG)` or default (by setting reduce_algo=3/5 in line before switch statement) -- the problem goes away and I'm getting a correct result.

PatriosTheGreat: NCCL explicitly supports in-place operations when recv and send buffers are the same:
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html?highlight=equal 

Does NVSHMEM support this?

olegshyshkov: Looks like NVSHMEM also supports in-place operations for AllReduce:

> Pointer arguments to NVSHMEM routines that point to non-const data must not overlap in memory with other arguments to the same NVSHMEM operation, with the exception of in-place reductions as described in Section [NVSHMEM_REDUCTIONS](https://docs.nvidia.com/nvshmem/api/gen/api/collectives.html#subsec-shmem-reductions)

https://docs.nvidia.com/nvshmem/api/gen/exec-model.html#invoking-nvshmem-operations

> The source and dest arguments must either be the same symmetric address, or two different symmetric addresses corresponding to buffers that do not overlap in memory. That is, they must be completely overlapping or completely disjoint.

https://docs.nvidia.com/nvshmem/api/gen/api/collectives.html#subsec-shmem-reductions

xla/backends/gpu/runtime/thunk.h
Tixxx: looks like there's some change in the CI for the formatting. this is automatically done by clang format, if i dont keep this formatting, the clang-format CI will fail. shall i just keep the failure as is?

xla/backends/gpu/runtime/thunk.cc
Tixxx: same issue with above

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
Tixxx: yes in-place update is supported by nvshmem when the input and output buffers have the same offset and size.
The address being the same is what nvshmem's symmetric allocation ensures, all the ranks have the same offset and base pointer for a given collective op. When registering with multicast with symmetric addresses, the latency of collectives is minimized.
Regarding the incorrect result, does it only happen with in-place allreduce?
Can i suggest to merge this pr first and work on upgrading to 3.2.5 separately? I'm investigating internally with nvshmem team to find the root cause. This scenario is being tested regularly within nvshmem though.

xla/service/gpu/gpu_executable.cc
Tixxx: Yea i had that initially but we'd only need to do 1 single barrier before the executable returns to make sure all the buffers are ready for the next execution. Doing that once per thunk is costly. Do you suggest that we add a nvshmem barrier thunk at the end of thunk sequence whenever there's a nvshmem collective thunk present?

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
Tixxx: I tried changing xla's data flow analysis to do out of place update for ar and the results looks correct for one-shot ar. Also confirmed that one shot in-place ar is tested only with device side APIs. I filed a bug to nvshmem team to fix it. I think for now, we can add a hook to data flow analysis to tell it when to do in-place ar and for nvshmem ar it will have to be out of place. Does this sound okay?

xla/backends/gpu/runtime/thunk.cc
Tixxx: i removed the extra space to show the formatting failure

xla/service/gpu/gpu_executable.cc
PatriosTheGreat: Since we already mutating collective_params to initialize need_barrier, we can use this just to signal from the first nvshmem thunk to all others that barrier was completed. So the `FinalizeExecution` or `PostExecutionOnStream` in nvshmem thunk base class can be something like:
```
if (collective_param.need_barrier) {
   NvshmemBarrier();
   collective_param.need_barrier = false;
}
```

In gpu executable this can be called as:

```
thunk_sequence.ExecuteOnStream(execute_params);
thunk_sequence.PostExecutionOnStream(execute_params);
```

This way barrier will be executed only once per execution.

Emitting a barrier thunk with nvshmem sounds like a better solution since then we won't need to make any changes in the runtime and all nvshmem related changes will be in one place (in the emitter code). So if this is not much harder to implement this would be preferable.

xla/backends/gpu/runtime/thunk.h
PatriosTheGreat: Ah that strange, I'll try to figure out why it has changed. So far you can keep whenever clang format does.

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
PatriosTheGreat: Thanks for a context.
We are using nvshmem 3.25 internally already and there is no easy way to go back to an old nvshmem. So with in-place all reduce this test is failing in our side.

Disabling in-place operation for nvshmem in dataflow analysis sounds a good workaround for now. I tried it locally and it fixes a wrong result issue.

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -0,0 +1,395 @@
...
+
+absl::Status NvshmemCollectiveTestBody(int rank_id, int num_ranks,
+                                       int input_data_type) {
+  xla::PrimitiveType data_type = (xla::PrimitiveType)input_data_type;
+  std::unique_ptr<xla::DistributedRuntimeService> service;
+  if (rank_id == 0) {
+    xla::CoordinationServiceImpl::Options service_options;
+    service_options.num_nodes = num_ranks;
+    TF_ASSIGN_OR_RETURN(service, xla::GetDistributedRuntimeService(
+                                     "[::]:123498", service_options));
```
PatriosTheGreat: Can we keep 12345 port here?

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
Tixxx: cool, let me open another pr for that

PatriosTheGreat (approved):

PR #25916, opened 2025-04-30
Added SafeStaticInit for deadlock free safe static initialization

vfdev-5:
Description:
- Added SafeStaticInit for deadlock free safe static initialization under free-threading
- Rewrote GetNumpyScalarTypes method using SafeStaticInit

cc @hawkinsp

### DIFF
--- a/xla/python/types.cc
+++ b/xla/python/types.cc
@@ -31,9 +31,9 @@ limitations under the License.
 #include "absl/strings/string_view.h"
 #include "absl/types/span.h"
 #include "nanobind/nanobind.h"
-#include "nanobind/ndarray.h"  // IWYU pragma: keep
-#include "nanobind/stl/shared_ptr.h"  // IWYU pragma: keep
-#include "nanobind/stl/string.h"  // IWYU pragma: keep
+#include "nanobind/ndarray.h"          // IWYU pragma: keep
+#include "nanobind/stl/shared_ptr.h"   // IWYU pragma: keep
+#include "nanobind/stl/string.h"       // IWYU pragma: keep
 #include "nanobind/stl/string_view.h"  // IWYU pragma: keep
 #include "xla/layout.h"
 #include "xla/literal.h"
@@ -41,6 +41,7 @@ limitations under the License.
 #include "xla/python/ifrt/dtype.h"
 #include "xla/python/nb_numpy.h"
 #include "xla/python/pjrt_ifrt/pjrt_dtype.h"
+#include "xla/python/safe_static_init.h"
 #include "xla/shape.h"
 #include "xla/shape_util.h"
 #include "xla/status_macros.h"
@@ -398,8 +399,9 @@ absl::StatusOr<nb_dtype> IfrtDtypeToDtypeWithTokenCanonicalization(
 }
 
 const NumpyScalarTypes& GetNumpyScalarTypes() {
-  static const NumpyScalarTypes* singleton = []() {
-    NumpyScalarTypes* dtypes = new NumpyScalarTypes();
+  auto init_fn = []() {
+    std::unique_ptr<NumpyScalarTypes> dtypes =
+        std::make_unique<NumpyScalarTypes>();
     nb::module_ numpy = nb::module_::import_("numpy");
     nb::module_ ml_dtypes = nb::module_::import_("ml_dtypes");
     dtypes->np_bool = nb::object(numpy.attr("bool_"));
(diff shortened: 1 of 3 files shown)


xla/util.h
```diff
@@ -1018,6 +1018,45 @@ struct FreeDeleter {
...
+// auto func = [](){
+//     // for example we call some python code using nanobind
+//     nb::module_ numpy = nb::module_::import_("numpy");
+//     auto np_int8 = nb::object(numpy.attr("int8"));
+//     SomeType* obj = new SomeType(np_uint8);
+//     return obj;
+// }
+// SomeType* p = SafeStaticInit<SomeType>(func);
+template<typename T, typename F>
+T * SafeStaticInit(F func) {
```
hawkinsp, line 1043: I'd be tempted to return a reference rather than a pointer. No strong reason, but it would read better to me.

xla/python/types.cc
```diff
@@ -446,7 +446,9 @@ const NumpyScalarTypes& GetNumpyScalarTypes() {
     dtypes->np_longlong = nb::object(numpy.attr("longlong"));
     dtypes->np_intc = nb::object(numpy.attr("intc"));
     return dtypes;
-  }();
+  };
+
+  const NumpyScalarTypes* singleton = SafeStaticInit<NumpyScalarTypes>(func);
```
hawkinsp, line 451: OOC, is the template parameter needed? I would have thought that template argument deduction would get it.

xla/util.h
```diff
@@ -1018,6 +1018,45 @@ struct FreeDeleter {
 std::unique_ptr<void, FreeDeleter> AlignedAlloc(std::size_t alignment,
                                                 std::size_t size);
 
+// Serialized static variable initialization from the `func` output.
+// It avoid a deadlock between static variable initialization lock
```
hawkinsp, line 1022: I'd note that:
a) this does not hold a lock while func() is called, which means that:
b) this may call func multiple times if invoked concurrently, but the output from all but one will be discarded.

It also occurs to me that it might be better to use a PEP 703 critical section here, since it would avoid the multiple call most of the time. For example in free threading mode, we could do this:

```
static nb::object* mutex_object = new nb::object();

nb::ft_object_guard lock(*mutex_object);
if (!initialized) {
  init();
}
```

I believe this will avoid the deadlock also because outer critical sections are released if you acquire an inner critical section.

xla/util.h
```diff
@@ -1018,6 +1018,45 @@ struct FreeDeleter {
...
+T * SafeStaticInit(F func) {
+  static absl::Mutex mutex;
+  static T* output = nullptr;
+  {
+    absl::MutexLock lock(&mutex);
+    if (output) {
+      return output;
+    }
+  }
+  T* p = func();
```
hawkinsp, line 1052: This leaks a copy of T if someone else has already populated output. I would change func to return `std::unique_ptr<T>` and then use `output = p.release()` below?

Then if we end up not needing `p`, it will be freed.

hawkinsp (changes requested):

xla/util.h
hawkinsp: (But note you'd need to use something like an absl::Mutex in non-FT mode.)

hawkinsp: Actually, never mind. You wouldn't need an `absl::Mutex`: the GIL would suffice.

xla/python/types.cc
vfdev-5: Compiler failed to get the type:
```
external/xla/xla/python/types.cc:451:39: error: no matching function for call to 'SafeStaticInit'
  451 |   const NumpyScalarTypes* singleton = SafeStaticInit(init_fn);         
      |                                       ^~~~~~~~~~~~~~                   
external/xla/xla/util.h:1052:4: note: candidate template ignored: couldn't infer template argument 'T'
```

xla/util.h
vfdev-5: We finally agreed in another chat to use absl::Mutex

xla/util.h
```diff
@@ -1018,6 +1018,53 @@ struct FreeDeleter {
 std::unique_ptr<void, FreeDeleter> AlignedAlloc(std::size_t alignment,
                                                 std::size_t size);
 
+// Serialized static variable initialization from the `init_fn` output.
```
hawkinsp, line 1021: I would not put this function here, instead I would put it in xla/python in its own file

Two reasons:
* util packages tend to grow a bunch of unrelated things, best to keep things small and focused.
* second: this idiom is really related to safe initialization in Python. Absent the GIL or similar we don't need it.

xla/util.h
```diff
@@ -1018,6 +1018,53 @@ struct FreeDeleter {
...
+T* SafeStaticInit(F init_fn) {
+  static absl::Mutex mutex;
+  static T* output = nullptr;
+  {
+    absl::MutexLock lock(&mutex);
+    if (output) {
+      return output;
+    }
+  }
+  std::unique_ptr<T> p(init_fn());
```
hawkinsp, line 1060: It is safer if init_fn() allocates and returns a `std::unique_ptr` rather than you wrapping it outside the function, consider exception safety for example.

hawkinsp (changes requested):

hawkinsp (approved):

PR #25953, opened 2025-05-01
[XLA:FFI][BugFix] Add Token Args and Rets for typed FFI on CPU backend

akhilgoe:
This PR adds support for token args and rets in the typed FFI for legacy (non-thunk) runtime. It also adds code to initialize the handler execution state when an instantiation callback is provided. These changes resolve crashes occurring in certain custom calls that use the typed FFI API.


ezhulenev (approved):

ezhulenev:
FYI I plan to start deleting old runtime soon, if you have any hard blockers for switching to a new one please let us know, we believe that we fixed almost all perf regressions (few smaller one are still there in different edge cases around while loops).

PR #26001, opened 2025-05-02
[NVIDIA GPU] Change UB register cache to include size in the key

Tixxx:
UB registration requires the exact pointer and size. The current cache doesn't use size as part of the key which causes NCCL to skip registration for some buffers.

### DIFF
--- a/xla/backends/gpu/runtime/collective_thunk.cc
+++ b/xla/backends/gpu/runtime/collective_thunk.cc
@@ -340,18 +340,17 @@ absl::Status RegisterBufferOnce(GpuCollectives* collectives,
           {executor->device_ordinal(), buffer.size(), comm, buffer.opaque()})) {
     // ncclCommRegister will internally get and use the base address/size of the
     // address we provide.
-    LOG(INFO) << "Registering " << buffer.opaque()
-              << " with size: " << buffer.size()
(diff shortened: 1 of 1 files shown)


xla/backends/gpu/runtime/collective_thunk.cc
```diff
@@ -337,13 +337,21 @@ absl::Status RegisterBufferOnce(GpuCollectives* collectives,
 
   absl::MutexLock lock(&all_registered.mu);
   if (!all_registered.records.contains(
-          {executor->device_ordinal(), comm, base_buffer.opaque()})) {
+          {executor->device_ordinal(), buffer.size(), comm, buffer.opaque()})) {
     // ncclCommRegister will internally get and use the base address/size of the
     // address we provide.
+    LOG(INFO) << "Registering " << buffer.opaque()
```
PatriosTheGreat, line 343: This might be too much to log. Could you change this and else logging to VLOG(2)?

PatriosTheGreat (changes requested):

xla/backends/gpu/runtime/collective_thunk.cc
Tixxx: yea that makes sense, i changed it to vlog(5)

PatriosTheGreat (approved):

PR #26006, opened 2025-05-02
port xla to meson

raghaven447:


dimitar-asenov:
Hi there. Thanks for your interest in XLA. At this point we do not plan to support additional build systems. If you want to start a discussion about this, please feel to comment on this PR or a new issue.

raghava447:
@dimitar-asenov: Thanks for your response.

I think that this is a good addition as it facilitates the TensorflowLite-C building easy with Meson and could be easily integrated with the frameworks which use TensorflowLite-C as dependency.

Please let me know your thoughts and further steps to support Meson build system.

dimitar-asenov:
@MichaelHudgins, what is your opinion here?

PR #26012, opened 2025-05-02
Convert computations to MakeNonfusionComputations in async wrappers

chaserileyroberts:
Using `module->computations` was never safe to use when adding new subcomputations (which is required when creating `AsyncStart` instructions), as that method would return a view to the `HloComputation`s, which is not memory safe and could occasionally cause segfaults.  

This change brings it inline with other similar passes. 

Example:
https://github.com/openxla/xla/blob/df11db787a1836f65f277d249f75425a7948a380/xla/service/gpu/transforms/stream_attribute_async_wrapper.cc#L64

See the description of `MakeNonfusionComputations` https://github.com/openxla/xla/blob/7f82d14ffbfdd1e344d1945b62ec91f0c3dff654/xla/hlo/ir/hlo_module.h#L354


reedwm (approved):
Approving on behalf of @frgossen who currently doesn't have phone to login to github.

PR #26046, opened 2025-05-05
Generalize dynamic memcpy support.

jreiffers:
Note: this optimization is currently not enabled.

Currently, dynamic memcpy fusions only trigger if the innermost fusion contains no offset computations, i.e. all dynamic-(update-)slice offsets are parameters, constants, or values that are clamped to 0. Unfortunately, this is not sufficient in general, since sometimes the offset computations are fused. 

To solve this, we need to generalize ResolveFunctionalDependencyOnInductionVariable. The previous logic traced parameters through the call stack until a while loop is reached, assuming all computations happen directly inside the while loop. The new logic is similar, but does not make the assumption that all computations happen there - for each layer of the stack, it computes the set of required parameters, still verifying that no side-effecting instructions are encountered on the way. Otherwise, arbitrary computations are allowed in each layer.

In the dynamic memcpy thunk, we then start in the while loop body, and for each call we evaluate the previously computed required parameters.

Apologies for the large PR, I didn't see a way to split this up.

### DIFF
--- a/xla/backends/gpu/runtime/copy_thunk.cc
+++ b/xla/backends/gpu/runtime/copy_thunk.cc
@@ -19,6 +19,7 @@ limitations under the License.
 #include <memory>
 #include <utility>
 
+#include "absl/container/node_hash_map.h"
 #include "absl/log/log.h"
 #include "absl/status/status.h"
 #include "absl/synchronization/mutex.h"
@@ -208,16 +209,22 @@ absl::Status CopyDoneThunk::ExecuteOnStream(const ExecuteParams& params) {
 
 namespace {
 
-absl::StatusOr<int64_t> EvaluateDynamicOffsets(
-    HloEvaluator& evaluator,
-    absl::Span<const DynamicMemcpyThunk::MemcpyDescriptor::DynamicOffset>
-        offsets) {
-  int64_t offset_sum = 0;
-  for (const auto& offset : offsets) {
+class DynamicOffsetEvaluator {
+ public:
+  absl::StatusOr<int64_t> EvaluateOffset(
+      const DynamicMemcpyThunk::MemcpyDescriptor::DynamicOffset& offset);
+
+ private:
+  absl::node_hash_map<const HloInstruction*, Literal> known_values_;
+};
+
+absl::StatusOr<int64_t> DynamicOffsetEvaluator::EvaluateOffset(
+    const DynamicMemcpyThunk::MemcpyDescriptor::DynamicOffset& offset) {
+  // Set up the value of the induction variable, if it's not known yet.
+  if (!known_values_.contains(offset.induction_variable)) {
     TF_ASSIGN_OR_RETURN(
         auto config,
         offset.while_loop->backend_config<xla::WhileLoopBackendConfig>());
-
     TF_RET_CHECK(config.has_known_init_step());
     TF_ASSIGN_OR_RETURN(int64_t iteration,
                         WhileThunk::CurrentLoopIteration(offset.while_loop));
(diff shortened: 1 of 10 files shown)


xla/service/gpu/ir_emission_utils.cc
```diff
@@ -865,13 +855,41 @@ ResolveFunctionalDependencyOnInductionVariable(
...
+      } else if (caller->opcode() == HloOpcode::kWhile) {
+        // We only support dependencies on the inner-most while loop's
+        // induction variable, so we don't need to continue the traversal here.
+        if (unique_param || !instruction->shape().IsTuple()) {
+          VLOG(5) << "Failed to match parameters.";
+          return std::nullopt;
+        }
+        while_loop = caller;
+        unique_param = instruction;
+        continue;
```
akuegel, line 882: This continue statement seems not needed?

xla/service/gpu/ir_emission_utils.h
```diff
@@ -362,28 +362,31 @@ absl::StatusOr<std::string> FingerprintWithBackendConfig(
...
-  // to have no other transitive dependencies (except constants).
-  const HloInstruction* derived_value;
+  // The dependency may be via multiple levels of intermediate calls. At each
+  // level, we need to know which parameters to evaluate, since not all of them
+  // may be relevant. The while loop's body is not included here, since the
+  // induction variable is implicitly the only dependency that is allowed.
+  // The size of the value is always the same as the number of parameters in the
+  // computation. We request a single element of inlined space, which will
+  // automatically pick the optimum value (16, usually).
+  absl::flat_hash_map<const HloComputation*,
```
akuegel, line 372: Seems we need a header include for flat_hash_map in this file. Pre-existing as flat_hash_map is already used without header include elsewhere in this file, but I guess ClangTidy will flag it.

xla/backends/gpu/runtime/copy_thunk.cc
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 #include <memory>
 #include <utility>
 
+#include "absl/container/node_hash_map.h"
```
akuegel, line 22: I guess this needs a corresponding dependency in the BUILD file?

jreiffers:
Thanks for the review! PTAL.

xla/service/gpu/ir_emission_utils.h
```diff
@@ -23,6 +23,7 @@ limitations under the License.
 #include <variant>
 #include <vector>
 
+#include "absl/container/flat_hash_map.h"
```
akuegel, line 26: This needs a BUILD file dependency

jreiffers: Oops, flat_hash_set != flat_hash_map. Sorry about that.

akuegel (approved):

xla/service/gpu/ir_emission_utils.cc
```diff
@@ -895,28 +912,7 @@ ResolveFunctionalDependencyOnInductionVariable(
...
-  }
-
-  // Find the while loop for 'unique_param'.
-  auto while_instr_it = std::find_if(
-      call_stack_it, call_stack_end, [&](const HloInstruction* instr) {
-        return instr->opcode() == HloOpcode::kWhile &&
-               unique_param == instr->while_body()->parameter_instruction(0);
-      });
-
-  if (while_instr_it == call_stack_end) {
```
akuegel, line 913: It seems we have lost this check. I didn't notice myself, but ClangTidy complained:

dereferencing a potentially nullable pointer (line 915)

I think it could be possible to construct a case where while_loop is still nullptr after the loop.

jreiffers: This is not possible, because while_loop is nullptr iff unique_param is nullptr, and the latter is checked. I'll add an explicit check to make that more obvious.

akuegel (approved):

xla/service/gpu/ir_emission_utils.cc
akuegel: Good point, but the extra check will also at least silence ClangTidy :)

PR #26057, opened 2025-05-05
[ROCM][NFC] redzone checker kernel uses unified implementation with a single in-process kernel

pemeliya:
Here I refactored redzone checker kernel which made us problems on ROCM side due to static variable used.
Now both implementations CUDA/ROCM are unified and use a simple in-process kernel, similar to buffer_comparator one.

@xla-rotation: can you have a look please?


xla/stream_executor/gpu/redzone_allocator.cc
```diff
@@ -269,10 +275,11 @@ absl::StatusOr<DeviceMemoryBase> RedzoneAllocator::CreateBuffer(
 absl::StatusOr<RedzoneCheckStatus> RedzoneAllocator::CheckRedzones() const {
   StreamExecutor* executor = stream_->parent();
 
-  TF_ASSIGN_OR_RETURN(ComparisonKernel * kernel,
-                      GetComparisonKernel(stream_->parent(), GpuAsmOpts()));
+  TF_ASSIGN_OR_RETURN(auto kernel, 
+      ComparisonKernel::FactoryType::Create(executor, "RedzoneCheckerKernel",
+                                      redzone_checker_kernel::kernel()));
```
beckerhe, line 280: Hey, thank you for the change. I like it. Actually I've tried something similar before as well.

But it got rolled back because it tanked the compiled-time on some Tensorflow models. The problem is that this line reloads the kernel on each call to `CheckRedzones`. That's why the old version had this annoying static cache in there.

A good solution would be to make the `kernel` variable a member in the `RedzoneAllocator` class, but that also doesn't work with Tensorflow since Tensorflow creates a new RedzoneAllocator for each redzone allocation.

That means if we want to ship this change we will probably need to bring back the static cache from LoadKernelOrGetPtr. But I haven't looked into how this is done in the best way.

pemeliya: ok, I see, thank you for the review. Let me think about it

pemeliya: I wonder if BufferComparator should then also suffer from this problem ?
I mean, in [gemm_algorithm_picker](https://github.com/openxla/xla/blob/5080f7d991af28a0ae2c15131ebcac35ebe33d57/xla/service/gpu/autotuning/gemm_algorithm_picker.cc#L367-L369), we basically reload BufferComparator kernel for each solution being tuned. Do this make a problem for us?

pemeliya: ah I think I got your point: CheckRedzones calls negatively affected the compilation time of some Tensorflow models because RedzoneAllocator kernel **was written in PTX**?  That's why it triggered jit-compilation everytime the kernel is instantiated (probably it was due to historical reasons when StreamExecutor only supported assembly kernels) ? 

But now I use **in-process** kernel symbol, so I assume redzone kernel instantiation shall be fast now?

beckerhe (approved):

PR #26100, opened 2025-05-06
Clean up special cases for AsyncStart in HloComputation.

jreiffers:
These are no longer needed, since computations now know their callers. I kept the utility function to check if a computation is async, since it is used from many places.

I had to remove one check from the AsyncStart constructor. Previously, it didn't actually verify that the computation isn't already called by an async-start, just that the special field is unset. I don't see a way to have this check and make the code in CommandBufferScheduling work.

After this, the only remaining special case is kFusion.


akuegel (approved):

jreiffers:
Looking into the test failures.

akuegel (approved):

PR #26101, opened 2025-05-06
Allow bitcasts in dynamic memcpy fusions.

jreiffers:
Both before and after the dynamic-(update-)slice.

This is in line with other copy fusions. I just missed this case previously.

### DIFF
--- a/xla/backends/gpu/codegen/copy.cc
+++ b/xla/backends/gpu/codegen/copy.cc
@@ -44,9 +44,13 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
-HloInstructionAdaptor SkipBitcast(HloInstructionAdaptor adaptor) {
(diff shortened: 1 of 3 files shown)


xla/backends/gpu/codegen/copy.cc
```diff
@@ -44,6 +44,15 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+HloInstructionAdaptor SkipOptionalBitcast(HloInstructionAdaptor adaptor) {
```
akuegel, line 47: Could this be in an anonymous namespace? Same below?

jreiffers: Yes, of course. Thanks!

akuegel (approved):

PR #26196, opened 2025-05-07
force delay scheduling start to extend overlap interval

Tixxx:
📝 Summary of Changes
This pr introduces a new heuristics to delay scheduling start based on the op type when the overlap limit is larger than 1 and default cost model is used.

🎯 Justification
The goal is to extend the overlap interval as much as possible to include more computes and provide the best out of the box scheduling.
🚀 Kind of Contribution
 ⚡️ Performance Improvement

hlo_llama31_8b_bf16_1x8.hlo -> 1359ms -> 1318.3ms
hlo_llama31_8b_bf16_2x8.hlo -> 1366ms -> 1321.4ms
hlo_llama31_8b_fp8_1x8.hlo -> 1123ms -> 1089ms
hlo_llama31_8b_fp8_2x8.hlo -> 1141ms -> 1.95.5ms
🧪 Unit Tests:
Added unit tests

🧪 Execution Tests:
included

### DIFF
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -585,8 +585,7 @@ absl::Status RunLatencyHidingSchedulerPasses(
   // If overlap limit is set to be greater than 1 and the default t-short size
   // estimator is used we will tell LHS to extend async-done intervals as much
   // as possible to start collectives as early as possible.
-  if (config.parallel_collective_overlap_limit > 1 &&
-      typeid(*estimator) == typeid(GpuLatencyEstimator)) {
+  if (config.parallel_collective_overlap_limit > 1) {
     config.prioritize_compute_over_async_start = true;
   }
   auto async_tracker = std::make_unique<GpuAsyncTracker>(config);
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -582,7 +582,13 @@ absl::Status RunLatencyHidingSchedulerPasses(
     pipeline.AddPass<PGLEAccuracyChecker>(
         dynamic_cast<ProfileGuidedLatencyEstimator&>(*estimator));
   }
-
+  // If overlap limit is set to be greater than 1 and the default t-short size
+  // estimator is used we will tell LHS to extend async-done intervals as much
+  // as possible to start collectives as early as possible.
+  if (config.parallel_collective_overlap_limit > 1 &&
+      typeid(*estimator) == typeid(GpuLatencyEstimator)) {
+    config.prioritize_compute_over_async_start = true;
+  }
   auto async_tracker = std::make_unique<GpuAsyncTracker>(config);
 
   std::shared_ptr<const SchedulingContext> scheduling_context =
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -585,8 +585,9 @@ absl::Status RunLatencyHidingSchedulerPasses(
   // If overlap limit is set to be greater than 1 and the default t-short size
   // estimator is used we will tell LHS to extend async-done intervals as much
   // as possible to start collectives as early as possible.
+  bool prioritize_compute_over_async_start;
   if (config.parallel_collective_overlap_limit > 1) {
-    config.prioritize_compute_over_async_start = true;
+    prioritize_compute_over_async_start = true;
   }
   auto async_tracker = std::make_unique<GpuAsyncTracker>(config);
 
@@ -594,10 +595,55 @@ absl::Status RunLatencyHidingSchedulerPasses(
       std::make_shared<const SchedulingContext>(
           module, std::move(estimator), std::move(async_tracker), alias_info,
           shape_size_in_bytes);
+
+  auto gpu_early_scheduling_rule =
+      [&prioritize_compute_over_async_start, &config](
+          DefaultSchedulerCore::ScheduleCandidate& a,
+          DefaultSchedulerCore::ScheduleCandidate& b)
+      -> std::optional<DefaultSchedulerCore::CandidateResult> {
+    if (config.aggressive_scheduling_policies &&
+        prioritize_compute_over_async_start) {
+      HloGraphNode* a_node = a.node;
+      HloGraphNode* b_node = b.node;
+      // If either one is an asyncDone, we proceed to the subsequent rules in
+      // the base LHS.
+      if (a_node->DoesOccupyAnyResource() || b_node->DoesOccupyAnyResource()) {
+        return std::nullopt;
+      }
+
+      bool a_has_async_resource =
+          a_node->DoesReleaseAnyResource() && !a.resource_constrained;
+      bool b_has_async_resource =
+          b_node->DoesReleaseAnyResource() && !b.resource_constrained;
+
+      if (!a_has_async_resource && b_has_async_resource) {
+        return DefaultSchedulerCore::CandidateResult{
+            a, "kDelayAsyncStartForCompute"};
+      } else if (a_has_async_resource && !b_has_async_resource) {
+        return DefaultSchedulerCore::CandidateResult{
+            b, "kDelayAsyncStartForCompute"};
+      }
+      if (a_has_async_resource && b_has_async_resource) {
+        // If 2 nodes are both async nodes, we prioritize the one
+        // with more depth to free up more computes to overlap
+        // with the one with less depth which can be launched
+        // early
+        if (a_node->GetDepth() > b_node->GetDepth()) {
+          return DefaultSchedulerCore::CandidateResult{
+              a, "kDelayAsyncStartForDepth"};
+        } else if (b_node->GetDepth() > a_node->GetDepth()) {
+          return DefaultSchedulerCore::CandidateResult{
+              b, "kDelayAsyncStartForDepth"};
+        }
+      }
+    }
+    return std::nullopt;
+  };
+
   auto scheduler_core = std::make_unique<DefaultSchedulerCore>(
       scheduling_context, config,
       /*target_scheduling_rule=*/nullptr,
-      /*early_target_scheduling_rule=*/nullptr,
+      /*early_target_scheduling_rule=*/gpu_early_scheduling_rule,
       /*post_processing_fn=*/nullptr);
 
   pipeline.AddPass<LatencyHidingScheduler>(scheduling_context,
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -585,7 +585,7 @@ absl::Status RunLatencyHidingSchedulerPasses(
   // If overlap limit is set to be greater than 1 and the default t-short size
   // estimator is used we will tell LHS to extend async-done intervals as much
   // as possible to start collectives as early as possible.
-  bool prioritize_compute_over_async_start;
+  bool prioritize_compute_over_async_start = false;
   if (config.parallel_collective_overlap_limit > 1) {
     prioritize_compute_over_async_start = true;
   }
--- a/xla/service/gpu/gpu_latency_hiding_scheduler_test.cc
+++ b/xla/service/gpu/gpu_latency_hiding_scheduler_test.cc
@@ -1087,8 +1087,8 @@ TEST_F(GpuLatencyHidingSchedulerBaseTest, ParallelThreadsShouldBeScheduled) {
 }
 
 TEST_F(GpuLatencyHidingSchedulerBaseTest,
-  MultipleParallelAsyncsExtendedOverAllComputes) {
-absl::string_view kHloModule = R"(
+       MultipleParallelAsyncsExtendedOverAllComputes) {
+  absl::string_view kHloModule = R"(
 HloModule m
 reduce {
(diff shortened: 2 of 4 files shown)


xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -584,7 +584,13 @@ absl::Status RunLatencyHidingSchedulerPasses(
     pipeline.AddPass<PGLEAccuracyChecker>(
         dynamic_cast<ProfileGuidedLatencyEstimator&>(*estimator));
   }
-
+  // If overlap limit is set to be greater than 1 and the default t-short size
+  // estimator is used we will tell LHS to extend async-done intervals as much
+  // as possible to start collectives as early as possible.
+  if (config.parallel_collective_overlap_limit > 1 &&
+      typeid(*estimator) == typeid(GpuLatencyEstimator)) {
```
frgossen: Using `typeid` seems wrong. Should this be a virtual member function instead?

frgossen (changes requested):

Can you provide speedups you measure on one of the HLO benchmarks in `compiler/xla/tools/benchmarks/hlo/`? 

Specifically for this PR: "force delay scheduling start to extend overlap interval #26196". Isn't this something the latency-hiding scheduler aims to do already? If it does not produce sufficient overlap, can we try to improve the cost model instead of forcing scheduling heuristics?

Tixxx:
> Following up on these PRs based on the discussions we had offline in our regular meetings. For this PR, we need to understand and track the performance implications.
> 
> Some general asks for all PRs (that also apply to this PR):
> 
> * What workload is this change motivated by and why is it important?
> * How can we measure and track the performance delta for this PR?
>   Can you provide speedups you measure on one of the HLO benchmarks in `compiler/xla/tools/benchmarks/hlo/`?
> * Can you add or point to an execution test that exercises this code path?
> 
> Specifically for this PR: "force delay scheduling start to extend overlap interval #26196". Isn't this something the latency-hiding scheduler aims to do already? If it does not produce sufficient overlap, can we try to improve the cost model instead of forcing scheduling heuristics?

To answer the questions,
* What workload is this change motivated by and why is it important?

This is motivated by a customer's model, the customer doesn't use PGLE due to compilation overhead, the analytical model also doesn't give the best perf on their cluster. We discovered that using a large overlap limit with this new heuristics produces the best out of the box result even with the default t-short size cost model in xla.

* How can we measure and track the performance delta for this PR? & Can you provide speedups you measure on one of the HLO benchmarks in `compiler/xla/tools/benchmarks/hlo/`?

I will work on getting perf numbers using the benchmarking HLOs

* Can you add or point to an execution test that exercises this code path?

The pr has an LHS unit test that exercises this path? Are you looking for something like a collective e2e test?

>Specifically for this PR: "force delay scheduling start to extend overlap interval https://github.com/openxla/xla/pull/26196". Isn't >this something the latency-hiding scheduler aims to do already? If it does not produce sufficient overlap, can we try to improve >the cost model instead of forcing scheduling heuristics?
Latency hiding scheduler only does it when the async depth is 0, we are enabling this logic optionally for other cases. The goal is to provide out of the box perf without using sophisticated cost models as those are not enabled by default and each has its own drawbacks.

frgossen:
> The pr has an LHS unit test that exercises this path? Are you looking for something like a collective e2e test?

Yes, I think an execution test would be great. 

> I will work on getting perf numbers using the benchmarking HLOs

Thank you! Let's merge this when we have these numbers

xla/service/gpu/gpu_latency_hiding_scheduler_test.cc
```diff
@@ -960,5 +960,81 @@ ENTRY main {
...
+
+  TF_EXPECT_OK(ScheduleModule(module.get(), /*num_parallel_resources=*/16));
+  auto schedule = module->schedule();
+  std::vector<HloInstruction*> instruction_sequence =
+      schedule.sequence(module->entry_computation()).instructions();
+  // With a lot of parallel resources and default latency estimator,
+  // LHS will try to extend all asyncs as much as possible.
+  // We expect all computes to be wrapped within all async start-done
+  // intervals.
+  EXPECT_TRUE(GetIndexByName(instruction_sequence, "add_0") >
```
felixwqp, line 1148: I pulled in this unit test into the internal system and run it without adding the change in lhs source code

```
schedule: HloSchedule
computation reduce:
  y
  x
  _
computation async_wrapped:
  async_param
  reduce-scatter
computation main:
  p1
  rs_0
  p6
  ar_2
  p0
  ar_0
  p2
  add_0
  p4
  p3
  div_0
  p5
  mul_0
  rs_1
  ar_3
  ar_1
  _.1
```

Looks  all computes have been included without lhs change? 

is this unit test expected to pass without lhs source change?

Tixxx: ah i think something changed in LHS earlier to not surface the scheduling issue. I have updated the unit test to be more specific.

felixwqp: I see. Thank you!

felixwqp:
> 📊 Benchmark (for Performance Improvements)
>  hlo_llama31_8b_bf16_1x8.hlo -> 3-5% speedup with this change on h100

can you also share the e2e wall time results for the rest of the benchmarks?

Tixxx:
> > 📊 Benchmark (for Performance Improvements)
> > hlo_llama31_8b_bf16_1x8.hlo -> 3-5% speedup with this change on h100
> 
> can you also share the e2e wall time results for the rest of the benchmarks?

As discussed offline, i put more benchmarking numbers of different configs in the desctiption.

felixwqp (approved):

Tixxx:
Hi @felixwqp , can we check why this was rolled back? Thanks

felixwqp:
@Tixxx  Hi Tj, as this may impact TPU logic, we would need TPU scheduling experts to provide some insights. 

@seherellis  may share some guidance on this CL.

Tixxx:
Thanks. The main logic is guarded by an option which is only enabled for gpu though.

Tixxx:
hi @seherellis  any comments?

seherellis:
Looking into this today. Will share my comments soon.

seherellis:
Thanks for waiting and sorry for the back and forth with the rollback.

So the reason why the compute ops are overlapped in the test is the `kDelayAsyncStartForDepth` rule. When `kDelayAsyncStartForCompute` is commented out, the outcome is the same.

Since this rule is very similar to the existing `kStartAtZeroDepth` rule, I'd like to extend that one instead of adding new rules. 

Could you please 
- not add any new rules,
- but modify the `kStartAtZeroDepth` rule so that it takes the depth information into account when both a and b have 0 async depth?

seherellis:
Other options would be
- defining a TargetSchedulingRule or 
- setting a Priority based on the depths.

Tixxx:
> Thanks for waiting and sorry for the back and forth with the rollback.
> 
> So the reason why the compute ops are overlapped in the test is the `kDelayAsyncStartForDepth` rule. When `kDelayAsyncStartForCompute` is commented out, the outcome is the same.
> 
> Since this rule is very similar to the existing `kStartAtZeroDepth` rule, I'd like to extend that one instead of adding new rules.
> 
> Could you please
> 
> * not add any new rules,
> * but modify the `kStartAtZeroDepth` rule so that it takes the depth information into account when both a and b have 0 async depth?

kDelayAsyncStartForCompute is something we find that can help on gpu in general when overlap limit is set to a larger number. It's beneficial to just delay scheduling async start so all the collectives can start as early as possible while not going over memory limit.
I tried the early target scheduling rule but it's too early. The preferable location would be after kscheduleDone.
If i introduce a hook after kScheduleDone ie after_schedule_done_target_scheduling_rule and move these new rules there, would it be okay?

seherellis:
Sounds like too much of a fine tuning for a specific use case.  See my other comment.

xla/service/latency_hiding_scheduler.cc
```diff
@@ -1348,6 +1371,16 @@ class ReadySetLt {
                    AsyncDepth0CandidateCondition(b, bn), "kStartAtZeroDepth");
```
seherellis, line 1426: Can you modify this function so that it takes the depth information into account when both a and b have 0 async depth?

seherellis (changes requested):

xla/service/latency_hiding_scheduler.cc
Tixxx: kDelayAsyncStartForDepth is targeting to launch async collectives that have less depth only when the overlap limit is set to > 1. I dont think it fits in this rule.

Tixxx:
> Sounds like too much of a fine tuning for a specific use case. See my other comment.

ReadySetLt being the core logic of LHS is really missing hooks for fine-tuning. We have found a few cases where we need to create downstream passes to fine-tune what LHS decides. It's working well in general but to squeeze out a few percentage of speedups, it should have a better interface for target-specific fine-tuning. Ideally backend should be able to override the comparator function and compose series of rules that work the best for the target.

xla/service/latency_hiding_scheduler.cc
seherellis: Is it not okay to have the rule I mentioned when the limit is 1? I don't see the concern here.

Based on the test case in this PR, it should work..

Can you demonstrate what you mean using an example?

Tixxx: I have been out of office for the past a couple of weeks, just came back.
With the limit set to 1, it's just not beneficial to apply these two new rules since any subsequent collectives will be blocked by async resource to launch. Also kStartAtZeroDepth only kicks in when candidates have 0 async depth. We'd like this rule to be applied to candidates that dont necessarily have 0 async depth. The test i added was for simplicity that they happened to have 0 async depth. In reality, this might not be the case.
We will still need to find a place to apply the kDelayAsyncStartForCompute rule though. i can put that in **early_target_scheduling_rule_**  for now to avoid any impact on other backends.

Tixxx: @seherellis  ping on this

seherellis: Sorry I missed the earlier comment from 3 weeks ago. I understand you want to skip kLessStall, but note that the new rules are overriding every other rule below that point, which is too much of a change. Adding config/flag-controlled changes in XLA is becoming more restricted. 

Could you add both rules in early_target_scheduling_rule from the GPU side?

Tixxx: Thanks for taking a look. I have moved the logic to early target scheduling rule. please take another look

pifon2a (changes requested):
It looks like this PR interferes with the logic for TPU. I would prefer not to merge it.

Tixxx:
> It looks like this PR interferes with the logic for TPU. I would prefer not to merge it.

Can you be more specific on which part interferes with logic for TPU? the new rules are injected through pre-defined hook from the gpu compiler.

xla/service/latency_hiding_scheduler.cc
seherellis: I will take a look today. Thanks for the patience.

xla/service/latency_hiding_scheduler.cc
```diff
@@ -1534,14 +1540,12 @@ class ReadySetLt {
...
-    if (cand.has_resource_constrained) {
-      return cand.resource_constrained;
-    }
+
+  void UpdateCandidateResourceConstrained(
+      DefaultSchedulerCore::ScheduleCandidate& cand,
+      const HloGraphNode* cand_node) const {
     if (cand_node->GetResources().empty()) {
       cand.set_resource_constrained(false);
-      return cand.resource_constrained;
```
seherellis: I think you should still return here. Otherwise this if block is redundant since you set it to false two lines below anyways.

seherellis (changes requested):

xla/service/latency_hiding_scheduler.cc
Tixxx: it a typo on my end, thanks for catching it! updated

seherellis (approved):
Thanks! Approving to create an internal code change to run our internal tests.

xla/service/latency_hiding_scheduler.cc
```diff
@@ -1552,10 +1557,14 @@ class ReadySetLt {
           max_it->second == 0 &&
           res_it != sched_state_.resource_users_in_queue.end() &&
           res_it->second > 0);
-      if (cand.resource_constrained) {
-        return cand.resource_constrained;
```
seherellis: We had some internal failure and it turns out we need this return too. It's because after we identify a resource which makes the node resource-constrained, we don't want to keep iterating over other resources and potentially set it back to "not resource constrained".

seherellis (changes requested):

xla/service/latency_hiding_scheduler.cc
Tixxx: i see that makes sense, i added back the early exit

seherellis (approved):
Approving to create an internal code change again.

golechwierowicz:
@Tixxx , can you take a look at ARM build failure? Thanks!

Tixxx:
> @Tixxx , can you take a look at ARM build failure? Thanks!

@golechwierowicz  thanks, Looks like some recent change caused it, i had to rebase to repro the error. It's fixed now.

seherellis (approved):

PR #26247, opened 2025-05-08
[XLA:GPU] Enable composite representation of block scaled dot, enable autotuning

sergey-kozub:
With the deprecation of custom calls in StableHLO, JAX will start lowering the block scaled dot operation as a composite op instead of a custom call. The custom call is still supported, but might be removed once JAX changes the lowering.

The cuDNN fusion compiler now supports dequantization nodes in the cuDNN graphs, so the custom implementation is now removed from "cuda_dnn.cc". The underlying tile propagation analyzer requires the `--xla_gpu_experimental_enable_subchannel_dequantisation_fusion` flag to be set (it will be enabled by default in a follow-up PR).

This PR also updates the autotuner to pick the more performant kernel (cuDNN or cuBLAS with dequantization performed separately). Triton lowering using tt.ScaledDotOp will be added in a follow-up PR (WIP).

The implementation converts the block scaled dot composite into a cuDNN fusion early in the compilation pipeline, this way it's not affected by most transformation passes. We want the fusion body to stay intact, as the fusion compiler uses a pattern matcher to produce dequantization nodes (one cuDNN graph node is represented by multiple HLO ops).


mooskagh:
Why would we want to create fusions so early, before layout assignment, dot normalizations and even SPMD passes?

In general we are aiming to have one place where fusions are created (e.g. PriorityFusion) and I feel this works in the opposite direction.

sergey-kozub:
> Why would we want to create fusions so early, before layout assignment, dot normalizations and even SPMD passes?
> 
> In general we are aiming to have one place where fusions are created (e.g. PriorityFusion) and I feel this works in the opposite direction.

We really want the composite body to be unaffected by the other passes - the fusion compiler needs to pattern-match "2x DQ -> dot" (which is ultimately a single PTX instruction on Blackwell).

One option would be to update the calls to the "CallInliner" pass with "composites_to_preserve" attribute, but this doesn't really help, as some passes (one example is LayoutAssignment) may still move instructions out of the composite body.

Creating a fusion early preserves the HLO graph (as the passes typically don't affect fusions), which is our intent here.

mooskagh:
Composites were meant to make HLO extensions more elegant, but it seems that using them is not very smooth yet, and workarounds arguably make the pipeline more ugly. Ideally, it should not be allowed for passes to modify the composite body by default (i.e. essentially composite should be a custom call with the fallback lowering attached).

Anyway, we've discussed this internally, and it seems that the best option for now is to rewrite the composite back to the custom-call in HLO, so that it's survives until the lowering intact. Would that work?

sergey-kozub:
> Composites were meant to make HLO extensions more elegant, but it seems that using them is not very smooth yet, and workarounds arguably make the pipeline more ugly. Ideally, it should not be allowed for passes to modify the composite body by default (i.e. essentially composite should be a custom call with the fallback lowering attached).
> 
> Anyway, we've discussed this internally, and it seems that the best option for now is to rewrite the composite back to the custom-call in HLO, so that it's survives until the lowering intact. Would that work?

The actual fusion around the block scaled dot can have more instructions, for example:
- transpose the input(s) are fused;
- the scaling factors are pre-padded, i.e. slice operations are fused.
Both these cases are supported by the cuDNN kernel.

So, rewriting back to a custom call would mean somehow attaching the HLO composite body, which defeats the purpose.
What issues do you think may arise if a fusion is added early in the compilation pipeline? I've seen no issues during the testing.

jprabhas:
> Ideally, it should not be allowed for passes to modify the composite body by default (i.e. essentially composite should be a custom call with the fallback lowering attached).

It would be great to have a discussion/write up on what could be done to improve the composite op experience so the graph is not mutated by other passes. @reichlfl for viz

mooskagh:
We are (very early) in the process of defining a consistent set of HLO graph requirements/restrictions at various stages of the pipeline. For example, we'd like to "formally" define whether and at which point we normalize dots (to have dimenions in certain order, or e.g. limiting number of contracting dimensions), whether to normalize singleton dimensions, when to allow non-normalized layouts etc.

One of such consistency checks would be disallowing fusions (and adding HloVerifyer rule) early in the pipeline. Ideally we'd like to have a single pass which would cut graph into fusions.

---

I'm not sure I understand how exactly the fusion would be used for cuDNN rewrite (i.e. in the current state of the PR)

- Composite gets rewritten into fusions
- Standard XLA fusions passes add more ops into the fusion
- Then the fusion gets rewritten into the cuDNN call
- But what happens if added ops cannot be expressed with the cuDNN call?

sergey-kozub:
> One of such consistency checks would be disallowing fusions (and adding HloVerifyer rule) early in the pipeline. Ideally we'd like to have a single pass which would cut graph into fusions.

I was not aware of that effort.
What about another idea: add a method to HloModule named "MakeNonfusionNoncompositeComputations" (in addition to "MakeNonfusionComputations"), and update all callers to use the new one?
This way the passes will not break the composites.

> I'm not sure I understand how exactly the fusion would be used for cuDNN rewrite (i.e. in the current state of the PR)
> 
> * Composite gets rewritten into fusions
> * Standard XLA fusions passes add more ops into the fusion

Standard XLA passes avoid changing fusions, so it remains intact.

sergey-kozub:
Would this work? https://github.com/openxla/xla/pull/27476

This makes all passes effectively ignore composite computations (same as fusions).
Also, do not inline composites by default.

sergey-kozub:
We need to agree on how the composite calls should be handled by XLA in general. I'll write a design doc to discuss.

PR #26358, opened 2025-05-12
Support optimization barrier in HLO evaluator

sergey-kozub:


akuegel (approved):

akuegel:
Something I missed when reviewing this: with this change, constant-folding will now look through optimization barriers, breaking the semantics. Therefore this change is being rolled back now. What was the goal of your change?

sergey-kozub:
The original goal was to be able to run a test which contains opt-barrier opcodes (to avoid float normalization).
However, I can also just disable the float normalization pass for that test, so we can drop this PR (not evaluating opt-barrier turns to be a feature, not a bug).

PR #26387, opened 2025-05-12
Fix module dumping functions

shraiysh:
There are two dumping functions: with and without buffer assignment. Both of them are expected to dump non-default debug options and return a list of paths of the files they dumped. This patch ensures that both adhere to this expectation.

Additionally, hlo-opt with xla_dump_to flag does not print files like `*.before_optimization.txt`, `*.after_optimizations.txt`, `*-buffer-assignment.txt`, `*.debug_options`, or other files that are dumped by other HLO tools. This patch adds that to the hlo-opt tool.


xla/service/dump.cc
```diff
@@ -904,25 +894,43 @@ std::string GetNonDefaultDebugOptions(const DebugOptions& debug_options) {
...
-  DumpToFileInDir(debug_options, filename, nonDefaultDebugOptions);
+  std::string filename = FilenameFor(module, "", suffix);
+  std::string nonDefaultDebugOptions = GetNonDefaultDebugOptions(debug_options);
+  return DumpToFileInDirImpl(filename, nonDefaultDebugOptions,
+                             CanonicalDebugOptions(debug_options));
 }
 
-std::vector<std::string> DumpHloModuleIfEnabled(
-    const HloModule& module, const BufferAssignment& buffer_assn,
+std::vector<std::string> DumpHloModuleIfEnabledInternal(
```
reedwm, line 906: Put this function in the anonymous namespace. Also, replace "Internal" with "Impl" for consistency with other functions in the file.

reedwm (changes requested):
@abhigunj can you review hlo_opt changes? Other than my one comment, the non-hlo_opt changes look good to me.

abhigunj (approved):

abhigunj:
Thanks for enabling the dump at the `hlo-opt`.

reedwm (approved):

reedwm (approved):

PR #26443, opened 2025-05-13
[NVIDIA GPU] Add support for nccl symmetric kernel

Tixxx:
This pr adds support for nccl symmetric kernels that are coming in 2.27.
To use symmetric kernels, buffers need to allocated using nccl allocator and then register using new registration API ncclCommWindowRegister.
ncclCommWindowRegister is a collective call and needs to be wrapped using nccl group sematics when running in 1process per node setup.
For small message allreduce, this will provide 30%-40% speedup vs non-symmetric kernels.


xla/service/gpu/compile_module_to_llvm_ir.cc
```diff
@@ -254,9 +254,12 @@ absl::StatusOr<std::unique_ptr<BufferAssignment>> RunBufferAssignment(
   const DebugOptions& options = module->config().debug_options();
   BufferAssigner::Colorer colorer =
       (options.xla_gpu_enable_nccl_user_buffers() ||
-       options.xla_gpu_experimental_enable_nvshmem())
-          ? CollectiveColorer(options.xla_gpu_enable_nccl_user_buffers(),
-                              options.xla_gpu_experimental_enable_nvshmem())
+       options.xla_gpu_experimental_enable_nvshmem() ||
+       options.xla_gpu_enable_nccl_symmetric_buffers())
+          ? CollectiveColorer(
+                options.xla_gpu_enable_nccl_user_buffers() ||
```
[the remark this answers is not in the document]
Tixxx: symmetric memory currently has a limitation where it wont work if a group call has more than 1 ops. This limitation will be lifted in 2.28. Before that, we'd need to keep the user buffer flag for now to utilize nvls optimizations in group calls. The goal is to get rid of legacy memory registration and only use symmetric in xla down the road.

xla/backends/gpu/collectives/nccl_communicator.cc
```diff
@@ -315,30 +343,58 @@ absl::StatusOr<size_t> NcclCommunicator::NumRanks() const {
 }
 
 absl::StatusOr<std::unique_ptr<Communicator::RegisteredBufferHandle>>
-NcclCommunicator::RegisterBuffer(stream_executor::DeviceMemoryBase buffer) {
+NcclCommunicator::RegisterBuffer(stream_executor::DeviceMemoryBase buffer,
```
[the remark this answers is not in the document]
jprabhas: And where do we handle registration for Nvshmem buffers?

PatriosTheGreat:
Hey sorry I was out for a few weeks, will take a look at this PR today/tomorrow.

xla/xla.proto
```diff
@@ -450,6 +450,9 @@ message DebugOptions {
   // Enable NCCL user buffers.
   bool xla_gpu_enable_nccl_user_buffers = 267;
 
+  // Enable NCCL symmetric buffers.
+  bool xla_gpu_enable_nccl_symmetric_buffers = 389;
```
PatriosTheGreat: This id is already taken, Please also increase next id: 390

xla/backends/gpu/runtime/collective_thunk.cc
```diff
@@ -356,39 +359,52 @@ absl::Status RegisterBufferOnce(GpuCollectives* collectives,
...
+                                          buffer.opaque()})) {
+      need_reg = true;
+    } else {
+      VLOG(5) << "Buffer: " << buffer.opaque()
+              << " with size: " << buffer.size()
+              << " and base pointer: " << base_buffer.opaque()
+              << " is already registered.";
+    }
+  }
+  if (need_reg) {
```
PatriosTheGreat, line 386: This looks like a race condition. Since the mutex is released at this point two threads can reach this point at the same time and register/add the same pointer twice. Why can't RegisterBuffer be done under the mutex?

xla/backends/gpu/runtime/collective_thunk.h
```diff
@@ -66,6 +66,8 @@ struct CollectiveConfig {
   void SetCollectiveOpKindAndID(const HloCollectivePermuteInstruction* instr);
   void SetCollectiveOpKindAndID(const HloSendRecvInstruction* instr);
   bool IsDegenerate(int64_t replica_count, int64_t partition_count) const;
+
+  bool use_symmetric_buffer;
```
PatriosTheGreat: Could you move this up to all other fields? Or move other fields here

PatriosTheGreat (changes requested):

xla/backends/gpu/runtime/collective_thunk.cc
Tixxx: symmetric registration is now a collective call, this lock is a process-wide lock so wrapping the collective call with a global lock leads to deadlocks as observed in many internal benchmarks.
I changed it to a thread local boolean with a check-check-set pattern. The first check is mutex-protected so for a given device ordinal and buffer, there could only be 1 thread that has need_reg set to true, then we call the register(could be a collective here), the last set will also be mutex protected.

PatriosTheGreat (approved):

Tixxx:
@PatriosTheGreat @frgossen  hey guys, wanted to check in on the status of this pr. Let me know if there are any issues you'd like to discuss.

PatriosTheGreat (approved):

xla/xla.proto
```diff
@@ -494,6 +494,9 @@ message DebugOptions {
   // Enable NCCL user buffers.
   optional bool xla_gpu_enable_nccl_user_buffers = 267;
 
+  // Enable NCCL symmetric buffers.
+  bool xla_gpu_enable_nccl_symmetric_buffers = 403;
```
PatriosTheGreat: Could you please add "optional" attribute?

Tixxx: added optional

PatriosTheGreat (approved):

PR #26445, opened 2025-05-13
[NFC] Clean up StreamKind and StreamId from GpuCliqueKey

chaserileyroberts:
In this PR, we completely remove `StreamKind` and `StreamId` types from the `GpuCliqueKey` class. While we only need `StreamKind` gone, I also took out `StreamId` since it doesn't seem to be used at all. 

This PR is intended to be a revival of https://github.com/openxla/xla/pull/22450, with the end goal of completely removing `AsyncStreamKind` entirely. There will be 2-3 more follow up PRs to achieve this without breaking anything.  

The biggest new change was adding a `is_p2p` boolean to `GpuCliqueKey`. The `is_p2p` is necessary since `resource_requests` depend upon whether the clique should be using p2p comms or not when setting the maxCTA config.

### DIFF
--- a/xla/backends/gpu/collectives/gpu_clique_key.cc
+++ b/xla/backends/gpu/collectives/gpu_clique_key.cc
@@ -129,12 +129,12 @@ std::string GpuCliqueKey::ToString() const {
 }
 
 void GpuCliqueKey::HashValue(absl::HashState state) const {
-  absl::HashState::combine(std::move(state), devices(), is_p2p_,
-                           participant_groups_, root_device_, incarnations_);
+  absl::HashState::combine(std::move(state), devices(), participant_groups_,
+                           root_device_, incarnations_);
 }
 
 bool operator==(const GpuCliqueKey& a, const GpuCliqueKey& b) {
-  return a.devices() == b.devices() && a.is_p2p_ == b.is_p2p_ &&
+  return a.devices() == b.devices() &&
          a.participant_groups_ == b.participant_groups_ &&
          a.num_local_participants_ == b.num_local_participants_ &&
          a.root_device_ == b.root_device_ && a.incarnations_ == b.incarnations_;
@@ -143,8 +143,7 @@ bool operator==(const GpuCliqueKey& a, const GpuCliqueKey& b) {
 // Constructs a tuple from the clique key for comparison purposes.
 static auto CmpKey(const GpuCliqueKey& key) {
   return std::make_tuple(key.devices().size(), key.devices(), key.root_device(),
-                         key.num_local_participants(), key.is_p2p(),
-                         key.incarnations());
+                         key.num_local_participants(), key.incarnations());
 }
 
 bool operator<(const GpuCliqueKey& a, const GpuCliqueKey& b) {
--- a/xla/backends/gpu/collectives/gpu_clique_key.cc
+++ b/xla/backends/gpu/collectives/gpu_clique_key.cc
@@ -35,6 +35,10 @@ limitations under the License.
 
 namespace xla::gpu {
 
+bool IsP2PStreamKind(AsyncStreamKind stream_kind) {
+  return stream_kind != AsyncStreamKind::kCollective;
+}
+
 CollectiveStreamId GetCollectiveStreamId(bool is_async,
                                          AsyncStreamKind stream_kind) {
   // TODO(ezhulenev): This implementation does not look correct as stream IDs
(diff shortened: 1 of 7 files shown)


frgossen:
AsyncStreamKind is needed in pipeline parallelism atm and for memcpy-based collectives (?). What do you want to use instead to distinguish between different collectives streams?

chaserileyroberts:
> AsyncStreamKind is needed in pipeline parallelism atm and for memcpy-based collectives (?). 

The end goal would still allow for these to still be scheduled on different streams, but to simplify the stream infrastructure. Rather than having distinct sets of streams for collectives vs compute, as they are today in `Thunk::CollectiveExecuteParams` and `Thunk::ExecuteParams`, they should just pull from one set of streams, simplifying the codebase and allowing the new stream annotation features to behave more predictably with them.

> What do you want to use instead to distinguish between different collectives streams?

Right now, the plan is to set `execution_stream_id` for the respective collective thunks, and have them dispatch on the corresponding compute stream. For things like pipeline parallelism and memcpy, we should be able to translate their old `AsyncStreamKind` ids to an equivalent `execution_stream_id` without changing core functionality. 

This change in the long run would have two main benefits: 

1) It would allow for multiple communication ops to run at once, allowing for better utilization of independent NVLink or Infiniband hardware (Think all-gathering on NVLink while simultaneously doing a reduce-scatter on Infiniband). This was the original goal that @terryysun had in #22450  

2) When GSPMD needs to inserts a communication op in a `compute_on("gpu_stream:#")` computation, that op currently doesn't respect the stream annotation and can cause timing corruptions. (This is mainly why I am helping here).

frgossen:
Thanks for explaining. I think there was a design doc for this, right? Do you mind linking it here?

chaserileyroberts:
For sure, here is the first doc on the multi-stream async collectives. 

https://docs.google.com/document/d/1Nxwr9j92yHx--ryBqFkQk2lW-Jx6_2ejjVD7XponOvI/edit?usp=sharing

We have another doc around the AsyncStreamKind cleanup from @terryysun , but I'm just now realizing that it was never shared externally. He'll get it approved to share first then I'll link it here.

terryysun:
> Thanks for explaining. I think there was a design doc for this, right? Do you mind linking it here?

Here's the link to the doc on stream assignment https://docs.google.com/document/d/1i1IzE-ycaWmWYN_0B-lxDwk2WO4CRs9huxVz0YAGgMU/edit?tab=t.0.

xla/backends/gpu/collectives/gpu_clique_key.cc
```diff
@@ -158,9 +160,7 @@ bool operator>(const GpuCliqueKey& a, const GpuCliqueKey& b) {
   if (a.num_local_participants_ > b.num_local_participants_) return true;
   if (b.num_local_participants_ > a.num_local_participants_) return false;
 
-  // We still use `<` to order by stream id as we want to acquire sync cliques
-  // before async ones.
-  return a.stream_id_.value() < b.stream_id_.value();
+  return a.is_p2p_ > b.is_p2p_;
```
frgossen: `is_p2p_` is a bool, no?

xla/backends/gpu/collectives/gpu_clique_key.h
```diff
@@ -84,7 +88,6 @@ class GpuCliqueKey : public CliqueKey {
 
   // Returns the stream kind for this clique key, stream kind will be used to
   // specify what configuration to pass for each type of operation.
-  AsyncStreamKind stream_kind() const { return stream_kind_; }
```
frgossen, line 89: Remove the function w/o the comment?

frgossen:
Looks like it fails to build in the CI. Does it build for you?

frgossen (changes requested):

xla/backends/gpu/collectives/gpu_clique_key.cc
chaserileyroberts: This code was removed when I fixed the merge conflicts.

xla/backends/gpu/collectives/gpu_clique_key.h
chaserileyroberts: Done

chaserileyroberts:
This is ready for a final pass.

frgossen (approved):

golechwierowicz:
This test fails: `collective_ops_e2e_test_2gpu`

```
AsyncCollectiveOps/AsyncCollectiveOps.AsyncAllToAllMemCpyWithoutSplitDim/sync
...
Value of: _status_or_value39.status().ok()
  Actual: false
Expected: true
NOT_FOUND: No clique found for clique key: devices=[0,1]; is_p2p=1; groups=[[0,1]]; root_device=-1; num_local_participants=2; incarnations=[]
```

chaserileyroberts:
@golechwierowicz I was able to replicate and fix it locally. The issue was that stream_kind was never included in `operator==` logic, so adding `is_p2p` broke some assumptions it seems.

PR #26633, opened 2025-05-16
[Proposal] Open source some simple custom call annotation handling.

yliu120:
As mentioned in the 3-way meeting between Google, NVIDIA and xAI, we actually want to upstream some infra features we built internally at xAI to avoid accumulating conflicts between us and upstream.

This can be a first attempt to see how much opinion conflicts we would have and how likely we can upstream our things which will benefit open source community as well.

The overall philosophy we have is to make transforms mostly use-controllable in JAX and treat XLA simply as a lowering/decomposing tool. Thus, we have some internally-used custom call passed into XLA majorly for annotating info that the compiler needs to generate optimized graphs.

In this PR, we have two custom calls:
1) AllocateBuffer, which simply exposes the internal XLA custom op to JAX so that JAX can allocate uninitialized buffer. With this custom call, we don't need to use the delegate jnp.zeros or broadcast to allocate tensors, which often times will be CSEed in JAX. This is quite single responsibility and as it is a custom call, there is no magic compiler optimization on itself.
2) AnnotateMemorySpace, this custom call can annotate the operand buffer to be in any memory space the JAX users want, for instance, the persistent temp, host or in the future SMEM. Unlike `jax.device_put`, this custom call doesn't rely on jax.shardings to take place. It can work with any temp buffers inside a JAX program without knowing the sharding.

A very small memory space propagation pass for propagating the annotated memory space in forward direction was added because our transformations in JAX already makes the semantics clear. We don't rely on any magic XLA pattern matching which can likely lead to unexplained compilation failures for feature dev, such as, activation offload.

We also open source the JAX-side use cases in small, non-production related examples to illustrate the power of this change, for instance, the pipelined scan which essentially works like a drop-in replacement of `jax.lax.scan` but automatically offload `saved` residuals to host:

https://github.com/yliu120/jax_examples/blob/main/magics/transforms/pipelined_scan.py
https://github.com/yliu120/jax_examples/blob/main/tests/pipelined_scan_test.py

The entire code that handles complicated pipelining and host offloading only spans 100+ lines in Python. This pattern can also be used for other use cases.

### DIFF
--- a/xla/service/gpu/transforms/memory_space_propagation.cc
+++ b/xla/service/gpu/transforms/memory_space_propagation.cc
@@ -1,0 +1,95 @@
+/* Copyright 2025 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/service/gpu/transforms/memory_space_propagation.h"
+
+#include <optional>
+#include <utility>
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/log/check.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "xla/hlo/analysis/hlo_dataflow_analysis.h"
+#include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_opcode.h"
+#include "xla/layout.h"
+#include "xla/layout_util.h"
+#include "xla/service/hlo_value.h"
+#include "xla/shape.h"
+#include "xla/shape_util.h"
+
+namespace xla::gpu {
+namespace {
+
+bool Propagate(ShapeIndexView index, HloInstruction* instruction,
+               const Shape& src_shape, const HloDataflowAnalysis* analysis) {
+  bool modified = false;
+
+  const HloValue& value =
+      analysis->GetValueDefinedAt(instruction, ShapeIndex(index));
+  auto memory_space_to_propagate = src_shape.layout().memory_space();
+  for (const HloPosition& position : value.positions()) {
+    auto* instruction = position.instruction;
+    Shape* shape = ShapeUtil::GetMutableSubshape(instruction->mutable_shape(),
+                                                 position.index);
+    if (!shape->has_layout() ||
+        shape->layout().memory_space() == memory_space_to_propagate) {
+      continue;
+    }
+    shape->mutable_layout()->set_memory_space(memory_space_to_propagate);
+    modified = true;
+  }
+
+  // We don't propagate from S(x) operand to S(x) output because it's
+  // pure memory transfer without any computation in S(x). Computations
+  // including transposes should be defined explicitly in JAX.
+  return modified;
+}
+
+}  // namespace
+
+absl::StatusOr<bool> MemorySpacePropagation::Run(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  bool modified = false;
+  TF_ASSIGN_OR_RETURN(auto dataflow_analysis,
+                      HloDataflowAnalysis::Run(*module, /*ssa_form=*/false,
+                                               /*bitcast_defines_value=*/true));
+
+  absl::flat_hash_set<HloInstruction*> visited_instructions;
+  for (HloComputation* computation :
+       module->MakeNonfusionComputations(execution_threads)) {
+    for (auto* instruction : computation->MakeInstructionPostOrder()) {
+      if (visited_instructions.contains(instruction)) {
+        continue;
+      }
+      visited_instructions.insert(instruction);
+      ShapeUtil::ForEachLeafShape(
+          instruction->shape(),
+          [&, analysis = dataflow_analysis.get()](const Shape& sub_shape,
+                                                  const ShapeIndex& index) {
+            if (sub_shape.has_layout() &&
+                sub_shape.layout().memory_space() !=
+                    xla::Layout::kDefaultMemorySpace &&
+                analysis->ValueIsDefinedAt(instruction, index)) {
+              modified |= Propagate(index, instruction, sub_shape, analysis);
+            }
+          });
+    }
+  }
+  return modified;
+}
+}  // namespace xla::gpu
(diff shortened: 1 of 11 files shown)


Tixxx:
@jprabhas  for visibility. However we still need google's feedback on this. @frgossen  Can you or folks at google side take a look at this proposal?

frgossen:
This is a limitation that was brought up internally, too. We have a design to address this and @bixia1 is working on this. I will ask her to share an external version of the design with you to see if it addresses all the issues you see. In principle, these changes make sense to me.

xla/service/gpu/transforms/memory_space_propagation.cc
```diff
@@ -0,0 +1,95 @@
...
+        shape->layout().memory_space() == memory_space_to_propagate) {
+      continue;
+    }
+    shape->mutable_layout()->set_memory_space(memory_space_to_propagate);
+    modified = true;
+  }
+
+  // We don't propagate from S(x) operand to S(x) output because it's
+  // pure memory transfer without any computation in S(x). Computations
+  // including transposes should be defined explicitly in JAX.
```
bixia1, line 58: Can you explain this comment? I don't understand how they are relevant to the code here. In this routine, we propagate the memory space from the definition of the value to the use of the value, right?

xla/service/gpu/transforms/memory_space_propagation_test.cc
```diff
@@ -0,0 +1,146 @@
...
+  %cosine.120 = f32[2,2]{1,0} cosine(%get-tuple-element.118)
+  %multiply.121 = f32[2,2]{1,0} multiply(%get-tuple-element.119, %cosine.120)
+  ROOT %tuple.122 = (f32[], f32[2,2]{1,0}) tuple(%reduce.63, %multiply.121)
+}
+  )";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(kHlo));
+  TF_ASSERT_OK_AND_ASSIGN(bool changed,
+                          RunHloPass(MemorySpacePropagation(), module.get()));
+  EXPECT_TRUE(changed);
```
bixia1, line 138: Is it possible to write a FileCheck test to verify the propagation result?

xla/service/gpu/transforms/post_layout_custom_call_rewriter.h
```diff
@@ -0,0 +1,43 @@
...
+#include "absl/container/flat_hash_set.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/hlo/pass/hlo_pass_interface.h"
+
+namespace xla {
+namespace gpu {
+
+// Rewrites custom calls after running layout assignment and normalization.
```
bixia1, line 27: Can you provide some details on the transformation that the rewriter performs?
It set up the memory space in the operand and use the operand to replace the custom call?

xla/service/gpu/transforms/memory_space_propagation.cc
```diff
@@ -0,0 +1,95 @@
...
+        continue;
+      }
+      visited_instructions.insert(instruction);
+      ShapeUtil::ForEachLeafShape(
+          instruction->shape(),
+          [&, analysis = dataflow_analysis.get()](const Shape& sub_shape,
+                                                  const ShapeIndex& index) {
+            if (sub_shape.has_layout() &&
+                sub_shape.layout().memory_space() !=
+                    xla::Layout::kDefaultMemorySpace &&
```
bixia1, line 86: This will propagate all memory space, not only the memory space introduced by PostLayoutCustomCallRewriter.
Can you put a comment here to explain why doing this is safe?

Alternatively, we can also performs PostLayoutCustomCallRewriter and let layout assignment to figure out the propagation. How would you compare your current approach with this approach?

xla/service/gpu/transforms/memory_space_propagation.cc
```diff
@@ -0,0 +1,95 @@
...
+                      HloDataflowAnalysis::Run(*module, /*ssa_form=*/false,
+                                               /*bitcast_defines_value=*/true));
+
+  absl::flat_hash_set<HloInstruction*> visited_instructions;
+  for (HloComputation* computation :
+       module->MakeNonfusionComputations(execution_threads)) {
+    for (auto* instruction : computation->MakeInstructionPostOrder()) {
+      if (visited_instructions.contains(instruction)) {
+        continue;
+      }
```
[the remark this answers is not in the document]
yliu120: Yeah. I guess this was something that I didn't remove when copying this out from the internal repo.
Thanks.

xla/service/gpu/transforms/memory_space_propagation.cc
yliu120: Yes. This is something we use only for host and persistent temp in GPU world.  it will propagate everything with non-default memory space. 

Memory space propagation happens at very late stage. I guess at this point all memory spaces we have here is pinned by users. So if users want it, we just give that.

xla/service/gpu/transforms/post_layout_custom_call_rewriter.h
yliu120: Internally we have a few different custom call here. but this one is just the one that annotates memory space for the output of the instruction.

The annotation is sharding-unaware, which means it decouples from sharding. The user doesn't need to know sharding when annotating memory space.

custom-call = annotate_memory_space(operand)

==> 

operand will have annotated memory space.

xla/service/gpu/transforms/memory_space_propagation.cc
yliu120: I think in principle putting inside layout assignment will work as well. but it is a more complicated solution than this one because you probably need to define some rules based on different type of instructions.

a = copy-start(a, b), a and b can have different memory space. It can be platform-dependent. In GPU, I guess if I want to copy from host memory space or persistent temp to default, this is valid. 

In TPU, copying from HBM to VMEM probably doable. This just potentially increases the complexity of the engineering work without too much benefits.

The philosophy here is to do simple but working things and move more controls to users. The compiler doesn't do too much fancy logic. Here the propagation is actually very simple. For instance, it just propagate to places where users feels hard to annotate, to all the HloPositions but not going too far beyond.

If users want to have the operand or HloUse also in that memory space, they should annotate but not rely on some automatic propagation that decides for users.

xla/service/gpu/transforms/memory_space_propagation.cc
yliu120: I replied in another comment. I just hope this memory space propagation be simple and not placing any unexpected behavior for the users. When user annotate a buffer to be in some memory space, its definition and all references should be annotated but nothing beyond.

xla/service/gpu/transforms/memory_space_propagation_test.cc
yliu120: Sorry I guess this PR is just proposing this idea and also exposing some philosophy on how we want to use XLA compiler in x.AI. If Google thinks we can merge, I will clean up tests but I didn't expect too much.

yliu120: BTW, I heard you are working on a design. so if this is included, we eventually can switch.

PR #26672, opened 2025-05-19
[XLA:GPU] Use fine grain operator update for updating command buffer executor. 

shawnwang18:


xla/backends/gpu/runtime/command_buffer_cmd.h
```diff
@@ -421,6 +425,10 @@ class CommandBufferCmdExecutor {
 
   // Buffer allocations indices referenced by commands in this sequence.
   absl::flat_hash_set<BufferAllocation::Index> allocs_indices_;
+
+  // Map from buffer allocation index to command ids that references it.
+  absl::flat_hash_map<BufferAllocation::Index, absl::flat_hash_set<CommandId>>
```
ezhulenev, line 430: can you instead keep a map:

`absl::flat_hash_map<CommandId, absl::flat_hash_set<BufferAllocation::Index>>`

and use to from command sequence to skip commands that reference buffer allocations that didn't change?

xla/backends/gpu/runtime/command_buffer_thunk.h
```diff
@@ -71,6 +71,10 @@ class CommandBufferThunk : public Thunk {
                                    const Thunk::ExecuteParams& params)
         ABSL_EXCLUSIVE_LOCKS_REQUIRED(mutex);
 
+    void RecordAllocations(const CommandBufferCmdExecutor& commands,
```
ezhulenev, line 74: can you instead of returning bool from `ShouldUpdateCommandBuffer`, return a flat_hash_set of buffer indices that changed. Empty set == no changes. Then this buffer can be forwarded to command buffer sequence.

xla/backends/gpu/runtime/command_buffer_cmd.h
```diff
@@ -202,6 +202,10 @@ class CommandBufferCmd {
     // An external state manager that gives efficient access to per-device state
     // to commands without a need to add expensive synchronization.
     StateManager& state;
+
+    // A reference to the vector of device memory bases that are previously
```
ezhulenev, line 206: Why it's needed at all? Command sequence only needs to know if allocation at given index changed or not, concrete address should not matter. And this information should stay only at CommandBufferCmdSequence level, individual commands should not care about that.

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -411,8 +413,54 @@ absl::Status CommandBufferCmdExecutor::RecordUpdate(
 
   // Keep a state associated with commands in the sequence in the state manager.
   CommandBufferCmd::StateManager& state = record_params.state;
+  const BufferAllocations* allocs = execute_params.buffer_allocations;
```
ezhulenev, line 416: `RecordUpdate` can take absl::flat_hash_set<Index> of indices that changes since last command update, and use that as predicate in the for loop (just a set intersection)

shawnwang18: Fixed

xla/backends/gpu/runtime/command_buffer_cmd.h
shawnwang18: Fixed

xla/backends/gpu/runtime/command_buffer_thunk.h
shawnwang18: Fixed

xla/backends/gpu/runtime/command_buffer_cmd.h
shawnwang18: Fixed

PR #26682, opened 2025-05-19
[ROCM] Fix buildbreak due to invalid loading of the rule

alekstheod:


tools/ci_build/gpu_build/BUILD
```diff
@@ -2,8 +2,6 @@
 # TensorFlow is a computational framework, primarily for use in machine
 # learning applications.
 
-load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
```
beckerhe, line 5: What's the breakage?

`sh_binary` will be removed from core Bazel, therefore we will need this load statement eventually and this is preparing for that.

alekstheod: The reason is that our [rocm] CI job is failing due to the fact we can't load this rule. Which version of bazel will effectively have this change? I assume then our issue is that we use a version of bazel that doesn't have "@rules_shell//shell:sh_binary.bzl? Which version would support it?

alekstheod: http://ml-ci.amd.com:21096/job/OpenXLA/job/XLA-multibranch/job/PR-26726/1/pipeline-console/log?nodeId=276

beckerhe (approved):

PR #26747, opened 2025-05-20
Precompute dynamic memcpy offsets.

jreiffers:
Currently, we compute the dynamic memcpy offsets during Thunk execution. There are a few problems with this:

- It makes the Thunk depend on HLO, which is not supposed to happen
- It's potentially slow, because HloEvaluator is slow
- It requires us to repeat the pattern matching
  - Which is brittle, because the pattern matching can fail due to changes introduced by command buffer scheduling.

This change moves the offset computation logic from copy_thunk to fusion_dynamic_memcpy_rewriter. The offsets are materialized in the backend config.

Note: the new code in xla/service/gpu/transforms/fusion_dynamic_memcpy_rewriter.cc was mostly moved there from copy_thunk.cc.


pifon2a (approved):
Very nice, thank you!

PR #26750, opened 2025-05-20
[ROCm] Delete communicator objects owned by gpu clique to avoid memor…

hsharsha:
…y leaks.

### DIFF
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -146,11 +146,14 @@ static void CheckClique(const GpuCliqueKey& clique_key,
   }
 }
 
+static std::atomic<bool> run_heartbeat{true};
 // TODO(ezhulenev): We need a mechanism to destroy whole clique when one of the
 // communicators is aborted to be able to recover from errors.
 static void GpuCliqueHeartBeatMonitorThread() {
   VLOG(5) << "Starting GPU cliques heart beat monitor";
   while (true) {
+    auto run = run_heartbeat.load();
+    if (!run) break;
     absl::SleepFor(absl::Seconds(30));
     ProcessGpuCliques& cliques = GetProcessGpuCliques();
     absl::MutexLock lock(&cliques.mu);
@@ -162,11 +165,10 @@ static void GpuCliqueHeartBeatMonitorThread() {
   }
 }
 
-static void StartGpuCliqueHeartBeatMonitor() {
-  static auto* monitor_thread = tsl::Env::Default()->StartThread(
+static tsl::Thread* StartGpuCliqueHeartBeatMonitor() {
+  return tsl::Env::Default()->StartThread(
       tsl::ThreadOptions(), "gpu_clique_heart_beat_monitor",
       GpuCliqueHeartBeatMonitorThread);
-  (void)monitor_thread;  // suppress unused variable warning
 }
 
 //===----------------------------------------------------------------------===//
@@ -234,7 +236,7 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
           << rank << "; num_local_participants=" << num_local_participants;
 
   // Start GPU clique heart beat monitor when create a first clique.
-  StartGpuCliqueHeartBeatMonitor();
+  tsl::Thread *heartbeat_thread = StartGpuCliqueHeartBeatMonitor();
 
   using RendezvousArg = std::pair<DeviceRank, /*synchronized=*/bool>;
 
@@ -348,10 +350,15 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
   // processes are not able to synchronize device activity.
   RendezvousArg rendezvous_arg = std::make_pair(device_rank, synchronized);
 
-  return Rendezvous<LockableGpuClique::Lock>(
+  auto ret_val = Rendezvous<LockableGpuClique::Lock>(
       initialization_rendezvous_name, rendezvous_key, rendezvous_arg,
       num_local_participants, initialize, WarnStuckTimeout(),
       TerminateTimeout());
+
+  run_heartbeat.store(false);
+  delete heartbeat_thread;
+
+  return ret_val;
 }
 
 // Computes a unique GPU communicator split color from a clique key. We use a
--- a/xla/backends/gpu/runtime/collective_thunk.cc
+++ b/xla/backends/gpu/runtime/collective_thunk.cc
@@ -490,9 +490,10 @@ absl::Status CollectiveThunk::ExecuteOnStream(const ExecuteParams& params) {
 
             ));
   }
-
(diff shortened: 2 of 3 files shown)


xla/backends/gpu/runtime/collective_thunk.cc
```diff
@@ -490,9 +490,10 @@ absl::Status CollectiveThunk::ExecuteOnStream(const ExecuteParams& params) {
 
             ));
   }
-
-  TF_ASSIGN_OR_RETURN(std::vector<Communicator*> comms, GetCommunicators(params));
-  for (auto comm : comms) { delete comm; }
+  if (!IsAsync()) {
```
hsharsha, line 493: Hi @ezhulenev I am only handling this for synchronous case. How can I pass the required context or CollectiveThunk object to CollectiveThunkDone so as to free comms after async operation is done.

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -348,10 +350,15 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
...
   RendezvousArg rendezvous_arg = std::make_pair(device_rank, synchronized);
 
-  return Rendezvous<LockableGpuClique::Lock>(
+  auto ret_val = Rendezvous<LockableGpuClique::Lock>(
       initialization_rendezvous_name, rendezvous_key, rendezvous_arg,
       num_local_participants, initialize, WarnStuckTimeout(),
       TerminateTimeout());
+
+  run_heartbeat.store(false);
+  delete heartbeat_thread;
```
mwhittaker, line 359: Sorry, but I'm having trouble understanding this change. You are starting the heartbeat thread, creating the clique, and then immediately stopping the heartbeat thread? This change also runs multiple heartbeat threads instead of one?

xla/tests/collective_ops_test.cc
```diff
@@ -386,7 +386,7 @@ XLA_TEST_F(CollectiveOpsTest, AllReduce_AllCombinations) {
   const int64_t kNumElems = 1024;
 
   for (std::vector<int64_t> devices :
-       PowerSetOfIota(std::min(num_devices(), static_cast<int64_t>(4)))) {
+       PowerSetOfIota(std::min(num_devices(), static_cast<int64_t>(10)))) {
```
mwhittaker, line 389: I'm not familiar with this test. Why was this change necessary?

xla/backends/gpu/runtime/collective_thunk.cc
```diff
@@ -490,7 +490,10 @@ absl::Status CollectiveThunk::ExecuteOnStream(const ExecuteParams& params) {
 
             ));
   }
-
+  if (!IsAsync()) {
+    TF_ASSIGN_OR_RETURN(std::vector<Communicator*> comms, GetCommunicators(params));
+    for (auto comm : comms) { delete comm; }
```
mwhittaker, line 495: I don't believe the communicators returned by `GetCommunicators` are owned, so it is an error to delete them.

mwhittaker (commented):
Hi Harsha, thanks for the PR. Would you mind elaborating on the problem you are trying to solve?

xla/backends/gpu/collectives/gpu_cliques.cc
hsharsha: I was running valgrind to check memory leaks in xla and the below test on our beefy system 1 node with 16 GPUs connected to it and found out that due to async threads and a chain of ownership movement of this shared_ptr (owing to std::move semantics) , these nccl Communicator objects are not getting destroyed and memory leaks.

xla/backends/gpu/runtime/collective_thunk.cc
hsharsha: I could not follow the RAII chain here but these objects are never getting deallocated.

xla/tests/collective_ops_test.cc
hsharsha: 10 is a sweet spot where the test can be completed within 1000 seconds and can expose the problem with memory leak which results in pthread failure as this test does all combinations of all reduce and generates power set of the number of gpus. Any number <= 8 masks this memory leak issue and test runs successfully.

hsharsha:
> Hi Harsha, thanks for the PR. Would you mind elaborating on the problem you are trying to solve?

Memory leak issue with ncclCommunicators when more than 8+ gpus are involved.

mwhittaker:
The `GpuCliques` are cached in a global map:

https://github.com/openxla/xla/blob/b2d0159e10901612433cecddd33d19e6fd93a77e/xla/backends/gpu/collectives/gpu_cliques.cc#L106-L110

Are these the objects you're seeing leak? I would have thought they would be destructed when the program terminates.

Your changes (e.g., deleting unowned memory and prematurely stopping the watchdog thread) unfortunately don't look correct to me. If you could pinpoint what is leaking and why, maybe we can make a deeper change to address the root issue.

hsharsha:
> The `GpuCliques` are cached in a global map:
> 
> https://github.com/openxla/xla/blob/b2d0159e10901612433cecddd33d19e6fd93a77e/xla/backends/gpu/collectives/gpu_cliques.cc#L106-L110
> 
> Are these the objects you're seeing leak? I would have thought they would be destructed when the program terminates.
> 
> Your changes (e.g., deleting unowned memory and prematurely stopping the watchdog thread) unfortunately don't look correct to me. If you could pinpoint what is leaking and why, maybe we can make a deeper change to address the root issue.

@mwhittaker  The main entry point of the thunk execution is [ExecuteThunksImpl](https://github.com/openxla/xla/blob/main/xla/service/gpu/gpu_executable.cc#L195) This function initializes gpu_cliques depending on the number of gpus and ncclCommunicator object to carry out collective operations across gpus. I expect at the end of this function all gpu cliques and ncclCommunicators to be destroyed.

AllReduce_AllCombinations test invokes the above function in a loop for 2^x where x is the number of gpus and we see memory bloat up there which indicates that cliques and ncclCommunicators are not destructed at the end of this ExecuteOnStream function.

mwhittaker:
> I expect at the end of this function all gpu cliques and ncclCommunicators to be destroyed.

The communicators are cached for later use. They shouldn't be destroyed at the end of `ExecuteThunksImpl`.

hsharsha:
> > I expect at the end of this function all gpu cliques and ncclCommunicators to be destroyed.
> 
> The communicators are cached for later use. They shouldn't be destroyed at the end of `ExecuteThunksImpl`.

What if the cliques are changing? Ideally we would have a fixed topology and caching the ncclCommunicators makes sense as the cliques do not change. But if we cache and cliques do change would it not be a memory leak?

golechwierowicz:
@mwhittaker @hsharsha is some follow up on this PR expected? if not, can we close the PR?

mwhittaker:
Let's close the PR for now. I think @hsharsha, you're suggesting that we delete cached communicators occasionally? I think the set of communicators is small, so garbage collecting the cache wouldn't have a huge effect?

PR #26772, opened 2025-05-20
[GPU] Memcpy all-to-all using stream dependency and pre-populated pointers

terryysun:
Memcpy all-to-all is using NCCL to sync execution progress and exchange pointers. This PR introduced following optimizations which significantly reduced the overhead:
1. Use stream dependency to sync execution progress
2. Pre-populate pointers during initialization.

In addition, this PR added progress sync after Memcpy, to ensure all data are ready before following usage.


xla/backends/gpu/runtime/all_to_all_thunk.cc
```diff
@@ -124,20 +131,75 @@ absl::Status AllToAllStartThunk::Initialize(const InitializeParams& params) {
...
+    std::optional<RankId> rank =
+        comm_handle.clique_key.rank(params.collective_params->global_device_id);
+    TF_ASSIGN_OR_RETURN(GpuCollectives * collectives,
+                        GetGpuCollectives(params));
+    size_t chunk_element_count = buffers_[0].element_count / num_ranks;
+    TF_ASSIGN_OR_RETURN(
+        std::vector<DeviceBufferPair> device_buffers,
+        ConvertToDeviceBuffers(params.buffer_allocations, buffers_,
+                               config_.config.operand_element_type));
+    for (int peer = 0; peer < num_ranks; ++peer) {
```
reedwm, line 160: The logic of this for-loop's body currently is you compute `buffer_idx = (rank+peer) % N`, use a rendezvous to exchange the values `output[buffer_idx]` into `rendezvous_results`, then grab `rendezvous_results[(rank - peer) % N]` to be the recv_ptr.

I find this somewhat hard to understand. The Rendezvous also gives every thread all the data, when it just needs the data from one other thread. 

How about this as an alternative: Outside the loop, you exchange the base destination buffers (`device_buffers`) via the Rendezvous, so each thread has all destination_buffers for all devices. Then each thread itself computes the correct output pointer for each device by calling `collectives->Slice`. So instead of calling `collectives->Slice` and exchanging the results, just exchange the base pointer and then call `collectives->Slice`.

I think this is more clear. It avoids sending redundant data, as the Rendezvous is called once outside the loop, which makes the logic easier to understand.

xla/backends/gpu/runtime/all_to_all_thunk.cc
```diff
@@ -124,20 +131,75 @@ absl::Status AllToAllStartThunk::Initialize(const InitializeParams& params) {
...
+      BufferRendezvousValue buffer_rendezvous_value;
+      buffer_rendezvous_value.rank = rank.value().value();
+      int buffer_idx = (rank.value().value() + peer) % num_ranks;
+      if (config_.has_split_dimension) {
+        buffer_rendezvous_value.buffer = reinterpret_cast<uint64_t>(
+            collectives
+                ->Slice(
+                    se::DeviceMemoryBase(device_buffers[0].destination_buffer),
+                    device_buffers[0].element_type,
```
reedwm, line 169: You call `device_buffers[0]` but can there be multiple device_buffers in the split_dimension case? Looks like the old logic does an all-to-all for each device buffer in the split_dimension case.

xla/backends/gpu/runtime/all_to_all_thunk.cc
```diff
@@ -289,22 +362,48 @@ static absl::Status RecvPtrFromPeer(void* ptr, RankId peer,
                           peer, GpuCollectives::On(stream));
 }
 
+// Syncs the execution progress across all devices.
+absl::Status SyncProgress(absl::string_view name,
```
reedwm, line 366: The similar logic in ra2a exchanges the events in the Rendezvous, instead of passing all the events to the function:

https://github.com/openxla/xla/blob/36dc59d1daf86f2bcad8a5a6a64d27b6980b9176/xla/backends/gpu/runtime/ragged_all_to_all_thunk.cc#L262-L271

The approach you do here is also good, but I think it's better to be consistent with what already exists. Eventually, we should unify the functions in a shared location to avoid code duplication, but no need to do that in this PR.

reedwm (changes requested):

xla/backends/gpu/runtime/all_to_all_thunk.cc
reedwm: > Thanks for the suggestion! Agreed that sounds cleaner but I don't think it's feasible. The lifecycle of `device_buffers` (or its elements) are local to the thread, only the underlying addresses (e.g. `device_buffers[0].destination_buffer`) are persistent. We need a loop of rendezvous anyways, plus some extra data structure to organize the rendezvous results from them. I played with it a bit locally and find it even more complicated than the current version if we are to do the rendezvous outside the loop.

I think it is feasible, although admittedly you would know better since you played with it locally. I don't see why we would need a loop of rendezvous anyway, and you can copy just `device_buffers[0].destination_buffer`.

In any case, no need to fix in this PR since you already tried to and found it more complicated. I might try to simplify it myself after it's submitted.

xla/backends/gpu/runtime/all_to_all_thunk.cc
reedwm: > Yes that's true. Do you know in which case we would want to have multiple device buffer for split-dim all-to-all? I was never able to trigger it with existing tests and can't think of a case for it. Plus the other mode doesn't seem to support this while they should be identical in terms of functionality. I'd prefer legalizing the buffer size if that's never used.

You're right, the split_dim case seems to require a single operand and output. The verifier checks this. So using device_buffers[0] is fine.

reedwm (approved):

PR #26948, opened 2025-05-23
SPMD Partial Windowed Einsums

philipphack:
Enables partial windowed einsums with multiple sharded operand dimensions as seen in simultaneous data and tensor parallelism. The tile assignments are of the form

dot(A([M,N,1]<=[M\*N]), B([1,N,M]<=[M,N]T(1,0) last_tile_dim_replicate)) -> D([M,1,N]<=[M\*N]),

i.e. the first operand of the dot has one sharded non-contracting dimension that maps to an identically sharded dimension in the output and one sharded non-contracting dimension that maps to a replicated dimension in the output. In the resulting windowed einsum, M parallel loops exchange the tiles of the first operand through N sequential collective permutes.

### DIFF
--- a/xla/service/gpu/gpu_spmd_pipeline.cc
+++ b/xla/service/gpu/gpu_spmd_pipeline.cc
@@ -120,7 +120,8 @@ void AddSPMDPasses(
           .debug_options()
           .xla_gpu_multi_streamed_windowed_einsum(),
       /*skip_checking_windowed_einsum_users=*/true,
-      /*disable_ag_rewrite_for_multiple_consumers=*/true, oper_size_threshold);
+      /*disable_ag_rewrite_for_multiple_consumers=*/true,
+      /*enable_partial_windowed_einsums=*/true, oper_size_threshold);
   spmd_pipeline.AddPass<CollectivePermuteMotion>();
 }
 
--- a/xla/service/spmd/BUILD
+++ b/xla/service/spmd/BUILD
@@ -100,6 +100,7 @@ xla_cc_test(
         ":sharding_format_picker",
         ":spmd_partitioner",
         ":spmd_prepare",
+        "//xla:literal_util",
         "//xla:shape_util",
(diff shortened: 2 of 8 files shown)


frgossen (changes requested):
This touches code that is not GPU-specific. Do we have a design doc for this? I'm mostly interested where you expect to see performance gains and how I can verify that.

abhinavgoel95:
This helps with collective matmul pattern matching.
Specifically, in cases where the LHS is sharded along two dimensions (DP/FSDP and TP).
The LHS shape would look like: [batch/DP, seq_len/TP, hidden_dim]

This is necessary for models with the above shape and collective matmul (which is very common).

Without this fix, the collective matmul does not even kick in models using DP/FSDP + TP.

In MaxText llama3-70B, FSDP=8 TP=2, we see a 3% E2E speedup.

reedwm:
> In MaxText llama3-70B, FSDP=8 TP=2, we see a 3% E2E speedup.

Can you provide instructions to reproduce the results here? Also, would it be possible to provide a repro which only uses 8 GPUs, as this is easier to run internally. Also, would it be possible to provide a single individual matmul that showcases the performance benefit of this PR, perhaps extracted from the llama3 model?

abhinavgoel95:
Yes, I can provide a repro on 8 GPUs.

frgossen (approved):

derdrdirk:
This breaks a unit test:

```
tensorflow/compiler/xla/service/spmd/stateful_rng_spmd_partitioner_test.cc:271
Expected equality of these values:
  CountInstructions(*module->entry_computation(), HloOpcode::kWhile)
    Which is: 0
  1
 ```
 
 Once fixed I can merge the PR.

philipphack:
Thanks @derdrdirk. Can you PTAL?

derdrdirk (approved):

PR #27127, opened 2025-05-27
Update kernel args limit for cuda

sfvaroglu:
We ran into an issue during internal testing that this PR addresses: the kernel args limit being too restrictive for CUDA.


xla/stream_executor/kernel.h
```diff
@@ -612,7 +617,13 @@ PackKernelArgs(absl::Span<const ArgType> args, uint32_t shared_mem_bytes) {
     return internal::PackKernelArgs<512>(args, shared_mem_bytes);
   }
 
-  return internal::PackKernelArgs<kKernelArgsLimit>(args, shared_mem_bytes);
+#if GOOGLE_CUDA && CUDA_VERSION >= 12010
```
jreiffers, line 620: Why not keep `return internal::PackKernelArgs<kKernelArgsLimit>(...)`?

jreiffers:
Why is this limit still being triggered? No such fusions should be created, so where did they come from?

xla/stream_executor/kernel.h
```diff
@@ -586,8 +586,13 @@ std::unique_ptr<KernelArgsPackedArrayBase> PackKernelArgs(
 template <typename ArgType>
 inline absl::StatusOr<std::unique_ptr<KernelArgsPackedArrayBase>>
 PackKernelArgs(absl::Span<const ArgType> args, uint32_t shared_mem_bytes) {
+#if GOOGLE_CUDA && CUDA_VERSION >= 12010
+  static constexpr int kKernelArgsLimit = 32764;
+#elif GOOGLE_CUDA
+  static constexpr int kKernelArgsLimit = 4096;
```
jreiffers, line 592: Where does this new number come from?

sfvaroglu: https://developer.nvidia.com/blog/cuda-12-1-supports-large-kernel-parameters/

jreiffers: Why 2^15-4? Also, the documentation says bytes. Each parameter is a pointer, which is 4-8 bytes depending on the pointer size. 

To me it sounds like 1024 is the limit assuming the old CUDA version and 4 byte pointers.

sfvaroglu: Thanks for catching this! Fixed.

sfvaroglu:
> Why is this limit still being triggered? No such fusions should be created, so where did they come from?

I'm investigating this, but I don't think it should block this PR.

xla/stream_executor/kernel.h
jreiffers: Wouldn't that be 4095 then? floor(32764/8).

jreiffers (approved):

xla/stream_executor/gpu/BUILD
```diff
@@ -904,3 +904,22 @@ xla_test(
         "@com_google_googletest//:gtest_main",
     ],
 )
+
+xla_test(
+    name = "gpu_kernel_args_test",
+    srcs = ["gpu_kernel_args_test.cc"],
+    backends = ["gpu"],
+    local_defines = if_cuda_is_configured(["GOOGLE_CUDA=1"]),
```
akuegel: Is this really needed? Can you maybe figure out at runtime whether it runs with cuda?

xla/stream_executor/kernel.h
```diff
@@ -586,8 +586,11 @@ std::unique_ptr<KernelArgsPackedArrayBase> PackKernelArgs(
 template <typename ArgType>
 inline absl::StatusOr<std::unique_ptr<KernelArgsPackedArrayBase>>
 PackKernelArgs(absl::Span<const ArgType> args, uint32_t shared_mem_bytes) {
+#if GOOGLE_CUDA && CUDA_VERSION >= 12010
```
akuegel: cuda specific changes should not be in this directory. I checked that this helper function is used only on GPU, maybe it can be moved to the GPU directory, and adjusted separately for ROCM and CUDA? Ideally, we should not use a GOOGLE_CUDA macro.

akuegel (changes requested):

akuegel:
> > Why is this limit still being triggered? No such fusions should be created, so where did they come from?
> 
> I'm investigating this, but I don't think it should block this PR.

I agree with Johannes, this limit should not be triggered right now. https://github.com/openxla/xla/pull/27128 seems like it will fix the issue you found. For regular fusions, we currently enforce an even smaller limit of 96 (but I was planning on trying to increase that limit).
I am open to making it possible in theory to use a higher limit on newer cuda versions, but I fear that will involve some refactoring.

xla/stream_executor/kernel.h
sfvaroglu: updated

sfvaroglu:
> > > Why is this limit still being triggered? No such fusions should be created, so where did they come from?
> > 
> > 
> > I'm investigating this, but I don't think it should block this PR.
> 
> I agree with Johannes, this limit should not be triggered right now. #27128 seems like it will fix the issue you found. For regular fusions, we currently enforce an even smaller limit of 96 (but I was planning on trying to increase that limit). I am open to making it possible in theory to use a higher limit on newer cuda versions, but I fear that will involve some refactoring.

Right, #27128 fixes the issue. I added the helper functions for ROCM and CUDA separately, but that seems to have triggered a lot of test failures due to missing header files. Given that the limit (1024 or 4095) probably won't be hit, I'm wondering if it's worth pursuing this PR further.

akuegel:
> > > > Why is this limit still being triggered? No such fusions should be created, so where did they come from?
> > > 
> > > 
> > > I'm investigating this, but I don't think it should block this PR.
> > 
> > 
> > I agree with Johannes, this limit should not be triggered right now. #27128 seems like it will fix the issue you found. For regular fusions, we currently enforce an even smaller limit of 96 (but I was planning on trying to increase that limit). I am open to making it possible in theory to use a higher limit on newer cuda versions, but I fear that will involve some refactoring.
> 
> Right, #27128 fixes the issue. I added the helper functions for ROCM and CUDA separately, but that seems to have triggered a lot of test failures due to missing header files. Given that the limit (1024 or 4095) probably won't be hit, I'm wondering if it's worth pursuing this PR further.

Yes, I think it might not be worth it to pursue this PR further (at least not for now). As far as I understand, in XLA we use a lower limit than possible for fusions because there is a risk of register pressure if there are too many fusion operands. So it could be we would need a heuristic for evaluating that if we want to increase the limit further. Given this, it seems unlikely we would make use of a higher limit here in StreamExecutor anytime soon.

sfvaroglu:
Thanks, closing this PR for now.

PR #27128, opened 2025-05-27
Make copy_fusion respect the budget limit

sfvaroglu:
copy_fusion does not respect the budget limit, this PR provides a fix.


xla/service/gpu/transforms/copy_fusion_test.cc
```diff
@@ -508,5 +506,54 @@ TEST_F(CopyFusionTest, CopyFusionWithFusionReturningTupleAndOtherUser) {
...
+  ASSERT_THAT(root, GmockMatch(m::Tuple(m::GetTupleElement(m::Fusion(&fusion)),
+                                        m::GetTupleElement())));
+  EXPECT_THAT(fusion->fused_expression_root(),
+              GmockMatch(m::Tuple(m::Negate(), m::Copy(), m::Copy())));
+}
+
+TEST_F(CopyFusionTest, FusionExceedsBudget) {
+  auto module = ParseAndReturnVerifiedModule(absl::StrCat(kModulePrefix, R"(
+    fused_computation {
+      p1.1 = f32[128,128,128]{2,1,0} parameter(0)
```
jreiffers, line 538: Hmm,  couple of things:

- Are you sure this test fails before your change? I would expect it to fail the OnlyElementwiseOpsReachableFromParams test, so there should be no copy fusion.
- This fusion doesn't use any shared memory (because none of the shared memory emitters support it), so it shouldn't actually exceed the budget. If it does, that's a bug.

Maybe you can create a fusion with many parameters with a loop. Something like this: https://github.com/openxla/xla/blob/159144edb8cc53ebc3bad4fea14737a5690dade3/xla/service/collective_ops_utils_test.cc#L155-L171. Your fusion kind should be `kLoop`, not `kInput`.

jreiffers (changes requested):

xla/service/gpu/transforms/copy_fusion_test.cc
```diff
@@ -533,26 +533,47 @@ TEST_F(CopyFusionTest, FusionFitsInBudget) {
 }
 
 TEST_F(CopyFusionTest, FusionExceedsBudget) {
```
jreiffers, line 535: Please add a companion test with a fusion that has one less parameter and verify that copy fusion triggers.

xla/service/gpu/transforms/copy_fusion_test.cc
```diff
@@ -533,26 +533,47 @@ TEST_F(CopyFusionTest, FusionFitsInBudget) {
...
-      broadcast = f32[128,128,128]{2,1,0} broadcast(reduce), dimensions={0}
-      transpose2 = f32[128,128,128]{2,1,0} transpose(broadcast), dimensions={2,1,0}
-      ROOT add = f32[128,128,128]{2,1,0} add(transpose2, transpose2)
-    }
+  auto module = CreateNewVerifiedModule();
+  HloComputation::Builder fusion_builder("fusion");
+  std::vector<HloInstruction*> fusion_params;
+  std::vector<Shape> param_shapes;
+
+  for (int i = 0; i < 97; ++i) {
```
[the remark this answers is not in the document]
sfvaroglu: done

xla/service/gpu/transforms/copy_fusion_test.cc
```diff
@@ -508,5 +507,93 @@ TEST_F(CopyFusionTest, CopyFusionWithFusionReturningTupleAndOtherUser) {
...
 }
 
+TEST_F(CopyFusionTest, FusionFitsInBudget) {
+  auto module = CreateNewVerifiedModule();
+  HloComputation::Builder fusion_builder("fusion");
+  std::vector<HloInstruction*> fusion_params;
+  std::vector<Shape> param_shapes;
+
+  for (int i = 0; i < MaxOperandsAndOutputsPerFusion() - 2; ++i) {
+    param_shapes.push_back(ShapeUtil::MakeShape(F32, {1}));
```
jreiffers: This is used quite a lot, so you could extract it. You could probably also initialize param_shapes like this and skip the `push_back`:

```
std::vector<Shape> param_shapes(MaxOperandsAndOutputsPerFusion() - 1, unit_shape);
```

If you feel like it.

xla/service/gpu/transforms/copy_fusion_test.cc
jreiffers: Let's extract the duplicated part into a helper function.

xla/service/gpu/transforms/copy_fusion_test.cc
sfvaroglu: Done

xla/service/gpu/transforms/copy_fusion_test.cc
```diff
@@ -24,24 +24,23 @@ limitations under the License.
...
-}
-
 class CopyFusionTest : public HloHardwareIndependentTestBase {
  public:
-  CopyFusionTest()
-      : device_description_(MakeDeviceDescriptor()), cf_(device_description_) {}
-  const stream_executor::DeviceDescription device_description_;
+  CopyFusionTest() : cf_(device_description_) {
+    stream_executor::GpuDeviceInfoProto proto;
+    proto.set_shared_memory_per_block(1024 * 1024);  // 1MB
```
jreiffers: This part can probably be reverted, the new tests don’t exercise the shmem logic.

xla/service/gpu/transforms/copy_fusion_test.cc
jreiffers: I meant a function that takes a parameter for the number of operands, so we don’t duplicate all that.

xla/service/gpu/transforms/copy_fusion_test.cc
```diff
@@ -508,5 +507,95 @@ TEST_F(CopyFusionTest, CopyFusionWithFusionReturningTupleAndOtherUser) {
...
+  HloInstruction* copy = builder.AddInstruction(
+      HloInstruction::CreateUnary(fusion->shape(), HloOpcode::kCopy, fusion));
+
+  HloComputation* computation =
+      module->AddEntryComputation(builder.Build(copy));
+
+  EXPECT_TRUE(cf_.Run(module.get()).value());
+}
+
+TEST_F(CopyFusionTest, FusionExceedsBudget) {
```
jreiffers, line 584: Looks like comments in reaolved threads don’t show up. We should still extract the common logic in the two tests into a function that takes the number of operands as a parameter, to get rid of the duplication and make it obvious what the difference of the two tests is.

xla/service/gpu/transforms/copy_fusion_test.cc
sfvaroglu: These tests fail, if reverted:
```
[  FAILED  ] CopyFusionTest.CopyFusionTransposeOfBroadcastedConstantTwoCopies
[  FAILED  ] CopyFusionTest.CopyFusionShouldRunWithUncopiedReduce
[  FAILED  ] CopyFusionTest.CopyFusionTransposeAndThreeCopies
```

xla/service/gpu/transforms/copy_fusion_test.cc
sfvaroglu: Done

xla/service/gpu/transforms/copy_fusion_test.cc
jreiffers: > These tests fail, if reverted

Ah sorry, I missed that. Let's just use TestGpuDeviceInfo::RTXA6000DeviceInfo then, if possible.

xla/service/gpu/transforms/copy_fusion_test.cc
```diff
@@ -24,25 +24,71 @@ limitations under the License.
...
+    proto.set_shared_memory_per_core(1024 * 1024);   // 1MB
+    device_description_ = stream_executor::DeviceDescription(proto);
+  }
+
+  stream_executor::DeviceDescription device_description_;
   CopyFusion cf_;
+
+  // Helper function to create a fusion with a specified number of operands.
+  // Returns true if the fusion was successful, false otherwise.
+  bool CreateAndTestFusion(int64_t num_operands) {
```
jreiffers: Thanks, this is a good start. Let's give it a different name, since this function doesn't really test in itself. We can also simplify the logic a bit, since `param_shapes` isn't really needed. We can also use `Build` without an explicit argument. Finally, we can start the sum at 0, which makes the logic a bit simpler. Something like this (not tested):

```
  bool CreateFusionAndRunCopyFusion(int64_t num_operands) {
    auto module = CreateNewVerifiedModule();
    Shape unit_shape = ShapeUtil::MakeShape(F32, {});
    
    // Create a fusion with `num_operands` parameters.
    HloComputation::Builder fusion_builder("fusion");
    HloInstruction* sum = fusion_builder.AddInstruction(
        HloInstruction::CreateConstant(LiteralUtil::CreateR0<float>(0.0)));
    for (int i = 0; i < num_operands; ++i) {
      HloInstruction* param =
          fusion_builder.AddInstruction(HloInstruction::CreateParameter(
              i, unit_shape, absl::StrFormat("param%d", i))));
      sum = fusion_builder.AddInstruction(
          HloInstruction::CreateBinary(unit_shape,
                                       HloOpcode::kAdd, sum, param));
   }
    HloComputation* fusion_computation =
        module->AddEmbeddedComputation(fusion_builder.Build());
   
    // Create an entry computation that calls the fusion and then makes a copy.
    HloComputation::Builder entry_builder(TestName());
    std::vector<HloInstruction*> entry_params;
    for (int i = 0; i < num_operands; ++i) {
      entry_params.push_back(
          entry_builder.AddInstruction(HloInstruction::CreateParameter(
              i, unit_shape, absl::StrFormat("param%d", i))));
    }
    HloInstruction* fusion =
        entry_builder.AddInstruction(HloInstruction::CreateFusion(
            unit_shape, HloInstruction::FusionKind::kLoop,
            params, fusion_computation));
    entry_builder.AddInstruction(
        HloInstruction::CreateUnary(fusion->shape(), HloOpcode::kCopy, fusion));
    module->AddEntryComputation(builder.Build());

    return cf_.Run(module.get()).value();
  }
```

jreiffers: (This changes the shape from {1} to {}. If that doesn't work for some reason, just use CreateR1 instead of CreateR0).

xla/service/gpu/transforms/copy_fusion_test.cc
sfvaroglu: Done

jreiffers (approved):
Thanks for the changes!

akuegel (approved):

PR #27188, opened 2025-05-28
[XLA:GPU] Hermetic Build oneAPI

mraunak:
This PR introduces support for a hermetic build for XLA using oneAPI toolkit. It enables Bazel to download and extract a prebuilt SYCL basekit automatically during the build process, ensuring consistent, reproducible environments without requiring a system-wide oneAPI installation.
It also adds basic build tests for CI. Execution of tests will be covered in the subsequent PRs

### DIFF
--- a/tensorflow.bazelrc
+++ b/tensorflow.bazelrc
@@ -260,10 +260,18 @@ build:rocm_ci_hermetic --repo_env="OS=ubuntu_22.04"
 build:rocm_ci_hermetic --repo_env="ROCM_VERSION=6.2.0"
 build:rocm_ci_hermetic --@local_config_rocm//rocm:rocm_path_type=hermetic
 
+# SYCL Configuration (non-hermetic)
 build:sycl --crosstool_top=@local_config_sycl//crosstool:toolchain
 build:sycl --define=using_sycl=true
 build:sycl --define=tensorflow_mkldnn_contraction_kernel=0
-build:sycl --repo_env TF_NEED_SYCL=1
+build:sycl --repo_env=TF_NEED_SYCL=1
+build:sycl --repo_env=SYCL_BUILD_HERMETIC=0
+
+# Hermetic SYCL Configuration
+build:sycl_hermetic --config=sycl
+build:sycl_hermetic --repo_env=SYCL_BUILD_HERMETIC=1
+build:sycl_hermetic --repo_env=ONEAPI_VERSION=2025.1
+build:sycl_hermetic --repo_env=OS=ubuntu_24.10
 
 # Options to disable default on features
 build:nonccl --define=no_nccl_support=true
--- a/third_party/gpus/sycl/BUILD.tpl
+++ b/third_party/gpus/sycl/BUILD.tpl
@@ -1,13 +1,13 @@
 package(default_visibility = ["//visibility:public"])
 
 # Intel(R) Software Development Tools Licensed under the Intel End User License Agreement for Developer Tools (Version August 2024)
-# Intel(R) oneAPI DPC++/C++ Compiler, Intel(R) Vtune(TM) Profiler
+# Tools -> Intel(R) oneAPI DPC++/C++ Compiler, Intel(R) Vtune(TM) Profiler
 # Intel(R) Software Development Tools Licensed under the Intel Simplified Software License (Version October 2022)  
-# oneAPI Math Kernel Library (oneMKL)
+# Tools -> oneAPI Math Kernel Library (oneMKL)
 # Intel(R) Software Development Tools Licensed under Open Source Licenses Apache License, Version 2.0 
-# oneAPI Deep Neural Network Library, Intel(R) oneAPI Data Analytics Library (oneDAL)
-# Apache License, Version 2.0 with LLVM Exception --Intel(R) oneAPI DPC++/C++ Compiler,Intel(R) oneAPI DPC++ Library (oneDPL)
-# The GNU General Public License v3.0 ---Intel(R) Distribution for GDB*
+# Tools -> oneAPI Deep Neural Network Library, Intel(R) oneAPI Data Analytics Library (oneDAL)
+# Apache License, Version 2.0 with LLVM Exception -- Tools ->Intel(R) oneAPI DPC++/C++ Compiler,Intel(R) oneAPI DPC++ Library (oneDPL)
+# The GNU General Public License v3.0 -> Tools-- Intel(R) Distribution for GDB*
 licenses(["restricted"])  
 
 config_setting(
--- a/third_party/gpus/sycl/BUILD.tpl
+++ b/third_party/gpus/sycl/BUILD.tpl
@@ -1,4 +1,5 @@
 package(default_visibility = ["//visibility:public"])
+
 # Intel(R) Software Development Tools Licensed under the Intel End User License Agreement for Developer Tools (Version August 2024)
 # Intel(R) Software Development Tools Licensed under the Intel Simplified Software License (Version October 2022)
 # Intel(R) Software Development Tools Licensed under Open Source Licenses Apache License, Version 2.0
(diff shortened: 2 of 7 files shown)


third_party/gpus/sycl/BUILD.tpl
```diff
@@ -1,5 +1,12 @@
 package(default_visibility = ["//visibility:public"])
 
+# Intel(R) Software Development Tools Licensed under the Intel End User License Agreement for Developer Tools (Version August 2024)
+# Intel(R) Software Development Tools Licensed under the Intel Simplified Software License (Version October 2022)
+# Intel(R) Software Development Tools Licensed under Open Source Licenses Apache License, Version 2.0
+# Apache License, Version 2.0 with LLVM Exception
+# The GNU General Public License v3.0
+licenses(["restricted"])
```
beckerhe: I'm not sure what this means. Can you clarify to that tools this applies to?

tensorflow.bazelrc
```diff
@@ -258,10 +258,18 @@ build:rocm_ci_hermetic --repo_env="OS=ubuntu_22.04"
 build:rocm_ci_hermetic --repo_env="ROCM_VERSION=6.2.0"
 build:rocm_ci_hermetic --@local_config_rocm//rocm:rocm_path_type=hermetic
 
+# SYCL Configuration (non-hermetic)
 build:sycl --crosstool_top=@local_config_sycl//crosstool:toolchain
 build:sycl --define=using_sycl=true
 build:sycl --define=tensorflow_mkldnn_contraction_kernel=0
-build:sycl --repo_env TF_NEED_SYCL=1
+build:sycl --repo_env=TF_NEED_SYCL=1
```
beckerhe, line 267: Is TF_NEED_SYCL going to be necessary? I'm asking because I would like to minimize the number of macros and would prefer it to be removed if it's not really, really necessary.

xla/stream_executor/sycl/BUILD
```diff
@@ -77,6 +77,10 @@ cc_library(
 
 cc_library(
     name = "stream_executor_sycl",
+    tags = [
+        "gpu",
+        "sycl-only",
```
beckerhe, line 91: Can you document the tag `sycl-only` in `build_tools/lint/tags.py`?

beckerhe (commented):
Hi, thanks a lot for this change. I've left some comments, but overall looks good to me.

mraunak:
Hi @beckerhe, I have made changes as commented. Please review and let me know if any further change is needed

tensorflow.bazelrc
beckerhe: So what't he conclusion? For what is this needed?

mraunak: This is a deciding factor whether to create a dummy repository or a fully configured SYCL repository
 https://github.com/Intel-tensorflow/xla/blob/mraunak/Hermetic_GCC/third_party/gpus/sycl_configure.bzl#L568


 However, we will try to eliminate the macros which are not necessary in the upcoming PRs

mraunak:
Hi @beckerhe since TF_NEED_SYCL=1 is currently the deciding factor to create a fully functional SYCL repository or a dummy repository. Hence, we are keeping it for now. However, I will try to change the code base further and try to remove TF_NEED_SYCL if possible in my upcoming PR. Thank you!

mraunak:
Hi @beckerhe, I have made changes as commented. Please review and let me know if any further change is needed.
Since the TF_NEED_SYCL flag currently determines whether a fully functional SYCL repository or a dummy repository is created, we are retaining it in this PR

third_party/gpus/sycl_configure.bzl
```diff
@@ -381,7 +409,33 @@ def _create_local_sycl_repository(repository_ctx):
...
+          sycl_basekit_path = install_path + "/oneapi/",
+          sycl_toolkit_path = install_path + "/oneapi/compiler/"+ oneapi_version,
+          # TODO(Intel) 
+          # hard coded for this oneAPI version 2025.1,
+          # change to auto-detect the sycl_version_number
+          sycl_version_number = "80000",
+          sycl_basekit_version_number = oneapi_version,
+        )
+    else:
+        install_path = repository_ctx.os.environ.get("SYCL_TOOLKIT_PATH", "") or "/opt/intel/oneapi/compiler/2025.1"
```
ybaturina: Here and in other places: I suggest using `repository_ctx.getenv(name)` - in this case the repository will be recreated if the env value changes.

third_party/gpus/sycl_configure.bzl
```diff
@@ -381,7 +409,33 @@ def _create_local_sycl_repository(repository_ctx):
...
+        repository_ctx.report_progress("Falling back to default SYCL path: %s" % install_path)
+        sycl_config = _get_sycl_config(repository_ctx, bash_bin)
 
     # Copy header and library files to execroot.
     copy_rules = [
```
ybaturina, line 441: I haven't seen previous PRs, so I have a question: why do you need to copy the files from one place to another?

mraunak: Copy_rules were implemented and used for both hermetic and non-hermetic builds to expose SYCL headers and libraries to Bazel in a structured, declarative way. This ensures Bazel can track them as inputs, apply caching correctly, and use them in BUILD rules reliably.

mraunak:
Thank you @ybaturina  for reviewing the PR. I have updated my PR as per your comments. Please let me know if there is a further change needed

third_party/gpus/sycl_configure.bzl
ybaturina: Just to clarify: are you using the same approach as in CUDA repository rules with LOCAL_CUDA_PATH?
I'm curious because I wonder if CUDA approach can be applied to SYCL as well, e.g. if we already have SYCL files somewhere on the machine, we can symlink the containing folders in SYCL repository, so no need to have additional copy rules.

mraunak: Hi @ybaturina, I’ve added the hermetic approach to the build. The non-hermetic support was implemented in earlier PRs following the CUDA-based approach, which used copy rules. Since both hermetic and non-hermetic logic share the same file(sycl_configure.bzl), we’ve continued using copy rules for consistency. However, for hermetic builds specifically, we can optimize this by using symlinks instead. I think non hermetic build might break due to using the symlinks, hence was not implemented earlier. However, if you strictly recommend using symlinks for both hermetic and non hermetic builds, I can dive deep and try to implement it.

https://github.com/openxla/xla/blob/main/third_party/gpus/cuda_configure.bzl#L896

ybaturina: OK, got it - so you were looking at the old `cuda_configure.bzl` that is not recommended for the usage anymore.
I had to create a new `hermetic/cuda_configure.bzl` to make it distinguishable from the old repository rules. I assume that this can be done for SYCL as well, but not in this PR.

third_party/gpus/sycl_configure.bzl
```diff
@@ -381,7 +409,33 @@ def _create_local_sycl_repository(repository_ctx):
     ]}
 
     bash_bin = get_bash_bin(repository_ctx)
-    sycl_config = _get_sycl_config(repository_ctx, bash_bin)
+   
+    hermetic = repository_ctx.getenv("SYCL_BUILD_HERMETIC") == "1"
+    if hermetic:
+        oneapi_version = repository_ctx.getenv("ONEAPI_VERSION") or ""
```
ybaturina: You can replace it with `repository_ctx.getenv("ONEAPI_VERSION", "")` here and in other places where you need a default value different from `None`.

ybaturina (approved):
Approved with a small comment.

Also I think it would be beneficial to have separate repo rule implementation than non-hermetic one, but it can be addressed later.

@vam-google please take a look too.

third_party/gpus/sycl_configure.bzl
mraunak: Thank you @ybaturina, I have made the above changes. I will work on implementing the separate repo rule for hermetic and non-hermetic

vam-google (approved):
LGTM

PR #27276, opened 2025-05-30
Add Support for Custom call with Called Computation and Layout Constraints

Cjkkkk:
Overload custom call creation to support custom call with both computation attached and layout constraint. Useful for cudnn sdpa custom call where layout constraint is needed and computation is used to extend cudnn functionality to support flex attention. See https://github.com/openxla/xla/pull/25298.

### DIFF
--- a/xla/hlo/ir/hlo_instruction.cc
+++ b/xla/hlo/ir/hlo_instruction.cc
@@ -1008,9 +1008,16 @@ absl::StatusOr<std::unique_ptr<HloInstruction>> HloInstruction::CreateFromProto(
         for (const ShapeProto& shape_proto : operand_shapes_with_layout) {
           operand_shapes.emplace_back(shape_proto);
         }
-        instruction =
-            CreateCustomCall(shape, all_operands(), proto.custom_call_target(),
-                             operand_shapes, proto.backend_config());
+        if (proto.called_computation_ids_size() == 1) {
+          instruction =
+              CreateCustomCall(shape, all_operands(), computations(0),
+                               proto.custom_call_target(), operand_shapes,
+                               proto.backend_config());
+        } else {
+          instruction = CreateCustomCall(
+              shape, all_operands(), proto.custom_call_target(), operand_shapes,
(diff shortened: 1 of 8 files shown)


xla/hlo/ir/hlo_instruction.cc
```diff
@@ -1008,9 +1008,16 @@ absl::StatusOr<std::unique_ptr<HloInstruction>> HloInstruction::CreateFromProto(
         for (const ShapeProto& shape_proto : operand_shapes_with_layout) {
           operand_shapes.emplace_back(shape_proto);
         }
-        instruction =
-            CreateCustomCall(shape, all_operands(), proto.custom_call_target(),
-                             operand_shapes, proto.backend_config());
+        if (proto.called_computation_ids_size() == 1) {
```
akuegel, line 1012: Can you please add a TF_RET_CHECK that proto.called_computation_ids_size() <= 1 ? Or alternatively add support for more than 1 called computation? It seems somewhat bad if someone tries to use a custom call with layout constraint and with more than one called computation and the called computations are silently dropped.
Given that mlir_hlo_to_hlo seems to support only 1 called computation, my preference would be a TF_RET_CHECK.

xla/service/gpu/custom_call_test.cc
```diff
@@ -611,6 +611,25 @@ TEST_F(CustomCallTest, WithCalledComputation) {
   EXPECT_THAT(result.data<float>(), ::testing::Each(42));
 }
 
+TEST_F(CustomCallTest, WithCalledComputationAndLayouts) {
+  auto shape = ShapeUtil::MakeShapeWithDenseLayout(F32, {128}, {0});
```
akuegel: Can you please adjust the test so that a layout constraint is actually meaningful? For rank 1 it doesn't matter.

akuegel (approved):
Thank you!

akuegel (approved):

Cjkkkk:
I just rebased and I think CI should be passing now.

PR #27371, opened 2025-06-03
PJRT_Executable_DeserializeAndLoad: plumb compile options

### LINKED ISSUE #18286: Deserialization of executables fails on non-zero ranks when deserializing single-device executable

olupton:
This plumbs overriden compilation options through `PJRT_Executable_DeserializeAndLoad` in protobuf format and into `LoadSerializedExecutable`. Previously these were dropped.

This should close https://github.com/openxla/xla/issues/18286 and address more recent flakes with the JAX compilation cache with the same root cause but a different error `jaxlib._jax.XlaRuntimeError: INVALID_ARGUMENT: Expected args to execute_sharded_on_local_devices to have 0 shards, got: ...`.

### DIFF
--- a/xla/pjrt/c/CHANGELOG.md
+++ b/xla/pjrt/c/CHANGELOG.md
@@ -1,11 +1,14 @@
 # PJRT C API changelog
 
-## 0.69
+## 0.70
+* Added `overridden_serialized_compile_options` and
+  `overridden_serialized_compile_options_size` fields to
+  `PJRT_Executable_DeserializeAndLoad_Args`.
 
+## 0.69
 * Implemented PjRtClient::CreateUninitializedBuffer
 
 ## 0.68
-
 * Changed the type of ``topology`` in
   ``PJRT_TopologyDescription_PlatformName_Args`` and
   ``PJRT_TopologyDescription_GetDeviceDescriptions_Args`` from
--- a/xla/pjrt/c/pjrt_c_api.h
+++ b/xla/pjrt/c/pjrt_c_api.h
@@ -82,7 +82,7 @@ PJRT_DEFINE_STRUCT_TRAITS(PJRT_Extension_Base, next);
 // Changes include:
 // * Adding a new field to the PJRT_Api or argument structs
 // * Renaming a method or argument (doesn't affect ABI)
-#define PJRT_API_MINOR 69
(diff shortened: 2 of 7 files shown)


jparkerh (approved):
This LGTM, thanks for plumbing this through!

olupton:
Rebased because the minor version changed underneath the patch.

jparkerh (approved):

ezhulenev (approved):

PR #27412, opened 2025-06-04
Command buffer respect control dependency of HloInstruction when running with concurrent mode.

shawnwang18:
When command buffer is running "xla_gpu_graph_enable_concurrent_region=true", i.e, command buffer will try run command sequence in parallelized mode if there are no data dependency. It will break the control dependency that was originally in XLA computation, because control dependency is originally guaranteed by running XLA thunks in serialized order, and when command buffer trying to parallelize thunk sequence, we should again respect the control dependency in HLO computation. 
 
This PR fixes the issue by bookkeeping the control dependency across thunks, and when lower thunk sequence into command buffer, uses the `token` resource for maintaining the control dependency between command buffer commands.


xla/backends/gpu/runtime/command_buffer_cmd.h
```diff
@@ -277,6 +279,13 @@ class CommandBufferCmd {
   // updates, thus they need to be consistent across calls to the function.
   virtual BufferUseVector buffers() const = 0;
 
+  absl::Status add_resource_use(ResourceUse use) {
```
ezhulenev: AddResourceUse as it's not a simple getter/setter

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
```diff
@@ -334,8 +335,40 @@ static absl::Status AppendCommands(CommandBufferCmdSequence& cmd_sequence,
...
+  for (const std::unique_ptr<Thunk>& thunk : sequence) {
+    int64_t previous_sequence_size = cmd_sequence.size();
     TF_RETURN_IF_ERROR(AppendCommands(cmd_sequence, *thunk, options));
+    if (cmd_sequence.size() > previous_sequence_size) {
+      thunk_to_cmd[thunk.get()] = cmd_sequence.back().get();
+    }
+  }
+
+  absl::flat_hash_map<CommandBufferCmd*, std::vector<CommandBufferCmd*>>
+      cmd_to_control_decessors;
```
ezhulenev: control_successors? (and everywhere below)

xla/backends/gpu/runtime/command_buffer_cmd.h
ezhulenev: I have a better idea, instead of mutating commands you can build thunk->resources mapping ahead of time from thunk sequence, and then pass them in constructor. Prefer immutable types when possible.

ezhulenev (approved):

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
shawnwang18: fixed

xla/backends/gpu/runtime/command_buffer_cmd.h
shawnwang18: Fixed.

xla/stream_executor/rocm/rocm_command_buffer.cc
```diff
@@ -338,6 +338,11 @@ absl::Status RocmCommandBuffer::UpdateKernelNode(
                   "Failed to set HIP graph kernel node params");
 }
 
+absl::StatusOr<GraphNodeHandle> RocmCommandBuffer::CreateEmptyNode(
```
thomasjoerg, line 341: This fails to compile with cuda-clang: `function definition is not allowed here`

thomasjoerg (changes requested):

xla/stream_executor/rocm/rocm_command_buffer.cc
shawnwang18: Sorry, I checked the code, and did not get why there is this error. Can you point to me?

ezhulenev (approved):

xla/stream_executor/rocm/rocm_command_buffer.cc
shawnwang18: @ezhulenev can you point to me why cuda-clang fails with the error `function definition is not allowed here` here?

ezhulenev (approved):

ezhulenev (approved):

PR #27469, opened 2025-06-05
[GPU] Enable copy engine for intra-node communications by default

terryysun:
Copy engine has better performance than NCCL kernels when available, plus it doesn't occupy SMs. This PR enables copy engine for local communications by default.


dimitar-asenov:
@frgossen Could you please take a look.

frgossen:
> Copy engine has better performance than NCCL kernels when available, plus it doesn't occupy SMs. This PR enables copy engine for local communications by default.

Can you quantify this? Does this show on maxtext configs or other reproducers? 
IIRC, memcpy_local_p2p causes crashes in the past. Can you add an execution test case?

PR #27498, opened 2025-06-05
[ROCm] Add new hip_runtime bazel target

mrodden:
In the jaxlib extension for ROCm, we need to link to the HIP runtime (libamdhip64.so) but without the extra links that are on rocm_hip, such as comgr, rocm-smi, and HSA runtime (which are linked by libamdhip itself)

Comgr shouldn't be directly linked by any HIP users either, but that hasn't been an issue for others apparently, so add this target that we can use from jaxlib and not intefere with rocm_hip.

### DIFF
--- a/third_party/gpus/rocm/BUILD.tpl
+++ b/third_party/gpus/rocm/BUILD.tpl
@@ -180,6 +180,24 @@ cc_library(
     ],
 )
 
+# NOTE(mrodden): used by jax_rocm_plugin to minimally link to hip runtime
+cc_library(
+    name = "hip_runtime",
+    visibility = ["//visibility:public"],
+    hdrs = glob(["%{rocm_root}/include/hip/**"]),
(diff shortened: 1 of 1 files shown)


mrodden:
@xla-rotation If you could take a look, I believe this should be a low risk change. Thanks

derdrdirk (approved):

third_party/gpus/rocm/BUILD.tpl
```diff
@@ -180,6 +180,24 @@ cc_library(
     ],
 )
 
+# NOTE(mrodden): used by jax_rocm_plugin to minimally link to hip runtime
```
thomasjoerg, line 183: `NOTE(username)` is not commonly used in the XLA codebase (unlike TODO). I'll remove this on import.

mrodden: Old habit from openstack days, and wasn't sure what the XLA convention was but I had seen TODO

PR #27721, opened 2025-06-11
Always evaluate while init in loop analysis.

yliu120:


yliu120 (commented):
I think the best way would be just evaluating the init value and materialize that value in HLO. 
It really depends on what upstream thinks is correct.

jreiffers (approved):
Thanks for the fix!

akuegel:
Based on the current code, it seems the intention is to have MatchLoopRangeWithKnownValues() work without HloEvaluator, and do the more costly evaluation separately. I see that ComputeWhileLoopTripCount calls MatchTrivialLoopTripCount(), but maybe it can be refactored to (also) call MatchLoopRangeWithKnownValues()? We would need to be able to pass the evaluated ind_var, so maybe extract the part of MatchLoopRangeWithKnownValues after we have the ind_var into another helper function? Then we could preserve that MatchLoopRangeWithKnownValues() only works with known values and does not need HloEvaluator.

jreiffers:
@akuegel I agree that this code could be refactored. Can we merge this for now and I'll look into the refactoring you suggested separately? I don't think the HloEvaluator invocation for the initial value is likely to have any meaningful performance impact.

akuegel:
> @akuegel I agree that this code could be refactored. Can we merge this for now and I'll look into the refactoring you suggested separately? I don't think the HloEvaluator invocation for the initial value is likely to have any meaningful performance impact.

sgtm

akuegel (approved):

jreiffers:
Thank you!

yliu120:
Yes. The code can be refactored and maybe redesigned in another PR.

Let's say if we can evaluate the induction var or the step or the max. We can actually just *materialize* them as a constant instruction in HLO.

PR #27816, opened 2025-06-13
 [ROCm] fixed broadcast_constant_block_dim_limit.hlo.test on rocm

amd-songpiao:
This is a fix to the failed //xla/backends/gpu/codegen/emitters/tests:loop/broadcast_constant_block_dim_limit.hlo.test.

@hsharsha could you first review my PR, please?


hsharsha:
@amd-songpiao  This effectively reverts the patch I had put in. It is the same code for both ROCm and CUDA. This was done https://github.com/openxla/xla/pull/19582

Owing to checks in https://github.com/openxla/xla/blob/main/xla/service/gpu/parallel_loop_emitter.cc#L169-L171 launch dimension can be of the form ((block.x, 1, 1), (thread.x, thread.y, 1)). And in ROCm it is expected that (block.x * thread.x) <= 0xFFFFFFFF

PR #27836, opened 2025-06-13
Add support for NVSHMEM put/get

sfvaroglu:


xla/backends/gpu/collectives/nvshmem_communicator.cc
```diff
@@ -442,7 +442,6 @@ absl::Status NvshmemCommunicator::RegisterBuffer(void* addr, size_t length) {
     LOG(ERROR) << absl::StrFormat(
         "Failed to register NVSHMEM buffer at %p with length %zu", addr,
         length);
-    return absl::InternalError("Failed to register NVSHMEM buffer");
```
PatriosTheGreat, line 445: Why this line was removed?

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
```diff
@@ -125,6 +125,25 @@ absl::Status IsValidNvshmemOperand(Shape shape, Thunk::Kind reduction_op);
 
 absl::StatusOr<xla::gpu::GpuCollectives*> GetNvshmemCollectivesFromRegistry();
 
+// NvshmemBufferAddresses provides a mechanism to store and retrieve buffer
+// addresses for NVSHMEM put/get operations. This is necessary because NVSHMEM
+// uses one-way put/get operations where both source and destination addresses
+// are required.
+struct NvshmemBufferAddresses {
```
allanrenucci: Prefer using a class: https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes

xla/backends/gpu/runtime/nvshmem_collective_thunk.cc
```diff
@@ -53,10 +53,31 @@ limitations under the License.
...
+}
+
+void NvshmemBufferAddresses::StoreNvshmemPtr(int device_ordinal,
+                                             void* buffer_addr) {
+  absl::MutexLock lock(&mu);
+  buffer_addrs[device_ordinal] = buffer_addr;
+}
+
+NvshmemBufferAddresses& g_nvshmem_buffer_addresses =
+    *new NvshmemBufferAddresses();
```
allanrenucci: Prefer the following pattern for declaring global objects:
```c++
NvshmemBufferAddresses& GlobalNvshmemBufferAddresses() {
  static NvshmemBufferAddresses* kAddresses = new NvshmemBufferAddresses();
  return *kAddresses;
}
```

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
```diff
@@ -0,0 +1,170 @@
...
+    }
+  } else {
+    VLOG(3) << "No source ID found, skipping Get operation";
+  }
+
+  return absl::OkStatus();
+}
+
+}  // namespace gpu
+}  // namespace xla
```
allanrenucci: Missing line at EOD

xla/backends/gpu/runtime/nvshmem_get_thunk.h
```diff
@@ -0,0 +1,57 @@
...
+  const NvshmemP2PConfig config_;
+  const CollectiveThunk::Buffer buffer_;
+  std::shared_ptr<NvshmemP2PExecutionCounters> execution_counters_;
+  std::string hlo_name_;
+};
+
+}  // namespace gpu
+}  // namespace xla
+
+#endif  // XLA_SERVICE_GPU_NVSHMEM_GET_THUNK_H_
```
allanrenucci: Missing line at EOD

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
```diff
@@ -0,0 +1,178 @@
...
+    }
+  } else {
+    VLOG(3) << "No target ID found, skipping Put operation";
+  }
+
+  return absl::OkStatus();
+}
+
+}  // namespace gpu
+}  // namespace xla
```
allanrenucci: Missing line at EOF

xla/backends/gpu/runtime/nvshmem_put_thunk.h
```diff
@@ -0,0 +1,55 @@
...
+  const NvshmemP2PConfig config_;
+  const CollectiveThunk::Buffer buffer_;
+  std::shared_ptr<NvshmemP2PExecutionCounters> execution_counters_;
+  std::string hlo_name_;
+};
+
+}  // namespace gpu
+}  // namespace xla
+
+#endif  // XLA_SERVICE_GPU_NVSHMEM_PUT_THUNK_H_
```
allanrenucci: Missing line at EOF

xla/backends/gpu/runtime/nvshmem_get_thunk.h
```diff
@@ -0,0 +1,57 @@
...
+
+ protected:
+  const CollectiveConfig& config() const override { return config_.config; }
+  absl::Status RunNvshmemCollective(const ExecuteParams& params,
+                                    se::Stream& stream) override;
+
+ private:
+  const NvshmemP2PConfig config_;
+  const CollectiveThunk::Buffer buffer_;
+  std::shared_ptr<NvshmemP2PExecutionCounters> execution_counters_;
```
allanrenucci: std::unique_ptr?

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
```diff
@@ -0,0 +1,178 @@
...
+                                 int64_t replica_count, int64_t partition_count,
+                                 const CollectiveThunk::Buffer& buffer)
+    : NvshmemCollectiveThunk(Thunk::kNvshmemPut, thunk_info,
+                             IsGPUSyncCollective(*instr)),
+      config_(GetNvshmemP2PConfigForPutGet(instr, instr->operand(0)->shape(),
+                                           replica_count, partition_count)),
+      buffer_(buffer),
+      execution_counters_(config_.validation_kind ==
+                                  NvshmemP2PConfig::ValidationKind::kConditional
+                              ? new NvshmemP2PExecutionCounters()
```
allanrenucci:
```suggestion
                              ? std::make_shared<NvshmemP2PExecutionCounters>()
```
Or `std::make_unique` if stored as `std::unique_ptr`

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
```diff
@@ -0,0 +1,178 @@
...
+    }
+  } else {
+    VLOG(3) << "No target ID found, skipping buffer registration";
+  }
+
+  if (target_id) {
+    bool should_run =
+        config_.validation_kind == NvshmemP2PConfig::ValidationKind::kInvalid
+            ? false
+            : true;
```
allanrenucci:
```suggestion
    bool should_run =
        config_.validation_kind != NvshmemP2PConfig::ValidationKind::kInvalid;
```

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
```diff
@@ -0,0 +1,178 @@
...
+          << ", current_id: " << current_id << ", group mode: "
+          << CollectiveOpGroupModeToString(config_.config.group_mode) << " ("
+          << hlo_name_ << ")";
+
+  const std::optional<int64_t> target_id = source_target.target;
+
+  VLOG(3) << absl::StreamFormat("%s : id = %d, target_id = %d", device_string,
+                                current_id, target_id.value_or(-1));
+
+  if (target_id) {
```
allanrenucci: Prefer early returns to avoid deeply nested if else branches. I.e.
```c++
if (!target_id) {
  VLOG() << ...;
  return absl::OkStatus();
}

if (!should_run) {
  VLOG() << ...;
  return absl::OkStatus();
}

...
```

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -2521,9 +2559,24 @@ absl::Status IrEmitterUnnested::EmitSendThunk(const HloSendInstruction* instr) {
       const HloInstruction* canonical_send_instr =
           FindCanonicalSendRecvStartOp(instr);
       if (collectives_async_events.contains(canonical_send_instr)) {
-        thunk->set_async_events(collectives_async_events[canonical_send_instr]);
+        if (IsNvshmemCollective(instr)) {
+          static_cast<NvshmemPutThunk*>(thunk.get())
```
allanrenucci: Let's replace these `static_cast` usages with [down_cast](https://github.com/openxla/xla/blob/5a073cb4d60c986479e2ba1ea933dff5425b9dc5/xla/tsl/platform/default/casts.h#L50);

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
```diff
@@ -0,0 +1,170 @@
...
+    VLOG(3) << "Storing destination device " << device_ordinal << " and buffer "
+            << buffer.destination_buffer.opaque();
+    g_nvshmem_buffer_addresses.StoreNvshmemPtr(
+        device_ordinal, buffer.destination_buffer.opaque());
+  } else {
+    VLOG(3) << "No source ID found, not storing buffer information";
+  }
+
+  // Get source buffer from source peer if needed.
+  if (source_id) {
```
allanrenucci: Prefer early returns to avoid deeply nested if else branches. I.e.
```c++
if (!source_id) {
  VLOG() << ...;
  return absl::OkStatus();
}

if (!should_run) {
  VLOG() << ...;
  return absl::OkStatus();
}

...
```

allanrenucci (changes requested):
Some C++ style comments. I'll defer to @PatriosTheGreat to review the logic.

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
sfvaroglu: Actually, in this case, we do not want to leave early.

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
sfvaroglu: Same, in this case, we do not want to leave early.

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
```diff
@@ -125,6 +125,30 @@ absl::Status IsValidNvshmemOperand(Shape shape, Thunk::Kind reduction_op);
...
+class NvshmemBufferAddresses {
+ public:
+  // Get buffer address for a device
+  absl::StatusOr<void*> GetNvshmemPtr(int device_ordinal);
+
+  // Store buffer address for a device
+  void StoreNvshmemPtr(int device_ordinal, void* buffer_addr);
+
+  // Get the global instance
+  static NvshmemBufferAddresses& GlobalNvshmemBufferAddresses() {
```
allanrenucci: Since this will be called as `NvshmemBufferAddresses::GlobalNvshmemBufferAddresses()` the name seems a bit repetitive. How about `NvshmemBufferAddresses::Global` or `NvshmemBufferAddresses::GlobalInstance`?

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
```diff
@@ -125,6 +125,30 @@ absl::Status IsValidNvshmemOperand(Shape shape, Thunk::Kind reduction_op);
...
+    static NvshmemBufferAddresses instance;
+    return instance;
```
allanrenucci: This is not sage initialisation of a static object because it is not trivially destructible: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables.

Hence why I suggested writing as:
```c++
static NvshmemBufferAddresses* instance = new NvshmemBufferAddresses();
return *instance;
```

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
allanrenucci: If I read the code correctly it is currently structured as follow:
```c++
if (source_id) {
  bool should_run = ...

  ...

  if (should_run) {
    ...
  } else {
    VLOG(3) << "Skipping Get operation";
  }
} else {
  VLOG(3) << "No source ID found, skipping Get operation";
}

return absl::OkStatus();
```
Is it not equivalent to:
```c++
if (!source_id) {
  VLOG(3) << "No source ID found, skipping Get operation";
  return absl::OkStatus();
}

bool should_run = ...
...

if (!should_run) {
  VLOG(3) << "Skipping Get operation";
  return absl::OkStatus();
}

...
```

xla/backends/gpu/runtime/nvshmem_get_thunk.h
allanrenucci: ping

allanrenucci (changes requested):

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
sfvaroglu: This change results in a deadlock in the test.

xla/backends/gpu/runtime/nvshmem_get_thunk.h
sfvaroglu: done

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
sfvaroglu: I'll take another look, but IIRC there are some edge cases.

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -87,17 +87,17 @@ absl::StatusOr<std::string> GetDataTypeString(xla::PrimitiveType data_type) {
   }
 }
 
-TEST(NvshmemGpuCollectivesTest, NvshmemCollectivePermuteFloat) {
-  constexpr int num_ranks = 2;
-
+void RunNvshmemTest(int num_ranks, PrimitiveType data_type,
+                    absl::string_view test_case) {
   tsl::SubProcess child[num_ranks];
```
PatriosTheGreat, line 92: Variable length array is not supported. std::vector can be used here instead, or you can inline 2 here instead of passing as a parameter

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -157,9 +169,25 @@ absl::Status NvshmemCollectiveTestBody(int rank_id, int num_ranks,
                 backend_config={"collective_backend_config":{"backend":"NVSHMEM"}}
           ROOT done = %s[] collective-permute-done(start)
         })",
-                                               data_type_str, data_type_str,
-                                               data_type_str, data_type_str);
-
+                               data_type_str, data_type_str, data_type_str,
+                               data_type_str);
+  } else if (test_case == "send_recv") {
+    kProgram = absl::StrFormat(R"(
```
PatriosTheGreat, line 174: Test is failing on out side with the following log at the rank 1 prcoess:

```
[1] NVSHMEM INFO mspace_malloc called with 2097152 bytes
[1] NVSHMEM INFO free chunk with size = 275773440 bytes found
[1] NVSHMEM INFO [1] type: 23nvshmemi_symmetric_heap allocated 2097152 bytes, 0 count, 0 alignment ptr: 0xa2f900800
nvshmem/src/host/mem/mem.cpp:buffer_register:67: Unable to register nvshmem heap memory. It is registered by default.

nvshmem/src/host/mem/mem.cpp:nvshmemx_buffer_register:240: Buffer addition for transport 0 failed with error 1

E0618 04:43:04.170032  140418 nvshmem_communicator.cc:442] Failed to register NVSHMEM buffer at 0xa2f900800 with length 4
E0618 04:43:04.170546  140418 pjrt_stream_executor_client.cc:2916] Execution of replica 1 failed: INTERNAL: Failed to register NVSHMEM buffer
xla/backends/gpu/collectives/nvshmem_communicator.cc:445
xla/backends/gpu/collectives/nvshmem_communicator.cc:353
xla/backends/gpu/collectives/nvshmem_communicator.cc:481
xla/backends/gpu/runtime/nvshmem_collective_thunk.cc:150
xla/backends/gpu/runtime/sequential_thunk.cc:100
xla/service/gpu/gpu_executable.cc:339
xla/service/gpu/gpu_executable.cc:891
xla/pjrt/gpu/se_gpu_pjrt_client.cc:2029
[1] NVSHMEM INFO mspace_free called on 0xa2f900800
[1] NVSHMEM INFO [1] freeing buf: 0xa2f900800 type: 23nvshmemi_symmetric_heap
```

xla/backends/gpu/runtime/nvshmem_collective_thunk.cc
```diff
@@ -193,5 +194,21 @@ absl::Status IsValidNvshmemOperand(Shape shape, Thunk::Kind reduction_op) {
...
+    int device_ordinal) {
+  absl::MutexLock lock(&mu_);
+  auto it = buffer_addrs_.find(device_ordinal);
+  if (it != buffer_addrs_.end()) {
+    return it->second;
+  }
+  return absl::NotFoundError("Buffer address not found for device");
+}
+
+void NvshmemBufferAddresses::StoreNvshmemPtr(int device_ordinal,
```
PatriosTheGreat, line 207: Will this work if several send/recv operations are working in parallel? This might happen since send-recv operations can be nested.

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
PatriosTheGreat: Keeping devices pointers for all local devices in each local process static memory is not an error-prone solution.  For example this would affect a use-case when the same XLA process is used to run several sequential models. Between those runs the static memory will not be cleaned which can lead to cross-model errors (when a recv operation from the next module uses a buffer stored by send operation of the previous module). We need to keep this registry somewhere in the context of gpu_executable::ExecuteThunks so this object is cleaned after execution is done. 

One possible solution is to pass a shared pointer to an instance of this registry to the nvshmem_get/put_thunk constructors and instantiate it in the ir_emmiter_unnested. So once the last thunk will be destructed the registry also will be destructed,

Another alternative is to add a registry to the ExecuteParams and instantiate it in the gpu_executable::ExecuteThunks. This solution seems to be a bit more complicated since ExecuteParams should need to keep mutated field relevant only for an nvshmem.

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
```diff
@@ -0,0 +1,169 @@
...
+namespace xla {
+namespace gpu {
+
+NvshmemGetThunk::NvshmemGetThunk(ThunkInfo thunk_info,
+                                 const HloRecvInstruction* instr,
+                                 int64_t replica_count, int64_t partition_count,
+                                 const CollectiveThunk::Buffer& buffer)
+    : NvshmemCollectiveThunk(Thunk::kNvshmemGet, thunk_info,
+                             IsGPUSyncCollective(*instr)),
+      config_(GetNvshmemP2PConfigForPutGet(instr,
```
PatriosTheGreat: NvshmemP2PConfig is a code-duplication of p2p_thunk_common::P2PConfig and the GetNvshmemP2PConfigForPutGet is the same as GetP2PConfigForSendRecv. Can you please switch to the P2PConfig instead?

NvshmemP2PGetSourceTargetPairs is the same as GetSourceTargetPairs

xla/backends/gpu/runtime/thunk.h
```diff
@@ -164,6 +164,10 @@ class Thunk {
     kNvshmemCollectivePermute,
     kNvshmemCollectivePermuteDone,
     kNvshmemCollectivePermuteStart,
+    kNvshmemGet,
```
PatriosTheGreat: Could you for the consistency call these methods Send/Recv instead of Get/Put? Also thunks as well.

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
```diff
@@ -0,0 +1,169 @@
...
+                                           instr->shape().tuple_shapes(0),
+                                           replica_count, partition_count)),
+      buffer_(buffer),
+      execution_counters_(config_.validation_kind ==
+                                  NvshmemP2PConfig::ValidationKind::kConditional
+                              ? std::make_unique<NvshmemP2PExecutionCounters>()
```
PatriosTheGreat: ExecutionCounters can be used instead since after which nvshmem_p2p_thunk_common can be removed.

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
```diff
@@ -0,0 +1,169 @@
...
+    TF_RETURN_IF_ERROR(execution_counters_->Initialize(
+        params.executor, params.collective_params->run_id));
+  }
+  return absl::OkStatus();
+}
+
+absl::Status NvshmemGetThunk::RunNvshmemCollective(const ExecuteParams& params,
+                                                   se::Stream& stream) {
+  TF_ASSIGN_OR_RETURN(auto* collectives, GetNvshmemCollectivesFromRegistry());
+  TF_ASSIGN_OR_RETURN(std::unique_ptr<Communicator> nvshmem_comm,
```
PatriosTheGreat: Could you please move these 2 lines down inside if (should_run)  block?

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
```diff
@@ -0,0 +1,176 @@
...
+    TF_RETURN_IF_ERROR(execution_counters_->Initialize(
+        params.executor, params.collective_params->run_id));
+  }
+  return absl::OkStatus();
+}
+
+absl::Status NvshmemPutThunk::RunNvshmemCollective(const ExecuteParams& params,
+                                                   se::Stream& stream) {
+  TF_ASSIGN_OR_RETURN(auto* collectives, GetNvshmemCollectivesFromRegistry());
+  TF_ASSIGN_OR_RETURN(std::unique_ptr<Communicator> nvshmem_comm,
```
PatriosTheGreat: Here also, could you please move these two lines above down in the if (should_run) block?

PatriosTheGreat (changes requested):

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
sfvaroglu: Updated

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
sfvaroglu: There is a difference between `!source` and `source_id.value_or(-1) == -1`. Simplified the code as much as possible.

xla/backends/gpu/runtime/nvshmem_collective_thunk.h
sfvaroglu: Updated

xla/backends/gpu/collectives/nvshmem_communicator.cc
sfvaroglu: Added back

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
sfvaroglu: Updated

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
sfvaroglu: Is this in a loop? It should be fixed now.

xla/backends/gpu/runtime/nvshmem_collective_thunk.cc
sfvaroglu: I don't think NCCL/NVSHMEM provides a way to specify explicit kernel launch dependency, so we cannot do interleaved p2p.

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
sfvaroglu: Done

xla/backends/gpu/runtime/thunk.h
sfvaroglu: Done

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
sfvaroglu: Done

xla/backends/gpu/runtime/nvshmem_get_thunk.cc
sfvaroglu: Done

xla/backends/gpu/runtime/nvshmem_put_thunk.cc
sfvaroglu: Done

xla/backends/gpu/collectives/nvshmem_communicator.cc
```diff
@@ -33,14 +35,14 @@ limitations under the License.
...
-#include "xla/tsl/platform/statusor.h"
-#include "xla/util.h"
 #include "xla/xla_data.pb.h"
-#include "tsl/platform/casts.h"
 
 namespace xla::gpu {
 
+// Used for global NVSHMEM buffer registration tracking
+std::mutex NvshmemCommunicator::registered_buffers_mutex_;
+std::unordered_set<void*> NvshmemCommunicator::registered_buffers_;
```
PatriosTheGreat: Could you please use absl::flat_hash_set instead?

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
PatriosTheGreat: Yes it works now, thanks!

xla/backends/gpu/collectives/nvshmem_communicator.h
```diff
@@ -138,6 +140,10 @@ class NvshmemCommunicator : public Communicator {
 
   NvshmemCollectives* collectives_;  // Parent NvshmemCollectives instance
   bool aborted_ = false;             // Has Abort() been called?
+
+  // Global tracking of registered buffers to avoid re-registration
+  static std::mutex registered_buffers_mutex_;
+  static std::unordered_set<void*> registered_buffers_;
```
PatriosTheGreat: This static array has the same issue as holding an address space in static area. An XLA process can be used to run several modules (or the same module several times). In this case eventually this set will contain all possible addresses and XLA will skip registering a re-allocated memory. If this is not expected (and I assume nvshmem needs to re-registed the memory if it was re-allocated) then we need to make something like RegisteredBufferRegistry, store it somewhere near to NvshmemBufferAddresses and pass it through Communicator instantiation interface.

PatriosTheGreat (changes requested):

xla/backends/gpu/collectives/nvshmem_communicator.h
sfvaroglu: NVSHMEM's buffer registration is global, per communicator registration will not work as intended--it will be global from NVSHMEM's point of view. If we run the same module several times, the current implementation avoids re-registration (NVSHMEM raises an error if we try to register the same buffer twice). Is there a unit case for the issue you mentioned?

PatriosTheGreat: Ah I see that NCCL memory registration implementation is currently similar: https://github.com/openxla/xla/blob/main/xla/backends/gpu/runtime/collective_thunk.cc#L351

So the multiple HLOs execution issue might be relevant right now for both NVSHMEM and NCCL. I think we can fix it in separate PR then.

For context  an issue can be reproduced for example with hlo runner using num_repeats parameter like this:
```
export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:~/nvshmem_src/src/lib"
export BENCHMARK_DIR="~/nvshmem_benchmarks"

# Should run once
# ./configure.py --backend=CUDA --cuda_compiler CLANG --nccl --cuda_compute_capabilities sm_90


xla_flags="--xla_gpu_experimental_enable_nvshmem=true"
hlo_file="$BENCHMARK_DIR/send_recv.hlo"
for i in $(seq 0 1);
do
    CUDA_VISIBLE_DEVICES=$i XLA_FLAGS="$xla_flags" bazel run -c opt --config=cuda --copt="-Wno-deprecated-declarations" --copt="-Wno-invalid-offsetof" //xla/tools/multihost_hlo_runner:hlo_runner_main -- \
        --use_spmd_partitioning=true --num_replicas=2 --num_partitions=1 --run_xla_backend_only=true --disable_all_hlo_passes=true --device_type=gpu \
        --address=127.0.0.1:12345 --num_nodes=2 --task_id=$i --num_repeats=10 # THIS one \
        $hlo_file
done
wait
```

With each iteration we should see more memory addresses in the registered set.

There is another problem here which I hit trying to reproduce this one. I hit this issue while running the HLO module with send_recv (the same which is used in the test):
```
I0626 09:23:26.447170  115723 nvshmem_collectives.cc:119] Initialized NVSHMEM on process 0; num_processes=2
I0626 09:23:26.447198  115723 nvshmem_collectives.cc:143] Start allocation of 2.00MiB (2097152 bytes) for NVSHMEM
I0626 09:23:26.447228  115723 nvshmem_collectives.cc:152] Allocated 2.00MiB (2097152 bytes). Address: 0x32f900800 from NVSHMEM
I0626 09:23:27.578686  115723 nvshmem_communicator.cc:145] Created NvshmemCommunicator(nvshmem_team_t=2)
I0626 09:23:27.578722  115723 nvshmem_communicator.cc:166] Barrier NVSHMEM communicator: NvshmemCommunicator(nvshmem_team_t=2)
I0626 09:23:27.578787  115723 nvshmem_communicator.cc:145] Created NvshmemCommunicator(nvshmem_team_t=2)
I0626 09:23:27.578795  115723 nvshmem_communicator.cc:166] Barrier NVSHMEM communicator: NvshmemCommunicator(nvshmem_team_t=2)
I0626 09:23:27.578809  115723 gpu_executable.cc:401] Join thunks initialization rendezvous with 1 local participants; device_ordinal=0
I0626 09:23:27.578824  115723 gpu_executable.cc:337] [0] Start GpuExecutable::ExecuteOnStream module: NvshmemSendRecv
I0626 09:23:27.578878  115723 nvshmem_communicator.cc:145] Created NvshmemCommunicator(nvshmem_team_t=2)
I0626 09:23:27.578885  115723 nvshmem_communicator.cc:485] Send NVSHMEM communicator: NvshmemCommunicator(nvshmem_team_t=2)
I0626 09:23:27.578889  115723 nvshmem_communicator.cc:359] Registering source buffer for send operation: 0x32f900b00
I0626 09:23:27.578892  115723 nvshmem_communicator.cc:451] Registering NVSHMEM buffer: 0x32f900b00, length: 4
I0626 09:23:27.578905  115723 nvshmem_communicator.cc:456] Registered buffers size: 0

Error:
nvshmem/src/host/mem/mem.cpp:buffer_register:67: Unable to register nvshmem heap memory. It is registered by default.

nvshmem/src/host/mem/mem.cpp:nvshmemx_buffer_register:240: Buffer addition for transport 0 failed with error 1
```

Looks like we are trying to register the memory marked as a heap memory for an NVSHMEM.
Is this an issue? Should we just ignore this type of errors when registring the buffer?

sfvaroglu: Oh, it might be due to https://github.com/openxla/xla/pull/28190. I'll look into it.

sfvaroglu: We don't need the registration logic anymore. With https://github.com/openxla/xla/pull/28190, the colorer colors buffers around collectives, so they are automatically allocated in nvshmem memory.

xla/backends/gpu/collectives/nvshmem_communicator.h
```diff
@@ -138,6 +138,10 @@ class NvshmemCommunicator : public Communicator {
 
   NvshmemCollectives* collectives_;  // Parent NvshmemCollectives instance
   bool aborted_ = false;             // Has Abort() been called?
+
+  // Global tracking of registered buffers to avoid re-registration
+  static std::mutex registered_buffers_mutex_;
+  static absl::flat_hash_set<void*> registered_buffers_;
```
allanrenucci: Unused?

xla/backends/gpu/runtime/nvshmem_recv_thunk.cc
```diff
@@ -0,0 +1,167 @@
...
+  if (source_id.value_or(-1) == -1) {
+    VLOG(3) << "Storing destination device " << device_ordinal << " and buffer "
+            << buffer.destination_buffer.opaque();
+    buffer_addresses_->StoreNvshmemPtr(device_ordinal,
+                                       buffer.destination_buffer.opaque());
+  }
+
+  if (source_id) {
+    bool should_run =
+        config_.validation_kind != P2PConfig::ValidationKind::kInvalid;
```
allanrenucci: I assume this should be moved outside the if statement

xla/backends/gpu/runtime/nvshmem_recv_thunk.h
```diff
@@ -0,0 +1,59 @@
...
+class NvshmemRecvThunk : public NvshmemCollectiveThunk {
+ public:
+  NvshmemRecvThunk(ThunkInfo thunk_info, const HloRecvInstruction* inst,
+                   int64_t replica_count, int64_t partition_count,
+                   const CollectiveThunk::Buffer& buffer,
+                   std::shared_ptr<NvshmemBufferAddresses> buffer_addresses);
+  absl::Status Initialize(const InitializeParams& params) override;
+
+  // Returns the group mode (cross-replica or cross-partition) for the operation
+  static CollectiveOpGroupMode GetGroupMode(const HloRecvInstruction* inst);
```
allanrenucci: I don't see a corresponding implementation in nvshmem_recv_thunk.cc. Is this used? Remove?

xla/backends/gpu/runtime/nvshmem_recv_thunk.cc
```diff
@@ -0,0 +1,167 @@
...
+  // For receive operations, we need to register the buffer with NVSHMEM
+  // and store it in the global struct for the send operation to use
+  if (source_id.value_or(-1) == -1) {
+    VLOG(3) << "Storing destination device " << device_ordinal << " and buffer "
+            << buffer.destination_buffer.opaque();
+    buffer_addresses_->StoreNvshmemPtr(device_ordinal,
+                                       buffer.destination_buffer.opaque());
+  }
+
+  if (source_id) {
```
allanrenucci: Let's change to early return, i.e.

```cpp
if (!source_id) {
  VLOG(3) << "No source ID found, skipping Recv operation";
  return absl::OkStatus()
}
```

xla/service/gpu/ir_emitter_unnested.cc
```diff
@@ -182,6 +184,8 @@ limitations under the License.
 #include "xla/xla_data.pb.h"
 #include "tsl/platform/human_readable_json.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
+#include "xla/backends/gpu/runtime/nvshmem_recv_thunk.h"
+#include "xla/backends/gpu/runtime/nvshmem_send_thunk.h"
```
allanrenucci: Duplicated?

xla/service/gpu/ir_emitter_unnested.h
```diff
@@ -51,6 +51,8 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+class NvshmemBufferAddresses;
```
allanrenucci: Let's include the relevant header instead.

allanrenucci (changes requested):

xla/backends/gpu/collectives/nvshmem_communicator.h
sfvaroglu: Removed

xla/backends/gpu/runtime/nvshmem_recv_thunk.h
sfvaroglu: Removed

xla/backends/gpu/runtime/nvshmem_recv_thunk.cc
sfvaroglu: Done

xla/service/gpu/ir_emitter_unnested.h
sfvaroglu: Done

xla/service/gpu/ir_emitter_unnested.cc
sfvaroglu: Removed

PatriosTheGreat (approved):
Thanks for addressing all issues and for the change!

allanrenucci (approved):

PR #27983, opened 2025-06-18
[ROCm] Fix rocm ci build break

alekstheod:
This fixes rocm CI build compilation errors introduced in the hermetic cc toolchain change.
https://github.com/openxla/xla/pull/27789

### DIFF
--- a/tensorflow.bazelrc
+++ b/tensorflow.bazelrc
@@ -229,6 +229,7 @@ build:asan --copt -g
 build:asan --copt -fno-omit-frame-pointer
 build:asan --linkopt -fsanitize=address
 
+build:rocm_base --config=clang_local
 build:rocm_base --copt=-Wno-gnu-offsetof-extensions
 build:rocm_base --crosstool_top=@local_config_rocm//crosstool:toolchain
 build:rocm_base --define=using_rocm_hipcc=true


alekstheod:
@xla-rotation please review!

tensorflow.bazelrc
```diff
@@ -229,6 +229,7 @@ build:asan --copt -g
 build:asan --copt -fno-omit-frame-pointer
 build:asan --linkopt -fsanitize=address
 
+build:rocm_base --config=clang_local
```
dimitar-asenov, line 232: I'm a little confused here. This setting seems to be clang-specific, but `rocm_base` is also used for `rocm_gcc`. Have you tried compiling with gcc after this change?

alekstheod: rocm-gcc is obsolete and we do not use it anymore, I think it doesn't even work right now.
I believe clang_local is misleading, it should be something like local_compiler.

alekstheod: We could also use these flags directly in our config but since we have that base config I thought we shall use it.

dimitar-asenov: OK, thanks for explaining.

dimitar-asenov (approved):

PR #27994, opened 2025-06-18
Multihost HLO Runner: Add CLI Option for Output Mode

Zoey-Cheng:
This PR adds a new CLI option `--output_mode` to **multihost_hlo_runner**, making the output mode configurable for the running function.

**Motivation:**
Previously, multihost_hlo_runner always returned outputs from all devices, which could cause unnecessary device-to-host transfers and increased memory usage. So here making output mode configurable. If outputs are not returned, outputs are still computed but the potentially slow device-to-host copy is skipped.

- Adds a new CLI option `--output_mode` to `multihost_hlo_runner`.
- Supports three modes:
  - `return_outputs` (default): Return outputs from all devices.
  - `not_return_outputs`: Do not return any outputs (outputs are computed but not transferred to host).
  - `return_device_0_outputs`: Return outputs only from logical device 0.
- Invalid values trigger a clear error message.

**What it Change**

The CLI arg is registeredin `hlo_runner_main.cc` with a notification. It call existing parsing logic from `functional_hlo_runner.h` (line 173), error logic form `functional_hlo_runner.cc`.

**Self Test**
- Try arg return_outputs, not_return_outputs, return_device_0_outputs; empty; error input. Search "Outputs for" in the log.
- Try test notification in "help"

`sample.hlo`

```
HloModule test

ENTRY main {
    p0 = f32[8,8] parameter(0)
    p1 = f32[8,8] parameter(1)
    ROOT add = f32[8,8] add(p0,p1)
}

```

Running Demo

```
bazel run xla/tools/multihost_hlo_runner/hlo_runner_main -- sample.hlo --log_output=true --output_mode=return_device_0_outputs

bazel run xla/tools/multihost_hlo_runner/hlo_runner_main -- --help
```


xla/tools/multihost_hlo_runner/hlo_runner_main.cc
```diff
@@ -163,6 +164,12 @@ RunningOptionsFromFlags(const HloRunnerConfig& opts) {
...
   TF_ASSIGN_OR_RETURN(out.module_argument_mode,
                       ArgumentModeFromString(opts.hlo_argument_mode));
+  // add parser for output mode (from functional_hlo_runner.h::AbslParseFlag)
+  if (std::string error; !FunctionalHloRunner::AbslParseFlag(
+          opts.output_mode_str, &out.module_output_mode, &error)) {
+    return absl::InvalidArgumentError(absl::StrCat(
+        "Invalid --output_mode: ", opts.output_mode_str, ". ", error));
+  }
 
   out.module_output_mode =
```
reedwm: This line will overwrite the parsed `module_output_mode`.

reedwm (changes requested):
What's the motivation for the change? Why not always return all outputs?

Also what does it mean to "return outputs" or to "not return outputs"? The HLO module must compute all outputs in any case. What are the outputs returned from?

sergachev:
> What's the motivation for the change? Why not always return all outputs?

This can be used to speed up debugging and benchmarking by avoiding device to host copies at the end of programs when output values are not verified but just discarded on the host either way. We already have and actively use the similar --hlo_argument_mode=uninitialized for the inputs.

> Also what does it mean to "return outputs" or to "not return outputs"? The HLO module must compute all outputs in any case. What are the outputs returned from?

Doing or not DtoH copies. This does not affect computing them on GPU.

reedwm:
That makes sense. @Zoey-Cheng, in the flag description, can you give details on what the flag does and the motivation? E.g. "Specify whether outputs are returned or not. Possible values: return_outputs, not_return_outputs, return_device0_outputs. If outputs are not returned, the outputs are still computed but a potentially-slow device-to-host copy is skipped."

Zoey-Cheng:
Sorry for the mistake I've made. I'll correct this, update the description, and check the output for each one

Zoey-Cheng:
I've updated and tested the result again in both local and gitlab MR, seems ok. Could you help review that?

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
```diff
@@ -163,9 +164,24 @@ RunningOptionsFromFlags(const HloRunnerConfig& opts) {
   FunctionalHloRunner::RunningOptions out;
   TF_ASSIGN_OR_RETURN(out.module_argument_mode,
                       ArgumentModeFromString(opts.hlo_argument_mode));
+  // add parser for output mode (from functional_hlo_runner.h::AbslParseFlag)
```
reedwm: Why not just call AbslParseFlag?

If there is a good reason, I'd remove this comment, since it doesn't add much value.

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
```diff
@@ -419,7 +435,15 @@ int main(int argc, char** argv) {
...
                 "proto message. The order of precedence: command line flags > "
-                "XLA_FLAGS > debug_options_file > default flags.")};
+                "XLA_FLAGS > debug_options_file > default flags."),
+      tsl::Flag(
+          "output_mode", &opts.output_mode_str,
+          "Specify whether outputs are returned after execution. "
+          "Possible values: return_outputs (default), not_return_outputs, "
+          "return_device0_outputs (return outputs only from logical device 0). "
+          "If outputs are not returned, outputs are still computed but the "
+          "potentially slow device-to-host copy is skipped. "),
```
reedwm: Remove the space at the end of the comment

reedwm (changes requested):

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
Zoey-Cheng: Changed this. Now directly call parser in `functional_hlo_runner.h`, and generate error through `functional_hlo_runner.cc`. Retest this

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
Zoey-Cheng: Deleted. Also tested this through `--help`

Zoey-Cheng:
Addressed previous comments:
- Replaced the explicit parser with a call to the parser defined in `functional_hlo_runner.h`, and error handling with the error handling implemented in `functional_hlo_runner.cc`.
- Retested with various parameter values and verified the `--help` output.
Requesting another review. Thank you!

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
```diff
@@ -163,9 +164,17 @@ RunningOptionsFromFlags(const HloRunnerConfig& opts) {
   FunctionalHloRunner::RunningOptions out;
   TF_ASSIGN_OR_RETURN(out.module_argument_mode,
                       ArgumentModeFromString(opts.hlo_argument_mode));
+  // add parser for output mode (matching from
+  // functional_hlo_runner.h::AbslParseFlag, error generate from
+  // functional_hlo_runner.cc)
```
reedwm: I don't think the comment is useful. It doesn't "match from functional_hlo_runner.h::AbslParseFlag" but directly calls that function. I would remove this comment.

reedwm (changes requested):

xla/tools/multihost_hlo_runner/hlo_runner_main.cc
Zoey-Cheng: Get. I deleted this

reedwm (approved):

PR #28184, opened 2025-06-24
Extend `WhileLoopAllReduceCodeMotion` pass with a new pattern (DUS)

sergey-kozub:
We're seeing a pattern in Llama3 model (fp8) where all-reduce happens inside the loop, but the results are only used outside of the loop. When there are many participating devices, this could result in a slowdown.

The pass now also matches all-reduces that are scattered into the loop output using "dynamic-update-slice" op with the loop induction variable as the index parameter.


akuegel (approved):

golechwierowicz:
Wait, it is not always profitable to sink all-reduces outside of the while loop. The point of keeping ARs in the loops is to amortize the cost of gradient accumulation and unlock compute-communication overlap. For example: if you pipeline ARs then when you run iteration `i` you can fully overlap `i-1` AR.

Sinking prevents that and can make models slower.

golechwierowicz (changes requested):
IMO, this is not always profitable. See my other comment.

sergey-kozub:
> Wait, it is not always profitable to sink all-reduces outside of the while loop. The point of keeping ARs in the loops is to amortize the cost of gradient accumulation and unlock compute-communication overlap. For example: if you pipeline ARs then when you run iteration `i` you can fully overlap `i-1` AR.

Running a single AR with more data should be faster than running N ARs with less data.
The AR after the loop could still be overlapped with other instructions.

As an alternative, we could introduce a threshold (e.g. 1Mb) and do not hoist the ARs above that size.
Note: the FP8 Llama3 case we're trying to fix here exchanges one element (i.e. 4 bytes) in each loop iteration.

golechwierowicz:
> Running a single AR with more data should be faster than running N ARs with less data.
The AR after the loop could still be overlapped with other instructions.

1) That is not true if you combined collectives up to certain threshold such that you saturate networking bandwidth.
2) There is usually no compute left after transformer layers finish running, the only thing left is updating the optimizer state which is very lightweight.

> As an alternative, we could introduce a threshold (e.g. 1Mb) and do not hoist the ARs above that size.
Note: the FP8 Llama3 case we're trying to fix here exchanges one element (i.e. 4 bytes) in each loop iteration.

If that's the case the threshold should be super small, but first I'd like to know about the case you are optimizing for.  Is it MaxText configuration? If so, can you share the config?

sergey-kozub:
> 1. That is not true if you combined collectives up to certain threshold such that you saturate networking bandwidth.

In our use case, the AR size is small enough, so that T(1)~=T(N)

> > As an alternative, we could introduce a threshold (e.g. 1Mb) and do not hoist the ARs above that size.

> If that's the case the threshold should be super small, but first I'd like to know about the case you are optimizing for.

Please suggest a threshold value then.

> Is it MaxText configuration? If so, can you share the config?

Yes, this is MaxText Llama3-8b. I don't have more specific details.

sergey-kozub:
I've added a size threshold, PTAL.

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -141,6 +151,166 @@ HloInstruction* GetEffectiveScalar(HloInstruction* instruction) {
...
+  if ((*loop_bound + 1) * ShapeUtil::ArraySize(all_reduce->shape()) >
+      kAllReduceMaxSizeBytes) {
+    VLOG(5) << "Resulting all-reduce exceeds the size threshold: "
+            << kAllReduceMaxSizeBytes;
+    return std::nullopt;
+  }
+
+  // Verify that the instruction is an iterator over the loop range.
+  // Allow reverse indexing (i.e. back-to-front).
+  auto is_induction_variable = [&](const HloInstruction* instr) {
```
golechwierowicz, line 207: Isn't this machinery repeating essentially everything that is available/being used in `dynamic_slice_fusion_rewriter`? We also match ops to dynamic slices there.

golechwierowicz (changes requested):
I still don't understand why we need it, to achieve overlap you have two options:

1. On XLA level you can pipeline ARs and let LHS overlap and (optionally) let combiners combine small AR with bigger ARs.
2. On JAX level you can unroll the loop and let LHS overlap and (optionally) let combiners combine the ARs.

xla/service/while_loop_all_reduce_code_motion.cc
sergey-kozub: We need this change in order to fix the regression when running an FP8 MaxText config with a large number of hosts. The AR is essentially exchanging the tensor global maximum (single value) between the hosts inside the loop. This AR is not picked up by the combiners.

I don't see how `dynamic_slice_fusion_rewriter` is related to this pass. Reading the comments in `dynamic_slice_fusion_rewriter.h` doesn't help - it's not clear why this is needed (wrapping slices into a fusion) or how it is related to the current PR.

golechwierowicz: Is it possible to re-use this? https://github.com/openxla/xla/blob/af2a3de2dd990ee818dbafa0cf4864732456d310/xla/service/gpu/dynamic_slicing_utils.h#L25-L30

This matches any op to a subsequent DUS and makes sure DUS operates only on induction variables. This is exactly what you're doing, right?

sergey-kozub: This code is indeed similar but it doesn't quite fit. For example, I need to match the update operand of DUS only, not just any DUS user.

golechwierowicz: Ok, but you can modify that code a little bit. We're trying to minimize the added code to the codebase here.

It makes more sense to modify a little bit something that mostly works, rather than rollout something on a side which is slightly different. Former adds little loc, latter adds more and duplicates it.

sergey-kozub: "A little bit" wouldn't cut it in this case. I need to significantly modify the code in `dynamic_slicing_utils.h` - it serves a very specific purpose and is not reusable in this case.

I will need to add more function parameters to `GetSlicedUserPaths` function to make it serve two purposes, which will make it harder to read and understand.

Please assign an additional reviewer to arbitrate. If they say to do it your way, I'll do the necessary adjustments.

golechwierowicz: I already modified it in https://github.com/openxla/xla/commit/33495f5f60d25af4ee9c425df460d30132885d0d, and just checking the operand should not pollute it much.

pifon2a: @sergey-kozub I think after Greg's changes it wouldn't require significant adjustments to the dynamic_slicing_utils.h. Please, adjust the code.

sergey-kozub: Updated the PR, now it uses a `dynamic_slicing_utils.h` helper.

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -141,6 +153,175 @@ HloInstruction* GetEffectiveScalar(HloInstruction* instruction) {
...
+    if (Match(instr, m::Subtract(m::ConstantScalar(&left), m::Op(&right))) &&
+        left->literal().GetFirstInteger().value() == *loop_bound) {
+      instr = right;
+    }
+    return Match(instr, m::GetTupleElement(m::Parameter(0), *indvar_idx));
+  };
+
+  // Do not hoist all-reduce ops if the resulting all-reduce is too large.
+  // The threshold value is chosen arbitrarily.
+  const int64_t kAllReduceMaxSizeBytes = 10240;
```
golechwierowicz: Change to constexpr.

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -141,6 +153,175 @@ HloInstruction* GetEffectiveScalar(HloInstruction* instruction) {
...
+         *loop_bound != range->max()->GetSignedValue())) {
+      VLOG(5) << "Only simple range loops are supported.";
+      return std::nullopt;
+    }
+    loop_bound = range->max()->GetSignedValue();
+  }
+
+  // Verify that the instruction is an iterator over the loop range.
+  // Allow reverse indexing (i.e. back-to-front).
+  auto is_induction_variable = [&](const HloInstruction* instr) {
```
golechwierowicz, line 207: After a while loop analysis there should be a WhileLoopBackendConfig attached to the while instruction and you should have a `known_induction_variable` which points to a tuple index. Isn't it enough?

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -141,6 +153,175 @@ HloInstruction* GetEffectiveScalar(HloInstruction* instruction) {
...
+    const CallGraph* call_graph) {
+  UpdateSliceContext context{/*update_slice_shape=*/all_reduce->shape(),
+                             /*update_slice_offsets=*/{},
+                             /*param_tuple_index=*/-1};
+
+  // Find the loop induction variable and match the loop range.
+  std::optional<int64_t> indvar_idx;
+  std::optional<int64_t> loop_bound;
+
+  for (HloInstruction* while_op : while_instructions) {
```
golechwierowicz, line 180: Can't you get a loop range you require from `WhileLoopBackendConfig`?

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -141,6 +153,175 @@ HloInstruction* GetEffectiveScalar(HloInstruction* instruction) {
...
+        context.update_slice_shape.set_element_type(
+            hlo->shape().element_type());
+        break;
+      default:
+        return std::nullopt;
+    }
+  }
+
+  // DUS ops may form a chain.
+  const HloInstruction* prev =
```
golechwierowicz, line 257: Is this true in practice that you can have such a long chain of dynamic update slices? DUS (DUS (DUS ...)))?

xla/service/while_loop_all_reduce_code_motion.cc
sergey-kozub: Do you mean the `WhileLoopTripCountAnnotator` pass? It runs after the current pass in the compilation pipeline, so not possible.

xla/service/while_loop_all_reduce_code_motion.cc
sergey-kozub: Ditto.

xla/service/while_loop_all_reduce_code_motion.cc
sergey-kozub: In the specific model we're trying to optimize there is a chain of two DUS.

xla/service/while_loop_all_reduce_code_motion.cc
golechwierowicz: I see, thanks!

golechwierowicz (approved):

dimitar-asenov:
@sergey-kozub We had to disable this change, because it broke an internal test. Here's the disable commit:

https://github.com/tensorflow/tensorflow/commit/f562379892537e9aeeda2aafc6196743905d4044

The breaking test is in JAX where grad of scan of allgather fails to compile on TPU (we haven't yet tried GPU or CPU). Xla fails with a RET_CHECK failure:

```
JaxStackTraceBeforeTransformation:
jax.jaxlib._jax.XlaRuntimeError: INTERNAL:
RET_CHECK failure (xla/hlo/ir/hlo_computation.cc:1615)
ShapeUtil::Compatible(old_instruction->shape(), new_instruction->shape()) "reduce-scatter.7" (f32[128]{0}) vs "multiply.15" (f32[1024]{0})
```

Here is the repro JAX:

```
import functools

import jax
import jax.numpy as jnp
import jax.random as jrandom
import jax.sharding as jsharding

mesh = jsharding.Mesh(jax.devices(), ('i',))

rng = jrandom.key(33)
x = jrandom.normal(rng, (4, 1024))
x_spec = jax.P(None, 'i')
x = jax.device_put(x, jsharding.NamedSharding(mesh, x_spec))

@functools.partial(
    jax.shard_map,
    mesh=mesh,
    in_specs=(x_spec,),
    out_specs=x_spec,
    check_vma=False,
)
@jax.grad
def dummy_loss(xs):
  def scan_fn(carry, x):
    x = jax.lax.all_gather(x, 'i', axis=0, tiled=True)
    new_carry = carry * x
    return new_carry, None

  init_carry = jnp.ones((1024,), dtype=xs.dtype)
  y, _ = jax.lax.scan(scan_fn, init_carry, xs)
  return jnp.square(y).mean()

dummy_loss(x)
```

Could you please look into this and see if you can debug the issue? Thanks!

sergey-kozub:
@dimitar-asenov Could you please provide the before-optimizations HLO?

I'm unable to reproduce the issue with the above JAX snippet on an 8-GPU instance.

Also, I only see one reference to `WhileLoopAllReduceCodeMotion`, in the `gpu_compiler.cc` - so I'm not sure how this could affect TPU.
Note: the changes to `dynamic_slicing_utils` shouldn't affect anything else (it only adds an argument with a default).

akuegel:
> @dimitar-asenov Could you please provide the before-optimizations HLO?
> 
> I'm unable to reproduce the issue with the above JAX snippet on an 8-GPU instance.
> 
> Also, I only see one reference to `WhileLoopAllReduceCodeMotion`, in the `gpu_compiler.cc` - so I'm not sure how this could affect TPU. Note: the changes to `dynamic_slicing_utils` shouldn't affect anything else (it only adds an argument with a default).

The `WhileLoopAllReduceCodeMotion` is used in internal TPU code, so not possible for you to see. I fear this is something that needs to be debugged internally, in case only TPU is affected @dimitar-asenov

dimitar-asenov:
To close the loop on this, we fixed the issue and reenabled this functionality. The fix was to limit the code motion to `all-reduce` only and exclude `reduce-scatter`, which the existing pass supported if the it was created with `enable_reduce_scatter` set to `true`.

PR #28190, opened 2025-06-24
[NVIDIA GPU] Add copies for collective memory ops if they are consuming from constant or module inputs

Tixxx:
Some collective allocators like ncclMemAlloc or nvshmemAlloc require specific alignment of allocated memory to be able to use underlying hardware feature, if inputs are coming from module inputs or constants which are pre-allocated by default bfc ahead of time, it will sometimes fail during runtime due to invalid permission or other memory errors.
This is to add an intermediate copy for those scenarios so the copy will have the collective color and allocated by appropriate allocator.


PatriosTheGreat (approved):
Thanks!

PR #28295, opened 2025-06-26
[NVIDIA GPU] Do out of place allreduce for nvshmem

Tixxx:
NVSHMEM's default one-shot allreduce algo requires separate buffers for input and output.
In nvshmem 3.3, the heuristics will change to use 2-shot algo when input and output buffers are the same.
This is a workaround before 3.3 upgrade.


frgossen:
Can you add test cases for these changes?

Tixxx:
> Can you add test cases for these changes?

@frgossen added a test for the data flow change. The clang format error is strange, i cannot repro that locally.

PatriosTheGreat (approved):

PR #28304, opened 2025-06-26
[GPU] Relax memory space enforcement for tuple shapes

terryysun:
Tuple shapes don't need layout thus should be excused from memory space check. Currently the check will assume all values without layout has default memory space, this could cause crash when a tuple is assigned to non-default buffer allocation, for example NCCL user buffer runs are crashing on MaxText main. This PR fixes the crash.


xla/pjrt/pjrt_executable.cc
```diff
@@ -279,10 +279,9 @@ void CompiledMemoryStats::PopulateBufferStatsFromAllocations(
     int64_t alloc_memory_space = -1;
     for (const auto& [value, _] : alloc.assigned_buffers()) {
       const HloPosition& defining_position = value->defining_position();
-      int64_t memory_space = Layout::kDefaultMemorySpace;
-      if (defining_position.shape().has_layout()) {
-        memory_space = defining_position.shape().layout().memory_space();
-      }
+      if (defining_position.shape().IsTuple()) continue;
+      CHECK_EQ(defining_position.shape().has_layout(), true);
+      int64_t memory_space = defining_position.shape().layout().memory_space();
```
jprabhas, line 293: Do we need to add an assertion after the loop before line 293 that the alloc_memory_space is not -1 and is indeed populated by some defining_position

terryysun: `alloc_memory_space` will be -1 iff all values in the allocation are tuples. We should still include it for stats purpose, and given `memory_space` is only used for determining is host or not, -1 just means not to treat this allocation as host.

gspschmid: Do tuple shapes ever have a layout? If so, shouldn't we still factor in their memory_space when available?

Also, do you have an example of where this fails? Fwiw, a simple example mixing different memory spaces in the same tuple seems to work: https://gist.github.com/gspschmid/1ed5078c722d23cfc0256f151ab6163a

terryysun: Tuple shapes don't have layout. This would fail when a tuple shape value and an array shape value colored to non-default memory space are in the same allocation, I don't have a small repro as allocation can't be controlled easily.

terryysun:
Hi @frgossen, I've added execution test for this fix, could you take a look? Thanks!

xla/pjrt/pjrt_executable.cc
```diff
@@ -280,10 +280,9 @@ void CompiledMemoryStats::PopulateBufferStatsFromAllocations(
     int64_t alloc_memory_space = -1;
     for (const auto& [value, _] : alloc.assigned_buffers()) {
       const HloPosition& defining_position = value->defining_position();
-      int64_t memory_space = Layout::kDefaultMemorySpace;
-      if (defining_position.shape().has_layout()) {
-        memory_space = defining_position.shape().layout().memory_space();
-      }
+      if (defining_position.shape().IsTuple()) continue;
+      CHECK_EQ(defining_position.shape().has_layout(), true);
```
frgossen: nit: Why not use CHECK?

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -3776,5 +3776,72 @@ INSTANTIATE_TEST_SUITE_P(
...
+  args.push_back(LiteralUtil::CreateR1<float>({8, 7, 6, 5, 4, 3, 2, 1}));
+
+  std::vector<std::vector<Literal*>> replica_args;
+  for (int i = 0; i < kNumReplicas; ++i) {
+    replica_args.push_back({&args[0], &args[1]});
+  }
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::vector<Literal> results,
+      HloTestBase::ExecuteReplicated(std::move(module),
```
frgossen, line 4164: Can you confirm that this test case crashes w/o the above changes?

xla/pjrt/pjrt_executable.cc
frgossen: True, tuples dont have layouts niut but the values in the tuple do. I think the right thing would be to iterate over the tuple elements. 

Does the fact that PJRT has to deal with tuples mean that they escape the HLO module? Is this supported?

frgossen (changes requested):

xla/pjrt/pjrt_executable.cc
terryysun: updated.

xla/pjrt/pjrt_executable.cc
terryysun: Yes only elements need to be iterated. It's unclear why we would see tuples at this point but it should be generally harmless since tuples don't really occupy space.

xla/tests/collective_ops_e2e_test.cc
terryysun: It's more of a regression guard. The failure is only observed with a MaxText full model training. The allocation is derived by XLA's private allocator which can't be controlled directly, while indirectly influencing the decision is very tricky.

xla/pjrt/pjrt_executable.cc
frgossen: Still think skipping tuples here is dangerous. Can you either assert that defining_position is not a tuple or support and test the case?

xla/tests/collective_ops_e2e_test.cc
frgossen: We should strive for a self-contained test case in XLA. Can you reproduce the failure with HLO dumped from maxtext? If so, that could be a starting point for a test case.

frgossen (changes requested):
You're relaxing a memory space verification for tuples here, which, based on the comment was intentionally put in place. We need a test case for this.

xla/pjrt/pjrt_executable.cc
terryysun: This chunk of code is deciding the alloc_memory_space, and the only effect is to count those mem space for device or host. Currently a tuple-only allocation would always be on device, i.e. the default setup, and skipping would trigger default. I've also added a check to make sure no future custom colorer could put it on host, since there doesn't seem to bring any benefits for an allocation of pointers.

terryysun:
Thank you @frgossen! I've added a stronger test and verified it would fail without the fix and pass with it. Could you take another look?

xla/pjrt/pjrt_executable.cc
frgossen: Not an expert in this code but it is not clear to me why skipping tuples is correct. Based on this comment, it seems wrong: 

// All logical buffers assigned to a buffer allocation share a color.
// With buffer assigner's default colorer the color happens to be the
// memory space of the underlying HLO value. Callers may choose other
// colorers, however, e.g.:
// https://github.com/openxla/xla/blob/50c6489cb058881cc65622605c9c55029abebc5b/xla/service/gpu/compile_module_to_llvm_ir.cc#L152
// Until buffer allocations provide a stronger guarantee about colors,
// we sanity-check that the default coloring behavior was used.

If tuples are always on the device, why can we not iterate over the tuple here and apply the same logic as for any other buffer?

xla/tests/collective_ops_e2e_test.cc
frgossen: Can you add a test case that crashes w/o the above changes?

frgossen (changes requested):

xla/pjrt/pjrt_executable.cc
terryysun: we are still iterating over tuples here, this part of code is just to decide if the buffer should be counted as host memory or not. So skipping tuples here essentially means we always count them on device. And the added check guarantees this is the case for all customer colorers.

xla/tests/collective_ops_e2e_test.cc
terryysun: yes new test is added at https://github.com/openxla/xla/pull/28304/files#diff-ce4a6f9e304d9bac11e6cacffa07b57f584ad705009e41aded11b4146ae536fdR1782, it fails without the changes

xla/stream_executor/stream_executor.h
```diff
@@ -54,7 +54,7 @@ limitations under the License.
 namespace stream_executor {
 
 // Identifies the memory space where an allocation resides.
-enum class MemoryType { kDevice = 0, kUnified, kCollective, kHost = 5 };
+enum class MemoryType { kDevice = 0, kCollective = 1, kUnified = 2, kHost = 5 };
```
frgossen, line 57: Why change the order?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1779,6 +1779,81 @@ TEST(StreamExecutorGpuClientTest, ExecutablePinnedHostOutputMemoryKindTest) {
...
+    }
+  )";
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto executable,
+      CompileExecutable(kProgramWithCollectiveAndTuple, *client, options));
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto memory_stats, executable->GetExecutable()->GetCompiledMemoryStats());
+  EXPECT_GE(memory_stats.output_size_in_bytes, 0);
```
frgossen, line 1852: Can you add the exact expected sizes here to help understand if it correct to skip tuples in the impl?

xla/tests/collective_ops_e2e_test.cc
frgossen: Thanks for adding the test! This helps a lot in understanding this change

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1779,6 +1779,81 @@ TEST(StreamExecutorGpuClientTest, ExecutablePinnedHostOutputMemoryKindTest) {
...
+      ar0 = f32[1024]{0} all-reduce(p4), channel_id=5, to_apply=region_0
+      ar1 = f32[1]{0} all-reduce(p5), channel_id=6, to_apply=region_0
+      
+      // Regular operations with default memory space
+      add0 = f32[512,128]{1,0} add(p0, p0)
+      add1 = f32[512,32,128]{2,1,0} add(p1, p1)
+      add2 = f32[512,8,128]{2,1,0} add(p2, p2)
+      
+      // Mix of all-gather results (memory space 1) and regular tensors (memory space 0)
+      ROOT tuple = (f32[4096,128]{1,0}, f32[4096,32,128]{2,1,0}, f32[4096,8,128]{2,1,0}, f32[4096,14336]{1,0},
```
frgossen, line 1823: Can you simplify this test case? IIUC, all you need is a mixed tuple result. Would one ag, one parameter, and one add be sufficient?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1779,6 +1779,81 @@ TEST(StreamExecutorGpuClientTest, ExecutablePinnedHostOutputMemoryKindTest) {
   EXPECT_EQ(memory_kinds[0][0], "pinned_host");
 }
 
+TEST(StreamExecutorGpuClientTest,
+     GetCompiledMemoryStatsWithTupleAndNcclUserBuffers) {
+  TF_ASSERT_OK_AND_ASSIGN(auto client,
+                          GetStreamExecutorGpuClient(DefaultOptions()));
+
+  xla::CompileOptions options;
```
frgossen, line 1787: Can you add a test case where the mixed tuple is not the root of the operation? I saw the root is a bit of a special case here.

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1779,6 +1779,81 @@ TEST(StreamExecutorGpuClientTest, ExecutablePinnedHostOutputMemoryKindTest) {
...
+      auto executable,
+      CompileExecutable(kProgramWithCollectiveAndTuple, *client, options));
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto memory_stats, executable->GetExecutable()->GetCompiledMemoryStats());
+  EXPECT_GE(memory_stats.output_size_in_bytes, 0);
+  EXPECT_GE(memory_stats.host_output_size_in_bytes, 0);
+  EXPECT_GE(memory_stats.peak_memory_in_bytes, 0);
+}
+
```
frgossen, line 1856: Can you also add a test case with many tuples. The memory statistics should show the memory usage of tuples, which I think it does not with your current change.

xla/pjrt/pjrt_executable.cc
frgossen: So are you still iterating over tuples or are you skipping them? I think you are skipping them. 

> this part of code is just to decide if the buffer should be counted...
This code also checks an invariant, which, based on the comment above is important. The counting happens below. 

I dug a little deeper and think I understand your point now. The tuple itself is allocated on device, in the kDefaultMemorySpace. It is not clear to me why the old implementation causes crashes for the test case above. I would assume it realizes that the tuple does not have a layout (has_layout is false) and records kDefaultMemorySpace for it (just like you fix it below). It seems all you do here is skip the sanity check on line 297, which is important. If `CHECK(alloc_memory_space == memory_space` fails for you, then you have two buffers with different memory spaces assigned to the same allocation, which is wrong. We should not remove this check here. If there is a special case that I'm missing, we should fix this sanity check for that specifically. 

In your test case (thanks for adding that!!), the only tuple is the result value. How does that crash on the old implementation? Which buffers in the same allocation have contradicting memory spaces?

xla/pjrt/pjrt_executable.cc
```diff
@@ -300,6 +299,14 @@ void CompiledMemoryStats::PopulateBufferStatsFromAllocations(
       }
     }
 
+    // In the case where all assignments in an allocation are tuples, prevent
+    // custom colorer from setting their color to host.
+    if (alloc_memory_space == -1) {
+      CHECK_EQ(alloc.color(), 0)
```
frgossen, line 305: Please use constants like kDefaultMemorySpace and do not hard-code 0 here.

xla/pjrt/pjrt_executable.cc
```diff
@@ -300,6 +299,14 @@ void CompiledMemoryStats::PopulateBufferStatsFromAllocations(
...
+          << "allocation contains only tuples, must be on device.";
+      alloc_memory_space = 0;
```
frgossen, line 307: same

frgossen (changes requested):
I think we have to understand the crash better. The change in this PR seems wrong to me in two ways:
 
- You skip an important sanity check: If two buffers with different memory spaces are assigned to the same allocation that is wrong. If we do not catch that here, it can be even more painful to debug. 
- If you skip over the tuples completely (not sure if that is the case), you might miss the allocations for tuples in the memory statistics.

xla/pjrt/pjrt_executable.cc
```diff
@@ -288,10 +288,9 @@ void CompiledMemoryStats::PopulateBufferStatsFromAllocations(
     int64_t alloc_memory_space = -1;
     for (const auto& [value, _] : alloc.assigned_buffers()) {
       const HloPosition& defining_position = value->defining_position();
-      int64_t memory_space = Layout::kDefaultMemorySpace;
-      if (defining_position.shape().has_layout()) {
-        memory_space = defining_position.shape().layout().memory_space();
-      }
+      if (defining_position.shape().IsTuple()) continue;
```
gspschmid, line 291: You pointed out that only arrays and buffers have a layout (https://github.com/openxla/xla/blob/186d61f0182f6561ae81a1e73c9217118fe405b2/xla/shape.h#L390C4-L392) -- instead of just tuples shouldn't we handle all non-array/buffer shapes like this?

xla/pjrt/pjrt_executable.cc
gspschmid: > How does that crash on the old implementation? Which buffers in the same allocation have contradicting memory spaces?

Agreed, I think this is important to clear up.

Is is that collective colorer ends up assigning `kCollective` to the tuple as well? If indeed tuples always live in device memory, lets make sure we only ever assign them to `kDefaultMemorySpace`. In that case the existing code ("assume `kDefaultMemorySpace` for anything that doesn't have a layout") becomes correct again and you won't need any additional checks/exemptions.

gspschmid: I just tried this out, simply doing the same thing as the default colorer for layout-less values fixes the issue, no need to relax the assertion: https://github.com/gspschmid/xla/commit/160e84758d878b2d98338ecfe8b9ba28275d89b1#diff-115ed306585431bb058ab9f96a390f001f21a486644a8272f4080a827cb21951R118-R120

We previously identified that CollectiveColorer should refine DefaultColorer's behavior in the sense that sometime when DefaultColorer assigns the default device memory space we instead want to assign `kCollective`. So the new insight here is that for layout-less values like tuples we should _always_ assign the default device memory space.

It's probably worth documenting all this in a comment in collective colorer.

gspschmid (changes requested):

xla/pjrt/pjrt_executable.cc
frgossen: I think tuples are the only layoutless value if you will. They do live in memory nevertheless, although most places look right through them. 

Thanks for investigating this! This is the right path forward.

frgossen (changes requested):
Should we drop this PR after the recent discussion?

xla/pjrt/pjrt_executable.cc
gspschmid: > I think tuples are the only layoutless value if you will. They do live in memory nevertheless, although most places look right through them.

@frgossen Do you know whether tokens have some runtime representation? See also https://github.com/openxla/xla/pull/28304#discussion_r2333313555

xla/pjrt/pjrt_executable.cc
frgossen: is there any non-array/buffer shape other than tuples?

frgossen (changes requested):
Following up on the offline discussion, should we convert this PR into just the test cases and keep the check as it was? Happy to land the test cases.

xla/pjrt/pjrt_executable.cc
gspschmid: Yes, at least tokens (and not sure what opaque is): https://github.com/openxla/xla/blob/186d61f0182f6561ae81a1e73c9217118fe405b2/xla/shape.h#L53-L55

frgossen: FWIW, I think tokens actually do not have a representation at runtime.

PR #28396, opened 2025-06-30
[NVIDIA GPU] Skip user buffer reg when the size is 1

Tixxx:
Using UB for single-element collectives brings no benefit, we skip those collectives here to avoid unnecessary reg and dereg process.

### DIFF
--- a/xla/backends/gpu/runtime/collective_thunk.cc
+++ b/xla/backends/gpu/runtime/collective_thunk.cc
@@ -380,6 +380,10 @@ absl::Status MaybeRegisterBuffers(se::StreamExecutor* executor,
                                   const std::vector<DeviceBufferPair>& buffers,
                                   Communicator* comm) {
   for (int i = 0; i < buffers.size(); ++i) {
+    if (buffers[i].element_count == 1) {
+      VLOG(5) << "Skipping buffer registration for single element collectives.";
+      continue;
+    }
     if (buffers[i].source_memory_space == kCollectiveMemorySpaceColor) {
       TF_RETURN_IF_ERROR(
           RegisterBufferOnce(executor, comm, buffers[i].source_buffer));


frgossen:
Can you add test cases for these PRs?

Tixxx:
> Can you add test cases for these PRs?

The registration is done at runtime, any suggestion on how to verify such runtime behavior? the user buffer registration cache is also not exposed outside of the thunks.

xla/backends/gpu/runtime/collective_thunk.cc
```diff
@@ -380,6 +380,11 @@ absl::Status MaybeRegisterBuffers(se::StreamExecutor* executor,
                                   const std::vector<DeviceBufferPair>& buffers,
                                   Communicator* comm) {
   for (int i = 0; i < buffers.size(); ++i) {
+    if (buffers[i].source_buffer.size() == 1 &&
```
PatriosTheGreat: Why the size 1 is special here? Would buffers of size 2 always benefit from using a user buffer registration? Or should this parameter be tunable?

PatriosTheGreat (changes requested):

xla/backends/gpu/runtime/collective_thunk.cc
Tixxx: i just realized that this should really be element count.
For a buffer with more than 1 elements, nccl will choose to use nvls+user buffer based on heuristics, there's currently no API to query nccl algo based on element size but for a single element collective, we know for sure that nccl will not use nvls with user buffer.

frgossen (changes requested):

Tixxx:
> Following up on these PRs based on the discussions we had offline in our regular meetings. For this PR, we need to understand and track the performance implications.
> 
> Some general asks for all PRs:
> 
> * What workload is this change motivated by and why is it important?
> * How can we measure and track the performance delta for this PR?
>   Can you provide speedups you measure on one of the HLO benchmarks in compiler/xla/tools/benchmarks/hlo/?
> * Can you add or point to an execution test that exercises this code path?

@frgossen  This change is mainly a minor improvement on memory utilization. We don't register the trivial size to avoid polluting the global memory registry with small registrations which don't have any runtime impact. We find in some fp8 cases there are many single-element allreduces on scales which add dozens of entries in the global registration cache but nccl doesn't use any nvls algo for single-element collectives.
This won't have any meaningful runtime perf impact though. Any suggestion to move forward with this type of change?

frgossen:
> > Following up on these PRs based on the discussions we had offline in our regular meetings. For this PR, we need to understand and track the performance implications.
> > Some general asks for all PRs:
> > 
> > * What workload is this change motivated by and why is it important?
> > * How can we measure and track the performance delta for this PR?
> >   Can you provide speedups you measure on one of the HLO benchmarks in compiler/xla/tools/benchmarks/hlo/?
> > * Can you add or point to an execution test that exercises this code path?
> 
> @frgossen This change is mainly a minor improvement on memory utilization. We don't register the trivial size to avoid polluting the global memory registry with small registrations which don't have any runtime impact. We find in some fp8 cases there are many single-element allreduces on scales which add dozens of entries in the global registration cache but nccl doesn't use any nvls algo for single-element collectives. This won't have any meaningful runtime perf impact though. Any suggestion to move forward with this type of change?

So the benefit here is better memory utilization, right? Can we measure this improvement (even if minor) on one of the  HLO benchmarks in compiler/xla/tools/benchmarks/hlo/?

Tixxx:
closing this since the registration has been changed to registering address ranges which addresses the same issue

PR #28401, opened 2025-06-30
[ROCm] Fix PackedTranspose for adapting to warp size 64

amd-jianli12:
Calculate ranges for dimensions in IndexingMap with kTileSize rather than WarpSize() since tile size is still expected to be vector_size * kNumShmemBanks which we preserve as 32.

Fixed failing test:
[  FAILED  ] Convolve2D_1x2x2x1024_2x2x128x512_Grouped_Valid/0.Types, where TypeParam = Eigen::half

### DIFF
--- a/xla/backends/gpu/codegen/emitters/transpose.cc
+++ b/xla/backends/gpu/codegen/emitters/transpose.cc
@@ -802,7 +802,7 @@ IndexingMap PackedTranspose::GetInputIndexing(MLIRContext* ctx) const {
       KernelFusionInterface::kIndexingMapThreadIdxDims[0], ctx);
   auto block_id =
       getAffineDimExpr(KernelFusionInterface::kIndexingMapBlockIdxDims[0], ctx);
-  auto warp_size = WarpSize(analysis_.device_info());
+  auto warp_size = kNumShmemBanks;
   auto lane_id = thread_id % warp_size;
   auto warp_id = thread_id.floorDiv(warp_size);
   std::vector<IndexingMap::Variable> dim_vars = DimVarsFromGPUGrid(
@@ -863,7 +863,7 @@ IndexingMap PackedTranspose::GetShmemWriteIndexing(
   // Dimensions variables.
   auto thread_id = getAffineDimExpr(
       KernelFusionInterface::kIndexingMapThreadIdxDims[0], ctx);
-  auto warp_size = WarpSize(analysis_.device_info());
+  auto warp_size = kNumShmemBanks;
   auto lane_id = thread_id % warp_size;
   auto warp_id = thread_id.floorDiv(warp_size);
   std::vector<IndexingMap::Variable> dim_vars = DimVarsFromGPUGrid(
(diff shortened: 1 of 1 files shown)


i-chaochen:
@xla-rotation this has fixed our ROCm CI

amd-songpiao:
Confirmed that the changes fixed the following two failed unit tests on rocm, too.

1. CollectiveOpsTestE2EWindowedNonWindowed.WindowedEinsumE2EAllGatherMultiConsumerF8
2. CollectiveOpsTestE2EWindowedNonWindowed.WindowedEinsumE2EAllToAllTransposeDecompose

e.g. 
bazel-bin/xla/tests/collective_ops_e2e_test_amdgpu_any --gtest_filter=CollectiveOpsTestE2EWindowedNonWindowed.WindowedEinsumE2EAllGatherMultiConsumerF8

xla/backends/gpu/codegen/emitters/transpose.cc
```diff
@@ -802,7 +802,7 @@ IndexingMap PackedTranspose::GetInputIndexing(MLIRContext* ctx) const {
       KernelFusionInterface::kIndexingMapThreadIdxDims[0], ctx);
   auto block_id =
       getAffineDimExpr(KernelFusionInterface::kIndexingMapBlockIdxDims[0], ctx);
-  auto warp_size = WarpSize(analysis_.device_info());
+  auto warp_size = kTileSize;
```
akuegel: While technically kTileSize replacement works, I believe logically it should be kNumShmemBanks

@pifon2a to confirm

pifon2a:
I am not very familiar with ROCm. If there are 64 threads per warp, does it mean that the shared memory has 64 x 4 byte banks? If yes, then all these tiles have to be tied to WarpSize.

amd-jianli12:
> I am not very familiar with ROCm. If there are 64 threads per warp, does it mean that the shared memory has 64 x 4 byte banks? If yes, then all these tiles have to be tied to WarpSize.

Yes, that was also my guess. But when I refer to [Chapter 11. Data Share Operations](https://www.amd.com/content/dam/amd/en/documents/instinct-tech-docs/instruction-set-architectures/amd-instinct-mi300-cdna3-instruction-set-architecture.pdf) of amd-instinct-mi300-cdna3-instruction-set-architecture.pdf, it says:

"There are 64 kB memory per compute unit, segmented into 32 banks of 512 Dwords, each bank being 32bits
wide."

That's the reason I'm trying to preserve kNumShmemBanks as 32 here. Seems CDNA 3 is warp-32 compatible through warp-64 HW?

xla/backends/gpu/codegen/emitters/transpose.cc
amd-jianli12: > While technically kTileSize replacement works, I believe logically it should be kNumShmemBanks
> 
> @pifon2a to confirm

Yes, kNumShmemBanks is more essential. Do you prefer I change kTileSize into kNumShmemBanks?

i-chaochen:
@xla-rotation wondering what do you expect us to do next? thanks in advance!

xla/backends/gpu/codegen/emitters/transpose.cc
pifon2a: Yes, let's call it kNumShmemBanks

pifon2a (approved):

i-chaochen:
Hi @xla-rotation can we merge this PR, please

pifon2a (approved):

PR #28406, opened 2025-06-30
[ROCm] Added support for handling variable-length sequences when using RNN layers.

ThorBl:
Implements support for variable-length (padded) sequences in the ROCm backend for RNN operations (LSTM, GRU, etc.). This resolves a feature gap compared to the cuDNN backend, which previously resulted in UNIMPLEMENTED errors when training RNN models with padded inputs on ROCm-enabled hardware.

**Refactored MIOpenRnnSequenceTensorDescriptor:**

* The class now uses a private constructor with public static Create factory methods, allowing for fallible initialization to be handled by returning an absl::StatusOr.
* The class always creates a base miopenTensorDescriptor_t and populates the legacy handles_ vector with pointers to it. This design choice ensures backward compatibility with existing functions like CheckRNNParameterSize that expect a non-empty handles() array.
* For variable-length sequences, the Create method also creates and configures a specialized miopenSeqTensorDescriptor_t.
* A new is_var_seq_lengths() method was added to provide a clear, explicit way for consumers to determine which execution path to take.

**Updated MIOpenSupport::CreateRnnSequenceTensorDescriptor:**

* The factory functions in MIOpenSupport have been updated to call the new static Create methods on the descriptor class.
* The TF_ASSIGN_OR_RETURN macro is now used for clean, idiomatic unpacking and error handling of the absl::StatusOr result.
 
**Updated MIOpenSupport::DoRnnForwardImpl:**

* The core RNN execution logic now implements a Strategy Pattern.
* It uses the new is_var_seq_lengths() method to check which type of descriptor was created.
* Based on the result, it calls the appropriate MIOpen kernel launch, passing either the specialized seq_tensor_descriptor() for padded sequences or the legacy handles() array for uniform-length sequences.

Example code that uses variable-length sequences (formerly failed with 'ROCm MIOpen only supports packed input output.' error): https://keras.io/examples/audio/ctc_asr/

### DIFF
--- a/xla/stream_executor/rocm/rocm_dnn.cc
+++ b/xla/stream_executor/rocm/rocm_dnn.cc
@@ -352,12 +352,18 @@ namespace wrap {
   __macro(miopenCreateRNNDescriptor)                                 \
   __macro(miopenSetRNNDescriptor)                                    \
   __macro(miopenSetRNNDescriptor_V2)                                 \
+  __macro(miopenSetRNNPaddingMode)                                   \
+  __macro(miopenSetRNNDataSeqTensorDescriptor)                       \
+  __macro(miopenCreateSeqTensorDescriptor)                           \
+  __macro(miopenDestroySeqTensorDescriptor)                          \
   __macro(miopenDestroyRNNDescriptor)                                \
   __macro(miopenGetRNNParamsSize)                                    \
   __macro(miopenGetRNNLayerParam)                                    \
   __macro(miopenGetRNNLayerBias)                                     \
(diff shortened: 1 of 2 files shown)


i-chaochen:
Hi @xla-rotation please check this PR. Thanks!

beckerhe (commented):
Hey, the code looks good to me in general but can you add tests for these changes (or enable existing that have so far been skipped for ROCm)

dimitar-asenov:
Closing due to inactivity. If you still want to merge this PR, please rebase and address the comment above.

PR #28421, opened 2025-06-30
Add memcopy async wrapper

chaserileyroberts:
This PR adds the `MemcopyAsyncWrapper` pass to gpu_compiler.cc.

Recently, there has been work from Nvidia to allow dynamic [update] slices that are contiguous to be dispatched from CUDA memcopies. This has the advantage that these operations do not need any SMs to run, and thus do not take away resources from other compute or H2D kernels that need them. However, to fully utilize these changes, it must be possible to run these operations in parallel with other work on different CUDA streams. Thankfully most of the infrastructure to do this already exists, so we just need to call `AsyncWrapper` on our specific fusion operations. 

From our testing, just this simple change results in a ~1% E2E speedup improvement of a Llama7B from MaxText on a single B200 node. Since this change is very general, we expect similar speedups across the board for models / hardware. However, we also find that the default overlap of these operations isn't perfect, and so further improvements to account for this change in the LHS will likely be the next follow ups. We expect improvements to the default scheduling could give another 1% improvement across the board. 

There has been some instability reported from customers around the memcopy feature, so we have this change broken out into an independent pass so that users can disable it using `--xla_disable_hlo_passes=MemcopyAsyncWrapper` without introducing another flag into XLA if they hit issues.

### DIFF
--- a/xla/service/gpu/transforms/async_wrapper.cc
+++ b/xla/service/gpu/transforms/async_wrapper.cc
@@ -59,8 +59,8 @@ absl::StatusOr<bool> AsyncWrapper::Run(
         // If the predicate matches, then wrap the instructions in async blocks.
         TF_RETURN_IF_ERROR(
             computation
-                ->CreateAsyncInstructions(instruction,
-                                          {ShapeUtil::MakeScalarShape(U32)})
+                ->CreateAsyncInstructions(
+                    instruction, {ShapeUtil::MakeScalarShape(U32)})
                 .status());
         changed = true;
         continue;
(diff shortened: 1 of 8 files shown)


xla/service/gpu/transforms/async_wrapper.cc
```diff
@@ -58,8 +58,8 @@ absl::StatusOr<bool> AsyncWrapper::Run(
         // If the predicate matches, then wrap the instructions in async blocks.
         TF_RETURN_IF_ERROR(
             computation
-                ->CreateAsyncInstructions(instruction,
-                                          {ShapeUtil::MakeScalarShape(U32)})
+                ->CreateAsyncInstructions(
+                    instruction, {ShapeUtil::MakeScalarShape(U32)}, "parallel")
```
chaserileyroberts: @abhigunj

frgossen (changes requested):
Thanks for adding this, it is intuitively a clear improvement.  

Some general asks for all PRs  (that also apply to this PR):

Specifically for this PR: "There has been some instability reported from customers around the memcopy feature" In this case, we should disable it by default and get to the bottom of it. Do you have a reproducer for such issues and can put that into a test case? We can add the test case in disabled form.

PR #28497, opened 2025-07-02
Create cudnn handle for all threads to reuse during cudnn compilation

serach24:
When multiple threads invoke cudnnCreate api in cuda_dnn.cc::GetLocalHandle() in the one-process-per-gpu setting, it causes a hang of 6-7 seconds, which causes a significant compilation overhead when `--xla_gpu_enable_triton_gemm=true` and `--xla_gpu_cudnn_gemm_fusion_level=1` is set on blackwell GPUs. 
This PR creates a separate handle for cudnn plan compilation only and reuses the same handle across threads.


derdrdirk (approved):
LGTM

PR #28615, opened 2025-07-08
[XLA:CPU][oneDNN] Implement oneDNN primitives for custom calls in Thunk runtime

othakkar:
This PR adds the implementation for oneDNN primitives that will be called when the thunk runtime is enabled for oneDNN custom calls (which will be added in a separate PR following this one). Currently, the `ExecuteOneDnn...` functions are not called or referenced anywhere and are meant to serve as a stub implementation or foundation for integrating oneDNN-backed custom calls into the thunk-based execution path.

These functions implement the low-level logic for various oneDNN operations and will be invoked once the dispatch logic is wired up. By landing this code first, we isolate the core implementation from integration concerns, making the follow-up changes smaller and easier to review.


othakkar:
@penpornk, can you please review the PR when you get a chance? Thanks!

xla/service/cpu/onednn_convolution.cc
```diff
@@ -208,6 +210,167 @@ CreateOneDnnPrimDesc<dnnl::convolution_forward::primitive_desc>(
       any_res_md, strides, rhs_dilations, pad_left, pad_right, attrs);
 }
 
+void ExecuteOneDnnConvolution(void* result, void* scratch, void** args,
```
penpornk, line 213: This is copy-pasting the old `__xla_cpu_runtime_OneDnnConvolution` with minor modifications. I think we should refactor `__xla_cpu_runtime_OneDnnConvolution` instead. Same for the other three `ExecuteOneDnn{Matmul,Layernorm,Softmax}` functions.

penpornk, line 213: The new function should use the new FFI interface instead of using `void*` pointers.

penpornk (changes requested):
Thank you for the PR! 

This is 400+ lines of code without tests though. I think it's better to have a self-contained PR. E.g., in addition to adding `ExecutingOneDnnMatMul` that's not called anywhere, the PR should show how you hook `ExecutingOneDnnMatMul` to existing code path, and add tests if it's not already covered by existing tests. 

If the PR gets too big, we can split the PR by primitives (one for matmul, one for conv,  one for layernorm, and one for softmax).

xla/service/cpu/onednn_convolution.cc
othakkar: @penpornk, thanks for the review. Could you please point me to a custom call example that uses the new FFI interface?

othakkar:
Marking this PR as "draft". I'll be upstreaming another set of PRs to enable support for custom call oneDNN in the thunk runtime based on @penpornk's review comments.

xla/service/cpu/onednn_convolution.cc
penpornk: Sorry for the late reply! 
JAX has plenty of examples in the example and jaxlib folder:

* https://github.com/jax-ml/jax/tree/main/examples/ffi
* https://github.com/jax-ml/jax/blob/main/jaxlib/cpu/lapack_kernels.cc
* https://github.com/jax-ml/jax/blob/main/jaxlib/cpu/sparse_kernels.cc
* https://github.com/jax-ml/jax/blob/main/jaxlib/gpu/hybrid_kernels.cc

For opaque types, you can just use `ffi::AnyBuffer`.

othakkar:
Closing this PR now, as other PRs enabling oneDNN custom call primitives in thunk runtime are merged.

PR #28671, opened 2025-07-09
[XLA:CPU] Enable mm-biasadd-add fusion as seen in some models

kanvi-nervana:
This PR enables matmul-biasadd-add fusion as seen in some models where dot is followed by a reshape. In such cases, the binary operands shape can be adjusted, making sure the number of elements match, enabling the fusion. This PR also adds a test for the same.

### DIFF
--- a/xla/service/cpu/onednn_contraction_rewriter.cc
+++ b/xla/service/cpu/onednn_contraction_rewriter.cc
@@ -19,6 +19,7 @@ limitations under the License.
 
 #include "xla/service/cpu/onednn_contraction_rewriter.h"
 
+#include "tsl/platform/logging.h"  // IWYU pragma: keep
 #include "xla/executable_run_options.h"
 #include "xla/hlo/evaluator/hlo_evaluator.h"
 #include "xla/hlo/ir/dfs_hlo_visitor_with_default.h"
@@ -36,7 +37,6 @@ limitations under the License.
 #include "xla/service/pattern_matcher.h"
 #include "xla/status_macros.h"
 #include "xla/tsl/util/onednn_threadpool.h"
-#include "tsl/platform/logging.h"  // IWYU pragma: keep
 
 namespace xla {
 namespace cpu {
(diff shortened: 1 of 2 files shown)


xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -391,6 +391,20 @@ absl::StatusOr<Shape> AdjustAddendShape(const HloInstruction* contraction,
...
+// Adjusting the operand shape to the dot's shape enables fusion in oneDNN.
+absl::StatusOr<Shape> AdjustBinaryOperandShape(
+    const HloInstruction* operand_instr, const Shape& dot_shape) {
+  if (ShapeUtil::ElementsIn(operand_instr->shape()) !=
+      ShapeUtil::ElementsIn(dot_shape)) {
+    return absl::CancelledError(
+        "Number of elements in operand and dot instruction do not match.");
+  }
+  Shape new_shape = dot_shape;
+  return new_shape;
```
penpornk: Nit: We can just return `dot_shape`.
```suggestion
  return dot_shape;
```

xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -716,6 +730,44 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
...
+      // enable the fusion. For example:
+      //      dot = f32[6304,3072] dot(...)
+      //      reshape = f32[32,197,3072] reshape(dot)
+      //      constant = f32[32,197,3072] constant(..)
+      //      add = f32[32,197,3072] add(reshape, constant)
+      // can become
+      //      dot = f32[6304,3072] dot(...)
+      //      constant = f32[32,197,3072] constant(..)
+      //      reshape1 = f32[6304,3072] reshape(constant)
+      //      add = f32[6304,3072] add(dot, reshape1)
```
penpornk, line 753: I think we need to add another `reshape` after `add` to get the output back to `[32,197,3072]`?

xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -716,6 +730,44 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
...
+      // and be replaced with the fusion
+      //      fused = f32[6304,3072] custom-call(..)
+      //      bitcast = f32[32,197,3072] bitcast(fused)
+      // clang-format on
```
penpornk: I don't see a `clang-format off`, so maybe this line isn't needed?

penpornk (changes requested):

kanvi-nervana:
@penpornk I have addressed the comments, thanks!

penpornk (approved):
Thank you for the changes!

PR #28716, opened 2025-07-09
[GPU] Make fabric info test compatible with lower CUDA driver versions 

terryysun:
Existing fabric info test assumes sufficient CUDA driver version (550+), which is not always the case. This PR make the test compatible with lower driver version. The enforcement of getting empty fabric info on Hopper is also removed as there's no guarantee for the status quo.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
@@ -1232,14 +1232,17 @@ TEST(StreamExecutorGpuClientTest, GetDeviceFabricInfo) {
           if (auto* cc = std::get_if<se::CudaComputeCapability>(
                   &executor->GetDeviceDescription().gpu_compute_capability())) {
             if (cc->IsAtLeastHopper()) {
-              TF_ASSERT_OK_AND_ASSIGN(
-                  std::string fabric_info,
-                  GetDeviceFabricInfo(executor->device_ordinal()));
-              // Hopper devices have empty fabric info, MNNVL Blackwell devices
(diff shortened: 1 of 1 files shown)


xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -1232,14 +1232,17 @@ TEST(StreamExecutorGpuClientTest, GetDeviceFabricInfo) {
...
+                  GetDeviceFabricInfo(executor->device_ordinal());
+              if (!fabric_info.ok()) {
+                auto error_message = fabric_info.status().message();
+                // Only allow failures due to insufficient CUDA driver version.
+                EXPECT_TRUE(
+                    (error_message.find("Failed to initialize NVML library.") !=
+                     std::string::npos) ||
+                    (error_message.find(
+                         "NVML library doesn't have required functions.") !=
+                     std::string::npos));
```
allanrenucci: You can express this with matchers. I.e.
```c++
using ::testing::HasSubstr;
using ::testing::AnyOf;

EXPECT_THAT(
    fabric_info.status().message(),
    AnyOf(HasSubstr("Failed to initialize NVML library."),
          HasSubstr("NVML library doesn't have required functions.")));
```

allanrenucci (changes requested):
Otherwise, LGTM.

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
terryysun: thanks! updated

allanrenucci (approved):

PR #28782, opened 2025-07-11
[XLA:GPU] Annotate cuBLAS/cuDNN outputs to avoid initcheck failures

sergey-kozub:
Upgrades NVTX to v3.2.1 and marks the outputs of cuBLAS/cuDNN as initialized (as compute-sanitizer may emit false positives for kernels using TMA).

### DIFF
--- a/xla/backends/gpu/runtime/thunk.proto
+++ b/xla/backends/gpu/runtime/thunk.proto
@@ -110,6 +110,7 @@ message PartitionIdThunkProto {
 message CudnnThunkProto {
   string fingerprint = 1;
   repeated xla.buffer_assignment.BufferAllocationSliceProto args = 2;
+  repeated bool output_args = 4;
   optional int64 sdpa_dropout_seed = 3;
 }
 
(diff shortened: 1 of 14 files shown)


olupton (approved):

akuegel (approved):

xla/backends/gpu/runtime/thunk.proto
```diff
@@ -110,6 +110,7 @@ message PartitionIdThunkProto {
 message CudnnThunkProto {
   string fingerprint = 1;
   repeated xla.buffer_assignment.BufferAllocationSliceProto args = 2;
+  repeated bool output_args = 4;
```
thomasjoerg, line 113: Please add some test coverage for this new proto field to make sure it's used correctly.

thomasjoerg (changes requested):

xla/backends/gpu/runtime/thunk.proto
sergey-kozub: Updated the tests, PTAL.

akuegel (approved):

PR #28877, opened 2025-07-14
[XLA]Clamp num_workers to avoid partition overflow

isharif168:
When invoking Worker::Parallelize in parallel_for(), the number of workers was previously set to thread_pool_->NumThreads(), which could exceed the number of tasks (n). This mismatch could lead to out-of-bounds access when workers attempt to access partitions that don't exist.

This ensures that:
- The number of worker threads does not exceed the number of tasks
- Each worker maps safely to a valid partition index
- Partition and worker ranges stay aligned, preventing undefined behavior

Change-Id: I5c3ab319f73009452d27b4db43f7905c5e136c2b

### DIFF
--- a/xla/backends/cpu/runtime/parallel_loop_runner.cc
+++ b/xla/backends/cpu/runtime/parallel_loop_runner.cc
@@ -86,15 +86,10 @@ ABSL_ATTRIBUTE_ALWAYS_INLINE void ParallelLoopRunner::ScheduleAll(
     size_t num_tasks, ParallelTask&& parallel_task) {
   DCHECK_GT(num_tasks, 1) << "Expected at least two task";
 
-  // Use at most `num_threads()` workers as we can't run more parallel workers
-  // than the number of threads in the thread pool.
-  size_t num_workers = std::min(std::min(num_tasks, num_threads()),
-                                size_t{std::numeric_limits<uint16_t>::max()});
-
   auto parallelize =
-      [this, num_workers, num_tasks,
+      [this, num_tasks,
        parallel_task = std::forward<ParallelTask>(parallel_task)](tsl::Chain) {
-        return Worker::Parallelize(device_.load()->getPool(), num_workers,
+        return Worker::Parallelize(device_.load()->getPool(), num_threads(),
                                    num_tasks, std::move(parallel_task));
       };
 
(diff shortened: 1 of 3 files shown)


cfRod:
@penpornk

penpornk (approved):
Thank you for the fix!

penpornk (changes requested):
Actually, I think we could fix this in `Worker::Parallelize` itself. 
https://github.com/openxla/xla/blob/f2039ffb2f5fa2aefbf28fedc57d3c8dbc44f25e/xla/backends/cpu/runtime/work_queue.h#L251

Please also add a test case in `WorkQueueTest.WorkQueuePartitions`:
https://github.com/openxla/xla/blob/f2039ffb2f5fa2aefbf28fedc57d3c8dbc44f25e/xla/backends/cpu/runtime/work_queue_test.cc#L38

cc: @ezhulenev

isharif168:
> Actually, I think we could fix this in `Worker::Parallelize` itself.
> 
> https://github.com/openxla/xla/blob/f2039ffb2f5fa2aefbf28fedc57d3c8dbc44f25e/xla/backends/cpu/runtime/work_queue.h#L251

XNNPack is already doing the clamping here
https://github.com/openxla/xla/blob/f2039ffb2f5fa2aefbf28fedc57d3c8dbc44f25e/xla/backends/cpu/runtime/parallel_loop_runner.cc#L91

But yes I think keeping it in `Worker::Parallelize` can avoid bugs if there are new path in future.
> 
> Please also add a test case in `WorkQueueTest.WorkQueuePartitions`:
> 
> https://github.com/openxla/xla/blob/f2039ffb2f5fa2aefbf28fedc57d3c8dbc44f25e/xla/backends/cpu/runtime/work_queue_test.cc#L38

We already have test for it but its not tested for oneDNN 

Thanks.

penpornk (changes requested):
XLA's policy requires adding / update tests for any PR that makes functional changes. This bug isn't hard to catch with a unit test so I think we should add one.

The easiest way is to fix in `Worker::Parallelize` and just update its unit test (adding one more sub-test). If you strongly prefer to fix in onednn_threadpool itself, you'll need to add a test case, probably in a new test file, e.g, onednn_threadpool_test.cc.

penpornk (approved):
Thank you for the changes!

PR #28919, opened 2025-07-15
[NVIDIA GPU] [XLA_GPU_MS_COLLECTIVE] Round-robin stream assignment for async communications

terryysun:
As part of the revival of https://github.com/openxla/xla/pull/22450, this PR adds round-robin stream assignment algorithm for async collectives. Followup PR will integrate the algorithm to the pipeline once stream assignment mechanism refactoring is in.


frgossen (approved):
Thank yoU!

PR #28928, opened 2025-07-15
Dump optimized HLO when deserializing

olupton:
This makes HLO dumps available when using the JAX compilation cache.

### DIFF
--- a/xla/pjrt/pjrt_stream_executor_client.cc
+++ b/xla/pjrt/pjrt_stream_executor_client.cc
@@ -3854,14 +3854,14 @@ PjRtStreamExecutorClient::LoadSerializedExecutable(
                       DeserializeToLocalExecutable(serialized, options));
   return LoadInternal(/*unoptimized_hlo_module_proto=*/std::nullopt,
                       std::move(local_executables_and_options.first),
-                      local_executables_and_options.second);
+                      local_executables_and_options.second, /*dump=*/true);
 }
 
 absl::StatusOr<std::unique_ptr<PjRtLoadedExecutable>>
 PjRtStreamExecutorClient::LoadInternal(
     std::optional<HloModuleProto> unoptimized_hlo_module_proto,
     std::vector<std::unique_ptr<LocalExecutable>> local_executables,
-    CompileOptions compile_options) {
+    CompileOptions compile_options, bool dump) {
   auto input_options = compile_options;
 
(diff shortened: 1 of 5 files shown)


xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3739,6 +3739,35 @@ PjRtStreamExecutorClient::LoadInternal(
...
+        VLOG(1) << "Replacing serialized debug options:\n"
+                << serialized_only_values << "with passed debug options:\n"
+                << GetNonDefaultDebugOptions(
+                       ex_options.debug_options(),
+                       hlo_module.config().debug_options());
+      }
+      hlo_module.mutable_config().mutable_debug_options() =
+          ex_options.debug_options();
+    }
+  }
```
beckerhe: I think this is too dangerous. There are probably various implicit dependencies between the runtime, the debug options, and the HLO module. Changing arbitrary DebugOptions might break things in unexpected ways.

I think a better way to do this is to change `DumpHloModuleIfEnabled` in a way that the DebugOptions can be supplied as a separate argument.

xla/pjrt/pjrt_stream_executor_client.cc
```diff
@@ -3739,6 +3739,35 @@ PjRtStreamExecutorClient::LoadInternal(
...
+  if (dump) {
+    for (std::unique_ptr<LocalExecutable>& local_executable :
+         local_executables) {
+      VLOG(1) << "Dumping deserialized executable";
+      // Does not quite match the naming convention of the dump during
+      // compilation, which includes a backend-specific prefix.
+      DumpHloModuleIfEnabled(local_executable->executable()->module(),
+                             kAfterOptimizationsDumpName);
```
beckerhe: `Executable::module()` might fail when no module is available. Can you wrap this in an if branch with `local_executable->executable()->has_module()` as the condition?

xla/pjrt/pjrt_stream_executor_client.cc
olupton: OK, I tried to plumb through the relevant options more explicitly.

xla/pjrt/pjrt_stream_executor_client.cc
olupton: Done 👍

beckerhe (approved):

PR #28962, opened 2025-07-16
Set KV store to null with mocked GPU processes

### LINKED ISSUE #28959: GPU mocking hangs the compiler (in autotuning)
When we use GPU mocking to mock multiple processes, the autotuner sharding will try to get autotuning results from the mock processes. That will hang.

JAX repro:
```python
import jax
import jax.numpy as jnp
from jax.sharding import Mesh, PartitionSpec as P, NamedSharding
import numpy as np

jax.config.update("jax_mock_gpu_topology", "4x8x1")

print(jax.devices())

devices = np.array(jax.devices()).reshape(4, 4, 2)
mesh = Mesh(devices, ('x', 'y', 'z'))
NS = lambda *s: NamedSharding(mesh, P(*s))
DP = lambda a, s: jax.device_put(a, NS(*s))

@jax.jit
def mm(x, y):
  return jnp.dot(x, y)

x = DP(jnp.ones((16, 16)), ('x', 'z'))
y = DP(jnp.ones((16, 16)), ('y', None))

# This hangs in the compiler's autotuner.
out = mm(x, y)
print(out.sharding)
```

jaro-sevcik:
With GPU mocking, there are no other processes, so any communication
over KV store will hang (e.g., sharding autotuning). Let us just set KV store
to null with GPU mocking on.

Fixes https://github.com/openxla/xla/issues/28959

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -1702,9 +1702,10 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
 
   std::shared_ptr<KeyValueStoreInterface> kv_store = options.kv_store;
   if (options.enable_mock_nccl) {
-    kv_store = std::make_shared<InMemoryKeyValueStore>();
+    kv_store = {};
   }
-  TF_RET_CHECK(options.num_nodes == 1 || kv_store != nullptr);
+  TF_RET_CHECK(options.enable_mock_nccl || options.num_nodes == 1 ||
+               kv_store != nullptr);
   TF_ASSIGN_OR_RETURN(
       DeviceTopologyPair device_topology_pair,
       BuildDistributedDevices(
--- a/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
@@ -1488,6 +1488,11 @@ TEST(StreamExecutorGpuClientTest, MockNcclClientWithGpuTopologyExecuteTest) {
   options.executable_build_options.set_num_partitions(8)
       .set_use_spmd_partitioning(true)
       .set_allow_spmd_sharding_propagation_to_output({true});
+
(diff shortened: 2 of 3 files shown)


jaro-sevcik:
@bchetioui Could you take a look?

What would be the best way to test this? Should we just duplicate one of the [autotuner tests](https://github.com/openxla/xla/blob/xla/service/gpu/autotuning/gemm_fusion_autotuner_test.cc) and check that the autotuner works with process_count > 1 and kv_store == null?

bchetioui:
Hi @jaro-sevcik.

I like the change in spirit, but agree this is not testable. How about we keep the logic in `gemm_fusion_autotuner.cc` as it was before (returning an error) and check that this is raised if `xla_gpu_shard_autotuning` is set at the same time as we use mock NCCL? That way, we'll need to also turn off `xla_gpu_shard_autotuning` whenever we use mock NCCL, but that seems reasonable tbh.

(Sorry for the delay, I was out of office when this was assigned to me---and just got a notification about it now.)

jaro-sevcik:
> Hi @jaro-sevcik.
> 
> I like the change in spirit, but agree this is not testable. How about we keep the logic in `gemm_fusion_autotuner.cc` as it was before (returning an error) and check that this is raised if `xla_gpu_shard_autotuning` is set at the same time as we use mock NCCL? That way, we'll need to also turn off `xla_gpu_shard_autotuning` whenever we use mock NCCL, but that seems reasonable tbh.
> 
> (Sorry for the delay, I was out of office when this was assigned to me---and just got a notification about it now.)

Thank you for getting back to me, your suggestion LGTM. I am a bit busy with other things now, but I'll try what you suggest in a couple of weeks.

jaro-sevcik:
Hi @bchetioui. I have updated the patch to raise an error when sharding autotuning with mock NCCL enabled. Is that what you had in mind?

PatriosTheGreat (approved):

golechwierowicz:
Following tests are failing

```
1. xla/pjrt/gpu:se_gpu_pjrt_client_test_2gpu_h100
2. xla/tools/multihost_hlo_runner:functional_hlo_runner_test_2gpu_h100
```

Can you please fix?

PR #28985, opened 2025-07-16
[XLA:GPU] Add shared_memory_per_block_optin device info member

andportnoy:
This allows us to query the amount of available shared memory in JAX, and use it to inform codegeneration of custom kernels.


xla/pjrt/gpu/se_gpu_pjrt_client.h
```diff
@@ -66,7 +66,8 @@ class StreamExecutorGpuDevice : public PjRtStreamExecutorDevice {
                           std::unique_ptr<LocalDeviceState> local_device_state,
                           std::string device_kind, std::string device_vendor,
                           std::string compute_capability, int core_count,
-                          int node_id, int slice_index = 0);
+                          int node_id, int shared_memory_per_block_optin,
```
bchetioui: nit: maybe move this before the node id so that the topology info can remain grouped together?

andportnoy: Done.

bchetioui (approved):
Thanks!

PR #28997, opened 2025-07-16
[Perf] Add expensive AllGather cost adjustment to default GPU Scheduler

andoorve:
All-gathers are also 1/2 as expensive as all-reduces, and should therefore be weighted accordingly.

### DIFF
--- a/xla/service/gpu/gpu_latency_hiding_scheduler.cc
+++ b/xla/service/gpu/gpu_latency_hiding_scheduler.cc
@@ -46,9 +46,11 @@ namespace {
 
 // A threshold for which we consider AR to be costly perf-wise.
 static constexpr int64_t kCostlyAllReduceThreshold = 30 * 1024 * 1024;
+static constexpr int64_t kCostlyAllGatherThreshold = 30 * 1024 * 1024;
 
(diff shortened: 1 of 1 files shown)


golechwierowicz (commented):
Can you add a test?

andoorve:
@golechwierowicz I took a look at https://github.com/openxla/xla/blob/main/xla/service/gpu/gpu_latency_hiding_scheduler_test.cc but don't see anything similar. Would you be able to point me to a similar test, for the expensive all reduce for example?

andoorve:
@golechwierowicz Gentle ping on this

golechwierowicz:
@frgossen can you take over?

frgossen (changes requested):
This PR changes the cost model, which we're trying to deprecate eventually. The extra cost for all-reduce was added, I believe, for data parallel training. 

While I understand the intention behind this change, we need to measure the performance impact of this. Can you run the benchmarks in xla/tools/benchmarks/hlo/... and report performance improvements regressions. 

What workload motivates your changes here?

golechwierowicz:
Closing this PR as we got rid of a T-shirt sized cost model a while ago.

PR #29073, opened 2025-07-18
SPMD Partitioning for MX Block Scaled Dots

philipphack:
Enables the SPMD partitioning of custom calls representing [block-scaled dots based on microscaling (MX) formats](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf) of the form

(A, B, X_A, X_B) -> D,

where X_A and X_B are the tensor block scales of dot operands A and B.

The implementation extends the present partitioning rules for regular dots to the block-scaled case by templating the functions of the SPMD dot handler. This is a non-functional change for convolutions and dots not operating on MX types.

### DIFF
--- a/xla/service/spmd/BUILD
+++ b/xla/service/spmd/BUILD
@@ -30,6 +30,7 @@ cc_library(
     hdrs = [
         "convolution_handler.h",
         "custom_call_handler.h",
+        "dot_handler.h",
         "spmd_partitioner.h",
         "spmd_partitioner_util.h",
     ],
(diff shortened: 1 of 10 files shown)


frgossen (changes requested):
Hi Philipp, 
Can you address the build failures in the CI, add an end to end execution test? 
Most importantly, which relevant workload benefits from this change? Can you provide a HLO to benchmark this?

frgossen (changes requested):

Some general asks for all PRs  (that also apply to this PR):

rosiezou (commented):
@philipphack please sync this PR to the latest copy in `main`, and then add any applicable tests in `dot_handler_test.cc`. In addition to @frgossen's suggestion of adding execution tests, please make sure that you include at least one execution test that demonstrates conversion from vector of scalar elements to MX-compliant format. For the execution test, it is sufficient to show two different HLOs that produce the same numerical results in the end. The tests you added in this current PR such as `MXCustomCall_BatchAndBatch` would be good candidates for execution tests. Also, is there an open PR and/or a recently closed PR that implements section 6.3 of [this document](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf)?

rosiezou (changes requested):
requesting changes on test coverage. please see previous comment for more details

rosiezou (approved):

PR #29116, opened 2025-07-21
Set names for XLA threads

jaro-sevcik:
Let us call `pthread_setname_np` for XLA threads so that the thread names are visible in gdb.

I should say that `pthread_setname_np` is not posix, but it should be provided on Linux and Mac.
If there is concern about this, we could put some `#if` around this.

### DIFF
--- a/xla/tsl/platform/default/env.cc
+++ b/xla/tsl/platform/default/env.cc
@@ -84,6 +84,9 @@ class PThread : public Thread {
     CHECK_EQ(ret, 0) << "Thread " << name
                      << " creation via pthread_create() failed.";
     pthread_attr_destroy(&attributes);
+  #if !defined(__APPLE__)
+    pthread_setname_np(thread_, name.c_str());
+  #endif
   }
 
   ~PThread() override {


dimitar-asenov (approved):

dimitar-asenov:
@jaro-sevcik Sorry for the delay on this. The current code fails to build on IOS 15 and possibly other IOS and Mac versions. The issue seems to be that on those platforms the signature for `pthread_setname_np` seems to only take one argument.

You can see the definition here:
https://github.com/apple-oss-distributions/libpthread/blob/main/include/pthread/pthread.h#L533-L535

Please add some macros to handle these cases. Thanks!

jaro-sevcik:
@dimitar-asenov I have added some conditional to avoid calling `pthread_setname_np` on Mac. Is that what you had in mind?

dimitar-asenov:
Yes, that's right. Thanks for making the change. We merged it.

wffurr:
This broke WebAssembly builds using XLA.  You should limit the macro usage to `#ifdef __linux__`.  XLA project, please consider adding a Wasm presubmit build.

PR #29157, opened 2025-07-22
[XLA:GPU] Add LHS schedule mode for command buffer

shawnwang18:
This PR adds LHS (Latency Hidding Scheduler) mode for command buffer scheduling, it will uses the same LHS scheduling results in the non-cuda-graph run.  For some cases, command buffer's automatic scheduling mode may not generate best perf, and user may want to try the legacy LHS scheduling.

### DIFF
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -255,10 +255,22 @@ message DebugOptions {
   // XLA:GPU options.
   //--------------------------------------------------------------------------//
   // clang-format off
-  // go/keep-sorted start newline_separated=yes skip_lines=2 ignore_prefixes=["optional AutotuneCacheMode","optional bool","optional float","optional int32","optional int64","optional LibNvJitLinkMode","map<string, string>","optional PGLEStrictnessLevel","optional PipelineParallelismOptLevel","repeated CollectiveOpType","repeated CommandBufferCmdType","repeated string","optional ShapeChecks","optional string","optional WhileLoopUnrolling","reserved","repeated GenericTritonEmitterFeature"] // NOLINT
+  // go/keep-sorted start newline_separated=yes skip_lines=2 ignore_prefixes=["optional AutotuneCacheMode","optional bool","optional float","optional int32","optional int64","optional LibNvJitLinkMode","map<string, string>","optional PGLEStrictnessLevel","optional PipelineParallelismOptLevel","repeated CollectiveOpType","repeated CommandBufferCmdType","repeated string","optional ShapeChecks","optional string","optional WhileLoopUnrolling","reserved","repeated GenericTritonEmitterFeature","optional CommandBufferSchedulingMode"] // NOLINT
   // clang-format on
 
   reserved 160;  // Was xla_gpu_enable_cudnn_frontend
+
+  // Command buffer scheduling mode. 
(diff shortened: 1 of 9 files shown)


xla/xla.proto
```diff
@@ -687,8 +687,16 @@ message DebugOptions {
   // profiling and debugging.
   optional bool xla_gpu_generate_line_info = 349;
 
-  // Identify concurrent regions in GPU graphs and execute them concurrently.
-  optional bool xla_gpu_graph_enable_concurrent_region = 215;
+  // Command buffer scheduling mode. 
+  // SERIALIZE: Serialize all commands in a command buffer.
+  // CONCURRENT: Identify concurrent across operator through data conflicts.
+  // LHS: Uses the same latency hidden scheulding results.
```
ezhulenev: typo

ezhulenev: Same as what? Have troubles parsing the documentation :)

ezhulenev (approved):

xla/xla.proto
shawnwang18: Thanks, fixed the comments.

xla/xla.proto
```diff
@@ -687,8 +687,18 @@ message DebugOptions {
...
+  // CONCURRENT: Identify concurrent across operator through data conflicts.
+  // LHS: Uses the same latency hidden scheduling results as adopted by
+  // LatencyHidingScheduler which hides latencies of operations that can run in
+  // parallel with other operations.
+  enum CommandBufferSchedulingMode {
+    SERIALIZE = 0;
+    CONCURRENT = 1;
+    LHS = 2;
+  }
+  optional CommandBufferSchedulingMode xla_gpu_command_buffer_scheduling_mode = 215;
```
penpornk: Our flag names are sorted alphabetically. Could you please make the following edits?

1. Move the enum definition to before `enum GenericTritonEmitterFeature` in [line 265](https://github.com/openxla/xla/blob/0ef8c05223a25c9d9c925ac1a508857fbb37baac/xla/xla.proto#L265).

2. Move the `xla_gpu_command_buffer_scheduling_mode` declaration to between `xla_gpu_collectives_use_persistent_cliques` ([line 370](https://github.com/openxla/xla/blob/0ef8c05223a25c9d9c925ac1a508857fbb37baac/xla/xla.proto#L370))and `xla_gpu_copy_insertion_use_region_analysis` ([line 372](https://github.com/openxla/xla/blob/0ef8c05223a25c9d9c925ac1a508857fbb37baac/xla/xla.proto#L372)).

3. Add "optional CommandBufferSchedulingMode" to the keep-sorted ignored keywords in line 258. (So that our checker will only sort the lines based on var name `xla_gpu_command_buffer_scheduling_mode`.)
https://github.com/openxla/xla/blob/0ef8c05223a25c9d9c925ac1a508857fbb37baac/xla/xla.proto#L258

shawnwang18: Thanks for your feedback, I made some change from your comments.

akuegel (approved):

akuegel (approved):

penpornk:
Just a quick update that this PR is undergoing an additional flag review internally. (Must be done each time an XLA flag is added/modified.) This could take a while.

PR #29204, opened 2025-07-23
[XLA:GPU] Remove Stream Id from command buffer

shawnwang18:
Command buffer now uses DAG to specify dependency, so stream id is totally not used, this PR remove the Stream ID from command buffer implementation.


ezhulenev (approved):

akuegel (approved):
Approving again to get this merged.

PR #29270, opened 2025-07-24
[GPU] Bubble up mismatched buffer color from donation

terryysun:
When user specifies to donate a input buffer while giving different output memory space via `out_shardings`, XLA buffer assignment check will fail. Because the error message is not bubbled up, JAX will just throw `Aborted (core dumped)`. This PR bubbles up the buffer assignment check message.

Specifically, compared to silently resolving this in XLA, we prefer a bubbled up message to the user. So it has two cases:
A. User specifies output memory space and donation, but provides memory space via `in_shardings` as well (user knows it's not possible to donate the buffer), we silently handle it and no error will be thrown.
B. User specifies output memory space and donation, we throw error to let user know that the donation is not possible.


terryysun:
JAX snippet to reproduce the error:
```
import jax
import jax.numpy as jnp
from jax.sharding import Mesh, NamedSharding as NS, PartitionSpec as P


mesh = Mesh(jax.devices(), ('x',))
s_host = NS(mesh, P()).with_memory_kind('pinned_host')
f = lambda x: jax.device_put(x, device=s_host)
f_jit = jax.jit(f, donate_argnums=(0,), out_shardings=s_host)
a0 = jnp.ones((128, 128))
print(f_jit.lower(a0).compile())
```

The updated message is manually checked but non-trivial to reproduce in XLA unit test. Preset assignment is probably the easiest way to approach but XLA has rigorous checks for adding it.

xla/service/buffer_assignment.cc
```diff
@@ -1941,8 +1954,8 @@ absl::Status BufferAssigner::AssignBuffersWithSequentialOrdering(
                   get_heap_algorithm(alignment), *private_stack_computation,
                   *instruction_sequence, assignment->alias_analysis(),
                   alias_info_, assignment->buffer_size_, &schedule, options));
-          AssignBuffersFromHeapSimulator(result, assignment, color,
-                                         isolation_options);
+          std::ignore = AssignBuffersFromHeapSimulator(
```
akuegel: Why do we ignore the returned status here? My guess is this would cause issues. At the very least, we should preserve the previous behavior and CHECK fail if the returned status here is not ok.

xla/service/buffer_assignment.cc
```diff
@@ -1952,8 +1965,8 @@ absl::Status BufferAssigner::AssignBuffersWithSequentialOrdering(
                                assignment->module(), schedule,
                                assignment->alias_analysis(), alias_info_,
                                assignment->buffer_size_, options));
-        AssignBuffersFromHeapSimulator(result, assignment, color,
-                                       isolation_options);
+        std::ignore = AssignBuffersFromHeapSimulator(result, assignment, color,
```
akuegel: Same here, we should preserve the previous behavior and CHECK fail if the returned status here is not ok.

xla/service/buffer_assignment.cc
```diff
@@ -1986,8 +1999,8 @@ absl::Status BufferAssigner::AssignBuffersWithSequentialOrdering(
                                *instruction_sequence,
                                assignment->alias_analysis(), alias_info_,
                                assignment->buffer_size_, options));
-        AssignBuffersFromHeapSimulator(result, assignment, color,
-                                       isolation_options);
+        std::ignore = AssignBuffersFromHeapSimulator(result, assignment, color,
```
akuegel: Same here, we should preserve the previous behavior and CHECK fail if the returned status here is not ok.

xla/service/buffer_assignment.cc
```diff
@@ -2332,7 +2347,8 @@ BufferAssigner::CreateAssignment(
     private_stack_colors.insert(color);
   }
 
-  assignment->CombineTempAllocations(private_stack_colors, temp_buffer_color);
+  std::ignore = assignment->CombineTempAllocations(private_stack_colors,
```
akuegel: Same here, we should preserve the previous behavior and CHECK fail if the returned status here is not ok.

akuegel (changes requested):
I agree that just CHECK failing when it is actually a user error is not nice. At the same time, we can also not allow that due to a user error, we run into a segfault or similar crash somewhere, a CHECK failure would still be better. So if it is too much work to bubble up all the errors at once in one PR, please at least add CHECK failures instead of using std::ignore.

akuegel (commented):
Thanks, looking good now. Can you please also add a test that shows that an error is bubbled up?

terryysun:
> Thanks, looking good now. Can you please also add a test that shows that an error is bubbled up?

Added a test to catch the error, could you take another look? Thanks!

akuegel (approved):
Thank you :)

PR #29392, opened 2025-07-28
[XLA:CPU][oneDNN][BugFix] Fix a failing FP16 matmul test

othakkar:
This PR fixes a failing FP16 matmul test (`SimpleTestF16WithBiasAndAddFusion`). SUM post-op works for FP16 (just not for BF16, as element-wise addition is not allowed by oneDNN for BF16), and the correct optimized HLO must compare with `fused_matmul_bias_add_sum_str_`.

### DIFF
--- a/xla/service/cpu/tests/onednn_matmul_test.cc
+++ b/xla/service/cpu/tests/onednn_matmul_test.cc
@@ -206,6 +206,10 @@ TEST_F(MatmulTest, SimpleTestBF16) {
   })";
 
   EXPECT_TRUE(RunAndCompare(matmul_module_str, ErrorSpec{1e-2, 1e-4}));
+  // For BF16, we match the optimized HLO with HLO containing
+  // BINARY_ADD instead of SUM as SUM post-op does not work for
+  // BF16 because element-wise addition is not allowed due to
+  // precision constraints by oneDNN.
   MatchOptimizedHlo(matmul_module_str, matmul_rewrite_str_);
 }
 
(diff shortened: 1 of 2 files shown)


othakkar:
@penpornk, please review this tiny bug fix PR when you get a chance. Thanks!

xla/service/cpu/tests/onednn_matmul_test.cc
```diff
@@ -1795,7 +1795,7 @@ TEST_F(MatmulTest, SimpleTestF16WithBiasAndAddFusion) {
   const std::string matmul_module_str =
       CreateMatmulBiasAddAndAddModuleText("f16", "f16");
   EXPECT_TRUE(RunAndCompare(matmul_module_str, ErrorSpec{1e-2, 1e-4}));
-  MatchOptimizedHlo(matmul_module_str, fused_matmul_bias_add_str_);
+  MatchOptimizedHlo(matmul_module_str, fused_matmul_bias_add_sum_str_);
```
penpornk, line 1802: Should the one in `SimpleTestBF16WithBiasAndAddFusion` (line 1787) be fixed too?

penpornk (commented):
Thank you for the PR!

xla/service/cpu/tests/onednn_matmul_test.cc
othakkar: `SimpleTestBF16WithBiasAndAddFusion` test matches the correct HLO. SUM post-op does not work for BF16 because element-wise addition is not allowed due to precision constraints by oneDNN. Hence, the SUM use case is fused as BINARY_ADD for BF16.

penpornk: Thank you for the clarification! Could you please help document this in the code (as comments)?

penpornk (changes requested):

xla/service/cpu/tests/onednn_matmul_test.cc
othakkar: Done!

penpornk (approved):
Thank you very much!

PR #28883, opened 2025-07-14
[XLA:CPU][oneDNN] Add build flag to enable asynchronous support in oneDNN

othakkar:
This PR optionally enables the oneDNN version that provides asynchronous execution support using a build flag.

PR #29459, opened 2025-07-29
[XLA:CPU][oneDNN] Refactor OneDnnThreadPool to support asynchronous execution using ParallelLoopRunner

othakkar:
**This PR depends on #28883.**

This PR refactors `OneDnnThreadPool` to support asynchronous execution via `ParallelLoopRunner`, enabling integration with the thunk runtime and oneDNN's async custom calls.

**Key updates**:
- Adds a new constructor to `OneDnnThreadPool` that accepts a `ParallelLoopRunner` instance.
- Updates `onednn_threadpool_test` to use a `ParallelLoopRunner`.
- When using a runner-backed thread pool, `OneDnnThreadPool` exposes the `ASYNCHRONOUS` flags and implements `wait()` to block on `runner->done_event()`.

This runner-based interface is necessary for future support of asynchronous oneDNN custom calls through the FFI handler, which will return a `future` (via `runner->done_event()`) to XLA’s async runtime.


othakkar:
@penpornk, please review this PR when you get a chance. Thanks!

xla/backends/cpu/runtime/onednn/onednn_threadpool_test.cc
```diff
@@ -100,6 +109,10 @@ TEST(OneDnnThreadPoolTest, Binary) {
   // Execute compiled oneDNN graph on the CPU stream.
   compiled_partitions[0].execute(stream, {src}, {dst});
 
+#ifdef ENABLE_ONEDNN_ASYNC
+  tsl::BlockUntilReady(runner.done_event());
```
penpornk: Let's call `threadpool.wait()` instead.

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
```diff
@@ -33,16 +35,60 @@ class OneDnnThreadPool final
  public:
   explicit OneDnnThreadPool(Eigen::ThreadPoolInterface* thread_pool)
       : thread_pool_(thread_pool) {}
+  explicit OneDnnThreadPool(ParallelLoopRunner* runner) : runner_(runner) {
+    use_runner_ = true;
+    dnnl_threadpool_interop_set_max_concurrency(runner_->num_threads());
+  }
```
penpornk: Since we can create `ParallelLoopRunner` from `Eigen::ThreadPoolInterface` and no `OneDnnThreadPool` users need to directly access `ParallelLoopRunner` anyway,  I think we can just pass a `bool async` parameter instead. E.g., change the existing constructor to

```suggestion
  explicit OneDnnThreadPool(Eigen::ThreadPoolInterface* thread_pool, bool is_async=false)
      : thread_pool_(thread_pool), is_async_(is_async) {
      if (is_async_) {
        runner_ = ...;  // Create runner.
        dnnl_threadpool_interop_set_max_concurrency(runner_->num_threads());
      }
  }
```

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
```diff
@@ -57,6 +103,13 @@ class OneDnnThreadPool final
 
  private:
   Eigen::ThreadPoolInterface* thread_pool_;
+  ParallelLoopRunner* runner_;
+  bool use_runner_ =
+      false;  // indicates if we are using ParallelLoopRunner
+              // which is used to return a future from the
+              // oneDNN custom call's FFI handler for asynchronous execution.
+              // TODO(intel-tf): Remove this flag when
+              // oneDNN supports asynchronous execution by default.
```
penpornk: Nit: The formatting is weird here. We prefer long comments to have their own lines, e.g.,
```suggestion
  // Indicates if we are using ParallelLoopRunner to return a future...
  // ...
  bool use_runner_ = false;  
```

But if we take `is_async` as a parameter, `use_runner_` should become `is_async_` instead.

penpornk (changes requested):
Thank you for the PR!

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
othakkar: The oneDNN handlers (e.g., `ExecuteOneDnnMatMul`) that will be added in the future PRs will need to directly access the `runner` object to return a `future` (`runner->done_event()`) to the [`CallAsync(...)` method](https://github.com/openxla/xla/blob/c6e57c046f1d938dd15770427f701150a8e64265/xla/ffi/ffi_api.cc#L238).

Similar to the earlier "non-blocking" version of `onednn_fusion_thunk.cc` - https://github.com/openxla/xla/blob/81335eabdc55f0a2ff02bde7f79da7adda7af2c9/xla/backends/cpu/runtime/onednn/onednn_fusion_thunk.cc#L108

othakkar: @penpornk, a friendly ping to check if using a new constructor that accepts a `ParallelLoopRunner` is still acceptable.

othakkar: @penpornk, a friendly ping again :)

penpornk:
@othakkar FYI, we have deleted `ParallelLoopRunner` (https://github.com/openxla/xla/commit/9fcbfb093b2bf35f8ece431747127f60176123ac) because neither XNNPACK nor oneDNN needs it now. oneDNN only needs simple 1D parallelization and can just call `Worker::Parallelize` directly instead. Here is an example:
https://github.com/openxla/xla/blob/62ccb4ab2367f186165937d257b904c4babd30f0/xla/backends/cpu/runtime/convolution_thunk_internal.h#L386-L396

 (XNNPACK used to need it, but doesn't anymore with their new runtime.)

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
penpornk: Sorry for the delay! Please see my reply here: https://github.com/openxla/xla/pull/29459#issuecomment-3168301986

othakkar:
@penpornk I updated the PR to remove the use of the `ParallelLoopRunner`. Please review again when you have a chance. Thanks!

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
```diff
@@ -28,32 +29,77 @@ limitations under the License.
...
   }
 
-  uint64_t get_flags() const final { return 0; }
+  uint64_t get_flags() const final {
+    if (is_async_) {
+      // set the ASYNCHRONOUS flag when using oneDNN
+      // with asynchronous execution support.
+      return ASYNCHRONOUS;
+    }
+    return 0;
```
penpornk: Nit: If this function isn't going to get more complex in the future, we can just use:
```suggestion
    return is_async_ ? ASYNCHRONOUS : 0;
```

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
```diff
@@ -28,32 +29,77 @@ limitations under the License.
...
+      // If we are using oneDNN with async support, we need to
+      // schedule the parallel loop using the done_event_.
+      // This allows us to return immediately and not block the caller thread.
+      auto parallelize = [this, n, fn](tsl::Chain) {
+        return Worker::Parallelize(
+            thread_pool_, thread_pool_->NumThreads(), n,
+            [fn, n](size_t i) { fn(static_cast<int>(i), n); });
+      };
+
+      done_event_ = done_event_.FlatMap(parallelize);
```
penpornk: `Worker::Parallelize` already returns an `AsyncValueRef`. Can't we just set `done_event_` to it? 
The caller should be responsible for chaining the events anyway.

```suggestion
      done_event_ = Worker::Parallelize(
            thread_pool_, thread_pool_->NumThreads(), n,
            [fn, n](size_t i) { fn(static_cast<int>(i), n); });
```

`Worker::Parallelize` itself is not blocking. We make it blocking by calling `tsl::BlockUntilReady` on the returned async value (like in line 105).

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
```diff
@@ -28,32 +29,77 @@ limitations under the License.
...
-  // TODO(intel-tf): Implement proper wait logic when thunk runtime
-  // with oneDNN is enabled.
-  void wait() final {}
+  // The wait() method only exists with oneDNN's experimental support
+  // for asynchronous execution determined by the ENABLE_ONEDNN_ASYNC.
+  void wait() override {
+    if (is_async_) {
+      // While performing asynchronous execution, wait() method is
+      // needed to notify the user that the output is ready.
+      // oneDNN will not call wait() inside the library to avoid deadlock.
```
penpornk: Nit: Could you please make the comments fill the full line before beginning each new line? Same for other comments in this PR.
```suggestion
      // While performing asynchronous execution, wait() method is needed to
      // notify the user that the output is ready. oneDNN will not call wait()
      // inside the library to avoid deadlock.
```

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
```diff
@@ -65,8 +111,21 @@ class OneDnnThreadPool final
...
+  tsl::AsyncValueRef<tsl::Chain> done_event() const {
+    return done_event_;
+  }
+
  private:
   Eigen::ThreadPoolInterface* thread_pool_;
+
+  // indicates if we are using oneDNN with async support.
+  // TODO(intel-tf): Remove this flag when
+  // oneDNN supports asynchronous execution by default.
```
penpornk: Nit: Please capitalize the beginning of a sentence.
```suggestion
  // Indicates if we are using oneDNN with async support.
  // TODO(intel-tf): Remove this flag when oneDNN supports asynchronous
  // execution by default.
```

penpornk (changes requested):
Thank you for the changes and sorry for the delay!

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
othakkar: Done.

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
othakkar: Done.

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
othakkar: Done.

xla/backends/cpu/runtime/onednn/onednn_threadpool.h
penpornk: We chatted offline and will proceed with this workaround.

penpornk (approved):
Thank you again for the PR!

PR #29506, opened 2025-07-30
[XLA:CPU][oneDNN] Refine conditions for Transpose Folding

akhilgoe:
This PR avoids folding transposes in cases where oneDNN does not dispatch to optimized algorithms. We can revisit these restrictions as and when oneDNN optimizes these cases.

Most likely fixes: https://github.com/jax-ml/jax/issues/29070

### DIFF
--- a/xla/service/cpu/onednn_contraction_rewriter.cc
+++ b/xla/service/cpu/onednn_contraction_rewriter.cc
@@ -997,8 +997,9 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
       // oneDNN does not dispatch to optimized algorithm in these cases, hence
       // avoid folding transposes here.
       // TODO(intel-tf): Reevaluate this condition with future oneDNN releases.
-      if (rank >= 5 || transpose->dimensions()[rank - 1] != rank - 1)
+      if (rank >= 5 || transpose->dimensions()[rank - 1] != rank - 1) {
         return absl::OkStatus();
+      }
       auto backend_config = custom_call->backend_config<BackendConfig>();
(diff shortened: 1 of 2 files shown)


xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -993,6 +993,12 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
     if (Match(instr,
               m::Copy(&copy, m::Transpose(&transpose,
                                           OneDnnMatmulInstr(&custom_call))))) {
+      auto rank = transpose->dimensions().size();
+      // oneDNN does not dispatch to optimized algorithm in these cases, hence
+      // avoid folding transposes here.
+      // TODO(intel-tf): Reevaluate this condition with future oneDNN releases.
+      if (rank >= 5 || transpose->dimensions()[rank - 1] != rank - 1)
+        return absl::OkStatus();
```
penpornk: Nit: Let's add braces even for single-line `if`s.
```suggestion
      if (rank >= 5 || transpose->dimensions()[rank - 1] != rank - 1) {
        return absl::OkStatus();
      }
```

penpornk (changes requested):
Thank you for the fix!

xla/service/cpu/onednn_contraction_rewriter.cc
akhilgoe: Done

akhilgoe:
Thanks @penpornk for the review. I have added a commit to address your comment. Please take a look.

penpornk (approved):
Thank you again for the PR and sorry for the delay!

PR #29511, opened 2025-07-30
SPMD Dot Tests

philipphack:
Adds end-to-end tests for the SPMD partitioning of dots.

### DIFF
--- a/xla/tests/collective_ops_e2e_test.cc
+++ b/xla/tests/collective_ops_e2e_test.cc
@@ -59,6 +59,7 @@ limitations under the License.
 #include "xla/tsl/platform/test.h"
 #include "xla/types.h"
 #include "xla/xla_data.pb.h"
+#include "tsl/platform/regexp.h"
 
 namespace xla {
 namespace {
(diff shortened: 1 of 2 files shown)


xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,282 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+    config.set_num_partitions(num_partitions);
+    TF_ASSERT_OK_AND_ASSIGN(auto module,
+                            ParseAndReturnVerifiedModule(hlo_text, config));
+
+    // Gather parameter and return value sharding information.
+    std::vector<std::vector<int64_t>> param_dims(num_parameters);
+    std::vector<std::vector<int64_t>> param_dims_per_shard(num_parameters);
+    std::vector<std::vector<int64_t>> param_sharded_dims(num_parameters);
+    for (int i = 0; i < num_parameters; ++i) {
+      param_dims[i] = std::vector(module->entry_computation()
```
frgossen: Can you extract the call to `module->entry_computation()
                                      ->parameter_instruction(i)
                                      ->shape()
                                      .dimensions()` and reuse it?

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,282 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+             k < parameter_sharding.tile_assignment().num_dimensions(); ++k) {
+          if (parameter_sharding.tile_assignment().dim(k) > 1) {
+            param_dims_per_shard[i][k] /=
+                parameter_sharding.tile_assignment().dim(k);
+            param_sharded_dims[i].push_back(k);
+          }
+        }
+      }
+    }
+    std::vector<int64_t> root_dims(module->entry_computation()
```
frgossen: same

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,282 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+                                      ->parameter_instruction(i)
+                                      ->shape()
+                                      .dimensions()
+                                      .begin(),
+                                  module->entry_computation()
+                                      ->parameter_instruction(i)
+                                      ->shape()
+                                      .dimensions()
+                                      .end());
+      param_dims_per_shard[i] = param_dims[i];
```
frgossen, line 1426: param_dims_per_shard could be generated in a helper function. Think that would make it easier to read.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,282 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+                                       ->root_instruction()
+                                       ->shape()
+                                       .dimensions()
+                                       .begin(),
+                                   module->entry_computation()
+                                       ->root_instruction()
+                                       ->shape()
+                                       .dimensions()
+                                       .end());
+    std::vector<int64_t> root_dims_per_shard = root_dims;
```
frgossen: This could be a helper function too. Likely possible to share code with the above for parameters.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,282 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
   EXPECT_FALSE(IsAsync(all_reduce));
 }
 
+// E2E tests comparing the results of sharded and unsharded execution.
+class CollectiveOpsTestE2EShardedUnsharded : public CollectiveOpsTestE2E {
+ public:
+  void CollectiveOpsCompareShardedUnsharded(std::string hlo_text) {
```
frgossen: This is a lot of code for a test constructor. Can you factor the two executuons out into two helper functions, one for the sharded version, one for the unsharded version? Maybe add a 3rd helper function to assert the equality and use that in the test cases rather than using this constructor.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,282 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+
+ENTRY entry {
+  lhs = f32[4,16,8]{2,1,0} parameter(0), sharding={devices=[4,1,1]<=[4]}
+  rhs = f32[4,4,8]{2,1,0} parameter(1), sharding={devices=[4,1,1]<=[4]}
+  ROOT dot = f32[4,16,4]{2,1,0} dot(lhs, rhs), lhs_batch_dims={0}, rhs_batch_dims={0}, lhs_contracting_dims={2}, rhs_contracting_dims={2}, sharding={devices=[1,4,1]<=[4]}
+})";
+  CollectiveOpsCompareShardedUnsharded(hlo_text);
+}
+
+TEST_F(CollectiveOpsTestE2EShardedUnsharded, DotBatchAndNonContracting) {
```
frgossen, line 1558: Can you update these test cases to use only two GPUs?

frgossen (changes requested):
Thanks for adding these test cases. This is very appreciated! 

Left a few comments. Most importantly, it would be ideal to have as many test cases as possible on 2 GPUs to save resources.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,267 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+  rhs = f32[1,4]{1,0} parameter(1), sharding={replicated}
+  ROOT dot = f32[4,1]{1,0} dot(lhs, rhs), lhs_contracting_dims={1}, rhs_contracting_dims={1}, sharding={devices=[2,1]<=[2]}
+})";
+  CollectiveOpsCompareShardedUnsharded(hlo_text);
+}
+
+TEST_F(CollectiveOpsTestE2EShardedUnsharded,
+       DotContractingNonContractingAndContractingNonContracting) {
+  const std::string hlo_text = R"(
+HloModule module, entry_computation_layout={(f32[16,8]{1,0}, f32[4,8]{1,0})->f32[16,4]{1,0}}, num_partitions=4
```
frgossen, line 1621: Can this be on 2 GPUs too?

xla/tests/collective_ops_e2e_test.cc
frgossen: Addressed.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,267 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
   EXPECT_FALSE(IsAsync(all_reduce));
 }
 
+// E2E tests comparing the results of sharded and unsharded execution.
+class CollectiveOpsTestE2EShardedUnsharded : public CollectiveOpsTestE2E {
+ public:
+  void CollectiveOpsCompareShardedUnsharded(const std::string& hlo_text,
```
frgossen, line 1347: I think RunAndCompareShardedAndUnsharded would be a good name. 
I still think breaking this into three functions would be a good idea.

frgossen (changes requested):

xla/tests/collective_ops_e2e_test.cc
philipphack: We need four devices to test with two sharded dimensions. Alternatively, we could remove this test.

xla/tests/collective_ops_e2e_test.cc
philipphack: The current name and structure follow CollectiveOpsTestE2EWindowedNonWindowed which compares windowed and non-windowed dots.

xla/tests/collective_ops_e2e_test.cc
frgossen: Makes sense. Thanks for checking.

frgossen (changes requested):

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,290 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+      }
+    } else {
+      EXPECT_TRUE(
+          LiteralTestUtil::Near(ref_results[0], results[0], error_spec));
+    }
+  }
+
+ private:
+  // Execute the replicated case.
+  absl::StatusOr<std::vector<Literal>> ExecuteReplicated(
```
frgossen: You are executing the unsharded case here, right? This is not replicated, is it? num_replicas=1 below. 
Maybe call this function `ExecuteUnsharded`

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,290 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+      const std::string& hlo_text) {
+    // Create the unsharded reference case by removing the sharding metadata
+    // from the HLO string.
+    std::string hlo_text_ref = hlo_text;
+    RE2::GlobalReplace(&hlo_text_ref, R"(, sharding=\{devices=\[[0-9,]*\].*\})",
+                       "");
+    RE2::GlobalReplace(&hlo_text_ref, R"(, sharding=\{replicated\})", "");
+
+    HloModuleConfig ref_config = GetModuleConfigForTest(/*replica_count=*/1);
+    auto ref_opts = GetDebugOptionsForTest();
```
frgossen: nit: Can you spell out the type instead of `auto`? Here and also in the other places.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,290 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+      param_dims[i] = std::vector(dimensions.begin(), dimensions.end());
+      param_dims_per_shard[i] = param_dims[i];
+      HloSharding parameter_sharding =
+          module->entry_computation()->parameter_instruction(i)->sharding();
+      EvaluateShardedDims(param_dims_per_shard[i], param_sharded_dims[i],
+                          parameter_sharding);
+    }
+
+    // Slice the tiled inputs to match the prescribed sharding.
+    auto fake_args = xla::MakeFakeArguments(module.get()).value();
```
frgossen, line 1434: Can you factor out the argument creation into its own function? It's fairly complex and completely separate from the rest of the function.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,290 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+    return HloTestBase::ExecuteReplicated(std::move(ref_module), ref_fake_ptrs,
+                                          /*num_replicas=*/1, &ref_assn,
+                                          /*run_hlo_passes=*/true,
+                                          /*use-threads=*/true);
+  }
+
+  // Execute the sharded case.
+  absl::StatusOr<std::vector<Literal>> ExecuteSharded(
+      const std::string& hlo_text, int64_t num_partitions,
+      int64_t num_replicas) {
```
frgossen: I think num_replicas is always 1 in this case. You are only dealing with multiple devices in the partition Dimension. You can just pass in a constant  1, e.g. `/*replica_count=*/1`

xla/tests/collective_ops_e2e_test.cc
frgossen: I dont think this is resolved.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,290 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+    }
+    return HloTestBase::ExecuteReplicated(std::move(module), fake_ptrs,
+                                          num_partitions,
+                                          /*run_hlo_passes=*/true, &assn);
+  }
+
+  void EvaluateShardedDims(std::vector<int64_t>& dims_per_shard,
+                           std::vector<int64_t>& sharded_dims,
+                           const HloSharding& sharding) {
+    if (!sharding.IsReplicated()) {
```
frgossen, line 1535: Is `IsReplicated` ever the case in this test file or can we simplify?

frgossen (changes requested):
PTAL

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,290 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+
+    // Slice the unsharded reference results and compare to the sharded case.
+    HloModuleConfig config =
+        GetModuleConfigForTest(/*replica_count=*/num_replicas);
+    auto opts = GetDebugOptionsForTest();
+    opts.set_xla_gpu_enable_triton_gemm(false);
+    config.set_debug_options(opts);
+    config.set_num_partitions(num_partitions);
+    TF_ASSERT_OK_AND_ASSIGN(auto module,
+                            ParseAndReturnVerifiedModule(hlo_text, config));
```
frgossen: Why parse the module here again? Could you just pass the needed info into the test case?

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,290 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+    TF_ASSERT_OK_AND_ASSIGN(std::vector<Literal> ref_results,
+                            ExecuteReplicated(hlo_text));
+    ASSERT_EQ(ref_results.size(), 1);
+
+    TF_ASSERT_OK_AND_ASSIGN(
+        std::vector<Literal> results,
+        ExecuteSharded(hlo_text, num_partitions, num_replicas));
+    ASSERT_EQ(results.size(), num_partitions);
+
+    // Slice the unsharded reference results and compare to the sharded case.
```
frgossen: I think this (" Slice the unsharded reference results and compare to the sharded case.") could also be a separate function with a descriptive name

frgossen (changes requested):

xla/tests/collective_ops_e2e_test.cc
philipphack: Passing the module into the functions isn't ideal since ParseAndReturnVerifiedModule returns a unique_ptr. Let me know if you want to change this anyway.

xla/tests/collective_ops_e2e_test.cc
philipphack: Some tests use replicated shardings.

frgossen (approved):
Thanks for addressing the comments. LGTM

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,295 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+
+ENTRY entry {
+  lhs = f32[16,8]{1,0} parameter(0), sharding={devices=[1,2]<=[2]}
+  rhs = f32[4,8]{1,0} parameter(1), sharding={devices=[1,2]<=[2]}
+  ROOT dot = f32[16,4]{1,0} dot(lhs, rhs), lhs_contracting_dims={1}, rhs_contracting_dims={1}, sharding={devices=[2,1]<=[2]}
+})";
+  CollectiveOpsCompareShardedUnsharded(hlo_text);
+}
+
+TEST_F(CollectiveOpsTestE2EShardedUnsharded, DotNonContractingAndContracting) {
```
penpornk, line 1582: This test crashed internally (running on 2 GPUs). Any quick guesses on why?
```
[ RUN      ] CollectiveOpsTestE2EShardedUnsharded.DotNonContractingAndContracting
*** SIGILL (@0x558221a2da23), received by PID 2659 (TID 2659) on cpu 8; stack trace: ***
xla::(anonymous namespace)::CollectiveOpsTestE2EShardedUnsharded::CollectiveOpsCompareShardedUnsharded()
FailureSignalHandler()
xla::(anonymous namespace)::CollectiveOpsTestE2EShardedUnsharded::CollectiveOpsCompareShardedUnsharded()
xla::(anonymous namespace)::CollectiveOpsTestE2EShardedUnsharded_DotNonContractingAndContracting_Test::TestBody()
testing::Test::Run()
testing::TestInfo::Run()
testing::TestSuite::Run()
testing::internal::UnitTestImpl::RunAllTests()
testing::UnitTest::Run()
main
...
```

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,295 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+        std::vector<int64_t> upper(param_dims_per_shard[k].begin(),
+                                   param_dims_per_shard[k].end());
+        for (int i = 0; i < num_partitions; ++i) {
+          fake_args_sliced[k].push_back(fake_args[k].Slice(lower, upper));
+          for (int m = param_sharded_dims.size() - 1; m >= 0; --m) {
+            if (upper[param_sharded_dims[k][m]] <
+                param_dims[k][param_sharded_dims[k][m]]) {
+              upper[param_sharded_dims[k][m]] +=
+                  param_dims_per_shard[k][param_sharded_dims[k][m]];
+              break;
```
frgossen, line 1449: Not the problem but this else is redundant.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -1340,6 +1341,295 @@ TEST_F(CollectiveOpsTestE2E, NoAsyncCollectives) {
...
+                                 root_dims_per_shard.end());
+      for (const Literal& result : results) {
+        Literal ref_results_slice = ref_results[0].Slice(lower, upper);
+        EXPECT_TRUE(
+            LiteralTestUtil::Near(ref_results_slice, result, error_spec));
+        for (int m = root_sharded_dims.size() - 1; m >= 0; --m) {
+          if (upper[root_sharded_dims[m]] < root_dims[root_sharded_dims[m]]) {
+            upper[root_sharded_dims[m]] +=
+                root_dims_per_shard[root_sharded_dims[m]];
+            break;
```
frgossen, line 1518: Not the problem but this else is redundant. 

Also, this code looks somewhat similar to the above. Do you see a way to factor this out into a function and share them impl?

frgossen (changes requested):

xla/tests/collective_ops_e2e_test.cc
philipphack: @penpornk all tests pass on my side.

philipphack: @penpornk can you try again with [f7bc551](https://github.com/openxla/xla/pull/29511/commits/f7bc551a60d741a52add7734dcaccad061d376ae)?

penpornk (approved):
This works. Thank you very much!

frgossen (approved):
Thanks for fixing this!

penpornk:
Could you please help resolve merge conflicts?

penpornk:
Never mind, I'll just merge conflicts internally. :)

PR #29622, opened 2025-07-31
[XLA:CPU][BugFix] Fix bug in populating oneDNN post-op arguments

akhilgoe:
This PR fixes a potential bug where an incorrect argument pod might get used in some long post-op chains.


penpornk (approved):
Thank you for the PR!

PR #29653, opened 2025-07-31
Changed the heading level for the TPU XLA flags section

mikegre-google:
Changed the heading level for the TPU XLA flags section so that it is at the same level as the GPU XLA flags section and will be displayed in the right-hand nav.


penpornk (approved):
Thank you for the fix!

PR #29766, opened 2025-08-04
[ROCm] Add bf16 starting from gfx11, bugfix & optimize RocmComputeCapability

Arech8:
* Bugfix and improve device_description.h::RocmComputeCompatibility

* Enable ALG_DOT_BF16* on rocm with HW support

### DIFF
--- a/xla/stream_executor/device_description.h
+++ b/xla/stream_executor/device_description.h
@@ -186,6 +186,9 @@ class RocmComputeCapability {
 
   bool has_nanoo_fp8_support() const { return gfx9_mi300(); }
 
+  /// \brief Invalid gfx id for default gcn_arch_name_ value and testing
+  static constexpr absl::string_view kInvalidGfx = "gfx000";
+
  private:
   /// \brief Takes one or more arrays of string-like objects and tests if the
   /// result of `gfx_version()` matches to any string in any of the arrays.
@@ -206,7 +209,7 @@ class RocmComputeCapability {
     });
   }
 
-  std::string gcn_arch_name_ = "gfx000";  // default to invalid arch.
+  std::string gcn_arch_name_{kInvalidGfx};  // default to invalid arch.
 };
 
 using GpuComputeCapability =
--- a/xla/stream_executor/device_description.h
+++ b/xla/stream_executor/device_description.h
@@ -143,10 +143,19 @@ class RocmComputeCapability {
                                                     "gfx1151"};
   bool gfx11_apu() const { return IsThisGfxInAnyList(kGfx11Apu); }
 
+  static constexpr absl::string_view kGfx11Rx7900[] = {"gfx1100", "gfx1101",
+                                                       "gfx1102"};
+  bool gfx11_rx7900() const {
+    // TODO(AMD/TF): instead of this, other gfx11*() methods might be better
+    return IsThisGfxInAnyList(kGfx11Rx7900);
+  }
+
   bool gfx12() const { return absl::StartsWith(gfx_version(), "gfx12"); }
 
   static constexpr absl::string_view kGfx12Discrete[] = {"gfx1200", "gfx1201"};
   bool gfx12_discrete() const { return IsThisGfxInAnyList(kGfx12Discrete); }
+
+  bool gfx12_rx8900() const { return gfx12_discrete(); }
 
   bool has_nhwc_layout_support() const { return gfx9_mi100_or_later(); }
 
--- a/xla/stream_executor/device_description.h
+++ b/xla/stream_executor/device_description.h
@@ -20,8 +20,10 @@ limitations under the License.
 #ifndef XLA_STREAM_EXECUTOR_DEVICE_DESCRIPTION_H_
 #define XLA_STREAM_EXECUTOR_DEVICE_DESCRIPTION_H_
 
+#include <algorithm>
 #include <cassert>
 #include <cstdint>
+#include <cstring>
 #include <string>
 #include <type_traits>
 #include <utility>
@@ -30,6 +32,7 @@ limitations under the License.
 
 #include "absl/algorithm/container.h"
 #include "absl/status/statusor.h"
+#include "absl/strings/match.h"
 #include "absl/strings/str_join.h"
 #include "absl/strings/str_split.h"
 #include "absl/strings/string_view.h"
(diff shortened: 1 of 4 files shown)


Arech8:
@xla-rotation can you please review the PR ?

Arech8:
@beckerhe JFYI: CI internal failure was fixed and it's fully green now.

xla/stream_executor/device_description.h
```diff
@@ -54,13 +57,45 @@ class RocmComputeCapability {
...
+  RocmComputeCapabilityProto ToProto() const {
+    RocmComputeCapabilityProto proto;
+    proto.set_gcn_arch_name(gcn_arch_name_);
+    return proto;
+  }
+
+  bool operator==(const RocmComputeCapability &other) const {
+    return gcn_arch_name_ == other.gcn_arch_name_;
+  }
+
```
beckerhe: Can you also define operator!= for symmetry reasons?

xla/stream_executor/device_description.h
```diff
@@ -54,13 +57,45 @@ class RocmComputeCapability {
...
   std::string gfx_version() const {
-    std::vector<std::string> tokens = absl::StrSplit(gcn_arch_name_, ':');
-    return tokens[0];
+    //  std::strchr() is faster for the case than std::string::find()
+    const char *const p_colon = std::strchr(gcn_arch_name_.c_str(), ':');
+    if (nullptr == p_colon) {
+      return gcn_arch_name_;  // likely it's the default invalid value
+    }
```
beckerhe, line 82: How is it relevant that this utility method is fast?

If this needs to be fast then we should do the split in the constructor and keep both parts around separately.

If you just want to make a bit faster we can avoid some temporary allocations:

```
std::pair<absl::string_view, absl::string_view> tokens = absl::StrSplit(gcn_arch_name_, ':');
return std::string(tokens.first);
```

But I'm a bit opposed to using raw C pointers and C library functions which makes the code harder to read.

xla/stream_executor/device_description.h
```diff
@@ -69,98 +104,104 @@ class RocmComputeCapability {
...
-  bool operator==(const RocmComputeCapability &other) const {
-    return gcn_arch_name_ == other.gcn_arch_name_;
+  /// \brief Template-less implementation of IsThisGfxInAnyList().
+  /// \warning Don't use directly!
+  bool implIsThisGfxInAnyList(const absl::string_view *beg,
+                              const absl::string_view *end,
+                              const std::string &gfx) const {
+    return std::any_of(beg, end, [&gfx = gfx](const absl::string_view &s) {
+      return gfx == s;
+    });
```
beckerhe, line 218: Can you makes this a lambda expression in the actual function?

xla/stream_executor/device_description.h
```diff
@@ -69,98 +104,104 @@ class RocmComputeCapability {
...
   }
 
-  bool operator==(const RocmComputeCapability &other) const {
-    return gcn_arch_name_ == other.gcn_arch_name_;
+  /// \brief Template-less implementation of IsThisGfxInAnyList().
+  /// \warning Don't use directly!
+  bool implIsThisGfxInAnyList(const absl::string_view *beg,
+                              const absl::string_view *end,
+                              const std::string &gfx) const {
+    return std::any_of(beg, end, [&gfx = gfx](const absl::string_view &s) {
```
beckerhe, line 216: Nit: I think it would be more idiomatic to just use `&` instead of `&gfx = gfx`.

beckerhe (commented):
Some minor comments, but in general it looks pretty straight forward to me.

beckerhe:
I just realized that there aren't any tests for any of the getters in RocmComputeCapability. Can you add those?

xla/stream_executor/device_description.h
Arech8: Ok, I will, though it doesn't look to me that "symmetry reasons" is a valid reason to add a code for which there was no demand during all the lifetime of the project. This isn't a new class with an uncertain scope of use. It's a core component that was here from the inception of AMD support.

xla/stream_executor/device_description.h
Arech8: Yes, it's important for the method to be fast, as even during compilation of a very trivial code, XLA may call it hundreds of times, unfortunately.

A much better solution would be to get rid of string comparisons entirely, but that requires a lot of additional work and is totally outside of the scope of the PR fixing BF16 support.

I also considered refactoring the gfx part of the `gcn_arch_name_` into a separate string data member, but decided to go with fixing the method only. XLA puts the class into a variant, and frequently passes it by value instead of just passing the variant as a reference to a const, so it's totally not obvious if such refactoring could improve, not deteriorate things. This also explains why having an index data member pointing to `:` and constructing a string from it and `gcn_arch_name_` isn't any better. Costs of doing `std::strchr()` might be offset by costs of frequent class object copying, plus encapsulation of knowing gfx part will be smeared across the whole class, instead of being restricted to a single method only.

With respect to pointers: harder or easier is a matter of knowledge and training, isn't?  Let's resolve to objectively measurable things instead, shall we? If you suppose that the method and the whole class needs to be read frequently, then why for all the lifetime of the project the method was so utterly bad and inefficient? Why there were very obvious bugs with `std::string::find()` for `gfx11()` and relevant implementations standing high in a plain sight, that weren't caught by any of multiple reviewers or other "readers"? Is it because everyone is incompetent, or just because external properties of a method is the main thing that matters and unless necessary, people tend to not look into an implementation at all?
Raw pointers are fully encapsulated here and never leave the method; if someone is to look inside - their use is safe and transparent. And there's literally no redundant work done here. Can you supply some data to justify the value of making a worse implementation?

xla/stream_executor/device_description.h
Arech8: In-place lambda here is a prvalue that will get inlined into the code and will never be called as a function on the most compilers, making the code very efficient. Refactoring it to a separate function might significantly decrease chances of that. Can you justify the value of doing that?

xla/stream_executor/device_description.h
Arech8: Implicit lambda captures, such as `&` or `=` are not required to capture only the variables used by the lambda. They could capture all variables of the scope, which will make the processor do unnecessary job. What is the value of that?

Second, but not less important: many coding style guides focusing on a code safety (such as AUTOSAR C++14 Coding Guidelines, or MISRA C++:2023) directly forbid implicit captures for a very good reason: you must know what variables you're going to use in the lambda body and that's how you document that, preventing accidental bugs in the lambda.

> Rule A5-1-2 (required, implementation, automated)
> Variables shall not be implicitly captured in a lambda expression.
> https://www.autosar.org/fileadmin/standards/R22-11/AP/AUTOSAR_RS_CPP14Guidelines.pdf

xla/stream_executor/device_description.h
Arech8: added

Arech8:
@beckerhe I have added `operator!=` and tests. I've replied to the other comments explaining why I don't think that suggested changes are beneficial for the case. If you still want them, please answer to the concerns stated.
//Aleksei

Arech8:
@beckerhe can you please finish with the review?

Arech8:
@beckerhe if you don't have capacity for the review, can you please assign someone else to do that?

Arech8:
Dear @xla-rotation , can you assign someone to review the PR? It's being stuck without a move for 12th day after a few requests to clarify previous reviewer's personal preferences.

i-chaochen:
Hi @xla-rotation please review this PR?

beckerhe (approved):

PR #29852, opened 2025-08-06
Add a new HloModule clone API that can return clone context

shawnwang18:
This PR adds a new API `CloneWithContext` to `HloModule`, so user can get the mapped instruction/computation inside cloned HloModule. 

This PR is mostly a refactor of existing implementation, so current tests should cover all its functionality.

### DIFF
--- a/xla/hlo/ir/hlo_module.cc
+++ b/xla/hlo/ir/hlo_module.cc
@@ -1243,19 +1243,16 @@ void CopyUniqueIds(const HloModule& source, HloModule* clone,
 
 }  // namespace
 
-std::unique_ptr<HloModule> HloModule::Clone(
-    const std::string& suffix,
-    std::optional<const HloModuleConfig> config_in) const {
-  auto module = CreateModule(suffix, config_in, *this);
-
-  HloCloneContext context(module.get(), suffix);
+void HloModule::Clone(const std::string& suffix, HloCloneContext* context,
+                      std::optional<const HloModuleConfig> config) const {
+  auto module = context->module();
   if (entry_computation_) {
-    auto cloned_computation = entry_computation_->Clone(suffix, &context);
+    auto cloned_computation = entry_computation_->Clone(suffix, context);
     module->AddEntryComputation(std::move(cloned_computation));
   }
 
   // Preserve original instruction and computation ids.
-  CopyUniqueIds(*this, module.get(), context);
+  CopyUniqueIds(*this, module, *context);
   module->next_unique_id_ = next_unique_id_;
 
   module->input_output_alias_config() = input_output_alias_config();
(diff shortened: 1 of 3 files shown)


xla/hlo/ir/hlo_module.h
```diff
@@ -711,6 +718,11 @@ class HloModule {
       std::unique_ptr<HloComputation> computation, bool is_entry,
       bool uniquify_identifiers, bool preserve_entry_layouts);
 
+  // Performs a deep clone of current module to context->module, and populate
```
akuegel, line 721: I think this can be a private method. I guess the new method CloneWithContext serves the need to get access to the mappings?

xla/hlo/ir/hlo_module.cc
```diff
@@ -1243,30 +1243,30 @@ void CopyUniqueIds(const HloModule& source, HloModule* clone,
...
 
-std::unique_ptr<HloModule> HloModule::Clone(
-    const std::string& suffix,
-    std::optional<const HloModuleConfig> config_in) const {
-  auto module = CreateModule(suffix, config_in, *this);
-
-  HloCloneContext context(module.get(), suffix);
+void HloModule::Clone(const std::string& suffix, HloCloneContext* context,
+                      std::optional<const HloModuleConfig> config) const {
+  CHECK(context != nullptr) << "HloCloneContext pointer must not be null.";
```
akuegel: If we make it a private method, I am not sure whether we really need the CHECKs here.

akuegel (commented):
Could you please add a test that calls the new CloneWithContext method?

xla/hlo/ir/hlo_module.cc
shawnwang18: Yes, make sense. I was making the fix to make copilot happy

xla/hlo/ir/hlo_module.h
shawnwang18: Yes, it is a private method now. The public method will only have the other two:

```
std::unique_ptr<HloModule> Clone(
      const std::string& suffix = "clone",
      std::optional<const HloModuleConfig> config = std::nullopt) const;

 
  std::pair<std::unique_ptr<HloModule>, std::unique_ptr<HloCloneContext>>
  CloneWithContext(
      const std::string& suffix,
      std::optional<const HloModuleConfig> config = std::nullopt) const;
```

You mean this is OK?

xla/hlo/ir/hlo_module.cc
akuegel: At least internally we try to avoid CHECKs if they are not really needed. So I would ignore the suggestion of Copilot here.

xla/hlo/ir/hlo_module.h
akuegel: Ah, right, it is already a private method. Yes, this is OK :)

shawnwang18:
> Could you please add a test that calls the new CloneWithContext method?

Thanks, just add a unittest for `CloneWithcontext`

akuegel (approved):

PR #28740, opened 2025-07-10
[XLA:GPU] Lowering dynamic update slice thunk into command buffer if it depends on loop iteration. 

shawnwang18:
This is PR tries to lower DynamicSliceThunk into command buffer, even if it depends on the loop iteration. 

The command buffer implementation will also use the same approach (HloEvaluator to get new allocation during runtime) as DynamicSLiceThunk to get the sliced allocations, and for each iteration, CommandBuffer will use HloEvalutor to get the new addresses, and doing graph update with the new address. 

The major changes to custom.cc file is to resolve the issue that when a module has been parsed by command buffer scheduler, it rewrites the module into nested calls, which breaks the while loop analysis pattern, and module extraction pattern, so the fix is trying to introduced a cloned inline module, and perform the loop analysis and module extraction from the inlined module.

PR #29884, opened 2025-08-06
[XLA:GPU] Add GetInlinedModule helper function in call_inliner

shawnwang18:
This PR add a function `GetInlinedModule`, which is  Given a module, this function first clone the module, then inline the module, and return the inlined module, clone context and inlined map in InlinedModule struct.

This class is required because we found that while_loop_analysis pass can not work on the module that has been parsed by command buffer rewriter, so to do loop analysis, we have to first inline the command buffer call. 

This PR is depended by :  https://github.com/openxla/xla/pull/28740

### DIFF
--- a/xla/service/call_inliner.cc
+++ b/xla/service/call_inliner.cc
@@ -30,7 +30,9 @@ limitations under the License.
 #include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
 #include "absl/types/span.h"
+#include "absl/algorithm/container.h"
 #include "xla/hlo/ir/dfs_hlo_visitor_with_default.h"
+#include "xla/hlo/ir/hlo_module.h"
 #include "xla/hlo/ir/hlo_instruction.h"
 #include "xla/hlo/ir/hlo_instructions.h"
 #include "xla/hlo/ir/hlo_opcode.h"
@@ -127,8 +129,8 @@ class SubcomputationInsertionVisitor : public DfsHloVisitorWithDefault {
             call_instruction_name =
                 call_original_value->leaf_begin()->second->instruction_name;
           }
-          absl::StrAppend(&original_array->instruction_name, "/",
-                          call_instruction_name);
+          original_array->instruction_name = absl::StrCat(
+              call_instruction_name, "/", original_array->instruction_name);
         }
       }
     }
@@ -344,8 +346,10 @@ bool CallInliner::ShouldInline(const CallGraph& call_graph,
 
 absl::StatusOr<bool> CallInliner::InlineAndLegalize(
     const CallGraph& call_graph, HloComputation* computation,
-    absl::Span<HloInstruction* const> instruction_sequence) const {
+    absl::Span<HloInstruction* const> instruction_sequence,
+    std::optional<InlinedInstructionMap*> inline_map) {
   HloModule* module = computation->parent();
+  VLOG(0) << "module has schedule: " << module->has_schedule();
   bool did_node_mutate = false;
   std::vector<HloInstruction*> inlined_instructions;
   for (HloInstruction* instruction : instruction_sequence) {
@@ -357,23 +361,29 @@ absl::StatusOr<bool> CallInliner::InlineAndLegalize(
       // The caller instruction will get removed after inlining. Record the
       // callee computation beforehand, so we can find its schedule.
       HloComputation* callee = instruction->to_apply();
-      TF_ASSIGN_OR_RETURN(CallInliner::InlinedInstructionMap inline_map,
-                          Inline(instruction));
+      TF_ASSIGN_OR_RETURN(
+          CallInliner::InlinedInstructionMap inline_map_cur_call,
+          Inline(instruction));
       if (module->has_schedule()) {
         for (HloInstruction* inlined_instruction :
              module->schedule().sequence(callee).instructions()) {
           // Parameters were already added to sequence as operands to the
           // call.
           if (inlined_instruction->opcode() != HloOpcode::kParameter) {
-            inlined_instructions.push_back(inline_map[inlined_instruction]);
+            inlined_instructions.push_back(
+                inline_map_cur_call[inlined_instruction]);
           }
         }
       }
       if (update_domain_) {
         HloDomainIsolator isolator([]() { return ShardingDomainCreator{}; });
-        for (const auto& [call_inst, inlined_inst] : inline_map) {
+        for (const auto& [call_inst, inlined_inst] : inline_map_cur_call) {
           TF_RETURN_IF_ERROR(isolator.UpdateDomains(inlined_inst).status());
         }
+      }
+      if (inline_map.has_value()) {
+        inline_map.value()->insert(inline_map_cur_call.begin(),
+                                   inline_map_cur_call.end());
       }
       did_node_mutate = true;
     } else if (module->has_schedule()) {
@@ -395,8 +405,8 @@ absl::StatusOr<bool> CallInliner::InlineAndLegalize(
   return did_node_mutate;
 }
 
-absl::StatusOr<bool> CallInliner::Run(
-    HloModule* module,
+absl::StatusOr<bool> CallInliner::RunWithInlineMap(
+    HloModule* module, std::optional<InlinedInstructionMap*> inline_map,
     const absl::flat_hash_set<absl::string_view>& execution_threads) {
   std::unique_ptr<CallGraph> call_graph = CallGraph::Build(module);
   // Because call graph nodes are visited in post-order (callees before callers)
@@ -414,12 +424,12 @@ absl::StatusOr<bool> CallInliner::Run(
               HloInstructionSequence& sequence =
                   module->schedule().GetOrCreateSequence(node.computation());
               return InlineAndLegalize(*call_graph, node.computation(),
-                                       sequence.instructions());
+                                       sequence.instructions(), inline_map);
             }
 
             return InlineAndLegalize(
                 *call_graph, node.computation(),
-                node.computation()->MakeInstructionPostOrder());
+                node.computation()->MakeInstructionPostOrder(), inline_map);
           }));
   if (did_mutate) {
     // Run DCE to remove called computations which are now becoming unused.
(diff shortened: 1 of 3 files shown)


pifon2a (commented):
How much did the compile time regress?

shawnwang18:
> How much did the compile time regress?

I did not see obvious change of compile time for this PR.  This is a preparation PR, and currently not used in the main pipeline.  I think for future use of it, we can only use it when it is necessary ( there are command buffer scheduler pass has been applied to the module)

xla/service/call_inliner.h
```diff
@@ -98,6 +104,27 @@ class CallInliner : public HloModulePass {
   std::optional<
       std::function<bool(const CallGraph& call_graph, HloInstruction*)>>
       should_inline_;
+  InlinedInstructionMap inline_map_;
```
allanrenucci: Is this unused?

xla/service/call_inliner.cc
```diff
@@ -355,24 +356,31 @@ absl::StatusOr<bool> CallInliner::InlineAndLegalize(
...
         HloDomainIsolator isolator([]() { return ShardingDomainCreator{}; });
-        for (const auto& [call_inst, inlined_inst] : inline_map) {
+        for (const auto& [call_inst, inlined_inst] : inline_map_cur_call) {
           TF_RETURN_IF_ERROR(isolator.UpdateDomains(inlined_inst).status());
         }
       }
+      if (inline_map.has_value()) {
+        for (const auto& [call_inst, inlined_inst] : inline_map_cur_call) {
+          (*inline_map.value())[call_inst] = inlined_inst;
+        }
```
allanrenucci: Maybe:
```c++
inline_map->insert(inline_map_cur_call.begin(), inline_map_cur_call.end());
```

xla/service/call_inliner.h
```diff
@@ -98,6 +104,27 @@ class CallInliner : public HloModulePass {
...
+};
+
+// ScopedClonedModuleCallInliner is used for the scenario where we can not
+// modify the original module, instead this RAII class clones the target module
+// and inlines it during initialization. 'inlined_module' returns the inlined
+// module, 'get_mapped_instruction' returns the inlined instruction for a
+// given original instruction.
+class ScopedClonedModuleCallInliner {
+ public:
+  ScopedClonedModuleCallInliner(const HloModule* module);
```
allanrenucci:
```suggestion
  explicit ScopedClonedModuleCallInliner(const HloModule* module);
```

xla/service/call_inliner.h
```diff
@@ -98,6 +104,27 @@ class CallInliner : public HloModulePass {
...
+  ~ScopedClonedModuleCallInliner() = default;
```
allanrenucci: I believe you can remove the explicit destructor.

xla/service/call_inliner.cc
```diff
@@ -433,4 +441,51 @@ absl::StatusOr<bool> CallInliner::Run(
...
+      }
+    }
+  }
+
+  CallInliner inliner;
+  auto status =
+      inliner.RunWithInlineMap(inlined_module_.get(), &inline_map_, {});
+  if (!status.ok()) {
+    LOG(ERROR) << "Failed to inline module: " << status.status();
+  }
```
allanrenucci: Instead of ignoring the error, how about we make the constructor private and have a factory function that returns a `absl::StatusOr<ScopedClonedModuleCallInliner>`?

xla/service/call_inliner.h
```diff
@@ -98,6 +104,27 @@ class CallInliner : public HloModulePass {
...
       should_inline_;
+  InlinedInstructionMap inline_map_;
+};
+
+// ScopedClonedModuleCallInliner is used for the scenario where we can not
+// modify the original module, instead this RAII class clones the target module
+// and inlines it during initialization. 'inlined_module' returns the inlined
+// module, 'get_mapped_instruction' returns the inlined instruction for a
+// given original instruction.
+class ScopedClonedModuleCallInliner {
```
allanrenucci: How about we just call it `InlinedClonedModule`? I am not sure what is "scoped". The lifetime of the cloned module is tied to this object but that is true for most fields in a C++ objects.

allanrenucci (changes requested):

akuegel:
I have taken a look at the PR that depends on this to see why we need this class, and now I am not so sure this is a good idea.
As far as I can tell, we want to evaluate *some* part of the module. Would it be possible to first extract the part of the module that is relevant, and then run CallInliner on it? Because right now, you will always clone the whole module which can be quite expensive, especially if you do it not just once but potentially many times (I guess at least once for each dynamic slice fusion)?

shawnwang18:
> I have taken a look at the PR that depends on this to see why we need this class, and now I am not so sure this is a good idea. As far as I can tell, we want to evaluate _some_ part of the module. Would it be possible to first extract the part of the module that is relevant, and then run CallInliner on it? Because right now, you will always clone the whole module which can be quite expensive, especially if you do it not just once but potentially many times (I guess at least once for each dynamic slice fusion)?

@akuegel @allanrenucci Thanks for the review comments, I agree with @akuegel that we can further optimize the cloned module call inliner. Will think how to do an further optimized call-inliner.

xla/service/call_inliner.h
shawnwang18: Yes, fixed. thanks.

xla/service/call_inliner.cc
shawnwang18: Fixed

shawnwang18:
> I have taken a look at the PR that depends on this to see why we need this class, and now I am not so sure this is a good idea. As far as I can tell, we want to evaluate _some_ part of the module. Would it be possible to first extract the part of the module that is relevant, and then run CallInliner on it? Because right now, you will always clone the whole module which can be quite expensive, especially if you do it not just once but potentially many times (I guess at least once for each dynamic slice fusion)?

@akuegel I did a large refactoring of the depending PR https://github.com/openxla/xla/pull/28740, the cost should have been greatly reduced. 

I think there are lots of redundant API calls when lowering dynamic update slice instruction, to reduce the costs, I create a separate PR for it: https://github.com/openxla/xla/pull/30397

could you review again?

xla/service/call_inliner.cc
```diff
@@ -344,8 +346,10 @@ bool CallInliner::ShouldInline(const CallGraph& call_graph,
 
 absl::StatusOr<bool> CallInliner::InlineAndLegalize(
     const CallGraph& call_graph, HloComputation* computation,
-    absl::Span<HloInstruction* const> instruction_sequence) const {
+    absl::Span<HloInstruction* const> instruction_sequence,
+    std::optional<InlinedInstructionMap*> inline_map) {
   HloModule* module = computation->parent();
+  VLOG(0) << "module has schedule: " << module->has_schedule();
```
akuegel: nit: maybe VLOG(2)
Or remove in case it is a leftover from debugging.

xla/service/call_inliner.h
```diff
@@ -95,6 +101,21 @@ class CallInliner : public HloModulePass {
       should_inline_;
 };
 
+// Returns true if the computation has instructioin that is inlinable.
```
akuegel: nit: instructioin -> instruction

xla/service/call_inliner.h
```diff
@@ -95,6 +101,21 @@ class CallInliner : public HloModulePass {
...
+bool IsInlineableComputation(HloComputation* computation);
+
+struct InlinedModule {
+  std::unique_ptr<HloModule> module;
+  std::unique_ptr<HloCloneContext> clone_context;
+  CallInliner::InlinedInstructionMap clone_inlined_map;
+  const HloInstruction* get_inlined_inst(const HloInstruction* inst);
+};
+
+// Given a module, this function first clone the module, then inline the module,
```
akuegel: nit: clone -> clones
inline -> inlines

xla/service/call_inliner.h
```diff
@@ -95,6 +101,21 @@ class CallInliner : public HloModulePass {
...
+// and return the inlined module, clone context and inlined map in InlinedModule
```
akuegel: return -> returns

akuegel:
> > I have taken a look at the PR that depends on this to see why we need this class, and now I am not so sure this is a good idea. As far as I can tell, we want to evaluate _some_ part of the module. Would it be possible to first extract the part of the module that is relevant, and then run CallInliner on it? Because right now, you will always clone the whole module which can be quite expensive, especially if you do it not just once but potentially many times (I guess at least once for each dynamic slice fusion)?
> 
> @akuegel I did a large refactoring of the depending PR #28740, the cost should have been greatly reduced.
> 
> I think there are lots of redundant API calls when lowering dynamic update slice instruction, to reduce the costs, I create a separate PR for it: #30397
> 
> could you review again?

Can you please quickly summarize how you addressed the problem with cost proportional to "number of dynamic slice fusions" times "number of instructions in the HloModule containing that fusion"? I still see the call to GetInlinedModule that is passed fusion.GetModule(), and that function clones the whole module that it gets passed. But maybe I am missing something, or looked at some outdated code?

shawnwang18:
> > > I have taken a look at the PR that depends on this to see why we need this class, and now I am not so sure this is a good idea. As far as I can tell, we want to evaluate _some_ part of the module. Would it be possible to first extract the part of the module that is relevant, and then run CallInliner on it? Because right now, you will always clone the whole module which can be quite expensive, especially if you do it not just once but potentially many times (I guess at least once for each dynamic slice fusion)?
> > 
> > 
> > @akuegel I did a large refactoring of the depending PR #28740, the cost should have been greatly reduced.
> > I think there are lots of redundant API calls when lowering dynamic update slice instruction, to reduce the costs, I create a separate PR for it: #30397
> > could you review again?
> 
> Can you please quickly summarize how you addressed the problem with cost proportional to "number of dynamic slice fusions" times "number of instructions in the HloModule containing that fusion"? I still see the call to GetInlinedModule that is passed fusion.GetModule(), and that function clones the whole module that it gets passed. But maybe I am missing something, or looked at some outdated code?

I think it should be proportional to number of DynamicSliceFusion instructions?  Because DynamicSliceFusion happens together with WHILE loop, so in LLM, basically it needs to process the process a single Transformer Layer. 

But maybe I can make static cloned module cache at here (https://github.com/openxla/xla/pull/28740/files#diff-c59b61734f1ef0976ce57bb9fa9f7047abda6e28117bfa10b09e676ed13653efR1409-R1412) , so for an HLO module, it just needs one clone, but this does not affect this PR. 

way like: 

```
 static absl::flat_hash_map<const HloModule*, InlinedModule>
      inlined_module_map;

  if (inlined_module_map.find(fusion.GetModule()) == inlined_module_map.end()) {
    TF_ASSIGN_OR_RETURN(auto inlined_module,
                        GetInlinedModule(fusion.GetModule()));
    inlined_module_map[fusion.GetModule()] = inlined_module;
  }
  auto inlined_module = inlined_module_map[fusion.GetModule()];
```

akuegel:
> > Can you please quickly summarize how you addressed the problem with cost proportional to "number of dynamic slice fusions" times "number of instructions in the HloModule containing that fusion"? I still see the call to GetInlinedModule that is passed fusion.GetModule(), and that function clones the whole module that it gets passed. But maybe I am missing something, or looked at some outdated code?
> 
> I think it should be proportional to number of DynamicSliceFusion instructions? Because DynamicSliceFusion happens together with WHILE loop, so in LLM, basically it needs to process the process a single Transformer Layer.
> 
> But maybe I can make static cloned module cache at here (https://github.com/openxla/xla/pull/28740/files#diff-c59b61734f1ef0976ce57bb9fa9f7047abda6e28117bfa10b09e676ed13653efR1409-R1412) , so for an HLO module, it just needs one clone, but this does not affect this PR.
> 
> way like:
> 
> ```
> static absl::flat_hash_map<const HloModule*, const HloModule*> cloned_inline_module_map 
> 
> if (cloned_inline_module_map.contains(fusion.GetModule()) {
>      cloned_inlined_module = cloned_inline_module_map[fusion.GetModule());
> } else {
>     cloned_inline_module_map[fusion.GetModule()) = GetInlinedModule(module)
> }
> ```

The cache would work if we can be sure that neither module nor cloned module got changed. For the cloned module you can enforce that by never giving access to a modifiable module, but for the module this seems difficult.
A possibly safer method could be to compute the InlinedModule in ir_emitter_unnested.cc in the EmitHloComputation() method and pass it around until the HloFusionInfo is constructed and passed to GetFusionEmitter, and from there to DynamicSliceFusion. With that, I think the performance concern would be addressed. EmitHloComputation() is called once per module (for the entry computation of the module).

akuegel (approved):
Now that we have discussed a way how to use this helper in a way that addresses the performance concern I had, I think this is ready to land.

PR #30072, opened 2025-08-11
[XLA:GPU]Improve Flag Handling while Linking for oneAPI

mraunak:
This PR addresses a linking failure caused by an overflow of command-line flags, resulting in an exit code 127 error during the linking stage. To resolve this, we introduced the following changes:

**Improved Handling of Whole-Archive Object Files**
Object files with .o or .lo extensions are now linked using the --whole-archive and --no-whole-archive flags. This forces the linker to include all symbols from these files, ensuring none are removed during linking. This change helps reduce the total number of linker flags while preserving necessary symbols, which in turn prevents command-line overflow issues.

For better debugging, we introduced support for the VERBOSE=1 environment variable. When set, it prints the full command line used to invoke the compiler, which helps with diagnosing cross-compilation issues and verifying correct toolchain usage.

### DIFF
--- a/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_sycl.tpl
+++ b/third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_sycl.tpl
@@ -164,18 +164,35 @@ def call_compiler(argv, link = False, sycl_compile = True):
         else:
             expanded_flags.append(s)
 
-    # Deduplicate flags
     seen = set()
-    deduped_flags = []
+    ordered_flags = []
+    
+    # Track the type of each unique flag and preserve order
     for s in expanded_flags:
         if s not in seen:
             seen.add(s)
-            if s.endswith(".o"):
+            flag_type = "whole" if s.endswith((".o", ".lo")) else "regular"
+            ordered_flags.append((flag_type, shlex.quote(s)))
+    
+    deduped_flags = []
+    in_whole_archive = False
+    
+    for flag_type, flag in ordered_flags:
+        if flag_type == "whole":
+            if not in_whole_archive:
                 deduped_flags.append("-Wl,--whole-archive")
-            deduped_flags.append(shlex.quote(s))
-            if s.endswith(".o"):
+                in_whole_archive = True
+            deduped_flags.append(flag)
+        else:
+            if in_whole_archive:
                 deduped_flags.append("-Wl,--no-whole-archive")
-
+                in_whole_archive = False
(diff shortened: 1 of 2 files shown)


third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_sycl.tpl
```diff
@@ -164,18 +164,35 @@ def call_compiler(argv, link = False, sycl_compile = True):
...
+    for flag_type, flag in ordered_flags:
+        if flag_type == "whole":
+            if not in_whole_archive:
                 deduped_flags.append("-Wl,--whole-archive")
-            deduped_flags.append(shlex.quote(s))
-            if s.endswith(".o"):
+                in_whole_archive = True
+            deduped_flags.append(flag)
+        else:
+            if in_whole_archive:
```
ybaturina, line 187: I'm a bit lost in the logic here.

The var `in_whole_archive` is initialized as `False`, and then the loop over the `ordered_flags` list is started. `in_whole_archive` is assigned `True` only inside the branch `if flag_type == "whole"`. When does `in_whole_archive` become `true` for the `else` branch?

mraunak: Hi @ybaturina,
The goal is to get 
[a.o, b.lo, c.a, d.a, e.o, f.lo] --->[-Wl,--whole-archive a.o b.lo -Wl,--no-whole-archive c.a d.a -Wl,--whole-archive e.o f.lo -Wl,--no-whole-archive]
To achieve this, we are using flag type - 'whole' for .o,.lo or 'regular' for .a or other extensions.
If flag_type is whole, we want to add -Wl,--whole-archive at the beginning and -Wl,--no-whole-archive at the end and if the flags are in sequence with flag_type = whole then they should be grouped, hence if the code enters the else part if the flag_type = regular and it will first add the no archive part at the end and then move on to handle regular flag type by exiting the 'for loop'

If the flag type is regular(for .a files) it should not enter the loop, it should only enter the loop for flag_type =whole for .o,.lo files

ybaturina (approved):

third_party/gpus/crosstool/clang/bin/crosstool_wrapper_driver_sycl.tpl
ybaturina: Thank you.

ddunl (approved):

PR #30179, opened 2025-08-13
Handle the API change of cuGraphAddNode in CUDA 13.

dimvar:
Starting in CUDA 13, cuGraphAddNode takes an extra argument named dependencyData.
https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__GRAPH.html#group__CUDA__GRAPH_1ge01208e62f72a53367a2af903bf17d23

I changed XLA to use cuGraphAddNode_v2, whose signature didn't change with CUDA 13. cuGraphAddNode_v2 was added in CUDA 12.3:
https://docs.nvidia.com/cuda/archive/12.9.1/cuda-driver-api/group__CUDA__GRAPH.html#group__CUDA__GRAPH_1gb26c1065578f8753c0dde6b877186b77
so this fix won't work for earlier CUDA versions.

If Google wants to support CUDA before 12.3, we would have to keep using cuGraphAddNode here, with some conditional compilation depending on the CUDA version. Let me know what is preferable.

### DIFF
--- a/xla/stream_executor/cuda/cuda_command_buffer.cc
+++ b/xla/stream_executor/cuda/cuda_command_buffer.cc
@@ -283,10 +283,10 @@ CudaCommandBuffer::CreateConditionalNode(
 
   std::vector<CUgraphNode> deps = ToCudaGraphHandles(dependencies);
   CUgraphNode node_handle = nullptr;
-  TF_RETURN_IF_ERROR(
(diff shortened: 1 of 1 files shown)


ezhulenev (approved):

dimvar:
This was merged in https://github.com/openxla/xla/commit/5a58b3230b6574b5ef922a8722f3e3602aaf3d8a

PR #30187, opened 2025-08-13
[XLA:GPU] Remove legacy gpu graph level flags

shawnwang18:
XLA flag `xla_gpu_graph_level` has been depreciated, and should use `xla_gpu_enable_command_buffer` flag instead.


ezhulenev (approved):

akuegel:
We still have internal users of the deprecated flag. We will temporarily revert this change until all users are migrated to the new flag. Just a FYI, you don't have to do anything, we will reland the change once it is possible.

akuegel:
The internal users of the deprecated flag have been migrated.
Relanded this change in https://github.com/openxla/xla/commit/add6972510a43f4b84c2554aef449e61deb29d62

PR #30239, opened 2025-08-13
[GPU] Tweak NVML library loading error message wording

terryysun:
Non-MNNVL clusters don't need some NVML libraries like `nvmlDeviceGetGpuFabricInfoV`, the current message wording includes `Error` thus causing some confusion for debugging. This PR removes the error keyword from the message and makes it a warning.

### DIFF
--- a/xla/service/gpu/model/gpu_collective_performance_model_test.cc
+++ b/xla/service/gpu/model/gpu_collective_performance_model_test.cc
@@ -13,12 +13,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-#include <gmock/gmock.h>
 #include <gtest/gtest.h>
-#include "absl/log/log_sink.h"
-#include "absl/log/scoped_mock_log.h"
 #include "xla/hlo/testlib/hlo_hardware_independent_test_base.h"
-#include "xla/service/gpu/model/gpu_collective_performance_model.h"
+#include "xla/service/gpu/backend_configs.pb.h"
 
 namespace xla {
 namespace gpu {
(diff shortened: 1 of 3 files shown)


xla/service/gpu/model/gpu_collective_performance_model_test.cc
allanrenucci, line 1: We don't require tests for logging changes. Let's drop this.

allanrenucci (changes requested):
Otherwise, LGTM

xla/service/gpu/model/gpu_collective_performance_model_test.cc
terryysun: Do we have a guideline on when we don't expect tests? just curious about the rationale, thanks!

allanrenucci: I don't know of any guideline available in OSS but internally we have this:

> We trust your judgement, but generally, code should have tests. Examples where tests may not be appropriate:
> * Pure refactoring changes that don't introduce behavior changes. If there are no tests for the code, write tests first so you can be confident it's behavior neutral!
> * Moving code from one directory to another
> * Cleaning unused dependencies
> * Deleting code! The bestest code!
> * Debug error messages / VLOG / LOG message changes
> * Comment updates
> * Config only changes (e.g. borg files)
> * BUILD file updates
> * Global renames
> * Adding benchmarks
> * Pure flag flips
> * Black box API usage which leaves output unchanged (e.g. moving to a new CUDA API where you'd have to mock the whole world)

terryysun: Thanks! dropped the test, could you take another look?

allanrenucci (approved):

xla/service/gpu/model/gpu_collective_performance_model_test.cc
jprabhas: @allanrenucci  this would be good to capture as a standardized guideline here https://github.com/openxla/xla/blob/main/docs/contributing.md

PR #30317, opened 2025-08-15
[XLA:GPU] Enable block_n=8 for triton dot fusion emitter

sergey-kozub:
This results in significant improvement for shapes with small non-contracting dim (N).

TritonDotFusionSearchSpace is updated, as it currently produces configs that fail at runtime:
{block_m=?, block_n=8, block_k=16, num_warps=8}

Related Triton change: https://github.com/triton-lang/triton/commit/19277de9a45dce943b246cd0875221a3a644c06a

### DIFF
--- a/xla/backends/gpu/codegen/triton/fusion_emitter_legacy_matmul.cc
+++ b/xla/backends/gpu/codegen/triton/fusion_emitter_legacy_matmul.cc
@@ -828,7 +828,7 @@ absl::Status ValidateMatMulConfig(const TritonGemmConfig& config,
   TF_RET_CHECK(config.split_k >= 1);
   TF_RET_CHECK(config.block_m >= 16);
   TF_RET_CHECK(config.block_k >= 16);
-  TF_RET_CHECK(config.block_n >= 16);
+  TF_RET_CHECK(config.block_n >= 8);
 
   const auto& dims = dot.dot_dimension_numbers();
   int num_batch_dims =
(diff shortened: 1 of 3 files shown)


akuegel:
Goran is probably the best reviewer for this change. There is a public holiday in Munich today, so you likely will have to wait until Monday.

appujee:
Not sure why clang-format error shows up here, maybe try rebasing on ToT

sergey-kozub:
> Not sure why clang-format error shows up here, maybe try rebasing on ToT

Rebased, but this doesn't help. clang-format triggers on all the files modified by a PR, and in this case the failure is unrelated (most likely caused by clang-format upgrade).

gflegar (approved):
LGTM. We should also add @nputikhin to check if all the internal tests pass. He was trying to enable N=8 before, and we got some crashes in Triton with that.

nputikhin:
Yes unfortunately there are two blocking items for enabling these sizes:
* A Triton bug that crashes on a common use case with block_n=8 and 16-bit types: https://github.com/triton-lang/triton/issues/7830
* A change to autotuner that prevents configs for block_n=8 from affecting split_k choice for other configs: https://github.com/openxla/xla/pull/30269

I'm working on a similar PR in https://github.com/openxla/xla/pull/29282 and slowly getting through various issues that pop up

nputikhin:
Now that Triton version in XLA got updated, we managed to merge https://github.com/openxla/xla/pull/29282 - it enables the new block size, adds tests to verify that Triton behaves correctly, and modifies dot search space to address perf regressions we've observed.

One thing missing from that PR compared to this one is the min tile size calculation in `AddContractingTiling` - there was some difference in microbenchmarks with it but it did not show up in end to end models.

PR #30397, opened 2025-08-19
[XLA:GPU] Refactor dynamic slice fusion lowering code to reduce the API calls

shawnwang18:
This PR should be able to remove many redundant calls to `GetLoopInductionVarTupleIdx`

### DIFF
--- a/xla/backends/gpu/codegen/custom.cc
+++ b/xla/backends/gpu/codegen/custom.cc
@@ -299,7 +299,8 @@ absl::Status CollectSliceInfo(
     std::vector<std::optional<uint64_t>>& offset_byte_sizes,
     std::vector<std::unique_ptr<HloModule>>& extracted_offset_modules,
     unsigned arg_idx, bool can_compute_indvar_on_host,
-    std::optional<const HloInstruction*> while_op) {
+    std::optional<const HloInstruction*> while_op,
+    std::optional<int64_t> indvar_idx) {
   if (!IsDynamicSliceOrDynamicUpdateSlice(slice_instrs[arg_idx])) {
     return absl::OkStatus();
   }
@@ -307,13 +308,6 @@ absl::Status CollectSliceInfo(
       Cast<HloDynamicIndexInstruction>(slice_instrs[arg_idx]);
   std::optional<HloInstruction*> async_caller =
       fusion_instr.parent()->GetUniqueCaller(HloOpcode::kAsyncStart);
-
-  std::optional<int64_t> indvar_idx = std::nullopt;
-  if (while_op != std::nullopt) {
-    CHECK(while_op.value() != nullptr)
-        << "GetWhileOp is not expected to return nullptr.";
-    indvar_idx = GetLoopInductionVarTupleIdx(*while_op);
-  }
 
   std::vector<DynamicSliceThunk::Offset> arg_offsets;
   for (auto idx_op : arg_slice_instr->index_operands()) {
@@ -543,10 +537,11 @@ absl::StatusOr<FusionEmissionResult> EmitGemm(
   std::optional<HloInstruction*> while_op =
       GetParentWhileOp(fusion, call_graph);
   std::unique_ptr<HloModule> init_module, update_module;
+  std::optional<int64_t> indvar_idx = std::nullopt;
   if (while_op != std::nullopt) {
     CHECK(while_op.value() != nullptr)
         << "GetWhileOp is not expected to return nullptr.";
-    std::optional<int64_t> indvar_idx = GetLoopInductionVarTupleIdx(*while_op);
+    indvar_idx = GetLoopInductionVarTupleIdx(*while_op);
     if (indvar_idx != std::nullopt) {
       init_module = ExtractWhileInitModule(*while_op, indvar_idx.value());
       update_module = ExtractWhileUpdateModule(*while_op, indvar_idx.value());
@@ -564,8 +559,8 @@ absl::StatusOr<FusionEmissionResult> EmitGemm(
   TF_RETURN_IF_ERROR(CollectSliceInfo(
       buffer_assignment, fusion, absl::Span<HloInstruction*>(slice_instrs),
       offset_buffer_indices, orig_shapes, sliced_shapes, offset_byte_sizes,
-      extracted_offset_modules, arg_idx++, can_compute_indvar_on_host,
-      while_op));
+      extracted_offset_modules, arg_idx++, can_compute_indvar_on_host, while_op,
+      indvar_idx));
 
   TF_ASSIGN_OR_RETURN(
       BufferAllocation::Slice rhs_slice,
@@ -575,8 +570,8 @@ absl::StatusOr<FusionEmissionResult> EmitGemm(
   TF_RETURN_IF_ERROR(CollectSliceInfo(
       buffer_assignment, fusion, absl::Span<HloInstruction*>(slice_instrs),
       offset_buffer_indices, orig_shapes, sliced_shapes, offset_byte_sizes,
-      extracted_offset_modules, arg_idx++, can_compute_indvar_on_host,
-      while_op));
+      extracted_offset_modules, arg_idx++, can_compute_indvar_on_host, while_op,
+      indvar_idx));
 
   BufferAllocation::Slice output;
   std::optional<BufferAllocation::Slice> workspace = std::nullopt;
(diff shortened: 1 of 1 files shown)


xla/backends/gpu/codegen/custom.cc
```diff
@@ -547,44 +537,50 @@ absl::StatusOr<FusionEmissionResult> EmitGemm(
...
   if (while_op != std::nullopt) {
     CHECK(while_op.value() != nullptr)
         << "GetWhileOp is not expected to return nullptr.";
-    init_module = ExtractWhileInitModule(*while_op);
-    update_module = ExtractWhileUpdateModule(*while_op);
+    indvar_idx = GetLoopInductionVarTupleIdx(*while_op);
+    if (indvar_idx != std::nullopt) {
+      init_module = ExtractWhileInitModule(*while_op, indvar_idx.value());
+      update_module = ExtractWhileUpdateModule(*while_op, indvar_idx.value());
+    }
```
shraiysh, line 554: nit: else error?

xla/backends/gpu/codegen/custom.cc
```diff
@@ -753,11 +750,15 @@ absl::StatusOr<FusionEmissionResult> EmitCustomCall(
...
```
shraiysh, line 767: nit: else error?

xla/backends/gpu/codegen/custom.cc
```diff
@@ -1055,11 +1057,18 @@ CollectSliceArgumentMetadataForCollectives(
...
         << "GetParentWhileOp is not expected to return nullptr.";
-    slice_data.init_module = ExtractWhileInitModule(*while_op);
-    slice_data.update_module = ExtractWhileUpdateModule(*while_op);
+    indvar_idx = GetLoopInductionVarTupleIdx(*while_op);
+    if (indvar_idx != std::nullopt) {
+      slice_data.init_module =
+          ExtractWhileInitModule(*while_op, indvar_idx.value());
+      slice_data.update_module =
+          ExtractWhileUpdateModule(*while_op, indvar_idx.value());
+    }
```
shraiysh, line 1077: nit: else error.

xla/backends/gpu/codegen/custom.cc
```diff
@@ -295,6 +278,11 @@ std::unique_ptr<HloModule> ExtractWhileInitModule(
   return init_module;
 }
 
+std::unique_ptr<HloModule> ExtractSliceModule(
```
shraiysh: why this indirection? can we simply rename the original function, or remove this and keep the name same?

shraiysh (approved):

xla/backends/gpu/codegen/custom.cc
shawnwang18: I understand while loop without a index variable is possible?  else is not an error.

xla/backends/gpu/codegen/custom.cc
shawnwang18: I was thinking to apply some codes in later PRs. but seems it is not necessary now, I will remove the indirection.

xla/backends/gpu/codegen/custom.cc
shawnwang18: Fixed

xla/backends/gpu/codegen/custom.cc
```diff
@@ -266,26 +257,18 @@ std::unique_ptr<HloModule> ExtractOffsetModule(
 // these conditions are ensured before fusing in
 // `dynamic_slice_fusion_rewriter.cc:IsValueFunctionOfLoopInductionVariable`.
 std::unique_ptr<HloModule> ExtractWhileUpdateModule(
-    const HloInstruction* while_op) {
```
bchetioui, line 269: Could you please update the docstring to explain what indvar_idx corresponds to?

xla/backends/gpu/codegen/custom.cc
```diff
@@ -266,26 +257,18 @@ std::unique_ptr<HloModule> ExtractOffsetModule(
...
-  return ExtractOffsetModule(update, while_op);
+      while_op->while_body()->root_instruction()->operand(indvar_idx);
+  return ExtractOffsetModule(update, indvar_idx);
 }
 
 // Extracts the while induction variable initialization module. This must have
 // no parameters and this condition was ensured before fusing in
 // `dynamic_slice_fusion_rewriter.cc:IsValueFunctionOfLoopInductionVariable`.
 std::unique_ptr<HloModule> ExtractWhileInitModule(
-    const HloInstruction* while_op) {
```
bchetioui, line 283: Could you please update the docstring to explain what indvar_idx corresponds to?

xla/backends/gpu/codegen/custom.cc
```diff
@@ -210,16 +210,7 @@ std::optional<HloInstruction*> GetParentWhileOp(const HloInstruction& op,
 // ensured before fusing in
 // `dynamic_slice_fusion_rewriter.cc:IsValueFunctionOfLoopInductionVariable`.
 std::unique_ptr<HloModule> ExtractOffsetModule(
-    const HloInstruction* offset_value,
```
bchetioui, line 213: Could you please update the docstring to explain what indvar_idx corresponds to?

xla/backends/gpu/codegen/custom.cc
```diff
@@ -311,7 +294,8 @@ absl::Status CollectSliceInfo(
     std::vector<std::optional<uint64_t>>& offset_byte_sizes,
     std::vector<std::unique_ptr<HloModule>>& extracted_offset_modules,
     unsigned arg_idx, bool can_compute_indvar_on_host,
-    std::optional<const HloInstruction*> while_op) {
+    std::optional<const HloInstruction*> while_op,
+    std::optional<int64_t> indvar_idx) {
```
bchetioui, line 309: Could you please add a docstring for this function, which explains the various parameters we're taking in, including what indvar_idx corresponds to?

At this point, it also seems like we might want to have a `struct` to carry these things around, if they're bound to either be both present of both optional.

bchetioui (changes requested):

xla/backends/gpu/codegen/custom.cc
shawnwang18: Thanks, I add a few comments explaining the parameter. 

> At this point, it also seems like we might want to have a struct to carry these things around, if they're bound to either be both present of both optional.

No, they may not be both present I understand, we can have a while instruction, but no index variable.

xla/backends/gpu/codegen/custom.cc
shawnwang18: Fixed

xla/backends/gpu/codegen/custom.cc
bchetioui: In which case is this variable set, and in which case isn't it? (Please also add a comment explaining that.)

xla/backends/gpu/codegen/custom.cc
```diff
@@ -547,44 +540,50 @@ absl::StatusOr<FusionEmissionResult> EmitGemm(
   std::optional<HloInstruction*> while_op =
       GetParentWhileOp(fusion, call_graph);
   std::unique_ptr<HloModule> init_module, update_module;
+  std::optional<int64_t> indvar_idx = std::nullopt;
```
bchetioui: nit: you don't need to do `=  std::nullopt`, this is equivalent to default initialization.

xla/backends/gpu/codegen/custom.cc
```diff
@@ -547,44 +540,50 @@ absl::StatusOr<FusionEmissionResult> EmitGemm(
...
   }
   bool can_compute_indvar_on_host =
       (init_module != nullptr && update_module != nullptr);
 
   unsigned arg_idx = 0;
-  TF_ASSIGN_OR_RETURN(BufferAllocation::Slice lhs_slice,
-                      GetOperandSlice(buffer_assignment, adaptor, fusion,
-                                      *custom_call.operand(arg_idx),
-                                      slice_instrs, /*shape_idx=*/{}, arg_idx));
+  TF_ASSIGN_OR_RETURN(
```
bchetioui, line 560: Unless I am missing something, I think this didn't change (?). In the future, please don't change the formatting of lines you didn't touch. This makes it harder to understand what actually changed (and pollutes the history).

bchetioui (changes requested):

xla/backends/gpu/codegen/custom.cc
shawnwang18: Yes, I think I run the whole file clang formatting, maybe later I will stick to lines format.

xla/backends/gpu/codegen/custom.cc
shawnwang18: I added some comments about it. I think: 
//  when there is a wrapping while loop, we may not have an index
// variable for the loop, in case when the loop count is not naively controlled
// by a single loop index, so we will have the while op but no indvar_idx.

shawnwang18: @bchetioui could you take another look at this PR?

bchetioui: Makes sense, thanks.

bchetioui (approved):

PR #30514, opened 2025-08-21
Rename sm_101 to sm_110. Naming change for Thor; takes effect on CUDA 13 onward.

dimvar:
sm_101 is the SM number for Thor prior to CUDA 13. It will be effectively unused. XLA can support Thor only on CUDA 13 onward.

### DIFF
--- a/xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
+++ b/xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
@@ -244,7 +244,7 @@ std::string GetSmName(se::CudaComputeCapability compute_capability) {
   // If the current compute capability isn't known, fallback to the
   // most recent version before it.
   constexpr stream_executor::CudaComputeCapability kSupportedVersions[] = {
-      {12, 1}, {12, 0}, {10, 3}, {10, 1}, {10, 0}, {9, 0}, {8, 9}, {8, 7},
+      {12, 1}, {12, 0}, {11, 0}, {10, 3}, {10, 0}, {9, 0}, {8, 9}, {8, 7},
       {8, 6},  {8, 0},  {7, 5},  {7, 2},  {7, 0},  {6, 2}, {6, 1}, {6, 0},
       {5, 3},  {5, 2},  {5, 0},  {3, 7},  {3, 5},  {3, 2}, {3, 0}};
   auto target_compute_capability = kSupportedVersions[0];
--- a/xla/service/gpu/llvm_gpu_backend/nvptx_backend_test.cc
+++ b/xla/service/gpu/llvm_gpu_backend/nvptx_backend_test.cc
@@ -43,11 +43,11 @@ TEST(UtilsTest, TestGetSmName) {
                 10, 0, FeatureExtension::kForwardCompatibleFeatures}),
             "sm_100f");
   ASSERT_EQ(nvptx::GetSmName(se::CudaComputeCapability{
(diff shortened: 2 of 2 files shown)


xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
```diff
@@ -244,7 +244,7 @@ std::string GetSmName(se::CudaComputeCapability compute_capability) {
   // If the current compute capability isn't known, fallback to the
   // most recent version before it.
   constexpr stream_executor::CudaComputeCapability kSupportedVersions[] = {
-      {12, 1}, {12, 0}, {10, 3}, {10, 1}, {10, 0}, {9, 0}, {8, 9}, {8, 7},
+      {12, 1}, {12, 0}, {11, 0}, {10, 3}, {10, 0}, {9, 0}, {8, 9}, {8, 7},
```
Artem-B, line 247: If I build XLA with cuda-12.9 and try targeting sm_101, which is supported by that version, what are my options after this patch? 

NVIDIA should've made sm_110 and alias to sm_101 (or vice versa), to avoid breaking existing users, but they didn't. So now it's up to all CUDA users to conditionally support both SM variants depending on the CUDA version they are building with, and that will have to be around for as long as XLA supports building with cuda-12.9.

dimvar: AFAIK the only viable support for Thor is with CUDA 13+ (that is, support is missing for Thor in several libraries prior to CTK 13). XLA doesn't support Thor yet (we had added the SM number prematurely but never really tested), so we won't lose any support by this rename.

As for the rename itself, I'm not quite sure about the motivation for it. I think the driver continues to support both SM numbers. The renaming only happens for CTK.

Google doesn't use Thor right? So this shouldn't be an issue for Google.

beckerhe (approved):

xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
johnnynunez: hello @Artem-B @dimvar , so please, remove it 10.1. cuda 13 is the baseline for Thor: https://forums.developer.nvidia.com/t/jetpack-7-0-jetson-linux-38-2-for-nvidia-jetson-thor-is-now-live/343128/3

With the legacy driver (nvgpu) used for CUDA 12.9, Thor was operating with SM 10.1.
This changes to SM 11.0 when the newer driver model (OpenRM), which is intended for CUDA 13.0, is introduced.
Thor 10.1 --> 11.0
Spark 12.1

johnnynunez:
could you merge it?
https://github.com/openxla/xla/pull/30723

xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
johnnynunez: https://github.com/jax-ml/jax/issues/31399#issuecomment-3236710109

hudson-ayers: > AFAIK the only viable support for Thor is with CUDA 13+ (that is, support is missing for Thor in several libraries prior to CTK 13).

This may be true for Jetson Thor, but the Drive AGX Thor has complete CUDA support for CUDA 12.8. The Drive Thor series also is behind the Jetson series in receiving updates for CUDA 13. So this is a bummer because it means XLA is not usable on the Drive Thor until CUDA 13 support is available at some unknown point in the future.

PR #30624, opened 2025-08-26
[XLA:GPU] Refactoring the command dependency implementation through token resources

shawnwang18:
📝 Summary of Changes
In command buffer, there are two places that relies on resource dependency to specify command order. 
1.  When command buffer preserves the control dependency that is original existed across HLO operators.
2.  When specifying the dependency for cuda-graph topology inferred from LHS scheduling. 

This PR introduces a unified way to specify the token dependency across commands: 
1.  Each CommandBufferCmd has a local variable of type `std::shared_ptr<Resource> token_`, 
2. On instantiating CommandBufferCmd, auto create a Write resource use for local token resources;
3. If other commands has execution dependency on current command, then other command needs to include a READ resource use of the token resource of current command in other commands' resoure_use vector. 

In this approach, LHS and control dependency has the unified way to describe the dependency across commands. 

🎯 Justification
More clean and easy to understand design. 

🚀 Kind of Contribution
♻️ Cleanup 

🧪 Unit Tests:
This PR reimplements existing feature, so it is already covered by current unit tests.
xla/backends/gpu/runtime/command_buffer_thunk_test.cc

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/backends/gpu/runtime/command_buffer_cmd.cc
@@ -275,7 +275,6 @@ class CommandOperation : public ExecutionGraph::Operation {
   CommandBufferCmd::BufferUseVector buffers_;
   const CommandBufferCmd* cmd_;
   ResourceUseVector resources_;
-  const CommandBufferCmd* cmd_;
 
   // The token resource is used to specify dependency other than buffer data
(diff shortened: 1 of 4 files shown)


shawnwang18:
@ezhulenev Could you help review this PR?

ezhulenev (approved):

PR #30664, opened 2025-08-26
fix collective permute clone

Cjkkkk:
📝 Summary of Changes
Collective permute instruction `CloneWithNewOperands` only clones first operand while Collective permute is variadic op. Fixed this by cloning whole operands list.

🎯 Justification
Required whenever collective permute that needs to be cloned has more than 1 operand. It is common to have more than 1 operands with collective permute combiner pass enabled. Observed this bug in maxtext gradient accumulation.

🚀 Kind of Contribution
🐛 Bug Fix

No performance improvements expected

🧪 Unit Tests:
Add minimum hlo unit test that makes sure the cloned instruction has same number of operands.

🧪 Execution Tests:
No execution tests added.

### DIFF
--- a/xla/hlo/ir/hlo_instruction_test.cc
+++ b/xla/hlo/ir/hlo_instruction_test.cc
@@ -146,6 +146,24 @@ ENTRY main {
   TF_EXPECT_OK(module->schedule().Verify());
 }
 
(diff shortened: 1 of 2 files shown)


derdrdirk (approved):
LGTM

frgossen (approved):

PR #30682, opened 2025-08-27
Disable float normalization for bitcasts.

sergachev:
Bitcasts can be used to express no-op data type conversions.

### DIFF
--- a/xla/hlo/transforms/simplifiers/float_normalization_test.cc
+++ b/xla/hlo/transforms/simplifiers/float_normalization_test.cc
@@ -583,6 +583,16 @@ TEST_F(FloatNormalizationTest, DoNotChangeBitcastConvert) {
   EXPECT_EQ(root->operand(0)->shape().element_type(), U16);
 }
 
+TEST_F(FloatNormalizationTest, DoNotChangeBitcast) {
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnVerifiedModule(R"(
+m {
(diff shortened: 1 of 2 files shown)


akuegel (approved):

xla/hlo/transforms/simplifiers/float_normalization_test.cc
```diff
@@ -583,6 +583,16 @@ TEST_F(FloatNormalizationTest, DoNotChangeBitcastConvert) {
   EXPECT_EQ(root->operand(0)->shape().element_type(), U16);
 }
 
+TEST_F(FloatNormalizationTest, DoNotChangeBitcast) {
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnVerifiedModule(R"(
+m {
+  a = s4[4,2] parameter(0)
```
loislo: it is only physically possible if the layout is packed.

loislo: from the pr it is not quite clear why do we need that thing happening.
I mean there is a tensor with a specific number of elements. I think there is an assumption that the bitcast op does not change the number of elements. Am i right?

loislo: should it be bitcast-convert?
https://openxla.org/xla/operation_semantics#bitcastconverttype

loislo (changes requested):
what do you think?

loislo (approved):

PR #30685, opened 2025-08-27
Dump HLO when deserializing CPU executables

olupton:
📝 Summary of Changes
This enables dumping optimized HLO when deserializing CPU executables, respecting dump-related options (`--xla_dump_to` etc.) passed in when deserializing. The same change for GPU executables already landed via #28928, but the proposed JAX test jax-ml/jax#31328 currently fails on CPU due to missing plumbing in that backend.

🎯 Justification
This is useful when working with the JAX persistent compilation cache.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

🧪 Unit Tests:
`PjRtCpuClientTest.DumpOnDeserialize`, similar to the existing GPU client test.

### DIFF
--- a/xla/pjrt/cpu/cpu_client.cc
+++ b/xla/pjrt/cpu/cpu_client.cc
@@ -473,6 +473,18 @@ PjRtCpuClient::LoadSerializedExecutable(absl::string_view serialized,
     }
   }
 
+  // Propagate env_option_overrides-> debug_options
+  TF_RETURN_IF_ERROR(compile_options.ApplyAllOptionOverrides());
+  // Override the debug_options() embedded in the module with those
+  // explicitly passed in when deserializing. This allows options such as
+  // --xla_dump_to to be changed.
+  if (executable->has_module()) {
+    DumpHloModuleIfEnabled(executable->module(), kAfterOptimizationsDumpName,
+                           build_options.has_debug_options()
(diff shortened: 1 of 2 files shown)


olupton:
I ran `//xla/python/ifrt_proxy/integration_tests:array_impl_test_tfrt_cpu` locally and it seems flaky, failing O(20%) of the time whether or not I include this patch.

olupton:
> I ran `//xla/python/ifrt_proxy/integration_tests:array_impl_test_tfrt_cpu` locally and it seems flaky, failing O(20%) of the time whether or not I include this patch.

It passed when I re-triggered the CI.

xla/pjrt/cpu/cpu_client.cc
```diff
@@ -473,6 +473,18 @@ PjRtCpuClient::LoadSerializedExecutable(absl::string_view serialized,
     }
   }
 
+  // Propagate env_option_overrides-> debug_options
```
frgossen: nit: whitespace

frgossen (approved):

xla/pjrt/cpu/cpu_client.cc
olupton: Fixed in 2821a56

PR #30695, opened 2025-08-27
[XLA:GPU] Lowering command buffer's `WhileCmd` into an unrolled cuda-graph.

shawnwang18:
📝 Summary of Changes

This PR enables lowering the `WhileCmd` of `CommandBufferCmd`  into an unrolled command buffer (sub cuda-graph without the loop node).  

The main idea includes below: 

1. We now assume that `CommandBufferCmdExecutor` (`CommandBufferCmd` sequence) can be lowered in the unrolling way, and the `RecordParams` struct has a new member variable `unroll_iteration`, which indicates the current unroll iteration.  Original normal `CommandBufferCmdExecutor::Record` lowering will just be assumed as a special case,  i.e, it just has the unrolling factor 1 .

2.  Original, the `StateManager` class maintains a one-one map from `CommandBufferCmd` to the set of cuda-graph nodes lowered from the command.  Our new assumption breaks this one-one mapping , now it means that one `CommandBufferCmd`  might be lowered into multiple sets of command buffer nodes (cuda-graph nodes), each set is associated with one unrolling iteration.   To support it, we update the `StateManager`, and now each `<const CommandBufferCmd*, const se::CommandBuffer*, unroll_itearation_value>` tuple will maps to one set of cuda-graph nodes: 

```
using Key = std::tuple<const CommandBufferCmd*, const se::CommandBuffer*,
                           TypeId, int64_t>;
absl::flat_hash_map<Key, std::unique_ptr<State>> state_;

CommandBufferCmd::State* CommandBufferCmd::StateManager::GetOrCreate(
    const CommandBufferCmd* cmd, const se::CommandBuffer* command_buffer,
    TypeId type_id, int64_t unroll_iteration,
    absl::FunctionRef<std::unique_ptr<State>()> create) {
  Key key = {cmd, command_buffer, type_id, unroll_iteration};
  if (auto it = state_.find(key); it != state_.end()) {
    return it->second.get();
  }
  return state_.try_emplace(key, create()).first->second.get();
}

```

3.  To lowering the `WhileCmd` into an unrolled cuda-graph, we needs to pre-known the trip count, this depends on whether the `WhileThunk` has pre-known trip count. 

4.  To unrolled the `WhileCmd` when the trip count is pre-known, we can implement the `WhileCmd:Record` function like below,  i.e, it lowers into cuda graph `ChildNode`, whose child graph is lowered from a sequence of `CommandBufferCmdExecutor`:  `cond_iteration0` ->`body-iteration0` ->`cond_iteration1` -> `body_iteration1` ......

```
   // Unroll the while loop body for `trip_count` times.
    // Unrolled execution sequence: cond -> body -> cond -> body -> ...
    // In the unrolled pattern, we still need to run the cond commands because
    // body commands might depends on the value of index variable that is
    // updated by condition commands.
    auto new_record_params = record_params;

    for (int64_t i = 0; i < trip_count_.value(); ++i) {
      new_record_params.unroll_iteration = i;
      if (i == 0) {
        // First iteration, cond_commands_ will not have dependencies.
        TF_RETURN_IF_ERROR(cond_commands_.Record(
            execute_params, new_record_params, CommandBufferCmd::RecordCreate{},
            child_command_buffer_.get(), false));
      } else {
        // Other iterations, cond_commands_ will have dependencies on the sink
        // commands from previous iteration's body.
        auto body_sink_commands = body_commands_.SinkCommands(
            new_record_params, child_command_buffer_.get(), i - 1);
        TF_RETURN_IF_ERROR(cond_commands_.Record(
            execute_params, new_record_params,
            CommandBufferCmd::RecordCreate{absl::MakeSpan(body_sink_commands)},
            child_command_buffer_.get(), false));
      }
      auto cond_sink_commands = cond_commands_.SinkCommands(
          new_record_params, child_command_buffer_.get(), i);
      TF_RETURN_IF_ERROR(body_commands_.Record(
          execute_params, new_record_params,
          CommandBufferCmd::RecordCreate{absl::MakeSpan(cond_sink_commands)},
          child_command_buffer_.get(),
          i == trip_count_.value() - 1 ? true : false));
    }
    return Handle(
        std::move(record_action),
        [&](absl::Span<const se::CommandBuffer::Command* const> dependencies) {
          return command_buffer->CreateChildCommand(
              se::CommandBuffer::ChildCommandType::kMoved,
              *child_command_buffer_, dependencies);
        },
        [&](const se::CommandBuffer::Command* command) {
          // Moved child command does not need to be updated.
          return absl::OkStatus();
        });

```

🎯 Justification

`WhileCmd` unrolling is a good way to work around several patterns that is not supported by cuda-graph `loop` operator,  e.g.: 

1.  nccl kernels can not be included in the body sub-graph of cuda graph `loop` node, because capturing NCCL custom will get un-supported loop body graph pattern. 
2.  DynamicSliceFusion command can not be wrapped by cuda-graph `loop` node, because memory pointers will change across loop iterations. 

Above issues are very difficult by fixing cuda. As scan layer is widely used in JAX LLM models, we need to find a work around. We hope that we can use the loop unrolling to lower the whole model (full loop, not just loop body) into one cuda-graph. 

🚀 Kind of Contribution
New feature. 

🧪 Unit Tests:
command_buffer_test.cc: CommandBufferUnrollTest.WhileLoop


ezhulenev (approved):

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -1449,22 +1468,73 @@ absl::StatusOr<const se::CommandBuffer::Command*> WhileCmd::Record(
...
+    }
+    return Handle(
+        std::move(record_action),
+        [&](absl::Span<const se::CommandBuffer::Command* const> dependencies) {
+          return command_buffer->CreateChildCommand(
+              se::CommandBuffer::ChildCommandType::kMoved,
+              *child_command_buffer_, dependencies);
+        },
+        [&](const se::CommandBuffer::Command* command) {
+          // Moved child command does not need to be updated.
```
ezhulenev: Hm, I don't think it's true, if buffer allocation changes you need to update all nodes that were created from an unrolled loop?

shawnwang18: It's just the child node itself does not need to do update, because child graph is moved.  nested graph update is performed by `cond_commands_.Record()` and `body_commands_.Record()` before the last `return Handle`, right?

shawnwang18: Update the comments to describe the update process more precisly.

dimitar-asenov:
Sorry for the delay on this, it will likely take until next week, as we need further internal approvals.

dimitar-asenov:
@shawnwang18 I just noticed that this PR is marked as a draft. Is there more you need to do on it before we merge it? If not, please mark it as open.

shawnwang18:
> @shawnwang18 I just noticed that this PR is marked as a draft. Is there more you need to do on it before we merge it? If not, please mark it as open.

Ah, just opened it, I also did not noticed that, the PR is ready for merge. Thanks.

shawnwang18:
@dimitar-asenov any updates of this PR?

dimitar-asenov:
@majnemer Could you please look into the newly defined flag?

jprabhas:
Should we rename the flag xla_gpu_command_buffer_unroll_loops to something like xla_experimental_gpu_command_buffer_unroll_loops ? I see this flag as a temporary guard until the feature is stable, and we can remove it or turn it on by default later.

dimitar-asenov:
@majnemer has to weigh in on the flag changes.

majnemer:
LGTM

majnemer (approved):

PR #30704, opened 2025-08-27
Algebraic simplifier: replace no-op bitcast-converts with bitcasts also for different type bit widths.

sergachev:


akuegel (changes requested):
This requires changes to the fusion emitter and the generic triton emitter.
Only if the bitcasts are not fused, we can support it right now.
See also Johannes' document that lists adding support to fusion emitter as a necessary step

sergachev:
Should be fine after https://github.com/openxla/xla/pull/30864.

akuegel (approved):

PR #30706, opened 2025-08-27
Expose Multi-Host HLO Runner in Python

mgoldfarb-nvidia:
📝 Summary of Changes
Exposes the multi-host runner via nanobind interface for calling by Python programs that register custom calls.

🎯 Justification
HLOs containing custom calls are not executable because the custom call targets are not linked. This change provides a straightforward path by allowing for registration of calls from python.

🚀 Kind of Contribution
✨ New Feature


xla/tools/multihost_hlo_runner/python_hlo_runner.cc
```diff
@@ -0,0 +1,448 @@
...
+    }
+    for (int i = 0; i < execution_profiles.size(); ++i) {
+      LOG(INFO) << "## Execution time, file=" << hlo_file << " repeat=" << i
+                << " duration=" << execution_profiles[i].compute_time_ns()
+                << "ns" << std::endl;
+    }
+  }
+}
+
+absl::Status RegisterCustomCallTarget(const std::string& fn_name, nb::object fn,
```
frgossen, line 218: Isnt custom call registration already exposed at the Python level? I found `PyRegisterCustomCallTarget` in jaxlib/xla_compiler.cc. Can we reuse that here instead?

xla/tools/multihost_hlo_runner/python_hlo_runner.cc
```diff
@@ -0,0 +1,448 @@
...
+        "Unsupported custom call target type for api_version=1");
+  }
+
+  return absl::UnimplementedError(absl::StrFormat(
+      "API version %d is not supported by RegisterCustomCallTarget. Supported "
+      "versions are 0 and 1.",
+      api_version));
+}
+
+nb::dict GetRegisteredCustomCallTargets(const std::string& platform) {
```
frgossen, line 288: Should this live in jaxlib too?

frgossen (changes requested):
Awesome! Happy to see this happen

xla/tools/multihost_hlo_runner/python_hlo_runner.cc
mgoldfarb-nvidia: I was thinking about that too and I think you are correct we can rely on that since AIUI the registry is global to the process. Let me remove this and test it out to make sure it still works but if so we can simplify.

mgoldfarb-nvidia: @frgossen I was looking into this but there is some trickery on linkage. The Jax APIs have their own copy of jaxlib and the FFI registry so they don't get shared across the library boundary.  It wasn't clear how exactly I can link to jaxlib from within XLA project maybe this is something the Google internal version does that we can't do as easily in OSS? 

Would we need to move the builder into the Jax project instead? That seems simpler if we want tighter integration with Jax. Otherwise we can duplicate the registration code (its really minimal) for this specific use case. We could separate out some of that code in to an xla library so it can be better shared.

WWYD?

frgossen: Thought about this again. Yes, we don't want a dependency on JAX here, so I guess the little bit of duplicate code is fine. Thanks for investigating

frgossen (approved):
Awesome, thanks!

PR #30713, opened 2025-08-27
[XLA:GPU][oneAPI] Stub for IntelGpuCompiler

mdfaijul:
This is a stub for IntelGpuCompiler. Each feature would extend its implementations in the future PRs.


mdfaijul:
Mistakenly added unnecessary commit.

PR #30721, opened 2025-08-28
[XLA:GPU] Command buffer allows multiple CommandBufferCmdExecutor to be recorded into one command buffer

shawnwang18:
📝 Summary of Changes
In Current implementation, one command buffer (cuda graph) can only be recorded from one command sequence (CommandBufferCmdExecutor), i.e, command buffer will be finalized after recording the last command from the command sequence.  But in some cases, we may want to recording multiple command sequence into the same command buffer.  This PR enables this feature. 

the new CommandBufferCmdExecutor::Record function will have the prototype:  
```
  absl::Status Record(const Thunk::ExecuteParams& execute_params,
                      const RecordParams& record_params,
                      CommandBufferCmd::RecordAction record_action,
                      se::CommandBuffer* command_buffer, bool finalize = true);
```

`record_action` is the dependency sets for source command of current sequence, `finalize` is a flag to indicate whether to finalize the cuda-graph after recording current sequence. 

For example, if we have 3 command sequences: A , B, C, and we want include them into one cuda-graph, the recorded order is A -> B -> C, where B's source commands depend on A's sink commands, and C's source commands depend on B's sink commands. 

🎯 Justification

This new record pattern will make constructing cuda-graph more easily in some cases.  For example, if we want to unroll the loop (do not use cuda-graph while node to modeling the loop), we can just recorded the loop body sequence into one command buffer in sequence very easily:  Iteration1.Record() -> Iteration2.Record() -> ....
   

🚀 Kind of Contribution
✨ New Feature

This is a pre-requisite feature for command buffer loop unrolling, so no perf number in this PR 

🧪 Unit Tests:
command_buffer_cmd_test.cc: CommandBufferCmdTest.RecordExecutorsWithDependencies

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/backends/gpu/runtime/command_buffer_cmd.cc
@@ -457,7 +457,8 @@ absl::Status CommandBufferCmdExecutor::Initialize(
 absl::Status CommandBufferCmdExecutor::Record(
     const Thunk::ExecuteParams& execute_params,
     const CommandBufferCmd::RecordParams& record_params,
-    se::CommandBuffer* command_buffer) {
+    CommandBufferCmd::RecordAction record_action,
+    se::CommandBuffer* command_buffer, bool finalize) {
   VLOG(3) << "Record " << commands_.size() << " commands into command buffer";
 
   if (command_buffer->state() == se::CommandBuffer::State::kFinalized) {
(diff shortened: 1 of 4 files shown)


shawnwang18:
@ezhulenev Could you help review this PR?

ezhulenev (approved):

PR #30735, opened 2025-08-28
[NFC] Remove META scope from triton fusion analysis

sergey-kozub:
📝 Summary of Changes
Remove `TritonFusionAnalysis::Scope::META` enum value, which is not used.
Next PR will add scopes for LHS/RHS scales.

🚀 Kind of Contribution
♻️ Cleanup


loislo (approved):

bchetioui:
Thank you @sergey-kozub!

PR #30805, opened 2025-09-01
[GPU] Fix and re-enable layout assignment tests for convolutions.

sergachev:
Current 3 test cases are always skipped because xla_cc_test is deviceless.


beckerhe (approved):
Nice. Thank you!

PR #30864, opened 2025-09-02
[GPU] Disable priority and multi-output fusion of bitwidth-changing bitcasts.

sergachev:
These better stay outside fusions to remain truly no-ops - inside fusions they can affect indexing and result in suboptimal access patterns.


akuegel (approved):
Agree, this is likely the best solution. Right now, it should be a no-op, but once AlgebraicSimplifier also turns those bitcast-converts into bitcasts, I will trigger a benchmark run just to verify that we don't lose performance.

PR #30967, opened 2025-09-04
[XLA:GPU] Allow split-k transformation on block scaled dot fusions

sergey-kozub:
📝 Summary of Changes
Modifies the SplitK GEMM rewriter to support fusions around scaled dot HLO op.

🎯 Justification
This PR will allow the triton legacy emitter to handle block scaled dot.

🚀 Kind of Contribution
✨ New Feature (part of a CL chain)

- If padding is needed, pad the scaling tensors first and multiply by the block size to get the contracting dimension size (note that block sizes may be different or missing, i.e. scalar scale);
- Adds DCE in case the rewriter fails after updating the computation (also affects regular dot fusion);
- Exposes block sizes in `TritonFusionAnalysis` (nullopt for scalars);
- Adds `MakeScaledDotHlo` helper and scaled dot pattern matcher;


xla/service/gpu/triton_fusion_analysis.cc
```diff
@@ -246,9 +266,16 @@ absl::Status TritonFusionAnalysis::ExecuteForDotFusion(
     if (operand_number >= dot.operand_count()) {
       continue;  // Scale operands are optional.
     }
-    if (is_scaled_dot_ && (operand_number == 1 || operand_number == 2)) {
+    if (is_scaled_dot_) {
       // Operands for scaled dot: (lhs, lhs_scale, rhs, rhs_scale)
```
mooskagh, line 292: Maybe make enum order in `Scope` the same as in the instruction? It doesn't look like anything (rather than function) depends on the index.

Or if it's not straightforward, a helper "ScopeToScaledDotOperandIdx" function would be cleaner.

xla/service/gpu/triton_fusion_analysis.h
```diff
@@ -97,14 +97,19 @@ class TritonFusionAnalysis {
...
+  std::optional<int64_t> lhs_block_size() const { return lhs_block_size_; }
+  std::optional<int64_t> rhs_block_size() const { return rhs_block_size_; }
 
  private:
   IterationSpecByInstructionByScopeMap iter_specs_;
   // HLO computation parameters per scope.
   std::map<Scope, ConstHloInstructionSet> parameters_;
   // Scaled dot has additional scale scopes.
   bool is_scaled_dot_ = false;
+  std::optional<int64_t> lhs_block_size_;
```
mooskagh: "block_size" is too overloaded term in dot.
Maybe "scale_block_size"?
It would also be nice to mention that that's the block size along the contracting dimension, maybe "`lhs_k_scale_block_size_`"?

xla/service/gpu/triton_fusion_analysis.cc
```diff
@@ -186,6 +186,22 @@ absl::Status FusionContext::PropagateDimensionOrdersToParameters(
...
+                                    TritonFusionAnalysis::Scope scope) {
+  CHECK(dot.opcode() == HloOpcode::kScaledDot);
+  CHECK(scope == TritonFusionAnalysis::Scope::LHS ||
+        scope == TritonFusionAnalysis::Scope::RHS);
+  int operand_number = scope == TritonFusionAnalysis::Scope::LHS ? 0 : 2;
+  const Shape& input = dot.operand(operand_number)->shape();
+  const Shape& scale = dot.operand(operand_number + 1)->shape();
+
+  if (!ShapeUtil::IsScalar(scale)) {
+    int dim_idx = ContractingDimensionIndex(dot, operand_number).value();
```
mooskagh: `TF_ASSIGN_OR_RETURN(`...

xla/service/gpu/split_k_gemm_rewriter.cc
```diff
@@ -366,8 +443,15 @@ absl::Status MakeDotSplitKBatch(HloInstruction* dot_fusion,
...
   const Layout output_layout = dot_fusion->shape().layout();
 
-  TF_RETURN_IF_ERROR(MakeDotComputationSplitKBatch(
-      dot_fusion->fused_instructions_computation(), config));
+  auto status = MakeDotComputationSplitKBatch(
+      dot_fusion->fused_instructions_computation(), config);
+  if (!status.ok()) {
+    TF_RETURN_IF_ERROR(
+        HloDCE()
+            .RunOnComputation(dot_fusion->fused_instructions_computation())
```
mooskagh, line 457: Why is this needed? (or rather, why did it start being needed after adding support for scales?)

mooskagh (changes requested):

xla/service/gpu/split_k_gemm_rewriter.cc
sergey-kozub: I just discovered and fixed a pre-existing bug. Running the "CleanupUncompilable" test at the repo HEAD will fail.

xla/service/gpu/triton_fusion_analysis.cc
sergey-kozub: RHS scope has different index for kDot and kScaledDot. Added the helper, as you suggest.

xla/service/gpu/triton_fusion_analysis.h
sergey-kozub: Renamed.

xla/service/gpu/triton_fusion_analysis.cc
sergey-kozub: Updated.

xla/service/gpu/split_k_gemm_rewriter.cc
mooskagh: Where exactly is the CleanupUncompilable test? I seem to unable to find this test in the codebase.

Having to run DCE on failure sounds wrong.. Ideally we would check whether the transformation is possible and then perform, rather than stopping halfway.
But if it's not easy in this case, maybe attempt the transformation on the clone, and then replace the final computation if everything is successful?

sergey-kozub: You can find it on the line 795 of "xla/service/gpu/split_k_gemm_rewriter_test.cc"

One can find may examples of the passes that run DCE (or other passes). In this specific case, this change is unrelated to the scaled dot PR chain. If you don't mind, I'll do what you suggest in a separate PR.

mooskagh: I've discussed that internally. We don't want to run passes from other passes, especially given that in theory DCE can clean up more instructions than were before (i.e. if there were already dangling instructions when the pass was falled), and we should clean up other places too.

In this case, if it's not easy to know in advance whether the transformation is possible, attempting it on a clone of the computation first sounds clean enough.

It's ok to do that in a follow up PR.

xla/service/gpu/split_k_gemm_rewriter.cc
```diff
@@ -265,29 +278,92 @@ absl::Status MakeDotComputationSplitKBatch(HloComputation* computation,
...
+        // Keep the precision of the accumulator type for the dot output.
+        TF_ASSIGN_OR_RETURN(
+            expanded, MakeDotHlo(lhs, rhs, new_dim_numbers,
+                                 dot->precision_config(), accumulator_dtype));
+      } else {
+        // Scaled dot operation.
+        // At least one scaling operand is not a scalar and will define the
+        // contracting dimension size.
+        std::optional<int64_t> contracting_size;
+        HloInstruction* lhs_scale = dot->mutable_operand(1);
```
mooskagh, line 329: The code for lhs and rhs is very similar.
Maybe it makes sense to factor it out into a lambda function (and e.g. pass operand number, l/rhs_contracting_idx, TritonFusionAnalysis::Scope) and call twice?

xla/service/gpu/split_k_gemm_rewriter.cc
```diff
@@ -265,29 +278,92 @@ absl::Status MakeDotComputationSplitKBatch(HloComputation* computation,
     }
   } while (true);
 
-  // Process the collected HLOs from computation root to dot.
+  // Keep track of whether any of the operands were padded.
   bool did_pad = false;
-  PrimitiveType accumulator_dtype = GetGemmAccumulatorType(dot);
+  auto is_padded = [](const HloInstruction* op) {
+    return op->operand_count() > 0 &&
```
mooskagh: Why do we check operand's operand and not operand itself?
Also, why do only check its first operand?

xla/service/gpu/split_k_gemm_rewriter.cc
```diff
@@ -265,29 +278,92 @@ absl::Status MakeDotComputationSplitKBatch(HloComputation* computation,
...
+        did_pad = is_padded(lhs) || is_padded(rhs);
+        // Keep the precision of the accumulator type for the dot output.
+        TF_ASSIGN_OR_RETURN(
+            expanded, MakeDotHlo(lhs, rhs, new_dim_numbers,
+                                 dot->precision_config(), accumulator_dtype));
+      } else {
+        // Scaled dot operation.
+        // At least one scaling operand is not a scalar and will define the
+        // contracting dimension size.
+        std::optional<int64_t> contracting_size;
```
mooskagh: Maybe rename `contracting_size` to something like `padded_k_size`, otherwise it's a bit confusing why we have both `k` and `contracting_size` and what is the relationship between them.

xla/service/gpu/split_k_gemm_rewriter.cc
sergey-kozub: Added a lambda.

xla/service/gpu/split_k_gemm_rewriter.cc
sergey-kozub: The operand is a bitcast op, it's first operand is maybe a pad. This logic hasn't changed compared to the previous version.

xla/service/gpu/split_k_gemm_rewriter.cc
sergey-kozub: Renamed.

xla/service/gpu/split_k_gemm_rewriter.cc
```diff
@@ -112,14 +113,14 @@ absl::Status UncompilableMatmul(absl::string_view explanation) {
 absl::StatusOr<HloInstruction*> MakeSplitKOperand(
     HloInstruction& dot, const TritonFusionAnalysis& analysis,
     const TritonGemmConfig& config, const int64_t contracting_dim_idx,
-    const int operand_number) {
+    TritonFusionAnalysis::Scope scope, const int operand_number,
+    std::optional<int64_t> contracting_size = std::nullopt) {
```
mooskagh: Sorry for not being clear, I meant rename `contracting_size` also in this function, not just as the field.

xla/service/gpu/split_k_gemm_rewriter.cc
mooskagh: Maybe then make it explicit? I.e. check that the op is kBitcast, and its operand is kPad?

mooskagh (approved):

xla/service/gpu/split_k_gemm_rewriter.cc
sergey-kozub: Renamed this one as well.

xla/service/gpu/split_k_gemm_rewriter.cc
sergey-kozub: Updated.

mooskagh (approved):

PR #30984, opened 2025-09-04
[NVIDIA GPU] Use both parallel and host threads for processing scheduling passes

Tixxx:
📝 Summary of Changes
This is to fix a bug where scheduling pipelines only use the main host execution threads(introduced in this https://github.com/openxla/xla/pull/30487).
It should use both main host and parallel threads as there are passes before scheduler that convert computes into async computes that use parallel threads

🎯 Justification
If async compute is present, the scheduling pass will error out with
"absl::raw_hash_map out of range"

🚀 Kind of Contribution

NA

🧪 Unit Tests:
Added scheduler test to verify changes

🧪 Execution Tests:

### DIFF
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -2235,6 +2235,7 @@ cc_library(
         "//xla/service/gpu/model:sol_latency_estimator",
         "//xla/service/gpu/transforms:pgle_accuracy_checker",
         "//xla/service/gpu/transforms:scheduling_instruction_annotator",
+        "//xla/service/gpu/transforms:stream_attribute_async_wrapper",
         "//xla/service/gpu/transforms/collectives:async_collective_annotator",
         "//xla/service/gpu/transforms/collectives:collective_ops_utils",
         "//xla/stream_executor:device_description",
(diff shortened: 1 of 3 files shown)


frgossen (approved):

penpornk (commented):
> Minimizing the original repro for an e2e test

Could you please clarify what this means? If it's covered by an existing e2e test, could you please specify the test target? Otherwise, could you please give a reason we don't need to update tests?

felixwqp (approved):

felixwqp:
I ran this CL inside Google environment using the CM maxtext repro shared from NV side. 
* Without this patch, CM example crashed at LHS
* with this patch, functionality-wise the CM maxtext workload can  pass.

Tixxx:
> > Minimizing the original repro for an e2e test
> 
> Could you please clarify what this means? If it's covered by an existing e2e test, could you please specify the test target? Otherwise, could you please give a reason we don't need to update tests?

Hi I just meant that i was working on the test by minimizing the original repro hlo.
Added a new commit with the test.

penpornk (approved):
Thank you very much!

PR #30996, opened 2025-09-04
[XLA:GPU] remove cudnn sdpa dbias constraint

Cjkkkk:
📝 Summary of Changes
Allow cudnn dbias sdpa with bias shape [1, 1, s, s], [b,1, s, s], [b, h, s, s] in addition to [1, h, s, s].

🎯 Justification
Support dbias computation with more types of bias shape broadcasting.

🚀 Kind of Contribution
✨ New Feature

No public hlo uses new shapes of bias.

🧪 Unit Tests:
No Hlo changes, mostly rely on cuDNN to do the job

🧪 Execution Tests:
Xla had 2 dbias tests **Flash_Attention_BMM1_Bias_Softmax_BMM2_BF16_Dbias** and **Flash_Attention_Training_BMM1_Bias_Softmax_BMM2_BF16** for same [1, h, s, s] shape bias, modify **Flash_Attention_BMM1_Bias_Softmax_BMM2_BF16_Dbias** test to use [1,1, s, s] shape bias.


Cjkkkk:
This also fixed the jax issue: https://github.com/jax-ml/jax/issues/31630

Cjkkkk:
@akuegel could you help take a look at this? Thanks!

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -4795,21 +4795,17 @@ absl::StatusOr<CudnnGraph> GetCudnnFlashAttentionBackwardOperationGraph(
...
                                         .set_stride(bias_strides)
                                         .set_uid(next_uid()));
     sdpa_backward_options.set_bias(bias_tensor);
 
-    // shapes [1, 1, s, s], [b, 1, s, s], [b, h, s, s] are not supported for
-    // dbias calculation but they are supported for forward bias calculation
-    // Set UID later: this is the last output tuple element.
+    // shapes [1, 1, s, s], [1, h, s, s], [b, 1, s, s], [b, h, s, s] are
+    // supported for dbias calculation.
+    // calculation Set UID later: this is the last output tuple element.
```
akuegel: nit: the word calculation at the beginning of this line should be removed.

akuegel (commented):
As far as I can tell, the change is to remove a DCHECK. For my understanding, how does this resolve the linked JAX issue where they complain about wrong results? I would expect that it would either lead to a CHECK failure (in debug mode) or already do the same thing as with the DCHECK removed (in opt mode).

akuegel:
> As far as I can tell, the change is to remove a DCHECK. For my understanding, how does this resolve the linked JAX issue where they complain about wrong results? I would expect that it would either lead to a CHECK failure (in debug mode) or already do the same thing as with the DCHECK removed (in opt mode).

Maybe you wanted to reference https://github.com/jax-ml/jax/issues/30593 ? It seems plausible that it will fix this issue.

Cjkkkk:
> > As far as I can tell, the change is to remove a DCHECK. For my understanding, how does this resolve the linked JAX issue where they complain about wrong results? I would expect that it would either lead to a CHECK failure (in debug mode) or already do the same thing as with the DCHECK removed (in opt mode).
> 
> Maybe you wanted to reference [jax-ml/jax#30593](https://github.com/jax-ml/jax/issues/30593) ? It seems plausible that it will fix this issue.

Right, both issues are related here. https://github.com/jax-ml/jax/issues/30593 should be the one that this PR addressed immediately. https://github.com/jax-ml/jax/issues/31630 also relies on this PR because JAX sets dbias to all 0 for [b, h, s, s] shape bias which leads to loss spike/ accuracy loss. To correctly fix https://github.com/jax-ml/jax/issues/31630, we first need to allow [b, h, s, s] shape dbias in JAX  and then we need to let XLA cuDNN SDPA supports this as well.

xla/stream_executor/cuda/cuda_dnn.cc
akuegel: @Cjkkkk In case you missed this comment, see above.

Cjkkkk: addressed, sorry missed that earlier.

Cjkkkk:
@akuegel Hi Adrian, could you take another look? Thanks!

akuegel (approved):

PR #30997, opened 2025-09-04
[XLA:CPU][oneDNN] Enable oneDNN MatMul Custom Calls in Thunk Runtime

othakkar:
This PR enables support for oneDNN MatMul operations in the XLA:CPU Thunk runtime, building upon the foundational implementation of `OneDnnOpThunk`.

Key changes:
- Added `onednn_op_thunk.cc` and `onednn_op_thunk.h` implementing the base `OneDnnOpThunk`.
- Extended `onednn_memory_util.h` with `OneDnnResources` to manage primitives and memory objects across async execution.
- Updated thunk emitter to emit `OneDnnOpThunk` for supported oneDNN custom call ops during compilation.
- Enabled oneDNN custom call rewrite for MatMul in `cpu_compiler.cc` and added support for oneDNN MatMul op via `ExecuteOneDnnMatMul(...)`.
- Updated Bazel build rules to register new library and test targets.


othakkar:
@penpornk I created another PR (https://github.com/openxla/xla/pull/31025) to add the flag to enable oneDNN Custom Calls in Thunk Runtime. If PR https://github.com/openxla/xla/pull/31025 can be merged first, then I can update this PR to use the new flag instead of the ifdef `ENABLE_ONEDNN_ASYNC` in `cpu_compiler.cc`. Please let me know what works best for you. Thanks!

xla/backends/cpu/runtime/onednn/onednn_op_thunk.h
```diff
@@ -0,0 +1,62 @@
...
+#include "xla/backends/cpu/runtime/custom_call_thunk.h"
+#include "xla/backends/cpu/runtime/thunk.h"
+#include "xla/runtime/object_pool.h"
+#include "xla/service/cpu/onednn_memory_util.h"
+#include "xla/service/cpu/onednn_util.h"
+#include "xla/tsl/concurrency/async_value_ref.h"
+
+namespace xla::cpu {
+
+typedef CustomCallThunk::OpBuffers OpBuffers;
```
penpornk: Let's just define the `OpBuffers` struct inside `OneDnnOpThunk` and not depend on `CustomCallThunk`. Eugene will clean up the duplicates later.

P.S. This is irrelevant now, but just an FYI: We prefer `using` instead of `typedef`, i.e., `using CustomCallThunk::OpBuffers`.

xla/backends/cpu/runtime/onednn/BUILD
```diff
@@ -110,3 +111,34 @@ onednn_graph_cc_test(
...
+    name = "onednn_op_thunk",
+    srcs = ["onednn_op_thunk.cc"],
+    hdrs = [
+        "onednn_op_thunk.h",
+    ],
+    copts = tsl_copts(),
+    visibility = ["//visibility:public"],
+    deps = [
+        ":onednn_threadpool",
+        "//xla/backends/cpu/runtime:custom_call_thunk",
```
penpornk: Let's not depend on `custom_call_thunk`.

xla/service/cpu/cpu_compiler.cc
```diff
@@ -673,6 +673,12 @@ absl::Status CpuCompiler::RunHloPassesThroughLayoutAssn(
   }
 #endif  // INTEL_MKL
 
+#ifdef ENABLE_ONEDNN_ASYNC
+  // TODO(intel-tf): Remove this ifdef after enabling the OneDnnOpsRewriter with
+  // thunk runtime.
+  is_onednn_compatible = !is_aot_compile;
+#endif  // ENABLE_ONEDNN_ASYNC
```
penpornk: I think the code is more complicated than it needs to be (considering all the `is_onednn_compatible` value setting before this. Can we just define an inline function:
```
inline bool IsOneDnnCompatible(bool is_aot_compile) {
#ifdef ENABLE_ONEDNN_ASYNC
  return !aot_compile;
#endif
  return false;
}
```
and call it in both `RunHloPassesThroughLayoutAssn` and `RunHloPassesAfterLayoutAssn`?

As for the `if` to add `OneDnnOpsRewriter` pass, you can just do `if (!is_thunk_runtime)`. Or please feel free to remove the whole thing and add the pass back later when it's enabled as well.

xla/service/cpu/onednn_matmul.cc
```diff
@@ -212,6 +212,89 @@ CreateOneDnnPrimDesc<dnnl::matmul::primitive_desc>(HloInstruction* instr) {
...
+    resources.scratch_mem =
+        memory(scratchpad_md, cpu_engine, scratch_minfo.Data());
+    matmul_args.insert({DNNL_ARG_SCRATCHPAD, resources.scratch_mem});
+  }
+
+  matmul_args.insert(resources.postop_args.begin(),
+                     resources.postop_args.end());
+
+  resources.primitive.execute(onednn_stream, matmul_args);
+}
```
penpornk, line 293: Nit: Please add a blank line after this line.

xla/service/cpu/onednn_matmul.cc
```diff
@@ -212,6 +212,89 @@ CreateOneDnnPrimDesc<dnnl::matmul::primitive_desc>(HloInstruction* instr) {
...
+    // input and output.
+    // TODO(intel-tf): Add support for prepacked weights for higher than 2D
+    // array.
+    weights_md =
+        memory::desc({input_md.get_dims().back(), output_md.get_dims().back()},
+                     weights_md.get_data_type(), weight_format);
+  }
+
+  const int64_t num_fused_operands =
+      arguments.size() - 2;  // excluding input and weights operands
```
penpornk: Nit: Let's put the comment in its own line to not break the statement into two lines.
```suggestion
  // Excluding input and weight operands.
  const int64_t num_fused_operands = arguments.size() - 2;  
```

xla/backends/cpu/runtime/onednn/onednn_op_thunk.cc
```diff
@@ -0,0 +1,195 @@
...
+OneDnnOpThunk::OneDnnRuntime::Invoke(Eigen::ThreadPoolInterface* thread_pool,
+                                     absl::Span<MemrefInfoHandler> arguments,
+                                     absl::Span<MemrefInfoHandler> results,
+                                     const std::string& config,
+                                     const std::string& target) {
+  // Update threadpool
+  threadpool->set_thread_pool(thread_pool);
+
+  // TODO(intel-tf): Add support for more oneDNN operations as needed.
+  static bool logged_once = false;
```
penpornk: Please use `absl::call_once` instead.
Example:
https://github.com/openxla/xla/blob/89661cf56f8480b15ea2886ad420e73df1424e9b/xla/backends/profiler/gpu/cupti_pm_sampler_impl.cc#L153-L165

xla/service/cpu/thunk_emitter.cc
```diff
@@ -1287,6 +1288,29 @@ static absl::StatusOr<CustomCallThunk::OpBuffers> GetCustomCallOpBuffers(
...
+    const HloInstruction* instruction) {
+  auto custom_call = Cast<HloCustomCallInstruction>(instruction);
+  auto custom_call_target = custom_call->custom_call_target();
+  auto backend_config = custom_call->backend_config<BackendConfig>();
+
+  std::string config_str;
+  if (custom_call_target == "__onednn$matmul") {
+    OneDnnMatMulConfig matmul_config;
+    matmul_config.CopyFrom(backend_config->onednn_matmul_config());
+    matmul_config.SerializeToString(&config_str);
```
penpornk: There is no need to serialize to string. You can pass the config as `OneDnnMatMulConfig` directly.

penpornk (changes requested):
Thank you very much for the PR!

xla/backends/cpu/runtime/onednn/onednn_op_thunk.h
othakkar: I defined `OpBuffers` struct inside `OneDnnOpThunk` and removed the dependency on `custom_call_thunk`, but that required me to make modifications to the `GetCustomCallOpBuffers` function in `thunk_emitter.cc`. Since the function only supported a return type of `CustomCallThunk::OpBuffers`, I templatized the function to support `OneDnnOpThunk::OpBuffers` also.

xla/backends/cpu/runtime/onednn/BUILD
othakkar: Done.

xla/service/cpu/cpu_compiler.cc
othakkar: Done. I used your suggestion to do `if (!is_thunk_runtime)` for `OneDnnOpsRewriter` as we plan to add support for that soon.

xla/service/cpu/onednn_matmul.cc
othakkar: Done.

xla/backends/cpu/runtime/onednn/onednn_op_thunk.cc
othakkar: Done.

xla/service/cpu/thunk_emitter.cc
othakkar: Done.

othakkar:
@penpornk thanks for reviewing the PR quickly! I addressed all of your review comments. Please re-review.

xla/service/cpu/onednn_matmul.cc
othakkar: done.

penpornk (approved):
Thank you for the changes!

othakkar:
@penpornk I added a fix for the "Failed to build" tests on the CI. Could you please help re-trigger the tests? Thanks.

penpornk (changes requested):
I have re-triggered the tests.

Could you please add onednn_thunk_test.cc from  https://github.com/openxla/xla/pull/30562 too (renaming it to onednn_op_thunk_test.cc)? We need unit tests to show that the code works.

xla/backends/cpu/runtime/onednn/onednn_op_thunk.h
```diff
@@ -0,0 +1,73 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef XLA_BACKENDS_CPU_RUNTIME_ONEDNN_ONEDNN_THUNK_H_
+#define XLA_BACKENDS_CPU_RUNTIME_ONEDNN_ONEDNN_THUNK_H_
```
penpornk:
```suggestion
#ifndef XLA_BACKENDS_CPU_RUNTIME_ONEDNN_ONEDNN_OP_THUNK_H_
#define XLA_BACKENDS_CPU_RUNTIME_ONEDNN_ONEDNN_OP_THUNK_H_
```

xla/backends/cpu/runtime/onednn/onednn_op_thunk.h
```diff
@@ -0,0 +1,73 @@
...
+  struct OneDnnRuntime;
+
+  OpBuffers op_buffers_;
+  OneDnnOpConfig config_;
+  std::string target_;
+};
+
+}  // namespace xla::cpu
+
+#endif  // XLA_BACKENDS_CPU_RUNTIME_ONEDNN_ONEDNN_THUNK_H_
```
penpornk:
```suggestion
#endif  // XLA_BACKENDS_CPU_RUNTIME_ONEDNN_ONEDNN_OP_THUNK_H_
```

xla/backends/cpu/runtime/onednn/onednn_op_thunk.cc
```diff
@@ -0,0 +1,192 @@
...
+  // TODO(intel-tf): Add support for more oneDNN operations as needed.
+  static absl::once_flag log_once_flag;
+  absl::call_once(log_once_flag, [&] {
+    VLOG(0) << absl::StreamFormat(
+        "Executing oneDNN thunk with target `%s`: num_args=%d, num_results=%d",
+        target, arguments.size(), results.size());
+  });
+
+  if (target == "__onednn$matmul") {
+    const auto& matmul_config = absl::get<OneDnnMatMulConfig>(config);
```
penpornk: For this function, we prefer the standard version.
```suggestion
    const auto& matmul_config = std::get<OneDnnMatMulConfig>(config);
```

penpornk (changes requested):
Some more nits please.

othakkar:
> I have re-triggered the tests.
> 
> Could you please add onednn_thunk_test.cc from #30562 too (renaming it to onednn_op_thunk_test.cc)? We need unit tests to show that the code works.

Thanks for re-triggering the tests. 
Regarding unit tests, would it make sense to rely on the existing `onednn_matmul_test` to verify the changes? Since the this non-FFI version of `onednn_op_thunk` currently only supports `matmul`, it seems more representative to test it through `onednn_matmul_test`, unlike the previous FFI-based `onednn_op_thunk` implementation in #30562, which could handle an arbitrary FFI handler. Please let me know your thoughts.

Although I realize that to run any test, you will need to build with `--define=build_with_onednn_async=true`.

xla/backends/cpu/runtime/onednn/onednn_op_thunk.h
othakkar: Sorry, I forgot to update this. Thanks, updated now.

xla/backends/cpu/runtime/onednn/onednn_op_thunk.h
othakkar: done.

xla/backends/cpu/runtime/onednn/onednn_op_thunk.cc
othakkar: Got it, done!

penpornk:
> Regarding unit tests, would it make sense to rely on the existing onednn_matmul_test to verify the changes?

No, because `onednn_matmul_test` relies on OneDnnContractionRewriter to change `dot` HLOs into oneDNN custom call. The pass doesn't have a way to enable in the thunk runtime right now (since the flag PR https://github.com/openxla/xla/pull/31025 is still ongoing). It also relies on thunk_emitter to emit the thunk properly from the HLOs.

I think we should have unit tests that show that the thunk works on its own, e.g., https://github.com/openxla/xla/blob/d2b2730b68a38886d613ce0f8e709d50a3b0e05e/xla/backends/cpu/runtime/xnnpack/xnn_fusion_thunk_test.cc

othakkar:
@penpornk thanks for the clarification. I added the `onednn_op_thunk_test`. Let me know if this looks good.

penpornk (approved):
Thank you for the changes!

PR #31003, opened 2025-09-05
[XLA::CPU][OneDnn]Async threadpool for onednn_fusion_thunk

sachinmuradi:
This PR updates the [OneDnnthreadpool ](https://github.com/openxla/xla/blob/09bb90bed5d0c036e1cdc67a6b4706016e7c8adc/xla/backends/cpu/runtime/onednn/onednn_fusion_thunk.cc#L68)used for creating OneDnnRuntime object inside onednn_fusion_thunk, to use threadpool with asynchronous execution. 
This results in improved the performance for onednn_fusion_thunks .

### DIFF
--- a/xla/backends/cpu/runtime/onednn/onednn_fusion_thunk.cc
+++ b/xla/backends/cpu/runtime/onednn/onednn_fusion_thunk.cc
@@ -49,9 +49,10 @@ struct OneDnnFusionThunk::OneDnnRuntime {
   OneDnnRuntime(OneDnnRuntime&&) = default;
   OneDnnRuntime& operator=(OneDnnRuntime&&) = default;
 
-  absl::Status Invoke(Eigen::ThreadPoolInterface* thread_pool,
-                      absl::Span<se::DeviceMemoryBase> arguments,
-                      absl::Span<se::DeviceMemoryBase> results);
+  tsl::AsyncValueRef<OneDnnFusionThunk::ExecuteEvent> Invoke(
+      Eigen::ThreadPoolInterface* thread_pool,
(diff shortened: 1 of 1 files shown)


xla/backends/cpu/runtime/onednn/onednn_fusion_thunk.cc
```diff
@@ -208,11 +211,14 @@ tsl::AsyncValueRef<OneDnnFusionThunk::ExecuteEvent> OneDnnFusionThunk::Execute(
...
-                                     absl::MakeSpan(arguments_buffers),
-                                     absl::MakeSpan(results_buffers)));
-
-  return OkExecuteEvent();
+  auto executed =
+      runtime->Invoke(thread_pool, absl::MakeSpan(arguments_buffers),
+                      absl::MakeSpan(results_buffers));
+  executed.AndThen([runtime = std::move(runtime)]() mutable {
+    // runtime will be destroyed here
+  });
```
penpornk: Nit: The comment being in the function makes it looks like a TODO when it's not.
```suggestion
  // Destroy the runtime after the task is done.
  executed.AndThen([runtime = std::move(runtime)] {});
```

penpornk (changes requested):
Thank you for the PR!

xla/backends/cpu/runtime/onednn/onednn_fusion_thunk.cc
sachinmuradi: Thanks @penpornk , Updated the file with suggested change.

penpornk (approved):
Thank you!

PR #31025, opened 2025-09-05
[XLA:CPU][oneDNN] Add a flag to enable oneDNN Custom Calls in Thunk Runtime

othakkar:
OneDNN custom calls are currently being enabled in an experimental mode. This PR adds a runtime flag (default `false`) to enable oneDNN rewrites via a custom call for the thunk runtime.


penpornk (approved):
Thank you for the PR!

PR #31027, opened 2025-09-05
Schedule cross-host transfers on XLA launcher threads to prevent deadlock

rao-ashish:
Summary of Changes
- Schedules cross-host data transfers using the XLA launcher threads instead of a separate thread pool.

Justification
- The current implementation can deadlock in the presence of other computations. An example of this can be found [here](https://gist.github.com/rao-ashish/e37d7460c7ef5225587a8291b63f6bfe), where a deadlock is observed in a program using cross-host device puts with `jax.experimental.multihost_utils.sync_global_devices()`.

Kind of Contribution
- Bug Fix

Tests
- Verified with [this script](https://gist.github.com/rao-ashish/e37d7460c7ef5225587a8291b63f6bfe) that deadlock occurs in the original implementation, but not with the change made by this PR.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -1039,7 +1039,7 @@ void StreamExecutorGpuClient::CopyToRemoteDevice(
       on_done(absl::OkStatus(), /*sends_were_enqueued=*/true);
     }
   };
-  thread_pool()->Schedule(send);
+  (*local_device)->execute_thread()->Schedule(send);
 }
 
 absl::StatusOr<std::vector<std::unique_ptr<PjRtBuffer>>>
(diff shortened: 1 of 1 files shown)


gspschmid:
Note that the context here is the relatively new ability of jax.device_put to perform cross-process transfers via NCCL send+recv on the GPU platform. The issue here is basically that both XLA itself might enqueue NCCL operations (via the `sync_global_devices` in Ashish's example) and Pjrt via the eager-mode jax.device_put. But since those operations are enqueued from different threads we might deadlock. The proposed resolution is to also enqueued jax.device_put-originated NCCL ops from the device's XLA launcher thread.

mjsML:
@PatriosTheGreat any feedback / input here?

gspschmid:
(We can close this one at it's ultimately been merged this via https://github.com/openxla/xla/pull/32074 .)

PR #31097, opened 2025-09-09
Add HNS Rename folder and storage layout API 

deepikarajani24:
📝 Summary of Changes
Adding rename folder API for HNS bucket in the existing rename method. In order to identify if bucket is HNS enabled using storage layout api 

🎯 Justification
Rename on HSN bucekt without this new API involves slow copy of every object and subfolders. This is an efficient, fast atomic operation which customers can use in their checkpointing workload 

🚀 Kind of Contribution
✨ New Feature, 🧪 Tests

N/A. This PR is a filesystem enhancement (improving GCS HNS rename operations) and does not directly affect the computational performance or compilation time of the HLO benchmarks.

🧪 Unit Tests:
transformed HLO.
N/A for the HLO examples, as this is a C++ filesystem change.
Unit tests were added to gcs_file_system_test.cc to validate the new rename logic and its fallbacks:
- The successful HNS rename path, including multi-poll LRO logic.
- Correct parsing of JSON error responses from the LRO.
- The critical fallback to the legacy copy/delete logic for non-HNS buckets.

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/tsl/platform/cloud/gcs_file_system.cc
+++ b/xla/tsl/platform/cloud/gcs_file_system.cc
@@ -103,8 +103,8 @@ constexpr size_t kMatchingPathsCacheDefaultMaxEntries = 1024;
 // Number of bucket locations cached, most workloads wont touch more than one
 // bucket so this limit is set fairly low
 constexpr size_t kBucketLocationCacheMaxEntries = 10;
-// Number of bucket storage layout cached, most workloads wont touch more than one
-// bucket so this limit is set fairly low
+// Number of bucket storage layout cached, most workloads wont touch more than
+// one bucket so this limit is set fairly low
 constexpr size_t kStorageLayoutCacheMaxEntries = 10;
 // LRUCache that has 30 mins expiration.
 constexpr uint64 kStorageLayoutCacheMaxAgeSecs = 30 * 60;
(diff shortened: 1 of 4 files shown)


xla/tsl/platform/cloud/gcs_file_system.cc
```diff
@@ -2122,6 +2207,99 @@ absl::Status GcsFileSystem::DeleteRecursively(const string& dirname,
...
+                              operation_response["error"].toStyledString());
+    }
+
+    if (operation_response.isMember("done")) {
+      bool done = false;
+      TF_RETURN_IF_ERROR(GetBoolValue(operation_response, "done", &done));
+      if (done) {
+        break; 
+      }
+    }
```
MichaelHudgins, line 2297: This is the first addition to long operations to the gcs_file_system.  @jakeharmon8 you are more familiar with tsl.  Is this consistent with other loops that can take significantly long periods of time or should this have a hard break condition?

deepikarajani24: That's a great question. To add some context, we decided against adding a timeout prematurely because the operation's length is unpredictable, and we wanted to avoid arbitrarily terminating legitimate, long-running renames. A fixed timeout could fail for very large folders.

jakeharmon8: Yeah, I think it makes sense that this would keep polling until GCS gives a status.

deepikarajani24: Appreciate the review. Let me know if there are any additional questions.

jakeharmon8 (approved):

PR #31284, opened 2025-09-12
[ROCm] Add PJRT_Triton_Extension support

amd-jianli12:
📝 Summary of Changes
This change is PJRT_Triton_Extension support for ROCm as counterpart of that for CUDA. Pallas Triton calls are lowered to HSACO directly rather than PTX on ROCm platform.

PR for JAX is at [jax-ml/jax#31768](https://github.com/jax-ml/jax/pull/31768)

🎯 Justification
There has PJRT_Triton_Extension support for CUDA from [openxla/xla#21483](https://github.com/openxla/xla/pull/21483) and this change is for ROCm.

🚀 Kind of Contribution
✨ New Feature

N/A

🧪 Unit Tests:
PjrtCAPIGpuExtensionTest.TritonCompile

🧪 Execution Tests:
N/A since this extension is adding compilation pipeline lowering Triton IR to HSACO.


xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+}
+
+using HashType = std::array<uint8_t, 32>;
+const std::string tempdir_default = "/tmp";
+
+struct HsacoCache {
+ private:
+  struct Hash64 {
+    size_t operator()(const HashType& s) const noexcept {
+      return *reinterpret_cast<const size_t*>(s.data());
```
beckerhe: This is casting an array with 32 bytes in a 8 bytes size_t? Is that intentional?

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+                     << hsaco_cache_dir_;
+      }
+    }
+  }
+  std::string find(const HashType& hash);
+  std::string emplace(const HashType& hash, std::string& hsaco_path);
+  static HsacoCache& i() {
+    static HsacoCache obj_;
+    return obj_;
+  }
```
beckerhe: Can you avoid making this a singleton? Singletons just mess with hermeticity and we had so many problems with poisened caches.

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+    if (hsaco_cache_dir_ != tempdir_default) {
+      if (!env->DeleteRecursively(hsaco_cache_dir_, &undeleted_files,
+                                  &undeleted_dirs)
+               .ok()) {
+        LOG(WARNING) << "Unable to delete HSACO cache directory "
+                     << hsaco_cache_dir_;
+      }
+    }
+  }
+  std::string find(const HashType& hash);
```
beckerhe: Can you make this function `const`?

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  auto debug_opts = xla::DefaultDebugOptionsIgnoringFlags();
+  auto llvm_opts = xla::gpu::amdgpu::GetAMDGPUBackendOptions(debug_opts);
+  llvm_ir::LLVMCommandLineOptionsLock llvm_lock(llvm_opts);
+
+  raw_sha256_ostream os;
+  llvm::WriteBitcodeToFile(*llvm_module, os);
+  os << arch_name;
+  auto opt_level = debug_opts.xla_backend_optimization_level();
```
beckerhe: Can you use the real type here?

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+    total_bytes_ += size;
+  }
+
+ public:
+  HashType sha256() {
+    flush();
+    return state_.final();
+  }
+  size_t current_pos() const override { return total_bytes_; }
+};
```
beckerhe: What's the purpose of having this type? Are the bitcode buffers so large that we have to stream into the SHA256 function directly? If not I would prefer if we make the BitcodeWriter write into a buffer which we then hash afterwards.

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+
+std::string HsacoCache::emplace(const HashType& hash, std::string& hsaco_path) {
+  auto* env = tsl::Env::Default();
+  absl::string_view hex_string(reinterpret_cast<const char*>(hash.data()),
+                               hash.size());
+  std::string hsaco_name = absl::BytesToHexString(hex_string);
+  std::string new_hsaco_path = tsl::io::JoinPath(hsaco_cache_dir_, hsaco_name);
+  if (env->FileExists(new_hsaco_path).ok()) {
+    new_hsaco_path = hsaco_path;
+    LOG(WARNING) << "HSACO file " << new_hsaco_path << "already exists!";
```
beckerhe: This is a hash collision, right? Why not abort in this case?

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+  if (auto it = hsaco_cache_.find(hash); it != hsaco_cache_.end()) {
+    return it->second;
+  }
+  return std::string();
+}
+
+std::string HsacoCache::emplace(const HashType& hash, std::string& hsaco_path) {
+  auto* env = tsl::Env::Default();
+  absl::string_view hex_string(reinterpret_cast<const char*>(hash.data()),
+                               hash.size());
```
beckerhe: To avoid undefined behaviour, can you change to not using reinterpret_cast? This should work:

```
std::array<char, sizeof(hash)> hex_string;
std::memcpy(hex_string.data(), hash.data(), hex_string.size());
std::string hsaco_name = absl::BytesToHexString(absl::string_view(hex_string.data(), hex_string.size());
```

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,309 @@
...
+                                  num_stages, &cluster_info));
+
+  auto shared_mem_bytes =
+      (*module_op)->getAttrOfType<mlir::IntegerAttr>("ttg.shared").getInt();
+
+  TF_ASSIGN_OR_RETURN(auto hsaco_path,
+                      LLVMToHSACO(*module_op, arch_name, num_warps));
+
+  return CompilationResult{
+      hsaco_path,
```
beckerhe: In the CUDA version we return PTX here, but it appears this is returning a filepath? How is that supposed to work?

xla/pjrt/triton_rocm.cc
amd-jianli12: Yes, to avoid singleton, maybe I can remove caching here and leave this part of work to JAX. Will update.

xla/pjrt/triton_rocm.cc
amd-jianli12: JAX is expecting compiled HSACO file name at [jaxlib/gpu/triton_kernels.cc#L120](https://github.com/jax-ml/jax/blob/main/jaxlib/gpu/triton_kernels.cc#L120) for ROCm. At run time, module binary is directly read from the temporary file.

xla/pjrt/triton_rocm.cc
amd-jianli12: Yes, will update to use real type instead.

xla/pjrt/triton_rocm.cc
amd-jianli12: Definitely. Will remove uses of reinterpret_cast.

xla/pjrt/triton_rocm.cc
amd-jianli12: Will update.

xla/pjrt/triton_rocm.cc
amd-jianli12: Will update.

xla/pjrt/triton_rocm.cc
amd-jianli12: Will update.

xla/pjrt/triton_rocm.cc
amd-jianli12: Will update.

xla/service/gpu/llvm_gpu_backend/amdgpu_backend.h
```diff
@@ -37,6 +37,11 @@ absl::StatusOr<std::vector<uint8_t>> CompileToHsaco(
     llvm::Module* module, stream_executor::GpuComputeCapability gpu_version,
     const DebugOptions& debug_options,
     const std::string& module_config_cache_key);
+// An overload of CompileToHsaco() that simply does compile and returns
+// compiled Hsaco file path.
+absl::StatusOr<std::string> CompileToHsaco(
```
beckerhe: Can you give this function a different name to make it very clear that it returns a filepath?

xla/pjrt/triton_rocm.cc
```diff
@@ -0,0 +1,158 @@
...
+      TritonToLLVM(*module_op, arch_name, num_warps, num_ctas, num_stages));
+
+  int64_t shared_mem_bytes =
+      (*module_op)->getAttrOfType<mlir::IntegerAttr>("ttg.shared").getInt();
+
+  TF_ASSIGN_OR_RETURN(std::string hsaco_path,
+                      LLVMToHSACO(*module_op, arch_name, num_warps));
+
+  // There is no clusters in ROCm for now.
+  return CompilationResult{
```
beckerhe, line 153: The first field in CompilationResult is `asm_text` and not a filepath. Misusing that for a filepath would be very confusing and not an option.

xla/pjrt/triton_rocm.cc
beckerhe: The right solution here is to change the JAX code. Misusing the field that clearly indicates that it contains the assembly as a filepath is not an option.

xla/pjrt/triton_rocm.cc
i-chaochen: considering we don't have ptx and use hsaco in essential, maybe we could change this part 

https://github.com/openxla/xla/blob/d1c168ca8f23d2df0bc0b5feb42e29863d76ae48/xla/pjrt/triton.h#L27-L33

as this following? 
```
struct AsmText { std::string value; };
struct HsacoPath { std::string value; };

struct CompilationResult {
  std::variant<AsmText, HsacoPath> compiled_output;
  int64_t smem_bytes = 0;
  int cluster_dim_x = 0;
  int cluster_dim_y = 0;
  int cluster_dim_z = 0;
};
```

@beckerhe @amd-jianli12  this is what we always doing in XLA code base.

amd-jianli12: Yes and besides, we may need much more refactoring in JAX:
1) C interface between XLA and JAX `PJRT_Triton_Compile_Args` at [xla/pjrt/c/pjrt_c_api_triton_extension.h#L39](https://github.com/openxla/xla/blob/main/xla/pjrt/c/pjrt_c_api_triton_extension.h#L39)
2) Protocol declaration of `CompilationResult` for compile handler at [jax/_src/lib/triton.py#L22](https://github.com/jax-ml/jax/blob/main/jax/_src/lib/triton.py#L22)
3) Binding of `TritonCompilationResult` at [jaxlib/gpu/gpu_plugin_extension.cc#L213](https://github.com/jax-ml/jax/blob/main/jaxlib/gpu/gpu_plugin_extension.cc#L213)
4) Definition of `Kernel` class at [jaxlib/gpu/triton_kernels.h#L58](https://github.com/jax-ml/jax/blob/main/jaxlib/gpu/triton_kernels.h#L58)
5) Handling for CUDA and ROCm at [jax/_src/pallas/triton/pallas_call_registration.py#L169](https://github.com/jax-ml/jax/blob/main/jax/_src/pallas/triton/pallas_call_registration.py#L169)

The refactoring may include splitting implementation for CUDA and ROCm into different files in JAX.

xla/service/gpu/llvm_gpu_backend/amdgpu_backend.h
amd-jianli12: Yes, will update.

xla/pjrt/triton_rocm.cc
beckerhe: > considering we don't have ptx and use hsaco in essential, maybe we could change this part
> 
> https://github.com/openxla/xla/blob/d1c168ca8f23d2df0bc0b5feb42e29863d76ae48/xla/pjrt/triton.h#L27-L33
> 
> as this following?
> 
> ```
> struct AsmText { std::string value; };
> struct HsacoPath { std::string value; };
> 
> struct CompilationResult {
>   std::variant<AsmText, HsacoPath> compiled_output;
>   int64_t smem_bytes = 0;
>   int cluster_dim_x = 0;
>   int cluster_dim_y = 0;
>   int cluster_dim_z = 0;
> };
> ```
> 
> @beckerhe @amd-jianli12 this is what we always doing in XLA code base.

That looks good to me!

amd-jianli12: Hi @beckerhe, I have updated this PR accordingly. Would you kindly take a look?

xla/pjrt/c/pjrt_c_api_triton_extension.h
```diff
@@ -38,6 +38,8 @@ struct PJRT_Triton_Compile_Args {
   int num_stages;
   const char* out_asm;  // owned
   size_t out_asm_size;
+  const char* out_path;  // owned
+  size_t out_path_size;
```
beckerhe, line 42: Can you add these new fields at the end so that we can still support the old API?

(You will also need to update the `last_field` to the PJRT_DEFINE_STRUCT_TRAITS macro below.

xla/pjrt/c/pjrt_c_api_triton_extension.h
beckerhe: Can you bump this version?

xla/pjrt/c/pjrt_c_api_triton_internal.cc
beckerhe, line 35: Can you adjust this check so that the old API is still supported?

It would basically mean: If the struct_size is missing the out_path fields, then at least fill the out_asm field if AsmText was returned and error out otherwise.

xla/pjrt/c/pjrt_c_api_triton_internal.cc
```diff
@@ -37,10 +37,24 @@ PJRT_Error* PJRT_Triton_Compile(PJRT_Triton_Compile_Args* args) {
...
 
-  auto* asm_copy = new char[result.asm_text.size()];
-  std::memcpy(asm_copy, result.asm_text.data(), result.asm_text.size());
-  args->out_asm = asm_copy;
-  args->out_asm_size = result.asm_text.size();
+  args->out_asm = nullptr;
+  args->out_asm_size = 0;
+  if (xla::triton::AsmText* ptr =
+          std::get_if<xla::triton::AsmText>(&result.compiled_output)) {
+    auto* asm_copy = new char[ptr->value.size()];
```
beckerhe: Nit: you can use `args->out_asm` instead of creating a temporary `asm_copy`

xla/pjrt/c/pjrt_c_api_triton_internal.cc
```diff
@@ -37,10 +37,24 @@ PJRT_Error* PJRT_Triton_Compile(PJRT_Triton_Compile_Args* args) {
...
+    std::memcpy(asm_copy, ptr->value.data(), ptr->value.size());
+    args->out_asm = asm_copy;
+    args->out_asm_size = ptr->value.size();
+  }
+  args->out_path = nullptr;
+  args->out_path_size = 0;
+  if (xla::triton::HsacoPath* ptr =
+          std::get_if<xla::triton::HsacoPath>(&result.compiled_output)) {
+    auto* path_copy = new char[ptr->value.size()];
```
beckerhe: Nit: You can allocate directly into `args->out_path` instead of using a temporary `path_copy`.

xla/pjrt/BUILD
```diff
@@ -1150,8 +1162,28 @@ cc_library(
...
+            "@llvm-project//mlir:Transforms",
+            "//xla:debug_options_flags",
+            "//xla/backends/gpu/codegen/triton:compilation_pipeline",
+            "//xla/service/gpu/llvm_gpu_backend:amdgpu_backend",
+            "//xla/service/gpu:target_constants",
+            "//xla/tsl/platform:errors",
+            "//xla/tsl/platform:statusor",
+        ]
     ),
 )
```
beckerhe, line 1288: Can you split this into 3 targets:

- `triton_stub`
- `triton_cuda`
- `triton_rocm`

Each of these targets should be tagged `manual` so that they don't get included into wildcard builds. This has the advantage that it's a lot easier to manage the dependencies for each of them separately. We also have some tooling that can automatically clean the dependencies if they are not hidden behind the `if_xyz_is_configured` macros.

Then you can have a single target `triton`, that conditionally depends on either of the other targets:

```
deps = if_cuda_or_rocm_is_configured([], [":triton_stub"]) + if_rocm_is_configured([":triton_rocm"]) + if_cuda_is_configured([":triton_cuda"])
```

The header `triton.h` should be listed in each of the `triton_{stub|cuda|rocm}` targets, but in the `srcs` attribute. In the `:triton` target `triton.h` should be listed in the `hdrs` attribute. This means that only the `:target` "exports" this headers and our automatic tooling then knows which target to depend on when some source file includes "triton.h".

xla/service/gpu/llvm_gpu_backend/amdgpu_backend.cc
```diff
@@ -571,17 +559,10 @@ absl::StatusOr<std::vector<uint8_t>> CompileToHsaco(
...
-  static absl::once_flag backend_init_flag;
-  // TODO(rocm) Ideally this would be refreshed if xla_gpu_cuda_data_dir
-  // changes.
-  static std::string rocdl_dir_path;  // NOLINT: static/global vars forbidden
-  absl::call_once(backend_init_flag, AMDGPUBackendInit, debug_options,
-                  rocdl_dir_path);
   auto llvm_opts = GetAMDGPUBackendOptions(debug_options);
   llvm_ir::LLVMCommandLineOptionsLock llvm_lock(llvm_opts);
 
   std::vector<uint8_t> hsaco;
```
beckerhe: Can you move this variable declaration down to when it's used for the first time?

beckerhe (commented):
Sorry for the late review. We would need to make the PJRT extension backwards compatible, otherwise LGTM.

xla/pjrt/c/pjrt_c_api_triton_extension.h
amd-jianli12: Yes, will relocate these fields and update.

xla/pjrt/c/pjrt_c_api_triton_extension.h
amd-jianli12: Yes, will update.

xla/pjrt/c/pjrt_c_api_triton_internal.cc
amd-jianli12: Yes, will add checks to support old API.

xla/pjrt/BUILD
amd-jianli12: Yes, will split into 3 targets and update.

xla/service/gpu/llvm_gpu_backend/amdgpu_backend.cc
amd-jianli12: Yes, will update.

amd-jianli12:
> Sorry for the late review. We would need to make the PJRT extension backwards compatible, otherwise LGTM.

Many thanks to @beckerhe for careful review and detailed explanation! I've updated this PR accordingly. Would you kindly have a look?

beckerhe (approved):
Thank you, Jian, for all the changes!

amd-jianli12:
I'm adding `"gpu"` into `tags` list of `triton_cuda` and `triton_rocm` target to fix Bazel Dependency Violations:
```
The following dependency violations were found:
 - @//xla/pjrt:triton_cuda (not tagged gpu) depends on @//xla/backends/gpu/codegen/triton:compilation_pipeline (tagged gpu)
 - @//xla/pjrt:triton_rocm (not tagged gpu) depends on @//xla/backends/gpu/codegen/triton:compilation_pipeline (tagged gpu)


There are a couple of potential solutions for this/these violation(s):

1. Tag the dependent target with the same tag as the dependee.

2. If unavoidable make the dependency selective using the
   'if_{gpu|cuda|rocm}_is_configured' macro. This is discouraged
   outside of stream_executor.
```

xla/pjrt/BUILD
```diff
@@ -1128,12 +1133,100 @@ cc_library(
...
+        "@com_google_absl//absl/status:statusor",
+        "@com_google_absl//absl/strings:string_view",
+        "@llvm-project//llvm:TargetParser",
+        "@tsl//tsl/platform:cuda_root_path",
+        "@tsl//tsl/platform:path",
+        "@com_google_absl//absl/status",
+    ],
+)
+
+cc_library(
```
thomasjoerg, line 1169: This breaks the following test among others: //xla/pjrt/c:pjrt_c_api_gpu_plugin.so 

All fail to build complaining about the //xla/backends/gpu/codegen/triton:compilation_pipeline dependency on Google's internal CI.

I wonder if you need to keep it behind `if_cuda_is_configured`, but am not quite sure.

@ddunl Can you advise maybe?

thomasjoerg (changes requested):

xla/pjrt/BUILD
amd-jianli12: Hi @thomasjoerg, Sorry for the trouble. I also noticed dependency violations with this PR in CI result.

Are you seeing the failure the same as:

> 
> ```
> The following dependency violations were found:
>  - @//xla/pjrt:triton_cuda (not tagged gpu) depends on @//xla/backends/gpu/codegen/triton:compilation_pipeline (tagged gpu)
>  - @//xla/pjrt:triton_rocm (not tagged gpu) depends on @//xla/backends/gpu/codegen/triton:compilation_pipeline (tagged gpu)
> 
> 
> There are a couple of potential solutions for this/these violation(s):
> 
> 1. Tag the dependent target with the same tag as the dependee.
> 
> 2. If unavoidable make the dependency selective using the
>    'if_{gpu|cuda|rocm}_is_configured' macro. This is discouraged
>    outside of stream_executor.
> ```
>

I have update this PR following suggestion 1. Tag the dependent target with the same tag as dependee:

```
diff --git a/xla/pjrt/BUILD b/xla/pjrt/BUILD
index 0a1772859d..dd9a724e5d 100644
--- a/xla/pjrt/BUILD
+++ b/xla/pjrt/BUILD
@@ -1120,7 +1120,10 @@ cc_library(
         "triton_cuda.cc",
         "triton.h",
     ],
-    tags = ["manual"],
+    tags = [
+        "manual",
+        "gpu",
+    ],
     visibility = internal_visibility([":friends"]),
     deps = [
         "//xla/tsl/platform:env",
@@ -1182,7 +1185,10 @@ cc_library(
         "triton_rocm.cc",
         "triton.h",
     ],
-    tags = ["manual"],
+    tags = [
+        "manual",
+        "gpu",
+    ],
     visibility = internal_visibility([":friends"]),
     deps = [
         "//xla/tsl/platform:env",
```

Hopefully this would fix the violation above. Do we still have other different violations?

amd-jianli12: Hi @thomasjoerg, Seems Bazel Dependency Violations are green with this PR now [3417434](https://github.com/openxla/xla/commit/3417434fbb982c230c5cb77d944b5422b9ce5592). Do you think it's good to have it merged?

amd-jianli12:
Rebased to the latest origin/main@f453a33.

amd-jianli12:
Rebased to the latest origin/main@9a941c5.

xla/pjrt/BUILD
thomasjoerg: I'm going to approve this PR again to import and test it against Google's CI. I'll report back.

thomasjoerg (approved):

amd-jianli12:
Rebased to the latest origin/main@6fd111d.

xla/pjrt/BUILD
thomasjoerg: No luck, still failing with the same error message.

amd-jianli12:
Rebased to the latest origin/main@74eea96.

hsharsha:
@beckerhe @thomasjoerg  Can you please approve latest change after rebase?

i-chaochen:
Hi @dimitar-asenov this PR seems stuck for a long while...wondering do you know the status? shall we rebase it again to resolve the conflict? or you can merge now since it's approved couple of times...

beckerhe (approved):

dimitar-asenov:
@i-chaochen Sorry for the delay on this. The PR is now approved, but it seems you need to rebase. Please rebase and I will make sure we merge it soon.

dimitar-asenov:
@amd-jianli12 Could you please rebase the PR and I will merge it. Thanks!

amd-jianli12:
> @amd-jianli12 Could you please rebase the PR and I will merge it. Thanks!

Hi @dimitar-asenov I have rebased this PR against the latest main. Would you help take a look? Thanks!

dimitar-asenov (approved):

dimitar-asenov:
@amd-jianli12 : Before we can merge this PR, we need a bit more info on the JAX side. I left you this comment https://github.com/jax-ml/jax/pull/31768#issuecomment-3743564314. Please take a look.

amd-jianli12:
> @amd-jianli12 : Before we can merge this PR, we need a bit more info on the JAX side. I left you this comment [jax-ml/jax#31768 (comment)](https://github.com/jax-ml/jax/pull/31768#issuecomment-3743564314). Please take a look.

Hi @dimitar-asenov, I have updated PR on JAX side accordingly. Besides, I also noticed CI on XLA side is complaining:
```
The following dependency violations were found:
 - @//xla/pjrt:triton_rocm (not tagged rocm-only) depends on @//xla/service/gpu/llvm_gpu_backend:amdgpu_backend (tagged rocm-only)


There are a couple of potential solutions for this/these violation(s):

1. Tag the dependent target with the same tag as the dependee.

2. If unavoidable make the dependency selective using the
   'if_{gpu|cuda|rocm}_is_configured' macro. This is discouraged
   outside of stream_executor.
```
So I tagged //xla/pjrt:triton_rocm with rocm-only to fix this violation in 1450606. Please kindly have a look. Thanks!

i-chaochen:
Hi @dimitar-asenov @beckerhe wondering do we still have unresolved issues on this PR? IIUC, we need to merge this in XLA side first and then jax side? https://github.com/jax-ml/jax/pull/31768 which seems is approved as well.

i-chaochen (approved):

dimitar-asenov:
With the latest changes to the JAX version and it getting reapproved, I think we should be able to make the necessary internal changes that will allow us to merge this PR. I hope to do this early next week, this week is very busy.

PR #31292, opened 2025-09-12
🐛 Bug Fix: propagate memory_space 

gwenzek:
🐛 Bug Fix

📝 Summary of Changes
Propagate memory space in `PJRT_Client_CreateUninitializedBuffer`

🎯 Justification
Currently `PJRT_Client_CreateUninitializedBuffer` doesn't honor the memory space argument it received.
`CreateUninitializedBuffer` is not passing memory_space to `AllocateDestinationBuffer`.
The argument then defaults to null, which is interpreted as the default memory space.


penpornk (changes requested):
Sorry for the delay! 

Both the `CreateUninitializedBuffer` and `BufferFromHostLiteral` functions have been removed from this file. Could you please sync to head? Is this PR still needed?

gwenzek:
It looks like the refactor that removed GPU specific `CreateUninitializedBuffer` and `BufferFromHostLiteral` in favor of `CommonPjRtClient` implementations fixed the bug. Thanks for checking.

PR #31300, opened 2025-09-12
[XLA:CPU][BugFix] Fix ODR Violation by Renaming Duplicate IsSupportedType Definition

othakkar:
Two different definitions of `xla::cpu::IsSupportedType(...)` in `onednn_util.h` and `convolution_lib.cc` cause ODR (One Definition Rule) violation. This occurs when modifying the `IsSupportedType(...)` function in `onednn_util.h` (e.g., adding logs), which causes the compiler not to inline the function and results in undefined behavior.

This PR fixes the issue by renaming the `IsSupportedType(...)` function in `convolution_lib.cc` (as changing the name in `onednn_util.h` will need changes in its multiple invocations).

### DIFF
--- a/xla/backends/cpu/runtime/convolution_lib.cc
+++ b/xla/backends/cpu/runtime/convolution_lib.cc
@@ -93,7 +93,7 @@ static absl::Status ValidateConvolutionShapes(
   return absl::OkStatus();
 }
 
-bool IsSupportedType(PrimitiveType primitive_type) {
+bool IsConvSupportedType(PrimitiveType primitive_type) {
   return primitive_type == PrimitiveType::F16 ||
          primitive_type == PrimitiveType::F32;
 }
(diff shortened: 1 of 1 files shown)


xla/backends/cpu/runtime/convolution_lib.cc
```diff
@@ -93,7 +93,7 @@ static absl::Status ValidateConvolutionShapes(
   return absl::OkStatus();
 }
 
-bool IsSupportedType(PrimitiveType primitive_type) {
+bool IsConvSupportedType(PrimitiveType primitive_type) {
```
penpornk: Nit: We can just make this an inline function.

penpornk (changes requested):
Thank you for the PR!

xla/backends/cpu/runtime/convolution_lib.cc
othakkar: Sure, done (https://github.com/openxla/xla/pull/31300/commits/2e2a5159f92d1a474ffb9d5b64ec91af34b0db0d).

penpornk (approved):
Thank you!

PR #31375, opened 2025-09-15
[XLA:GPU] Add NVLink domain check to CollectiveBackendAssigner

sfvaroglu:
📝 Summary of Changes
This PR updates the CollectiveBackendAssigner pass to account for NVLink domain connectivity when deciding between NVSHMEM and DEFAULT backends. It does this by adding a slice_size parameter to the compilation pipeline and introducing an IsIntraNVLinkDomain check.  

🎯 Justification
The CollectiveBackendAssigner now uses NVSHMEM not only for single-host scenarios, but also when all devices are within the same NVLink domain. 

🚀 Kind of Contribution
⚡️ Performance Improvement, 🧪 Tests

H100
|  | NVSHMEM enabled | NVSHMEM disabled |
|----------|----------|----------|
| llama31_8b_fp8_1x8    | 1095330 us   | 1093816 us    |
| llama31_8b_bf16_2x8    | 1368948 us   | 1370896 us   |
| llama31_8b_fp8_2x8    | 1096447 us   | 1092437 us   |
| llama31_70b_fp8_16x8    | 9723821 us   | 9707544 us    |

🧪 Unit Tests:
Added unit tests to xla/service/gpu/transforms/collectives/collective_backend_assigner_test.cc

🧪 Execution Tests:
Tested with llama3-8b on 2 GB200 nodes (fsdp = 8). The average step time in NVSHMEM case was 3.69s (vs. 3.76s in the default case).


frgossen:
> N/A for NVSHMEM
Can you provide some measurements here? We will not track this on our end but I think there would still be value in quantifying the impact of this PR.

frgossen (changes requested):
> N/A for NVSHMEM

Can you provide some measurements here? We will not track this on our end but I think there would still be value in quantifying the impact of this PR.

sfvaroglu:
> > N/A for NVSHMEM
> 
> Can you provide some measurements here? We will not track this on our end but I think there would still be value in quantifying the impact of this PR.

@frgossen Added HLO measurements without NVSHMEM, as NVSHMEM requires a one-process-per-device setup, which is not currently supported by our existing tools.

frgossen (changes requested):
> @frgossen Added HLO measurements without NVSHMEM, as NVSHMEM requires a one-process-per-device setup, which is not currently supported by our existing tools.

What speedups do you see on your end? What is the impact of this PR?

sfvaroglu:
> Tested with llama3-8b on 2 GB200 nodes (fsdp = 8). The average step time in NVSHMEM case was 3.69s (vs. 3.76s in the default case).

> > @frgossen Added HLO measurements without NVSHMEM, as NVSHMEM requires a one-process-per-device setup, which is not currently supported by our existing tools.
> 
> What speedups do you see on your end? What is the impact of this PR?

Is this sufficient? "Tested with llama3-8b on 2 GB200 nodes (fsdp = 8). The average step time in NVSHMEM case was 3.69s (vs. 3.76s in the default case)."

frgossen:
> > Tested with llama3-8b on 2 GB200 nodes (fsdp = 8). The average step time in NVSHMEM case was 3.69s (vs. 3.76s in the default case).
> 
> > > @frgossen Added HLO measurements without NVSHMEM, as NVSHMEM requires a one-process-per-device setup, which is not currently supported by our existing tools.
> > 
> > 
> > What speedups do you see on your end? What is the impact of this PR?
> 
> Is this sufficient? "Tested with llama3-8b on 2 GB200 nodes (fsdp = 8). The average step time in NVSHMEM case was 3.69s (vs. 3.76s in the default case)."

Ah, this slipped me. If it's not too much work it'd be great to measure the speedup for all the benchmarks, also to make sure there are no regressions.

sfvaroglu:
> > > Tested with llama3-8b on 2 GB200 nodes (fsdp = 8). The average step time in NVSHMEM case was 3.69s (vs. 3.76s in the default case).
> > 
> > 
> > > > @frgossen Added HLO measurements without NVSHMEM, as NVSHMEM requires a one-process-per-device setup, which is not currently supported by our existing tools.
> > > 
> > > 
> > > What speedups do you see on your end? What is the impact of this PR?
> > 
> > 
> > Is this sufficient? "Tested with llama3-8b on 2 GB200 nodes (fsdp = 8). The average step time in NVSHMEM case was 3.69s (vs. 3.76s in the default case)."
> 
> Ah, this slipped me. If it's not too much work it'd be great to measure the speedup for all the benchmarks, also to make sure there are no regressions.

@frgossen I ran a few benchmarks with NVSHMEM enabled and disabled, but I don’t think the results add much value. NVSHMEM heuristics mainly apply to small-size AllReduce and CollectivePermute operations, and I believe these models don’t include such cases that would significantly impact performance.

xla/service/gpu/transforms/collectives/collective_backend_assigner_test.cc
```diff
@@ -70,7 +72,9 @@ TEST_F(CollectiveBackendAssignerTest, SmallAllReduceUsesNvshmem) {
   )";
 
   TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(kHloText));
-  EXPECT_THAT(RunCollectiveBackendAssigner(module.get()),
+
+  EXPECT_THAT(RunCollectiveBackendAssigner(
+                  module.get(), /*num_devices_per_host=*/1, /*slice_size=*/0),
```
felixwqp, line 77: can you help me understand why `/*num_devices_per_host=*/1` would have nvshmem enabled? 

I think nvshmem is only available for multi-device platform, right?

pifon2a (approved):

xla/service/gpu/transforms/collectives/collective_backend_assigner_test.cc
sfvaroglu: > can you help me understand why `/*num_devices_per_host=*/1` would have nvshmem enabled?
> 
> I think nvshmem is only available for multi-device platform, right?

It's 1 process per device.

PR #31477, opened 2025-09-18
Support building with Bzlmod

meteorcloudy:
Context: https://bazel.build/external/migration

This PR adds support for building xla with Bzlmod. `bazel test --config=bzlmod //xla/tests:all` passes most of the test cases except for one.

- Introduced common bazel dependencies as bazel modules
- Introduced some C++ dependencies with module extensions, which can later be converted to Bazel modules ideally.
- All extensions added located under `third_party/extensions`, which mainly reuses repo definitions in the WORKSPACE build.


meteorcloudy:
Still working on the python toolchains

MODULE.bazel
```diff
@@ -0,0 +1,172 @@
...
+    # TODO: what about other platforms?
+    "python_3_11_x86_64-unknown-linux-gnu",
+)
+register_toolchains("@python_3_11_toolchains//:all")
+
+python_init_pip = use_extension("//third_party/extensions/python:python_init_pip.bzl", "python_init_pip_ext")
+use_repo(python_init_pip, "pypi")
+
+install_deps = use_extension("//third_party/extensions/python:install_deps.bzl", "install_deps_ext")
+use_repo(install_deps, "pypi_numpy", "pypi_lit")
```
rickeylev, line 138: What necessitated having to do this instead of using pip.parse directly?

IIRC, the only customization was to add some "annotations" (patches). I think all that's needed is to use pip.whl_mods: https://rules-python.readthedocs.io/en/latest/howto/pypi-headers.html#bzlmod-setup

MODULE.bazel
```diff
@@ -0,0 +1,172 @@
...
+
+python_init_toolchains = use_extension("//third_party/extensions/python:python_init_toolchains.bzl", "python_init_toolchains_ext")
+use_repo(python_init_toolchains,
+    "python_3_11",
+    "python_3_11_host",
+    "python_3_11_toolchains",
+    # TODO: what about other platforms?
+    "python_3_11_x86_64-unknown-linux-gnu",
+)
+register_toolchains("@python_3_11_toolchains//:all")
```
rickeylev, line 132: It'd be best if we could use python.toolchains() instead of calling the workspace macros.

What is using the `python_3_11` and `python_3_11_<platform>` and `python_3_11_host` repos? I can see the register_toolchains() call.

Looking at https://github.com/openxla/xla/blob/main/third_party/py/python_repo.bzl

I don't see anything custom it's doing to the runtimes and toolchains themselves. i.e. `python.toolchain(python_version="3.11"); use_repo(python, "python_3_11_host")` should be functionally equivalent.

The pieces of logic I see it doing are:

Transforming the requirements files. This looks like it could move into its own repo rule, something like:

```
requirements = use_extension(...)
requirements.override(
  local_wheels={...},
  requirements = "//:requirements_xxx.txt"
)
use_repo(requirements, "requirements")
pip.parse(requirements = "@requirements//:requirements_xxx.txt", ...)
```

-----

In any case, ultimately, I don't think putting the versioned repo names into MODULE.bazel is going to be feasible. They support multiple python versions, and IIRC, they basically operate in a "one python version for the whole build" type of design, so have a single `@python` repo (or named something like that) for whatever was selected by the HERMETIC_PYTHON_FOO env variables (note that the `HERMETIC_PYTHON_*` vars are an XLA-ism, not a rules_python-ism).

rickeylev (commented):
I know this is currently draft, but wanted to point some things out / questions.

Quick brain dump:
docs: https://rules-python.readthedocs.io/en/latest/api/rules_python/python/extensions/python.html#python

features of note:

python.defaults: set python version from env var or file.

python.single_version_override and python.single_version_platform_override -- this basically allow specifying an arbitrary URL.

I think those API might be lacking a bit, though, since they want literal strings, but so much of this is driven by environment variable. I'm not quite sure how much of this environment variable driven design in XLA is appropriate as general features, though -- it strongly smells of the problems workspace has that bzlmod generally doesn't.

meteorcloudy:
@rickeylev Thanks for talking a look! I have another PR at https://github.com/openxla/xla/pull/31718 with the native rules_python toolchains, now I can build xla with the toolchains, but I still need to figure out how to respect the HERMETIC_PYTHON_FOO env variables (or whether we have to? @vam-google ).

meteorcloudy:
Closing in favor of https://github.com/openxla/xla/pull/32055

Issue #9010, opened 2024-01-31
Proto config fails for specific "round" decimal numbers.


jmaksymc:
In XLA:CPU backend_config.proto using decimal values as float and double causes issues. Specific round decimal values are retrieved from the config as 0 and seem to break further values. Example:

Proto file:
```
float sample_float_one = 6;
float sample_float_two = 7;
float sample_float_three = 8;
float sample_float_four = 9;
```
Input values:
```
backend_config->mutable_onednn_matmul_config()->set_sample_float_one(0.49);
backend_config->mutable_onednn_matmul_config()->set_sample_float_two(0.51);
backend_config->mutable_onednn_matmul_config()->set_sample_float_three(0.50);
backend_config->mutable_onednn_matmul_config()->set_sample_float_four(0.53);
```
Retrieving values from the proto config:
```
float sample_float1 = matmul_config.sample_float_one();
float sample_float2 = matmul_config.sample_float_two();
float sample_float3 = matmul_config.sample_float_three();
float sample_float4 = matmul_config.sample_float_four();
```
After retrieving these values:
```
sample_float1 = 0.49
sample_float2 = 0.51
sample_float3 = 0
sample_float4 = 0
```
I have not tested it broadly but it seems to be failing for specific round numbers such as 0.5 or 0.125 which is used in the test for MatMul and Linear fusion. (https://github.com/openxla/xla/pull/8583) (In this fusion a workaround by casting decimal value to int32 has been used)

cheshire:
Sounds really confusing. Could you add a unit test demonstrating this behavior?

kredd2506:
Hi, I investigated this and it appears this issue was resolved by PRs #31511 and #32800 (both merged via Copybara in Oct-Nov 2025). The `alpha_typecast` int32 workaround was removed after switching to the thunk-based execution model, which bypasses the JSON serialization path that caused the protobuf float parsing issue. The native `float alpha` field is now used directly. Can this issue be closed?

penumallas1261:
Hi @jmaksymc ,

Closing the issue because we believe the question has been answered, Please feel free to reopen the issue if you still have a concern. Thanks!

PR #31511, opened 2025-09-18
[XLA:CPU][oneDNN] Enable oneDNN Convolution Custom Calls in Thunk Runtime

othakkar:
This PR enables support for oneDNN Convolution operations in the XLA:CPU Thunk runtime, building upon the foundational implementation of `OneDnnOpThunk`.

Key changes:
- Updated thunk emitter to emit `OneDnnOpThunk` for oneDNN convolution during compilation.
- Enabled oneDNN custom call rewrite for Convolution in `onednn_contraction_rewriter.cc`. 
- Added support for oneDNN Convolution op via `ExecuteOneDnnConvolution(...)` in `onednn_convolution.cc`.


othakkar:
@penpornk, please review this PR when you have a chance. Thanks!

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
```diff
@@ -106,6 +106,124 @@ TEST(OneDnnOpThunkTest, SimpleOneDnnMatMulThunk) {
   EXPECT_EQ(out_literal, expected);
 }
 
+// A very small 2D NHWC convolution test exercising the oneDNN convolution path
+// through OneDnnOpThunk. We build a 1x1 kernel with weight=2 so the expected
```
penpornk: Nit: Can we do a 2x2 kernel instead? 1x1 seems too simple.

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
```diff
@@ -106,6 +106,124 @@ TEST(OneDnnOpThunkTest, SimpleOneDnnMatMulThunk) {
...
+      Array4D<float>(1, 2, 2, 1, {0.f, 0.f, 0.f, 0.f}));
+
+  // Expected = input * 2
+  Literal expected_literal = LiteralUtil::CreateR4FromArray4D<float>(
+      Array4D<float>(1, 2, 2, 1, {2.f, 4.f, 6.f, 8.f}));
+
+  // Buffer allocations
+  auto input_alloc = CreateBufferAllocation(0, input_literal);
+  auto weight_alloc = CreateBufferAllocation(1, weight_literal);
+  auto output_alloc = CreateBufferAllocation(2, output_literal);
```
penpornk: Nit: We can use the more-compact form:
```suggestion
  auto [input_alloc, weight_alloc, output_alloc] =
      CreateBufferAllocation(input_literal, weight_literal, output_literal);
```

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
```diff
@@ -106,6 +106,124 @@ TEST(OneDnnOpThunkTest, SimpleOneDnnMatMulThunk) {
...
+
+  auto input_slice = CreateBufferAllocationSlice(input_alloc);
+  auto weight_slice = CreateBufferAllocationSlice(weight_alloc);
+  auto output_slice = CreateBufferAllocationSlice(output_alloc);
```
penpornk: Same here. We can use the more-compact form:
```suggestion
auto [input_slice, weight_slice, output_slice] =
      CreateBufferAllocationSlice(input_alloc, weight_alloc, output_alloc);
```

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
```diff
@@ -106,6 +106,124 @@ TEST(OneDnnOpThunkTest, SimpleOneDnnMatMulThunk) {
...
+
+  BufferAllocations allocations =
+      CreateBufferAllocations(input_literal, weight_literal, output_literal);
+
+  // Build a minimal OneDnnConvolutionConfig proto.
+  OneDnnConvolutionConfig conv_config;
+  conv_config.set_dims(4);  // rank
+  // Input layout NHWC: batch=0, feature=3, spatial (H,W) = (1,2) zero-based.
+  auto* inp = conv_config.mutable_input();
```
penpornk: Nit: Please explicitly specify types of each variable when possible, so new readers can easily see the type of each variable.

```suggestion
  OneDnnTensorLayoutProto* inp = conv_config.mutable_input();
```

Reference: [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Type_deduction)
```
Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type.
```

Example exceptions: When specifying the type is redundant.
The following line from `onednn_op_thunk` is fine because the type was already in `std::get`.
```
const auto& conv_config = std::get<OneDnnConvolutionConfig>(config);
```

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
```diff
@@ -106,6 +106,124 @@ TEST(OneDnnOpThunkTest, SimpleOneDnnMatMulThunk) {
...
+
+  // Kernel layout assumed HWIO (H,W,In,Out):
+  auto* ker = conv_config.mutable_kernel();
+  ker->set_dims(4);
+  auto* filter = ker->mutable_filter();
+  filter->set_input_feature_dim(2);   // zero-based index of IC
+  filter->set_output_feature_dim(3);  // zero-based index of OC
+  // Spatial dims (H,W) one-based: (0->1,1->2) => 1,2
+  filter->add_spatial_dims(1);
+  filter->add_spatial_dims(2);
```
penpornk, line 191: We do we use mixed-bases (zero-based and one-based) here? Is it a constraint from oneDNN? Would you mind pasting a link for the documentation on this in the comment in the code?

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
```diff
@@ -106,6 +106,124 @@ TEST(OneDnnOpThunkTest, SimpleOneDnnMatMulThunk) {
...
+  win->add_strides(2);
+  win->add_strides(2);
+  // Pads store (actual +1) so 1 -> 0 actual pad.
+  win->add_pad_left(1);
+  win->add_pad_left(1);
+  win->add_pad_right(1);
+  win->add_pad_right(1);
+  // Dilations store (actual +2) so 3 -> 1 actual dilation.
+  win->add_window_dilations(3);
+  win->add_window_dilations(3);
```
penpornk: Are `actual + 1` and `actual + 2` what oneDNN expects too? Would you mind putting a link to oneDNN documentation on this in the comment in the code?

xla/service/cpu/onednn_convolution.cc
```diff
@@ -208,6 +209,147 @@ CreateOneDnnPrimDesc<dnnl::convolution_forward::primitive_desc>(
       any_res_md, strides, rhs_dilations, pad_left, pad_right, attrs);
 }
 
+void ExecuteOneDnnConvolution(absl::Span<MemrefInfoHandler> arguments,
```
penpornk, line 217: Can we just make `__xla_cpu_runtime_OneDnnConvolution` call this function to avoid having redundant code? Or should we just delete `__xla_cpu_runtime_OneDnnConvolution` since there is no way (flag) to run the old runtime anymore?

penpornk (changes requested):

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
othakkar: Done (https://github.com/openxla/xla/pull/31511/commits/c103dd964d6a59319c30c9aabb4345dfa4a882c7).

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
othakkar: Done (https://github.com/openxla/xla/pull/31511/commits/c103dd964d6a59319c30c9aabb4345dfa4a882c7).

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
othakkar: Done (https://github.com/openxla/xla/pull/31511/commits/c103dd964d6a59319c30c9aabb4345dfa4a882c7).

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
othakkar: Thanks for sharing this. Done (https://github.com/openxla/xla/pull/31511/commits/c103dd964d6a59319c30c9aabb4345dfa4a882c7).

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
othakkar: This encoding is not a oneDNN requirement, but an internal encoding defined in [onednn_contraction_rewriter.cc](https://github.com/openxla/xla/blob/main/xla/service/cpu/onednn_contraction_rewriter.cc#L681). This was done to reserve `0` to denote "unset" (for example, zero padding).
Updated the comments to address this better in https://github.com/openxla/xla/blob/main/xla/service/cpu/onednn_contraction_rewriter.cc#L681.

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
othakkar: This encoding is not a oneDNN requirement, but an internal encoding defined in [onednn_contraction_rewriter.cc](https://github.com/openxla/xla/blob/main/xla/service/cpu/onednn_contraction_rewriter.cc#L681). This was done to reserve `0` to denote "unset" (for example, zero padding).
Updated the comments to address this better in https://github.com/openxla/xla/blob/main/xla/service/cpu/onednn_contraction_rewriter.cc#L681.

xla/service/cpu/onednn_convolution.cc
othakkar: I plan to submit a cleanup PR for all four oneDNN primitives to remove all dead code related to the legacy runtime, later, once the thunk runtime changes are merged. This is to avoid making this (and other thunk) PRs longer than needed. Let me know if that is okay.

penpornk: I don't think the PR will appear too much longer, because most of the code is still the same. The diff should just look like we change the signature of `__xla_cpu_runtime_OneDnnConvolution` and rename some of the variables there. Could you please try with this PR? If it actually looks hard to read, we don't need to do it with the LayerNorm and Softmax thunks.

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
penpornk: Thank you for the explanation!

I think we can check for unset values by checking the size of each array, e.g., if `pad_left` is empty, etc. For `spatial_dims`, what does it mean to have an unset value (0) in the middle of dimension sizes? We should look into cleaning this up later.

xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -554,10 +554,6 @@ bool OneDnnContractionRewriter::ShouldRewriteDot(
 
 bool OneDnnContractionRewriter::ShouldRewriteConv(
     const HloInstruction* conv_instr) {
-  // TODO(intel-tf): remove this restriction after enabling oneDNN convolution
-  // support in thunk runtime.
-  return false;
-
   // NOLINTBEGIN(clang-diagnostic-unreachable-code)
```
penpornk: Please remove this line and its matching `NOLINTEND` line.

penpornk (changes requested):

xla/service/cpu/onednn_contraction_rewriter.cc
othakkar: Done (https://github.com/openxla/xla/pull/31511/commits/9a4ba20c1624fb3d51e3a43cb805e1f4ce7dede7)

othakkar:
@penpornk I added a commit (https://github.com/openxla/xla/pull/31511/commits/1bc5c7364e81e4e867d4763ae57eb1c15907d01e) to remove the code related to convolution for the legacy runtime. Please let me know if this looks okay. I addressed your other comments. Thanks again for the review!

xla/backends/cpu/runtime/onednn/onednn_op_thunk_test.cc
akhilgoe: Hi @penpornk, I don't know if you remember but we encountered an issue in the past where storing a "0" in a repeated field occasionally caused misinterpretation of the serialized buffers. Something similar to https://github.com/openxla/xla/issues/9010. That is the reason why we switched the subspace of such affected repeated fields from whole numbers to natural numbers.
Now that we use thunks, we can reevaluate if such a transformation is still necessary and do appropriate cleanup.

penpornk: @akhilgoe You're right. We have recently done a major protobuf version update -- from 3.21.9 to 5.28.3 in June (https://github.com/openxla/xla/commit/1fc2a846d259d3c896a6177485abcc68d97627a9), and to v6.31.1 in August (https://github.com/openxla/xla/pull/29658/commits/7abd4055b1629815d53dff1647a3ec15b7a79dc6). So let's test if the issue was gone with the upgrade or if it still persists.

penpornk (approved):
Thank you again for the PR! Please rebase because this PR conflicts with https://github.com/openxla/xla/commit/968b761b26d600620be5f4458aa3376cb6210295 that recently went in.

othakkar:
@penpornk, thanks for approving the PR! I rebased the PR with `main`. (https://github.com/openxla/xla/pull/31511/commits/0b32837218aa6dec8ce7efe965fa4feefe81ac9e)

penpornk (approved):
Thank you again!

allanrenucci:
Can you rebase this PR? Thanks

othakkar:
@allanrenucci Done!

allanrenucci (approved):

PR #31525, opened 2025-09-19
[GPU] Optimize collectives on sub-byte types.

sergachev:
The new pass replaces suitable collectives on packed sub-byte types with equivalent ones on the same data casted to S8, which eliminates unnecessary conversions and reduces the volume of moved data.


xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  HloInstruction* input = hlo->mutable_operand(0);
+  const Shape& input_shape = input->shape();
+  const PrimitiveType input_type = input_shape.element_type();
+  constexpr PrimitiveType casted_type = S8;
+  const int64_t ratio = primitive_util::BitWidth(casted_type) /
+                        primitive_util::BitWidth(input_type);
+
+  if (!(primitive_util::IsSubByteNonPredType(input_type) &&
```
PatriosTheGreat: Could you please revert this condition?

`if (!primitive_util::IsSubByteNonPredType(input_type) || !ShapeUtil::LastDimIsMinorMost(input_shape)) || ...`

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization_test.cc
```diff
@@ -0,0 +1,150 @@
...
+#include <gtest/gtest.h>
+#include "xla/hlo/testlib/hlo_hardware_independent_test_base.h"
+
+namespace xla {
+
+class SubByteCollectiveNormalizationTest
+    : public HloHardwareIndependentTestBase {};
+
+TEST_F(SubByteCollectiveNormalizationTest, SkipNonSubByteTypes) {
+  EXPECT_TRUE(RunAndCheckHloRewrite(R"(
```
olegshyshkov: nit: `TF_ASSERT_OK` is also common in our codebase. Here and below.

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization_test.cc
```diff
@@ -0,0 +1,150 @@
...
+e {
+ a = s8[4,8]{1,0} parameter(0)
+ b = s8[8,8]{1,0} all-gather(a), dimensions={0}
+})",
+                                    std::move(SubByteCollectiveNormalization{}),
```
olegshyshkov: move of temp object in unnecessary. Also functions argument in const&.

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
```diff
@@ -0,0 +1,180 @@
...
+  absl::Status DefaultAction(HloInstruction*) override;
+  bool changed() const { return changed_; }
+
+ private:
+  bool changed_ = false;
+};
+
+absl::Status SubByteCollectiveNormalizationVisitor::DefaultAction(
+    HloInstruction* hlo) {
+  if (!HloPredicateIsOp<HloOpcode::kAllGather, HloOpcode::kAllToAll,
```
olegshyshkov: Maybe it would be better to move all the logic below into a helper method and implement separate handler method in the visitor.

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
```diff
@@ -0,0 +1,180 @@
...
+                        HloOpcode::kCollectiveBroadcast,
+                        HloOpcode::kCollectivePermute>(hlo) ||
```
olegshyshkov: Maybe a taste thing, but it might be better to split the if into two, because the conditions are independent.

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
olegshyshkov: Could you move it into a helper with a readable name?

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
```diff
@@ -0,0 +1,180 @@
...
+            primitive_util::BitWidth(input_shape.element_type()) &&
+        input_shape.dimensions().back() % ratio == 0)) {
+    return absl::OkStatus();
+  }
+  if (hlo->opcode() == HloOpcode::kAllToAll) {
+    const auto* all_to_all = Cast<HloAllToAllInstruction>(hlo);
+    if (all_to_all->split_dimension()) {
+      TF_ASSIGN_OR_RETURN(
+          CollectiveOpGroupMode group_mode,
+          GetCollectiveOpGroupMode(all_to_all->channel_id().has_value(),
```
olegshyshkov: There's a another versions of the function that accept just an instruction. `GetCollectiveOpGroupMode(all_to_all)` is enough.

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
```diff
@@ -0,0 +1,180 @@
...
+    new_collective_shape.mutable_layout()->set_element_size_in_bits(0);
+  }
+  HloInstruction* new_collective =
+      hlo->parent()->AddInstruction(hlo->CloneWithNewOperands(
+          new_collective_shape,
+          {ReshapeAndCastToWiderType(input, casted_type)}));
+  TF_RETURN_IF_ERROR(hlo->parent()->ReplaceInstructionWithDifferentShape(
+      hlo, CastToNarrowerTypeAndReshape(new_collective, hlo->shape())));
+
+  changed_ = true;
```
olegshyshkov: `MarkAsChanged()`

PatriosTheGreat (changes requested):

olegshyshkov:
A general question, what happens right now? Where do we get the conversions and increased data transfer size? Is it not just a bitcast?

olegshyshkov (changes requested):

sergachev:
> A general question, what happens right now? Where do we get the conversions and increased data transfer size? Is it not just a bitcast?

Currently float normalization inserts converts around to a larger type and back. Here we are adding bitcast-converts, not directly bitcasts for generality - some of them are not no-ops. Ones that are indeed no-ops are further replaced with bitcasts in the algebraic simplifier: https://github.com/openxla/xla/pull/30704 .

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
sergachev: Not sure I understand the idea yet.

frgossen (changes requested):
How can we measure this speedup? Can you measure this speedup on one of the benchmarks in google3/third_party/tensorflow/compiler/xla/tools/benchmarks/hlo/ ?

sergachev:
> How can we measure this speedup? Can you measure this speedup on one of the benchmarks in google3/third_party/tensorflow/compiler/xla/tools/benchmarks/hlo/ ?

On a microbenchmark probably. I don't see any sub-byte types in those HLOs.

felixwqp:
What's the motivation of this normalization? XLA HLO supports s4, but collective backend didn't support(e.g. NCCL)?

sergachev:
> What's the motivation of this normalization? XLA HLO supports s4, but collective backend didn't support(e.g. NCCL)?

Collectives on sub-byte types are currently implemented as conversion to a larger type -> collective on that type -> conversion back, which is at least 2 extra kernels + more bytes to move with the collective. Also see https://github.com/openxla/xla/pull/31525#issue-3433358985 and https://github.com/openxla/xla/pull/31525#issuecomment-3312348295.

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
olegshyshkov: Instead of implementing `DefaultAction` and then exiting conditionally, I suggest to implement `HandleAllGather`, `HandleAllReduce` and other methods. Since the implementation will be the same, common logic can be moved to a shared helper.

olegshyshkov:
> > What's the motivation of this normalization? XLA HLO supports s4, but collective backend didn't support(e.g. NCCL)?
> 
> Collectives on sub-byte types are currently implemented as conversion to a larger type -> collective on that type -> conversion back, which is at least 2 extra kernels + more bytes to move with the collective. Also see [#31525 (comment)](https://github.com/openxla/xla/pull/31525#issue-3433358985) and [#31525 (comment)](https://github.com/openxla/xla/pull/31525#issuecomment-3312348295).

I see the value in this change. Converting to a larger type just to do a collective operation is wasteful. I'm generally supportive of this change, but want to challenge it slight to understand it's the best place for the solution. Is it easier to do an HLO rewrite than adding support to runtime thunks? `AllGather` and `AllToAll` are just data transfers, so NCCL should not really care about the data type in the buffer, only the amount of data to be transferred.

sergachev:
> I see the value in this change. Converting to a larger type just to do a collective operation is wasteful. I'm generally supportive of this change, but want to challenge it slight to understand it's the best place for the solution. Is it easier to do an HLO rewrite than adding support to runtime thunks? `AllGather` and `AllToAll` are just data transfers, so NCCL should not really care about the data type in the buffer, only the amount of data to be transferred.

I considered the way you described too. Not every all-gather on a sub-byte type can be directly replaced with one on a larger type. For example, if the concatenated dimension has size 1, an optimal way is to use a transpose after an all-gather on a larger type (but still not with a convert like XLA does currently). The current approach with a simple HLO rewrite can handle those cases too via composition with other existing passes and expresses the intended optimization clearly in HLO.

sergachev:
Microbenchmark:

```
e {
  a = s4[1024,512]{1,0:E(4)} parameter(0)
  b = s4[8192,512]{1,0:E(4)} all-gather(a), dimensions={0}
}
```

```
bazel run xla/tools/multihost_hlo_runner/hlo_runner_main -- --hlo_argument_mode=uninitialized --output_mode=not_return_outputs --profile_execution --num_repeats=10 x.hlo
```

Before:
```
## Execution time, file=x.hlo repeat=1 duration=50432ns
## Execution time, file=x.hlo repeat=2 duration=50368ns
## Execution time, file=x.hlo repeat=3 duration=50816ns
## Execution time, file=x.hlo repeat=4 duration=49952ns
## Execution time, file=x.hlo repeat=5 duration=50848ns
```

After:
```
## Execution time, file=x.hlo repeat=1 duration=27968ns
## Execution time, file=x.hlo repeat=2 duration=28704ns
## Execution time, file=x.hlo repeat=3 duration=28064ns
## Execution time, file=x.hlo repeat=4 duration=28160ns
## Execution time, file=x.hlo repeat=5 duration=27968ns
```

Measured on 8xH100 DGX.

@frgossen

frgossen (approved):
Thanks!

allanrenucci:
Jax tests fail internally. E.g. https://github.com/jax-ml/jax/blob/7b4aee01849918430ca1b1b57d8a8326ee0d468b/tests/pjit_test.py#L433

```
*** SIGFPE (@0x55b368ed807c), si_code=1, see [go/stacktraces#s15](http://goto.google.com/stacktraces#s15) received by PID 4962 (TID 4962) on cpu 58; stack trace: ***
PC: @     0x55b368ed807c  (unknown)  xla::(anonymous namespace)::SubByteCollectiveNormalizationVisitor::HandleAllToAll()
    @     0x55b38a101be5       1904  FailureSignalHandler()
    @     0x7f8c2a4c4e80  (unknown)  (unknown)
    @     0x55b368ed807c         96  xla::(anonymous namespace)::SubByteCollectiveNormalizationVisitor::HandleAllToAll()
    @     0x55b3872c96e7        496  xla::PostOrderDFS<>()
    @     0x55b3872c72d6        128  xla::HloInstruction::Accept<>()
    @     0x55b367c1f22d        144  xla::HloComputation::Accept<>()
    @     0x55b368ed7d4c        128  xla::SubByteCollectiveNormalization::Run()
...
```

sergachev:
> Jax tests fail internally.

Thanks! I see, a check needed to be duplicated during the requested refactoring.

allanrenucci (approved):

allanrenucci:
The following test is failing: https://github.com/openxla/xla/blob/f9d3c91c5a3d626295322997bfbc3f76f573c1e9/xla/tests/collective_ops_test.cc#L1786
```
[ RUN      ] CollectiveOpsTest.AllToAll16BitInt
*** SIGFPE (@0x5564517982d8), si_code=1, see [go/stacktraces#s15](http://goto.google.com/stacktraces#s15) received by PID 4569 (TID 4569) on cpu 50; stack trace: ***
PC: @     0x5564517982d8  (unknown)  xla::(anonymous namespace)::SubByteCollectiveNormalizationVisitor::HandleAllToAll()
    @     0x55646ca3ad25       1904  FailureSignalHandler()
    @     0x7f0d8c891e80  (unknown)  (unknown)
    @     0x5564517982d8         96  xla::(anonymous namespace)::SubByteCollectiveNormalizationVisitor::HandleAllToAll()
    @     0x55646ad53cc7        496  xla::PostOrderDFS<>()
    @     0x55646ad518b6        128  xla::HloInstruction::Accept<>()
    @     0x5564514fb20d        144  xla::HloComputation::Accept<>()
    @     0x556451797f8c        128  xla::SubByteCollectiveNormalization::Run()
    @     0x556452ff501e        560  xla::HloPassPipeline::RunPassesInternal<>()
    @     0x556452ff488f       1104  xla::HloPassPipeline::Run()
    @     0x5564515e361b       2480  xla::gpu::GpuCompiler::OptimizeHloPostLayoutAssignment()
    @     0x5564515d05b6        384  xla::gpu::NVPTXCompiler::OptimizeHloPostLayoutAssignment()
    @     0x5564515ddae9       1344  xla::gpu::GpuCompiler::OptimizeHloModule()
    @     0x5564515e638d       3104  xla::gpu::GpuCompiler::RunHloPasses()
    @     0x556452e5df32        208  xla::LLVMCompiler::Compile()
    @     0x556451544046        672  xla::Compiler::Compile()
    @     0x5564515439e0        640  xla::HloRunner::CreateExecutableWithBufferAssignment()
    @     0x55645154154e         80  xla::HloRunner::ExecuteReplicated()
    @     0x55645154375b        128  xla::HloRunner::ExecuteReplicated()
    @     0x556451538ed9        480  xla::HloRunnerAgnosticTestBase::ExecuteReplicated()
    @     0x556451529f83       2176  xla::(anonymous namespace)::CollectiveOpsTest_AllToAll16BitInt_Test::TestBody()
    @     0x55645159619a         48  testing::Test::Run()
    @     0x556451596d93         64  testing::TestInfo::Run()
    @     0x556451597c77         80  testing::TestSuite::Run()
    @     0x5564515a887a        272  testing::internal::UnitTestImpl::RunAllTests()
    @     0x5564515a7658         64  testing::UnitTest::Run()
    @     0x556451562d9a        176  main
    @     0x7f0d8c7223d4        192  __libc_start_main
    @     0x5564514fb02a  (unknown)  [../sysdeps/x86_64/start.S:120](https://cs.corp.google.com/piper///depot/sysdeps/x86_64/start.S?l=120&ws=tap-presubmit-server/450223044&snapshot=2) _start
```

sergachev:
> The following test is failing:

Thanks! More duplication of checks was needed.

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
```diff
@@ -0,0 +1,208 @@
...
+absl::Status
+SubByteCollectiveNormalizationVisitor::ProcessCollectiveInstruction(
+    HloInstruction& hlo) {
+  if (hlo.operand_count() != 1) {
+    // Variadic ones are not supported yet.
+    return absl::OkStatus();
+  }
+  if (!CanBeRepresentedAs(hlo.operand(0)->shape(), casted_type_)) {
+    return absl::OkStatus();
+  }
```
allanrenucci: Can we factor out this logic instead of duplicating it? I.e.
```c++
bool SubByteCollectiveNormalizationVisitor::ShouldProcessInstruction(const HloInstruction& hlo) const {
  return hlo.operand_count() == 1 && CanBeRepresentedAs(hlo.operand(0)->shape(), casted_type_);
}
```

allanrenucci (changes requested):

xla/hlo/transforms/simplifiers/sub_byte_collective_normalization.cc
```diff
@@ -0,0 +1,210 @@
...
+    HloInstruction* hlo) {
+  return ProcessCollectiveInstruction(*hlo);
+}
+
+absl::Status
+SubByteCollectiveNormalizationVisitor::ProcessCollectiveInstruction(
+    HloInstruction& hlo) {
+  // HandleAllToAll already calls ShouldProcessInstruction, the other handlers
+  // don't.
+  if (hlo.opcode() != HloOpcode::kAllToAll && !ShouldProcessInstruction(hlo)) {
```
allanrenucci: This small optimization does not pull its weight IMO.

```suggestion
  if (!ShouldProcessInstruction(hlo)) {
```

allanrenucci (changes requested):
One small nit. Otherwise LGTM!

allanrenucci (approved):
Thank you!

sergachev:
Rebased to resolve conflicts.

allanrenucci (approved):

PR #31679, opened 2025-09-22
[XLA:GPU] Support block scaled dot fusions in triton GEMM legacy emitter

sergey-kozub:
📝 Summary of Changes
Update legacy Triton GEMM emitter to allow fusions around kScaledDot (not just kDot), and emit the Triton's DotScaled op instead (supports MXFP8 only).

🎯 Justification
This change will allow running the autotuner and select the faster implementation of scaled dot (Triton or cuDNN).

🚀 Kind of Contribution
✨ New Feature (part of PR chain)

### DIFF
--- a/xla/backends/gpu/codegen/triton/emitter_helpers.cc
+++ b/xla/backends/gpu/codegen/triton/emitter_helpers.cc
@@ -106,10 +106,9 @@ absl::StatusOr<Type> TritonType(EmitterLocOpBuilder& b, PrimitiveType t) {
       return b.getType<mlir::Float8E5M2Type>();
     case F8E4M3FN:
       return b.getType<mlir::Float8E4M3FNType>();
+    // This is a storage-only data type, use integer type to represent it.
     case F8E8M0FNU:
-      return b.getType<mlir::Float8E8M0FNUType>();
-    case F4E2M1FN:
-      return b.getType<mlir::Float4E2M1FNType>();
+      return b.getI8Type();
     default:
       return absl::UnimplementedError(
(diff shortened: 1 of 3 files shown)


loislo (approved):

xla/backends/gpu/codegen/triton/emitter_helpers.cc
```diff
@@ -106,10 +106,9 @@ absl::StatusOr<Type> TritonType(EmitterLocOpBuilder& b, PrimitiveType t) {
       return b.getType<mlir::Float8E5M2Type>();
     case F8E4M3FN:
       return b.getType<mlir::Float8E4M3FNType>();
+    // This is a storage-only data type, use integer type to represent it.
     case F8E8M0FNU:
-      return b.getType<mlir::Float8E8M0FNUType>();
-    case F4E2M1FN:
-      return b.getType<mlir::Float4E2M1FNType>();
+      return b.getI8Type();
```
loislo, line 111: it should be the regular Float8e8m0 type. We have a rewriter that converts it from f8 to i8 right after emitting

bchetioui:
Hey @sergey-kozub, thanks for the PR!

However, we are looking to delete `fusion_emitter_legacy_matmul.cc` in short order. As a result, I don't think we want to accept this PR.

Note, @loislo has made changes to support block-scaled dot fusions in the generic Triton emitter---so this should also not be necessary?

sergey-kozub:
> As a result, I don't think we want to accept this PR.

Ok, converted to Draft.

PR #31708, opened 2025-09-23
[XLA:CPU][oneDNN graph] Enable dots with mismatched dimension ranks

akhilgoe:
This PR enables oneDNN graph rewrite for cases where operand dimension ranks are unequal.


xla/backends/cpu/onednn_support.cc
```diff
@@ -110,4 +88,68 @@ absl::StatusOr<bool> IsOneDnnDotSupported(
          !dot_canonical_dims.rhs_column_major;
 }
 
+// TODO(intel-tf): Refactor this once oneDNN thunks are implemented
```
akhilgoe: Once oneDNN CC thunks are fully implemented, I'll refactor this to use a shared implementation for similar cases across oneDNN graphs and primitives based code paths.

akhilgoe:
Hi @penpornk, I have updated the PR with latest changes. Please take a look when you get time. Thanks!

xla/backends/cpu/transforms/BUILD
```diff
@@ -21,7 +21,7 @@ cc_library(
     name = "library_rewriter",
     srcs = ["library_rewriter.cc"],
     hdrs = ["library_rewriter.h"],
-    local_defines = if_graph_api(["XLA_ONEDNN_USE_GRAPH_API=1"]),
+    defines = if_graph_api(["XLA_ONEDNN_USE_GRAPH_API=1"]),
```
penpornk, line 24: Just curious, why does `local_defines` not work here?

xla/service/cpu/tests/onednn_fusion_test.cc
```diff
@@ -31,6 +33,67 @@ inline constexpr bool IsOneDnnGraphEnabled() {
...
+  void SetUp() override {
+    if (!IsOneDnnGraphEnabled()) {
+      GTEST_SKIP() << "oneDNN graph is disabled!";
+    }
+    if (!IsSupportedType(dtype_)) {
+      GTEST_SKIP() << "CPU does not support " << dtypeString_;
+    }
+  }
+
+  void AdjustToleranceForDtype(PrimitiveType for_type, float atol, float rtol) {
```
penpornk, line 82: This function is not used?

penpornk (commented):
High-level question please: Why do we need to change the HLO before matching? This affects other libraries' matching too. Can't we just relax the dimension checks in `IsOneDnnDotSupported` in onednn_support.cc?

xla/service/cpu/tests/onednn_fusion_test.cc
akhilgoe: Yes, not used right now. Just added as a utility in case we'd like to adjust tolerance for any particular test (and any particular dtype).

xla/backends/cpu/transforms/BUILD
akhilgoe: `local_defines` don't seem to be visible inside header files, and this macro is being used in `library_rewriter.h`

https://github.com/openxla/xla/blob/610351c8a2aeff0ad8ffd6679b1434f0570d6ce7/xla/backends/cpu/transforms/library_rewriter.h#L65-L71

akhilgoe:
Thanks for the review @penpornk!

Yes, you are right! That is what we had originally planned i.e. changing HLO as we build the `fusion` (https://github.com/openxla/xla/pull/31708/commits/b8fb1a44bed2522a00510ff4eb5214d13d6a0f7e). 
However, this approach poses some challenges, especially in the context of the data movement ops like `Bitcast` and `Reshape` that might appear between say a `Dot` and a fusible unary/binary op like `Exp`. oneDNN graph currently cannot readily absorb such ops while building the post-op chain so we have to modify the HLO. Adding logic to check and move such ops while building the fusion can become a little messy and hard to debug in case of any bugs. That is why we thought of having a separate canonicalization pipeline before building fusions.   

This pipeline is currently guarded by [the library feature flags](https://github.com/openxla/xla/pull/31708/files#diff-87757202e9bec4d58425c2d03af78b660e847d09450bc95aac3ea661d7ada07eR947-R951), so these passes are executed only when `xnnpack` is explicitly disabled and `onednn` is enabled. Also, none of the optimizations we plan to add here are library specific. 
Few things that we plan to do in this pass pipeline:
1. Contract consecutive bitcasts.
2. Move data movement ops around. (possibly expand this to `Convert` too)
3. Some transformations to bring certain ops to canonical form. This helps increase the pool of ops that can be rewritten to oneDNN fusions. 

Please let me know what you think, hopefully I was able to explain the issue with the original approach.

xla/backends/cpu/transforms/BUILD
penpornk: I see. Thank you for the explanation!

xla/service/cpu/tests/onednn_fusion_test.cc
penpornk: I think we should just remove it and add it in a PR that uses it.

penpornk (changes requested):

penpornk:
> This pipeline is currently guarded by [the library feature flags](https://github.com/openxla/xla/pull/31708/files#diff-87757202e9bec4d58425c2d03af78b660e847d09450bc95aac3ea661d7ada07eR947-R951), so these passes are executed only when xnnpack is explicitly disabled and onednn is enabled. Also, none of the optimizations we plan to add here are library specific.

Just because it isn't library-specific doesn't mean we need one. I still don't see the need for an extra pass, especially since it will introduce extra bitcasts (and other extra ops) for matmuls that aren't matched. 

I think it makes more sense to add extra ops into the fusion op (e.g., `LibraryRewriter` already handles output conversion [here](https://github.com/openxla/xla/blob/32771aea675fe9d0497733ef0d63c668d4e974da/xla/backends/cpu/transforms/library_rewriter.cc#L282)), or just put the op in the fusion op as-is and push the complexity into `onednn_emitter`.

> Few things that we plan to do in this pass pipeline:
> 1. Contract consecutive bitcasts.
> 2. Move data movement ops around. (possibly expand this to Convert too)
> 3. Some transformations to bring certain ops to canonical form. This helps increase the pool of ops that can be rewritten to oneDNN fusions.

Could you please give example HLO snippets for each of these case? (Both before and after pre-processing.)

akhilgoe:
>> This pipeline is currently guarded by [the library feature flags](https://github.com/openxla/xla/pull/31708/files#diff-87757202e9bec4d58425c2d03af78b660e847d09450bc95aac3ea661d7ada07eR947-R951), so these passes are executed only when xnnpack is explicitly disabled and onednn is enabled. Also, none of the optimizations we plan to add here are library specific.

> Just because it isn't library-specific doesn't mean we need one.

Agreed! I was just trying to address your initial concern regarding this affecting patterns for other libraries.

As for your question on this pass adding additional bitcasts / ops. That is a valid concern and that is part of the reason why we would want to have a separate optimization / canonicalization pipeline where initially we would modify the HLO to potentially increase the pool of candidate fusion centroids, and then we would clean up all the remnants of the modifications that we made. This pass is just 1 of the passes that we would want to have here. This will be followed by other simplification passes.

An example, consider this pseudo-HLO:

```
op0 = bf16[256,1024], parameter(0)
op1 = bf16[1024], parameter(1)
addend = bf16[16,16], parameter(2)
multiplier = bf16[256], parameter(3)
dot0 = bf16[256] dot(op0, op1), lhs_contracting_dim=1, rhs_contracting_dim=0
b0 = bf16[16,16] bitcast(dot0)
add = bf16[16,16] add(b0, addend)
b1 = bf16[256] bitcast(add)
m0 = bf16[256] multiply(b1, multiplier)
```

As of today, XLA does not allow BF16 arithmetic on CPU, so we can expect the input to the `library_rewriter` (or the input to the proposed canonicalization pipeline) to be something like below:

```
op0 = bf16[256,1024], parameter(0)
op1 = bf16[1024], parameter(1)
addend = bf16[16,16], parameter(2)
addend_f32 = f32[16,16] convert(addend)
multiplier = bf16[256], parameter(3)
multiplier_f32 = f32[256] convert(multiplier)
dot0 = bf16[256] dot(op0, op1), lhs_contracting_dim=1, rhs_contracting_dim=0
dot0_f32 = f32[256] convert(dot0)
b0_f32 = f32[16,16] bitcast(dot0_f32)
add = f32[16,16] add(b0_f32, addend_f32)
add_bf16 = bf16[16,16] convert(add)
add_f32 = f32[16,16] convert(add_bf16)
b1 = f32[256] bitcast(add_f32)
m0 = f32[256] multiply(b1, multiplier_f32)
m0_bf16 = bf16[256] convert(m0)
```

After the changes from this PR (whether in the form of a pass or as part of the `library_rewriter` processing):

```
op0 = bf16[256,1024], parameter(0)
op1 = bf16[1024], parameter(1)
**op1_new = bf16[1024,1] bitcast(op1)**
addend = bf16[16,16], parameter(2)
addend_f32 = f32[16,16] convert(addend)
multiplier = bf16[256], parameter(3)
multiplier_f32 = f32[256] convert(multiplier)
dot0 = bf16[256,1] dot(op0, op1_new), lhs_contracting_dim=1, rhs_contracting_dim=0
**dot0_res = bf16[256] bitcast(dot0)**
dot0_f32 = f32[256] convert(dot0_res)
b0_f32 = f32[16,16] bitcast(dot0_f32)
add = f32[16,16] add(b0_f32, addend_f32)
add_bf16 = bf16[16,16] convert(add)
add_f32 = f32[16,16] convert(add_bf16)
b1 = f32[256] bitcast(add_f32)
m0 = f32[256] multiply(b1, multiplier_f32)
m0_bf16 = bf16[256] convert(m0)
```

Here, there is significant opportunity for simplification and canonicalization, particularly by addressing the multiple bitcasts and conversions that are coming in the way of expanding the post-op chains.

After all the simplifications and canonicalization related to the data movement ops (to be implemented as part of future pass(es) in this pipeline), we expect the HLO to be something like below:

```
op0 = bf16[256,1024], parameter(0)
op1 = bf16[1024], parameter(1)
op1_new = bf16[1024,1] bitcast(op1)
addend = bf16[16,16], parameter(2)
addend_up = bf16[256] bitcast(addend)
addend_up_f32 = f32[256] convert(addend_up)
multiplier = bf16[256], parameter(3)
multiplier_f32 = f32[256] convert(multiplier)
dot0 = bf16[256,1] dot(op0, op1_new), lhs_contracting_dim=1, rhs_contracting_dim=0
dot0_res = bf16[256] bitcast(dot0)
dot0_f32 = f32[256] convert(dot0_res)
add = f32[256] add(dot0_f32, addend_up_f32)
m0 = f32[256] multiply(add, multiplier_f32)
m0_bf16 = bf16[256] convert(m0)
```

Graphically:

Before (current input to `library_rewriter`):
Note: This does not include the changes from this PR and the dot here cannot be re-written to a `onednn-fusion`.

pre




After (changes from this PR and future passes in the canonicalization pipeline):

post


If the `library_rewriter` receives this HLO as opposed to the initial input, it would be a lot easier to identify and match fusions. Doing everything in `library_rewriter` might limit our ability to fuse nodes in complex scenarios and pushing complexities to the emitter might be risky. What do you think? Would love to discuss this in detail if needed.

akhilgoe:
Closing for now, will re-open with updated implementation.

PR #31718, opened 2025-09-23
Support building with Bzlmod (native rules_python toolchain)

meteorcloudy:


xla/mlir_hlo/tests/BUILD
```diff
@@ -27,7 +27,7 @@ package(
         tags = [
             "nomsan",  # The execution engine doesn't work with msan, see b/248097619.
         ],
-        deps = ["@pypi_lit//:pkg"],  # copybara:comment
+        deps = ["@pypi//lit"],  # copybara:comment
```
rickeylev, line 30: fyi: this can probably be done before bzlmod is used. The `@pypi//lit` label should also work under workspace

MODULE.bazel
```diff
@@ -0,0 +1,157 @@
+module(name = "xla")
+
+##############################################################
+# Bazel module dependencies
```
rickeylev, line 4: I wonder which are dev-only dependencies 😬 

The XLA project is a bit funny because it's more a "library" module than a terminal module. It has a bunch of things that downstream projects re-use. I think it makes sense to ignore this problem for now, until XLA itself is using bzlmod.

MODULE.bazel
```diff
@@ -0,0 +1,157 @@
...
+python.toolchain(python_version = "3.11")
+
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+pip.parse(
+    hub_name = "xla_pypi",
+    python_version = "3.11",
+    requirements_lock = "//:requirements_lock_3_11.txt",
+)
+
+use_repo(pip, pypi = "xla_pypi")
```
rickeylev, line 109: Overall, the python and pip configuration here looks roughly correct.

third_party/py/BUILD.bazel
```diff
@@ -1,44 +1,44 @@
...
-    name = "requirements",
-    extra_args = [
-        "--allow-unsafe",
-        "--build-isolation",
-    ],
-    generate_hashes = True,
-    requirements_in = "requirements.in",
-    requirements_txt = REQUIREMENTS,
-)
+# compile_pip_requirements(
```
rickeylev: Just curious: why did these need to be commented out?

Not a big deal to drop usage of pip. I saw something uv above, but I don't see any usage of uv?

rules_python has an experimental uv module extension and rules. I'm not sure if using it as part of this project will help or not, though.

In any case, pip.parse doesn't care how requirements.txt is generated, so its fine to use any tool to do it.

third_party/py/numpy/BUILD.bazel
```diff
@@ -2,14 +2,25 @@ licenses(["restricted"])
...
+
 alias(
     name = "numpy",
-    actual = "@pypi_numpy//:pkg",
+    actual = "@pypi//numpy",
 )
 
-alias(
+whl_filegroup(
+    name = "numpy_includes",
```
rickeylev, line 14: Hm. Are these numpy headers used only in testing contexts? If so, all good, no concerns.

If not, then, potential problem: XLA will have its own pypi closure, using its own definition of the numpy headers. If those are out of sync with what another consumer also building against numpy headers, then version skew issues could occur.

This stems from the workspace/bzlmod difference of repo visibility. In workspace, a "library" project (like xla) would late-bind to a name by using e.g. `@pypi//whatever`, and then any consumer just has to ensure `@pypi//whatever` is defined. Under bzlmod with pip.parse, the repo's are part of the pip extension, and it has code that'll give an error if two modules try to use the same hub name. Allowing that sort of late-binding is something we've yet to get around to implementing in rules_python. Worst case, we can probably patch the pip.parse code to allow a duplicate hub_name, giving preference to earlier BFS modules

MODULE.bazel
```diff
@@ -0,0 +1,157 @@
...
+      "//third_party/absl:check_op.patch",
+      "//third_party/absl:check_op_2.patch",
+  ],
+  patch_strip = 1,
+)
+
+##############################################################
+# C++ dependencies
+
+# TODO: most of them can be Bazel modules, but need a release strategy for them
```
rickeylev, line 65: For C++ stuff, turning them into modules makes sense. Doing so has become the defacto standard with bzlmod.

For python packages, especially ones publish to pypi, I don't think so. It makes more sense to get them from pypi instead via requirements.

third_party/py/BUILD.bazel
meteorcloudy: Because this target depends on the `python_version_repo` to pass through the current requirement file being used, which I didn't introduce in the bzlmod yet. I will have to think through how to better set up the requirement update targets in the bzlmod build.

third_party/py/numpy/BUILD.bazel
meteorcloudy: So I think the numpy headers defined here is actually transitively used by JAX.
```
//jaxlib/tools:jaxlib_wheel (d614af8)
//jaxlib/tools:jaxlib_sources (d614af8)
//jaxlib/tools:jaxlib_sources_py (d614af8)
//jaxlib/tools:jaxlib_sources_py_gather (d614af8)
//jaxlib:jaxlib (d614af8)
//jaxlib/cpu:_sparse (d614af8)
//jaxlib/cpu:_sparse.so (d614af8)
//jaxlib:kernel_nanobind_helpers (d614af8)
@xla//xla/tsl/python/lib/core:numpy (d614af8)
@xla//third_party/py/numpy:headers (d614af8)
```

That means we do need to make sure the numpy headers match the current python toolchain selected by the root module. We can use `override_repo(pip, xla_pypi = "pypi")` to override the `xla_pypi` repo to the one from root, but I would consider this as the last resort if we cannot find a better alternative. Allowing a common hub name for those ML projects might be worth it?

meteorcloudy:
Closing in favor of https://github.com/openxla/xla/pull/32055

PR #31746, opened 2025-09-23
Change the padded index value to be invalid in prefix scan of scatter_determinism_expander

serach24:
📝 Summary of Changes
Fixed a critical bug in the ScatterDeterminismExpander where the prefix scan algorithm was using zero padding for indices, causing false matches with valid index 0. This led to incorrect accumulation in scatter operations, particularly affecting scatter_set operations.
Key changes:
Modified CreateScanWithIndices to use operand tensor size as padding value instead of zero
Updated function signatures to pass operand_dims instead of full scatter instruction
Fixed HLO verification test expectations to match the corrected generated code

🎯 Justification
This fix is critical for correctness of scatter operations, especially:
Scatter_set operations where zero is a valid update value
TensorFlow scatter operations that were failing due to incorrect accumulation
The bug was causing incorrect results in scenarios like:
Scattering [1, 2, 0, 0] into positions [0, 1, 2, 3] would incorrectly produce [0, 2, 0, 0] instead of [1, 2, 0, 0]
Consecutive duplicate indices like [0, 0] were not being properly accumulated

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests
Updated existing tests:
ScatterAddHloVerificationTest: Updated HLO verification patterns to match corrected generated code
All existing scatter determinism expander tests now pass with the fix
Test coverage:
Scalar scatter operations with duplicate indices
Multi-dimensional scatter operations (InversePermutation test)
Scatter-set vs scatter-add operations
Out-of-bounds index handling

🧪 Execution Tests:
Existing tests that now pass:
TensorFlowScatterV1_UpdateTwice: Tests consecutive duplicate indices [0, 0]
Test scenarios covered:
Scatter operations with index 0 (previously causing false matches)
Scatter operations with consecutive duplicate indices
Multi-dimensional scatter operations with various index patterns
Both scatter_add and scatter_set operations
The fix ensures that all scatter operations now produce correct, deterministic results regardless of whether they use index 0 or have consecutive duplicate indices.

### DIFF
--- a/xla/service/scatter_determinism_expander.cc
+++ b/xla/service/scatter_determinism_expander.cc
@@ -311,7 +311,7 @@ static std::vector<HloInstruction*> SortIndicesAndUpdates(
 //   input for the next iteration.
 static absl::StatusOr<HloInstruction*> CreateScanWithIndices(
     HloComputation* parent, HloInstruction* updates, HloInstruction* indices,
-    HloComputation* to_apply) {
+    HloComputation* to_apply, absl::Span<const int64_t> operand_dims) {
   const Shape& updates_shape = updates->shape();
   const Shape& indices_shape = indices->shape();
   // Get the length of the input array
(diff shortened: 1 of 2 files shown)


xla/service/scatter_determinism_expander.cc
```diff
@@ -368,15 +376,6 @@ static absl::StatusOr<HloInstruction*> CreateScanWithIndices(
     auto* indices_mask = parent->AddInstruction(HloInstruction::CreateCompare(
         ShapeUtil::MakeShape(PRED, {num_updates}), indices,
         concatenated_indices, ComparisonDirection::kEq));
-    auto* first_element_false = parent->AddInstruction(
```
akuegel, line 371: Makes sense to remove this now. When I saw this fix, I was a bit sceptical it would work in all cases, which is how I came up with the additional test. It was a good quick fix though for the "normal" cases.

akuegel (approved):
Thanks for the additional fix, looks good to me.

PR #31745, opened 2025-09-23
[XLA:CPU][oneDNN] Enable oneDNN Softmax Custom Calls in Thunk Runtime

othakkar:
This PR enables support for oneDNN Softmax operations in the XLA:CPU Thunk runtime, building upon the foundational implementation of `OneDnnOpThunk`.

Key changes:
- Updated thunk emitter to emit `OneDnnOpThunk` for oneDNN softmax during compilation.
- Enabled oneDNN custom call rewrite for Softmax in `cpu_compiler.cc` via `OneDnnOpsRewriter` (while temporarily disabling the rewrite for oneDNN Layer Norm).
- Added support for oneDNN Softmax op via `ExecuteOneDnnSoftmax(...)` in `onednn_softmax.cc`.

PR #31747, opened 2025-09-23
[XLA:CPU][oneDNN] Use Async Threadpool for Pre-Computing Blocked Weights

othakkar:
This PR replaces the use of the synchronous (legacy) oneDNN threadpool interface with the asynchronous oneDNN threadpool interface to reorder weights for pre-computing blocked weights.


othakkar:
@penpornk Please review this (bug-fix) PR when you have a chance, thanks!

allanrenucci:
If this a bug fix, can you add a test that exercises this bug?

othakkar:
@allanrenucci, added a unit test in [a1ffd8d](https://github.com/openxla/xla/pull/31747/commits/a1ffd8dd669c537f95e194e258f115607fb05b8f).

xla/service/cpu/tests/onednn_matmul_test.cc
```diff
@@ -1647,6 +1647,32 @@ TEST_F(MatmulTest, WeightsPrepackAndScratch) {
   )");
 }
 
+TEST_F(MatmulTest, PrepackLarge2DWeights) {
```
penpornk, line 1651: Is this test actually passing in OSS? We haven't set the flag to turn the pass on, so it shouldn't pass yet right? Can we expedite the PR to fix the test and get it in first so we can actually test this and other future PRs?

penpornk (commented):
Thank you for the PR!

xla/service/cpu/tests/onednn_matmul_test.cc
othakkar: @penpornk, thanks for the feedback! I just want to make sure I understand your request correctly.

As of now, oneDNN tests (including the `PrepackLarge2DWeights` test) won't pass in OSS unless XLA is built with the `--define=build_with_onednn_async=true` flag. If that's set and the unit test is run with `XLA_FLAGS="--xla_cpu_experimental_onednn_custom_call=true"`, the test will fail on the current `main`. However, with the changes in this PR and those flags, the test should pass successfully.

Please let me know if this clarifies things or if there's another specific PR you're referring to that needs to be expedited.

penpornk: Can we set the `DebugOptions` here so we don't need to set `XLA_FLAGS`?

penpornk (changes requested):

xla/service/cpu/tests/onednn_matmul_test.cc
othakkar: Sure. I added the flag setting (https://github.com/openxla/xla/pull/31747/commits/46bc2e644eb57fa7c482d058c76449c3de03d187) for oneDNN MatMul and Conv tests (as Conv is merged already). I will add it for Layernorm and Softmax in the Softmax PR (https://github.com/openxla/xla/pull/31745).

penpornk (approved):
Thank you for the changes!

PR #31795, opened 2025-09-24
[GPU] Assign default color to tuples

terryysun:
📝 Summary of Changes
[Downstream check](https://github.com/openxla/xla/blob/main/xla/pjrt/pjrt_executable.cc#L288-L299) assumes tuples on default memory space, force assign default color to tuples will get around the check.

🎯 Justification
NCCL user buffer runs are crashing on MaxText main. This PR fixes the crash.

🚀 Kind of Contribution
🐛 Bug Fix

N/A.

🧪 Unit Tests:
Existing unit tests.

🧪 Execution Tests:
Added multiple execution tests.


xla/service/gpu/gpu_memory_space_assignment.h
```diff
@@ -115,6 +115,12 @@ inline BufferAssigner::Colorer CollectiveColorer(bool use_user_buffers,
           value->set_color(BufferValue::Color(memory_space));
           continue;
         }
+      } else if (defining_position.shape().IsTuple()) {
+        // Tuples do not have layout while downstream allocation consistency
```
frgossen: Tuples do not have a layout but they do live in memory. Maybe just say that tuples live in the default memory space?

xla/service/gpu/gpu_memory_space_assignment.h
```diff
@@ -115,6 +115,12 @@ inline BufferAssigner::Colorer CollectiveColorer(bool use_user_buffers,
...
+        // check will assume default memory space for them, force assign to
+        // avoid downstream errors.
+        value->set_color(0);
```
frgossen: Can you add a constant for the default memory space? We already have some, see kTempBufferMemorySpaceColor

frgossen (changes requested):
Thanks for fixing this!

xla/service/gpu/gpu_memory_space_assignment.h
terryysun: updated

xla/service/gpu/gpu_memory_space_assignment.h
terryysun: added

frgossen (approved):
Thanks!

terryysun:
Hi @frgossen could you help kick off copybara? Thanks!

PR #31804, opened 2025-09-25
Fix class vs struct ambiguity

srivatsav1998:
📝 Summary of Changes
The `BufferView` class was defined as a `struct`, but in the `LogicalIndexView`, it was redefined as `class` which caused build failure on my machine. 

🎯 Justification
This change will fix the ambiguity and thereby prevent any build failures where warnings are treated as errors.

🚀 Kind of Contribution
Please remove what does not apply: 🐛 Bug Fix, ♻️ Cleanup

N/A

🧪 Unit Tests:
N/A

🧪 Execution Tests:
N/A


allanrenucci:
Thanks! Already fixed in #31819.

PR #31810, opened 2025-09-25
[XLA:CPU][oneDNN] Select rewriting pass for dot operation based on runtime flags for oneDNN

yimeisun123:
When runtime flag, xla_cpu_use_onednn, is enabled for oneDNN, rewrite the dot operation through DotLibraryRewriter instead of OneDnnContractionRewriter.


xla/backends/cpu/transforms/BUILD
```diff
@@ -21,7 +21,7 @@ cc_library(
     name = "library_rewriter",
     srcs = ["library_rewriter.cc"],
     hdrs = ["library_rewriter.h"],
-    local_defines = if_graph_api(["XLA_ONEDNN_USE_GRAPH_API=1"]),
+    defines = if_graph_api(["XLA_ONEDNN_USE_GRAPH_API=1"]),
```
penpornk, line 24: Just curious, why does `local_defines` not work here?

xla/service/cpu/cpu_compiler.cc
```diff
@@ -908,15 +908,18 @@ absl::Status CpuCompiler::RunHloPassesAfterLayoutAssn(
...
     // Run SimplifyFPConversions pass to simplify the BF16 pattern and make it
     // easier to match.
     // Remove `f32 -> bf16 -> f32` casts inserted by bf16 normalization.
     if (debug_options.xla_allow_excess_precision()) {
       pipeline.AddPass<SimplifyFPConversions>();
     }
-    pipeline.AddPass<OneDnnContractionRewriter>(max_parallelism,
-                                                compile_options.thread_pool);
+    pipeline.AddPass<OneDnnContractionRewriter>(
+        max_parallelism, compile_options.thread_pool, use_onednn_graph);
```
penpornk, line 922: We can just skip adding this pass instead, right?
```suggestion
    if (!use_onednn_graph) {
      pipeline.AddPass<OneDnnContractionRewriter>(
          max_parallelism, compile_options.thread_pool, use_onednn_graph);
    }
```

penpornk (changes requested):

xla/backends/cpu/transforms/BUILD
yimeisun123: With local_defines, the defined flag is only local to the cc file, but not available to "library_rewriter.h" which actually check for this flag.

xla/service/cpu/cpu_compiler.cc
yimeisun123: The OneDnnContractionRewriter includes both dot and convolution custom call rewrite. We only want to block the the dot rewrite part, the convolution will go through custom call.

xla/backends/cpu/transforms/BUILD
penpornk: Thank you for the clarification!

xla/service/cpu/cpu_compiler.cc
penpornk: Got it. Just a heads up that `LibraryRewriter` will support convolution too in the future.

penpornk (approved):

PR #31886, opened 2025-09-26
Fix libdevice search

Flamefire:
📝 Summary of Changes
This enhances the search for the CUDA libdevice path:
- Fix an invalid empty path added when `TF_CUDA_TOOLKIT_PATH` which may be empty
- Fix invalid paths based on runtime folders: `runfiles_dir.substr(0, runfiles_ind + runfiles_suffix.length())` is not meaningful when `runfiles_ind` isn't valid, i.e. `std::string::npos`
- Add `$CUDA_HOME` to the search paths. This is also used in TensorFlow already

🎯 Justification
Without this the libdevice file won't be found if CUDA isn't installed in a standard location or e.g. an updated version is available in a different location.   
This is the case for e.g. HPC systems where multiple CUDA versions are available side-by-side.

🚀 Kind of Contribution
🐛 Bug Fix, ♻️ Cleanup

Fixes #28590

🧪 Unit Tests:

Simple test that when `CandidateCudaRoots` returns anything it contains `$CUDA_HOME`

### DIFF
--- a/xla/stream_executor/cuda/assemble_compilation_provider_test.cc
+++ b/xla/stream_executor/cuda/assemble_compilation_provider_test.cc
@@ -15,6 +15,7 @@ limitations under the License.
 
 #include "xla/stream_executor/cuda/assemble_compilation_provider.h"
 
+#include <algoritm>
 #include <memory>
 #include <string>
 
@@ -25,6 +26,7 @@ limitations under the License.
 #include "xla/stream_executor/cuda/compilation_provider_options.h"
 #include "xla/stream_executor/cuda/nvjitlink_support.h"
 #include "xla/stream_executor/cuda/ptx_compiler_support.h"
+#include "xla/tsl/platform/env.h"
 #include "xla/tsl/platform/status_matchers.h"
 #include "xla/tsl/platform/statusor.h"
 #include "tsl/platform/cuda_root_path.h"
@@ -36,6 +38,18 @@ namespace {
 using ::testing::AllOf;
 using ::testing::HasSubstr;
 using ::tsl::testing::StatusIs;
+
+TEST(AssembleCompilationProviderTest,
+     CandidateCudaRootsConsidersCUDA_HOME) {
+  const std::string cuda_home = "/my/cuda/home";
+  tsl::setenv("CUDA_HOME", cuda_home.c_str(), 1);
+  auto roots = tsl::CandidateCudaRoots();
+  if (roots.empty()) {
+    GTEST_SKIP() << "Test only when any root is set";
+  }
+  EXPECT_THAT(std::find(roots.begin(), roots.end(), cuda_home),
+              Ne(roots.end()));
+}
 
 TEST(AssembleCompilationProviderTest,
      ReturnsErrorIfNoCompilationProviderIsAvailable) {
--- a/xla/stream_executor/cuda/assemble_compilation_provider_test.cc
+++ b/xla/stream_executor/cuda/assemble_compilation_provider_test.cc
@@ -43,12 +43,10 @@ TEST(AssembleCompilationProviderTest,
      CandidateCudaRootsConsidersCUDA_HOME) {
   const std::string cuda_home = "/my/cuda/home";
   tsl::setenv("CUDA_HOME", cuda_home.c_str(), 1);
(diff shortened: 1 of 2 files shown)


xla/stream_executor/cuda/assemble_compilation_provider_test.cc
```diff
@@ -37,6 +39,18 @@ using ::testing::AllOf;
...
+TEST(AssembleCompilationProviderTest,
+     CandidateCudaRootsConsidersCUDA_HOME) {
+  const std::string cuda_home = "/my/cuda/home";
+  tsl::setenv("CUDA_HOME", cuda_home.c_str(), 1);
+  auto roots = tsl::CandidateCudaRoots();
+  if (roots.empty()) {
+    GTEST_SKIP() << "Test only when any root is set";
+  }
+  EXPECT_THAT(std::find(roots.begin(), roots.end(), cuda_home),
+              Ne(roots.end()));
```
beckerhe: Can you use `EXPECT_THAT(roots, ::testing::Contains(cud_home));` here?

This will produce a better error message when it fails.

xla/tsl/platform/default/cuda_root_path.cc
```diff
@@ -56,14 +56,21 @@ std::vector<std::string> CandidateCudaRoots() {
...
+          runfiles_dir.substr(0, runfiles_ind + runfiles_suffix.length()),
+          cuda_dir_name);
+      roots.push_back(cuda_dir);
+    }
   }
 
-  roots.push_back(TF_CUDA_TOOLKIT_PATH);
+  const char* cuda_home = getenv("CUDA_HOME");
+  if (cuda_home)
+    roots.emplace_back(cuda_home);
```
beckerhe, line 70: Can you add curly braces here? (The google style guide wants us to use those everywhere.)

xla/tsl/platform/default/cuda_root_path.cc
```diff
@@ -56,14 +56,21 @@ std::vector<std::string> CandidateCudaRoots() {
...
+  std::string cuda_toolkit_path = TF_CUDA_TOOLKIT_PATH;
+  if (!cuda_toolkit_path.empty())
+    roots.push_back(std::move(cuda_toolkit_path));
```
beckerhe: Same here.

xla/stream_executor/cuda/assemble_compilation_provider_test.cc
```diff
@@ -37,6 +39,18 @@ using ::testing::AllOf;
...
 using ::tsl::testing::StatusIs;
 
+TEST(AssembleCompilationProviderTest,
+     CandidateCudaRootsConsidersCUDA_HOME) {
+  const std::string cuda_home = "/my/cuda/home";
+  tsl::setenv("CUDA_HOME", cuda_home.c_str(), 1);
+  auto roots = tsl::CandidateCudaRoots();
+  if (roots.empty()) {
+    GTEST_SKIP() << "Test only when any root is set";
+  }
```
beckerhe: I don't understand this SKIP. Shouldn't `CandidateCudaRoots` always at least return what you set in CUDA_HOME?

Flamefire: I copied that from similar tests. A quick inspection shows that the whole function always returns an empty vector when `PLATFORM_GOOGLE` is defined.

Judging by the runtime checks in other tests it might not be defined for tests or we could if-def out the whole test

beckerhe: Oh I see. This is probably because we have a different implementation of CandidateCudaRoots internally that also doesn't take CUDA_HOME into account. Therefore the skip is actually needed.

But it would be better to do it at the very top of the test case and to check whether we're building the OSS version of XLA. Something like that:

```
if (!tsl::kIsOpenSource) {
  GTEST_SKIP() << "CUDA_HOME is only being considered in the OSS build of XLA.";
}
```

Flamefire: Done, also fixed 2 typos in the check

beckerhe (approved):

PR #29167, opened 2025-07-22
Draft: [GPU] L2 prefetch.

sergachev:

PR #31994, opened 2025-09-29
[NFC] Move computation simplification methods from command buffer scheduling to a new library.

sergachev:


xla/hlo/transforms/simplifiers/computation_canonicalizers.h
```diff
@@ -0,0 +1,40 @@
...
+// the construction of command buffer computations because we don't need to deal
+// with parameters and constants that have users outside of a command buffer.
+// Returns true if there is a change in the order of instructions, false
+// otherwise.
+absl::StatusOr<bool> MoveParametersAndConstantsToFront(HloComputation&);
+
+// Moves GetTupleElement instructions to right after the instruction that
+// produces the tuple. Returns whether the computation was changed. This is run,
+// for instance, before command buffer scheduling.
+//
```
bchetioui: nit: remove empty comment line

xla/hlo/transforms/simplifiers/computation_canonicalizers.h
```diff
@@ -0,0 +1,40 @@
...
+absl::StatusOr<bool> MoveGTEsRightAfterTupleDefinition(HloComputation&);
```
bchetioui, line 35: Do you intend to use this anywhere else? This was previously private, but now is accessible for other use cases; I'm not opposed to it if there's a use case, but I'd like to understand whether we do need to extend the "public" API. WDYT?

Note, if we do indeed want to expose this, then this will need accompanying tests in `computation_canonicalizers_test.cc`.

xla/hlo/transforms/simplifiers/computation_canonicalizers_test.cc
```diff
@@ -0,0 +1,75 @@
+
```
bchetioui: nit: empty line at the top of the file

bchetioui (changes requested):

xla/hlo/transforms/simplifiers/computation_canonicalizers.h
sergachev: Yes, it's used in https://github.com/openxla/xla/pull/29167.

bchetioui (approved):
Thanks!

PR #32003, opened 2025-09-29
[GPU][NFC] Merge methods querying fusion kind.

sergachev:


dimitar-asenov (approved):

xla/service/gpu/hlo_fusion_analysis.h
```diff
@@ -32,6 +32,8 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+bool IsFusionKind(const HloInstruction& hlo, absl::string_view kind);
```
bchetioui: This probably should be called IsGpuFusionKind and have a docstring. Otherwise, it may make us think that we are querying the fusion kind in `HloInstruction` which is different from the one in GpuBackendConfig :|

Alternatively, we could call this IsCustomFusionWithKind... and then check for kCustom + underlying FusionBackendConfig.

Could you please do one of those?

bchetioui (changes requested):

bchetioui (approved):
Thanks Ilia!

xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
```diff
@@ -68,6 +68,7 @@ limitations under the License.
 #include "xla/service/gpu/autotuning/redzone_buffers.h"
 #include "xla/service/gpu/backend_configs.pb.h"
 #include "xla/service/gpu/gpu_float_support.h"
+#include "xla/service/gpu/hlo_fusion_analysis.h"
```
akuegel, line 71: The gemm_fusion_autotuner target needs the corresponding dependency that provides this header.

akuegel (changes requested):

akuegel (approved):

PR #32055, opened 2025-09-30
Support building with Bzlmod

meteorcloudy:
- Added MODULE.bazel and module extensions to introduce external dependencies with Bzlmod.
- Added a CI config for Linux CPU build with Bzlmod (enabled by `--config=bzlmod`)

TODOs:
- Support overriding Python runtime
- Support build with more configs and platforms


meteorcloudy:
Rocm related changes will be upstreamed at https://github.com/openxla/xla/pull/32416

fcoUnda (approved):

third_party/extensions/python_version.bzl
```diff
@@ -0,0 +1,34 @@
+"""
+A much simplified version of third_party/py/python_configure.bzl, which generates the "python_version_repo" repo.
+
+This is just to keep the current build compatible with both WORKSPACE and Bzlmod, we may not need this in future.
+"""
+
+def _python_version_repo_impl(repository_ctx):
+    version = repository_ctx.os.environ.get("HERMETIC_PYTHON_VERSION", "3.11")
+    use_pywrap_rules = bool(
+        repository_ctx.os.environ.get("USE_PYWRAP_RULES", False),
```
rickeylev, line 15: env vars are strings, so returning False changes the data type, which is a bit misleading

bool() treats any non-empty string as true -- does the match the original logic?

third_party/extensions/python_version.bzl
```diff
@@ -0,0 +1,34 @@
...
+    )
+    repository_ctx.file("BUILD", "")
```
rickeylev: nit: BUILD.bazel

third_party/extensions/python_version.bzl
```diff
@@ -0,0 +1,34 @@
...
+    repository_ctx.file(
+        "py_version.bzl",
+        """
+HERMETIC_PYTHON_VERSION = "{version}"
+USE_PYWRAP_RULES = {use_pywrap_rules}
+""".format(
+            version = version,
+            use_pywrap_rules = use_pywrap_rules,
+        )
```
rickeylev: nit: move content into a module-level constant for readability. I find the dedent and indent hard to read. IME, it also makes it hard to maintain over time since the triple quotes context gets lost (e.g. in the github diff ui)

rickeylev (approved):

third_party/extensions/python_version.bzl
meteorcloudy: Yeah, copied from https://github.com/openxla/xla/blob/c4ff0fc760316d9dfd1b14e6dd4745c4fea7d678/third_party/py/python_repo.bzl#L79-L81

I think `bool(False)` is not that confusing?

third_party/extensions/python_version.bzl
meteorcloudy: Done

third_party/extensions/python_version.bzl
meteorcloudy: Good point, thanks!

meteorcloudy:
I will import this change manually

meteorcloudy:
This is merged at https://github.com/openxla/xla/commit/88e6fc7d12d0c75951f03f5ebe6867bb3f24364e

PR #32074, opened 2025-09-30
Proposed PjRt API Changes for Improved Cross-Process Transfers on GPUs

rao-ashish:
📝 Summary of Changes
This PR is a sketch outlining potential modifications to the PjRt API to better support cross-process transfers on GPUs, intended for discussion with @emilyfertig and @mwhittaker.

🎯 Justification
The current implementation of cross-process device-puts on GPUs does not cache communicators for reuse across multiple transfers between the same sets of devices, and does not aggregate transfers of multiple arrays together with ncclGroup calls. This PR changes the PjRt API so that the MakeCrossHostReceiveBuffers and CopyToRemoteDevice functions now receive information about the global device IDs involved in the data transfer so that communicators can be cached.

🚀 Kind of Contribution
Performance Improvement

This PR only makes API changes necessary for future performance improvements such as communicator caching, so no benchmarks were ran for this PR. Follow-up PRs implementing performance improvements will include benchmarks.

🧪 Execution Tests:

### DIFF
--- a/xla/pjrt/common_pjrt_client.cc
+++ b/xla/pjrt/common_pjrt_client.cc
@@ -310,7 +310,8 @@ absl::StatusOr<xla::Shape> CommonPjRtClient::MakeDefaultShapeForMemorySpace(
 }
 
 void CommonPjRtBufferImpl::CopyToRemoteDevice(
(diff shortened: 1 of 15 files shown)


emilyfertig:
Closing this since we merged a different version.

PR #32076, opened 2025-09-30
Use AcquireCollectiveCliques to Cache Communicators for Cross-Process Device-Puts

rao-ashish:
📝 Summary of Changes
This PR is a sketch outlining how communicators might be cached for cross-process device-puts, intended for discussion with @emilyfertig and @mwhittaker. This is PR 2 in a sequence of 2 PRs; the [first PR](https://github.com/openxla/xla/pull/32074) sketches out PjRt API changes needed to implement communicator caching.

🎯 Justification
The current implementation of cross-process device-puts on GPUs does not cache communicators for reuse across multiple transfers between the same sets of devices, reducing performance. This PR uses the existing communicator caching code in resource_requests.cc (in particular, AcquireCollectiveCliques) to cache and reuse communicators.

🚀 Kind of Contribution
Performance Improvement

📊 Benchmark
Benchmarks were ran on these [three toy programs](https://gist.github.com/rao-ashish/28155ffee98406b7c041037c20fc75e6) using cross-process device-puts.
<meta charset="utf-8"><div dir="ltr" style="margin-left:0pt;" align="left">
Test Name | Implementation | Time taken
-- | -- | --
example_basic_overlap.py | Current JAX | 27.500 sec / 10 iters
  | Prototype | 47.622 ms / 10 iters
example_overlap.py | Current JAX | 200.519 sec / 10 iters
  | Prototype | 580.936 ms / 10 iters
example_pp.py | Current JAX | 366.920 sec / 10 iters
  | Prototype | 2.485 sec / 10 iters
</div>

🧪 Execution Tests

### DIFF
--- a/xla/pjrt/common_pjrt_client.h
+++ b/xla/pjrt/common_pjrt_client.h
@@ -237,7 +237,7 @@ class CommonPjRtClient : public PjRtClient {
       tsl::RCReference<CommonPjRtRawBuffer> raw_buffer,
       std::vector<tsl::RCReference<tsl::AsyncValue>> definition_events,
       tsl::RCReference<PjRtDeviceEventPromise> usage_event_promise,
-      Future<std::string> serialized_descriptor,
+      PjRtGlobalDeviceId dst_global_device_id, CrossHostTransferId transfer_id,
       PjRtBuffer::RemoteSendCallback on_done);
 };
 
@@ -278,8 +278,9 @@ class CommonPjRtBufferImpl : public CommonPjRtBuffer {
   // device to host transfer to read the metadata of dynamic shape.
   absl::StatusOr<Shape> logical_on_device_shape() override;
 
-  void CopyToRemoteDevice(Future<std::string> serialized_descriptor,
-                          RemoteSendCallback on_done) override;
+  void CopyToRemoteDevice(PjRtGlobalDeviceId dst_global_device_id,
+                          CrossHostTransferId transfer_id,
+                          PjRtBuffer::RemoteSendCallback on_done) override;
 
   absl::StatusOr<std::unique_ptr<PjRtBuffer>> CopyToMemorySpace(
       PjRtMemorySpace* dst_memory_space) override;
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -947,9 +947,69 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
   return absl::OkStatus();
 }
 
+std::string CrossHostTransferName(PjRtGlobalDeviceId src_global_device_id,
+                                  PjRtGlobalDeviceId dst_global_device_id,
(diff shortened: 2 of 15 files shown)


xla/pjrt/common_pjrt_client.h
```diff
@@ -237,7 +237,7 @@ class CommonPjRtClient : public PjRtClient {
       tsl::RCReference<CommonPjRtRawBuffer> raw_buffer,
       std::vector<tsl::RCReference<tsl::AsyncValue>> definition_events,
       tsl::RCReference<PjRtDeviceEventPromise> usage_event_promise,
-      Future<std::string> serialized_descriptor,
+      PjRtGlobalDeviceId dst_global_device_id, int64_t transfer_id,
```
ezhulenev: Prefer strongly typed aliases instead of int64_t ids, it gets confusing very quickly with many different int64_t ids in the same context.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -947,76 +948,141 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
...
-  }
-  gpu::GpuCollectives* gpu_collectives =
-      tsl::down_cast<gpu::GpuCollectives*>(*collectives);
-  if (gpu_collectives == nullptr) {
-    on_done(absl::InternalError("Failed to get GPU collectives"),
-            /*sends_were_enqueued=*/false);
+// Returns a CollectiveCliques object for use in a cross-process transfer; the
+// CollectiveCliques object will hold a mapping from ranks to NCCL communicators
+// that should be used to perform the transfer.
+absl::StatusOr<gpu::Thunk::CollectiveCliques> GetTransferCollectiveCliques(
```
ezhulenev: because you always have a single clique key, prefer returning `absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>>` and call API `AcquireTransferGpuClique`gpu_cliques.h`

xla/pjrt/common_pjrt_client.h
rao-ashish: Thanks for taking a look! Added `CrossHostTransferId` as a strongly typed alias for transfer IDs.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Refactored the code to directly use `AcquireGpuClique`; this has been inlined inside `GetGpuTransferContext` instead of having a separate function.

xla/pjrt/common_pjrt_client.h
gspschmid: I think Eugene meant something like https://github.com/openxla/xla/blob/401cb62669d09784e3fcde64b4d722f062f455a1/xla/pjrt/pjrt_common.h#L40-L42 instead -- `using CrossHostTransferId = int64_t` just aliases the type, i.e. is not strongly-typed, meaning the type checker wouldn't prevent you from confusing `CrossHostTransferId` with any other int64_t.

PR #32115, opened 2025-10-01
[XLA:CPU] Make rendezvous timeouts configurable via flags

gerlero:
📝 Summary of Changes
As discussed in https://github.com/openxla/xla/pull/26921#issuecomment-3299642657, add configurable options for setting the timeout (and also the time before the "stuck" warning shows) when running in parallel on CPU.

🎯 Justification
Configurable timeouts are required for longer parallel workloads with uneven work distributions (i.e., the inputs on some CPUs take longer to process than on others).

Note: I've set the default values to follow the equivalent GPU case. These are a bit longer than the current hardcoded values, but it made sense to me to have the same defaults regardless of the device type.

Closes https://github.com/jax-ml/jax/issues/31818

🚀 Kind of Contribution
✨ New Feature

N/A

🧪 Unit Tests:
None

🧪 Execution Tests:
None


gerlero:
Updated to resolve conflicts

ezhulenev (approved):

PR #32187, opened 2025-10-02
PTX version 9.0 is supported starting with CUDA 13.0.

dimvar:
https://docs.nvidia.com/cuda/parallel-thread-execution/#release-notes

Also a slight refactoring of the code.

### DIFF
--- a/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
+++ b/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
@@ -39,7 +39,7 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
     // CUDA 12.6 -> PTX 8.5
     // CUDA 12.9 -> PTX 8.8
     return {cuda_version.major() - 4, cuda_version.minor() - 1, 0};
-  } else if (cuda_version <= stream_executor::SemanticVersion{13, 0, 0}) {
+  } else if (cuda_version < stream_executor::SemanticVersion{13, 1, 0}) {
     // Examples:
     // CUDA 11.0 -> PTX 7.0
     // CUDA 12.4 -> PTX 8.4
--- a/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
+++ b/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
@@ -19,7 +19,7 @@ namespace xla::gpu::nvptx {
 
 namespace {
 constexpr stream_executor::SemanticVersion kFallbackPtxVersion{6, 5, 0};
-constexpr stream_executor::SemanticVersion kMaxPtxVersion{8, 8, 0};
+constexpr stream_executor::SemanticVersion kMaxPtxVersion{9, 0, 0};
 }  // namespace
 
 stream_executor::SemanticVersion
@@ -33,20 +33,18 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
 
   // Mapping determined from
   // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes
-  // Examples:
-  // CUDA 11.0 -> PTX 7.0
-  // CUDA 11.1 -> PTX 7.1
-  // CUDA 12.0 -> PTX 8.0
-  // CUDA 12.4 -> PTX 8.4
-  // This versioning scheme is valid until CUDA 12.6
-  if (cuda_version < stream_executor::SemanticVersion{12, 6, 0}) {
-    return {cuda_version.major() - 4, cuda_version.minor(), 0};
-  }
-  // CUDA 12.6 -> PTX 8.5
-  // CUDA 12.8 -> PTX 8.7
-  // CUDA 12.9 -> PTX 8.8
-  if (cuda_version < stream_executor::SemanticVersion{12, 10, 0}) {
+  if (cuda_version >= stream_executor::SemanticVersion{12, 6, 0} &&
+      cuda_version < stream_executor::SemanticVersion{12, 10, 0}) {
+    // Examples:
+    // CUDA 12.6 -> PTX 8.5
+    // CUDA 12.9 -> PTX 8.8
     return {cuda_version.major() - 4, cuda_version.minor() - 1, 0};
+  } else if (cuda_version < stream_executor::SemanticVersion{13, 0, 0}) {
+    // Examples:
+    // CUDA 11.0 -> PTX 7.0
+    // CUDA 12.4 -> PTX 8.4
+    // CUDA 13.0 -> PTX 9.0
+    return {cuda_version.major() - 4, cuda_version.minor(), 0};
   }
 
   // Return maximum known PTX version.
--- a/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
+++ b/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
@@ -39,7 +39,7 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
     // CUDA 12.6 -> PTX 8.5
     // CUDA 12.9 -> PTX 8.8
     return {cuda_version.major() - 4, cuda_version.minor() - 1, 0};
-  } else if (cuda_version < stream_executor::SemanticVersion{13, 0, 0}) {
+  } else if (cuda_version <= stream_executor::SemanticVersion{13, 0, 0}) {
     // Examples:
     // CUDA 11.0 -> PTX 7.0
     // CUDA 12.4 -> PTX 8.4


xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
```diff
@@ -30,6 +30,13 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
     // We don't support CUDA below 11 anymore.
     return kFallbackPtxVersion;
   }
+  if (cuda_version > stream_executor::SemanticVersion{13, 0, 0}) {
```
Artem-B: I think it should be `>=`

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
```diff
@@ -38,18 +45,18 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
...
+  if (cuda_version >= stream_executor::SemanticVersion{12, 6, 0} &&
+      cuda_version < stream_executor::SemanticVersion{12, 10, 0}) {
+    minor_version = cuda_version.minor() - 1;
+  } else {
+    minor_version = cuda_version.minor();
   }
-
-  // Return maximum known PTX version.
-  return kMaxPtxVersion;
+  return {major_version, minor_version, 0};
```
Artem-B: I'd just collapse it all into a 
```
switch(major_version *10 + minor_version) {
case 120: return {8,0,0};
case 121: return {8,1,0};
,,,
case 126: return {8,5,0};
case 130: return {9,0,0};
}
```

It makes the mapping obvious, so there's no need for the reader to stumble and figure out what's the magic behind subtraction of one from the CUDA version, sometimes. And you can also drop the per-version comments.

dimvar: Done, thanks.

I left the earlier check as a > because I read it as "if the cuda version is newer than the latest version XLA knows, do this". >= works as well I just find > a little bit cleaner.

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
Artem-B: > I left the earlier check as a > because I read it as "if the cuda version is newer than the latest version XLA knows, do this". >= works as well I just find > a little bit cleaner.

In that case, the check should be moved down to the `default` case of the switch below. There it would be obvious that we're handling some version we didn't enumerate.

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
```diff
@@ -30,26 +32,32 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
...
+    case 126:
+      return {8, 5, 0};
+    case 128:
+      return {8, 7, 0};
+    case 129:
+      return {8, 8, 0};
+    case 130:
+      return {9, 0, 0};
+    default:
+      LOG(FATAL) << "Unsupported CUDA version: " << cuda_version;
```
Artem-B: Nit: Printing major.minor would look a bit nicer.

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
dimvar: done

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
```diff
@@ -33,23 +33,26 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
...
+    case 126:
+      return {8, 5, 0};
+    case 128:
+      return {8, 7, 0};
+    case 129:
+      return {8, 8, 0};
+    case 130:
+      return {9, 0, 0};
+    default:
+      // Return maximum known PTX version.
```
Artem-B: I think we still need a version check. As the unknown version may be CUDA 11.8 or any other older, unenumerated version. For those we should have an assertion, so such errors are caught at build time.

Also, the enumerated version list should probably be limited to the CUDA SDK versions supported by XLA. I'm not sure if  CUDA-11 is supported any more. Even if it is, it's likely limited to cuda-11.8, which introduced sm_90.

dimvar: Can we keep the behavior unrelated to this patch unchanged, to avoid score creep?

Before CUDA 11, the PR keeps the same behavior as before:
https://github.com/openxla/xla/blob/55e120e80374007e0298320da3690ad61a379b7e/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc#L28

Same for CUDA 11.0 and 11.1:
https://github.com/openxla/xla/blob/55e120e80374007e0298320da3690ad61a379b7e/xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc#L37

If you'd like to change that, let's do it in a separate patch that doesn't delay this one. WDYT?

Artem-B: I'm all for keeping the existing behavior. The problem with the current state of the patch is that you've implemented enumeration of the known versions based on the *comments* in the code, not on the actual code behavior. If the patch had enumerated minor versions of cuda-11 and 12, and added the checks on upper/lower boundary in the default case, the existing behavior would be preserved.

If you want to revert to the original state of the code, based on calculating the PTX version based on specific CUDA version, I'm fine with that, as long as the code is reasonably readable. The first iteration pushed things beyond the point of being reasonable for such a straightforward version map helper. If you want to go that route, just add a simple `if (cuda >= 13) return ptx-9.0` and be done with it.

dimvar: > If the patch had enumerated minor versions of cuda-11 and 12

Good catch, thanks. I'll revert to the IF statement version.

> added the checks on upper/lower boundary in the default case, the existing behavior would be preserved.

I don't think that would be needed because the lower boundary is checked in the beginning, and there is no upper boundary; beyond the latest case it's just the max known version. (It doesn't matter though since I'll remove the switch. Just pointing this out.)

I'll do one more pass and ping you.

Artem-B: >   and there is no upper boundary; beyond the latest case it's just the max known version. 

Yet. But we do know we will see a new CUDA version eventually. Considering that that new CUDA version will continue to work with PTX-9.0, there's no point erroring out.

dimvar: PTAL

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
```diff
@@ -33,20 +33,18 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
...
+    // Examples:
+    // CUDA 12.6 -> PTX 8.5
+    // CUDA 12.9 -> PTX 8.8
     return {cuda_version.major() - 4, cuda_version.minor() - 1, 0};
+  } else if (cuda_version < stream_executor::SemanticVersion{13, 0, 0}) {
+    // Examples:
+    // CUDA 11.0 -> PTX 7.0
+    // CUDA 12.4 -> PTX 8.4
+    // CUDA 13.0 -> PTX 9.0
+    return {cuda_version.major() - 4, cuda_version.minor(), 0};
```
Artem-B: Almost there. Except now the condition *excludes* cuda-13.0, yet the comment explicitly lists cuda-13.0 as the example of the versions supposedly handled by the branch.

If the intent is to handle known versions that have the same version offset, then we should be comparing with `< {13,1,0}`.

We do happen to return 9.0 for cuda-13.0 below, but I really want to see the code do what it says without forcing me to stop and think what's going on here and whether the mismatch is intentional or accidental.

dimvar: Sorry, typo! Fixed.

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
```diff
@@ -33,20 +33,18 @@ DetermineHighestSupportedPtxVersionFromCudaVersion(
...
-  // CUDA 12.8 -> PTX 8.7
-  // CUDA 12.9 -> PTX 8.8
-  if (cuda_version < stream_executor::SemanticVersion{12, 10, 0}) {
+  if (cuda_version >= stream_executor::SemanticVersion{12, 6, 0} &&
+      cuda_version < stream_executor::SemanticVersion{12, 10, 0}) {
+    // Examples:
+    // CUDA 12.6 -> PTX 8.5
+    // CUDA 12.9 -> PTX 8.8
     return {cuda_version.major() - 4, cuda_version.minor() - 1, 0};
+  } else if (cuda_version <= stream_executor::SemanticVersion{13, 0, 0}) {
```
Artem-B: I hate to keep harping on the edge cases, but the reason I suggested comparing upper end of the range va vs `3.1.0` is that we do want to cover patch releases of cuda-13.0, too. 
As written, `cuda_version <= stream_executor::SemanticVersion{13, 0, 0}` will be false for cuda-`13.0.1` and that's not what we want. I'd even take comparison vs `3.0.1` which would be unnecessarily strict (I don't think NVIDIA ever bumped PTX version in a patch release), but would correctly check for all currently known versions of CUDA-13.

Artem-B (approved):
LGTM modulo the edge case in cuda-13 patch release handling.

xla/service/gpu/llvm_gpu_backend/ptx_version_util.cc
dimvar: Done. Thanks for being thorough.

PR #32217, opened 2025-10-03
[Refactor] Completely Remove AsyncStreamKind

chaserileyroberts:
Coauthored with @terryysun 

📝 Summary of Changes

In this PR, we completely remove the AsyncStreamKind type and all of its uses. Instead, we rely only on using `ExecutionStreamId`s to determine op stream placement. In the places where `AsyncStreamKind` were explicitly used (such as in the send/recv ops), we did our best to replicate the behavior with `GetStreamIdOverride`. 

🎯 Justification

This is a continuation of the multi stream collective work.

Relevant design docs
https://docs.google.com/document/d/1i1IzE-ycaWmWYN_0B-lxDwk2WO4CRs9huxVz0YAGgMU/edit?usp=sharing
https://docs.google.com/document/d/1Nxwr9j92yHx--ryBqFkQk2lW-Jx6_2ejjVD7XponOvI/edit?usp=sharing 

🚀 Kind of Contribution: ♻️ Cleanup

This is a non functional change, so no benchmarks are expected to be impacted. In our next follow up PR, we will enable overlapped collective scheduling, which will then have performance impact. This was broken out to reduce the size of this PR.

🧪 Unit Tests:
We removed dead tests in `gpu_clique_key_test.cc`. Updated test in `command_buffer_conversion_pass_test.cc`.

🧪 Execution Tests:
We added new e2e execution tests in `collective_ops_e2e_test.cc`

### DIFF
--- a/xla/backends/gpu/runtime/all_to_all_thunk.cc
+++ b/xla/backends/gpu/runtime/all_to_all_thunk.cc
@@ -54,10 +54,10 @@ limitations under the License.
 #include "xla/stream_executor/stream.h"
 #include "xla/tsl/platform/errors.h"
 #include "xla/tsl/platform/logging.h"
+#include "xla/tsl/platform/status_macros.h"
 #include "xla/tsl/platform/statusor.h"
 #include "xla/util.h"
 #include "xla/xla_data.pb.h"
-#include "xla/tsl/platform/status_macros.h"
 
 namespace xla {
 namespace gpu {
@@ -83,7 +83,7 @@ AllToAllStartThunk::AllToAllStartThunk(
     const AllToAllConfig& config, std::vector<CollectiveThunk::Buffer> buffers,
     bool p2p_memcpy_enabled)
     : CollectiveThunk(Thunk::kAllToAllStart, thunk_info, async_events,
-                      AsyncStreamKind::ASYNC_STREAM_KIND_COLLECTIVE),
+                      p2p_memcpy_enabled),
       config_(config),
       buffers_(std::move(buffers)),
       p2p_memcpy_enabled_(p2p_memcpy_enabled) {
@@ -134,12 +134,10 @@ absl::Status AllToAllStartThunk::Initialize(const InitializeParams& params) {
       << "Local device count : " << device_count_;
 
   if (is_local() && p2p_memcpy_enabled_) {
-    AsyncStreamKind stream_kind = GetAsyncStreamKind();
-
     TF_ASSIGN_OR_RETURN(
         GpuCliqueKey clique_key,
         GetGpuCliqueKey(*params.collective_params, config().replica_groups,
-                        config().group_mode, stream_kind));
+                        config().group_mode, p2p_memcpy_enabled_));
 
     TF_ASSIGN_OR_RETURN(
         Communicator * comm,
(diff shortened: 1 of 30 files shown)


frgossen (changes requested):
Thanks for cleaning this up. I am almost sure, as is, this change will break pipeline parallelism in Maxtext. Can you run a pipeline parallelism config and check that the overlap is still guaranteed?

frgossen:
So this will only show in perf. Can you add some step time comparison with and w/o this patch for a maxtext pipeline parallelism config?

xla/backends/gpu/runtime/all_to_all_thunk.cc
```diff
@@ -246,11 +244,13 @@ absl::StatusOr<bool> AllToAllStartThunk::RunCollective(
   return true;
 }
 
-AsyncStreamKind AllToAllStartThunk::GetAsyncStreamKind() const {
+const std::optional<ExecutionStreamId> AllToAllStartThunk::GetStreamIdOverride()
+    const {
   if (is_local() && p2p_memcpy_enabled_) {
-    return AsyncStreamKind::ASYNC_STREAM_KIND_MEMCPYP2P;
+    // TODO: Remove this specialized value when it is safe to do so.
+    return ExecutionStreamId(3);
```
felixwqp: Can you share how these magic numbers are determined?  I guess it comes from `ASYNC_STREAM_KIND_MEMCPYP2P`? 

Would it make sense to name them with a `enum: uint8` with a more meaningful name?

terryysun: Yes it's from the ASYNC_STREAM_KIND_MEMCPYP2P. Part of the goal of this PR is to get rid of the enum so we don't plan to add new ones. This is temporary to maintain the existing behavior, will be deleted soon after e2e testing of round-robin assignment.

felixwqp (approved):
I will temporarily approve this PR to create a local change, to verify the behavior locally.

In parallel, please raise the issue of pipeline parallelism on head. We should target submit this pr and ensure pipeline parallelism functionality together.

chaserileyroberts:
Here is that bug https://github.com/AI-Hypercomputer/maxtext/issues/2739

felixwqp (approved):
Re-approval to trigger the internal PR generation, so as to verify the pipeline parallelism functionality from HLO level.

felixwqp (approved):

xla/backends/gpu/runtime/collective_group_thunk.h
```diff
@@ -55,8 +55,8 @@ class CollectiveGroupThunk : public Thunk {
 
  private:
   ThunkSequence thunks_;
-  AsyncStreamKind stream_kind_;
   std::shared_ptr<CollectiveThunk::AsyncEvents> async_events_;
+  bool is_p2p_;
```
felixwqp: From `ClangTidy`: private field 'is_p2p_' is not used, 

can you fix?

xla/backends/gpu/runtime/collective_thunk.h
```diff
@@ -168,12 +157,14 @@ class CollectiveThunk : public Thunk {
                                              Communicator& comm) = 0;
 
   virtual const CollectiveConfig& config() const = 0;
-  virtual AsyncStreamKind GetAsyncStreamKind() const { return stream_kind_; }
   virtual CollectiveStreamId GetAsyncStreamId() const { return stream_id_; }
+  virtual const std::optional<ExecutionStreamId> GetStreamIdOverride() const {
```
felixwqp: From ClangTidy: `return type 'const std::optional<ExecutionStreamId>' (aka 'const std::optional< ::tsl::gtl::IntType<ExecutionStreamId_tag_, unsigned long>>') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness`

can you please fix?

chaserileyroberts: Fixed.

xla/backends/gpu/runtime/collective_group_thunk.h
chaserileyroberts: Fixed.

chaserileyroberts:
@felixwqp This should be ready for a final pass.

felixwqp:
There is a conflict in gpu/runtime/collective_thunk.cc#47,48, could you rebase? 

Thank you!

PR #32226, opened 2025-10-03
Relax the error spec to make BitcastReduceWithStride1Tiling pass on Spark.

dimvar:


xla/backends/gpu/codegen/triton/fusion_emitter_device_test.cc
```diff
@@ -782,7 +782,7 @@ CHECK-COUNT-1:  triton_xla.extract
 CHECK: tt.reduce
 CHECK-COUNT-2:  triton_xla.insert
 )"));
-  EXPECT_TRUE(RunAndCompareNoHloPasses(kHloText, kExactMatch));
+  EXPECT_TRUE(RunAndCompareNoHloPasses(kHloText, ErrorSpec{/*aabs=*/1e-4, /*arel=*/1e-4}));
```
bchetioui: This looks reasonable---the test was too strict---but could still hide a bug.

Instead of changing the tolerance, WDYT if we make the reducer a `maximum` (and the init constant a `-inf`)? That way, if the results are not bitwise equivalent, this is a clear bug.

bchetioui (changes requested):

xla/backends/gpu/codegen/triton/fusion_emitter_device_test.cc
dimvar: Done. Thanks for the suggestion.

dimvar:
@bchetioui PTAL

bchetioui (approved):
Thank you @dimvar for the change! (And the ping, this slipped my mind.)

PR #32229, opened 2025-10-04
Changes to the compute capabilities to account for the two different Blackwell Edge GPUs.

dimvar:
RTX PRO 6000 has CC 12.0.
Spark has CC 12.1.

Removed the IsAtLeastBlackwellPro method because there is no guarantee that future data center GPUs will have CC higher than 12.0.

Also skipped the latency estimator test on Edge GPUs because it uses the collective performance model and crashes here:
https://github.com/openxla/xla/blob/784702574ee3f2df06116e7f7e4b837150c8694a/xla/service/gpu/model/gpu_collective_performance_model.cc#L239

### DIFF
--- a/xla/service/gpu/model/analytical_latency_estimator_test.cc
+++ b/xla/service/gpu/model/analytical_latency_estimator_test.cc
@@ -119,6 +119,12 @@ TEST_F(AnalyticalLatencyHidingSchedulerTest, TestAnalyticalLatencyEstimator) {
       if (!c.IsAtLeast(se::CudaComputeCapability::kPascal)) {
         GTEST_SKIP() << "This test is for Pascal+ GPUs.";
       }
+      if (c.major == 12 && c.minor == 1) {
+        // Skip this test for Spark. Because of the AllReduce, the test uses
+        // gpu_collective_performance_model, which only makes sense in a
+        // datacenter network setting.
+        GTEST_SKIP() << "This test is for datacenter GPUs.";
+      }
     } else if (!std::is_same_v<stream_executor::RocmComputeCapability, cc>) {
       GTEST_SKIP() << "This test is for Pascal+ GPUs.";
     }
--- a/xla/stream_executor/cuda/cuda_compute_capability.h
+++ b/xla/stream_executor/cuda/cuda_compute_capability.h
@@ -70,7 +70,7 @@ struct CudaComputeCapability {
     kAmpere = 8,
     kHopper = 9,
     kBlackwell = 10,
-    kBlackwellPro = 12
+    kBlackwell_12 = 12
   };
 
   constexpr CudaComputeCapability() = default;
@@ -138,20 +138,6 @@ struct CudaComputeCapability {
                                  FeatureExtension::kForwardCompatibleFeatures};
   }
 
-  // Includes all GPUs with compute capability 12.x. When comparing with
-  // `IsAtLeast` this will true for all compute capabilities of 12.0 or higher.
-  constexpr static CudaComputeCapability BlackwellPro() {
-    return CudaComputeCapability{kBlackwellPro, 0, FeatureExtension::kNone};
-  }
-
-  // Includes all GPUs with compute capability 12.x. When comparing with
-  // `IsAtLeast` this will true for all 12.x compute capabilities but not for
-  // compute capabilities with a higher major version.
-  constexpr static CudaComputeCapability BlackwellProGenerationOnly() {
-    return CudaComputeCapability{kBlackwellPro, 0,
-                                 FeatureExtension::kForwardCompatibleFeatures};
-  }
-
   // Returns true if the compute capability is at least
   // `other_major.other_minor`. It is equivalent to
   // this->SupportsAllFeaturesOf(CudaComputeCapability{other_major,
@@ -183,10 +169,6 @@ struct CudaComputeCapability {
     return major >= CudaComputeCapabilities::kBlackwell;
   }
 
-  bool IsAtLeastBlackwellPro() const {
-    return major >= CudaComputeCapabilities::kBlackwellPro;
-  }
-
   bool IsPascal() const { return major == CudaComputeCapabilities::kPascal; }
 
   bool IsVolta() const { return major == CudaComputeCapabilities::kVolta; }
@@ -202,10 +184,6 @@ struct CudaComputeCapability {
 
   bool IsBlackwell() const {
     return major == CudaComputeCapabilities::kBlackwell;
-  }
-
-  bool IsBlackwellPro() const {
-    return major == CudaComputeCapabilities::kBlackwellPro;
   }
 
   // Returns true if a kernel compiled for compute capability `other` can be run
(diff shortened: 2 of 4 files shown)


xla/stream_executor/cuda/cuda_compute_capability.h
```diff
@@ -205,7 +187,15 @@ struct CudaComputeCapability {
...
-    return major == CudaComputeCapabilities::kBlackwellPro;
+    return major == CudaComputeCapabilities::kBlackwellEdge && minor == 0;
+  }
+
+  bool IsBlackwellSpark() const {
+    return major == CudaComputeCapabilities::kBlackwellEdge && minor == 1;
+  }
+
+  bool IsEdgeGpu() const {
+    return major == CudaComputeCapabilities::kBlackwellEdge;
```
Artem-B: > RTX PRO 6000 has CC 12.0.
> Spark has CC 12.1.

While the functions above blackwell refer to GPU architectures, these new functions appear to refer to GPU cards, so it's somewhat confusing. It may be OK for one-off things like DGX Spark if it's theonly product based on sm_121, but for sm_120 it's ambiguous, as according to https://developer.nvidia.com/cuda-gpus, `GeForce RTX 50**` are also sm_120

Given that we're checking for a compute capability, we should probably stick with the chip/architecture names and avoid commingling them with the GPU card names. Is there a canonical documentation on all those new GPU architecture variant naming? 

I personally would just stick with something as straightforward as `isBlackwell(enum variant)`, with variant enumbs being self-explanatory, someghing like `SM_10X`/`SM_12X`/`SM_121`. It's much easier to look up in the PTX docs than the "Pro/Edge/Spark" that are never mentioned there.

dimvar: (Note that I didn't introduce the IsBlackwellPro function, it was introduced in https://github.com/openxla/xla/pull/31545).

Nice catch on the ambiguity of the naming. I changed it to IsBlackwellRtx, which I think covers everything in that page.

IMO, we shouldn't do `isBlackwell(enum variant)`, because we already have `CudaComputeCapability`, and we'd introduce the new enum to express basically the same information, but only use it for Blackwell.

> Is there a canonical documentation on all those new GPU architecture variant naming?

The page you already linked to is the only one I know.

Artem-B: Naming -- one of the hardest unsolved problems in computer science. :-) 

> we shouldn't do isBlackwell(enum variant)

Fair enough. I don't have a strong preference on this, just speculating what those functions are intended to provide. We generally do not care about card variants, but do care about GPU variants/CC. If we already have helpers for checking compute capability, do we need product-name-based helpers at all? I'm mostly curious. If they are useful, I'm fine with that.

> The page you already linked to is the only one I know.

It tends to lag behind the GPU architecture variants exposed by CUDA SDK. That would be one more reason to stick with just CC/sm_XXX nomenclature, IMO.

dimvar: > do we need product-name-based helpers at all?

For the test I'm fixing, I just need "IsEdgeGpu".

I was assuming Google needs IsBlackwellPro somewhere, that's why it got added. But I'm happy to delete it and also IsBlackwellRtx if they're not needed yet.

xla/service/gpu/model/analytical_latency_estimator_test.cc
```diff
@@ -119,6 +119,12 @@ TEST_F(AnalyticalLatencyHidingSchedulerTest, TestAnalyticalLatencyEstimator) {
       if (!c.IsAtLeast(se::CudaComputeCapability::kPascal)) {
         GTEST_SKIP() << "This test is for Pascal+ GPUs.";
       }
+      if (c.IsEdgeGpu()) {
+        // Because of the AllReduce, the test uses
+        // gpu_collective_performance_model, which only makes sense in a
+        // datacenter network setting.
+        GTEST_SKIP() << "This test is for datacenter GPUs.";
```
Artem-B: I have to nitpick on the naming, again.

Based on the explanation message we're actually checking whether it's a datacenter GPU, yet the check itself is for something opaque called "EdgeGpu" that tells nothing to the reader of the code. I think that the condition check function should reflect the condition it checks for, not the marketing name of the device that happens to satisfy that condition. If we want to exclude all sm_12* GPUs, we should check for that. We already using `se::CudaComputeCapability` in this test, and I assume that's the standard practice in the rest of the XLA code, and I think the trivial check implemented by `isEdgeGpu` just obscures things without improving readability of the code.

dimvar: "Edge" is not a marketing term. It's widely adopted naming for non-cloud (i.e., non-datacenter) devices.
I'm fine checking for sm_12* here though. I don't have a strong preference.

Regarding the IsBlackwellPro and IsBlackwellRtx, are you OK with me deleting both?

dimvar: Also, I do need to keep kBlackwellEdge even if we delete the IsEdgeGPU function, to check for it in a few places where we switch on the compute capability.

Artem-B: > "Edge" is not a marketing term.

That's the part I find confusing. If it would be the marketing name referring to sm_120 GPUs, I'd sort of understand it's use here. However, if you use it in a sense of "not datacenter", then its use doesn't make much sense to me. By that logic, sm_61 and any other GPU used in consumer cards would fit that description. Also, what are we to do with RTX6000 PRO then? Is that a datacenter GPU or not? It is being deployed in datacenters, and it ships as a stand-alone card to the users. What about sm_60, sm_70 and sm_75? They are also used in datacenters, and in the cards you can use in a regular PC.

dimvar: I misspoke earlier. RTX6000 is for datacenter use cases. Only Spark (sm_121) is an edge device. There is also Jetson Thor (sm_110) which is an edge device, and I'll be making some changes in XLA related to that soon.

Of the older SM numbers that you mention, did XLA ever support the edge versions of those, or only the datacenter versions? I thought we'd only supported datacenter GPUs until now. (Things may work on edge devices, but we don't really test on them or have any CI.)

As far as this patch goes, how about I create a function `IsEdgeGpu(CC)`, which would currently only be true for Spark, but we will eventually add more?

Also, `CudaComputeCapabilities::kBlackwellPro` needs renaming since Spark uses sm_12x and is not one of the "PRO" cards. I can't think of a good name for it. Is `CudaComputeCapabilities::kBlackwell12` too awkward?

dimvar: Updates based on chat discussion.

* No IsEdgeGpu method. Skip the test by checking the CC in the test only.
* Rename kBlackwellPro to kBlackwell_12.

Fixed another test also to support Spark.

Artem-B (approved):

xla/service/gpu/model/analytical_latency_estimator_test.cc
Artem-B: The gist of the chat discussion:
What we ultimately needed is to give the non-hermetic tests a hint that some GPU variants are unlikely to be used in an environment expected by the test. The compiler itself does not have any particular need to distinguish between edge/datacenter devices.
There appears to be no canonical for the sm_12x Blackwell variant. "Pro" is ambiguous as sm_120 is used on "Pro" and RTX cards. So for lack of better name, "_12" would serve to distinguish sm_100 blackwell from sm_120 blackwell.

PR #32295, opened 2025-10-06
Cross-Host Transfer PjRt API Changes: Move CrossHostSendBuffers and CrossHostReceiveBuffers into PjRt API

rao-ashish:
📝 Summary of Changes
This PR is a sketch outlining potential modifications to the PjRt API to better support cross-process transfers on GPUs, intended for discussion with @emilyfertig and @mwhittaker.

A [previous PR](https://github.com/openxla/xla/pull/32074) had also proposed PjRt API modifications for cross-process transfers, but this PR takes a different approach which simplifies the API, and which would allow us to aggregate transfers with ncclGroup calls in the future.

🎯 Justification
The current implementation of cross-process device-puts on GPUs does not cache communicators for reuse across multiple transfers between the same sets of devices, and does not aggregate transfers of multiple arrays together with ncclGroup calls. This PR changes the PjRt API to make CrossHostSendBuffers and CrossHostReceiveBuffers virtual functions inside xla::PjRtClient, allowing for client-specific implementations. Follow-on PRs can build on these API changes to implement communicator caching and aggregating transfers with ncclGroup calls.

🚀 Kind of Contribution
Performance Improvement

This PR only makes API changes necessary for future performance improvements such as communicator caching and aggregating transfers with ncclGroup calls, so no benchmarks were ran for this PR. Follow-up PRs implementing performance improvements will include benchmarks.

🧪 Execution Tests:

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
@@ -2657,6 +2657,285 @@ TEST(StreamExecutorGpuClientTest, MultipleDeviceShareDmaMapping) {
   TF_EXPECT_OK(client->DmaUnmap(host_dma_ptr.get()));
 }
 
+TEST(StreamExecutorGpuClientTest, FailedCrossHostSendArgsSizeMismatch) {
+  // Create the client.
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<PjRtClient> client,
+                          GetStreamExecutorGpuClient(DefaultOptions()));
+
+  // Create a buffer to try to send.
+  std::vector<int32_t> data(256);
+  std::iota(data.begin(), data.end(), 1);
+
+  Shape shape = ShapeUtil::MakeShape(S32, {256});
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<PjRtBuffer> buffer,
+      client->BufferFromHostBuffer(
+          data.data(), shape.element_type(), shape.dimensions(),
+          /*byte_strides=*/std::nullopt,
+          PjRtClient::HostBufferSemantics::kImmutableOnlyDuringCall, nullptr,
+          *client->addressable_devices()[0]->default_memory_space(),
+          /*device_layout=*/nullptr));
+
+  // Try to send some data, giving an extra dst_global_device_id.
+  std::vector<Future<>> send_futures = client->CrossHostSendBuffers(
+      {buffer.get()}, {PjRtGlobalDeviceId(1), PjRtGlobalDeviceId(2)},
+      CrossHostTransferKey(0));
+
+  EXPECT_EQ(send_futures.size(), 1);
+
+  EXPECT_THAT(
+      send_futures[0].Await(),
+      absl_testing::StatusIs(
+          absl::StatusCode::kInternal,
+          ::testing::StrEq("CrossHostSendBuffers: dst_global_device_ids "
+                           "must be the same size as buffers.")));
+}
+
+TEST(StreamExecutorGpuClientTest, FailedCrossHostSendMoreThanOneBuffer) {
+  // Create the client.
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<PjRtClient> client,
+                          GetStreamExecutorGpuClient(DefaultOptions()));
+
+  // Create a buffer to try to send.
+  std::vector<int32_t> data(256);
+  std::iota(data.begin(), data.end(), 1);
+
+  Shape shape = ShapeUtil::MakeShape(S32, {256});
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<PjRtBuffer> buffer,
+      client->BufferFromHostBuffer(
+          data.data(), shape.element_type(), shape.dimensions(),
+          /*byte_strides=*/std::nullopt,
+          PjRtClient::HostBufferSemantics::kImmutableOnlyDuringCall, nullptr,
+          *client->addressable_devices()[0]->default_memory_space(),
+          /*device_layout=*/nullptr));
+
+  // Try to send some data, giving more than one buffer.
+  std::vector<Future<>> send_futures = client->CrossHostSendBuffers(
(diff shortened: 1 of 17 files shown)


thomasjoerg:
XLA PR rotation here: Sounds like this PR is a sketch for discussion, and hence, I am going to turn this into a draft PR. Please turn it back if you want to have it reviewed for merging. And please add some unit test coverage before doing so. Thank you!

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -2463,6 +2463,304 @@ TEST(StreamExecutorGpuClientTest, MultipleDeviceShareDmaMapping) {
...
+                              << sender_stdout << "\nsender stderr:\n"
+                              << sender_stderr;
+  EXPECT_EQ(receiver_status, 0) << "receiver stdout:\n"
+                                << receiver_stdout << "\nreceiver stderr:\n"
+                                << receiver_stderr;
+}
+
+absl::Status SuccessfulCrossHostTransferTestSenderBody() {
+  // Make sure only device 0 is visible.
+  tsl::setenv("CUDA_VISIBLE_DEVICES", "0", 1);
```
gspschmid: Can we use `options.allowed_devices = {options.node_id};` instead of setting `CUDA_VISIBLE_DEVICES`?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -2463,6 +2463,304 @@ TEST(StreamExecutorGpuClientTest, MultipleDeviceShareDmaMapping) {
...
+
+  EXPECT_EQ(sender_status, 0) << "sender stdout:\n"
+                              << sender_stdout << "\nsender stderr:\n"
+                              << sender_stderr;
+  EXPECT_EQ(receiver_status, 0) << "receiver stdout:\n"
+                                << receiver_stdout << "\nreceiver stderr:\n"
+                                << receiver_stderr;
+}
+
+absl::Status SuccessfulCrossHostTransferTestSenderBody() {
```
gspschmid: Maybe just have a single `SuccessfulCrossHostTransferTestBody(bool is_sender)` since `SenderBody` and `ReceiverBody` are pretty similar to begin with?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -2693,13 +2992,29 @@ INSTANTIATE_TEST_SUITE_P(
...
   std::string cache_dir;
   bool use_xla_computation = false;
+
+  // Variables used by SuccessfulCrossHostTransfer.
+  std::string cross_host_test_role;
+
   std::vector<tsl::Flag> flag_list = {
+      tsl::Flag("test_to_run", &test_to_run,
+                "Which multiprocess test to execute. Allowed values: '' (runs "
+                "all tests), 'ShardedAutotuningWorksHelper' or "
```
gspschmid, line 3371: all tests -> all single-process tests?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
rao-ashish: Thanks for taking a look! Changed the code to use `allowed_devices` instead of `CUDA_VISIBLE_DEVICES`.

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
rao-ashish: Got it, merged the two functions.

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
rao-ashish: Modified the wording slightly, but if `test_to_run` is empty, the code should end up calling `RUN_ALL_TESTS()`, which will execute multiprocess tests like `ShardedAutotuningWorks` and `SuccessfulCrossHostTransfer` in addition to the single-process ones.

xla/python/pjrt_ifrt/pjrt_client.cc
```diff
@@ -1337,25 +1337,43 @@ PjRtClient::CopyArraysForCrossHost(absl::Span<ArrayRef> arrays,
             dst_devices->devices()[i]->DebugString()));
       }
 
-      PjRtArray::PjRtBuffers send_buffers;
+      // Create vector of dst devices; we send each array to
+      // dst_devices->devices()[i].
+      TF_ASSIGN_OR_RETURN(
+          xla::PjRtGlobalDeviceId dst_global_device_id,
+          GetPjRtGlobalDeviceId(dst_devices->devices()[i]->Id()));
+      std::vector<PjRtGlobalDeviceId> dst_global_device_ids(
```
emilyfertig, line 1345: Does this have to be a vector? (Same with `src_global_device_ids` below)

rao-ashish: You're right, for this implementation of `CopyArraysForCrossHost`, since each array is being sent between the same src and dst devices, we could modify `CrossHostSendBuffers` / `CrossHostReceiveBuffers` to take in a single src / dst device id instead of a vector.

However, the motivation for `CrossHostSendBuffers` / `CrossHostReceiveBuffers` to work with vectors of src / dst devices rather than individual ids was to enable a future performance optimization. If we want to send two arrays between two potentially distinct (src, dst) GPUs, it would be beneficial to surround these transfers with ncclGroup calls to reduce kernel launch overheads and allowing NCCL to better optimize the transfers. With the proposed API, one could imagine a future implementation of `CopyArraysForCrossHost` that calls `CrossHostSendBuffers` / `CrossHostReceiveBuffers` only once for all arrays, and the cross host transfer methods would be able to aggregate all the transfers with ncclGroup calls.

emilyfertig:
@rao-ashish The overall plan for this looks good. Could you please make a version that adds the new APIs/tests, but does not remove any existing APIs, and omits the changes to PjRt-IFRT? Once that's merged, I'll implement the new APIs in the TPU backend, then we can merge the IFRT changes.

PR #32297, opened 2025-10-06
[XLA:GPU] Add host offloading support to collective pipeliner

sfvaroglu:
📝 Summary of Changes
1. Add host offloading support to collective pipeliner to enable memory optimization for collective operations by offloading intermediate results to host memory.

2. Add dynamic variable detection for CollectivePipeliner-transformed loops to enable FusionDynamicMemcpyRewriter optimization, allowing better memory management for dynamically shaped computations.

🎯 Justification
When host memory offloading is enabled, activations are copied asynchronously; device to host in the forward pass and back for the backward pass. These copies often happen at loop boundaries, limiting overlap with compute. This PR adds host offloading support to collective pipeliner so that we can move these copies to nearby iterations to reduce dependencies and improve overlap. This feature is enabled via `xla_gpu_enable_pipelined_host_offloading` flag. 

🚀 Kind of Contribution
⚡️ Performance Improvement, ✨ New Feature

The changes in this PR are gated by a flag. The HLOs under `xla/tools/benchmarks/hlo` don't affect offloading behavior, so we shouldn't expect any impact.

🧪 Unit Tests:
Added unit tests.

🧪 Execution Tests:
Saw ~12% speedup on GB200 with llama3-8b, fsdp=8 using MLP offloading plus collective pipelining, compared to just offloading.


xla/service/collective_pipeliner_test.cc
```diff
@@ -5281,5 +5282,145 @@ ENTRY entry {
   EXPECT_EQ(fusion_count, 4);
 }
 
+TEST_F(CollectivePipelinerTest, HostOffloading) {
+  constexpr absl::string_view hlo_string = R"(
+HloModule jit_scanned, entry_computation_layout={(f32[10,1000,8000]{2,1,0}, f32[10,8000,1000]{2,1,0}, f32[1000,1000]{1,0})->(f32[10,1000,8000]{2,1,0}, f32[10,8000,1000]{2,1,0})}, allow_spmd_sharding_propagation_to_parameters={true,true,true}, allow_spmd_sharding_propagation_to_output={true,true}
```
felixwqp: I didn't see any collective in the two-while loop example you shared here. Is this intended? 

If so, can you share some context why collective pipeliner is blocking the host offloading?

felixwqp:
Do you have a hlo for host offload to demonstrate this speed-up? Adding it into benchmark suite could guard host offloading features from regression for future development.

sfvaroglu:
> Do you have a hlo for host offload to demonstrate this speed-up? Adding it into benchmark suite could guard host offloading features from regression for future development.

Added. I got 1294739 us with this HLO on H100.

xla/service/collective_pipeliner_test.cc
felixwqp: Reminder about this comment.

sfvaroglu: Sorry, I missed this. I sent the design doc for host offloading pipelining in GChat.

xla/debug_options_flags.cc
```diff
@@ -2758,8 +2766,7 @@ FlagStatus GetFlagStatus(absl::string_view flag_name) {
           "xla_gpu_all_reduce_combine_threshold_bytes",
           "xla_gpu_autotune_level",
           "xla_gpu_collective_permute_decomposer_threshold",
-          "xla_gpu_cublas_fallback",
-          "xla_gpu_dot_merger_threshold_mb",
+          "xla_gpu_cublas_fallback", "xla_gpu_dot_merger_threshold_mb",
```
felixwqp, line 2878: is this change expected?

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +295,77 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
 
+bool IsHostOffloadingInstruction(const HloInstruction* instr) {
+  if (!instr) {
+    return false;
+  }
+
+  if (instr->IsCustomCall("MoveToHost")) {
+    // Check if MoveToHost is used by DynamicUpdateSlice (forward pattern)
+    std::vector<const HloInstruction*> to_check = {instr};
+    std::set<const HloInstruction*> visited;
```
felixwqp: recommend s/std::set/absl::btree_set; as inside google codebase, we have some workflow to keep updating data structure usage to more performant one, suggesting `absl::btree_set`

```
The Abseil container library contains ordered containers generally adhering to the STL container API contract, but implemented using (generally more efficient) B-trees rather than binary trees (as used in std::map et al):
```
ref: https://abseil.io/docs/cpp/guides/container#b-tree-ordered-containers

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +295,77 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+  } else if (instr->IsCustomCall("MoveToDevice")) {
+    // Check if MoveToDevice uses DynamicSlice (backward pattern)
+    std::vector<const HloInstruction*> to_check = {instr};
+    std::set<const HloInstruction*> visited;
+
+    while (!to_check.empty()) {
+      const HloInstruction* current = to_check.back();
+      to_check.pop_back();
+
+      if (visited.insert(current).second) {
```
felixwqp: for the sake of readability, can you check the inserted or not more explicitly

e.g. 

```
auto [_, inserted] = visited.insert(current);
 if (inserted) {

}
```

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +295,77 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+        for (const HloInstruction* operand : current->operands()) {
```
felixwqp: for the sake of readability, can you please check the existence of `current->operands()` before using it? e.g. 

```
if (!current->has_operands()) {
  continue;
}
```

xla/service/collective_pipeliner_test.cc
```diff
@@ -5463,5 +5464,145 @@ ENTRY entry {
...
+
+  std::vector<HloInstruction*> while_loops;
+  for (auto* instr : module->entry_computation()->instructions()) {
+    if (instr->opcode() == HloOpcode::kWhile) {
+      while_loops.push_back(instr);
+    }
+  }
+  ASSERT_EQ(while_loops.size(), 2) << "Expected 2 while loops in the module";
+
+  std::set<std::set<int64_t>> expected_dynamic_sets = {{5, 0}, {8, 0}};
```
felixwqp: can you educate me what you are trying to test here for host offloading in collective pipeliner? 

if my understanding is correct,  the purpose of collective pipeliner is to change, taking the forward pass as an example: 

from 
```
while (i < LAYERS) {
  p = param(0)
  x = computation(p)
  MoveToHost cutomcall(x)+DUS
}

```

into something like 

```
x_prev = computation(p)
i = i + 1
while (i < LAYERS, x_prev) {
  MoveToHost cutomcall(x_prev)+DUS
  p = param(0)
  x = computation(p)
  x_prev = x
}
 MoveToHost cutomcall(x_prev)+DUS

```


is my understanding correct here?

felixwqp (changes requested):

xla/service/collective_pipeliner_test.cc
sfvaroglu: Yes, correct. This test is checking that after pipelining, the two while loops are correctly annotated with dynamic variable indices - positions {5, 0} and {8, 0} respectively. These annotations tell the FusionDynamicMemcpyRewriter which tuple elements contain the pipelined buffers that get sliced/updated dynamically each iteration. The rewriter then finds the DS/DUS ops operating on these indices and converts them into optimized memcpy operations with precomputed offsets. Without correctly marking indices as dynamic, the rewriter wouldn't know where to look for the DS/DUS patterns that need optimization.

xla/debug_options_flags.cc
sfvaroglu: `clang-format` makes this change.

xla/service/collective_pipeliner_test.cc
```diff
@@ -5463,5 +5464,145 @@ ENTRY entry {
...
+}
+)";
+
+  auto module = ParseAndReturnUnverifiedModule(hlo_string, config_).value();
+
+  EXPECT_TRUE(
+      RunOptimizer(module.get(), /*last_run=*/false, 0,
+                   /*pipeline_use_tree=*/true,
+                   /*process_different_sized_ops=*/true,
+                   collective_pipeliner_utils::PipeliningDirection::kForward,
```
felixwqp: can you add the fields name here? e.g. /*direction*/, /*should_process*/?

xla/service/collective_pipeliner_test.cc
```diff
@@ -5463,5 +5464,145 @@ ENTRY entry {
...
+                   host_offload_utils::IsHostOffloadingInstruction)
+          .value());
+  XLA_VLOG_LINES(1, "After forward pipelining:\n" + module->ToString());
+
+  EXPECT_TRUE(
+      RunOptimizer(module.get(), /*last_run=*/true, 0,
+                   /*pipeline_use_tree=*/true,
+                   /*process_different_sized_ops=*/true,
+                   collective_pipeliner_utils::PipeliningDirection::kBackward,
```
felixwqp: same here.

xla/debug_options_flags.cc
```diff
@@ -2803,8 +2811,7 @@ FlagStatus GetFlagStatus(absl::string_view flag_name) {
...
```
felixwqp, line 2878: Looks this change is not necessary? can you revert?

felixwqp (commented):
Could you also add a unit test to demonstrate how "dynamic_variable_tuple_indices" config is used under FusionDynamicMemcpyRewriter? 

or will they be on the upcoming PR? if so, can you describe how you gonna to achive the rewrite via this config? 

thank you!

xla/service/gpu/ir_emission_utils.cc
```diff
@@ -667,6 +667,27 @@ bool IsInductionVariable(const HloInstruction* maybe_variable,
...
+    if (dynamic_idx == tuple_idx) {
+      return true;
+    }
+  }
+  return false;
+}
+
 // Attempts to find the induction variable of `loop` in `dependencies`. If there
 // are any dependencies on non-induction variable loop-carried variables,
 // returns nullopt.
```
felixwqp, line 683: Does the comment need update for dynamic variable?

xla/service/collective_pipeliner_test.cc
felixwqp: Thank you Sevin, 

I want to learn more about how {5,0},{8,0} come from.

e.g. in 1st while loop, 

```
%region_0.40 (arg_tuple.13: (s32[], f32[1000,1000], f32[10,1000,1000], f32[10,1000,8000], f32[10,8000,1000])) -> (s32[], f32[1000,1000], f32[10,1000,1000], f32[10,1000,8000], f32[10,8000,1000]) {
  %arg_tuple.13 = (s32[], f32[1000,1000]{1,0}, f32[10,1000,1000]{2,1,0}, f32[10,1000,8000]{2,1,0}, f32[10,8000,1000]{2,1,0}) parameter(0)
  %get-tuple-element.14 = s32[] get-tuple-element(%arg_tuple.13), index=0
   ... 
  %dynamic-update-slice.37 = f32[10,1000,1000]{2,1,0} dynamic-update-slice(%get-tuple-element.16, %reshape.36, %get-tuple-element.14, %constant.20, %constant.20)
  ROOT %tuple.39 = (s32[], f32[1000,1000]{1,0}, f32[10,1000,1000]{2,1,0}, f32[10,1000,8000]{2,1,0}, f32[10,8000,1000]{2,1,0}) tuple(%add.38, %dot.1, %dynamic-update-slice.37, %get-tuple-element.17, %get-tuple-element.18)
}
```
my understanding is only index 0 should be captured, e.g. `%get-tuple-element.14`, so where the 5 comes from here?

xla/service/collective_pipeliner_test.cc
```diff
@@ -5463,5 +5464,145 @@ ENTRY entry {
...
+  std::vector<HloInstruction*> while_loops;
+  for (auto* instr : module->entry_computation()->instructions()) {
+    if (instr->opcode() == HloOpcode::kWhile) {
+      while_loops.push_back(instr);
+    }
+  }
+  ASSERT_EQ(while_loops.size(), 2) << "Expected 2 while loops in the module";
+
+  std::set<std::set<int64_t>> expected_dynamic_sets = {{5, 0}, {8, 0}};
+  std::set<std::set<int64_t>> actual_dynamic_sets;
```
felixwqp: could you use absl container for consistency? thank you!

xla/service/collective_pipeliner_test.cc
sfvaroglu: After forward pipelining, it becomes:
```
I0000 00:00:1762803610.495774   16160 collective_pipeliner_test.cc:5571] %region_0.40.clone (loop_peel_param: (s32[], f32[1000,1000], f32[10,1000,1000], f32[10,1000,8000], f32[10,8000,1000], /*index=5*/s32[], f32[1000,1000])) -> (s32[], f32[1000,1000], f32[10,1000,1000], f32[10,1000,8000], f32[10,8000,1000], /*index=5*/s32[], f32[1000,1000]) {
I0000 00:00:1762803610.495775   16160 collective_pipeliner_test.cc:5571]   %loop_peel_param = (s32[], f32[1000,1000]{1,0}, f32[10,1000,1000]{2,1,0}, f32[10,1000,8000]{2,1,0}, f32[10,8000,1000]{2,1,0}, /*index=5*/s32[], f32[1000,1000]{1,0}) parameter(0)
I0000 00:00:1762803610.495776   16160 collective_pipeliner_test.cc:5571]   %get-tuple-element.6 = s32[] get-tuple-element(%loop_peel_param), index=0
...
I0000 00:00:1762803610.495796   16160 collective_pipeliner_test.cc:5571]   %get-tuple-element.11 = s32[] get-tuple-element(%loop_peel_param), index=5
I0000 00:00:1762803610.495798   16160 collective_pipeliner_test.cc:5571]   %custom-call.1 = s32[] custom-call(%get-tuple-element.11, %constant.7), custom_call_target="InsertedByPreviousStep"
I0000 00:00:1762803610.495800   16160 collective_pipeliner_test.cc:5571]   %dynamic-update-slice.1 = f32[10,1000,1000]{2,1,0} dynamic-update-slice(%get-tuple-element.10, %reshape.5, %custom-call.1, %constant.9, %constant.9)
}
```

Similar pattern for backward pipelining.

xla/debug_options_flags.cc
sfvaroglu: Clang format automatically does this. Otherwise, the PR fails format check.

xla/debug_options_flags.cc
felixwqp: Sorry I missed your original comments. feel free to ignore.

sfvaroglu:
> Could you also add a unit test to demonstrate how "dynamic_variable_tuple_indices" config is used under FusionDynamicMemcpyRewriter?

Added to `copy_test` as DynamicMemcpyFusion::GetMemcpyDescriptorForFusion is in copy.cc

bixia1 (approved):
I am approving this so that "the process" will create a copy inside google for me to play with. If you need to sync the tree and resolve conflict, please do help.
Once I got a copy inside google, I will remove the approval and I will start reviewing the PR.

bixia1:
I have a high level question: from the PR description, the benchmark performance data, it is not alway true that the runtimes are decreased, there are like 5 cases where the runtimes increase, an example 
llama31_70b_fp8_32x8 | 9710313 us | 10520533 us

What would you say about this?

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,85 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
   host_instruction.set_frontend_attributes(frontend_attributes);
 }
 
+bool IsHostOffloadingInstruction(const HloInstruction* instr) {
```
bixia1: Is it better to split this into two routines, this can help avoid duplicating part of the implementation here in collective_pipeliner.cc.

Suggested names:
    IsMovedToHostFollowedWithDusUsers
    IsMovedToDeviceWithDsOperands

Please also add unit tests to host_offload_utils_test.cc

sfvaroglu:
> I have a high level question: from the PR description, the benchmark performance data, it is not alway true that the runtimes are decreased, there are like 5 cases where the runtimes increase, an example llama31_70b_fp8_32x8 | 9710313 us | 10520533 us
> 
> What would you say about this?

That's just noise. The changes in this PR are gated by a flag, and the measurements were taken without that flag enabled. These HLOs don't affect offloading behavior, so we shouldn't expect any impact.

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,85 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+            to_check.push_back(operand);
+          }
+        }
+      }
+    }
+  }
+  return false;
+}
+
+bool HasHostOffloadingOperations(const HloComputation* computation) {
```
bixia1: I don't feel that you need this routine in host_offload_utils.cc, as there is currently only use place that uses it.
But if you really want to have this, would HasDusOrDsRelatedHostOffloadingOperations?

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,85 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+  if (!computation) {
+    return false;
+  }
```
bixia1: I don't think this check is necessary. In a valid HLO module, there is no places that store nullptr for computation?

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,85 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
   host_instruction.set_frontend_attributes(frontend_attributes);
 }
 
+bool IsHostOffloadingInstruction(const HloInstruction* instr) {
+  if (!instr) {
+    return false;
+  }
+
+  if (instr->IsCustomCall("MoveToHost")) {
```
bixia1: Please use memory_annotations::constant like you do in collective_pipeliner.cc

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,85 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
   host_instruction.set_frontend_attributes(frontend_attributes);
 }
 
+bool IsHostOffloadingInstruction(const HloInstruction* instr) {
+  if (!instr) {
```
bixia1: I don't think this check is necessary, in a valid HLO module, there is no places that store nullptr as an HloInstruction.

bixia1 (commented):
I have reviewed host_offload_utils.h/cc.
My suggestion is that to split these two files into a child PR (or just a new PR), and add unit tests to host_offload_utils_test.cc, so that we can move forward with submitting them.

bixia1:
> > I have a high level question: from the PR description, the benchmark performance data, it is not alway true that the runtimes are decreased, there are like 5 cases where the runtimes increase, an example llama31_70b_fp8_32x8 | 9710313 us | 10520533 us
> > What would you say about this?
> 
> That's just noise. The changes in this PR are gated by a flag, and the measurements were taken without that flag enabled. These HLOs don't affect offloading behavior, so we shouldn't expect any impact.

Thanks for the explanation. In that case, I would think replacing that block of benchmark performance data with somethink similar to what you have just said is better.

sfvaroglu:
> Thanks for the explanation. In that case, I would think replacing that block of benchmark performance data with somethink similar to what you have just said is better.

The benchmark data is required by the PR template (using default flags). I've also included an e2e case in the description showing the performance improvement. Happy to add more context if that would be helpful.

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+
+  std::vector<const HloInstruction*> to_check = {instr};
+  absl::btree_set<const HloInstruction*> visited;
+
+  while (!to_check.empty()) {
+    const HloInstruction* current = to_check.back();
+    to_check.pop_back();
+
+    auto [_, inserted] = visited.insert(current);
+    if (inserted) {
```
bixia1: We can reduce the indentation by
if (!inserted) {
  continue;
}

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
   host_instruction.set_frontend_attributes(frontend_attributes);
 }
 
+bool IsMoveToHostWithDynamicUpdateSlice(const HloInstruction* instr) {
+  if (!instr->IsCustomCall(kMoveToHostCustomCallTarget)) {
+    return false;
+  }
+
+  std::vector<const HloInstruction*> to_check = {instr};
+  absl::btree_set<const HloInstruction*> visited;
```
bixia1: It is more efficient to use absl::flat_hash_set.

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+
+  while (!to_check.empty()) {
+    const HloInstruction* current = to_check.back();
+    to_check.pop_back();
+
+    auto [_, inserted] = visited.insert(current);
+    if (inserted) {
+      if (current->user_count() == 0) {
```
bixia1: This check is not necessary, as the loop below will have 0 iteration in such a case.

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+        continue;
+      }
+      for (const HloInstruction* user : current->users()) {
+        if (user->opcode() == HloOpcode::kDynamicUpdateSlice) {
+          return true;
+        }
+        if (user->opcode() == HloOpcode::kReshape ||
+            user->opcode() == HloOpcode::kBroadcast ||
+            user->opcode() == HloOpcode::kTranspose) {
```
bixia1: Prefer to use if (HloPredicateIsOp<HloOpcode::KReshape, HloOpcode::kBroadcast, HloOpcode::kTranspose>(user)) {

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+  return false;
+}
+
+bool IsMoveToDeviceWithDynamicSlice(const HloInstruction* instr) {
+  if (!instr->IsCustomCall(kMoveToDeviceCustomCallTarget)) {
+    return false;
+  }
+
+  std::vector<const HloInstruction*> to_check = {instr};
+  absl::btree_set<const HloInstruction*> visited;
```
bixia1: Similar to above, please use absl::flat_hash_set instead.

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+
+  while (!to_check.empty()) {
+    const HloInstruction* current = to_check.back();
+    to_check.pop_back();
+
+    auto [_, inserted] = visited.insert(current);
+    if (inserted) {
```
bixia1: Similar to the above, use
if (!inserted) {
  condinue;
}

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+      if (current->operand_count() == 0) {
+        continue;
+      }
```
bixia1: Similar to the above, not need for this check.

xla/service/host_offload_utils.cc
```diff
@@ -294,5 +296,69 @@ void SetHostComputeFrontendAttribute(HloInstruction& host_instruction) {
...
+      for (const HloInstruction* operand : current->operands()) {
+        if (operand->opcode() == HloOpcode::kDynamicSlice) {
+          return true;
+        }
+        if (operand->opcode() == HloOpcode::kReshape ||
+            operand->opcode() == HloOpcode::kBroadcast ||
+            operand->opcode() == HloOpcode::kTranspose) {
```
bixia1: Also, use HloPredicateIsOp<>() instead.

bixia1 (commented):
Would you please split host_offload_utils* to a PR, we can move forward with submitting that PR.
Smaller PRs are preferred for many reasons, such as easy submission, we can still keep this big PR for full context of the work.

sfvaroglu:
> Would you please split host_offload_utils* to a PR, we can move forward with submitting that PR. Smaller PRs are preferred for many reasons, such as easy submission, we can still keep this big PR for full context of the work.

Created https://github.com/openxla/xla/pull/34118.

xla/service/gpu/ir_emission_utils.cc
```diff
@@ -659,6 +659,27 @@ bool IsInductionVariable(const HloInstruction* maybe_variable,
          maybe_variable->tuple_index() == loop.induction_variable_index;
 }
 
+// Returns true if `maybe_variable` is marked as a dynamic variable.
+bool IsDynamicVariable(const HloInstruction* maybe_variable,
```
bixia1: (1) When we collect dynamic_variable_tuple_indices , we make sure it is a set on indices from GTE that are "connected" to DS/DUS (through copy or arithmetic ops) inside a loop.
In the routine here,  "maybe_variable" is a GTE and it is inside a loop. Do we still need to make sure it is "connected" to DS/DUS (through copy or arithmetic ops) ?

(2) on variable naming: can we just name the instruction "gte", or "variable". The "maybe_" part of the name doesn't contribute much to help understand the code.

xla/service/gpu/ir_emission_utils.cc
```diff
@@ -659,6 +659,27 @@ bool IsInductionVariable(const HloInstruction* maybe_variable,
...
+                       const VerifiedLoop& loop) {
+  if (!IsLoopCarriedVariable(maybe_variable, loop)) {
+    return false;
+  }
```
bixia1: This check is not needed if you swap the branch ordering around the callsite of this routine, that is, move this branch 
if (IsLoopCarriedVariable(gte, loop)) {...} ahead.

xla/service/collective_pipeliner.cc
```diff
@@ -1631,6 +1663,88 @@ HloInstruction* CreateZero(HloComputation* comp, const Shape& shape,
...
+      indices.insert(op_indices.begin(), op_indices.end());
+    }
+  }
+
+  return indices;
+}
+
+// Scans while loop body for DS/DUS, traces their index operands back to GTEs
+// and marks corresponding tuple indices as dynamic variables.
+absl::Status MarkDynamicVariables(HloInstruction* while_loop) {
```
bixia1: Is it true that dynamic_variable_tuple_indices is only used in ir_emission?
If that is the case, what would be the problem if we move the functionality of MarkDynamicVariables to ir_emission, and maybe avoid using WhileLoopBackendConfig?

If we cann't avoid using WhileLoopBackendConfig and can't move it to ir_emission, maybe you can consider using this field in the CollectivePipeliner config "HloPostprocessor postprocess_backward_peeled_trailing_op"?

xla/service/collective_pipeliner.cc
```diff
@@ -1631,6 +1663,88 @@ HloInstruction* CreateZero(HloComputation* comp, const Shape& shape,
...
+    return absl::OkStatus();
+  }
+
+  WhileLoopBackendConfig config;
+  TF_ASSIGN_OR_RETURN(config,
+                      while_loop->backend_config<WhileLoopBackendConfig>());
+
+  config.clear_dynamic_variable_tuple_indices();
+
+  absl::flat_hash_set<int64_t> dynamic_slice_indices;
```
bixia1: flat_hash_set doesn't preserve insertion order and if we use it here, it will become a source of 
indeterministic HLO. Maybe we should use std::set?

xla/hlo/transforms/while_loop_trip_count_annotator.cc
```diff
@@ -44,6 +44,13 @@ absl::StatusOr<bool> WhileLoopTripCountAnnotator::RunImpl(
         // The following analyses all need the induction variable index.
         WhileLoopBackendConfig config;
 
+        // Preserve existing backend config data
+        if (auto existing_config =
+                instr->backend_config<WhileLoopBackendConfig>();
+            existing_config.ok()) {
+          config = *existing_config;
```
bixia1, line 51: (1) Shall we check that the config doesn't already have the information that we intend to collect in the current pass?
(2) This change should go to a separated PR and can be submitted before this parent PR.

xla/service/gpu/transforms/double_buffer_loop_unrolling.cc
```diff
@@ -374,8 +374,18 @@ absl::StatusOr<bool> FullyUnroll(HloInstruction* while_instr,
...
+  TF_ASSIGN_OR_RETURN(old_config,
+                      while_instr->backend_config<WhileLoopBackendConfig>());
+
   WhileLoopBackendConfig new_config;
   new_config.mutable_known_trip_count()->set_n(1);
+
+  // Preserve dynamic_variable_tuple_indices from original config
+  for (int64_t idx : old_config.dynamic_variable_tuple_indices()) {
+    new_config.add_dynamic_variable_tuple_indices(idx);
+  }
```
bixia1: Following this logic, each time we will add a new field to WhileLoopBackendConfig we will need to come to here to copy the field?

I think a better way is to copy over the old_config, without having to know what is inside, then clear up the fields that the double_buffer_loop unroller need to touch.

xla/service/collective_pipeliner.cc
```diff
@@ -527,6 +530,35 @@ std::optional<std::vector<HloInstruction*>> CollectIndependentOperandChain(
...
+        if (current->opcode() == HloOpcode::kReshape ||
+            current->opcode() == HloOpcode::kBroadcast ||
+            current->opcode() == HloOpcode::kTranspose) {
+          for (HloInstruction* operand : current->operands()) {
+            to_check.push_back(operand);
+          }
+        }
+      }
+    }
+  }
```
bixia1: We want to not hide device specific detail from this pass.
Is it possible to add a field to CollectivePipeline::Config, something similar this existing field "HloPredicate acceptable_formatting"?

xla/service/collective_pipeliner_test.cc
```diff
@@ -5465,5 +5466,150 @@ ENTRY entry {
   EXPECT_EQ(fusion_count, 4);
 }
 
+TEST_F(CollectivePipelinerTest, HostOffloading) {
```
bixia1: Can you split the tests? For example DS and DUS don't have to be in the same test, also, avoid having the test to perform transformation on a program produced by a previous transformation.

No need for layout in the HLO

xla/service/collective_pipeliner.cc
sfvaroglu: This function is post-transformation metadata annotation that captures what CollectivePipeliner created, rather than a standalone optimization. When pipelining host offloading, CollectivePipeliner changes loop structure and creates new DS/DUS patterns. `MarkDynamicVariables` annotates which tuple indices became dynamic variables as a result of these transformations. It runs only when `changed=true` as part of the cleanup phase, and the metadata flows from CollectivePipeliner → FusionDynamicMemcpyRewriter.

xla/service/gpu/ir_emission_utils.cc
sfvaroglu: No, we don't need an additional connection check in `IsDynamicVariable`. The `dynamic_variable_tuple_indices` set is already pre-filtered during collection to only include indices that are connected to DS/DUS through arithmetic/copy ops (via `FindTupleIndicesInOperand`). `IsDynamicVariable` is just checking membership in this already-validated set. Re-checking the connection would be redundant.

xla/hlo/transforms/while_loop_trip_count_annotator.cc
sfvaroglu: This pass doesn't recompute `dynamic_variable_tuple_indices`, so we preserve the existing config.

xla/service/collective_pipeliner.cc
bixia1: Then maybe we can use this field in the CollectivePipeliner config "HloPostprocessor postprocess_backward_peeled_trailing_op" and implement the marking in the functor?

xla/service/gpu/ir_emission_utils.cc
bixia1: Can we add tests to gpu/ir_emission_utils_test.cc? In particular, include a test with irrelevant GET that don't correct DUS/DS whose index matches the meta data?

xla/service/collective_pipeliner.cc
sfvaroglu: `postprocess_backward_peeled_trailing_op` isn't suitable because it's called per-operation rather than per-loop, and doesn't have access to the `dynamic_slice_indices` set collected during the pass. Marking dynamic variables requires updating the `WhileLoopBackendConfig` once per loop after all indices are collected.

xla/hlo/transforms/while_loop_trip_count_annotator.cc
bixia1: Can you split this to a new PR and we can submit this change regardless the status of this parent PR.

xla/service/collective_pipeliner.cc
bixia1: I don't understand what you said. Here routine MarkDynamicVariables(HloInstruction* while_loop) is called for each while_loop, and when postprocess_backward_peeled_trailing_op is called, a transformed while_loop is passed to it. We only need to check the while_loop that is transformed by the pipeliner, is this correct?
If we don't come to an agreement on this, maybe we can set up a GVC to discuss about this. Please message me on chat for this.

xla/service/collective_pipeliner.cc
```diff
@@ -527,6 +531,33 @@ std::optional<std::vector<HloInstruction*>> CollectIndependentOperandChain(
...
```
bixia1: We want to move this block of code to the function provided by config, so that when we call the function with a DS we want to add to the chain, the function returns true.

This is similar to how CheckStoreIntoSliceIsCompatible use config.acceptable_formatting for the forward pass. We may just use this same field acceptable_formatting for the backward pass?

bixia1 (commented):
Did you remove the benchmark you add to this PR? we actually want to have such benchmarks.

sfvaroglu:
> Did you remove the benchmark you add to this PR? we actually want to have such benchmarks.

Merged as a separate PR https://github.com/openxla/xla/pull/34335

xla/service/collective_pipeliner.cc
sfvaroglu: Done

xla/service/collective_pipeliner.cc
sfvaroglu: Updated per our chat conversation.

xla/hlo/transforms/while_loop_trip_count_annotator.cc
bixia1: Sorry, I couldn't figure out what is your response to this request, to split this file into another PR.

xla/service/collective_pipeliner.cc
```diff
@@ -527,6 +533,14 @@ std::optional<std::vector<HloInstruction*>> CollectIndependentOperandChain(
         return !IsLoopIterator(instr, loop_iter) &&
                !loop_invariant_params.count(instr);
       };
+
+  if (find_dynamic_slice_operand) {
+    auto maybe_dynamic_slice = find_dynamic_slice_operand(instr, visited_set);
```
bixia1: (1) In your earlier code, the visited set is only internal to the code that find DS, can you explain why you add that to the function pointer?
(2) Shall we rename find_dynamic_slice_operand in the config to something more general? Something like additional_chain_start_op?

xla/hlo/transforms/while_loop_trip_count_annotator.cc
sfvaroglu: Sorry, I missed this. Here: https://github.com/openxla/xla/pull/34707

xla/service/collective_pipeliner.cc
sfvaroglu: (1) To avoid redundant traversal
(2) Done

xla/service/collective_pipeliner.cc
```diff
@@ -527,6 +533,15 @@ std::optional<std::vector<HloInstruction*>> CollectIndependentOperandChain(
         return !IsLoopIterator(instr, loop_iter) &&
                !loop_invariant_params.count(instr);
       };
+
+  if (additional_chain_start_op_finder) {
+    auto maybe_additional_op =
+        additional_chain_start_op_finder(instr, visited_set);
```
bixia1: Can we make the function returning a visited_set without passing a set into the function?

sfvaroglu: We could return the set instead, but passing by reference is better here. The function adds just one instruction to visited_set, so returning and merging a whole set would be wasteful. Also, this function mutates shared state used by the main algorithm, pass-by-reference makes this explicit and matches the original inline code behavior.

Let me know if this is what you prefer:
```
// Current
auto maybe_op = finder(instr, visited_set);

// Alternative
auto [maybe_op, to_merge] = finder(instr);
visited_set.insert(to_merge.begin(), to_merge.end());
```

sfvaroglu: Updated

bixia1 (approved):

bixia1 (approved):

xla/service/gpu/gpu_compiler.cc
```diff
@@ -157,6 +159,7 @@ limitations under the License.
 #include "xla/service/collective_pipeliner_utils.h"
 #include "xla/service/collective_utils.h"
 #include "xla/service/compiler.h"
+#include "xla/service/memory_annotations.h"
```
thomasjoerg, line 163: Please add the new dep to the BUILD target too (xla/service/gpu:gpu_compiler)

thomasjoerg (changes requested):

xla/service/gpu/gpu_compiler.cc
sfvaroglu: Done

bixia1 (approved):

thomasjoerg (approved):

xla/xla.proto
```diff
@@ -549,6 +549,8 @@ message DebugOptions {
 
   optional bool xla_gpu_enable_pipelined_reduce_scatter = 231;
 
+  optional bool xla_gpu_enable_pipelined_host_offloading = 439;
```
bixia1: Can you move this line ahead, before this line so that they are in sorted order?
optional bool xla_gpu_enable_pipelined_p2p = 246;

bixia1 (changes requested):

xla/xla.proto
sfvaroglu: Done

bixia1 (approved):

xla/xla.proto
```diff
@@ -545,6 +545,8 @@ message DebugOptions {
 
   optional bool xla_gpu_enable_pipelined_collectives = 239 [deprecated = true];
 
+  optional bool xla_gpu_enable_pipelined_host_offloading = 439;
```
bixia1: Field number 439 has already been used in "xla.DebugOptions" by field "xla_allow_h2h_copy_when_automatic_host_compute_offload_disabled"

bixia1 (changes requested):

bixia1 (approved):

xla/service/collective_pipeliner.cc
```diff
@@ -42,6 +42,7 @@ limitations under the License.
 #include "absl/types/span.h"
 #include "xla/comparison_util.h"
 #include "xla/hlo/analysis/tuple_points_to_analysis.h"
+#include "xla/hlo/analysis/while_loop_analysis.h"
```
bixia1: this is not used anymore

xla/service/collective_pipeliner.cc
```diff
@@ -60,6 +61,8 @@ limitations under the License.
 #include "xla/service/collective_ops_utils.h"
 #include "xla/service/collective_pipeliner_utils.h"
 #include "xla/service/constant_value.h"
+#include "xla/service/host_offload_utils.h"
```
bixia1: this line is not needed anymore

xla/service/collective_pipeliner.cc
```diff
@@ -60,6 +61,8 @@ limitations under the License.
...
+#include "xla/service/memory_annotations.h"
```
bixia1: this line is not needed anymore

xla/service/collective_pipeliner_test.cc
```diff
@@ -5625,5 +5630,209 @@ ENTRY entry {
...
+
+  std::vector<HloInstruction*> while_loops;
+  for (auto* instr : module->entry_computation()->instructions()) {
+    if (instr->opcode() == HloOpcode::kWhile) {
+      while_loops.push_back(instr);
+    }
+  }
+  ASSERT_EQ(while_loops.size(), 1) << "Expected 1 while loop in the module";
+
+  VLOG(1) << "Transformed while body:\n"
```
bixia1: no header providing "VLOG" is directly included

xla/service/collective_pipeliner_test.cc
```diff
@@ -5625,5 +5630,209 @@ ENTRY entry {
...
+          << while_loops[0]->while_body()->ToString();
+
+  TF_ASSERT_OK_AND_ASSIGN(
+      WhileLoopBackendConfig config,
+      while_loops[0]->backend_config<WhileLoopBackendConfig>());
+
+  std::set<int64_t> dynamic_indices(
```
bixia1, line 5721: no header providing "std::set" is directly included

xla/service/gpu/transforms/double_buffer_loop_unrolling_test.cc
```diff
@@ -1498,6 +1497,94 @@ TEST_F(GpuLoopDoubleBufferTransformerTest, UpdateInitStepEvenTripCount) {
...
+
+  DoubleBufferLoopUnrolling double_buffer(
+      DoubleBufferLoopUnrolling::UnrollStrategy::kDoubleBuffer);
+  TupleSimplifier tuple_simplifier;
+
+  TF_ASSERT_OK_AND_ASSIGN(bool changed, double_buffer.Run(module.get()));
+  ASSERT_TRUE(changed);
+  TF_ASSERT_OK_AND_ASSIGN(changed, tuple_simplifier.Run(module.get()));
+
+  std::vector<HloInstruction*> while_loops;
```
bixia1, line 1550: no header providing "std::vector" is directly included

bixia1 (changes requested):

xla/service/BUILD
```diff
@@ -461,6 +461,8 @@ cc_library(
         ":collective_ops_utils",
         ":collective_pipeliner_utils",
         ":constant_value",
+        ":host_offload_utils",
```
bixia1: Please also add host_offload_utils to collective_pipeliner_test target

bixia1 (changes requested):

bixia1 (approved):

xla/service/gpu/transforms/double_buffer_loop_unrolling_test.cc
```diff
@@ -1498,6 +1498,94 @@ TEST_F(GpuLoopDoubleBufferTransformerTest, UpdateInitStepEvenTripCount) {
...
+
+  ASSERT_FALSE(while_loops.empty())
+      << "Expected at least one while loop after double buffering";
+
+  for (HloInstruction* while_loop : while_loops) {
+    TF_ASSERT_OK_AND_ASSIGN(
+        WhileLoopBackendConfig config,
+        while_loop->backend_config<WhileLoopBackendConfig>());
+
+    std::set<int64_t> dynamic_indices(
```
bixia1, line 1567: no header providing "std::set" is directly included

bixia1 (changes requested):

bixia1 (approved):

xla/service/host_offload_utils.cc
```diff
@@ -353,5 +353,88 @@ bool IsMoveToDeviceWithDynamicSlice(const HloInstruction* instr) {
...
+      has_host_offloading = true;
+      break;
+    }
+  }
+  if (!has_host_offloading) {
+    return absl::OkStatus();
+  }
+
+  WhileLoopBackendConfig config;
+  TF_ASSIGN_OR_RETURN(config,
```
bixia1, line 412: no header providing "TF_ASSIGN_OR_RETURN" is directly included

xla/service/host_offload_utils.cc
```diff
@@ -353,5 +353,88 @@ bool IsMoveToDeviceWithDynamicSlice(const HloInstruction* instr) {
...
+                      while_loop->backend_config<WhileLoopBackendConfig>());
+
+  config.clear_dynamic_variable_tuple_indices();
+
+  std::set<int64_t> dynamic_slice_indices;
```
bixia1, line 417: no header providing "std::set" is directly included

xla/service/host_offload_utils.cc
```diff
@@ -353,5 +353,88 @@ bool IsMoveToDeviceWithDynamicSlice(const HloInstruction* instr) {
...
+
+  for (auto* instr : while_loop->while_body()->instructions()) {
+    if (instr->opcode() == HloOpcode::kDynamicUpdateSlice ||
+        instr->opcode() == HloOpcode::kDynamicSlice) {
+      int first_index_operand =
+          (instr->opcode() == HloOpcode::kDynamicUpdateSlice)
+              ? Cast<HloDynamicUpdateSliceInstruction>(instr)
```
bixia1, line 424: no header providing "xla::Cast" is directly included

xla/service/host_offload_utils.cc
```diff
@@ -353,5 +353,88 @@ bool IsMoveToDeviceWithDynamicSlice(const HloInstruction* instr) {
...
+        dynamic_slice_indices.insert(op_indices.begin(), op_indices.end());
+      }
+    }
+  }
+
+  for (int64_t tuple_idx : dynamic_slice_indices) {
+    config.add_dynamic_variable_tuple_indices(tuple_idx);
+  }
+
+  TF_RETURN_IF_ERROR(while_loop->set_backend_config(config));
```
bixia1, line 441: o header providing "TF_RETURN_IF_ERROR" is directly included

bixia1 (changes requested):

xla/service/host_offload_utils.cc
sfvaroglu: Done

xla/service/host_offload_utils.cc
sfvaroglu: Done

xla/service/host_offload_utils.cc
sfvaroglu: Done

xla/service/host_offload_utils.cc
sfvaroglu: Done

PR #32357, opened 2025-10-07
Update gemma2 keras benchmark script - fix ttft, and use tokenizer

gaurides:
📝 Summary of Changes

1. Update calculation for TTFT to be the time to first generated token. This will also impact TPOT calculations.
2. Use tokenizer to count the number of tokens generated instead of counting words using space

🎯 Justification
Currently the script computes TTFT as time to first token which is from the prompt and still in prefill stage.

🚀 Kind of Contribution
🐛 Bug Fix

### DIFF
--- a/xla/backends/cpu/benchmarks/e2e/gemma2/keras/benchmark.py
+++ b/xla/backends/cpu/benchmarks/e2e/gemma2/keras/benchmark.py
@@ -74,9 +74,7 @@ def run(gemma_lm, tokenizer, max_len):
   output = gemma_lm.generate(_QUERY, max_length=total_output_len + 1)
   warmup_time = (time.time() - start) * 1000
   num_actual_output_tokens = len(tokenizer(output))
-  gen_tokens = 0
-  if num_actual_output_tokens > in_tokens_len:
-    gen_tokens = num_actual_output_tokens - in_tokens_len
+  gen_tokens = max(num_actual_output_tokens - in_tokens_len, 0)
 
   print(f"Warmup: Number of generated output tokens: ", gen_tokens)
 
@@ -92,9 +90,6 @@ def run(gemma_lm, tokenizer, max_len):
     elapsed_time = (time.time() - start) * 1000
     assert num_actual_output_tokens == len(tokenizer(output))
     times.append(elapsed_time)
-    gen_tokens = 0
-    if num_actual_output_tokens > in_tokens_len:
-      gen_tokens = num_actual_output_tokens - in_tokens_len
 
     if _VERBOSE:
       print("%d: %lf ms" % (i, elapsed_time))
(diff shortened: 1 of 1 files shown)


xla/backends/cpu/benchmarks/e2e/gemma2/keras/benchmark.py
```diff
@@ -51,21 +51,34 @@ def compute_stats(array):
...
   # Warm up.
   start = time.time()
-  output = gemma_lm.generate(_QUERY, max_length=max_len + 1)
-  num_actual_output_tokens = len(output.split(" "))
+  output = gemma_lm.generate(_QUERY, max_length=total_output_len + 1)
   warmup_time = (time.time() - start) * 1000
+  num_actual_output_tokens = len(tokenizer(output))
+  gen_tokens = 0
+  if num_actual_output_tokens > in_tokens_len:
+    gen_tokens = num_actual_output_tokens - in_tokens_len
```
penpornk: Nit: We can make this more concise
```suggestion
  gen_tokens = max(num_actual_output_tokens - in_tokens_len, 0)
```

xla/backends/cpu/benchmarks/e2e/gemma2/keras/benchmark.py
```diff
@@ -75,27 +88,34 @@ def run(gemma_lm, max_len):
...
     start = time.time()
-    output = gemma_lm.generate(_QUERY, max_length=max_len + 1)
-    assert num_actual_output_tokens == len(output.split(" "))
+    output = gemma_lm.generate(_QUERY, max_length=total_output_len + 1)
     elapsed_time = (time.time() - start) * 1000
+    assert num_actual_output_tokens == len(tokenizer(output))
     times.append(elapsed_time)
+    gen_tokens = 0
+    if num_actual_output_tokens > in_tokens_len:
+      gen_tokens = num_actual_output_tokens - in_tokens_len
```
penpornk: We don't need these lines since both `num_actual_output_tokens` and `in_tokens_len` are unchanged. (The assert in line 93 made sure `num_actual_output_tokens` stays the same.)

xla/backends/cpu/benchmarks/e2e/gemma2/keras/benchmark.py
```diff
@@ -75,27 +88,34 @@ def run(gemma_lm, max_len):
...
 
 
 def main():
   if _VERBOSE:
     print("Query: %s" % _QUERY)
 
   gemma_lm = keras_nlp.models.GemmaCausalLM.from_preset("gemma2_2b_en")
-  mean_1, diff_1, _ = run(gemma_lm, 1)
-  mean_n, diff_n, num_output_tokens = run(gemma_lm, _NUM_OUTPUT_TOKENS)
+  tokenizer = keras_nlp.models.GemmaTokenizer.from_preset("gemma2_2b_en")
```
penpornk: Maybe add a variable to store "gemma2_2b_en" and pass it for both `gemma_lm` and `tokenizer`?

penpornk (changes requested):
Thank you for the fix!

xla/backends/cpu/benchmarks/e2e/gemma2/keras/benchmark.py
gaurides: Shall we remove the assert instead? and i can make these lines concise as you suggested above. 
Because the generated output is not always same and num_actual_output_tokens can vary. It somehow works in this example with a small _NUM_OUTPUT_TOKENS = 30, but it doesn't when this number is changed.

gaurides:
@penpornk : thanks for the review. I addressed your comments.

penpornk (approved):
Thank you for the changes!

PR #32378, opened 2025-10-07
[XLA:CPU][oneDNN][BugFix] Fix unsafe subtraction from unsigned values

akhilgoe:
A recent change updated the method used to get the size of the dimensions of a `Shape` (from `dimensions_size()` to `dimensions().size()`) in `onednn_contraction_rewriter.cc`. The new method returns an unsigned value whereas the earlier one returned a signed integer. As a result, we need to make some modifications here to avoid any unsigned underflow.

### DIFF
--- a/xla/service/cpu/onednn_contraction_rewriter.cc
+++ b/xla/service/cpu/onednn_contraction_rewriter.cc
@@ -400,8 +400,11 @@ absl::StatusOr<Shape> AdjustAddendShape(const HloInstruction* contraction,
   // If rank(new_shape) > rank(instr), extra dimensions with value = 1 can be
   // deleted from the new_shape.
   auto instr_shape = contraction->shape();
-  int64_t rank_difference =
-      new_shape.dimensions().size() - instr_shape.dimensions().size();
+  int64_t rank_difference = 0;
+  if (new_shape.dimensions().size() > instr_shape.dimensions().size()) {
+    rank_difference =
+        new_shape.dimensions().size() - instr_shape.dimensions().size();
+  }
   auto new_dims = new_shape.dimensions();
   std::vector<int64_t> dims_to_delete;
   for (int i = 0; i < rank_difference; ++i) {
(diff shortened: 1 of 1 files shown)


xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -400,8 +400,11 @@ absl::StatusOr<Shape> AdjustAddendShape(const HloInstruction* contraction,
   // If rank(new_shape) > rank(instr), extra dimensions with value = 1 can be
   // deleted from the new_shape.
   auto instr_shape = contraction->shape();
-  int64_t rank_difference =
-      new_shape.dimensions().size() - instr_shape.dimensions().size();
+  int64_t rank_difference = 0;
+  if (new_shape.dimensions().size() > instr_shape.dimensions().size()) {
+    rank_difference =
+        new_shape.dimensions().size() - instr_shape.dimensions().size();
+  }
```
penpornk, line 410: Nit: This can be condensed to 
```suggestion
  int64_t rank_difference = max(0, new_shape.dimensions().size() - instr_shape.dimensions().size());
```

(Need reformatting.)

penpornk (changes requested):
Thank you for the fix!

xla/service/cpu/onednn_contraction_rewriter.cc
akhilgoe: Hi @penpornk, your suggestion might not work in case of an unsigned underflow. Because the operands are unsigned, the difference might loop around to a huge value.

akhilgoe:
@penpornk, a gentle ping.

xla/service/cpu/onednn_contraction_rewriter.cc
penpornk: Sorry for the delay! I see. Could you please add this comment in the code so that it won't be changed accidentally in the future?

akhilgoe: Done

akhilgoe:
Thanks @penpornk for the review! I have added a comment as suggested. Please take a look whenever you can.

penpornk (approved):
Thank you again for the PR!

PR #32389, opened 2025-10-08
[NVIDIA GPU] Change collective pipeliner to honor opt-barrier

Tixxx:
📝 Summary of Changes
The current collective pipeliners don't honor opt-barrier in their user/producer chain.
This pr changes the behavior to teach collective pipeliners to use acceptable_formatting function to filter out what cannot be pipelined.
🎯 Justification
Without this, opt-barriers connected with target collectives will be peeled and pipelined to other loop iterations which changes the semantic of the whole program.

🚀 Kind of Contribution
 🐛 Bug Fix

NA

🧪 Unit Tests:
Added unit tests

🧪 Execution Tests:
NA


frgossen (approved):
Awesome, thanks!

Tixxx:
can someone provide more info on the revert? thanks

PR #32454, opened 2025-10-09
Respect debug options override in LHS.

yliu120:
Any place inside the compiler has to respect the debug options override set by the users.


yliu120:
@frgossen please take a look. thx.

frgossen (approved):
Thanks!

PR #32521, opened 2025-10-10
[XLA:CPU][oneDNN] Update dot plus elementwise op fusion for oneDNN

yimeisun123:
Add constrains on fusing dot with elementwise ops to align with current OneDNN capabilities. 
Refactor onednn fusion test and onednn support code.


xla/backends/cpu/onednn_support.cc
```diff
@@ -15,57 +15,57 @@ limitations under the License.
...
+    return false;
   }
-
-  return false;
+  return (std::all_of(hlo->operands().begin(), hlo->operands().end(),
+                      [cpu_features](const HloInstruction* op) {
+                        return (IsOneDnnSupportedDType(
+                                    op->shape().element_type(), cpu_features) &&
+                                IsOneDnnSupportedLayout(op->shape()));
+                      }));
```
penpornk: Can we define the lambda as a variable and call it in the first `if` as well?

```
auto is_supported = [cpu_features](...){...};
if (!is_supported(hlo)) {
  return false
}
return std::all_of(...);
```

yimeisun123:
@penpornk , @ezhulenev , could you please let me know if you had a chance on reviewing this change and if any comment? Thanks

golechwierowicz:
Hey @penpornk is this PR still relevant?

xla/backends/cpu/transforms/library_rewriter.cc
```diff
@@ -261,13 +271,26 @@ absl::Status LibraryRewriter::FuseNeighbors(HloFusionInstruction* fusion,
     // If `instr` is another fusion of the same library type, fuse it.
     // We don't need to add its neighbors to the frontier because anything that
     // can be fused would have already been fused into `instr`.
-    if (IsCustomFusionWithKind(instr, lib->fusion_kind())) {
+    // TODO(intel-tf): Check if merge fusions are beneficial for oneDNN.
+    if (lib->fusion_kind() != kOneDnnFusionKind &&
```
penpornk: I'd rather keep oneDNN-specific stuff out of the common code. For this, we can add a function `ShouldMergeFusions()` and check `lib->ShouldMergeFusions()` instead.

xla/backends/cpu/transforms/library_rewriter.cc
```diff
@@ -261,13 +271,26 @@ absl::Status LibraryRewriter::FuseNeighbors(HloFusionInstruction* fusion,
...
                           MergeFusionInstructions(
                               fusion, Cast<HloFusionInstruction>(instr), dir));
       continue;
     }
 
+    if (lib->fusion_kind() == kOneDnnFusionKind &&
+        instr->opcode() == HloOpcode::kDot) {
+      VLOG(4) << "  Only one dot op is allowed in oneDNN fusion";
+      break;
+    }
```
penpornk: If it's greedy mode (fusion grown from eltwise op), then there could have been no dots in the fusion yet.

We should also make this not oneDNN-specific. I think we can make oneDNN matcher track the fusion state (e.g., count #dots) and add a `lib->ShouldFuse(fusion, instr)` function that we can check on (where we check `lib->IsOpSupported` right now). 

I'm fine with keeping this code for now. Please add a TODO to hide oneDNN-specific checks in oneDNN matcher. I can do that after this PR goes in.

penpornk (changes requested):
Belated happy new year! I'm very sorry for the delay.

xla/backends/cpu/transforms/library_rewriter.cc
yimeisun123: Ok, I added the TODO comment and we can make the changes later.

yimeisun123:
@penpornk, thank you very much for your review and suggestions. I made updates as suggested, please let me know if any other change is needed. Thanks

penpornk (approved):
Thank you for the changes!

xla/backends/cpu/onednn_support.cc
```diff
@@ -15,57 +15,57 @@ limitations under the License.
...
 #include "absl/log/log.h"
 #include "absl/status/statusor.h"
-#include "dnnl.hpp"  // NOLINT: for DNNL_MAX_NDIMS
+#include "oneapi/dnnl/dnnl.hpp"  // NOLINT: for DNNL_MAX_NDIMS
+#include "oneapi/dnnl/dnnl_graph.hpp"
+#include "tsl/platform/cpu_info.h"
 #include "xla/backends/cpu/codegen/target_machine_features.h"
 #include "xla/backends/cpu/runtime/dot_dims.h"
+#include "xla/hlo/ir/hlo_casting_utils.h"
+#include "xla/hlo/ir/hlo_instruction.h"
```
thomasjoerg, line 31: Please update the BUILD file deps accordingly.

thomasjoerg (changes requested):

xla/backends/cpu/onednn_support.cc
yimeisun123: Update the BUILD file with dependency for these two. Thanks!

thomasjoerg (approved):

thomasjoerg (approved):

yimeisun123:
@thomasjoerg , I pushed a commit fixing the window's build error (should also fix the other variants' build error), and a commit of merging to master yesterday. Could you please check? Thanks

yimeisun123:
@penpornk , @thomasjoerg , could you please check if the last commit is ok for the builds of different variants? I verified for the window's build. Thanks

yimeisun123:
@penpornk , I made additional update on the BUILD file, does the onednn_fusion_test work for you now? It passed on our end of testing. In case you still encounter the issue, could you please share the error traces? Thanks!

yimeisun123:
@penpornk , thanks for triggering the CI jobs, I see 8 failing pull requests but not sure what is the cause for those, the logs don't seem to be related to this PR changes, please let me know if anything is needed from me. Thanks!

yimeisun123:
@xla-rotation , reach out to check on the 8 failing pull request failures, I checked the logs and don't seem to be related to this PR changes, appreciate if any advice on how to proceed from here. Thanks

yimeisun123:
@penpornk , @thomasjoerg , please let me know if any action needed from me to move this PR? i.e., are the latest 8 pull request errors due to the baseline outdated and I need to rebase? Thanks

dimitar-asenov:
@yimeisun123, Sorry for the delay. Could you please rebase this PR and ping me afterwards! Thanks.

yimeisun123:
@dimitar-asenov , I merged to the latest base (public portal for intel), please let me know if anything else is needed. Thanks!

dimitar-asenov (approved):

dimitar-asenov:
@yimeisun123 The PR contains a couple of blocks guarded by a macro:

```
#if XLA_ONEDNN_USE_GRAPH_API
  if (lib->fusion_kind() == kOneDnnFusionKind) {
  ...
```

We really want to avoid macros if at all possible. For the code above, I suppose there are two options:
- Can we remove the `#if` completely and just rely on `if (lib->fusion_kind() == kOneDnnFusionKind)`? Are there some undefined symbols there?
- If the above is not possible, is it possible to do another run-time check that doesn't require macros?

yimeisun123:
> @yimeisun123 The PR contains a couple of blocks guarded by a macro:
> 
> ```
> #if XLA_ONEDNN_USE_GRAPH_API
>   if (lib->fusion_kind() == kOneDnnFusionKind) {
>   ...
> ```
> 
> We really want to avoid macros if at all possible. For the code above, I suppose there are two options:
> 
> * Can we remove the `#if` completely and just rely on `if (lib->fusion_kind() == kOneDnnFusionKind)`? Are there some undefined symbols there?
> * If the above is not possible, is it possible to do another run-time check that doesn't require macros?

Yes, the macro is needed as without it, some builder job fails ( I couldn't remember exactly which build failed before). Similar macro is in "xla/service/cpu/thunk_emitter.cc".
We have additional feature changes gated by this one, I agree that we will need to discuss on how to best accommodate the specific backend need without special flags if possible. 

There are 6 failing checks (one is timeout), I am not sure if they are related to my changes though. Please let me know if anything to be fixed to address those failed check. 
Thank you for your help on moving this PR!

dimitar-asenov:
When I try to merge this I get two failures:

### OneDnnFusionUnaryOpTestSuite/OneDnnFusionUnaryOpTest.UnaryOp/log_f32:

```
onednn_fusion_test.cc:180

Value of: RunAndCompare(hlo_unary_str, ErrorSpec{atol_, rtol_})
  Actual: false (
Mismatch count 9 (22.5000%) in shape f32[40] (40 elements), abs bound 0.0001, rel bound 0.0001
  actual              83.666893, expected            -2.11341548, index {34}, rel error     40.6, abs error     85.8
  actual             82.9737473, expected            -2.81624508, index {4}, rel error     30.5, abs error     85.8
  actual             82.9737473, expected            -2.83174324, index {15}, rel error     30.3, abs error     85.8
  actual             81.5874557, expected            -4.19633627, index {10}, rel error     20.4, abs error     85.8
  actual             81.5874557, expected            -4.19806051, index {18}, rel error     20.4, abs error     85.8


Expected literal:
f32[40] {nan, nan, -2.06607842, -3.28088307, -2.81624508, nan, nan, -2.26664448, -2.26539588, -1.71253502, -4.19633627, -2.88553476, -1.90186715, nan, nan, -2.83174324, -2.28922653, nan, -4.19806051, nan, -3.67909575, -2.24401617, -2.567945, -1.71963811, -1.8717767, -2.84595728, nan, -2.34157848, -3.69691944, -2.2040329, -1.75390732, -2.05058646, nan, nan, -2.11341548, nan, -2.41026425, -2.06417346, -1.62297297, -4.64582491}

Actual literal:
f32[40] {nan, nan, -2.06608129, -3.28088498, 82.9737473, nan, nan, -2.26664543, -2.26539683, -1.68579042, 81.5874557, -2.88553476, -1.90186882, nan, nan, 82.9737473, -2.28922749, nan, 81.5874557, nan, -3.6790967, -2.24401712, -2.56794691, -1.67511582, -1.87177861, -2.84595871, nan, -2.34158158, -3.69692254, -2.2040329, -1.72456455, -2.05058956, nan, nan, 83.666893, nan, -2.36438727, -2.06417632, -1.6229763, -4.64582682})
```

### OneDnnFusionMatMulFusionTestSuite/OneDnnFusionMatMulFuseUnaryTest.MatmulFuseWith/log_f32

```
onednn_fusion_test.cc:293

Value of: RunAndCompare(hlo_fusion_str, ErrorSpec{atol_, rtol_})
  Actual: false (
Mismatch count 62327 (20.7757%) in shape f32[1000,300] (300000 elements), abs bound 0.0001, rel bound 0.0001
  actual             85.7463379, expected        -0.000669464993, index {851,105}, rel error 1.28e+05, abs error     85.7
  actual             85.7463379, expected         -0.00108091813, index {811,246}, rel error 7.93e+04, abs error     85.7
  actual             85.7463379, expected         -0.00144442881, index {67,110}, rel error 5.94e+04, abs error     85.7
  actual             85.7463379, expected         -0.00258379895, index {315,105}, rel error 3.32e+04, abs error     85.7
  actual             85.7463379, expected         -0.00357401511, index {479,105}, rel error  2.4e+04, abs error     85.7
Absolute magnitude breakdown of actual values:
  0      <= x < 0.0001 :       0 (  0.0000%)
  0.0001 <= x < 0.001  :       4 (  0.0013%)
  0.001  <= x < 0.01   :      14 (  0.0047%)
  0.01   <= x < 0.1    :     366 (  0.1220%)
  0.1    <= x < 1      :  230819 ( 76.9397%), mismatches 26720
  1      <= x < inf    :   68797 ( 22.9323%), mismatches 35607
Elements exceeding abs error bound 0.0001: 62327 (20.7757%)
Relative error breakdown of elements exceeding abs error bound:
  <  0.0001 :       0 (0.0000%)
  >= 0.0001 :   62327 (100.0000%)
  >= 0.001  :   62327 (100.0000%)
  >= 0.01   :   62323 (99.9936%)
  >= 0.1    :   37847 (60.7233%)
  >= 1      :   29911 (47.9904%)
Elements exceeding rel error bound 0.0001: 62377 (20.7923%)
Absolute error breakdown of elements exceeding rel error bound:
  <  0.0001 :      50 (0.0802%)
  >= 0.0001 :   62327 (99.9198%)
  >= 0.001  :   62327 (99.9198%)
  >= 0.01   :   62327 (99.9198%)
  >= 0.1    :   29911 (47.9520%)
  >= 1      :   29911 (47.9520%)


Expected literal:
[TRUNCATED, Literal with more than 1000 values]

Actual literal:
[TRUNCATED, Literal with more than 1000 values])
```

Can you please look into these? Is this change expected?

yimeisun123:
@dimitar-asenov , I am not seeing the test case failures. Our internal CI run also passed on my last merge commit before I pushed the commit. Could you please let me know the test run command that having the above failures? just wonder if there is a difference of flag settings used for running the test. Thanks

dimitar-asenov:
@yimeisun123 The command is simply:

```
blaze test xla/service/cpu/tests:onednn_fusion_test --compilation_mode=opt 
```

Let me know if you still can't reproduce and I can have a closer look on our end.

yimeisun123:
> @yimeisun123 The command is simply:
> 
> ```
> blaze test xla/service/cpu/tests:onednn_fusion_test --compilation_mode=opt 
> ```
> 
> Let me know if you still can't reproduce and I can have a closer look on our end.

@dimitar-asenov , I am not able to reproduce this on my side with your bazel test command option. 
Which kind of linux machine is the test run on? I'd appreciate if you could check on your end. Also, if you could reproduce the error, could you please also try to add the "--define=build_with_onednn_async=true" to the test command and see if it prevents the failure? Thanks

seantalts:
maybe adding `--compilation_mode=opt --//third_party/tensorflow/compiler/xla/tsl:ci_build=true` could help repro? 
it's an amd k8 cpu.

yimeisun123:
> maybe adding `--compilation_mode=opt --//third_party/tensorflow/compiler/xla/tsl:ci_build=true` could help repro? it's an amd k8 cpu.

1) I am not able to add '--//third_party/tensorflow/compiler/xla/tsl:ci_build=true', it doesn't exist in the xla repo.
2) I tried running the tests on amd turion, and the tests pass too.

I am running the test under the xla repo with the command
 **bazel test //xla/service/cpu/tests:onednn_fusion_test --compilation_mode=opt --test_output=all -t-**
 
 Anything else that may be different in your project or test command option?

seantalts:
Sorry, remove the Google project structure part of that and start it with 'xla.' That's all I can think of...

yimeisun123:
> xla/tsl:ci_build=true

I added the option, and no failure either on amd turion. Below is the last part of test completion, and the command used is at the end. 

`
[==========] 25 tests from 4 test suites ran. (7161 ms total)
[  PASSED  ] 24 tests.
[  SKIPPED ] 1 test, listed below:
[  SKIPPED ] OneDnnFusionBinaryOpTestSuite/OneDnnFusionBinaryOpTest.BinaryOp/dot_f16

INFO: Found 1 test target...
Target //xla/service/cpu/tests:onednn_fusion_test up-to-date:
  bazel-bin/xla/service/cpu/tests/onednn_fusion_test
INFO: Elapsed time: 16.501s, Critical Path: 8.33s
INFO: 2 processes: 1 internal, 1 local.
INFO: Build completed successfully, 2 total actions
//xla/service/cpu/tests:onednn_fusion_test                               PASSED in 7.2s

There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
(env_test) yimeisun@amd-01:~/intel_pub_xla$ bazel test xla/service/cpu/tests:onednn_fusion_test --compilation_mode=opt --//xla/tsl:ci_build=true --test_output=all -t-
`

dimitar-asenov:
The failures happen on AMD Milan EPYC. Is this something you can test on your end?

dimitar-asenov:
Also, the specific command doesn't seem to matter on our end, with or without `--c opt` it still fails. E.g. even the plain `bazel test xla/service/cpu/tests:onednn_fusion_test` fails with the numerics mismatch.

yimeisun123:
@dimitar-asenov , does adding _--define=build_with_onednn_async=true_ option help with the test cases? i.e., 

 bazel test xla/service/cpu/tests:onednn_fusion_test --define=build_with_onednn_async=true --compilation_mode=opt --test_output=all -t-
 
 I'm trying to find AMD Milan EPYC on my side. 
 Thanks

dimitar-asenov:
Sorry for the delay, `--define=build_with_onednn_async=true` makes no difference and it still fails in the same way with that flag.

dimitar-asenov:
We also ran the test on an Intel Skylake CPU and it failed in the same way (with and without the flag). So maybe there is some other difference, that's independent of the CPU architecture.

yimeisun123:
@dimitar-asenov , I tried on AMD EPYC 7B13, the test case pass (regardless of compilation opt or async build options). I will try on skylake and update.

dimitar-asenov:
@penpornk Do you have any idea what could be causing the internal failures?

penpornk:
@dimitar-asenov I suspected it's because we are using oneDNN v3.3 internally (vs v3.7.3 in OSS). I was holding off this PR to test when we update to v3.7.3. I have informed Intel offline but didn't post here. Sorry about that.

The update is taking a while and still not working properly yet. So I'd suggest disabling the failing Log test case for now.

xla/backends/cpu/onednn_support.cc
```diff
@@ -110,4 +110,122 @@ absl::StatusOr<bool> IsOneDnnDotSupported(
          !dot_canonical_dims.rhs_column_major;
 }
 
+const absl::flat_hash_map<HloOpcode, op::kind>& GetOneDnnUnaryOpMap() {
+  static absl::NoDestructor<absl::flat_hash_map<HloOpcode, op::kind>>
+      unary_op_map({
+          {HloOpcode::kAbs, op::kind::Abs},
+          {HloOpcode::kExp, op::kind::Exp},
+          {HloOpcode::kLog, op::kind::Log},
```
penpornk: Please disable the log op for now since we can't guarantee that it will work correctly.

```suggestion
          // TODO(b/514636963): Enable Log op once we fix the wrong-answer issue internally.
          // {HloOpcode::kLog, op::kind::Log},
```

penpornk (changes requested):

xla/backends/cpu/onednn_support.cc
yimeisun123: I will disable the log op, and will also check if I can get the onednn version at runtime to enable or disable the test. Thanks.

penpornk: Thank you! I don't think we need to disable the test since other ops seem fine.

xla/service/cpu/tests/BUILD
penpornk, line 1: Why are we deleting these test targets?

xla/service/cpu/tests/restricted/onednn_fusion_test.cc
```diff
@@ -22,7 +22,7 @@ limitations under the License.
 #include "xla/backends/cpu/onednn_support.h"
 #include "xla/error_spec.h"
 #include "xla/service/cpu/onednn_util.h"
-#include "xla/tests/hlo_test_base.h"
+#include "xla/tests/restricted/hlo_test_base.h"
```
penpornk: This line was changed in a recent commit (to `hlo_test_base_legacy.h`). Please rebase again?

xla/backends/cpu/transforms/library_rewriter.cc
```diff
@@ -257,7 +257,21 @@ absl::Status LibraryRewriter::FuseNeighbors(HloFusionInstruction* fusion,
...
+
+  // TODO(intel-tf): Restrict fusion direction for oneDNN till future
+  // release of oneDNN library with both fusion direction support.
+#if XLA_ONEDNN_USE_GRAPH_API
+  if (lib->fusion_kind() == kOneDnnFusionKind) {
+    direction = FusionDirection::kDown;
+  }
+#endif  // XLA_ONEDNN_USE_GRAPH_API
+
+  AddFusionCandidates(fusion, fusion, direction, frontier);
```
penpornk: Can we add another function to LibraryMatcher for the starting fusion direction? Default value should be `kBoth` and the `OneDnnMatcher` overrides the function to return `kDown`.
```suggestion
  AddFusionCandidates(fusion, fusion, lib->fusion_direction(), frontier);
```

xla/backends/cpu/transforms/library_rewriter.cc
```diff
@@ -257,7 +257,21 @@ absl::Status LibraryRewriter::FuseNeighbors(HloFusionInstruction* fusion,
...
+
+  FusionDirection direction = FusionDirection::kBoth;
+
+  // TODO(intel-tf): Restrict fusion direction for oneDNN till future
+  // release of oneDNN library with both fusion direction support.
+#if XLA_ONEDNN_USE_GRAPH_API
+  if (lib->fusion_kind() == kOneDnnFusionKind) {
+    direction = FusionDirection::kDown;
+  }
+#endif  // XLA_ONEDNN_USE_GRAPH_API
```
penpornk: Please remove the `#ifdef` guard. We don't need it. The oneDNN matcher backend is only run when oneDNN is built with Graph API anyway:
https://github.com/openxla/xla/blob/9c76ff29f31c62217964d5441cde76b928f3a5c9/xla/backends/cpu/transforms/library_rewriter.h#L67-L73

penpornk (changes requested):

xla/service/cpu/tests/restricted/onednn_fusion_test.cc
yimeisun123: Yes, our internal sync is little late, i am doing it now. thanks

xla/backends/cpu/transforms/library_rewriter.cc
yimeisun123: Done

xla/backends/cpu/onednn_support.cc
yimeisun123: Done with commenting out the log op.

xla/backends/cpu/transforms/library_rewriter.cc
yimeisun123: Done

xla/service/cpu/tests/BUILD
yimeisun123: Most of the onednn tests are moved from "tests" to a newly created sub-folder of "restricted". Is that the question here?

xla/backends/cpu/transforms/library_rewriter.cc
yimeisun123: We have couple of more fusion changes that are built on top of this PR, we will discuss with you about bringing in those once this PR is merged.

yimeisun123:
@penpornk , I've addressed the review comments and merged up to the latest as of now, please let me know if additional update is needed. Thanks

xla/backends/cpu/transforms/library_rewriter.cc
penpornk: We recently added this check in https://github.com/openxla/xla/commit/0a70932fbf0f8492a20e74d1b7e7792d2393e138 Please use this check instead.
We can put the condition in its own `if` and add a `VLOG` message.

Please add `MaxFusionSize()` to library_matcher with default value to `kMaxInstructionsInFusion` (100) and override it in onednn_matcher. Please help rename `kMaxInstructionsInFusion` to `kMaxFusionSize`.

xla/backends/cpu/transforms/library_rewriter.cc
penpornk: Please remove the `#if XLA_ONEDNN_USE_GRAPH_API` guard.

penpornk (changes requested):
A few more comments please. Sorry for the inconvenience!

xla/backends/cpu/transforms/library_rewriter.cc
yimeisun123: Please let me know if the new commit addresses this change request. Thanks

yimeisun123:
@penpornk, I added the following changes to the matchers to address the review comments, i.e.,
1) MaxFusionSize() to set the limit of fusion size.
2) ShouldFuse(...) to check if an instruction can be added to a fusion.
3) Removed #if XLA_ONEDNN_USE_GRAPH_API guard in the library_rewriter.
4) Updated library_rewriter_test

Please check and let me know if additional updates are needed. 
Thanks

p.s., if you are using v3.3 oneDNN library internally, the fusions (i.e., dot plus binary, or unary) may not be supported.

penpornk (approved):
Thank you for the changes!

dimitar-asenov (approved):

thcmbs:
Hi, it seems like the CI is failing with your change, so we cannot merge it. Can you PTAL?

yimeisun123:
From the failed builder logs, I need to add back the onednn guard in the library_matcher.h. I will push a commit soon.

yimeisun123:
@thcmbs , new commit pushed to address the errors from some builders and clang tidy. 
For clang-format, I run locally with style=Google, and I didn't see format error on changed codes, not sure which update is not complied to the clang-format check on pull request.

dimitar-asenov (approved):

dimitar-asenov:
@yimeisun123  The `library_rewriter_test.cc` file introduces or extends an `#if`, `#ifdef`, or `#ifndef` macro. The XLA team is actively working to reduce the number of these macros in the codebase as they hurt maintainability.

Could you please revise these tests so that they don't have this kind of conditional compilation? If needed, you can introduce new test targets that have conditional dependencies. In general, it's better to push these kind of conditions to the `BUILD` file and out of the C++ code.

yimeisun123:
> @yimeisun123 The `library_rewriter_test.cc` file introduces or extends an `#if`, `#ifdef`, or `#ifndef` macro. The XLA team is actively working to reduce the number of these macros in the codebase as they hurt maintainability.
> 
> Could you please revise these tests so that they don't have this kind of conditional compilation? If needed, you can introduce new test targets that have conditional dependencies. In general, it's better to push these kind of conditions to the `BUILD` file and out of the C++ code.

@dimitar-asenov , @penpornk , the `library_rewriter_test.cc` file already has the #if XLA_ONEDNN_USE_GRAPH_API guarded code section and this PR adds couple more. If we need to remove the macro completely from this test file, are you ok with -
split the content from this file into two, i.e., 
  xla_cc_test(
      name = "library_rewriter_test",
      srcs = ["library_rewriter_test.cc"],
      deps = [
          # ... existing deps without onednn_matcher
      ],
  )

  onednn_graph_cc_test(  # This macro automatically sets XLA_ONEDNN_USE_GRAPH_API
      name = "library_rewriter_onednn_test",
      srcs = ["library_rewriter_onednn_test.cc"],
      deps = [
          # ... same deps plus onednn_matcher
      ],
  )

Remove the test cases that require the #ifdef guard from the library_rewriter_test, and cover those in the library_rewriter_onednn_test. There will be some duplicates on the test code, but no macro usage. 

Is this what you are suggesting? I can make the change if that aligns with your thought.

I checked the most recent failed workflows, 
- the clang tidy reported the need of header inclusion. 
- the clang-format failure, I can't tell which file failed the check. In my local check, couple of files need to reorder the header file inclusion, e.g., move up the include <gtest/gtest.h>, but the order is pre-existing, not touched by this PR and many files have the similar out of order header inclusion. Should I change the header file inclusion order even if they are not touched by this PR.
(also, notice that clang tidy and clang-format are skipped for other PRs). 
- the other workflow failures are not related to this PR change, I can rebase this PR to newer baseline and maybe those failures will be gone.

dimitar-asenov:
I think @penpornk is submitting a change that will help you avoid the `#if`. @penpornk please post here once your change is in.

yimeisun123:
@dimitar-asenov, thanks for sharing @penpornk 's changes. 
1) With Penporn's change, the existing #ifdef guard usage in the library_rewriter_test is eliminated. Thanks for this. I will do a merge once the change is released in base.
2) In this PR, I am adding new test case to check the max fusion size, and the limit is different between default (ynn) and onednn. Can I also move the "kMaxFusionSize" and "kMaxOneDnnFusionSize" to the "library_fusion_kinds.h" file?

Thanks

xla/backends/cpu/transforms/onednn_matcher.h
```diff
@@ -22,14 +22,18 @@ limitations under the License.
...
 #include "absl/strings/string_view.h"
+#include "tsl/platform/protobuf.h"
 #include "xla/backends/cpu/codegen/target_machine_features.h"
 #include "xla/backends/cpu/onednn_support.h"
 #include "xla/backends/cpu/transforms/library_matcher.h"
 #include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_instructions.h"
 #include "xla/hlo/ir/hlo_opcode.h"
-#include "tsl/platform/protobuf.h"
+#include "xla/hlo/utils/hlo_query.h"
```
penpornk, line 33: Please add a dep "//xla/hlo/utils:hlo_query" to `onednn_matcher` build target.

penpornk (changes requested):
> With Penporn's change, the existing #ifdef guard usage in the library_rewriter_test is eliminated. Thanks for this. I will do a merge once the change is released in base.

Thank you! Please also add a missing dependency for `onednn_matcher`. 

> In this PR, I am adding new test case to check the max fusion size, and the limit is different between default (ynn) and onednn. Can I also move the "kMaxFusionSize" and "kMaxOneDnnFusionSize" to the "library_fusion_kinds.h" file?

We can either do this or add a function in LibraryRewriter to look up a lib backend and return its limit. For simplicity, let's just move the constants to library_fusion_kinds.h in this PR for now. Thank you for checking!

dimitar-asenov (approved):

PR #32628, opened 2025-10-13
DOC: errors doc pages

mtsokol:
📝 Summary of Changes

This PR introduces errors overview docs and error codes pages.

🎯 Justification

Errors and error code docs.

🚀 Kind of Contribution

📚 Documentation


docs/errors_overview.md
```diff
@@ -0,0 +1,53 @@
...
+a list of an additional context other than the error message, which will be
+attached to each error within the category.
+
+### E01xx - Runtime data transfer/allocation failures
+
+Failures when allocating memory on accelerator (such as HBM or Vmem) or host’s memory (DRAM):
+- Memory space,
+- Available memory,
+- Summarized view of what other large buffers/programs are occupying memory,
+- Backtrace - stacktrace at the point where the error is thrown.
```
dubstack: We don't want to say anything about the context attached to these errors yet, we will gradually add them as we add support for them in the backend.

docs/errors_overview.md
```diff
@@ -0,0 +1,53 @@
+# XLA Errors Overview
+
+XLA errors are categorized into different XLA error sources. Each source has
+a list of an additional context other than the error message, which will be
+attached to each error within the category.
+
```
dubstack, line 6: Maybe give an example of how an error message would look?

docs/errors_overview.md
```diff
@@ -0,0 +1,53 @@
+# XLA Errors Overview
+
+XLA errors are categorized into different XLA error sources. Each source has
```
dubstack, line 3: We should clarify that for a period of time not all error messages would have an attached error code as we gradually migrate errors in XLA.

docs/errors_overview.md
```diff
@@ -0,0 +1,53 @@
...
+a list of an additional context other than the error message, which will be
+attached to each error within the category.
+
+### E01xx - Runtime data transfer/allocation failures
```
dubstack: We might be restricting ourselves by telling people how we plan to group these errors, maybe skip these sections completely for now?

SandSnip3r: Agree. This and the below sections will be redundant when compared to the actual error pages.

I think what should be here is instead some commentary on the mechanism via which errors are thrown in XLA (Statuses vs CHECK failures). This openxla documentation is not just for the users of XLA, but also the developers. Explaining when a developer should be returning a Status in the case of an error compared to triggering a CHECK would be useful.

mtsokol: I added a short paragraph on statuses and CHECK failures, and removed all errors' sections from the main page.

docs/errors_overview.md
mtsokol: Added a note it's work-in-progress.

docs/errors_overview.md
mtsokol: Removed these details.

docs/errors_overview.md
mtsokol: I copied an error log example from one of the codes. Is it sufficient?

SandSnip3r (approved):

docs/errors/error_03xx.md
SandSnip3r, line 1: Actually, all of these documents get added, but they're empty. We should not add the documents until we actually have content for them

docs/error_codes.md
```diff
@@ -0,0 +1,19 @@
...
+
+- [E1000](./errors/error_1000.md)
+- [E1001](./errors/error_1001.md)
+- [E1002](./errors/error_1002.md)
+- [E03xx](./errors/error_03xx.md)
+- [E0100](./errors/error_0100.md)
+- [E0101](./errors/error_0101.md)
+- [E0102](./errors/error_0102.md)
+- [E2xxx](./errors/error_2xxx.md)
+- [E11xx](./errors/error_11xx.md)
```
SandSnip3r: Similar to above about adding empty pages. If we're not going to introduce empty pages, we should not be adding any links yet

SandSnip3r (changes requested):

docs/errors/error_03xx.md
mtsokol: I just removed them.

docs/error_codes.md
mtsokol: Done!

mtsokol:
@SandSnip3r Let's merge it?

dubstack (approved):

docs/errors_overview.md
```diff
@@ -0,0 +1,36 @@
+# XLA Errors Overview
+
+XLA errors are categorized into different XLA error sources. Each source has
+a list of an additional context other than the error message, which will be
+attached to each error within the category.
+
+🚧 Please note that this standarization effort is a work in progress so not all
```
dubstack: Style notice: Don't use the term "Please note".

Reference URL: https://developers.google.com/style/word-list#please

mtsokol: Done!

dubstack (approved):

dubstack (approved):

docs/error_codes.md
```diff
@@ -0,0 +1,3 @@
+# XLA Error codes
+
+This page is a list of all error codes emitted by the XLA compiler.
```
appujee, line 3: We should add at least a few error codes to get started.

PR #32724, opened 2025-10-15
Disable only the test cases that are failing and enable 3 test targets on B200.

dimvar:


akuegel (approved):

IllogicalMoose (approved):

dimvar:
@IllogicalMoose  PTAL. I disabled the test case in cudnn_test that fails for you, and I kept gpu_compiler_test as "broken".

xla/backends/gpu/codegen/cudnn_test.cc
```diff
@@ -895,7 +895,7 @@ ENTRY Test {
                             ErrorSpec{/*aabs=*/1e-3, /*arel=*/1e-5}));
 }
 
-TEST_F(CuDnnFusionExecutionTest, ConvWgradWithNHWCLayoutExecutesCorrectly) {
+TEST_F(CuDnnFusionExecutionTest, DISABLED_ConvWgradWithNHWCLayoutExecutesCorrectly) {
```
akuegel: Why do you disable the test completely? Shouldn't it be sufficient to skip it on B200?

dimvar: Oversight. Fixed.

akuegel (approved):

PR #30127, opened 2025-08-12
[ROCM] Splitting gridDim.x for very large reductions / loop fusion kernels

pemeliya:
We had several failures on ROCM side due to very large gridDim.x values for fusion kernels.
Here is the proposal to split gridDim.x into gridDim.x/y in these cases.

1. added gridDim.x splitting for reduction and loop fusions to handle large grid sizes
2. extended gpu_kernel_tiling test and buffer_comparator_test
3. improved error logging in rocm_stream.cc in case of kernel launch failure
4. **fixing dangling reference bugs in emitter_base.cc**

We already had some rocm-related fixes for this issue: with this PR all fixes shall be unified

@xla-rotation: could you have a look please ?

PR #32748, opened 2025-10-15
[ROCM] Adding grid-stride loops for buffer comparator and redzone checker kernels

pemeliya:
📝 Summary of Changes
Added grid-stride loops to buffer_comparator and redzone checker kernels,
extended gpu_kernel_tiling test and buffer_comparator_test, removed rocm-specific hacks

🎯 Justification
When the input shape is too large, buffer_comparator/redzone_checker kernels may fail to launch due to grid size limitations

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
transformed HLO.

🧪 Execution Tests:

This PR was extracted as part of the original one: https://github.com/openxla/xla/pull/30127

@xla-rotation: can you have a look, please ?

### DIFF
--- a/xla/backends/gpu/runtime/buffer_comparator_test.cc
+++ b/xla/backends/gpu/runtime/buffer_comparator_test.cc
@@ -449,37 +449,41 @@ TEST_F(BufferComparatorTest, BF16) {
 }
 
 TEST_F(BufferComparatorTest, VeryLargeArray) {
-  constexpr PrimitiveType number_type = U16;
+  constexpr PrimitiveType number_type = U8;
   using NT = primitive_util::PrimitiveTypeToNative< number_type >::type;
 
-  // Set non-power-of-two element count on purpose
-  int64_t element_count = (1LL << 34) - 11;
+  // Set non-power-of-two element count on purpose, use aligned byffer siuze
+  int64_t n_elems = (1LL << 33) - 11,
+  // Buffer size must be 4-bytes aligned for Memset32
+          buf_size = (((n_elems + 1) * sizeof(NT)) + 3) & ~3;
   auto stream = stream_exec_->CreateStream().value();
 
   // Use host memory here since there is a limitation of 4GB per test on
   // device memory alloc
   TF_ASSERT_OK_AND_ASSIGN(
     auto base, 
-    stream_exec_->HostMemoryAllocate((element_count + 1) * sizeof(NT)));
+    stream_exec_->HostMemoryAllocate(buf_size));
 
   // We use overlapping lhs and rhs arrays to reduce memory usage, also this 
   // serves as an extra test for possible pointer aliasing problems
-  se::DeviceMemoryBase lhs(base->opaque(), base->size() - sizeof(NT)),
+  se::DeviceMemoryBase lhs(base->opaque(), n_elems * sizeof(NT)),
                        rhs(static_cast< NT *>(base->opaque()) + 1, lhs.size());
 
-  TF_CHECK_OK(stream->Memset32(&lhs, 0xABCDABCD, base->size()));
+  constexpr uint32_t pattern = 0xABABABAB;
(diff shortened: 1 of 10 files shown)


i-chaochen:
Hi @derdrdirk this PR is splited as you requested, could you please check. Thanks!

olegshyshkov (approved):

pemeliya:
@xla-rotation it seems that this PR is not merged yet.. Are there some internal checks failing ?

xla/backends/gpu/runtime/buffer_comparator_test.cc
```diff
@@ -448,6 +448,45 @@ TEST_F(BufferComparatorTest, BF16) {
                    .value());
 }
 
+TEST_F(BufferComparatorTest, VeryLargeArray) {
```
thomasjoerg, line 452: Heads-up: This test OOMs on Google's CI system. What is the smallest amount of memory you need to test what grid-stride loops properly?

thomasjoerg (changes requested):

xla/backends/gpu/runtime/buffer_comparator_test.cc
pemeliya: I think we can limit it to 4GB, would it be fine ? I have already reduced mem requirements, currently the test uses 1<<33 which is 8 GB

thomasjoerg: Yes, 4GB should work.

thomasjoerg (approved):

xla/backends/gpu/runtime/buffer_comparator_test.cc
pemeliya: yes, I have reduced the mem usage in the test to 4GB

thomasjoerg:
Heads-up: We are reverting this change because it breaks [/jax/tests:checkify_test](https://github.com/jax-ml/jax/blob/main/tests/checkify_test.py) under ASAN.

thomasjoerg:
```
SUMMARY: AddressSanitizer: stack-use-after-scope /xla/backends/gpu/codegen/emitters/emitter_base.cc:250 in xla::gpu::EmitterBase::EmitThreadId(mlir::ImplicitLocOpBuilder&, int) const
```

pemeliya:
> ```
> SUMMARY: AddressSanitizer: stack-use-after-scope /xla/backends/gpu/codegen/emitters/emitter_base.cc:250 in xla::gpu::EmitterBase::EmitThreadId(mlir::ImplicitLocOpBuilder&, int) const
> ```
@thomasjoerg are there any logs I can check to see what is going on? which tests were failing ? This is odd, the PR has nothing to do with XLA codegen emitters.

pemeliya:
I think I know what it is.. There is a long-term bug in emitter_base.cc [here](https://github.com/openxla/xla/blob/f243fdf62b3f12ee04940788351761a23a48c13d/xla/backends/gpu/codegen/emitters/emitter_base.cc#L228-L254). 

That is:
const auto& counts = launch_dimensions().block_counts(); 
taking a reference of temporary since launch_dimensions() is defined in **KernelFusionInterface** as:

virtual LaunchDimensions launch_dimensions() const = 0;

It used to be fine before I made LaunchDimensions::block_counts() and LaunchDimensions::thread_counts_per_block() functions to return const references.
Yes, it still shall work fine since C++ extends the lifetime of 'LaunchDimensions' temporary but it makes ASAN to complain..

I have fixed those in my original PR: https://github.com/openxla/xla/pull/30127/files#diff-d2ff1b3a972f7d24b800938e5a06fd414c65aa1bb2a5a9e5d69065bb034851dbL227-L250

but after it has been split upon request into two PRs, these changes appeared to be in different PRs which created the issue.

PR #32800, opened 2025-10-16
[XLA:CPU][oneDNN] Undo proto workarounds after switching to thunks based execution

akhilgoe:
This PR removes the proto workarounds that were added to make the code work with the legacy runtime. Specifically, it undoes:
1. Using 1-based indexing for proto repeated fields.
2. Type punning for floats.


akhilgoe (commented):
@penpornk this PR addresses the feedback you provided here

> @akhilgoe You're right. We have recently done a major protobuf version update -- from 3.21.9 to 5.28.3 in June (https://github.com/openxla/xla/commit/1fc2a846d259d3c896a6177485abcc68d97627a9), and to v6.31.1 in August (https://github.com/openxla/xla/pull/29658/commits/7abd4055b1629815d53dff1647a3ec15b7a79dc6). So let's test if the issue was gone with the upgrade or if it still persists.

_Originally posted by @penpornk in https://github.com/openxla/xla/pull/31511#discussion_r2394419894_

penpornk (approved):
Thank you for the changes!

PR #32836, opened 2025-10-17
[GPU] Dispatch S-curve model to single-partition multi-host topology

terryysun:
📝 Summary of Changes
Updated SINGLE_HOST communication type to SINGLE_PARTITION (fast-interconnect domain) to meet the need of multi-node NVLink (MNNVL) topology. Piped auto-detected partition size for communication type determination, also exposed partition size in SolGPUCostModel::Config for AOT compilation.

🎯 Justification
S-curve model cannot handle NVLink latency, single fast-interconnect domain including MNNVL topology should use latency table model. This PR updates the routing mechanism so that MNNVL will be treated as a single partition, while previously host is assumed equivalent to partition.

🚀 Kind of Contribution
✨ New Feature

N/A

🧪 Unit Tests:
Added unit tests for model dispatching mechanism.

🧪 Execution Tests:
Behavior unchanged for non-MNNVL topology, N/A.


xla/service/collective_utils.h
```diff
@@ -70,6 +70,10 @@ constexpr char kSolChunkSizeBytes[] = "chunk_size_bytes";
 // cost model.
 constexpr char kSolGpusPerNode[] = "gpus_per_node";
 
+// Defines the default value for the partition size (number of devices per
+// fast-interconnect domain), used by the SoL cost model.
+constexpr char kSolPartitionSize[] = "partition_size";
```
felixwqp, line 76: the const string in this file defines the flag name passed in via s-curve cost model flag. 

What's the use case that we want to customize the `partition_size`? shouldn't this be auto detected?

xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -203,10 +203,16 @@ absl::StatusOr<absl::Duration> DispatchEstimation(
   TF_ASSIGN_OR_RETURN(auto num_groups_and_devices,
                       GetReplicaGroupCountAndSize(&instr));
 
+  int64_t partition_size =
```
felixwqp: nit: can you abstract this section wtih 319-325 into a utility function to make the maintenance easier?

xla/service/collective_utils.h
terryysun: It's for AOT compilation when auto-detection is unavailable.

felixwqp: Good to know, Can I assume the user would need to set the flags for `partition_size` during AOT?

felixwqp: Forward offline discussion here: 

if this is not really used at this moment, how about keeping it simple by just removing this part? 

if you think this will have a use case soon, can you share an example about how this will be used? and  add comments to make the purpose of this flag more understandable for users?

xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -189,6 +189,16 @@ absl::StatusOr<absl::Duration> DCNCollectiveDuration(
   return result;
 }
```
felixwqp, line 191: the nested Ternary Operator takes high cognitive cost to understand and is harder to debug.  And it doesn't align with google internal codebase standard, 

 https://stackoverflow.com/questions/160218/to-ternary-or-not-to-ternary


can you refactor it with if-else to make it more easy to read?

terryysun: refactored

xla/service/collective_utils.h
terryysun: It's also useful when CUDA driver version too low for auto detection, which we encountered a couple of times before. Updated the doc string with more details.

xla/service/gpu/transforms/collectives/collective_ops_utils.h
```diff
@@ -34,13 +34,13 @@ enum class GPUCommunicationType {
   // the involved hosts has only a subset of its devices participating.
   NON_RAIL_ALIGNED = 2,
   // All devices participating in the collective operation reside on the same
-  // host machine.
-  SINGLE_HOST = 3
+  // fast-interconnect domain.
```
felixwqp, line 38: Thanks Terry! 

With this, we basically, merged the single-slice and single-host to the same concept. 

do you have plan to distinguish single-partition v.s. single-host in long term, in the scope of cost model?

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
```diff
@@ -66,68 +66,69 @@ bool SameParticipantCounts(const absl::flat_hash_map<int64_t, size_t>& lhs,
...
   }
 
-  return CommunicationMetadata{node_to_participant_count, num_devices_per_host,
+  return CommunicationMetadata{partition_to_participant_count, partition_size,
                                instr.GetModule()->config().replica_count()};
 }
 
-bool IsSingleHost(const CommunicationMetadata& pattern) {
-  if (pattern.node_to_participant_count.size() == 1) {
+bool IsSingleSlice(const CommunicationMetadata& pattern) {
```
felixwqp: Would `SinglePartition` align with the type naming here?

xla/service/gpu/transforms/collectives/collective_ops_utils.h
terryysun: For cost model, we care about the boundary of fast-interconnect domain instead of host/node, so single-host is no longer relevant.

xla/service/gpu/transforms/collectives/collective_ops_utils.cc
terryysun: Updated to partition, slice/partition is kind of used interchangeably in the code base at the moment, planning to align to partition.

felixwqp: sounds great! Thanks!

xla/service/gpu/transforms/collectives/collective_ops_utils.h
felixwqp: I see. I don't have access to GB200 now,  in GB200, in the same partition, can I assume NVLink between nodes and NVLink between partitions share the same network charecteristic?

felixwqp (approved):

felixwqp:
There are some merge conflict on HEAD.

Could you rebase? 

Thank you so much!

xla/service/gpu/transforms/collectives/collective_ops_utils.h
terryysun: Not sure if I follow the question but partition is for representing the NVLink domain, i.e. there's no NVLink between any partitions. Within the same partition, the connections across nodes should be the same as those within node.

terryysun:
@felixwqp could you help trigger copybara? Thanks!

felixwqp (approved):

PR #32838, opened 2025-10-17
Fix family-conditional logic

olupton:
📝 Summary of Changes
The fallback logic now correctly identifies the highest known compatible architecture when given an unknown architecture as input.

🎯 Justification
Previously the logic would propose an incompatible architecture in this case.

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
Added a new test case showing the previously-failing case (it used to propose `sm_110`)

### DIFF
--- a/xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
+++ b/xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
@@ -248,12 +248,12 @@ std::string GetSmName(se::CudaComputeCapability compute_capability) {
       {8, 6},  {8, 0},  {7, 5},  {7, 2},  {7, 0},  {6, 2}, {6, 1}, {6, 0},
       {5, 3},  {5, 2},  {5, 0},  {3, 7},  {3, 5},  {3, 2}, {3, 0}};
   auto target_compute_capability = kSupportedVersions[0];
-
   for (const auto& v : kSupportedVersions) {
-    if (!gpu_compute_capability.CanRunOn(v)) {
+    target_compute_capability = v;
+    if (gpu_compute_capability.SupportsAllFeaturesOf(
+            target_compute_capability)) {
       break;
     }
-    target_compute_capability = v;
   }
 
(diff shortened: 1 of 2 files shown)


xla/service/gpu/llvm_gpu_backend/nvptx_backend.cc
```diff
@@ -248,12 +248,12 @@ std::string GetSmName(se::CudaComputeCapability compute_capability) {
       {8, 6},  {8, 0},  {7, 5},  {7, 2},  {7, 0},  {6, 2}, {6, 1}, {6, 0},
       {5, 3},  {5, 2},  {5, 0},  {3, 7},  {3, 5},  {3, 2}, {3, 0}};
   auto target_compute_capability = kSupportedVersions[0];
-
   for (const auto& v : kSupportedVersions) {
-    if (!gpu_compute_capability.CanRunOn(v)) {
+    target_compute_capability = v;
```
apivovarov: target_compute_capability is assigned to kSupportedVersions[0] twice. Before the loop and inside it. What do you think about the following refactoring? It initializes target_compute_capability to the least supported version as a safe fallback, and then updates it only if a more advanced supported version is found:

```cpp
// Initialize to the least supported version, which acts as a safe fallback
target_compute_capability = kSupportedVersions[std::size(kSupportedVersions) - 1]; 

for (const auto& v : kSupportedVersions) {
  if (gpu_compute_capability.SupportsAllFeaturesOf(v)) {
    // Found the most advanced supported capability
    target_compute_capability = v;
    break;
  }
}
```

olupton: I don't feel strongly. They come to the same thing in terms of fallback behavior. I pushed your version in case you feel strongly.

apivovarov: LGTM. Thank you.

apivovarov (approved):

Issue #32845, opened 2025-10-17
Verifier failure on collective permute with mixed floating point precision (on multi-GPU)


jaro-sevcik:
If collective permutes of mixed precisions get combined, the verifier will fail on the resulting code.

JAX repro (on a node with at least two GPUs, JAX version 0.7.2, but also repros on ToT JAX):

```python
from functools import partial

import numpy as np

import jax
import jax.numpy as jnp
from jax.sharding import Mesh, PartitionSpec as P
from jax.experimental.shard_map import shard_map

d = jax.devices()
mesh = jax.sharding.Mesh(np.array(d[:2]), ("x",))

a = jnp.ones((1024,), dtype=jnp.float32)
b = jnp.ones((1024,), dtype=jnp.bfloat16) + 2

@partial(shard_map, mesh=mesh, in_specs=(P('x'), P('x')),
         out_specs=(P('x'), P('x')))
def f(i, j):
  i = jax.lax.ppermute(i, 'x', [(0, 1), (1, 0)])
  j = jax.lax.ppermute(j, 'x', [(0, 1), (1, 0)])
  return i + 1, j + 2
print(jax.jit(f).lower(a, b).compile().as_text())
```

Error message:
```
jax.errors.JaxRuntimeError: INTERNAL: during context [hlo verifier]: Seen floating point types of different precisions in %collective-permute-start = ((f32[512]{0}, bf16[512]{0}), (f32[512]{0}, bf16[512]{0})) collective-permute-start(%param.2, %param.3), channel_id=1, source_target_pairs={{0,1},{1,0}}, backend_config={"operation_queue_id":"0","wait_on_operation_queues":[],"collective_backend_config":{"is_sync":false,"is_pipelined":false,"backend":"DEFAULT"},"force_earliest_schedule":false,"reification_cost":[],"device_type":"DEVICE_TYPE_INVALID"}, but mixed precision is disallowed.
```

I believe this can be fixed by simply adding `kCollectivePermute` to [the list of opcodes](https://github.com/openxla/xla/blob/d6d451e6b691b0677c44da8cff62920fcd2026b7/xla/service/hlo_verifier.cc#L1914) that are allowed to have mixed precision operands.

PR #32846, opened 2025-10-17
Allow mixed precision operands for collective permute

jaro-sevcik:
📝 Summary of Changes
Allow mixed precision collective-permute in the verifier.

🎯 Justification
Partially addresses https://github.com/openxla/xla/issues/32845

🚀 Kind of Contribution
🐛 Bug Fix

N/A

🧪 Unit Tests:
Tests that verifier passes on mixed precision collective-permute.

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/service/hlo_verifier.cc
+++ b/xla/service/hlo_verifier.cc
@@ -1915,6 +1915,7 @@ absl::Status CheckMixedPrecisionOperands(const HloInstruction* instruction) {
     case HloOpcode::kAsyncDone:
     case HloOpcode::kAsyncUpdate:
     case HloOpcode::kAsyncStart:
+    case HloOpcode::kCollectivePermute:
     case HloOpcode::kCopyDone:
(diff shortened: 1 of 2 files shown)


akuegel (approved):
I take it you tried out that codegen also works? In any case, it seems likely that it works.

jaro-sevcik:
> I take it you tried out that codegen also works? In any case, it seems likely that it works.

Now that I tested with JAX again, I see that I also need to relax the verifier for kCollectivePermuteStart and kCollectivePermuteDone (I am confused why I have not seen that previously) - with those ops allowed, I am getting the correct result from JAX.

jaro-sevcik:
Let me prepare a separate PR to also allow mixed precision for collective-permute-start and collective-permute-done.

jaro-sevcik:
Prepared the async version here: https://github.com/openxla/xla/pull/32905 (one of these PRs will have to be rebased when the other one lands, I think)

PR #32872, opened 2025-10-17
[DOC] New Performance Optimization Guide

athurdekoos:
📝 Summary of Changes
New document - `Performance Optimizing Operations`
-  Top level view of ops that can aid with performance optimization

🚀 Kind of Contribution
Please remove what does not apply: 📚 Documentation

### DIFF
--- a/docs/performance_optimizing_operations.md
+++ b/docs/performance_optimizing_operations.md
@@ -16,11 +16,9 @@ The following ops distribute work and/or aggregates results across replicas or d
 * [`CrossReplicaSum`](https://openxla.org/xla/operation_semantics#crossreplicasum) \- Performs [`AllReduce`](https://openxla.org/xla/operation_semantics#allreduce) with a summation computation.  
 * [`ReduceScatter`](https://openxla.org/xla/operation_semantics#reducescatter) \- A collective operation that effectively does an [`AllReduce`](https://openxla.org/xla/operation_semantics#allreduce) and then scatters the result by splitting it into shard blocks and replica groups.
 
-
 ## Reduced Precision
 
 Ops performing truncating low-order mantissa bits and/or exponent bits to emulate lower precision.
- 
 
 * [ReducePrecision](https://openxla.org/xla/operation_semantics#reduceprecision) \- Models the effect of converting floating-point values to a lower-precision format (such as IEEE-FP16) and back to the original format.
 
--- a/docs/performance_optimizing_operations.md
+++ b/docs/performance_optimizing_operations.md
@@ -1,0 +1,67 @@
+# Performance Optimizing Operations 
+
+This is a high level landing page that describes what operations or variants of operations can help make the computation faster, cheaper, or more memory efficient. Detailed information on all operations can be found on the [operation semantics](https://openxla.org/xla/operation_semantics) page.
+
+Operations on this page are organized by the inherent optimization of the operation.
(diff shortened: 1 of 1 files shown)


docs/performance_optimizing_operations.md
```diff
@@ -0,0 +1,65 @@
+# Performance Optimizing Operations
```
pifon2a, line 1: I am not sure what it means. What operations? HLO? but HLO by themselves do not optimize the performance, they just imply some computation.

docs/performance_optimizing_operations.md
```diff
@@ -0,0 +1,65 @@
+# Performance Optimizing Operations 
+
+This is a high level landing page that describes what operations or variants of operations can help make the computation faster, cheaper, or more memory efficient. Detailed information on all operations can be found on the [operation semantics](https://openxla.org/xla/operation_semantics) page.
```
pifon2a, line 3: Operations don't make anything faster or cheaper. Optimization passes do.

docs/performance_optimizing_operations.md
```diff
@@ -0,0 +1,65 @@
...
+
+Operations on this page are organized by the inherent optimization of the operation.
```
pifon2a, line 5: what's an "inherent" optimization?

pifon2a (commented):
What is the idea behind this doc? If it is a "performance opimization guide", then I would expect to see some useful information for users about the optimizations that they can tweak/add/remove/debug. There is a list of collectives here and different flavors of dots, but I am not sure if it is very helpful.

athurdekoos:
> What is the idea behind this doc? If it is a "performance optimization guide", then I would expect to see some useful information for users about the optimizations that they can tweak/add/remove/debug. There is a list of collectives here and different flavors of dots, but I am not sure if it is very helpful.

I agree, thank you for the feedback.  The doc as it stands is more of a collated list of ops that could be useful in optimization rather than a comprehensive guide. 

As for the idea behind the doc: 
My understanding is that from a user perspective, [operation semantics](https://openxla.org/xla/operation_semantics) is a tad dense and long. Having a landing page to quickly give a 10,000 ft overview of the op may be useful to the user to know where to start looking.

docs/performance_optimizing_operations.md
athurdekoos: Would it be safe or reasonable to say that utilizing operations that specialize in a specific task incurs less overhead than using multiple operations to complete the same task? 

I'm not well versed in the optimization passes to know if those are always optimized out.

pifon2a:
> I agree, thank you for the feedback. The doc as it stands is more of a collated list of ops that could be useful in optimization rather than a comprehensive guide.

All ops are useful. The program is a graph of ops and the optimizations are transformations of one graph into another, but with a better performance. That's why just listing the ops does not work.

> As for the idea behind the doc:
My understanding is that from a user perspective, [operation semantics](https://openxla.org/xla/operation_semantics) is a tad dense and long. Having a landing page to quickly give a 10,000 ft overview of the op may be useful to the user to know where to start looking.

I think this particular documentation page should be written by some engineer. Or at least a draft should be provided. I am not sure it is even possible to write smth like that without context.

athurdekoos:
Task to be rescoped/reworked on with the Engineering team and a new PR will be submitted.

PR #32934, opened 2025-10-21
[XLA:CPU][oneDNN] Fix failing oneDNN tests and F16 regressions

akhilgoe:
This PR addresses the side effects of a previous XLA commit (https://github.com/openxla/xla/commit/bd3f15e30fd588a53c8d7a594bafaa34d4f4adb5) that prevented any fusions for oneDNN F16 custom calls. In particular, this PR:
1. Expands the type promotion condition to include checks for oneDNN CCs too.
2. Use oneDNN float support for both oneDNN CCs and graph based execution.

### DIFF
--- a/xla/service/cpu/cpu_compiler.cc
+++ b/xla/service/cpu/cpu_compiler.cc
@@ -684,6 +684,9 @@ absl::Status CpuCompiler::RunHloPassesThroughLayoutAssn(
           .debug_options()
           .xla_cpu_experimental_onednn_custom_call() &&
       IsOneDnnCompatible(is_aot_compile);
+  bool use_onednn_graph =
+      module->config().debug_options().xla_cpu_use_onednn() &&
+      IsOneDnnCompatible(is_aot_compile);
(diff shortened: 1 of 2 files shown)


penpornk (approved):
Sorry for the delay and thank you for the PR!

xla/service/cpu/cpu_compiler.cc
```diff
@@ -684,6 +684,9 @@ absl::Status CpuCompiler::RunHloPassesThroughLayoutAssn(
           .debug_options()
           .xla_cpu_experimental_onednn_custom_call() &&
       IsOneDnnCompatible(is_aot_compile);
+  bool use_onednn_graph =
```
akuegel: Can you move that down below into the #ifdef XLA_ONEDNN block? Otherwise we get a unused variable warning in builds without XLA_ONEDNN defined.

akhilgoe: Done

akuegel (approved):

PR #32970, opened 2025-10-21
[NVIDIA GPU] Set nccl max channels to 32 for blackwell

Tixxx:
📝 Summary of Changes
Starting NCCL 2.28, nccl reduced the default max channels for collectives on blackwell from 32 to 16.
We need to pin it to 32 in xla to avoid surprise perf regressions.

🎯 Justification
Without this, default xla will incur perf regression after nccl 2.28 on blackwell.

🚀 Kind of Contribution
🐛 Bug Fix

Ran all benchmarks on blackwell and observed no perf regressions with this change.

🧪 Unit Tests:
NA

🧪 Execution Tests:
All benchmarks under tools/benchmark

### DIFF
--- a/xla/backends/gpu/collectives/nccl_collectives.cc
+++ b/xla/backends/gpu/collectives/nccl_collectives.cc
@@ -105,7 +105,8 @@ NcclCollectives::GetCliqueIdCallback(const CliqueIdCallback* clique_id_callback,
   return local_callback;
 }
 
-static ncclConfig_t AsNcclConfig(const GpuCollectives::Config& config) {
+static ncclConfig_t AsNcclConfig(const GpuCollectives::Config& config,
+                                 const se::StreamExecutor* stream_executor) {
   ncclConfig_t comm_config = NCCL_CONFIG_INITIALIZER;
   comm_config.blocking = config.blocking_communicators ? 1 : 0;
 #if !defined(TENSORFLOW_USE_ROCM) || TF_ROCM_VERSION > 50700
@@ -114,6 +115,16 @@ static ncclConfig_t AsNcclConfig(const GpuCollectives::Config& config) {
   if (config.max_nchannels > 0) {
     VLOG(1) << "Maximum number of channels is set to: " << comm_config.maxCTAs;
     comm_config.maxCTAs = config.max_nchannels;
+  } else if (stream_executor->GetDeviceDescription()
+                 .cuda_compute_capability()
+                 .IsAtLeastBlackwell()) {
+#if (NCCL_VERSION_CODE >= 22800)
+    // Future NCCL versions will reduce the default max number of channels on
+    // Blackwell to 16. We need to manually set it to 32 here to avoid surprise
+    // perf regressions.
+    VLOG(1) << "Setting max number of channels to 32 on Blackwell.";
+    comm_config.maxCTAs = 32;
+#endif  // NCCL_VERSION_CODE >= 22800
   }
   return comm_config;
 }
@@ -154,7 +165,6 @@ NcclCollectives::CreateCommunicators(const CliqueKey& clique_key,
         "async_execution is false. Non-blocking communicators require "
         "asynchronous execution.");
   }
-  ncclConfig_t comm_config = AsNcclConfig(gpu_config);
 
   // make_comm returns a new ncclComm_t.
   auto make_comm = [&](int i) -> absl::StatusOr<ncclComm_t> {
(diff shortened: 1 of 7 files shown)


xla/backends/gpu/collectives/nccl_collectives.cc
```diff
@@ -114,6 +115,16 @@ static ncclConfig_t AsNcclConfig(const GpuCollectives::Config& config) {
   if (config.max_nchannels > 0) {
     VLOG(1) << "Maximum number of channels is set to: " << comm_config.maxCTAs;
     comm_config.maxCTAs = config.max_nchannels;
+  } else if (stream_executor->GetDeviceDescription()
+                 .cuda_compute_capability()
+                 .IsAtLeastBlackwell()) {
```
felixwqp: In case future generation doesn't need 32 channel, for the sake of maintenance, should we limit the scope to blackwell?  s/`IsAtLeastBlackwell`/`IsBlackwell`

felixwqp:
Thank you TJ for enabling this! 

I understand we haven't onboarded 2.28 yet for XLA. But I want to learn is NVidia side can test this change with NCCL 2.28 on blackwell to understand the 32 channel's impact on representative benchmark suite?

felixwqp (commented):
Thank you TJ!

Tixxx:
> Thank you TJ for enabling this!
> 
> I understand we haven't onboarded 2.28 yet for XLA. But I want to learn is NVidia side can test this change with NCCL 2.28 on blackwell to understand the 32 channel's impact on representative benchmark suite?

Hi Felix, yes i have verified this change on blackwell including B200 and GB200 for benchmarks hlo_llama31_8b_bf16_1x8.hlo, hlo_llama31_8b_fp8_1x8.hlo and hlo_mixtral_8x7b_bf16_1x8.hlo, there's no regression observed.
Without this change, there would be 1-3% regressions for the benchmarks mentioned above.

xla/backends/gpu/collectives/nccl_collectives.cc
Tixxx: Ok sure, let me change it

felixwqp (approved):

xla/backends/gpu/collectives/nccl_collectives.cc
```diff
@@ -114,6 +115,16 @@ static ncclConfig_t AsNcclConfig(const GpuCollectives::Config& config) {
   if (config.max_nchannels > 0) {
     VLOG(1) << "Maximum number of channels is set to: " << comm_config.maxCTAs;
     comm_config.maxCTAs = config.max_nchannels;
+  } else if (stream_executor->GetDeviceDescription()
+                 .cuda_compute_capability()
+                 .IsBlackwell()) {
+#if (NCCL_VERSION_CODE >= 22800)
```
beckerhe: As far as I understand this doesn't need to be a preprocessor `#if`. Can you just make this a normal C++ if-branch. That makes testing this code easier.

Tixxx: thanks, i updated it

xla/backends/gpu/collectives/nccl_collectives.cc
```diff
@@ -114,6 +115,15 @@ static ncclConfig_t AsNcclConfig(const GpuCollectives::Config& config) {
   if (config.max_nchannels > 0) {
     VLOG(1) << "Maximum number of channels is set to: " << comm_config.maxCTAs;
     comm_config.maxCTAs = config.max_nchannels;
+  } else if (stream_executor->GetDeviceDescription()
+                 .cuda_compute_capability()
+                 .IsBlackwell() &&
+             NCCL_VERSION_CODE >= NCCL_VERSION(2, 28, 0)) {
```
beckerhe: Now that I'm thinking about this: This should probably be a runtime check using ncclGetVersion because for XLA-users like JAX NCCL is dynamically linked and the version we compile against is not necessarily the version the is used at runtime. WDYT?

Tixxx: thanks, i think that makes sense. updated

felixwqp (approved):

xla/backends/gpu/collectives/nccl_collectives.cc
```diff
@@ -105,15 +105,28 @@ NcclCollectives::GetCliqueIdCallback(const CliqueIdCallback* clique_id_callback,
   return local_callback;
 }
 
-static ncclConfig_t AsNcclConfig(const GpuCollectives::Config& config) {
+static absl::StatusOr<ncclConfig_t> AsNcclConfig(
+    const GpuCollectives::Config& config,
+    const se::StreamExecutor* stream_executor) {
```
felixwqp, line 111: nit: no header providing "stream_executor::StreamExecutor" is directly included

can you add `#include "third_party/tensorflow/compiler/xla/stream_executor/stream_executor.h"`

Tixxx: thanks. added the include path

beckerhe (approved):

PR #32993, opened 2025-10-22
[XLA:GPU] Change the NCCL command to have the highest priority in command buffer

shawnwang18:
📝 Summary of Changes
Change the collective command buffer cmd to have the highest priority by default.

🎯 Justification
collective has the highest priority should have more compute-communicate overlap

🚀 Kind of Contribution
⚡️ Performance Improvement,

All collective command buffer unitests should have covered the fix in this PR.

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/backends/gpu/runtime/command_buffer_cmd.cc
@@ -1999,7 +1999,7 @@ CommandBufferCmd::BufferUseVector CustomCallCmd::buffers() const {
 CollectiveCmd::CollectiveCmd(
     CommandBufferCmdType cmd_type, CollectiveConfig config,
     std::shared_ptr<CollectiveThunk::AsyncEvents> async_events)
-    : CommandBufferCmd(cmd_type),
+    : CommandBufferCmd(cmd_type, se::StreamPriority::Highest),
       config_(std::move(config)),
       async_events_(std::move(async_events)) {}
 


felixwqp:
- What workload is this change motivated by and why is it important?  (: Can I assume it's due to the potential regression brought up during the summit, right? 
- How can we measure and track the performance delta for this PR?  Can you provide speedups you measure on one of the HLO benchmarks in `compiler/xla/tools/benchmarks/hlo/`? 


Specifically for this PR: `collective has the highest priority should have more compute-communicate overlap`, does this optimization requires the latency hiding scheduler to see the performance benefit?

What's the current use case that collective is scheduled as a command buffer for now?

shawnwang18:
> Some general asks for all PRs (that also apply to this PR):
> 
> * What workload is this change motivated by and why is it important?  (: Can I assume it's due to the potential regression brought up during the summit, right?
> * How can we measure and track the performance delta for this PR?  Can you provide speedups you measure on one of the HLO benchmarks in `compiler/xla/tools/benchmarks/hlo/`?
> 
> Specifically for this PR: `collective has the highest priority should have more compute-communicate overlap`, does this optimization requires the latency hiding scheduler to see the performance benefit?
> 
> What's the current use case that collective is scheduled as a command buffer for now?

Hi @felixwqp The motivation for this PR is that I observed some overlapping pattern on the device like below: 

image

Where you can see that NCCL kernel starts later than compute kernel 'nvjet_sm100.....', it seems that if NCCL starts earlier , than the overlapping looks better. 

I am checking whether change the graph node's priority can help here, and please hold for a moment for this PR to confirm the perf impact of this PR change.

PR #33016, opened 2025-10-22
[ROCm] correct LHS logic with force_earliest_schedule()

i-chaochen:
📝 Summary of Changes
To be consistent with `force_earliest_schedule` logic.

🎯 Justification

Since we want `gpu_config->force_earliest_schedule()` to schedule the node as early as possible, we shouldn't use `node.SetForceDelay()` here.

🚀 Kind of Contribution

@xla-rotation please check this PR, thanks!


i-chaochen:
Hi @xla-rotation @felixwqp @frgossen could you please look this PR? thanks!

i-chaochen:
Hi @xla-rotation please check this PR

seherellis (changes requested):
LHS schedules instructions from ROOT to parameters, so "delay" means it will be executed as early as possible. There is no logical error.

You can check the explanation in ProfileGuidedCostModelWithForceEarliestSchedule test with that specific backend config.

i-chaochen:
Thanks @seherellis explain! Yes, it's counterintuitive in the beginning, now I c as it's backward-scheduling and this is correct one.

PR #33073, opened 2025-10-24
[XLA:GPU] add a warm up iteration for command buffer thunk

shawnwang18:
📝 Summary of Changes

This PR fix the issue that NCCL may call cuda-graph un-supported host side API during graph capturing (e.g. `cuCtxEnablePeerAccess`), this will break XLA cuda graph run.  To work around the issue, this PR introduces a warm up iteration for command buffer thunk, during warm up iteration, command buffer thunk are executed through normal thunks.  The warm up iteration will do the proper NCCL setup, so later iterations running through command buffer does not need to call NCCL setup APIs.

So for command buffer thunk, the execution phase are: 

1st iteration:   warm up (run through normal thunks)
2nd iteration:   cuda graph creation iteration. 
3rd- iteration:  possible cuda graph update iteration
 

🎯 Justification
The warm up iteration is required to setup proper NCCL connections, so the afterwards cuda-graph iterations does not need to call un-supported cuda APIs. 

🚀 Kind of Contribution

🧪 Execution Tests:
xla/service/gpu/tests/command_buffer_test.cc

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_thunk.cc
+++ b/xla/backends/gpu/runtime/command_buffer_thunk.cc
@@ -157,6 +157,13 @@ absl::Status CommandBufferThunk::Initialize(const InitializeParams& params) {
     TF_RETURN_IF_ERROR(thunks_->Initialize(params));
   }
 
+  // If there are no thunks, or command buffer does not require initialization,
+  // we can mark warm up as done immediately.
+  if ((!thunks_ || !commands_.requires_initialization()) &&
+      !cmd_buffer->warmup_done) {
+    cmd_buffer->warmup_done = true;
+  }
+
   // Construct ExecuteParams with empty fields for everything that is not needed
   // for recording commands.
   Thunk::ExecuteParams execute_params(
@@ -177,9 +184,9 @@ absl::Status CommandBufferThunk::Initialize(const InitializeParams& params) {
   // If commands require initialization, we also record them into the command
   // buffer before execution. This is required to guarantee that collective
   // commands recorded on all participating ranks to avoid deadlocks.
-  if (cmd_buffer->command_buffer->state() ==
-          se::CommandBuffer::State::kCreate ||
-      commands_.requires_initialization()) {
+  if (cmd_buffer->warmup_done && (cmd_buffer->command_buffer->state() ==
+                                      se::CommandBuffer::State::kCreate ||
+                                  commands_.requires_initialization())) {
     VLOG(3) << "Initialize command buffer on device #"
             << params.executor->device_ordinal()
             << " by recoding command buffer cmd sequence"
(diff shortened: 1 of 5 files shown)


apivovarov:
The PR title and description seem misleading. Could you please update it to indicate that this is a test-only change and that no changes to actual XLA behavior are expected?

shawnwang18:
> The PR title and description seem misleading. Could you please update it to indicate that this is a test-only change and that no changes to actual XLA behavior are expected?

My bad, the PR lost some commits on the design file, let me push that.

shawnwang18:
@apivovarov just submitted the implementation files, but seems the pipeline is broken?

shawnwang18:
@apivovarov can you take a look again?

xla/backends/gpu/runtime/command_buffer_thunk.cc
```diff
@@ -238,6 +245,14 @@ absl::Status CommandBufferThunk::ExecuteOnStream(const ExecuteParams& params) {
 
   absl::MutexLock lock(cmd_buffer->mutex);
 
+  // warm up iteration, run through thunks if they are present.
+  if (!cmd_buffer->warmup_done && thunks_) {
+    VLOG(2) << "Executing warm up iteration of command buffer thunk";
+    TF_RETURN_IF_ERROR(thunks_->ExecuteOnStream(params));
```
apivovarov, line 251: I'm a bit concerned that we holding the command buffer mutex while calling ExecuteOnStream. I'd release the mutex before executing the thunks, then re-acquire to set the flag. @ezhulenev What you think?

shawnwang18: I think holding the lock here to do warm up is OK, because if do not hold the lock the here, then another thread may also tries to do the warm up during warm up running, that's not necessary. 

The warm up is only performed during the first iteration, even hold the hold to do the warm up, that is acceptable.

apivovarov: Eugene Zhulenev, Yesterday 3:28 PM
I think that's ok, just for the warmup

shawnwang18: @apivovarov Could you approve this PR if there are no other unresolved issues?

apivovarov (approved):

xla/backends/gpu/runtime/command_buffer_thunk.cc
apivovarov: done

xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -17,10 +17,13 @@ limitations under the License.
 #include <utility>
 #include <variant>
 #include <vector>
+#include <memory>
+#include <optional>
 
 #include <gtest/gtest.h>
 #include "absl/strings/ascii.h"
 #include "absl/strings/string_view.h"
+#include "absl/types/span.h"
```
akuegel, line 26: This needs a corresponding dependency in the BUILD file

akuegel (approved):

PR #33106, opened 2025-10-24
Rename NumericOptions to EngineOptions and add field to only select cuDNN plans that support CUDA graphs.

dimvar:
Defaults to false.

This fixes two test failures on GB200 and Spark,
//xla/backends/gpu/runtime:cuda_command_buffer_thunk_test_nvgpu_any //xla/stream_executor/cuda:cuda_command_buffer_test_nvgpu_any where cuDNN was selecting the cuBLAS backend that does not support CUDA graph.

Also, in cuda_dnn.cc, moves the notes after the call to create_execution_plans, otherwise they have no effect.

Thanks to Emil Gilliam from the cuDNN team for guidance.


dimvar:
CC @egilliam-nv

xla/stream_executor/numeric_options.h
egilliam-nv, line 1: None of the following is blocking this PR, just some possible suggestions:

* This struct now contains "numeric" and "behavior" options together, which cuDNN considers 2 separate things. Consider either separating them into 2 structs, or just renaming this struct (and the file) to something like `EngineOptions`.
* Many of your call sites include the default values of the fields, which is error-prone (if you get one of the defaults wrong at the call-site), and now you've had to change some call sites to include `require_command_buffer`. Consider one of the following:
    
    * Builder pattern: Provide "set_" methods, like this:

```
struct EngineOptions {
  // If true, the op must be deterministic
  bool require_determinism = false;
  // If true, float32 inputs can be rounded to TensorFloat-32 precision
  bool allow_tf32 = true;
  // If true, the execution plan selected must support command buffer
  // construction.
  bool require_command_buffer = false;

  EngineOptions& set_require_determinism(bool value) {
      require_determinism = value;
      return *this;
  }
  // etc.
}

Then at the call site:

`..., EngineOptions().set_the_one_field_I_need(true), ...`

   * If you have C++20: make the struct just contain the fields, then you can use this notation at the callsite setting only the non-default fields:

`..., EngineOptions{.require_determinism=true}`

dimvar:
I like the name EngineOptions. I think it's ok to keep this as a single struct in XLA; we don't need to be exactly duplicating the definitions in the cudnn frontend, esp. since it's unlikely that we'll ever need to support all options in the notes. @ezhulenev WDYT about renaming to EngineOptions?

https://github.com/NVIDIA/cudnn-frontend/blob/0b1577c8c83401237d601d0d0db5210506705396/include/cudnn_frontend_utils.h#L608

Re: the builder, I'm OK either way, but even if we do it, it should be a separate patch. It's unrelated to the goal of this PR.

ezhulenev:
+1 to rename the struct to something else, `EngineOptions` SGTM, because it's certainly not a numeric options anymore

dimvar:
Did the renaming. @ezhulenev PTAL

ezhulenev (approved):

PR #33117, opened 2025-10-25
Rename "forward compatible" capabilities to "family compatible", per NVIDIA naming.

dimvar:
See:
https://developer.nvidia.com/blog/nvidia-blackwell-and-nvidia-cuda-12-9-introduce-family-specific-architecture-features/

Family compatible supported was introduced in https://github.com/openxla/xla/commit/16b9d957ff9c4b177ccb0e736b9e5135b392749a.
It's not clear to me how someone can actually configure XLA to use sm_100f. Will look into that next.


beckerhe (approved):
This is great. Thank you!

PR #33125, opened 2025-10-27
Update .clang-format

aws-alexk:
📝 Summary of Changes
I updated the .clang-format file to have valid values.

🎯 Justification
change.
.clang-format file is giving me issues with my machine even from clean source
Example of error:
YAML:6:3: error: unexpected scalar Kind: Always ^ YAML:6:3: error: invalid boolean Kind: Always ^ Error reading .clang-format: Invalid argument

🚀 Kind of Contribution
Please remove what does not apply: ♻️ Cleanup

N/A

🧪 Unit Tests:
transformed HLO.
N/A

🧪 Execution Tests:
N/A

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,5 @@
 BasedOnStyle: Google
 Language: Cpp
 PointerBindsToType: true
-SortIncludes: Never
-AlignTrailingComments:
-  Kind: Always
+SortIncludes: false
+AlignTrailingComments: true


ddunl:
What version of clang-format was used to get the error? When we run it on CI we use 17, so I think if these changes are due to deprecations that happened after clang-format 17 we should change this + update CI.

(Example run from today where it passes with clang-format 17: https://github.com/openxla/xla/actions/runs/18846357364/job/53771346243)

aws-alexk:
Oops, I thought I was using a recent version but my project path was set using a very old clang-format version 11.
After updating to a newer version, I can confirm the original clang-format works.
Sorry for the confusion!

PR #33188, opened 2025-10-28
[ROCm] Fix hermetic tests when executing on rbe worker

alekstheod:
📝 Summary of Changes
Fix tests when executing on rbe

🎯 Justification
Fix hermetic dependencies so right libs are loaded.
Libs are put as a data dependency as many of the rocm libs may have their own
RPATH=$ORIGIN/..., they will try to load the libs from the place where they are.
This leads to a situation that one lib located in _solid_data dir will try to find its 
dependencies in the same dir. However dependency might be located in the data 
directory. Then either test can't load the lib or tries to load one from the system libs.

🚀 Kind of Contribution

Build fix

🧪 Unit Tests:
Build fix

🧪 Execution Tests:
Build fix


beckerhe (approved):

chsigg (commented):
`external/local_config_rocm` probably shouldn't be hardcoded.

`linkopts` supports various expansions (https://bazel.build/reference/be/c-cpp#cc_binary.linkopts), so you can use `$(location)` etc.

Also, `cc_import` might be easier to use than cc_library.

Instead of sprinkling these replacements across the BUILD file, it would be better to have a separate file that just defines these, and template-expand that file only.

These are just some suggestions if you are inclined to clean up this corner a bit.

alekstheod:
> `cc_import`

I am not sure this would serve our needs. The reason we use that "hack" put the lib into the data and have rpath pointing to it is that this lib expects some data files being located relative to its location. Not sure how cc_import can solve this issue.

alekstheod:
> `linkopts` supports various expansions (https://bazel.build/reference/be/c-cpp#cc_binary.linkopts), so you can use `$(location)` etc.

I think I may try this? linkopts = ["-Wl,-rpath,$ORIGIN"]
I would do it in separate pr though...

chsigg:
Yes please, separate PR. Sorry, I forgot to be explicit about this.

PR #33212, opened 2025-10-28
[ROCm] Enable HloEvaluator for more complex ops

nurmukhametov:
This PR enables the HloEvaluator to handle complex numbers in more operations (trigonometric and hyperbolic). Tests are added to check if folding works for these operations.

It also disables `constant_folding` in `complex_unary_op_test.cc` test, as it was intended to check the accuracy of backends implementations. If constant folding remains enabled, the test ends up checking the accuracy of `libstdc++` implementations for `tan`, `asin`, and `asinh` instead

### DIFF
--- a/xla/tests/complex_unary_op_test.cc
+++ b/xla/tests/complex_unary_op_test.cc
@@ -40,6 +40,13 @@ class ComplexUnaryOpTest
     : public ClientLibraryTestRunnerMixin<
           HloPjRtInterpreterReferenceMixin<HloPjRtTestBase>> {
  protected:
+  // Disable constant folding to ensure we test the actual backend implementation.
+  // Otherwise, constant folding pre-computes results using HloEvaluator's
+  // reference implementation (std c++), not the backend under test.
+  void SetUp() override {
+    ClientLibraryTestRunnerMixin::SetUp();
+    mutable_debug_options()->add_xla_disable_hlo_passes("constant_folding");
+  }
   template <typename T, size_t index, typename... Types>
   std::vector<T> get_column(const std::vector<std::tuple<Types...>>& table) {
     std::vector<T> column;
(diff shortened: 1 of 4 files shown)


xla/tests/complex_unary_op_test.cc
```diff
@@ -40,6 +40,13 @@ class ComplexUnaryOpTest
     : public ClientLibraryTestRunnerMixin<
           HloPjRtInterpreterReferenceMixin<HloPjRtTestBase>> {
  protected:
+  // Disable constant folding to ensure we test the actual backend implementation.
+  // Otherwise, constant folding pre-computes results using HloEvaluator's
+  // reference implementation (std c++), not the backend under test.
+  void SetUp() override {
```
bchetioui, line 46: This file has no other change. Is this change necessary, and if so can you please explain why?

If not, please remove this from the diff.

bchetioui (changes requested):

xla/tests/complex_unary_op_test.cc
nurmukhametov: Yes, it is necessary. If not changed, tests from this file fail.

This test file is designed in a way that it contains code that can be folded by `HloEvaluator`. This PR enables `HloEvaluator` for `Tan`/`Asin`/`Asinh`, so it folds here using the C++/C standard library implementation. At least in my environment, the results of stdlib `Tan` were different from the expected ones. I don't think that this test file needs to test C/C++ library accuracy, because it was designed to test [these implementations](https://github.com/openxla/xla/blob/01a696b2048a1197b6aae747c95e584a7f34c9c6/xla/service/elemental_ir_emitter.cc#L1262).

However, these tests can actually verify not only the mentioned implementation in `elemental_ir_emitters.cc` but also [implementations](https://github.com/nurmukhametov/llvm-project/blob/72616c5ca99bcf4d93d423642387016cb88409bb/mlir/lib/Conversion/ComplexToStandard/ComplexToStandard.cpp#L704) from MLIR depending on the value of the `--xla_cpu_use_fusion_emitters` flag.

bchetioui: Ah, I see; thank you for explaining.

bchetioui (approved):

bchetioui:
Thanks! Apologies for the late stamp, I lost the notification about this PR.

PR #33238, opened 2025-10-28
Clear multi-stream attributes in GemmFusionAutotuner

sfvaroglu:
📝 Summary of Changes
Clear multi-stream attributes when extracting HLO instructions into new modules for autotuning. These stream IDs reference streams from the original execution context that don't exist in the extracted module's context, causing segfaults. We see this behavior when we enable collective-matmul and triton gemms (`xla_gpu_threshold_for_windowed_einsum_mib=0, xla_gpu_enable_triton_gemm=true`) in llama3-8b with tensor parallelism.

🎯 Justification
During autotuning, instructions are extracted into standalone modules that run in a new execution context with only a single default stream. The original operation_queue_id values (e.g., 131, 8, 7) map to streams in the original module's `ExecuteParams::additional_compute_streams` map, which is empty in the autotuner's execution context. Resetting stream IDs to 0 ensures all operations use the default stream, which always exists and is appropriate for autotuning isolated kernels.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

N/A

🧪 Unit Tests:
Added a unit test.

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
+++ b/xla/service/gpu/autotuning/gemm_fusion_autotuner.cc
@@ -273,6 +273,23 @@ HloCostAnalysis::Options PriorityFusionOptions() {
   return {.count_multiple_input_accesses = true};
 }
 
+// Clear stream IDs that reference streams in the original execution context,
+// which don't exist in the extracted module's execution context.
+void ClearMultiStreamAttributes(HloModule* module) {
+  for (HloComputation* comp : module->computations()) {
+    for (HloInstruction* instr : comp->instructions()) {
+      if (instr->has_backend_config()) {
+        auto gpu_config = instr->backend_config<GpuBackendConfig>();
+        if (gpu_config.ok()) {
+          gpu_config->set_operation_queue_id(0);
+          gpu_config->clear_wait_on_operation_queues();
+          (void)instr->set_backend_config(*gpu_config);
+        }
+      }
+    }
+  }
+}
+
 absl::StatusOr<std::unique_ptr<HloModule>> TritonGemmAutotuneExtractor(
     const TritonGemmConfig& config,
     const se::DeviceDescription& gpu_device_info,
@@ -287,6 +304,7 @@ absl::StatusOr<std::unique_ptr<HloModule>> TritonGemmAutotuneExtractor(
         false);
   }
   new_module->mutable_config().set_debug_options(debug_opts);
+  ClearMultiStreamAttributes(new_module.get());
 
   HloComputation* entry_computation = new_module->entry_computation();
   HloInstruction* cloned_dot_fusion = entry_computation->root_instruction();
@@ -335,6 +353,7 @@ absl::StatusOr<std::unique_ptr<HloModule>> CublasGemmAutotuneExtractor(
   std::unique_ptr<HloModule> new_module =
       ExtractComputationIntoNewModule(*fusion_computation);
   new_module->mutable_config().set_debug_options(debug_opts);
+  ClearMultiStreamAttributes(new_module.get());
 
   auto* dot = hlo_query::GetFirstInstructionWithOpcode(
       *new_module->entry_computation(), HloOpcode::kDot);
@@ -391,6 +410,7 @@ absl::StatusOr<std::unique_ptr<HloModule>> CustomFusionKernelAutotuneExtractor(
   std::unique_ptr<HloModule> new_module =
       ExtractComputationIntoNewModule(*fusion_computation);
   new_module->mutable_config().set_debug_options(debug_opts);
+  ClearMultiStreamAttributes(new_module.get());
 
   CustomKernelFusionRewriter rewriter(&config.GetDeviceDescription());
   PriorityFusion fusion_pass(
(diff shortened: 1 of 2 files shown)


akuegel (commented):
I think that maybe this logic should be put into hlo_decomposer? First, we may face the same problem when we extract an instruction or computation from a module for debugging reasons (using the methods from hlo_decomposer). Second, your added test would pass even without the new logic, as far as I can tell. When calling ClearMutliStreamAttributes from ExtractInstructionIntoNewModule and ExtractComputationIntoNewModule, it also becomes testable.

sfvaroglu:
> I think that maybe this logic should be put into hlo_decomposer? First, we may face the same problem when we extract an instruction or computation from a module for debugging reasons (using the methods from hlo_decomposer). Second, your added test would pass even without the new logic, as far as I can tell. When calling ClearMutliStreamAttributes from ExtractInstructionIntoNewModule and ExtractComputationIntoNewModule, it also becomes testable.

This change is specific to GPUs, that's why I didn't put it into hlo_decomposer.

The unit test fails without this change. Are you seeing otherwise?

akuegel:
> > I think that maybe this logic should be put into hlo_decomposer? First, we may face the same problem when we extract an instruction or computation from a module for debugging reasons (using the methods from hlo_decomposer). Second, your added test would pass even without the new logic, as far as I can tell. When calling ClearMutliStreamAttributes from ExtractInstructionIntoNewModule and ExtractComputationIntoNewModule, it also becomes testable.
> 
> This change is specific to GPUs, that's why I didn't put it into hlo_decomposer.
> 
> The unit test fails without this change. Are you seeing otherwise?

I only looked at the test, and it seemed to check only something I expect to be happening even without the change: that the triton fusion preserves the attribute. But I copied the test and noticed that it indeed does fail with a segfault without your change, but this is sort of subtle and does not have anything to do with the CHECK lines there. I debugged it and found that the bug is in this line: https://github.com/openxla/xla/blob/5aff798706c762c9ff125092cb8021ccddd31b6b/xla/service/gpu/transforms/stream_attribute_annotator.cc#L159
I have sent out a fix for this. Maybe that already resolves the issue with autotuning that you were seing, and this PR becomes obsolete? In any case, with my fix applied, your added test would succeed without the additional changes from this PR.

akuegel:
> > > I think that maybe this logic should be put into hlo_decomposer? First, we may face the same problem when we extract an instruction or computation from a module for debugging reasons (using the methods from hlo_decomposer). Second, your added test would pass even without the new logic, as far as I can tell. When calling ClearMutliStreamAttributes from ExtractInstructionIntoNewModule and ExtractComputationIntoNewModule, it also becomes testable.
> > 
> > 
> > This change is specific to GPUs, that's why I didn't put it into hlo_decomposer.
> > The unit test fails without this change. Are you seeing otherwise?
> 
> I only looked at the test, and it seemed to check only something I expect to be happening even without the change: that the triton fusion preserves the attribute. But I copied the test and noticed that it indeed does fail with a segfault without your change, but this is sort of subtle and does not have anything to do with the CHECK lines there. I debugged it and found that the bug is in this line:
> 
> https://github.com/openxla/xla/blob/5aff798706c762c9ff125092cb8021ccddd31b6b/xla/service/gpu/transforms/stream_attribute_annotator.cc#L159
> 
> 
> I have sent out a fix for this. Maybe that already resolves the issue with autotuning that you were seing, and this PR becomes obsolete? In any case, with my fix applied, your added test would succeed without the additional changes from this PR.

The fix has been merged (https://github.com/openxla/xla/commit/6048a869f1585fa7bf0a507a549ef0e07eec90c0). You can try it out, the test case you added should work now without your additional changes. If you think your additional changes are needed, please make them in a way so that they are testable.

sfvaroglu:
> > > > I think that maybe this logic should be put into hlo_decomposer? First, we may face the same problem when we extract an instruction or computation from a module for debugging reasons (using the methods from hlo_decomposer). Second, your added test would pass even without the new logic, as far as I can tell. When calling ClearMutliStreamAttributes from ExtractInstructionIntoNewModule and ExtractComputationIntoNewModule, it also becomes testable.
> > > 
> > > 
> > > This change is specific to GPUs, that's why I didn't put it into hlo_decomposer.
> > > The unit test fails without this change. Are you seeing otherwise?
> > 
> > 
> > I only looked at the test, and it seemed to check only something I expect to be happening even without the change: that the triton fusion preserves the attribute. But I copied the test and noticed that it indeed does fail with a segfault without your change, but this is sort of subtle and does not have anything to do with the CHECK lines there. I debugged it and found that the bug is in this line:
> > https://github.com/openxla/xla/blob/5aff798706c762c9ff125092cb8021ccddd31b6b/xla/service/gpu/transforms/stream_attribute_annotator.cc#L159
> > 
> > I have sent out a fix for this. Maybe that already resolves the issue with autotuning that you were seing, and this PR becomes obsolete? In any case, with my fix applied, your added test would succeed without the additional changes from this PR.
> 
> The fix has been merged ([6048a86](https://github.com/openxla/xla/commit/6048a869f1585fa7bf0a507a549ef0e07eec90c0)). You can try it out, the test case you added should work now without your additional changes. If you think your additional changes are needed, please make them in a way so that they are testable.

Thank you for the fix, closing this PR.

PR #33240, opened 2025-10-28
Adding a delayMoveToHost heuristic to LHS and related tests. 

mingxu1067:
📝 Summary of Changes

- Adding a heurisitic to GPU-scheduler for having better MoveToHost overlapping.

🎯 Justification
This could help hide D2H/H2D data movement behind computations.

🚀 Kind of Contribution
✨ New Feature

📊 Benchmark (for Performance Improvements)

🧪 Unit Tests:
Added unit-testes.

🧪 Execution Tests:
Maxtext/DeepSeek3-671B

PR #33269, opened 2025-10-29
Addin a knob to control the limitation of async-compute resource.

mingxu1067:
📝 Summary of Changes

-  Addin a knob to control the limitation of async-compute resource. This switch provides ample flexibility for control, enabling more asynchronous computations to execute concurrently. In host-offloading experiments, increasing this value effectively overlaps device-to-host (D2H) transfers with other computations, resulting in improved performance.﻿

🎯 Justification
This could allow users to control how many in-fligh async-computation in LHS.

🚀 Kind of Contribution
✨ New Feature,

### DIFF
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -809,6 +809,8 @@ message DebugOptions {
   //   coll.2-done = collective(coll.2-start)
   optional int32 xla_gpu_experimental_parallel_collective_overlap_limit = 336;
 
+  optional int32 xla_gpu_experimental_parallel_async_compute_limit = 465;
+
   optional PipelineParallelismOptLevel
       xla_gpu_experimental_pipeline_parallelism_opt_level = 351;
 
@@ -1436,7 +1438,7 @@ message DebugOptions {
   // Note: when adding a new flag, please add it to one of the hardware-specific
   // or hardware-agnostic sections at the top of this proto message.
 
-  // Next id: 465
+  // Next id: 466
 
   // Extra options to pass to the compilation backend (e.g. LLVM); specific
   // interpretation of these values is left to the backend.
--- a/xla/xla.proto
+++ b/xla/xla.proto
@@ -809,6 +809,11 @@ message DebugOptions {
   //   coll.2-done = collective(coll.2-start)
   optional int32 xla_gpu_experimental_parallel_collective_overlap_limit = 336;
 
+  // This controls the limitation of async-compute resource.
+  // This provides ample flexibility for control, enabling more asynchronous
+  // computations to execute concurrently. In host-offloading experiments,
+  // increasing this value effectively overlaps device-to-host (D2H) transfers
+  // with other computations, resulting in improved performance.
   optional int32 xla_gpu_experimental_parallel_async_compute_limit = 465;
 
   optional PipelineParallelismOptLevel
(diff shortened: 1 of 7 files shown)


xla/xla.proto
```diff
@@ -685,6 +685,8 @@ message DebugOptions {
   //   coll.2-done = collective(coll.2-start)
   optional int32 xla_gpu_experimental_parallel_collective_overlap_limit = 336;
 
+  optional int32 xla_gpu_experimental_parallel_async_compute_limit = 423;
```
felixwqp: can you add a comments to explain the purpose of this flag? 

like how the CL above does.

felixwqp:
Trying to understand the purpose of this CL: 

for `, increasing this value effectively overlaps device-to-host (D2H) transfers with other computations`, can I assume both compute and D2H are treated as the async-compute now, so this is helpful the overlap. 



Besides,  Some general asks for all PRs (that also apply to this PR):
- How can we measure and track the performance delta for this PR? Can you provide speedups you measure on one of the HLO benchmarks in compiler/xla/tools/benchmarks/hlo/?

xla/xla.proto
mingxu1067: Done

mingxu1067:
> for , increasing this value effectively overlaps device-to-host (D2H) transfers with other computations, can I assume both compute and D2H are treated as the async-compute now, so this is helpful the overlap.

IIUC, only DUS fused with D2H would be treated as async-compute. But we have many that kernels, and defulat value (2) would blokc scheduler to open `start`.

> How can we measure and track the performance delta for this PR? Can you provide speedups you measure on one of the HLO benchmarks in compiler/xla/tools/benchmarks/hlo/?

We measured in DeepSeekV3-671B implemented with Maxtext and are seeing ~10% speedup end-2-end.
> Can you add or point to an execution test that exercises this code path?

Could you advise what tests should we have for adding a flag?

felixwqp:
> We measured in DeepSeekV3-671B implemented with Maxtext and are seeing ~10% speedup end-2-end.
> 
> >

similar to the other CL, we would like adding this benchmark into our suite, so as to guard the optimization from this PR and related PR from future changes.

felixwqp:
> > Can you add or point to an execution test that exercises this code path?
> 
> Could you advise what tests should we have for adding a flag?

my understanding is this PR opens us user from setting parallel compute on more than 2 streams. If so, can you add test for this scenario, e.g. 3 stream or 4? 



Regarding the location, i think it depends on the use case you are trying to support here, e.g. in @Tixxx 's https://github.com/openxla/xla/pull/7854, test is added in /transforms/stream_attribute_annotator_test.cc for 2 stream cases. 


my intention is to understand the use case here more concretely so it can benefit users from wider community.

mingxu1067:
> similar to the other CL, we would like adding this benchmark into our suite, so as to guard the optimization from this PR and related PR from future changes.

I added a HLO in [pull/33240](https://github.com/openxla/xla/pull/33240).

> my understanding is this PR opens us user from setting parallel compute on more than 2 streams. If so, can you add test for this scenario, e.g. 3 stream or 4?

Could we wait [pull/33240](https://github.com/openxla/xla/pull/33240) merged then extend its tests for testing #stream more than 2?

felixwqp:
due to priority adjustment, pifon2a@(or his delegate) will collaborate  on the PR review as reviewer going forward. 

Feel free to ping me if any additional information is needed.

golechwierowicz (approved):

golechwierowicz:
@mingxu1067 can you rebase?

mingxu1067:
Hi @golechwierowicz, I rebased on the latest main. Kindly let me know if any modification suggestions. Thanks.

golechwierowicz (approved):

mingxu1067:
Hi @golechwierowicz, I rebased this PR on top of the latest main. Could you help review and merge? Thanks.

golechwierowicz (approved):

PR #33284, opened 2025-10-29
PjRt API changes for improved cross host transfers

rao-ashish:
📝 Summary of Changes
This PR adds methods `CrossHostSendBuffers` and `CrossHostReceiveBuffers` to PjRt clients as a new cross-host data transfer API. This new API enables optimizations when running on GPUs such as communicator caching and aggregating transfers with NCCL group calls. The new API evolved from discussions with @emilyfertig (for eg see discussion in #32074 , #32076 , #32295 ). 

🎯 Justification
The current implementation of cross-process device-puts on GPUs does not cache communicators for reuse across multiple transfers between the same sets of devices, and does not aggregate transfers of multiple arrays together with NCCL group calls. This PR adds cross-host transfer methods `CrossHostSendBuffers` and `CrossHostReceiveBuffers` that now have information about the global device IDs involved in the data transfer so that communicators can be cached. Further, since these methods operate on multiple arrays, transfers can be aggregated with NCCL group calls.

🚀 Kind of Contribution
⚡️ Performance Improvement

This PR only makes API changes necessary for future performance improvements such as communicator caching, so no benchmarks were ran for this PR. Follow-up PRs implementing performance improvements will include benchmarks.

🧪 Unit Tests:
Unit tests for the new API have been added inside `se_gpu_pjrt_client_test.cc`.

🧪 Execution Tests:
When building XLA with the code in this PR along with [a patch for `pjrt_ifrt/pjrt_client.cc`](https://gist.github.com/rao-ashish/ba5cd773906c8e428147bb80a3568912) so that it uses the new API, I verified that [these 4 correctness checks](https://gist.github.com/rao-ashish/24ac0df0cb18243c649ac535964b31b8) pass. The IFRT changes will be included in a follow-up PR once other PjRt clients have implemented the new API.

### DIFF
--- a/xla/pjrt/c_api_client/pjrt_c_api_client.cc
+++ b/xla/pjrt/c_api_client/pjrt_c_api_client.cc
@@ -989,6 +989,61 @@ PJRT_Transfers_CrossHostRecvNotifierInfo CppCrossHostRecvNotifierToC(
       }};
 }
 
+// Helper struct and method used to serialize shapes past the C API boundary.
+struct ShapesInfo {
+  std::vector<size_t> shape_num_dims;
+  std::vector<PJRT_Buffer_MemoryLayout*> layout_list;
+  std::vector<const int64_t*> num_dims;
+  std::vector<PJRT_Buffer_Type> element_type_list;
+};
+
+ShapesInfo MakeShapesInfo(absl::Span<const Shape> shapes) {
+  std::vector<size_t> shape_num_dims;
+  shape_num_dims.reserve(shapes.size());
+  std::vector<PJRT_Buffer_MemoryLayout*> layout_list;
+  layout_list.reserve(shapes.size());
+  std::vector<const int64_t*> num_dims;
+  num_dims.reserve(shapes.size());
+  std::vector<PJRT_Buffer_Type> element_type_list;
+  element_type_list.reserve(shapes.size());
+
+  for (int i = 0; i < shapes.size(); ++i) {
+    shape_num_dims.push_back(shapes[i].dimensions().size());
+
+    num_dims.push_back(shapes[i].dimensions().data());
+    element_type_list.push_back(
+        pjrt::ConvertToPjRtBufferType(shapes[i].element_type()));
+    // TODO(b/434246423): Enable this once ASAN failure is fixed.
+    // if (shapes[i].has_layout()) {
+    //   // this is messed up
+    //   auto& layout = shapes[i].layout();
+    //   TF_ASSIGN_OR_RETURN(
+    //       pjrt::BufferMemoryLayoutData c_layout_data,
+    //       pjrt::ConvertToBufferMemoryLayoutData(layout));
+    //   layout_list.push_back(&(c_layout_data.c_layout));
+    layout_list.push_back(nullptr);
+  }
+
+  return ShapesInfo{
+      .shape_num_dims = std::move(shape_num_dims),
+      .layout_list = std::move(layout_list),
+      .num_dims = std::move(num_dims),
+      .element_type_list = std::move(element_type_list),
+  };
+}
+
+// Helper method to convert a list of PJRT_Buffer* to a list of PjRtBuffer*.
+std::vector<std::unique_ptr<PjRtBuffer>> MakePjRtBuffersFromPJRT_Buffers(
+    PjRtCApiClient* client, PJRT_Buffer** c_buffers, size_t num_buffers) {
+  std::vector<std::unique_ptr<PjRtBuffer>> buffers;
+  buffers.reserve(num_buffers);
+  for (int i = 0; i < num_buffers; ++i) {
+    buffers.emplace_back(
+        std::make_unique<PjRtCApiBuffer>(client, c_buffers[i]));
+  }
+  return buffers;
+}
+
 absl::StatusOr<std::vector<std::unique_ptr<PjRtBuffer>>>
 PjRtCApiClient::MakeCrossHostReceiveBuffers(
     absl::Span<const Shape> shapes, PjRtDevice* device,
@@ -1006,35 +1061,13 @@ PjRtCApiClient::MakeCrossHostReceiveBuffers(
       PJRT_Transfers_PJRT_Client_MakeCrossHostReceiveBuffers_Args_STRUCT_SIZE;
   args.extension_start = nullptr;
   args.client = c_client_.get();
-  args.num_shapes = shapes.size();
-  std::vector<size_t> shape_num_dims;
-  shape_num_dims.reserve(shapes.size());
-  std::vector<PJRT_Buffer_MemoryLayout*> layout_list;
-  layout_list.reserve(shapes.size());
-  std::vector<const int64_t*> num_dims;
-  num_dims.reserve(shapes.size());
-  std::vector<PJRT_Buffer_Type> element_type_list;
-  element_type_list.reserve(shapes.size());
-  for (int i = 0; i < shapes.size(); ++i) {
-    shape_num_dims.push_back(shapes[i].dimensions().size());
 
-    num_dims.push_back(shapes[i].dimensions().data());
-    element_type_list.push_back(
-        pjrt::ConvertToPjRtBufferType(shapes[i].element_type()));
-    // TODO(b/434246423): Enable this once ASAN failure is fixed.
-    // if (shapes[i].has_layout()) {
-    //   // this is messed up
-    //   auto& layout = shapes[i].layout();
-    //   TF_ASSIGN_OR_RETURN(
-    //       pjrt::BufferMemoryLayoutData c_layout_data,
-    //       pjrt::ConvertToBufferMemoryLayoutData(layout));
-    //   layout_list.push_back(&(c_layout_data.c_layout));
-    layout_list.push_back(nullptr);
-  }
-  args.shape_num_dims = shape_num_dims.data();
-  args.num_dims = num_dims.data();
-  args.element_types = element_type_list.data();
-  args.layouts = layout_list.data();
+  ShapesInfo shapes_info = MakeShapesInfo(shapes);
+  args.num_shapes = shapes.size();
+  args.shape_num_dims = shapes_info.shape_num_dims.data();
+  args.num_dims = shapes_info.num_dims.data();
+  args.element_types = shapes_info.element_type_list.data();
+  args.layouts = shapes_info.layout_list.data();
 
   args.notifier = CppCrossHostRecvNotifierToC(c_api, std::move(notifier));
   args.device = tensorflow::down_cast<PjRtCApiDevice*>(device)->c_device();
@@ -1044,13 +1077,107 @@ PjRtCApiClient::MakeCrossHostReceiveBuffers(
   RETURN_STATUS_IF_PJRT_ERROR(
       extension->PJRT_Transfers_PJRT_Client_MakeCrossHostReceiveBuffers(&args),
       c_api);
-  std::vector<std::unique_ptr<PjRtBuffer>> buffers;
-  buffers.reserve(args.num_buffers);
-  for (int i = 0; i < args.num_buffers; ++i) {
-    buffers.emplace_back(std::unique_ptr<PjRtBuffer>(
-        std::make_unique<PjRtCApiBuffer>(this, args.buffers[i])));
+
+  return MakePjRtBuffersFromPJRT_Buffers(this, args.buffers,
+                                         temp_buffers.size());
+}
+
+absl::StatusOr<std::vector<Future<>>> PjRtCApiClient::CrossHostSendBuffers(
+    absl::Span<PjRtBuffer* const> buffers,
+    absl::Span<const PjRtGlobalDeviceId> dst_global_device_ids,
+    std::vector<CrossHostTransferKey> transfer_keys) {
+  // Get C API extension.
+  const PJRT_Api* c_api = pjrt_c_api();
+  PJRT_CrossHostTransfers_Extension* extension =
+      FindExtension<PJRT_CrossHostTransfers_Extension>(
+          PJRT_Extension_Type::PJRT_Extension_Type_CrossHostTransfers);
+  if (extension == nullptr) {
+    absl::Status error = absl::UnimplementedError(
+        "CrossHostSendBuffers is not implemented in this PJRT plugin.");
+    std::vector<Future<>> futures;
+    futures.reserve(buffers.size());
+    for (int i = 0; i < buffers.size(); ++i) {
+      futures.push_back(Future<>(error));
+    }
+    return futures;
   }
-  return buffers;
+
+  // Form inputs.
+  PJRT_Transfers_PJRT_Client_CrossHostSendBuffers_Args args;
+  args.struct_size =
+      PJRT_Transfers_PJRT_Client_CrossHostSendBuffers_Args_STRUCT_SIZE;
+  args.extension_start = nullptr;
+  args.client = c_client_.get();
+  args.num_buffers = buffers.size();
+
+  std::vector<PJRT_Buffer*> c_buffers;
+  c_buffers.reserve(buffers.size());
+  for (PjRtBuffer* buffer : buffers) {
+    c_buffers.push_back(
+        tensorflow::down_cast<const PjRtCApiBuffer*>(buffer)->c_buffer());
+  }
+
+  args.buffers = c_buffers.data();
+  args.dst_global_device_ids = dst_global_device_ids.data();
+  args.transfer_keys = transfer_keys.data();
+
+  auto send_events = std::vector<PJRT_Event*>(args.num_buffers);
+  args.send_events = send_events.data();
+
+  RETURN_STATUS_IF_PJRT_ERROR(
+      extension->PJRT_Transfers_PJRT_Client_CrossHostSendBuffers(&args), c_api);
+
+  std::vector<Future<>> send_futures;
+  for (int i = 0; i < args.num_buffers; ++i) {
+    send_futures.push_back(
+        pjrt::ConvertCEventToCppFuture(args.send_events[i], c_api));
+  }
+
+  return send_futures;
+}
+
+absl::StatusOr<std::vector<std::unique_ptr<PjRtBuffer>>>
+PjRtCApiClient::CrossHostReceiveBuffers(
+    xla::PjRtDevice* device, absl::Span<const xla::Shape> shapes,
+    absl::Span<const PjRtGlobalDeviceId> src_global_device_ids,
+    std::vector<CrossHostTransferKey> transfer_keys) {
+  // Get C API extension.
+  const PJRT_Api* c_api = pjrt_c_api();
+  PJRT_CrossHostTransfers_Extension* extension =
+      FindExtension<PJRT_CrossHostTransfers_Extension>(
+          PJRT_Extension_Type::PJRT_Extension_Type_CrossHostTransfers);
+  if (extension == nullptr) {
+    return absl::UnimplementedError(
+        "CrossHostReceiveBuffers is not implemented in this PJRT plugin.");
+  }
+
+  // Form inputs.
+  PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers_Args args;
+  args.struct_size =
+      PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers_Args_STRUCT_SIZE;
+  args.extension_start = nullptr;
+  args.client = c_client_.get();
+
+  ShapesInfo shapes_info = MakeShapesInfo(shapes);
+  args.num_shapes = shapes.size();
+  args.shape_num_dims = shapes_info.shape_num_dims.data();
+  args.num_dims = shapes_info.num_dims.data();
+  args.element_types = shapes_info.element_type_list.data();
+  args.layouts = shapes_info.layout_list.data();
+
+  args.device = tensorflow::down_cast<PjRtCApiDevice*>(device)->c_device();
+  args.src_global_device_ids = src_global_device_ids.data();
+  args.transfer_keys = transfer_keys.data();
+
+  std::vector<PJRT_Buffer*> temp_buffers(shapes.size());
+  args.buffers = temp_buffers.data();
+
+  RETURN_STATUS_IF_PJRT_ERROR(
+      extension->PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers(&args),
+      c_api);
+
+  return MakePjRtBuffersFromPJRT_Buffers(this, args.buffers,
+                                         temp_buffers.size());
 }
 
 class PjRtCApiAsyncHostToDeviceTransferManager
--- a/xla/pjrt/extensions/cross_host_transfers/pjrt_c_api_cross_host_transfers_extension.h
+++ b/xla/pjrt/extensions/cross_host_transfers/pjrt_c_api_cross_host_transfers_extension.h
@@ -30,10 +30,58 @@ extern "C" {
 // are supported with the PjRtClient::MakeCrossHostReceiveBuffers() and
 // PjRtBuffer::CopyToRemoteDevice() APIs.
 
-#define PJRT_API_CROSS_HOST_TRANSFERS_EXTENSION_VERSION 1
+// Version 2 adds an alternate API for cross-host transfers:
+// CrossHostSendBuffers and CrossHostReceiveBuffers. These methods allow PjRt
+// clients to implement various optimizations for cross-host transfers.
+
+#define PJRT_API_CROSS_HOST_TRANSFERS_EXTENSION_VERSION 2
 
 // ---------------------------------- Methods ----------------------------------
 
+// Structs and methods prefixed with
+// PJRT_Transfers_PJRT_Client_CrossHost{Send,Receive}Buffers correspond to the
+// second cross-host transfers API.
+struct PJRT_Transfers_PJRT_Client_CrossHostSendBuffers_Args {
+  size_t struct_size;
+  PJRT_Extension_Base* extension_start;
+  PJRT_Client* client;
+  size_t num_buffers;
+  PJRT_Buffer** buffers;
+  const xla::PjRtGlobalDeviceId*
+      dst_global_device_ids;                       // Has size num_buffers.
+  const xla::CrossHostTransferKey* transfer_keys;  // Has size num_buffers.
+  PJRT_Event** send_events;  // Output; has size num_buffers.
+};
+
+PJRT_DEFINE_STRUCT_TRAITS(PJRT_Transfers_PJRT_Client_CrossHostSendBuffers_Args,
+                          send_events);
+
+typedef PJRT_Error* PJRT_Transfers_PJRT_Client_CrossHostSendBuffers(
+    PJRT_Transfers_PJRT_Client_CrossHostSendBuffers_Args* args);
+
+struct PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers_Args {
+  size_t struct_size;
+  PJRT_Extension_Base* extension_start;
+  PJRT_Client* client;
+  size_t num_shapes;
+  size_t* shape_num_dims;
+  const int64_t** num_dims;
+  PJRT_Buffer_Type* element_types;
+  PJRT_Buffer_MemoryLayout** layouts;
+  PJRT_Device* device;
+  const xla::PjRtGlobalDeviceId* src_global_device_ids;  // Has size num_shapes.
+  const xla::CrossHostTransferKey* transfer_keys;        // Has size num_shapes.
+  PJRT_Buffer** buffers;  // Output; has size num_shapes.
+};
+
+PJRT_DEFINE_STRUCT_TRAITS(
+    PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers_Args, buffers);
+
+typedef PJRT_Error* PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers(
+    PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers_Args* args);
+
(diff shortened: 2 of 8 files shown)


xla/pjrt/pjrt_client.h
```diff
@@ -1011,6 +1022,44 @@ class PjRtClient {
         "DmaUnmap not supported on platform %s", platform_name()));
   }
 
+  // CrossHostSendBuffers, CrossHostReceiveBuffers, and CancelCrossHostTransfer
+  // are part of the new cross-host transfers API.
+
+  // Send buffers to remote devices specified by dst_global_device_ids.
+  virtual std::vector<Future<>> CrossHostSendBuffers(
```
emilyfertig: I think we should return a StatusOr here since it would somewhat simplify the overriding methods (e.g. use of `TF_ASSIGN_OR_RETURN` instead of checking `status.ok`), as well as the code below. Also `CrossHostReceiveBuffers` returns a `StatusOr` of a vector of futurey things.

xla/pjrt/c_api_client/pjrt_c_api_client.cc
```diff
@@ -1047,8 +1047,135 @@ PjRtCApiClient::MakeCrossHostReceiveBuffers(
...
+  // Form inputs.
+  PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers_Args args;
+  args.struct_size =
+      PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers_Args_STRUCT_SIZE;
+  args.extension_start = nullptr;
+  args.client = c_client_.get();
+
+  args.num_shapes = shapes.size();
+
+  std::vector<size_t> shape_num_dims;
```
emilyfertig: Would it make sense to factor out repeated code here and `MakeCrossHostReceiveBuffers`?

xla/pjrt/common_pjrt_client.h
```diff
@@ -278,6 +281,9 @@ class CommonPjRtBufferImpl : public CommonPjRtBuffer {
   // device to host transfer to read the metadata of dynamic shape.
   absl::StatusOr<Shape> logical_on_device_shape() override;
 
+  // TODO(asrao): CopyToRemoteDevice is part of the legacy cross-host transfers
```
emilyfertig: Let's remove this TODO, I think the old APIs might have to stay indefinitely for TPU (though that's TBD).

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -353,6 +353,308 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
   return absl::OkStatus();
 }
 
+std::string CrossHostTransferName(PjRtGlobalDeviceId src_global_device_id,
+                                  PjRtGlobalDeviceId dst_global_device_id,
+                                  RunId transfer_run_id) {
+  std::stringstream ss;
```
emilyfertig: Use `absl::StrCat` instead?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -353,6 +353,308 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
...
+  ss << "cross_host_transfer-" << src_global_device_id.value() << "_to_"
+     << dst_global_device_id.value() << "-run_" << transfer_run_id.ToInt();
+  return ss.str();
+}
+
+absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
+    LocalDeviceState* local_device, gpu::GpuCollectives* gpu_collectives,
+    CliqueId clique_id, bool is_sender) {
+  VLOG(2) << "Creating a new communicator for cross host transfer, is_sender = "
```
emilyfertig: Should this be `VLOG(3)` instead? (Looked to me like that might be more aligned with the other logging statements).

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -353,6 +353,308 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
...
+  if (dst_global_device_ids.size() != buffers.size() ||
+      transfer_keys.size() != buffers.size()) {
+    return make_error_futures(InvalidArgument(
+        "CrossHostSendBuffers: buffers, "
+        "dst_global_device_ids, and transfer_keys "
+        "must have the same length, but got %d, %d, and %d.",
+        buffers.size(), dst_global_device_ids.size(), transfer_keys.size()));
+  }
+
+  std::vector<Future<>> out_futures;
```
emilyfertig, line 411: reserve `buffers.size()`?

xla/pjrt/pjrt_client.h
```diff
@@ -1011,6 +1022,44 @@ class PjRtClient {
...
+  CrossHostReceiveBuffers(
+      xla::PjRtDevice* device, absl::Span<const xla::Shape> shapes,
+      absl::Span<const PjRtGlobalDeviceId> src_global_device_ids,
+      absl::Span<const CrossHostTransferKey> transfer_keys) {
+    return absl::UnimplementedError(
+        "Cross-host data transfers are not supported.");
+  }
+
+  // Cancel cross host transfers associated with a given transfer_key.
+  virtual absl::Status CancelCrossHostTransfer(
```
emilyfertig: Is there a prototype for how cancellation will work? If the transfer key will become irrelevant in `CrossHost{Send,Receive}Buffers`, will it be ignored here?

I heard from another team who's very interested in this work and asked about cancellation semantics for failures, i.e. whether it will handle one transfer's failure in isolation without affecting other concurrent requests.

(On TPU, cancellation is handled by a callback passed to the notifier callback in `MakeCrossHostReceiveBuffers`, so as long as the new API on TPU is implemented in terms of the old one, it's unclear to me how/if we could make this work on TPU.)

emilyfertig (changes requested):
This is looking good! I left a few comments, but I still need to think about the `std::vector<CrossHostTransferKey>` question we talked about earlier, and I haven't looked at the tests much yet. I'll try to do that tomorrow.

xla/pjrt/pjrt_client.h
```diff
@@ -1011,6 +1020,44 @@ class PjRtClient {
...
+
+    return futures;
+  }
+
+  // Places buffers from a cross-host send onto device.
+  virtual absl::StatusOr<std::vector<std::unique_ptr<PjRtBuffer>>>
+  CrossHostReceiveBuffers(
+      xla::PjRtDevice* device, absl::Span<const xla::Shape> shapes,
+      absl::Span<const PjRtGlobalDeviceId> src_global_device_ids,
+      absl::Span<const CrossHostTransferKey> transfer_keys) {
```
emilyfertig: I took another look at how this works on TPU, and I think we should stick with passing a list of keys instead of a single key. I think we could make a single key work in the TPU client by using a counter per src/dst device pair, along with the src/dst device IDs and the single transfer key, to generate a unique KV store key. That strikes me as not worth the extra complexity, what do you think?

Also, could you use `std::vector` instead of `absl::Span` for the keys? On TPU, the notifier callback needs to take ownership of the keys (or copy them), or else they'll go out of scope when the callback is called. We can `std::move` the vector into `CrossHost{Send,Receive}Buffers` and into the callback, like we do currently in IFRT. (`{src,dst}_global_device_ids` would theoretically have the same issue, though they're currently unused on TPU).

xla/pjrt/pjrt_client.h
rao-ashish: Thanks for reviewing this! Changed to return `absl::StatusOr<std::vector<Future<>>>`.

xla/pjrt/c_api_client/pjrt_c_api_client.cc
rao-ashish: Added helper functions `MakeShapesInfo` and `MakePjRtBuffersFromPJRT_Buffers` to reduce repeated code.

xla/pjrt/common_pjrt_client.h
rao-ashish: Got it, I've removed this and some similar TODOs I left.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Changed to use `absl::StrCat`

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Makes sense; changed this to `VLOG(3)`.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Added a reserve here and similarly in `CrossHostReceiveBuffers`.

xla/pjrt/pjrt_client.h
rao-ashish: For TPU, I think it should be possible to implement the new API using a similar notifier / callback system as in the old API. The old API constructs `notifier` inside the IFRT client and passes this into `MakeCrossHostReceiveBuffers`, but I think you could just construct this notifier inside the PJRT client directly. The only information held by the IFRT client that `notifier` is currently using seems to be the KV store, which should also be available inside the PJRT client (at least, it is for `StreamExecutorGpuClient`). So, I'm imagining that the TPU implementation of `PjRtClient::CrossHostReceiveBuffers` could construct the notifier callback, call `MakeCrossHostReceiveBuffers` with that callback, and `MakeCrossHostReceiveBuffers` could still return cancel notifiers to `PjRtClient::CrossHostReceiveBuffers` as it is doing now. Basically, this is like moving `ifrt::PjRtClient::CrossHostSendBuffers` and `ifrt::PjRtClient::CrossHostReceiveBuffers` down into the `xla::PjRtClient`.

(From the above, it strikes me that with such an implementation, cancellation becomes something entirely contained inside the PJRT client. Because of this, maybe we don't need to expose cancellation at all on the IFRT layer, and we can remove this method?)

For GPU, we don't have a prototype implemented of cancellation yet. 

RE transfer keys, yes, in the optimized version, we will ignore the transfer keys (for now it is still used to construct the transfer name to form the keys in the KV store).

emilyfertig: Thanks -- I was thinking along the same lines for TPU. At first (and maybe indefinitely) the new TPU APIs will just call the old ones. The old ones offer the advantage of allowing users to handle communication of the buffer descriptors themselves, which some users want, so I'm not sure we'll delete them (though it probably makes sense to delete their GPU implementation).

I mostly wanted to make sure that, on the GPU side, we'll have a mechanism for gracefully handling the failure of a single transfer in isolation without e.g. restarting the process, and that any API we choose has that in mind. It sounds like handling cancellation in the PJRT client will for sure enable us to do that, right? (Sorry for the basic question, I'm new to some of the concepts here.)

I think handling cancellation in PJRT would be preferred, and we can remove this method.

emilyfertig:
Thank you! I'll find someone on the OpenXLA side to review who has the right permissions to pull this in and run the presubmit workflows.

Small request -- in the future could you please push individual commits, and then squash at the end when it's ready to merge? Makes it easier to see the diffs in review.

xla/pjrt/pjrt_client.h
rao-ashish: Got it thanks, I've removed the method. I haven't thought too much about cancellation on the GPU side, but yes it does seem to me that it can be internal to the PJRT client; I don't see anything that would be needed for it from the IFRT side. Maybe @mwhittaker could confirm if this API would work fine for cancellation on GPU?

xla/pjrt/pjrt_client.h
rao-ashish: I agree, I think just using a list of keys is probably the way to go. Also, as you suggested, I've changed the methods to take in a vector of transfer keys instead of a span.

rao-ashish:
Awesome, thanks so much for all your help on this! Will keep the note about individual commits in mind for next time.

xla/pjrt/pjrt_client.h
```diff
@@ -1011,6 +1020,37 @@ class PjRtClient {
         "DmaUnmap not supported on platform %s", platform_name()));
   }
 
+  // CrossHostSendBuffers, CrossHostReceiveBuffers, and CancelCrossHostTransfer
```
emilyfertig: Remove mention of CancelCrossHostTransfer?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -353,6 +353,298 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
...
+      src_global_device_id, dst_global_device_id, RunId(transfer_key.value()));
+
+  // Get the buffer's shape.
+  TF_ASSIGN_OR_RETURN(Shape shape, buffer->HostShape());
+
+  auto [promise, future] = Future<>::MakePromise();
+
+  // Acquire a hold on the buffer.
+  auto* buffer_impl = tensorflow::down_cast<CommonPjRtBufferImpl*>(buffer);
+  CommonPjRtBufferImpl::ScopedHold hold =
```
pschuh: hold must be checked for errors. also need to wait on definition events. the usage_event is useless if it isn't set on the hold via usage_event. Consider using: AcquireScopedRawBuffer().

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -353,6 +353,298 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
...
+
+          return absl::OkStatus();
+        };
+
+        absl::Status s = f();
+        if (!s.ok()) SetEventAsError(usage_event, s);
+        promise.Set(s);
+      };
+
+  local_device->execute_thread()->Schedule(std::move(send));
```
pschuh, line 535: Why run on the execute thread?

xla/pjrt/pjrt_client.h
```diff
@@ -964,16 +973,16 @@ class PjRtClient {
...
-  // cross host transfers using `client` on `device'. Asynchronously calls
-  // `notifier` once receive descriptors are ready to be communicated to the
-  // sender. `shapes` must be the exact shapes, with identical layouts,
-  // corresponding to the buffers that will be sent. When resources for the
-  // transfer are available, notifier will be called with a vector of
-  // PjRtCrossHostRecvDescriptors structs, one for each shape in `shapes`. Each
-  // struct contains an opaque string that should be transmitted to the sending
-  // host and used in a call to CopyToRemoteDevice. None of the recv buffers
-  // will become ready until *all* of the sends have completed.
+  // Part of legacy cross-host transfers API. Returns a vector of PjRtBuffers
```
emilyfertig: I think to avoid confusion on TPU, we should call these something besides "new" and "legacy", though I can't think of good descriptors. Maybe just "first" and "second"?

Also it might be worth adding a note like "See subclass documentation for platform-specific tradeoffs between the two cross-host transfer methods" just to orient people a little bit for why there are two.

emilyfertig (changes requested):
Had one more thought about the docstrings, then this looks good.

xla/pjrt/pjrt_client.h
```diff
@@ -1011,6 +1020,37 @@ class PjRtClient {
...
   }
 
+  // CrossHostSendBuffers, CrossHostReceiveBuffers, and CancelCrossHostTransfer
+  // are part of the new cross-host transfers API.
+
+  // Send buffers to remote devices specified by dst_global_device_ids.
+  virtual absl::StatusOr<std::vector<Future<>>> CrossHostSendBuffers(
+      absl::Span<PjRtBuffer* const> buffers,
+      absl::Span<const PjRtGlobalDeviceId> dst_global_device_ids,
+      std::vector<CrossHostTransferKey> transfer_keys) {
```
mwhittaker, line 1033: Just to make sure I'm understanding, `buffers`, `dst_global_device_ids`, and `transfer_keys` should all be the same length? And `CrossHostSendBuffers` will send `buffers[i]` to `dst_global_device_ids[i]` with unique key `transfer_keys[i]`. And the receiving device has to call `CrossHostReceiveBuffers` before the call to `CrossHostSendBuffers`?

xla/pjrt/extensions/cross_host_transfers/pjrt_c_api_cross_host_transfers_extension.h
```diff
@@ -81,7 +125,10 @@ typedef void PJRT_Buffer_CopyToRemoteDevice(
 
 typedef struct PJRT_CrossHostTransfers_Extension {
   PJRT_Extension_Base base;
-
+  PJRT_Transfers_PJRT_Client_CrossHostReceiveBuffers*
```
mwhittaker, line 135: I'm not an expert at the C API stuff, but do these added fields have to go at the end of the struct? @emilyfertig, maybe you know?

mwhittaker (commented):
Looks awesome! Thanks so much!

xla/pjrt/extensions/cross_host_transfers/pjrt_c_api_cross_host_transfers_extension.h
emilyfertig: Oh I think you're right, thanks for catching that.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Thanks for taking a look! I was trying to wait on the definition events via `WaitForBufferDefinitionEventsOnStream` inside `f`, and I thought that adding `usage_event.AndThen([mem]() {});` was sufficient to make sure the buffer stays allocated.

I've changed things to use a helper `AcquireHeldSendBuffer` wrapping `AcquireScopedRawBuffer`, does this work to use holds correctly now? I based this off of `CommonPjRtBufferImpl::CopyToRemoteDevice`. It seemed like the logic around `usage_event_promise` could be removed in the `AcquireHeldSendBuffer` case if we just pass in the usage event. 

In this updated version, am I waiting on definition events correctly? I wasn't able to find an analogous call to the `WaitForBufferDefinitionEventsOnStream` I had earlier inside the `CopyToRemoteDevice`/`ScheduleRemoteSend` code path, so I removed it under the assumption that the waiting happens elsewhere (eg in `event->GetErrorIfPresent()`?)

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: I was observing that the original implementation could deadlock when `device_put` was called in the presence of other computations. For example, [here](https://gist.github.com/rao-ashish/e37d7460c7ef5225587a8291b63f6bfe) is a minimal example that was deadlocking. Scheduling transfers on the execute thread solved this issue.

As per the NCCL documentation: “... to remain deadlock free, users must ensure the order of host-side launches matches for all devices.” ([link](https://docs.nvidia.com/deeplearning/nccl/archives/nccl_2277/user-guide/docs/usage/communicators.html#using-multiple-nccl-communicators-concurrently)). If we schedule the sends/recvs on a thread pool instead of the execute thread, the order of host-side launches may no longer match for senders and receivers. Some more details on this are given in [this document](https://docs.google.com/document/d/1KLczScQVV0qD7ve4Gj4QBOU-3qFa57qJiO52wQX4qtU/edit?usp=sharing) that was testing out the original implementation.

xla/pjrt/pjrt_client.h
rao-ashish: Oops, removed this.

xla/pjrt/pjrt_client.h
rao-ashish: I've changed the comments to move away from "legacy" / "new"; I sometimes use "original' when referring to the first API.

xla/pjrt/pjrt_client.h
rao-ashish: Yes, `CrossHostSendBuffers` will send `send buffers[i]` to `dst_global_device_ids[i]`, and that transfer is associated with a unique key `transfer_keys[i]`.

Regarding the ordering of `CrossHostSendBuffers` / `CrossHostReceiveBuffers`, it should be fine if the sending device calls `CrossHostSendBuffers` before the receiving device `CrossHostReceiveBuffers`; I believe that the one which executes first will just block the corresponding device's execute thread until the second runs.

xla/pjrt/extensions/cross_host_transfers/pjrt_c_api_cross_host_transfers_extension.h
rao-ashish: Got it, I've moved these fields to the end of struct.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
pschuh: wow, I think this is pretty limiting. This means that you need to precisely predict when buffers will be available in the global ordering and schedule the device_puts accordingly, and even when you do this, it will block the compute queue which will prevent compute communicate overlap. What if we had a separate thread just for scheduling these (that way they would be ordered still)?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -353,6 +353,324 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
...
+
+  auto send = [this, gpu_collectives, promise = std::move(promise),
+               usage_event = std::move(usage_event),
+               held_send_buffer = std::move(held_send_buffer), local_device,
+               cross_host_transfer_name, shape]() mutable {
+    se::Stream* stream = local_device->GetDeviceToDeviceStream();
+
+    auto f = [&]() -> absl::Status {
+      // Wait until the buffer we want to send is fully materialized.
+      for (const auto& event : held_send_buffer.definition_events)
```
pschuh: I think you need to block on the events:

for (const auto& event : held_send_buffer.definition_events) {
  tsl::BlockUntilReady(event);
  if (auto* status = event->GetErrorIfPresent()) return *status;
}

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: > What if we had a separate thread just for scheduling these (that way they would be ordered still)?

The issue is that the "normal" SPMD executables launched by XLA on the execute threads can also use NCCL collectives, and we need a consistent order between the `device_put` NCCL transfers and these. So even if we have a dedicated thread to just schedule cross-host transfers, its execution could be interleaved with the XLA launcher thread, and we could end up without a consistent order.

What if we still launch these on the execute thread, but instead of calling `send_future.Await()` here, we launch something on `this->thread_pool()` that will wait on `send_future` and fulfill `usage_event` and `promise` when`send_future` is done? That way we won't block the execute thread?

pschuh: It sounds like there is no way around it. Either you need a central coordinator which picks a defined order for all ops, you stall the execute thread here, or fix nccl to allow certain collectives to run unordered.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Thanks, added this, realized that `ScheduleWhenReady` before was taking care of this.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: > It sounds like there is no way around it.

Just to clarify, do you mean that there is no way around scheduling this on the execute thread? Also, do you think that having a separate thread wait on `send_future` and fulfill `usage_event` and `promise` could help?

Also, I was hoping that this PR could add the new cross-host data transfer API, and that follow-up PRs could implement client-specific optimizations. As such, do you think that we could address the issue of reducing the amount of time the execute thread is blocked in a follow-up PR, since this is internal to `StreamExecutorGpuClient`'s implementation of the new API?

xla/pjrt/extensions/cross_host_transfers/pjrt_c_api_cross_host_transfers_extension.h
emilyfertig, line 1: @rao-ashish I forgot, could you please increment the version number on line 33 and add a brief update to the docstring above it?

rao-ashish: Updated the version number.

emilyfertig (approved):
Thank you!

emilyfertig:
@rao-ashish Could you take a look at the presubmit failures? Most of them look unrelated but this one e.g. looks real: https://github.com/openxla/xla/actions/runs/19079235052/job/54518987561?pr=33284

rao-ashish:
Ah sorry about that, the relevant failures seem to have been because I was not checking the output of `kv_store_->Set(...)` in `se_gpu_pjrt_client.cc` line 641. I've just updated the commit to surround this with `TF_RETURN_IF_ERROR`, which should solve the issue.

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -2908,6 +2908,283 @@ TEST(StreamExecutorGpuClientTest, LinkedEventPromise) {
...
+    EXPECT_EQ(send_futures.size(), num_arrays);
+    for (int i = 0; i < num_arrays; ++i) {
+      LOG(INFO) << log_prefix << ": waiting for send " << i << " to complete";
+      TF_RETURN_IF_ERROR(send_futures[i].Await());
+      LOG(INFO) << log_prefix << ": send " << i << " completed";
+    }
+  }
+
+  // Receiver logic.
+  else {
```
emilyfertig, line 3142: Could you move this `else` up so it's on the same line as the preceding `}`?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -3131,5 +3439,25 @@ int main(int argc, char* argv[]) {
     }
     return result.raw_code();
   }
-  return RUN_ALL_TESTS();
-}
+
+  else if (test_to_run == "SuccessfulCrossHostTransferHelper") {
```
emilyfertig, line 3443: Move this else up too?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -3131,5 +3439,25 @@ int main(int argc, char* argv[]) {
...
+                                                   num_arrays);
+    } else {
+      LOG(ERROR) << "cross_host_test_role must be 'sender' or 'receiver'.";
+      return 1;
+    }
+    if (!s.ok()) LOG(ERROR) << s;
+    return s.raw_code();
+  }
+
+  else {
```
emilyfertig, line 3459: And this one?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -3131,5 +3439,25 @@ int main(int argc, char* argv[]) {
...
+    LOG(ERROR) << "Unrecognized multiprocess test name " << test_to_run << ".";
+    return 1;
+  }
+}
```
emilyfertig, line 3463: Add a newline at the end of the file?

emilyfertig (commented):
Just a few more things our internal linter complained about. Sorry for the back and forth but I think this should be the last of it.

@pschuh to double check is there anything else you think should be addressed in this PR as opposed to a follow-up?

emilyfertig:
@rao-ashish Actually maybe hold off on that, it looks like someone on the OpenXLA team is fixing it internally.

xla/pjrt/pjrt_client.h
```diff
@@ -1012,6 +1023,37 @@ class PjRtClient {
...
+        "Cross-host data transfers are not supported by this client.");
+
+    std::vector<Future<>> futures;
+    futures.reserve(buffers.size());
+
+    for (size_t i = 0; i < buffers.size(); ++i) {
+      futures.emplace_back(error);
+    }
+
+    return futures;
```
allanrenucci, line 1044: Why is this not simply:
```suggestion
    return absl::InternalError(
        "Cross-host data transfers are not supported by this client.");
```

If it is intentional, I'd say it deserves a comment explaining why in the code.

xla/pjrt/c_api_client/pjrt_c_api_client.cc
```diff
@@ -1044,13 +1077,107 @@ PjRtCApiClient::MakeCrossHostReceiveBuffers(
...
+          PJRT_Extension_Type::PJRT_Extension_Type_CrossHostTransfers);
+  if (extension == nullptr) {
+    absl::Status error = absl::UnimplementedError(
+        "CrossHostSendBuffers is not implemented in this PJRT plugin.");
+    std::vector<Future<>> futures;
+    futures.reserve(buffers.size());
+    for (int i = 0; i < buffers.size(); ++i) {
+      futures.push_back(Future<>(error));
+    }
+    return futures;
```
allanrenucci, line 1102: Same comment as above, why not simply:

```suggestion
    return absl::UnimplementedError(
        "CrossHostSendBuffers is not implemented in this PJRT plugin.");
```

allanrenucci (changes requested):

xla/pjrt/pjrt_client.h
emilyfertig: You're right, this method used to return `std::vector<Future<>>` and I missed that this wasn't updated after we changed it to return the `StatusOr`. We've cloned this PR internally to make linting fixes before submitting, so I'll include these changes too.

PR #33363, opened 2025-10-30
[XLA:GPUI] dump command buffer graph to the directory specified by enviroment variable

shawnwang18:
📝 Summary of Changes
This PR enables dumping the command buffer dot file into the folder specified through environment variable COMMAND_BUFFER_DUMP_DIR, instead to the /tmp directory, which is inconvenient to fetch the files for runs like slurm. 

🎯 Justification
Easy debug

🚀 Kind of Contribution
📚 Documentation

### DIFF
--- a/xla/stream_executor/gpu/gpu_command_buffer.cc
+++ b/xla/stream_executor/gpu/gpu_command_buffer.cc
@@ -46,6 +46,7 @@ limitations under the License.
 #include "xla/tsl/platform/statusor.h"
 #include "tsl/platform/casts.h"
 #include "tsl/platform/path.h"
+#include "xla/tsl/util/env_var.h"
 
 namespace stream_executor::gpu {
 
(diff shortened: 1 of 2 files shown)


xla/stream_executor/gpu/gpu_command_buffer.cc
```diff
@@ -571,7 +572,18 @@ absl::Status GpuCommandBuffer::Finalize() {
   // Maybe dump created GPU graph to a dot file for debugging.
   if (state_ == State::kCreate &&
       (VLOG_IS_ON(10) || (VLOG_IS_ON(9) && mode_ == Mode::kPrimary))) {
-    std::string path = tsl::io::GetTempFilename(/*extension=*/"dot");
+    std::string path;
+    std::string dump_dir;
+    if (!tsl::ReadStringFromEnvVar("COMMAND_BUFFER_DUMP_DIR", "", &dump_dir)
```
apivovarov, line 577: xla already provides a convention and flag for dumping debug data 
```
--xla_dump_to=""  string  A directory to dump xla debug data to.
```

@ezhulenev Do you think we need to introduce separate environment variable COMMAND_BUFFER_DUMP_DIR

apivovarov: before_validation.ttir.txt dump example
[xla/backends/gpu/codegen/triton/fusion_emitter.cc](https://github.com/openxla/xla/blob/main/xla/backends/gpu/codegen/triton/fusion_emitter.cc#L1980)

```
// dump.h
// Consolidated utilities for logging information during compilation, usually
// based on the options specified in the DebugOptions proto.
//
// Most functions here take an HloModule and read the DebugOptions from the
// module's config.
```
[xla/service/dump.h](https://github.com/openxla/xla/blob/main/xla/service/dump.h)

shawnwang18:
--xla-dump-to are  flags only used during compile time, gpu command buffer are runtime thinge, so I think we can not use the flag


获取Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #33448, opened 2025-11-03
[DOC] added memory space identifiers

athurdekoos:
📝 Summary of Changes

- New memory space identifiers subsection
- Short sentence identifying memory space in the intro.

🚀 Kind of Contribution
Please remove what does not apply: 📚 Documentation

### DIFF
--- a/docs/operation_semantics.md
+++ b/docs/operation_semantics.md
@@ -11,7 +11,7 @@ N-dimensional array holding elements of some uniform type (such as 32-bit
 float). Throughout the documentation, *array* is used to denote an
 arbitrary-dimensional array. For convenience, special cases have more specific
 and familiar names; for example a *vector* is a 1-dimensional array and a
-*matrix* is a 2-dimensional array.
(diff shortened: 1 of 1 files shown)


docs/operation_semantics.md
```diff
@@ -3381,6 +3381,19 @@ in detail on the [broadcasting page](broadcasting.md).
 For StableHLO information see
 [StableHLO - maximum](https://openxla.org/stablehlo/spec#maximum).
 
+## Memory Space Identifiers
```
pavithraes, line 3384: @athurdekoos - Thanks for this PR! However, I don't think this is the right place to document this information. Even with the note in the introduction, this makes this section looks like it's an Op, and makes the info difficult to find.

I think we can move this whole section to the intro, with more context around Shapes, Layouts, and Tiling. But would be even better if we document it with an example close to those pages (Shapes, Layouts, and Tiling) and link to those sections in this doc.

These are quite a few changes, so I'm happy to take over and open a separate PR. Please feel free to close this one. :)

PR #33464, opened 2025-11-03
Add a new version of ShapeUtil::ByteStrides() and fix its use in the GPU compiler.

sergachev:
📝 Summary of Changes
Disable incorrect GPU copy fusions on packed sub-byte types using incorrect stride calculations.
Multiple uses in PJRT rely on the existing behavior of ByteStrides() which ignores element_size_in_bits - this is now called UnpackedByteStrides().

🎯 Justification
Fixes GPU DynamicMemcpyFusion which relies on ByteStrides().

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
Yes.

🧪 Execution Tests:
No.

### DIFF
--- a/xla/shape_util.cc
+++ b/xla/shape_util.cc
@@ -2201,8 +2201,8 @@ Shape ShapeUtil::DeviceShapeToHostShape(Shape s) {
 }
 
 /*static*/
-absl::Status ShapeUtil::ByteStrides(const Shape& shape,
-                                    absl::Span<int64_t> strides) {
+absl::Status ShapeUtil::UnpackedByteStrides(const Shape& shape,
+                                            absl::Span<int64_t> strides) {
   TF_RET_CHECK(shape.IsArray());
   TF_RET_CHECK(shape.has_layout());
   TF_RET_CHECK(shape.dimensions().size() == strides.size());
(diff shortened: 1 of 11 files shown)


xla/shape_util.cc
```diff
@@ -2201,8 +2201,8 @@ Shape ShapeUtil::DeviceShapeToHostShape(Shape s) {
 }
 
 /*static*/
-absl::Status ShapeUtil::ByteStrides(const Shape& shape,
```
akuegel, line 2204: Can we keep this method around but mark it with ABSL_DEPRECATE_AND_INLINE() and make it call UnpackedByteStrides? There are a few internal callers of this method. I think if we mark it like that, the remaining users will be changed automatically by internal tools.

xla/shape_util.cc
```diff
@@ -2216,15 +2216,23 @@ absl::Status ShapeUtil::ByteStrides(const Shape& shape,
...
-  if (!ByteStrides(shape, absl::MakeSpan(strides)).ok()) {
+  if (!UnpackedByteStrides(shape, absl::MakeSpan(strides)).ok()) {
     return std::nullopt;
   }
   return strides;
 }
 
+/*static*/ std::optional<absl::InlinedVector<int64_t, 4>>
+ShapeUtil::ByteStrides(const Shape& shape) {
+  if (shape.layout().element_size_in_bits() % CHAR_BIT != 0) {
```
akuegel, line 2236: Should this possibly be a TF_RET_CHECK in the UnpackedByteStrides method above?

sergachev: UnpackedByteStrides keeps the current behavior of ByteStrides which ignores element_size_in_bits. PJRT uses rely on this, they fail with this check.

akuegel (approved):

PR #33531, opened 2025-11-04
[ROCm] Introduce rocm_ci github action

alekstheod:
📝 Summary of Changes
Introduce github action for rocm ci job

🎯 Justification
We will replace rocm jenkins job with a github action which shall make better
our response time on prs

🚀 Kind of Contribution
Please remove what does not apply: ⚡️ Performance Improvement

CI performance improvement, not relevant to xla performance

🧪 Unit Tests:
not relevant

🧪 Execution Tests:
Not relevant


.github/workflows/rocm_xla_ci.yml
```diff
@@ -0,0 +1,55 @@
...
+            --security-opt=seccomp=unconfined \
+            -v ${{ github.workspace }}:/workspace \
+            -w /workspace \
+            ${{ env.DOCKER_IMAGE }}
+
+      - name: Test XLA
+        run: |
+          docker exec ${{ env.CONTAINER_NAME }} bash -c "
+            build_tools/rocm/run_xla_ci_build.sh \
+              --config=rocm_ci \
```
alekstheod: Lets put the 8 targets we do run currently on CI

.github/workflows/rocm_xla_ci.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      # Pre-clean up in case a previous run crashed
+      - name: Cleanup container
+        run: docker rm -f ${{ env.CONTAINER_NAME }} || true
+
+      - uses: actions/checkout@v4
+
+      - name: Start Container
+        run: |
+          docker run -dt \
```
alekstheod, line 50: rbe certs are required here. We do get them from rocAutomation

.github/workflows/rocm_xla_ci.yml
```diff
@@ -0,0 +1,60 @@
+name: '[ROCm] OpenXla CI'
```
MichaelHudgins: Bit of a nit but i would prefer something like `CI ROCm` to be more in line with current workflows.  (For instance i think its assumed this is for OpenXLA)

.github/workflows/rocm_xla_ci.yml
```diff
@@ -0,0 +1,60 @@
...
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
```
MichaelHudgins: Do you believe you have enough runners to move to `cancel-in-progress: ${{ github.ref != 'main' }}`. Without it you can run into the case where the jobs on main keep getting canceled when new commits land, but you would need enough hardware to allow for that.

.github/workflows/rocm_xla_ci.yml
```diff
@@ -0,0 +1,60 @@
+name: '[ROCm] OpenXla CI'
+
```
MichaelHudgins: Please add permissions explicitly.  Such as 
https://github.com/openxla/xla/blob/main/.github/workflows/ci.yml#L16C1-L17C17

.github/workflows/rocm_xla_ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    runs-on: linux-x64-mi210-64-2gpu-amd
+    env:
+      DOCKER_IMAGE: rocm/tensorflow-build@sha256:7fcfbd36b7ac8f6b0805b37c4248e929e31cf5ee3af766c8409dd70d5ab65faa
+      CONTAINER_NAME: xla-runner
+    steps:
+      # Pre-clean up in case a previous run crashed
+      - name: Cleanup container
+        run: docker rm -f ${{ env.CONTAINER_NAME }} || true
+
+      - uses: actions/checkout@v4
```
MichaelHudgins: We in general require actions to be sha locked.  I use https://github.com/sethvargo/ratchet for that typically.

MichaelHudgins (changes requested):

.github/workflows/rocm_xla_ci.yml
MichaelHudgins: (If not i think it would be better to have a cron trigger than an on: push: main trigger)

.github/workflows/rocm_xla_ci.yml
MichaelHudgins: Edited above suggestion, doing something like CI ROCm would at least make those results appear next to other current CI results in the PR list

.github/workflows/rocm_xla_ci.yml
alekstheod: Done

MichaelHudgins (approved):

alekstheod:
@MichaelHudgins do we need some other approval to merge this in?

MichaelHudgins (approved):

MichaelHudgins:
> @MichaelHudgins do we need some other approval to merge this in?

Thanks for the poke, automation didn't pull it in like it should have.  I've sent it for the internal review

.github/workflows/rocm_xla_ci.yml
```diff
@@ -0,0 +1,61 @@
+name: CI ROCm
```
thomasjoerg, line 16: Please add an Apache License header to this file.

thomasjoerg (changes requested):

thomasjoerg (approved):

PR #32107, opened 2025-10-01
Expand tiling docs

mtsokol:
📝 Summary of Changes

This PR expands tiling docs with Motivation section and Tiling formats.

🎯 Justification

Expand Tiling docs

🚀 Kind of Contribution

📚 Documentation

PR #33820, opened 2025-11-11
Update `shapes.md` with example and memory location info

pavithraes:
**📝 Summary of Changes**

- Added an example for how to read an Op, derived from [JAX scaling book](https://jax-ml.github.io/scaling-book/profiling/#how-to-read-an-xla-op).
- Added description for Memory Space Identifier (also derived from the book), because it was not documented in the XLA docs.
- Added relevant cross-links between pages.

**🎯 Justification**

While there are notes about Shape, Layout, and Tiling, there was some gap in contextualizing this information with an example. Additionally, there was no information about Memory Space Identifiers.

These were originally noted/flagged by @GleasonK and @harini-sridhar.

**🚀 Kind of Contribution**

📚 Documentation

### DIFF
--- a/docs/operation_semantics.md
+++ b/docs/operation_semantics.md
@@ -13,6 +13,8 @@ arbitrary-dimensional array. For convenience, special cases have more specific
 and familiar names; for example a *vector* is a 1-dimensional array and a
 *matrix* is a 2-dimensional array.
 
+Learn more about the structure of an Op in [_Shapes and layout_](shapes.md) and [_Tiled Layout_](tiled_layout.md).
+
 ## Abs
 
 See also
--- a/docs/shapes.md
+++ b/docs/shapes.md
@@ -5,27 +5,41 @@
 Consider an example HLO:
 
 ```
+add.936 = bf16[8,1,1280,16384]{3,2,0,1:T(8,128)(2,1)}
+          add(exponential.183, broadcast.3115)
+```
+
+This consists of the following components:
+
+* Op Name: `add.936`
+  * This is the unique name for the operation.
+* Shape: `bf16[8,1,1280,16384]`
+  * This is the output shape of the Op. Here the dtype is [bf316](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) and the shape is `[8,1,1280,16384]`.
+* Layout (with Tiling): `3,2,0,1:T(8,128)(2,1)`
+  * This describes how the array is stored in memory. `3,2,0,1` denotes the order of the axes in memory (i.e., column major, row major, etc.) and `T(8,128)(2,1)` denotes the tiling & padding used.
+  * Layout is optional. If not specified, there is no tiling and the dimensions are assumed to be ordered from most-major to most-minor.
+* Operation: `add`
+  * The operation being performed. Here, it is [Add](operation_semantics.md#add), which is also mentioned in the Op name.
+* Arguments: `exponential.183`, `broadcast.3115`
+  * This operation takes two arguments, specified with their unique names.
+
+Let's consider another example, a fusion Op:
+
+```
 %fusion.3 = bf16[32,32,4096]{2,1,0:T(8,128)(2,1)S(1)}
             fusion(bf16[32,32,8192]{2,1,0:T(8,128)(2,1)S(1)} %fusion.32),
             kind=kCustom, calls=%all-reduce-scatter.3
 ```
 
-This consists on the following components:
+In addition to the previously described components, this consists of:
 
-* Op Name: `%fusion.3`
-  * This is the unique name for the operation. Here, the name indicates this is a fusion operation (which combines multiple operations into a single kernel). The `%` sign is part of the name, and may or may not be present in all operations. Learn about other Ops in [Operation Semantics](operation_semantics.md).
-* Shape: `bf16[32,32,4096]`
-  * This is the output shape of the op. Here the dtype is bf16 (2 bytes per parameter) and the shape is `[32,32,4096]`. The following sections have more details about Shape.
-* Layout (with Tiling): `{2,1,0:T(8,128)(2,1)}`
-  * This describes how the array is stored in memory. `2,1,0` denotes the order of the axes in memory (column major, row major, etc.) and `T(8,128)(2,1)` denotes the tiling & padding used. The following sections have more details about Layout, you can also learn more in [Tiled Layout](tiled_layout.md).
-  * This is optional. If not specified, there is no tiling and the dimensions are assumed to be ordered from most-major to most-minor.
+* Attributes: `kind` and `calls`
+  * These provide more information about the operation being performed, in this case: fusion.
 * Memory location (memory space identifier): `S(1)`
-  * This denotes the memory space/location where the array is stored. `S(1)` denotes this array lives in VMEM (on a TPU). More [memory space identifiers are listed below](#memory-space-identifiers).
-* Arguments: `bf16[32,32,8192]{2,1,0:T(8,128)(2,1)S(1)} %fusion.32`
-  * This op has one input, a bf16 array called `fusion.32` with a particular shape (as well as layout, tiling, and memory location): ` bf16[32,32,8192]{2,1,0:T(8,128)(2,1)S(1)}`. This tells us what function feeds into this one.
-* Attributes:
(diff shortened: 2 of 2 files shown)


docs/shapes.md
```diff
@@ -1,11 +1,36 @@
...
+
+Consider an example Op:
+
+```
+%fusion.3 = bf16[32,32,4096]{2,1,0:T(8,128)(2,1)S(1)} fusion(bf16[32,32,8192]{2,1,0:T(8,128)(2,1)S(1)} %fusion.32), kind=kCustom, calls=%all-reduce-scatter.3
+```
+
+This consists on the following main components:
+
+* Op Name: `fusion.3`
```
dimitar-asenov: The `%` sign is actually part of the name and may or may not be there. It does not have any special meaning.

docs/shapes.md
```diff
@@ -1,11 +1,36 @@
...
+  * A dot or fusion op is a set of operations containing at most 1 matrix multiplication and possibly a bunch of related pointwise VPU-ops.
```
dimitar-asenov: Why focus on dot of fusion ops, this is a general doc for all ops, right? Also this sentence isn't quite right. A dot is a standalone op and not a set of operations, a fusion is also a standalone op that happens to point to  a computation. That computation is a set of ops.

Could you please clarify what you wanted to explain here?

docs/shapes.md
```diff
@@ -1,11 +1,36 @@
...
+* Shape: `bf16[32,32,4096]`
+  * This is the output shape of the op. Here the dtype is bf16 (2 bytes per parameter) and the shape is `[32,32,4096]`. More details in the following sections.
+* Layout (with Tiling): `{2,1,0:T(8,128)(2,1)}`
```
dimitar-asenov: It is worth mentioning that the layout is optional. If it's not specified, there is no tiling and the dimensions are assumed to be ordered from most-major to most-minor.

docs/shapes.md
```diff
@@ -1,11 +1,36 @@
...
+  * This describes how the array is stored in memory. More details in the following sections.
+  * `2,1,0` denotes the order of the axes in memory (column major, row major, etc.).
+  * `T(8,128)(2,1)` denotes the tiling & padding used. Learn more in [Tiled Layout](tiled_layout.md).
+* Memory location: `S(1)`
+  * `S(1)` denotes this array lives in VMEM. More details at the end of page.
```
dimitar-asenov: VMEM is a TPU-specific concept.  Mention here that this is the memory space.

dimitar-asenov: Also please link to the "end of the page" section where this is explained.

docs/shapes.md
```diff
@@ -1,11 +1,36 @@
...
+* Arguments:`bf16[32,32,8192]{2,1,0:T(8,128)(2,1)S(1)} %fusion.32`
```
dimitar-asenov: After the arguments come the attributes. In this case `kind`, `calls`. Please document these as well.

dimitar-asenov (changes requested):

docs/shapes.md
pavithraes: Yes, added.

docs/shapes.md
pavithraes: Updated the description, please let me know if it's better now.

docs/shapes.md
pavithraes: Updated :)

docs/shapes.md
pavithraes: Thanks! Included this in the description.

docs/shapes.md
pavithraes: Thanks for pointing out, this is an oversight from initially coping the example from the JAX scaling book. I've still added a brief explanation of fusion (mostly to not leave it as jargon, and because I didn't see it described anywhere else in the docs), but de-emphasized. Instead, this description now focuses more on naming.

docs/operation_semantics.md
```diff
@@ -13,6 +13,8 @@ arbitrary-dimensional array. For convenience, special cases have more specific
 and familiar names; for example a *vector* is a 1-dimensional array and a
 *matrix* is a 2-dimensional array.
 
+Learn more about the structure of an Op in [_Shapes and layout_](shapes.md) and [_Tiled Layout_](tiled_layout.md).
```
pavithraes, line 16: Kevin also mentioned potentially merging these pages, but that's partially blocked on #32107.

docs/shapes.md
```diff
@@ -137,3 +166,12 @@ index for each dimension. Linear indices are a single `int64` value which
 indexes into the buffer holding the array. See `shape_util.h` and
 `layout_util.h` in the same directory for utilities that simplify creation and
 manipulation of shapes and layouts.
+
+## Memory Space Identifiers
```
pavithraes, line 184: Is this the right terminology to use here?

docs/shapes.md
```diff
@@ -137,3 +166,12 @@ index for each dimension. Linear indices are a single `int64` value which
...
+
+In HLO, each array may be annotated with a memory space identifier, written as S(n).
```
pavithraes, line 186: Also, looks like all of these are TPU-specific, but I'd like to confirm. If so, I can add a clear note about this.

docs/shapes.md
```diff
@@ -1,11 +1,40 @@
...
+
+Consider an example HLO:
+
+```
+%fusion.3 = bf16[32,32,4096]{2,1,0:T(8,128)(2,1)S(1)}
+            fusion(bf16[32,32,8192]{2,1,0:T(8,128)(2,1)S(1)} %fusion.32),
+            kind=kCustom, calls=%all-reduce-scatter.3
+```
+
+This consists on the following components:
```
dimitar-asenov: "of" instead of "on"

docs/shapes.md
dimitar-asenov: I've never seen any GPU-HLO that contains memory spaces, but we do use them internally during compilation. I'm not sure whether these may be used in the source HLO, but I don't see why not.

@GleasonK Do you know?

docs/shapes.md
```diff
@@ -1,11 +1,40 @@
 # Shapes and layout
 
+## Structure of an XLA Op
+
+Consider an example HLO:
```
dimitar-asenov, line 5: I think this section needs to be tweaked. Right now it feels very specific and the explanation overfits the specific example. Here are some ideas how to make the section more general:

- Remove unnecessary details, e.g. pick a nice op name without `%` and `.3`. These are irrelevant and distract from the structure. The only important thing about the names are that each name must be unique.
- When discussing the attributes, it doesn't become clear if there might be other attributes (there can be). Instead of just describing these two specific attributes, it's better to explain what attributes are in general. (e.g. attributes further specify the behavior of the op, some are required some are optional)
- Not just for attributes, but also for arguments and results it's good to say more generally what could they look like.
- Explain the difference between attributes and op inputs (behavior vs data, compile-time vs run-time, etc.)
- Consider providing not one but 2 or 3 examples so that whoever is reading this gets a more complete picture. At least one of the examples should be something simpler (e.g. an `add` op).

dimitar-asenov (changes requested):

dimitar-asenov:
@pifon2a Could you please suggest what else we need to document here. I added some comments, but I feel there are more things that I'm missing.

docs/shapes.md
pavithraes: Thanks for this note and your patience. I agree that the example+explanation was too specific. I've made some updates.

docs/shapes.md
```diff
@@ -1,11 +1,54 @@
...
+            kind=kCustom, calls=%all-reduce-scatter.3
+```
+
+In addition to the previously described components, this consists of:
+
+* Attributes: `kind` and `calls`
+  * These provide more information about the operation being performed, in this case: fusion.
+* Memory location (memory space identifier): `S(1)`
+  * This denotes the memory space/location where the array is stored. `S(1)` here denotes this array lives in VMEM (on a TPU).
+* Shape and layout details for the input argument `%fusion.32`
```
dimitar-asenov, line 40: Note that these are optional, and if provided must exactly match the actual shape and layout of each argument.

dimitar-asenov (approved):

dimitar-asenov:
@pifon2a Do you think we can merge this PR in the current state? Do we need to explain more things?

PR #33909, opened 2025-11-13
[ROCm] Make multi gpu tests exclusive if executed locally

alekstheod:
📝 Summary of Changes
Make multigpu tests being executed exclusively.

🎯 Justification
Multigpu tests have to run exclusively to not clash the utilization of gpus in parallel.
This PR marks multigpu tests as such by adding 'multi_gpu' tag, and in case of rocm
make them run exclusively if executed locally.

🚀 Kind of Contribution
Please remove what does not apply:  ♻️ Cleanup

Not relevant, no logic change

🧪 Unit Tests:
Not relevant, no logic change

🧪 Execution Tests:
Not relevant, no logic change

### DIFF
--- a/xla/service/BUILD
+++ b/xla/service/BUILD
@@ -283,6 +283,7 @@ cc_library(
 xla_cc_test(
     name = "collective_permute_cycle_test",
     srcs = ["collective_permute_cycle_test.cc"],
+    tags = ["multi_gpu"],
     deps = [
         ":collective_permute_cycle",
         ":source_target_pairs",
@@ -472,6 +473,7 @@ cc_library(
 xla_cc_test(
     name = "collective_pipeliner_test",
     srcs = ["collective_pipeliner_test.cc"],
+    tags = ["multi_gpu"],
     deps = [
         ":collective_pipeliner",
         ":collective_pipeliner_utils",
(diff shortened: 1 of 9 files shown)


ddunl (approved):

xla/service/BUILD
```diff
@@ -283,6 +283,7 @@ cc_library(
 xla_cc_test(
     name = "collective_permute_cycle_test",
     srcs = ["collective_permute_cycle_test.cc"],
+    tags = ["multi_gpu"],
```
ddunl: I think these should be wrapped in `if_rocm` IIUC

alekstheod:
@beckerhe had a comment on that. It seems that xla_cc_tests are not really gpu tests but cpu only tests. I will have to remove the tag and adjust accordingly!

xla/service/BUILD
alekstheod: Seems that cc tests are not really gpu tests, they are wrongly assigned in our multigpu tests bazelrc configs. I will remove this tag from these kind of tests.

alekstheod: are we sure we want to add this tag only for rocm? I think if the test requires more than one gpu it has to be marked as multi_gpu test. I will check which of them are truly multi_gpu tests.

alekstheod: I did adaptions. Seems that all these tests are not really multi-gpu tests. I removed the tag and left only the exclusive tag to make them execute exclusively when running locally.

ddunl (approved):

alekstheod:
@xla-rotation please check this PR

PR #33910, opened 2025-11-13
Implement cross host transfer optimizations in StreamExecutorGpuClient

rao-ashish:
📝 Summary of Changes
This PR adds three optimizations to the `CrossHost{Send,Receive}Buffers` methods in `StreamExecutorGpuClient`: 
1. Communicators are cached for re-use across transfers with `AcquireGpuClique`.
2. Transfers of multiple arrays are aggregated with NCCL group calls using `GpuCommunicator::GroupExecute`.
3. Usage / definition events and promises that signal the completion of cross-host transfers are fulfilled on `StreamExecutorGpuClient::thread_pool()` instead of blocking the XLA execute thread.

This PR follows up on #33284 and is based on discussions with @emilyfertig, @mwhittaker, and @pschuh.

🎯 Justification
#33284 added `CrossHost{Send,Receive}Buffers` to the PjRt Client API to enable the optimizations implemented in this PR. The old implementation does not cache communicators, aggregate transfers, or asynchronously fulfill events / promises, leading to reduced performance.

🚀 Kind of Contribution
⚡️ Performance Improvement

Benchmarks were ran on these [three toy programs](https://gist.github.com/rao-ashish/28155ffee98406b7c041037c20fc75e6).

Workload | Implementation | Time
-- | -- | --
example_basic_overlap | Unoptimized | 26.968 sec
  | Optimized | 43.124 ms
example_overlap | Unoptimized | 369.541 sec
  | Optimized | 733.071 ms
example_pp | Unoptimized | 369.0924 sec
  | Optimized | 2.433 sec

🧪 Unit Tests:
The unit tests added in #33284 continue to pass. The new implementation requires that cross-host transfers are always between a local device and a remote device; a test has been added to make sure that the implementation throws an InvalidArgument error if this is not the case.

🧪 Execution Tests:
When building XLA with the code in this PR along with [a patch for pjrt_ifrt/pjrt_client.cc](https://gist.github.com/rao-ashish/ba5cd773906c8e428147bb80a3568912) so that it uses the new cross-host transfers API, I verified that [these 4 correctness checks](https://gist.github.com/rao-ashish/24ac0df0cb18243c649ac535964b31b8) pass. These IFRT changes will be included in a follow-up PR once other PjRt clients have implemented the new API.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -597,10 +597,11 @@ absl::StatusOr<PreparedReceive> PrepareReceive(
       client->AllocateRawBuffer(memory_space, on_device_bytes_count,
                                 /*retry_on_oom=*/true,
                                 /*allocate_after=*/{}));
-  TF_ASSIGN_OR_RETURN(
-      std::unique_ptr<PjRtBuffer> buffer,
-      client->DefineBuffer(on_device_shape, raw_buffer, {definition_event},
-                           /*raw_buffer_is_mutable=*/true));
+
+  TF_ASSIGN_OR_RETURN(std::unique_ptr<PjRtBuffer> buffer,
+                      client->DefineBuffer(on_device_shape, memory_space,
+                                           raw_buffer, {definition_event},
+                                           /*raw_buffer_is_mutable=*/true));
   definition_event->AndThen([raw_buffer]() {});
 
   return PreparedReceive{
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -512,7 +512,8 @@ absl::StatusOr<PreparedSend> PrepareSend(
     se::Stream* stream, PjRtBuffer* buffer,
     PjRtGlobalDeviceId dst_global_device_id, CrossHostTransferKey transfer_key,
     gpu::AcquiredCliquesMap& acquired_cliques_map,
-    std::shared_ptr<Future<>::Promise> promise) {
+    std::shared_ptr<Future<>::Promise> promise,
+    tsl::RCReference<PjRtStreamExecutorDeviceEvent> usage_event) {
   // Form the GPU clique key.
   // TODO(asrao, mwhittaker): Supply correct incarnations when creating the
   // clique key.
@@ -526,11 +527,6 @@ absl::StatusOr<PreparedSend> PrepareSend(
       AcquiredCliqueAndCommunicator clique_and_communicator,
       AcquireCliqueAndCommunicator(client, gpu_collectives, clique_key,
                                    acquired_cliques_map, RankId(0), stream));
-
-  // Create a usage event to track when the send is done.
-  tsl::RCReference<PjRtStreamExecutorDeviceEvent> usage_event =
-      tsl::MakeRef<PjRtStreamExecutorDeviceEvent>(
-          BufferSequencingEvent::Create(client->thread_pool()));
 
   // Acquire a hold on this buffer. The hold is held as long as usage_event is
   // not fulfilled; this behavior is achieved by registering a 'dummy' closure
@@ -571,7 +567,8 @@ absl::StatusOr<PreparedReceive> PrepareReceive(
     StreamExecutorGpuClient* client, gpu::GpuCollectives* gpu_collectives,
     se::Stream* stream, PjRtDevice* device, PjRtMemorySpace* memory_space,
     PjRtGlobalDeviceId src_global_device_id, CrossHostTransferKey transfer_key,
-    Shape shape, gpu::AcquiredCliquesMap& acquired_cliques_map) {
+    Shape shape, gpu::AcquiredCliquesMap& acquired_cliques_map,
+    tsl::RCReference<PjRtStreamExecutorDeviceEvent> definition_event) {
   // Form the GPU clique key.
   // TODO(asrao, mwhittaker): Supply correct incarnations when creating the
   // clique key.
@@ -585,11 +582,6 @@ absl::StatusOr<PreparedReceive> PrepareReceive(
       AcquiredCliqueAndCommunicator clique_and_communicator,
       AcquireCliqueAndCommunicator(client, gpu_collectives, clique_key,
                                    acquired_cliques_map, RankId(1), stream));
-
-  // Create a definition event to track when the receive is done.
-  tsl::RCReference<PjRtStreamExecutorDeviceEvent> definition_event =
-      tsl::MakeRef<PjRtStreamExecutorDeviceEvent>(
-          BufferSequencingEvent::Create(client->thread_pool()));
 
   // Allocate an uninitialized buffer. The buffer will be populated with data
   // received from the sending process.
@@ -742,27 +734,28 @@ void StreamExecutorGpuClient::ScheduleSendsOnLocalDevice(
     const std::vector<CrossHostTransferKey> transfer_keys,
     std::vector<std::shared_ptr<Future<>::Promise>> promises) {
   // Get the local device state, transfer stream, and prepare the send
-  // buffers.
+  // buffers. We associate the group of sends with a single usage_event.
   LocalDeviceState* local_device_state;
   se::Stream* stream;
   std::vector<PreparedSend> prepared_sends;
   prepared_sends.reserve(buffers.size());
+  tsl::RCReference<PjRtStreamExecutorDeviceEvent> usage_event;
 
   auto setup_sends = [&]() -> absl::Status {
     TF_ASSIGN_OR_RETURN(local_device_state, GetLocalDeviceState(device));
     stream = local_device_state->GetDeviceToDeviceStream();
     gpu::GpuCollectives* gpu_collectives = gpu::GpuCollectives::Default();
+    usage_event = tsl::MakeRef<PjRtStreamExecutorDeviceEvent>(
+        BufferSequencingEvent::Create(this->thread_pool()));
 
     gpu::AcquiredCliquesMap acquired_cliques_map;
     for (int i = 0; i < buffers.size(); ++i) {
       absl::StatusOr<PreparedSend> prepared_send = PrepareSend(
           this, gpu_collectives, stream, buffers[i], dst_global_device_ids[i],
-          transfer_keys[i], acquired_cliques_map, promises[i]);
+          transfer_keys[i], acquired_cliques_map, promises[i], usage_event);
 
       if (!prepared_send.ok()) {
-        for (PreparedSend& prev_prepared_send : prepared_sends)
-          SetEventAsError(prev_prepared_send.usage_event->event(),
-                          prepared_send.status());
+        SetEventAsError(usage_event->event(), prepared_send.status());
         return prepared_send.status();
       }
 
@@ -807,11 +800,16 @@ void StreamExecutorGpuClient::ScheduleSendsOnLocalDevice(
   // Form the closure to schedule on the device's execute thread.
   auto execute_sends_fn = [this, local_device_state, stream,
                            prepared_sends = std::move(prepared_sends),
-                           launch_send_group =
-                               std::move(launch_send_group)]() mutable {
+                           launch_send_group = std::move(launch_send_group),
+                           usage_event = std::move(usage_event)]() mutable {
     // Group transfers by GPU clique.
     absl::flat_hash_map<gpu::GpuCliqueKey, std::vector<PreparedSend>>
         grouped_sends = GroupSendsByCliqueKey(std::move(prepared_sends));
+
+    // Transfers for a particular clique are executed as a group. This vector
+    // holds group futures for each clique_key in grouped_sends.
+    std::vector<Future<>> group_futures;
+    group_futures.reserve(grouped_sends.size());
 
     for (auto& [clique_key, curr_sends] : grouped_sends) {
       // Get the communicator on which we will execute this group of
@@ -822,28 +820,31 @@ void StreamExecutorGpuClient::ScheduleSendsOnLocalDevice(
           curr_sends[0].clique_and_communicator.second;
 
       // Launch the group of transfers.
-      Future<> group_future = gpu_communicator->GroupExecute(
+      group_futures.push_back(gpu_communicator->GroupExecute(
           [&launch_send_group, &curr_sends,
            stream](gpu::GpuCommunicator* gpu_comm) -> absl::Status {
             return launch_send_group(gpu_comm, absl::MakeSpan(curr_sends),
                                      stream);
-          });
+          }));
+    }
 
-      // On a separate thread pool, await the group future and fulfill buffer
-      // sequencing events and promises.
-      group_future.OnReady(
-          *this->thread_pool()->AsExecutor(),
-          [this, local_device_state, stream,
-           curr_sends =
-               std::move(curr_sends)](const absl::Status& status) mutable {
+    // On a separate thread pool, await group futures and fulfill buffer
+    // sequencing events and promises.
+    Future<> all_sends_future = JoinFutures(group_futures);
+
+    all_sends_future.OnReady(
+        *this->thread_pool()->AsExecutor(),
+        [this, local_device_state, stream,
+         grouped_sends = std::move(grouped_sends),
+         usage_event](const absl::Status& status) mutable {
+          absl::Status fulfill_status = FulfillDeviceEvent(
+              this, local_device_state, stream, usage_event, status);
+          for (auto& [clique_key, curr_sends] : grouped_sends) {
             for (PreparedSend& prepared_send : curr_sends) {
-              absl::Status fulfill_status =
-                  FulfillDeviceEvent(this, local_device_state, stream,
-                                     prepared_send.usage_event, status);
               prepared_send.promise->Set(status.ok() ? fulfill_status : status);
             }
-          });
-    }
+          }
+        });
   };
 
   // Schedule transfers on the execute thread.
@@ -931,13 +932,14 @@ StreamExecutorGpuClient::CrossHostReceiveBuffers(
   }
 
   // Get the local device state, transfer stream, and prepare the receive
-  // buffers.
+  // buffers. We associate the group of receives with a single definition_event.
   LocalDeviceState* local_device_state;
   se::Stream* stream;
   std::vector<std::unique_ptr<PjRtBuffer>> buffers;
   buffers.reserve(shapes.size());
   std::vector<PreparedReceive> prepared_receives;
   prepared_receives.reserve(shapes.size());
+  tsl::RCReference<PjRtStreamExecutorDeviceEvent> definition_event;
 
   auto setup_receives = [&]() -> absl::Status {
     TF_ASSIGN_OR_RETURN(local_device_state, GetLocalDeviceState(device));
@@ -945,18 +947,18 @@ StreamExecutorGpuClient::CrossHostReceiveBuffers(
     TF_ASSIGN_OR_RETURN(PjRtMemorySpace * memory_space,
                         device->default_memory_space());
     gpu::GpuCollectives* gpu_collectives = gpu::GpuCollectives::Default();
+    definition_event = tsl::MakeRef<PjRtStreamExecutorDeviceEvent>(
+        BufferSequencingEvent::Create(this->thread_pool()));
 
     gpu::AcquiredCliquesMap acquired_cliques_map;
     for (int i = 0; i < shapes.size(); ++i) {
       absl::StatusOr<PreparedReceive> prepared_receive =
           PrepareReceive(this, gpu_collectives, stream, device, memory_space,
                          src_global_device_ids[i], transfer_keys[i], shapes[i],
-                         acquired_cliques_map);
+                         acquired_cliques_map, definition_event);
 
       if (!prepared_receive.ok()) {
-        for (PreparedReceive& prev_prepared_receive : prepared_receives)
-          SetEventAsError(prev_prepared_receive.definition_event->event(),
-                          prepared_receive.status());
+        SetEventAsError(definition_event->event(), prepared_receive.status());
         return prepared_receive.status();
       }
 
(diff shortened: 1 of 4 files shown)


xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1421,8 +1840,9 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
       pjrt_platform_name, xla_client, std::move(device_topology_pair.first),
       options.node_id, std::move(allocator), std::move(host_memory_allocator),
       options.should_stage_host_to_device_transfers, std::move(gpu_run_options),
-      std::move(kv_store), options.abort_collectives_on_failure,
-      std::move(gpu_topology), options.num_nodes);
+      std::move(kv_store), std::move(options.distributed_runtime_client),
```
mwhittaker: Ah, I unfortunately think this won't work for a very subtle reason. Are you familiar with the [PjRt C API](https://openxla.org/xla/pjrt)? Long story short, the GPU client gets created through a plugin system that goes from C++ to C and back to C++. Here's [the code that creates the GPU client](https://github.com/openxla/xla/blob/1bba37da00bc59a79ad4095c1b58fc933de4eb32/xla/pjrt/c/pjrt_c_api_gpu_internal.cc#L200). This means that every argument here has to be serialized from C++ to C and back to C++. This is very hard to do for the distributed runtime client. 

I think the `distributed_runtime_client` field should actually be removed. I added it to do something very similar to what you're doing, but later realized that it doesn't work :cry:

How about you change the CL to use an empty set of incarnations for now? And later we can work together to pipe through the correct incarnations? The code should still work with an empty set of incarnations; it just won't be fault tolerant. But I think that's fine for now.

rao-ashish: Got it thanks; I've just changed the code to use an empty set of incarnations for now.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -407,140 +665,183 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
+  // Form the closure to schedule on the device's execute thread.
+  auto execute_sends_fn = [this, local_device_state, stream,
+                           prepared_sends = std::move(prepared_sends),
+                           launch_send_group =
+                               std::move(launch_send_group)]() mutable {
+    // Group transfers by GPU clique.
+    absl::flat_hash_map<gpu::GpuCliqueKey, std::vector<PreparedSend>>
+        grouped_sends = GroupByCliqueKey(std::move(prepared_sends));
+
+    for (auto&& [clique_key, curr_sends] : std::move(grouped_sends)) {
```
emilyfertig: Why loop over the `std::move`d container like this instead of e.g. using an iterator?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -593,91 +895,139 @@ StreamExecutorGpuClient::CrossHostReceiveBuffers(
...
+      // Wait until the receive buffer is allocated.
+      WaitForAllocation(stream, *prepared_receive.raw_buffer);
+
+      // Launch the receive.
+      auto mem = tensorflow::down_cast<PjRtStreamExecutorRawBuffer*>(
+                     prepared_receive.raw_buffer.get())
+                     ->device_buffer();
+      TF_RETURN_IF_ERROR(gpu_communicator->LaunchRecv(
+          /*recv_buffer=*/mem->mem(),
+          /*dtype=*/xla::PrimitiveType::U8,
```
emilyfertig, line 1017: How come the dtype is hard coded?

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
```diff
@@ -2917,6 +2917,51 @@ TEST(StreamExecutorGpuClientTest, FailedCrossHostSendArgsSizeMismatch) {
                            "must have the same length, but got 1, 1, and 2.")));
 }
 
+TEST(StreamExecutorGpuClientTest, FailedCrossHostTransferSrcAndDstAddressable) {
```
emilyfertig, line 2935: I noticed `SuccessfulCrossHostTransferTest` only tests lists of arrays where each array has the same data, do you think it's worth varying that to ensure that send/recv order is preserved? (It might also be nice to test e.g. sends from one device to multiple destination devices, though I think our CI tests are limited to 2 GPUs so it couldn't run.)

emilyfertig (commented):
This looks good to me (though Michael knows the content better). Your code comments are really helpful.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+  PreparedReceive(const PreparedReceive&) = delete;
+  PreparedReceive& operator=(const PreparedReceive&) = delete;
+  PreparedReceive(PreparedReceive&&) = default;
+  PreparedReceive& operator=(PreparedReceive&&) = default;
+};
+
+// Acquire the GPU clique and communicator for a given clique key.
+absl::StatusOr<AcquiredCliqueAndCommunicator> AcquireCliqueAndCommunicator(
+    StreamExecutorGpuClient* client, gpu::GpuCollectives* gpu_collectives,
+    gpu::GpuCliqueKey clique_key, gpu::AcquiredCliquesMap& acquired_cliques_map,
```
mwhittaker:
```suggestion
    const gpu::GpuCliqueKey& clique_key, gpu::AcquiredCliquesMap& acquired_cliques_map,
```

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+                         /*executor=*/stream->parent(), RunId(0), clique_key,
+                         clique_id_callback, rank_id, acquired_cliques_map,
+                         /*max_nchannels=*/0));
+  }
+  std::shared_ptr<gpu::LockableGpuClique::Lock> clique =
+      acquired_cliques_map[clique_key];
+
+  // Get the communicator to use for this receive.
+  std::optional<Communicator*> maybe_communicator =
+      (*clique.get())->comm(rank_id);
```
mwhittaker:
```suggestion
      clique->comm(rank_id);
```

Does this work?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+  if (!maybe_communicator.has_value())
+    return absl::InternalError(
+        "AcquireCliqueAndCommunicator: Unable to get communicator from "
+        "acquired GPU clique.");
```
mwhittaker:
```suggestion
  if (!maybe_communicator.has_value()) {
    return absl::InternalError(
        "AcquireCliqueAndCommunicator: Unable to get communicator from "
        "acquired GPU clique.");
   }
```

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+      /*buffer=*/std::move(buffer),
+      /*raw_buffer=*/std::move(raw_buffer),
+      /*definition_event=*/std::move(definition_event),
+      /*clique_and_communicator=*/std::move(clique_and_communicator),
+  };
+}
+
+// Groups Prepared{Send,Receive} structs that have a `.clique_key` attribute.
+template <typename PreparedT>
+static absl::flat_hash_map<gpu::GpuCliqueKey, std::vector<PreparedT>>
```
mwhittaker: I don't think you need `static` if you're in an anonymous namespace?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+GroupByCliqueKey(std::vector<PreparedT>&& prepareds) {
+  absl::flat_hash_map<gpu::GpuCliqueKey, std::vector<PreparedT>> grouped;
+  grouped.reserve(prepareds.size());
```
mwhittaker: Am I understanding correctly that `grouped` can be smaller than `prepared.size()`? If so, we can skip the `reserve` call?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+      /*clique_key=*/std::move(clique_key),
+      /*buffer=*/std::move(buffer),
+      /*raw_buffer=*/std::move(raw_buffer),
+      /*definition_event=*/std::move(definition_event),
+      /*clique_and_communicator=*/std::move(clique_and_communicator),
+  };
+}
+
+// Groups Prepared{Send,Receive} structs that have a `.clique_key` attribute.
+template <typename PreparedT>
```
mwhittaker: Would you mind un-templating this function and maybe having two copies: one for `PreparedSend` and one for `PreparedReceive`? I think the duplication might be worth avoiding the complexities of templating?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+    se::Stream* stream,
+    tsl::RCReference<PjRtStreamExecutorDeviceEvent> device_event,
+    const absl::Status& status) {
+  if (!status.ok()) {
+    client->SetEventAsError(device_event->event(), status);
+    return absl::OkStatus();
+  }
+  absl::Status s = client->AllocateAndRecordEvent(device_event->event(),
+                                                  local_device_state, stream);
+  if (!s.ok()) client->SetEventAsError(device_event->event(), s);
```
mwhittaker:
```suggestion
  if (!s.ok()) { 
     client->SetEventAsError(device_event->event(), s);
  }
```

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -407,140 +665,183 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
-// and get its raw buffer and definition events. This is used to ensure that the
-// buffer is not deleted while the send is in progress.
-struct HeldSendBuffer {
-  tsl::RCReference<CommonPjRtRawBuffer> raw_buffer;
-  std::vector<tsl::RCReference<tsl::AsyncValue>> definition_events;
-};
+  // Group the sends by local device.
+  absl::flat_hash_map<PjRtDevice*, std::vector<int>> sends_by_device;
+  for (int i = 0; i < buffers.size(); ++i)
+    sends_by_device[buffers[i]->device()].push_back(i);
```
mwhittaker:
```suggestion
  for (int i = 0; i < buffers.size(); ++i) {
    sends_by_device[buffers[i]->device()].push_back(i);
  }
```

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -354,14 +355,28 @@ absl::Status StreamExecutorGpuClient::UpdateCompileOptionsInternal(
...
+      tensorflow::down_cast<PjRtStreamExecutorDevice*>(device);
+  return pjrt_se_device->GetLocalDeviceState();
 }
 
+// Get the device to device stream for a given PjRtDevice.
+absl::StatusOr<se::Stream*> GetDeviceToDeviceStream(PjRtDevice* device) {
+  TF_ASSIGN_OR_RETURN(LocalDeviceState * local_device,
+                      GetLocalDeviceState(device));
+  return local_device->GetDeviceToDeviceStream();
+}
```
mwhittaker: Is this function unused?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -407,140 +665,183 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
+        for (PreparedSend& prev_prepared_send : prepared_sends)
+          SetEventAsError(prev_prepared_send.usage_event->event(),
+                          prepared_send.status());
+        return prepared_send.status();
+      }
 
-  // Create an event to track when the send is done.
-  auto usage_event = tsl::MakeRef<PjRtStreamExecutorDeviceEvent>(
-      BufferSequencingEvent::Create(this->thread_pool()));
+      prepared_sends.push_back(std::move(*prepared_send));
```
mwhittaker:
```suggestion
      prepared_sends.push_back(*std::move(prepared_send));
```

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+  StreamExecutorGpuClient* client;
+  gpu::GpuCliqueKey clique_key;
+  tsl::RCReference<CommonPjRtRawBuffer> raw_buffer;
+  std::vector<tsl::RCReference<tsl::AsyncValue>> definition_events;
+  tsl::RCReference<PjRtStreamExecutorDeviceEvent> usage_event;
+  AcquiredCliqueAndCommunicator clique_and_communicator;
+  std::shared_ptr<Future<>::Promise> promise;
+
+  ~PreparedSend() {
+    if (!usage_event || usage_event->event()->IsDefined()) return;
```
mwhittaker:
```suggestion
    if (!usage_event || usage_event->event()->IsDefined()) { 
      return;
    }
```

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -393,6 +408,249 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+
+  StreamExecutorGpuClient* client;
+  gpu::GpuCliqueKey clique_key;
+  std::unique_ptr<PjRtBuffer> buffer;
+  tsl::RCReference<CommonPjRtRawBuffer> raw_buffer;
+  tsl::RCReference<PjRtStreamExecutorDeviceEvent> definition_event;
+  AcquiredCliqueAndCommunicator clique_and_communicator;
+
+  ~PreparedReceive() {
+    if (!definition_event || definition_event->event()->IsDefined()) return;
```
mwhittaker:
```suggestion
    if (!definition_event || definition_event->event()->IsDefined()) {
        return;
    }
```

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -407,140 +665,183 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
+                              se::Stream* stream) -> absl::Status {
+    for (PreparedSend& prepared_send : prepared_sends) {
       // Wait until the buffer we want to send is fully materialized.
-      for (const auto& event : held_send_buffer.definition_events) {
+      for (const auto& event : prepared_send.definition_events) {
         tsl::BlockUntilReady(event.get());
-        if (auto* status = event->GetErrorIfPresent()) {
-          return *status;
-        }
+        if (auto* status = event->GetErrorIfPresent()) return *status;
```
mwhittaker:
```suggestion
        if (auto* status = event->GetErrorIfPresent(); status != nullptr) {
          return *status;
        }
```

mwhittaker (commented):
Looks great. @pschuh, could you take a look at the event-related stuff?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Thanks for taking a look! There was no specific reason for this; I was just using a range-based loop because it seemed slightly less verbose. I've changed this line to `for (auto& [clique_key, curr_sends] : grouped_sends)` (and analogously for the receives) because I think the two should be functionally equivalent and the latter is a bit simpler.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: This was actually what the original `ScheduleRemoteSend` and `MakeReceiveBuffers` did, so I kept this behavior. It's still correct for any array dtype because we are passing the number of bytes as the element count (`mem->mem()`). I suppose the reason for doing this in the original implementation was just to make things a bit simpler / to not have to add code that dynamically selects the dtype based on the input.

xla/pjrt/gpu/se_gpu_pjrt_client_test.cc
rao-ashish: Good idea, I've changed the test to vary the data in the arrays (it now uses [0 ... 255], [1000 ... 1255], [2000 ... 2255]). For now, I haven't added the send from one device to multiple destinations due to the 2 GPU limit you mentioned.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Thanks for taking a look! Changed to `const gpu::GpuCliqueKey& `.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: No, that ends up giving a compile error. But `(*clique)->comm(rank_id);` does work though, so I've removed the `.get()`.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Ah yes, thanks, removed the unnecessary `static`s.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Separated into `GroupSendsByCliqueKey`, `GroupReceivesByCliqueKey`.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Oops it wasn't; I've removed it.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: You're right, `prepared.size()` is an upper bound. I was thinking it would be safer to reserve space as per this upper bound to avoid reallocating and moving nodes if / as the map grows.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -593,91 +906,140 @@ StreamExecutorGpuClient::CrossHostReceiveBuffers(
...
+          });
+
+      // On a separate thread pool, await the group future and fulfill buffer
+      // sequencing events.
+      group_future.OnReady(
+          *this->thread_pool()->AsExecutor(),
+          [this, local_device_state, stream,
+           curr_receives =
+               std::move(curr_receives)](const absl::Status& status) mutable {
+            for (PreparedReceive& prepared_receive : curr_receives) {
```
pschuh: I think we should maybe use CreateLinkedEventPromise when building the definition events (we would then store promises instead of events) so that we only need to call FulfilDeviceEvent once (which creates a single cuda event).

rao-ashish: Thanks for reviewing this! I see what you mean, it would be better to create only one event per group of sends / receives. Instead of CreateLinkedEventPromise, I think we can get the same behavior by preallocating a single usage / definition event, passing it into Prepare{Sends/Receives}, and having each Prepared{Send/Receive} object re-use this event. I've added a commit that does this, let me know if this works.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
mwhittaker: Got it, sounds good to me!

mwhittaker (approved):

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -46,6 +46,7 @@ limitations under the License.
 #include "absl/time/time.h"
 #include "absl/types/span.h"
 #include "mlir/IR/BuiltinOps.h"
+#include "xla/backends/gpu/collectives/gpu_communicator.h"
```
emilyfertig, line 50: @rao-ashish This is missing from the BUILD file, could you please add it, as well as the other new deps in `se_gpu_pjrt_client.h`?

rao-ashish:
I've added `gpu_communicator` as a dependency; this should fix the `gpu_communicator.h` include in `se_gpu_pjrt_client.cc`. `se_gpu_pjrt_client.h` added includes `gpu_clique_key.h`, `gpu_cliques.h`, and `se_raw_buffer.h`, which should be covered by the existing dependencies in the BUILD file on `gpu_clique_key`, `gpu_cliques`, and `pjrt_stream_executor_client`.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
pschuh: Yea, that also works.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -392,6 +400,264 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+// Helper structs / classes for second cross-host transfers API.
+using AcquiredCliqueAndCommunicator =
+    std::pair<std::shared_ptr<gpu::LockableGpuClique::Lock>,
+              gpu::GpuCommunicator*>;
+
+class PreparedSend {
+ public:
+  PreparedSend() = delete;
+
+  StreamExecutorGpuClient* client;
```
emilyfertig: Could you please give the class member variable names here (and in `PreparedReceive`) trailing underscores? Internal clang-tidy is complaining https://google.github.io/styleguide/cppguide.html#Variable_Names

xla/pjrt/gpu/se_gpu_pjrt_client.h
```diff
@@ -206,9 +209,11 @@ class StreamExecutorGpuClient : public xla::PjRtStreamExecutorClient {
   // Helpers for cross host transfers.
   absl::Duration cross_host_transfer_timeout_ = absl::Minutes(3);
 
-  absl::StatusOr<Future<>> CrossHostSendBuffer(
-      PjRtBuffer* buffer, PjRtGlobalDeviceId dst_global_device_id,
-      CrossHostTransferKey transfer_key);
+  void ScheduleSendsOnLocalDevice(
+      PjRtDevice* device, std::vector<PjRtBuffer*> buffers,
+      const std::vector<PjRtGlobalDeviceId> dst_global_device_ids,
```
emilyfertig: Could you remove the const qualifiers in the declaration here?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -392,6 +400,252 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
   return std::move(communicators[0]);
 }
 
+// Helper structs / classes for second cross-host transfers API.
+using AcquiredCliqueAndCommunicator =
+    std::pair<std::shared_ptr<gpu::LockableGpuClique::Lock>,
```
emilyfertig, line 408: Add `#include "xla/backends/gpu/collectives/gpu_clique.h"` for this?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -392,6 +400,252 @@ absl::StatusOr<std::unique_ptr<Communicator>> CreateTransferCommunicator(
...
+      client->gpu_run_options(dummy_execute_options);
+  gpu::CliqueIdCallback clique_id_callback =
+      dummy_gpu_run_options->clique_id_callback();
+
+  // Acquire the GPU clique for this receive.
+  if (!acquired_cliques_map.contains(clique_key)) {
+    TF_ASSIGN_OR_RETURN(
+        acquired_cliques_map[clique_key],
+        AcquireGpuClique(gpu_collectives,
+                         /*executor=*/stream->parent(), RunId(0), clique_key,
```
emilyfertig: Rename `executor` to `device`? (Linter says "argument name 'executor' in comment does not match parameter name 'device'")

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -406,140 +660,208 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
+      // Get the communicator on which we will execute this group of
+      // transfers. We assume each clique key is associated with a unique
+      // communicator, so we just take the communicator of the first
+      // transfer_idx of this clique key.
+      gpu::GpuCommunicator* gpu_communicator =
+          curr_sends[0].clique_and_communicator.second;
+
+      // Launch the group of transfers.
+      group_futures.push_back(gpu_communicator->GroupExecute(
+          [&launch_send_group, &curr_sends,
```
emilyfertig: Linter says "captured structured bindings are incompatible with C++ standards before C++20" for `&curr_sends`. I think it can be fixed by using `&curr_sends = curr_sends` in the capture instead. (Same with `curr_receives` on 1029).

Sorry for adding these linter comments piecemeal, I think that's the last of it.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
emilyfertig: If you leave these as structs, could you remove the trailing underscore again? Seems like that's only the convention for classes.

PR #21800, opened 2025-01-24
[XLA:GPU] Add block scaling rewriter pass

sergey-kozub:
This PR adds a transformation pass that supports custom calls to block quantize/dequantize/dot ops.
Such calls are replaced by an equivalent sequence of HLO operations.

This pass is supposed to support MX scaling formats, such as MXFP8, but is not limited to those and can be used with any data types and block sizes.
The quantization op sequence matches the one described in the section 6.3 of the MX spec:
https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf

Once cuDNN frontend 1.10 is released, a lowering to a cuDNN graph will be enabled for the hardware that supports block scaled dot natively (i.e. Blackwell). This pass will stay disabled until then.

I also plan on introducing a new HLO op, "block-scaled-dot", which will be more generic than a custom call - for example, will have configurable dimensions numbers akin to the general dot op. This will follow in a separate PR, once that is approved, I'll replace the custom call "__op$block_scaled_dot" with it.

WANT_LGTM=any

PR #28173, opened 2025-06-24
[ROCm] Enable mx data type for ROCm

ScXfjiang:
In this PR, we support MX datatypes for the ROCm platform. While the CUDA path uses **cuDNN**,  ROCm implements MX datatypes through **hipBLASLt**.

### Key updates
* **BlockScalingRewriter**  
  * Rewrites `__op$block_scaled_dot` into an HLO graph that invokes the hipBLASLt custom call `__cublaslt$matmul$mx`.

* **CublasLtMatmulThunk**  
  * Extends the existing lowering stage so the new custom call is finally dispatched to a hipBLASLt kernel.

Note:
* `GemmAlgorithmPicker` isn't enabled for MX datatypes. We will support it in a separate PR.

### Reference

- https://arxiv.org/pdf/2310.10537
- https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf
- https://github.com/openxla/xla/pull/21800
- https://github.com/openxla/xla/pull/22256

PR #33947, opened 2025-11-14
Ci dev mx datatype rocm

ScXfjiang:
In this PR, we support MX datatypes for the ROCm platform. While the CUDA path uses **cuDNN**,  ROCm implements MX datatypes through **hipBLASLt**.

### Key updates
* **BlockScalingRewriter**  
  * Rewrites `__op$block_scaled_dot` into an HLO graph that invokes the hipBLASLt custom call `__cublaslt$matmul$mx`.

* **CublasLtMatmulThunk**  
  * Extends the existing lowering stage so the new custom call is finally dispatched to a hipBLASLt kernel.

Note:
* `GemmAlgorithmPicker` isn't enabled for MX datatypes. We will support it in a separate PR.

### Reference

- https://arxiv.org/pdf/2310.10537
- https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf
- https://github.com/openxla/xla/pull/21800
- https://github.com/openxla/xla/pull/22256

### DIFF
--- a/xla/stream_executor/gpu/gpu_blas_lt.h
+++ b/xla/stream_executor/gpu/gpu_blas_lt.h
@@ -129,6 +129,7 @@ struct GemmConfig {  // plain GemmConfig which is extended with create functions
   std::optional<int64_t> algorithm;
   bool grad_x;
   bool grad_y;
+  bool mx_mode;
   std::optional<blas::ComputationType> compute_type;
 
   static absl::StatusOr<GemmConfig> FromProto(
--- a/xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
+++ b/xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
@@ -344,7 +344,7 @@ static absl::Status AppendCommands(CommandBufferCmdSequence& cmd_sequence,
       return Internal(
           "Error trying to emit command for a CommandBufferThunk. Input HLO "
           "must already contain command buffers and XLA should not run command "
-          "buffer scheduling pass the second time. It it happens in the test, "
+          "buffer scheduling pass the second time. If it happens in the test, "
           "try explicitly disabling command buffers in tested HLO module.");
 
     default:
--- a/xla/backends/gpu/runtime/command_buffer_thunk_test.cc
+++ b/xla/backends/gpu/runtime/command_buffer_thunk_test.cc
@@ -718,7 +718,7 @@ TEST(CommandBufferThunkTest, GemmCmd) {
       ShapeUtil::MakeShape(PrimitiveType::F32, {4, 3}), {}, {0},
       ShapeUtil::MakeShape(PrimitiveType::F32, {2, 3}), 1.0, 0.0, 0.0,
       PrecisionConfig::ALG_UNSET, std::nullopt,
-      se::blas::kDefaultComputePrecision, false, false,
+      se::blas::kDefaultComputePrecision, false, false, false,
       stream_executor->GetDeviceDescription().gpu_compute_capability());
   ASSERT_TRUE(config.ok());
 
@@ -839,7 +839,7 @@ TEST(CommandBufferThunkTest, ChildGemmCmd) {
       ShapeUtil::MakeShape(PrimitiveType::F32, {4, 3}), {}, {0},
       ShapeUtil::MakeShape(PrimitiveType::F32, {2, 3}), 1.0, 0.0, 0.0,
       PrecisionConfig::ALG_UNSET, std::nullopt,
-      se::blas::kDefaultComputePrecision, false, false,
+      se::blas::kDefaultComputePrecision, false, false, false,
       stream_executor->GetDeviceDescription().gpu_compute_capability());
   ASSERT_TRUE(config.ok());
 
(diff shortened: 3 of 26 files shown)


ScXfjiang:
Hi @mooskagh @loislo, could you take a look please? Thanks!

ScXfjiang:
Hi @mooskagh, could you take a look please?

xla/stream_executor/gpu/gpu_blas_lt.h
```diff
@@ -129,6 +129,7 @@ struct GemmConfig {  // plain GemmConfig which is extended with create functions
   std::optional<int64_t> algorithm;
   bool grad_x;
   bool grad_y;
+  bool mx_mode;
```
mooskagh, line 132: Why do we need an `mx_mode` as a GemmConfig option?

Cannot we just look at the Gemm operands whether they are FP8 types or not?

mooskagh:
As far as I understand, the plan is to get rid of backend-specific custom calls, and instead use a new `scaled-dot` instruction that was recently added.

@loislo is likely to have more context.

xla/stream_executor/gpu/gpu_blas_lt.h
ScXfjiang: Hi @mooskagh, as I understand, it's hard to tell if the matmul is a block-scaling mamtul or a pure FP8 matmul by checking the data type and # of operands, because for some situations, they are exactly the same. I prefer to add a new filed in the config.

For example,
The operands of a FP8 matmul can be: a (FP8), b (FP8), a_scale, b_scale.
The operands of a BLock-Scaling Matmul can be the same.

ScXfjiang:
> As far as I understand, the plan is to get rid of backend-specific custom calls, and instead use a new `scaled-dot` instruction that was recently added.
> 
> @loislo is likely to have more context.

Thanks for the info. Yes, that was why I asked in the old PR: https://github.com/openxla/xla/pull/28173#issuecomment-3517821245. Could you confirm about the dev plan here? Many thanks! @mooskagh @loislo

ScXfjiang:
I was wondering could we merge this PR first and I will take care of the incoming changes for ROCm, otherwise, I have to update the PR again and again here.

xla/stream_executor/gpu/gpu_blas_lt.h
mooskagh: If types and shapes are the same, how are these matmuls different?  
Is there any difference in semantics, or it's just whatever machinery is used underneath?

ScXfjiang: The types and shapes can be the same for only a few scenarios, 

A typical example:
- Both FP8 matmul and block-scaling matmul use FP8 input -- input shapes & types are the same.
- The inputs matrices are small so the scaling factors (`a_scale`, `b_scale`) are only scalars of the same type.

But I believe for most of the time, they are different, especially for the scaling factors.
- FP8 matmul use only scalar scaling factors.
- Block-scaling factors use multiple scaling factors for multiple blocks.

So it's hard to tell if the matmul is FP8 matmul or block-scaling matmul in RewritePass or IREmitter. My solution is to set a new field to do this.

ScXfjiang: Yes, they have different semantics from op level, not just different backend implementations.

ScXfjiang: FYI: here is how we enable mx mode for hipblaslt matmul:

https://github.com/ROCm/xla/blob/e9269f1b6d1bcc3205ca38788880b8ce3ae30b9e/xla/stream_executor/rocm/hip_blas_lt.cc#L270-L279

ScXfjiang:
Hi @mooskagh @loislo , could you take a look please. Many thanks!

mooskagh:
I've asked around, and here's my understanding of the final state we are aiming for:
- The `block_scaling_rewriter.cc`, rewriting from/to custom_call, will go away.
- The scaled dot will be expressed as `scaled-dot` instruction (it may enter XLA as a composite or custom-call but then immediately rewritten).
- The `scaled-dot` would be fused into normal fusions in a central place like Priority Fusion or in general fusion pipeline, aiming to form a fusion that would be supported by the main fusion backend (currently, Triton-based emitter).
- After that, different backends (cudnn, triton, cublas) would be run through autotuner to determine the fastest one.

I'm not sure what is the expected timeline for these changes.

ScXfjiang:
> I've asked around, and here's my understanding of the final state we are aiming for:
> 
> * The `block_scaling_rewriter.cc`, rewriting from/to custom_call, will go away.
> * The scaled dot will be expressed as `scaled-dot` instruction (it may enter XLA as a composite or custom-call but then immediately rewritten).
> * The `scaled-dot` would be fused into normal fusions in a central place like Priority Fusion or in general fusion pipeline, aiming to form a fusion that would be supported by the main fusion backend (currently, Triton-based emitter).
> * After that, different backends (cudnn, triton, cublas) would be run through autotuner to determine the fastest one.
> 
> I'm not sure what is the expected timeline for these changes.

Hi @mooskagh, thanks so much for the update, and I'm looking forward to the final version! Will you update the community, maybe by a presentation or an official post, once it's done?

ScXfjiang:
The PR is deprecated. ROCm scaled dot implementation: https://github.com/openxla/xla/pull/38333

PR #34034, opened 2025-11-16
[XLA:GPU] Add conv fusion rewriter

Cjkkkk:
📝 Summary of Changes
*	Added fusion model to grow fusion staring from operands and users. It enables epilogue activation fusion/integer conv fusion/ fp8 conv fusion/multi-output conv fusion for fp8 amax.
*	Added float normalization support to skip adding type conversion for conv fusion.
*	Added layout assignment for conv fusion to enforce NHWC layout.
*	Added window(stride/padding/dilation) construction for backward conv because cuDNN requires backward conv to use same window as forward convolution.
*	Added layout normalization to normalize layout to have descending layout.
*	Reused cudnn_fused_conv_rewriter_test.cc in conv_fusion_rewriter_test.cc to test new conv fusion, need to rewrite fp8 patterns because new cudnn frontend doesn’t allow prologue fusion on weight. 

🎯 Justification
XLA cuDNN conv refactoring to express conv as cuDNN fusion instead of custom call.

🚀 Kind of Contribution
✨ New Feature

No major speed up is expected as this is refactoring of cuDNN conv. It does enable more fusions with conv so there might be some speed up/regression.

🧪 Unit Tests:
reused unit tests from cudnn_fused_conv_rewriter_test.cc in conv_fusion_rewriter_test.cc

🧪 Execution Tests:
reused unit tests from cudnn_fused_conv_rewriter_test.cc in conv_fusion_rewriter_test.cc


xla/service/gpu/transforms/layout_assignment.cc
```diff
@@ -493,6 +624,11 @@ absl::Status GpuLayoutAssignment::AddBackendConstraints(
           Cast<HloCustomCallInstruction>(instruction), constraints));
     }
 
+    if (IsCuDNNConvolutionFusion(*instruction)) {
```
mooskagh: There's already IsGpuFusionKind() in `service/gpu/hlo_fusion_analysis.h`, could it be reused?

xla/service/gpu/conv_layout_normalization.cc
```diff
@@ -181,6 +184,130 @@ absl::StatusOr<std::optional<HloInstruction*>> UpdateLayoutForCudnnConvolution(
...
+  const Shape& output_shape = conv->shape();
+
+  auto MaybeBitcast = [](HloInstruction* hlo,
+                         const Shape& original_shape) -> HloInstruction* {
+    return hlo->shape() == original_shape ? hlo
+                                          : MakeBitcastHlo(hlo, original_shape);
+  };
+
+  bool performed_normalization = false;
+  // Normalize fusion instrcution shape if needed
```
derdrdirk: Nit: instruction.

xla/service/gpu/conv_layout_normalization.cc
```diff
@@ -181,6 +184,130 @@ absl::StatusOr<std::optional<HloInstruction*>> UpdateLayoutForCudnnConvolution(
...
+    }
+    normalized_shape = ShapeUtil::MakeTupleShape(new_tuple_shape);
+  } else {
+    normalized_shape =
+        ShapeUtil::MakeShapeWithDescendingLayoutAndSamePhysicalLayout(
+            hlo->shape());
+  }
+  performed_normalization |= normalized_shape != hlo->shape();
+
+  // Normalize fusion instrcution operand shape if needed
```
derdrdirk: Nit: instruction

xla/service/gpu/conv_layout_normalization.cc
```diff
@@ -181,6 +184,130 @@ absl::StatusOr<std::optional<HloInstruction*>> UpdateLayoutForCudnnConvolution(
...
+  for (int idx = 0; idx < hlo->operand_count(); idx++) {
+    HloInstruction* operand = hlo->mutable_operand(idx);
+    Shape normalized_operand_shape =
+        ShapeUtil::MakeShapeWithDescendingLayoutAndSamePhysicalLayout(
+            operand->shape());
+    performed_normalization |= normalized_operand_shape != operand->shape();
+    normalized_operands.emplace_back(
+        MaybeBitcast(operand, normalized_operand_shape));
+  }
+  // No normalized required, don't do anything
```
derdrdirk: Nit: No normalization

derdrdirk (changes requested):
Hi Shanbin,
thanks for the changes.

Could you move the ConvFusionRewriter to [FusionPipeline](https://github.com/openxla/xla/blob/main/xla/service/gpu/fusion_pipeline.cc) rather than running it as a pre-layout pass. We aim to keep all fusion passes after layout assignment and normalization.

To handle the layout requirements (e.g., NHWC vs NCHW) for the convolution operators:

1. We should add the necessary operator constraints to the kConvolution instruction itself during LayoutAssignment.
2. Then, when the FusionPipeline runs, we can check the assigned layouts of the convolution and its producers/consumers to determine if they are compatible for fusion.

Cjkkkk:
> Hi Shanbin, thanks for the changes.
> 
> Could you move the ConvFusionRewriter to [FusionPipeline](https://github.com/openxla/xla/blob/main/xla/service/gpu/fusion_pipeline.cc) rather than running it as a pre-layout pass. We aim to keep all fusion passes after layout assignment and normalization.
> 
> To handle the layout requirements (e.g., NHWC vs NCHW) for the convolution operators:
> 
> 1. We should add the necessary operator constraints to the kConvolution instruction itself during LayoutAssignment.
> 2. Then, when the FusionPipeline runs, we can check the assigned layouts of the convolution and its producers/consumers to determine if they are compatible for fusion.

Thanks @derdrdirk for the feedback , one concern of adding constraints for the conv is that cudnn fusion doesn't allow transpose in the fusion. But one WAR I can think of is that we can treat the transpose as the real conv and fuse users of transpose and then move transpose outside fusion.

derdrdirk:
Can you clarify what you mean by "treating the transpose as the real conv"?

In principle layout assignment should handle the layout propagation and we should only fuse operators with supported layouts. 
Let's assume the following order of instructions `input -> operator_conv -> conv -> bias`, with input having a layout of NCHW and con being constrained to NHWC then layout assignment should insert the transform before the operator: `input (NCHW) -> transform (NCHW to NHWC) -> operator_conv (NHWC) -> conv (NHWC) -> bias`. In this case we can fuse the prologue (operator_conv). If the transform would be before the conv we can only fuse the epilogue (bias).

Cjkkkk:
what is this operator conv here? From my understanding, for example, for fp8 conv, assuming conv takes NCHW originally and layout assignment wants to assign NHWC layout:

`conv(input * scale, filter * scale) + bias`
=> layout assignment
`transpose_to_NCHW(conv(transpose_to_NHWC( input * scale), transpose_to_NHWC( filter * scale))) + bias `

But cudnn doesn't allow transpose in the graph, which makes scale and bias not fusible.

If transpose can be inserted before input scale and after bias, it allows everything to be fusible and transpose will be outside of fusion
`transpose_to_NCHW(conv(transpose_to_NHWC( input) * scale, transpose_to_NHWC( filter) * scale) + bias)`

We can do this by fusing everything(including transpose) into cudnn fusion, then move transpose out of the graph afterwards.

derdrdirk:
Thanks for the clarification. IMO we should let layout_assignment insert the copy/ transpose before the scales. Layout assignment can add constraints to the operator and if we give it a high enough priority it will propagate the transpose up the producer chain. Then you would have no transpose and could cleanly fuse the convolution.

IMO the convolution fusion rewriter shouldn't know about transposes. If there is a transpose, i.e. an operator that it cannot fuse it just should not fuse it.

---
 
E.g. copy after scale and before conv (cannot be fused):
image

E.g. copy before scale (can be fused):
image

---

I did an experiment using Gemini generating the following code snippet
```
// [layout_assignment.cc](http://_vscodecontentref_/2)

// ... inside AddMandatoryConstraints ...

    } else if (instruction->opcode() == HloOpcode::kConvolution) {
      const ConvolutionDimensionNumbers& dnums =
          instruction->convolution_dimension_numbers();
      // Input layout: NHWC.
      std::vector<int64_t> input_layout;
      input_layout.push_back(dnums.input_feature_dimension());
      for (int i = dnums.input_spatial_dimensions_size() - 1; i >= 0; --i) {
        input_layout.push_back(dnums.input_spatial_dimensions(i));
      }
      input_layout.push_back(dnums.input_batch_dimension());
      // Increase priority to propagate layout up to producers
      TF_RETURN_IF_ERROR(SetArrayOperandLayout(
          LayoutUtil::MakeLayout(input_layout), instruction, 0,
          /*mandatory=*/true, /*dfs=*/true,
          /*priority=*/current_priority_ + 1));

      // Filter layout: OHWI.
      std::vector<int64_t> filter_layout;
      filter_layout.push_back(dnums.kernel_input_feature_dimension());
      for (int i = dnums.kernel_spatial_dimensions_size() - 1; i >= 0; --i) {
        filter_layout.push_back(dnums.kernel_spatial_dimensions(i));
      }
      filter_layout.push_back(dnums.kernel_output_feature_dimension());
      TF_RETURN_IF_ERROR(SetArrayOperandLayout(
          LayoutUtil::MakeLayout(filter_layout), instruction, 1,
          /*mandatory=*/true, /*dfs=*/true,
          /*priority=*/current_priority_ + 1));

    } else if (instruction->opcode() == HloOpcode::kParameter) {
       // ... existing checks ...
       if (parameter_layout.AnyLayoutIsSet()) {
          // ...
          // Use computation constraint priority to respect fixed layouts (e.g. entry params)
          Shape param_shape = parameter_layout.shape();
          TF_RETURN_IF_ERROR(SetInstructionLayout(
              param_shape, instruction, /*mandatory=*/true, /*dfs=*/true,
              /*allow_alias=*/true,
              /*priority=*/constraints->computation_constraint().priority()));
          // ...
       }
    }
```

using the following hlo
```
HloModule Test

ENTRY Test {
  zero = f16[] constant(0)
  zeros = f16[1,32,9,9] broadcast(zero), dimensions={}

  input = f16[1,17,9,9] parameter(0)
  filter = f16[3,3,17,32] parameter(1)

  conv = f16[1,32,9,9] convolution(input, filter), window={size=3x3 pad=1_1x1_1}, dim_labels=bf01_01io->bf01, feature_group_count=1
  ROOT relu = f16[1,32,9,9] maximum(zeros, conv)
}
```

Generating after layout_assignment
```
HloModule Test, entry_computation_layout={(f16[1,17,9,9]{3,2,1,0}, f16[3,3,17,32]{3,2,1,0})->f16[1,32,9,9]{3,2,1,0}}

ENTRY %Test (input: f16[1,17,9,9], filter: f16[3,3,17,32]) -> f16[1,32,9,9] {
  %zero = f16[] constant(0)
  %zeros = f16[1,32,9,9]{3,2,1,0} broadcast(%zero), dimensions={}
  %input = f16[1,17,9,9]{3,2,1,0} parameter(0)
  %copy = f16[1,17,9,9]{1,3,2,0} copy(%input)
  %input_scale_scalar = f16[] constant(0.5)
  %input_scale = f16[1,17,9,9]{1,3,2,0} broadcast(%input_scale_scalar), dimensions={}
  %scaled_input = f16[1,17,9,9]{1,3,2,0} multiply(%copy, %input_scale)
  %filter = f16[3,3,17,32]{3,2,1,0} parameter(1)
  %copy.1 = f16[3,3,17,32]{2,1,0,3} copy(%filter)
  %filter_scale = f16[3,3,17,32]{2,1,0,3} broadcast(%input_scale_scalar), dimensions={}
  %scaled_filter = f16[3,3,17,32]{2,1,0,3} multiply(%copy.1, %filter_scale)
  %conv = f16[1,32,9,9]{3,2,1,0} convolution(%scaled_input, %scaled_filter), window={size=3x3 pad=1_1x1_1}, dim_labels=bf01_01io->bf01
  ROOT %relu = f16[1,32,9,9]{3,2,1,0} maximum(%zeros, %conv)
}
```

Cjkkkk:
@derdrdirk thanks for the explanation! I am not aware of the priority usage here, I will try what you suggested.

Cjkkkk:
Hi @derdrdirk, I got some time this week and tried to work on moving the conv fusion after layout assignment. One issue during the refactoring is that in order to assign NHWC or NCHW layout, we need to know the kind of conv (fprop/dgrad/wgrad) which determines the correspondence between each dim and N, H, W, C.
 For example, consider a conv:
```
conv = f32[1,2,2,2] convolution(inputs_f32, filters_f32),
               window={size=1x1}, dim_labels=bf01_oi01->bf01
```
bf01, oi01 is NCHW, NCHW for fprop, NCHW, CNHW for dgrad, CNHW, CNHW  for wgrad. However we only choose conv kind during rewriting conv to conv fusion(add kind in the fusion config).

We could isolate the rewriter to two parts: rewriting to fusion with single conv and assign a conv kind (before layout assignment) and a conv fusion pass which fuses more ops (after layout assignment). Does this sound reasonable to you?

derdrdirk:
I see.

> We could isolate the rewriter to two parts: rewriting to fusion with single conv and assign a conv kind (before layout assignment) and a conv fusion pass which fuses more ops (after layout assignment). Does this sound reasonable to you?

Having a conv pass that determines the convolution type  (fprop, dgrad, wgrad) and storing the type in the convolution backend_config before layout assigment makes sense to me. Then layout assigment can assign and propagate the convolution type dependent layout and the fusion pass can be kept simple.

Cjkkkk:
Hi @derdrdirk, happy new year. Some updates here: I tried what you suggested and everything works except 2 issues:
* move_copy_to_user pass will move copy after convert which makes convert not fusible, conv(convert(copy), convert(copy)) =>conv(copy(convert), copy(convert)). I added new checks to not move copy after convert if convert is consumed by conv. 
 * algebra_simplifier will move copy before reduce which makes amax pattern fail to fuse, copy(reduce(conv)) => reduce(copy(conv)).  still thinking of a WAR for this.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -1903,6 +1904,16 @@ absl::Status GpuCompiler::OptimizeHloPostLayoutAssignment(
...
                                                        gpu_version);
 
+  // Rewrite convs into conv fusions.
+  if (!debug_options.xla_gpu_experimental_disable_binary_libraries() &&
+      debug_options.xla_gpu_experimental_enable_conv_fusion()) {
+    se::dnn::VersionInfo dnn_version = gpu_target_config.dnn_version_info;
+    if (stream_exec != nullptr) {
+      TF_ASSIGN_OR_RETURN(dnn_version, GetDnnVersionInfo(stream_exec));
+    }
+    pipeline.AddPass<ConvFusionRewriter>(gpu_version, dnn_version);
```
derdrdirk: Can we move the ConvFusionRewriter to the [FusionPipeline](https://github.com/openxla/xla/blob/5a0da8d3a33305415d85a3ed01cfbea14e98f6df/xla/service/gpu/fusion_pipeline.cc)? It could be run before the other passes like priority fusion to greedily fuse convolutions first.

xla/service/gpu/transforms/conv_kind_assignment.cc
```diff
@@ -0,0 +1,595 @@
...
+
+  std::vector<HloInstruction*> new_operands = {lhs, rhs};
+  auto new_conv = conv->CloneWithNewOperands(output_shape, new_operands);
+  new_conv->set_feature_group_count(num_groups);
+  new_conv->set_batch_group_count(1);
+  new_conv->set_convolution_dimension_numbers(dim_numbers);
+  return computation->AddInstruction(std::move(new_conv));
+}
+
+static absl::StatusOr<HloInstruction*> AssignConvKindHelper(
```
derdrdirk: Nit: This could be renamed to AssignConvKind.

xla/service/gpu/transforms/conv_kind_assignment.cc
derdrdirk, line 1: Can we add some unit test to this new pass?

xla/service/gpu/conv_utils.h
derdrdirk, line 1: We should add some unit tests to cover conv_utils.

derdrdirk (changes requested):
In general LGTM. Added some more comments.

derdrdirk (approved):
LGTM :)

xla/xla.proto
```diff
@@ -660,6 +660,9 @@ message DebugOptions {
   // Experimentally disables binary libraries in GPU compiler passes.
   optional bool xla_gpu_experimental_disable_binary_libraries = 329;
 
+  // Experimentally enable rewriting conv as fusion in GPU compiler passes.
```
thomasjoerg: Please keep this file sorted!

thomasjoerg (changes requested):

xla/xla.proto
Cjkkkk: Sorted.

thomasjoerg (approved):

Cjkkkk:
Converted back to draft since there are other smaller PRs ongoing. Reopen if needed.

PR #34110, opened 2025-11-18
[pjrt] Add cuda_version attribute to GPU plugin

gspschmid:
📝 Summary of Changes
Adds a C API plugin attribute `cuda_version` that's set to the cuda runtime version the plugin was compiled against. The verison may be queried from JAX, e.g.,
```bash
$ python3 -c "import jax.extend; print(jax.extend.backend.get_backend().cuda_version)"
13000
```

🎯 Justification
This allows CUDA-version-specific packages to detect the CUDA version present in JAX. See also https://github.com/jax-ml/jax/issues/32729

🚀 Kind of Contribution
New Feature

N/A

🧪 Unit Tests:
None

🧪 Execution Tests:
None

cc @skye

### DIFF
--- a/xla/pjrt/c/BUILD
+++ b/xla/pjrt/c/BUILD
@@ -427,7 +427,10 @@ cc_library(
     name = "pjrt_c_api_gpu_internal",
     srcs = ["pjrt_c_api_gpu_internal.cc"],
     hdrs = ["pjrt_c_api_gpu_internal.h"],
-    local_defines = if_rocm_is_configured(["TENSORFLOW_USE_ROCM=1"]),
+    local_defines = (
+        if_cuda_is_configured(["GOOGLE_CUDA=1"]) +
+        if_rocm_is_configured(["TENSORFLOW_USE_ROCM=1"])
+    ),
     visibility = ["//visibility:public"],
     deps = [
         ":pjrt_c_api_custom_partitioner_extension_hdrs",
@@ -472,7 +475,9 @@ cc_library(
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/strings:str_format",
-    ],
+    ]+ if_cuda_is_configured([
+        "@local_config_cuda//cuda:cuda_headers",
+    ]),
 )
 
 cc_library(
--- a/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
+++ b/xla/pjrt/c/pjrt_c_api_gpu_internal.cc
@@ -59,6 +59,10 @@ limitations under the License.
 #include "xla/service/compiler.h"
 #include "xla/service/custom_call_target_registry.h"
 
+#if GOOGLE_CUDA
+#include "third_party/gpus/cuda/include/cuda_runtime_api.h"
+#endif  // GOOGLE_CUDA
+
 namespace pjrt {
 namespace gpu_plugin {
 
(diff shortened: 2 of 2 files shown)


gspschmid:
@skye I believe this is the shortest path to exposing the CUDA version as an attribute. I don't love the addition of platform-specific code in xla/pjrt/c/pjrt_c_api_gpu_internal.cc, but given the current setup of the C API I don't see a neat alternative. Some elaboration below on what might be a cleaner solution.

---

At the moment there are some confusing differences between the plugin path and the monolithic (Google-internal) path: The attributes exposed to Python are drawn from `pjrt_client->plugin_attributes()` (https://github.com/openxla/xla/blob/dab69f82aac191376ee2e9f340fb09d43d3e6bdb/xla/python/pjrt_ifrt/pjrt_client.cc#L108-L115).

**Without plugins** that call (presumably) goes directly to the implementation in, say, `StreamExecutorGpuClient::plugin_attributes()` (https://github.com/openxla/xla/blob/dab69f82aac191376ee2e9f340fb09d43d3e6bdb/xla/pjrt/gpu/se_gpu_pjrt_client.cc#L271-L278).

**With plugins** that call would go to the C API wrapper (https://github.com/openxla/xla/blob/7ed824646630c95dd4552f5f54e9a9f98e527f44/xla/pjrt/c_api_client/pjrt_c_api_client.cc#L335-L339). `PjRtCApiClient::plugin_attributes()` is populated by assigning some attributes ad-hoc and adding them to ones coming out of the C API call `PJRT_Plugin_Attributes` (https://github.com/openxla/xla/blob/dab69f82aac191376ee2e9f340fb09d43d3e6bdb/xla/pjrt/c_api_client/pjrt_c_api_client.cc#L281-L295).

But, at least for the GPU backend, `PJRT_Plugin_Attributes` is only configured to provide XLA "generic" attributes (https://github.com/openxla/xla/blob/dab69f82aac191376ee2e9f340fb09d43d3e6bdb/xla/pjrt/c/pjrt_c_api_gpu_internal.cc#L473) and never takes `StreamExecutorGpuClient::plugin_attributes()` into account:
https://github.com/openxla/xla/blob/dab69f82aac191376ee2e9f340fb09d43d3e6bdb/xla/pjrt/gpu/se_gpu_pjrt_client.cc#L271-L278

---

The current `PJRT_Plugin_Attributes` C API unfortunately doesn't pass the instantiated PjrtClient in, so we can't call `plugin_attributes()` from within it. Presumably that's also the reason the `supports_cross_host_transfers` and `serialize_with_sdy` attributes are set both in the C API wrapper and the pjrt_ifrt code rather than some client-specific code.

All that is to say, a better long term solution might be to move away from `PJRT_Plugin_Attributes` and add a new C API `PJRT_Client_Attributes` that is allowed to call `client->plugin_attributes()` under the hood.

xla/pjrt/c/BUILD
```diff
@@ -466,7 +469,10 @@ cc_library(
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/strings:str_format",
-    ],
+    ]+ if_cuda_is_configured([
+        # keep sorted
```
skye: remove boilerplate comment?

skye (approved):
Very nice. I'm fine with this fix, although agree PJRT_Client_Attributes would make sense and be cleaner.

skye:
I'm not an XLA maintainer, can someone trigger the workflows and/or merge?

hawkinsp (approved):

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
```diff
@@ -322,6 +326,41 @@ PJRT_Error* PJRT_GpuDeviceTopology_Create(
...
+PJRT_Error* PJRT_Plugin_Attributes_Gpu(PJRT_Plugin_Attributes_Args* args) {
+  PJRT_RETURN_IF_ERROR(ActualStructSizeIsGreaterOrEqual(
+      "PJRT_Plugin_Attributes_Args", PJRT_Plugin_Attributes_Args_STRUCT_SIZE,
+      args->struct_size));
+  const std::vector<PJRT_NamedValue>& base_attributes =
+      pjrt::GetXlaPluginCAttributes();
+
+#if GOOGLE_CUDA && defined(CUDART_VERSION)  // cuda
+  // Attributes have the lifetime of the process, so we leak the array here.
+  std::vector<PJRT_NamedValue>* attributes = new std::vector<PJRT_NamedValue>();
```
skye, line 333: Google has an internal presubmit that doesn't like this leak. Can you use a static variable like this? https://github.com/openxla/xla/blob/93e006ef34ca95731b1e3b28969185b2c7077297/xla/pjrt/c/pjrt_c_api_helpers.cc#L660

This is technically against the Google C++ style guide: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables But we already did it once so I think it's ok

gspschmid: Note that our `static c_values`'s initializer isn't entirely trivial -- we copy from `GetXlaPluginCAttributes()` and add the `"cuda_version"` attribute. So that code should likely go into a helper function like `MakeCudaPluginCAttributes`, and called as the initializer for the static local variable: `static const std::vector<PJRT_NamedValue>* c_values = MakeCudaPluginCAttributes()`. I fear the `new` inside that helper would still violate your presubmit lint?

gspschmid: @skye I made the above change and rebased. Let me know if you still run into the presubmit error. If it is unhappy with the dynamic allocation inside the helper I suppose we could wrap that code in the constructor of an auxiliary struct and instead do something like `static const CudaPluginCAttributes* c_values = new CudaPluginCAttributes()`?

skye: Let's try it

skye (approved):

xla/pjrt/c/pjrt_c_api_gpu_internal.cc
skye: That apparently thwarted the leak checker. However, now `//xla/pjrt/c:pjrt_c_api_gpu_internal` is failing to build internally:
```
xla/pjrt/c/pjrt_c_api_gpu_internal.cc:346:27: error: use of undeclared identifier 'CUDART_VERSION'
  346 |     c_value.int64_value = CUDART_VERSION;
      |                           ^~~~~~~~~~~~~~
1 error generated.
```
Taking a look, but I'm mostly out of office today so feel free to take over if you don't hear more from me.

skye:
Ok yeah you just need to update the header guard: https://github.com/tensorflow/tensorflow/pull/105869/files I can get that in directly but probably not until tomorrow, so if you just wanna make that or similar change here that might be easiest. I think we're almost there!!

gspschmid:
Oh right, I didn't make `MakeCudaPluginCAttributes()` conditional on `GOOGLE_CUDA`! (I guess that's what you wanted to link to?) Let's see if third time's the charm :-)

skye (approved):

PR #34143, opened 2025-11-19
[GPU] Add all-to-all support to S-curve model

terryysun:
📝 Summary of Changes
Added all-to-all support to S-curve model.

🎯 Justification
S-curve model doesn't support all-to-all, fallback may lead to bad performance, benchmarking justified that the added all-to-all model can improve performance for models with cross-NVL domain all-to-all.

🚀 Kind of Contribution
⚡️ Performance Improvement/✨ New Feature

| Branch | End-to-end execution time mean on mixtral_8x7b_bf16_2x8 |
| :------- | :------: |
| main     | 1128328 us   |
| terryysun/a2a_s_curve (this branch)   | 1009397 us |

Speedup over main: 11.78%.

🧪 Unit Tests:
Added exact-matching unit tests to guard the estimation value.

🧪 Execution Tests:
Added execution tests to guard the comm-compute overlapping behavior.


xla/service/gpu/model/sol_gpu_cost_model.h
```diff
@@ -73,6 +74,16 @@ class SolGPUCostModel {
                                              const CollectiveType& coll_type,
                                              int num_communicators) const;
 
+  // Returns the latency of an AllToAll collective across multiple nodes.
+  //
+  // `buff_size_bytes`: the size of the message to be transferred.
+  // `num_nodes`: the number of nodes participating in the all-to-all.
+  // `num_communicators`: the number of communicators participating in the
+  // all-to-all.
+  absl::StatusOr<absl::Duration> AllToAllLatency(int64_t buff_size_bytes,
```
felixwqp, line 83: To improve consistency, since we're adding AllToAllLatency, perhaps we should rename RingLatency to RingCollectiveLatency? This would clarify that both interfaces deal with  communication operation name.

xla/service/gpu/model/sol_gpu_cost_model_test.cc
```diff
@@ -27,48 +27,80 @@ limitations under the License.
...
+                              /*num_nodes=*/1, absl::Microseconds(284)},
+                             {SolGPUCostModel::CollectiveType::kReduceScatter,
+                              /*num_nodes=*/2, absl::Microseconds(485)},
+                             {SolGPUCostModel::CollectiveType::kReduceScatter,
+                              /*num_nodes=*/4, absl::Microseconds(885)},
+                             {SolGPUCostModel::CollectiveType::kSendRecv,
+                              /*num_nodes=*/1, absl::Microseconds(292)},
+                             {SolGPUCostModel::CollectiveType::kSendRecv,
+                              /*num_nodes=*/2, absl::Microseconds(485)},
+                             {SolGPUCostModel::CollectiveType::kAllToAll,
```
felixwqp, line 97: can you double check to make sure the single-host(intra-host) a2a is using the perf-table and multi-host(cross-partition) is using the new model you are adding? 



I ask because  function [CommunicationType](https://github.com/openxla/xla/blob/ad4c7233420aef4237c5ee2bdeda2872d7224aa3/xla/service/gpu/model/sol_latency_estimator.cc#L325) is used to determine the type of alltoall; and which model to use to estimate this alltoall.


we may need to make sure the alltoall is properly classified in `CommunicationType`

with that said, can you add another unit test into xla/service/gpu/model/sol_latency_estimator_test.cc? we need to ensure the alltoall is properly classified by different underlying model(s-curve or perf-table)

felixwqp (commented):
thank you Terry!

xla/service/gpu/model/sol_gpu_cost_model.h
terryysun: I think ring is more of a pattern name. I haven't found a good pattern name for a2a so using collective name here -- it's one of it kind so there should not be much ambiguity.

xla/service/gpu/model/sol_gpu_cost_model_test.cc
terryysun: added model dispatching test.

xla/service/gpu/model/sol_latency_estimator_test.cc
```diff
@@ -536,6 +534,55 @@ INSTANTIATE_TEST_SUITE_P(SolLatencyEstimatorTests, SolLatencyEstimatorTest,
...
+  EXPECT_FALSE(SolLatencyEstimator::ComputeCollectiveTime(
+                   *nvl_instr, gpu_info, shape_size_fn, sol_flags, &mlir_ctx,
+                   /*collective_interpolator=*/nullptr)
+                   .ok());
+  EXPECT_TRUE(SolLatencyEstimator::ComputeCollectiveTime(
+                  *nvl_instr, gpu_info, shape_size_fn, sol_flags, &mlir_ctx,
+                  interpolator.get())
+                  .ok());
+
+  // IB collective should use S-curve model (world-level across 2 hosts).
```
felixwqp: can you rephrase it to "cross-partition" collective? 

IB may not apply to all Xla user's transport.

terryysun: Rephased.

felixwqp (approved):

PR #34196, opened 2025-11-20
[NVIDIA GPU] Fix a deadlock when doing comm split

Tixxx:
📝 Summary of Changes
If participant group of a collective is empty, doing split can sometimes cause deadlocks since we dont know what are the participating ranks. Instead it should proceed to a regular comm init.

🎯 Justification
A target case would be if there are multiple communicators cached from different modules that use different device groups.
We rely one a rendezvous to return early if a communicator has been cached. Since participant groups are part of the key, there will be inconsistent cache hits among ranks leading to some ranks go to split but other won't.  ncclCommSplit requires all ranks to be call the api otherwise it will hang.
🚀 Kind of Contribution

NA
🧪 Unit Tests:
can only be tested with execution test

🧪 Execution Tests:
added

### DIFF
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -494,9 +494,13 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
 
   // Synchronize the device to make sure no other collectives are
   // running before we do the split.
-  if (!device->SynchronizeAllActivity()) {
-    return Internal(
-        "Failed to synchronize GPU before splitting communicators.");
+  {
+    tsl::profiler::TraceMe trace("SynchronizeAllActivityBeforeSplit");
+    if (!device->SynchronizeAllActivity()) {
+      return Internal(
+          "Failed to synchronize GPU before splitting communicators.");
+    }
+    VLOG(3) << "Synchronized device before splitting";
   }
 
   // Current approach for communicator splitting works because of XLAs SPMD
@@ -732,6 +736,9 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
         return InitializeGpuClique(collectives, device, run_id, clique_key,
                                    acquired_clique, num_local_participants,
                                    rank, config);
+      } else if (clique_key.ParticipantGroups().empty()) {
+        LOG(WARNING) << "Found empty participant groups."
+                     << " Skip splitting communicators.";
       }
     }
   }
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -492,6 +492,13 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
   GpuCollectives::DeviceRank device_rank = {&gpu_device, rank};
   RankPair rank_pair = {parent_rank, device_rank};
 
+  // Synchronize the device to make sure no other collectives are
+  // running before we do the split.
+  if (!device->SynchronizeAllActivity()) {
+    return Internal(
+        "Failed to synchronize GPU before splitting communicators.");
+  }
+
   // Current approach for communicator splitting works because of XLAs SPMD
   // programming model where all collective operations have replica groups that
   // include all ranks. This property guarantees that we'll split each
@@ -718,7 +725,10 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
 
   if (enable_nccl_comm_splitting) {
     for (auto& [acquired_clique_key, acquired_clique] : acquired_cliques) {
-      if (clique_key.IsSubsetOf(acquired_clique_key)) {
+      // If the participant group is empty, we won't know if there are other
+      // ranks involved in the split. Proceed to normal initialization.
+      if (clique_key.IsSubsetOf(acquired_clique_key) &&
+          !clique_key.ParticipantGroups().empty()) {
         return InitializeGpuClique(collectives, device, run_id, clique_key,
                                    acquired_clique, num_local_participants,
                                    rank, config);
--- a/xla/tests/collective_ops_e2e_test.cc
+++ b/xla/tests/collective_ops_e2e_test.cc
@@ -2899,10 +2899,10 @@ TEST_F(CollectiveOpsTestE2E, MultipleModuleDifferentDeviceGroupsShouldRun) {
 
   const int64_t kNumReplicas_1 = 2;
   const int64_t kNumReplicas_2 = 4;
-
-  ASSERT_GE(hlo_runner_->device_count(), kNumReplicas_2)
-      << "Test requires at least " << kNumReplicas_2 << " devices ("
-      << hlo_runner_->device_count() << " available)";
+  if (hlo_runner_->device_count() < kNumReplicas_2) {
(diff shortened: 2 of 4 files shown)


xla/tests/collective_ops_e2e_test.cc
```diff
@@ -4833,5 +4833,104 @@ TEST_F(CollectiveMetadataTest, ConstructCollectiveMetadataWithReplicaGroup) {
...
+      GetModuleConfigForTest(/*replica_count=*/kNumReplicas_2);
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module_1,
+                          ParseAndReturnVerifiedModule(kModuleStr_1, config_1));
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module_2,
+                          ParseAndReturnVerifiedModule(kModuleStr_2, config_2));
+
+  int64_t num_elements_1 = module_1->entry_computation()
+                               ->root_instruction()
```
felixwqp: Looks like you are using the root of the module to init the size of input parameters.  Is this intended? 

for this case above the shape between input params and output root are consistent, so it's working well functionality. 

but it looks a bit counter-intuitive. Should we capture the num_elements from the input parameter of each module?

Tixxx: good catch, yea i changed it to inspect the input param's count

Tixxx:
hi @felixwqp  could you take another look at this? thanks

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -492,6 +492,13 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
   GpuCollectives::DeviceRank device_rank = {&gpu_device, rank};
   RankPair rank_pair = {parent_rank, device_rank};
 
+  // Synchronize the device to make sure no other collectives are
+  // running before we do the split.
+  if (!device->SynchronizeAllActivity()) {
```
felixwqp: as there is an additional synchronization during the initialization process, can you run the suite of benchmark to make sure this has not performance implication? 

thank you!

felixwqp: @ezhulenev  Hi Eugene, could you also share some insights about this runtime side change? 

thank you!

Tixxx: ok i can run this on a few benchmarks to confirm but this is done during initialization so it wont have much impact on thunk execution.

ezhulenev: It should be fine, it's done once per process lifetime. I'd just add TraceMe here to see it in profiles + maybe a VLOG.

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -718,7 +725,10 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
 
   if (enable_nccl_comm_splitting) {
     for (auto& [acquired_clique_key, acquired_clique] : acquired_cliques) {
-      if (clique_key.IsSubsetOf(acquired_clique_key)) {
+      // If the participant group is empty, we won't know if there are other
```
ezhulenev, line 732: Can you also log (WARN maybe?) that comm splitting was ignored because "reasons".

ezhulenev:
LGTM. Few comments about improving observability and logging a little bit.

xla/backends/gpu/collectives/gpu_cliques.cc
Tixxx: thanks! added trace and vlog

xla/backends/gpu/collectives/gpu_cliques.cc
Tixxx: added

ezhulenev (approved):

felixwqp (approved):

Tixxx:
@felixwqp  looks like it says some internal checks failed, could you check what the failures were? Thanks

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -2863,5 +2863,96 @@ INSTANTIATE_TEST_SUITE_P(
...
+    all-reduce.2 = f32[65536] all-reduce(all-reduce.1), to_apply=apply_op, replica_groups={{0,1}, {2,3}}
+    all-reduce.3 = f32[65536] all-reduce(all-reduce.2), to_apply=apply_op, replica_groups={{0,1}, {2,3}}
+    ROOT all-reduce.4 = f32[65536] all-reduce(all-reduce.3), to_apply=apply_op, replica_groups={{0,1,2,3}}
+  }
+  )";
+
+  const int64_t kNumReplicas_1 = 2;
+  const int64_t kNumReplicas_2 = 4;
+
+  ASSERT_GE(hlo_runner_->device_count(), kNumReplicas_2)
```
felixwqp: Internal test failed because some of the internal CI doesn't have enough GPU, can you skip this test if the GPU is not sufficient? 

more specifically,  instead of "ASSERT_GE", 

can you use something like 

```
  if (hlo_runner_->device_count() < kNumReplicas) {
    GTEST_SKIP() << "Test requires at least " << kNumReplicas << " devices ("
                 << hlo_runner_->device_count() << " available)";
  }
```
to skip the test, if it requires more than 4 GPUs to run.

Tixxx: ok i see, i changed it to skip if less than 4 devices. Can you give it another try?

derdrdirk:
Would it be possible to rewrite the test `MultipleModuleDifferentDeviceGroupsShouldRun` to only use 2 GPUs (instead of 4)?

Tixxx:
> Would it be possible to rewrite the test `MultipleModuleDifferentDeviceGroupsShouldRun` to only use 2 GPUs (instead of 4)?

hi @derdrdirk  this is only reliably reproducible when running modules with different number of GPUs in a sequence, 2 and 4 are the smallest combination of number of gpus in this case.
What's the issue with 4 gpus? there are many tests in this file that use 4 gpus.

xla/tests/collective_ops_e2e_test.cc
```diff
@@ -2863,5 +2863,96 @@ INSTANTIATE_TEST_SUITE_P(
...
+    y = f32[] parameter(1)
+    ROOT apply_op = f32[] add(x, y)
+  }
+
+  ENTRY test_computation {
+    param_0 = f32[65536] parameter(0)
+    all-reduce.1 = f32[65536] all-reduce(param_0), to_apply=apply_op, replica_groups={{0,1}, {2,3}}
+    all-reduce.2 = f32[65536] all-reduce(all-reduce.1), to_apply=apply_op, replica_groups={{0,1}, {2,3}}
+    all-reduce.3 = f32[65536] all-reduce(all-reduce.2), to_apply=apply_op, replica_groups={{0,1}, {2,3}}
+    ROOT all-reduce.4 = f32[65536] all-reduce(all-reduce.3), to_apply=apply_op, replica_groups={{0,1,2,3}}
```
PatriosTheGreat: Would smaller shapes cover this test-case? f32[8] for example.
Initializing and checking parameters for big shapes increases test time quite dramatically.

PatriosTheGreat (changes requested):

xla/tests/collective_ops_e2e_test.cc
Tixxx: sure, updated the size to 8

derdrdirk (approved):
LGTM

PatriosTheGreat (approved):

PR #34228, opened 2025-11-20
Refactor the heuristic for max unroll factor.

dimvar:
A recent change introduced max unroll factor = 8 on Blackwell under certain conditions. But the function MaxUnrollFactor is still hardcoded to return 4, which can be confusing to the reader. This patch moves the heuristic into MaxUnrollFactor. There is no functional change.


dimvar:
CC @akuegel Can you take a look at this PR?

akuegel (approved):

PR #34320, opened 2025-11-24
Migrate from native built-ins to Starlark rule definitions

cerisier:
📝 Summary of Changes

- Replaces all occurrences of native.rule by their starlark version.
- Add missing load statements where native rules were implicit.

🎯 Justification

This is a preparation PR for Bazel 9 where implicits must be loaded from starlark.

🚀 Kind of Contribution


cerisier:
<s>abseil-cpp seems not affected by the changes in this PR, however the CI fails ?</s>

cerisier:
@ddunl any luck ? Bazel 9 already in testing for us. Cheers

MODULE.bazel
```diff
@@ -21,6 +21,7 @@ bazel_dep(name = "pybind11_bazel", version = "2.13.6")
 bazel_dep(name = "pybind11_protobuf", version = "0.0.0-20250210-f02a2b7")
 bazel_dep(name = "re2", version = "2024-07-02.bcr.1", repo_name = "com_googlesource_code_re2")
 bazel_dep(name = "rules_cc", version = "0.2.0")
+bazel_dep(name = "rules_java", version = "8.16.1")
```
ddunl, line 24: Is this needed? otherwise looks good I think

cerisier: Thanks for the review.
Yes because of the java_runtime a bit later in the CL.
I added the explicit load where it's used.

ddunl: Ah, sorry about that. I was searching for native.java_runtime but it wouldn't work that way in the `BUILD` file. Sorry about that! LGTM.

ddunl (approved):
This one will probably not be able to automatically merge because of weirdness of the tools/toolchains dir in the monorepo, so I will have to do some extra patching on our end before it will land.

ddunl:
Whoops, this did automerge, but just without the `tools/toolchains` bits. I'll add another commit that corrects that (it will still be attributed to you on GitHub!). Thanks again for the PR :)

PR #34467, opened 2025-11-26
[ROCm] Fix ExtractThreadDims for AMD targets

mfrancepillois:
📝 Summary of Changes
AMD/ROCm Triton backend does not support warp specialization. `ThreadDims` are therefore calculated from module attributes and not retrieved from `nvvm.reqntid`.

🎯 Justification
As warp specialization is not currently supported by the AMD/ROCm Triton backend, this backend ignores the `nvvm.reqntid` attribute. Therefore, this attribute does not contain a correct value, as the currently Triton implementation assumes the number of threads per warp is always 32, which is not the case for some AMD targets (see https://github.com/triton-lang/triton/blob/49e174c6856aed1d36b85fb2b398ffaa32a80aa8/lib/Conversion/TritonGPUToLLVM/FuncOpToLLVM.cpp#L204C53-L204C68).
Consequently, the `ExtractThreadDims` has been adapted to calculate `ThreadDims` only based on attributes used and updated by the AMD triton backend.

🚀 Kind of Contribution

Not relevant

🧪 Unit Tests:
Fixes failures of type:
```
xla/backends/gpu/codegen/triton/fusion_emitter_device_test.cc:4234: Failure
Value of: RunAndCompareNoHloPasses(kHloText, ErrorSpecForDotAlgorithm(algorithm))
 Actual: false (INTERNAL: Expected total threads as per reqntid attribute to be 32 but got 64 as per ttg.total-num-warps and tt.threads-per-warp attributes.)
Expected: true
```
for Triton Tests when targeting AMD GPUs.

🧪 Execution Tests:
Not relevant

### DIFF
--- a/xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
+++ b/xla/backends/gpu/codegen/triton/compilation_pipeline_rocm.cc
@@ -123,10 +123,6 @@ static void MakeLLIR(mlir::OpPassManager* pm,
                      const stream_executor::RocmComputeCapability& rocm_cc,
                      int num_stages) {
   const int custom_lds_size = 0;
-  // The `createTritonGPUAllocateWarpGroups` pass is not implemented in the
-  // upstream Triton, but is necessary for `ExtractThreadDims` in emitter
-  // helpers. It adds the `ttg.total-num-warps` attribute.
-  pm->addPass(mt::gpu::createTritonGPUAllocateWarpGroups());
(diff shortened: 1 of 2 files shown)


mfrancepillois:
@zoranjovanovic-ns could you please take a look at this PR when you have time?

zoranjovanovic-ns (approved):

i-chaochen (approved):

akuegel (approved):

PR #34493, opened 2025-11-26
[GPU] NCCL all-to-all integration

terryysun:
📝 Summary of Changes
Integrate NCCL all-to-all API to XLA.

🎯 Justification
NCCL supports all-to-all natively since 2.28U2, leveraging this API can simplify corresponding logics in XLA.

🚀 Kind of Contribution
✨ New Feature

N/A

🧪 Unit Tests:
Existing all-to-all tests guarding identical behavior.

🧪 Execution Tests:
Added execution tests.

### DIFF
--- a/xla/backends/gpu/collectives/nccl_communicator.cc
+++ b/xla/backends/gpu/collectives/nccl_communicator.cc
@@ -683,10 +683,10 @@ absl::Status NcclCommunicator::LaunchAllToAll(
 
   TF_ASSIGN_OR_RETURN(ncclDataType_t nccl_dtype, ToNcclDataType(dtype, false));
 
+#if NCCL_VERSION_CODE >= 22800  // ncclAlltoAll requires NCCL 2.28+
   // Try using native ncclAlltoAll if available.
   // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#ncclalltoall
   bool can_use_native_alltoall = true;
-
   size_t element_size = primitive_util::ByteWidth(dtype);
 
   // Non-contiguous buffers are not supported by native ncclAlltoAll.
@@ -713,8 +713,7 @@ absl::Status NcclCommunicator::LaunchAllToAll(
   }
 
   if (can_use_native_alltoall) {
-#if NCCL_VERSION_CODE >= 22800  // ncclAlltoAll requires NCCL 2.28+
-    LOG(INFO) << absl::StreamFormat(
+    VLOG(1) << absl::StreamFormat(
         "[%d] NCCL AllToAll: Using native ncclAlltoAll API",
         stream->parent()->device_ordinal());
     TF_RETURN_IF_ERROR(XLA_NCCL_STATUS(
@@ -725,13 +724,13 @@ absl::Status NcclCommunicator::LaunchAllToAll(
       TF_RETURN_IF_ERROR(PollUntilDone());
     }
     return absl::OkStatus();
-#else
-    VLOG(1) << absl::StreamFormat(
-        "[%d] NCCL AllToAll: ncclAlltoAll requires NCCL 2.28+, "
-        "falling back to Send/Recv (current NCCL version: %d)",
-        stream->parent()->device_ordinal(), NCCL_VERSION_CODE);
-#endif
   }
+#else
+  VLOG(1) << absl::StreamFormat(
+      "[%d] NCCL AllToAll: ncclAlltoAll requires NCCL 2.28+, "
+      "falling back to Send/Recv (current NCCL version: %d)",
+      stream->parent()->device_ordinal(), NCCL_VERSION_CODE);
+#endif
 
   // Fall back to Send/Recv for in-place, non-contiguous, or old NCCL version.
   VLOG(2) << absl::StreamFormat("[%d] NCCL AllToAll: Using Send/Recv fallback",
--- a/xla/backends/gpu/collectives/nccl_communicator.cc
+++ b/xla/backends/gpu/collectives/nccl_communicator.cc
@@ -26,7 +26,6 @@ limitations under the License.
 
 #include "absl/container/inlined_vector.h"
 #include "absl/functional/any_invocable.h"
-#include "xla/debug_options_flags.h"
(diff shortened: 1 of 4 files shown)


xla/backends/gpu/collectives/nccl_communicator.cc
```diff
@@ -683,6 +683,59 @@ absl::Status NcclCommunicator::LaunchAllToAll(
...
+      if (send_buffers[i].opaque() == recv_buffers[i].opaque()) {
+        can_use_native_alltoall = false;
+        break;
+      }
+    }
+  }
+
+  if (can_use_native_alltoall) {
+#if NCCL_VERSION_CODE >= 22800  // ncclAlltoAll requires NCCL 2.28+
+    LOG(INFO) << absl::StreamFormat(
```
olegshyshkov: VLOG

xla/backends/gpu/collectives/nccl_communicator.cc
```diff
@@ -683,6 +683,59 @@ absl::Status NcclCommunicator::LaunchAllToAll(
 
   TF_ASSIGN_OR_RETURN(ncclDataType_t nccl_dtype, ToNcclDataType(dtype, false));
 
+  // Try using native ncclAlltoAll if available.
+  // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#ncclalltoall
+  bool can_use_native_alltoall = true;
+
+  size_t element_size = primitive_util::ByteWidth(dtype);
+
+  // Non-contiguous buffers are not supported by native ncclAlltoAll.
```
olegshyshkov, line 692: So `ncclAllToAll` can only be used when we have `all-to-all` HLO with split_dimension. When we have `all-to-all` with multiple input and tuple output, it's highly unlikely (not impossible though) that we'll get contiguous buffers. If we only can support `split_dimension`, then the check below sounds very unnecessary. We, ourselves, slices buffers into multiple chunks in `AllToAllThunk` and now checking if those slices are contiguous and reconstructing input back? How about we add another `AllToAll` method to `NcclCommunicator` that is closer to `ncclAlltoAll`.

xla/backends/gpu/collectives/nccl_communicator.cc
```diff
@@ -683,6 +683,59 @@ absl::Status NcclCommunicator::LaunchAllToAll(
...
+    for (size_t i = 0; i < send_buffers.size(); ++i) {
+      if (send_buffers[i].opaque() == recv_buffers[i].opaque()) {
+        can_use_native_alltoall = false;
+        break;
+      }
+    }
+  }
+
+  if (can_use_native_alltoall) {
+#if NCCL_VERSION_CODE >= 22800  // ncclAlltoAll requires NCCL 2.28+
```
olegshyshkov: How long would we need to maintain this `#if`? Do we need this just because Google internally didn't integrate NCCL 2.28+ yet or are there other clients that would build NCCL with another version?

@PatriosTheGreat

olegshyshkov (changes requested):

xla/backends/gpu/collectives/nccl_communicator.cc
PatriosTheGreat: This usually stays for a long time, since external users are allowed to use XLA with non-default NCCL package.

xla/backends/gpu/collectives/nccl_communicator.cc
```diff
@@ -683,6 +683,59 @@ absl::Status NcclCommunicator::LaunchAllToAll(
...
   TF_ASSIGN_OR_RETURN(ncclDataType_t nccl_dtype, ToNcclDataType(dtype, false));
 
+  // Try using native ncclAlltoAll if available.
+  // https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#ncclalltoall
+  bool can_use_native_alltoall = true;
+
+  size_t element_size = primitive_util::ByteWidth(dtype);
+
+  // Non-contiguous buffers are not supported by native ncclAlltoAll.
+  for (size_t i = 1; i < send_buffers.size(); ++i) {
```
PatriosTheGreat, line 693: Can you move these checks under ifdef NCCL_VERSION_CODE >= 22800? I assume we don't need to do this if the NCCL version doesn't support this.

PatriosTheGreat (changes requested):

xla/backends/gpu/collectives/nccl_communicator.cc
terryysun: Agreed that would make this method cleaner, but I don't think it worth a separate interface for avoiding this check. The constraints of NCCL native API may also change over time, having a centralized place can make future updates easier.

xla/backends/gpu/collectives/nccl_communicator.cc
terryysun: updated.

xla/backends/gpu/collectives/nccl_communicator.cc
terryysun: updated.

xla/backends/gpu/collectives/nccl_communicator.cc
olegshyshkov: It's hard for me to image what kind of potential future NCCL API changes would benefit from the code in this state. Do you have anything specific in mind?

Right now we end up doing extra work for no gain. We get perfectly fine, contiguous buffers in [0] and slice them into multiple buffers to use Send/Recv. Then here, we realize that we can use the new NCCL API and try to reconstruct if original buffer was contiguous or not. It would be much cleaner if nccl_communicator had two versions of AllToAll. If we mode slicing logic here, contiguous-AllToAll can call non-contiguous-AllToAll when necessary.

[0] https://github.com/openxla/xla/blob/adc132f8b7b38e1f3b146567f2facddc8ca43d98/xla/backends/gpu/runtime/all_to_all_thunk.cc#L315

olegshyshkov:
A PR with similar functionality landed in https://github.com/openxla/xla/pull/35463. Should we close this PR?

PR #34499, opened 2025-11-27
Fix cublasLT bug when swap operands

shawnwang18:
🎯 Justification
This pull request introduces support for FP8 matmul operations with swapped operands in the GPU autotuner and ensures correct handling of scaling factors when operands are swapped. The main changes include updates to the autotuner test suite, the CUDA BLAS Lt matmul implementation, and build dependencies.

🚀 Kind of Contribution
🐛 Bug Fix,

🧪 Unit Tests:
CublasLtBackendTest.CompileFp8SwapOperands

### DIFF
--- a/xla/backends/gpu/autotuner/cublaslt_test.cc
+++ b/xla/backends/gpu/autotuner/cublaslt_test.cc
@@ -30,6 +30,7 @@ limitations under the License.
 #include "xla/hlo/testlib/filecheck.h"
 #include "xla/hlo/testlib/hlo_hardware_independent_test_base.h"
 #include "xla/service/compiler.h"
+#include "xla/service/executable.h"
 #include "xla/service/gpu/nvptx_compiler.h"
 #include "xla/service/platform_util.h"
 #include "xla/stream_executor/blas.h"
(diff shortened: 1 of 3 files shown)


xla/backends/gpu/autotuner/cublaslt_test.cc
```diff
@@ -196,5 +197,46 @@ TEST_F(CublasLtBackendTest, ApplyConfig) {
               absl_testing::IsOkAndHolds(true));
 }
 
+TEST_F(CublasLtBackendTest, CompileFp8SwapOperands) {
+  if (!stream_executor_->GetDeviceDescription()
+           .cuda_compute_capability()
+           .IsAtLeast(8, 9)) {
+    GTEST_SKIP() << "FP8 requires compute capability 8.9 or higher";
+  }
+  const char kFp8MatmulWithSwapHlo[] = R"(
```
derdrdirk, line 210: Nit: Would be good to add a comment why this Hlo contains a swap. Is it because lhs, rhs and ROOT are row-major, i.e. they have {1,0} layout, while cublas preferes column-major ({0,1}) layout?

xla/backends/gpu/autotuner/cublaslt_test.cc
```diff
@@ -196,5 +197,46 @@ TEST_F(CublasLtBackendTest, ApplyConfig) {
               absl_testing::IsOkAndHolds(true));
 }
 
+TEST_F(CublasLtBackendTest, CompileFp8SwapOperands) {
```
derdrdirk, line 200: The test ideally should be close to cuda_blas_lt.cc. Is there a way to create a test file cuda_blas_lt_test.cc that checks for the same behavior?

derdrdirk (changes requested):
Overall LGTM. Added two comments.

xla/backends/gpu/autotuner/cublaslt_test.cc
shawnwang18: Yes, I added some comments on why it needs a operand swap.

xla/backends/gpu/autotuner/cublaslt_test.cc
shawnwang18: Fixed, moved the test to a separate file.

derdrdirk (approved):
LGTM. Thanks for the changes.

xla/stream_executor/cuda/BUILD
```diff
@@ -2494,3 +2494,35 @@ cuda_library(
...
+    ],
+    deps = [
+        "//xla:autotuning_proto_cc",
+        "//xla:xla_proto_cc",
+        "//xla/backends/autotuner:codegen_backend",
+        "//xla/backends/gpu/autotuner:cublaslt",
+        "//xla/hlo/ir:hlo",
+        "//xla/hlo/testlib:filecheck",
+        "//xla/hlo/testlib:hlo_hardware_independent_test_base",
+        "//xla/service:compiler",
```
thomasjoerg: Please do not add dependencies from stream_executor back to the XLA compiler, which depends on stream executor itself.

thomasjoerg (changes requested):

xla/stream_executor/cuda/BUILD
shawnwang18: Thanks for the info, I relocate the test, could you help review again?

akuegel (approved):

PR #34705, opened 2025-12-02
Enable the use of tcgen05 instructions on Thor by Triton pass

and-ivanov:
📝 Summary of Changes

This PR fixes a failure in `//xla/backends/gpu/codegen/triton:fusion_emitter_device_test_b200` on Thor devices (`sm110`). The failure occurs because the Triton pass responsible for enabling the `tcgen05` instruction (`mxf8f6f4.block_scale.scale_vec::1X`) was not previously enabled for the `sm110` architecture.

🎯 Justification for implementing in XLA vs. Triton

While this logic ideally belongs in the Triton repo, we are implementing it in XLA for the following reasons:
- Support for `sm110` was introduced in [PTX ISA 9.0](https://docs.nvidia.com/cuda/parallel-thread-execution/#changes-in-ptx-isa-version-9-0), which relies on CUDA 13.0.
- Upstream Triton [does not yet support CUDA 13.0](https://github.com/triton-lang/triton/blob/main/cmake/nvidia-toolchain-version.json), and support may not land in the immediate future.
- We want to ensure XLA is fully functional on Thor devices irrespective of the Triton status.

We prefer to upstream this patch to ensure Thor is fully supported rather than keeping it in an internal development branch.

### DIFF
--- a/third_party/triton/temporary/series.bzl
+++ b/third_party/triton/temporary/series.bzl
@@ -14,5 +14,6 @@ those to this list.
 """
 
 temporary_patch_list = [
+    "//third_party/triton:temporary/thor_support.patch",
     # Add new patches just above this line
 ]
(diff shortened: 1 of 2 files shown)


chsigg (approved):
Thanks!

vwbaker:
Did you try to upstream this to triton? I think they should be able to allow you to test with this before it's fully supported (it's not fully supported in xla either) so I'm not sure how the status between XLA & triton differ. If triton upstream refuses, then we can merge this patch.

and-ivanov:
closing: fix is implemented in Triton

PR #34715, opened 2025-12-02
[onednn] Update onednn custom call name

gaurides:
📝 Summary of Changes
Update custom_call op_name with target to easily differentiate diff onednn custom ops.

🎯 Justification
This will help debugging and viewing the ops in profile traceview. So, instead of `custom-call.2993.clone` it will show as `custom-call.2993.clone__onednn$matmul`. It will be easier to view the timeline trace.

This doesn't affect performance.


gaurides:
hi @penpornk - can you please review this small PR? Thanks!

penpornk (approved):
Done. Sorry for the delay!

PR #34789, opened 2025-12-04
[XLA:GPU] Fix cuDNN SDPA test to use 0 as workspace size to work universally on all archs

Cjkkkk:
📝 Summary of Changes
use 0 as default workspace size and query later so it works universally on all archs, cuDNN paged attention reference doesn't do this like other cuDNN sdpa tests, it fails on B200 in NV internal CI. Therefore the fix.

🎯 Justification
use 0 as default workspace size and query later so it works universally on all archs, cuDNN paged attention reference doesn't do this like other cuDNN sdpa tests, it fails on B200 in NV internal CI. Therefore the fix.

🚀 Kind of Contribution
🐛 Bug Fix

None

🧪 Unit Tests:
None

🧪 Execution Tests:
None


derdrdirk (approved):
LGTM. Thanks :)

Cjkkkk:
Hi @akuegel, could you help approve this test fix? Thanks.

PR #34853, opened 2025-12-04
[xla:emitters][oneAPI] Enable math.log1p in MathToLLVMConversionPatterns

mdfaijul:
📝 Summary of Changes
This PR fixes lowering `mlir::math::Log1pOp` on Intel GPU targets. The NVVM and ROCDL lower math.log1p directly via their GPU pattern sets, but he SPIR-V pipeline does not. For Intel GPUs we therefore enable approximate log1p option in the generic MathToLLVM conversion.

🚀 Kind of Contribution: 🐛 Bug Fix

🧪 Unit Tests:
Added the following lit test: `xla/backends/gpu/codegen/emitters/transforms/tests/lower_math_oneapi.mlir`

### DIFF
--- a/xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
+++ b/xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
@@ -123,7 +123,12 @@ class LowerToLLVMGPUPass
       return mlir::success();
     };
 
-    if (mlir::failed(LowerToLLVM(getOperation(), populate_patterns))) {
+    // NVVM and ROCDL lower math.log1p directly via their GPU pattern sets, but
+    // the SPIR-V pipeline does not. For Intel GPUs we therefore fall back to
+    // the generic MathToLLVM conversion, hence enabling approximate log1p.
+    if (mlir::failed(
+            LowerToLLVM(getOperation(), populate_patterns,
+                        /* approximateLog1p */ device_spec_.IsIntelGpu()))) {
       signalPassFailure();
     }
   }
--- a/xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
+++ b/xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
@@ -126,9 +126,9 @@ class LowerToLLVMGPUPass
     // NVVM and ROCDL lower math.log1p directly via their GPU pattern sets, but
     // the SPIR-V pipeline does not. For Intel GPUs we therefore fall back to
     // the generic MathToLLVM conversion, hence enabling approximate log1p.
+    bool lower_math_log1p = device_spec_.IsIntelGpu();
     if (mlir::failed(
-            LowerToLLVM(getOperation(), populate_patterns,
-                        /* approximateLog1p */ device_spec_.IsIntelGpu()))) {
+            LowerToLLVM(getOperation(), populate_patterns, lower_math_log1p))) {
       signalPassFailure();
     }
   }
--- a/xla/codegen/tools/emitters_opt.cc
+++ b/xla/codegen/tools/emitters_opt.cc
@@ -41,6 +41,7 @@ limitations under the License.
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Tools/mlir-opt/MlirOptMain.h"
 #include "mlir/Transforms/Passes.h"
+#include "stablehlo/dialect/StablehloOps.h"
 #include "xla/backends/cpu/codegen/emitters/ir/xla_cpu_dialect.h"
 #include "xla/backends/cpu/codegen/emitters/transforms/passes.h"
 #include "xla/backends/cpu/codegen/tiled/transforms/passes.h"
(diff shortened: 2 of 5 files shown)


xla/codegen/emitters/transforms/lower_to_llvm_test.cc
```diff
@@ -0,0 +1,76 @@
...
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Support/LogicalResult.h"
+#include "xla/codegen/emitters/transforms/passes.h"
+#include "xla/stream_executor/device_description.h"
+#include "xla/stream_executor/device_description.pb.h"
+
+// Validates LowerToLLVMPass for Intel GPU targets. This lives as a gtest rather
+// than a lit test because the current gpu_device_info option only understands
+// CUDA and ROCm compute capabilities, so we can't express oneAPI metadata there
+// yet.
```
cota: Perhaps we should add a proto for it already. The proto doesn't need to have much metadata yet -- if you look at the Rocm one it just has a name. This would allow us to write a lit test which has a lot less boilerplate.

cota (commented):
Apologies for the delay. I just landed some changes to lower_to_llvm.cc (and I have one more change pending) so in order to not waste your time I'd like you to wait until that second change lands and then rebase. In the meantime, please address my comment about OneApiComputeCapability since that change would not conflict with the LowerToLLVM pass. Thanks!

xla/codegen/emitters/transforms/lower_to_llvm_test.cc
mdfaijul: @cota Thanks! I will wait for second change on `lower_to_llvm.cc`. Meanwhile, we will create OneAPIComupteCapability.

cota (changes requested):
My changes have now landed ([1](https://github.com/openxla/xla/commit/349a1d4e00c), [2](https://github.com/openxla/xla/commit/bb219c8f9a)) -- please rebase. Note that the file you want to edit now is [lower_to_llvm_gpu.cc](https://github.com/openxla/xla/blob/main/xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc).
Also please go ahead with the proto addition so that we get a lit test for this. Thanks!

mdfaijul:
Thanks! @cota  We will first submit a PR for OneApiComputeCapability and the reopen this PR. Currently changing into draft.

mdfaijul:
@cota  This PR implements OneAPIComputeCapability https://github.com/openxla/xla/pull/35735

mdfaijul:
@cota I have updated the PR with OneApiComputeCapability. Please take a look.

xla/codegen/tools/emitters_opt.cc
```diff
@@ -41,20 +41,21 @@ limitations under the License.
...
 #include "xla/backends/gpu/codegen/emitters/transforms/passes.h"
 #include "xla/codegen/emitters/ir/xla_dialect.h"
+#include "xla/codegen/emitters/transforms/lower_to_llvm_gpu.h"
 #include "xla/codegen/emitters/transforms/pass_pipelines.h"
 #include "xla/codegen/emitters/transforms/passes.h"
 #include "xla/codegen/xtile/ir/transforms/passes.h"
 #include "xla/codegen/xtile/ir/xtile_dialect.h"
 #include "xla/mlir_hlo/mhlo/IR/hlo_ops.h"
 #include "xla/service/gpu/gpu_device_info_for_tests.h"
+#include "stablehlo/dialect/StablehloOps.h"
```
cota: Why has this line been reordered?

xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
```diff
@@ -123,7 +123,12 @@ class LowerToLLVMGPUPass
       return mlir::success();
     };
 
-    if (mlir::failed(LowerToLLVM(getOperation(), populate_patterns))) {
+    // NVVM and ROCDL lower math.log1p directly via their GPU pattern sets, but
+    // the SPIR-V pipeline does not. For Intel GPUs we therefore fall back to
+    // the generic MathToLLVM conversion, hence enabling approximate log1p.
```
cota, line 128: Suggestion to make this a bit more readable: define here a bool right after the comment, and then pass it to LowerToLLVM. No need to add the `/*approximateLog1p*/` argument comment anymore because the bool will be self-explanatory.

cota (changes requested):
Thanks! Just two minor comments to address.

xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
mdfaijul: Added a bool.

xla/codegen/tools/emitters_opt.cc
mdfaijul: It was due to running clang-format locally. I reverted this ordering. Thanks!

mdfaijul:
@cota Thanks! I have addressed the comments.

cota (approved):
Thanks!

PR #34917, opened 2025-12-05
Turn on the scatter determinism expander by default

serach24:
📝 Summary of Changes
Enable xla_gpu_enable_scatter_determinism_expander flag by default (change from false to true).
🎯 Justification
The scatter determinism expander provides significant performance improvements for deterministic scatter operations (up to 9000x speedup for certain input sizes compared to the sequential while-loop approach).
With recent fixes for batched scatter support and proper handling of scatter_dims_to_operand_dims, the pass is now robust enough to be enabled by default.
Users who experience issues can still disable it with --xla_gpu_enable_scatter_determinism_expander=false.
🚀 Kind of Contribution
⚡️ Performance Improvement
🧪 Unit Tests
All existing scatter tests pass with the flag enabled by default:
//xla/service:scatter_determinism_expander_test
//xla/tests:scatter_test

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -441,7 +441,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_enable_fast_math(false);
   opts.set_xla_gpu_experimental_parallel_collective_overlap_limit(1);
   opts.set_xla_pjrt_allow_auto_layout_in_hlo(false);
-  opts.set_xla_gpu_enable_scatter_determinism_expander(false);
+  opts.set_xla_gpu_enable_scatter_determinism_expander(true);
   opts.set_xla_gpu_unsupported_enable_ragged_all_to_all_decomposer(false);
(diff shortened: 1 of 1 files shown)


akuegel (approved):

akuegel:
@serach24 This change triggered a problem in our GemmFusion pass. While the ScatterDeterminismExpander is not directly responsible for this, it makes it easier to roll back for now until we have fixed the other issue.
At the same time I am investigating another issue that might be related to ScatterDeterminismExpander itself, I will let you know when I can confirm whether it is a bug.

akuegel:
> At the same time I am investigating another issue that might be related to ScatterDeterminismExpander itself, I will let you know when I can confirm whether it is a bug.

This other issue was a rewrite of scatters involving S4 types. This triggered a bug in our sort emitter that didn't handle packed types correctly. I have found a fix for that, and once the other issue in NestGemmFusion pass is fixed, I can likely roll forward the flag flip.

PR #34956, opened 2025-12-08
 [ROCm] flush rocprofiler buffer when disabling RocmTracer

Eetusjo:
Fixes a bug in RocmTracer where events not reaching the rocprofiler watermark are not captured when disabling the tracer. Happens e.g. when the workload is very small.

Added an explicit buffer flush in and a relevant test that fails without the flush: `//xla/backends/profiler/gpu:rocm_tracer_test`

🚀 Kind of Contribution
🐛 Bug Fix


Eetusjo:
@xla-rotation could I get a review, please?

beckerhe (approved):

PR #34957, opened 2025-12-08
[ROCm] Enable matmul perf table implementation for ROCm

Eetusjo:
Implements HloOpProfiler for ROCm for use in `//xla/tools:matmul_perf_table_gen`. Also a small fix for hlo_op_profiles to return correct arch name. 

Tests enabled and passing on ROCm:
`//xla/service/gpu/model:hlo_op_profiler_test`
`//xla/service/gpu/model:hlo_op_profiles_test`
`//xla/tools:matmul_perf_table_gen_test`

Matmul perf table generation works:
```
> bazel run //xla/tools:matmul_perf_table_gen_main --config=rocm_clang_official -- ...
entries {
  key: "gfx942"
  value {
    entries {
...
```

### DIFF
--- a/xla/service/gpu/model/BUILD
+++ b/xla/service/gpu/model/BUILD
@@ -785,7 +785,6 @@ cc_library(
         "//xla/backends/profiler/gpu:cupti_collector",
         "//xla/backends/profiler/gpu:cupti_tracer",
     ]) + if_rocm_is_configured([
-        # keep sorted
         "//xla/backends/profiler/gpu:rocm_collector",
         "//xla/backends/profiler/gpu:rocm_tracer",
     ]),
--- a/xla/service/gpu/model/BUILD
+++ b/xla/service/gpu/model/BUILD
@@ -720,6 +720,7 @@ cc_library(
         "//xla/service:hlo_proto_cc",
         "//xla/stream_executor:device_description",
         "//xla/stream_executor/cuda:cuda_compute_capability",
+        "//xla/stream_executor/rocm:rocm_compute_capability",
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/log:check",
         "@com_google_absl//absl/memory",
--- a/xla/service/gpu/model/BUILD
+++ b/xla/service/gpu/model/BUILD
@@ -1,5 +1,6 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@local_config_cuda//cuda:build_defs.bzl", "if_cuda")
+load("@local_config_rocm//rocm:build_defs.bzl", "if_rocm_is_configured")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 load("//xla:xla.default.bzl", "xla_cc_test")
 
@@ -747,7 +748,7 @@ cc_library(
     srcs = ["hlo_op_profiler.cc"],
     hdrs = ["hlo_op_profiler.h"],
     compatible_with = get_compatible_with_portable(),
-    local_defines = if_cuda(["GOOGLE_CUDA"]),
+    local_defines = if_cuda(["GOOGLE_CUDA"]) + if_rocm_is_configured(["TENSORFLOW_USE_ROCM"]),
     deps = [
         ":hlo_op_profile_proto_cc",
         "//xla:debug_options_flags",
@@ -782,6 +783,10 @@ cc_library(
         "//xla/backends/profiler/gpu:cupti_buffer_events",
         "//xla/backends/profiler/gpu:cupti_collector",
         "//xla/backends/profiler/gpu:cupti_tracer",
+    ]) + if_rocm_is_configured([
+        # keep sorted
+        "//xla/backends/profiler/gpu:rocm_collector",
+        "//xla/backends/profiler/gpu:rocm_tracer",
     ]),
 )
 
@@ -844,7 +849,7 @@ xla_test(
     # TODO(b/372714955): Fix the memory leak!
     args = if_google(["--heap_check="]),
     backends = ["gpu"],
-    local_defines = if_cuda(["GOOGLE_CUDA"]),
+    local_defines = if_cuda(["GOOGLE_CUDA"]) + if_rocm_is_configured(["TENSORFLOW_USE_ROCM"]),
     deps = [
         ":hlo_op_profiler_lib",
         "//xla:xla_data_proto_cc",
--- a/xla/service/gpu/model/BUILD
+++ b/xla/service/gpu/model/BUILD
@@ -1,7 +1,8 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
-load("@local_config_cuda//cuda:build_defs.bzl", "if_cuda")
 load("@local_config_rocm//rocm:build_defs.bzl", "if_rocm_is_configured")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("//xla/stream_executor:build_defs.bzl", "if_cuda_or_rocm_is_configured")
+load("//xla/tsl/platform/default:cuda_build_defs.bzl", "if_cuda_is_configured")
 load("//xla:xla.default.bzl", "xla_cc_test")
 
 # Libraries for performance modeling of HLO.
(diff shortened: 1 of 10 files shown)


Eetusjo:
@xla-rotation could I get a review for this PR, please

xla/service/gpu/model/hlo_op_profiler_test.cc
```diff
@@ -29,10 +29,20 @@ namespace xla {
...
 
+#if defined(GOOGLE_CUDA)
+constexpr int kMinClockCyclesAddF32 = 0;
+constexpr int kMinClockCyclesDivideF64 = 280;
+constexpr int kMinClockCyclesSqrtC128 = 1000;
+#elif defined(TENSORFLOW_USE_ROCM)
+constexpr int kMinClockCyclesAddF32 = 0;
+constexpr int kMinClockCyclesDivideF64 = 100;
+constexpr int kMinClockCyclesSqrtC128 = 1000;
+#endif
```
beckerhe: This breaks the ONEAPI build.
Instead of using GOOGLE_CUDA and TENSORFLOW_USE_ROCM compile flags,
can you rather check `HloTestBase::GetTestPlatform()->id()` and then set those values accordingly?

xla/service/gpu/model/hlo_op_profiler_test.cc
```diff
@@ -29,10 +29,20 @@ namespace xla {
...
+
 class HloOpProfilerTest : public HloTestBase {
   void SetUp() override {
-#ifndef GOOGLE_CUDA
-    GTEST_SKIP() << "Not built with --config=cuda";
+#if !defined(GOOGLE_CUDA) && !defined(TENSORFLOW_USE_ROCM)
+    GTEST_SKIP() << "Not built with --config=cuda or --config=rocm";
```
beckerhe: Can you also replace this by a check of `HloTestBase::GetTestPlatform()`?

xla/service/gpu/model/hlo_op_profiler_test.cc
```diff
@@ -96,11 +106,27 @@ TEST_F(HloOpProfilerTest, AllSupportedCombinationsAreMeasurable) {
...
+  const std::unordered_set<HloOpcode> skip_on_rocm = {
+      HloOpcode::kPopulationCount,
+      HloOpcode::kRoundNearestAfz,
+      HloOpcode::kRoundNearestEven,
+  };
+
   FloatTypes.insert(MeasurebleInFloat.begin(), MeasurebleInFloat.end());
   HloOpProfiler profiler(test_runner_as_hlo_runner());
+
+  const bool is_rocm = backend()
```
beckerhe: The usage of the `backend()` function is deprecated, but you can use `HloTestBase::GetTestPlatform()` instead.

xla/service/gpu/model/hlo_op_profiles_test.cc
```diff
@@ -115,6 +128,19 @@ TEST_F(HloOpProfilesTest, GetProfileH100) {
...
+  auto hlo_op_profiles = HloOpProfiles::Load(kDeviceHloOpProfilesTest,
+                                             /*default_profile_name=*/"sm_80");
+  auto device_info_mi210 = TestGpuDeviceInfo::AMDMI210DeviceInfo();
+
+  const auto& op_profile = hlo_op_profiles->GetProfile(device_info_mi210);
+  ASSERT_TRUE(op_profile.contains(
+      std::make_pair(HloOpcode::kMultiply, PrimitiveType::F32)));
+  EXPECT_EQ(
+      op_profile.at(std::make_pair(HloOpcode::kMultiply, PrimitiveType::F32)),
+      123);
```
beckerhe: Can you use `EXPECT_THAT` together with the ::testing::Contains matcher here? This will result in better error messages when something breaks.

xla/tools/matmul_perf_table_gen_test.cc
```diff
@@ -34,12 +34,12 @@ namespace {
 
 class MatmulPerfTableGenTest : public HloTestBase {
   void SetUp() override {
-    if (!backend()
+    auto compute_capability = backend()
```
beckerhe: The style guide wants us to use the full type (no auto) here.

xla/tools/matmul_perf_table_gen_test.cc
```diff
@@ -314,5 +314,36 @@ TEST_F(MatmulPerfTableGenTest, MergeGemmTables) {
...
+  cfg.n_spec.step = 1;
+  cfg.dry_run = false;
+  cfg.dtypes.emplace_back(
+      MatmulPerfTableGen::DataTypeSpec{"f32", "f32", "f32"});
+
+  MatmulPerfTableGen gen(cfg);
+  DeviceHloInstructionProfiles profiles = gen.ComputeTable();
+
+  EXPECT_EQ(profiles.entries_size(), 1);
+  EXPECT_EQ(profiles.entries().begin()->second.entries_size(), 1);
```
beckerhe: If the first EXPECT fails, then the second crashes. Can you use `EXPECT_THAT` with `::testing::ElementsAre` here? This also leads to better error messages when something fails.

beckerhe: In fact it should be possible to make all of this one EXPECT_THAT expression.

Not sure if that's easy to reason about, so I'll leave this up to you.

xla/service/gpu/model/hlo_op_profiler_test.cc
Eetusjo: Changed from flags to checking HloTestBase::GetTestPlatform()->id()

xla/service/gpu/model/hlo_op_profiler_test.cc
Eetusjo: Also changed to HloTestBase::GetTestPlatform

xla/service/gpu/model/hlo_op_profiler_test.cc
Eetusjo: Changed to HloTestBase::GetTestPlatform

xla/service/gpu/model/hlo_op_profiles_test.cc
Eetusjo: Collapsed the two checks into a single EXPECT_THAT with ::testing::Contains

xla/tools/matmul_perf_table_gen_test.cc
Eetusjo: Fixed

xla/tools/matmul_perf_table_gen_test.cc
Eetusjo: Not exactly sure what your idea behind EXPECT_THAT+ElementsAre is. I now changed these check to ASSERT_THATs since the later ones do not make sense if these fail. Since these are meant to check the existence of a single profile and a single entry I think these achieve that.

Let me know if I'm misunderstanding something.

i-chaochen:
Hi @beckerhe @dimitar-asenov this PR is stuck for a long while, could you have a look, please?

dimitar-asenov (approved):

dimitar-asenov (approved):

Eetusjo:
@dimitar-asenov @beckerhe I rebased this again, could you please check? Looks like it was already approved after the fixes but wasn't moving forward.

dimitar-asenov (approved):

akuegel (approved):

dimitar-asenov:
@beckerhe Do you see a nice way to reduce the amount of `#ifdef` and similar macros in this PR?

allanrenucci:
> @beckerhe Do you see a nice way to reduce the amount of #ifdef and similar macros in this PR?

IMO an easy improvement would be to split ROCM and CUDA implementations into separate libraries and in `hlo_op_profiler.cc` instantiate the relevant one (i.e. `RocmKernelTracer` or `CuptiKernelTracer`).

Eetusjo:
Hey @dimitar-asenov , checking on this PR: Is this refactoring into separate libraries (or similar) something you're looking into doing internally, or are there any asks for me at this moment to get this moving forward?

dimitar-asenov:
Hi @Eetusjo. Sorry for the delay on this PR. I just briefly discussed it with Henning and we agreed that the best way to move forward is to implement the split that Allan suggested. Here's a few more details:

- In general, we want to avoid `#ifdef` and friends. We've already refactored many existing occurences. Other cases should be gradually removed as the need arises, such as in this PR.
- The goal is to move all platform selection to the `BUILD` file and leave each individual C++ file as a self-contained unit without conditional code blocks.
- With that in mind, we need to split the `hlo_op_profiler.cc` into three files:
  - `hlo_op_profiler_cuda.cc`
  - `hlo_op_profiler_rocm.cc`
  - `hlo_op_profiler_stub.cc` - This should implement the same API, but simply return an error. We need this internally.
- A great example to follow is https://github.com/openxla/xla/blob/main/xla/pjrt/BUILD#L1356-L1371

Could you please make the necessary changes?

Eetusjo:
> Hi @Eetusjo. Sorry for the delay on this PR. I just briefly discussed it with Henning and we agreed that the best way to move forward is to implement the split that Allan suggested. Here's a few more details:
> ...

Thanks for the information @dimitar-asenov, I'll do this refactoring.

dimitar-asenov (approved):

dimitar-asenov:
Thanks for making the changes. Will let you know if there is some issue detected internally, but I think it looks good.

xla/service/gpu/model/BUILD
```diff
@@ -777,12 +787,116 @@ cc_library(
...
+    ],
+)
+
+cc_library(
+    name = "hlo_op_profiler_lib",
+    srcs = [],
+    hdrs = ["hlo_op_profiler.h"],
+    compatible_with = get_compatible_with_portable(),
+    deps = if_cuda_or_rocm_is_configured(
+        [],
```
akuegel, line 905: You also need to add the deps here for the header includes in hlo_op_profiler.h

Eetusjo: Fixed

i-chaochen: @akuegel wondering do we need to re-approve it again?

akuegel (approved):

xla/service/gpu/model/BUILD
```diff
@@ -798,11 +746,17 @@ xla_cc_test(
 )
 
 cc_library(
-    name = "hlo_op_profiler_lib",
-    srcs = ["hlo_op_profiler.cc"],
+    name = "hlo_op_profiler_cuda",
```
thomasjoerg, line 749: The BUILD deps violate the `no-cuda-only-targets-in-cpu-build` check:
https://github.com/openxla/xla/actions/runs/23739699012/job/69153360514

thomasjoerg (changes requested):

xla/service/gpu/model/BUILD
Eetusjo: Added missing cuda-only tag

thomasjoerg (approved):

xla/service/gpu/model/hlo_op_profiler_rocm.cc
```diff
@@ -0,0 +1,113 @@
...
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "absl/log/check.h"
+#include "absl/log/log.h"
+#include "absl/strings/match.h"
+#include "xla/backends/profiler/gpu/rocm_collector.h"
+#include "xla/backends/profiler/gpu/rocm_tracer.h"
```
thomasjoerg, line 28: Internal CI reports the following build breakage. I'm looking into this. Please let me know if you have a suspicion as to what could cause this.

```
In file included from //xla/service/gpu/model/hlo_op_profiler_rocm.cc:26
In file included from //xla/backends/profiler/gpu/rocm_tracer.h:24
In file included from //xla/stream_executor/rocm/roctracer_wrapper.h:25
In file included from //gpus/rocm/include/rocprofiler-sdk/buffer_tracing.h:29
In file included from //gpus/rocm/include/rocprofiler-sdk/kfd/kfd_id.h:26
In file included from //gpus/rocm/include/rocprofiler-sdk/hsa.h:37
In file included from //gpus/rocm/include/rocprofiler-sdk/hsa/api_trace_version.h:28
In file included from //gpus/rocm/include/hsa/hsa_api_trace.h:50
//gpus/rocm/include/hsa/hsa_ext_amd.h:2324
error: unknown type name 'hsa_handle_t'
 2324 |                                         hsa_handle_t interop_handle,
      |                                         ^
1 error generated.
```

Eetusjo: Interesting, couldn't reproduce this yet at least. Just to make sure: Does your CI run this exact branch, or is it rebased from main or something similar?

akuegel: We might be building with a different version of ROCM. Currently we have imported version 7.2.1

Eetusjo: Thanks, ~double checked on 7.2.1, but don't see this breaking there either. I'll ask around and continue investigating.~. Nvm, got fat fingers. I'll get back to you once I have tested on 7.2.1.

Eetusjo: Yeah, can't reproduce on our 7.2.1 either. I'll continue investigating.

draganmladjenovic: > We might be building with a different version of ROCM. Currently we have imported version 7.2.1

Is this hermetic build? Or one on 7.2.1 docker image?

akuegel: It is in our internal build, so it seems possible that you cannot reproduce that yourself. We import ROCM in a way that we just have the headers and stubs for the implementation, as we don't run any code with ROCM. I checked that hsa_handle_t is defined in the `gpus/rocm/include/rocprofiler-sdk/hsa.h` header. Maybe a missing BUILD dependency?

draganmladjenovic: https://github.com/ROCm/rocm-systems/blob/fb5d0a11f604ec05a3092385b10fcb2e4941aec6/projects/rocr-runtime/runtime/hsa-runtime/inc/hsa.h#L5691 And the header that complains is from the same library. Probably referring to <hsa/hsa.h>

Eetusjo: Double checking, since you mention "I checked that hsa_handle_t is defined in the gpus/rocm/include/rocprofiler-sdk/hsa.h".

I believe the `hsa_handle_t` in `gpus/rocm/include/hsa/hsa_ext_amd.h` should be defined in `gpus/rocm/include/hsa/hsa.h`, not rocprofiler-sdk. hsa_ext_amd.h includes hsa/hsa.h which contains the definition.

akuegel: Yes, you are right, I got the path wrong.

akuegel: I think I figured out a fix. The new BUILD target still contained a copy/pasted `# keep_sorted` comment, which actually prevented the dependencies after that to be sorted into the right lexicographic order by the buildifier tool. After the deps were sorted, for some reason the compile error was gone. I don't really know why, but at least merging this PR is now unblocked.

PR #34979, opened 2025-12-08
Introduce xla_googletest_wrapper with module extension instead of a bazel_dep

meteorcloudy:
So that other projects depending on XLA doesn't have to provide the `local_path_override`


dextero:
#34763 removes the local_path_override instead, and should fix this problem too :)

dextero (changes requested):
#34763 removed the local_path_override. Can you please check if this is still needed?

dextero:
#34763 removed the `local_path_override` (and the entire `xla_googletest_wrapper`) so this isn't necessary any more.

PR #35098, opened 2025-12-10
[GPU] Wrap single instructions in fusions before autotuning.

sergachev:
📝 Summary of Changes
Wrap single instructions into fusions before autotuning so that they get picked up.

🎯 Justification
Creating fusions out of single instructions earlier enables more autotuning.

🚀 Kind of Contribution
⚡️ Performance Improvement ♻️ Cleanup

🧪 Unit Tests:
yes

🧪 Execution Tests:
no

### DIFF
--- a/xla/backends/gpu/codegen/triton/triton_gemm_fusion_test.cc
+++ b/xla/backends/gpu/codegen/triton/triton_gemm_fusion_test.cc
@@ -685,7 +685,8 @@ ENTRY e {
 
   MatchOptimizedHlo(kHloText, R"(
 ; CHECK: ENTRY
-; CHECK: transpose
+; CHECK: fusion
+; CHECK-SAME: kind=kLoop
 ; CHECK: fusion
 ; CHECK-SAME: kind=kCustom
 ; CHECK-SAME: "__triton_nested_gemm_fusion"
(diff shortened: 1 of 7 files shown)


vwbaker:
I remember there was a reason that we couldn't move FusionWrapper all the way up this high, but this is simply adding the pass in two places to add more autotuning, which should be fine functionally. Adding Oleg to double check - this could affect compile time negatively, but performance should only be positive; should we run benchmarks before merging this?

sergachev:
Current behaviour is inconsistent because for example transpose(parameter) is not getting autotuned but reshape(transpose(parameter)) is, same for instructions that stay alone in larger graphs.

olegshyshkov:
This change looks reasonable. We'll get more transposes and other instruction tuned, so performance should improve. But we need to run benchmarks anyway, just in case.

> I remember there was a reason that we couldn't move FusionWrapper all the way up this high

It's more like we need to have FusionWrapper all the way at the end, because any of the passes can insert a stray instruction that needs to be wrapped, because it can be passes to emitters. Those instructions still wouldn't be tuned unfortunately.

olegshyshkov (approved):

PR #35101, opened 2025-12-10
[XLA:GPU] Update memory allocator API to support VA allocation, VA/PA remap

shawnwang18:


ezhulenev:
We want to build new memory allocation APIs around memory_allocation.h, and keep DeviceAddressAllocator as is (it allocates addresses, if they are backed by VMM or not is an implementation details). We'll have something to share soon.

PR #35132, opened 2025-12-11
[XLA:GPU] Update HLO cublas workspace size after autotuner select the algorithm

shawnwang18:
📝 Summary of Changes

This PR introduces a pass that updates the workspace size for cuBLAS/cuBLASLt GEMM operations after autotuning has selected a specific algorithm. The GemmRewriter pass conservatively allocates workspace before autotuning. After autotuning,we know the exact algorithm selected and can query its actual workspace requirement, potentially reducing memory usage.

🎯 Justification
Potentially reducing memory usage.

🚀 Kind of Contribution

🧪 Unit Tests:
Existing gemm tests should cover the workspace size config.

### DIFF
--- a/xla/backends/gpu/runtime/gpublas_lt_matmul_thunk.cc
+++ b/xla/backends/gpu/runtime/gpublas_lt_matmul_thunk.cc
@@ -44,6 +44,7 @@ CublasLtMatmulThunk::CublasLtMatmulThunk(const CublasLtMatmulThunk& rhs)
       gemm_config_(rhs.gemm_config_),
       epilogue_(rhs.epilogue_),
       algorithm_idx_(rhs.algorithm_idx_),
+      autotune_workspace_size_(rhs.autotune_workspace_size_),
       canonical_hlo_(rhs.canonical_hlo_),
       a_(rhs.a_),
       b_(rhs.b_),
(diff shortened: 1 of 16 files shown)


xla/backends/gpu/runtime/gpublas_lt_matmul_thunk.cc
```diff
@@ -135,10 +138,8 @@ CublasLtMatmulThunk::GetCachedMatmulPlan(const ExecuteParams& params) {
 
     TF_ASSIGN_OR_RETURN(auto plan,
                         blas_lt->GetMatmulPlan(gemm_config_, epilogue_));
-    // if workspace buffer is not provided, consider only the algorithms which
-    // do not require a scratch space
-    int64_t max_workspace =
-        workspace_.has_value() ? workspace_.value().size() : 0;
+    // Use the workspace size that was determined during autotuning
+    int64_t max_workspace = autotune_workspace_size_;
```
akuegel, line 145: Maybe I misunderstand something, but shouldn't we use the original (non-adjusted) workspace size when doing the query for GetAlgorithms down below? If we use the possibly smaller adjusted workspace size, some algorithms could be filtered out which can invalidate the algorithm index.

shawnwang18: Yes, the code is trying to use the same workspace size when thunk call GetAlgorithm as the autotuner.  But the comments is misleading, fixed the comments.

akuegel (commented):
Thanks for adjusting the comment. Can you please also add a test for GemmWorkspaceRewriter?

shawnwang18:
> Thanks for adjusting the comment. Can you please also add a test for GemmWorkspaceRewriter?

Done, add a new test: xla/service/gpu/transforms/gemm_workspace_rewriter_test.cc

akuegel (approved):

PR #35138, opened 2025-12-11
[ROCm] Fix rocm build due to inverted condition

alekstheod:
📝 Summary of Changes
Looks there is invalid check introduced in this commit:
https://github.com/openxla/xla/commit/349a1d4e00cf33dbabd1db5771d03b0c5a4aea61
image
 

🎯 Justification
This commit breaks all the rocm tests, we start getting ir code for nvidia gpus. 
Inverting the check fixes them.

🚀 Kind of Contribution
🐛 Bug Fix

Not relevant

🧪 Unit Tests:
All rocm u-tests

🧪 Execution Tests:
CI

### DIFF
--- a/xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
+++ b/xla/codegen/emitters/transforms/lower_to_llvm_gpu.cc
@@ -77,7 +77,7 @@ class LowerToLLVMGPUPass
       : device_spec_(device_description) {}
 
   void runOnOperation() override {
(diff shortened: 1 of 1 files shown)


akuegel (approved):
Thanks for the fix. Seems we didn't catch it because it did not matter for Nvidia GPU code path.

cota (approved):
Oops sorry for the breakage and thanks for the fix!

PR #35211, opened 2025-12-12
[ROCm] Reimplement register spilling detection

nurmukhametov:
Replace amd_comgr library with LLVM's native API to find NT_AMDGPU_METADATA note sections and extract the stack usage and register spill counts from there.

Add detection for dynamic stack usage.

Add VLOG(2) dumps for per-kernel stats as well as register counts.

Change the logic of discarding the module. The module is discarded only if the stack is used, i.e., either .private_segment_fixed_size is not zero or .uses_dynamic_stack is true. There are examples where there are SGPR spills, but they are saved to VGPRs and not to the stack.

Add tests in amdgpu_register_spilling_test.cc which cover cases where no spills, VGPR-only spills, SGPR-only spills, or dynamic stack usage occur. For that, the following LLVM IR inputs are added:
- amdgpu_no_spills.ll: Simple kernel with minimal register usage
- amdgpu_vgpr_spills.ll: High VGPR pressure with limited VGPRs (64)
- amdgpu_sgpr_spills.ll: High SGPR pressure with limited SGPRs (32)
- amdgpu_dynamic_stack.ll: Indirect function call requiring dynamic stack

### DIFF
--- a/third_party/gpus/rocm/BUILD.tpl
+++ b/third_party/gpus/rocm/BUILD.tpl
@@ -579,7 +579,6 @@ alias(
         threshold = 71000,
         value = rocm_version_number(),
     ),
-    visibility = ["//visibility:public"],
 )
 
 cc_library(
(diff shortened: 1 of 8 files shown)


nurmukhametov:
@xla-rotation could I get a review for this PR, please?

akuegel (approved):

PR #35280, opened 2025-12-15
[ROCm] Make autotuner cublaslt_test platform independent and enable amdgpu backends

Eetusjo:
In order to enable autotuner cublaslt tests on ROCm
- Removed dependency on `NVPTXCompiler` and changed to using `Compiler::GetForPlatform`
- Removed cuda-only tag
- Added amdgpu backend

@xla-rotation could I get a review for this PR, please?

### DIFF
--- a/xla/backends/gpu/autotuner/BUILD
+++ b/xla/backends/gpu/autotuner/BUILD
@@ -242,9 +242,10 @@ xla_test(
         "a100",
         "h100",
         "b200",
+        "amdgpu_any",
     ],
     tags = [
-        "cuda-only",
+        "gpu",
         "no_mac",
     ],
     deps = [
@@ -258,9 +259,9 @@ xla_test(
         "//xla/service:compiler",
         "//xla/service:executable",
         "//xla/service:platform_util",
-        "//xla/service/gpu:nvptx_compiler_impl",
         "//xla/stream_executor:blas",
         "//xla/stream_executor:device_description_proto_cc",
+        "//xla/stream_executor:platform",
         "//xla/stream_executor:stream_executor_h",
         "//xla/tsl/lib/core:status_test_util",
         "//xla/tsl/platform:statusor",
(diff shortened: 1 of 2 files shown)


derdrdirk:
@Eetusjo the CublasLt backend is intended to be used exclusively for Nvidia's CublasLt library. The current usage of Cublas and CublasLt backends in amdgpu_compiler.cc should be replaced by a proper AMD ROC BLAS backend:

```
absl::Status AMDGPUCompiler::AddConvAndGemmAutotuningPasses(
    HloPassPipeline* pipeline, const se::GpuComputeCapability& gpu_version,
    const CompileOptions& options, HloModule* hlo_module,
    AutotuneConfig& autotune_config, tsl::thread::ThreadPool* thread_pool,
    se::StreamExecutor* stream_exec,
    const Compiler::GpuTargetConfig* target_config) {
...
  std::vector<std::unique_ptr<CodegenBackend>> backends;
  // TODO(b/407494793): - Add proper support for ROCM. Currently the Cublas
  // backend uses the same API as rocBLAS.
  backends.push_back(std::make_unique<CublasBackend>(
      stream_exec, &debug_options, this, target_config));
  backends.push_back(std::make_unique<CublasLtBackend>(
      stream_exec, &debug_options, this, target_config));
  auto should_autotune = [](const HloInstruction& instruction) -> bool {
    return instruction.opcode() == HloOpcode::kCustomCall &&
           IsCublasGemm(instruction);
  };
  ...
}
```

Is there a plan to implement the corresponding backends?

draganmladjenovic:
@derdrdirk There are no plans to introduce rocm specific blas backends since we share exactly the same interface. Divergence happens in gemm rewriter.

derdrdirk:
I understand that the ROCm and CUDA BLAS interfaces currently share code, but I must insist on separating the autotuning backends. There are ideas from NVIDIA to unify cuBLAS, cuBLASLt, and cuDNN, which would require an AMD ROCm backends to not loose AMD autotuning support. Furthermore we have also seen in the past from MiOpen and CuDNN, that despite sharing the same interface, that they could be developed and maintained easier separately. 

Please add a proper backend for hipBLAS to enable testing.

PR #35315, opened 2025-12-16
[XLA:GPU] Avoid command buffer update through virtual address remapping. 

shawnwang18:
A slide about the idea: 
[update_free_graph_impl.pdf](https://github.com/user-attachments/files/25188093/update_free_graph_impl.pdf)

Some perf results on GB200 cluster: 
image

📝 Summary of Changes

XLA currently uses update base approach to run graph executor, i.e, when memory allocations change, the XLA runtime will try to update the graph executor with the new address before running the executor.

The primary issue of current approach is that it suffers from significant performance penalties that can negate the benefits of using CUDA graphs

This PR implements a design to avoid doing graph update even if memory allocations change. The center idea is separating Virtual Address (VA) ranges from the underlying Physical Address (PA) memory. Instead of rebuilding the graph when memory locations change, the system maintains a stable set of virtual addresses for the graph while dynamically remapping the physical memory behind them.

Without any optimization, when running with VA remapping mode, the execution flow requires these sequential steps:
1.	Reserve VA range for the GpuExecutable
2.	Map VA range to physical addresses (PA) for the current iteration
3.	Execute the GpuExecutable on the GPU
4.	Unmap VA range after execution completes
5.	Go step 2

The issue of this flow is that before unmapping a VA range, we must wait for the GPU to complete its work. This synchronization creates a device bubble — idle GPU time while the CPU performs the VA→PA remapping for the next iteration.

The Solution: Multiple VA Ranges with Interleaving

By reserving multiple VA ranges (typically 2), we can interleave their usage so that:
•	While the GPU executes using VA1, the CPU prepares VA2 for the next iteration
•	While the GPU executes using VA2, the CPU unmaps and re-maps VA1

This eliminates GPU idle time caused by waiting for VA remapping.
Execution Timeline 

```                 
                  +---------------------+---------------------++---------------------+---------------------+
GPU               |  VA1 Execute        |  VA2 Execute        ||  VA1 Execute        |  VA2 Execute        |
                  +---------------------+---------------------++---------------------+---------------------+
        +---------++---------+           +---------++---------+ +---------++---------++---------+           +---------+
CPU     | VA1 Map || VA2 Map |           |VA1 UnMap|| VA1 Map | |VA2 UnMap|| VA2 Map ||VA1 UnMap|           |VA2 UnMap|
        +---------++---------+           +---------++---------+ +---------++---------++---------+           +---------+
```

To implement this overlapping idea, we reserve multiple sets (controlled by `kNumVaInterleaveSlots`) of VAs for each GpuExecutable, and GpuExecutable::ExecuteThunksImpl interleaves the using of VA ranges.  For CommandBufferThunk, it means that I will also create multiple graph executors, and each is corresponding to one VA set.  During runtime, it choose the graph executor based on the VA index passed through execution parameters. 

Key Data Structures
```
VaRanges
    allocation_va_map: Maps each BufferAllocation::Index to its reserved DeviceAddressBase. This guarantees consistent VA addresses across multiple calls to the same GpuExecutable — critical for command buffers that don’t need address updates.
    va_unmap_event: A synchronization event that the device signals upon task completion. The host waits on this event before remapping the VA range to new physical addresses, ensuring safe reuse.
DeviceVaRange
    va_ranges_idx_map: Holds multiple VaRanges instances (indexed by integer), enabling the interleaving scheme.
    idx: Tracks which VA range slot to use next, cycling through available ranges.
```

🚀 Kind of Contribution
⚡️ Performance Improvement,

🧪 Unit Tests:
Reimplements command buffer, existing command buffers should pass.


ezhulenev:
I like the direction, I think more bits can be pushed to SE level, but I'll look into it more closely next week. In general we want VaRanges to be a first class concept in SE and not inside XLA.

ezhulenev:
I.e. why wouldn't we allocate the temp from "physcial" allocator and map it to VA range, then if command buffer doesn't capture any of the parameters, it will always be stable (assuming one execution in flight).

We can have pool of command buffers and VA allocations to have always stable pointers up to POOL_SIZE.

I think this idea needs some investigation. First it would require XLA to change default allocator to use cuMemCreate.

shawnwang18:
> I.e. why wouldn't we allocate the temp from "physcial" allocator and map it to VA range, then if command buffer doesn't capture any of the parameters, it will always be stable (assuming one execution in flight).

I think this make sense, and it requires a minor change.   The implementation in this PR is very similar, the good thing of current approach is that you don't need to do any distinction of the buffer types (just allocate VMM based allocations, and query its physical, then remap to the new addresses).     

Directly allocating temp through `cuMemCreate` requires a bit different path, but that's not a big change.

shawnwang18:
> I.e. why wouldn't we allocate the temp from "physcial" allocator and map it to VA range, then if command buffer doesn't capture any of the parameters, it will always be stable (assuming one execution in flight).
> 
> We can have pool of command buffers and VA allocations to have always stable pointers up to POOL_SIZE.
> 
> I think this idea needs some investigation. First it would require XLA to change default allocator to use cuMemCreate.

Not sure I understand your comments correctly, the update-free cuda-graph works on all buffer types (including parameters).  If we change the default allocator to be VMM based, then it means that we can get the physical address of the parameter buffer, then we can also map the parameter buffer to stable VA ranges, and then all thunks are supported by update-free cuda-graph.

yliu120:
> I.e. why wouldn't we allocate the temp from "physcial" allocator and map it to VA range, then if command buffer doesn't capture any of the parameters, it will always be stable (assuming one execution in flight).
> 
> We can have pool of command buffers and VA allocations to have always stable pointers up to POOL_SIZE.
> 
> I think this idea needs some investigation. First it would require XLA to change default allocator to use cuMemCreate.

I agree with what Eugene said here. I think this should be the correct approach.
The only thing you need is stable pointers to make the SetParams thing removed.

yliu120:
> > I.e. why wouldn't we allocate the temp from "physcial" allocator and map it to VA range, then if command buffer doesn't capture any of the parameters, it will always be stable (assuming one execution in flight).
> > We can have pool of command buffers and VA allocations to have always stable pointers up to POOL_SIZE.
> > I think this idea needs some investigation. First it would require XLA to change default allocator to use cuMemCreate.
> 
> I agree with what Eugene said here. I think this should be the correct approach. The only thing you need is stable pointers to make the SetParams thing removed.

Also discussed with @Tixxx and I think no need to have any pools for stable pointers as long as the buffer addresses won't change during all executions. We will explore some new ideas as well.

apivovarov:
Hi Shawn
This PR includes 40 files with lines stats: +2,569 , -97.
To keep the review manageable and code quality high It should be split into multiple smaller PRs. Try to aim around 300 lines in each PR.

shawnwang18:
> Hi Shawn This PR includes 40 files with lines stats: +2,569 , -97. To keep the review manageable and code quality high It should be split into multiple smaller PRs. Try to aim around 300 lines in each PR.

Hi, Alexander. Yes, this is the planned work.  I've created several smaller PRs that was being reviewed.   

This PR is intended for e2e testing.

PR #35330, opened 2025-12-16
Respect print_metadata option when dumping HLO

olupton:
📝 Summary of Changes
https://github.com/openxla/xla/pull/34060 caused more stack trace metadata to be included in the text representation that is used when calculating HLO module fingerprints. This trickles through into the `fingerprint_before_lhs` attribute, which is used by profiler tooling to group executions of the same program across different processes.

🎯 Justification
Without this change, getting fingerprints to match across processes is unnecessarily difficult, requiring that (for JAX programs) all absolute Python paths and stack frames strictly match - including the parts outside the JIT.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

n/a

🧪 Unit Tests:
New unit test in `xla/service/gpu/gpu_hlo_schedule_test.cc` verifying that the fingerprint is independent of the metadata.

🧪 Execution Tests:
n/a

### DIFF
--- a/xla/service/gpu/gpu_hlo_schedule_test.cc
+++ b/xla/service/gpu/gpu_hlo_schedule_test.cc
@@ -124,12 +124,20 @@ class GpuHloScheduleTest : public HloTestBase {
                                        GetModuleConfig(test_config));
   }
 
-  static bool HasValidFingerprint(HloModule* module) {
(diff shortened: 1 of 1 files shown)


olupton:
To be rebased when https://github.com/openxla/xla/pull/35299 lands.

pifon2a (approved):

PR #35353, opened 2025-12-16
[WIP ROCm] Fix flaky PersistedAutotuningTest.SingleOperationGetsAutotuned

nurmukhametov:
Clear the shared autotune cache in PersistedAutotuningTest::SetUp.

The test was randomly passing or failing depending on test execution order due to a shared global autotune_cache that persists across test executions.

With this fix, the test now consistently fails for ROCm/AMDGPU, which is the expected behavior since transpose autotuning (implemented in #35098) is not yet supported on ROCm/AMDGPU. The test was occasionally and incorrectly passing when it inherited autotune results from other tests via the shared cache.

### DIFF
--- a/xla/service/gpu/gpu_compiler_test.cc
+++ b/xla/service/gpu/gpu_compiler_test.cc
@@ -395,6 +395,13 @@ ENTRY e {
 
 class PersistedAutotuningTest : public HloTestBase {
  protected:
+  void SetUp() override {
+    AutotunerUtil::ClearAutotuneResults();
+    xla_gpu_dump_autotune_results_to_ = GetUniqueTempFilePath(".txt");
+  }
+
+  void TearDown() override { AutotunerUtil::ClearAutotuneResults(); }
+
   static constexpr absl::string_view kHloText = R"(
 HloModule t
 
@@ -428,7 +435,6 @@ ENTRY e {
 
 TEST_F(PersistedAutotuningTest, WriteResultsOnEachCompilation) {
   constexpr absl::string_view kInvalidTextProto = "Invalid!";
-  xla_gpu_dump_autotune_results_to_ = GetUniqueTempFilePath(".txt");
 
   // Check that it writes the results on the first compilation.
   TF_EXPECT_OK(GetOptimizedModule(kHloText).status());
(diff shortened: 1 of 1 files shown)


xla/service/gpu/gpu_compiler_test.cc
```diff
@@ -395,6 +395,11 @@ ENTRY e {
 
 class PersistedAutotuningTest : public HloTestBase {
  protected:
+  void SetUp() override {
```
alekstheod, line 398: Which tests autotuner results it was inheriting from? It seems a right solution is to fix those other tests and put that in TearDown so they don't leave artifacts behind. Maybe there is a test base class which we can do it to avoid modifying all of them?

alekstheod: HloTestBase?

nurmukhametov: > Which tests autotuner results it was inheriting from?

- When `WriteResultsOnEachCompilation` runs first, it populates the shared cache with autotune results from dot operations. When `SingleOperationGetsAutotuned` runs second, it dumps all cache entries into its cache file, including results from the previous test. This causes the test to incorrectly pass because it only checks whether results are non-empty.
- When `SingleOperationGetsAutotuned` runs first with an empty cache, it correctly fails because transpose operations don't generate autotune results.

nurmukhametov: > HloTestBase?

I am not sure it is reasonable. There are probably lots of tests inherited form this class and not all of them generate any autotune results. It is probably make sense to add `ClearAutotuneResults` to `TearDown` too as `GpuCompilerTestWithAutotuneDb` does ([link](https://github.com/ROCm/xla/blob/4eac412ee2ab227489167a54781171a0167eeb8d/xla/service/gpu/gpu_compiler_test.cc#L634)).

alekstheod: Is this is the only tests which does it then lets do it in TearDown instead. Tests shall keep invariants unchanged.

nurmukhametov: I have added the cache clearing in the TearDown.

alekstheod (approved):

derdrdirk (approved):
LGTM

PR #35354, opened 2025-12-16
[ROCM] bug-fixing SortRewriter and fixing self_adjoint_test on ROCM

pemeliya:
📝 Summary of Changes
- This reenables SortRewriter which was erroneously disabled on ROCM (wrong platform).
- SortRewriter is called after EighExpander to make sure sort op inserted by EighExpander gets properly rewritten
- Reenables sort_rewriter_test on ROCM which was marked 'cuda-only'

This shall also improve the perf of EighExpander on CUDA since [this sort op](https://github.com/ROCm/xla/blob/855c22fafe82b400429380621fa8596a523a2014/xla/hlo/transforms/expanders/eigh_expander.cc#L375) gets now rewritten to CUB custom-call

🚀 Kind of Contribution
🐛 Bug Fix, ⚡️ Performance Improvement,

🧪 Unit Tests:
The existing self_adjoint_test was failing on ROCM due to disabled SortRewriter.
Also sort_rewriter_test is now enabled on ROCM too

@xla-rotation could you have a look please?

### DIFF
--- a/xla/service/gpu/transforms/BUILD
+++ b/xla/service/gpu/transforms/BUILD
@@ -2629,7 +2629,6 @@ xla_test(
         "gpu",
     ],
     tags = [
-        "cuda-only",
         "test_migrated_to_hlo_runner_pjrt",
     ],
     deps = [
--- a/xla/service/gpu/transforms/sort_rewriter_test.cc
+++ b/xla/service/gpu/transforms/sort_rewriter_test.cc
@@ -439,6 +439,9 @@ ENTRY %main {
   ROOT %sort = f32[$0,100000] sort(%input), dimensions={1}, to_apply=%compare
 })";
 
+  if (xla::PlatformUtil::CanonicalPlatformName("gpu").value() == "rocm") {
+    GTEST_SKIP() << "Skipping CUDA-specific test";
+  }
   auto pass = SortRewriter(TestGpuDeviceInfo::RTXH100SXMDeviceInfo(), "CUDA");
 
   // Batch 1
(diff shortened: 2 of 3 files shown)


xla/service/gpu/gpu_compiler.cc
```diff
@@ -613,19 +613,6 @@ absl::Status RunOptimizationPasses(
   // would do.
   pipeline.AddPass<PermutationSortExpander>();
 
-  // SortRewriter needs to ask the device how much scratch space is needed,
```
akuegel: We need to keep this here, as we want to run SortRewriter before ComparisonExpander has changed the comparison computation into something that SortRewriter does not match anymore.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -656,6 +643,21 @@ absl::Status RunOptimizationPasses(
   pipeline.AddPass<QrExpander>();
   pipeline.AddPass<EighExpander>();
 
+  // SortRewriter needs to ask the device how much scratch space is needed,
```
akuegel: I don't think this is needed. We do already run SortRewriter a second time after DynamicPadder (which also adds a sort). And DynamicPadder runs after EighExpander. I don't think we need to rewrite the sort as early as possible, after DynamicPadder should still be fine?

akuegel (changes requested):

xla/service/gpu/gpu_compiler.cc
pemeliya: Indeed, the 2nd call to SortRewriter used to be missing on older branches, but now there is no need to move SortRewriter call.

akuegel (approved):

pemeliya:
Hi @akuegel, @ezhulenev 
Replicating my comment here to make sure it won't get missed.
In this PR, I fixed cubsort FFI handler registration for ROCM platform but it was again broken by refactoring in https://github.com/openxla/xla/pull/39083. 

So we will have to create another bugfixing PR.

PR #35356, opened 2025-12-16
[ROCM] fixing warp size for multioutput fusion test for rocm platform

pemeliya:
📝 Summary of Changes
Fixes UT for ROCM platform when warp_size = 64

🎯 Justification
There is one subtest which implicitly assumes that warp_size=32 which breaks on ROCM platform when warp_size=64. 
This is because the decision to use Elemental IR emitter for reductions (**IsUnnestedReductionFasterThanElemental**) uses warp_size provided by DeviceDescription

🚀 Kind of Contribution
🧪 Tests

🧪 Unit Tests:
xla/tests/multioutput_fusion_test.cc

@xla-rotation could you have a look please?

### DIFF
--- a/xla/tests/multioutput_fusion_test.cc
+++ b/xla/tests/multioutput_fusion_test.cc
@@ -21,6 +21,7 @@ limitations under the License.
 #include "xla/tests/xla_test_backend_predicates.h"
 #include "absl/log/check.h"
 #include "absl/strings/str_cat.h"
+#include "absl/strings/substitute.h"
 #include "absl/types/span.h"
 #include "xla/error_spec.h"
 #include "xla/hlo/ir/hlo_computation.h"
@@ -28,6 +29,7 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/literal.h"
 #include "xla/literal_util.h"
+#include "xla/service/platform_util.h"
 #include "xla/shape.h"
 #include "xla/shape_util.h"
 #include "xla/tests/hlo_pjrt_interpreter_reference_mixin.h"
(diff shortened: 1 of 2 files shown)


xla/tests/multioutput_fusion_test.cc
```diff
@@ -378,22 +380,33 @@ TEST_F(MultiOutputFusionTest, MultiOutputReduceFusionMinorWithExtraOutput) {
...
     ENTRY reduce {
-      p = f32[32,32,2]{2,1,0} parameter(0)
-      ROOT fusion = (f32[32,2]{1,0}, f32[32,32,2]{2,1,0}, f32[32,2]{1,0})
+      p = f32[$0,$0,2]{2,1,0} parameter(0)
+      ROOT fusion = (f32[$0,2]{1,0}, f32[$0,$0,2]{2,1,0}, f32[$0,2]{1,0})
         fusion(p), kind=kInput, calls=fused_reduce
     })");
-  auto module = ParseAndReturnVerifiedModule(testcase).value();
+  
+  // This test implicitly assumes that warp_size=32, otherwise such a fusion
```
dimitar-asenov: The comment here is true for the old version of the test. With your fixes there are no assumptions anymore. Please adjust the comment to say that we need to adjust the HLO to match the warp size which is different between CUDA and ROCm.

dimitar-asenov (changes requested):

dimitar-asenov (approved):

xla/tests/multioutput_fusion_test.cc
pemeliya: done!

dimitar-asenov:
I will merge this very soon. I need to make a few adjustments to ensure that it is using the correct APIs for fetching the threads per warp. Let me know if it does not work as expected on ROCm.

PR #35363, opened 2025-12-16
Fixed triton_support_test on rocm.

zoranjovanovic-ns:
📝 Summary of Changes
Updated functions from triton/support.c and triton/support_test to execute
correctly on ROCm.

🎯 Justification
support_test was failing on ROCm.

🚀 Kind of Contribution

🧪 Unit Tests:
Used existing triton/support_test 

🧪 Execution Tests:

### DIFF
--- a/xla/backends/gpu/codegen/triton/support.cc
+++ b/xla/backends/gpu/codegen/triton/support.cc
@@ -47,6 +47,8 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+namespace {
+
 bool IsTritonSupportedDataType(PrimitiveType type,
                                const se::GpuComputeCapability& gpu_version) {
   switch (type) {
@@ -74,8 +76,6 @@ bool IsTritonSupportedDataType(PrimitiveType type,
       return false;
   }
 }
-
-namespace {
 
 // Set of unary elementwise ops that are genuinely supported by Triton.
 absl::flat_hash_set<HloOpcode> TritonSupportedUnaryElementwiseOps(
--- a/xla/backends/gpu/codegen/triton/support.cc
+++ b/xla/backends/gpu/codegen/triton/support.cc
@@ -47,8 +47,6 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
-namespace {
-
 bool IsTritonSupportedDataType(PrimitiveType type,
                                const se::GpuComputeCapability& gpu_version) {
   switch (type) {
@@ -76,6 +74,8 @@ bool IsTritonSupportedDataType(PrimitiveType type,
       return false;
   }
 }
+
+namespace {
 
 // Set of unary elementwise ops that are genuinely supported by Triton.
 absl::flat_hash_set<HloOpcode> TritonSupportedUnaryElementwiseOps(
--- a/xla/backends/gpu/codegen/triton/support.cc
+++ b/xla/backends/gpu/codegen/triton/support.cc
@@ -79,7 +79,7 @@ bool IsTritonSupportedDataType(PrimitiveType type,
 
 // Set of unary elementwise ops that are genuinely supported by Triton.
 absl::flat_hash_set<HloOpcode> TritonSupportedUnaryElementwiseOps(
-    PrimitiveType element_type) {
+    PrimitiveType element_type, const se::GpuComputeCapability& gpu_version) {
   if (element_type == PrimitiveType::PRED) {
     return {HloOpcode::kNot, HloOpcode::kCopy};
   }
@@ -97,8 +97,8 @@ absl::flat_hash_set<HloOpcode> TritonSupportedUnaryElementwiseOps(
   if (element_type != PrimitiveType::F8E5M2 &&
       element_type != PrimitiveType::F8E4M3FN &&
       element_type != PrimitiveType::F8E8M0FNU &&
-      element_type != PrimitiveType::F8E5M2FNUZ &&
-      element_type != PrimitiveType::F8E4M3FNUZ) {
+      !(gpu_version.IsRocm() && element_type == PrimitiveType::F8E5M2FNUZ) &&
+      !(gpu_version.IsRocm() && element_type == PrimitiveType::F8E4M3FNUZ)) {
     ret.insert(HloOpcode::kNegate);
   }
 
@@ -201,8 +201,8 @@ absl::flat_hash_set<HloOpcode> TritonSupportedBinaryElementwiseOps(
   if (element_type == PrimitiveType::S4 || element_type == PrimitiveType::U16 ||
       element_type == PrimitiveType::F8E5M2 ||
       element_type == PrimitiveType::F8E4M3FN || 
-      element_type == PrimitiveType::F8E5M2FNUZ ||
-      element_type == PrimitiveType::F8E4M3FNUZ) {
+      (gpu_version.IsRocm() && element_type == PrimitiveType::F8E5M2FNUZ) ||
+      (gpu_version.IsRocm() && element_type == PrimitiveType::F8E4M3FNUZ)) {
     return {};
   }
 
@@ -256,8 +256,8 @@ absl::flat_hash_set<HloOpcode> TritonSupportedTernaryElementwiseOps(
 
   if (element_type == PrimitiveType::F8E5M2 ||
       element_type == PrimitiveType::F8E4M3FN || 
-      element_type == PrimitiveType::F8E5M2FNUZ ||
-      element_type == PrimitiveType::F8E4M3FNUZ) {
+      (gpu_version.IsRocm() && element_type == PrimitiveType::F8E5M2FNUZ) ||
+      (gpu_version.IsRocm() && element_type == PrimitiveType::F8E4M3FNUZ)) {
     return {HloOpcode::kSelect};
   }
 
@@ -270,7 +270,8 @@ absl::flat_hash_set<HloOpcode> TritonSupportedTernaryElementwiseOps(
 // device of interest.
 bool IsTritonSupportedElementwise(HloOpcode opcode, PrimitiveType element_type,
                                   const se::GpuComputeCapability& gpu_version) {
-  return TritonSupportedUnaryElementwiseOps(element_type).contains(opcode) ||
+  return TritonSupportedUnaryElementwiseOps(element_type, gpu_version)
+             .contains(opcode) ||
          TritonSupportedBinaryElementwiseOps(element_type, gpu_version)
              .contains(opcode) ||
          TritonSupportedTernaryElementwiseOps(element_type, gpu_version)
@@ -286,8 +287,10 @@ CodegenDecision CanTritonHandleReduce(
     const se::GpuComputeCapability& gpu_version) {
   if (reduce.shape().element_type() == PrimitiveType::F8E4M3FN ||
       reduce.shape().element_type() == PrimitiveType::F8E5M2 ||
-      reduce.shape().element_type() == PrimitiveType::F8E5M2FNUZ ||
-      reduce.shape().element_type() == PrimitiveType::F8E4M3FNUZ) {
+      (gpu_version.IsRocm() &&
+       reduce.shape().element_type() == PrimitiveType::F8E5M2FNUZ) ||
+      (gpu_version.IsRocm() &&
+       reduce.shape().element_type() == PrimitiveType::F8E4M3FNUZ)) {
     return CodegenDecision::Forbid(
         "F8E4M3FN and F8E5M2 are not supported for reductions.");
   }
@@ -353,8 +356,7 @@ absl::Status CheckSupportedCheckDotDimensions(const HloDotInstruction& dot) {
   return absl::OkStatus();
 }
 
-bool IsSupportedDotAlgorithm(PrecisionConfig::Algorithm algorithm,
-                             const se::GpuComputeCapability& gpu_version) {
+bool IsSupportedDotAlgorithm(PrecisionConfig::Algorithm algorithm) {
   switch (algorithm) {
     case PrecisionConfig::ALG_UNSET:
     case PrecisionConfig::ALG_DOT_F16_F16_F16:
@@ -394,8 +396,12 @@ CodegenDecision AreTypesSupportedByAlgUnsetDot(
     }
   }
 
-  auto supported_float_types = {BF16, F16, F32, F64, F8E4M3FN, F8E5M2,
-                                F8E4M3FNUZ, F8E5M2FNUZ};
+  std::vector<PrimitiveType> supported_float_types = {BF16, F16, F32, F64,
+                                                      F8E4M3FN, F8E5M2};
+  if (gpu_version.IsRocm()) {
+    supported_float_types.insert(supported_float_types.end(),
+                                 {F8E4M3FNUZ, F8E5M2FNUZ});
+  }
   if (absl::c_linear_search(supported_float_types, input_type)) {
     return CodegenDecision::Allow();
   }
@@ -531,7 +537,7 @@ CodegenDecision IsTritonSupportedDot(
   const PrecisionConfig& precision_config = dot.precision_config();
   const PrecisionConfig::Algorithm algorithm = precision_config.algorithm();
 
-  if (!IsSupportedDotAlgorithm(algorithm, gpu_version)) {
+  if (!IsSupportedDotAlgorithm(algorithm)) {
     return CodegenDecision::Forbid(
         absl::StrCat("Unsupported dot algorithm: ",
                      PrecisionConfig::Algorithm_Name(algorithm)));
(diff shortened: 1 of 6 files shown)


xla/backends/gpu/codegen/triton/support.cc
```diff
@@ -337,7 +353,8 @@ absl::Status CheckSupportedCheckDotDimensions(const HloDotInstruction& dot) {
   return absl::OkStatus();
 }
 
-bool IsSupportedDotAlgorithm(PrecisionConfig::Algorithm algorithm) {
+bool IsSupportedDotAlgorithm(PrecisionConfig::Algorithm algorithm,
+                             const se::GpuComputeCapability& gpu_version) {
```
dimitar-asenov: This function now has an extra argument, but it seems unused. Please remove it, or use it.

xla/backends/gpu/codegen/triton/support.cc
```diff
@@ -377,7 +394,8 @@ CodegenDecision AreTypesSupportedByAlgUnsetDot(
     }
   }
 
-  auto supported_float_types = {BF16, F16, F32, F64, F8E5M2, F8E4M3FN};
+  auto supported_float_types = {BF16, F16, F32, F64, F8E4M3FN, F8E5M2,
+                                F8E4M3FNUZ, F8E5M2FNUZ};
```
dimitar-asenov: Shouldn't the support for the extra new types depend on `gpu_version`?

xla/backends/gpu/codegen/triton/support.cc
```diff
@@ -671,6 +690,8 @@ CodegenDecision IsTritonSupportedInstructionImpl(
     return CodegenDecision(
         element_type != PrimitiveType::F8E4M3FN &&
             element_type != PrimitiveType::F8E5M2 &&
+            element_type != PrimitiveType::F8E4M3FNUZ &&
```
dimitar-asenov: Like above, shouldn't this depend on `gpu_version?`

xla/backends/gpu/codegen/triton/support_test.cc
```diff
@@ -609,12 +614,22 @@ ENTRY triton_computation {
...
+         data_type == PrimitiveType::F8E4M3FN)) {
+      fail_mode = ExpectedFailMode::kCrash;
+    }
+  } else {
+    if (((opcode == HloOpcode::kMaximum || opcode == HloOpcode::kMinimum) &&
+         (data_type == PrimitiveType::F8E5M2 ||
+          data_type == PrimitiveType::F8E4M3FN || 
+          data_type == PrimitiveType::F8E5M2FNUZ ||
+          data_type == PrimitiveType::F8E4M3FNUZ))) {
+      fail_mode = ExpectedFailMode::kFailOrCrash;
```
dimitar-asenov: Why `FailOrCrash`? What happens exactly? We should avoid this mode if possible and prefer something mroe specific.

xla/backends/gpu/codegen/triton/support_test.cc
```diff
@@ -643,11 +658,21 @@ ENTRY triton_computation {
...
+         data_type == PrimitiveType::F8E4M3FN)) {
+      fail_mode = ExpectedFailMode::kCrash;
+    }
+  } else {
+    if (((opcode == HloOpcode::kMaximum || opcode == HloOpcode::kMinimum) &&
+         (data_type == PrimitiveType::F8E5M2 ||
+          data_type == PrimitiveType::F8E4M3FN ||
+          data_type == PrimitiveType::F8E5M2FNUZ ||
+          data_type == PrimitiveType::F8E4M3FNUZ))) {
+      fail_mode = ExpectedFailMode::kFailOrCrash;
```
dimitar-asenov: Why `FailOrCrash`? What happens exactly? We should avoid this mode if possible and prefer something mroe specific.

xla/backends/gpu/codegen/triton/support_test.cc
```diff
@@ -701,7 +726,20 @@ ENTRY triton_computation {
...
+  if (cc.IsRocm()) {
+    skip_failure_branch_to_avoid_crash =
+        (opcode == HloOpcode::kClamp || opcode == HloOpcode::kSelect) &&
+        (data_type == PrimitiveType::F8E5M2 ||
+         data_type == PrimitiveType::F8E4M3FN ||
+         data_type == PrimitiveType::F8E5M2FNUZ ||
+         data_type == PrimitiveType::F8E4M3FNUZ);
+  }
+
+  RunSupportTest(std::move(ti), /*output_tile_sizes=*/{1, 32}, cc,
```
dimitar-asenov: Why `FailOrCrash`? What happens exactly? We should avoid this mode if possible and prefer something mroe specific.

xla/backends/gpu/codegen/triton/support_test.cc
```diff
@@ -2324,6 +2378,12 @@ ENTRY triton_computation {
   if (absl::c_linear_search(std::vector{F8E5M2, F8E4M3FN, S8}, data_type)) {
     fail_mode = ExpectedFailMode::kFailOrCrash;
   }
+  if (cc.IsRocm()) {
+    if (absl::c_linear_search(std::vector{F8E4M3FNUZ, F8E5M2FNUZ, F8E4M3FN,
+                                          S8, S16, S32, S64}, data_type)) {
+      fail_mode = ExpectedFailMode::kFailOrCrash;
```
dimitar-asenov: Why `FailOrCrash`? What happens exactly? We should avoid this mode if possible and prefer something mroe specific.

xla/backends/gpu/codegen/triton/support_test.cc
```diff
@@ -2416,6 +2476,15 @@ ENTRY triton_computation {
                             data_type)) {
     fail_mode = ExpectedFailMode::kFailOrCrash;
   }
+  if (cc.IsRocm()) {
+    if (absl::c_linear_search(std::vector{F8E4M3FN, F8E5M2FNUZ, F8E4M3FNUZ, F64},
+                                data_type) ||
+        (absl::c_linear_search(std::vector{S64, S32, S16, BF16, F16, F32},
+                                data_type)  &&
+         algorithm == xla::PrecisionConfig::ALG_DOT_F64_F64_F64)) {
+      fail_mode = ExpectedFailMode::kFailOrCrash;
```
dimitar-asenov: Why `FailOrCrash`? What happens exactly? We should avoid this mode if possible and prefer something mroe specific.

dimitar-asenov (changes requested):
Thanks, I left you some comments. For the ones asking about `FailOrCrash`, please either adjust the code, or add a brief comment to the code explaining why `FailOrCrash` is the right choice.

zoranjovanovic-ns:
@dimitar-asenov updated PR (applied review comments).

dimitar-asenov (approved):

penpornk:
[Linux x86 GPU CI failed](https://github.com/openxla/xla/actions/runs/20316907944/job/58428882612?pr=35363). Could you please help take a look?
```
[ RUN      ] BinaryElementwiseTestSuite/BinaryElementwiseTest.IsTritonSupportedBinaryElementwise/f8e4m3fnuz_divide_100
LLVM ERROR: Unsupported bitwidth
/b/f/w/bazel-out/k8-opt/bin/xla/backends/gpu/codegen/triton/support_test.runfiles/xla/build_tools/ci/parallel_gpu_execute: line 74:    28 Aborted                 (core dumped) "$TEST_BINARY" $@
```

zoranjovanovic-ns:
Already looking at the issue.

zoranjovanovic-ns:
Cuda tests are failing because of changes in emitter_helpers.cc - adding support for F8E5M2FNUZ and F8E4M3FNUZ data types.
Before changes, CreateTritonModule always failed (resulting in test failure) and now it manages to create the module, but after that Triton pipeline crashes because those types are not supported.
Basically, number of tests that used to fail (in XLA) now crashes (in Triton passes) and that results in test failure.
Triton passes do not always crash with F8E5M2FNUZ and F8E4M3FNUZ - sometimes just fail.
This, most probably, means that it will be necessary to use ExpectedFailMode::kFailOrCrash for those tests (some tests fail and some tests crash).
Will try to find some additional time tomorrow to fix the tests for Cuda.

penpornk:
Thank you for the update!

zoranjovanovic-ns:
@penpornk  updated PR.

zoranjovanovic-ns:
@xla-rotation Can someone take a look at updated version?

dimitar-asenov (changes requested):
Thanks for looking into the issue.

Please do not use `GTEST_SKIP` or any other way to skip these tests. They should not be skipped, but rather the behavior that actually happens should be tested (if it is a crash then check for a crash).

The reason is this:
- We want the Triton support tests to precisely reflect the current known state of the implemenetation.
- If you skip a certain test combination, then we get a hole in our tests. Imagine that at a later time, Cuda starts supporting these types. Now our tests won't detect this in any way and we are very likely to forget to update them.
- Instead, if you specifically test that these types do not work under the CUDA (e.g. by testing for a crash) and at some later point Cuda implements them, then our tests will detect the discrepancy and we will update them. This will ensure that we're testing the full possible extend of the Triton code-gen capabilities.

zoranjovanovic-ns:
> Please do not use `GTEST_SKIP` or any other way to skip these tests. They should not be skipped, but rather the behavior that actually happens should be tested (if it is a crash then check for a crash).

Thank you, will modify the tests.

zoranjovanovic-ns:
> Thanks for looking into the issue.
> 
> Please do not use `GTEST_SKIP` or any other way to skip these tests. They should not be skipped, but rather the behavior that actually happens should be tested (if it is a crash then check for a crash).
> 
> The reason is this:
> 
> * We want the Triton support tests to precisely reflect the current known state of the implemenetation.
> * If you skip a certain test combination, then we get a hole in our tests. Imagine that at a later time, Cuda starts supporting these types. Now our tests won't detect this in any way and we are very likely to forget to update them.
> * Instead, if you specifically test that these types do not work under the CUDA (e.g. by testing for a crash) and at some later point Cuda implements them, then our tests will detect the discrepancy and we will update them. This will ensure that we're testing the full possible extend of the Triton code-gen capabilities.

Having several issues with this.
One of the problems is with conversions: from F8E4M3FNUZ to F8E4M3FNUZ and from F8E5M2FNUZ to F8E5M2FNUZ.
This should not be supported on Cuda (everything with those 2 types should fail), but those just (accidentally) pass in Cuda Triton - resulting in issues with conversion test.
Will additionally investigate this.

zoranjovanovic-ns:
@dimitar-asenov As far as I can see, the problem is that some operations just happen to work on Cuda Triton for F8E4M3FNUZ and F8E5M2FNUZ data types.

It seems to me that, with current tests logic, only way to have some tests to pass is to modify some of the functions from support.c (e.g. IsTritonSupportedConversion) to return **true** for those data types. Otherwise, test expects those cases to fail (it does not matter with fail or crash) but some pass - resulting in test failures.

In my opinion, we should not modify IsTritonSupportedConversion and similar functions to return true for cuda and those data types, because there is no intentional support for F8E4M3FNUZ and F8E5M2FNUZ in cuda Triton.

dimitar-asenov:
Thanks for the details.

I see two options:
- If the generated code really works and that's not a bug, then we should change `IsTritonSupportedConversion`.
- If the generated code works, but that's more of an accident and seems broken, we can change the implementation of the codegen to check for these types under CUDA and if encountered raise and explicit error. This will enable us to properly write the test without modifying `IsTritonSupportedConversion`.

Could you please clarify which exact tests seem to pass? I'm not that familiar with the FP8 support on CUDA and Triton, what do you mean when you say that it's unintentional?

zoranjovanovic-ns:
> I'm not that familiar with the FP8 support on CUDA and Triton, what do you mean when you say that it's unintentional?

I meant that generated code works by accident, or because conversion from one fp8 type to the same fp8 does not require much code :) .

Will check second option from your comment.

dimitar-asenov:
If the "accidentally working" conversion cases are only when the input and output types are identical (so it's a noop), then this could be something we declare supported. Technically we care about what Triton supports. So even if CUDA does not support some FP8 type, if Triton implements support for it (in this case by just doing nothing), then that's good enough.

zoranjovanovic-ns:
Hi @dimitar-asenov , updated PR.

xla/backends/gpu/codegen/triton/support.h
```diff
@@ -66,6 +66,11 @@ CodegenDecision IsTritonSupportedComputation(
 // `kTritonGemmFusionKind`.
 bool IsTritonFusedComputation(const HloComputation& computation);
 
+// Returns `CodegenDecision`'s equivalent of `true` if data type
+// is supported by the Triton emitters for the given compute capability.
+bool IsTritonSupportedDataType(PrimitiveType type,
```
dimitar-asenov: I'm not sure why we expose this and would prefer not to. We should only need to call the functions above. See my comment in `xtile_compiler.cc`

xla/backends/gpu/codegen/triton/xtile_compiler.cc
```diff
@@ -229,12 +229,35 @@ absl::Status IsTritonSupportedFusion(const HloFusionInstruction& fusion,
   return absl::OkStatus();
 }
 
+absl::Status ValidateUsedTypes(
```
dimitar-asenov, line 232: Below, this new function is called after you have already verified that the fusion is supported by Triton. `IsTritonSupportedFusion` (code just above) internally calls `IsTritonSupportedInstruction` which in turn calls `IsTritonSupportedDataType`.

So if `IsTritonSupportedFusion` returns true, there should be no need to call `IsTritonSupportedDataType` also here. Please remove this function and do not expose `IsTritonSupportedDataType`. Ideally, things will just work,. but if that somehow breaks your code, then there might be a need to adjust one of `IsTritonSupportedFusion` or `IsTritonSupportedInstruction`.

dimitar-asenov (changes requested):

xla/backends/gpu/codegen/triton/xtile_compiler.cc
zoranjovanovic-ns: IsTritonSupportedFusion  is not called for all instructions.
It is called only for kFusion (and only for nested fusions) and kPad instructions.
It is never called for e.g. convert instructions, resulting in returning True for some instructions that are using fp8...fnuz  data types. Those instructions are converted to triton ir and then we have issue that I have described previously (in some cases triton codegen crashes, in some fails, but in some cases it just pass - conversion from one fp8 to the same fp8 type - which is not expected by the test and it fails).

With current implementation, additional check is necessary to detect presence od fp8...fnuz data types in triton ir before executing triton passes on it - in order to have codegen to fail (not to pass when it works with not supported fp8 types) - to be able to handle it with kFail of kCrash.

I could replace usage of IsTritonSupportedDataType with IsTritonSupportedInstruction to avoid exporting another function from support.cc.

zoranjovanovic-ns:
@dimitar-asenov updated PR.
Modified implementation to use IsTritonSupportedInstruction (instead of IsTritonSupportedDataType) to check all instructions before creating triton module,
As result number of test cases with kCrash significantly reduced.

dimitar-asenov:
Sorry for the delay on this. We would like to submit a related cleanup internally and then it should be possible to merge this PR. Will need a bit more time. @WillFroom is doing the cleanup.

zoranjovanovic-ns:
> Sorry for the delay on this. We would like to submit a related cleanup internally and then it should be possible to merge this PR. Will need a bit more time. @WillFroom is doing the cleanup.

@dimitar-asenov @WillFroom Thank you for the info.
Any raw estimate on when this will be ready?
It would be good to know this in order to better organize our work.

I can see some related changes here:
https://github.com/openxla/xla/commit/6443e3c3d7a462af8d8ece499c931b771cc3d136

Is it going to be something more or this is it?

WillFroom:
Hello @zoranjovanovic-ns sorry for the delay, I am looking at this and I should hopefully should have my refactor merged by early next week

zoranjovanovic-ns:
Thank you for the info.

zoranjovanovic-ns:
> Hello @zoranjovanovic-ns sorry for the delay, I am looking at this and I should hopefully should have my refactor merged by early next week

Hi @dimitar-asenov  @WillFroom,  just to verify, does this commit contains refactoring that you have mentioned:
https://github.com/openxla/xla/commit/0c70fc107ebb3f5803070256d96a393ff93fa1b3 ?

Is something else expected or I can rebase my PR to proceed with upstreaming my changes?

dimitar-asenov:
The commit you linked is independent to the cleanups we wanted to do. Regarding those we explored one path internally but that turned out to not be quite what we want, so now we have to do something else. Sorry for the delay.

zoranjovanovic-ns:
> The commit you linked is independent to the cleanups we wanted to do. Regarding those we explored one path internally but that turned out to not be quite what we want, so now we have to do something else. Sorry for the delay.

Thank you for the fast reply.
Any estimate on when this refactoring will be ready (to be able to plan my activities on upstreaming changes)?

dimitar-asenov:
@zoranjovanovic-ns The refactoring is in:
https://github.com/openxla/xla/pull/37640

Please rebase on top and adjust this PR. To reiterate the important points:
- The code-gen layer should not call into any `IsSupported...` function.
- Your code-gen should avoid explicit checks of support conditions and just lets lower levels  (Triton/LLVM) fail gracefully.
- There are two cases where writing extra checks in your code-gen is reasonable:
  - If the check is necessary to decide how to codegen.
  - If the lower levels fail with a crash.
  - If you write checks here, they should not call into the general `IsSupport...` function.
- The support test should accurately reflect what is supported, what fails gracefully, and what crashes.

Hope this helps.

zoranjovanovic-ns:
Created new PR to replace this one:
https://github.com/openxla/xla/pull/38507

PR #35456, opened 2025-12-17
Manage events for intra-process data transfers after enqueue of dependencies instead of completion

rao-ashish:
📝 Summary of Changes
This PR modifies `PjRtStreamExecutorRawBuffer` to allocate and record the definition event of the receive buffer / the usage event of the send buffer after the definition event of the send buffer has been recorded instead of waiting for it to complete. This is a fix similar to #35113 but for the single-process case.

🎯 Justification
Intra-process data transfers currently allocate and record events only after the buffers they need to transfer have been materialized, instead of just after their definition events have been recorded. This means that if the host tries to launch an executable that depends on the result of the transfer, the host thread will block until the buffer being sent by the transfer has been fully materialized.

[Here](https://gist.github.com/rao-ashish/855d4536af5c96413a1826955728237a) is a toy program illustrating the issue, which transfers an array from one set of devices to another, does a matmul, transfers the result back to the first set of devices, and does another matmul. A profile shows that the host blocks while enqueuing the second matmul until the first matmul is complete:

image

After this fix, we see that the host can enqueue the computation without blocking:

image

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
The previously added unit tests inside `se_gpu_pjrt_client_test.cc` continue to pass (the existing `AsyncCopyToDevice` test in particular already tests the relevant functionality).

🧪 Execution Tests: Verified that the implementation still works on these [four end-to-end tests](https://gist.github.com/rao-ashish/aeaf2a1166bcce60e54186a004acf0d2).

### DIFF
--- a/xla/pjrt/se_raw_buffer.cc
+++ b/xla/pjrt/se_raw_buffer.cc
@@ -417,8 +417,8 @@ void PjRtStreamExecutorRawBuffer::CopyTo(
   if (dst_raw_buffer->memory_space()->client() == memory_space()->client()) {
     IntraClientCopyToWithDependencies(
         /*dependencies=*/{}, std::move(dst_raw_buffer),
-        std::move(definition_event_promise),
-        std::move(src_usage_event_promise));
+        std::move(definition_event_promise), std::move(src_usage_event_promise),
+        std::move(allocation_event));
   } else if (auto* src_ptr = GetHostPointer()) {
     auto h2d_event = dst_raw_buffer->CopyRawHostToDeviceAndReturnEvent(
         src_ptr, 0, GetOnDeviceSizeInBytes());
@@ -507,24 +507,10 @@ void PjRtStreamExecutorRawBuffer::ScheduleCopyTo(
          definition_event_promise = std::move(definition_event_promise),
          src_usage_event_promise = std::move(src_usage_event_promise),
          allocation_event = std::move(allocation_event)]() mutable {
-          for (const auto& av : transfer_dependency_avs) {
-            if (auto* error = av->GetErrorIfPresent()) {
-              auto status = *error;
-              if (allocation_event) {
-                allocation_event.SetError(status);
-              }
-              definition_event_promise->SetError(status);
-              src_usage_event_promise->SetError(status);
-              return;
-            }
-          }
-          if (allocation_event) {
-            allocation_event.SetStateConcrete();
-          }
           this_ref->IntraClientCopyToWithDependencies(
               std::move(transfer_dependency_avs), std::move(dst_raw_buffer),
               std::move(definition_event_promise),
-              std::move(src_usage_event_promise));
+              std::move(src_usage_event_promise), std::move(allocation_event));
         });
     return;
   }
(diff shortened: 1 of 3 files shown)


pschuh (approved):

emilyfertig (approved):

emilyfertig:
Hi @allanrenucci can we pull this in? It looks like pschuh and I don't have the right permissions

allanrenucci (approved):

gspschmid:
@allanrenucci Could you retrigger the import? The jobs above seem to have failed for unrelated reasons, e.g.
```
ERROR: Error computing the main repository mapping: Label '@@local_xla//third_party/cccl:workspace.bzl' is invalid because 'third_party/cccl' is not a package; perhaps you meant to put the colon here: '@@local_xla//third_party:cccl/workspace.bzl'?
```

allanrenucci:
It was imported internally. It is pending review.

thomasjoerg (changes requested):
Hi there, I'm looking into merging the PR at Google. We typically ask for unit tests for changes of this size. Please add some coverage unless existing tests do already provide it.

rao-ashish:
Thanks for taking a look! Two tests that I believe already test the modified code are [`CopyToMemorySpace`](https://github.com/rao-ashish/xla/blob/b5534d35f8ec247410600530d6cb72f67253c5c1/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc#L1328) and [`AsyncCopyToDevice`](https://github.com/rao-ashish/xla/blob/b5534d35f8ec247410600530d6cb72f67253c5c1/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc#L1120) inside `se_gpu_pjrt_client_test.cc`. Both these tests call into `CommonPjRtBufferImpl::CopyToMemorySpace`, which calls `CommonPjRtBufferImpl::DirectCopyToMemorySpace` , which in turn will call the `PjRtStreamExecutorRawBuffer::ScheduleCopyTo` / `PjRtStreamExecutorRawBuffer::CopyTo` code that the PR modifies.`AsyncCopyToDevice` confirms that the modified code still correctly waits for dependencies before executing transfers, and `CopyToDevice` confirms that the modified code works in the simpler case when the buffers to send are already ready. Let me know if this is sufficient, but happy to add more tests if needed!

thomasjoerg:
> Thanks for taking a look! Two tests that I believe already test the modified code are [`CopyToMemorySpace`](https://github.com/rao-ashish/xla/blob/b5534d35f8ec247410600530d6cb72f67253c5c1/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc#L1328) and [`AsyncCopyToDevice`](https://github.com/rao-ashish/xla/blob/b5534d35f8ec247410600530d6cb72f67253c5c1/xla/pjrt/gpu/se_gpu_pjrt_client_test.cc#L1120) inside `se_gpu_pjrt_client_test.cc`. Both these tests call into `CommonPjRtBufferImpl::CopyToMemorySpace`, which calls `CommonPjRtBufferImpl::DirectCopyToMemorySpace` , which in turn will call the `PjRtStreamExecutorRawBuffer::ScheduleCopyTo` / `PjRtStreamExecutorRawBuffer::CopyTo` code that the PR modifies.`AsyncCopyToDevice` confirms that the modified code still correctly waits for dependencies before executing transfers, and `CopyToDevice` confirms that the modified code works in the simpler case when the buffers to send are already ready. Let me know if this is sufficient, but happy to add more tests if needed!

Sounds plausible, but I am not very familiar with this part of the codebase. I'm going to defer to @pschuh here.

pschuh:
This is covered by CopyToMemorySpace (but not the error paths). Looking at the code closely, I think some of this logic might not work with error events populated after some time. We should possibly test that directly. We're also blocking the async runner threads waiting for xla/pjrt/buffer_sequencing_event.h;l=157 to become ready. I think this new path probably should probably call ScheduleWhenReady on these instead.

rao-ashish:
@pschuh Got it, thanks. I've just added a commit that should fix the error handling. It also adds tests `CopyErrorBufferToDevice` and `CopyDelayedErrorBufferToDevice` to test the error handling, and modifies `IntraClientCopyToWithDependencies` to use `ScheduleWhenReady`. Let me know if this looks good now.

xla/pjrt/se_raw_buffer.cc
```diff
@@ -525,4 +489,138 @@ void PjRtStreamExecutorRawBuffer::CopyTo(
...
+
+          if (auto* error = av_to_check->GetErrorIfPresent()) {
+            auto status = *error;
+            client->SetEventAsError(usage_event, status);
+            if (allocation_event) {
+              allocation_event.SetError(status);
+            }
+            return;
+          }
+        }
```
pschuh, line 574: Not sure allocation_events are really used in the GPU client, but 
if (allocation_event) {
  allocation_event.SetStateConcrete();
}
needs to be called before client->WaitForAllocation(stream, *dst_raw_buffer).

rao-ashish: If `allocation_event` is set to concrete before `client->WaitForAllocation`, how do we propagate the status from `stream->MemcpyD2D` if it's an error? Is it alright in that case for `allocation_event` to be set to concrete, but just propagate the error through `usage_event`?

Also, seems like inside `PjRtStreamExecutorClient::AllocateRawBuffer`, the allocation event `allocate_after` is [asserted to be nullptr](https://github.com/openxla/xla/blob/b22a8638122c17082a83ca4958f19f729831b8a6/xla/pjrt/pjrt_stream_executor_client.cc#L483). Given this, should we just also assert that it is nullptr here instead of modifying it?

pschuh: I guess, its not to hard to implement properly though in case we want to support allocate_after properly on GPU. It is ok to just allocate the buffer in that circumstance because the only errors should be inside the actual operation.

rao-ashish: Ok, I've added a commit to move `SetStateConcrete` before `WaitForAllocation(dst_raw_buffer)` (also noticed that the allocation_event was being set to concrete twice in `CopyTo()` for the intra-client case, so fixed that).

pschuh (approved):

rao-ashish:
@thomasjoerg this should be ready now (see the discussion above, tests `CopyErrorBufferToDevice` and `CopyDelayedErrorBufferToDevice` have been added to test the error handling of the new code).

rao-ashish:
@thomasjoerg Just following up on this -- I've added the error handling tests discussed above and @pschuh approved the changes; is there anything else needed before this can be merged?

thomasjoerg (approved):

PR #35463, opened 2025-12-17
[xla:gpu] Support ncclAlltoall directly for contiguous send/recv buffers

ezhulenev:
With latest NCCL we can use `ncclAlltoall` API directly without having to launch grouped send and recv operations.

### DIFF
--- a/xla/backends/gpu/collectives/nccl_communicator.cc
+++ b/xla/backends/gpu/collectives/nccl_communicator.cc
@@ -640,6 +640,32 @@ absl::Status NcclCommunicator::LaunchAllGather(
   return absl::OkStatus();
 }
 
+// If all buffers are contiguous returns a device address range that covers all
+// of them, otherwise returns an empty optional.
+static std::optional<se::DeviceAddressBase> IsContinguous(
+    absl::Span<const se::DeviceAddressBase> buffers) {
+  if (buffers.empty()) {
+    return std::nullopt;
+  }
+
+  if (buffers.size() == 1) {
+    return buffers[0];
+  }
+
+  size_t total_size = buffers[0].size();
+  for (size_t i = 1; i < buffers.size(); ++i) {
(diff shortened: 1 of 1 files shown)


ezhulenev:
@PatriosTheGreat @olegshyshkov (still can't assign reviewers on my PRs)

xla/backends/gpu/collectives/nccl_communicator.cc
```diff
@@ -640,6 +640,32 @@ absl::Status NcclCommunicator::LaunchAllGather(
   return absl::OkStatus();
 }
 
+// If all buffers are contiguous returns a device address range that covers all
+// of them, otherwise returns an empty optional.
+static std::optional<se::DeviceAddressBase> IsContinguous(
```
olegshyshkov, line 645: There is already another PR to do this, but it hasn't been updated for a week. The same comment [0] applies here, could we maybe not try to reconstructs if buffers are contiguous and instead introduce another override of `NcclCommunicator::LaunchAllToAll` that accepts a contiguous buffer? It will fit nicely in AllToAllThunk and we can move the slicing logic into NcclCommunicator.

[0] https://github.com/openxla/xla/pull/34493/files#r2569443694

olegshyshkov (changes requested):

xla/backends/gpu/collectives/nccl_communicator.cc
ezhulenev: It will require changing base `Communicator::AllToAll` API, and it will touch a lot of other implementation and in general will make it less flexible. I like keeping `AllToAll` flexible and falling back on faster implementations internally if possible.

ezhulenev: FWIW checking buffers is super cheap

olegshyshkov (approved):

hawkinsp:
This needs to be added to xla/tsl/cuda/nccl.symbols.

PR #35482, opened 2025-12-18
Correctly generate compile_commands.json

ezhulenev:
Sometime json incorrectly parse compile commands from blaze, and we end up passing them as

```
"-isystem path/to/includes"
```

to `clangd`, and these flags parsed incorrectly

### DIFF
--- a/build_tools/lint/generate_compile_commands.py
+++ b/build_tools/lint/generate_compile_commands.py
@@ -67,7 +67,11 @@ class CompileCommand:
       if arg.endswith(".cc"):
         cc_file = arg
 
-      filtered_args.append(arg)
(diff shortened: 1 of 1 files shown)


penpornk (approved):
Nice to see you and thanks for the fix, Eugene! :)

penpornk:
You forgot to change your Github email, by the way.

ezhulenev:
Still figuring out how is it to work from the other side 😊 what’s up with copybara? Is it some real error?

penpornk:
> what’s up with copybara? Is it some real error?

I was waiting for @ddunl just in case. But I've manually verified your change so I guess I can just merge it.

PR #35540, opened 2025-12-18
[XLA:CPU][codegen] Fix llvm intrinsic lowering in CPU pipeline

akhilgoe:
LLVM now has an explicit dependency on library lowering analysis info for libcall lowering decisions, [link](https://github.com/llvm/llvm-project/commit/081a99735c04b2018eeb7e74f9860e1d0e1b#diff-680aac42e651b506634e9cf972770658b7a5119e233a168c11ff7c8edc443e56R666). Hence, this PR adds a pass to the codegen pipeline to make that information available to the compiler. In the absence of the analysis, target intrinsics are marked unsupported and emitted as inline loops.


seantalts (approved):
Looks great, thank you. Just to confirm, the test you've added fails without the pass added, right?

akhilgoe:
@seantalts, thanks for the review! Yes that's right, the test fails without the pass. I have only tested it on x86 but I'd expect similar results on aarch64 too.

PR #35568, opened 2025-12-19
Fix crash in LatencyHidingScheduler with compute offload

zhenying-liu:
📝 Summary of Changes
When running JAX unit tests with compute offload (e.g., test_compute_host_loop), the LatencyHidingScheduler(LHS) crashes with std::out_of_range exceptions when trying to access schedules and pressure states for host computations.

Host computations created during compute offload don't have device schedules (they execute separately on the host CPU), but the scheduler assumed all computations have schedules and pressure states, leading to crashes. So we add defensive checks to skip host computations that don't have schedules or pressure states.

🎯 Justification
This CL prevents JAX compute offload code from crashing in LHS when LHS is enabled. Without this fix, any HLO module containing host computations (created during compute offload) would crash during scheduling.

🚀 Kind of Contribution
 🐛 Bug Fix

N/A.

🧪 Unit Tests:
A new unit test was added: HostOffloadComputationsWithoutSchedule in xla/service/latency_hiding_scheduler_test.cc.

🧪 Execution Tests:
 The fix was verified with JAX end-to-end execution tests. jax/tests/memories_test.py contains compute offload tests in the ComputeOffload test class (e.g.,  test_compute_host_loop) that previously failed due to the scheduler crash. After the fix, these tests pass, confirming the scheduler correctly handles host computations without device schedules.

### DIFF
--- a/xla/service/latency_hiding_scheduler.cc
+++ b/xla/service/latency_hiding_scheduler.cc
@@ -828,6 +828,13 @@ BufferInfoTracker::BufferInfoTracker(
   std::function<void(const HloComputation*)> process_computation =
       [&process_computation, module, alias_analysis, this,
        &shape_size_bytes](const HloComputation* computation) {
+        // Skip computations that don't have schedules (e.g., host computations
+        // created during compute offload that are executed separately).
+        // Note: We don't need to track memory usage on CPU for now. If needed,
+        // it can be added later.
+        if (!module->schedule().is_computation_scheduled(computation)) {
+          return;
+        }
         const HloInstructionSequence& sequence =
             module->schedule().sequence(computation);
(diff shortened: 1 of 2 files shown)


zhenying-liu:
Please shed some lights on [AMD ROCm -- Community CI Build](http://aus-navi3x-03.amd.com:8080/job/OpenXLA/job/XLA-multibranch/job/PR-35568/1/display/redirect), as I can't access.

i-chaochen:
Hi @zhenying-liu no worries about ROCm CI, the red is irrelevant to your PR

seherellis (approved):
Approving to create an internal code change, so we can run our internal tests. Will update here once the tests finish.

seherellis (approved):
Re-approving.

PR #35585, opened 2025-12-19
[ROCm] Add fusion autotuning implementation for AMD GPUs

nurmukhametov:
AMD GPUs were missing the AddFusionAutotuningPass() override, causing PersistedAutotuningTest.SingleOperationGetsAutotuned to fail because no autotuning results were generated.

This change adds the same autotuning implementation to AMDGPUCompiler that NVPTXCompiler uses, enabling autotuning between native and block-level fusion emitters for transpose and reduce operations.

Fixes: PersistedAutotuningTest.SingleOperationGetsAutotuned on AMD GPUs


derdrdirk (approved):
LGTM

PR #35706, opened 2025-12-23
[XLA:GPU] Refactoring command buffer

shawnwang18:
This pull request refactors and modernizes the command buffer recording infrastructure in the GPU runtime by simplifying the command recording API, improving type safety with absl nullability annotations, and enhancing dependency management for command execution. The changes focus on making the codebase more robust, maintainable, and easier to reason about, especially around command buffer recording and dependencies between commands and executors.

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/backends/gpu/runtime/command_buffer_cmd.cc
@@ -156,9 +156,19 @@ static se::CommandBuffer::CreateCommands CreateCommands(
     const Thunk::ExecuteParams* execute_params,
     const CommandBufferCmd::RecordParams* record_params) {
   return [=](se::CommandBuffer* command_buffer,
-             absl::Span<const se::CommandBuffer::Command* const> dependencies) {
-    return commands->RecordCreate(*execute_params, *record_params,
-                                  command_buffer, dependencies);
+             absl::Span<const se::CommandBuffer::Command* const> dependencies)
+             -> absl::StatusOr<std::vector<const se::CommandBuffer::Command*>> {
+    auto new_record_params = *record_params;
+    new_record_params.command_buffer = command_buffer;
+    new_record_params.executor_dependencies.assign(dependencies.begin(),
+                                                   dependencies.end());
+    // Don't finalize from within CreateCommands callbacks - the parent
+    // (CreateWhile/CreateCase) handles finalization of nested command buffers.
+    new_record_params.is_finalize = false;
+    // Set the executor for dependency resolution within this executor.
+    new_record_params.executor = commands;
+    TF_RETURN_IF_ERROR(commands->Record(*execute_params, new_record_params));
+    return commands->SinkCommands(new_record_params);
   };
 }
 
@@ -181,8 +191,14 @@ static se::CommandBuffer::UpdateCommands UpdateCommands(
     const Thunk::ExecuteParams* execute_params,
     const CommandBufferCmd::RecordParams* record_params) {
   return [=](se::CommandBuffer* command_buffer) {
-    return commands->RecordUpdate(*execute_params, *record_params,
-                                  command_buffer);
+    auto new_record_params = *record_params;
+    new_record_params.command_buffer = command_buffer;
+    // Don't finalize from within UpdateCommands callbacks - the parent
+    // (UpdateWhile/UpdateCase) handles the nested command buffer lifecycle.
+    new_record_params.is_finalize = false;
+    // Set the executor for dependency resolution within this executor.
+    new_record_params.executor = commands;
+    return commands->Record(*execute_params, new_record_params);
   };
 }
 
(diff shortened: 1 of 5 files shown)


xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -156,9 +156,19 @@ static se::CommandBuffer::CreateCommands CreateCommands(
     const Thunk::ExecuteParams* execute_params,
     const CommandBufferCmd::RecordParams* record_params) {
   return [=](se::CommandBuffer* command_buffer,
-             absl::Span<const se::CommandBuffer::Command* const> dependencies) {
-    return commands->RecordCreate(*execute_params, *record_params,
-                                  command_buffer, dependencies);
+             absl::Span<const se::CommandBuffer::Command* const> dependencies)
+             -> absl::StatusOr<std::vector<const se::CommandBuffer::Command*>> {
+    auto new_record_params = *record_params;
```
ezhulenev: add type + don't use `new` in variable names
`CommandBufferCmd::RecordParams nested_record_params = `

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -181,8 +191,14 @@ static se::CommandBuffer::UpdateCommands UpdateCommands(
     const Thunk::ExecuteParams* execute_params,
     const CommandBufferCmd::RecordParams* record_params) {
   return [=](se::CommandBuffer* command_buffer) {
-    return commands->RecordUpdate(*execute_params, *record_params,
-                                  command_buffer);
+    auto new_record_params = *record_params;
```
ezhulenev: ditto

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -1298,15 +1134,15 @@ absl::StatusOr<const se::CommandBuffer::Command*> CustomKernelLaunchCmd::Record(
   se::KernelArgsDeviceMemoryArray kernel_args(
       buffers, custom_kernel_.shared_memory_bytes());
 
-  return Handle(
-      std::move(record_action),
-      [&](absl::Span<const se::CommandBuffer::Command* const> dependencies) {
-        return command_buffer->CreateLaunch(
+  return HandleCmdCreateOrUpdate(
+      record_params,
+      [&]() {
```
ezhulenev: `[&]() {` -->> `[&] {` in this file

xla/backends/gpu/runtime/command_buffer_cmd.h
```diff
@@ -277,24 +279,30 @@ class CommandBufferCmd {
...
-    absl::Span<const se::CommandBuffer::Command* const> dependencies;
-  };
+    // The command buffer that is recording the commands.
+    se::CommandBuffer* command_buffer = nullptr;
 
-  // Update previously recorded commands in the command buffer.
-  struct RecordUpdate {
-    const se::CommandBuffer::Command* command;
+    // The dependencies across CommandBufferCmdExecutor
+    std::vector<const se::CommandBuffer::Command*> executor_dependencies;
```
ezhulenev: Move it after executor and explain how exactly executor can end up with dependencies, I suspect it's because of nested control flow (If, While)?

Also maybe call it `external_dependencies`? 

"External dependencies that are defined by the parent command buffer executor, i.e. if `executor` defines command for a `while` operation, this will be a set of commands that must finish execution before the while operation starts"

If I got it correctly. Basically this is quite non-trivial part of command buffer recording and needs a bit more explanation.

xla/backends/gpu/runtime/command_buffer_cmd.h
```diff
@@ -277,24 +279,30 @@ class CommandBufferCmd {
...
     // unroll_iteration to locate the commands for current unroll iteration.
     int64_t unroll_iteration = 0;
-  };
 
-  // Create new commands in the command buffer using the given dependencies.
-  struct RecordCreate {
-    absl::Span<const se::CommandBuffer::Command* const> dependencies;
-  };
+    // The command buffer that is recording the commands.
+    se::CommandBuffer* command_buffer = nullptr;
```
ezhulenev: can you sprinke absl_nullable/absl_nonnull annotation in this file, I suspect that command buffer is always non-null, but hard to tell

xla/backends/gpu/runtime/command_buffer_cmd.h
shawnwang18: Add below comments, it is used when we do while cmd unrolling: 

   // External dependencies that source commands in this executor must wait on.
    // When multiple CommandBufferCmdExecutors are recorded into the same command
    // buffer (e.g., A -> B -> C), the source commands of executor B must depend
    // on the sink commands of executor A. These external dependencies are passed
    // in via this field to establish cross-executor ordering.
    //
    // Example: WhileCmd loop unrolling. When a WhileCmd has a known trip count,
    // we unroll the loop by recording cond_commands and body_commands executors
    // multiple times into a single command buffer:
    //   [cond_0] -> [body_0] -> [cond_1] -> [body_1] -> ... -> [cond_N] -> [body_N]
    // Each iteration's body_commands must wait on the preceding cond_commands,
    // and each cond_commands (except the first) must wait on the preceding
    // body_commands. The external_dependencies field carries these cross-executor
    // dependencies between unrolled iterations.

xla/backends/gpu/runtime/command_buffer_cmd.h
shawnwang18: Done

xla/backends/gpu/runtime/command_buffer_cmd.cc
shawnwang18: Fixed.

xla/backends/gpu/runtime/command_buffer_cmd.cc
shawnwang18: Fixed.

xla/backends/gpu/runtime/command_buffer_cmd.cc
shawnwang18: Fixed

ezhulenev (approved):

appujee (approved):

cota (approved):

PR #35735, opened 2025-12-24
[XLA:GPU][oneAPI] Add oneAPI compute capability interface

akhilgoe:
This PR adds an initial implementation of the `OneAPIComputeCapability` interface. Full implementation will follow in a subsequent PR.


xla/stream_executor/device_description.cc
```diff
@@ -140,9 +155,9 @@ bool ThreadDimOk(const DeviceDescription &device_description,
   return ok;
 }
 
-void CalculateDimensionality(const DeviceDescription &device_description,
-                             int64_t element_count, int64_t *threads_per_block,
-                             int64_t *block_count) {
+void CalculateDimensionality(const DeviceDescription& device_description,
```
cota: Please don't do unrelated style fixes in the PR -- it pollutes the "blame" history unnecessarily. Just do the changes you need. That said, if you're changing certain lines because you need them for your change, feel free to fix the style on those lines.

cota (changes requested):

xla/stream_executor/device_description.cc
akhilgoe: Thanks for the review! I reverted the file-wide clang-format changes

xla/stream_executor/device_description.h
```diff
@@ -32,16 +32,22 @@ limitations under the License.
 #include "xla/stream_executor/launch_dim.h"
 #include "xla/stream_executor/rocm/rocm_compute_capability.h"
 #include "xla/stream_executor/semantic_version.h"
+#include "xla/stream_executor/sycl/oneapi_compute_capability.h"
 
 namespace stream_executor {
 
+using sycl::OneAPIComputeCapability;
```
cota: There should be no "using ..." in header files.

Consider just dropping the sycl namespace for OneAPIComputeCapability -- AFAICT others, e.g. rocm, don't have it. If you have a good reason to keep it, you'll have to spell out sycl::OneAPIComputeCapability in all header files.

akhilgoe: Done

xla/stream_executor/sycl/oneapi_compute_capability.cc
```diff
@@ -0,0 +1,78 @@
...
+  // TODO(intel-tf): Add proper implementation for the proto message.
+  proto.set_variant("");
+  return proto;
+}
+
+std::string OneAPIComputeCapability::ToString() const {
+  return absl::StrCat(generation_, ".", version_);
+}
+
+const std::pair<uint32_t, uint32_t>
```
beckerhe: Nit: The `const` has no meaning here.

xla/stream_executor/sycl/oneapi_compute_capability.cc
```diff
@@ -0,0 +1,78 @@
...
+OneAPIComputeCapability::BaseVersionTupleFor(absl::string_view name) {
+  auto architecture = absl::AsciiStrToLower(name);
```
beckerhe: The style guide would like us to use the real type (std::string) here.

xla/stream_executor/sycl/oneapi_compute_capability.cc
```diff
@@ -0,0 +1,78 @@
...
+  } else if (architecture == "bmg") {
+    return {0x14, 0x1};
+  } else if (architecture == "dg2") {
+    return {0xc, 0x37};
+  } else {
+    return {0, 0};
+  }
+}
+
+const uint32_t OneAPIComputeCapability::GenericIPVersionFor(
```
beckerhe: Nit: The const here has no meaning

xla/stream_executor/sycl/oneapi_compute_capability.cc
```diff
@@ -0,0 +1,78 @@
...
+  auto architecture = absl::AsciiStrToLower(name);
+  if (architecture == "pvc") {
+    return {0xc, 0x3c};
+  } else if (architecture == "bmg") {
+    return {0x14, 0x1};
+  } else if (architecture == "dg2") {
+    return {0xc, 0x37};
+  } else {
+    return {0, 0};
+  }
```
beckerhe, line 60: We try to avoid no `else` after a `return` branch in the code base. Would be great if you could adjust this here as well.

xla/stream_executor/device_description.h
```diff
@@ -393,6 +419,10 @@ class DeviceDescription {
     gpu_compute_capability_ = RocmComputeCapability(std::move(gcn_arch_name));
   }
 
+  void set_oneapi_compute_capability(uint32_t ip_version) {
+    gpu_compute_capability_ = OneAPIComputeCapability(std::move(ip_version));
```
beckerhe: Nit: The `std::move` for a uint32_t is not doing anything. You can remove it.

xla/stream_executor/device_description.proto
```diff
@@ -70,7 +73,8 @@ message RuntimeVersionProto {
 message GpuTargetConfigProto {
   GpuDeviceInfoProto gpu_device_info = 1;
   reserved 2, 3;
-  reserved "cuda_compute_capability", "rocm_compute_capability";
+  reserved "cuda_compute_capability", "rocm_compute_capability",
+      "oneapi_compute_capability";
```
beckerhe: There is no need to add `oneapi` here. `reserved` just means that the field names mentioned here existed at some point and shouldn't be reintroduced to avoid accidental wrong interpretation of old protobufs. `oneapi_compute_capability` never existed, therefore this risk doesn't exist.

beckerhe (commented):
Thank you. In general looks good to me.

xla/stream_executor/sycl/oneapi_compute_capability.cc
akhilgoe: Done

xla/stream_executor/sycl/oneapi_compute_capability.cc
akhilgoe: Done

xla/stream_executor/sycl/oneapi_compute_capability.cc
akhilgoe: Done...I originally marked the return type as `const` to prevent accidental modification of constants. But because this function is private and only called by other private helpers, dropping the const qualifier should be safe I think.

xla/stream_executor/device_description.h
akhilgoe: Makes sense, thanks! Done

xla/stream_executor/device_description.proto
akhilgoe: Done

xla/stream_executor/sycl/oneapi_compute_capability.cc
akhilgoe: Done...similar to above, it should be fine here as well, since this private function is only called by a constructor.

akhilgoe:
Thanks @beckerhe and @cota for your reviews. I have added a commit to address your comments. Please take a look whenever you get time.

xla/stream_executor/sycl/oneapi_compute_capability.cc
cota: I still see auto here

cota (commented):
LGTM (modulo the nit I just commented on) although I'll leave it to @beckerhe to approve since he is more familiar with this part of XLA. Thanks!

xla/stream_executor/sycl/oneapi_compute_capability.cc
akhilgoe: Fixed!
Apologies, I thought you had asked to replace `absl::string_view` with `std::string`

cota (approved):
Approving since Henning is OOO today. We will have to approve internally later anyway.

akhilgoe:
@cota @beckerhe it seems there was a failure for a test on CPU due to missing mock override of the `GetOneAPIComputeCapability()` function. I have added a commit to fix it. Please take a look.

penpornk (approved):
Thank you for the fix! I'm approving the fix and rerunning the workflows.

akhilgoe:
Fixed conflicts with main, this might need a fresh :+1:, thanks!

PR #35740, opened 2025-12-25
Add xla_cpu_ftz flag to control flush-to-zero on CPU

### LINKED ISSUE #22858: Feature Request: An option to disable flushing to zero on CPU
I would like a way to disable the code below,
https://github.com/openxla/xla/blob/a41222c72b0774d05ec3ad4eb0fe22517272e133/xla/service/llvm_ir/llvm_util.cc#L718

Xla supports disabling fastmath (e.g. the `--xla_cpu_enable_fast_math` flag) and we seem to disable flushing to zeros in the code below. So controlling whether we set `denormal-fp-math` seems reasonable to me. https://github.com/openxla/xla/blob/a41222c72b0774d05ec3ad4eb0fe22517272e133/xla/service/llvm_compiler.cc#L41-L51.

Based on https://github.com/llvm/llvm-project/issues/81204#issue-2126249114 and https://groups.google.com/g/llvm-dev/c/TDGKHFU4hzE/m/k-LEa3NvBQAJ I don't believe on x86_64 that setting `denormal-fp-math` actually sets the floating point register and instead might only affect constant folding.

yashwantbezawada:
Fixes #22858

I ran into needing to control denormal handling on CPU similar to how
xla_gpu_ftz works for GPU. Currently the CPU backend unconditionally sets
the denormal-fp-math=preserve-sign attribute on all functions.

This adds an xla_cpu_ftz flag (field 493 in DebugOptions) that controls
whether this attribute gets applied in CreateCpuFunction. It defaults to
true so existing behavior is preserved. Users who want to disable
flush-to-zero can set --xla_cpu_ftz=false.

Usage:
```
XLA_FLAGS="--xla_cpu_ftz=false" ./program
```

### DIFF
--- a/xla/service/llvm_ir/llvm_util.cc
+++ b/xla/service/llvm_ir/llvm_util.cc
@@ -776,10 +776,11 @@ llvm::Function* CreateCpuFunction(llvm::FunctionType* function_type,
   // created by the JIT compiled code.
   function->setUWTableKind(llvm::UWTableKind::Default);
 
-  // Tensorflow always flushes denormals to zero, let LLVM know that flushing
-  // denormals is safe. This allows vectorization using ARM's neon instruction
-  // set.
-  function->addFnAttr("denormal-fp-math", "preserve-sign");
+  // Optionally flush denormals to zero. This allows vectorization using ARM's
+  // NEON instruction set. Controlled by the xla_cpu_ftz flag.
+  if (module_config.debug_options().xla_cpu_ftz()) {
+    function->addFnAttr("denormal-fp-math", "preserve-sign");
+  }
 
   // Add the optimize attribute to the function if optimizing for size. This
   // controls internal behavior of some optimization passes (e.g. loop
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -240,6 +240,9 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_cpu_fast_math_honor_functions(true);
   opts.set_xla_cpu_fast_math_honor_division(true);
 
+  // Default to true for backwards compatibility (always flush denormals).
+  opts.set_xla_cpu_ftz(true);
+
   opts.set_xla_gpu_fused_attention_use_cudnn_rng(false);
 
   // By default, copy TF's Eigen style min_max behavior with nans.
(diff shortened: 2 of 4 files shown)


xla/service/llvm_ir/llvm_util.cc
```diff
@@ -768,10 +768,11 @@ llvm::Function* CreateCpuFunction(llvm::FunctionType* function_type,
...
 
-  // Tensorflow always flushes denormals to zero, let LLVM know that flushing
-  // denormals is safe. This allows vectorization using ARM's neon instruction
-  // set.
-  function->addFnAttr("denormal-fp-math", "preserve-sign");
+  // Optionally flush denormals to zero. This allows vectorization using ARM's
+  // NEON instruction set. Controlled by the xla_cpu_ftz flag.
+  if (module_config.debug_options().xla_cpu_ftz()) {
+    function->addFnAttr("denormal-fp-math", "preserve-sign");
+  }
```
penpornk, line 783: Would you mind adding a unit test for this? E.g., set the flag to true/false and check for attributes.
https://github.com/openxla/xla/blob/189a63c1e028d7d0caf7dde17b7323ce480aabf7/xla/service/gpu/tests/gpu_codegen_test.cc#L42

penpornk (changes requested):
Thank you for the PR!

xla/service/llvm_ir/llvm_util.cc
yashwantbezawada: Added unit tests in llvm_util_test.cc. One test checks that the attribute gets added when the flag is true, another checks it's omitted when false.

yashwantbezawada:
Rebased onto latest main and resolved the merge conflict in the test file. The unit tests for the flag are still in place. Ready for another look when you get a chance.

dimitar-asenov:
@penpornk Could you please review this PR?

penpornk (commented):
Thank you for adding the test and I'm so sorry for the delay!

@seantalts is looking at CPU numerics and the flags to control them and is the better reviewer for this PR.

seantalts (approved):
lgtm but could you please rebase? ty!

yashwantbezawada:
@seantalts Rebased onto latest main. Field 443 got reserved in DebugOptions since I last updated this, so I moved xla_cpu_ftz to 493 and bumped the next id comment. The test file also moved to restricted/ on main, so the tests went along with it. No other changes, should be good to go now.

seantalts (approved):

PR #35750, opened 2025-12-26
[se] Add support for custom kernel argument packing via template specialization

ezhulenev:
Main changes are in `xla/stream_executor/kernel_args.h`.

* Consistently use `KernelArgPacking` to pack custom arguments to device kernels. 
* Add a test for overriding default packing mechanism via template specializations.
* Rename `KernelArgument` to `KernelArg` for consistency with other types
* Rename `KernelArgPackingRelocation::Type` to `Kind` for a bit more consistency with other parts of XLA, and because `Type` is used in many places to reference a C++ type
* Simplify how `KernelArgsPackedArray` stores POD arguments and remove a bunch of templates.

This change is needed to be able to pass NCCL device communicators to custom kernels without leaking NCCL headers from public XLA headers, and instead hiding kernel argument packing inside NCCL collectives library.


beckerhe (approved):
LGTM. Thanks for all the renames!

PR #35759, opened 2025-12-27
[ROCm] Remove ConvAlgorithmPicker

draganmladjenovic:
📝 Summary of Changes
Default to MIOpen immediate mode allowing regular convolution to work w/o autotuning and any workarounds in convolution_thunk. Remove fallback path for RocmFusedConvRunner and revert fused convs back to regular ones if no algorithm has been found during the autotuning. Implement miopen autotuning backend and remove ConvAlgorithmPicker as rocm was its sole user.

🎯 Justification
Solves one TODO on b/407494793

🚀 Kind of Contribution
♻️ Cleanup

🧪 Unit Tests:
Added //xla/service/gpu/transforms:cudnn_fused_conv_rewriter_autotune_disabled_test to stress out
fused conv decomposing path.


draganmladjenovic:
@xla-rotation @derdrdirk PTAL

xla/service/gpu/amdgpu_compiler.cc
```diff
@@ -245,35 +246,49 @@ absl::Status AMDGPUCompiler::AddConvAndGemmAutotuningPasses(
     se::StreamExecutor* stream_exec,
     const Compiler::GpuTargetConfig* target_config) {
   const DebugOptions& debug_options = hlo_module->config().debug_options();
+  // TODO(rocm) Add more uses of xla_gpu_experimental_disable_binary_libraries
```
derdrdirk, line 249: What are "more uses of xla_gpu_experimental-disable_binary_libraries"?

derdrdirk (approved):
LGTM. 

Thanks a lot for this awesome cleanup! -2315 lines :)

xla/service/gpu/amdgpu_compiler.cc
draganmladjenovic: This is the only place we check for xla_gpu_experimental-disable_binary_libraries in amdgpu_compiler, so it is kind of half broken. I think right now we would still use cublas etc., but not have autotuning instead of not using it at all.

derdrdirk: I see. Might make sense to add the corresponding TODO before the cublas backend.

draganmladjenovic:
@derdrdirk Do I need to rebase? I see copybara is doing something.

derdrdirk:
IIUC you introduced macros in `rocm_dnn.cc`, e.g. `#if (TF_ROCM_VERSION >= 70000)`, which blocked the submission and need to be manually submitted. I added an exemption and the PR should be submitted soon. 

Nit: "Within XLA, you can now use RETURN_IF_ERROR/ASSIGN_OR_RETURN from util/task/status_macros.h rather than their TSL TF_ equivalents."

PR #30036, opened 2025-08-11
[XLA:GPU] Add parent pointer for command buffer to track CommandBuffer/NestedCommandBuffer

shawnwang18:
This PR mainly made two changes to CommandBuffer:
1. introduce the `parent_` pointer in CudaCommandBuffer to point to the parent.
2. When calling CreateChildNode, auto link nested CudaCommandBuffer with parent CommandBuffer by setting the `parent_` pointer. 
3. When updating nodes in nested command buffer, the graph executor can be fected through traversaling the `parent_` pointer to the top level CudaCommandBuffer. 

This is a simplified design, and made possible for CommandBufferCmd to have NestedCmd (introduced in other PRs), where the nested command sequence can be updated explicitly (not through tracing).

PR #35760, opened 2025-12-27
[XLA:GPU] Set parent_ pointer for conditional command buffer nodes

dstoutamire:
# Summary

PR #30036 introduced the `parent_` pointer to track parent-child relationships
between command buffers. However, `CreateConditionalNode` was not updated to
set this pointer, causing crashes on the second execution of while loops when
command buffers are enabled.

## Bug Description

**Symptoms:**
- First execution of `lax.while_loop` with command buffers enabled works
- Second execution crashes with: `graph_exec_ is nullptr for top level cuda command buffer`

**Crash flow:**
1. `CreateWhile` creates a conditional node with a nested command buffer
2. First execution: graph is created and finalized successfully
3. Second execution: `UpdateWhile` → `body->Update()` → `CheckCanBeUpdated()`
4. `graph_exec()` walks up the `parent_` chain to find the instantiated graph
5. But `parent_` is nullptr, so it stays at the nested buffer level
6. Nested buffer's `graph_exec_` is nullptr (only top-level buffers call `InstantiateGraph`)
7. CHECK fails

## Fix

Set `nested_buffer->parent_ = this` in `CreateConditionalNode`, matching what
`CreateChildNode` already does (line 428).

## Testing

Reproducer (requires CUDA 12.3+ and `--xla_gpu_enable_command_buffer=WHILE`):

```python
import jax
import jax.numpy as jnp
import jax.lax as lax

def while_loop_fn(x):
    def cond(state):
        i, val = state
        return i < 10
    def body(state):
        i, val = state
        return (i + 1, val + x)
    init = (jnp.array(0), jnp.array(0.0))
    _, result = lax.while_loop(cond, body, init)
    return result

compiled = jax.jit(while_loop_fn).lower(jnp.array(1.0)).compile()
result1 = compiled(jnp.array(1.0))  # First execution: OK
result1.block_until_ready()
result2 = compiled(jnp.array(1.0))  # Second execution: CRASH without fix
result2.block_until_ready()
```

## Related

- PR #30036: Introduced `parent_` pointer but missed `CreateConditionalNode`
- `CreateChildNode` (line 421-428) correctly sets `parent_`

### DIFF
--- a/xla/stream_executor/cuda/cuda_command_buffer.cc
+++ b/xla/stream_executor/cuda/cuda_command_buffer.cc
@@ -275,12 +275,18 @@ CudaCommandBuffer::CreateConditionalNode(
   VLOG(2) << "Created conditional CUDA graph "
           << cu_params.conditional.phGraph_out[0];
 
+  // Create the nested command buffer and set its parent so that graph_exec()
+  // can walk up the parent chain to find the instantiated graph.
+  // Without this, UpdateWhile crashes with "graph_exec_ is nullptr".
+  auto* nested_buffer = new CudaCommandBuffer(
+      Mode::kNested, stream_exec_, cuda_context_,
+      cu_params.conditional.phGraph_out[0],
+      /*is_owned_graph=*/false);
+  nested_buffer->parent_ = this;
+
   return GraphConditionalNodeHandle{
       FromCudaGraphHandle(node_handle),
-      std::unique_ptr<CudaCommandBuffer>(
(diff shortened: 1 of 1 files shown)


dstoutamire:
Closing this PR as the fix has been incorporated via #36135.

I verified that the main branch now passes the original reproduction test - multiple executions of while loops with --xla_gpu_enable_command_buffer=WHILE complete successfully without the "graph_exec_ is nullptr" crash.

Thanks to the XLA team for addressing this!

PR #35762, opened 2025-12-28
Fix permutation sort expander dtype mismatch

samwithwicky:
---

### 🐛 Fix scatter update dtype handling in permutation sort expander

The permutation sort expander previously hardcoded `S32` when constructing the
scatter update computation. This caused an invalid update computation when the
updates were `S64`, leading to HLO verifier failures on GPU.

This change constructs the update computation scalar shape using the **actual
element type of the updates**, instead of assuming `S32`.

---

### 📝 Summary of Changes

* Fixes a dtype mismatch in the permutation sort expander by deriving the scatter
  update computation scalar shape from the update element type.
* Adds a unit test to verify that the scatter update computation parameters use
  the same element type as the updates (e.g. `S64`).

---

### 🎯 Justification

The permutation sort expander previously assumed `S32` scalars when building the
scatter update computation. When the updates are `S64` (for example, when using
64-bit indices), this resulted in an invalid update computation and triggered an
HLO verifier failure on GPU.

This fix ensures the update computation matches the actual update element type,
restoring correctness while preserving existing behavior for the default `S32`
case.

---

### 🚀 Kind of Contribution

🐛 Bug Fix

---

### 🧪 Unit Tests

* Added a new unit test that explicitly checks that the scatter update computation
  parameters use the same element type as the updates.
* The test fails with the previous hardcoded `S32` behavior and passes with this
  fix.

---

### 🧪 Execution Tests

* No new execution tests added.
* The fix is validated by existing HLO verifier checks during compilation.

### DIFF
--- a/xla/hlo/transforms/expanders/permutation_sort_expander.cc
+++ b/xla/hlo/transforms/expanders/permutation_sort_expander.cc
@@ -131,7 +131,9 @@ absl::StatusOr<HloInstruction*> PermutationSortExpander::ExpandInstruction(
   // Construct the updates computation, which simply replaces the operand
   // values with the update values.
   HloComputation::Builder b("update_replace_computation");
-  Shape scalar_shape = ShapeUtil::MakeShape(S32, {});
+  // Use the element type of the updates to avoid S32/S64 mismatches under x64.
+  PrimitiveType elem_ty = update_shape.element_type();
+  Shape scalar_shape = ShapeUtil::MakeShape(elem_ty, {});
   b.AddInstruction(
       HloInstruction::CreateParameter(0, scalar_shape, "scalar_lhs"));
   HloInstruction* scalar_rhs = b.AddInstruction(
--- a/xla/hlo/transforms/expanders/permutation_sort_expander_test.cc
+++ b/xla/hlo/transforms/expanders/permutation_sort_expander_test.cc
@@ -169,40 +169,71 @@ TEST_F(PermutationSortExpanderTest, DontReplacePermutationSortIfNonIntegral) {
   EXPECT_THAT(PermutationSortExpander().Run(module.get()), IsOkAndHolds(false));
 }
 
-TEST_F(PermutationSortExpanderTest, DontReplacePermutationSortWrongDimensions) {
-  // Same as ReplacePermutationSortWithScatter except that the sort dimensions
-  // don't match.
+TEST_F(PermutationSortExpanderTest,
+       ScatterUpdateComputationUsesSameElementTypeAsUpdates) {
   const char* hlo_string = R"(
-   HloModule permutation_sort
+    HloModule permutation_sort_int64
 
-    lt_f32 {
-      p.0.lhs = f32[] parameter(0)
-      p.0.rhs = f32[] parameter(1)
-      p.1.lhs = s32[] parameter(2)
-      p.1.rhs = s32[] parameter(3)
-      ROOT lt = pred[] compare(p.0.lhs, p.0.rhs), direction=LT, type=TOTALORDER
+    lt_f32_s64 {
+      k_lhs = f32[] parameter(0)
+      k_rhs = f32[] parameter(1)
+      v_lhs = s64[] parameter(2)
+      v_rhs = s64[] parameter(3)
+      ROOT lt = pred[] compare(k_lhs, k_rhs), direction=LT, type=TOTALORDER
     }
 
-    lt_s32 {
-      p.0.lhs = s32[] parameter(0)
-      p.0.rhs = s32[] parameter(1)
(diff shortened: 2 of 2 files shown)


xla/hlo/transforms/expanders/permutation_sort_expander.cc
```diff
@@ -131,7 +131,9 @@ absl::StatusOr<HloInstruction*> PermutationSortExpander::ExpandInstruction(
   // Construct the updates computation, which simply replaces the operand
   // values with the update values.
   HloComputation::Builder b("update_replace_computation");
-  Shape scalar_shape = ShapeUtil::MakeShape(S32, {});
+  // Use the element type of the updates to avoid S32/S64 mismatches under x64.
```
penpornk, line 134: Could you please help add a new unit test to catch this behavior in this PR?

penpornk (changes requested):
Thank you for the fix!

xla/hlo/transforms/expanders/permutation_sort_expander.cc
samwithwicky: Done — I’ve added a unit test. Thanks!

xla/hlo/transforms/expanders/permutation_sort_expander_test.cc
```diff
@@ -169,40 +169,71 @@ TEST_F(PermutationSortExpanderTest, DontReplacePermutationSortIfNonIntegral) {
   EXPECT_THAT(PermutationSortExpander().Run(module.get()), IsOkAndHolds(false));
 }
 
-TEST_F(PermutationSortExpanderTest, DontReplacePermutationSortWrongDimensions) {
```
penpornk, line 172: I don't think we need to remove this existing test case. This and the new one test different things. Please keep this test case?

xla/hlo/transforms/expanders/permutation_sort_expander_test.cc
```diff
@@ -169,40 +169,71 @@ TEST_F(PermutationSortExpanderTest, DontReplacePermutationSortIfNonIntegral) {
...
 
     ENTRY sort_computation {
-      keys = f32[64,8732]{1,0} parameter(0)
-      values = s32[64,8732]{1,0} iota(), iota_dimension=1
-      sort = (f32[64,8732]{1,0}, s32[64,8732]{1,0}) sort(keys, values), dimensions={1}, to_apply=lt_f32
-      gte = s32[64,8732]{1,0} get-tuple-element(sort), index=1
-      ROOT sort2 = (s32[64,8732]{1,0}, s32[64,8732]{1,0}) sort(gte, values), dimensions={0}, to_apply=lt_s32
+      keys = f32[10]{0} parameter(0)
+      values = s64[10]{0} iota(), iota_dimension=0
+
```
penpornk, line 196: Nit: We can remove the blank lines around the `sort1` command.

xla/hlo/transforms/expanders/permutation_sort_expander_test.cc
```diff
@@ -169,40 +169,71 @@ TEST_F(PermutationSortExpanderTest, DontReplacePermutationSortIfNonIntegral) {
...
+      sort1 = (f32[10]{0}, s64[10]{0})
+          sort(keys, values), dimensions={0}, to_apply=lt_f32_s64
+
+      perm = s64[10]{0} get-tuple-element(sort1), index=1
+
```
penpornk, line 201: Nit: We can remove the blank line here too.

xla/hlo/transforms/expanders/permutation_sort_expander_test.cc
```diff
@@ -169,40 +169,71 @@ TEST_F(PermutationSortExpanderTest, DontReplacePermutationSortIfNonIntegral) {
...
+  HloInstruction* scatter = nullptr;
+  for (HloComputation* comp : module->computations()) {
+    for (HloInstruction* instr : comp->instructions()) {
+      if (instr->opcode() == HloOpcode::kScatter) {
+        scatter = instr;
+        break;
+      }
+    }
+  }
```
penpornk, line 222: Nit: You can use `FindInstruction` for this.
https://github.com/openxla/xla/blob/189a63c1e028d7d0caf7dde17b7323ce480aabf7/xla/hlo/testlib/hlo_hardware_independent_test_base.h#L73-L75

penpornk (changes requested):
Thank you very much! I have a few minor nits.

hawkinsp:
Any updates?

dimitar-asenov:
@samwithwicky Could you please address the latest review comments?

dimitar-asenov:
Nevermind, I implemented the fixes and merged the PR. Thanks for fixing the type mismatch.

PR #35784, opened 2025-12-30
[XLA:GPU] phase-1 moving CommandBufferCmd to Thunk

shawnwang18:
PR design doc: https://docs.google.com/document/d/1MHTeuK8mUtrcPtWZzqBzcMJ7ve8AQjS74cP_8gg2cx0/edit?usp=sharing

### DIFF
--- a/xla/backends/gpu/runtime/BUILD
+++ b/xla/backends/gpu/runtime/BUILD
@@ -81,6 +81,58 @@ xla_cc_test(
 )
 
 cc_library(
+    name = "command_buffer_params",
+    hdrs = ["command_buffer_params.h"],
+    deps = [
+        ":command_state",
+        "//xla/service:buffer_assignment",
+        "//xla/stream_executor:command_buffer",
+        "//xla/tsl/lib/gtl:int_type",
+        "@com_google_absl//absl/base:nullability",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/functional:function_ref",
+    ],
+)
+
+# Base class for commands that can be recorded into command buffers.
+# Also includes CommandBufferCmdSequence and CommandBufferCmdExecutor.
+cc_library(
+    name = "command_thunk",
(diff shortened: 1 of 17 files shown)


xla/backends/gpu/runtime/BUILD
```diff
@@ -57,9 +57,51 @@ cc_library(
...
+        "//xla/service:buffer_assignment",
+        "//xla/stream_executor:command_buffer",
+        "//xla/tsl/lib/gtl:int_type",
+        "@com_google_absl//absl/base:nullability",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/functional:function_ref",
+    ],
+)
+
+# Header-only target for forward declarations and class definitions.
```
ezhulenev: This is bad. Is this temporary? Do you see how this ciruclar dep will be resolved in the end?

xla/backends/gpu/runtime/thunk.h
```diff
@@ -356,26 +362,53 @@ class Thunk {
 
     int64_t execution_id = 0;
 
+    mutable CommandBufferParams* record_params = nullptr;
+
+    // RAII helper to temporarily change record_params and automatically
+    // restore it when the scope exits.
+    class ScopedCommandBufferParams {
```
ezhulenev, line 366: Instead of this hack with `mutable` it should be possible to add something like:


```
ExecuteParams ExecuteParams::CopyWith(CommandBufferParams*);
```

all members are cheap to copy and it's less confusing

xla/backends/gpu/runtime/BUILD
shawnwang18: Removed the circular dependency in latest commit.

xla/backends/gpu/runtime/thunk.h
shawnwang18: Let me think a bit more whether any other good ways.

PR #35790, opened 2025-12-30
fix(graph_dumper): Fix HLO HTML dump segfaults after gemm_fusion_autotuner

QZero233:
📝 Summary of Changes
- There is a segfault bug when try to dump html after gemm_fusion_autotuner (See #35786 )
- This segfault happens when `IsFused()` returns `true` while its `FusionInstruction()` is still `nullptr`
- To fix this problem, we add a new function `IsFusedStrict()`, which will return `true` iff `IsFused()` is `true` and `FusionInstruction()` is not `nullptr`.
- We also changed the code in `hlo_graph_dumper.cc`, to adopt our `IsFusedStrict()`

This change will **only affect `hlo_graph_dumper.cc`**, leaving elsewhere untouched.

🎯 Justification
HLO HTML dump will not work  in some cases without this fix (detailed case see #35786 )

🚀 Kind of Contribution
- 🐛 Bug Fix

### DIFF
--- a/xla/hlo/ir/hlo_instruction.cc
+++ b/xla/hlo/ir/hlo_instruction.cc
@@ -4465,11 +4465,6 @@ bool HloInstruction::IsFused() const {
   return parent_ != nullptr && parent_->IsFusionComputation();
 }
 
-bool HloInstruction::IsFusedStrict() const {
-  return parent_ != nullptr && parent_->IsFusionComputation() &&
-         parent_->FusionInstruction() != nullptr;
-}
-
 bool HloInstruction::IsCustomCall(absl::string_view target) const {
   return opcode() == HloOpcode::kCustomCall && custom_call_target() == target;
 }
--- a/xla/hlo/ir/hlo_instruction.cc
+++ b/xla/hlo/ir/hlo_instruction.cc
@@ -4465,6 +4465,11 @@ bool HloInstruction::IsFused() const {
   return parent_ != nullptr && parent_->IsFusionComputation();
 }
 
+bool HloInstruction::IsFusedStrict() const {
+  return parent_ != nullptr && parent_->IsFusionComputation() &&
+         parent_->FusionInstruction() != nullptr;
+}
+
 bool HloInstruction::IsCustomCall(absl::string_view target) const {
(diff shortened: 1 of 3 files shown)


xla/hlo/ir/hlo_instruction.cc
```diff
@@ -4465,6 +4465,11 @@ bool HloInstruction::IsFused() const {
   return parent_ != nullptr && parent_->IsFusionComputation();
 }
 
+bool HloInstruction::IsFusedStrict() const {
```
olegshyshkov: I would make this a private helper function inside `hlo_graph_dumper.cc`, because `IsFusedStrict` has semantics very specific to the use case and I'm afraid that it will be misused (and I think it's already misused in this PR, see my comment in `hlo_graph_dumper`).

xla/service/hlo_graph_dumper.cc
```diff
@@ -1236,7 +1236,7 @@ ColorScheme HloDotDumper::GetInstructionColor(const HloInstruction* instr) {
     case HloOpcode::kBitcast:
       // Unfused bitcast is free, but fused bitcast should count as a non-free
       // data-movement op (e.g. requires linearization of indices on GPU).
-      if (!instr->IsFused()) {
+      if (!instr->IsFusedStrict()) {
```
olegshyshkov: I think this should remain `IsFused`. From the perspective of the bitcast, it doesn't matter if the computation has a caller instruction, it's still a part of fusion computation. If we have unattached computation, that just means that it's unreachable and should be deleted, but colors should be the same.

olegshyshkov (changes requested):

xla/hlo/ir/hlo_instruction.cc
QZero233: Thanks for your comment, I've moved this function inside `hlo_graph_dumper.cc` as a helper function in the newest version.

xla/service/hlo_graph_dumper.cc
QZero233: Sorry for the misuse. I've checked the scenarios where strict mode is applied in the latest version. 
If there are any other errors, please let me know.

olegshyshkov (approved):

PR #35808, opened 2025-12-31
[xla:gpu] Extract Command, CommandState and CommandExecutor into separate targets

ezhulenev:
`CommandBufferCmd` is too big and complicated, split it into separate targets with a long term plan to unify commands and thunks:

* `Command` - new base class for all commands
* `CommandState` - state management for commands
* `CommandExecutor` - takes a sequence of commands and records them into the command buffer (or updates them)

This PR doesn't have any functional changes, it simply splits `command_buffer_cmd` into 4 separate build targets and avoids circular dependencies between them by adding new command abstractions (`AsyncStartCommand` and `AsyncDoneCommand`)


xla/backends/gpu/runtime/command_buffer_cmd.cc
shawnwang18: I am wondering if this is correct with your new StateManager here. child_commands_ should be recorded into a separate command buffer, which is created by cuda driver. But in you code, I understand that you are actually recording to the parent graph?

shawnwang18: Ah, it should be correct.

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -2840,7 +2807,7 @@ absl::StatusOr<const se::CommandBuffer::Command*> DynamicSliceFusionCmd::Record(
                       execute_params.stream->parent()->CreateCommandBuffer(
                           se::CommandBuffer::Mode::kNested));
 
-  StateManager state;
+  CommandStateManager state(command_buffer);
```
shawnwang18, line 2126: Do we actually need to create a new StateManager here? Is there any conflict chance for any CommandBufferCmd ptr from embedded_commands_ conflicts with commands in the parent sequences?

ezhulenev: I'm pretty sure that no, but didn't want to make any functional changes in this PR.

PR #35868, opened 2026-01-04
Fix crashes due to kernel cache confusion. (#33220)

### LINKED ISSUE #33220: [XLA:GPU] Jax example produces CUDA_ERROR_INVALID_VALUE: invalid argument
Original bug: https://github.com/jax-ml/jax/issues/32882

The following JAX program results in XLA:GPU generating a kernel with invalid arguments:

```
import jax
import jax.numpy as jnp
import numpy as np


def sink(*args, **kwargs):
    jax.experimental.io_callback(lambda *args, **kwargs: None, None, *args, **kwargs)


@jax.jit
def main(int_value, input_params, slice_index):
    def inner_fn(shape, params):
        h1_dims = (shape[0], max(shape[1], 16))
        h1 = jnp.concatenate(
            [
                jnp.ones((*h1_dims, 127)) * jnp.array([0.0])[0],
                jnp.zeros((*h1_dims, 1)),
            ],
            axis=-1,
        )
        h2 = params["p1"] * h1
        h3 = h2 / (h2**2).sum(axis=-1, keepdims=True)
        h4 = params["p2"] / (jnp.linalg.norm(params["p2"], axis=-1, keepdims=True))
        h5 = h4[:, :-1].reshape((-1,))
        h6 = h3 * (np.zeros((128,)) * h5)
        return jax.lax.dynamic_slice_in_dim(h6, slice_index, shape[1], axis=-2)

    def outer_fn_1():
        sink(inner_fn((1, 1), input_params))

    def outer_fn_2():
        def loss_fn(params):
            return inner_fn((8, 17), params).sum()

        sink(jax.value_and_grad(loss_fn)(input_params))

    def main_loop_fn(*args, **kwargs):
        jax.lax.cond(int_value != 0, outer_fn_1, outer_fn_1)
        jax.lax.cond(int_value != 0, outer_fn_2, outer_fn_2)

    jax.lax.fori_loop(0, 1, main_loop_fn, None)


main(0, {"p1": jnp.zeros((1,)), "p2": jnp.zeros((32, 5))}, jnp.array(0))
```

Result:
```
E1027 14:49:00.537894  832649 pjrt_stream_executor_client.cc:2839] Execution of replica 0 failed: INTERNAL: CUDA error: Failed to add kernel node to a CUDA graph: CUDA_ERROR_INVALID_VALUE: invalid argument
  File .../jax_bug.py", line 44, in <module>
    main(0, {"p1": jnp.zeros((1,)), "p2": jnp.ones((32, 5))}, jnp.array(0))
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
jaxlib.xla_extension.XlaRuntimeError: INTERNAL: CUDA error: Failed to add kernel node to a CUDA graph: CUDA_ERROR_INVALID_VALUE: invalid argument
--------------------
For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

System Information:
```
jax:    0.8.0
jaxlib: 0.8.0
numpy:  2.3.4
python: 3.13.9 | packaged by Anaconda, Inc. | (main, Oct 21 2025, 19:16:10) [GCC 11.2.0]
device info: NVIDIA GeForce RTX 3080 Ti-1, 1 local devices"
process_count: 1
platform: uname_result(system='Linux', node='...', release='6.8.0-85-generic', version='#85~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 19 16:18:59 UTC 2', machine='x86_64')
$ nvidia-smi
Mon Oct 27 14:45:38 2025       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.95.05              Driver Version: 580.95.05      CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
```

nshepperd:
Use cached launch dimensions in EmitterBase and add the emitter class name to the discriminator when the accessing kernel cache. Fixes https://github.com/openxla/xla/issues/33220. 

To recap, sharing of the kernel cache between TritonFusion and EmitterBase (which handles loop fusions) allows for the different kernel emitters to return each other's kernels. In #33220 this was observed in the form of crashes when the launch dimensions were different -- because EmitterBase returns its own launch dimensions, not those that the cached kernel was compiled against. In most cases this simply causes a runtime error because the generated ptx has an assertion for the launch dimensions inserted into it, but could also manifest as segfaults or data corruption if the CTA shapes happen to be the same. 

I made two changes to fix this:
1. EmitterBase uses the cached launch dimensions. This just makes sense, since we're using the cached kernel.
2. Both EmitterBase and TritonFusion add their names to the discriminator string, which prevents more exotic crashes due to reasons such as the different handling of cluster_dims or tma objects (I don't know how to cause those crashes on demand though). But this does mean that an identical HLO computation can potentially end up being compiled twice, if it's used in both loop and triton fusion contexts.

I also added a unit test that reliably reproduces the issue, on cuda devices:

```
E0000 00:00:1767504254.670422  599411 tfrt_gpu_executable.cc:729] Calling RunAsync failed for executable test_module on device cuda:0, status = INTERNAL: CUDA error: Failed to launch CUDA kernel: triton_result; block dims: 128x1x1; grid dims: 1x1x1; shared memory size: 0: CUDA_ERROR_INVALID_VALUE: invalid argument
xla/backends/gpu/codegen/emitters/kernel_cache_test.cc:87: Failure
Value of: RunAndCompareNoHloPasses(kHloText, ErrorSpec{ 1e-6, 1e-6})
  Actual: false (INTERNAL: CUDA error: Failed to launch CUDA kernel: triton_result; block dims: 128x1x1; grid dims: 1x1x1; shared memory size: 0: CUDA_ERROR_INVALID_VALUE: invalid argument)
Expected: true
```

### DIFF
--- a/xla/backends/gpu/codegen/emitters/BUILD
+++ b/xla/backends/gpu/codegen/emitters/BUILD
@@ -1,5 +1,6 @@
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 load("//xla:xla.default.bzl", "xla_cc_test")
+load("//xla/tests:build_defs.bzl", "xla_test")
 
 package(
     # copybara:uncomment default_applicable_licenses = ["//tensorflow:license"],
--- a/xla/backends/gpu/codegen/emitters/mlir_kernel_emitter.cc
+++ b/xla/backends/gpu/codegen/emitters/mlir_kernel_emitter.cc
@@ -378,7 +378,8 @@ absl::StatusOr<FusionEmissionResult> MlirKernelFusion::Emit(
                                   GetDefaultBufferAlignment(), &fusion));
   auto [future_entry, cached] = ir_emitter_context.kernel_cache().GetWithStatus(
       fusion.fused_instructions_computation(), args.args(),
-      /*discriminator=*/"", [&]() -> xla::Future<KernelReuseCache::Entry> {
+      /*discriminator=*/"MlirKernelFusion",
+      [&]() -> xla::Future<KernelReuseCache::Entry> {
         std::string kernel_name = ir_emitter_context.GetSanitizedUniqueName(
             std::string(fusion.name()));
         return EmitLlvmModule(fusion, kernel_name, ir_emitter_context)
--- a/xla/backends/gpu/codegen/triton/fusion.cc
+++ b/xla/backends/gpu/codegen/triton/fusion.cc
@@ -217,7 +217,7 @@ absl::StatusOr<TritonFusion::EmitResult> TritonFusion::Emit(
   auto [status_or_entry, was_cached] =
       ir_emitter_context.kernel_cache().GetWithStatus(
           hlo_computation, kernel_arguments.args(),
-          /*discriminator=*/"", generate);
+          /*discriminator=*/"TritonFusion", generate);
   ASSIGN_OR_RETURN(const KernelReuseCache::Entry* entry,
                    status_or_entry.Await());
   return EmitResult{
--- a/xla/backends/gpu/tests/BUILD
+++ b/xla/backends/gpu/tests/BUILD
@@ -713,6 +713,19 @@ xla_test(
 )
 
 xla_test(
+    name = "kernel_cache_regression_test",
+    srcs = ["kernel_cache_regression_test.cc"],
+    backends = ["gpu"],
+    deps = [
+        "//xla:error_spec",
+        "//xla/tests:hlo_pjrt_interpreter_reference_mixin",
+        "//xla/tests:hlo_pjrt_test_base",
+        "@com_google_absl//absl/strings:string_view",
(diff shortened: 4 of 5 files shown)


nshepperd:
(Removed the superfluous second unit test I had added to test the testing.)

akuegel:
Sorry, it seems this PR fell through the cracks. If this fix is still needed, can you please rebase this PR?

pifon2a (approved):

akuegel:
@nshepperd Can you please rebase again?

akuegel (approved):

PR #36086, opened 2026-01-08
[xla:gpu] Remove accidental always-on loop unrolling

ezhulenev:
`enable_loop_unroll_` should not be hardcoded to always true. This was hiding a segfault in one of the command buffer tests.

### DIFF
--- a/xla/service/gpu/tests/command_buffer_test.cc
+++ b/xla/service/gpu/tests/command_buffer_test.cc
@@ -623,6 +623,10 @@ ENTRY main.49 {
 }
 
 TEST_P(CommandBufferTest, DynamicSliceCopyFusionCmd) {
+  if (GpuExecutor() != nullptr) {
+    GTEST_SKIP() << "This test leads to segfault in CommandBuffer";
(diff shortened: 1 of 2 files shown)


xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -623,6 +623,10 @@ ENTRY main.49 {
 }
 
 TEST_P(CommandBufferTest, DynamicSliceCopyFusionCmd) {
+  if (GpuExecutor() != nullptr) {
+    GTEST_SKIP() << "This test leads to segfault in CommandBuffer";
```
apivovarov: Should we open an issue and add TODO with Issue number here?

ezhulenev: https://github.com/openxla/xla/issues/36087

apivovarov: Thank you, Eugene!

apivovarov (approved):

PR #36135, opened 2026-01-08
Fix test CommandBufferTest.DynamicSliceCopyFusionCmd  failure

shawnwang18:
There are 3 bugs in XLA which accidently makes test CommandBufferTest.DynamicSliceCopyFusionCmd passed.

1. When lowering the command buffer WhileCmd, the nested body command buffer is not setting its parent properly, this leads to assertion failure when doing the update. 

2. In WhileCmd::Initialize(), enable_loop_unroll_ is hardcoded to 1, which was original for debug purpose, but forget to remove it. 

3. The command_buffer_conversion_pass is not passing the debug flags xla_gpu_command_buffer_unroll_loops down to the command_buffer_cmd_emitter. 

This PR fixed the above issue, and the test PASSed as expected.

### DIFF
--- a/xla/stream_executor/cuda/cuda_command_buffer.cc
+++ b/xla/stream_executor/cuda/cuda_command_buffer.cc
@@ -26,6 +26,7 @@ limitations under the License.
 #include "absl/functional/any_invocable.h"
 #include "absl/log/check.h"
 #include "absl/log/log.h"
+#include "absl/memory/memory.h"
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/str_cat.h"
@@ -275,12 +276,13 @@ CudaCommandBuffer::CreateConditionalNode(
   VLOG(2) << "Created conditional CUDA graph "
           << cu_params.conditional.phGraph_out[0];
 
-  return GraphConditionalNodeHandle{
-      FromCudaGraphHandle(node_handle),
-      std::unique_ptr<CudaCommandBuffer>(
-          new CudaCommandBuffer(Mode::kNested, stream_exec_, cuda_context_,
-                                cu_params.conditional.phGraph_out[0],
-                                /*is_owned_graph=*/false))};
+  auto nested_cmd_buffer = absl::WrapUnique(new CudaCommandBuffer(
+      Mode::kNested, stream_exec_, cuda_context_,
+      cu_params.conditional.phGraph_out[0], /*is_owned_graph=*/false));
+  nested_cmd_buffer->parent_ = this;
+
+  return GraphConditionalNodeHandle{FromCudaGraphHandle(node_handle),
+                                    std::move(nested_cmd_buffer)};
 #else
   return absl::UnimplementedError("unsupported node type");
 #endif  // CUDA_VERSION >= 12030
--- a/xla/backends/gpu/runtime/command_buffer_cmd.cc
+++ b/xla/backends/gpu/runtime/command_buffer_cmd.cc
@@ -1523,7 +1523,11 @@ absl::Status WhileCmd::Initialize(const Thunk::InitializeParams& params,
       cond_commands_.support_loop_unroll() && trip_count_ != std::nullopt) {
     is_unrolled_loop_ = true;
   }
-  VLOG(3) << "while command trip_count: " << trip_count_.value_or(-1);
+  VLOG(3) << "WhileCmd::Initialize: enable_loop_unroll_=" << enable_loop_unroll_
+          << ", body_support=" << body_commands_.support_loop_unroll()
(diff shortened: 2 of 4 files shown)


xla/stream_executor/cuda/cuda_command_buffer.cc
```diff
@@ -275,12 +275,15 @@ CudaCommandBuffer::CreateConditionalNode(
   VLOG(2) << "Created conditional CUDA graph "
           << cu_params.conditional.phGraph_out[0];
 
+  auto* nested_cmd_buffer =
+      new CudaCommandBuffer(Mode::kNested, stream_exec_, cuda_context_,
```
apivovarov: While your use of new followed by a `std::unique_ptr` constructor is technically correct, it’s generally safer to use `absl::WrapUnique` or `std::make_unique`
```cpp
auto nested_cmd_buffer = absl::WrapUnique(new CudaCommandBuffer(
      Mode::kNested, stream_exec_, cuda_context_,
      cu_params.conditional.phGraph_out[0],
      /*is_owned_graph=*/false));
  nested_cmd_buffer->parent_ = this;

  return GraphConditionalNodeHandle{
      FromCudaGraphHandle(node_handle),
      std::move(nested_cmd_buffer)};
```

xla/stream_executor/cuda/cuda_command_buffer.cc
```diff
@@ -275,12 +275,15 @@ CudaCommandBuffer::CreateConditionalNode(
...
+                            cu_params.conditional.phGraph_out[0],
+                            /*is_owned_graph=*/false);
+  nested_cmd_buffer->parent_ = this;
```
ezhulenev, line 282: Prefer passing parent via constructor vs mutation

ezhulenev: In general is is_owned_graph==false then command buffer must have a parent that also is and owner, because all command buffers from a tree structure, with topmost graph the owner of all nested command buffers + owner of the graph executable

xla/stream_executor/cuda/cuda_command_buffer.cc
shawnwang18: Thanks, fixed the code to use std::make_unique here.

xla/stream_executor/cuda/cuda_command_buffer.cc
shawnwang18: Changing the constructor will affect many files across the whole class hierarchy, I would like to send a nother PR for this change to make this PR solution more easy to get.

ezhulenev: Why would it? Just make `parent = nullptr, is_owned_graph = false` default constructor, and `CudaCommandBuffer` created only in one file. Is it even possible to be is_owned == true && parent != nullptr? I think `is_owned_graph_` is a redundant field.

xla/stream_executor/cuda/cuda_command_buffer.cc
ezhulenev: `absl::WrapUnique` is more widely used in stream executor code base where we want to hide constructors

xla/stream_executor/cuda/cuda_command_buffer.cc
shawnwang18: This is the PR for adding parent to constructor: https://github.com/openxla/xla/pull/36159

shawnwang18: > `Why would it` 

Why would what? I mean if we specify the parent command buffer in the constructor, then we will need add the parent parameter to the constructor for CommandBuffer, GpuCommandBuffer, and CudaCommandBuffer. Because only changing the constructor of CudaCommandBuffer, only from cuda_command_buffer, we have no way to know which is the parente CudaCommandBuffer if is_owned_graph = false.  

For example, when command_buffer_cmd wants to create a new command buffer for tracing, we needs to specify the parent when creating the trace command buffer.

shawnwang18: >  I think is_owned_graph_ is a redundant field.

I think so, but should we leave this PR just for fixing the test failure?  removing `Is_owned_graph_` can be in another separate PR that focus on it.

apivovarov (approved):

xla/stream_executor/cuda/cuda_command_buffer.cc
apivovarov: @ezhulenev refactoring is not that small (#36159 - 18 files touched, +86 −52). Can we keep this PR as-as and do API change in a separate PR?

PR #36198, opened 2026-01-09
Algebraic simplifier: handle bitcast(unary_elementwise(bitcast())).

sergachev:
📝 Summary of Changes
Squash pairs of bitcasts.

🎯 Justification
Speeds up some fusions by simplifying index computations. This fusion for example gets 35-45% faster from squashing tmp_16 into tmp_18:
```
f {
  tmp_0 = bf16[16384,14336] parameter(1)
  tmp_1 = bf16[2,8192,14336] bitcast(tmp_0)
  tmp_2 = bf16[] constant(1)
  tmp_3 = f32[] convert(tmp_2)
  tmp_4 = f32[2,8192,14336] broadcast(tmp_3), dimensions={}
  tmp_5 = bf16[2,8192,14336] negate(tmp_1)
  tmp_6 = bf16[2,8192,14336] exponential(tmp_5)
  tmp_7 = bf16[2,8192,14336] broadcast(tmp_2), dimensions={}
  tmp_8 = bf16[2,8192,14336] add(tmp_6, tmp_7)
  tmp_9 = f32[2,8192,14336] convert(tmp_8)
  tmp_10 = f32[2,8192,14336] divide(tmp_4, tmp_9)
  tmp_11 = bf16[2,8192,14336] convert(tmp_10)
  tmp_12 = bf16[2,8192,14336] multiply(tmp_1, tmp_11)
  tmp_13 = bf16[16384,14336] parameter(0)
  tmp_14 = bf16[2,8192,14336] bitcast(tmp_13)
  tmp_15 = bf16[2,8192,14336] multiply(tmp_12, tmp_14)
  tmp_16 = bf16[16384,14336] bitcast(tmp_15)
  tmp_17 = bf16[16384,14336] abs(tmp_16)
  tmp_18 = bf16[28672,8192] bitcast(tmp_17)
  tmp_19 = bf16[] constant(-inf)
  tmp_20 = bf16[28672] reduce(tmp_18, tmp_19), dimensions={1}, to_apply={
    _ = bf16[] maximum(bf16[] parameter(0), bf16[] parameter(1))
  }
  ROOT tmp_21 = (bf16[28672], bf16[2,8192,14336]) tuple(tmp_20, tmp_15)
}
```

🚀 Kind of Contribution
⚡️ Performance Improvement

| Benchmark                                              | Speedup |
|--------------------------------------------------------|---------|
| gpu_hlo.hlo                                            |   1.01x |
| hlo_llama31_8b_bf16_1x8.hlo                            |   0.99x |
| hlo_llama31_8b_fp8_1x8.hlo                             |   1.01x |
| hlo_llama3_8b_bf16_activation_offloading_1x8.hlo       |   1.00x |
| hlo_mixtral_8x7b_bf16_1x8.hlo                          |   1.00x |
| nv_maxtext_1n1g_jit_train_step_before_optimization.hlo |   1.00x |

🧪 Unit Tests:
yes

🧪 Execution Tests:
no

### DIFF
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
@@ -73,11 +73,11 @@ limitations under the License.
 #include "xla/shape_util.h"
 #include "xla/status_macros.h"
 #include "xla/tsl/platform/errors.h"
+#include "xla/tsl/platform/status_macros.h"
 #include "xla/tsl/platform/statusor.h"
 #include "xla/util.h"
 #include "xla/window_util.h"
 #include "xla/xla_data.pb.h"
-#include "xla/tsl/platform/status_macros.h"
 
 namespace xla {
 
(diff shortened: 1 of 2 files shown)


dimitar-asenov:
LGTM, but I'm not sure if there is some subtlety that we need to be aware of. @pifon2a could you please also review this?

sergachev:
ping

mooskagh (approved):

akuegel:
> LGTM, but I'm not sure if there is some subtlety that we need to be aware of. @pifon2a could you please also review this?

There is indeed a subtlety. On GPU we do allow to replace bitcast-convert with bitcast. But this new simplification pattern does not work correctly if it squashes bitcasts that change element type. Elementwise ops can behave differently on different element types. @sergachev I believe we should skip any bitcasts that change element type for this pattern.

akuegel (approved):

akuegel:
There are still some failing internal tests, it complains about bitcasts from `(s32[2,512]{1,0:T(2,128)})` to `(s32[512,2]{1,0:T(8,128)})`. So I guess we should also skip bitcasts that involve any of these tiling specifications.

akuegel (approved):
Thank you :)

akuegel:
I will roll forward myself with a fix. The problem was that IsElementwise also returns true for loop fusions. But the pattern would not update the computation shape of the fusion computation, so we should not apply the pattern for a fusion op between two bitcasts.

akuegel:
The roll forward will likely take a while, some of our presubmits are currently broken and block submits.

sergachev:
Thank you, Adrian!

Does it really make sense that IsElementwise is true for loop fusions?

akuegel:
> Thank you, Adrian!
> 
> Does it really make sense that IsElementwise is true for loop fusions?

Probably not. It has been there for a while, and I do believe it is a footgun. I found the change that added it (predates open sourcing). But it seems likely that just removing it could break something. Apparently at the time, considering such fusions as elementwise allowed a buffer reuse.

PR #36224, opened 2026-01-09
Support all-reduce hoisting for scatter-based accumulation pattern

sfvaroglu:
📝 Summary of Changes
This PR enables hoisting all-reduce operations out of while loops for scatter-based gradient accumulation patterns, commonly used in ZeRO-1 with gradient accumulation.

🎯 Justification
This optimization improves ZeRO-1 gradient accumulation performance by replacing all-reduce operations inside a loop with one after the loop, reducing communication overhead. 

🚀 Kind of Contribution
✨ New Feature, 🧪 Tests

The public HLOs in `xla/tools/benchmarks/hlo/` do not have this pattern.

🧪 Unit Tests:
Added to `xla/service/while_loop_all_reduce_code_motion_test.cc`

🧪 Execution Tests:
N/A


xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -715,6 +715,110 @@ MovableAllReduceContext IsAllReduceMovable(
   // Finds all accumulation contexts of the given all-reduce instruction
   // if it is movable.
   std::vector<AccumulationContext> accumulation_contexts;
+
+  // Scatter pattern: select(pred, 0, buffer) → scatter → all-reduce
+  // Unlike other patterns, pred can be non-replicated (e.g., partition-id)
+  // since all-reduce synchronizes results across replicas.
+  auto find_scatter_pattern = [&]() -> std::optional<AccumulationContext> {
```
olegshyshkov: Could you please move make lambda into a helper function? I don't this we gain much from lambda here.

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -715,6 +715,110 @@ MovableAllReduceContext IsAllReduceMovable(
...
+    std::stack<HloInstruction*> operand_chain;
```
olegshyshkov: I don't think you need this `stack`, it never growth bigger than 1. Judging from the code, you want to follow a chain of convert, bitcast, reshape, transpose and slice until you find a scatter and then there's a completely different logic. I think you should split that part in two.

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -715,6 +715,110 @@ MovableAllReduceContext IsAllReduceMovable(
...
+    operand_chain.push(all_reduce->mutable_operand(0));
+
+    while (!operand_chain.empty()) {
+      HloInstruction* current = operand_chain.top();
+      operand_chain.pop();
+
+      switch (current->opcode()) {
+        case HloOpcode::kScatter: {
+          std::stack<HloInstruction*> base_chain;
```
olegshyshkov: Same question why is this a stack. It never growth bigger than 1. I think you can streamline the logic here as well.

olegshyshkov (changes requested):

xla/service/while_loop_all_reduce_code_motion.cc
sfvaroglu: Done

xla/service/while_loop_all_reduce_code_motion.cc
sfvaroglu: Done

xla/service/while_loop_all_reduce_code_motion.cc
sfvaroglu: Done

sfvaroglu:
@olegshyshkov ping

olegshyshkov (approved):

allanrenucci:
Can you please rebase this PR?

allanrenucci (approved):

akuegel:
@sfvaroglu Can you please rebase?

allanrenucci (approved):

PR #36345, opened 2026-01-13
[XLA:GPU] Add experimental flag to allow all fusions to be autotuned with Triton

and-ivanov:
📝 Summary of Changes
This addition of the flag "xla_gpu_experimental_autotune_all_fusions_with_block_level_emitter" enables all fusions (including elementwise) to be implemented through Triton backend. 

🎯 Justification
This can be useful for evaluating the effect of enabling TMA through Triton.

### DIFF
--- a/xla/service/gpu/nvptx_compiler.cc
+++ b/xla/service/gpu/nvptx_compiler.cc
@@ -426,10 +426,7 @@ NVPTXCompiler::GetCodegenBackends(
 
 namespace {
 
-// Returns true if the instruction is a fusion that would go through the native
-// emitter, but may benefit from going through the block-level emitter.
-// Currently, we only do this for reductions and transposes.
-bool ShouldAutotuneBetweenFusionEmitters(const HloInstruction& instruction) {
+bool ShouldAutotuneBetweenFusionEmittersAny(const HloInstruction& instruction) {
   if (instruction.opcode() != HloOpcode::kFusion) {
     return false;
   }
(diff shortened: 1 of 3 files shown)


vwbaker (approved):
Cool! This is something we were also considering looking into, thanks for the change.

Moerafaat:
The change seems reasonable, thanks!

Moerafaat (approved):

Moerafaat (approved):

golechwierowicz (approved):

xla/service/gpu/nvptx_compiler.cc
```diff
@@ -450,14 +456,19 @@ absl::Status NVPTXCompiler::AddFusionAutotuningPass(
       /*use_default_config=*/true);
   backends.push_back(std::move(ble_backend));
 
+  auto should_autotune =
+      debug_options
+              .xla_gpu_experimental_autotune_all_fusions_with_block_level_emitter()
```
thomasjoerg: I'm looking at merging this PR at Google. We have a strict requirement lines to be <= 80 chars wide. The flag name you picked is too long to do that and keep the code formatted (with ClangFormat). Please pick a shorter Flag name.

thomasjoerg (changes requested):

xla/service/gpu/nvptx_compiler.cc
and-ivanov: I renamed it to set_xla_gpu_experimental_all_fusions_with_triton

thomasjoerg (approved):

xla/xla.proto
```diff
@@ -651,6 +651,9 @@ message DebugOptions {
   // Specifies the directory to store per kernel autotuning cache.
   optional string xla_gpu_experimental_autotuner_cache_dir = 407;
 
+  // If true, autotune all fusions with block level emitter.
```
thomasjoerg: Please keep the flags sorted. Thank you!

thomasjoerg (changes requested):

xla/xla.proto
and-ivanov: done, now it should be sorted

thomasjoerg (approved):

PR #36346, opened 2026-01-13
[XLA] Add missing return to prevent nullptr dereference

apach301:
Hi, I was testing TensorFlow code with Svace static analyzer and found possible null dereference in XLA.

Possible null dereference may occur because of missing return in MsaAlgorithm::UpdateAllocationRequirementForUseAliases(). There is a case when `aliased_allocation` variable could be null, and it could be dereferenced in AddAliasedRequiredAssignment().

https://github.com/tensorflow/tensorflow/issues/99907


allanrenucci (approved):
Thanks!

PR #36388, opened 2026-01-14
[xla:gpu] Remove persistent collective cliques

ezhulenev:
Persistent collective cliques is an unsafe escape hatch to cache NCCL cliques and avoid rendezvous before each execution. This is unused and easily leads to deadlocks. Remove it from XLA.

This is a rollback of experimental feature added in https://github.com/openxla/xla/commit/d93ec4a2f3573e8d3e07f5a628666188dd03e60d#diff-2aa478b8c4a8c178582b43774b52591df68d9f49fed0118db02a2be508cf2c49, it didn't find any users and it adds unneeded complexity to XLA.

### DIFF
--- a/xla/service/gpu/gpu_executable.cc
+++ b/xla/service/gpu/gpu_executable.cc
@@ -486,11 +486,7 @@ absl::Status ExecuteThunksImpl(
   if (!mock_collectives) {
     TF_ASSIGN_OR_RETURN(
         collective_cliques,
-        AcquireCollectiveCliques(
-            collective_params, clique_requests,
-            debug_options
-                ? debug_options->xla_gpu_collectives_use_persistent_cliques()
-                : false));
+        AcquireCollectiveCliques(collective_params, clique_requests));
   }
 
   TF_RETURN_IF_ERROR(multimem_registry.Build());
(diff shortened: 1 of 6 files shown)


xla/service/gpu/gpu_executable.cc
```diff
@@ -512,16 +508,6 @@ absl::Status ExecuteThunksImpl(
...
   }
 
-  // Maybe join a round of rendezvous after thunk initialization. We do this
-  // only in presence of newly acquired collective cliques which means that we
-  // have collective operations and clique initialization is famous for
-  // introducing deadlocks if we try to execute it concurrently with other
-  // potentially memory-allocating operations.
-  if (collective_cliques.num_transient_cliques() > 0) {
-    TF_RETURN_IF_ERROR(
-        RendezvousAfterInitialization(run_options, debug_options));
```
apivovarov, line 522: Just want to double check that the deadlock scenarios described in the deleted comment are no longer possible. Can you confirm?

ezhulenev: Ah, good catch, reverted it back to correct state. This is basically a rollback of: https://github.com/openxla/xla/commit/d93ec4a2f3573e8d3e07f5a628666188dd03e60d#diff-2aa478b8c4a8c178582b43774b52591df68d9f49fed0118db02a2be508cf2c49

apivovarov (approved):
Thank you, Eugene!

penpornk (approved):

PR #36417, opened 2026-01-14
[xla:gpu] Add device comms requirements to clique requests

ezhulenev:
Add an e2e test that requests collective cliques with device comms using the `CollectiveCliqueRequests` API during the prepare execution stage.

Device communicators allocated lazily when requested and added to the per-process `GpuClique` instance.


ezhulenev:
@olegshyshkov @PatriosTheGreat can you check this one please, rebased to latest `main`

olegshyshkov (approved):

xla/backends/gpu/collectives/nccl_collectives.cc
```diff
@@ -84,6 +84,14 @@ bool NcclCollectives::IsGlobalConfig() const {
   return nccl_comm_id != nullptr;
 }
 
+bool NcclCollectives::SupportsDeviceComm() const {
+#if NCCL_VERSION_CODE >= 22800
+  return true;
+#else
+  return false;
+#endif  // NCCL_VERSION_CODE >= 22800
+}
```
allanrenucci, line 89: Let's avoid the usage of macro here. Otherwise we need to bypass the internal Presubmit error.

```suggestion
bool NcclCollectives::SupportsDeviceComm() const {
  return NCCL_VERSION_CODE >= 22800;
}
```

allanrenucci (changes requested):

xla/backends/gpu/collectives/nccl_collectives.cc
ezhulenev: Done.

penpornk (approved):

PR #36441, opened 2026-01-15
Support all-reduce hoisting with scalar multiplication pattern

sfvaroglu:
📝 Summary of Changes
This PR enables hoisting all-reduces out of loops when gradients are scaled by a scalar all-reduce before accumulation. 

🎯 Justification
This pattern commonly appears in ZeRO-1 gradient accumulation with weight normalization:
```
    grads_accum = tree_map(lambda g, b: AR(g) * AR(total_weights) + b, grads, grads_accum)
```

Before: Each iteration performs ARs on gradients and weights
After: Local accumulation in loop, ARs after loop

🚀 Kind of Contribution
✨ New Feature, 🧪 Tests

The HLOs in `compiler/xla/tools/benchmarks/hlo/` do not have this pattern.

🧪 Unit Tests:
Added to `xla/service/while_loop_all_reduce_code_motion_test.cc`

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/service/while_loop_all_reduce_code_motion.cc
+++ b/xla/service/while_loop_all_reduce_code_motion.cc
@@ -768,6 +768,44 @@ MovableAllReduceContext IsAllReduceMovable(
           }
           break;
         }
+        case HloOpcode::kMultiply: {
+          if (!is_reduce_scatter) {
+            HloInstruction* other_operand =
+                user->mutable_operand(1 - user->operand_index(instruction));
+            HloInstruction* unwrapped_other = other_operand;
+            bool other_is_broadcast = false;
+            while (unwrapped_other->opcode() == HloOpcode::kBroadcast ||
+                   unwrapped_other->opcode() == HloOpcode::kConvert) {
+              if (unwrapped_other->opcode() == HloOpcode::kBroadcast) {
+                other_is_broadcast = true;
+              }
+              unwrapped_other = unwrapped_other->mutable_operand(0);
+            }
+            bool current_is_scalar = ShapeUtil::IsScalar(all_reduce->shape());
+            bool other_is_all_reduce =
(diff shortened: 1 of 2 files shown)


olegshyshkov (approved):

sfvaroglu:
@olegshyshkov Any ideas why this PR is blocked?

olegshyshkov (approved):

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -778,6 +768,44 @@ MovableAllReduceContext IsAllReduceMovable(
...
+              }
+              unwrapped_other = unwrapped_other->mutable_operand(0);
+            }
+            bool current_is_scalar = ShapeUtil::IsScalar(all_reduce->shape());
+            bool other_is_all_reduce =
+                unwrapped_other->opcode() == HloOpcode::kAllReduce ||
+                unwrapped_other->opcode() == HloOpcode::kReduceScatter;
+
+            if (current_is_scalar && other_is_all_reduce) {
+              VLOG(4) << "Scalar all-reduce " << all_reduce->name()
```
akuegel, line 790: It looks (based on the VLOG statement here) that this block should also set `is_all_reduce_movable = false;` ?

We got a test failure internally, so we will have to revert the PR. But unfortunately even adding that line does not fix the test. Maybe some other condition is also not handled yet?

sfvaroglu: @akuegel Could you share the failing test or a reproducer HLO?

akuegel: It is not a test we can share, also not sure whether we can share the HLO. But as a first step, you could increase test coverage and try to think about all edge cases. Once you have another PR with increased test coverage, we can try it out with the internal test and possibly do some debugging.

PR #36462, opened 2026-01-15
[NVIDIA GPU] Enable nccl symmetric kernels by default

Tixxx:
📝 Summary of Changes
Enables nccl symmetric kernels by default

🎯 Justification
Symmetric kernels have smaller latencies compared legacy kernels, they also consume less SMs. NCCL symmetric kernels have been integrated into xla for some time and we have been running with it on. This pr attempts to enable it by default.
Note that symmetric buffers are allocated in the collective memory space which is outside of xla's default bfc allocator pool. XLA_CLIENT_MEM_FRACTION needs to be lowered if it's manually set to a high ratio. This needs xla to upgrade to nccl 2.28.

🚀 Kind of Contribution
⚡️ Performance Improvement,

hlo_llama31_8b_bf16_1x8.hlo -> 7% speedup
hlo_llama31_8b_fp8_1x8.hlo -> 8% speedup

🧪 Unit Tests:
all existing unit tests

🧪 Execution Tests:
all existing execution tests

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -282,7 +282,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_nccl_termination_timeout_seconds(-1);
   opts.set_xla_gpu_enable_shared_constants(true);
   opts.set_xla_gpu_enable_nccl_user_buffers(false);
-  opts.set_xla_gpu_experimental_enable_nccl_symmetric_buffers(false);
+  opts.set_xla_gpu_experimental_enable_nccl_symmetric_buffers(true);
   opts.set_xla_gpu_experimental_enable_nvshmem(false);
   opts.set_xla_gpu_enable_nccl_comm_splitting(true);
   opts.set_xla_gpu_nccl_init_max_rank_per_root_ratio(0);


PatriosTheGreat:
Unfortunately we can't turn this on for now by default, because we faced with some illegal memory access issues with NCCL

Tixxx:
> Unfortunately we can't turn this on for now by default, because we faced with some illegal memory access issues with NCCL

Yea this needs 2.28 of nccl which is progress, i suggest to keep it open until 2.28 is officially supported?

thomasjoerg (approved):
Approving this for import. Some testing will be required before merge. I'll keep you posted.

thomasjoerg:
Some internal benchmarks failing with:

```
INTERNAL: NCCL operation ncclGroupEnd() failed: unhandled cuda error (run with NCCL_DEBUG=INFO for details). Last NCCL warning(error) log entry (may be unrelated) 'Cuda failure 1 'invalid argument
```

I'm investigating. Let me know if you have a suspicion what could be causing this.

olegshyshkov:
> INTERNAL: NCCL operation ncclGroupEnd() failed: unhandled cuda error (run with NCCL_DEBUG=INFO for details). Last NCCL warning(error) log entry (may be unrelated) 'Cuda failure 1 'invalid argument

I was playing around with NCCL device API on Friday and got the same error. I don't remember exactly what fixed it (and I even found a fix), but I have a feeling that NCCL device API might be incompatible with the way XLA manager either memory or cliques:

1. Registering the same device address with `ncclCommWindowRegister` invalidated some handlers.
2. When we do NCCL comm splitting, it unregisters device memory handlers, but I'm not sure if they don't stay cached somewhere.

Tixxx:
> Some internal benchmarks failing with:
> 
> ```
> INTERNAL: NCCL operation ncclGroupEnd() failed: unhandled cuda error (run with NCCL_DEBUG=INFO for details). Last NCCL warning(error) log entry (may be unrelated) 'Cuda failure 1 'invalid argument
> ```
> 
> I'm investigating. Let me know if you have a suspicion what could be causing this.

This likely means there's an OOM error when nccl communicator tries to allocate memory outside of xla's mem pools. We usually lower the XLA_PYTHON_CLIENT_MEM_FRACTION when we enable symmetric memory. Can we try that with the failing workload?

Tixxx:
> > INTERNAL: NCCL operation ncclGroupEnd() failed: unhandled cuda error (run with NCCL_DEBUG=INFO for details). Last NCCL warning(error) log entry (may be unrelated) 'Cuda failure 1 'invalid argument
> 
> I was playing around with NCCL device API on Friday and got the same error. I don't remember exactly what fixed it (and I even found a fix), but I have a feeling that NCCL device API might be incompatible with the way XLA manager either memory or cliques:
> 
> 1. Registering the same device address with `ncclCommWindowRegister` invalidated some handlers.
> 2. When we do NCCL comm splitting, it unregisters device memory handlers, but I'm not sure if they don't stay cached somewhere.

For 1, it's not allowed in NCCL to symmetrically register the same address multiple times though. We should prevent that from happening in xla by re-using the handle cache in communicator.
I'm not aware that xla will unregister symmetric windows upon splitting comms, we set resource_sharing in the split config so the child communicators will share the same resource including registered windows with the parent comm, it doesn't make sense to me as to why it would unregister that window.

thomasjoerg:
> > Some internal benchmarks failing with:
> > ```
> > INTERNAL: NCCL operation ncclGroupEnd() failed: unhandled cuda error (run with NCCL_DEBUG=INFO for details). Last NCCL warning(error) log entry (may be unrelated) 'Cuda failure 1 'invalid argument
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm investigating. Let me know if you have a suspicion what could be causing this.
> 
> This likely means there's an OOM error when nccl communicator tries to allocate memory outside of xla's mem pools. We usually lower the XLA_PYTHON_CLIENT_MEM_FRACTION when we enable symmetric memory.

I ran open-source Gemma benchmarks. Gemma 3 Prefill did OOM, but the error message is different:
`hlo_runner_main.cc:629] Check failed: (::tsl::TfCheckOkDeprecationMarker(), xla::RealMain(argc, argv)) is OK (RESOURCE_EXHAUSTED: Out of memory while trying to allocate 832.00MiB. [tf-allocator-allocation-error='']`

>  Can we try that with the failing workload?

Sure, but this PR seeks to enable nccl symmetric kernels by default. In it's current state, this will be a breaking change to XLA users. Is there a way to avoid this?

Tixxx:
> I ran open-source Gemma benchmarks. Gemma 3 Prefill did OOM, but the error message is different:
hlo_runner_main.cc:629] Check failed: (::tsl::TfCheckOkDeprecationMarker(), xla::RealMain(argc, argv)) is OK (RESOURCE_EXHAUSTED: Out of memory while trying to allocate 832.00MiB. [tf-allocator-allocation-error='']

ok thanks, both collective allocator and nccl's own allocator perform allocations outside of XLA_PYTHON_CLIENT_MEM_FRACTION so looks like either will have a chance to OOM without tweaking XLA_PYTHON_CLIENT_MEM_FRACTION.

> Sure, but this PR seeks to enable nccl symmetric kernels by default. In it's current state, this will be a breaking change to XLA users. Is there a way to avoid this?

One way is to estimate how much memory will be needed for collectives and if that falls under OOM limit, we will use symmetric memory otherwise disable it, the challenge here is to know how much will be used by nccl comms since it's controlled by nccl lib. We can ask nccl to take a user-supplied memory cap.
Another way is to consolidate the memory pools in xla to do all allocation under XLA_PYTHON_CLIENT_MEM_FRACTION, we have some exploration in progress in this regard.

thomasjoerg:
> > I ran open-source Gemma benchmarks. Gemma 3 Prefill did OOM, but the error message is different:
> > hlo_runner_main.cc:629] Check failed: (::tsl::TfCheckOkDeprecationMarker(), xla::RealMain(argc, argv)) is OK (RESOURCE_EXHAUSTED: Out of memory while trying to allocate 832.00MiB. [tf-allocator-allocation-error='']
> 
> ok thanks, both collective allocator and nccl's own allocator perform allocations outside of XLA_PYTHON_CLIENT_MEM_FRACTION so looks like either will have a chance to OOM without tweaking XLA_PYTHON_CLIENT_MEM_FRACTION.
> 
> > Sure, but this PR seeks to enable nccl symmetric kernels by default. In it's current state, this will be a breaking change to XLA users. Is there a way to avoid this?
> 
> One way is to estimate how much memory will be needed for collectives and if that falls under OOM limit, we will use symmetric memory otherwise disable it, the challenge here is to know how much will be used by nccl comms since it's controlled by nccl lib. We can ask nccl to take a user-supplied memory cap. Another way is to consolidate the memory pools in xla to do all allocation under XLA_PYTHON_CLIENT_MEM_FRACTION, we have some exploration in progress in this regard.

The second approach seems more robust. We need a solution for the OOM issues before we can merge this PR.

thomasjoerg (changes requested):
See comment above

Tixxx:
converting this to draft, we are addressing the memory space consolidation in a separate pr.

PR #36525, opened 2026-01-16
Fix entry_computation_layout memory space for host offloaded outputs

zhenying-liu:
📝 Summary of Changes
When the compute_on custom call's outputs reside in host memory, this information must be propagated to the results of entry_computation_layout.

🎯 Justification
This change is important for workloads such as MaxText, as it enables compute offload by correctly handling outputs in host memory.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

This is primarily a bug fix rather than a functional change. It enables MaxText to offload optimizer and weight update computations to the host, which may facilitate future performance improvements, but does not directly impact benchmark speed at this stage.

🧪 Unit Tests:
Three new unit tests were added to host_offloader_test.cc to cover the following cases: (1) MoveToHost custom call with a single output, and (2) tuple outputs where elements can reside in different memory spaces (host or device).

🧪 Execution Tests:
Currently, only compute_on() has test cases, and its outputs are all on the device. After this fix, new end-to-end tests will be added in jax/tests/memories_test.py to cover compute_on2() cases where outputs can be in host memory spaces.

### DIFF
--- a/xla/hlo/transforms/host_offloader.cc
+++ b/xla/hlo/transforms/host_offloader.cc
@@ -68,6 +68,53 @@ void SetMemorySpace(Shape* shape, int64_t memory_space_color) {
   shape->mutable_layout()->set_memory_space(memory_space_color);
 }
 
+// Returns true if the layout has explicit customizations that indicate it was
+// explicitly set by the user rather than being a default layout. We check:
+// - tiles_size(): User-specified tiling configuration (e.g., T(2,128))
+// - index_primitive_type(): Custom indexing for sparse/ragged arrays
+// - pointer_primitive_type(): Custom pointer representation
+// - element_size_in_bits(): Packed/compressed element sizes
+// If any of these are set, we must not modify the memory space, as doing so
+// could break user expectations or cause layout incompatibilities.
+bool IsLayoutExplicitlyCustomized(const Layout& layout) {
+  return layout.tiles_size() > 0 ||
+         layout.index_primitive_type() != PRIMITIVE_TYPE_INVALID ||
+         layout.pointer_primitive_type() != PRIMITIVE_TYPE_INVALID ||
+         layout.element_size_in_bits() != 0;
+}
+
+// Validates that the entry computation layout at the given shape_index can
+// accept host memory space, and updates it if needed. Returns true if the
+// layout was updated. Returns an error if the layout was explicitly customized
+// by the user with a non-host memory space.
+absl::StatusOr<bool> EnsureEntryLayoutIsHostMemory(
+    ComputationLayout* comp_layout, const ShapeIndex& shape_index,
+    absl::string_view context_for_error) {
+  const Shape& output_shape =
+      ShapeUtil::GetSubshape(comp_layout->result_layout().shape(), shape_index);
+  CHECK(output_shape.has_layout())
+      << "Expecting output shape of entry computation to have a layout.";
+
+  // Already host memory - no change needed.
+  if (output_shape.layout().memory_space() == Layout::kHostMemorySpace) {
+    return false;
+  }
+
+  // Check if user explicitly customized the layout.
+  if (IsLayoutExplicitlyCustomized(output_shape.layout())) {
+    return error::CompileTimeHostOffloadOutputLocationMismatch(
+        "Tensor which is moved to host (%s) is returned from the entry "
+        "computation but the layout for this output is not set to host memory.",
+        context_for_error);
+  }
+
+  // Safe to update - layout was not explicitly customized.
+  Layout new_layout = output_shape.layout();
+  new_layout.set_memory_space(Layout::kHostMemorySpace);
+  comp_layout->mutable_result_layout()->ResetLayout(new_layout, shape_index);
+  return true;
+}
+
 bool SetBuffersToMemorySpaceColor(
     const std::vector<InstructionAndShapeIndex>& buffers_to_set_to_host_memory,
     int64_t memory_space_color) {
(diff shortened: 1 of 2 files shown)


zhenying-liu:
Can I ask the failed message as I don't have access to it?
/__w/.cache/bazel/bazel/_bazel_root/d1cb79a7d71e7631affefe2251b38575/execroot/xla/bazel-out/k8-opt/testlogs/xla/hlo/transforms/host_offloader_test/shard_12_of_12/test.log

zhenying-liu:
The log showed the reason of CI ROCm / execute-xla-ut (pull_request) failure:
`ERROR: /workspace/xla/tests/BUILD:1479:9: Testing //xla/tests:convolution_test_amdgpu_any (shard 3 of 50) failed: (Exit 34): NOT_FOUND: [third_party/bazel/src/main/native/unix_jni.cc:1259] write (No space left on device)`

mingyaoy:
We should not update entry computation layout in this pass. Entry computation layout is given to XLA by runtime and XLA needs to respect the entry computation layout (which is how arguments will be passed by runtime). There might be a missing host offloading annotation in the original module that leads to some assert of memory space but it should not be fixed by modifying entry computation layout.

zhenying-liu:
@mingyaoy Thanks for your feedback. You mentioned that "entry computation layout is given to XLA by runtime and XLA needs to respect it." However, in xla/service/layout_assignment.cc, there are explicit comments:                                                                                                                                                                                                                                                                                                  

```
  // Allow some parameter/result layouts to be unset in the entry                                                                                                
  // computation.                                                                                                                                                
  if (parameter_layout.AnyLayoutIsSet()) {                                                                                                                       
    // Clear out memory space in layout. Host offloader will do the                                                                                              
    // analysis later.                                                                                                                                           
    TF_RETURN_IF_ERROR(ResetMemorySpaceInLayout(parameter_layout));     
```
 Layout assignment clears memory spaces and defers to host offloader for analysis. My understanding was that this analysis should include updating entry_computation_layout to reflect the results, but I recognize the comment doesn't explicitly state that.                                                    
                                                                                                                                                                 
 This patch fixes a real issue with JAX's @compute_on2 decorators where users explicitly specify output memory spaces via the out_memory_spaces parameter:              
                                                                                                                                                                 
```
  @compute_on2(compute_type='device_host', out_memory_spaces=(                                                                                                   
      jax.memory.Space.Device,                                                                                                                                   
      jax.memory.Space.Host                                                                                                                                      
  ))                                                                                                                                                             
  def tuple_outputs(x):                                                                                                                                          
      return x * 2.0, x + 1.0                                                                                                                                    
                                                                                                                                                                 
  inp_b = jnp.array([1.0, 2.0, 3.0])                                                                                                                             
  result_b1, result_b2 = tuple_outputs(inp_b)  # FAILS without this patch  
```
The user explicitly requests the second output in host memory. JAX lowers this to MoveToHost custom calls. Layout assignment clears the memory spaces expecting host offloader to analyze them later. But currently, host offloader never updates entry_computation_layout, causing a mismatch between the HLO graph (host memory) and the interface contract (device memory).                                                                                                            
                                                                                                                                                                 
The patch only updates entry_computation_layout when the layout has no explicit customizations (tiles, custom primitive types, element sizes). If the layout is explicitly customized, it returns an error, preserving the original behavior. This ensures we don't override user-specified layouts, only populate unset memory spaces that layout_assignment intentionally left for host offloader.                                                                                    
                                                                                                                                                                 
 I'm working on enabling this for MaxText GPU compute offloading. I initially considered fixing this in GPU layout assignment's PropagateResultConstraint(), but that also has complications with ResetMemorySpaceInLayout(). Could you advise on where this propagation should happen? The runtime (JAX) is providing the intent via out_memory_spaces for compute_on2(), but I'm unsure which XLA pass should reflect that in entry_computation_layout.

zhenying-liu:
The failed CI ROCm/execute-xla-ut has the error message:
```
ERROR: /workspace/xla/tests/BUILD:1481:9: Testing //xla/tests:convolution_test_amdgpu_any (shard 43 of 50) failed: (Exit 34): INVALID_ARGUMENT: Unable to pull image 'rocm/tensorflow-build@sha256:a2672ff2510b369b4a5f034272a518dc93c2e492894e3befaeef19649632ccaa' on worker 'ctr-cx65-mi300x-23 (/10.7.200.77:8081)':
'docker pull rocm/tensorflow-build@sha256:a2672ff2510b369b4a5f034272a518dc93c2e492894e3befaeef19649632ccaa' failed with Command 'docker pull rocm/tensorflow-build@sha256:a2672ff2510b369b4a5f034272a518dc93c2e492894e3befaeef19649632ccaa' failed in 900.012s (15m 0s) with exit code 143: docker.io/rocm/tensorflow-build@sha256:a2672ff2510b369b4a5f034272a518dc93c2e492894e3befaeef19649632ccaa: Pulling from rocm/tensorflow-build
```
It seems the docker pull issue, not from my commit.
@dubstack Please help take a look at the code review. Thanks.

mingyaoy:
>   // Allow some parameter/result layouts to be unset in the entry                                                                                                
  // computation.                                                                                                                                                
  if (parameter_layout.AnyLayoutIsSet()) {                                                                                                                       
    // Clear out memory space in layout. Host offloader will do the                                                                                              
    // analysis later.                                                                                                                                           
    TF_RETURN_IF_ERROR(ResetMemorySpaceInLayout(parameter_layout)); 

This doesn't change entry computation layout. Currently if
> @compute_on2(compute_type='device_host', out_memory_spaces=(                                                                                                   
      jax.memory.Space.Device,                                                                                                                                   
      jax.memory.Space.Host                                                                                                                                      
  )) 

returns an array instead of a tuple, a copy is added. So we can either mimic that's done for a non-tuple which isn't ideal. Or the real issue is in the before-optimization hlo xla receives. The entry computation layout should have S(5) in it for the output. I confirmed with Jax team and they said they'll fix it.

mingyaoy:
Jax team just fixed it. So the output should have S(5) to begin with.

zhenying-liu:
Cool. Thanks @mingyaoy! I'll test the latest jax and close this.

yashk2810:
https://github.com/jax-ml/jax/pull/34658 should fix your minimal repro so you probably don't need this change.

PR #36612, opened 2026-01-20
[ROCm] Fix reduction emitter tests to work on ROCm

nurmukhametov:
The tests mof_scalar_variadic.hlo and side_output_broadcast.hlo were designed to test the Reduction emitter with Multi-Output Fusions (MOF) containing variadic reductions and side outputs. However, they used f32[6,6] (36 elements) which fails the
IsUnnestedReductionFasterThanElemental heuristic on ROCm (warp_size=64) while passing on CUDA (warp_size=32).

When the heuristic fails, the fusion is routed to the Loop emitter instead of the Reduction emitter. The Loop emitter cannot handle fusions with incompatible output shapes (e.g., f32[] and f32[6,6] in the same tuple), causing a segmentation fault in GetBitcastMap/ApplyIndexing.

Fix by increasing the test dimensions from 6x6 to 8x8 (64 elements), which satisfies the heuristic on both platforms. This ensures the tests exercise the Reduction emitter as originally intended on both CUDA and ROCm.

### DIFF
--- a/xla/backends/gpu/codegen/emitters/tests/reduce_row/mof_scalar_variadic.hlo
+++ b/xla/backends/gpu/codegen/emitters/tests/reduce_row/mof_scalar_variadic.hlo
@@ -16,11 +16,14 @@
   ROOT tuple = (f32[], f32[]) tuple(add0, add1)
 }
 
-fusion {
-  %p0 = f32[6,6] parameter(0)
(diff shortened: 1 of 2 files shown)


nurmukhametov:
@xla-rotation, could I get a review on this?

nurmukhametov:
@jreiffers, could you please review this PR?

pifon2a (approved):

PR #36633, opened 2026-01-20
[xla] Use xla/util errors instead of absl errors to capture stack trace

ezhulenev:
[xla] Use xla/util errors instead of absl errors to capture stack trace


mwhittaker (approved):
:shipit:

ermilovmaxim (approved):

PR #36708, opened 2026-01-22
[XLA:GPU][oneAPI] Device description for sycl stream executor

mdfaijul:
This PR populates various fields of `DeviceDescription` mostly using level zero API and adds it to sycl executor.


mdfaijul:
@beckerhe Could you please help reviewing this PR.

mdfaijul:
@beckerhe @penpornk Could you please help reviewing this PR.

xla/stream_executor/sycl/sycl_device_description.cc
```diff
@@ -0,0 +1,292 @@
...
+    }                                                            \
+  } while (0)
+
+namespace {
+
+// Current level zero release drivers report incorrect memory properties on some
+// platforms. Therefore, we use values from the official product specifications.
+// TODO(intel-tf): Implement a proper memory bandwidth estimation based on
+// level zero memory properties.
+inline int64_t EstimateMemoryBandwidth(
```
beckerhe: The `inline` keyword on all these functions are a bit redundant since they are defined in a single compilation unit in an anonymous namespace. I would suggest to remove them for brevity.

xla/stream_executor/sycl/sycl_device_description.cc
```diff
@@ -0,0 +1,292 @@
...
+  }
+  return 32;  // Default subgroup size to conform to CUDA warp size.
+}
+
+// Parses version strings of the form "major.minor.patch+build#". For example,
+// "1.2.3+456" will be parsed as {1, 2, 3}. If parsing fails, returns {0, 0, 0}.
+inline SemanticVersion ParseOrDefaultDriverVersion(
+    absl::string_view version_str) {
+  // Split by '+' to remove build# if present.
+  std::vector<std::string> version_str_parts = absl::StrSplit(version_str, '+');
```
beckerhe: You can avoid some heap allocations here by making the return type `std::array<absl::string_view, 1>`

xla/stream_executor/sycl/sycl_device_description.cc
```diff
@@ -0,0 +1,292 @@
...
+  if (version_str_parts.empty()) {
+    return SemanticVersion{0, 0, 0};
+  }
```
beckerhe: This can only be empty if `version_str` is empty. I think it would be clearer if you were checking whether `version_str` is empty at the beginning of this function.

xla/stream_executor/sycl/sycl_device_description.cc
```diff
@@ -0,0 +1,292 @@
...
+
+  auto [major, minor, patch] =
+      std::tuple<unsigned, unsigned, unsigned>(0, 0, 0);
+  if (!absl::SimpleAtoi(version_parts[0], &major) ||
+      !absl::SimpleAtoi(version_parts[1], &minor) ||
+      !absl::SimpleAtoi(version_parts[2], &patch)) {
+    return SemanticVersion{0, 0, 0};
+  }
+
+  return SemanticVersion{major, minor, patch};
```
beckerhe: There is `SemantionVersion::ParseFromString` which you can use here.

xla/stream_executor/sycl/sycl_device_description.cc
```diff
@@ -0,0 +1,292 @@
...
+  zeDriverGetExtensionFunctionAddress(
+      driver, "zeIntelGetDriverVersionString",
+      reinterpret_cast<void**>(&driver_version_func));
+  if (driver_version_func) {
+    size_t driver_version_string_size = 0;
+    driver_version_func(driver, nullptr, &driver_version_string_size);
+    std::vector<char> driver_version_string(driver_version_string_size);
+    driver_version_func(driver, driver_version_string.data(),
+                        &driver_version_string_size);
+    return ParseOrDefaultDriverVersion(driver_version_string.data());
```
beckerhe: This implicitly constructs an absl::string_view and determines the length by calling `strlen`. Could you explicitly provide the length to avoid calling strlen? I suspect we need to pass in `driver_version_string_size - 1` since there is probably a NULL character. The documentation doesn't specify that which is also the reason why I would like to avoid relying on strlen here.

xla/stream_executor/sycl/sycl_device_description.cc
```diff
@@ -0,0 +1,292 @@
...
+  }
+  return SemanticVersion{0, 0, 0};
+}
+
+inline std::string GetPciBusId(ze_device_handle_t lz_device) {
+  ze_pci_ext_properties_t pci_props{ZE_STRUCTURE_TYPE_PCI_EXT_PROPERTIES};
```
beckerhe, line 141: This file is using a bunch of `ze_*` functions but is not including any level zero headers. We try to follow the include-what-you-use scheme, so this file should include the headers which declare all the used symbols. Could you add those includes?

xla/stream_executor/sycl/sycl_executor.cc
```diff
@@ -36,19 +34,22 @@ limitations under the License.
...
 #include "xla/stream_executor/stream.h"
 #include "xla/stream_executor/stream_executor.h"
+#include "xla/stream_executor/sycl/sycl_device_description.h"
 #include "xla/tsl/util/env_var.h"
 #include "tsl/platform/env.h"
 #include "tsl/platform/errors.h"
 #include "tsl/platform/fingerprint.h"
 #include "tsl/platform/numbers.h"
 #include "tsl/platform/statusor.h"
+#include <unistd.h>
```
beckerhe: I believe clang-format will not be happy with the include down here. Could you move it back up?

beckerhe (commented):
Hi @mdfaijul, The change looks good. Thank you! Just minor comments from my side.

xla/stream_executor/sycl/sycl_device_description.cc
mdfaijul: Done.

xla/stream_executor/sycl/sycl_device_description.cc
mdfaijul: Done.

xla/stream_executor/sycl/sycl_device_description.cc
mdfaijul: Done.

xla/stream_executor/sycl/sycl_device_description.cc
mdfaijul: Thanks! Reused this code.

xla/stream_executor/sycl/sycl_device_description.cc
mdfaijul: I have verified that level zero uses array of char without null terminator. I have put comments on this.
Here is some level zero documentation with example code https://github.com/intel/compute-runtime/blob/master/level_zero/doc/experimental_extensions/GET_DRIVER_VERSION_STRING.md

xla/stream_executor/sycl/sycl_device_description.cc
mdfaijul: Done.

xla/stream_executor/sycl/sycl_executor.cc
mdfaijul: Done. Moved at the top. However, if I run the following command `git diff main -- '*.cc' '*.h' | clang-format-diff.py -binary clang-format -p1 -style=file`, it moves  `#include <unistd.h>` between `absl/.*` and `xla/.*`

mdfaijul:
@beckerhe Thanks for the review. I have addressed the comments. Please check.

mdfaijul:
@beckerhe Could you please check review comments addressed?

xla/stream_executor/sycl/sycl_device_description.cc
beckerhe: Okay, so when it's not null-terminated then we can't rely on strlen. Can you change it to `return ParseOrDefaultDriverVersion(absl::string_view(driver_version_string.data(), driver_version_string_size));`?

mdfaijul: Fixed.

mdfaijul:
@beckerhe Thanks! for the review. I have updated with `return ParseOrDefaultDriverVersion(absl::string_view(driver_version_string.data(), driver_version_string_size));`

beckerhe (approved):

mdfaijul:
@beckerhe Thanks for approving the PR. I noticed few CI failures Linux {x86, ARM64} CPU. Those failures I think unrelated to the PR. Let me know if something needs to be taken care of from my side.

beckerhe (approved):

xla/stream_executor/sycl/BUILD
```diff
@@ -488,3 +489,39 @@ xla_test(
         "@com_google_googletest//:gtest_main",
     ],
 )
+
+cc_library(
```
thcmbs: fme: should this be sycl_library ?

mdfaijul: Yes, you are right. Thanks for the suggestion. Updated with sycl_library.

mdfaijul:
@thcmbs I have addressed your comments. Please check.

mdfaijul:
@thcmbs @mraunak The internal PRs from [copybara-service](https://github.com/apps/copybara-service) added some artifacts which were not in this XLA PR.

Here are the artifacts added.

- The dependency target "//third_party/intel_level_zero:level_zero_loader" is incorrect and not needed.
- `#include <ze_api.h>` is not correct, it should be `#include <level_zero/ze_api.h>`.

Those artifacts are causing build failures. @thcmbs Could you please help removing those artifacts.

PR #36755, opened 2026-01-23
autotuner: iterate in deterministic order

olupton:
📝 Summary of Changes
Avoid iterating in a run-to-run-unstable order (https://abseil.io/docs/cpp/guides/container#iteration-order-instability). This was causing different processes in distributed JAX runs to use inconsistent operation names.

🎯 Justification
PGLE expects consistent operation names.

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
None

🧪 Execution Tests:
None


shyamliagrawal (approved):
Thanks for the fix!

PR #36792, opened 2026-01-24
Fix/triton gemm small batch

anshuljoshi:
📝 Summary of Changes
Skip nested Triton GEMM conversion when symbolic tile analysis fails, avoiding a compile-time crash. It will fix this: https://github.com/jax-ml/jax/issues/34010. Also added a unit test covering the failure path.

🎯 Justification
This prevents XLA GPU compilation failures for small-batch (64, 32) CNN workloads where symbolic tile analysis fails in Triton GEMM conversion. It keeps Triton GEMM available for other supported cases.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

N/A (not a performance change).

🧪 Unit Tests:
Added SkipsConversionWhenTileAnalysisFails in `xla/service/gpu/transforms/convert_triton_gemm_config_test.cc`

🧪 Execution Tests:
None added yet.

### DIFF
--- a/xla/service/gpu/transforms/convert_triton_gemm_config.cc
+++ b/xla/service/gpu/transforms/convert_triton_gemm_config.cc
@@ -128,6 +128,16 @@ class ConvertTritonGemmConfigVisitor : public DfsHloRewriteVisitor {
       return absl::OkStatus();
     }
 
+    absl::StatusOr<BlockLevelParameters> block_level_parameters =
+        FindBlockLevelParameters(dot, config, mlir_context_,
+                                 device_description_);
+    if (!block_level_parameters.ok()) {
+      VLOG(2) << "Skipping Triton nested GEMM conversion for "
+              << fusion->name() << ": "
+              << block_level_parameters.status();
+      return absl::OkStatus();
+    }
+
     // Annotate the dot with the contraction tile size.
     ASSIGN_OR_RETURN(auto tile_sizes, dot->backend_config<Tile>());
     tile_sizes.add_sizes(config.block_k);
@@ -140,13 +150,8 @@ class ConvertTritonGemmConfigVisitor : public DfsHloRewriteVisitor {
         *gpu_config.mutable_fusion_backend_config();
     backend_config.clear_triton_gemm_config();
     backend_config.set_kind(kTritonNestedGemmFusionKind);
-
-    ASSIGN_OR_RETURN(BlockLevelParameters block_level_parameters,
-                     FindBlockLevelParameters(dot, config, mlir_context_,
-                                              device_description_));
-
     *backend_config.mutable_block_level_fusion_config() =
-        block_level_parameters.ToBlockLevelFusionConfig();
+        block_level_parameters->ToBlockLevelFusionConfig();
     RETURN_IF_ERROR(fusion->set_backend_config(gpu_config));
 
     MarkAsChanged();
--- a/xla/service/gpu/transforms/convert_triton_gemm_config_test.cc
+++ b/xla/service/gpu/transforms/convert_triton_gemm_config_test.cc
@@ -153,5 +153,45 @@ ENTRY entry {
 )"));
 }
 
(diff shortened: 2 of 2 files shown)


penpornk:
Please see the instructions below on how to sign the CLA:
https://github.com/openxla/xla/blob/main/docs/contributing.md#sign-the-contributor-license-agreement

anshuljoshi:
Hi @penpornk, I have already signed the CLA as per the guide. The CLA check is successful. Please let me know if something is still missing.

penpornk:
Hi @anshuljoshi,
I have manually checked your CLA. We've got it and you are good to go. Sorry for the confusion! I should have checked whether you have signed the CLA after @google-cla's message first.

anshuljoshi:
@penpornk no worries at all. Thanks.

anshuljoshi:
Hi @pifon2a, can you please review?

metaflow (changes requested):
Thank you for the PR! Have you checked that the jax issue will be resolved with this? I am confused as FindBlockLevelParameters call in convert_triton_gemm config is called behind experimental flag (that I would not recommend to use), you probably want to change the callsite in nest_gemm_fusion.cc - but that would also be quite tricky to do as we are rewriting HLO and can't go back at this point and will fail compilation.

As a workaround you can try to not use triton to emit this fusion: through the flags or using autotuner to find the configuration that will work. Is that an option?

anshuljoshi:
@metaflow Apologies, I was out for sometime.
Thanks for reviewing and your comments.

I realize now that `ConvertTritonGemmConfig` is not in the pipeline with the default flag configuration (`xla_gpu_unsupported_disable_nested_gemm_fusions=false`), so my fix in `convert_triton_gemm_config.cc` is unreachable for the default code path. The actual crash path goes through `NestGemmFusion` -> `MakeNestedFusionFromGemmFusion()` -> `FindBlockLevelParameters()`.

You are right that fixing `nest_gemm_fusion.cc` is tricky since the HLO has already been restructured by the time `FindBlockLevelParameters` is called. Would you consider the following options:
1. **Pre-check approach**: Call `FindBlockLevelParameters` in `NestGemmFusionVisitor::HandleFusion()` *before* calling `RewriteFusion()`. If it fails on the original computation, it can skip the fusion gracefully. This is conservative (the post-nesting analysis might succeed since nesting encapsulates ops like `pad` inside nested fusions) but it is safe as it never leaves the HLO in a partially modified state.
2. Your suggested workaround - configure the autotuner to avoid Triton for these fusions is a valid short-term solution. 

If you agree, I will update the PR to add the pre-check in `nest_gemm_fusion.cc`. As the upstream has since changed the architecture (both passes unconditionally registered, flag default changed to `true`, code moved to `backends/gpu/transforms/`), I will rebase to pick up those changes which would make the `convert_triton_gemm_config.cc` fix also effective for the new default path.

metaflow:
Hi @anshuljoshi ! I have not got a notification about this message from you. Will review changes today.

metaflow:
alright, so we going to keep the '__triton_gemm' fusion kind and skip rewriting the config. That is fine when looking at the pass in isolation. But what will happen next? We don't have the "legacy emitter" that understands this fusions. https://github.com/openxla/xla/blob/main/xla/backends/gpu/codegen/triton/xtile_compiler.cc#L282 explicitly rejects this fusion kind. So for me it seems that by not falling in rewrite config we are just pushing the error away for a few moments. Or maybe I miss something? Have you checked that the jax issue will be resolved with this?

It is a known issue that gemm_fusion sometimes produces fusion that cannot be emitter later as we have replaced the emitter. We are actively working on improving the gemm_fusion pass to be aligned with the emitter but it will take some time. Meanwhile disabling triton emitter path completely or updating the source model might be a workaround.

anshuljoshi:
Hi @metaflow, thanks for your response. It did solve the jax issue. Thanks for the context that you are actively working on improving the `gemm_fusion` pass to be aligned with the emitter. I will continue with your suggested workaround of disabling triton emitter path. I will now close this PR.

PR #36806, opened 2026-01-26
[XLA:CPU][oneDNN] Refactor oneDNN Resources in oneDNN Thunk

othakkar:
This PR refactors `OneDnnResources` to support resources for both oneDNN primitives and graph.


penpornk (approved):

PR #36826, opened 2026-01-26
[Rocm] Add flag to control hipblaslt swish activation fusion.

mfrancepillois:
📝 Summary of Changes
This PR adds a flag to allow users to disable this fusion if they wish.

🎯 Justification
We notice that performance of certain LLM models decreases when the hipblaslt swish activation fusion was enabled. We therefore add a new flag to allow users to choose when to use this fusion.

🚀 Kind of Contribution

🧪 Unit Tests:
2 tests have been added to test the flag behavior.

🧪 Execution Tests:
Tests pass.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -374,6 +374,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_nccl_collective_max_nchannels(0);
   opts.set_xla_gpu_nccl_p2p_max_nchannels(0);
   opts.set_xla_gpu_multi_streamed_windowed_einsum(true);
+  opts.set_xla_gpu_experimental_enable_cublaslt_swish_fusion(true);
 
   opts.set_xla_gpu_experimental_stream_annotation(true);
   // Minimum combined size of matrices in matrix multiplication to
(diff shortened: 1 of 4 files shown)


mfrancepillois:
@xla-rotation could you please review this PR?

xla/xla.proto
```diff
@@ -814,6 +814,8 @@ message DebugOptions {
   // rewrite).
   optional int64 xla_gpu_gemm_rewrite_size_threshold = 283;
 
+  optional bool xla_gpu_experimental_enable_cublaslt_swish_fusion = 502;
```
dimitar-asenov, line 817: Please explain in a comment in this file what a swish fusion is and that it's only available on certain AMD hardware (and also, why we might want to disable it).

xla/debug_options_flags.cc
```diff
@@ -2966,8 +2973,7 @@ FlagStatus GetFlagStatus(absl::string_view flag_name) {
           "xla_gpu_all_reduce_combine_threshold_bytes",
           "xla_gpu_autotune_level",
           "xla_gpu_collective_permute_decomposer_threshold",
-          "xla_gpu_cublas_fallback",
-          "xla_gpu_dot_merger_threshold_mb",
+          "xla_gpu_cublas_fallback", "xla_gpu_dot_merger_threshold_mb",
```
dimitar-asenov, line 2976: Please leave out this formatting change from the PR.

dimitar-asenov (changes requested):

xla/xla.proto
mfrancepillois: Thanks for reviewing this PR. Your comment has led me to reconsider the necessity of this flag. Initially, this flag was needed for performance reasons, but since PR https://github.com/openxla/xla/pull/37287 fixes the actual source of the observed performance regression, the flag is no longer necessary once PR https://github.com/openxla/xla/pull/37287 is merged. I therefore suggest we close this PR.

mfrancepillois:
PR https://github.com/openxla/xla/pull/37287 fixes the actual source of the observed performance regression. The flag introduced by this PR is no longer necessary. We can therefore close this PR.

PR #36970, opened 2026-01-28
[XLA][GPU] Hardcode memory clock and bus width values on Thor when driver returns 0

and-ivanov:
This is a temporary workaround (until the fix in CUDA driver) for a bug on Thor devices (CC 11.0) reporting mem_clock_khz and mem_bus_width_bits as 0.


xla/stream_executor/cuda/cuda_executor.cc
```diff
@@ -1812,6 +1812,15 @@ CudaExecutor::CreateDeviceDescription(int device_ordinal) {
   absl::StatusOr<int> mem_bus_width_bits = GetDeviceAttribute(
       CU_DEVICE_ATTRIBUTE_GLOBAL_MEMORY_BUS_WIDTH, device_ordinal);
   if (mem_clock_khz.ok() && mem_bus_width_bits.ok()) {
+    // Temporary fix when driver reports 0
```
ermilovmaxim: Please add comment about affected driver versions and when workaround can be safely removed

and-ivanov: added

pifon2a (approved):

Issue #36953, opened 2026-01-28
Add flag for showing inline stack frame in HLO output


Findus23:
This is unrelated to JAX, but I am using it to illustrate what I mean:

Since JAX 0.9.0 (so bb760b047bdbfeff962f0366ad5cc782c98657e0) and I think specifically since #34060 XLA seems to be using a more compact format for stack metadata in HLO output.

So e.g. in 
```python
import jax
from jax import numpy as jnp

jitted_fn = jax.jit(lambda x: (jnp.sin(x) / x).sum())

input = jnp.zeros(10)

comp = jitted_fn.lower(input).compile()
print(comp.as_text())
```
I get the following HLO output:
```mlir
HloModule jit__lambda, is_scheduled=true, entry_computation_layout={(f32[10]{0})->f32[]}, allow_spmd_sharding_propagation_to_parameters={true}, allow_spmd_sharding_propagation_to_output={true}

FileNames
1 "/.../tmptest.py"

FunctionNames
1 "<module>"
2 "<lambda>"

FileLocations
1 {file_name_id=1 function_name_id=1 line=8 end_line=8 column=7 end_column=29}
2 {file_name_id=1 function_name_id=2 line=4 end_line=4 column=30 end_column=52}
3 {file_name_id=1 function_name_id=2 line=4 end_line=4 column=31 end_column=41}

StackFrames
1 {file_location_id=1 parent_frame_id=1}
2 {file_location_id=2 parent_frame_id=2}
3 {file_location_id=3 parent_frame_id=2}


%region_0.1 (reduce_sum.3: f32[], reduce_sum.4: f32[]) -> f32[] {
  %reduce_sum.3 = f32[] parameter(0), metadata={op_name="reduce_sum"}
  %reduce_sum.4 = f32[] parameter(1), metadata={op_name="reduce_sum"}
  ROOT %reduce_sum.5 = f32[] add(%reduce_sum.3, %reduce_sum.4), metadata={op_name="jit(<lambda>)/reduce_sum" stack_frame_id=2}
}

%fused_computation (param_0.3: f32[10]) -> f32[] {
  %param_0.3 = f32[10]{0} parameter(0)
  %sin.0 = f32[10]{0} sine(%param_0.3), metadata={op_name="jit(<lambda>)/sin" stack_frame_id=3}
  %div.0 = f32[10]{0} divide(%sin.0, %param_0.3), metadata={op_name="jit(<lambda>)/div" stack_frame_id=3}
[...]

PR #36975, opened 2026-01-28
[HLO] Add option to print inline stack frames. 

hittanshubhanderi20:
Issue No.: #36953 

📝 Summary of Changes 
Added a new debug option xla_hlo_print_inline_stack_frames to DebugOptions. When enabled, the HLO printer resolves stack_frame_ids and prints the source file and line number directly inline with the instruction's metadata. This bypasses the default behavior of printing compact stack frame IDs that reference a separate index.

🎯 Justification 
Current HLO output separates stack frame information into a specific section at the end of the module to reduce file size. While efficient, this makes debugging difficult for users (especially JAX/Python users) who need to quickly correlate HLO instructions with their source code location. This change provides an opt-in flag to restore the inline display of source locations, significantly improving the debugging experience for manual HLO inspection.

🚀 Kind of Contribution 
✨ New Feature

📊 Benchmark N/A - This is a debug-only feature affecting text dumping.

🧪 Unit Tests: Manual verification was performed to ensure the flag correctly triggers the inline printing logic in HloInstruction::ToString. No new unit tests were added as the change relies on GetModule() access which is standard pattern in HloInstruction.

🧪 Execution Tests: N/A

### DIFF
--- a/xla/hlo/ir/hlo_instruction.cc
+++ b/xla/hlo/ir/hlo_instruction.cc
@@ -4074,10 +4074,25 @@ void HloInstruction::PrintWithCanonicalNameMap(
       (metadata_ != nullptr &&
        (!metadata_->op_type().empty() || !metadata_->op_name().empty() ||
         !metadata_->source_file().empty() ||
-        !metadata_->scheduling_name().empty()))) {
+        !metadata_->scheduling_name().empty() ||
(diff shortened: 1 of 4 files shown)


seantalts (approved):

pifon2a (approved):
Thank you!

PR #36981, opened 2026-01-28
[xla:gpu] Fix comm split deadlocks

ezhulenev:
Fix deadlocks as a result of calling (or not calling!) communicator split APIs on different ranks

1. `participant_groups` (renamed to `device_groups`) is not a property of the `GpuClique` itself, it only makes sense during communicator acquisition call
2. We can safely do the split **ONLY** if all ranks of the split-from clique will joint the process, so make it explicit check!
3. Add a little bit more logging to help with debugging comm split deadlocks.
4. Add timing to logging for individual communicators initialization


xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -724,6 +743,28 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
                              {"clique_key", clique_key}});
   });
 
+  // Maybe find if we acquired a clique with communicators that we can split.
+  static const int64_t enable_nccl_comm_splitting =
+      GetDebugOptionsFromFlags().xla_gpu_enable_nccl_comm_splitting();
+
+  // To be able to split from existing clique, clique key must be a subset of
+  // the already acquired clique, and device groups must cover all devices in
+  // the split-form clique, as all ranks must make a split call. XLA currently
```
mwhittaker:
```suggestion
  // the split-from clique, as all ranks must make a split call. XLA currently
```

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -743,10 +784,39 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
...
+              //   (1) [0,1] do an all-reduce (device_groups=[[0,1]])
+              //   (2) [0,1,2,3] do an all-reduce (device_groups=[[0,1,2,3]])
+              //   (3) [0,1,2,3] do an all-reduce (device_groups=[[0,1],[2,3]])
+              //
+              // When (3) will be executed, ranks [0,1] will already have a
+              // clique acquired earlier, but ranks [2,3] will try to initiate
+              // a communicator split. This will lead to deadlock.
+              if (split_from &&
+                  it->second.created() < (*split_from)->created()) {
+                state.cliques.erase(clique_key);
```
mwhittaker, line 861: Do we need to abort this clique before we erase it? What if there is a pending operation on the clique?

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -743,10 +784,39 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
...
+              }
+
+              // If the clique is stale (one one participating processes already
+              // died), we return a nullptr, as the clique will have to be
+              // re-created.
               absl::Status stale =
                   CheckCliqueIsntStaleImpl(state.task_state_infos, clique_key);
-              return it == state.cliques.end() || !stale.ok() ? nullptr
-                                                              : &it->second;
+              return !stale.ok() ? nullptr : &it->second;
```
mwhittaker:
```suggestion
              return stale.ok() ? &it->second : nullptr;
```

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -743,10 +784,39 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
...
+              //
+              //   (1) [0,1] do an all-reduce (device_groups=[[0,1]])
+              //   (2) [0,1,2,3] do an all-reduce (device_groups=[[0,1,2,3]])
+              //   (3) [0,1,2,3] do an all-reduce (device_groups=[[0,1],[2,3]])
+              //
+              // When (3) will be executed, ranks [0,1] will already have a
+              // clique acquired earlier, but ranks [2,3] will try to initiate
+              // a communicator split. This will lead to deadlock.
+              if (split_from &&
+                  it->second.created() < (*split_from)->created()) {
```
mwhittaker: I'm probably missing something obvious, but why are we checking the creation time of the two cliques? In your example, Eugene, imagine we flipped the order of (1) and (2):

```
//   (2) [0,1,2,3] do an all-reduce (device_groups=[[0,1,2,3]])
//   (1) [0,1] do an all-reduce (device_groups=[[0,1]])
//   (3) [0,1,2,3] do an all-reduce (device_groups=[[0,1],[2,3]])
```

Then `it->second.created() > (*split_from)->created()` and we would `return it->second`, but wouldn't this  lead to a deadlock where devices 0 and 1 are re-using an existing clique and devices 2 and 3 are trying to split?

xla/backends/gpu/runtime/collective_clique_requests.cc
```diff
@@ -19,20 +19,47 @@ limitations under the License.
...
+  // the first device. We need this for determenistic check below.
+  absl::c_for_each(device_groups, [](auto& group) { absl::c_sort(group); });
+  absl::c_sort(device_groups, [](const auto& a, const auto& b) {
+    CHECK(!a.empty() && !b.empty()) << "Replica groups must not be empty";
+    return a[0] < b[0];
+  });
+
+  VLOG(5) << absl::StreamFormat(
+      "Add collective clique request: %v; device_groups=[%s]; "
+      "requirementes=%v",
```
mwhittaker:
```suggestion
      "requirements=%v",
```

xla/backends/gpu/collectives/gpu_cliques.cc
ezhulenev: It will be owned by whoever still runs with it, and will successfully finish execution, and will be destroyed after that. And all next executions will run re-acquired clique. All of that needs better unit testing, it has very light coverage today, I'll send some PRs later this or next week.

xla/backends/gpu/collectives/gpu_cliques.cc
ezhulenev: Ugh, yes, this is a problem, we need to tack clique provenance :( I'll think how to do it properly.

ezhulenev: So I added an explicit `parent` field to `GpuClique` and now can safely decide if all ranks will split from the same parent... I think, it all seems to work with SPMD programming model, but I still can't convince myself that it also works with MPMD and Jax PP, but I'll find out soon. Anyway it's certainly a better state than today as our internal workloads do not deadlock anymore.

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -106,7 +107,7 @@ struct ProcessGpuCliques {
   absl::Mutex mu;
 
   // Constructed GPU cliques, keyed by GpuCliqueKey.
-  absl::node_hash_map<GpuCliqueKey, LockableGpuClique> cliques
+  absl::node_hash_map<GpuCliqueKey, std::shared_ptr<LockableGpuClique>> cliques
```
mwhittaker: Can we change `absl::node_hash_map` to an `absl::flat_hash_map` now?

xla/backends/gpu/collectives/gpu_clique.h
```diff
@@ -44,7 +44,8 @@ class GpuClique : public Clique {
   GpuClique(
       GpuCliqueKey key, std::optional<CliqueIds> ids,
       absl::btree_map<RankId, std::unique_ptr<Communicator>> communicators,
-      bool peer_access_enabled, std::shared_ptr<CancellationToken> cancel);
+      bool peer_access_enabled, std::shared_ptr<CancellationToken> cancel,
+      const GpuClique* parent = nullptr);
```
mwhittaker, line 48: Sorry, I'm having trouble finding where this constructor is called with a non-null `parent`. Could you point me to it?

xla/backends/gpu/collectives/gpu_cliques.cc
mwhittaker: Is it possible for a process to fail while it is executing, and so it will not successfully finish execution? If so, I think we would need to abort it?

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -743,10 +810,38 @@ absl::StatusOr<std::shared_ptr<LockableGpuClique::Lock>> AcquireGpuClique(
...
+              // Example: imagine 4 distributed processes
+              //
+              //   (1) [0,1] do an all-reduce (device_groups=[[0,1]])
+              //   (2) [0,1,2,3] do an all-reduce (device_groups=[[0,1,2,3]])
+              //   (3) [0,1,2,3] do an all-reduce (device_groups=[[0,1],[2,3]])
+              //
+              // When (3) will be executed, ranks [0,1] will already have a
+              // clique acquired earlier, but ranks [2,3] will try to initiate
+              // a communicator split. This will lead to deadlock.
+              if (it->second->HasParent((*split_from)->parent())) {
```
mwhittaker: Hm, I'm having trouble understanding this too. I thought `it->second` is clique (1) in your example, and `split_from` is clique (2) in your example? But neither of these cliques have parents?

xla/backends/gpu/collectives/gpu_cliques.cc
ezhulenev: Done.

xla/backends/gpu/collectives/gpu_clique.h
ezhulenev: Line 703 when we split from parent (`/*parent=*/&**parent_clique`)

xla/backends/gpu/collectives/gpu_cliques.cc
ezhulenev: Yes, my bad, not sure how it even passed the test. Updated the code and introduced variables with names + logging.

xla/backends/gpu/collectives/gpu_cliques.cc
ezhulenev: Yes. I added an explicit locking step to make sure that all other users are done, and then it's safe to erase it + added logging + comments explaining what's going on.

mwhittaker: You mean the lock on line 812:

```
absl::MutexLock lock(state.mu);
```

or the Acquire on line 857?

```
return lockable_clique ? lockable_clique->Acquire()
                       : LockableGpuClique::Lock();
```

Because I thought `state.mu` is not held during a collective, so grabbing it doesn't guarantee that a collective isn't ongoing. And we acquire the lockable clique only after we've erased it?

ezhulenev: `auto safety_lock = to_lock->Acquire();` on line 848, we acquire the shared lock for a clique in the per-process state, and in means that no one else is using it. Only after that we can safely erase it and return nullptr to the caller.

ezhulenev: Added a bit more details to comment.

ezhulenev: I don't think that in practice we can even have a concurrent execution with collectives, it will quickly lead to deadlocks, especially as clique cancellation across ranks requires cross-process coordination that we conviniently decided non to do.

Assuming that all processes execute identical JAX programs on all ranks, and they all will acquire the same set of cliques in the same exact order, it should work.

In practice we worry about previous "sequential" JAX execution to be fully completed once we start the next one, and current mechanism prevents this well. The worlds of many concurrent requests... I hope we don't really need it, especially across multiple processes.

mwhittaker (approved):
Looks great, Eugene! I know the code is very subtle and not easy to change, so I appreciate you working through all the subtlety and improving things.

pifon2a (approved):

pifon2a (approved):

pifon2a (approved):

hawkinsp:
I bisected a deadlock in JAX on 8xH100 to this PR:

https://github.com/jax-ml/jax/actions/runs/23050465655/job/66950084165

Repro via JAX test suite:
```
python tests/pjit_test.py ArrayPjitTest.test_prng_sharding_propagation_with_nested_jit ShardingInTypesTest.test_auto_gather_out_sharding ShardingInTypesTest.test_tile
```

I asked Gemini to try to fix it and it came up with this. Caution: I have not verified this makes any sense whatsoever, but I did verify it fixes the original problem, and that the test case here fails without the fix and passes with.

> 
> The Bug (Deadlock Root Cause)
> The bug was caused by an inconsistent caching state across different subgroups participating in an ncclCommSplit operation.
> 
> When a root communicator (e.g. [0,1,2,3]) is split into subgroups (e.g. [0,1] and [2,3]), XLA tries to reuse previously cached cliques for those subgroups if they exist and mathematically map to the parent.
> 
> The hang happens when:
> 
> Subgroup [0, 1] finds a matching cached clique with the correct parent, so it decides to bypass the split and simply reuse the cache.
> Subgroup [2, 3] does not have a cache hit with the correct parent (e.g., if its cache was overwritten or evicted). Since it has no valid cache, it decides to participate in the NCCL split operation.
> ncclCommSplit is a globally synchronous operation that requires all devices in the root parent communicator [0,1,2,3] to participate. Because devices [0, 1] reused their cache and skipped the split call, devices [2, 3] hang indefinitely waiting for [0, 1] to enter the collective.
> 
> The Fix
> The fix prevents this by enforcing an "All or Nothing" policy across all participating subgroups.
> 
> Before deciding whether to abandon or reuse a cached clique, the code now checks the cache validity for all subgroups in the split plan (i.e. both [0,1] and [2,3]). If any subgroup is missing a valid cached clique, it forces every subgroup to abandon their caches and participate in the full ncclCommSplit operation together. This ensures that either everyone reuses a cache, or everyone splits, eliminating the mismatch that causes the NCCL deadlock.

```
diff --git a/xla/backends/gpu/collectives/gpu_cliques.cc b/xla/backends/gpu/collectives/gpu_cliques.cc
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -837,30 +837,62 @@ absl::StatusOr<std::shared_ptr<LockableG
               // When (3) will be executed, ranks [0,1] will already have a
               // clique acquired earlier, but ranks [2,3] will try to initiate
               // a communicator split. This will lead to deadlock.
-              if (split_from && !to_lock->HasParent(&**split_from)) {
-                VLOG(5) << absl::StrFormat(
-                    "Abandon previously acquired clique %v to re-acquire it by "
-                    "splitting from %v",
-                    clique_key, *split_from_key);
+              if (split_from) {
+                bool should_abandon = false;
+                for (const auto& group : device_groups) {
+                  int64_t group_local_participants = 0;
+                  for (GlobalDeviceId dev : group) {
+                    std::optional<RankId> parent_rank =
+                        (*split_from)->key().rank(dev);
+                    if (parent_rank && (*split_from)->comm(*parent_rank)) {
+                      group_local_participants++;
+                    }
+                  }
+
+                  if (group_local_participants == 0) continue;

-                // First we have to acquire a lock to this clique, to make sure
-                // that all other users of this clique finished execution.
-                //
-                // `LockableGpuClique::Lock` is acquired by GPU executable
-                // before the program starts running, and automatically released
-                // when execution is done. By waiting for lock here, we are
-                // essentially waiting for completion of the other run that
-                // uses `to_lock` clique (if such run exists). We do not release
-                // `state.mu` as we don't want any concurrent modification of
-                // the per-proces state.
-                auto safety_lock = to_lock->Acquire();
+                  // If it has local communicators, check if it's cached and
+                  // valid.
+                  std::vector<IncarnationId> incarnations(
+                      clique_key.incarnations().begin(),
+                      clique_key.incarnations().end());
+                  GpuCliqueKey group_key(group, group_local_participants,
+                                         clique_key.is_p2p(), incarnations);
+                  auto group_it = state.cliques.find(group_key);
+                  if (group_it == state.cliques.end() ||
+                      !group_it->second->HasParent(&**split_from)) {
+                    should_abandon = true;
+                    break;
+                  }
+                }

-                // Only when we know that the clique doesn't have any other
-                // users we can delete it. It will be destroyed when `to_lock`
-                // will get out of scope.
-                state.cliques.erase(clique_key);
+                if (should_abandon) {
+                  VLOG(5) << absl::StrFormat(
+                      "Abandon previously acquired clique %v to re-acquire it "
+                      "by "
+                      "splitting from %v",
+                      clique_key, *split_from_key);

-                return nullptr;
+                  // First we have to acquire a lock to this clique, to make
+                  // sure that all other users of this clique finished
+                  // execution.
+                  //
+                  // `LockableGpuClique::Lock` is acquired by GPU executable
+                  // before the program starts running, and automatically
+                  // released when execution is done. By waiting for lock here,
+                  // we are essentially waiting for completion of the other run
+                  // that uses `to_lock` clique (if such run exists). We do not
+                  // release `state.mu` as we don't want any concurrent
+                  // modification of the per-proces state.
+                  auto safety_lock = to_lock->Acquire();
+
+                  // Only when we know that the clique doesn't have any other
+                  // users we can delete it. It will be destroyed when `to_lock`
+                  // will get out of scope.
+                  state.cliques.erase(clique_key);
+
+                  return nullptr;
+                }
               }

               // If the clique is stale (one one participating processes already
diff --git a/xla/backends/gpu/collectives/gpu_cliques_test.cc b/xla/backends/gpu/collectives/gpu_cliques_test.cc
--- a/xla/backends/gpu/collectives/gpu_cliques_test.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques_test.cc
@@ -301,4 +301,139 @@ TEST(GpuCliquesTest, SplitCliquesNoDeadl
   }
 }

+TEST(GpuCliquesTest, SplitCliquesNoDeadlock2) {
+  auto cleanup = absl::MakeCleanup([] { internal::DestroyAcquiredCliques(); });
+
+  ASSERT_OK_AND_ASSIGN(se::Platform * platform,
+                       se::PlatformManager::PlatformWithName("CUDA"));
+
+  if (platform->VisibleDeviceCount() < 4) {
+    GTEST_SKIP() << "Test requires at least 4 GPUs";
+  }
+
+  tsl::thread::ThreadPool pool(tsl::Env::Default(), "collectives", 4);
+  tsl::Executor& exec = *pool.AsExecutor();
+
+  RunId run_id(0);
+
+  GpuCliqueKey key0123({kD0, kD1, kD2, kD3}, 4);
+  GpuCliqueKey key01({kD0, kD1}, 2);
+  GpuCliqueKey key23({kD2, kD3}, 2);
+
+  std::vector<std::vector<GlobalDeviceId>> group01 = {{kD0, kD1}};
+  std::vector<std::vector<GlobalDeviceId>> group0123 = {{kD0, kD1, kD2, kD3}};
+  std::vector<std::vector<GlobalDeviceId>> group01_23 = {{kD0, kD1},
+                                                         {kD2, kD3}};
+
+  ASSERT_OK_AND_ASSIGN(std::vector<se::StreamExecutor*> executors_vec,
+                       CreateExecutors(platform, 4));
+  absl::Span<se::StreamExecutor*> executors(executors_vec);
+
+  std::vector<AcquiredCliquesMap> acquired_cliques_vec(4);
+  absl::Span<AcquiredCliquesMap> acquired_cliques(acquired_cliques_vec);
+
+  // (1) [0, 1, 2, 3] do an all-reduce.
+  {
+    auto futures =
+        AcquireCliques(exec, executors, key0123, group0123, acquired_cliques);
+    ASSERT_OK_AND_ASSIGN(auto cliques, WaitCliques(std::move(futures)));
+
+    for (size_t i = 0; i < 4; ++i) {
+      acquired_cliques.at(i).emplace(key0123, cliques.at(i));
+    }
+  }
+
+  // (2) [0, 1, 2, 3] do an all-reduce SPLITTING FROM [0, 1, 2, 3].
+  // Populates cache for [0, 1] and [2, 3] WITH the parent [0, 1, 2, 3].
+  {
+    auto futures0 = AcquireCliques(exec, executors.first(2), key01, group01_23,
+                                   acquired_cliques.first(2));
+    auto futures1 = AcquireCliques(exec, executors.last(2), key23, group01_23,
+                                   acquired_cliques.last(2));
+    ASSERT_OK_AND_ASSIGN(auto cliques0, WaitCliques(std::move(futures0)));
+    ASSERT_OK_AND_ASSIGN(auto cliques1, WaitCliques(std::move(futures1)));
+
+    for (size_t i = 0; i < 2; ++i) {
+      acquired_cliques.at(i).emplace(key01, cliques0.at(i));
+    }
+    for (size_t i = 0; i < 2; ++i) {
+      acquired_cliques.at(i + 2).emplace(key23, cliques1.at(i));
+    }
+  }
+
+  // MUST drop the locks from acquired_cliques so step 3 can re-acquire without
+  // test deadlock!
+  for (size_t i = 0; i < 4; ++i) {
+    acquired_cliques.at(i).clear();
+  }
+
+  // (3) Create a NEW root clique [0, 2, 3] and split [0] and [2, 3] from it!
+  // This forces [2, 3] to abandon its cache (which has parent [0,1,2,3]) and
+  // recreate it with parent [0,2,3]. But [0, 1]'s cache is UNTOUCHED and
+  // retains parent [0,1,2,3]!
+  {
+    std::vector<GlobalDeviceId> dev023_vec = {kD0, kD2, kD3};
+    GpuCliqueKey key023(dev023_vec, 3, false, {});
+    std::vector<std::vector<GlobalDeviceId>> group_0_23 = {{kD0}, {kD2, kD3}};
+    GpuCliqueKey key0({kD0}, 1, false, {});
+
+    std::vector<se::StreamExecutor*> exec023_vec = {executors[0], executors[2],
+                                                    executors[3]};
+    absl::Span<se::StreamExecutor* const> exec023(exec023_vec);
+    std::vector<AcquiredCliquesMap> acquired023_vec(3);
+    absl::Span<AcquiredCliquesMap> acquired023(acquired023_vec);
+
+    auto futures_root =
+        AcquireCliques(exec, exec023, key023, {dev023_vec}, acquired023);
+    ASSERT_OK_AND_ASSIGN(auto cliques_root,
+                         WaitCliques(std::move(futures_root)));
+    for (size_t i = 0; i < 3; ++i) {
+      acquired023.at(i).emplace(key023, cliques_root.at(i));
+    }
+
+    auto futures_0 = AcquireCliques(exec, exec023.subspan(0, 1), key0,
+                                    group_0_23, acquired023.subspan(0, 1));
+    auto futures_23 = AcquireCliques(exec, exec023.subspan(1, 2), key23,
+                                     group_0_23, acquired023.subspan(1, 2));
+    ASSERT_OK(WaitCliques(std::move(futures_0)).status());
+    ASSERT_OK(WaitCliques(std::move(futures_23)).status());
+  }
+
+  // Ensure all locks are dropped before step 4!
+  for (size_t i = 0; i < 4; ++i) {
+    acquired_cliques.at(i).clear();
+  }
+
+  // (4) Re-acquire [0, 1, 2, 3] so we can split from it!
+  {
+    auto futures =
+        AcquireCliques(exec, executors, key0123, group0123, acquired_cliques);
+    ASSERT_OK_AND_ASSIGN(auto cliques, WaitCliques(std::move(futures)));
+    for (size_t i = 0; i < 4; ++i) {
+      acquired_cliques.at(i).clear();
+      acquired_cliques.at(i).emplace(key0123, cliques.at(i));
+    }
+  }
+
+  // (5) [0, 1, 2, 3] do an all-reduce SPLITTING FROM [0, 1, 2, 3].
+  // Before fix:
+  // - [0, 1] finds cached clique (from step 2) with NO PARENT mismatch ->
+  // REUSES.
+  // - [2, 3] finds cached clique (from step 3) WITH PARENT mismatch -> ABANDONS
+  // & SPLITS.
+  // -> DEADLOCK!
+  // After fix:
+  // - [0, 1] and [2, 3] check both subgroups. Subgroup [2, 3] violates parent
+  // match.
+  // - Therefore BOTH [0, 1] and [2, 3] ABANDON & SPLIT safely!
+  {
+    auto futures0 = AcquireCliques(exec, executors.first(2), key01, group01_23,
+                                   acquired_cliques.first(2));
+    auto futures1 = AcquireCliques(exec, executors.last(2), key23, group01_23,
+                                   acquired_cliques.last(2));
+    ASSERT_OK_AND_ASSIGN(auto cliques0, WaitCliques(std::move(futures0)));
+    ASSERT_OK_AND_ASSIGN(auto cliques1, WaitCliques(std::move(futures1)));
+  }
+}
+
 }  // namespace xla::gpu
```

ezhulenev:
It's roughly correct. but I think subtly wrong :) https://github.com/openxla/xla/pull/38105 is an attempt to fix and https://github.com/openxla/xla/pull/37063#pullrequestreview-3947930784 is another one, and the ROCM version looks correct to me.

PR #37024, opened 2026-01-30
[xla:gpu] NCCL DevComm: add e2e test for symmetric memory and device comms

ezhulenev:
Allocate collected symmetric memory requests during executable initialization stage. Node that every buffer allocation becomes a separate symmetric memory, and multiple symmetric address ranges (i.e. if model has multiple collectives) will be mapped to symmetric memory just once.

This is still far from optimal, as symmetric allocations are not cached, however judging from timing it is very fast (~10ms) on a single host. Will look into optimization performance once collect data from larger clusters.

Collective memory follows collective cliques request/acquire model:
1. Thunks request device addresses to become symmetric
2. GpuExecutable acquires symmetric memory
3. At run time Thunks can fetch symmetric memory corresponding to the requested device address.


xla/service/gpu/gpu_executable.cc
```diff
@@ -509,6 +510,13 @@ absl::Status ExecuteThunksImpl(
                                               collective_clique_requests));
   }
 
+  // Acquire collective memories requested by thunks.
+  ASSIGN_OR_RETURN(
+      CollectiveMemory collective_memory,
+      AcquireCollectiveMemory(collective_params, collective_cliques,
+                              collective_memory_requests));
+  (void)collective_memory;
```
PatriosTheGreat: Is this line needed?

xla/backends/gpu/runtime/collective_memory.cc
```diff
@@ -0,0 +1,153 @@
...
+    //    them.
+    // 3. Cache symmetric memories in a process-level cache.
+    //
+    // Currently it's very simple proof of concept.
+
+    ASSIGN_OR_RETURN(GpuCommunicator * comm, cliques.GetComm(r.key, *rank));
+    for (BufferAllocation::Index i : r.allocations) {
+      ASSIGN_OR_RETURN(
+          std::unique_ptr<SymmetricMemory> symm,
+          comm->CreateSymmetricMemory(requests.buffers().GetDeviceAddress(i)));
```
PatriosTheGreat, line 138: There is a implicit limitation here that the buffer should be allocated by the VMM API (in this case most likely via ncclMemAlloc). So the buffer assigner should mark buffers used for symmetrical memory as a collective buffers and the copy should be emitted for an input parameters of the module, since we can't guarantee that they were allocated in the collectives memory space.

Since this PR is a PoC, we can fix this as a follow-up.

xla/backends/gpu/runtime/BUILD
```diff
@@ -2081,6 +2082,34 @@ xla_cc_test(
     ],
 )
 
+cc_library(
```
PatriosTheGreat, line 2152: Please add compatible_with = get_compatible_with_portable(), property to this target.

xla/service/gpu/gpu_executable.cc
```diff
@@ -808,9 +817,21 @@ absl::StatusOr<se::DeviceAddressBase> GpuExecutable::BufferForAllocation(
...
     se::DeviceAddressBase buffer_address;
     if (buffer_size > 0) {
+      // BFC allocator ingores memory alignment and always allocates 256 byte
+      // aligned buffers, however for collective memory underlying libraries
+      // require larger alignment. We conservatively round up all allocation
+      // sizes to 4096 bytes. Proper fix must be done in BFC allocator and all
+      // the other allocator adaptors that we have in XLA, but this is left
+      // as an excercise for curious reader. The raw memory allocator that backs
+      // the BFC allocator uses correct granularity and alignment.
+      static constexpr int64_t kCollectiveMemoryColor = 1;
```
PatriosTheGreat: Please use existing MemorySpaceColor::kCollective

xla/service/gpu/gpu_executable.cc
```diff
@@ -808,9 +817,21 @@ absl::StatusOr<se::DeviceAddressBase> GpuExecutable::BufferForAllocation(
...
+      static constexpr int64_t kCollectiveGranularity = 4096;
```
PatriosTheGreat: This is a NCCL specific requirement (I assume this corresponds to NCCL_REC_PAGE_SIZE). Could we at least hide this behind GPUCollectives?

PatriosTheGreat (changes requested):

xla/service/gpu/gpu_executable.cc
ezhulenev: Removed.

xla/backends/gpu/runtime/collective_memory.cc
ezhulenev: Yes, my assumption that buffer coloring was already done by compiler, and parameters were copied into the correct memory space. I think we need memory space to be a part of `DeviceAddressBase`, so we can pass it to FFI handlers, so that FFI can safely return an error if address range is in the wrong space.

It's a huge change, but the right one. I'll check what it would take.

xla/backends/gpu/runtime/BUILD
ezhulenev: It can't be compatible with portable because it has non-compatible dependencies, had this problem in earlier PR. It needs a larger cleanup of all transitive deps, and I can't reproduce it in OSS :(

xla/service/gpu/gpu_executable.cc
ezhulenev: I wanted to do that, but it brings a ton of compiler dependencies into the runtime (transitive deps via BufferAssignemnt). It needs a separate change to extract all enums.

xla/service/gpu/gpu_executable.cc
ezhulenev: Done. Added a virtual API to GpuCollectives.

ezhulenev: Also 4096 bytes seems to be enough, NCCL_REC_PAGE_SIZE is much larger (4mb), and documentation says 32 bytes or 256 bytes, can't find what is the real requirement.

xla/backends/gpu/runtime/collective_cliques.cc
```diff
@@ -117,9 +146,9 @@ absl::StatusOr<CollectiveCliques> AcquireCollectiveCliques(
     const CollectiveCliqueRequests::CliqueRequest& r = ordered_cliques[i];
     VLOG(2) << absl::StreamFormat(
         "[%d]    clique #%d (global device %v): "
```
PatriosTheGreat: Could you please here and everywhere where device_id logging is needed use XLA_LOG_DEVICE?
This allows us to change format quickly

xla/backends/gpu/collectives/nccl_collectives.cc
```diff
@@ -228,6 +228,12 @@ bool NcclCollectives::SupportsDeviceComm() const {
   return NCCL_VERSION_CODE >= 22800;
 }
 
+size_t NcclCollectives::SymmetricMemoryAlignment() const {
+  // TODO(ezhulenev): Query memory requirements from NCCL. Alignment might not
+  // be the only one.
+  return 4096;
```
PatriosTheGreat, line 234: We need to query [CU_MULTICAST_GRANULARITY_MINIMUM](https://docs.nvidia.com/cuda/archive/12.6.1/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggcdeff171670a788001418262a0f88378790ecc4b1f2c2f6e2c9bc21b230872a9) here, since this alignment is platform-dependent.

We are already doing this in cuda_executor for recommended granularity during multicast object construction, so I assume this should also be done by stream_executor.

We can do this as a follow up, just add a comment here for now.

xla/service/gpu/gpu_executable.cc
```diff
@@ -843,23 +859,49 @@ static absl::Status CheckAlignment(const BufferAllocation& allocation,
...
+    const ServiceExecutableRunOptions* run_options, VariantArguments arguments,
     const GpuExecutable::BufferAllocToDeviceMemoryMap* globals,
     se::DeviceAddressAllocator* const memory_allocator, int device_ordinal) {
   tsl::profiler::TraceMe hlo_module_activity(
       [&] { return std::string("Build buffer allocations"); },
       tsl::profiler::TraceMeLevel::kInfo);
 
+  absl::flat_hash_map<LogicalBuffer::Color, int64_t> allocate_granularity;
+  if (auto* collectives = ResolveGpuCollectives(run_options)) {
+    // BFC allocator ingores memory alignment and always allocates 256 byte
```
PatriosTheGreat: Nit: typo, ignores

xla/service/gpu/gpu_executable.cc
```diff
@@ -843,23 +859,49 @@ static absl::Status CheckAlignment(const BufferAllocation& allocation,
...
+    // aligned buffers, however for collective memory underlying libraries
+    // require larger alignment. We conservatively round up all allocation
+    // sizes to the alignment requirement. Proper fix must be done in BFC
+    // allocator and all the other allocator adaptors that we have in XLA, but
+    // this is left as an excercise for curious reader. The raw memory allocator
```
PatriosTheGreat: Nit: typo, exercise

xla/service/gpu/gpu_executable.cc
```diff
@@ -509,6 +511,12 @@ absl::Status ExecuteThunksImpl(
                                               collective_clique_requests));
   }
 
+  // Acquire collective memories requested by thunks.
+  ASSIGN_OR_RETURN(
+      CollectiveMemory collective_memory,
+      AcquireCollectiveMemory(collective_params, collective_cliques,
```
PatriosTheGreat, line 517: Could you add todo here to cover this with the test?
We might be able to check this with one-shot thunks after multimem migration.

PatriosTheGreat (changes requested):
Overall looks good, I only have a few nits.

xla/backends/gpu/runtime/collective_cliques.cc
ezhulenev: Done.

xla/backends/gpu/collectives/nccl_collectives.cc
ezhulenev: Added TODO, I'm not 100% sure that NCCL and multicast have same requirements, though I suspect they probably are the same.

xla/service/gpu/gpu_executable.cc
ezhulenev: fixed

xla/service/gpu/gpu_executable.cc
ezhulenev: done

xla/service/gpu/gpu_executable.cc
ezhulenev: It's covered by ffi test today, though only with latest NCCL, and it's covered by multimem test in next PR.

penpornk (approved):
Approving since @PatriosTheGreat's last round of comments were all addressed.

penpornk (approved):
Approving again for the build fix.

penpornk (approved):

PR #37026, opened 2026-01-30
[XLA:GPU][oneAPI][Bugfix] Fix scalar or empty global constants.

mdfaijul:
The XLA embeds (initializes) scalar or empty global constants in the LLVM IR, regardless of the platform ([llvm_emitter.cc](https://github.com/openxla/xla/blob/963d2228023fb9b887c89b7c77763b9156847ef4/xla/backends/gpu/codegen/llvm/llvm_emitter.cc#L700)). On Sycl platform, since unused constants in a SPIRV module are removed by the level zero API, `GpuExecutable` fails to resolve constants by invoking `StreamExecutor::GetSymbol` API. Global constants more than one elements do not have issue as `GpuExecutable` uses zero initializers and copies the contents of constants to shared allocated buffers.

This PR fixes the issue by using zero initializers for scalar and empty global constants when the backend is SPIRV (i.e., sycl platform).

Lot of tests fail on SYCL platform without this fix. A minimal test that is covered by this PR is `//xla/tests:constants_test`.


mdfaijul:
The fix does not apply with recent changes.

PR #37063, opened 2026-01-30
[ROCM] Invalidate cached subset cliques to fix NCCL comm split dea…

phambinhfin:
…dlock

When `xla_gpu_enable_nccl_comm_splitting` is enabled (default), it is possible for a deadlock to occur if a subset clique was already created and cached, and those devices reuse it from the clique map, while other devices initiate a split and wait forever for the cached devices to join.
For example, this script produces the deadlock
When 

1. 2-GPU operations run with different mesh axes (DP, then TP)
2. 4-GPU operations run afterward
3. Sharding patterns trigger collectives on hidden dimension

```
"""
import os
os.environ["JAX_TRACEBACK_FILTERING"] = "off"

import jax
import jax.numpy as jnp
import numpy as np
from jax import random
from jax.sharding import Mesh, NamedSharding, PartitionSpec as P
from jax._src.pjit import pjit

jax.config.update("jax_use_shardy_partitioner", False)


def run_test(n_devices, mesh_shape, axes, dp_res, tp_res):
    """Run pure JAX operations with specific sharding to trigger collectives."""
    print(f"Running {n_devices}-GPU test with axes={axes}...")
    
    devices = np.array(jax.devices()[:n_devices]).reshape(mesh_shape)
    mesh = Mesh(devices, axes)
    
    # Shard on hidden dim (like bad_sharding=True) to force collectives
    x_spec = P(dp_res, None, None, tp_res)
    
    x = random.normal(random.PRNGKey(0), [32, 12, 128, 128], jnp.float16)
    
    # Function 1: softmax + reduction (forces all-reduce)
    def fn1(x):
        y = jax.nn.softmax(x, axis=-1)  # softmax on sharded dim
        return jnp.mean(y)
    
    # Function 2: different pattern - layer norm style
    def fn2(x):
        mean = jnp.mean(x, axis=-1, keepdims=True)
        var = jnp.var(x, axis=-1, keepdims=True)
        y = (x - mean) / jnp.sqrt(var + 1e-5)
        return jnp.mean(y)
    
    with mesh:
        x_ = jax.device_put(x, NamedSharding(mesh, x_spec))
        
        # Run first function with grad
        jit1 = pjit(jax.value_and_grad(fn1), in_shardings=x_spec, out_shardings=(None, x_spec))
        fwd1, grad1 = jit1(x_)
        
        # Get HLO (triggers compilation)
        _ = jit1.lower(x_).compile().as_text()
        
        # Run second function with grad - different collective pattern
        jit2 = pjit(jax.value_and_grad(fn2), in_shardings=x_spec, out_shardings=(None, x_spec))
        fwd2, grad2 = jit2(x_)
        grad2.block_until_ready()
    
    print(f"  {n_devices}-GPU PASSED")


if __name__ == "__main__":
    print(f"JAX version: {jax.__version__}")
    print(f"Devices: {jax.devices()[:4]}")
    print()
    
    # Same sequence as TE reproducer:
    # 1. 2-GPU with DP axis
    run_test(2, (2,), ("dp",), "dp", None)
    
    # 2. 2-GPU with TP axis  
    run_test(2, (2,), ("tp",), None, "tp")
    
    # 3. 4-GPU with DP+TP axes -> Does it deadlock?
    run_test(4, (2,2), ("dp","tp"), "dp", "tp")
    
    print("\nAll passed!")
```

This was occurring in JAX TransformerEngine tests with pjit when running tests with different mesh configurations sequentially. The scenario is:

1. Test 1: Creates clique [0,1] with 2 GPUs
2. Test 2: Creates clique [0,1] with 2 GPUs (different mesh axis)
3. Test 3: Creates clique [0,1,2,3] with 4 GPUs, then needs [2,3]
   - Ranks 2,3 try to split [2,3] from [0,1,2,3]
   - But ranks 0,1 reuse the cached standalone [0,1] clique
   - NCCL comm split requires ALL ranks of parent clique to participate
   - Ranks 0,1 don't join the split because they reused cached clique
   - DEADLOCK: Ranks 2,3 wait forever for 0,1 to join

The fix invalidates any cached cliques that are proper subsets of a newly created clique. This ensures subsequent operations will create split communicators from the parent clique rather than reusing stale standalone communicators.

This is complementary to the participant_groups fix in PR #15935, which prevents clique key collisions. This fix handles the case where the cached clique was created before the larger clique existed.

### DIFF
--- a/xla/backends/gpu/collectives/gpu_clique_key_test.cc
+++ b/xla/backends/gpu/collectives/gpu_clique_key_test.cc
@@ -221,4 +221,42 @@ TEST(GpuCliqueIdStringTest, ToString) {
   }
 }
 
+// Test that IsSubsetOf correctly identifies subset relationships for clique
+// invalidation. This is important for preventing deadlocks when cached subset
+// cliques need to be invalidated after a larger clique is created.
+TEST(GpuCliqueKeyTest, IsSubsetOfForCliqueInvalidation) {
+  GlobalDeviceId id0 = GlobalDeviceId(0);
+  GlobalDeviceId id1 = GlobalDeviceId(1);
+  GlobalDeviceId id2 = GlobalDeviceId(2);
+  GlobalDeviceId id3 = GlobalDeviceId(3);
+
+  // Scenario: First we create clique [0,1], then later create [0,1,2,3].
+  // The cached [0,1] clique should be identified as a subset of [0,1,2,3]
+  // so it can be invalidated to prevent deadlock during comm splitting.
+  GpuCliqueKey small_clique({id0, id1}, /*num_local_participants=*/2, false);
+  GpuCliqueKey large_clique({id0, id1, id2, id3}, /*num_local_participants=*/4,
+                            false);
+
+  // [0,1] is a subset of [0,1,2,3]
+  EXPECT_TRUE(small_clique.IsSubsetOf(large_clique));
+
+  // [0,1,2,3] is not a subset of [0,1]
+  EXPECT_FALSE(large_clique.IsSubsetOf(small_clique));
+
+  // A clique is a subset of itself
+  EXPECT_TRUE(small_clique.IsSubsetOf(small_clique));
+  EXPECT_TRUE(large_clique.IsSubsetOf(large_clique));
+
+  // [2,3] is also a subset of [0,1,2,3]
+  GpuCliqueKey other_small({id2, id3}, /*num_local_participants=*/2, false);
+  EXPECT_TRUE(other_small.IsSubsetOf(large_clique));
+
+  // [0,1] is not a subset of [2,3] (disjoint)
+  EXPECT_FALSE(small_clique.IsSubsetOf(other_small));
+
+  // Different is_p2p flag should prevent subset relationship
+  GpuCliqueKey p2p_clique({id0, id1}, /*num_local_participants=*/2, true);
+  EXPECT_FALSE(p2p_clique.IsSubsetOf(large_clique));
+}
+
 }  // namespace xla::gpu
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -442,6 +442,48 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
       VLOG(3) << absl::StrFormat("[%s] [ranks=%s] Created new clique %v",
                                  DeviceOrdinalsToString(ranks),
                                  DeviceRanksToString(ranks), clique_key);
+
+      // When NCCL comm splitting is enabled, invalidate any cached cliques
+      // that are proper subsets of the newly created clique. This ensures
+      // that subsequent operations using those subset device groups will
+      // create split communicators from this parent clique rather than
+      // reusing stale standalone communicators.
+      //
+      // Without this, if clique [0,1] exists from a previous run and we
+      // create [0,1,2,3], a subsequent [2,3] operation would try to split
+      // from [0,1,2,3] but [0,1] operations would use the old standalone
+      // clique, causing rendezvous deadlocks because NCCL comm split
+      // requires all ranks of the parent clique to participate.
+      static const bool enable_nccl_comm_splitting =
+          xla::GetDebugOptionsFromFlags().xla_gpu_enable_nccl_comm_splitting();
+      if (enable_nccl_comm_splitting) {
+        std::vector<GpuCliqueKey> keys_to_remove;
+        for (auto& [cached_key, cached_clique] : state.cliques) {
+          // Skip the clique we just created and check if cached is a proper
+          // subset of the new clique (same devices would mean equal keys).
+          if (!(cached_key == clique_key) &&
+              cached_key.IsSubsetOf(clique_key)) {
+            VLOG(3) << absl::StrFormat(
+                "[%s] [ranks=%s] Invalidating cached subset clique %v "
+                "(subset of new clique %v)",
+                DeviceOrdinalsToString(ranks), DeviceRanksToString(ranks),
+                cached_key, clique_key);
+            keys_to_remove.push_back(cached_key);
+          }
+        }
+        for (const auto& key : keys_to_remove) {
+          auto it = state.cliques.find(key);
+          if (it != state.cliques.end()) {
+            // Abort the communicators before removing to ensure clean
+            // shutdown of NCCL resources.
+            if (auto status = it->second.Abort(); !status.ok()) {
+              LOG(WARNING) << "Failed to abort subset clique "
+                           << key.ToString() << ": " << status;
+            }
+            state.cliques.erase(it);
+          }
+        }
+      }
     }
 
     return emplaced.first->second.Acquire();


i-chaochen:
Hi @beckerhe wondering could you review this issue? Thanks!

beckerhe:
> Hi @beckerhe wondering could you review this issue? Thanks!

I'm not the right person to give an effective review, but I'm going to find reviewers.

olegshyshkov:
We have a similar parallel PR: https://github.com/openxla/xla/pull/38105. The other PR had some activity, but also isn't ready to be merged. We should probably focus on one to avoid duplicate work.

@ezhulenev what do you think?

ezhulenev (approved):

ezhulenev (approved):

ezhulenev:
@olegshyshkov I think this is the right approach, a little bit too aggressive, but at the same time it also guarantees that NCCL resources are shared efficiently across all cliques

xla/backends/gpu/collectives/gpu_clique_key_test.cc
```diff
@@ -221,4 +221,42 @@ TEST(GpuCliqueIdStringTest, ToString) {
   }
 }
 
+// Test that IsSubsetOf correctly identifies subset relationships for clique
+// invalidation. This is important for preventing deadlocks when cached subset
+// cliques need to be invalidated after a larger clique is created.
+TEST(GpuCliqueKeyTest, IsSubsetOfForCliqueInvalidation) {
```
ezhulenev, line 227: There is also https://github.com/openxla/xla/blob/main/xla/backends/gpu/collectives/gpu_cliques_test.cc, I think you should be able to add a mini reproducer there that will catch this problem.

ezhulenev (approved):

hawkinsp:
I manually verified that this PR fixes the deadlock in the JAX test suite on 8xH100 I reported here: https://github.com/openxla/xla/pull/36981#issuecomment-4059168227

If we could finish this up and merge this, it would be great!

i-chaochen:
> I manually verified that this PR fixes the deadlock in the JAX test suite on 8xH100 I reported here: [#36981 (comment)](https://github.com/openxla/xla/pull/36981#issuecomment-4059168227)
> 
> If we could finish this up and merge this, it would be great!

Thanks for the verification! @hawkinsp I guess now we can merge this PR? seems this PR still requires 1 approving review with write access...and this PR even doesn't trigger the CI jobs.

We have found out this `ncclCommSplit` deadlock issue last year with the hotfix to disable this feature. Then, we had this fix upstream PR on Jan 2026.......Yes, if PR review process is better we could  avoid this issue on upstream earlier.

i-chaochen (approved):

hawkinsp (approved):

hawkinsp:
It looks like there's a build failure:
```
xla/backends/gpu/collectives/gpu_cliques.cc:506:41: error: no member named 'Abort' in 'std::shared_ptr<xla::gpu::LockableGpuClique>'; did you mean to use '->' instead of '.'?
  506 |             if (auto status = it->second.Abort(); !status.ok()) {
```

Can you rebase on head and fix? (Just needs to be `->`)

hawkinsp:
Actually, I'll just do it.

hawkinsp:
Even after this PR, I'm seeing NCCL deadlocks running this pair of JAX tests on 8xH100:

```
python tests/pjit_test.py ArrayPjitTest.test_wsc_vmap_abstract_mesh ShardingInTypesTest.test_conv_general_dilated
```

As far as I can tell things are hanging on the new `Abort()` call.

ezhulenev:
I'll look into this today

hawkinsp:
AI gave me this test case, for however much we can trust AI:

```
TEST(GpuCliquesTest, SplitCliquesNoDeadlock2) {
  auto cleanup = absl::MakeCleanup([] { internal::DestroyAcquiredCliques(); });

  ASSERT_OK_AND_ASSIGN(se::Platform * platform,
                       se::PlatformManager::PlatformWithName("CUDA"));

  if (platform->VisibleDeviceCount() < 8) {
    GTEST_SKIP() << "Test requires at least 8 GPUs";
  }

  tsl::thread::ThreadPool pool(tsl::Env::Default(), "collectives", 8);
  tsl::Executor& exec = *pool.AsExecutor();

  RunId run_id(0);

  GpuCliqueKey key0123({kD0, kD1, kD2, kD3}, 4);
  GpuCliqueKey key01({kD0, kD1}, 2);
  GpuCliqueKey key23({kD2, kD3}, 2);
  GpuCliqueKey key01234567({kD0, kD1, kD2, kD3, kD4, kD5, kD6, kD7}, 8);

  std::vector<std::vector<GlobalDeviceId>> group0123 = {{kD0, kD1, kD2, kD3}};
  std::vector<std::vector<GlobalDeviceId>> group01_23 = {{kD0, kD1},
                                                         {kD2, kD3}};
  std::vector<std::vector<GlobalDeviceId>> group01234567 = {
      {kD0, kD1, kD2, kD3, kD4, kD5, kD6, kD7}};

  ASSERT_OK_AND_ASSIGN(std::vector<se::StreamExecutor*> executors_vec,
                       CreateExecutors(platform, 8));
  absl::Span<se::StreamExecutor*> executors(executors_vec);

  std::vector<AcquiredCliquesMap> acquired_cliques_vec(8);
  absl::Span<AcquiredCliquesMap> acquired_cliques(acquired_cliques_vec);

  {  // Acquire clique for devices [0,1,2,3].
    auto futures = AcquireCliques(exec, executors.first(4), key0123, group0123,
                                  acquired_cliques.first(4));
    ASSERT_OK_AND_ASSIGN(auto cliques, WaitCliques(std::move(futures)));

    for (size_t i = 0; i < 4; ++i) {
      acquired_cliques.at(i).emplace(key0123, cliques.at(i));
    }
  }

  {  // Split [0,1] and [2,3] from [0,1,2,3].
    auto futures0 = AcquireCliques(exec, executors.first(2), key01, group01_23,
                                   acquired_cliques.first(2));
    auto futures1 = AcquireCliques(exec, executors.subspan(2, 2), key23,
                                   group01_23, acquired_cliques.subspan(2, 2));
    ASSERT_OK_AND_ASSIGN(auto cliques0, WaitCliques(std::move(futures0)));
    ASSERT_OK_AND_ASSIGN(auto cliques1, WaitCliques(std::move(futures1)));
  }

  for (auto& m : acquired_cliques_vec) {
    m.clear();
  }

  {  // Acquire clique for all 8 devices. This should invalidate the cached
    // [0,1], [0,1,2,3], and [2,3] cliques without deadlocking.
    auto futures = AcquireCliques(exec, executors, key01234567, group01234567,
                                  acquired_cliques);
    ASSERT_OK_AND_ASSIGN(auto cliques, WaitCliques(std::move(futures)));
  }
}
```

hawkinsp:
Also here's the failure in JAX CI (pjit_test_gpu):

https://github.com/jax-ml/jax/actions/runs/23149029500/job/67245195107

I derived the two test cases above by minimizing the set that ran in the failing case linked there.

hawkinsp:
@ezhulenev any progress? We're trying to release JAX approximately now and I'm wondering if this state is safe to release.

olupton:
We are seeing rendezvous timeouts in both JAX and TransformerEngine tests that I suspect are related.

ezhulenev:
AMD folks are taking care of it: https://github.com/openxla/xla/pull/39268, @hawkinsp does it fix your deadlock?

ezhulenev:
I suspect that maybe it was a wrong change :) debugging the original deadlock because of a split devision divergence, hopefully will get a better fix today

ezhulenev:
https://github.com/openxla/xla/pull/39372 this should fix the comm split

PR #37074, opened 2026-01-30
[ROCm] Reenable miopen autotune when xla_gpu_autotune_level == 0

draganmladjenovic:
📝 Summary of Changes
Adapt GpuCompiler::AddConvAndGemmAutotuningPass to match pre refactor behavior of AMDGPUCompiler. 

🎯 Justification
For ROCm we need to run miopen backend even when autotuning is disabled in order to decompose back unsupported fused convolutions. There is no runtime fallback.

🚀 Kind of Contribution
🐛 Bug Fix

N\A

🧪 Unit Tests:
None

🧪 Execution Tests:
None

### DIFF
--- a/xla/backends/gpu/autotuner/miopen.h
+++ b/xla/backends/gpu/autotuner/miopen.h
@@ -39,7 +39,8 @@ class MIOpenBackend : public GpuCodegenBackend {
                          const DebugOptions* debug_options, Compiler* compiler,
                          const Compiler::GpuTargetConfig* target_config)
       : GpuCodegenBackend(autotuner::Backend::MIOPEN, debug_options, compiler,
-                          target_config, stream_executor) {}
+                          target_config, stream_executor),
+        do_not_autotune_(debug_options->xla_gpu_autotune_level() == 0) {}
 
   absl::StatusOr<std::vector<std::unique_ptr<BackendConfig>>>
   GetSupportedConfigs(const HloInstruction& instr) override;
@@ -52,6 +53,7 @@ class MIOpenBackend : public GpuCodegenBackend {
 
  private:
   bool IsSupported(const HloInstruction& instr) override;
+  bool do_not_autotune_;
 };
 
 }  // namespace gpu
--- a/xla/service/gpu/gpu_compiler.cc
+++ b/xla/service/gpu/gpu_compiler.cc
@@ -3267,16 +3267,26 @@ absl::Status GpuCompiler::AddConvAndGemmAutotuningPass(
                       GetCodegenBackends(stream_exec, target_config, alias_info,
                                          debug_options, mlir_context));
 
-  bool do_not_autotune_cublas_and_cudnn =
+  bool do_not_autotune_cublas =
       debug_options.xla_gpu_experimental_disable_binary_libraries() ||
       debug_options.xla_gpu_autotune_level() == 0 ||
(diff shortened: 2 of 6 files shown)


draganmladjenovic:
@derdrdirk Gentle ping.

derdrdirk:
Is this actually needed? When xla_gpu_autotune_level == 0 we select the first config that compiles. Does the compilation of the MIOpen kernel succeed, but then is not executable? In this case is it possible to to let the compilation fail?

draganmladjenovic:
When xla_gpu_autotune_level == 0 autotuner does not run at all.

draganmladjenovic:
@derdrdirk Ping as gentle as a soft breeze.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -3267,16 +3267,24 @@ absl::Status GpuCompiler::AddConvAndGemmAutotuningPass(
...
-  bool do_not_autotune_cublas_and_cudnn =
+  bool do_not_autotune_cublas =
       debug_options.xla_gpu_experimental_disable_binary_libraries() ||
       debug_options.xla_gpu_autotune_level() == 0 ||
       debug_options.xla_gpu_exclude_nondeterministic_ops();
-  auto should_autotune = [do_not_autotune_cublas_and_cudnn](
+  // We need to run miopen autotune to decompose unsuported fused convolutions
+  bool do_not_autotune_cudnn =
+      debug_options.xla_gpu_experimental_disable_binary_libraries() ||
+      (do_not_autotune_cublas && !gpu_version.IsRocm());
```
derdrdirk, line 3279: Can you add a TODO here that we can merge do_not_autotune_cudnn into do_not_autotune_cublas once we use HloFusions for convolutions that can be decomposed using the FissionBackend.

derdrdirk:
I understand the issue now. In general LGTM as I do not have a better solution for now. In general the fission / decomposition of the convolution should not happen in the apply config. We have a FissionBackend to decompose/ fission HloFusions. This would apply to fused convolutions as well, but unfortunately convolutions are not HloFusions, but custom calls. Note that we are currently refactoring to express fused convolutions as HloFusions instead of CustomCalls so I asked you to add a TODO to improve the code once this has landed.

derdrdirk (changes requested):

derdrdirk (approved):

xla/backends/gpu/autotuner/miopen.h
```diff
@@ -52,6 +53,7 @@ class MIOpenBackend : public GpuCodegenBackend {
 
  private:
   bool IsSupported(const HloInstruction& instr) override;
+  bool do_not_autotune_;
```
shyamliagrawal, line 56: @draganmladjenovic IIUC, miopen.h/cc changes should not be required. The Autotuner will automatically use the Default config when autotune_level=0. Can you please try removing this change and see if it works for you?

draganmladjenovic: It will pick up the first one. Default is used for deviceless compilation.

shyamliagrawal: You are right, I am planning to unify the default and first options, we can clean this up then.

shyamliagrawal (approved):

PR #37081, opened 2026-01-30
Fix/generalize AR+Slice matching

jfc4050:
📝 Summary of Changes
Reduce scatter creator was missing some replacement opportunities. In my case there were (provably) no-op clamps that were making it think the multiplier and offset index were different.

🎯 Justification
In pretty standard 3D parallel workload, missing AR->RS replacements causes 50% in TP comms and 50% increase in DP comms.

🚀 Kind of Contribution

🧪 Unit Tests:
Added unit tests for expected matching patterns, also a minimized version of the matching that was previously failing for me.

🧪 Execution Tests:

### DIFF
--- a/xla/service/collective_opt_utils.cc
+++ b/xla/service/collective_opt_utils.cc
@@ -15,11 +15,11 @@ limitations under the License.
 
 #include "xla/service/collective_opt_utils.h"
 
+#include <algorithm>
 #include <cstdint>
-#include <functional>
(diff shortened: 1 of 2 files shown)


pifon2a (approved):
Let me get back to you when the benchmarks are complete.

pifon2a (approved):
The benchmarks did not regress. Also it is a nice clean-up. Thank you!

PR #37110, opened 2026-01-31
[CPU] Fix Use-After-Free in JitCompiler caused by zombie tasks in ThreadPool

jimmyorourke:
📝 Summary of Changes
This PR fixes a use-after-free race condition in `xla::cpu::JitCompiler` where tasks lingering in the thread pool extend the lifetime of JIT resources beyond the lifetime of the compiler itself.

The issue stems from how the `Eigen::ThreadPool` used by `JitCompiler` (via `TaskRunner`) manages task lifetimes. Worker threads copy tasks to their local stack for execution, but the original std::function often remains alive in the thread pool's circular buffer until it is eventually overwritten. 
If the JitCompiler (and its underlying `llvm::orc::ExecutionSession`) is destroyed while these zombie tasks are still in the queue, the `llvm::orc::Task` objects inside them stay alive, causing two distinct crash signatures depending on the task type:
- Use-After-Free (Segfault): If the zombie task is destroyed after the `ExecutionSession` is freed, its destructor may try to access the dead session (via `~MaterializationResponsibility`), causing a crash in `recursive_mutex`.
- Assertion Failure (`SymbolStringPool`): If the zombie task holds a `SymbolStringPtr` (eg a `MaterializationTask`), it keeps the `SymbolStringPool` reference count above zero. When the `ExecutionSession` tries to destroy the pool, it hits an assertion: `Assertion Pool.empty() && "Dangling references at pool destruction time"` failed.

This change introduces a `TaskHolder` indirection  to decouple the lifetime of the `llvm::orc::Task` from the lifetime of the thread pool queue entry. We wrap the `std::unique_ptr<Task>` in a `std::shared_ptr<TaskHolder>`.
Both the thread pool queue and the worker thread hold references to this shared holder.
Immediately after `task->run()` completes, we explicitly call `task.reset()` inside the wrapper.
This ensures that the Task resources are released synchronously at the end of execution. Even if the wrapper struct remains alive in the thread pool queue, it no longer keeps a Task alive with it, and can be safely destroyed at any time thereafter.

🎯 Justification
Fixes recurring crashes on resource-constrained environments (like ChromeOS/CrosVM) where the ThreadPool queue wraps around frequently when there are multiple calls to `PjRtClient.CompileAndLoad`.
Fixes occasional assertion failure crashes on destruction.

🚀 Kind of Contribution
🐛 Bug Fix

N/A

🧪 Unit Tests:
New CPU JitCompiler unit test added which mimics the behavior of the Eigen thread pool by holding a copy of a task in a vector while executing a local copy. The test asserts that the task's resources are freed immediately after execution, even while the vector retains a reference to the task wrapper.
Without the fix, this test fails (resources are retained by the queue). With the fix, it passes.

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/backends/cpu/codegen/jit_compiler.cc
+++ b/xla/backends/cpu/codegen/jit_compiler.cc
@@ -237,7 +237,7 @@ void JitCompiler::TaskDispatcher::dispatch(
 
     if (task_holder->task) {
       // We run and explicitly destroy the task before decrementing the
-      // counterand notifying the condition variable, to ensure that the
+      // counter and notifying the condition variable to ensure that the
       // task is fully executed and cleaned up before task dispatcher shut
       // down.
       task_holder->task->run();
--- a/xla/backends/cpu/codegen/jit_compiler.cc
+++ b/xla/backends/cpu/codegen/jit_compiler.cc
@@ -224,15 +224,30 @@ void JitCompiler::TaskDispatcher::dispatch(
     ++num_dispatched_tasks_;
   }
 
-  task_runner_([this, task = std::shared_ptr<llvm::orc::Task>(
+  // Wrap the move-only task in a shared struct. This satisfies the thread
+  // pool's requirement for copyable tasks while enabling explicit control of
+  // the task's lifetime.
+  struct TaskHolder {
+    std::unique_ptr<llvm::orc::Task> task;
+  };
+
+  task_runner_([this, task_holder = std::make_shared<TaskHolder>(
                           std::move(task))]() mutable {
     TraceMe trace("TaskDispatcher::dispatch");
 
-    // We run and explicitly destroy the task before decrementing the counter
-    // and notifying the condition variable, to ensure that the task is fully
-    // executed and cleaned up before task dispatcher shut down.
-    task->run();
-    task.reset();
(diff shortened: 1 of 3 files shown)


xla/backends/cpu/codegen/jit_compiler.cc
```diff
@@ -224,15 +224,30 @@ void JitCompiler::TaskDispatcher::dispatch(
...
     TraceMe trace("TaskDispatcher::dispatch");
 
-    // We run and explicitly destroy the task before decrementing the counter
-    // and notifying the condition variable, to ensure that the task is fully
-    // executed and cleaned up before task dispatcher shut down.
-    task->run();
-    task.reset();
+    if (task_holder->task) {
+      // We run and explicitly destroy the task before decrementing the
+      // counterand notifying the condition variable, to ensure that the
```
penpornk: Nit:
```suggestion
      // counter and notifying the condition variable to ensure that the
```

penpornk (changes requested):
Sorry for the delay and thank you for the fix!

jimmyorourke:
Superseded by https://github.com/openxla/xla/pull/37984

PR #37129, opened 2026-02-01
[Trivial] Fixed a bug in the handling of recv callbacks in PJRT.

eaplatanios:
📝 Summary of Changes

Switched from `options.send_callbacks.size()` to `options.recv_callbacks.size()` in a couple places that dealt with receive callbacks instead of send callbacks. I'm guessing this was probably a copy-pasta bug.

🎯 Justification

I discovered this because I've been developing Rust bindings for PJRT and I kept getting segfaults while trying to use a send callback with a program that had a send operation and no receive operations. I verified that this fix resolved the segfault I was hitting.

🚀 Kind of Contribution

🐛 Bug Fix


allanrenucci (approved):
Thanks!

PR #37157, opened 2026-02-02
[ROCm] Enable the xla-opt binary build

nurmukhametov:
- Add visibility patch for Triton test libraries needed by AllPassesAndDialects
- Fix RegisterTritonDialects.h include path


chsigg (approved):
Great, thanks for enabling this.

PR #37287, opened 2026-02-04
[ROCm] Skip swish fusion if pre-activation output used

mfrancepillois:
Hipblaslt does not support SILU_AUX yet. So, we do not fuse the activation if the pre-activation output is used.
This avoid duplicating the gemm operation (1 with activation + 1 without activation).

### Motivation

The swish activation makes "only" 2 references to the gemm ouput (Swish(x)=x∗sigmoid(βx)).
Consequently, the `aux` condition should be:
```
bool has_aux = gemm->user_count() > 2;
```
But that is not sufficient. Because in the Llama training workload for example, the gemm output does not have more that 2 references, because the output is first "bitcasted" before being used to calculate the swish function.
So, without gradient, we have something like:
```
dot.67 (gemm output)
  ↓
bitcast.85 (reshape to higher rank)
  ↓ (used in 2 places)
  ├─ neg.16 → exp.10 → add.326 → div.147 (sigmoid computation)
  └─ mul.617 (multiply with sigmoid result)
dot.67.user_count() = 1  (only bitcast.85)
bitcast.85.user_count() = 2  (neg.16 and mul.617)
```
Consequently, we must search for final users and not direct users. That is why the `CountFinalUsers` function has been introduced.

## Test Plan

UT added in this PR

## Test Result

Test pass.


dimitar-asenov (approved):

mfrancepillois:
@dimitar-asenov Thanks for reviewing this PR. I had to manually rebase  it on top of main due to conflicts. Could you please take another look?

dimitar-asenov (approved):

xla/backends/gpu/transforms/gemm_rewriter.cc
```diff
@@ -542,6 +542,24 @@ auto OptionalBitcast(HloInstruction** optional_bitcast, Pattern pattern) {
                                   std::move(pattern));
 }
 
+// Counts the number of final users of an instruction, recursively traversing
+// through slice and bitcast operations. If a user is a slice or bitcast, we
+// recursively count its users. Otherwise, we count it as a final user.
+static inline uint32_t CountFinalUsers(const HloInstruction* instr) {
```
allanrenucci: Please remove static an inline. static is redundant in the anonymous namespace and it does not make sense to inline a recursive function (the compiler will most likely ignore it).

allanrenucci (changes requested):

xla/backends/gpu/transforms/gemm_rewriter.cc
mfrancepillois: I completely agree. Thank you for noticing. The code has been changed.

allanrenucci (approved):
Thanks

PR #37289, opened 2026-02-04
[ROCm] Fix FloatNormalization to use gpu_compute_capability

nurmukhametov:
Use gpu_compute_capability() instead of cuda_compute_capability() when constructing GpuFloatSupport for FloatNormalization in the autotuner. This could cause incorrect type support detection or wrong type conversions.


nurmukhametov:
@xla-rotation, could I get a review on this?

allanrenucci (approved):

PR #37511, opened 2026-02-09
[ROCm] Explicitly skip failing test cases 

Eetusjo:
Adds GTEST_SKIPs to unit tests failing on ROCm, making the skips explicit, instead of relying on e.g. excluding test targets when running the full test suite.

🚀 Kind of Contribution
Please remove what does not apply: ♻️ Cleanup, 🧪 Tests

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_conversion_pass_test.cc
+++ b/xla/backends/gpu/runtime/command_buffer_conversion_pass_test.cc
@@ -625,6 +625,9 @@ TEST(CommandBufferConversionPassTest, DontConvertIfNotMinGraphSize) {
 }
 
 TEST(CommandBufferConversionPassTest, ConvertWhileThunk) {
+  if (GetPlatformName() == "ROCM") {
+    GTEST_SKIP() << "Not supported on ROCm";
+  }
   CommandBufferConversionPass pass{"test"};
 
   std::vector<std::unique_ptr<Thunk>> thunks;
@@ -736,6 +739,9 @@ TEST(CommandBufferConversionPassTest,
 }
 
 TEST(CommandBufferConversionPassTest, ConvertWhileThunkWithAsyncPair) {
+  if (GetPlatformName() == "ROCM") {
+    GTEST_SKIP() << "Not supported on ROCm";
+  }
   CommandBufferConversionPass pass{"test"};
 
   std::vector<std::unique_ptr<Thunk>> thunks;
--- a/xla/backends/gpu/transforms/cudnn_fused_conv_rewriter_test.cc
+++ b/xla/backends/gpu/transforms/cudnn_fused_conv_rewriter_test.cc
@@ -127,6 +127,13 @@ class CudnnFusedConvRewriterTest : public GpuCodegenTest {
         .gpu_compute_capability()
         .IsCuda();
   }
+  bool IsRocm() const {
+    return backend()
+        .default_stream_executor()
+        ->GetDeviceDescription()
+        .gpu_compute_capability()
+        .IsRocm();
+  }
   se::CudaComputeCapability GetCudaComputeCapability() const {
     return backend()
         .default_stream_executor()
@@ -318,6 +325,7 @@ class CudnnFusedConvRewriterTest : public GpuCodegenTest {
   } while (0)
 
 TEST_F(CudnnFusedConvRewriterTest, TestConvOnly) {
+  if (IsRocm()) GTEST_SKIP() << "Skipped on ROCm";
   // max(0, conv(x, w));
   TestMatchWithAllTypes(R"(
     HloModule Test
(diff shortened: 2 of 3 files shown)


Eetusjo:
@xla-rotation could I get a review for this PR, please

dimitar-asenov:
@olegshyshkov Can you please weigh in on this PR? While the intention is reasonable, I'm not sure if we should have so many `GTEST_SKIP()`. Particularly in `triton/support_test.cc`, we don't normally skip based on architecture or version.

Eetusjo:
Thanks @dimitar-asenov. Probably could have explicated this in the PR description as well: The intention of this PR isn't to say "Not supported on ROCm" for each of the features under test that are now skipped. Hoewver, this helps us get a clean baseline on ROCm when running the whole test suite. The intention is to start fixing all of the tests that are not clearly platform-specific (if there are such). I can also add TODO comments on the skips introduced in this PR to make it clear we're coming back to these.

Regarding triton/support_test specifically, https://github.com/openxla/xla/pull/35363 fixes that concern but it has been pending for quite a while.

cc @i-chaochen

dimitar-asenov:
With respect to Triton support tests, please see my comment at https://github.com/openxla/xla/pull/35363#pullrequestreview-3626532807

TLDR: we should not skip these tests. If something is not supported on RoCM then the test should execute and test that it is not supported.

i-chaochen:
@dimitar-asenov thanks for the feedback. We don't want to skip any tests and is going to fix all. As you see some of them are pending PRs to be reviewed and merged. [`GTEST_SKIP` is used many places in XLA code base](https://github.com/search?q=repo%3Aopenxla%2Fxla+GTEST_SKIP+path%3A%2F%5Exla%5C%2Ftests%5C%2F%2F&type=code&p=1) so we use it as temp solution. We used to add `no_rocm` tag on bazel build target but seems `GTEST_SKIP` is fine grained.

If you have any other better suggestion please let us know.

Would be better to add TODO comments for ROCm team on these skipped ones to make it clearer?

Eetusjo:
> TLDR: we should not skip these tests. If something is not supported on RoCM then the test should execute and test that it is not supported.

@dimitar-asenov  just to check, do I understand correctly that this PR won't get merged with the skips, or are we still waiting for input from @olegshyshkov?

dimitar-asenov:
@olegshyshkov Should weigh in on this PR.

For many of the files you modify, `GTEST_SKIP` is OK. That's because those tests are focused on what works and things don't always work on RoCM.

However, for the `support_test.cc` in particular, I think we should not skip any tests. These tests do not simply check that a specific functionality works (positive case), but they also test the negative case. In that sense you can already write these tests correctly by ensuring that for RoCM they verify that things are not (yet) supported. If and when a particular feature becomes supported, then the test should change to reflect that.

Skipping support tests for unsupported features means that we are not testing whether the `IsSupported...` functions correctly return `false` for the features in question. Ideally, we should avoid this reduced coverage.

Hope this clarifies things.

xla/backends/gpu/codegen/triton/fusion_emitter_int4_device_test.cc
```diff
@@ -125,6 +125,9 @@ ENTRY entry_computation {
 }
 
 TEST_F(TritonTest, FuseSubchannelDequantizationWithTranspose) {
+  if (device_desc().gpu_compute_capability().IsRocm()) {
+    GTEST_SKIP() << "Skipped on ROCm";
```
olegshyshkov: Why is this test skipped? Does it not fuse on ROCm? Or we need a different FileCheck?

xla/backends/gpu/codegen/triton/fusion_emitter_parametrized_test.cc
```diff
@@ -271,6 +271,10 @@ TEST_P(UnaryElementwiseTest, ElementwiseUnaryOpExecutesCorrectly) {
   float tolerance;
   std::tie(data_type, opcode, tolerance) = GetParam();
 
+  if (is_built_with_rocm_ && data_type == F16 && opcode == HloOpcode::kCos) {
+    GTEST_SKIP() << "Skipped on ROCm";
```
olegshyshkov: Could you please add a comment why the test is skipped? Here and below in this file.

olegshyshkov (commented):
I agree with @dimitar-asenov, most of the skips are reasonable. There are a few places where I would like to see a more detailed comment why the test is skipped to understand if it's a fundamental limitation, a feature you're working on and will be enabled later or the test just works differently on ROCm and we should adapt the test rather than disabling it.

As for `support_test.cc` skips, can we just land https://github.com/openxla/xla/pull/35363? Maybe I missed something in the discussion, but looks like the last comment from @dimitar-asenov suggests that PR is unblocked.

Eetusjo:
@olegshyshkov I ended up removing most of the skips: Looks like the PR that fixes `support_test` is now getting merged (https://github.com/openxla/xla/pull/38521) and we should have fixes in the pipeline for the other triton-related issues as well. Many of the other ones have now been fixed in other PRs that are either merged or in review:

- https://github.com/openxla/xla/pull/38258
- https://github.com/openxla/xla/pull/38208
- https://github.com/openxla/xla/pull/38028
- https://github.com/openxla/xla/pull/37981
- https://github.com/openxla/xla/pull/38028

etc.

olegshyshkov (approved):

PR #37601, opened 2026-02-10
[xla:ffi] Extract separate library for handler registration

ezhulenev:
Extract separate library for handler registration in preparation with fixing global static hell.

`xla::ffi::Invoke` now takes explicit `XLA_FFI_APi*` parameter because we want users to pass FFI API to all XLA:FFI methods and do not rely on linking order to get correct pointer.

This is one more step in the direction of fixing duplicate FFI registries in XLA and JAX.


bchetioui (approved):
Thank you Eugene!

After this, `ffi_api` must be linked a single time per plugin to make this safe.

ezhulenev:
Yes, `ffi_api` visibility should be restricted to XLA, but it requires cleaning up JAX first, and maybe few other places in g3.

It will be linked into PJRT plugin via transitive deps from one of the XLA targets.

bchetioui (approved):

bchetioui (approved):

mwhittaker (approved):
✉️

cota (approved):

cota (approved):

cota (approved):

bchetioui (approved):

PR #37629, opened 2026-02-11
[ROCm] Propagate register spill info for autotuner filtering

nurmukhametov:
AMD's GPU backend already detected register spilling from HSACO metadata, but this information wasn't being passed through ModuleStats to the autotuner. This caused the autotuner's
xla_gpu_filter_kernels_spilling_registers_on_autotuning flag to have no effect on ROCm, as the spill check in autotuner_compile_util.cc always saw empty ModuleStats.

This change:
- Refactors RegisterSpillInfo from aggregate-max into per-kernel KernelSpillInfo, so that kernel names are preserved for ModuleStats
- Adds HsacoResult and HsacoFileResult structs to return both HSACO bytes/path and ModuleStats from CompileToHsaco and CompileToHsacoAndReturnFilePath
- Converts per-kernel spill info to ModuleStats format by mapping private_segment_size to store_bytes_spilled and load_bytes_spilled
- Updates AMDGPUCompiler::CompileTargetBinary to pass module_stats through BackendCompileResult
- Ensures the HSACO cache-hit path also returns correct ModuleStats by re-parsing spill info from the cached binary

This aligns ROCm behavior with NVIDIA, where kernels that spill registers are correctly filtered out during autotuning.

Fixes TritonFusionNumericsVerifierTest.
CompilationSucceedsEvenIfKernelWillSpillRegisters on AMD GPUs.


xla/service/gpu/llvm_gpu_backend/amdgpu_backend.cc
```diff
@@ -172,6 +176,60 @@ struct RegisterSpillInfo {
...
+
+  // Convert to ModuleStats format for autotuner filtering.
+  //
+  // Only kernels with private_segment_size > 0 (i.e. spills that use stack
+  // scratch memory) are included. SGPR-only spills that are saved to VGPRs
+  // don't hit memory, so they are not reported here. This means the
+  // autotuner's ilter
+  // (xla_gpu_filter_kernels_spilling_registers_on_autotuning) will not discard
+  // them. Such spills are still caught by the hard-fail path
+  // (xla_gpu_fail_ptx_compilation_on_register_spilling) via HasStackUsage() in
```
draganmladjenovic: Are they? I thought we only do hard fail on HasStackUsage?

xla/service/gpu/llvm_gpu_backend/amdgpu_backend.cc
```diff
@@ -927,26 +990,27 @@ absl::StatusOr<std::vector<uint8_t>> CompileToHsaco(
...
+    remove(file_result.hsaco_path.c_str());
   }
   HsacoCache::Add(str, hash, gcn_arch_name, hsaco);
 
-  return hsaco;
+  return HsacoResult{std::move(hsaco), std::move(file_result.module_stats)};
 }
 
-absl::StatusOr<std::string> CompileToHsacoAndReturnFilePath(
+absl::StatusOr<HsacoFileResult> CompileToHsacoAndReturnFilePath(
```
draganmladjenovic, line 1015: To be honest have no idea why we have CompileToHsacoAndReturnFilePath. The fact that we use tmp files is implementation detail and should not have been exposed (we probably leak a file on disk too). ⁠

nurmukhametov: It was introduced recently in https://github.com/openxla/xla/pull/31284.

xla/service/gpu/llvm_gpu_backend/amdgpu_backend.cc
nurmukhametov: Right, I wanted to mention the dynamic stack usage difference, but I lost track of that point after a few edits. Should be fixed now.

xla/service/gpu/llvm_gpu_backend/amdgpu_backend.cc
```diff
@@ -29,6 +30,7 @@ limitations under the License.
 #include <variant>
 #include <vector>
 
+#include "absl/algorithm/container.h"
```
akuegel, line 33: This requires another dependency.

nurmukhametov: I have added `@com_google_absl//absl/algorithm:container` dependency

derdrdirk (approved):
LGTM

Nit: autotuner_compile_util.cc and gemm_fusion_autotuner.cc are not used anymore. The register check happens in xla/backends/autotuner/autotuner.cc.

```
absl::Status Autotuner::IsValidExecutable(
    const absl::StatusOr<std::unique_ptr<Executable>>& executable) const {
  if (!executable.ok()) {
    return absl::Status(
        executable.status().code(),
        absl::StrCat("Compilation failed: ", executable.status().message()));
  }

  if (!autotune_config_.allow_reg_spills && executable.value()) {
    const auto spills_registers = [](const auto& pair) {
      const KernelStats& kernel_stats = pair.second;
      return kernel_stats.store_bytes_spilled > 0 ||
             kernel_stats.load_bytes_spilled > 0;
    };
    ModuleStats module_stats = executable.value()->module_stats();
    if (absl::c_any_of(module_stats, spills_registers)) {
      return absl::ResourceExhaustedError(
          "Discarding compilation due to register spilling.");
    }
  }
  return absl::OkStatus();
}
```

PR #37670, opened 2026-02-11
[xla:ffi] Expose static handler and type registries via internal XLA FFI API

ezhulenev:
Update "internal" ffi library `xla/ffi/ffi.h` to use type registry from the XLA FFI API instance for registering type ids. After this change it's "safe" to use "internal" FFI for building custom call libraries that will be later linked dynamically into JAX/XLA. It is still unsafe to do that with different versions of XLA as pointers to C+ structs will cross the ABI boundary, but as long as custom calls built from same XLA commit and same toolchain it will be safe, type and handler registrations will use the same global static registries in the same object file that defines XLA FFI API.


bchetioui (approved):
Thank you!

bchetioui (approved):

PR #37732, opened 2026-02-12
[ROCm] Fix tools/xla_*_compile_lib_tests

Eetusjo:
📝 Summary of Changes
Add platform detection and platform-specific device configs to the following test targets to fix the tests on ROCm: 
`//xla/tools:xla_deviceless_compile_lib_test`
`//xla/tools:xla_gpu_compile_lib_test`

🎯 Justification
Expands ROCm unit test coverage

🚀 Kind of Contribution


Eetusjo:
@xla-rotation could I get a review for this, please

dimitar-asenov (approved):

xla/tools/xla_deviceless_compile_lib_test.cc
```diff
@@ -45,9 +46,13 @@ using ::testing::IsEmpty;
 using ::testing::Not;
 
 absl::StatusOr<Compiler::GpuTargetConfig> GetGpuTargetConfig() {
+  const std::string spec_file =
+      stream_executor::PlatformManager::PlatformWithName("ROCM").ok()
```
allanrenucci: Seems wrong to add a dependency on `stream_executor::PlatformManager` for a test that is is meant to exercise deviceless compilation.

allanrenucci (changes requested):

xla/tools/xla_deviceless_compile_lib_test.cc
akuegel: It is not only the dependency, but also the fact that we would actually initialize the GPU platform with this call. You could instead use CanonicalPlatformName("gpu") == "rocm" check which would still require a dependency on the stream_executor platform, but at least it would not initialize the platform.
Theoretically this helper method could also be moved out of platform_util, then we could use it without depending on platform at all.

Eetusjo: Thanks for the comments. Changed to using PlatformUtil::CanonicalPlatformName("gpu")

allanrenucci (approved):

PR #37769, opened 2026-02-12
[XLA:GPU] Support sub-byte types in DynamicMemcpyFusion

sfvaroglu:
📝 Summary of Changes
DynamicMemcpyFusion now supports sub-byte types when every dimension has a byte-aligned stride. A fallback computes strides when ShapeUtil::ByteStrides returns null, and the fusion is accepted only when each dimension’s stride is byte-aligned.

🎯 Justification
ShapeUtil::ByteStrides returns null for sub-byte types, so DynamicMemcpyFusion previously rejected all of them. Many sub-byte layouts still have byte-aligned strides per dimension and are valid for memcpy. This change enables the memcpy path for those cases and still rejects any dimension with a non–byte-aligned stride.

🚀 Kind of Contribution
✨ New Feature, 🧪 Tests

N/A

🧪 Unit Tests:
Added. 

🧪 Execution Tests:
Tested with llama3-8b.

### DIFF
--- a/xla/backends/gpu/codegen/copy_test.cc
+++ b/xla/backends/gpu/codegen/copy_test.cc
@@ -382,22 +382,77 @@ TEST_F(CopyFusionTest, BuildDescriptorWithDynamicVariable) {
   EXPECT_EQ(offset.byte_stride, 8 * 8 * sizeof(float));
 }
 
-TEST_F(CopyFusionTest, PackedSubByteTypesAreNotSupported) {
-  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(R"(
+TEST_F(CopyFusionTest, PackedSubByteSliceNonByteAlignedStride) {
+  auto module = ParseAndReturnVerifiedModule(R"(
     dynamic_slice {
-      a = s4[20]{0:E(4)} parameter(0)
-      c = s32[] constant(10)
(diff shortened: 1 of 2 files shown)


xla/backends/gpu/codegen/copy_test.cc
```diff
@@ -382,24 +382,79 @@ TEST_F(CopyFusionTest, BuildDescriptorWithDynamicVariable) {
...
+TEST_F(CopyFusionTest, PackedSubByteSliceNonByteAlignedStride) {
+  auto module = ParseAndReturnVerifiedModule(R"(
     dynamic_slice {
-      a = s4[20]{0:E(4)} parameter(0)
-      c = s32[] constant(10)
-      s = s4[10] dynamic-slice(a, c), dynamic_slice_sizes={10}
+      p0 = s4[32,4,2,8]{3,2,1,0:E(4)} parameter(0)
+      c0 = s32[] constant(0)
+      c4 = s32[] constant(4)
+      ROOT slice = s4[32,4,2,4]{3,2,1,0:E(4)} dynamic-slice(p0, c0, c0, c0, c4), dynamic_slice_sizes={32,4,2,4}
```
akuegel, line 391: I guess technically we could consider this to be also aligned. But it would need additional checks, e.g. whether offset is byte aligned, and whether the dynamic slice size is byte aligned.

akuegel (approved):

PR #37783, opened 2026-02-12
[xla:ffi] Add support for setting and getting state in all execution stages

ezhulenev:
Add support for setting and getting state in all execution stages.

There are 2 kinds of state that can be attached to FFI handler (by FFI handler itself):

1. Per-instance state: state attached to a custom call thunk and shared by all XLA executions
2. Per-execution: every call to GpuExecutable::Execute has a chance to create its own state that will have a lifetime matching the XLA execution.


xla/backends/gpu/runtime/thunk.h
```diff
@@ -246,6 +247,24 @@ class Thunk {
...
+  // synchronized). `ExecutionScopedState` is a container that allows thunks to
+  // put an arbitrary state that will have an execution scope, i.e. it will be
+  // automatically destroyed when GpuExecutable finishes execution. This allows
+  // thunks to pass arbitrary state between stages (from prepare to initialize
+  // and then to execute), without having to create a globally synchronized map
+  // and it also guarantees correct state life time, as leaving state in a map
+  // might lead to "leaks", as the map will be destroyed only when the
+  // executable is destroyed. It also thread-safe by construction as all thunks
+  // for a GPU program run sequentially from a single thread.
+  using ExecutionScopedState = absl::flat_hash_map<const Thunk*, std::any>;
```
sohaibiftikhar: This puts a restriction on execution scoped state to be copyable. Is this intended?

xla/backends/gpu/runtime/custom_call_thunk.cc
```diff
@@ -464,21 +474,38 @@ InvokeContext CustomCallThunk::BuildInvokeContext(
...
   }
 
+  // Lookup per-execution state for prepare and init stages.
+  ffi::ExecutionState* prepare_state = nullptr;
+  ffi::ExecutionState* initialize_state = nullptr;
+
+  if (execution_scoped_state) {
+    std::any& state = (*execution_scoped_state)[this];
+    if (!state.has_value()) {
+      state = std::make_shared<PrepareAndInitState>();
```
sohaibiftikhar: I see that you get around this by wrapping it in a shared pointer. This makes ownership semantics for the shared state not ideal imho.

Perhaps we should add a tsl::MoveOnlyAny or something similar (not as part of this change). Wdyt?

xla/backends/gpu/runtime/thunk.h
ezhulenev: It's unfortunate and annoying side-effect of `std::any`, but I don't think it's wort doing our own any for this, and I didn't find anything live move-only-any in. tsl/absl.

xla/backends/gpu/runtime/custom_call_thunk.cc
ezhulenev: Hm, let me see how hard it would be, indeed I'd find it super useful to XLA, and I recall other places in XLA with same problem with any.

ezhulenev: I found that tfrt alrady did that as UniqueAny, will do similar in tsl.

ezhulenev: Will migrate to https://github.com/openxla/xla/pull/37821 once it's submitted

sohaibiftikhar: Thank you.

ezhulenev: Done.

sohaibiftikhar (approved):

PR #37816, opened 2026-02-13
Build and use xxd from source

cerisier:
This is small yet important change since it was the only non hermetic, non posix tool used as part of the PJRT plugins build graph.

Building xxd from source enables fully remote, fully hermetic building of the PJRT plugins (at least) on a vanilla debian docker image (with a few other config modifications and toolchain registration).

See https://app.buildbuddy.io/invocation/41378ea1-d350-4c19-b08a-27ec865d95af 

If this is appealing enough, I'm willing to submit a few other patches to enable fully remote hermetic builds using user provided CC toolchain with full backward compliance with the existing config setup.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -28,6 +28,7 @@ bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "rules_python", version = "1.6.0")
 bazel_dep(name = "rules_shell", version = "0.6.1")
 bazel_dep(name = "snappy", version = "1.2.1")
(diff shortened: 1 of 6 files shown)


cerisier:
xxd needs to be fetched from source.

So for the CI to pass without bzlmod, `https://github.com/vim/vim/archive/refs/tags/v9.1.0917.tar.gz` would need to be added to the tf mirror (not sure about the process).

**This URL is the same that is used in the Bazel Central Registry, I only need that for workspace support.**

pifon2a (approved):
That's amazing. Thank you!

PR #37936, opened 2026-02-17
[NVIDIA GPU] Reduce rendezvous frequency in gpu executable

Tixxx:
📝 Summary of Changes
Rendezvous is needed in gpu executable to make sure all ranks have arrived and finished NCCL initialization for collectives.
This is not needed if all cliques in the clique request have done their first rendezvous. For repeated runs of already initialized cliques, we can skip rendezvous.
This also includes the Rendezvous before and after init.


PatriosTheGreat (commented):
I tried something similar a year ago and back then I faced with deadlocks on some multi-host benchmarks. I'll try to run the same benchmarks with this CL

xla/backends/gpu/runtime/collective_cliques.cc
```diff
@@ -126,6 +126,32 @@ absl::StatusOr<bool> CollectiveCliques::peer_access_enabled(
...
+}
+
+absl::StatusOr<bool> AllFirstRendezvousCompleted(
+    const CollectiveCliques& collective_cliques,
+    const std::vector<GpuCliqueKey>& requested_clique_keys) {
+  return collective_cliques.empty() ||
+         absl::c_all_of(
+             requested_clique_keys, [&](const GpuCliqueKey& clique_key) {
+               auto rend_flags =
+                   (*collective_cliques.GetCliqueFirstRendezvousFlags(
```
PatriosTheGreat: GetCliqueFirstRendezvousFlags returns StatusOr which seems to be ignored here. Could you please add error handling here?

PatriosTheGreat (changes requested):

xla/backends/gpu/runtime/collective_cliques.cc
Tixxx: good catch. added a basic check here

PatriosTheGreat (approved):

Tixxx:
@PatriosTheGreat  I changed it to be clique-module-based instead of just the clique. Can you take another look? Thanks

PatriosTheGreat:
Could you please fix clang-tidy complains from: https://github.com/openxla/xla/actions/runs/25391353818/job/74466400860?pr=37936

And could you check if JAX test ragged_collective_test_gpu works with this change?

I can share instructions how to run JAX tests with pending XLA if needed.

Tixxx:
> Could you please fix clang-tidy complains from: https://github.com/openxla/xla/actions/runs/25391353818/job/74466400860?pr=37936
> 
> And could you check if JAX test ragged_collective_test_gpu works with this change?
> 
> I can share instructions how to run JAX tests with pending XLA if needed.

@PatriosTheGreat  it looks like the clangTidy reports a bunch of errors not related to my change. It runs in all the files i touch and errors out on some old changes. I'm not sure if that makes sense to address all of them in this pr.
Yes the latest commit to use named modules to store rendezvous flags is to fix those tests, the main idea is to call rend once for each module rather than each clique since cliques are cached for the program lifetime which is why the tests time out

dimitar-asenov:
Once https://github.com/openxla/xla/pull/42218 is submitted, the clang-tidy output should be filtered to only report issues with changed lines.

dimitar-asenov:
The PR above has now landed.

@Tixxx Could you please rebase and check whether clang-tidy still complains?

dimitar-asenov (approved):

xla/backends/gpu/runtime/collective_cliques.cc
```diff
@@ -126,6 +129,35 @@ absl::StatusOr<bool> CollectiveCliques::peer_access_enabled(
...
+                    clique_key.ToString());
+  }
+  return (*clique->second)->GetFirstRendezvousFlags(module_name);
+}
+
+absl::StatusOr<bool> AllFirstRendezvousCompleted(
+    const CollectiveCliques& collective_cliques,
+    const std::vector<GpuCliqueKey>& requested_clique_keys,
+    const absl::string_view module_name) {
+  return !module_name.empty() &&
```
allanrenucci: AI review says:

> If `collective_cliques.empty()` is true, we should probably return true immediately to skip the rendezvous, even if `module_name` is empty. The current logic will return false if `module_name` is empty but collective_cliques is empty, which causes an unnecessary `RendezvousAfterInitialization` in `gpu_executable.cc` compared to the original behavior.

Seems reasonable. WDYT?

seantalts:
There are also some DWYU & CI failures

xla/backends/gpu/runtime/collective_cliques.cc
Tixxx: Yea that makes sense, i changed it

Tixxx:
> There are also some DWYU & CI failures

Hi, i think the DWYU failures are unrelated to my changes. The CI logs also dont show what the exact failure is. It seems to say the executor aborted something at the end. Trying to get the full logs gives me 403 permission error.

seantalts:
As mentioned on another PR unfortunately the way this works it that the internal tool flags issues in anything you touch. I believe this is intentional on their part.

seantalts (approved):

Tixxx:
> As mentioned on another PR unfortunately the way this works it that the internal tool flags issues in anything you touch. I believe this is intentional on their part.

Ok do i need to fix those in my PR?

seantalts:
Oops, didn't mean to press that. But it looks like the relevant CI errors have cleared up, just need to fix DWYU errors.

seantalts:
> Ok do i need to fix those in my PR?

Yes please, that will get it past the internal build.

Tixxx:
> > Ok do i need to fix those in my PR?
> 
> Yes please, that will get it past the internal build.

ok let me push the fix

Tixxx:
> > Ok do i need to fix those in my PR?
> 
> Yes please, that will get it past the internal build.

so i can't easily fix the failures below in this pr:
buildozer 'add deps :select_k_exec_stub' //xla/backends/gpu/runtime:select_k_exec_raft_test
buildozer 'add deps :select_k_exec_stub' //xla/backends/gpu/runtime:select_k_thunk

select_k_thunk has the dependency listed as if_cuda_configured but buildozer doesnt recognize that condition
select_k_exec_raft_test cannot have select_k_exec_stub as a permanent dependency because it's only for when gpu is absent and the test is cuda-only. If i add it, the test will fail with unimplemented error.

dimitar-asenov (approved):

dimitar-asenov:
Let me see if I can merge this internally.

belitskiy:
This was reverted due to segfaults in three tests in a JAX H100 multi-GPU job

Repro'd at the suspect commit
https://github.com/jax-ml/jax/actions/runs/25816222522/job/75845367932

Clean at the previous one
https://github.com/jax-ml/jax/actions/runs/25817760727/job/75850648858

ezhulenev:
This is per-executable rendezvous, I don't think it's even wort optimizing, so many different moving parts can lead to a deadlock that imho it's easier to pay the penalty of rendezvous on first call and sleep well :)

Tixxx:
> This is per-executable rendezvous, I don't think it's even wort optimizing, so many different moving parts can lead to a deadlock that imho it's easier to pay the penalty of rendezvous on first call and sleep well :)

The rendezvous eats up a significant portion of runtime though when the same executable is run in a loop for small models. For 8-gpu case, it costs on average 30 us with max up to 60 us. For a small model that only runs for a few hundred us per iteration, it's quite large. All the collectives are synchronizing devices, we should just eliminate any unnecessary host side syncs.
For training scenarios, the penalty is negligible. But has this never been an issue for your inference cases?

ezhulenev:
But it should be only once when the model called the first time? In steady state it should be only 1 rendezvous after thunk initialization to make sure that all resources are allocated.

Tixxx:
> But it should be only once when the model called the first time? In steady state it should be only 1 rendezvous after thunk initialization to make sure that all resources are allocated.

No, the rendezvous after init is called every time the module is called.

ezhulenev:
It also can be updated to use rendezvous flag to run only once, but it was added there for some reason... I think it had something to do with cuda graphs <-> NCCL deadlocks, when one rank gets too far ahead and launched NCCL kernels, and other ranks are in cuda graph create/update, it was causing a deadlock inside CUDA.

I'd try to add rendezvous once flag to executable, not to the clique.

Tixxx:
> It also can be updated to use rendezvous flag to run only once, but it was added there for some reason... I think it had something to do with cuda graphs <-> NCCL deadlocks, when one rank gets too far ahead and launched NCCL kernels, and other ranks are in cuda graph create/update, it was causing a deadlock inside CUDA.
> 
> I'd try to add rendezvous once flag to executable, not to the clique.

I heard about some issues with graph create, i think there's a warm up iteration added for command buffer to address that. I haven't heard about update causing issues though. I ran some benchmarks with cuda graph enabled for collectives but didnt run into that issue.

I remember there was a comment somewhere for those first rendezvous flags to move them to cliques, i see that the comments were removed recently so I guess that's not a TODO anymore. Adding a run once flag to executable makes sense too, and it will make it easier. I can update it.

ezhulenev:
I added it ~2 years ago, will not be too surprised of most of the issues already fixed.

dimitar-asenov (approved):

PatriosTheGreat:
Hi @Tixxx 
This change seems to lead to deadlocks with our benchmarks. I might need to revert it and make sure deadlock doesn't appear again. If this change is the problem I'll try to provide reproducer.

PR #38007, opened 2026-02-18
Gracefully handle missing handler on legacy custom call path

draganmladjenovic:
📝 Summary of Changes
Add error checking when doing lookup for legacy custom call handler

🎯 Justification
Prevents segmentation fault when trying to invoke legacy custom call with no handler registred

🚀 Kind of Contribution
🐛 Bug Fix

N\A

🧪 Unit Tests:
Ne test case added to //xla/service/gpu:custom_call_test

🧪 Execution Tests:
N\A


beckerhe (approved):
Thank you!

PR #38009, opened 2026-02-18
[xla:pjrt:gpu] Pass network nodes to LocalTopologyProto

ezhulenev:
Don't forget to pass network nodes to local topology proto, so that the coordinator process can use them to do network-topology-optimized global device assignment.

JAX PR that allows keeping devices sorted by assigned global device id: https://github.com/jax-ml/jax/pull/35178

### DIFF
--- a/xla/pjrt/distributed/topology_util.cc
+++ b/xla/pjrt/distributed/topology_util.cc
@@ -339,8 +339,20 @@ absl::Status ExchangeTopologies(absl::string_view platform, int node_id,
         kv_store->Get(global_topology_key, get_global_topology_timeout));
     global_topology->ParseFromString(global_topology_str);
   }
-  VLOG(3) << "Global topology for platform " << platform << ":\n"
-          << global_topology->DebugString();
+
+  // Because we might do global topology assignment based on network proximity
+  // of XLA processes, the process id might not be ordered anymore, however it
+  // does not matter for XLA at run time, as we always lookup replica and
+  // partition id based on global topology we compute here.
+  if (VLOG_IS_ON(3)) {
+    VLOG(3) << "Global topology for platform " << platform
+            << ": num_processes=" << global_topology->processes_size();
+    for (size_t rank = 0; rank < global_topology->processes_size(); ++rank) {
+      VLOG(3) << "topology for process rank #" << rank << ":\n"
+              << global_topology->processes(rank).DebugString();
+    }
+  }
+
   return absl::OkStatus();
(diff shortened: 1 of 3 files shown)


mwhittaker (approved):

xla/python/pjrt_ifrt/pjrt_client.cc
```diff
@@ -797,14 +809,14 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> PjRtClient::Create(
                           client.get(), pjrt_client, global_topology));
 
   if (options.sort_devices_by_process_index) {
-    absl::c_sort(devices, [](const std::unique_ptr<PjRtDevice>& a,
-                             const std::unique_ptr<PjRtDevice>& b) {
+    absl::c_stable_sort(devices, [](const std::unique_ptr<PjRtDevice>& a,
```
hyeontaek: Thanks Eugene!

Could you elaborate the context of this change? It is unclear to me if stable sorting helps. Since IFRT requires all IFRT devices IDs to be unique within a single IFRT client, and `PjRtDevice` here is `xla::ifrt::PjRtDevice`, not `xla::PjRtDevice`, I think both sorting below would not see any duplicate keys, and thus stable sorting is not necessary.

ezhulenev: You are right, forgot what was my motivation last week :) reverted it

xla/pjrt/distributed/topology_util.cc
```diff
@@ -339,8 +339,18 @@ absl::Status ExchangeTopologies(absl::string_view platform, int node_id,
         kv_store->Get(global_topology_key, get_global_topology_timeout));
     global_topology->ParseFromString(global_topology_str);
   }
-  VLOG(3) << "Global topology for platform " << platform << ":\n"
-          << global_topology->DebugString();
+
+  // Because we might do global topology assignment based on network proximity
```
hyeontaek, line 343: Nit: When a loop is only doing VLOG, consider using the following pattern so that we can skip looping completely when debug logging is not enabled:

```
if (VLOG_IS_ON(3)) {
  VLOG(3) << ...;
  for (...) {
    VLOG(3) << ...;
  }
}
```

ezhulenev: Done

hyeontaek (approved):

penpornk (approved):

PR #38105, opened 2026-02-20
[GPU] Erase stale sibling sub-cliques to avoid deadlocks

terryysun:
📝 Summary of Changes
When a cached sub-clique is evicted due to a parent communicator mismatch, also evict all sibling sub-cliques that were split from the same old parent. This prevents asymmetric cache state where some siblings skip `ncclCommSplit` (cache hit) while others trigger it (cache miss), violating the collective requirement that all parent ranks participate simultaneously.

🎯 Justification
`ncclCommSplit` is collective over all ranks of the parent communicator — every rank must call it. The process-global clique cache `ProcessGpuCliques` can develop asymmetric state when different computations create the same child clique from different parents:

  1. Computation A (8-device mesh): creates parent {0,...,7}, splits children {0,1}, {2,3}, {4,5}, {6,7} — all with parent_ = P8.
  2. Computation B (4-device mesh): needs child {0,1}, finds it cached with P8, but wants to split from P4 — evicts and re-creates {0,1} with parent_ = P4. Children {4,5}, {6,7} are not accessed and retain parent_ = P8.
  3. Computation A repeats: {0,1} has wrong parent → evict → call `ncclCommSplit` on P8. But {4,5} still matches P8 → cache hit → skip `ncclCommSplit`. Deadlock: only 4 of 8 parent ranks called the collective.

This deadlock is reproducible by running `pytest -s tests/pmap_test.py tests/pjit_test.py` in the JAX test suite .

🚀 Kind of Contribution
🐛 Bug Fix

N/A.

🧪 Unit Tests:
N/A.

🧪 Execution Tests:
Verified with aforementioned JAX tests.

### DIFF
--- a/xla/backends/gpu/collectives/gpu_clique.cc
+++ b/xla/backends/gpu/collectives/gpu_clique.cc
@@ -146,6 +146,10 @@ bool LockableGpuClique::HasParent(const GpuClique* parent) const {
   return this->value().parent() == parent;
 }
 
+const GpuClique* LockableGpuClique::parent() const {
+  return this->value().parent();
+}
+
 std::string LockableGpuClique::DebugString() const {
   return absl::StrFormat("LockableGpuClique: %s", value().DebugString());
 }
--- a/xla/backends/gpu/collectives/gpu_clique.h
+++ b/xla/backends/gpu/collectives/gpu_clique.h
@@ -124,6 +124,9 @@ class LockableGpuClique : public Lockable<GpuClique, GpuClique::LockableName> {
   // Returns if *this lockable clique has a given parent.
   bool HasParent(const GpuClique* parent) const;
 
+  // Returns the parent clique pointer, or nullptr if not created by splitting.
+  const GpuClique* parent() const;
+
   std::string DebugString() const;
 
   // Checks for async errors for all the communicators in the clique without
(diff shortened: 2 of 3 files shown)


ezhulenev:
I think that eviction logic is wring, it's too aggressive. If `split_from` is a subset of `to_lock->parent()` then we can guarantee that all ranks that will split from `split_from` must see exactly the same parent.

Otherwise if we'll alternate between computations A and B we'll keep re-constructing nccl comms on each call.

Running B and then A is the real problem, then split/no-split {0,1} can lead to deadlock, but if we are running computations starting from the ones that span larger number of decides, it should be safe to reuse comms.

terryysun:
> I think that eviction logic is wring, it's too aggressive. If `split_from` is a subset of `to_lock->parent()` then we can guarantee that all ranks that will split from `split_from` must see exactly the same parent.
> 
> Otherwise if we'll alternate between computations A and B we'll keep re-constructing nccl comms on each call.
> 
> Running B and then A is the real problem, then split/no-split {0,1} can lead to deadlock, but if we are running computations starting from the ones that span larger number of decides, it should be safe to reuse comms.

Thanks! Agreed that running B then A is the real problem. For alternating between A and B, this cache eviction will be necessary for correctness. If we always run larger computation there should not be a problem and the added logic will never be triggered, but I'm not sure if there's a good way to enforce that. Any suggestions?

ezhulenev:
I think that https://github.com/openxla/xla/pull/37063 has a correct solution. When we create a larger clique it means that all devices went through local rendezvous process and global nccl comm bootstrap, so it must be safe to abort and erase cliques which are subset of newly create big clique. It's certainly true for all practical training workloads, it might be not always true in presence of concurrent XLA execution, but I'm sure we have many other bugs when it happens.

PR #37806, opened 2026-02-13
[GPU] Emitters: improve loop unrolling heuristic.

sergachev:
📝 Summary of Changes
Make the heuristic traverse xla.pure_call computations while estimating the
cost of unrolling. Increase cost of math.exp.

🎯 Justification
Speeds up some fusions by preventing excessive unrolling and heavy register pressure it causes.
This one gets 15% faster on B200:
```
f {
  tmp_0 = bf16[16384,14336] parameter(1)
  tmp_1 = bf16[2,8192,14336] bitcast(tmp_0)
  tmp_2 = bf16[] constant(1)
  tmp_3 = f32[] convert(tmp_2)
  tmp_4 = f32[2,8192,14336] broadcast(tmp_3), dimensions={}
  tmp_5 = bf16[2,8192,14336] negate(tmp_1)
  tmp_6 = bf16[2,8192,14336] exponential(tmp_5)
  tmp_7 = bf16[2,8192,14336] broadcast(tmp_2), dimensions={}
  tmp_8 = bf16[2,8192,14336] add(tmp_6, tmp_7)
  tmp_9 = f32[2,8192,14336] convert(tmp_8)
  tmp_10 = f32[2,8192,14336] divide(tmp_4, tmp_9)
  tmp_11 = bf16[2,8192,14336] convert(tmp_10)
  tmp_12 = bf16[2,8192,14336] multiply(tmp_1, tmp_11)
  tmp_13 = bf16[16384,14336] parameter(0)
  tmp_14 = bf16[2,8192,14336] bitcast(tmp_13)
  tmp_15 = bf16[2,8192,14336] multiply(tmp_12, tmp_14)
  tmp_16 = bf16[2,8192,14336] abs(tmp_15)
  tmp_17 = bf16[28672,8192] bitcast(tmp_16)
  tmp_18 = bf16[] constant(-inf)
  tmp_19 = bf16[28672] reduce(tmp_17, tmp_18), dimensions={1}, to_apply={
    _ = bf16[] maximum(bf16[] parameter(0), bf16[] parameter(1))
  }
  ROOT tmp_20 = (bf16[28672], bf16[2,8192,14336]) tuple(tmp_19, tmp_15)
}
```

🚀 Kind of Contribution
⚡️ Performance Improvement

📊 Benchmark (for Performance Improvements)
8x B200:

| Benchmark                                          | Speedup | Memory | Compilation speedup |
|----------------------------------------------------|---------|--------|---------------------|
[...]

PR #38174, opened 2026-02-23
[GPU] Unroll and vectorize up to 32 elements / 256 bits on Blackwell.

sergachev:
📝 Summary of Changes
Increase maximum unroll factor and maximum vector width.

🎯 Justification
Speeds up kernels.

🚀 Kind of Contribution
⚡️ Performance Improvement

B200:

| Benchmark                                           | Speedup | Error | Memory | Compilation speedup |
|-----------------------------------------------------|---------|-------|--------|---------------------|
|  gpu_hlo                                            |   1.03x | 0.01x |  1.00x |               0.89x |
|  hlo_llama3_8b_bf16_activation_offloading_1x8       |   0.99x | 0.01x |  1.00x |               0.98x |
|  hlo_llama3_8b_cp_bf16_1x4                          |   0.99x | 0.01x |  1.00x |               0.92x |
|  hlo_llama31_8b_bf16_1x8                            |   1.01x | 0.01x |  1.00x |               0.99x |
|  hlo_llama31_8b_fp8_1x8                             |   0.99x | 0.03x |  1.00x |               0.98x |
|  hlo_mixtral_8x7b_bf16_1x8                          |   1.05x | 0.01x |  1.00x |               0.96x |
|  nv_maxtext_1n1g_jit_train_step_before_optimization |   1.01x | 0.00x |  1.00x |               0.95x |
|  gemma2_2b_keras_jax                                |   1.04x | 0.02x |  1.00x |               1.03x |
|  gemma3_1b_flax_call                                |   1.01x | 0.11x |  1.00x |               1.05x |
|  gemma3_1b_flax_sample_loop                         |   1.37x | 0.43x |  1.00x |               0.99x |
|  gemma3_4b_flax_call                                |   1.39x | 0.17x |  1.00x |               0.99x |
|  gemma3_4b_flax_sample_loop                         |   0.92x | 0.17x |  1.00x |               0.93x |
|  gemma3_12b_flax_call                               |   1.03x | 0.01x |  1.00x |               1.07x |
|  gemma3_12b_flax_sample_loop                        |   1.02x | 0.06x |  1.00x |               1.08x |
|  4                                                  |   3.43x | 0.02x |  1.00x |               1.22x |
|  8                                                  |   2.50x | 0.03x |  1.00x |               0.94x |
|  16                                                 |   1.38x | 0.03x |  1.00x |               1.30x |
|  32                                                 |   1.39x | 0.02x |  1.00x |               1.15x |
|  64                                                 |   1.01x | 0.02x |  1.00x |               1.12x |
|  128                                                |   1.00x | 0.01x |  1.00x |               1.15x |
|  ComputeLoopFusionConfig32Bit                       |   1.26x | 0.03x |  1.00x |               1.15x |
|  ComputeLoopFusionConfig                            |   1.16x | 0.04x |  1.00x |               1.05x |
|  pr38174_1                                          |   1.00x | 0.00x |  1.00x |               0.98x |
|  pr38174_2                                          |   1.46x | 0.00x |  1.00x |               1.03x |

ComputeLoopFusionConfig* are taken from affected cases of gpu_fusible_test. pr38174_1 and pr38174_2 were provided by Adrian in the comments to the PR. HLOs for microbenchmarks are in the first comment.

🧪 Unit Tests:
yes

🧪 Execution Tests:
no

### DIFF
--- a/xla/codegen/emitters/transforms/vectorize_loads_stores.cc
+++ b/xla/codegen/emitters/transforms/vectorize_loads_stores.cc
@@ -132,7 +132,7 @@ int64_t GetAlignmentOfRemainder(mlir::AffineExpr expr,
 // Attempts to extract the vector type for the given loop. This means:
 // - checks that the lower bound is 0
 // - checks that the step is 1
-// - checks that the upper bound is 2, 4, or 8.
+// - checks that the upper bound is a power of 2 between 2 and 32.
 // Returns a vector type with the given upper bound and the tensor's element
 // type.
 // All tensors are 1D after flatten-tensors pass.
@@ -151,10 +151,14 @@ mlir::VectorType GetVectorType(mlir::RankedTensorType tensor_type,
       mlir::getConstantIntValue(loop.getLowerBound()) != 0) {
     return nullptr;
   }
-  std::optional<int> vector_size =
+  std::optional<unsigned int> vector_size =
       mlir::getConstantIntValue(loop.getUpperBound());
-  if (vector_size != 2 && vector_size != 4 && vector_size != 8) {
+  if (vector_size < 2 || vector_size > 32 ||
+      !absl::has_single_bit(*vector_size)) {
     return nullptr;  // Unsupported vector size.
+  }
+  if (tensor_type.getElementTypeBitWidth() * *vector_size > 256) {
+    return nullptr;  // Vector size is too large.
   }
   if (tensor_type.getShape().back() % *vector_size) {
     return nullptr;  // Misaligned start indices.
--- a/xla/codegen/emitters/transforms/vectorize_loads_stores.cc
+++ b/xla/codegen/emitters/transforms/vectorize_loads_stores.cc
@@ -20,6 +20,7 @@ limitations under the License.
 #include <utility>
 
 #include "absl/log/check.h"
+#include "absl/numeric/bits.h"
 #include "llvm/ADT/APInt.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
@@ -151,10 +152,10 @@ mlir::VectorType GetVectorType(mlir::RankedTensorType tensor_type,
       mlir::getConstantIntValue(loop.getLowerBound()) != 0) {
     return nullptr;
   }
-  std::optional<unsigned int> vector_size =
+  std::optional<int64_t> vector_size =
       mlir::getConstantIntValue(loop.getUpperBound());
   if (vector_size < 2 || vector_size > 32 ||
-      !absl::has_single_bit(*vector_size)) {
+      !absl::has_single_bit(static_cast<unsigned int>(*vector_size))) {
     return nullptr;  // Unsupported vector size.
   }
   if (tensor_type.getElementTypeBitWidth() * *vector_size > 256) {
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -3019,6 +3019,7 @@ cc_library(
         "//xla/hlo/utils:hlo_traversal",
         "//xla/service:instruction_fusion",
         "//xla/stream_executor:device_description",
+        "//xla/stream_executor:semantic_version",
         "@com_google_absl//absl/algorithm:container",
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/container:flat_hash_set",
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -3007,6 +3007,7 @@ cc_library(
         ":ir_emission_utils",
         ":launch_dimensions",
         ":reduction_utils",
+        "//xla:comparison_util",
         "//xla:permutation_util",
         "//xla:shape_util",
         "//xla:side_effect_util",
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -3007,7 +3007,6 @@ cc_library(
         ":ir_emission_utils",
         ":launch_dimensions",
         ":reduction_utils",
-        "//xla:comparison_util",
         "//xla:permutation_util",
         "//xla:shape_util",
         "//xla:side_effect_util",
--- a/xla/service/gpu/gpu_fusible.cc
+++ b/xla/service/gpu/gpu_fusible.cc
@@ -39,6 +39,7 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/hlo/utils/hlo_traversal.h"
 #include "xla/permutation_util.h"
+#include "xla/primitive_util.h"
 #include "xla/service/gpu/alias_info.h"
 #include "xla/service/gpu/backend_configs.pb.h"
 #include "xla/service/gpu/hlo_fusion_analysis.h"
@@ -112,8 +113,9 @@ int ComputeMaxUnrollFactor(int64_t num_elements, int64_t max_unroll) {
 }  // namespace
 
 int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
+  constexpr int kDefaultUnrollFactor = 4;
   if (analysis == nullptr) {
-    return 4;
+    return kDefaultUnrollFactor;
   }
 
   // On Blackwell we would like to increase the maximum unroll factor to 8, as
@@ -140,30 +142,52 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
   // don't face register pressure. This part of the heuristic may need
   // improvements.
   constexpr int kMaxNumOutputsForFullUnrolling = 1;
-  // On PTX level, we can vectorize with v4.b32, but not with v8.b32. So
-  // higher unroll factor does not make sense with 32 bit or more.
-  constexpr int kMaxBitsToVectorizeWithVectorSize4 = 32;
+  // On PTX level, we can efficiently vectorize up to 256 bits (e.g. v8.b32).
+  // So we allow aggressive (>4x) unrolling for types smaller than 64 bits
+  // to reach this width.
+  constexpr int kMaxBitsForAggressiveUnrolling = 64;
   DebugOptions debug_options = analysis->fusion_root(0)
                                    .instruction()
                                    .GetModule()
                                    ->config()
                                    .debug_options();
   if (analysis->device_info().cuda_compute_capability().IsBlackwell() &&
-      analysis->emitter_fusion_kind() ==
-          HloFusionAnalysis::EmitterFusionKind::kLoop &&
+      (analysis->emitter_fusion_kind() ==
+           HloFusionAnalysis::EmitterFusionKind::kLoop ||
+       analysis->emitter_fusion_kind() ==
+           HloFusionAnalysis::EmitterFusionKind::kTranspose) &&
       analysis->input_output_info().smallest_output_dtype_bits <
-          kMaxBitsToVectorizeWithVectorSize4 &&
+          kMaxBitsForAggressiveUnrolling &&
       analysis->fusion_root_count() <= kMaxNumOutputsForFullUnrolling &&
-      debug_options.xla_gpu_experimental_allow_unroll_factor_eight() &&
-      !HloAnyOf(
-          analysis->fusion(),
-          [](HloInstructionAdaptor node) {
-            return node.opcode() == HloOpcode::kReduce;
-          }) &&
-      !ContainsTransposeWithSmallMostMinorDim(analysis->fusion(), 8)) {
-    return 8;
+      !HloAnyOf(analysis->fusion(), [](HloInstructionAdaptor node) {
+        return node.opcode() == HloOpcode::kReduce;
+      })) {
+    int64_t max_dtype_bits = 0;
+    for (const HloInstruction* param : analysis->fusion().GetParameters()) {
+      max_dtype_bits = std::max(
+          max_dtype_bits, static_cast<int64_t>(primitive_util::StorageBitWidth(
+                              param->shape().element_type())));
+    }
+    for (const HloInstructionAdaptor& root : analysis->fusion_roots()) {
+      max_dtype_bits = std::max(
+          max_dtype_bits, static_cast<int64_t>(primitive_util::StorageBitWidth(
+                              root.instruction().shape().element_type())));
+    }
+
+    int64_t unroll_factor =
+        debug_options.xla_gpu_experimental_max_unroll_factor() * 8 /
+        std::max(int64_t{8}, max_dtype_bits);
+    unroll_factor =
+        1LL << (absl::bit_width(static_cast<uint64_t>(unroll_factor)) - 1);
+    while (unroll_factor > kDefaultUnrollFactor) {
+      if (!ContainsTransposeWithSmallMostMinorDim(analysis->fusion(),
+                                                  unroll_factor)) {
+        return unroll_factor;
+      }
+      unroll_factor /= 2;
+    }
   }
-  return 4;
+  return kDefaultUnrollFactor;
 }
 
 bool IsPhysicallyTransposing(const HloInstruction& instr) {
--- a/xla/service/gpu/gpu_fusible.cc
+++ b/xla/service/gpu/gpu_fusible.cc
@@ -51,6 +51,7 @@ limitations under the License.
 #include "xla/shape_util.h"
 #include "xla/side_effect_util.h"
 #include "xla/stream_executor/device_description.h"
+#include "xla/stream_executor/semantic_version.h"
 #include "xla/util.h"
 #include "xla/xla.pb.h"
 
@@ -142,22 +143,23 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
   // don't face register pressure. This part of the heuristic may need
   // improvements.
   constexpr int kMaxNumOutputsForFullUnrolling = 1;
-  // On PTX level, we can efficiently vectorize up to 256 bits (e.g. v8.b32).
-  // So we allow aggressive (>4x) unrolling for types smaller than 64 bits
-  // to reach this width.
-  constexpr int kMaxBitsForAggressiveUnrolling = 64;
-  DebugOptions debug_options = analysis->fusion_root(0)
-                                   .instruction()
-                                   .GetModule()
-                                   ->config()
-                                   .debug_options();
+  // On PTX level, we can efficiently vectorize up to 128/256 bits (v4/v8.b32).
+  // So we allow aggressive unrolling for narrow types to reach this width.
+  const int64_t max_vector_bit_width =
+      (analysis->device_info().cuda_compute_capability().IsBlackwell() &&
+       analysis->device_info().compile_time_toolkit_version() >=
+           stream_executor::SemanticVersion(12, 9, 0))
+          ? 256
+          : 128;
+  const int max_bits_for_aggressive_unrolling =
+      max_vector_bit_width / kDefaultUnrollFactor;
   if (analysis->device_info().cuda_compute_capability().IsBlackwell() &&
       (analysis->emitter_fusion_kind() ==
            HloFusionAnalysis::EmitterFusionKind::kLoop ||
        analysis->emitter_fusion_kind() ==
            HloFusionAnalysis::EmitterFusionKind::kTranspose) &&
-      analysis->input_output_info().smallest_output_dtype_bits <
-          kMaxBitsForAggressiveUnrolling &&
+      analysis->input_output_info().smallest_output_dtype_bits <=
+          max_bits_for_aggressive_unrolling &&
       analysis->fusion_root_count() <= kMaxNumOutputsForFullUnrolling &&
       !HloAnyOf(analysis->fusion(), [](HloInstructionAdaptor node) {
         return node.opcode() == HloOpcode::kReduce;
@@ -173,10 +175,17 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
           max_dtype_bits, static_cast<int64_t>(primitive_util::StorageBitWidth(
                               root.instruction().shape().element_type())));
     }
+    // Do not unroll sub-byte types further for now.
+    max_dtype_bits = std::max(int64_t{8}, max_dtype_bits);
 
     int64_t unroll_factor =
-        debug_options.xla_gpu_experimental_max_unroll_factor() * 8 /
-        std::max(int64_t{8}, max_dtype_bits);
+        std::min(int64_t(analysis->fusion_root(0)
+                             .instruction()
+                             .GetModule()
+                             ->config()
+                             .debug_options()
+                             .xla_gpu_experimental_max_unroll_factor()),
+                 max_vector_bit_width / max_dtype_bits);
     unroll_factor =
         1LL << (absl::bit_width(static_cast<uint64_t>(unroll_factor)) - 1);
     while (unroll_factor > kDefaultUnrollFactor) {
--- a/xla/service/gpu/gpu_fusible.cc
+++ b/xla/service/gpu/gpu_fusible.cc
@@ -114,12 +114,12 @@ int ComputeMaxUnrollFactor(int64_t num_elements, int64_t max_unroll) {
 }  // namespace
 
 int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
-  constexpr int kDefaultUnrollFactor = 4;
+  constexpr int64_t kDefaultUnrollFactor = 4;
   if (analysis == nullptr) {
     return kDefaultUnrollFactor;
   }
 
-  // On Blackwell we would like to increase the maximum unroll factor to 8, as
+  // On Blackwell we would like to increase the maximum unroll factor, as
   // we need more vectorization for full performance.
   // However we need to check additional conditions:
   //   - Unrolling is potentially bad for fusions with reductions, where one
@@ -151,8 +151,10 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
            stream_executor::SemanticVersion(12, 9, 0))
           ? 256
           : 128;
+  // Apply unroll factors >= 2 * kDefaultUnrollFactor only to narrow enough data
+  // types.
   const int max_bits_for_aggressive_unrolling =
-      max_vector_bit_width / kDefaultUnrollFactor;
+      max_vector_bit_width / (2 * kDefaultUnrollFactor);
   if (analysis->device_info().cuda_compute_capability().IsBlackwell() &&
       (analysis->emitter_fusion_kind() ==
            HloFusionAnalysis::EmitterFusionKind::kLoop ||
@@ -195,7 +197,7 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
                                                   unroll_factor)) {
       unroll_factor /= 2;
     }
-    return unroll_factor;
+    return std::max(unroll_factor, kDefaultUnrollFactor * 2);
   }
   return kDefaultUnrollFactor;
 }
--- a/xla/service/gpu/gpu_fusible.cc
+++ b/xla/service/gpu/gpu_fusible.cc
@@ -164,7 +164,8 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
       !HloAnyOf(analysis->fusion(), [](HloInstructionAdaptor node) {
         return node.opcode() == HloOpcode::kReduce;
       })) {
-    int64_t max_dtype_bits = 0;
+    // Do not unroll sub-byte types further for now.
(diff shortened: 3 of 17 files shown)


sergachev:
```
ComputeLoopFusionConfig {
  p0 = f16[2048,1024]{1,0} parameter(0)
  ROOT res = f16[2048,1024]{1,0} negate(p0)
}

ComputeLoopFusionConfig32Bit {
  p0 = f32[2048,1024]{1,0} parameter(0)
  ROOT res = f32[2048,1024]{1,0} negate(p0)
}

m4 {
  t = s4[14336,8,512]{2,1,0:E(4)} transpose(s4[8,14336,512]{2,1,0:E(4)} parameter(0)), dimensions={1,0,2}
}

m8 {
  t = s8[14336,8,256]{2,1,0} transpose(s8[8,14336,256]{2,1,0} parameter(0)), dimensions={1,0,2}
}

m16 {
  t = s16[14336,8,128]{2,1,0} transpose(s16[8,14336,128]{2,1,0} parameter(0)), dimensions={1,0,2}
}

m32 {
  t = s32[14336,8,64]{2,1,0} transpose(s32[8,14336,64]{2,1,0} parameter(0)), dimensions={1,0,2}
}

m64 {
  t = s64[14336,8,32]{2,1,0} transpose(s64[8,14336,32]{2,1,0} parameter(0)), dimensions={1,0,2}
}

m128 {
  t = c128[14336,8,16]{2,1,0} transpose(c128[8,14336,16]{2,1,0} parameter(0)), dimensions={1,0,2}
}
```

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +142,52 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
-          }) &&
-      !ContainsTransposeWithSmallMostMinorDim(analysis->fusion(), 8)) {
-    return 8;
+      !HloAnyOf(analysis->fusion(), [](HloInstructionAdaptor node) {
+        return node.opcode() == HloOpcode::kReduce;
+      })) {
+    int64_t max_dtype_bits = 0;
+    for (const HloInstruction* param : analysis->fusion().GetParameters()) {
+      max_dtype_bits = std::max(max_dtype_bits,
+                                static_cast<int64_t>(primitive_util::BitWidth(
```
akuegel: I suggest do use primitive_util::StorageBitWidth. Otherwise you would get size 1 for PRED type, which is wrong as we store it as 1 byte.

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +142,52 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
   // don't face register pressure. This part of the heuristic may need
   // improvements.
   constexpr int kMaxNumOutputsForFullUnrolling = 1;
-  // On PTX level, we can vectorize with v4.b32, but not with v8.b32. So
-  // higher unroll factor does not make sense with 32 bit or more.
-  constexpr int kMaxBitsToVectorizeWithVectorSize4 = 32;
+  // On PTX level, we can efficiently vectorize up to 256 bits (e.g. v8.b32).
```
akuegel: Interesting. I thought I had tested it and never saw v8.b32 being used. Has that changed recently in the PTX backend?

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +142,52 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
+                                static_cast<int64_t>(primitive_util::BitWidth(
+                                    param->shape().element_type())));
+    }
+    for (const HloInstructionAdaptor& root : analysis->fusion_roots()) {
+      max_dtype_bits = std::max(
+          max_dtype_bits, static_cast<int64_t>(primitive_util::BitWidth(
+                              root.instruction().shape().element_type())));
+    }
+
+    int64_t unroll_factor =
```
akuegel: Even with unroll factor 8 I have hit register pressure quite easily for fusions containing e.g. exponential, or something like this. Here is an example:

```
HloModule m

ENTRY main {
  %param_3.3093 = f32[2908900,128]{1,0} parameter(3)
  %slice.33.14.clone.1 = f32[2899200,128]{1,0} slice(%param_3.3093), slice={[9700:2908900], [0:128]}
  %param_2.4305 = f32[128]{0} parameter(2)
  %broadcast.2897.40.clone.1 = f32[2899200,128]{1,0} broadcast(%param_2.4305), dimensions={1}
  %add.54.21.clone.1 = f32[2899200,128]{1,0} add(%slice.33.14.clone.1, %broadcast.2897.40.clone.1)
  %compare.8.3.clone.1 = pred[2899200,128]{1,0} compare(%add.54.21.clone.1, %add.54.21.clone.1), direction=NE
  %constant_670_4_clone_1 = f32[] constant(-1.38629436)
  %broadcast.547.22.clone.1 = f32[2899200,128]{1,0} broadcast(%constant_670_4_clone_1), dimensions={}
  %add.8.3.clone.1 = f32[2899200,128]{1,0} add(%add.54.21.clone.1, %broadcast.547.22.clone.1)
  %constant_266_8_clone_1 = f32[] constant(-0.693147182)
  %broadcast.324.62.clone.1 = f32[2899200,128]{1,0} broadcast(%constant_266_8_clone_1), dimensions={}
  %add.53.3.clone.1 = f32[2899200,128]{1,0} add(%add.54.21.clone.1, %broadcast.324.62.clone.1)
  %maximum.2.3.clone.1 = f32[2899200,128]{1,0} maximum(%add.53.3.clone.1, %broadcast.324.62.clone.1)
  %abs.2.7.clone.1 = f32[2899200,128]{1,0} abs(%add.54.21.clone.1)
  %negate.2.5.clone.1 = f32[2899200,128]{1,0} negate(%abs.2.7.clone.1)
  %exponential.2.3.clone.1 = f32[2899200,128]{1,0} exponential(%negate.2.5.clone.1)
  %negate.2.3.clone.1 = f32[2899200,128]{1,0} negate(%exponential.2.3.clone.1)
  %add.9.3.clone.1 = f32[2899200,128]{1,0} add(%maximum.2.3.clone.1, %negate.2.3.clone.1)
  %select.9.1.clone.1 = f32[2899200,128]{1,0} select(%compare.8.3.clone.1, %add.8.3.clone.1, %add.9.3.clone.1)
  %param_1.6924 = f32[2899200,128]{1,0} parameter(1)
  %add.2923.1 = f32[2899200,128]{1,0} add(%select.9.1.clone.1, %param_1.6924)
  %convert.2924.1 = f64[2899200,128]{1,0} convert(%add.2923.1)
  %param_4.2558 = f32[2908900,128]{1,0} parameter(4)
  %slice.36.14.clone.1 = f32[2899200,128]{1,0} slice(%param_4.2558), slice={[0:2899200], [0:128]}
  %add.60.21.clone.1 = f32[2899200,128]{1,0} add(%slice.36.14.clone.1, %broadcast.2897.40.clone.1)
  %compare.11.3.clone.1 = pred[2899200,128]{1,0} compare(%add.60.21.clone.1, %add.60.21.clone.1), direction=NE
  %add.14.3.clone.1 = f32[2899200,128]{1,0} add(%add.60.21.clone.1, %broadcast.547.22.clone.1)
  %add.59.3.clone.1 = f32[2899200,128]{1,0} add(%add.60.21.clone.1, %broadcast.324.62.clone.1)
  %maximum.5.3.clone.1 = f32[2899200,128]{1,0} maximum(%add.59.3.clone.1, %broadcast.324.62.clone.1)
  %abs.5.7.clone.1 = f32[2899200,128]{1,0} abs(%add.60.21.clone.1)
  %negate.5.5.clone.1 = f32[2899200,128]{1,0} negate(%abs.5.7.clone.1)
  %exponential.5.3.clone.1 = f32[2899200,128]{1,0} exponential(%negate.5.5.clone.1)
  %negate.5.3.clone.1 = f32[2899200,128]{1,0} negate(%exponential.5.3.clone.1)
  %add.15.3.clone.1 = f32[2899200,128]{1,0} add(%maximum.5.3.clone.1, %negate.5.3.clone.1)
  %select.12.1.clone.1 = f32[2899200,128]{1,0} select(%compare.11.3.clone.1, %add.14.3.clone.1, %add.15.3.clone.1)
  %param_0.5662 = f32[2899200,128]{1,0} parameter(0)
  %add.3409.1 = f32[2899200,128]{1,0} add(%select.12.1.clone.1, %param_0.5662)
  %convert.3410.1 = f64[2899200,128]{1,0} convert(%add.3409.1)
  %add.3411.1 = f64[2899200,128]{1,0} add(%convert.2924.1, %convert.3410.1)
  %convert.3412.1 = f32[2899200,128]{1,0} convert(%add.3411.1)
  ROOT %tuple.103 = (f32[2899200,128]{1,0}, f32[2899200,128]{1,0}, f32[2899200,128]{1,0}) tuple(%convert.3412.1, %select.9.1.clone.1, %select.12.1.clone.1)
}
```

Can you please try whether your change would possibly regress that? And if yes, any ideas what we can do?

xla/service/gpu/gpu_fusible.cc
sergachev: Requires PTX ISA 8.8 and sm_100 indeed.

xla/service/gpu/gpu_fusible.cc
sergachev: I ran this HLO on B200 and it uses only 4xf32 vectors already before my change - so it is not affected. I recognise that I might need to narrow down the scope of this change if we see regressions somewhere. Speaking of exp and register pressure, I have another change you might be interested in - https://github.com/openxla/xla/pull/37806.

xla/service/gpu/gpu_fusible.cc
akuegel: In that case, I think we should guard it that we don't use it for cuda version < 12.9 (as far as I understand we need 12.9 or above for PTX ISA 8.8). Internally we still use 12.8.1 by default, so this change would cause regressions.

xla/service/gpu/gpu_fusible.cc
akuegel: It uses 4xf32 because I added the logic to only allow unroll factor 8 for types with < 32 bits. As far as I can tell, with your change you are now allowing also for types with 32 bits (by using kMaxBitsForAggressiveUnrolling = 64). Maybe there is yet another condition that makes sure that still unroll factor of 4 is used that I forgot?

Regarding exp and register pressure: that seems for a later unrolling pass. However I think we may already unroll before that pass (based on the decision here from gpu_fusible). Why we have this logic in two places, I don't know. In any case, from the debugging I did it seems that somewhat small unroll factor is still fine for exp, but at some point we hit register pressure. When exactly probably depends on other things as well. It seems quite hard to avoid regressions. Likely we need to find a good balance of improvements and some minor regressions.

akuegel (approved):
I am approving this to have an easier way to run our own benchmarks. Will report back whether we hit regressions. I will try running with a new enough cuda version with a flag.

akuegel (approved):
Approving to get the latest state, thanks for adding the guard. That makes it easier for me to run the benchmarks :)

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +143,60 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
+  const int max_bits_for_aggressive_unrolling =
+      max_vector_bit_width / kDefaultUnrollFactor;
   if (analysis->device_info().cuda_compute_capability().IsBlackwell() &&
-      analysis->emitter_fusion_kind() ==
-          HloFusionAnalysis::EmitterFusionKind::kLoop &&
-      analysis->input_output_info().smallest_output_dtype_bits <
-          kMaxBitsToVectorizeWithVectorSize4 &&
+      (analysis->emitter_fusion_kind() ==
+           HloFusionAnalysis::EmitterFusionKind::kLoop ||
+       analysis->emitter_fusion_kind() ==
```
akuegel: This causes a regression for this fusion:

```
HloModule m

ENTRY main {
  param_0.11291 = bf16[16,8512,16]{2,1,0} parameter(0)
  param_1.9571 = bf16[16,1,8512]{2,0,1} parameter(1)
  reshape.2319.5 = bf16[16,8512]{1,0} reshape(param_1.9571)
  broadcast.36384.5 = bf16[16,8512,16]{2,1,0} broadcast(reshape.2319.5), dimensions={0,1}
  multiply.10892.3 = bf16[16,8512,16]{2,1,0} multiply(param_0.11291, broadcast.36384.5)
  reshape.16364.1 = bf16[1,16,8512,16]{3,2,1,0} reshape(multiply.10892.3)
  ROOT transpose.7316.1 = bf16[1,16,16,8512]{3,2,1,0} transpose(reshape.16364.1), dimensions={0,1,3,2}
}
```

I had noticed this already when I worked on the heuristic, but didn't want to spend too much time to deep dive into it because we do already reroute fusions that contain transposes to triton, which often helps. I suggest to keep this change out of the PR and possibly investigate/improve the heuristic for transpose fusions separately.

xla/codegen/emitters/transforms/vectorize_loads_stores.cc
```diff
@@ -151,11 +151,15 @@ mlir::VectorType GetVectorType(mlir::RankedTensorType tensor_type,
       mlir::getConstantIntValue(loop.getLowerBound()) != 0) {
     return nullptr;
   }
-  std::optional<int> vector_size =
+  std::optional<unsigned int> vector_size =
```
akuegel: While fixing this, please change to `std::optional<int64_t>` to match what the called method returns.

xla/codegen/emitters/transforms/vectorize_loads_stores.cc
```diff
@@ -151,11 +151,15 @@ mlir::VectorType GetVectorType(mlir::RankedTensorType tensor_type,
...
       mlir::getConstantIntValue(loop.getUpperBound());
-  if (vector_size != 2 && vector_size != 4 && vector_size != 8) {
+  if (vector_size < 2 || vector_size > 32 ||
+      !absl::has_single_bit(*vector_size)) {
```
akuegel: This requires the header include #include "third_party/absl/numeric/bits.h" + corresponding dependency in the BUILD file

xla/codegen/emitters/transforms/vectorize_loads_stores.cc
```diff
@@ -151,11 +151,15 @@ mlir::VectorType GetVectorType(mlir::RankedTensorType tensor_type,
...
     return nullptr;  // Unsupported vector size.
   }
+  if (tensor_type.getElementTypeBitWidth() * *vector_size > 256) {
```
akuegel, line 161: Isn't the 256 here also dependent on the GPU and which PTX version we have?

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +143,60 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
+    for (const HloInstructionAdaptor& root : analysis->fusion_roots()) {
+      max_dtype_bits = std::max(
+          max_dtype_bits, static_cast<int64_t>(primitive_util::StorageBitWidth(
+                              root.instruction().shape().element_type())));
+    }
+    // Do not unroll sub-byte types further for now.
+    max_dtype_bits = std::max(int64_t{8}, max_dtype_bits);
+
+    int64_t unroll_factor =
+        std::min(int64_t(analysis->fusion_root(0)
```
akuegel: Please use static_cast here

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +143,60 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
+                 max_vector_bit_width / max_dtype_bits);
+    unroll_factor =
+        1LL << (absl::bit_width(static_cast<uint64_t>(unroll_factor)) - 1);
+    while (unroll_factor > kDefaultUnrollFactor) {
+      if (!ContainsTransposeWithSmallMostMinorDim(analysis->fusion(),
+                                                  unroll_factor)) {
+        return unroll_factor;
+      }
+      unroll_factor /= 2;
+    }
```
akuegel, line 199: Missing here: `return unroll_factor`.

I think this explains why with the example I gave you above you still got vectorization by 4.

akuegel (commented):
Sorry for taking so long to reply. I was trying to check whether performance is ok, and only very late noticed that in fact there is a missing return which essentially changes that we will always return the default unroll factor. Once fixing that, I will need to recheck everything. But maybe you can check on your side, first?

xla/service/gpu/gpu_fusible.cc
sergachev: I don't think so. The loop exits either when ContainsTransposeWithSmallMostMinorDim becomes false for the current value (and it gets returned) or the unroll factor gets down to == kDefaultUnrollFactor for which case the further return at line 199 is fine.

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +143,60 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
+          max_dtype_bits, static_cast<int64_t>(primitive_util::StorageBitWidth(
+                              param->shape().element_type())));
+    }
+    for (const HloInstructionAdaptor& root : analysis->fusion_roots()) {
+      max_dtype_bits = std::max(
+          max_dtype_bits, static_cast<int64_t>(primitive_util::StorageBitWidth(
+                              root.instruction().shape().element_type())));
+    }
+    // Do not unroll sub-byte types further for now.
+    max_dtype_bits = std::max(int64_t{8}, max_dtype_bits);
```
akuegel: For simplicity, this could be moved above where max_dtype_bits is initialized (instead of initializing it to 0).

xla/service/gpu/gpu_fusible.cc
akuegel: Ah, good point. Maybe add a break and return to make it easier to read?

So then the question is, why do we get vectorization size 4 for the example I gave above? I think we do return 8 from here now (if I can trust the debugging I added). I want to avoid that we rely on something that we do not directly control (e.g. LLVM vectorization).

xla/codegen/emitters/transforms/vectorize_loads_stores.cc
sergachev: Generating a 256b vector here does not require PTX support for 256b vectors, it gets split into smaller ones if necessary.

0_demo.mlir: tensor<8xf32>

`emitters_opt 0_demo.mlir -xla-vectorize-loads-stores -xla-gpu-test-to-llvm -cse -canonicalize > 1_demo_llvm_dialect.mlir`

1_demo_llvm_dialect.mlir: vector<8xf32>

`mlir-translate --mlir-to-llvmir 1_demo_llvm_dialect.mlir > 2_demo.ll`

2_demo.ll: load <8 x float>

`llc 2_demo.ll -march=nvptx64 -mcpu=sm_90a -o 3_demo.ptx`

3_demo.ptx: ld.v4.f32 + ld.v4.f32

[3_demo.ptx.txt](https://github.com/user-attachments/files/25720603/3_demo.ptx.txt)
[2_demo.ll.txt](https://github.com/user-attachments/files/25720606/2_demo.ll.txt)
[1_demo_llvm_dialect.mlir.txt](https://github.com/user-attachments/files/25720607/1_demo_llvm_dialect.mlir.txt)
[0_demo.mlir.txt](https://github.com/user-attachments/files/25720608/0_demo.mlir.txt)

xla/service/gpu/gpu_fusible.cc
sergachev: In your example the fusion has 3 outputs which limits it to use kDefaultUnrollFactor = 4 - the check for the number of outputs stays unchanged: https://github.com/openxla/xla/pull/38174/changes#diff-57d9a17e200887e87b43997953e3946b887978492fcc0972bf078d0a869ca844R163

akuegel: Thanks for tracking this down, that makes sense :)

xla/codegen/emitters/transforms/vectorize_loads_stores.cc
akuegel: True, it gets split, but likely it is not as efficient then. Anyway, I guess it doesn't matter so much.

akuegel (approved):

xla/service/gpu/gpu_fusible.cc
```diff
@@ -50,6 +51,7 @@ limitations under the License.
 #include "xla/shape_util.h"
 #include "xla/side_effect_util.h"
 #include "xla/stream_executor/device_description.h"
+#include "xla/stream_executor/semantic_version.h"
```
akuegel, line 54: This needs a corresponding BUILD dependency.

xla/xla.proto
```diff
@@ -1044,6 +1038,12 @@ message DebugOptions {
   // the results of regular emitters.
   optional bool xla_gpu_verify_triton_fusion_numerics = 291;
 
+  // Controls max unroll factor on Blackwell architectures. This is also
+  // guarded with a heuristic, but the heuristic is not perfect, so changing
+  // this flag can cause both performance improvements and performance
+  // regressions.
+  optional int32 xla_gpu_experimental_max_unroll_factor = 458;
```
akuegel: Please move it to the right place in lexicographic order inside the XLA:GPU options section. See go/keep-sorted start which makes the linter complain when this is not sorted.

akuegel:
I found a fusion that seems to regress quite a bit, here the HLO before optimizations (that will form one fusion later):

```
HloModule m

ENTRY main {
  param_0.679 = pred[16,156]{1,0} parameter(0)
  broadcast.119.1 = pred[1,16,156,156]{3,2,1,0} broadcast(param_0.679), dimensions={1,2}
  broadcast.120.3 = pred[1,16,156,156]{3,2,1,0} broadcast(param_0.679), dimensions={1,3}
  and.14.3 = pred[1,16,156,156]{3,2,1,0} and(broadcast.119.1, broadcast.120.3)
  constant_48_2 = pred[] constant(false)
  pad.27.1 = pred[1,512,156,156]{3,2,1,0} pad(and.14.3, constant_48_2), padding=0_0x0_496x0_0x0_0
  reshape.32.1 = pred[512,156,156]{2,1,0} reshape(pad.27.1)
  ROOT broadcast_in_dim.42.1 = pred[512,8,156,156]{3,2,1,0} broadcast(reshape.32.1), dimensions={0,2,3}
}
```

There are other benchmarks that might have regressions that I still need to check. This is from the one that regressed the most, as far as I can tell mainly due to this fusion. On B200, we go from 1.87 TB/s to 210 GB/s

One thing that might be worth checking is whether it is related to the Pad op. Often it blocks vectorization of the loads due to the if construct that our emitter does.

sergachev:
Thank you for the input! I've reproduced it and found a trivial fix which makes the fusion ~1.5x faster instead of regressing it, will test it now on the benchmarks I have.

sergachev:
I made the OptimizeLoops pass use the unroll factor calculated through ComputeLoopFusionConfig(MaxUnrollFactor(HloFusionAnalysis)) instead of the default constant returned by MaxUnrollFactor(nullptr) and updated the benchmark results.

akuegel:
> I made the OptimizeLoops pass use the unroll factor calculated through ComputeLoopFusionConfig(MaxUnrollFactor(HloFusionAnalysis)) instead of the default constant returned by MaxUnrollFactor(nullptr) and updated the benchmark results.

This sound like it might have been a pre-existing problem as well, thanks for the fix. I will start another benchmark run on our side.

akuegel (approved):

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +143,58 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
-          }) &&
-      !ContainsTransposeWithSmallMostMinorDim(analysis->fusion(), 8)) {
-    return 8;
+      !HloAnyOf(analysis->fusion(), [](HloInstructionAdaptor node) {
+        return node.opcode() == HloOpcode::kReduce;
+      })) {
+    // Do not unroll sub-byte types further for now.
+    int64_t max_dtype_bits = 8;
+    for (const HloInstruction* param : analysis->fusion().GetParameters()) {
+      max_dtype_bits = std::max(
```
akuegel, line 175: Can you skip scalar parameters here? I found a fusion where we got a regression because currently we unroll to vector size 8, but now we don't because it has a scalar f32[] parameter (all others and the single output is bf16).
Edit: forgot to mention, this happens with old cuda 12.8.1. With cuda 12.9 or above, I guess we would still unroll to vector size 8, but likely should unroll to vector size 16.

akuegel (commented):
Benchmarks in general look relatively good, but I found at least one regression.

xla/service/gpu/BUILD
```diff
@@ -2997,6 +2995,7 @@ cc_library(
         ":ir_emission_utils",
         ":launch_dimensions",
         ":reduction_utils",
+        "//xla:comparison_util",
```
akuegel: This dependency is not actually needed, we get the header through shape_util already.

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +143,61 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
-          }) &&
-      !ContainsTransposeWithSmallMostMinorDim(analysis->fusion(), 8)) {
-    return 8;
+      !HloAnyOf(analysis->fusion(), [](HloInstructionAdaptor node) {
+        return node.opcode() == HloOpcode::kReduce;
+      })) {
+    // Do not unroll sub-byte types further for now.
+    int64_t max_dtype_bits = 8;
+    for (const HloInstruction* param : analysis->fusion().GetParameters()) {
+      if (ShapeUtil::IsScalar(param->shape())) {
```
akuegel, line 172: So while it certainly makes sense to skip scalars, we seem to have also cases where we have parameters that have fewer elements than the maximum number of elements of a parameter, and there was one fusion where the somewhat smaller parameters had type f32 and the parameter with maximum number of elements was s8. In this case, unrolling by factor 8 turned out to be better than 4.

So getting this tradeoff right (which parameters to ignore, which do count) seems tricky again. If we want to avoid regressions, it might be easier to ensure that we always return at least unroll factor 8 if we enter this if block for Blackwell. I changed line 198 to return std::max(unroll_factor, int64_t{8}), and benchmarks looked quite good.

But maybe you also have another idea? Maybe we shouldn't look at maximum bit widths in outputs/parameters but instead for minimum? I guess this would err on the side of causing too much register pressure?

@pifon2a suggested that maybe we should just autotune unroll factor because it seems to be so tricky to get a heuristic working.

xla/service/gpu/gpu_fusible.cc
```diff
@@ -140,30 +143,61 @@ int64_t MaxUnrollFactor(const HloFusionAnalysis* analysis) {
...
   if (analysis->device_info().cuda_compute_capability().IsBlackwell() &&
-      analysis->emitter_fusion_kind() ==
-          HloFusionAnalysis::EmitterFusionKind::kLoop &&
-      analysis->input_output_info().smallest_output_dtype_bits <
-          kMaxBitsToVectorizeWithVectorSize4 &&
+      (analysis->emitter_fusion_kind() ==
+           HloFusionAnalysis::EmitterFusionKind::kLoop ||
+       analysis->emitter_fusion_kind() ==
+           HloFusionAnalysis::EmitterFusionKind::kTranspose) &&
+      analysis->input_output_info().smallest_output_dtype_bits <=
```
akuegel, line 163: This was a `<` instead of `<=` before. Why change this? The idea before was we enter this block if we expect to use unroll factor 8. In combination to my other suggested change to return at least 8 in this code block, we should also change this back to `<`.

sergachev: I read this several more times - you are right.

xla/service/gpu/gpu_fusible.cc
sergachev: I agree the heuristic is still far from ideal. I just got another example where one parameter is read multiple time through slices and this currently isn't analyzed properly.

Agree that autotuning is ultimately better for performance. I think we have enough microbenchmark data showing that getting the unroll factor right is important to justify at least trying it.

Shall we still try to land this change with the improvements you suggested?

akuegel: Latest benchmark run on our side looks good, geomean improvement of 1% for device time excluding collectives, and the regressions seem to be noise (we compared two runs, and the regressions in one run are not a regression in the other run). So I think we can land it now with the two latest changes I suggested.
Regarding autotuning, I think it is good if we have a heuristic that already proposes a somewhat good unroll factor. Then we might get away with just comparing with unroll / 2 and unroll * 2 when autotuning. So thanks for your work on improving the heuristic.

PR #38224, opened 2026-02-24
[ROCm] Use VLOG(1) for tracer teardown logs

AratiGanesh:
📝 Summary of Changes
The ROCm tracer was using LOG(INFO) in toolFinalize(). Those run during tracer teardown, so the messages were always printed and could still appear after logging was turned off. That broke JAX  tests that expect no log output once logging is disabled i.e tests/logging_test.py::LoggingTest::test_subprocess_toggling_logging_level. The teardown message were switched from LOG(INFO) to VLOG(1)

🚀 Kind of Contribution

image

### DIFF
--- a/xla/backends/profiler/gpu/rocm_tracer.cc
+++ b/xla/backends/profiler/gpu/rocm_tracer.cc
@@ -490,7 +490,7 @@ int RocmTracer::toolInit(rocprofiler_client_finalize_t fini_func,
 
 void RocmTracer::toolFinalize(void* tool_data) {
   auto& obj = RocmTracer::GetRocmTracerSingleton();
-  LOG(INFO) << "Calling toolFinalize!";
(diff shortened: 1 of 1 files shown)


cj401-amd (commented):
Hey @Aratiganesh123, should we replace all LOG(INFO) with VLOG(1) as LOG(INFO) might cause a JAX test failure? https://github.com/ROCm/xla/pull/442/changes

AratiGanesh:
Hi @cj401-amd 
I've changed only the function that actually run after logging is turned off in that JAX test - toolFinalize (“Calling toolFinalize!”)  to VLOG(1). The other LOG(INFO) calls run during init before the JAX test turns off logging, so they don’t affect this test. I can switch the rest to VLOG(1) as well for consistency if you prefer.

dimitar-asenov (approved):

PR #38255, opened 2026-02-25
DRAFT: [xla:gpu] Add support for GPU coredumps on detected hangs

ezhulenev:
When `HangWatchdog` detects that `GpuExecutable` stuck, add a way to dump the GPU state to a file. For CUDA platform it enables `cuda-dbg` to look at what GPU kernels were running on device at the time of detected deadlock.


ezhulenev:
Closing it for now as I was not able to make it work because of CUDA issues, progress tracking PR achives similar result in a much lighter way.

PR #36158, opened 2026-01-09
[GPU] Add NCCL FP8 support

terryysun:
📝 Summary of Changes
Enable FP8 support.

🎯 Justification
Starting from Blackwell, NCCL supports multimem reduction which makes the FP8 benefit significant. This PR enables NCCL FP8 support in XLA.

🚀 Kind of Contribution
✨ New Feature

📊 Benchmark (for Performance Improvements)
N/A.

🧪 Unit Tests:
N/A.

🧪 Execution Tests:
Added training step execution test.

PR #38269, opened 2026-02-25
[ROCm] Add RCCL FP8 support

mfrancepillois:
📝 Summary of Changes
Enable FP8 support on RCCL.

🎯 Justification
GPU architecture `gfx9_mi350` and subsequent generations (ROCm 7.0+) support FP8 datatype in RCCL.
This PR enables FP8 support in a similar way to what was done in PR https://github.com/openxla/xla/pull/36158 for NCCL.

🚀 Kind of Contribution

🧪 Execution Tests:
Extends the tests added in  https://github.com/openxla/xla/pull/36158 to run on ROCm targets as well.

### DIFF
--- a/xla/backends/gpu/collectives/rccl_communicator.cc
+++ b/xla/backends/gpu/collectives/rccl_communicator.cc
@@ -82,12 +82,15 @@ static size_t ToNcclCount(PrimitiveType dtype, size_t count) {
   return primitive_util::IsComplexType(dtype) ? count * 2 : count;
 }
 
-static absl::StatusOr<ncclDataType_t> ToNcclDataType(PrimitiveType dtype,
-                                                     bool is_reduction_op) {
+static absl::StatusOr<ncclDataType_t> ToNcclDataType(
+    PrimitiveType dtype, bool is_reduction_op,
+    se::RocmComputeCapability rocm_cc) {
   switch (dtype) {
     case S8:
     case F8E5M2:
+      return rocm_cc.has_ocp_fp8_support() ? ncclFloat8e5m2 : ncclInt8;
     case F8E4M3FN:
+      return rocm_cc.has_ocp_fp8_support() ? ncclFloat8e4m3 : ncclInt8;
     case F8E5M2FNUZ:
     case F8E4M3FNUZ:
     case F8E8M0FNU:
@@ -554,7 +557,11 @@ absl::Status RcclCommunicator::LaunchAllReduce(
       recv_buffer.opaque(), primitive_util::LowercasePrimitiveTypeName(dtype),
       count, reduction_kind, comm_, stream);
 
-  TF_ASSIGN_OR_RETURN(ncclDataType_t nccl_dtype, ToNcclDataType(dtype, false));
+  TF_ASSIGN_OR_RETURN(
+      ncclDataType_t nccl_dtype,
+      ToNcclDataType(
+          dtype, false,
+          stream->parent()->GetDeviceDescription().rocm_compute_capability()));
 
   TF_RETURN_IF_ERROR(XLA_RCCL_STATUS(ncclAllReduce(
       send_buffer.opaque(), recv_buffer.opaque(), ToNcclCount(dtype, count),
@@ -582,7 +589,11 @@ absl::Status RcclCommunicator::LaunchBroadcast(
       recv_buffer.opaque(), primitive_util::LowercasePrimitiveTypeName(dtype),
       count, root.value(), comm_, stream);
 
-  TF_ASSIGN_OR_RETURN(ncclDataType_t nccl_dtype, ToNcclDataType(dtype, false));
+  TF_ASSIGN_OR_RETURN(
+      ncclDataType_t nccl_dtype,
+      ToNcclDataType(
+          dtype, false,
+          stream->parent()->GetDeviceDescription().rocm_compute_capability()));
 
   TF_RETURN_IF_ERROR(XLA_RCCL_STATUS(ncclBroadcast(
       send_buffer.opaque(), recv_buffer.opaque(), ToNcclCount(dtype, count),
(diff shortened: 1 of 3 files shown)


xla/backends/gpu/collectives/rccl_communicator.cc
```diff
@@ -82,12 +82,15 @@ static size_t ToNcclCount(PrimitiveType dtype, size_t count) {
...
+static absl::StatusOr<ncclDataType_t> ToNcclDataType(
+    PrimitiveType dtype, bool is_reduction_op,
+    se::RocmComputeCapability rocm_cc) {
   switch (dtype) {
     case S8:
     case F8E5M2:
+      return rocm_cc.has_ocp_fp8_support() ? ncclFloat8e5m2 : ncclInt8;
     case F8E4M3FN:
+      return rocm_cc.has_ocp_fp8_support() ? ncclFloat8e4m3 : ncclInt8;
     case F8E5M2FNUZ:
```
draganmladjenovic, line 94: What about FNUZ types?

xla/tests/all_reduce_e2e_test.cc
```diff
@@ -91,9 +91,18 @@ class AllReduceTestNoParams : public CollectiveOpsWithFlagsBase {
...
     CollectiveOpsE2ETestBase::SetUp();
-    if (!IsAmpereAndHigher()) {
-      GTEST_SKIP() << "Test requires Ampere or newer architecture since it's "
-                      "using triton.";
+    // Check for Triton support: Ampere+ for CUDA, MI300+ for ROCm
+    bool has_triton_support = false;
+    if (Capability().IsCuda()) {
+      has_triton_support = IsAmpereAndHigher();
+    } else if (Capability().IsRocm()) {
+      has_triton_support =
```
draganmladjenovic: I don't think we have such limitation.

xla/backends/gpu/collectives/rccl_communicator.cc
mfrancepillois: I'm not sure if FNUZ types are natively supported  by RCCL/NCCL. The message of the commit introducing RCCL FP8 support mentioned:

> Add fp8 support
          * Add support for e5m2 and e4m3 8-bit floating point operations.
 
And the header file only contains the datatypes `ncclFloat8e4m3` and `ncclFloat8e5m2 `.
https://github.com/ROCm/rccl/blob/release/rocm-rel-7.0/src/nccl.h.in#L417-L418 

I have therefore limited RCCL FP8 to the same types as those handled by NCCL FP8. (https://github.com/terryysun/xla/blob/main/xla/backends/gpu/collectives/nccl_communicator.cc#L99-L104)

xla/tests/all_reduce_e2e_test.cc
mfrancepillois: Thanks for noticing. The code has been updated.

xla/backends/gpu/collectives/rccl_communicator.cc
draganmladjenovic: Can you check it out? Use ncclFloat8e4m3 and ncclFloat8e5m2  for fnuz types and see it they work on gfx942.

mfrancepillois: I tested the FNUZ types on gfx942, but the produced outputs did not pass the correctness check. So, it seems that these datatypes are not (correctly) supported.

dimitar-asenov (approved):

xla/backends/gpu/collectives/rccl_communicator.cc
```diff
@@ -82,12 +82,15 @@ static size_t ToNcclCount(PrimitiveType dtype, size_t count) {
   return primitive_util::IsComplexType(dtype) ? count * 2 : count;
 }
 
-static absl::StatusOr<ncclDataType_t> ToNcclDataType(PrimitiveType dtype,
-                                                     bool is_reduction_op) {
+static absl::StatusOr<ncclDataType_t> ToNcclDataType(
+    PrimitiveType dtype, bool is_reduction_op,
+    se::RocmComputeCapability rocm_cc) {
   switch (dtype) {
     case S8:
```
thcmbs: Is it expected to return ncclFloat8e5m2 for S8 ? Shouldn't this branch to to int8?

xla/backends/gpu/collectives/rccl_communicator.cc
```diff
@@ -637,7 +652,11 @@ absl::Status RcclCommunicator::LaunchAllGather(
       recv_buffer.opaque(), primitive_util::LowercasePrimitiveTypeName(dtype),
       count, comm_, stream);
 
-  TF_ASSIGN_OR_RETURN(ncclDataType_t nccl_dtype, ToNcclDataType(dtype, false));
+  TF_ASSIGN_OR_RETURN(
+      ncclDataType_t nccl_dtype,
+      ToNcclDataType(
+          dtype, false,
```
thcmbs: fme: why do we pass false here? Isn't this a reduction op?

xla/backends/gpu/collectives/rccl_communicator.cc
```diff
@@ -554,7 +557,11 @@ absl::Status RcclCommunicator::LaunchAllReduce(
       recv_buffer.opaque(), primitive_util::LowercasePrimitiveTypeName(dtype),
       count, reduction_kind, comm_, stream);
 
-  TF_ASSIGN_OR_RETURN(ncclDataType_t nccl_dtype, ToNcclDataType(dtype, false));
+  TF_ASSIGN_OR_RETURN(
+      ncclDataType_t nccl_dtype,
+      ToNcclDataType(
+          dtype, false,
```
thcmbs: fme: why do we pass false here? Isn't this a reduction op?
+ nit: /*is_reduction_op=*/true/false would help readability.

thcmbs (commented):
Hi, thanks for the contribution. I'm responsible for merging the PR and had a few QQ. Apologies if these are obvious! Wanted to double check before merging.

xla/backends/gpu/collectives/rccl_communicator.cc
mfrancepillois: Nice catch! The code has been fixed. Thanks

thcmbs: Thanks!

thcmbs (approved):

xla/backends/gpu/collectives/rccl_communicator.cc
mfrancepillois: The file change history shows that this parameter has only been set since this commit https://github.com/openxla/xla/commit/e82d31fc0070e09dbda0f24cdfed60e4c36a9617 to let NCCL to throw the error. In our case, we do not use it in the same way, as invalid types are not caught by RCCL. That's why I kept this parameter as it was in the initial PR.
But I agree that it is reduction op, so theoretically this parameter should be set to true. I therefore updated the code (it doesn't really have any effect, as U16 are indeed promoted to U32 anyway).

xla/backends/gpu/collectives/rccl_communicator.cc
mfrancepillois: Same as above

beckerhe (approved):

PR #38415, opened 2026-02-27
[xla:gpu] Add thunk-free library for supporting while loops at run time

ezhulenev:
Refactor tracking of the while loop iterations and move it outside of `WhileThunk` in preparation for reuse in command buffer implementation. Also add `while_loop.h` documentation that explains the intended use case (to be implemented in followup PRs).

### DIFF
--- a/xla/backends/gpu/runtime/while_loop.h
+++ b/xla/backends/gpu/runtime/while_loop.h
@@ -1,0 +1,110 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef XLA_BACKENDS_GPU_RUNTIME_WHILE_LOOP_H_
+#define XLA_BACKENDS_GPU_RUNTIME_WHILE_LOOP_H_
+
+#include <cstddef>
+#include <optional>
+#include <string>
+
+#include "absl/strings/string_view.h"
+
+namespace xla::gpu {
+
+// A library that supports running while loops in the XLA:GPU runtime.
+//
+// See StableHLO documentation for the while loop semantics:
+// https://openxla.org/stablehlo/spec#while
+//
+// This library supports both Thunks and Commands. When a while loop executes
+// thunks (via the WhileThunk), some of the nested thunks might depend on the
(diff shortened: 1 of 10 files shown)


xla/backends/gpu/runtime/while_loop.h
```diff
@@ -0,0 +1,105 @@
...
+// Example:
+//
+//   1. WhileThunk(cond, body); name="while.0", trip_count=2
+//
+// Thunk sequence after unrolling:
+//
+//   1. WhileEnterThunk: name="while.0", trip_count=2
+//   2. SequentialThunk(cond)
+//   3. SequentialThunk(body)
+//   4. WhileIncrementThunk
```
thomasjoerg, line 60: I've seen the increment happening in the while loop body. Would this still happen but be ignored?

ezhulenev: What do you mean? AFAIK loop iteration is not available to XLA computation on device, only whatever cond computes is present as device buffer. For nested while loops it will be multiple `WhileEnterThunk` and each inc will increment the inner-most counter.

thomasjoerg: From what I've seen, the `WhileLoopTripCountAnnotator` HLO pass infers `trip_count` and induction variable from HLO while loops (using pattern matching and other techniques). In that sense, the induction variable / loop iteration is available to the device and fed into the `cond`. At least, the tests of `WhileLoopTripCountAnnotator` show that pattern.
I was just wondering if the induction variable increment and the `cond` checking the loop bound would be dead code on the device and should be elided. It's maybe out of scope for this PR, but something I wondered about before, hence my question.

thomasjoerg (approved):

xla/backends/gpu/runtime/while_loop.h
ezhulenev: cond checking is dead code, but induction variable might be used for dynamic slices, so we must always run it. But I'm not sure we can reliable skip cond checking kernel, it might be fused with something useful.

thcmbs:
Thanks for the contribution! Can you please add the header guards for `while_loop.h` ? 
```
#ifndef TENSORFLOW_COMPILER_XLA_BACKENDS_GPU_RUNTIME_WHILE_LOOP_H_
#define TENSORFLOW_COMPILER_XLA_BACKENDS_GPU_RUNTIME_WHILE_LOOP_H_

/* ...  */

#endif  // TENSORFLOW_COMPILER_XLA_BACKENDS_GPU_RUNTIME_WHILE_LOOP_H_
```

xla/backends/gpu/runtime/while_loop.h
thomasjoerg: command_buffer_conversion_pass.cc;l=537-539 seems to enqueue only the while body but, not the condition, IIUC.
For very small while loop bodies, the overhead of condition checking and induction variable increments can add up. Hence, it'd be nice to avoid it were we can. However, replacing the induction variable with constants seems feasible only in HLO-level unrolling and constant folding. Do you agree?

ezhulenev:
Fixed + rebased.

PR #38473, opened 2026-03-01
[xla:gpu] Extract ThunkExecutor out of SequentialThunk

ezhulenev:
`SequentialThunk` has a special meaning in XLA:GPU runtime, and it's confusing as it mixes the "thunk AST" (graph) structure and `Thunk` execution.

Extract the execution bits into `ThunkExecutor` which makes it consistent with `CommandExecutor` counterpart from commands land, and also consistent with XLA:CPU.

In the follow up PR `SequentialThunk` will loose it special meaning in `GpuExecutable`, as the the `executable` must be a `ThunkExecutor` that corresponds to the top level thunk sequence.

This PR is **NFC**, just extracting code into a separate target and updating all users. Longer term we might want to add more "execution optimizations" to executor, i.e. handling multi-stream execution in one central place (similar to how ThunkExecutor in CPU launches tasks into thread pool), today it's spread thin across different individual thunks.


pifon2a (approved):

xla/backends/gpu/runtime/thunk_executor.cc
```diff
@@ -0,0 +1,242 @@
...
+  std::vector<ExecutedThunkEntry> entries;
+  entries.reserve(events_->map.size());
+  for (auto& [thunk, progress] : events_->map) {
+    if (progress.executed != absl::InfinitePast()) {
+      entries.push_back({thunk, &progress});
+    }
+  }
+
+  // Sort by executed time before checking event status.
+  absl::c_sort(entries, [most_recent_first](const auto& a, const auto& b) {
```
thcmbs: Hey Eugene, 
I am responsible for merging your PR. ClangTidy complains about non deterministic sorting order here. Granted, timestamps are unlikely to be equal, but should we add a tie breaker like

if (a.progress->executed != b.progress->executed) {
    return most_recent_first ? a.progress->executed > b.progress->executed
                             : a.progress->executed < b.progress->executed;
  }
  // Tie-breaker: use static execution index for determinism.
  return most_recent_first ? a.progress->index > b.progress->index
                           : a.progress->index < b.progress->index;

to guarantee determinis,?

ezhulenev: Changed to `c_stable_sort` as in this case the order doesn't really matter for the caller.

pifon2a (approved):

pifon2a (approved):

xla/backends/gpu/runtime/BUILD
```diff
@@ -2726,27 +2726,20 @@ cc_library(
         ":annotation",
         ":collective_thunk",
```
penpornk: Please remove unused deps:
```
:collective_thunk
//xla:util 
```

penpornk (changes requested):

xla/backends/gpu/runtime/BUILD
ezhulenev: Done.

penpornk (approved):
Thank you!

penpornk (commented):
There are merge conflicts on //xla/service/gpu/BUILD. Please help resolve?

penpornk (approved):
Thank you!

beckerhe (approved):

pifon2a (approved):

PR #38504, opened 2026-03-02
Add StreamExecutorVmmAllocator and VMM interface to StreamExecutor

shawnwang18:
Extract stream_executor_vmm_allocator.h/cc and its test from the update_free_cmd_buffer branch, along with the VMM virtual methods added to StreamExecutor (ReserveAddress, FreeAddress, MapRawAddress, VmmAllocateMemory, VmmDeallocateMemory, etc.) and the corresponding BUILD targets.

📝 Summary of Changes

🎯 Justification
change.

🚀 Kind of Contribution

🧪 Unit Tests:
transformed HLO.

🧪 Execution Tests:

### DIFF
--- a/xla/stream_executor/stream_executor.h
+++ b/xla/stream_executor/stream_executor.h
@@ -59,6 +59,8 @@ limitations under the License.
 
 namespace stream_executor {
 
+TSL_LIB_GTL_DEFINE_INT_TYPE(RawAddressHandle, uint64_t);
+
 /// The StreamExecutor is a single-device abstraction for:
 //
 // * Loading/launching data-parallel-kernels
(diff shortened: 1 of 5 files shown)


xla/stream_executor/stream_executor.h
```diff
@@ -59,6 +59,8 @@ limitations under the License.
 
 namespace stream_executor {
 
+TSL_LIB_GTL_DEFINE_INT_TYPE(RawAddressHandle, uint64_t);
```
ezhulenev, line 62: Stream executor has https://github.com/openxla/xla/blob/main/xla/stream_executor/memory_allocation.h, instead of adding integer-typed handles let's use the memory allocation abstraction to encapsulate physical memory allocations.

xla/stream_executor/stream_executor.h
```diff
@@ -175,6 +177,63 @@ class StreamExecutor {
   // Deallocation of a nullptr-representative value is permitted.
   virtual void Deallocate(DeviceAddressBase* mem) = 0;
 
+  // Reserves size bytes on the underlying platform and returns a
+  // DeviceAddressBase representing that reservation.
+  virtual absl::StatusOr<DeviceAddressBase> ReserveAddress(
```
ezhulenev, line 182: Instead of abusing `DeviceAddressBase` let's distinguish between addessable devices memory and virtual. 

I'd recommend adding `MemoryReservation` (memory_reservation.h) to SE APIs, and document it as a block of virtual addressable memory reservcation.

Then the physical `MemoryAllocation` can be mapped to `MemoryReservation`. `MemoryAllocation` already has an `address()` API that returns an optional address.

In general I want `se::DeviceAddess(Base)` to be always an address that can be dereferenced one device.

Let's do something like this:

```
class MemoryReservation {
  virtual ~MemoryReservation();
  // Maps a memory reservation to a physical allocation and returns a device address that
  // is accessible on device.  On destructing `ScopedDeviceAddress` it unmaps the memory
  // reservation from the physical allocation.
  absl::StatusOr<ScopedDeviceAddress<T>> MapTo(MemoryAllocation* allocation);
};
```

`MemoryReservation` itself should not give `se::DeviceMemory` address to anyone who needs it, it should use RAII to proeprly manager lifetime and mapped-to relationship. In general XLA and SE should use more RAII concepts and less Allocate/Free/Map/Unmap/etc. APIs.

@PatriosTheGreat @beckerhe @pifon2a

PR #38505, opened 2026-03-02
[Autotuner] Provide a better error message when no configs are found

sergey-kozub:
📝 Summary of Changes
Improves error logging (do not discard error messages).

🎯 Justification
Currently, if the autotuner fails due to OOM (for example), it's not easy to see that from the error logs.
This PR adds autotuner error messages to the logs.

### DIFF
--- a/xla/backends/autotuner/autotuner.cc
+++ b/xla/backends/autotuner/autotuner.cc
@@ -618,15 +618,23 @@ absl::StatusOr<Autotuner::ConfigResult> Autotuner::PickBestConfig(
     std::vector<ConfigResult>& results) {
   absl::Duration min_duration = absl::InfiniteDuration();
   ConfigResult* best_result = nullptr;
+  std::vector<std::string> failures;
   for (ConfigResult& result : results) {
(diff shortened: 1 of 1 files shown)


derdrdirk (approved):
LGTM

xla/backends/autotuner/autotuner.cc
```diff
@@ -618,15 +618,23 @@ absl::StatusOr<Autotuner::ConfigResult> Autotuner::PickBestConfig(
...
       min_duration = result.duration;
       best_result = &result;
     }
   }
 
   if (best_result == nullptr) {
-    return absl::NotFoundError("No valid config found!");
+    std::string message = "No valid config found!";
+    if (!failures.empty()) {
+      message += absl::StrCat(" Failures: ", failures.size(), "\n");
```
thcmbs: Please use StrAppend instead of StrCat when appending to an existing string

Thanks!

sergey-kozub: Updated.

PR #38570, opened 2026-03-03
[XLA:GPU] Add CUDA VMM RAII wrappers: CudaRawMemoryAllocation and CudaMemoryReservation

shawnwang18:
Add CUDA Virtual Memory Management RAII wrappers

  Introduces building blocks for CUDA VMM (Virtual Memory Management API):

  - MemoryReservation — new abstract base class in stream_executor representing a reserved virtual address range, with MapToAllocation, SetAccess, and UnMap as the core interface.
  - CudaRawMemoryAllocation — RAII wrapper for a physical memory handle created via cuMemCreate. The handle is not yet mapped to any virtual address; lifetime is managed automatically.
  - CudaMemoryReservation — RAII wrapper for a virtual address range reserved via cuMemAddressReserve. Supports mapping a CudaRawMemoryAllocation into a sub-range, enabling device access via cuMemSetAccess, and explicit unmapping. The destructor handles cleanup automatically.

  All classes follow the CudaEvent design pattern (static Create factory, move semantics, context activation via StreamExecutor). Unit tests are included for both CUDA classes covering creation, address invariants, move semantics, and the full map+access workflow.

### DIFF
--- a/xla/stream_executor/cuda/cuda_memory_reservation.cc
+++ b/xla/stream_executor/cuda/cuda_memory_reservation.cc
@@ -1,0 +1,132 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/stream_executor/cuda/cuda_memory_reservation.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+
+#include "absl/log/log.h"
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "third_party/gpus/cuda/include/cuda.h"
+#include "xla/stream_executor/activate_context.h"
+#include "xla/stream_executor/cuda/cuda_raw_memory_allocation.h"
+#include "xla/stream_executor/cuda/cuda_status.h"
+#include "xla/stream_executor/device_address.h"
+#include "xla/stream_executor/memory_allocation.h"
+#include "xla/stream_executor/stream_executor.h"
+#include "xla/util.h"
+#include "tsl/platform/statusor.h"
+
+namespace stream_executor::gpu {
+namespace {
+
+CUmemAllocationProp BuildAllocationProperties(CUdevice device) {
+  CUmemAllocationProp props = {};
+  props.type = CU_MEM_ALLOCATION_TYPE_PINNED;
+  props.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  props.location.id = device;
+  props.requestedHandleTypes =
+      static_cast<CUmemAllocationHandleType>(CU_MEM_HANDLE_TYPE_NONE);
+  return props;
+}
+
+}  // namespace
+
+absl::StatusOr<std::unique_ptr<CudaMemoryReservation>>
+CudaMemoryReservation::Create(StreamExecutor* executor, uint64_t size) {
+  std::unique_ptr<ActivateContext> activation = executor->Activate();
+
+  CUdevice device;
+  TF_RETURN_IF_ERROR(
+      cuda::ToStatus(cuDeviceGet(&device, executor->device_ordinal())));
+
+  CUmemAllocationProp props = BuildAllocationProperties(device);
+
+  size_t granularity = 0;
+  TF_RETURN_IF_ERROR(cuda::ToStatus(cuMemGetAllocationGranularity(
+      &granularity, &props, CU_MEM_ALLOC_GRANULARITY_RECOMMENDED)));
+
+  uint64_t padded_size = xla::RoundUpTo<uint64_t>(size, granularity);
+
+  CUdeviceptr ptr;
+  TF_RETURN_IF_ERROR(cuda::ToStatus(
+      cuMemAddressReserve(&ptr, padded_size, granularity, 0, 0)));
+
+  return std::unique_ptr<CudaMemoryReservation>(
+      new CudaMemoryReservation(executor, ptr, padded_size));
+}
+
+CudaMemoryReservation::CudaMemoryReservation(StreamExecutor* executor,
+                                             CUdeviceptr ptr, uint64_t size)
+    : executor_(executor), ptr_(ptr), size_(size) {}
+
+DeviceAddressBase CudaMemoryReservation::address() const {
+  return DeviceAddressBase(reinterpret_cast<void*>(ptr_), size_);
+}
+
+absl::Status CudaMemoryReservation::Map(size_t reservation_offset,
+                                        size_t allocation_offset, size_t size,
+                                        MemoryAllocation& allocation) {
+  auto* cuda_alloc = dynamic_cast<CudaRawMemoryAllocation*>(&allocation);
+  if (cuda_alloc == nullptr) {
+    return absl::InvalidArgumentError(
+        "CudaMemoryReservation::Map requires a CudaRawMemoryAllocation");
+  }
+  std::unique_ptr<ActivateContext> activation = executor_->Activate();
+  return cuda::ToStatus(cuMemMap(ptr_ + reservation_offset, size,
+                                 allocation_offset, cuda_alloc->GetHandle(),
+                                 0));
+}
+
+absl::Status CudaMemoryReservation::SetAccess(uint64_t reservation_offset,
+                                              size_t size) {
+  std::unique_ptr<ActivateContext> activation = executor_->Activate();
+  CUmemAccessDesc desc = {};
+  desc.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  desc.location.id = static_cast<int>(executor_->device_ordinal());
+  desc.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+  return cuda::ToStatus(
+      cuMemSetAccess(ptr_ + reservation_offset, size, &desc, 1));
+}
+
+absl::Status CudaMemoryReservation::UnMap(size_t offset, size_t size) {
+  std::unique_ptr<ActivateContext> activation = executor_->Activate();
+  return cuda::ToStatus(cuMemUnmap(ptr_ + offset, size));
+}
+
+CudaMemoryReservation::~CudaMemoryReservation() {
+  if (ptr_ == 0) {
+    return;
+  }
+  std::unique_ptr<ActivateContext> activation = executor_->Activate();
+  // Attempt to unmap the full range before freeing the virtual address space.
+  // Sub-ranges already unmapped by ScopedMapping destructors will cause this
+  // call to fail; the error is logged and the address range is freed anyway.
+  auto unmap_status =
+      cuda::ToStatus(cuMemUnmap(ptr_, size_), "Error unmapping CUDA memory");
+  if (!unmap_status.ok()) {
+    LOG(ERROR) << unmap_status.message();
+  }
+  auto free_status = cuda::ToStatus(cuMemAddressFree(ptr_, size_),
+                                    "Error freeing CUDA address range");
+  if (!free_status.ok()) {
+    LOG(ERROR) << free_status.message();
+  }
+}
+
+}  // namespace stream_executor::gpu
--- a/xla/stream_executor/cuda/cuda_raw_memory_allocation.h
+++ b/xla/stream_executor/cuda/cuda_raw_memory_allocation.h
@@ -1,0 +1,62 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef XLA_STREAM_EXECUTOR_CUDA_CUDA_RAW_MEMORY_ALLOCATION_H_
+#define XLA_STREAM_EXECUTOR_CUDA_CUDA_RAW_MEMORY_ALLOCATION_H_
+
+#include <cstdint>
+
+#include "absl/status/statusor.h"
+#include "third_party/gpus/cuda/include/cuda.h"
+#include "xla/stream_executor/device_address.h"
+#include "xla/stream_executor/memory_allocation.h"
+#include "xla/stream_executor/stream_executor.h"
+
+namespace stream_executor::gpu {
+
+// RAII wrapper for a physical memory allocation created via cuMemCreate.
+// The handle is not mapped to any virtual address; this class only manages
+// the lifetime of the CUmemGenericAllocationHandle.
+class CudaRawMemoryAllocation : public MemoryAllocation {
+ public:
+  // Creates a physical memory allocation of at least `size` bytes using
+  // cuMemCreate. StreamExecutor is used only for context activation.
+  static absl::StatusOr<std::unique_ptr<CudaRawMemoryAllocation>> Create(
+      StreamExecutor* executor, uint64_t size);
+
+  // Returns a DeviceAddressBase whose opaque() holds the raw
+  // CUmemGenericAllocationHandle cast to void*, and size() is the
+  // padded allocation size.
+  DeviceAddressBase address() const override;
+
+  CUmemGenericAllocationHandle GetHandle() const { return handle_; }
+
+  ~CudaRawMemoryAllocation() override;
+  CudaRawMemoryAllocation(const CudaRawMemoryAllocation&) = delete;
+  CudaRawMemoryAllocation& operator=(const CudaRawMemoryAllocation&) = delete;
+
+ private:
+  explicit CudaRawMemoryAllocation(StreamExecutor* executor,
+                                   CUmemGenericAllocationHandle handle,
+                                   uint64_t size);
+
+  StreamExecutor* executor_;
+  CUmemGenericAllocationHandle handle_;  // 0 means moved-from / released
+  uint64_t size_;
+};
+
+}  // namespace stream_executor::gpu
+
+#endif  // XLA_STREAM_EXECUTOR_CUDA_CUDA_RAW_MEMORY_ALLOCATION_H_
--- a/xla/stream_executor/memory_reservation.cc
+++ b/xla/stream_executor/memory_reservation.cc
@@ -1,0 +1,146 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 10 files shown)


xla/stream_executor/memory_reservation.h
```diff
@@ -0,0 +1,65 @@
+/* Copyright 2025 The OpenXLA Authors.
```
ezhulenev: 2026 in in all other files as well

xla/stream_executor/memory_reservation.h
```diff
@@ -0,0 +1,65 @@
...
+// physical allocations are mapped into it via MapToAllocation.
+//
+// MemoryReservation is the base class for platform-specific implementations
+// (e.g. CUDA virtual memory management via cuMemAddressReserve).
+class MemoryReservation {
+ public:
+  MemoryReservation() = default;
+  virtual ~MemoryReservation() = default;
+
+  MemoryReservation(const MemoryReservation&) = delete;
```
ezhulenev: deleting only move constructor/assignment is enough, copy will be removed auotmatically

xla/stream_executor/memory_reservation.h
```diff
@@ -0,0 +1,65 @@
...
+  // Returns the base address and size of the reserved virtual address range.
+  // The returned address is a valid device virtual address, but may not be
+  // accessible until physical memory is mapped via MapToAllocation and
+  // access is enabled via SetAccess.
+  virtual DeviceAddressBase address() const = 0;
+
+  // Maps a physical allocation into the sub-range [offset, offset+size) of
+  // this reservation. `offset` is a byte offset from the base address of this
+  // reservation. `size` must not exceed the size of `allocation`.
+  virtual absl::Status MapToAllocation(uint64_t offset, uint64_t size,
```
ezhulenev: Instead of Map/UnMap in public API (it should be hard to forget to call UnMap):

```
class MemoryReservation {
  // RAII handler for mapped address range. Physical allocation automatically unmapped when
  // handles goes out of scope. 
  class ScopedMapping {
    public:
      ~ScopedMapping();

      se::DeviceAddressBase() mapped_address();
      MemoryAllocation* allocation();
      uint64_t offset();
      uint64_t size();
    private:
      friend class MemoryReservation;
      ScopedMapping(... we don't want constructor to be private);
      MemoryReservation* reservation_; // must be alive
      MemoryAllocation* allocation_; // must be alive
      ...
  };

  absl::StatusOr<ScopedMapping> MapToAllocation(...) {
    implement it on top of private APIs
  }

  private:
    virtual absl::Status MapToAllocation(uint64_t offset, uint64_t size, ...);
    virtual absl::Status UnMap(uint64_t offset, uint64_t size)
};
```

xla/stream_executor/cuda/cuda_memory_reservation.cc
```diff
@@ -0,0 +1,164 @@
...
+}
+
+CudaMemoryReservation::CudaMemoryReservation(CudaMemoryReservation&& other)
+    : executor_(other.executor_), ptr_(other.ptr_), size_(other.size_) {
+  other.executor_ = nullptr;
+  other.ptr_ = 0;
+  other.size_ = 0;
+}
+
+CudaMemoryReservation& CudaMemoryReservation::operator=(
```
ezhulenev: I don't think we should allow move assignemnt or construction for memory reservation, it complicates everythig.

xla/stream_executor/cuda/cuda_memory_reservation.cc
```diff
@@ -0,0 +1,164 @@
...
+
+CudaMemoryReservation::~CudaMemoryReservation() {
+  if (ptr_ == 0) {
+    return;
+  }
+  std::unique_ptr<ActivateContext> activation = executor_->Activate();
+  // Unmap the full range before freeing the virtual address space.
+  // Any sub-ranges that were not explicitly unmapped are cleaned up here.
+  auto unmap_status =
+      cuda::ToStatus(cuMemUnmap(ptr_, size_), "Error unmapping CUDA memory");
```
ezhulenev, line 121: We shouldn't be doing it here, all ScopedMappings must goi out of scope before we destroy the reservation, if ScopedMappins are alive it means it's a user program error.

xla/stream_executor/cuda/cuda_memory_reservation.cc
```diff
@@ -0,0 +1,164 @@
...
+  props.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  props.location.id = device;
+  props.requestedHandleTypes =
+      static_cast<CUmemAllocationHandleType>(CU_MEM_HANDLE_TYPE_NONE);
+  return props;
+}
+
+}  // namespace
+
+/*static*/ absl::StatusOr<CudaMemoryReservation> CudaMemoryReservation::Create(
```
ezhulenev: no need to add /*static*/ (we should clean up that from XLA)

xla/stream_executor/cuda/cuda_raw_memory_allocation.h
```diff
@@ -0,0 +1,64 @@
...
+namespace stream_executor::gpu {
+
+// RAII wrapper for a physical memory allocation created via cuMemCreate.
+// The handle is not mapped to any virtual address; this class only manages
+// the lifetime of the CUmemGenericAllocationHandle.
+class CudaRawMemoryAllocation : public MemoryAllocation {
+ public:
+  // Creates a physical memory allocation of at least `size` bytes using
+  // cuMemCreate. StreamExecutor is used only for context activation.
+  static absl::StatusOr<CudaRawMemoryAllocation> Create(
```
ezhulenev: It's ok to have it this way for now, but proper solution is to add CudaRawMemoryAllocator that implements MemoryAllocator API, but this PR is already big, so it's up to you if you want to add it here or in the next one.

xla/stream_executor/memory_reservation.h
shawnwang18: This might not work, because there are cases when a single reservation has mapping of multiple allocations, then you will have multiple mapping, but you only need to do one Unmap for the reservation, so this ScopedMapping pattern does not work

xla/stream_executor/cuda/cuda_memory_reservation.cc
shawnwang18: Per  the comment on ScoppedMapping will not work, I think it still make sense to put the Unmap operation here? 

```This might not work, because there are cases when a single reservation has mapping of multiple allocations, then you will have multiple mapping, but you only need to do one Unmap for the reservation, so this ScopedMapping pattern does not work```

File: xla/stream_executor/cuda/cuda_memory_reservation.cc (hunk 6 above)
Comment (shawnwang18) [reply xla/stream_executor/cuda/cuda_memory_reservation.cc]: fixed.

File: xla/stream_executor/cuda/cuda_memory_reservation.cc (hunk 4 above)
Comment (shawnwang18) [reply xla/stream_executor/cuda/cuda_memory_reservation.cc]: fixed.

File: xla/stream_executor/cuda/cuda_raw_memory_allocation.h (hunk 7 above)
Comment (shawnwang18) [reply xla/stream_executor/cuda/cuda_raw_memory_allocation.h]: OK, let's have a seperate PR for doing this.

File: xla/stream_executor/memory_reservation.h (hunk 1 above)
Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: Fixed.

File: xla/stream_executor/memory_reservation.h (hunk 2 above)
Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: Fixed.

File: xla/stream_executor/memory_reservation.h (hunk 3 above)
Comment (ezhulenev) [reply xla/stream_executor/memory_reservation.h]: `[cuMemMap](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html#group__CUDA__VA_1gff1d395423af5c5c75375516959dae56) can only create mappings on VA range reservations that are not currently mapped.` - looks like you can have at most one `ScopedMapping` alive at a time?

Also I though that offset and size are from reservation, but it doesn't change the picture.

If it's possible to create multiple mappings for non-overlapped ranges? Then `ScopedMapping` still can be an exclusive owner of the reservation range.

Also if it's possible to create multiple mappings, then the API should be like this?

```
// Maps a reservation range [reservation_offset, reservation_offset+size) to memory allocation range
// [allocation_offset, allocation_offset+size).
absl::StatusOr<ScopedMapping> MapToAllocation(size_t reservation_offset, 
  MemoryAllocation* allocation, size_t allocation_offset, size_t size ) {
    implement it on top of private APIs
  }
```

Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: Yes, you can have multiple non-overlapping mappings for a single reservation with multiple CudaRawMemoryAllocation (each raw allocation has the offset 0), in this case, when doing the unmapping of the reservation, you don't need to call N times of Unmap for each mapping, Instead you can call a single Unmap on the whole reservation, then cuda will unmap all the mappings currently mapped to the reservation.  

It seems that ScopedMapping does not support this pattern.

Comment (ezhulenev) [reply xla/stream_executor/memory_reservation.h]: `The range must be the entire contiguous address range that was mapped to`

Imagine this:

1. reservation [0 ..... 1000]
2. map [0 ... 500] to allocation#0
3. map [501....1000] to allocation#1
4. unmap [0...500] - this will automatically make mapping 2 invalid? I'm not even sure you can call unmap [0....1000] because it will be 2 ranges, and I suspect coMemUnmap will fail.

Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: 4. unmap [0...500] - this will automatically make mapping 2 invalid? I'm not even sure you can call unmap [0....1000] because it will be 2 ranges, and I suspect coMemUnmap will fail.

If you call unmap[0..500], mapping 2 is still valid. 

But you can also call unmap[0..1000] without any failure, I have tested it, and my update-free design in this PR (https://github.com/openxla/xla/pull/35315) uses it a lot.

Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: > If you call unmap[0..500], mapping 2 is still valid.

Sorry this maybe not true.  I have not tested this pattern.

Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: > > If you call unmap[0..500], mapping 2 is still valid.
> 
> Sorry this maybe not true. I have not tested this pattern.

Hmmm, from the doc, it seems that it support it?  

```
The range must be the entire contiguous address range that was mapped to
```

It might mean that we can not call unmap[0..200] ?

Comment (ezhulenev) [reply xla/stream_executor/memory_reservation.h]: I think it means that you can unmap exactly what you mapped, so it means that every ScopedMapping will unmap it's own portion of reservation and everything should work. And if only one mapping goes out of scope every other mapping is still valid.

Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: > I think it means that you can unmap exactly what you mapped, so it means that every ScopedMapping will unmap it's own portion of reservation and everything should work. And if only one mapping goes out of scope every other mapping is still valid.

Yes, but if I want to call Unmap[0..1000], then ScoppedMapping does not support, right?

Comment (ezhulenev) [reply xla/stream_executor/memory_reservation.h]: Yes, and it's fine, because it's too low level API, and too easy to misuse. ScopedMapping RAII mechanism guarantees that users don't forget to call unmap on various error paths.

Comment (shawnwang18) [reply xla/stream_executor/memory_reservation.h]: Maybe we can use the public API map/unmap in this PR?  

I think in later PR, we can also add the ScopedMapping, so supports both map/unmap API and ScopedMapping. 

if the user does use the first unmap pattern Unmap[0..500], then it can use ScoppedMapping.

Comment (ezhulenev) [reply xla/stream_executor/memory_reservation.h]: Let's do the other way :) if you **really** need low level API, then let's make it public later, but I hope we don't need to do it at all, because it's unsafe. If you mix Scoped/Map/Unmap, you can easily get into illegal state. If you want to coalesce all unmappings then we can use refernce count, but I'm not sure what's the value, because we should support arbitrary Map/Unmap cycles.

File: xla/stream_executor/memory_reservation.h [hunk 8]
```diff
@@ -0,0 +1,112 @@
...
+  // Unmaps the mapped range from the reservation on destruction.
+  class ScopedMapping {
+   public:
+    ScopedMapping() = default;
+    ~ScopedMapping();
+
+    ScopedMapping(ScopedMapping&&) noexcept;
+    ScopedMapping& operator=(ScopedMapping&&) noexcept;
+
+    ScopedMapping(const ScopedMapping&) = delete;
```
ezhulenev: It can be removed as custom move assignemtn deletes it autoamtically.

xla/stream_executor/memory_reservation.h
```diff
@@ -0,0 +1,112 @@
...
+
+  // Maps multiple physical allocations to a contiguous memory reservation
+  // range. The descriptors must form a contiguous range in the reservation.
+  // Enables device access to the full contiguous range before returning.
+  // For non-contiguous mappings, use separate MapTo calls instead.
+  absl::StatusOr<ScopedMapping> MapTo(
+      absl::Span<const MappingDescriptor> mappings);
+
+ private:
+  virtual absl::Status Map(size_t reservation_offset, size_t allocation_offset,
```
ezhulenev, line 103: add size arg?

xla/stream_executor/memory_reservation.h
```diff
@@ -0,0 +1,112 @@
...
+                           size_t size, MemoryAllocation& allocation) = 0;
+
+  virtual absl::Status SetAccess() = 0;
```
ezhulenev: SetAccess needs offset + size, because we must set access only on the range that was mapped.

xla/stream_executor/memory_reservation.h
```diff
@@ -0,0 +1,112 @@
...
+
+  virtual absl::Status UnMap(size_t offset, size_t size) = 0;
```
ezhulenev: reservation_offset for consistencyu

xla/stream_executor/memory_reservation.cc
```diff
@@ -0,0 +1,134 @@
...
+
+MemoryReservation::ScopedMapping::ScopedMapping(MemoryReservation* reservation,
+                                                size_t reservation_offset,
+                                                size_t size)
+    : reservation_(reservation),
+      reservation_offset_(reservation_offset),
+      size_(size) {}
+
+MemoryReservation::ScopedMapping::~ScopedMapping() {
+  if (reservation_ == nullptr) return;
```
ezhulenev: google style requires {} around all ifs

xla/stream_executor/memory_reservation.cc
```diff
@@ -0,0 +1,134 @@
...
+    : reservation_(other.reservation_),
+      reservation_offset_(other.reservation_offset_),
+      size_(other.size_) {
+  other.reservation_ = nullptr;
+}
+
+MemoryReservation::ScopedMapping& MemoryReservation::ScopedMapping::operator=(
+    ScopedMapping&& other) noexcept {
+  if (this != &other) {
+    if (reservation_ != nullptr) {
```
ezhulenev, line 67: Extract it into static Unmap helper function? Used in destructor as well.

xla/stream_executor/memory_reservation.cc
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// MemoryReservation::MapTo
+
+absl::StatusOr<MemoryReservation::ScopedMapping> MemoryReservation::MapTo(
+    size_t size, size_t reservation_offset, size_t allocation_offset,
+    MemoryAllocation& allocation) {
+  absl::Status map_status =
+      Map(reservation_offset, allocation_offset, size, allocation);
+  if (!map_status.ok()) return map_status;
```
ezhulenev: RETURN_IF_ERROR (available in tsl/platform/statuc_macros)

xla/stream_executor/memory_reservation.cc
```diff
@@ -0,0 +1,134 @@
...
+
+  absl::Status access_status = SetAccess();
```
ezhulenev: RETURN_IF_ERROR

xla/stream_executor/memory_reservation.cc
```diff
@@ -0,0 +1,134 @@
...
+  if (!access_status.ok()) {
+    absl::Status unmap_status = UnMap(reservation_offset, size);
```
ezhulenev, line 91: this can be done with absl::Cleanup that you reset if you get to return statement

xla/stream_executor/memory_reservation.cc
```diff
@@ -0,0 +1,134 @@
...
+                 << unmap_status.message();
+    }
+    return access_status;
+  }
+  return ScopedMapping(this, reservation_offset, size);
+}
+
+absl::StatusOr<MemoryReservation::ScopedMapping> MemoryReservation::MapTo(
+    absl::Span<const MappingDescriptor> mappings) {
+  if (mappings.empty()) {
```
ezhulenev, line 106: Add IsContiguous check?

xla/stream_executor/memory_reservation.cc
```diff
@@ -0,0 +1,134 @@
...
+    return absl::InvalidArgumentError("MapTo: mappings must not be empty");
+  }
+
+  size_t start_offset = mappings[0].reservation_offset;
+  size_t total_size = 0;
+
+  for (const MappingDescriptor& desc : mappings) {
+    absl::Status map_status = Map(desc.reservation_offset, desc.allocation_offset,
+                                  desc.size, *desc.allocation);
+    if (!map_status.ok()) {
```
ezhulenev: again, once you have successfull map push absl::Cleanup to temp container, if you reach the end you can clear all cleanup objects

xla/stream_executor/cuda/cuda_raw_memory_allocation.h
```diff
@@ -0,0 +1,64 @@
...
+  // CUmemGenericAllocationHandle cast to void*, and size() is the
+  // padded allocation size.
+  DeviceAddressBase address() const override;
+
+  CUmemGenericAllocationHandle GetHandle() const { return handle_; }
+
+  ~CudaRawMemoryAllocation() override;
+  CudaRawMemoryAllocation(const CudaRawMemoryAllocation&) = delete;
+  CudaRawMemoryAllocation& operator=(const CudaRawMemoryAllocation&) = delete;
+  CudaRawMemoryAllocation(CudaRawMemoryAllocation&&);
```
ezhulenev: delete it? We never move RawMemoryAllocations, treat is as std::mutex, non-moveable and non-copyable, in practice it's always passed as std::unique_ptr<MemoryAllocation>

xla/stream_executor/cuda/cuda_raw_memory_allocation.h
```diff
@@ -0,0 +1,64 @@
...
+namespace stream_executor::gpu {
+
+// RAII wrapper for a physical memory allocation created via cuMemCreate.
+// The handle is not mapped to any virtual address; this class only manages
+// the lifetime of the CUmemGenericAllocationHandle.
+class CudaRawMemoryAllocation : public MemoryAllocation {
+ public:
+  // Creates a physical memory allocation of at least `size` bytes using
+  // cuMemCreate. StreamExecutor is used only for context activation.
+  static absl::StatusOr<CudaRawMemoryAllocation> Create(
```
ezhulenev: return absl::StatusOr<std::unique_ptr<MemoryAllocation>> and then you don't need move constructor/assignemt, and it's consistent with real API uses that will go via allocator

xla/stream_executor/memory_reservation.cc
shawnwang18: added

xla/stream_executor/memory_reservation.cc
shawnwang18: fixed

xla/stream_executor/memory_reservation.cc
shawnwang18: done

xla/stream_executor/memory_reservation.h
shawnwang18: there is a size arg after allocation_offset, I am assuming the size of the reservation and allocation is the same.

ezhulenev: oops missed it

ezhulenev (approved):

pifon2a (approved):

PR #38689, opened 2026-03-04
[xla] Update xla::concurrency package to use standard concurrency primitives

ezhulenev:
Use standard `xla::Future` and `tsl::Executor` to implement parallel `ForEach` for HLO computations.

Deprecate `TskTaskExecutor`, as all users should prefer the default one which respects the `NPROC` env variable used everywhere in XLA to control the number of threads.


pifon2a (approved):

pifon2a (approved):

pifon2a (approved):

thcmbs:
Hi Eugene, apologies for the delay here. 
FYI: this CL is non trivial to merge internally because xla::concurrency::ForEach(...) is used in non OSS components. I need to find time to figure out how to propagate your PR there.

ezhulenev:
Thanks for looking into that. I tried to make it backward compatible, and unless internally ForEach is used with custom executors it should just work.

PR #38732, opened 2026-03-05
[ROCm] Support hipblaslt group-gemm 1/4

mfrancepillois:
**PR 1/4: Upper layers of the support**

PR 1/4 : https://github.com/openxla/xla/pull/38732  <==
PR 2/4 : https://github.com/openxla/xla/pull/38735
PR 3/4 : https://github.com/openxla/xla/pull/38736
PR 4/4 : https://github.com/openxla/xla/pull/38737

📝 Summary of Changes
Add HipBlasLT-Ext GroupedGemm support for RaggedDot operation implemented for matrices with and without batch dimension.
Three ragged modes supported:
    - ragged in Non-Contracting dimension
    - ragged in Contracting dimension
    - ragged in Batch dimension
This support takes advantage of the hip group-gemm support (https://rocm.docs.amd.com/projects/hipBLASLt/en/develop/reference/ext-reference.html#groupedgemm).

Please note that the HIP group-gemm support is not fully optimized yet and some configurations still need to be fine-tuned by the HIP team. Also, not all datatypes are supported yet. On MI300, only FP16 and BF16 are supported. On MI350, only FP16 is supported but not tuned yet. For these reasons, we have added a flag allowing users to enable or disable this feature.

Note that this PR is the first out of four PRs. 
It specifically adds the higher layers of the group-gemm support:
-	Changes to the existing compilation passes to let `RaggedDot` reaches `gemm_rewriter` where it will be rewritten into custom-call.

🎯 Justification
Group-gemm support enables faster execution of ragged dot operations while reducing the memory usage. Indeed, the baseline handles ragged-dot operation by duplicating the ragged input matrices in multiple matrices, then perform regular dot operations. With the group-gemm support, the ragged inputs are no longer duplicated but the hardware support fetches the data directly from the original ragged matrix, reducing the memory usage of such operations.
See below (benchmark section) for more details on performance improvements.

🚀 Kind of Contribution
Please remove what does not apply:⚡️ Performance Improvement, ✨ New Feature,

📊 Benchmark (for Performance Improvements)

Performance evaluation has been carried out on MI300.
Grou-Gemm performances were compared to performances of the ragged-dot rewritten into regular dot operation by the ragged-dot-rewriter pass (baseline).

### Micro-benchmark
We run JAX `lax.ragged_dot` operation with different input sizes, with and without using the group-gemm support.
After a a warm-up of 10 execution, we run 10 times 10  `lax.ragged_dot` operations and take the mean runtime.

#### num_experts = 16

[...]

PR #38735, opened 2026-03-05
[ROCm] Support hipblaslt group-gemm 2/5

mfrancepillois:
**PR 2/5: Middle layers of the support: transform and IR emission**

PR 1/5 : https://github.com/openxla/xla/pull/38732  
PR 2/5 : https://github.com/openxla/xla/pull/38735 <==
PR 3/5 : https://github.com/openxla/xla/pull/38909 
PR 4/5 : https://github.com/openxla/xla/pull/38736
PR 5/5 : https://github.com/openxla/xla/pull/38737

📝 Summary of Changes
This is the second PR out of five.
It introduces some of the middle layers of the group-gemm support:
- the remaining transform passes
- IR emission passes

### DIFF
--- a/xla/backends/gpu/transforms/gemm_fusion.cc
+++ b/xla/backends/gpu/transforms/gemm_fusion.cc
@@ -884,6 +884,10 @@ class GemmFusionVisitor : public DfsHloRewriteVisitor {
             .xla_gpu_experimental_use_ragged_dot_grouped_gemm() &&
         module->config().debug_options().xla_gpu_enable_cublaslt();
     if (has_grouped_gemm) {
+      // At the moment, if Gpublaslt support is available, it is prefered
+      // over triton fused ragged-dot. Therefore, we skip this pass and
+      // does not fused the ragged-dot op if the Gpublaslt support
+      // is available for this operation.
       return absl::OkStatus();
     }
 
--- a/xla/backends/gpu/transforms/gemm_rewriter.cc
+++ b/xla/backends/gpu/transforms/gemm_rewriter.cc
@@ -716,6 +716,77 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
     return absl::OkStatus();
   }
 
+  absl::Status HandleRaggedDot(HloInstruction* instr) override {
+    if (!IsGpublasLtSupportedGroupedMatMul(*instr)) {
+      return absl::OkStatus();
+    }
+    HloRaggedDotInstruction* ragged_dot =
+        DynCast<HloRaggedDotInstruction>(instr);
+    if (ragged_dot == nullptr) {
+      return absl::OkStatus();
+    }
+    const auto& ragged_dims = ragged_dot->ragged_dot_dimension_numbers();
+    const auto& dot_dims = ragged_dims.dot_dimension_numbers();
+    if (ragged_dims.lhs_ragged_dimensions().size() != 1) {
+      return absl::UnimplementedError("lhs_ragged_dimensions must have size 1");
+    }
+    int lhs_ragged_dim = ragged_dims.lhs_ragged_dimensions(0);
+
+    auto isLhsRaggedDimInContractingDim = [](int lhs_ragged_dim,
+                                             const DotDimensionNumbers& dnums) {
+      return std::any_of(dnums.lhs_contracting_dimensions().begin(),
+                         dnums.lhs_contracting_dimensions().end(),
+                         [&](auto dim) { return dim == lhs_ragged_dim; });
+    };
+
+    auto isLhsRaggedDimInBatchDim = [](int lhs_ragged_dim,
+                                       const DotDimensionNumbers& dnums) {
+      return std::any_of(dnums.lhs_batch_dimensions().begin(),
+                         dnums.lhs_batch_dimensions().end(),
+                         [&](auto dim) { return dim == lhs_ragged_dim; });
+    };
+
+    if (!isLhsRaggedDimInContractingDim(lhs_ragged_dim, dot_dims) &&
+        !isLhsRaggedDimInBatchDim(lhs_ragged_dim, dot_dims) &&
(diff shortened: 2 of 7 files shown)


xla/backends/gpu/transforms/gemm_fusion.cc
```diff
@@ -877,6 +877,16 @@ class GemmFusionVisitor : public DfsHloRewriteVisitor {
   }
 
   absl::Status HandleRaggedDot(HloInstruction* ragged_dot) override {
+    auto module = ragged_dot->GetModule();
+    const bool has_grouped_gemm =
+        module->config()
+            .debug_options()
+            .xla_gpu_experimental_use_ragged_dot_grouped_gemm() &&
+        module->config().debug_options().xla_gpu_enable_cublaslt();
+    if (has_grouped_gemm) {
```
vwbaker, line 886: add a comment saying we don't want to fuse ragged dots that would be better suited for gpublaslt?

xla/service/gpu/ir_emission_utils.cc
```diff
@@ -74,6 +74,24 @@ limitations under the License.
 namespace xla {
 namespace gpu {
 
+absl::StatusOr<bool> IsCublasLtSupportedGroupedMatMul(
```
vwbaker: IsGpublasLtSupportedGroupedMatMul

vwbaker: at what point might this return a status rather than a boolean? let's return a boolean if possible

xla/backends/gpu/transforms/gemm_rewriter.cc
```diff
@@ -716,6 +716,85 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+      return absl::OkStatus();
+    }
+    const auto& ragged_dims = ragged_dot->ragged_dot_dimension_numbers();
+    const auto& dot_dims = ragged_dims.dot_dimension_numbers();
+    if (ragged_dims.lhs_ragged_dimensions().size() != 1) {
+      return absl::UnimplementedError("lhs_ragged_dimensions must have size 1");
+    }
+    int lhs_ragged_dim = ragged_dims.lhs_ragged_dimensions(0);
+
+    auto isLhsRaggedDimInContractingDim = [](int lhs_ragged_dim,
```
vwbaker, line 735: nit: this would be cleaner:

return std::any_of(dnums.lhs_contracting_dimensions().begin(), dnums.lhs_contracting_dimensions().end(), [&](auto dim) {return dim == lhs_ragged_dim});

for below as well.

xla/backends/gpu/transforms/gemm_rewriter.cc
```diff
@@ -716,6 +716,85 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
...
+      }
+      return false;
+    };
+
+    if (!isLhsRaggedDimInContractingDim(lhs_ragged_dim, dot_dims) &&
+        !isLhsRaggedDimInBatchDim(lhs_ragged_dim, dot_dims) &&
+        ragged_dims.rhs_group_dimensions().size() != 1) {
+      return absl::UnimplementedError(
+          "rhs_group_dimensions must have size equal to 1 when lhs ragged "
+          "dimension is a non-contracting dimension");
```
vwbaker, line 754: I don't see a similar check happening in ragged dot rewriter; can we be making sure it doesn't make it this far? if it's going to throw an unimplemented error later, it should just rewrite the ragged dot to an expanded dot.

vwbaker (commented):
This is quite large & will require multiple people to review. Can you separate the runtime changes from this PR? I can review the transforms and ir_emission_utils, but the runtime, thunk, and stream_executor stuff should be able to be separated.

xla/backends/gpu/transforms/gemm_rewriter.cc
mfrancepillois: This condition is actually enforced by the instruction itself (https://github.com/openxla/xla/blob/34d7ec76e5ede12e8daef505a9e001f0beca2501/xla/service/shape_inference.cc#L1144-L1149).
So, here we are just double-checking that this condition is valid before transforming the instruction to a custom-call. But, I can remove it if you prefer.

xla/backends/gpu/transforms/gemm_fusion.cc
mfrancepillois: A comment has been added.

xla/service/gpu/ir_emission_utils.cc
mfrancepillois: Renamed 👍

mfrancepillois: Return type changed 👍

xla/backends/gpu/transforms/gemm_rewriter.cc
mfrancepillois: Thanks for the suggestion. The code has been updated.

mfrancepillois:
> This is quite large & will require multiple people to review. Can you separate the runtime changes from this PR? I can review the transforms and ir_emission_utils, but the runtime, thunk, and stream_executor stuff should be able to be separated.

The PR has been split.

xla/backends/gpu/transforms/gemm_rewriter.cc
vwbaker: Ok, then this is fine; thanks.

vwbaker (approved):

xla/service/gpu/backend_configs.proto
thomasjoerg: Please update the Next-Id

thomasjoerg (changes requested):

xla/service/gpu/backend_configs.proto
mfrancepillois: Thanks for noticing. This has been updated.

thomasjoerg (approved):

PR #38736, opened 2026-03-05
[ROCm] Support hipblaslt group-gemm 4/5

mfrancepillois:
**PR 4/5: Hipblaslt Group-Gemm support**

PR 1/5 : https://github.com/openxla/xla/pull/38732  
PR 2/5 : https://github.com/openxla/xla/pull/38735 
PR 3/5 : https://github.com/openxla/xla/pull/38909 
PR 4/5 : https://github.com/openxla/xla/pull/38736 <==
PR 5/5 : https://github.com/openxla/xla/pull/38737

📝 Summary of Changes
This is the fourth PR out of five.
 It introduces the low layers of the group-gemm support:
-	Hipblaslt support for group-gemm
-	Adds e2e tests
This support takes advantage of the hip group-gemm support (https://rocm.docs.amd.com/projects/hipBLASLt/en/develop/reference/ext-reference.html#groupedgemm).
As the user-defined arguments are only now at runtime, the group-gemm arguments (user-arguments) are set by additional kernels that set these arguments at runtime directly in the GPU memory.

🧪 Execution Tests:
e2e tests are included in this PR. The tests verify:
- the translation of the custom-call into a CublasLtGroupedMatmulThunk and execute correctly on device, checking results correctness against the baseline.


dextero:
Does this PR include the changes from previous parts as well?

If so, would you mind rebasing this PR on the current state of #38909 to make it easier to see the new changes only? Thank you!

mfrancepillois:
> Does this PR include the changes from previous parts as well?
> 
> If so, would you mind rebasing this PR on the current state of #38909 to make it easier to see the new changes only? Thank you!

Thank you for taking a look at this PR. The problem is that the branches of these PRs come from a fork. So, in order for the PRs to be visible in `openxla/xla` repository, I had to open the PRs against the main branch of `openxla/xla`. So, I think I need to wait for the [PR2 ](https://github.com/openxla/xla/pull/38735) and [PR3 ](https://github.com/openxla/xla/pull/38909) to be merged into main, and then rebased this branch/PR.

xla/backends/gpu/runtime/gpublas_lt_matmul_thunk.cc
```diff
@@ -322,41 +322,42 @@ absl::StatusOr<ThunkProto> CublasLtMatmulThunk::ToProto() const {
   cublas_lt_matmul_thunk->set_epilogue(
       stream_executor::gpu::BlasLt::EpilogueToProto(epilogue_));
   cublas_lt_matmul_thunk->set_algorithm_idx(algorithm_idx_);
-  cublas_lt_matmul_thunk->set_autotune_workspace_size(autotune_workspace_size_);
   cublas_lt_matmul_thunk->set_canonical_hlo(canonical_hlo_);
-  ASSIGN_OR_RETURN(*cublas_lt_matmul_thunk->mutable_a(), a_.ToProto());
```
dextero, line 327: Please remove the macro renames, so that the actual logic changes are easier to spot.

xla/backends/gpu/runtime/gpublas_lt_matmul_thunk.h
```diff
@@ -38,16 +38,19 @@ namespace gpu {
...
-      std::optional<ShapedSlice> a_scale, std::optional<ShapedSlice> b_scale,
-      std::optional<ShapedSlice> c_scale, std::optional<ShapedSlice> d_scale,
-      std::optional<ShapedSlice> d_amax,
-      std::optional<const ShapedSlice> workspace);
+  CublasLtMatmulThunk(Thunk::ThunkInfo thunk_info, std::string canonical_hlo,
+                      GemmConfig gemm_config,
+                      se::gpu::BlasLt::Epilogue epilogue, int64_t algorithm_idx,
+                      int64_t autotune_workspace_size, ShapedSlice a,
+                      ShapedSlice b, ShapedSlice c, ShapedSlice d,
+                      std::optional<ShapedSlice> bias /* may be null */,
```
dextero: Nit: the `may be null` comments are not needed anymore.

xla/backends/gpu/transforms/gemm_rewriter_test_lib.cc
```diff
@@ -71,6 +71,11 @@ bool GemmRewriteTestBase::SkipGpuBlasLtTest() {
          GetDebugOptionsForTest().xla_gpu_enable_cublaslt();
 }
 
+bool GemmRewriteTestBase::SkipGroupedGemmTest() {
+  // Grouped GEMM is only supported on ROCm with hipBLASLt
+  return IsCuda() || !Capability().rocm_compute_capability()->has_hipblaslt();
```
dextero: Nit: won't `has_hipblaslt` evaluate to false for non-ROCm platforms anyway?

I think this could be just `return !Capability().rocm_compute_capability()->has_hipblaslt();`.

xla/backends/gpu/transforms/cublas_gemm_rewriter_test.cc
```diff
@@ -3713,6 +3639,708 @@ ENTRY test (x: f32[2,3,4], y: f32[4,5,7], z: f32[7]) -> f32[2,3,5,7] {
...
+    if (SkipGroupedGemmTest()) {
+      GTEST_SKIP() << "Grouped GEMM is only supported on ROCm with hipBLASLt";
+    }
+  }
+};
+
+TEST_F(GroupedGemmRewriteTest, CheckCustomCallTargetGroupedGemm) {
+  if (SkipGpuBlasLtTest()) {
+    GTEST_SKIP() << "BlasLt is not supported on this GPU architecture";
+  }
```
dextero: Nit: is this a requirement for all grouped gemm tests? If it is, how about including the check in `SkipGroupedGemmTest`, or `GroupedGemmRewriteTest` to not duplicate it in each test?

mfrancepillois:
> Does this PR include the changes from previous parts as well?
> 
> If so, would you mind rebasing this PR on the current state of #38909 to make it easier to see the new changes only? Thank you!

The PR has finally been rebased. It now includes only the changes introduced by this PR related to the low-level layers and ROCm implementation.

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -214,14 +224,49 @@ absl::Status BlasLt::Init() {
...
+    absl::MutexLock lock(&blas_lt->mu_);
+
+    std::unique_ptr<ActivateContext> activation = blas_lt->parent_->Activate();
+
+    auto problem = grouped_gemm_->getProblemTypes()[0];
+
+    grouped_gemm_->setMaxWorkspaceBytes(max_workspace_size);
+
+    SE_HIPBLAS_RETURN_IF_ERROR(
+        getAllAlgos(blas_lt->blas_lt_.get(),
```
dextero: Can we please prefix it with `hipblaslt_ext::`, so that we don't have to rely on ADL to figure out the namespace?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -214,14 +224,49 @@ absl::Status BlasLt::Init() {
...
+    VLOG(2) << "Total heuristics found: " << heuristicResult.size();
+    std::vector<MatmulAlgorithm> algorithms;
+    algorithms.reserve(max_algorithm_count);
+    for (hipblasLtMatmulHeuristicResult_t &result : heuristicResult) {
+      size_t workspace_size = 0;
+      if ((result.state == HIPBLAS_STATUS_SUCCESS) &&
+          (grouped_gemm_->isAlgoSupported(result.algo, workspace_size) ==
+           HIPBLAS_STATUS_SUCCESS)) {
+        algorithms.push_back({result.algo, result.workspaceSize});
+        if (algorithms.size() >= max_algorithm_count) break;
```
dextero: In the pathological case of calling the function with `max_algorithm_count == 0` this may return 1 algorithm. Is that intentional, or should we rather check against `max_algorithm_count` before `push_back`?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -214,14 +224,49 @@ absl::Status BlasLt::Init() {
   return std::move(desc);
 }
 
-auto BlasLt::MatmulPlan::GetAlgorithms(const Stream* stream,
+auto BlasLt::MatmulPlan::GetAlgorithms(const Stream *stream,
                                        size_t max_algorithm_count,
                                        size_t max_workspace_size) const
     -> absl::StatusOr<std::vector<MatmulAlgorithm>> {
+  // Handle grouped matmul case
+  if (is_grouped()) {
```
dextero: Nit, optional: the grouped and non-grouped cases are completely independent, and the function is pretty long already. Can we split this into `GetAlgorithmsForGroupedMatmul` and `GetAlgorithmsForNonGroupedMatmul`, similar to how we've got two `Execute*Matmul`?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+            cfg.precision_algorithm, primitive_type_a, primitive_type_d,
+            cfg.compute_precision,
+            parent_->GetDeviceDescription().gpu_compute_capability()));
+  }
+  if (!compute_type) {
+    return absl::InternalError(
+        "This algorithm requires a non-zero compute_type!");
+  }
+
+  bool must_swap = cfg.must_swap_operands;
```
dextero: Nit: this isn't used anywhere else than in the `make_unique` below, which takes the bool by value. How about inlining it?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+    plan->grouped_gemm_ = std::make_unique<hipblaslt_ext::GroupedGemm>(
+        blas_lt_.get(), AsHipblasOperation(plan->cfg_->trans_a),
+        AsHipblasOperation(plan->cfg_->trans_b),
+        AsHipblasDataType(plan->cfg_->type_a),
+        AsHipblasDataType(plan->cfg_->type_b),
+        AsHipblasDataType(plan->cfg_->type_c),
+        AsHipblasDataType(plan->cfg_->type_d),
+        AsHipblasComputeType(*compute_type));
+  }
+  auto &ggemm = plan->grouped_gemm_;
```
dextero: Nit: This is only used once, way below. Can it be inlined?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+      break;
+    }
+  }
+
+  // TODO: recover GemmEpilogues from args
+  std::vector<hipblaslt_ext::GemmEpilogue> epilogue(plan->cfg_->group_count);
+  std::vector<hipblaslt_ext::GemmInputs> inputs(plan->cfg_->group_count);
+
+  // TODO Improve alpha and beta conversion (similarly to done in the
+  // MatmulPlan)
```
dextero: Any plans to clean up those TODOs?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+      AsHipblasDataType(plan->cfg_->type_c),
+      AsHipblasDataType(plan->cfg_->type_d),
+      AsHipblasComputeType(*compute_type));
+
+  // Set the Matrix orders does not seem to change anything.
+  // This unexpected behavior is worth investigating further.
+  // For the moment, we do not defined a specific order and
+  // go with the default order (i.e., COLUMN-MAJOR)
+  // problem.setOrderA(HIPBLASLT_ORDER_COL);
+  // problem.setOrderB(HIPBLASLT_ORDER_COL);
```
dextero: Can you elaborate on what is this about? The commented code appears to unconditionally set the values to the default column-major.

Is the issue that needs investigating is a hipBLASlt-specific behavior? And it gets investigated, do we expect to need to set the row ordering here?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+  {
+    absl::MutexLock lock(&mu_);
+    SE_HIPBLAS_RETURN_IF_ERROR(ggemm->setProblem(
+        v_m, v_n, v_k, v_batch_count, v_lda, v_ldb, v_ldc, v_ldd, v_strideA,
+        v_strideB, v_strideC, v_strideD, epilogue, inputs, problem));
+  }  // end block
```
dextero: Nit: the comment doesn't contain any meaningful information, can we please remove it?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+  if (algorithm_must_be_initialized_ ||
+      !args.workspace.IsSameAs(saved_address_workspace_)) {
+    void *addr_workspace = static_cast<void *>(
+        static_cast<uint8_t *>(args.workspace.opaque()) +
+        sizeof(hipblaslt_ext::UserArguments) * cfg_->group_count);
+    SE_HIPBLAS_RETURN_IF_ERROR(
+        grouped_gemm_->initialize(*palgo, addr_workspace));
+    algorithm_must_be_initialized_ = false;
+    saved_address_workspace_ = args.workspace;
+  }
```
dextero, line 886: Initializing things in ExecuteOnStream is a bit awkward, and as you noticed requires the use of `mutable` members.

Do you expect `MatmulPlan` to be reused in a way that the workspace pointer is going to change from call to call?

If not, `CublasLtMatmulThunk::Initialize` stage sounds like a better fit than `ExecuteOnStream` for the initialize() call, and would let us remove both `algorithm_must_be_initialized_` and `saved_address_workspace_`.

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+      case blas::DataType::kF8E4M3FN:    // 8-bit float: 1 sign, 4 exponent, 3
+                                         // mantissa bits (FN variant)
+      case blas::DataType::kF8E5M2FNUZ:  // 8-bit float: E5M2 FNUZ variant
+      case blas::DataType::kF8E4M3FNUZ:  // 8-bit float: E4M3 FNUZ variant
+      case blas::DataType::kF8E4M3:      // 8-bit float: 1 sign, 4 exponent, 3
+                                         // mantissa bits
+      case blas::DataType::kF8E3M4:      // 8-bit float: 1 sign, 3 exponent, 4
+                                         // mantissa bits
+      case blas::DataType::kF8E8M0FNU:   // 8-bit float: 8 exponent, 0 mantissa
+                                         // bits (FNU variant)
```
dextero: Nit, optional: the definition of the constants sounds like a better place to document what the enum values mean. I'd personally skip the comments here, but up to you.

xla/stream_executor/rocm/rocm_helpers.cu.cc
```diff
@@ -157,4 +159,474 @@ INSTANTIATE_BIAS_ACTIVATION(float, float)
...
+void GroupGemmUpdateArgs(
+    hipStream_t stream, hipblaslt_ext::UserArguments* args, const void* a,
+    const void* b, void* d, const void* group_sizes,
+    uint8_t group_size_bytewidth, uint8_t log2_byte_width_elem_a,
+    uint8_t log2_byte_width_elem_b, uint8_t log2_byte_width_elem_d,
+    uint32_t stride_ragged_dim, uint32_t stride_group_dim,
+    uint32_t output_stride_ragged_dim, bool must_swap_operands, uint32_t m,
+    uint32_t n, uint32_t k, uint32_t batch, uint32_t strideA1,
+    uint32_t strideA2, uint32_t strideB1, uint32_t strideB2, uint32_t strideD1,
+    uint32_t strideD2, const uint8_t ragged_mode, uint32_t num_gemms) {
```
dextero: Is there a reason to pass `ragged_mode` as `uint8_t` rather than the enum? If we could `switch` on an enum, we wouldn't need to comment the hard-coded integers.

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+  uint32_t strideD1 = cfg_->output_leading_dim_stride;
+  uint32_t strideD2 = cfg_->m * cfg_->n;
+
+  auto hip_stream =
+      absl::bit_cast<hipStream_t>(stream->platform_specific_handle().stream);
+
+  GroupGemmUpdateArgs(
+      hip_stream,
+      static_cast<hipblaslt_ext::UserArguments *>(d_userArgs->opaque()),
+      a.opaque(), b.opaque(), args.d.opaque(), args.group_sizes.opaque(),
```
dextero: Please use `DeviceMemory<>` instead of raw pointers when passing the device pointers around host code, and convert with `.opaque()` when launching the kernel.

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+        return 4;
+      default:
+        LOG(FATAL) << "Unknown DataType " << static_cast<int32_t>(ty);
+    }
+  };
+
+  DeviceMemoryBase a = cfg_->must_swap_operands ? args.b : args.a;
+  DeviceMemoryBase b = cfg_->must_swap_operands ? args.a : args.b;
+  DeviceMemoryBase *d_userArgs =
+      const_cast<DeviceMemoryBase *>(&args.workspace);
```
dextero: `const_cast` is not necessary, `DeviceMemoryBase::opaque()` returns non-const void regardless of the object's constness.

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+          cfg_->k * cfg_->n * cfg_->batch_count * cfg_->group_count});
+  std::unique_ptr<EventBasedTimer> timer;
+
+  if (profile_result != nullptr) {
+    TF_ASSIGN_OR_RETURN(timer, stream->CreateEventBasedTimer(
+                                   profile_result->warmup_run_executed()));
+  }
+
+  // ! The on-device update has not been updated to handle transposed matrices
+  // nor the different ragged modes.
```
dextero: What do you mean by "on-device update" here?

xla/stream_executor/rocm/hip_blas_lt.h
```diff
@@ -151,6 +183,9 @@ class BlasLt : public gpu::BlasLt {
       gpu::GroupedGemmConfig& config,
       const std::vector<Epilogue>& epilogues) const override;
 
+  void ClearGroupedMatmulPlanCache();
+  size_t GetGroupedMatmulPlanCacheSize() const;
+
```
dextero: These two don't seem to be implemented. Are they needed?

xla/stream_executor/rocm/hip_blas_lt.cc
```diff
@@ -686,14 +731,269 @@ absl::Status BlasLt::MatmulPlan::ExecuteOnStream(
...
+            cfg.compute_precision,
+            parent_->GetDeviceDescription().gpu_compute_capability()));
+  }
+  if (!compute_type) {
+    return absl::InternalError(
+        "This algorithm requires a non-zero compute_type!");
+  }
+
+  bool must_swap = cfg.must_swap_operands;
+  auto plan = std::make_unique<MatmulPlan>(std::move(cfg), must_swap);
```
dextero: The code in this function below this line operates almost exclusively on `MatmulPlan` internals to complete the initialization.

Did you consider putting this code in a grouped gemm specific `MatmulPlan` constructor or static factory method? This way we would avoid exposing `MatmulPlan` private data to `BlasLt`.

xla/stream_executor/rocm/hip_blas_lt.h
```diff
@@ -118,25 +132,43 @@ class BlasLt : public gpu::BlasLt {
...
+    std::optional<MatrixLayout> a_desc_;
+    std::optional<MatrixLayout> b_desc_;
+    std::optional<MatrixLayout> c_desc_;
+    std::optional<MatrixLayout> d_desc_;
+    std::optional<xla::complex128> alpha_;
+    std::optional<double> beta_;
     bool must_swap_operands_;
     std::optional<MatmulAlgorithm> algorithm_;  // selected algorithm
+    friend class BlasLt;
+    using GroupedGemmPtr = std::unique_ptr<hipblaslt_ext::GroupedGemm>;
```
dextero: Nit: if this is only used once, I'd consider inlining it.

xla/stream_executor/rocm/rocm_helpers.cu.cc
```diff
@@ -157,4 +159,474 @@ INSTANTIATE_BIAS_ACTIVATION(float, float)
...
+
+// Inline device function for copying data from shared memory to global memory
+// Uses vectorized uint4 copy for efficiency, with fallback for remaining bytes
+__device__ __forceinline__ void copy_shared_to_global(void* shared_src,
+                                                      void* global_dest,
+                                                      size_t total_bytes) {
+  size_t count_uint4 = total_bytes / sizeof(uint4);
+
+  // Vectorized copy using uint4 (16 bytes per iteration)
+  if (count_uint4 > 0) {
```
dextero: Optional: the `if` doesn't seem to make a difference, the loop 2 lines below would handle the zero-sized case.

xla/stream_executor/rocm/rocm_helpers.cu.cc
```diff
@@ -157,4 +159,474 @@ INSTANTIATE_BIAS_ACTIVATION(float, float)
...
+  auto* sharedUserArgs =
+      reinterpret_cast<hipblaslt_ext::UserArguments*>(shared_mem);
+
+  // Last thread initialize cumulative offset
+  // No need to syncthread here as this variable will be
+  // updated later by the last thread before
+  // being read by other threads.
+  if (threadIdx.x == BLOCK_SIZE - 1) {
+    cumulative_offset = 0;
+  }
```
dextero, line 485: We assume that
- No thread will read `cumulative_offset` value before the last thread reaches this line, and
- The last thread won't ever reach `cumulative_offset += offset_in_batch + group_size;` before other ones get to calculate `offset_group` based on `cumulative_offset`.
 
How do we guarantee the ordering?

xla/backends/gpu/transforms/gemm_fusion_test.cc
```diff
@@ -1425,6 +1425,31 @@ ENTRY main {
 )");
 }
 
+TEST_F(GemmFusionTest, RaggedDotNotBecomesFusion) {
```
vwbaker: RaggedDotDoesNotBecomeFusionUsingCublasLt has since been added - is this test doing anything different than that one?

xla/backends/gpu/transforms/cublas_gemm_rewriter_test.cc
```diff
@@ -3713,6 +3639,708 @@ ENTRY test (x: f32[2,3,4], y: f32[4,5,7], z: f32[7]) -> f32[2,3,5,7] {
...
+
+ protected:
+  void SetUp() override {
+    if (SkipGroupedGemmTest()) {
+      GTEST_SKIP() << "Grouped GEMM is only supported on ROCm with hipBLASLt";
+    }
+  }
+};
+
+TEST_F(GroupedGemmRewriteTest, CheckCustomCallTargetGroupedGemm) {
```
vwbaker: Can you change all the names - instead of "CheckCustomCallTarget" to be "CustomCallTargets" - prefer active verb for description.

Are there also things it shouldn't target that should get tested?

third_party/gpus/rocm/BUILD.tpl
```diff
@@ -636,6 +617,24 @@ cc_library(
     ]),
 )
 
+cc_library(
```
vwbaker: Is this different than the one removed above? Is it supposed to be? Perhaps move it to the same location so the difference can be seen better.

vwbaker (commented):
Mostly LGTM. I reviewed the backends/gpu/transforms files. I'll let @dextero look at the stuff under stream_executor.

xla/backends/gpu/transforms/gemm_fusion_test.cc
mfrancepillois: No, you are right. I missed that during the rebased. I,ve removed the test `RaggedDotNotBecomesFusion` which was redundant.

xla/backends/gpu/transforms/cublas_gemm_rewriter_test.cc
mfrancepillois: Tests have been renamed.
`ragged-dot` that cannot be lowered to `__cublas$lt$groupedMatmul` should have been caught earlier in the pipeline (these cases are tested in `ragged_dot_rewriter_test.cc`).

xla/backends/gpu/runtime/gpublas_lt_matmul_thunk.cc
mfrancepillois: Sorry, these changes came back during the rebase but have been removed.

xla/backends/gpu/runtime/gpublas_lt_matmul_thunk.h
mfrancepillois: Same. This has been been removed

xla/backends/gpu/transforms/gemm_rewriter_test_lib.cc
mfrancepillois: This has been updated

xla/backends/gpu/transforms/cublas_gemm_rewriter_test.cc
mfrancepillois: You're right. You do not need it as we already check `SkipGroupedGemmTest` in the `SetUp`.

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: Thanks. This has been fixed.

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: This function has been split as suggested.

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: The first TODO will be addressed in a following PR when the Group-Gemm support is fully merged.

The second TODO has been removed. I think this improvement is not really needed.

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: This is a Hipblaslt specific limitation/feature.
The documentation mentions this `setOrderA/B` API but in practice it does not appear to work. Matrices are expected to be in COLUMN-MAJOR order anyway.
This comment was added during the development phase, but is no longer needed, because even if matrix order could be changed with this API, this would require many changes in the way pointers and strides are calculated. I have therefore changed the comment so that it only states the expected order of matrices.

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: The workspace is defined for a specific operation, while the plan is defined for the “signature” of an operation (data types, operands sizes, …). So, we can execute different operations using the same plan (i.e., calling multiple times ExecuteOnStream with the same plan but different workspaces and operands).

xla/stream_executor/rocm/rocm_helpers.cu.cc
mfrancepillois: This has been changed.

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: This has been changed

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: My apologies. It was an old comment that I forgot to delete.

xla/stream_executor/rocm/hip_blas_lt.h
mfrancepillois: You’re right. Generic methods in `gpu_blas_lt.cc` are enough.

xla/stream_executor/rocm/hip_blas_lt.cc
mfrancepillois: Thanks for the suggestion. The code has been changed this way.

xla/stream_executor/rocm/rocm_helpers.cu.cc
mfrancepillois: First point: Theads other than the last one do not read cumulative_offset in the first loop iteration, but only in the subsequent loop. Therefore, all threads will be synchronized at the end of the first loop (__barrier) before other threads will read cumulative_offset. So, ` cumulative_offset` will be initialized when other threads will read it.

Second point: I agree. The access ordering was not guaranteed. The ` cumulative_offset += offset_in_batch + group_size;` has been moved between the barrier to ensure that all threads get the value relative to the current loop iteration before updating it.

xla/stream_executor/rocm/hip_blas_lt.cc
dextero: Thank you for the clarification. I don't really see a better way right now, so let's leave it as is.

xla/stream_executor/rocm/rocm_helpers.cu.cc
dextero: LGTM, thank you!

dextero (approved):
Looks great. Thank you for the contribution, and for the exhaustive test suite!

dextero (approved):

xla/backends/gpu/transforms/gemm_rewriter_test_lib.cc
```diff
@@ -71,6 +71,12 @@ bool GemmRewriteTestBase::SkipGpuBlasLtTest() {
          GetDebugOptionsForTest().xla_gpu_enable_cublaslt();
 }
 
+bool GemmRewriteTestBase::SkipGroupedGemmTest() {
+  // Grouped GEMM is only supported on ROCm with hipBLASLt on gfx942 or gfx950
+  return !Capability().rocm_compute_capability()->has_hipblaslt() ||
+         !Capability().rocm_compute_capability()->gfx9_mi300_series();
```
dextero: I missed this before. I think this needs to be something like `!has_hipblas || (has_hipblas && !gfx9_mi300_series)`? Right now we end up running the tests on NVIDIA gpus and that doesn't quite work:


```
[ RUN      ] GroupedGemmRewriteTest.CustomCallTargetGroupedGemmLargeGroupCount600
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1774454599.812343      27 service.cc:154] XLA service 0x565487604a90 initialized for platform CUDA (this does not guarantee that XLA will be used). Devices:
I0000 00:00:1774454599.824899      27 service.cc:170]   StreamExecutor [0]: Tesla T4, Compute Capability 7.5 (Driver: 13.0.0[580.82.7]; Runtime: 12.9.0; Toolkit: 12.9.0; DNN: 9.8.0)
I0000 00:00:1774454599.837066      27 se_gpu_pjrt_client.cc:1505] Using BFC allocator.
I0000 00:00:1774454599.839518      27 gpu_helpers.cc:144] XLA backend will use up to 11727814656 bytes on device 0 for BFCAllocator.
I0000 00:00:1774454599.844655      27 gpu_helpers.cc:183] XLA backend will use up to 3909271552 bytes on device 0 for CollectiveBFCAllocator.
I0000 00:00:1774454599.894047      27 cuda_dnn.cc:461] Loaded cuDNN version 90800
/b/f/w/bazel-out/k8-opt/bin/xla/backends/gpu/transforms/cublas_gemm_rewriter_test_nvgpu_any.runfiles/xla/build_tools/ci/parallel_gpu_execute: line 74:    27 Segmentation fault      (core dumped) "$TEST_BINARY" $@
```

mfrancepillois: Thanks for noticing! In fact, we also need `IsCuda()` in the condition because `.rocm_compute_capability()` return null if it is not a rocm target. This has been fixed.

vwbaker (approved):

xla/backends/gpu/transforms/cublas_gemm_rewriter_test.cc
vwbaker: nit; should be CustomCallTarget*s* with an s to make it an active verb for custom call.

Also, just FYI: https://github.com/openxla/xla/pull/39853. I think it has the same effect that you're going for, but just a warning since you've been updating this code.

xla/backends/gpu/transforms/gemm_rewriter_test_lib.cc
vwbaker: what about has_rocm_compute_capability()?

mfrancepillois: I cannot find such an API in `stream_executor`. `IsRocm` and `IsCuda` nevertheless serve the same purpose https://github.com/openxla/xla/blob/c913e9c3c4d2644fc10ce78928146fb3a4c110d3/xla/stream_executor/device_description.h#L127-L133 . As `GemmRewriteTestBase::SkipGpuBlasLtTest()` is using `IsCuda`   I’ve opted for that over `IsRocm`, but I can change it if you’d like.

vwbaker: no lgtm

vwbaker:
looks like DWYU needs a fix: 

buildozer 'add deps //xla/stream_executor:memory_allocation' //xla/stream_executor/rocm:amdhipblaslt_plugin

dextero (approved):

PR #38777, opened 2026-03-05
[ROCm] Add error checking and status propagation to ROCm profiler

magaonka-amd:
📝 Summary of Changes
- Add RETURN_IF_ROCPROFILER_ERROR macro for toolInit() calls
- Change Enable() to return absl::Status instead of void, propagating rocprofiler_start_context errors to the caller
- Update GpuTracer::DoStart() to check Enable() return via TF_RETURN_IF_ERROR

🎯 Justification
Currently most of rocm profiler calls are fire and forget type where return status from it is not checked so if some steps fail profiler continues with invalid data which could lead to undefined behavior. This will leave user with no info on what happened so adding changes to mitigate it. While building this we need to take how rocprof-sdk returns errors.

🚀 Kind of Contribution
🐛 Bug Fix,


magaonka-amd:
Hello @cj401-amd , this patch adds error checking for rocm profiler steps. In my testing this adds near zero overhead to profiler. I think it is nice to have feature where we know if profiler steps fail they are reported correctly. I would love to hear from you about this at your convenience.

cj401-amd:
> Hello @cj401-amd , this patch adds error checking for rocm profiler steps. In my testing this adds near zero overhead to profiler. I think it is nice to have feature where we know if profiler steps fail they are reported correctly. I would love to hear from you about this at your convenience.

@magaonka-amd, thanks for adding this up.

i-chaochen (commented):
Hi @magaonka-amd can you create a PR to rocm/xla:main first?

xla/backends/profiler/gpu/rocm_tracer.cc
```diff
@@ -48,6 +49,18 @@ limitations under the License.
 #include "xla/tsl/profiler/backends/cpu/annotation_stack.h"
 #include "tsl/platform/abi.h"
 
+#define RETURN_IF_ROCPROFILER_ERROR(expr)                             \
```
dimitar-asenov: Please avoid defining custom alternatives to `RETURN_IF_ERROR`/`ASSIGN_OR_RETURN`. If the `rocprofiler_status_t` is incompatible with the existing macros, implement a function that converts it to a `absl::Status`.

dimitar-asenov (changes requested):

xla/backends/profiler/gpu/rocm_tracer.cc
magaonka-amd: Sorry for late response. I updated my patch. I added conversion function to avoid custom error handling.

also my commit shows 2 unrelated DWYU fails.  CI check somehow can't resolve if DWYU dependency in conditional branching. If I try to resolve it, I get hit with duplicate dependency error from bazel.

Please review at your convenience

dimitar-asenov (approved):

PR #38779, opened 2026-03-05
[XLA:GPU] Add a VMM based memory allocator. 

shawnwang18:
# Add `DeviceAddressVmmAllocator`: VMM-based device memory allocator

## Motivation

XLA currently uses `cuMemAlloc`-style allocation, which ties virtual address reservation and physical memory allocation together as a single indivisible operation. This prevents the physical memory backing a buffer from being re-mapped under a different virtual address — a capability required for features like command buffer reuse where the same compiled kernel graph must be replayed with different input buffers mapped at the same VA range.

This PR introduces `DeviceAddressVmmAllocator`, a new `DeviceAddressAllocator` implementation built on CUDA Virtual Memory Management (VMM) APIs. It separates physical memory creation from virtual address reservation, making it possible for callers to remap physical allocations to arbitrary virtual addresses independently.

We need to introduce a separate memory allocator, mainly because VMM CUDA APIs have the following two restrictions which other allocators do not handle:

### Restriction 1: `cuMemMap` does not support a non-zero offset into the physical allocation

The CUDA VMM `cuMemMap(vaPtr, size, offset, handle, ...)` call has the `offset` parameter but in practice it must be zero — you cannot remap an interior slice of a physical allocation to an arbitrary VA location. This means each physical allocation must be mapped starting at offset 0. Therefore the allocator must create one physical allocation per logical buffer (via `cuMemCreate`) rather than sub-allocating from a shared slab. This is why `DeviceAddressVmmAllocator` stores one `CudaRawMemoryAllocation` per buffer rather than using a traditional bump/pool allocator.

### Restriction 2: All accesses to a pointer must complete before `cuMemFree`

`cuMemFree` is synchronous with respect to device execution — if you call it while a GPU kernel is still reading that memory, you get corruption or undefined behavior. The ordinary allocator call path in XLA calls `Deallocate()` from the CPU immediately after submitting work, so the GPU may still be running. This allocator supports asynchronous deallocation: when Deallocate() is called, it records a GPU timeline write on the device's stream and defers the actual deallocation until the GPU reaches that point in the stream. This allows callers to deallocate memory while device kernels may still be consuming the data.

## Design

The allocator decomposes each allocation into three distinct objects, each with its own RAII lifetime:

1. `CudaRawMemoryAllocation` — wraps a CUDA physical memory handle created via `cuMemCreate`. Has no associated virtual address; exposes the raw handle via `address().opaque()`. Size is rounded up to the VMM allocation granularity (typically 2 MB).
2. `CudaMemoryReservation` — wraps a virtual address range reserved via `cuMemAddressReserve`. The VA range is not accessible until a physical allocation is mapped into it.
3. `MemoryReservation::ScopedMapping` — an RAII token returned by `MemoryReservation::MapTo()`. Its constructor calls `cuMemMap` + `cuMemSetAccess` to make the physical memory accessible at the reserved VA. Its destructor calls `cuMemUnmap` to unmap it, ensuring the VA range and physical memory handles can be reused or released independently.

`DeviceAddressVmmAllocator` owns all three objects per live allocation and destroys them in the correct order on deallocation: unmap first, release VA reservation second, release physical memory last.

## 🚀 Kind of contribution

✨ New Feature

## 🧪 Unit tests

`xla/stream_executor/stream_executor_vmm_allocator_test.cc`

### DIFF
--- a/xla/stream_executor/BUILD
+++ b/xla/stream_executor/BUILD
@@ -323,6 +323,57 @@ cc_library(
 )
 
 cc_library(
+    name = "stream_executor_vmm_allocator",
+    srcs = ["stream_executor_vmm_allocator.cc"],
+    hdrs = ["stream_executor_vmm_allocator.h"],
+    deps = [
+        ":device_address",
+        ":device_address_allocator",
+        ":memory_allocation",
+        ":memory_reservation",
+        ":platform",
+        ":stream",
+        ":stream_executor_h",
+        "//xla/tsl/platform:statusor",
+        "@com_google_absl//absl/base:core_headers",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/container:flat_hash_set",
+        "@com_google_absl//absl/log",
+        "@com_google_absl//absl/log:check",
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/status:statusor",
+        "@com_google_absl//absl/strings:str_format",
+        "@com_google_absl//absl/synchronization",
+        "@com_google_absl//absl/time",
+        "@com_google_absl//absl/types:span",
+    ],
+)
+
+xla_cc_test(
+    name = "stream_executor_vmm_allocator_test",
+    srcs = ["stream_executor_vmm_allocator_test.cc"],
+    tags = [
+        "cuda-only",
+        "gpu",
+        "requires-gpu-nvidia",
+    ],
+    deps = [
+        ":device_address",
+        ":platform",
+        ":platform_manager",
+        ":stream",
+        ":stream_executor_h",
+        ":stream_executor_vmm_allocator",
+        "//xla/stream_executor/cuda:cuda_device_address_vmm_allocator",
+        "//xla/stream_executor/cuda:cuda_platform",
+        "//xla/tsl/platform:statusor",
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/status:status_matchers",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
+
+cc_library(
     name = "launch_dim",
     srcs = ["launch_dim.cc"],
     hdrs = ["launch_dim.h"],
--- a/xla/stream_executor/BUILD
+++ b/xla/stream_executor/BUILD
@@ -323,9 +323,9 @@ cc_library(
 )
 
 cc_library(
-    name = "stream_executor_vmm_allocator",
-    srcs = ["stream_executor_vmm_allocator.cc"],
-    hdrs = ["stream_executor_vmm_allocator.h"],
+    name = "vmm_device_address_allocator",
+    srcs = ["vmm_device_address_allocator.cc"],
+    hdrs = ["vmm_device_address_allocator.h"],
     deps = [
         ":device_address",
         ":device_address_allocator",
@@ -346,30 +346,6 @@ cc_library(
         "@com_google_absl//absl/synchronization",
         "@com_google_absl//absl/time",
         "@com_google_absl//absl/types:span",
-    ],
-)
-
-xla_cc_test(
-    name = "stream_executor_vmm_allocator_test",
-    srcs = ["stream_executor_vmm_allocator_test.cc"],
-    tags = [
-        "cuda-only",
-        "gpu",
-        "requires-gpu-nvidia",
-    ],
-    deps = [
-        ":device_address",
-        ":platform",
-        ":platform_manager",
-        ":stream",
-        ":stream_executor_h",
-        ":stream_executor_vmm_allocator",
-        "//xla/stream_executor/cuda:cuda_device_address_vmm_allocator",
-        "//xla/stream_executor/cuda:cuda_platform",
-        "//xla/tsl/platform:statusor",
-        "@com_google_absl//absl/status",
-        "@com_google_absl//absl/status:status_matchers",
-        "@com_google_googletest//:gtest_main",
     ],
 )
 
--- a/xla/stream_executor/stream_executor_vmm_allocator.cc
+++ b/xla/stream_executor/stream_executor_vmm_allocator.cc
@@ -1,0 +1,456 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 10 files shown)


xla/stream_executor/stream_executor_vmm_allocator.h
```diff
@@ -0,0 +1,174 @@
...
+// The allocator tracks the ScopedMapping and underlying MemoryAllocation and
+// MemoryReservation objects for each returned DeviceAddressBase. Callers can
+// retrieve these via GetRawAllocation() and GetReservation().
+//
+// This allocator supports asynchronous deallocation: when Deallocate() is
+// called, it records an event on the stream and defers the actual deallocation
+// until the event completes. This allows callers to deallocate memory while
+// device kernels may still be consuming the data.
+//
+// This allocator is NOT thread-safe. All operations must be performed from a
```
ezhulenev: AFAIK All other allocators are thread safe, and if we want to swap XLA allocation to this one by default it must be thread safe as well?

xla/stream_executor/vmm_device_address_allocator.h
```diff
@@ -0,0 +1,235 @@
...
+// allows callers to deallocate memory while device kernels may still be
+// consuming the data.
+//
+// Requires compute capability >= 7.0 (Volta and later) for
+// cuStreamWriteValue64 support. Use Create() to obtain an instance; it
+// returns an error if the device does not meet this requirement.
+//
+// This allocator is thread-safe for concurrent use by multiple threads across
+// any registered devices.
+class DeviceAddressVmmAllocator : public DeviceAddressAllocator {
```
ezhulenev, line 77: Because it's very CUDA-specific, let's move it to `cuda_vmm_address_allocator.{h,cc}` under cuda folder. See https://github.com/openxla/xla/pull/39021/changes#diff-5cc5aedb2b5522a7b9894458648e8862689e4e28fb01b8c8ba3a8be4c09c5139, I'm adding `cuda_vmm_allocator.{h,cc}` for physical alocations. Eventually we want to build address allocators on top of raw memory allocators.

xla/stream_executor/stream_executor_vmm_allocator.h
ezhulenev: NVM, this was for an older PR version.

xla/stream_executor/stream_executor_vmm_allocator.cc
```diff
@@ -0,0 +1,556 @@
...
+    state.pending_deallocations.pop_front();
+  }
+}
+
+// ABSL_NO_THREAD_SAFETY_ANALYSIS because clang's thread-safety analysis cannot
+// reason through manual Unlock()/Lock() pairs. The declaration in the header
+// retains no ABSL_EXCLUSIVE_LOCKS_REQUIRED annotation to reflect that this
+// method manages locking internally.
+void DeviceAddressVmmAllocator::WaitPendingDeallocationsToComplete(
+    PerDeviceState& state, uint64_t size) ABSL_NO_THREAD_SAFETY_ANALYSIS {
```
ezhulenev: I'm worried about `ABSL_NO_THREAD_SAFETY_ANALYSIS`, are you sure it's a clang problem here and not real bugs? Any other ways to rewrite thus function to keep thread annotation?

xla/stream_executor/vmm_device_address_allocator.h
shawnwang18: Yes, let's move to cuda folder. 

I'd like to build `cudaDeviceAddressVmmAllocator` upon `cudaMemoryReservation`, and `cudaRawAllocation` wrapper,  these wrappers have already wraps the vmm allocation APIs. 

And also the allocator needs to provide the API for fetching the `MemoryAllocation` handle for the reservation, so  I think we need built it through Reservation/Allocation APIs.

shawnwang18: fixed

xla/stream_executor/BUILD
```diff
@@ -322,6 +322,57 @@ cc_library(
...
+        "@com_google_absl//absl/time",
+        "@com_google_absl//absl/types:span",
+    ],
+)
+
+xla_cc_test(
+    name = "stream_executor_vmm_allocator_test",
+    srcs = ["stream_executor_vmm_allocator_test.cc"],
+    tags = [
+        "cuda-only",
```
ezhulenev: move this to cuda folder as it requires cuda platform?

xla/stream_executor/vmm_device_address_allocator.h
```diff
@@ -0,0 +1,248 @@
...
+//
+// Concrete subclasses implement the platform-specific virtual methods
+// (InitializeDeviceState, CreateAllocation, CreateReservation,
+// EnqueueDeferredDeallocation) and expose platform-specific Create() factories.
+// Subclasses must also set PerDeviceState::destroy_fn in InitializeDeviceState
+// to release platform-specific resources (e.g. pinned timeline memory).
+//
+// This allocator is thread-safe for concurrent use by multiple threads across
+// any registered devices.
+class DeviceAddressVmmAllocator : public DeviceAddressAllocator {
```
ezhulenev, line 77: Rename file to vmm_address_allocator, and type to VmmAddressAllocator (or add vmm_device.../VmmDevice...), no need to spell stream_executor here, we are already in stream_executor folder.

shawnwang18: fixed.

xla/stream_executor/BUILD
shawnwang18: fixed.

ezhulenev (approved):
LGTM. It doesn't add of change any public APIs to StreamExecutor, so I think it should be ok to submit it and test it on real workloads. It looks correct, but it's too complex to be able to tell fore sure that it covers all edge cases.

akuegel:
@beckerhe Eugene already reviewed this, bug maybe you also want to take a look?

beckerhe (approved):

seantalts:
Sorry for the delay, my first week on rotation. Would you mind removing the unused dep on compiler/xla/stream_executor:device_address  from the target stream_executor/cuda:cuda_device_address_vmm_allocator ? Thank you

shawnwang18:
> Sorry for the delay, my first week on rotation. Would you mind removing the unused dep on compiler/xla/stream_executor:device_address from the target stream_executor/cuda:cuda_device_address_vmm_allocator ? Thank you

Thanks, just did the update, could you help approve and submit the CI flow?

akuegel (approved):

PR #34572, opened 2025-11-28
[ROCM] Added command buffers support for convolutions 2nd attempt

pemeliya:
Added CommandBuffer support for Convolution ops
Graph capture of convolutions is by default disabled (since not all convolutions can be captured). 
It can be enabled through a newly added flag: '**--xla_gpu_enable_command_buffers=+convolution**"

🎯 Justification
This op was missing for whatever reason: this results in graph fragmentation especially for large models. Hence one gets several (sometimes many) execution graphs instead of just one.

🚀 Kind of Contribution
✨ New Feature

🧪 Unit Tests:
Added new subtest to xla/service/gpu/transforms/command_buffer_scheduling_test.cc and xla/backends/gpu/runtime/command_buffer_conversion_pass_test.cc

This is a fork of the original PR which was reverted due to conv problems: https://github.com/openxla/xla/pull/32053
Now convolutions capture is by default disabled.

@beckerhe , @ezhulenev, @dimitar-asenov  please have another look !

PR #38802, opened 2026-03-06
[ROCM] improving RunAndCompareThreeIterations in command_buffer_test

pemeliya:
📝 Summary of Changes
Rewirting RunAndCompareThreeIterations function

🎯 Justification
The above function clones input literals with the idea that this will cause command buffer updates but this does not really happen.
The point is, the function [ExecuteWithExecutable](https://github.com/openxla/xla/blob/5e9258842fe13c6c155684d7263604d3d2963380/xla/service/hlo_runner_pjrt.cc#L509-L512) creates and destroys device buffers on every call, hence just cloning the input literals won't help to force command buffer update. This is because BFC allocator used by PJRT runner would likely to allocate device buffers on exact same locations as before. So, I had to rewrite this function to really keep 2 sets of input arguments allocated at the same time.
Though, to get it working I had to use a small hack to [downcast runner pointer](https://github.com/openxla/xla/blob/5e9258842fe13c6c155684d7263604d3d2963380/xla/service/gpu/tests/command_buffer_test.cc#L157) to HloRunnerPjRt type..

🚀 Kind of Contribution
 🧪 Tests

Extracted as a separate PR from https://github.com/openxla/xla/pull/34572

### DIFF
--- a/xla/service/gpu/tests/command_buffer_test.cc
+++ b/xla/service/gpu/tests/command_buffer_test.cc
@@ -157,7 +157,7 @@ class CommandBufferTest
         auto exec, CreateExecutable(std::move(module), run_hlo_passes));
 
     auto* pjrt_runner = tsl::down_cast<HloRunnerPjRt*>(&test_runner());
-    EXPECT_TRUE(pjrt_runner != nullptr);
+    ASSERT_TRUE(pjrt_runner != nullptr);
 
     // Create two copies of device buffers to make sure command buffer saved
     // pointers really get updated during the last "update run".
@@ -171,7 +171,7 @@ class CommandBufferTest
 
     static constexpr absl::string_view kPhases[] = {"warm-up", "create",
                                                     "update"};
-    for (int i = 0; i < 3; i++) {
+    for (size_t i = 0; i < std::size(kPhases); i++) {
       BufferSet current_set = (i < 2) ? kInitial : kUpdated;
       TF_ASSERT_OK_AND_ASSIGN(auto output_buffers,
                               pjrt_runner->ExecuteWithDeviceBuffers(
--- a/xla/service/gpu/tests/command_buffer_test.cc
+++ b/xla/service/gpu/tests/command_buffer_test.cc
@@ -36,6 +36,7 @@ limitations under the License.
 #include "xla/literal_util.h"
 #include "xla/service/hlo_module_config.h"
 #include "xla/service/hlo_runner_interface.h"
+#include "xla/service/hlo_runner_pjrt.h"
 #include "xla/service/platform_util.h"
 #include "xla/stream_executor/device_address.h"
 #include "xla/stream_executor/platform_manager.h"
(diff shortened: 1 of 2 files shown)


xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -130,68 +131,57 @@ class CommandBufferTest
...
-        exec.get(), absl::MakeSpan(cloned_arg_ptrs));
-    if (!r3.ok()) return ::testing::AssertionFailure() << r3.status();
-    if (!LiteralTestUtil::NearOrEqual(*reference, *r3, error))
-      return ::testing::AssertionFailure() << "Mismatch on update run";
-
-    return ::testing::AssertionSuccess();
+    for (int i = 0; i < 3; i++) {
+      TF_ASSERT_OK_AND_ASSIGN(auto output_buffers,
+                              pjrt_runner->ExecuteWithDeviceBuffers(
+                                  exec.get(), argument_handles[i < 2 ? 0 : 1]));
```
apivovarov: Lets use 
```
enum BufferSet { kInitial = 0, kUpdated = 1 };
BufferSet current_set = (i < 2) ? kInitial : kUpdated;
```

xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -130,68 +131,57 @@ class CommandBufferTest
...
+
+      TF_ASSERT_OK_AND_ASSIGN(auto result,
+                              pjrt_runner->TransferLiteralsFromDevice(
+                                  output_buffers, untuple_results));
+      if (!LiteralTestUtil::NearOrEqual(reference, result, error)) {
+        ::testing::AssertionFailure() << "Mismatch on "
```
apivovarov: This will not fail the test. Creating an AssertionFailure object without returning it or passing it to a macro does nothing. Use EXPECT_TRUE with a custom message so the test actually fails and reports the specific iteration.
```
EXPECT_TRUE(LiteralTestUtil::NearOrEqual(reference, result, error))
          << "Mismatch on " << phase_name << " run (iteration " << i << ")";
```

xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -130,68 +131,57 @@ class CommandBufferTest
...
+                                      << (i == 0   ? "warm-up run"
+                                          : i == 1 ? "create run"
+                                                   : "update run");
```
apivovarov: To improve readability we can get phase_name as the following
```
static constexpr absl::string_view kPhases[] = {"warm-up", "create", "update"};
absl::string_view phase_name = kPhases[i];
```

pemeliya:
Thanks for the suggestions, it makes sense

xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -130,68 +131,58 @@ class CommandBufferTest
...
-    std::vector<const Literal*> cloned_arg_ptrs;
-    cloned_arg_ptrs.reserve(arg_ptrs.size());
-    for (const Literal* a : arg_ptrs) {
-      cloned_args_storage.push_back(a->Clone());
-      cloned_arg_ptrs.push_back(&cloned_args_storage.back());
+    TF_ASSERT_OK_AND_ASSIGN(
+        auto exec, CreateExecutable(std::move(module), run_hlo_passes));
+
+    auto* pjrt_runner = tsl::down_cast<HloRunnerPjRt*>(&test_runner());
+    EXPECT_TRUE(pjrt_runner != nullptr);
```
apivovarov: `EXPECT_TRUE` is a "non-fatal" failure. If the downcast returns nullptr, the test will record a failure but continue executing. The very next line calls pjrt_runner->TransferLiteralsToDefaultDevice, which will cause a segmentation fault and crash the entire test suite. ASSERT_TRUE / ASSERT_NE ensures the test stops immediately

xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -130,68 +131,58 @@ class CommandBufferTest
...
-    absl::StatusOr<Literal> r3 = test_runner().ExecuteWithExecutable(
-        exec.get(), absl::MakeSpan(cloned_arg_ptrs));
-    if (!r3.ok()) return ::testing::AssertionFailure() << r3.status();
-    if (!LiteralTestUtil::NearOrEqual(*reference, *r3, error))
-      return ::testing::AssertionFailure() << "Mismatch on update run";
-
-    return ::testing::AssertionSuccess();
+    static constexpr absl::string_view kPhases[] = {"warm-up", "create",
+                                                    "update"};
+    for (int i = 0; i < 3; i++) {
```
apivovarov: Nit: `i < std::size(kPhases)`

apivovarov (approved):

xla/service/gpu/tests/command_buffer_test.cc
```diff
@@ -36,6 +36,7 @@ limitations under the License.
 #include "xla/literal_util.h"
 #include "xla/service/hlo_module_config.h"
 #include "xla/service/hlo_runner_interface.h"
+#include "xla/service/hlo_runner_pjrt.h"
```
thomasjoerg, line 39: Please update the BUILD deps accordingly

thomasjoerg (changes requested):

thomasjoerg (approved):

PR #38836, opened 2026-03-06
[ROCm] Add minimal ROCm support to ExecutableAbiVersion

magaonka-amd:
Hot-Fix for breaking changes added in 5150bfd950

📝 Summary of Changes
The ExecutableAbiVersion::FromDeviceDescription was only implemented for CUDA, returning UNIMPLEMENTED for all other platforms. This broke every ROCm compilation path after 5150bfd950 introduced mandatory ABI version derivation in GpuCompiler::RunBackend.

Add a CreateForRocm handler that returns a minimal ABI version with
only platform_name set to "ROCm" and no platform-specific version
info. This preserves pre-existing ROCm behavior (no compatibility
checks) until proper ABI versioning is designed for ROCm.

🎯 Justification
without this change JAX tests on ROCm platform fail with error:
```bash
jax.errors.JaxRuntimeError: UNIMPLEMENTED: Deriving the executable ABI version from the device description is not implemented for the target platform.
```
🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
added rocm test to test //xla/stream_executor/abi:executable_abi_version_test

### DIFF
--- a/xla/stream_executor/abi/BUILD
+++ b/xla/stream_executor/abi/BUILD
@@ -41,6 +41,7 @@ xla_cc_test(
         "//xla/stream_executor:device_description",
         "//xla/stream_executor:semantic_version",
         "//xla/stream_executor/cuda:cuda_compute_capability",
+        "//xla/stream_executor/rocm:rocm_compute_capability",
         "//xla/tsl/util/proto:proto_matchers",
         "@com_google_googletest//:gtest_main",
     ],
--- a/xla/stream_executor/abi/executable_abi_version.cc
+++ b/xla/stream_executor/abi/executable_abi_version.cc
@@ -39,6 +39,16 @@ static absl::StatusOr<ExecutableAbiVersion> CreateForCuda(
   return ExecutableAbiVersion::FromProto(std::move(proto));
 }
 
+// Returns a minimal ABI version for ROCm with no platform-specific version
+// info. Compatibility checks will treat this as always-compatible, preserving
+// pre-existing ROCm behavior until proper ABI versioning is designed.
+static absl::StatusOr<ExecutableAbiVersion> CreateForRocm(
+    const DeviceDescription& /*device_description*/) {
+  ExecutableAbiVersionProto proto;
+  proto.set_platform_name("ROCm");
+  return ExecutableAbiVersion::FromProto(std::move(proto));
+}
+
 absl::StatusOr<ExecutableAbiVersion> ExecutableAbiVersion::FromProto(
     const ExecutableAbiVersionProto& proto) {
   return ExecutableAbiVersion(proto);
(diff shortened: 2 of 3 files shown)


draganmladjenovic:
I'm not sure this change is correct for us. @beckerhe What is the aim of executable ABI version? For cuda I don't see enough info to for example make sure that the autotuned algos are valid when run on different cublas/cublaslt etc. I see that on cuda we use compile-time versions of runtime/cuddn/cub libraries. Is it because there might be ifdefs in the code using those?

@magaonka-amd If this is critical. I'd rather go with empty abi version and runtime version whose IsCompatible impl returns true. That would still preserve the old behavior until we figure our what abi version check is supposed to guard against.

magaonka-amd:
Thank you for the review. That's really good point @draganmladjenovic . Yes going with empty abi version now is better approach. There is a second order effect to my fix when someone adds compatibility checks it will cause issues.  Fix is critical for JAX CI, I have made suggested changes.
We can add follow up fixes when we have more clarity on how we want to handle this for ROCm.

nurmukhametov:
Do we need a similar change for RuntimeAbiVersion https://github.com/openxla/xla/commit/418a49d8447780761896ef89c0942ab70ea8060a?

magaonka-amd:
@nurmukhametov  right now https://github.com/openxla/xla/commit/418a49d8447780761896ef89c0942ab70ea8060a is not a breaking change because ROCm doesn't call GetRuntimeAbiVersion().
But yes we will need this as well but not needed as part of this fix. we can add separate PR for it. agree?

nurmukhametov:
> @nurmukhametov right now [418a49d](https://github.com/openxla/xla/commit/418a49d8447780761896ef89c0942ab70ea8060a) is not a breaking change because ROCm doesn't call GetRuntimeAbiVersion(). But yes we will need this as well but not needed as part of this fix. we can add separate PR for it. agree?

Yeah, sure, it is fine with me.

beckerhe:
> I'm not sure this change is correct for us. @beckerhe What is the aim of executable ABI version? For cuda I don't see enough info to for example make sure that the autotuned algos are valid when run on different cublas/cublaslt etc. I see that on cuda we use compile-time versions of runtime/cuddn/cub libraries. Is it because there might be ifdefs in the code using those?
> 
> @magaonka-amd If this is critical. I'd rather go with empty abi version and runtime version whose IsCompatible impl returns true. That would still preserve the old behavior until we figure our what abi version check is supposed to guard against.

The main purpose of the ABI system is for infrastructure that uses AOT compiled PjRt executables to determine whether a previously compiled executable is still compatible with the current version of the runtime.

There are 2 types. ExecutableAbiVersion and RuntimeAbiVersion. The latter type implements all the logic about checking compatibility. The former should record all the version that might become relevant in the future to detect a breaking change.

That all being said, there is currently no plan to make XLA or JAX directly use these new APIs. We have some infrastructure internally that's using it and everyone is of course free to build something that's using it as well.

The chang in itself looks fine to me, even though you probably want to record some library version numbers so that you can detect breaking changes that might occur (intentionally or unintentionally) in the future.

Let me know how you wanna proceed, I'm happy to approve the PR in the current form. But feel also free to close it or to add some version fields.

magaonka-amd:
Thanks @beckerhe . I feel we can merge this PR and unblock JAX CI and then I can figure out how we want to handle ABI versioning in ROCm and make follow up fixes.

beckerhe (approved):

PR #38921, opened 2026-03-09
[xla] Add Bant macro expansions for XLA test targets

ezhulenev:
This allows buildcleaner-like analysis of XLA tests to detect missing dependencies.

### DIFF
--- a/.bant-macros
+++ b/.bant-macros
@@ -1,0 +1,10 @@
+# Bant macro expansions for XLA custom rules.
+#
+# Syntax: https://github.com/hzeller/bant?tab=readme-ov-file#macros.
+#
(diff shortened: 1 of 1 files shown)


ermilovmaxim (changes requested):
Since the PR adds bant specific file to codebase, could you add usage examples to documentation?

ezhulenev:
Added a link to XLA documentation

ezhulenev:
@ermilovmaxim ping

ermilovmaxim (approved):

ddunl:
I think this doesn't import because we enumerate top level files in the Copybara config, I'll try to patch it in this afternoon

PR #38943, opened 2026-03-09
[XLA:GPU] Add Put, Signal, and WaitSignal APIs to Communicator for NCCL

sfvaroglu:
📝 Summary of Changes
Adds the Put, Signal, and WaitSignal APIs to the Collectives/Communicator so the NCCL one-sided put path (e.g. ragged all-to-all with symmetric buffers) can be implemented on top of the existing CollectiveMemory and SymmetricMemory APIs.

🎯 Justification
Declare Put/Signal/WaitSignal on the communicator so the NCCL one-sided put path (e.g. ragged all-to-all with symmetric buffers) can be implemented later. Stub implementations only on the NCCL path.

🚀 Kind of Contribution
✨ New Feature/API

N/A

🧪 Unit Tests:
N/A

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/backends/gpu/collectives/gpu_communicator.h
+++ b/xla/backends/gpu/collectives/gpu_communicator.h
@@ -189,11 +189,14 @@ class GpuCommunicator : public Communicator {
     return Unimplemented("LaunchPut is not implemented");
   }
 
-  virtual absl::Status LaunchSignal(const Executor& executor) {
+  virtual absl::Status LaunchSignal(RankId peer, const SignalDesc& desc,
+                                    const Executor& executor) {
     return Unimplemented("LaunchSignal is not implemented");
   }
 
-  virtual absl::Status LaunchWaitSignal(RankId peer, const Executor& executor) {
+  virtual absl::Status LaunchWaitSignal(RankId peer, int op_cnt,
+                                        const SignalDesc& desc,
+                                        const Executor& executor) {
     return Unimplemented("LaunchWaitSignal is not implemented");
   }
 };
--- a/xla/backends/gpu/collectives/gpu_communicator.h
+++ b/xla/backends/gpu/collectives/gpu_communicator.h
@@ -181,6 +181,21 @@ class GpuCommunicator : public Communicator {
   virtual absl::Status LaunchRecv(se::DeviceAddressBase recv_buffer,
                                   PrimitiveType dtype, size_t count,
                                   RankId peer, const Executor& executor) = 0;
+
+  virtual absl::Status LaunchPut(se::DeviceAddressBase send_buffer,
+                                 SymmetricMemory* recv_buffer, size_t offset,
+                                 size_t count, RankId peer,
+                                 const Executor& executor) {
+    return Unimplemented("LaunchPut is not implemented");
+  }
+
+  virtual absl::Status LaunchSignal(const Executor& executor) {
+    return Unimplemented("LaunchSignal is not implemented");
+  }
+
+  virtual absl::Status LaunchWaitSignal(RankId peer, const Executor& executor) {
+    return Unimplemented("LaunchWaitSignal is not implemented");
+  }
 };
 
 }  // namespace xla::gpu
--- a/xla/backends/gpu/collectives/gpu_communicator.h
+++ b/xla/backends/gpu/collectives/gpu_communicator.h
@@ -189,13 +189,13 @@ class GpuCommunicator : public Communicator {
     return Unimplemented("LaunchPut is not implemented");
   }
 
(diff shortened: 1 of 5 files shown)


xla/core/collectives/communicator.h
```diff
@@ -161,6 +162,20 @@ class Communicator {
     return Unimplemented("One-way recv is not implemented");
   }
 
+  virtual Future<> Put(se::DeviceAddressBase send_buffer,
```
ezhulenev, line 173: A little bit of documentation will help, just a line or two to explain what it does and how it's different from Send. Same for Signal and WaitSignal

xla/backends/gpu/collectives/gpu_communicator.h
```diff
@@ -181,6 +181,21 @@ class GpuCommunicator : public Communicator {
...
+                                 size_t count, RankId peer,
+                                 const Executor& executor) {
+    return Unimplemented("LaunchPut is not implemented");
+  }
+
+  virtual absl::Status LaunchSignal(const Executor& executor) {
+    return Unimplemented("LaunchSignal is not implemented");
+  }
+
+  virtual absl::Status LaunchWaitSignal(RankId peer, const Executor& executor) {
```
Tixxx: i think nccl's wait is a counted wait so we will need a count in the parameter

xla/core/collectives/communicator.h
```diff
@@ -161,6 +162,20 @@ class Communicator {
...
+                       SymmetricMemory* recv_buffer, size_t offset,
+                       size_t count, RankId peer, const Executor& executor) {
+    return Unimplemented("Put is not implemented");
+  }
+
+  virtual Future<> Signal(const Executor& executor) {
```
Tixxx: it would be good to put some documentation in the interface to explain how to use these. ie 
put is for sending actual data
signal doenst send any data, it can be used as a barrier and etc

xla/backends/gpu/collectives/gpu_communicator.h
ezhulenev: Also there is `sigIdx` and `ctx`, but instead of hardcoding it as ints in this API, let's do the same trick as `Executor` and define virtual base `Signal` (`SignalContext`? I'd prefer to not use context, too many contexts already. `SignalDesc`).

xla/backends/gpu/collectives/nccl_communicator.h
ezhulenev, line 157: Put also needs a signal desc

xla/core/collectives/communicator.h
```diff
@@ -55,6 +55,11 @@ class Communicator {
     virtual ~Executor() = default;
   };
 
+  class SignalDesc {
```
ezhulenev, line 58: Few lines of documentation what signal desc is about

xla/core/collectives/communicator.h
```diff
@@ -162,17 +167,29 @@ class Communicator {
...
                        SymmetricMemory* recv_buffer, size_t offset,
                        size_t count, RankId peer, const Executor& executor) {
     return Unimplemented("Put is not implemented");
   }
 
-  virtual Future<> Signal(const Executor& executor) {
+  // Sends a signal to peer without transferring data. Can be used as a barrier
+  // or to notify peer that prior Puts (and Signals) to the same descriptor have
+  // completed. desc carries backend-specific signal identity (e.g. sigIdx, ctx).
+  virtual Future<> Signal(RankId peer, const SignalDesc& desc,
```
ezhulenev: instead of `desc`, `signal_desc` is self-documenting, just `desc` will be harder to parse later without knowing thetype

xla/core/collectives/communicator.h
sfvaroglu: Done

xla/backends/gpu/collectives/gpu_communicator.h
sfvaroglu: Done

xla/core/collectives/communicator.h
sfvaroglu: Done

xla/core/collectives/communicator.h
sfvaroglu: Done

xla/core/collectives/communicator.h
sfvaroglu: Done

ezhulenev (approved):

Tixxx (approved):

ermilovmaxim (approved):

PR #38946, opened 2026-03-09
[xla:gpu] Add a generic Async{Start/Done}Thunk to support async execution

ezhulenev:
Introduces a generic mechanism for executing nested thunk sequences asynchronously on dedicated streams with event-based synchronization.

#### AsyncExecution

Manages per-execution event lifecycle with an object pool. Enforces exactly one open Start/Done scope via a counter, and DCHECKs for missing Done at scope destruction.

#### AsyncStartThunk / AsyncDoneThunk

  - AsyncStartThunk wraps a ThunkExecutor for nested thunks, creates a compute→async stream dependency via event, executes nested thunks on the async stream, and records a completion event via RAII guard.
  - AsyncDoneThunk synchronizes the compute stream with the async completion event.
  - AsyncKind enum distinguishes compute vs communication stream resolution. Long term we should unify async comm with compute.


beckerhe (approved):

pifon2a (approved):

seantalts (approved):

xla/backends/gpu/runtime/async_thunk_test.cc
```diff
@@ -0,0 +1,151 @@
...
+  ASSIGN_OR_RETURN(se::Platform * platform,
+                   se::PlatformManager::PlatformWithName(platform_name));
+  return platform->ExecutorForDevice(0);
+}
+
+// Test that 4 async start thunks each memset a quarter of a buffer on separate
+// streams, and 4 async done thunks synchronize back to the main stream.
+TEST(AsyncThunkTest, ConcurrentMemsets) {
+  ASSERT_OK_AND_ASSIGN(se::StreamExecutor * executor, CreateExecutor());
+  ASSERT_OK_AND_ASSIGN(auto stream, executor->CreateStream());
```
thomasjoerg, line 59: Running into compile issues here:

```
error: expected '(' for function-style cast or type construction
   58 |   ASSERT_OK_AND_ASSIGN(se::StreamExecutor * executor, CreateExecutor());
      |                        ~~~~~~~~~~~~~~~~~~ ^
//xla/backends/gpu/runtime/async_thunk_test.cc:58 
error: use of undeclared identifier 'executor'; did you mean 'CollectiveParams::executor'?
   58 |   ASSERT_OK_AND_ASSIGN(se::StreamExecutor * executor, CreateExecutor());
      |                                             ^~~~~~~~
      |                                             CollectiveParams::executor
//xla/backends/gpu/runtime/collective_params.h:51
note: 'CollectiveParams::executor' declared here
   51 |   se::StreamExecutor* executor;
      |                       ^
//xla/backends/gpu/runtime/async_thunk_test.cc:58
error: invalid use of non-static data member 'executor'
   58 |   ASSERT_OK_AND_ASSIGN(se::StreamExecutor * executor, CreateExecutor());
```

thomasjoerg (changes requested):

xla/backends/gpu/runtime/async_thunk_test.cc
ezhulenev: Can't reproduce it, but pushed a change that uses `ASSERT_OK_AND_ASSIGN(auto` as I guess spaces + * trips macro expansion.

beckerhe (approved):

xla/backends/gpu/runtime/async_thunk_test.cc
beckerhe: I suspect that it doesn't recognize ASSERT_OK_AND_ASSIGN as a macro, so the problem might be a missing include. If it works here in OSS, then it might be an issue internally.

ezhulenev: I also added both `#include <gmock/gmock.h> #include <gtest/gtest.h>`, I suspect it comes from one of them internally

ermilovmaxim (approved):

seantalts (approved):

PR #38992, opened 2026-03-10
Integrate Triton version 5a9cfa6ab3e59608d2f210e41e024994626a838c

zoranjovanovic-ns:
📝 Summary of Changes
Integrate new Triton version 5a9cfa6ab3e59608d2f210e41e024994626a838c
PR created just for testing purposes

🎯 Justification
Some fixes on Triton side necessary for ROCm/

🚀 Kind of Contribution

🧪 Unit Tests:
transformed HLO.

🧪 Execution Tests:


vwbaker:
> PR created just for testing purposes

removing my review because it says it's just for testing.

If this isn't the case, we will likely be unable to accept any Triton integration PRs from OSS - we generally have to update a lot more internally. Is there a reason you need such an up-to-date Triton? The  Triton as of right now is only 4 days old. 

If something is needed urgently, we can do a cherry-pick and add a patch, but this would also likely require us doing it internally in order to update our version as well.

i-chaochen:
Hi @vwbaker thanks a lot for your reply!

The reason we need this bump up is because last week we found we need this patch https://github.com/triton-lang/triton/commit/f47029e15e941162fd4bd9063eb7addbecf58bb5 but I can see the triton commit is already updated so we don't need this PR anymore. 

Thanks!

PR #39014, opened 2026-03-10
[xla:pjrt] Unify AsyncWorkRunner API with tsl::Executor

ezhulenev:
Refactor `AsyncWorkRunner` to own thread pools by value and unify with `tsl::Executor`

Summary:
  - Make `AsyncWorkRunner` extend `tsl::Executor` with a single `Execute(Task)` virtual method, replacing the old `Schedule/ScheduleWhenReady/EnqueueWork` APIs. Delete async_work_runner.cc.
  - `ThreadPoolAsyncWorkRunner` and `UnboundedAsyncWorkRunner` now own their underlying thread pool/work queue by value, fixing ASAN/MSAN use-after-free bugs where `ExecuteWhenReady` callbacks outlived the runner.
  - Update all callers (cpu_client, tfrt_gpu_client, se_gpu_pjrt_client, common_pjrt_client).


derdrdirk (approved):
LGTM

Thanks for the fix!

xla/pjrt/async_work_runner.h
```diff
@@ -28,24 +29,31 @@ namespace xla {
...
+  void ExecuteWhenReady(
+      absl::Span<const tsl::RCReference<tsl::AsyncValue>> dependencies,
+      Task task) {
+    tsl::RunWhenReady(dependencies, [this, task = std::move(task)]() mutable {
+      Execute(std::move(task));
+    });
+  }
+
+  ABSL_DEPRECATE_AND_INLINE()
+  void Schedule(absl::AnyInvocable<void() &&> work) {
```
thomasjoerg, line 47: Google-internal code overrides this method. Can you keep this as `virtual` so that's still feasible?

thomasjoerg (changes requested):

xla/pjrt/async_work_runner.h
ezhulenev: It shouldn't, instead it should override Execute and conform to `tsl::Executor` API. Also instead of overriding ScheduleWhenReady it should use default version. The idea is that there are too many virtual APIs that look and feel like tsl::Executor, FWIW I talked with Pathways team and the also planned to use tsl::Executor instead of PathwaysExecutor.

ezhulenev: Dirk had a patched CL that got rolled back, I assume it had all internal fixes.

thomasjoerg: Alright, I'll check with Dirk!

thomasjoerg (approved):

PR #39083, opened 2026-03-12
[xla:gpu] Remove CubSortThunk and invoke FFI custom calls directly

ezhulenev:
Remove `CubSortThunk;` emit CUB sort as generic FFI custom call

Simplify CUB sort infrastructure by removing the specialized `CubSortThunk` and instead emitting `CUB` sort operations as generic `FFI `custom calls handled by `CustomCallThunk`. This eliminates a dedicated thunk type in favor of the existing `FFI` handler framework, which is already powerful enough to support scratch buffer computation (via instantiate stage) and runtime dispatch. Also don't abuse passing a pointer via attributes to get scratch size, use proper state mechanism.

####  Key changes:
  - Consolidate ~30 per-type FFI handler registrations into two single handlers (`xla.gpu.ext.cub_sort_keys` and `xla.gpu.ext.cub_sort_pairs`) with runtime type dispatch via `AnyBuffer::element_type()`
  - Sort rewriter emits `API_VERSION_TYPED_FFI` custom calls with MLIR dict backend config (`descending`, `batch_size` attributes)
  - `EstimateCubScratchSize` pass invokes FFI instantiate handler directly to compute scratch buffer size
  - `Delete CubSortThunk`, its proto, tests, and thunk kind enum entry
  - Fix `TypeRegistry::GetTypeId<T>()` to use `GetOrAssignTypeId` instead of `AssignTypeId`, preventing type ID conflicts when FFI handlers and direct `ExecutionState` access register the same type
  - Add segment offset copying in FFI execute handlers for batched sort
  - Add all int32_t key pair instantiations (s32_b16, s32_b32, s32_b64) for completeness

This approach is better than creating a custom thunk for each extension operation because FFI custom calls already provide all needed runtime capabilities: instantiate stage for scratch size computation, typed buffer access, attribute passing, and stream management via CustomCallThunk.


pifon2a (approved):

xla/backends/gpu/runtime/thunk.proto
```diff
@@ -421,7 +413,7 @@ message ThunkProto {
     FftThunkProto fft_thunk = 27;
     Memset32BitValueThunkProto memset32bit_value_thunk = 28;
     CustomCallThunkProto custom_call_thunk = 30;
-    CubSortThunkProto cub_sort_thunk = 31;
+    // 31 was CubSortThunkProto, removed.
```
allanrenucci: Should we reserve this field to avoid reuse?

ezhulenev: Done. Also rebased to HEAD.

allanrenucci (approved):

seantalts (approved):

akuegel:
FYI: I needed to keep the change out for now to allow int,short and int,long key-value pairs. A JAX test was complaining that the JAX wheels size becomes to big. I will prepare a separate change for adding those, and then ask someone from JAX to approve.

pemeliya:
@akuegel, @ezhulenev 
Note that, this PR has broken our ROCM cubsort FFI handler registration which I previously fixed in: https://github.com/openxla/xla/pull/35354

So we will have to create another PR to fix it.

akuegel:
> @akuegel, @ezhulenev Note that, this PR has broken our ROCM cubsort FFI handler registration which I previously fixed in: #35354
> 
> So we will have to create another PR to fix it.

@pemeliya Just to avoid any misunderstanding, will this PR with a fix be coming from you or someone else at AMD, or are you expecting us to fix it? As it was not my change, I am not planning to create a fix for this. It is somewhat complicated for us to test it as we don't run ROCM internally.

pemeliya:
> > @akuegel, @ezhulenev Note that, this PR has broken our ROCM cubsort FFI handler registration which I previously fixed in: #35354
> > So we will have to create another PR to fix it.
> 
> @pemeliya Just to avoid any misunderstanding, will this PR with a fix be coming from you or someone else at AMD, or are you expecting us to fix it? As it was not my change, I am not planning to create a fix for this. It is somewhat complicated for us to test it as we don't run ROCM internally.

yes, I understand, we are going to fix it on ROCM, already working on it. I just wanted to keep you informed.

PR #39155, opened 2026-03-13
Fix UB in FFI State Get/Set Backwards Compatibility Case 

mgoldfarb-nvidia:
📝 Summary of Changes
Replaces `reinterpret_cast` with `std::memcpy` to avoid U.B.
Implementation of Get_State was not setting state pointer when using the backward compatibility path

🎯 Justification
Support state with V0.2 FFI in latest JAX.

🚀 Kind of Contribution
 🐛 Bug Fix

### DIFF
--- a/xla/ffi/ffi_api.cc
+++ b/xla/ffi/ffi_api.cc
@@ -349,13 +349,15 @@ static XLA_FFI_Error* XLA_FFI_State_Set(XLA_FFI_State_Set_Args* args) {
   // If struct size matches the legacy struct layout, always assume that we set
   // the state for instantiation stage.
   if (args->struct_size == XLA_FFI_State_Args_V02_STRUCT_SIZE) {
-    auto* v02 = reinterpret_cast<XLA_FFI_State_Args_V02*>(args);
+    // memcpy to safely avoid U.B.
+    XLA_FFI_State_Args_V02 v02{};
+    std::memcpy(&v02, args, XLA_FFI_State_Args_V02_STRUCT_SIZE);
 
     XLA_FFI_State_Set_Args compat = {XLA_FFI_State_Set_Args_STRUCT_SIZE};
-    compat.ctx = v02->ctx;
+    compat.ctx = v02.ctx;
     compat.stage = XLA_FFI_ExecutionStage_INSTANTIATE;
-    compat.type_id = v02->type_id;
-    compat.state = v02->state;
+    compat.type_id = v02.type_id;
+    compat.state = v02.state;
 
     return XLA_FFI_State_Set(&compat);
   }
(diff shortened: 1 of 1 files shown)


mgoldfarb-nvidia:
FYI @ezhulenev

xla/ffi/ffi_api.cc
```diff
@@ -349,13 +349,15 @@ static XLA_FFI_Error* XLA_FFI_State_Set(XLA_FFI_State_Set_Args* args) {
   // If struct size matches the legacy struct layout, always assume that we set
   // the state for instantiation stage.
   if (args->struct_size == XLA_FFI_State_Args_V02_STRUCT_SIZE) {
-    auto* v02 = reinterpret_cast<XLA_FFI_State_Args_V02*>(args);
+    // memcpy to safely avoid U.B.
+    XLA_FFI_State_Args_V02 v02{};
+    std::memcpy(&v02, args, XLA_FFI_State_Args_V02_STRUCT_SIZE);
```
ezhulenev, line 354: Why memcpy and what UB it avoids?

xla/ffi/ffi_api.cc
```diff
@@ -380,15 +382,23 @@ static XLA_FFI_Error* XLA_FFI_State_Get(XLA_FFI_State_Get_Args* args) {
...
+    compat.type_id = v02.type_id;
+    compat.state = v02.state;
 
-    return XLA_FFI_State_Get(&compat);
+    XLA_FFI_Error* err = XLA_FFI_State_Get(&compat); 
+    if (ABSL_PREDICT_FALSE(err != nullptr)) {
+      return err;
+    }
+
+    args->state = compat.state;
```
ezhulenev: this is wrong (maybe?) you set state via old struct layout, `v02->state = compat.state` should be enough here?

xla/ffi/ffi_api.cc
mgoldfarb-nvidia: https://en.cppreference.com/w/cpp/language/reinterpret_cast.html

Strict aliasing between two unrelated types causes UB. So reads and writes through `args` may not reflect . The `memcpy` is the pre-C++20 compliant way to circumvent this rule. In many cases compiler should optimize away an actual copy.

xla/ffi/ffi_api.cc
mgoldfarb-nvidia: I actually just updated this due to copy-paste failure on my part. Indeed we can not write directly to `args->state` instead we need to `v02.state = compat.state` then `memcpy` to safely update.

ezhulenev (approved):

xla/ffi/ffi_api.cc
beckerhe: Can you use `absl::bit_cast` instead? It's a backport of std::bit_cast and is basically a wrapper around the memcpy pattern; plus some additional checks.

xla/ffi/ffi_api.cc
```diff
@@ -380,15 +382,25 @@ static XLA_FFI_Error* XLA_FFI_State_Get(XLA_FFI_State_Get_Args* args) {
...
+    compat.state = v02.state;
 
-    return XLA_FFI_State_Get(&compat);
+    XLA_FFI_Error* err = XLA_FFI_State_Get(&compat); 
+    if (ABSL_PREDICT_FALSE(err != nullptr)) {
+      return err;
+    }
+
+    v02.state = compat.state;
+    std::memcpy(args, &v02, XLA_FFI_State_Args_V02_STRUCT_SIZE);
```
beckerhe, line 401: and here.

xla/ffi/ffi_api.cc
```diff
@@ -380,15 +382,25 @@ static XLA_FFI_Error* XLA_FFI_State_Get(XLA_FFI_State_Get_Args* args) {
   // If struct size matches the legacy struct layout, always assume that we get
   // the state for instantiation stage.
   if (args->struct_size == XLA_FFI_State_Args_V02_STRUCT_SIZE) {
-    auto* v02 = reinterpret_cast<XLA_FFI_State_Args_V02*>(args);
+    // memcpy to safely avoid U.B.
+    XLA_FFI_State_Args_V02 v02{};
+    std::memcpy(&v02, args, XLA_FFI_State_Args_V02_STRUCT_SIZE);
```
beckerhe, line 387: Same here.

xla/ffi/ffi_api.cc
mgoldfarb-nvidia: One problem with `absl::bit_cast` (and `std::bit_cast`) is it (looks like it) requires the type sizes to be equal for writes. At least in the fallback path that uses `std::memcpy`

```
inline Dest bit_cast(const Source& source) {
  Dest dest;
  memcpy(static_cast<void*>(std::addressof(dest)),
         static_cast<const void*>(std::addressof(source)), sizeof(dest));
  return dest;
}
```

Its possible we could modify this function s.t. we pass `void*` from the FFI API then first cast to `size_t` to determine the size then cast to one concrete type. However, this could still be a U.B. violation and a bigger change than makes sense unless we start adding more cases where we are making ABI breaking changes.

beckerhe: Oh, I see. In that case the current version is fine.

beckerhe (approved):

PR #39190, opened 2026-03-13
[ROCm] Enable ROCm Triton backend for AllReduce

mfrancepillois:
📝 Summary of Changes
This PR enables the ROCm Triton backend for AllReduce (collective emitter).
To this end:
- Add generic passes to lower `triton_xla.block_barrier` and atomic operations `triton_xla.atomic_write` and `triton_xla.atomic_spin_wait`. These passes use only triton operations. thereby avoiding the use of target specific inline assembly.
- Add missing API to rocm_executor `RocmExecutor::CanEnablePeerAccessTo(int other_device_ordinal)` (this API is required to enable collective_emitter thunk).

🎯 Justification
Prior to this PR, the `triton_xla.block_barrier`, `triton_xla.atomic_write` and `triton_xla.atomic_spin_wait` operations were lowered using PTX assembly. Therefore, AllReduce triton backend was only available for CUDA target.
This PR adds a generic way to lower these operations only Triton operations.
Thanks to that, the triton backend for AllReduce is now available for ROCm target.

🚀 Kind of Contribution

🧪 Unit Tests:
This PR includes a LIT test checking the lowering of atomic operations.

🧪 Execution Tests:
The existing test `xla/tests:all_reduce_e2e_test` is relevant for testing this support.


xla/backends/gpu/codegen/triton/transforms/triton_xla_lower_atomics_generic_pass.cc
```diff
@@ -0,0 +1,311 @@
...
+            tensor_type.getShape(), loop_builder.getI1Type());
+
+        // Use the predicate that was determined outside the lambda
+        mlir::Value comparison_result = mlir::arith::CmpIOp::create(
+            loop_builder, predicate, cas_results, expected_tensor);
+
+        // Apply mask if provided: arith.select %mask, %comparison_result,
+        // %false
+        mlir::Value masked_comparison = comparison_result;
+        if (mask) {
```
draganmladjenovic, line 209: This seems wrong. Can yo provide LLVM IR for this whole op. At fisrt glance I don't see memory acesses being guarded by mask.

xla/backends/gpu/codegen/triton/transforms/triton_xla_lower_atomics_generic_pass.cc
```diff
@@ -0,0 +1,311 @@
...
+        // Extend booleans to i32 for reduction (matching Triton Python)
+        // arith.extui %masked_comparison : tensor<Nxi1> to tensor<Nxi32>
+        auto i32_tensor_type = mlir::RankedTensorType::get(
+            tensor_type.getShape(), loop_builder.getI32Type());
+        mlir::Value comparison_i32 = mlir::arith::ExtUIOp::create(
+            loop_builder, i32_tensor_type, masked_comparison);
+
+        // Reduce with ADD (matching Triton Python, not OR)
+        // tt.reduce %comparison_i32 {axis = 0} : tensor<Nxi32> -> i32
+        auto reduce_op = triton::ReduceOp::create(
```
draganmladjenovic, line 223: Seem inefficient. Can you provide LLVM IR? If it is optimized out then its ok. Otherwise it seem that you chose a wrong level of abstraction and this should have been done one on LLVM lowering level. Just comparing the tensor agains other tensor cannot serve as the loop condition source?

xla/backends/gpu/codegen/triton/compilation_pipeline.cc
```diff
@@ -32,13 +32,23 @@ namespace xla::gpu {
...
+  // targets
+  if (gpu_cc.cuda_compute_capability() != nullptr) {
+    // CUDA-specific passes using PTX inline assembly
+    pm->addPass(mlir::triton::xla::CreateTritonXLALowerBlockBarrierPass());
+    pm->addPass(mlir::triton::xla::CreateTritonXLALowerAtomicsPass());
+    pm->addPass(mlir::triton::xla::CreateTritonXLALowerGetTidPass());
+  } else {
+    // Generic passes using pure Triton ops (ROCm and other targets)
+    pm->addPass(
+        mlir::triton::xla::CreateTritonXLALowerBlockBarrierGenericPass());
```
draganmladjenovic, line 49: Why we need this whole new pass?

xla/backends/gpu/codegen/triton/transforms/tests/triton_xla_lower_atomics_generic.mlir
```diff
@@ -0,0 +1,61 @@
+// RUN: xla-opt %s -triton-xla-atomics-generic | FileCheck %s
+
+// Test lowering of AtomicWriteOp to pure Triton AtomicRMWOp using generic Triton ops
+
+// CHECK-LABEL: @test_atomic_write
+tt.func @test_atomic_write(%ptr: !tt.ptr<i32>, %value: i32) {
+  // CHECK-NOT: triton_xla.atomic_write
+  // CHECK: tt.atomic_rmw exch
+  triton_xla.atomic_write sys, release, %ptr, %value : (!tt.ptr<i32>, i32) -> ()
```
draganmladjenovic, line 9: I think LLVM backend does so, but please check if this ends up plain atomic store.

xla/backends/gpu/codegen/triton/compilation_pipeline.cc
draganmladjenovic: Ideally if generic impl lowers to same asm code we should just replace ptx asm with generic IR level impl.

xla/backends/gpu/codegen/triton/transforms/triton_xla_lower_atomics_generic_pass.cc
mfrancepillois: You are right. Memory accesses were not guarded by the mask. The code has been updated to address this.

xla/backends/gpu/codegen/triton/transforms/triton_xla_lower_atomics_generic_pass.cc
mfrancepillois: The problem is that there is no Triton API that compares two tensors and return a scalar or Boolean. `CmpIOp` on tensors return a tensor. So, we need to reduce it at some point to obtain or loop condition. If we don't want to have this `ReduceOp`, we must resort to `element_wise` ops, so either `tt.elementwise_inline_asm` where target specific assembly  is defined in XLA or `tt.extern_elementwise` where we rely on a Triton extension to generate target specific assembly or llvm intrinsics if available.
But generally  I think that it is a good thing to have a generic pass that can target any GPUs, even though we later add an AMD-optimized pass.

xla/backends/gpu/codegen/triton/compilation_pipeline.cc
mfrancepillois: The CUDA TritonXLALowerBlockBarrier pass differs from the generic one, as it uses `GetTidOp` to get thread ID and then wraps the barrier logic in an `if (tid < world_size)` block.
So, even though we could probably merge the two passes (i.e., the cuda one and the generic one) into a single pass, IMO the code is clearer and more modular with two different passes that correspond to the two strategy for lowering this block-barrier (one depending on the thread id and the second one not depending on any target specific code.)

xla/backends/gpu/codegen/triton/transforms/tests/triton_xla_lower_atomics_generic.mlir
mfrancepillois: This is lowered by triton into:
```
%199 = cmpxchg ptr addrspace(1) %196, i32 %arg3.load77, i32 %arg3.load77 acquire monotonic, align 4, !noalias !8
```
However, Triton API does not allow us to perform atomic write except with [this op](https://triton-lang.org/main/dialects/TritonOps.html#tt-atomic-rmw-triton-atomicrmwop): `tt.atomic_rmw exch`

draganmladjenovic: Can you check the llvm IR before isel phase or assembly itself. This seems like a CAS which seems wrong.

xla/backends/gpu/codegen/triton/compilation_pipeline.cc
draganmladjenovic: If our change resutls in the same code there is no reason to keep the ptx assembly one.

xla/backends/gpu/codegen/triton/transforms/triton_xla_lower_atomics_generic_pass.cc
draganmladjenovic: There is not need a for a scalar. Can triton loop on tensor of bools? extern_elementwise  is generic too. LLVM IR is good enough generic abstraction.

xla/stream_executor/rocm/rocm_executor.cc
```diff
@@ -657,6 +656,23 @@ absl::Status RocmExecutor::Init() {
...
+  for (int i = 0; i < device_count; ++i) {
+    if (i == device_ordinal()) {
+      // A device can always access its own memory
+      peer_access_cache_[i] = true;
+      continue;
+    }
+
+    // Check peer access capability and cache the result
+    TF_ASSIGN_OR_RETURN(hipDevice_t peer_device, GetDevice(i));
+    peer_access_cache_[i] = CanEnablePeerAccess(device_, peer_device);
```
draganmladjenovic, line 673: I don't think there is an need to cache this. just call the underlying API when needed.

mfrancepillois: This mimics the CUDA implementation. And since this function is called every time a Collective Emitter thunk is launched, I think it’s better to use a cache rather than calling the API each time.

xla/backends/gpu/codegen/triton/transforms/tests/triton_xla_lower_atomics_generic.mlir
mfrancepillois: My previous reply was incorrect.
Indeed `triton_xla.atomic_write` is lowered to:
Triton IR
```
%50 = "tt.atomic_rmw"(%48, %49) <{atomic_rmw_op = 10 : i32, scope = 3 : i32, sem = 3 : i32}> : (tensor<8x!tt.ptr<i32>>, tensor<8xi32>) -> tensor<8xi32> "all-reduce-start"
```
which is lowered to:
LLVM IR
```
%187 = atomicrmw xchg ptr addrspace(1) %186, i32 %113 release, align 4, !noalias !8
```
(not the `cmpxchg` mentioned previously),

And this is lowered to:
Assembly
```
global_atomic_swap v[14:15], v3, off sc1
```

mfrancepillois:
The PR is replaced by https://github.com/openxla/xla/pull/39924, which implements a more optimized support for ROCm triton backend for AllReduce

PR #39253, opened 2026-03-16
Add Triton GEMM default configs for consumer Blackwell (SM 12.0)

blasphemetheus:
## 📝 Summary of Changes

`GetDefaultTritonConfigs()` uses `IsAtLeastBlackwell()` to select B200 (SM 10.0) configs for all Blackwell GPUs. But consumer Blackwell GPUs (RTX 5090, SM 12.0) have different hardware characteristics (170 SMs, different cache/warp configs), and the B200-tuned configs fall outside SM 12.0's valid config search space. This causes `OptimizeConfigSet` to reject all hints:

```
dot_search_space.cc:240] All configs were filtered out because none of them
sufficiently match the hints.
```

Changes:
- Add `sm120.txtpb` with configs generated by exhaustive autotuning on an RTX 5090
- Add `kBlackwellConsumer` to `TritonConfigsPlatform` enum
- Update `GetDefaultTritonConfigs` to use `IsBlackwell()` for SM 10.0 (datacenter) and fall through to `kBlackwellConsumer` for SM 12.0+ (consumer)

## 🎯 Justification

Consumer Blackwell GPUs (RTX 5090, RTX 5080, etc.) are SM 12.0, which has different hardware characteristics from datacenter Blackwell (B200, SM 10.0). The existing `IsAtLeastBlackwell()` check applies B200 configs to both, but these configs are invalid for SM 12.0's search space. This causes the autotuner to reject all hints on every Triton GEMM compilation, logging a warning and falling back to brute-force search of the full config set. With SM 12.0-specific configs, the autotuner can use its hint-based filtering as designed, reducing first-compilation overhead.

## 🚀 Kind of Contribution

🐛 Bug Fix — consumer Blackwell GPUs get invalid hint configs, causing fallback to full search on every Triton GEMM compilation.

## 📊 Benchmark

Not a performance optimization per se — this fixes the hint selection so the autotuner works as designed on SM 12.0. The configs were generated by running `--xla_gpu_exhaustive_tiling_search=true` with representative GEMM sizes on:

- NVIDIA GeForce RTX 5090 (SM 12.0a, 170 SMs, 2.467 GHz, 1792 GB/s memory bandwidth, 96 MB L2 cache)
- JAX 0.9.1, cuDNN 9.13.0
- GEMM sizes: square (8×8 to 2048×2048), non-square transformer shapes (128×4096, 1024×4096), skinny projections (8192×64, 4096×128)

## 🧪 Unit Tests

No new tests added. The existing `triton_configs_test.cc` should cover the new platform enum. The `dot_search_space_test.cc` tests verify that `OptimizeConfigSet` works with valid hint configs — this change ensures SM 12.0 gets valid hints.

## 🧪 Execution Tests

Verified on RTX 5090 that the "All configs were filtered out" warning no longer appears when using SM 12.0-specific configs. No new execution tests added — the existing autotuner test infrastructure covers the config selection path.

## Related

- #39250 — f32 GEMM accuracy and autotuner issues on Blackwell (SM 12.0a)


blasphemetheus:
I signed the CLA, may need time to register somewhere tho, idk

blasphemetheus:
@mooskagh For the SM 12.0 configs in the PR, I generated them by running JAX matmuls of various sizes with exhaustive autotuning:

  ```bash
  XLA_FLAGS="--xla_gpu_exhaustive_tiling_search=true \
             --xla_gpu_dump_autotune_results_to=/tmp/sm120_autotune.textproto" \
  JAX_ENABLE_X64=1 python generate_configs.py
```
  The script runs jnp.dot(a, b) for 16 representative GEMM shapes (square 8×8 to 2048×2048, non-square transformer shapes like 128×4096 and 1024×4096, skinny projections like 8192×64) on an RTX 5090. I extracted the unique winning Triton configs from the autotune dump — that gave 10 configs.

  The B200 config file has significantly more entries. Should I run a more comprehensive set of shapes?

  Or is there a standard benchmark suite you'd prefer I run on this hardware?

blasphemetheus:
## Benchmark: Default (B200 fallback) vs Exhaustive (true optimal) on RTX 5090

  I benchmarked the same GEMM shapes with default config selection (B200 hints → "all filtered out" →
  fallback) vs exhaustive tiling search (finds true optimal):

  | Shape | Default (ms) | Exhaustive (ms) | Diff |
  |---|---|---|---|
  | f32 128x128 | 0.025 | 0.021 | -16% |
  | f32 256x256 | 0.023 | 0.022 | ~0% |
  | f32 512x512 | 0.031 | 0.038 | +18% |
  | f32 1024x1024 | 0.058 | 0.067 | +13% |
  | f32 2048x2048 | 0.234 | 0.213 | -9% |
  | f32 4096x4096 | 1.395 | 1.378 | ~0% |
  | f32 2048x3072x768 | 0.177 | 0.143 | -19% |
  | bf16 4096x4096 | 0.649 | 0.654 | ~0% |

  Performance is within noise — the B200 fallback picks kernels roughly as good as exhaustive search. So
  this PR is about correctness of the hint selection path (eliminating the warning, providing proper SM 12.0
   hints) rather than a performance fix.

  Re: Ampere configs — given that default fallback already matches optimal, comparing against A100 configs
  seems unnecessary. The fallback mechanism works fine; this PR just makes it work through the intended hint
   path instead of the warning+fallback path.

mooskagh:
Using pure matmuls without any fused ops is usually not representative for the actual configs, as they are memory bandwidth bound, and also usually call cuBLAS backend which is highly optimized for such cases.

Internally, we use a set of ≈2000 fusions, extracted from real models, but I don't think it's easily shareable. I'll check whether there are any open sourced fusions.

blasphemetheus:
@mooskagh gotcha, pure matmuls not representative. I found the open-sourced HLO benchmarks at `xla/tools/benchmarks/hlo/` (Llama, Mixtral, MaxTest). Would running those through exhaustive autotuning on the RTX 5090 produce better configs? I'd need to build the HLO runner but have the Bazel toolchain set up already.

`triton_gemm_fusion_test.cc` has ~193 test fusions that might be better than raw matmuls.

blasphemetheus:
Updated the configs with fused operations — matmul + bias + activation (GELU/ReLU/SiLU), attention Q/K/V projections, attention scores (Q@K^T/sqrt(d)), SwiGLU MLP, shapes from `triton_gemm_fusion_test.cc`, and mixed precision.

52 out of 58 fused operations chose the Triton path (vs mostly cuBLAS for pure matmuls, as you predicted). Combined with the earlier pure matmul configs, we now have 193 unique configs.

blasphemetheus:
Compilation time benchmark — loading cached autotune results vs default:

  | | Default | With cached results | Speedup |
  |---|---|---|---|
  | 10 fused ops total | 2.82s | 1.63s | 42% faster |
  | SwiGLU 32x4096x14336 | 0.47s | 0.085s | 5.5x |
  | linear 8x2048x8192 | 0.19s | 0.053s | 3.6x |

  This was using `--xla_gpu_load_autotune_results_from` with the fused autotune dump. The hint configs in `sm120.txtpb` would provide a similar benefit through the `OptimizeConfigSet` path by narrowing the search space.

pifon2a (approved):

PR #39293, opened 2026-03-16
[xla] Add DWYU (Depend on What You Use) check

ezhulenev:
Add `bant`-based (https://github.com/hzeller/bant) presubmit that does roughly the same check as google-internal buildcleaner check, but visible for OSS contributors. It might not be always correct, but it catches most of the common problems.


ddunl:
LGTM, I will let Henning review too to make sure there's no concerns on his end

ermilovmaxim (approved):

PR #39302, opened 2026-03-16
Fix dynamic memcpy offset computation for host offloading with collective pipelining

sfvaroglu:
📝 Summary of Changes
Track per-variable init/step metadata in WhileLoopBackendConfig so that FusionDynamicMemcpyRewriter, collective pipelining, and double buffer unrolling correctly compute dynamic slice offsets for host-offloaded loops.

🎯 Justification
FusionDynamicMemcpyRewriter uses a single global known_init_step for all dynamic variables, even when they have different init/step values. After collective pipelining and double buffering transform the loop, non-primary variables carry stale metadata, producing wrong slice offsets and potential data corruption in host-offloaded workloads.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

N/A

🧪 Unit Tests:
Added

🧪 Execution Tests:
N/A


xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils_test.cc
```diff
@@ -119,7 +119,11 @@ TEST_F(CollectiveCombinerUtilsTest,
       });
 
   hlo_query::ForEachInstructionWithPred(
-      *module, HloPredicateIsNotOp<HloOpcode::kAllReduce>,
+      *module,
+      [](const HloInstruction* instr) {
```
olegshyshkov: `HloPredicateIsNotOp` can have multiple template arguments. `HloPredicateIsNotOp<HloOpcode::kAllReduce, HloOpcode::kWhile>`

xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils_test.cc
```diff
@@ -208,7 +212,11 @@ TEST_F(CollectiveCombinerUtilsTest,
...
```
olegshyshkov: HloPredicateIsNotOp can have multiple template arguments. HloPredicateIsNotOp<HloOpcode::kAllReduce, HloOpcode::kWhile>

xla/xla_data.proto
```diff
@@ -1428,6 +1428,18 @@ message WhileLoopBackendConfig {
   // Variables that should be treated as induction variables for dynamic memcpy
   // analysis, even though they are not the primary induction variable.
   repeated int64 dynamic_variable_tuple_indices = 4;
+
+  message DynamicVariable {
+    int64 tuple_index = 1;
```
olegshyshkov, line 1434: If we already have `tuple_index` here, then maybe we can remove `dynamic_variable_tuple_indices` to avoid duplication?

sfvaroglu: I considered unifying them but it's not straightforward. `host_offload_utils.cc` populates `dynamic_variable_tuple_indices` by scanning DUS/DS instructions without loop analysis info, so it can't compute init/step. If we create DynamicVariable entries with default init=0/step=0, the memcpy rewriter would use those instead of falling back to known_init_step, producing wrong offsets. We'd need to make init/step optional in the proto to distinguish "unknown" from "zero." Happy to do that as a follow-up to keep this PR focused on the correctness fix.

olegshyshkov (approved):

allanrenucci (approved):
Trying to import internally

allanrenucci:
This break tests on TPU. Looking into it

xla/service/collective_pipeliner.cc
```diff
@@ -2225,6 +2233,24 @@ absl::StatusOr<HloInstruction*> TransformLoopForward(
         absl::MakeSpan(loop_output_to_replace), output_stacked_data));
   }
   TF_RETURN_IF_ERROR(loop_computation->parent()->RemoveUnusedComputations());
+  if (!induction_var_pipeline_counter_positions.empty() &&
+      loop_analysis.GetLoopStart().has_value() &&
+      loop_analysis.GetLoopIncrement().has_value()) {
+    WhileLoopBackendConfig config;
+    auto config_or = new_while_loop->backend_config<WhileLoopBackendConfig>();
```
allanrenucci: This breaks the TPU compiler. Downstream TPU passes expect a backend specific config to be attached to while loop instructions.
```
INTERNAL: Could not convert JSON string to proto: invalid JSON in  xla.jellyfish.BackendConfig, near 1:35 (offset 34):   no such   field: 'dynamic_variable_tuple_indices'
third_party/tensorflow/compiler/xla/tsl/platform/errors.h:1247
third_party/tensorflow/compiler/xla/hlo/ir/backend_config.cc:109
third_party/tensorflow/compiler/xla/hlo/ir/hlo_instruction.h:2069
platforms/xla/service/jellyfish/lowering_emitter.cc:3006
```
It is not clear to me what the right solution is. Will discuss internally.

allanrenucci:
Instead of updating `WhileLoopBackendConfig` as we transform the while loop, can we instead create the config after we are done transforming the loop? This is what `WhileLoopTripCountAnnotator` does: https://github.com/openxla/xla/blob/bf54fff31c2144fbfde6399c5b75a780eaec9727/xla/service/gpu/gpu_compiler.cc#L1186-L1189

Can this be done as part of `WhileLoopTripCountAnnotator` instead of `DoubleBufferLoopUnrolling`/`CollectivePipeliner`? This would address the TPU compiler failures since `WhileLoopTripCountAnnotator` is not run as part of the TPU compiler pipeline.

Updating the loop metadata across multiple passes seems very fragile.

sfvaroglu:
> Instead of updating `WhileLoopBackendConfig` as we transform the while loop, can we instead create the config after we are done transforming the loop? This is what `WhileLoopTripCountAnnotator` does:
> 
> https://github.com/openxla/xla/blob/bf54fff31c2144fbfde6399c5b75a780eaec9727/xla/service/gpu/gpu_compiler.cc#L1186-L1189
> 
> Can this be done as part of `WhileLoopTripCountAnnotator` instead of `DoubleBufferLoopUnrolling`/`CollectivePipeliner`? This would address the TPU compiler failures since `WhileLoopTripCountAnnotator` is not run as part of the TPU compiler pipeline.
> 
> Updating the loop metadata across multiple passes seems very fragile.

Please review.

xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils_test.cc
```diff
@@ -125,7 +125,7 @@ TEST_F(CollectiveCombinerUtilsTest,
       });
 
   hlo_query::ForEachInstructionWithPred(
-      *module, HloPredicateIsNotOp<HloOpcode::kAllReduce>,
+      *module, HloPredicateIsNotOp<HloOpcode::kAllReduce, HloOpcode::kWhile>,
```
allanrenucci: Why this change?

xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils_test.cc
```diff
@@ -214,7 +214,7 @@ TEST_F(CollectiveCombinerUtilsTest,
...
```
allanrenucci: ditto

xla/service/host_offload_utils.cc
```diff
@@ -475,8 +473,26 @@ absl::Status MarkDynamicVariables(HloInstruction* while_loop) {
...
+  absl::flat_hash_map<int64_t,
+                      std::pair<std::optional<int64_t>, std::optional<int64_t>>>
+      existing;
+  for (const auto& dv : config.dynamic_variables()) {
+    std::optional<int64_t> init =
+        dv.has_init() ? std::optional<int64_t>(dv.init()) : std::nullopt;
+    std::optional<int64_t> step =
+        dv.has_step() ? std::optional<int64_t>(dv.step()) : std::nullopt;
+    existing[dv.tuple_index()] = {init, step};
+  }
```
allanrenucci: Is this dead code? I believe the config should be empty here.

If we move the logic below to `WhileLoopTripCountAnnotator` then we wouldn't need this.

xla/service/host_offload_utils.cc
allanrenucci: Can we move this to WhileLoopTripCountAnnotator?

xla/hlo/transforms/while_loop_trip_count_annotator.h
```diff
@@ -24,13 +24,24 @@ limitations under the License.
...
 //
 // The annotation is stored as a backend-config on the while loop node.
 //
 // This pass should run after all passes that might semantically modify a while
 // loop, e.g. by unrolling it.  Otherwise, a loop could end up with a
-// backend-config that doesn't match its true trip-count.
+// backend-config that doesn't match its true trip-count, or with stale
+// `dynamic_variables` init/step entries for unrolled/pipelined slots.
+//
+// This pass is the *only* writer of `WhileLoopBackendConfig::known_init_step`
```
allanrenucci: I believe `DoubleBufferLoopUnrolling` also mutates `known_init_step`. It is not causing a problem because it is a GPU specific pass.

xla/hlo/transforms/while_loop_trip_count_annotator.h
```diff
@@ -24,13 +24,24 @@ limitations under the License.
...
+// and of `init`/`step` on `dynamic_variables`. Other passes that need to
+// register dynamic-variable tuple positions (e.g. host-offload pipelining)
+// should write only the `tuple_index` and rely on this pass to fill in the
+// affine values once the loop is final. This keeps `WhileLoopBackendConfig`
+// off the critical path for backends (e.g. TPU) that use a different backend
+// config proto and don't run this pass.
```
allanrenucci: Ideally I think we should move the MarkDynamicVariables logic into WhileLoopBackendConfig. Is that practical?

xla/hlo/transforms/while_loop_trip_count_annotator.cc
```diff
@@ -76,50 +192,61 @@ absl::StatusOr<bool> WhileLoopTripCountAnnotator::RunImpl(
...
-                instr->backend_config<WhileLoopBackendConfig>();
-            existing_config.ok()) {
-          config = *existing_config;
-        }
+      WhileLoopBackendConfig config;
+      if (auto existing_config =
+              instr->backend_config<WhileLoopBackendConfig>();
+          existing_config.ok()) {
+        config = *existing_config;
+      }
```
allanrenucci: Can we return an error if `WhileLoopBackendConfig` is non empty? I don't think we should handle the case where `WhileLoopBackendConfig` is pre-populated since we are trying to consolidate the `WhileLoopBackendConfig` creation logic into `WhileLoopTripCountAnnotator`.

Something along the lines of:
```cc
RETURN_IF_ERROR(auto config, instr->backend_config<WhileLoopBackendConfig>());
if (config != WhileLoopBackendConfig::default_instance()) {
  return absl::FailedPreconditionError(...);
}
```

allanrenucci (changes requested):

xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils_test.cc
sfvaroglu: Fixed

xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils_test.cc
sfvaroglu: Fixed

xla/service/host_offload_utils.cc
sfvaroglu: Done

xla/service/host_offload_utils.cc
sfvaroglu: Done

xla/hlo/transforms/while_loop_trip_count_annotator.h
sfvaroglu: Updated

xla/hlo/transforms/while_loop_trip_count_annotator.h
sfvaroglu: Done

xla/hlo/transforms/while_loop_trip_count_annotator.cc
sfvaroglu: Done

xla/backends/gpu/transforms/double_buffer_loop_unrolling_test.cc
```diff
@@ -1587,6 +1587,90 @@ ENTRY main {
...
+    }
+
+    ASSERT_TRUE(dv_init_step.contains(0));
+    EXPECT_EQ(dv_init_step[0].second, 2)
+        << "Primary variable step should be doubled after double buffering";
+
+    ASSERT_TRUE(dv_init_step.contains(3));
+    EXPECT_EQ(dv_init_step[3].second, 2)
+        << "Dynamic variable step should be doubled after double buffering";
+  }
```
allanrenucci: Can we use `EqualsProto` to make this more readable? I.e.

```cpp
EXPECT_THAT(config, EqualsProto(R"pb(
  known_init_step { ... }
  dynamic_variables { ... }
  ...
)pb"));
```

xla/hlo/transforms/while_loop_trip_count_annotator_test.cc
```diff
@@ -224,6 +225,107 @@ TEST_F(TripCountAnnotatorTest, Int64Overflow) {
...
+  TF_ASSERT_OK_AND_ASSIGN(auto config,
+                          m->entry_computation()
+                              ->root_instruction()
+                              ->backend_config<WhileLoopBackendConfig>());
+  ASSERT_EQ(config.dynamic_variables_size(), 1);
+  EXPECT_EQ(config.dynamic_variables(0).tuple_index(), 1);
+  ASSERT_TRUE(config.dynamic_variables(0).has_init());
+  ASSERT_TRUE(config.dynamic_variables(0).has_step());
+  EXPECT_EQ(config.dynamic_variables(0).init(), 5);
+  EXPECT_EQ(config.dynamic_variables(0).step(), 1);
```
allanrenucci: Can we use EqualsProto to make this more readable?

xla/hlo/transforms/while_loop_trip_count_annotator_test.cc
```diff
@@ -224,6 +225,107 @@ TEST_F(TripCountAnnotatorTest, Int64Overflow) {
...
+  TF_ASSERT_OK_AND_ASSIGN(auto config,
+                          m->entry_computation()
+                              ->root_instruction()
+                              ->backend_config<WhileLoopBackendConfig>());
+  ASSERT_EQ(config.dynamic_variables_size(), 1);
+  EXPECT_EQ(config.dynamic_variables(0).tuple_index(), 1);
+  ASSERT_TRUE(config.dynamic_variables(0).has_init());
+  ASSERT_TRUE(config.dynamic_variables(0).has_step());
+  EXPECT_EQ(config.dynamic_variables(0).init(), 0);
+  EXPECT_EQ(config.dynamic_variables(0).step(), 1);
```
allanrenucci: Can we use EqualsProto to make this more readable?

xla/hlo/transforms/while_loop_trip_count_annotator_test.cc
```diff
@@ -517,5 +619,38 @@ TEST_F(TripCountAnnotatorTest, InductionVarNonZeroTupleIndexForwarded) {
...
+      ROOT while_loop = (s32[]) while(initial_tuple), condition=Cond, body=Body,
+        backend_config={"known_trip_count":{"n":"99"}}
+    })";
+
+  ASSERT_OK_AND_ASSIGN(auto m, ParseAndReturnVerifiedModule(kModuleStr));
+  WhileLoopTripCountAnnotator pass;
+  auto status_or_changed = RunHloPass(&pass, m.get());
+  EXPECT_FALSE(status_or_changed.ok());
+  EXPECT_EQ(status_or_changed.status().code(),
+            absl::StatusCode::kFailedPrecondition);
```
allanrenucci: Can be combined into a single statement. I.e. `EXPECT_THAT(RunHloPass(&pass, m.get()), StatusIs(absl::StatusCode::kFailedPrecondition))`

xla/service/collective_pipeliner_test.cc
```diff
@@ -5719,12 +5717,12 @@ ENTRY %main.117 (Arg_0.1: f32[10,1000,8000], Arg_1.2: f32[10,8000,1000], Arg_2.3
...
 
-  std::set<int64_t> dynamic_indices(
-      config.dynamic_variable_tuple_indices().begin(),
-      config.dynamic_variable_tuple_indices().end());
-
-  std::set<int64_t> expected_indices = {0, 5};
+  absl::flat_hash_set<int64_t> dynamic_indices =
+      host_offload_utils::CollectDynamicVariableTupleIndices(while_loops[0]);
+  absl::flat_hash_set<int64_t> expected_indices = {0, 5};
   EXPECT_EQ(dynamic_indices, expected_indices);
```
allanrenucci: You can do:
```cpp
EXPECT_THAT(dynamic_indices, UnorderedElementsAre(0, 5));
```

xla/service/collective_pipeliner_test.cc
```diff
@@ -5828,12 +5824,12 @@ ENTRY %main.117 (Arg_0.1: f32[10,1000,8000], Arg_1.2: f32[10,8000,1000], Arg_2.3
...
 
-  std::set<int64_t> dynamic_indices(
-      config.dynamic_variable_tuple_indices().begin(),
-      config.dynamic_variable_tuple_indices().end());
-
-  std::set<int64_t> expected_indices = {0, 8};
+  absl::flat_hash_set<int64_t> dynamic_indices =
+      host_offload_utils::CollectDynamicVariableTupleIndices(while_loops[0]);
+  absl::flat_hash_set<int64_t> expected_indices = {0, 8};
   EXPECT_EQ(dynamic_indices, expected_indices);
```
allanrenucci: Let's use UnorderedElementsAre

xla/service/host_offload_utils_test.cc
```diff
@@ -363,6 +365,88 @@ TEST_F(HostOffloadUtilsTest, ConditionalReusedComputationPredecessorsTest) {
...
+  ROOT loop = (s32[], f32[10,8], f32[10,8]) while(init), body=body, condition=cond
+}
+)hlo";
+  TF_ASSERT_OK_AND_ASSIGN(auto module,
+                          ParseAndReturnVerifiedModule(hlo_string));
+  HloInstruction* loop = FindInstruction(module.get(), "loop");
+  ASSERT_NE(loop, nullptr);
+  absl::flat_hash_set<int64_t> got = CollectDynamicVariableTupleIndices(loop);
+  absl::flat_hash_set<int64_t> expected = {0};
+  EXPECT_EQ(got, expected);
```
allanrenucci, line 447: Let' use `UnorderedElementsAre`

xla/hlo/transforms/while_loop_trip_count_annotator.cc
```diff
@@ -31,10 +35,126 @@ limitations under the License.
...
+  }
+  std::optional<LinearTerm> term =
+      DecomposeLinear(root->operand(tuple_idx), param);
+  if (!term.has_value() || !term->source_tuple_idx.has_value()) {
+    return std::nullopt;
+  }
+  return AffineMatch{*term->source_tuple_idx, term->constant_offset};
+}
+
+std::optional<std::pair<int64_t, int64_t>> ResolveAffine(
```
allanrenucci, line 143: It is not clear to me what this function does from the name only. Can you add a doc comment?

allanrenucci (changes requested):
LGTM overall. Just a few style nits. Thanks!

xla/backends/gpu/transforms/double_buffer_loop_unrolling_test.cc
sfvaroglu: Done

allanrenucci (approved):
Thank you!

xla/service/host_offload_utils_test.cc
```diff
@@ -15,10 +15,12 @@ limitations under the License.
 
 #include "xla/service/host_offload_utils.h"
 
+#include <cstdint>
 #include <memory>
 #include <vector>
 
 #include <gtest/gtest.h>
+#include "absl/container/flat_hash_set.h"
```
allanrenucci, line 23: This is missing the corresponding dependency in the BUILD file

xla/hlo/transforms/while_loop_trip_count_annotator.cc
```diff
@@ -76,50 +216,75 @@ absl::StatusOr<bool> WhileLoopTripCountAnnotator::RunImpl(
...
+
+      TF_ASSIGN_OR_RETURN(WhileLoopBackendConfig existing_config,
+                          instr->backend_config<WhileLoopBackendConfig>());
+      if (existing_config.ByteSizeLong() != 0) {
+        return absl::FailedPreconditionError(absl::StrFormat(
+            "WhileLoopTripCountAnnotator is the sole creator of "
+            "WhileLoopBackendConfig, but %s already has a non-empty "
+            "backend_config: %s",
+            instr->name(), existing_config.ShortDebugString()));
+      }
```
allanrenucci, line 232: This is causing `functional_hlo_runner_test` to fails. #23786 introduced a use case for overwriting the `WhileLoopBackendConfig`. How about we turn this into a warning instead? And we document that `WhileLoopTripCountAnnotator` will overwrite any existing config.

allanrenucci (changes requested):
Please fix CI errors. DYWY seems unrelated and can be ignored.

xla/service/host_offload_utils_test.cc
sfvaroglu: Done

xla/hlo/transforms/while_loop_trip_count_annotator.cc
sfvaroglu: Done

allanrenucci (approved):
Let's try internal import again

allanrenucci:
@sfvaroglu, can you please rebase this PR?

sfvaroglu:
> @sfvaroglu, can you please rebase this PR?

https://github.com/openxla/xla/pull/42396 is causing NaNs in MaxText-llama3 training. I'll do a rebase once it gets fixed/reverted so I can test this PR end-to-end.

sfvaroglu:
@allanrenucci Rebased, please take a look

allanrenucci (approved):

PR #39268, opened 2026-03-16
[ROCm] Fix invalid isSubset implementation, missing incarnations_ comparison

alekstheod:
📝 Summary of Changes

🎯 Justification
Add missing incarnations comparison in isSubsetOf the clique key.
Fix timeout on jax test target //tests:image_test_gpu

This change: https://github.com/openxla/xla/pull/37063
Uncovered a bug in IsSubsetOf implementation where comparison
to incarnations was missing leading to triggering ncclAbort on the old
cached cliques and consequently to a deadlock.

```
  Test Run 1:
  - JAX backend created with incarnation=X
  - Creates clique [GPU0, incarnation=X]
  - Clique stored in process-level cache
  - Test finishes

  Test Run 2 (same process):
  - JAX backend re-created with NEW incarnation=Y
  - Tries to create clique [GPU0, incarnation=Y]
  - XLA finds old clique [GPU0, incarnation=X] in cache
  - Buggy IsSubsetOf() thinks [GPU0, X] is subset of [GPU0, Y]
    (because it only checks devices, not incarnations)
  - Calls Abort() on old clique → HANG!

  The cache persistence is intentional because:
  - Destroying NCCL/RCCL communicators during process lifetime can be dangerous
  - The cliques might still be in use by pending operations
  - It's designed to handle process restarts via incarnations, not in-process reinitializations

  Bottom line: The clique cache is NOT supposed to be cleaned up between test runs in the same process. The incarnation mechanism is specifically designed to handle this - but the bug in IsSubsetOf() broke that mechanism by not checking incarnations when determining if cliques are subsets of each other.
```

✻ Brewed for 3m 1s

🚀 Kind of Contribution
Please remove what does not apply: 🐛 Bug Fix

📊 Benchmark (for Performance Improvements)
Not relevant

🧪 Unit Tests:
Jax-u-tests pipeline (got new test target timeout)

🧪 Execution Tests:
CI tests

PR #39353, opened 2026-03-17
[ROCm] Handle empty incranation case in IsSubsetOf function

alekstheod:
📝 Summary of Changes
Follow up of this fix: https://github.com/openxla/xla/pull/39268
There might be a case with an empty incarnations list.
In this IsSubsetOf must return false.

🎯 Justification
Issue reported by NV in this PR: https://github.com/openxla/xla/pull/39268
From the logs we see the incarnations list is empty. In this case we handle the clique
as a subset, which leads to a deadlock.

🚀 Kind of Contribution

Not relevant

🧪 Unit Tests:
CI

🧪 Execution Tests:
CI.

### DIFF
--- a/xla/backends/gpu/collectives/gpu_clique_key.cc
+++ b/xla/backends/gpu/collectives/gpu_clique_key.cc
@@ -99,6 +99,7 @@ bool GpuCliqueKey::IsSubsetOf(const CliqueKey& other) const {
                           return absl::c_linear_search(other_gpu->devices(),
                                                        id);
                         }) &&
+         !incarnations_.empty() &&
          absl::c_all_of(incarnations_, [&](IncarnationId id) {
            return absl::c_linear_search(other_gpu->incarnations_, id);
          });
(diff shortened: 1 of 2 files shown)


alekstheod:
@olupton can you check if this patch will fix your issue?

hawkinsp:
I can confirm this PR fixes this repro that I described in #37063  (`python tests/pjit_test.py ArrayPjitTest.test_wsc_vmap_abstract_mesh ShardingInTypesTest.test_conv_general_dilated`).

The entire `pjit_test.py` also now passes on 8xH100, which is a good sign.

xla/backends/gpu/collectives/gpu_clique_key.cc
```diff
@@ -99,6 +99,7 @@ bool GpuCliqueKey::IsSubsetOf(const CliqueKey& other) const {
                           return absl::c_linear_search(other_gpu->devices(),
                                                        id);
                         }) &&
+         !incarnations_.empty() &&
```
ezhulenev, line 102: Hm, I'm not sure it's correct. It's ok to have empty incarnations, with your change cliques with empty incarnations always return false, which is wrong.

I think existing check is correct, and the root cause is in some other 0place.

alekstheod:
Closing according to @ezhulenev

PR #39372, opened 2026-03-17
[xla:gpu] Fix NCCL comm split deadlock

ezhulenev:
Fixx deadlock in GPU collective clique acquisition caused by overly aggressive abandon predicate.

**Root cause:** When sub-cliques were split from a large parent (e.g. 8-device), and later re-acquired with a smaller `split_from` candidate (e.g. 4-device), the old `HasParent()` method compared raw `GpuClique*` pointers. Since the parent pointer (8-device clique) didn't match the split_from pointer (4-device clique), the code unnecessarily abandoned the existing sub-clique and attempted to re-split. This caused deadlock because only some ranks would abandon while others kept their cliques, breaking the collective rendezvous.

**Fix:** Replace pointer-based `HasParent()` with key-based `IsParentSupersetOf()` on `LockableGpuClique`. The new method checks whether the existing clique's parent key is a superset of the proposed `split_from` key. If so, the clique was created from a bigger (or equal) split and is still valid — no abandon needed. Only abandon when the new split candidate is strictly larger than the existing parent.


hawkinsp (approved):
I verified this solved the original issue with the JAX pjit_test.py on 8xH100.

PR #39386, opened 2026-03-18
Fix use-after-free of temporary Literal in LinearizeInto

shawnwang18:
LinearizeInto captured a LiteralSlice (non-owning view) by value in an async H2D transfer closure scheduled on a thread pool. When the caller passed a temporary Literal (e.g. via BufferFromHostLiteral), the temporary was destroyed after the call returned — before the thread pool task ran cuMemcpyHtoDAsync — leaving the closure with a dangling pointer to freed heap memory that could be reallocated for a subsequent Literal.

Fix by cloning the literal into an owned Literal before scheduling the async task, ensuring the data remains valid for the duration of the H2D transfer.

Add BufferFromHostLiteralTemporaryE2ETest to reproduce the bug: two scalar buffers created from temporaries, whose heap memory can alias under the default BFC allocator, causing incorrect values without the fix.

  ---
  📝 Summary of Changes

  PjRtStreamExecutorClient::LinearizeInto now clones the LiteralSlice argument into an owned Literal before scheduling the async H2D transfer closure on the thread pool. The owned copy is captured by move into the closure, guaranteeing the literal data remains alive until cuMemcpyHtoDAsync completes.

  🎯 Justification

  This is a correctness bug affecting any caller of BufferFromHostLiteral that passes a temporary Literal. The use-after-free is a data race: the freed heap memory can be reallocated for a second Literal before the first H2D transfer runs, causing the device buffer to silently receive the wrong data. The bug is latent in all allocator configurations (BFC, platform, VMM) since the race depends on thread pool
  scheduling, not the allocator.

  🚀 Kind of Contribution

  🐛 Bug Fix

  N/A

  🧪 Unit Tests

  No new unit tests. The fix is a one-line change to ownership semantics in LinearizeInto; correctness is validated by the execution test below.

  🧪 Execution Tests

  Added StreamExecutorGpuClientTest.BufferFromHostLiteralTemporaryE2ETest in xla/pjrt/gpu/se_gpu_pjrt_client_test.cc.

  The test creates a GPU client with the default BFC allocator (single GPU), calls BufferFromHostLiteral twice with temporary scalars 1.0f and 2.0f, runs a scalar-add HLO, and asserts the result is 3.0f. Without the fix, the two temporaries' heap allocations alias under the BFC allocator, causing both device buffers to hold 2.0f and the result to be 4.0f instead.

### DIFF
--- a/xla/pjrt/common_pjrt_client.cc
+++ b/xla/pjrt/common_pjrt_client.cc
@@ -155,11 +155,20 @@ CommonPjRtClient::BufferFromHostLiteral(const LiteralSlice& literal,
                       AllocateRawBuffer(memory_space, on_device_bytes_count,
                                         /*retry_on_oom=*/true,
                                         /*allocate_after=*/{}));
+  // Clone the literal so its data remains valid until the H2D stream has
+  // processed all enqueued operations. GpuTransferManager stages the copy via
+  // DoHostCallback, which runs asynchronously on the stream, so the data must
+  // outlive the lambda in LinearizeInto. The caller may have passed a temporary
+  // whose lifetime does not extend past this function call.
+  auto literal_owner = std::make_shared<Literal>(literal.Clone());
   TF_ASSIGN_OR_RETURN(
       auto definition_event,
-      LinearizeInto(literal, device_shape,
+      LinearizeInto(*literal_owner, device_shape,
                     HostBufferSemantics::kImmutableUntilTransferCompletes,
                     raw_buffer));
+  // Keep the clone alive until the definition event fires (after the stream
+  // has completed all transfers, including DoHostCallback-based staging).
+  definition_event->AndThen([literal_owner = std::move(literal_owner)] {});
   return DefineBuffer(device_shape, memory_space, std::move(raw_buffer),
                       {std::move(definition_event)});
 }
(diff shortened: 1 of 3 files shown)


akuegel (approved):

pifon2a (approved):

thcmbs (approved):

akuegel:
It looks like there are some test failures after this change. For example //xla/backends/gpu/codegen/triton:fusion_emitter_device_test

Can you please check whether you can reproduce the problem?

akuegel (approved):

cezheng:
I am going to revert this PR because it breaks our CI since it undesirably doubles the memory usage.

The API contract clearly stated it is the caller's responsibility to keep the Literal alive, not unconditionally forcing a copy inside the API: https://github.com/openxla/xla/blob/3e1945dc92ec3bd80965272c246474eabbd04128/xla/pjrt/pjrt_client.h#L959-L961

akuegel:
> I am going to revert this PR because it breaks our CI since it undesirably doubles the memory usage.
> 
> The API contract clearly stated it is the caller's responsibility to keep the Literal alive, not unconditionally forcing a copy inside the API:
> 
> https://github.com/openxla/xla/blob/3e1945dc92ec3bd80965272c246474eabbd04128/xla/pjrt/pjrt_client.h#L959-L961

Good point, I missed that when reviewing.

PR #39435, opened 2026-03-18
[xla:gpu] Add async thunk passes for optimizing async execution

ezhulenev:
Thunk passes that correct LHS scheduling decisions if buffer assignment permits it:

- **Implement `RemoveRedundantAsyncThunkPass`** — removes redundant async start/done thunk pairs where an `AsyncDoneThunk` immediately follows its matching `AsyncStartThunk`. In this case there is no actual asynchronous execution (the done thunk just waits for an event recorded right before it), so we inline the nested thunk sequence from the start thunk, avoiding the overhead of creating an async execution scope (recording events and synchronizing streams).

- **Implement `ExpandAsyncScopeThunkPass`** — a new thunk pass that widens async execution scopes by moving `AsyncStartThunk` as far up the thunk sequence as possible and `AsyncDoneThunk` as far down as possible, maximizing the overlap window between async operations (e.g., collectives) and compute.

- **Conflict detection** uses `BufferUse::ReadWriteSet` and `ResourceUse::ReadWriteSet`, with buffer and resource uses collected transitively via the `Thunk::Walk` API. `AsyncDoneThunk` (which reports no uses of its own) inherits the uses of all matching `AsyncStartThunk`s to handle pipelined async chains correctly.

### DIFF
--- a/xla/backends/gpu/runtime/async_thunk_passes_test.cc
+++ b/xla/backends/gpu/runtime/async_thunk_passes_test.cc
@@ -1,0 +1,511 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/backends/gpu/runtime/async_thunk_passes.h"
+
+#include <cstdint>
+#include <memory>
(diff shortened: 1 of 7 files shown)


xla/backends/gpu/runtime/async_thunk_passes_test.cc
seantalts, line 1: Could you add a couple of tests (I think these might fail today)

1. Redundant pair nested inside a non-redundant AsyncStartThunk: An AsyncStartThunk whose nested sequence contains [start_inner, done_inner], where the outer start and done are not adjacent. The inner pair should still get inlined.
2. Same for ExpandAsyncScopeThunkPass: An AsyncStartThunk whose nested sequence contains [kernel, start, kernel, done, kernel]. The inner scope should get expanded.

Probably also good to have a test that combines the passes and makes sure they interact well, and also test multiple independent async pairs.

seantalts (changes requested):
otherwise looks good to me

ezhulenev:
Looks like when we already have thunk sequence it's too late to optimize it, it must be done earlier at HLO level. Abandoning it for now.

PR #12050, opened 2024-05-02
[XLA:GPU] Force nccl command buffer to update even memory pointers do not change

shawnwang18:
This PR force command buffer to do cuda graph update if it contains nccl collective operators, because NCCL requires all devices to make the collective call even in the cuda-graph capturing mode.

PR #39449, opened 2026-03-18
[ROCm] Cache traced command buffers for collective operations

phambinhfin:
## Summary of Changes
CollectiveCmd::RecordTracedCommand calls TraceCommandBufferFactory::Create on every update, which triggers a full stream capture cycle (StreamBeginCapture -> collective op -> StreamEndCapture -> GraphInstantiate) each time buffer allocations change. Since allocations change frequently between iterations, this re-capture happens on nearly every iteration.

TracedCommandBufferCmd (used by GEMM, cuBLAS, cuDNN commands) already solves this with a TracedCommandBuffer LRU cache that stores previously captured graphs keyed by buffer addresses and only re-traces on cache miss. This change applies the same caching to CollectiveCmd, but with a critical addition: a xla::Rendezvous gate that coordinates the cache decision across all local participants.

Unlike GEMM or cuDNN, collective operations require all ranks to call into NCCL together during stream capture (PR #12050 documents this: "NCCL requires all devices to make the collective call even in the cuda-graph capturing mode"). A per-rank cache without coordination could cause some ranks to use cached graphs while others re-trace, leading to NCCL deadlocks.

**The Rendezvous gate works as follows:**

1. Each rank checks its local TracedCommandBuffer cache
2. All local participants vote via xla::Rendezvous: "do I have a cache hit?"
3. If ALL ranks have cache hits, all use their cached graphs (no NCCL call needed)
4. If ANY rank has a cache miss, ALL ranks re-trace together (NCCL calls are symmetric)
**Additional changes:**

- TracedCommandBuffer::HasEntry() for read-only cache lookup before the vote

- force_update() on CollectiveCmd to ensure all ranks always reach the Rendezvous (prevents skip_command_update from causing asymmetric participation)
- GpuCliqueKey passed through all collective subclass Record methods for Rendezvous keying
- Non-local cliques (multi-host) fall back to always-trace since in-process Rendezvous cannot coordinate across hosts
Performance on ROCm improves significantly:

Small workloads (1-2 collectives) | Improvement
-- | --
allreduce_only (1 psum) | 10.3× faster
matmul+allreduce (1 GEMM + 1 psum) | 9.4× faster
mlp+2x_allreduce (2 GEMM + 2 psum) | 13.9× faster

Heavy workloads (many compute + collective) | Improvement
-- | --
transformer_10L (~60 GEMM + 20 psum) | 8.4× faster

## 🎯 Justification

Without caching, each collective update triggers a full stream capture and graph instantiation cycle. With caching, subsequent iterations with the same buffer addresses return the cached graph instantly, eliminating the dominant overhead in collective command buffer execution.


i-chaochen:
Hi @ekuznetsov139 could you review this PR, please. Thanks!

i-chaochen (approved):

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -1439,28 +1457,98 @@ absl::Status CollectiveCmd::Prepare(const Thunk::PrepareParams& params) {
...
+              execute_params.command_buffer_trace_stream, trace));
+      nested_cmd_ptr = nested_cmd_owned.get();
+    }
+  } else {
+    TF_ASSIGN_OR_RETURN(
+        nested_cmd_owned,
+        se::TraceCommandBufferFactory::Create(
+            execute_params.stream->parent(),
+            execute_params.command_buffer_trace_stream, trace));
+    nested_cmd_ptr = nested_cmd_owned.get();
```
pemeliya, line 554: Looks like these two branches are identical - one can unify them somehow ? Also, I see that if we do not get a hit, the child graphs cache is not used? So, I suppose we need then a separate function to update the child graphs cache ?

Also, what would happen if the clique is not local? Then we are not going to use cache?

xla/backends/gpu/runtime/command_buffer_cmd.h
```diff
@@ -553,13 +556,16 @@ class CollectiveCmd : public AsyncStartCommand {
 
   bool requires_initialization() const final { return true; }
 
+  bool force_update() const final { return true; }
```
pemeliya: I think we do not need it now since the same functionality is achieved with 'requires_initialization' function ?

xla/backends/gpu/runtime/command_buffer_cmd.cc
phambinhfin: yes, make sense, let me work on this more!

xla/backends/gpu/runtime/command_buffer_cmd.h
phambinhfin: Here , requires_initialization() covers the setup path, force_update() covers the execution path. Both are needed.

requires_initialization() ensures collectives are always recorded during Initialize() (which runs before execution begins). But Initialize() is called once — subsequent iterations go through ExecuteOnStream(), where only force_update() controls whether the update phase runs.

xla/backends/gpu/runtime/command_buffer_cmd.cc
phambinhfin: 1. Unified the duplicate branches — restructured using a use_cache flag so TraceCommandBufferFactory::Create appears only once.

2. Fixed the cache not being populated on miss — added ForceTraceCommandBuffer() which always traces AND updates the cache. On a coordinated miss, all ranks call this, so the cache is populated for subsequent iterations.

3. Non-local cliques trace without caching. This is intentional — xla::Rendezvous is process-local and cannot coordinate across hosts.

phambinhfin:
@ezhulenev , could you please review this PR when you have a chance?
If there’s any concern with the current approach, I’d be happy to revise it.
Thanks!

ezhulenev:
No, it seems to be the good one, passing to @shawnwang18 who is unifying thunks and commands + traced commands. Will look closer tomorrow as well.

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -309,6 +310,82 @@ absl::StatusOr<se::CommandBuffer*> TracedCommandBuffer::GetOrTraceCommandBuffer(
...
+  }
+
+  for (size_t i = 0; i < capacity_; ++i) {
+    if (absl::c_equal(entries_[i].recorded_allocs, allocs) &&
+        entries_[i].command_buffer) {
+      return true;
+    }
+  }
+  return false;
+}
```
shawnwang18: This HasEntry implementation has code duplication with GetOrTraceCommandBuffer, try refactoring to avoid code duplicate.

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -309,6 +310,82 @@ absl::StatusOr<se::CommandBuffer*> TracedCommandBuffer::GetOrTraceCommandBuffer(
...
+
+absl::StatusOr<se::CommandBuffer*>
+TracedCommandBuffer::ForceTraceCommandBuffer(
```
shawnwang18: This function is mostly identical with GetOrTraceCommandBuffer, try do the refactoring.

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -1439,28 +1516,111 @@ absl::Status CollectiveCmd::Prepare(const Thunk::PrepareParams& params) {
...
+  se::CommandBuffer* nested_cmd_ptr = nullptr;
+  std::unique_ptr<se::CommandBuffer> nested_cmd_owned;
+
+  // For local cliques, coordinate the cache decision across all participants
+  // via Rendezvous to maintain NCCL call symmetry: either all ranks use
+  // cached graphs or all ranks re-trace together.
+  bool use_cache = false;
+  TracedCommandBuffer* traced_cmd = nullptr;
+
+  if (clique_key.is_local()) {
```
shawnwang18, line 943: If record_action is Create state, then no need to rendezvously vote, just treat it as cache miss.

shawnwang18:
The TracedCommandBufferCmd has been renamed to TracedCommand, and moved to traced_command.h/cc files, please rebase.

xla/backends/gpu/runtime/traced_command_buffer.h
```diff
@@ -55,13 +59,56 @@ class TracedCommandBuffer : public CommandState {
       se::Stream* stream, absl::FunctionRef<absl::Status(se::Stream*)> trace,
       se::StreamPriority priority = se::StreamPriority::Default);
 
- private:
-  std::vector<BufferAllocation::Index> allocs_indices_;
+  // Returns true if a cache entry already exists for the buffer addresses
+  // implied by `buffer_allocation`. Read-only; does not trace or mutate the
+  // cache. Used by collective operations to vote on cache state across ranks
+  // before deciding whether to use a cached graph or re-trace together.
+  bool HasEntry(const BufferAllocations* buffer_allocation) const;
```
ezhulenev, line 66: nit: HasEntry should be first function to keep tracing variants close to each other

xla/backends/gpu/runtime/traced_command_buffer.h
```diff
@@ -55,13 +59,56 @@ class TracedCommandBuffer : public CommandState {
...
+
+  // Shared implementation for `GetOrTraceCommandBuffer` and
+  // `ForceTraceCommandBuffer`.
+  // - When `force_retrace` is false: if a matching entry exists, returns the
+  //   cached command buffer; otherwise traces into an empty/evicted slot.
+  // - When `force_retrace` is true: always invokes `trace`. If a matching
+  //   entry exists it is overwritten in place; otherwise traces into an
+  //   empty/evicted slot. Used to keep ranks symmetric on a coordinated
+  //   cache miss.
+  absl::StatusOr<se::CommandBuffer*> TraceImpl(
```
ezhulenev, line 106: fwiw I hate `Impl` suffixes, just call it `Trace` :)

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -898,28 +901,123 @@ absl::Status CollectiveCmd::Prepare(const Thunk::PrepareParams& params) {
...
+    // Rendezvous vote stays symmetric and falls through to the
+    // ForceTraceCommandBuffer path (which populates the cache for subsequent
+    // RecordUpdate iterations).
+    if (!std::holds_alternative<RecordCreate>(record_action)) {
+      bool local_hit = traced_cmd->HasEntry(execute_params.buffer_allocations);
+
+      CollectiveTraceCacheKey rendezvous_key{clique_key, this};
+      TF_ASSIGN_OR_RETURN(
+          std::shared_ptr<bool> all_hit,
+          xla::Rendezvous<bool>(
```
ezhulenev, line 965: See [gpu_clique_rendezvous](https://github.com/openxla/xla/blob/main/xla/backends/gpu/collectives/gpu_clique_rendezvous.h), I think it will be a little shorter

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -898,28 +901,123 @@ absl::Status CollectiveCmd::Prepare(const Thunk::PrepareParams& params) {
...
+              },
+              /*warn_stuck_timeout=*/absl::Seconds(10),
+              /*terminate_timeout=*/absl::Seconds(30)));
+
+      use_cache = *all_hit;
+    }
+  }
+
+  if (use_cache) {
+    VLOG(5) << "Collective trace cache hit: all ranks using cached graph";
```
ezhulenev, line 980: instead of VLOG prefer XLA_VLOG_DEVICE when you ahve device ordinal in scope (I think here it must be accessible from params)

xla/backends/gpu/runtime/command_buffer_cmd.cc
ezhulenev: And extract it into static helper, if it will be shorter, if you have to pass nearly all parameteres, then don't bother, just let's try to keep rendezvous details our of the main code path

xla/backends/gpu/runtime/command_buffer_cmd.h
pemeliya: Ah, nope, I think we have
thunk_sequence.Initialize(...)
before every call to 
thunk_sequence.ExecuteOnStream(...)

So, in this case, collectives will be recorded twice for each CommandBufferThunk execution. 
The only difference is WhileThunk where Initialize() is called once but ExecuteOnStream() is called in a loop.

Otherwise, we won't have: 
bool requires_initialization() const final { return true; }
for collectives.

Anyway, I would check how many times collectives are getting recorded for each CommandBufferThunk execution.

phambinhfin: Thanks, let me check for it

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -860,28 +901,147 @@ absl::Status CollectiveCmd::Prepare(const Thunk::PrepareParams& params) {
...
+    // On RecordCreate the (this, command_buffer) TracedCommandBuffer cache was
+    // just constructed and is necessarily empty on every rank, so local_hit is
+    // uniformly false across the clique. SPMD execution guarantees every rank
+    // reaches RecordCreate for this command together, so skipping the
+    // Rendezvous vote stays symmetric and falls through to the
+    // ForceTraceCommandBuffer path (which populates the cache for subsequent
+    // RecordUpdate iterations).
+    if (!std::holds_alternative<RecordCreate>(record_action)) {
+      bool local_hit = traced_cmd->HasEntry(execute_params.buffer_allocations);
+      if (clique_key.num_local_participants() <= 1) {
```
ezhulenev: rendezvous has 1-participant optimization and skipped (needs fact checking, but I'm pettu sure it's there, at least for some versions), I'd just remove this comment + extra condition, it already gets pretty nested and complicated

xla/backends/gpu/runtime/command_buffer_cmd.cc
```diff
@@ -860,28 +901,147 @@ absl::Status CollectiveCmd::Prepare(const Thunk::PrepareParams& params) {
...
+        // Single local participant means there's nothing to coordinate: there
+        // is exactly one rank in this process for this clique, so its local
+        // decision is by definition the clique-wide decision. Skipping the
+        // process-local Rendezvous avoids one synchronization per iteration
+        // in the common production layout (one process per GPU).
+        use_cache = local_hit;
+      } else {
+        TF_ASSIGN_OR_RETURN(
```
ezhulenev: prefer regular ASSIGN_OR_RETURN on all new code + migrate existing ones in toucehd files

xla/backends/gpu/runtime/command_buffer_cmd.h
```diff
@@ -314,13 +331,30 @@ class CollectiveCmd : public Command {
...
   bool requires_initialization() const final { return true; }
 
+  // We intentionally do NOT force the update path on every ExecuteOnStream().
+  // requires_initialization() above already records the command on every
+  // execution, so the Rendezvous gate fires once per execution. Forcing an
+  // additional record from the Execute path is redundant (it doubles the
+  // number of Rendezvous round-trips) and was observed to corrupt parent
+  // command buffer state on long-running models like MaxText LLaMA 3 8B,
+  // causing segfaults after a few iterations.
+  bool force_update() const final { return false; }
```
ezhulenev: I think this is gone from base command after muy recent change?

ezhulenev (approved):
Few minor comments, but overall LGTM

phambinhfin:
@ezhulenev  @shawnwang18 Could you check the change again with the recent change

xla/backends/gpu/runtime/command_buffer_cmd.cc
phambinhfin: You're right — rendezvous.h:301-305 has the 1-thread fast path. Dropped the branch + comment.

xla/backends/gpu/runtime/command_buffer_cmd.cc
phambinhfin: Done

xla/backends/gpu/runtime/command_buffer_cmd.cc
phambinhfin: I extracted it into static helper to make codes more organize

ezhulenev (approved):

ezhulenev:
LGTM

ermilovmaxim (approved):

phambinhfin:
@beckerhe  @apivovarov  Could you take a look? thanks

i-chaochen:
Hi @beckerhe could you check this PR

dimitar-asenov (approved):

seantalts:
I see:

```
In file included from [third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc:16](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc?l=16&ws=tap-presubmit-server/596578162&snapshot=2):
[./third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.h:29](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.h?l=29&ws=tap-presubmit-server/596578162&snapshot=2):10: error: module //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer does not directly depend on a module exporting 'third_party/absl/types/span.h', which is part of indirectly-used module third_party/absl/types/span.h
see [http://go/cpp-features#layering_check](https://www.google.com/url?q=http://go/cpp-features%23layering_check&sa=D); to fix run:
	build_cleaner //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer
   29 | #include "third_party/absl/types/span.h"
      |          ^
[third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc:26](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc?l=26&ws=tap-presubmit-server/596578162&snapshot=2):10: error: module //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer does not directly depend on a module exporting 'third_party/absl/base/optimization.h', which is part of indirectly-used module third_party/absl/base/optimization.h
see [http://go/cpp-features#layering_check](https://www.google.com/url?q=http://go/cpp-features%23layering_check&sa=D); to fix run:
	build_cleaner //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer
   26 | #include "third_party/absl/base/optimization.h"
      |          ^
[third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc:34](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc?l=34&ws=tap-presubmit-server/596578162&snapshot=2):10: error: module //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer does not directly depend on a module exporting 'third_party/absl/types/span.h', which is part of indirectly-used module third_party/absl/types/span.h
see [http://go/cpp-features#layering_check](https://www.google.com/url?q=http://go/cpp-features%23layering_check&sa=D); to fix run:
	build_cleaner //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer
   34 | #include "third_party/absl/types/span.h"
      |          ^
[third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc:35](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/backends/gpu/runtime/traced_command_buffer.cc?l=35&ws=tap-presubmit-server/596578162&snapshot=2):10: error: module //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer does not directly depend on a module exporting 'third_party/gloop/util/status/status_macros.h', which is part of indirectly-used module third_party/gloop/util/status/status_macros.h
see [http://go/cpp-features#layering_check](https://www.google.com/url?q=http://go/cpp-features%23layering_check&sa=D); to fix run:
	build_cleaner //third_party/tensorflow/compiler/xla/backends/gpu/runtime:traced_command_buffer
   35 | #include "third_party/gloop/util/status/status_macros.h"
      |          ^
4 errors generated.
```

shawnwang18:
Hi, @phambinhfin We have migrated the CollectiveCmd to CollectiveThunk as part of migration process which move all command buffer functions to each Thunk.  and CollectiveCmd is no longer used, could you help also migrate your design to CollectiveThunk ? Really sorry for the inconvient.

phambinhfin:
@shawnwang18  Thanks for informing, I am working on it

phambinhfin:
Close this one since it does not gain performance well in the popular models.

PR #39475, opened 2026-03-19
[XLA:GPU] Adds a new construction mode (kCapture) for command buffer

shawnwang18:
Introduces a new kCapture construction mode for CommandExecutor alongside the existing kExplicit mode, enabling more efficient CUDA graph construction via inline stream capture.

  - ConstructionMode enum (kExplicit / kCapture) added to Command and exposed on CommandExecutor. Mode is automatically derived at construction time: kCapture is selected only when all commands in the sequence support it; otherwise falls back to kExplicit.
  - kExplicit (existing default): records each command individually as a graph node.
  - kCapture (new): executes all commands via ExecuteOnStream() within a single CommandBuffer::Trace() call, capturing them inline into the outer CUDA graph. This avoids child-graph overhead since captured nodes become direct nodes in the outer graph.
  - Command::Record() base implementation handles the kCapture path: on RecordCreate, it drives the trace using each command's ExecuteOnStream(); RecordUpdate is rejected in capture mode (inlined nodes cannot be atomically swapped).
  - CommandBuffer::Trace() moved to public API and its return type changed from vector<Command*> to absl::StatusOr<const Command*> (a single sink node, since Trace always produces exactly one GpuTraceCommand).
  - ToGraphNodeDependencies updated to handle GpuTraceCommand (mirrors existing GpuCaseCommand pattern), so trace commands can appear as dependencies without hitting the fatal fallthrough.

  🎯 Justification

  The existing kExplicit mode records each command as a separate CUDA child graph node. For command sequences that execute simple stream operations (e.g., kernel launches, memcpy), the child-graph indirection adds overhead. kCapture mode uses CUDA 12.3+ inline stream capture to fold these commands directly into the outer graph as first-class nodes, eliminating child-graph overhead and enabling more efficient graph
  execution for eligible command sequences.

  🚀 Kind of Contribution

  - Performance Improvement
  - New Feature

  (To be filled in — kCapture mode eliminates child-graph node overhead for eligible command sequences. Benchmarks pending.)

  🧪 Unit Tests

  (To be filled in.)

  🧪 Execution Tests

  (To be filled in.)

### DIFF
--- a/xla/backends/gpu/runtime/command.cc
+++ b/xla/backends/gpu/runtime/command.cc
@@ -16,6 +16,11 @@ limitations under the License.
 #include "xla/backends/gpu/runtime/command.h"
 
 #include <string>
+#include <variant>
+
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "xla/stream_executor/command_buffer.h"
 
 namespace xla::gpu {
 
@@ -27,6 +32,32 @@ std::string CommandTypeString(CommandType type) {
     XLA_GPU_COMMAND_LIST(CASE_CMD_STRING)
 #undef CASE_CMD_STRING
   }
+}
+
+absl::StatusOr<const se::CommandBuffer::Command*> Command::Record(
+    const Thunk::ExecuteParams& execute_params,
+    const RecordParams& /*record_params*/, RecordAction record_action,
+    se::CommandBuffer* command_buffer) {
+  if (construction_mode_ == ConstructionMode::kExplicit) {
+    return absl::UnimplementedError("Record is not implemented");
+  }
+
+  // kCapture mode
+  if (std::holds_alternative<RecordUpdate>(record_action)) {
+    return absl::UnimplementedError(
+        "RecordUpdate is not supported in kCapture construction mode.");
+  }
+
+  // RecordCreate: capture ExecuteOnStream into the command buffer via Trace.
+  auto& create = std::get<RecordCreate>(record_action);
+  se::Stream* trace_stream = execute_params.command_buffer_trace_stream;
+
+  return command_buffer->Trace(
+      trace_stream,
+      [this, &execute_params]() -> absl::Status {
+        return ExecuteOnStream(execute_params);
+      },
+      create.dependencies);
 }
 
 bool IsCollectiveCommand(CommandType type) {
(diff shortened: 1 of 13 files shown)


ezhulenev:
I don't think we should do that, this can be achieved today my "auto-generating" record function with TracedCommand. But even that is not a good idea, because how different resources are managed in thunks vs commands is different. In thunks "resources" attached to streams and owned by thunks, in commands some resources (i.e. traced graphs?) attached to commands.

Also this change assumes that update-free command buffers is the default option, but we'll have to see how well it works and if this can be a viable default. For now having an efficient update path that can update parts of the cuda graph that changed is crucial for perf.

shawnwang18:
> I don't think we should do that, this can be achieved today my "auto-generating" record function with TracedCommand. But even that is not a good idea, because how different resources are managed in thunks vs commands is different. In thunks "resources" attached to streams and owned by thunks, in commands some resources (i.e. traced graphs?) attached to commands.
> 
> Also this change assumes that update-free command buffers is the default option, but we'll have to see how well it works and if this can be a viable default. For now having an efficient update path that can update parts of the cuda graph that changed is crucial for perf.

@ezhulenev I would go back to this PR, with this PR, I think will have two more benefits. 

# 1. Can avoid host node bubble inside child graph

The main difference with the implementation in this PR vs original traced based commands is that the new PR allows to capture the commands into the **top level graph** , without the need to have a **Child** node to wrap them.  Becuase command buffer now can now support stable address, this is truly achievable. 

There is an perf issue with original **Child** wrapped graph is that cuda driver scoped the scheduling only within the child-graph, and as NCCL capturing will get a  host node, this will introduce a device bubble for host.  Capturing the NCCL operator directly into the top-level graph will allow cuda driver to scheduling the host node so can overlap with other kernels in the graph, this will remove the bubble introduced by host.

# 2. Gpu thunks now have a default command buffer support
 
  Gpu thunks naturally supports command buffer lowering if it can be cuda-graph captured, we can just have an default mode to lowering them, and if we still want to do manual lower, we can create an explicit rule to lower them. 

  For now, there are roughly 28 thunks out of 56 total thunk types  that have command buffer support, having a default lowering will make the remaining thunks command buffer lowerable.

akuegel:
@ezhulenev what do you think about the reply by @shawnwang18 ?

shawnwang18:
> @ezhulenev what do you think about the reply by @shawnwang18 ?

Maybe we can go back to discuss about this after the migration from command buffer command to thunk is finished, then it will be easy to support this mode I think.

PR #39537, opened 2026-03-20
[xla:gpu] Add hang watchdog to PJRT GPU client

ezhulenev:
Add a hang watchdog that aborts the process if PJRT client can't create NCCL communicator in a reasonable timeframe. Rendezvous termination does't catch timed out communicator initialization in PJRT client because for `device_put` we always create comms with a single local participant, and in this case there is no other threads that can watch the NCCL communicator init for time outs.

Re-use existing hang watchdog mechanism to guard agains potential infinite hangs.


mwhittaker (approved):
Nice!

akuegel (approved):
The internal version of this PR got deleted, reapproving to trigger a reimport.

PR #39693, opened 2026-03-23
Fix tracked_gpu_device_buffer_test in OSS by adding cuda_platform dep

and-ivanov:
📝 Summary of Changes
- Add `//xla/stream_executor/cuda:cuda_platform` dependency to `tracked_gpu_device_buffer_test`
- The test was failing with `RET_CHECK failure: !platforms.empty() No platforms found` because no GPU platform was registered

🎯 Justification
We successfully run many `no_oss`-tagged GPU tests and are working to separate the ones that genuinely cannot work in OSS from those that simply need minor fixes. The `no_oss` tag means "do not execute in public openxla/xla CI" but does not imply the test *cannot* work outside of it. This test only needed a missing platform dependency.


golechwierowicz (approved):

xla/pjrt/gpu/tfrt/BUILD
akuegel: I think you can remove this no_oss tag here. In this case it was added because at the time when the test was written, we did not support tests that require 2 GPUs to run. I think by now we support it, I saw other tests with requires-gpu-nvidia:2 and no no_oss tag.

xla/pjrt/gpu/tfrt/BUILD
```diff
@@ -357,6 +357,7 @@ xla_cc_test(
         "//xla/service:shaped_buffer",
         "//xla/stream_executor:device_address",
         "//xla/stream_executor:device_address_allocator",
+        "//xla/stream_executor/cuda:cuda_platform",
```
thomasjoerg, line 361:
```
The following dependency violations were found:
 - @//xla/pjrt/gpu/tfrt:tracked_gpu_device_buffer_test (not tagged cuda-only) depends on @//xla/stream_executor/cuda:cuda_platform (tagged cuda-only)


There are a couple of potential solutions for this/these violation(s):

1. Tag the dependent target with the same tag as the dependee.

2. If unavoidable make the dependency selective using the
   'if_{gpu|cuda|rocm}_is_configured' macro. This is discouraged
   outside of stream_executor.
```

thomasjoerg (changes requested):

xla/pjrt/gpu/tfrt/BUILD
and-ivanov: removed no_oss tag

xla/pjrt/gpu/tfrt/BUILD
and-ivanov: I added cuda-only tag, it seems to be in line with existing tags "gpu" and ""requires-gpu-nvidia:2". However, it is possible to remove cuda dependency from the test by not relying on [LocalClient](https://github.com/openxla/xla/blob/6d75a1ab6c97270c6b2963e3c97c6848c86f0ea6/xla/pjrt/gpu/tfrt/tracked_gpu_device_buffer_test.cc#L137) for shape computation.

akuegel: Adding cuda-only tag is ok for now.

akuegel (approved):

thomasjoerg (approved):

thomasjoerg:
This PR seems to break some tests that look related. Please have a look at:
https://github.com/openxla/xla/actions/runs/23595448043/job/68710869999

akuegel:
> This PR seems to break some tests that look related. Please have a look at: https://github.com/openxla/xla/actions/runs/23595448043/job/68710869999

@and-ivanov Ok, seems I was wrong and this test still does not work in our oss CI. Please re-add the no_oss tag.

thomasjoerg (approved):

akuegel (approved):

PR #39694, opened 2026-03-23
Make tfrt_gpu_buffer_test pass in OSS

and-ivanov:
📝 Summary of Changes
  - Fix `tfrt_gpu_buffer_test` to build and pass outside of google-internal CI
  - Replace `absl::down_cast` with `tsl::down_cast` (fixes build: `absl::down_cast` is not available in the OSS absl)
  - Add missing `cuda_platform` dep so the GPU platform is registered at test startup
  - Remove unused `absl/base` dep

🎯 Justification
We successfully run many `no_oss`-tagged GPU tests and are working to separate the ones that genuinely cannot work in OSS from those that simply need minor fixes. The `no_oss` tag means "do not execute in public openxla/xla CI" but does not imply the test *cannot* work outside of it. This test only needed a build fix and a missing dependency.

### DIFF
--- a/xla/pjrt/gpu/tfrt/tfrt_gpu_buffer_test.cc
+++ b/xla/pjrt/gpu/tfrt/tfrt_gpu_buffer_test.cc
@@ -22,7 +22,6 @@ limitations under the License.
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
-#include "absl/base/casts.h"
 #include "absl/status/status.h"
 #include "absl/status/status_matchers.h"
 #include "absl/status/statusor.h"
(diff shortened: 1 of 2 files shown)


xla/pjrt/gpu/tfrt/tfrt_gpu_buffer_test.cc
```diff
@@ -137,8 +134,7 @@ TEST_F(TfrtGpuBufferTest, AcquireExternalReference) {
 
 TEST_F(TfrtGpuBufferTest, ReleaseDeviceMemoryOwnershipNoWait) {
   Shape on_device_shape = ShapeUtil::MakeShapeWithType<int32_t>({4, 4});
-  TfrtGpuDevice* device =
-      absl::down_cast<TfrtGpuDevice*>(client_->devices()[0]);
+  TfrtGpuDevice* device = tsl::down_cast<TfrtGpuDevice*>(client_->devices()[0]);
```
golechwierowicz, line 137: We've been internally migrating to absl::down_cast on purpose because we want to deprecate tsl::down_cast.

Why is it not working in OSS? I see that it's supposed to be available: https://github.com/abseil/abseil-cpp/blob/master/absl/base/casts.h#L256.

and-ivanov: Currently XLA uses version [LTS 20250814.1](https://github.com/openxla/xla/blob/main/third_party/absl/workspace.bzl#L10) of absl, which [doesn't include](https://github.com/abseil/abseil-cpp/blob/d38452e1ee03523a208362186fd42248ff2609f6/absl/base/casts.h) down_cast.

golechwierowicz: Hey, it looks like the version of absl got updated in OSS and now there's `tsl::down_cast`. Can you rebase on main and confirm this change is no longer necessary?

akuegel:
I am pretty sure this PR is now obsolete, as we have switched to the new Abseil version that has absl::down_cast.

PR #39695, opened 2026-03-23
Fix xla aot compile gpu test

and-ivanov:
📝 Summary of Changes
  - Add missing `cuda_platform` dep so the GPU platform is registered
  - Disambiguate `LocalExecutable::Run()` calls with explicit `absl::Span<const ShapedBuffer* const>` (fixes build: overload resolution was ambiguous in
  OSS)
  - Gracefully skip tests when the GPU's compute capability doesn't match the AOT-compiled binary instead of failing with an opaque error

🎯 Justification
We successfully run many `no_oss`-tagged GPU tests and are working to separate the ones that genuinely cannot work in OSS from those that simply need minor fixes. The `no_oss` tag means "do not execute in public openxla/xla CI" but does not imply the test *cannot* work outside of it. This test needed a build fix, a missing dependency, and a skip for compute-capability mismatches (the AOT artifacts are compiled for a specific GPU architecture).

### DIFF
--- a/xla/service/BUILD
+++ b/xla/service/BUILD
@@ -6237,6 +6237,7 @@ xla_cc_test(
         "//xla/pjrt/proto:compile_options_proto_cc",
         "//xla/service/cpu:cpu_compiler",
         "//xla/service/gpu:backend_configs_cc",
+        "//xla/stream_executor/cuda:cuda_platform",
         "//xla/tsl/lib/core:status_test_util",
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/types:span",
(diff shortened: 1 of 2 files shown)


beckerhe (commented):
Hi @and-ivanov,
thanks for the fix.

I think a better approach would be to convert the xla_aot_compile_gpu_test from an `xla_cc_test` in Bazel to a `xla_test` with the backends set to `h100`. Then the CC mismatch checking won't be needed and `xla_test` also automatically includes `cuda_platform`, therefore adding the dependency is not needed.

and-ivanov:
Hi @beckerhe ! I have implemented the suggested changes. Let me know if you’d like any adjustments.

beckerhe (approved):

PR #39725, opened 2026-03-23
[xla] Add global per-process hang watchdog

ezhulenev:
There is no need to create hang watchdogs for different part of XLA, just one per-process instance is enough for most of the use cases.


kredd2506:
Since the tests now use the shared `HangWatchdog::Global()` instead of local instances, could there be interference between tests if they run in parallel? Or does gtest run them sequentially by default?

ezhulenev:
All tests run sequentially, though in randomized order, but they all are order-independent.

seantalts (approved):

ezhulenev:
Rebased + pushed

PR #39763, opened 2026-03-24
[ROCm] Extend select_threshold macro to be more selective

alekstheod:
📝 Summary of Changes
Extend select threshold macro to accept dictionary

🎯 Justification
We use select_threshold to handle changes in rocm interfaces changes in bazel 
dependencies. It appears that we need to handle more than one interface change
in currently supported distros. Therefore we introduce the handling through the dict

🚀 Kind of Contribution
Please remove what does not apply: ♻️ Cleanup

Not relevant

🧪 Unit Tests:
Not relevant

🧪 Execution Tests:
Not relevant

### DIFF
--- a/third_party/gpus/rocm/BUILD.tpl
+++ b/third_party/gpus/rocm/BUILD.tpl
@@ -601,9 +601,11 @@ cc_library(
 alias(
     name = "amd_comgr",
     actual = select_threshold(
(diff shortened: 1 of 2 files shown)


mminutoli:
@alekstheod : I like your solution better. There was a minor issue and I went ahead and fixed it.

beckerhe (approved):

PR #39788, opened 2026-03-24
[ROCM][NFC] Improving & making nvshmem tests platform-agnostic

pemeliya:
This PR makes nvshmem tests platform-agnostic in preparation for integrating shmem-like GPU library for ROCM (nvshmem counterpart).

Besides, I have also improved xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc to use "self-verifying" HLOs because the original version was written in a "very verbose manner".

I was also able to fix [nvshmem sendrecv subtest](https://github.com/ROCm/xla/blob/9dd07605b92d27cc642a32453f0d327c3f6f5098/xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc#L241). Yet, this requires changing [CALL_NVSHMEM_P2P](https://github.com/ROCm/xla/blob/d8fcde684f40e11a19b15955f2c1959e647faf8f/xla/backends/gpu/collectives/nvshmem_communicator.cc#L370). The point is, nvshmem are one-sided collectives: e.g. there is no need to call 'Recv' when the other side calls 'Send'. But I think this shall be modified by NV side.

🚀 Kind of Contribution
♻️ Cleanup🧪 Tests


xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -50,10 +54,13 @@ limitations under the License.
 
 namespace xla {
 namespace {
+
 class NvshmemGpuCollectivesTest : public ::testing::Test {};
 
 static const char* test_binary_name;
 
+// static const absl::flat_hash_map< absl::string_view, xla::ResourceType >
```
beckerhe, line 62: This looks unintended

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -166,6 +302,8 @@ absl::Status NvshmemCollectiveTestBody(int rank_id, int num_ranks,
   TF_QCHECK_OK(distributed_client->Connect());
 
   GpuClientOptions client_options;
+  // client_options.allocator_config.kind = GpuAllocatorConfig::Kind::kPlatform;
```
beckerhe, line 305: Unintended?

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
```diff
@@ -177,390 +315,19 @@ absl::Status NvshmemCollectiveTestBody(int rank_id, int num_ranks,
   xla::CompileOptions options;
   options.executable_build_options.mutable_debug_options()
       ->set_xla_gpu_experimental_enable_nvshmem(true);
-  options.executable_build_options.set_run_backend_only(true);
+  // options.executable_build_options.set_run_backend_only(true);
```
beckerhe, line 318: Unintended?

xla/service/gpu/gpu_memory_space_assignment.cc
```diff
@@ -45,6 +45,10 @@ const absl::NoDestructor<absl::flat_hash_set<HloOpcode>>
         HloOpcode::kAllGatherStart,
         HloOpcode::kAllGatherDone,
         HloOpcode::kReduceScatter,
+        HloOpcode::kRecv,
```
beckerhe, line 48: How is this related?

xla/pjrt/gpu/nvshmem_gpu_collectives_test.cc
pemeliya: Indeed, this was a left over after debugging, need to fix it

xla/service/gpu/gpu_memory_space_assignment.cc
pemeliya: This is needed for send-recv nvshmem tests. Otherwise, the buffers for send/recv won’t be placed to collective memory space. 

Though, I heard that nvshmem is going to be deprecated since NCCL offers the same functionality. So I am wondering if it makes sense to fix these nvshmem tests now?

pemeliya:
Converted this PR to draft since nvshmem is getting retired and we are not going to follow the same integration path with MORI

PR #39871, opened 2026-03-25
[ROCm] Fix bf16 upcast handling for libdevice calls.

zoranjovanovic-ns:
📝 Summary of Changes
arguments always up/down casted for bf16 - there are no native libdevice functions for bf16

🎯 Justification
with current implementation up/down cast are omitted for bf16, which is not correct

🚀 Kind of Contribution
Please remove what does not apply: 🐛 Bug Fix 🧪 Tests

🧪 Unit Tests:
Added new test:
//xla/backends/gpu/codegen/triton/transforms/tests:triton_xla_math_to_libdevice_rocm.mlir

🧪 Execution Tests:

### DIFF
--- a/xla/service/gpu/target_util.cc
+++ b/xla/service/gpu/target_util.cc
@@ -346,8 +346,8 @@ struct TargetDeviceFunction GetDeviceFunctionRoot(
 }  // namespace
 
 bool HasF16Implementation(TargetDeviceFunctionID func_id,
-                          llvm::Triple target_triple, PrimitiveType type) {
-  return target_triple.isAMDGPU() && type == F16 &&
+                          llvm::Triple target_triple) {
+  return target_triple.isAMDGPU() &&
          (func_id == TargetDeviceFunctionID::kAtan2 ||
           func_id == TargetDeviceFunctionID::kCbrt ||
           func_id == TargetDeviceFunctionID::kCos ||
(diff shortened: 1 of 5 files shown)


xla/service/gpu/target_util.cc
```diff
@@ -346,8 +346,8 @@ struct TargetDeviceFunction GetDeviceFunctionRoot(
 }  // namespace
 
 bool HasF16Implementation(TargetDeviceFunctionID func_id,
-                          llvm::Triple target_triple) {
-  return target_triple.isAMDGPU() &&
+                          llvm::Triple target_triple, PrimitiveType type) {
```
beckerhe: This change is not clear to me. The function is called `HasF16Implementation` - why does we need to check a type here?

zoranjovanovic-ns: Idea was to check for available implementations for all float 16 data types (e.g. F16 and BF16).

beckerhe: Ah okay. That makes sense, but I think you either need to rename this function to reflect that it also intended for BF16, or check the type at the call site. I would prefer the latter.

zoranjovanovic-ns: @beckerhe updated PR.

beckerhe (approved):
Thank you!

xla/service/gpu/target_util.h
```diff
@@ -29,6 +29,7 @@ limitations under the License.
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Value.h"
 #include "llvm/TargetParser/Triple.h"
+#include "mlir/IR/Types.h"
```
thomasjoerg: Please update the BUILD deps accordingly.

thomasjoerg (changes requested):

xla/service/gpu/target_util.h
zoranjovanovic-ns: This include is not necessary with latest version, thus removed.

zoranjovanovic-ns:
@thomasjoerg updated PR. Removed unnecessary file inclusion

thomasjoerg (approved):

PR #39924, opened 2026-03-26
[ROCm] Refactor Triton backend for AllReduce and Enable ROCm support

mfrancepillois:
📝 Summary of Changes
This PR enables the ROCm Triton backend for AllReduce (collective emitter).
To this end:
- Add a new passes to lower atomic operations `triton_xla.atomic_write` and `triton_xla.atomic_spin_wait` and `triton_xla.get_tid`. These passes rely on `extern_elementwise` triton operations. thereby avoiding the use of target specific inline assembly. The `extern_elementwise` ops are then caught later in the compiler pipeline and replaced by llvm intrinsics.
- Add missing API to rocm_executor RocmExecutor::CanEnablePeerAccessTo(int other_device_ordinal) (this API is required to enable collective_emitter thunk).

🎯 Justification
Prior to this PR, the `triton_xla.get_tid`, `triton_xla.atomic_write` and `triton_xla.atomic_spin_wait` operations were lowered using PTX assembly. Therefore, AllReduce triton backend was only available for CUDA target.
This PR replace the existing CUDA-only lowering strategy to lower these operations via Triton operations (`tt.extern_elementwise`), and then catch this operations, at the end of the triton pipeline, implement them using LLVM target dependent intrinsics.
This unifies the lowering of  `triton_xla.atomic_write` and `triton_xla.atomic_spin_wait` and `triton_xla.get_tid` operations for CUDA and ROCm targets. Thanks to that, the triton backend for AllReduce is now available for ROCm target.

🚀 Kind of Contribution

🧪 Unit Tests:
This PR includes a LIT test checking the lowering of atomic operations.

🧪 Execution Tests:
The existing test xla/tests:all_reduce_e2e_test is relevant for testing this support.


mfrancepillois:
The PTX assembly generated using LLVM intrinsics looks similar to one generated by the old pass (using inline assembly):
#### 1. __Thread ID Retrieval__ 

__Old (PTX Assembly):__

```ptx
// begin inline asm

    mov.u32 %r8, %tid.x;
  
// end inline asm
```

__New (Intrinsics):__

```ptx
mov.u32 	%r2, %tid.x;
```

#### 2. __Atomic Store__ 

__Old (PTX Assembly):__

```ptx
// begin inline asm

    st.global.sys.release.u32 [%rd25], %r12;
  
// end inline asm
```

__New (Intrinsics):__

```ptx
st.release.sys.global.b32 	[%rd30], %r4;
```

#### 3. __Atomic Spin-Wait Loop__ 

__Old (PTX Assembly):__

```ptx
// begin inline asm

    {
    .reg .pred %p<1>;
    .reg .b32 %r<1>;
    wait:
      ld.global.sys.acquire.u32 %r0, [%rd28];
      setp.lt.u32 %p0, %r0, %r12;
      @%p0 bra wait;
    }
  
// end inline asm
```

__New (Intrinsics):__

```ptx
$L__BB0_2:
	ld.acquire.sys.global.b32 	%r14, [%rd7];
	setp.lt.u32 	%p2, %r14, %r4;
	@%p2 bra 	$L__BB0_2;
```

sohaibiftikhar:
I haven't reviewed everything in detail but can we use the assembly counterpart for ROCm instead of carrying it all the way down using extern_elementwise?

mfrancepillois:
> I haven't reviewed everything in detail but can we use the assembly counterpart for ROCm instead of carrying it all the way down using extern_elementwise?

Thanks for your feedback.
We prefer not to use inline assembly for the ROCm backend. Another option is to use native Triton ops only (as was the case in [[ROCm] Enable ROCm Triton backend for AllReduce by mfrancepillois · Pull Request #39190 · openxla/xla](https://github.com/openxla/xla/pull/39190)) but we prefer this extern_elementwise->LLVM instrinsics approach which is more optimized.

However, for the CUDA backend, we have indeed several options:
1.	Keep the passes as there were, i.e., inline PTX inserted before the Triton pipeline
2.	Insert triton extern_elementwise ops before triton pipeline and then replace them by inline PTX at the end of the LLIR phase (this solution is available in commit [[ROCm] Refactor Triton backend for AllReduce and Enable ROCm support by mfrancepillois · Pull Request #39924 · openxla/xla](https://github.com/openxla/xla/pull/39924/changes/22565523bd8c0003c4303460b0dc03e4d4d5130c) ). This allows us to unify the `triton_xla_lower_atomics_pass` and `triton_xla_lower_get_tid_pass` for both backends.
3.	The solution presented in this PR, i.e., extern_elementwise->LLVM instrinsics approach for CUDA backend as well.

sohaibiftikhar:
I have thought about this a bit and I agree that the current approach using llvm intrinsics is more principled at least from the point of view of supporting intrinsics for multiple platforms. It does make testing a bit tricky for us since transformations happen a lot lower but we can solve that later.

I have a few high level comments.
1. I would like to split this change. The first part should only change the lowering passes but not the emitter itself. It would be a pure refactor of the passes that affects only CUDA. Then we add ROCM as a followup.
2. Create a centralized TritonExternElementwise function name utility that does the magic string handling. One way would be that we invoke parse on it and get back a std::variant of one of the accepted instructions. Then the caller visits each possible instruction type. The memory semantic validation and instruction validation stay within the parsing utility. A ToString or Serialize method on all types of this variant would take care of converting the type to its representation in high level pass. Then magic string parsing stays in a single method. You can add it to codegen/triton/. This can also be a separate precursor change to the other things here. That way we can test it in isolation and merging it does not break anything.
3. Use op rewrite patterns instead of walking the modules in the extern lowering implementation and doing string matching there. Unless I am missing something obvious this should be possible?
4. Avoid the _extern_ on the higher level lowering passes. That way we don't rename files so its easier to follow the history.

Nit: We also don't need `target_` inside the implementation as a class member. Just use target_backend directly as a class member.

mfrancepillois:
This PR can be closed as it has been reworked and split into multiple PRs as suggested. 
PR 1/4: https://github.com/openxla/xla/pull/40460
PR 2/4: https://github.com/openxla/xla/pull/40462
PR 3/4: https://github.com/openxla/xla/pull/40463
PR 4/4: https://github.com/openxla/xla/pull/40464

PR #39925, opened 2026-03-26
[ROCm] Remove deprecated hipCtx_t and HIP context APIs

phambinhfin:
## Summary
Remove deprecated hipCtx_t and HIP context APIs from the ROCm backend
- Simplify RocmContext to a header-only class with inline methods
- Make RocmContext a value field of RocmExecutor (no heap allocation)
- Remove 7 deprecated HIP context API calls
- Remove 6 deprecated API wrappers from rocm_driver_wrapper.h
- Delete rocm_context.cc
## Motivation
AMD has deprecated all hipCtx* and hipDevicePrimaryCtx* APIs since ROCm 1.9
with the warning "might not be supported in future releases." On ROCm,
hipCtx_t is a thin wrapper around a device ordinal -- the entire context
lifecycle is a no-op. A future ROCm release could remove these APIs and
break the build.
## Test plan
- Tested GPU device discovery (8x MI300X)
- Tested matmul, memory allocation, multi-GPU peer transfer, 
  device synchronization, rapid context switching across 8 GPUs


xla/stream_executor/rocm/rocm_executor.cc
```diff
@@ -1024,16 +1012,32 @@ fft::FftSupport* RocmExecutor::AsFft() {
...
   RocmExecutor* rocm_other = static_cast<RocmExecutor*>(other);
-  return EnablePeerAccess(rocm_context_, rocm_other->rocm_context_);
+  return EnablePeerAccess(&rocm_context_, &rocm_other->rocm_context_);
 }
 
-bool RocmExecutor::DeviceMemoryUsage(int64_t* free, int64_t* total) const {
-  return rocm_context_->GetDeviceMemoryUsage(free, total);
+bool RocmExecutor::DeviceMemoryUsage(int64_t* free_out,
+                                     int64_t* total_out) const {
+  CHECK_OK(ToStatus(wrap::hipSetDevice(rocm_context_.device_ordinal()),
```
draganmladjenovic: Use ScopedActivateContext.

phambinhfin: switched to ScopedActivateContext(&rocm_context_)

xla/stream_executor/rocm/rocm_executor.h
```diff
@@ -203,8 +204,11 @@ class RocmExecutor : public GpuExecutor {
   // GPU ISA version for device_.
   int version_;
 
-  // RocmContext for this device.
-  RocmContext* rocm_context_;
+  // RocmContext for this device.  Owned as a value — on ROCm a "context"
+  // is just a device ordinal, so there is no heavyweight object to manage.
+  // Mutable because SetActive() (hipSetDevice) is logically const — it does
+  // not change RocmContext state, but ScopedActivateContext takes non-const.
```
dimitar-asenov, line 210: I'm a little confused about this comment. Let's ignore the fact that the context is just an ordinal, that's an implementation detail. Doesn't activating the context "change it"? I don't fully understand the mismatch between the places where we are const-correct and where not.

To me it feels like SetActive should be logically non-const, but I'm not that familiar with that part of the code. Can you please clarify?

phambinhfin: Yes, that true, SetActive is logically non-const, it changes global GPU state via hipSetDevice().

The mutable is needed because DeviceMemoryUsage() is declared const in the StreamExecutor base class, but internally needs ScopedActivateContext, which takes a non-const Context*.

The original code had the same issue -- RocmExecutor::DeviceMemoryUsage() const called rocm_context_->GetDeviceMemoryUsage(), which called ScopedActivateContext(this) (non-const).

beckerhe (approved):

PR #39950, opened 2026-03-26
[ROCm] Fix hipblasLt Int8 GEMM support and autotuner output comparison

cj401-amd:
📝 Summary of Changes

- Fix autotuner workspace buffer comparison: restore extracting only the first tuple element from profiler output, excluding the workspace scratch buffer whose contents are non-deterministic across algorithm runs. Without this, all hipblasLt autotuning candidates are rejected as "WRONG RESULTS". (Regressed by  698df50c9e)                                                                                                                                                     
 - Add kI32 and kF64 computation types to AsHipblasComputeType() in hip_blas_utils.cc, matching the CUDA equivalent AsCublasComputeType(). Previously these hit LOG(FATAL), crashing on Int8 or F64 GEMMs routed through hipblasLt by the new autotuner.                                                                        
 - Add Int8 TYPED_MATMUL entries (HIP_R_8I/HIP_R_32I) to the hipblasLt matmul execution dispatch in hip_blas_lt.cc, matching cuda_blas_lt.cc. Without these, Int8 GEMMs failed at execution time with "Unexpected dtype".    

- //xla/backends/gpu/transforms:gemm_rewriter_test_amdgpu_any


xla/stream_executor/rocm/hip_blas_utils.cc
```diff
@@ -89,12 +89,19 @@ hipDataType AsHipblasDataType(blas::DataType type) {
...
-  else if (type == blas::ComputationType::kTF32AsF32)
-    return HIPBLAS_COMPUTE_32F_FAST_TF32;
-  else
-    LOG(FATAL) << "unsupported hipblaslt computation type";
+  switch (type) {
+    case blas::ComputationType::kF32:
+      return HIPBLAS_COMPUTE_32F;
+    case blas::ComputationType::kTF32AsF32:
+      return HIPBLAS_COMPUTE_32F_FAST_TF32;
+    case blas::ComputationType::kF64:
```
draganmladjenovic, line 97: You checked this on gfx90a?

shyamliagrawal:
> Fix autotuner workspace buffer comparison: restore extracting only the first tuple element from profiler output, excluding the workspace scratch buffer whose contents are non-deterministic across algorithm runs. Without this, all hipblasLt autotuning candidates are rejected as "WRONG RESULTS".

There was the same issue with cuBLAS(lt) and hence we added an option to set uses_last_output_for_scratch in  GpuCodegenBackend constructor, can you please use that for hipblasLt.

xla/stream_executor/rocm/hip_blas_utils.cc
cj401-amd: Yes. it works on gfx90a as shown

```
INFO: Analyzed target //xla/backends/gpu/transforms:gemm_rewriter_test_amdgpu_any (456 packages loaded, 45663 targets configured).
INFO: Found 1 test target...
Target //xla/backends/gpu/transforms:gemm_rewriter_test_amdgpu_any up-to-date:
  bazel-bin/xla/backends/gpu/transforms/gemm_rewriter_test_amdgpu_any
INFO: Elapsed time: 1108.794s, Critical Path: 1008.13s
INFO: 31972 processes: 15471 internal, 16501 local.
INFO: Build completed successfully, 31972 total actions
//xla/backends/gpu/transforms:gemm_rewriter_test_amdgpu_any              PASSED in 232.8s

root@hyd-7c-ZT13-03:/work/xla# ls /opt
rocm  rocm-7.1.1
```

cj401-amd:
> > Fix autotuner workspace buffer comparison: restore extracting only the first tuple element from profiler output, excluding the workspace scratch buffer whose contents are non-deterministic across algorithm runs. Without this, all hipblasLt autotuning candidates are rejected as "WRONG RESULTS".
> 
> There was the same issue with cuBLAS(lt) and hence we added an option to set uses_last_output_for_scratch in GpuCodegenBackend constructor, can you please use that for hipblasLt.

now we use `uses_last_output_for_scratch=true` in HipblasLtBackend as well.

shyamliagrawal (approved):

PR #39991, opened 2026-03-27
Add test coverage for BatchNormExpander untested code paths

kredd2506:
## Summary
Fixes #39930

- Add test for `BatchNormInference` expansion (previously had zero test coverage)
- Add test for `BatchNormInference` with sharding propagation
- Add test for `BatchNormGrad` with sharding propagation (training had it, grad didn't)
- Add test for disabled rewrite flags (verify pass is a no-op when all flags are false)

## Test plan
- [ ] CI verification (unable to build full CPU backend locally — ARM Mac under x86 emulation)
- [x] Tests follow existing patterns in the file (`HloPjRtTestBase`)
- [x] Follow Google C++ style

### DIFF
--- a/xla/service/batchnorm_expander_test.cc
+++ b/xla/service/batchnorm_expander_test.cc
@@ -157,6 +157,116 @@ ENTRY entry {
   }
 }
 
+// Test that we expand BatchNormInference.
+TEST_F(BatchNormExpanderTest, BatchNormInference) {
+  const char* module_str = R"(
+HloModule module
+ENTRY entry {
+  %param.0 = f32[2,2,2,2] parameter(0)
+  %param.1 = f32[2] parameter(1)
+  %param.2 = f32[2] parameter(2)
+  %param.3 = f32[2] parameter(3)
+  %param.4 = f32[2] parameter(4)
+  ROOT %batch-norm-inference = f32[2,2,2,2]
+    batch-norm-inference(f32[2,2,2,2] %param.0, f32[2] %param.1, f32[2] %param.2,
+                         f32[2] %param.3, f32[2] %param.4),
+    epsilon=0.001, feature_index=3
+})";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto m, ParseAndReturnVerifiedModule(module_str));
+  BatchNormExpander rewriter(/*rewrite_training_op=*/false,
+                             /*rewrite_inference_op=*/true,
+                             /*rewrite_grad_op=*/false);
+  ASSERT_TRUE(rewriter.Run(m.get()).value());
+  // BatchNormInference expands to a single result (not a tuple).
+  EXPECT_NE(m->entry_computation()->root_instruction()->opcode(),
+            HloOpcode::kBatchNormInference);
+}
+
+TEST_F(BatchNormExpanderTest, BatchNormInferenceSharding) {
+  const char* module_str = R"(
+HloModule module
+ENTRY entry {
+  %param.0 = f32[8,4] parameter(0)
+  %param.1 = f32[4] parameter(1)
+  %param.2 = f32[4] parameter(2)
+  %param.3 = f32[4] parameter(3)
+  %param.4 = f32[4] parameter(4)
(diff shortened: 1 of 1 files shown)


akuegel (commented):
This PR contains tests for AllGatherRemoveDegenerateDims. This does not seem to match the PR description?

xla/service/batchnorm_expander_test.cc
```diff
@@ -157,6 +157,140 @@ ENTRY entry {
...
+      continue;
+    }
+    auto device = instruction->sharding_unique_device();
+    ASSERT_TRUE(device);
+    EXPECT_EQ(*device, 1);
+  }
+}
+
+TEST_F(BatchNormExpanderTest, DisabledFlagsAreNoop) {
+  Shape input_shape = ShapeUtil::MakeShape(F32, {2, 2, 2, 2});
```
akuegel: Could you use HloText here like for the other two tests above?

xla/service/batchnorm_expander_test.cc
```diff
@@ -157,6 +157,140 @@ ENTRY entry {
   }
 }
 
+// Test that we expand BatchNormInference.
+TEST_F(BatchNormExpanderTest, BatchNormInference) {
+  Shape input_shape = ShapeUtil::MakeShape(F32, {2, 2, 2, 2});
```
akuegel: Could you please use HloText, like in the test below?

xla/hlo/transforms/collectives/all_gather_remove_degenerate_dims_test.cc
```diff
@@ -123,5 +123,59 @@ TEST_F(AllGatherRemoveDegenerateDimsTest, OnlyDropsMixedMinorDims) {
                             AllGatherRemoveDegenerateDims());
 }
 
+TEST_F(AllGatherRemoveDegenerateDimsTest, OnlyMinorDegenerateDims) {
```
akuegel, line 126: I see this is part of https://github.com/openxla/xla/pull/39931

Please revert from this PR

kredd2506:
Thanks for the review! Will fix all three — remove the AllGather file, and switch both tests to HLO text format. Pushing an update shortly.

kredd2506:
Updated — removed the AllGather file and rewrote both tests to use HLO text format.

Quick question: is there a preferred convention for when to use HLO text vs C++ builder in tests? I noticed the older tests in this file use the builder API — is the project moving towards HLO text for new tests?

akuegel:
> Updated — removed the AllGather file and rewrote both tests to use HLO text format.
> 
> Quick question: is there a preferred convention for when to use HLO text vs C++ builder in tests? I noticed the older tests in this file use the builder API — is the project moving towards HLO text for new tests?

Yes, we are moving towards HLO text for new tests. There are a lot of old tests that still use builder, but it is too much effort to migrate them. Maybe at some point it can be done with AI tools.
There is one possible exception I can think of where using builder API makes sense: if you want to construct a large HLO graph, it can be more efficient to do that via test logic and builder API.

akuegel (approved):
Thank you :)

kredd2506:
Thanks for the review and the insight on HLO text vs builder! That's really helpful.

I noticed the two older tests in this file (`BatchNormTraining` and `BatchNormGrad`) still use the builder API. Would a PR migrating those to HLO text be welcome? Seems like a good starting point for the migration you mentioned.

akuegel:
> Thanks for the review and the insight on HLO text vs builder! That's really helpful.
> 
> I noticed the two older tests in this file (`BatchNormTraining` and `BatchNormGrad`) still use the builder API. Would a PR migrating those to HLO text be welcome? Seems like a good starting point for the migration you mentioned.

It depends. If you do it manually, it is maybe not time well spent. Technically HLO text is a bit more readable, but normally an existing test doesn't have to be adjusted very often. So increasing the readability of existing tests does not have that large of an impact. However for newly added tests, it makes sense to use HLO text as that also makes it easier for the reviewer.

PR #40017, opened 2026-03-27
[xla:gpu] Correctly track thunk progress in presence of loops

ezhulenev:
The previous progress tracking design pre-allocated one `se::Event` per thunk and overwrote its timestamp on each execution. This meant that for thunks inside while loops, only the *last* iteration's state was visible -- earlier iterations were silently lost. When diagnosing deadlocks inside loops, the tracker could not tell which iteration hung or show the full execution history.

This PR redesigns the progress tracker to append a new event record for every thunk execution, preserving the complete chronological history including loop iteration state.

**Test coverage**: Extended `thunk_executor_test` to verify double-execution accumulation (2x event count, correct timestamps) and while loop nest tracking (trip count, loop depth, iteration number recorded per body thunk execution).

### DIFF
--- a/xla/backends/gpu/runtime/thunk_executor_test.cc
+++ b/xla/backends/gpu/runtime/thunk_executor_test.cc
@@ -30,6 +30,8 @@ limitations under the License.
 #include "xla/backends/gpu/runtime/memset_thunk.h"
 #include "xla/backends/gpu/runtime/sequential_thunk.h"
 #include "xla/backends/gpu/runtime/thunk.h"
+#include "xla/backends/gpu/runtime/while_loop.h"
+#include "xla/backends/gpu/runtime/while_thunk.h"
 #include "xla/service/buffer_assignment.h"
 #include "xla/service/gpu/buffer_allocations.h"
 #include "xla/service/platform_util.h"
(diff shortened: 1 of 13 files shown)


ermilovmaxim (approved):

xla/backends/gpu/runtime/thunk_executor_test.cc
```diff
@@ -134,6 +136,95 @@ TEST(SequentialThunkProgressTrackerTest, TrackProgress) {
...
+  body_thunks.push_back(std::make_unique<MemzeroThunk>(
+      ThunkInfo("loop_body_memzero"), ShapedSlice{slice, shape}));
+
+  ThunkSequence outer_sequence;
+  outer_sequence.push_back(std::make_unique<WhileThunk>(
+      ThunkInfo("while_loop"), slice, std::move(condition_thunks),
+      std::move(body_thunks), /*trip_count=*/kTripCount));
+
+  ThunkExecutor executor(std::move(outer_sequence));
+  ASSERT_OK(executor.Prepare(Thunk::PrepareParams()));
```
thomasjoerg, line 202: SanitizerError
UndefinedBehaviorSanitizer: null-pointer-use //xla/backends/gpu/runtime/thunk.h:254

Can you please take a look Eugene?

PR #40021, opened 2026-03-27
Add BF16 fallback for FP8 cuDNN convolutions without valid plans

kasper0406:
📝 **Summary of Changes**

Some FP8 convolution configurations have no cuDNN execution plans on some GPUs — e.g. grouped convolutions with fewer than 16 channels per group on sm_120 (RTX 5090 / RTX PRO 6000). Today the conv-fusion pipeline fails these with a hard error when the autotuner enumerates plans for the fusion.

This adds a `ConvFp8Fallback` pass that probes cuDNN at compile time and rewrites such FP8 conv fusions to BF16 — a clone of the `__cudnn$fusion` whose F8 shapes are BF16, with converts on the F8 operands and converts back to the original F8 types on the outputs — keeping the model compilable at reduced precision instead of failing.

This is a full rework of the original version of this PR, which modified the HLO from inside the autotuner. Per @derdrdirk's review the decision is now a compile-time pass driven by a cuDNN support probe, and the autotuner itself is untouched.

**Probing.** The pass probes with `CuDnnFusionCompiler::SupportsFusionDeviceless` (added in #45534, now merged), which needs only the target's `stream_executor::DeviceDescription`. The decision therefore works identically during JIT and AOT compilation, and the verdict cannot drift from what compiling the same fusion would do — probe and execution prepare plans through the same deviceless cuDNN graph path.

Verdicts other than `kUnsupported` leave the fusion unchanged. `kUnknown` (a cuDNN runtime too old for the target — cuDNN >= 9.8 in general, >= 9.19 for Blackwell targets — a graph the deviceless probe cannot model, or a cuDNN frontend failure) must not trigger a rewrite on a guess. The rewrite only fires if the BF16 replacement itself probes as supported; otherwise the FP8 fusion is kept and a warning is logged.

BF16 rather than F16 is the fallback because it preserves FP8's exponent range, so the rewrite cannot introduce overflow that the FP8 convolution would not have had.

**Placement.** The pass runs at the head of the autotuner pipeline (`NVPTXCompiler` now overrides `AddAutotunerPass`): after `ConvFusionRewriter` has formed the `__cudnn$fusion` instructions, and before the autotuner enumerates plans for them. Rewriting the convolution any earlier is not robust — `ConvFusionRewriter` runs inside the fusion pipeline, immediately ahead of `PriorityFusion` / `HloCSE` / `MultiOutputFusion`, and the inserted boundary converts are themselves fusable and would be folded straight back into an FP8 fusion. `FusionWrapper` in the pre-scheduling pipeline wraps the boundary converts for codegen.

The pass is a no-op unless conv fusions exist, i.e. under `--xla_gpu_experimental_enable_conv_fusion`. The legacy conv custom-call pipeline is deliberately not covered: its engine enumeration differs from the fusion pipeline's, so the probe verdict would not apply to it. This follows the plan in the `TODO(b/487265446)` in `nvptx_compiler.cc` for `ConvFusionRewriter` to become the default.

🎯 **Justification**

FP8 grouped convolutions with fewer than 16 channels per group have no valid cuDNN plans on RTX 5090 systems and cause JAX compilation failures. Falling back to BF16 keeps such models running instead of crashing.

🚀 **Kind of Contribution**

🐛 Bug Fix

📊 **Benchmark (for Performance Improvements)**

Not applicable — this trades precision for compilability on configurations that otherwise fail to compile at all. Convolutions that do have FP8 plans are left untouched.

🧪 **Unit Tests**

All deviceless, against checked-in target-config specs; run for every NVIDIA GPU generation via `backends = ["gpu"]`.

- Pure rewrite tests for the bare-FP8-conv, scale/clamp epilogue, and two-output amax fusion shapes produced by the real `ConvKindAssignment` + `ConvFusionRewriter` passes (so the tested shapes cannot drift from pipeline output): F8 stays only on the fusion boundary, the f32 accumulate and scale stay f32, and the modules still verify.
- Full-pass tests: f32 fusions and supported FP8 fusions (H100) are kept; FP8 fusions fall back to BF16 on a non-FP8-capable target (A100); and on a target where the BF16 replacement has no plans either (V100) the FP8 fusion is kept and the BF16 clone built for probing is removed again.
- sm_120 tests pinning the RTX 5090 / RTX PRO 6000 grouped-conv support surface that motivated the pass: 16 channels/group has plans and is kept FP8; 8 and 2 channels/group have no plans the conv-fusion pipeline can execute and fall back to BF16. These skip on cuDNN runtimes < 9.19, whose deviceless heuristics cannot probe Blackwell-generation targets.

🧪 **Execution Tests**

No convolution kernels are executed; every case is a compile-time probe.

The sm_120 cases skip on XLA's hermetic cuDNN (pinned at 9.8.0, which cannot probe Blackwell targets devicelessly), so they were additionally run against a local cuDNN 9.19 (`LOCAL_CUDNN_PATH`) on an L40:

- cuDNN 9.19: 9/9 pass, including both sm_120 grouped-conv cases.
- hermetic cuDNN 9.8.0: 7 pass, and the 2 sm_120 cases skip with "cuDNN runtime cannot probe conv graphs devicelessly for sm_120 targets (requires cuDNN >= 9.19)".


derdrdirk:
This changes lets the autotuner modify the HLO by adding FP8 to BF16 converts. The autotuner should not modify the HLO, but solely select a config. If cudnn doesn't support FP8 the autotuner should return 0 configs in GetSupportedConfigs. 

You should update the conv_rewriter.cc and xla/backends/gpu/transforms/conv_rewriter.cc  and xla/backends/gpu/transforms/conv_fusion_rewriter.h to to handle not supported fp8 oprations.

kasper0406:
Hi @derdrdirk, thanks for having a look! Makes sense. The original reason for me not including it in the compiler, was that I did not seem to find an easy way to probe actual runtime CUDA capabilities from there.
Currently `conv_rewriter.cc` and `conv_fusion_rewriter.h` do not have a stream-executor passed through, so I added a dedicated conv_fp8_fallback` pass which only job is to prove runtime convolution capabilities and do a fallback if required. What do you think about this approach?

derdrdirk:
The pass you added relies on the stream executor and thus would not work without a device. We are currently working on enabling ahead-of-time compilation for CuDNN, which would enable us querying CuDNN plans for convolutions without a device. The ETA for CuDNN AOT to land is ~1-2 weeks. If it would be alright with you I would like to wait for CuDNN AOT to land before merging this change.

dimitar-asenov:
@derdrdirk Can you please provide an update here. Can the PR now be reviewed/merged?

kasper0406:
@dimitar-asenov my understand is that we are waiting for https://github.com/openxla/xla/pull/41021 to merge to be able to implement this cleanly

derdrdirk:
https://github.com/openxla/xla/pull/41021 has finally landed. If you clean up the PR and give me another ping I could take another look.

kasper0406:
Sounds good @derdrdirk ! It seems like it will be in two steps:
1. Expose the AOT probing to convolutions
2. Do the rewriting from this PR based on the result

I hope to have some PRs ready by the weekend. I’ll let you know.

kasper0406:
Added a PR for device less probing convolutional capabilities here which this PR now builds upon: https://github.com/openxla/xla/pull/45534

akuegel:
@kasper0406 What is the state of this PR? I see that the PR that you mentioned as a prerequisite has been closed already, so is this ready for another review?

kasper0406:
@akuegel @derdrdirk yes, this is good to review. It now uses the just added conv probing to figure out if replacing FP8 convolutions with BF16 makes plans available, and if so swaps to the BF16 version.

derdrdirk (approved):
LGTM. Thanks!

PR #40053, opened 2026-03-29
[xla:ifrt] Allow passing PjRt client for autotuning during cross-compile

ezhulenev:
Cross-compilation can happen on a machine with identical GPUs to the worker nodes, in this case we want to be able to run cross-compilation + autotuning on real hardware to select optimal kernels. Plumb autotuning client via `XlaCompileOptions`, so it can be passed to `PjRtCompile` that already supports cross-compile + autotuning mode.

### DIFF
--- a/xla/python/pjrt_ifrt/xla_compiler.h
+++ b/xla/python/pjrt_ifrt/xla_compiler.h
@@ -56,6 +56,12 @@ struct XlaCompileOptions
   DeviceListRef devices;
   std::vector<tsl::RCReference<LoadedHostCallback>> loaded_host_callbacks;
 
+  // Optional PjRtClient for GPU autotuning during cross-compilation. When
+  // compiling for a virtual topology (e.g. 1024 H100s) via a compile-only
+  // client, this provides access to local GPU device(s) that can be used for
+  // kernel autotuning. Topology must have compatible device type.
+  xla::PjRtClient* autotuning_client = nullptr;
+
   // CompileOptions implementation.
 
   ~XlaCompileOptions() override = default;
--- a/xla/python/pjrt_ifrt/pjrt_compiler.cc
+++ b/xla/python/pjrt_ifrt/pjrt_compiler.cc
@@ -159,16 +159,17 @@ tsl::Future<ExecutableRef> PjRtCompiler::Compile(
   if (pjrt_topology == nullptr) {
     return absl::InvalidArgumentError("PjRtCompiler requires a PjRtTopology");
   }
-  auto compile =
-      [program = std::move(program), xla_program = std::move(xla_program),
-       xla_compile_options = std::move(xla_compile_options), pjrt_topology,
-       user_context = UserContextScope::current()]() mutable {
-        UserContextScope scope(std::move(user_context));
-        return PjRtExecutable::Create(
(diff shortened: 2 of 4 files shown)


beckerhe:
I wouldn't be surprised if this broke something internally, but let's try!

beckerhe (approved):

xla/python/pjrt_ifrt/xla_compiler.h
```diff
@@ -56,6 +56,12 @@ struct XlaCompileOptions
   DeviceListRef devices;
   std::vector<tsl::RCReference<LoadedHostCallback>> loaded_host_callbacks;
 
+  // Optional PjRtClient for GPU autotuning during cross-compilation. When
+  // compiling for a virtual topology (e.g. 1024 H100s) via a compile-only
+  // client, this provides access to local GPU device(s) that can be used for
+  // kernel autotuning. Topology must have compatible device type.
+  xla::PjRtClient* autotuning_client = nullptr;
```
hyeontaek, line 63: I think this is not suitable to be propagated via `xla::ifrt::XlaCompileOptions`. `xla::PjRtClient*` is not serializable in a version compatible manner or reverse-proxyable (possible in theory but PJRT C API would be used in both directions, which is an unusual setup), so this would break the serializability of `xla::ifrt::XlaCompileOptions`.

The optional autotuning client seems more like what `xla::ifrt::PjRtCompiler` could track and use, but it will require moving `xla::PjRtCompile()` call (so that it is `xla::ifrt::PjRtCompiler` who supplies autotuning client to the compilation) and other metadata extraction logic from `PjRt(Loaded)Executable::Create` to the `xla::ifrt::PjRtCompiler` side.

BTW, is there some reason why the autotuning client is not just used as the main PjRt client and do non-AoT-style compilation from the IFRT perspective, to effectively do AoT compilation in the end? i.e., Does using the autotuning client instead of nullptr for the compile-only IFRT client have the same outcome? https://github.com/openxla/xla/blob/3707d0f82eb7e61e000f154437968b255fe2b67a/xla/python/compile_only_ifrt/client.h#L370

ezhulenev: The reason is I think how topology is reported from the client. We have topology with 1k GPUs that we want to compile for, but real client sees only 8 local devices.

hyeontaek: OK, so it is actually the other way around; AoT is being done for 1k GPU, but autotuning is utilizing local GPU devices.

First, I am not sure why the autotuning client is being handed over to `xla::PjRtCompile()` via `client` parameter because I would think `client` as the target client for the compilation, not an auxiliary system resource (`PjRtClient`) used for a part of the compilation (autotuning). I think it should have been a separate parameter or option. For the scope of this PR, let me set this issue aside.

For the IFRT side, it should still be `xla::ifrt::PjRtCompiler` who gets and uses the `autotuning_client` for the compilation. `xla::PjRtClient*` (whether it is an autotuning client or a real client), `xla::ifrt::XlaCompileOptions` needs full serializability or at least proxyability (from the compiler side), and `xla::PjRtClient*` does not meet this bar.

PR #40132, opened 2026-03-30
[xla:gpu] Allow cross-compilation via PjRtCompiler::Compile

ezhulenev:
Allow passing `PjRtClient` to `PjRtCompile` when cross compiling for topology different from the clients topology (the cross compilation use case).

Corresponding JAX change: https://github.com/jax-ml/jax/pull/36350

### DIFF
--- a/xla/python/pjrt_ifrt/pjrt_executable.h
+++ b/xla/python/pjrt_ifrt/pjrt_executable.h
@@ -95,10 +95,13 @@ class PjRtExecutable final
     : public llvm::RTTIExtends<PjRtExecutable, PjRtCompatibleExecutable> {
  public:
   // Creates PjRtExecutable from an MLIR module. Internally, it compiles the
-  // provided MLIR module into an `xla::PjRtExecutable`.
+  // provided MLIR module into an `xla::PjRtExecutable`. When `compile_client`
+  // is non-null, it is passed to `xla::PjRtCompile` to request
+  // cross-compilation.
   static absl::StatusOr<ExecutableRef> Create(
       xla::MaybeOwningMlirModule module, xla::CompileOptions compile_options,
-      const xla::PjRtTopologyDescription& topology);
+      const xla::PjRtTopologyDescription& topology,
+      xla::PjRtClient* compile_client = nullptr);
 
   // PjRtCompatibleExecutable implementation.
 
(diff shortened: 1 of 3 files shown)


xla/python/pjrt_ifrt/pjrt_executable.h
```diff
@@ -95,10 +95,13 @@ class PjRtExecutable final
...
   // Creates PjRtExecutable from an MLIR module. Internally, it compiles the
-  // provided MLIR module into an `xla::PjRtExecutable`.
+  // provided MLIR module into an `xla::PjRtExecutable`. When `client` is
+  // non-null, it is passed to PjRtCompile to enable cross-compilation with
+  // autotuning on real hardware.
   static absl::StatusOr<ExecutableRef> Create(
       xla::MaybeOwningMlirModule module, xla::CompileOptions compile_options,
-      const xla::PjRtTopologyDescription& topology);
+      const xla::PjRtTopologyDescription& topology,
+      xla::PjRtClient* client = nullptr);
```
hyeontaek: Nit: Naming the new parameter to `compile_client` (here and other places adding this parameter) as done in `xla/python/pjrt_ifrt/pjrt_compiler.cc` can signify that it is used only for compilation and is unrelated to any execution or the client that would be used in the future for loading.

xla/python/pjrt_ifrt/pjrt_executable.h
```diff
@@ -95,10 +95,13 @@ class PjRtExecutable final
     : public llvm::RTTIExtends<PjRtExecutable, PjRtCompatibleExecutable> {
  public:
   // Creates PjRtExecutable from an MLIR module. Internally, it compiles the
-  // provided MLIR module into an `xla::PjRtExecutable`.
+  // provided MLIR module into an `xla::PjRtExecutable`. When `client` is
+  // non-null, it is passed to PjRtCompile to enable cross-compilation with
```
hyeontaek: I think autotuning is a bit too much low level detail because it may or may not happen, and it is a PjRt compiler detail.

We can say: `When `client` is non-null, it is passed to `xla::PjRtCompile` to request cross-compilation.`

xla/python/pjrt_ifrt/pjrt_executable.h
ezhulenev: Done

xla/python/pjrt_ifrt/pjrt_executable.h
ezhulenev: Done

hyeontaek (approved):

seantalts (approved):

PR #40136, opened 2026-03-30
Fix XProf program_id metadata mapping

zhenying-liu:
📝 Summary of Changes
  This change fixes XProf operation/layer attribution by preserving HLO
  program_id as the /host:metadata event_metadata key throughout the
  profiling pipeline. It restores integer program_id-based metadata
  insertion in metadata_utils.h, preserves those metadata IDs when merging
  /host:metadata planes in xplane_utils.cc, and adds targeted unit tests
  covering both metadata insertion and metadata-plane merge behavior.

🎯 Justification
  XProf’s Operations view relies on GPU kernel program_id values matching
  /host:metadata event_metadata keys so kernels can be attributed back to
  HLO modules and higher-level layers. Recent changes broke that mapping in
  two places: HLO metadata insertion stopped keying entries by integer
  program_id, and metadata-plane merge could also reassign those IDs. This
  change restores that mapping. The main workload validated for this fix is
  MaxText profiling with XPlane/XProf, including a llama3.1-8b
  reproducer that previously showed missing layer attribution.

🚀 Kind of Contribution

Not applicable. This change is a correctness fix for profiler metadata 
mapping, not a performance optimization.

🧪 Unit Tests:
  
Added:
  - //xla/backends/profiler/cpu:metadata_utils_test: verifies that HLO
    metadata uses the actual program_id as the key. It also checks that
    name and display name are preserved, kProgramId is stored, repeated
    insertion for the same program_id is idempotent, and the scoped
    builder flow does not corrupt program_id keys.
  - //xla/tsl/profiler/utils:xplane_utils_test: adds a targeted test that
    merging /host:metadata preserves program_id as the event_metadata
    key.

🧪 Execution Tests:
  No new checked-in execution test was added in this PR. External end-to-end
  validation was run with a MaxText reproducer on a single machine/GPU setup
  using llama3.1-8b.

### DIFF
--- a/xla/backends/profiler/cpu/BUILD
+++ b/xla/backends/profiler/cpu/BUILD
@@ -127,6 +127,21 @@ cc_library(
 )
 
 xla_cc_test(
+    name = "metadata_utils_test",
+    srcs = ["metadata_utils_test.cc"],
+    deps = [
+        ":metadata_utils",
+        "//xla/service:hlo_proto_cc",
+        "//xla/tsl/platform:test",
+        "//xla/tsl/profiler/convert:xla_op_utils",
+        "//xla/tsl/profiler/utils:xplane_builder",
+        "//xla/tsl/profiler/utils:xplane_schema",
+        "@com_google_googletest//:gtest_main",
+        "@tsl//tsl/profiler/protobuf:xplane_proto_cc",
+    ],
+)
+
+xla_cc_test(
     name = "host_tracer_test",
     srcs = ["host_tracer_test.cc"],
     deps = [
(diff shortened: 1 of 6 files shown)


zhenying-liu:
The current bug:
current-bug

After the fix:
after-fix

zhenying-liu:
The CI ROCm/execute-xla-ut failure seems not related with the PR: Error response from daemon: No such container: xla-runner-23772681715-1.

zhenying-liu:
I verified Xprof 2.22.2 fixed the issue. So we don't need this PR.
[profile-xprof-2 22 2
](url)

PR #40173, opened 2026-03-31
[XLA:GPU] Command buffer supports partial update-free for trace commands. 

shawnwang18:
📝 Summary of Changes

  Replaces the boolean flag xla_gpu_enable_command_buffer_va_remapping with a three-value enum CommandBufferUpdateMode in DebugOptions:

  - FULL_UPDATE (default): VA remapping disabled; all command buffer commands are updated on every execution.
  - FULL_UPDATE_FREE: All buffer allocations are VA-remapped to fixed virtual addresses; the command buffer is recorded once and replayed without any updates.
  - CUSTOM_LIBRARY_UPDATE_FREE: Only traced commands (TracedCommandBufferCmd subclasses such as GEMM, cuDNN) and collective operations (CollectiveCmd subclasses) are VA-remapped. Non-library commands (e.g., memcpy, elementwise kernels) still update normally. This enables update-free replay for the expensive custom-library kernels while retaining correctness for all others.

  Additionally introduces VA range multiplexing (kNumVaReservationSets = 2): two VA reservation sets alternate between executions, allowing the CPU to remap one VA range while the GPU executes on the other, enabling CPU/GPU overlap during remapping.

  The CommandBufferThunk state map key is updated from se::StreamExecutor* to (se::StreamExecutor*, void*) — the second element is the physical address of the first VA-mapped allocation, used to identify which reservation set is active.

  🎯 Justification

  The original boolean flag only covered the all-or-nothing case (FULL_UPDATE_FREE): either every allocation is VA-remapped or none are. This is insufficient for models that mix custom library calls (GEMM, cuDNN, NCCL collectives) with regular XLA-emitted kernels, because:

  1. Custom library commands (cuBLAS, cuDNN, NCCL) use opaque traced command buffers that cannot be updated by XLA after recording — they must be VA-remapped to be replay-safe.
  2. Regular XLA kernels are cheap to update and do not need VA remapping.

  CUSTOM_LIBRARY_UPDATE_FREE covers this common case: it eliminates redundant updates for the expensive library calls while keeping full update semantics for everything else. This reduces per-execution overhead on models dominated by GEMM and collective operations, while can also greatly reduce the cost of cuda VMM API.

Some benchmarks on GB200: 
image

  🚀 Kind of Contribution

  ⚡️ Performance Improvement, ✨ New Feature

  📊 Benchmark

  N/A — the performance benefit is execution-latency reduction on models with repeated identical executions (e.g., training loops). The overhead eliminated is the per-step command buffer update cost for cuBLAS/cuDNN/NCCL commands. Benchmark numbers to be added separately against representative training HLOs.

  🧪 Unit Tests

  No new compiler unit tests — the change is in runtime thunk behavior, not HLO transformations. Existing CommandBufferThunk unit tests cover the update/record path.

  🧪 Execution Tests

  Six new end-to-end execution tests added in xla/pjrt/gpu/se_gpu_pjrt_client_test.cc, all exercising the VA remapping modes with 1 GPU:

  - CommandBufferVaRemappingFusionOps — elementwise fusion with FULL_UPDATE_FREE
  - CommandBufferVaRemappingGemmOps — GEMM with FULL_UPDATE_FREE
  - CommandBufferVaRemappingConditional — conditional op with FULL_UPDATE_FREE
  - CommandBufferVaRemappingWhileLoop — while loop with FULL_UPDATE_FREE
  - CommandBufferVaRemappingDynamicSliceFusion — dynamic slice fusion with FULL_UPDATE_FREE
  - CommandBufferVaRemappingMultiplexing — concurrent execution across two VA reservation sets to verify CPU/GPU overlap correctness

### DIFF
--- a/xla/service/gpu/gpu_executable.cc
+++ b/xla/service/gpu/gpu_executable.cc
@@ -120,6 +120,7 @@ limitations under the License.
 #include "xla/tsl/platform/env.h"
 #include "xla/tsl/platform/env_time.h"
 #include "xla/tsl/platform/logging.h"
+#include "xla/tsl/platform/status_macros.h"
 #include "xla/tsl/util/sorted_range.h"
 #include "xla/util.h"
 #include "xla/util/split_proto/split_executable_and_options_writer.h"
@@ -128,7 +129,6 @@ limitations under the License.
 #include "tsl/platform/random.h"
 #include "tsl/profiler/lib/scoped_annotation.h"
 #include "tsl/profiler/lib/traceme.h"
-#include "xla/tsl/platform/status_macros.h"
 
 namespace xla {
 namespace gpu {
@@ -173,7 +173,7 @@ class GpuExecutableThunkPassBufferAllocator : public ThunkPassBufferAllocator {
       BufferAllocation::Index start_idx)
       : next_idx_(start_idx) {}
 
-  absl::StatusOr<BufferAllocation* absl_nonnull> NewEmptyAllocation(
+  absl::StatusOr<BufferAllocation * absl_nonnull> NewEmptyAllocation(
       int64_t size) override {
     allocations_.push_back(BufferAllocation(next_idx_++, size, /*color=*/0));
     return &allocations_.back();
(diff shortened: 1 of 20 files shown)


xla/service/gpu/gpu_executable.cc
```diff
@@ -361,25 +361,49 @@ GpuExecutable::GpuExecutable(
...
+                           .xla_gpu_command_buffer_update_mode()
+                     : DebugOptions::FULL_UPDATE;
+
+    if (update_mode == DebugOptions::FULL_UPDATE_FREE ||
+        update_mode == DebugOptions::CUSTOM_LIBRARY_UPDATE_FREE) {
+      CHECK_OK(thunk_executor_->thunks().WalkNested(
+          [this, update_mode](const Thunk* t) -> absl::Status {
+            auto* cbt = dynamic_cast<const CommandBufferThunk*>(t);
+            if (cbt == nullptr) return absl::OkStatus();
+            return cbt->WalkCommands([this, update_mode](
```
ezhulenev: capture by ref [&], it's a sync lambda anyway

xla/xla.proto
```diff
@@ -357,6 +357,19 @@ message DebugOptions {
     CONCURRENT_REGIONS = 3;
   }
 
+  // Controls the VA remapping strategy for command buffer thunks.
```
ezhulenev, line 350: FWIW I don't like the names, but don't have better suggestions, leaving it to @thomasjoerg

ezhulenev (approved):
Still don't fully understand all the details, but change looks minimally invasive and reasonable. Passing to @thomasjoerg and the team for final judgement!

xla/xla.proto
shawnwang18: AI suggestions: 

 ```
  Option A — Focus on what's remapped (VA scope)
  NO_VA_REMAPPING          // was FULL_UPDATE
  FULL_VA_REMAPPING        // was FULL_UPDATE_FREE
  PARTIAL_VA_REMAPPING     // was CUSTOM_LIBRARY_UPDATE_FREE
  Clean and symmetric. Describes the feature being controlled (VA remapping scope) rather than the side effect.

  ---
  Option B — Focus on update behavior (command buffer updates)
  ALWAYS_UPDATE            // was FULL_UPDATE
  NEVER_UPDATE             // was FULL_UPDATE_FREE
  SELECTIVE_UPDATE         // was CUSTOM_LIBRARY_UPDATE_FREE
  Very readable but loses the connection to VA remapping.
```

I prefer option B, what do you think?

**Comment by shawnwang18** [comment]
> Still don't fully understand all the details, but change looks minimally invasive and reasonable. Passing to @thomasjoerg and the team for final judgement!

Here is an explain on the implementation detail. 

 NEVER_UPDATE

  All buffer allocations are VA-remapped to fixed addresses before the first execution. The command buffer is recorded once (is_first_record = true when state is kCreate) with these fixed VA addresses baked in. On all subsequent executions, the physical memory behind the VA changes (via VMM), but the recorded GPU commands never need updating — the addresses they reference are always valid.

  The lookup key for GetOrCreateCommandBuffer is (executor, first_alloc[0].opaque()) — the physical address of the first allocation, which identifies which of the two VA reservation sets is currently active.

  ---
  CAPTURE_CMD_NEVER_UPDATE

  A partial version of NEVER_UPDATE. Only traced commands (e.g. cuBLAS via GemmCmd/CublasLtCmd) and collective commands get VA-remapped; all other commands still update normally (needs_update is true for this mode, same as ALWAYS_UPDATE).

  At construction time, it precomputes first_traced_cmd_alloc_idx_ — the minimum buffer allocation index belonging to the first traced command. This cached index is then used as the GetOrCreateCommandBuffer lookup key instead of allocs_indices()[0], because only traced command allocations are VA-remapped (non-traced ones change freely). The command_executor.cc side skips re-recording for traced commands during
  update (they don't need it), while non-traced commands are re-recorded normally.

File: xla/service/gpu/gpu_executable.cc (hunk 1 above)
Comment (shawnwang18) [reply xla/service/gpu/gpu_executable.cc]: Fixed.

File: xla/xla.proto (hunk 2 above)
Comment (shawnwang18) [reply xla/xla.proto:350]: I changed the name to : 
```
  ALWAYS_UPDATE                    // was FULL_UPDATE
  NEVER_UPDATE                     // was FULL_UPDATE_FREE
  CAPTURE_CMD_NEVER_UPDATE         // was CUSTOM_LIBRARY_UPDATE_FREE
```

This looks better?

pifon2a (approved):
Thank you @shawnwang18 ! Very nice!

shawnwang18:
> Thank you @shawnwang18 ! Very nice!

Thanks for the review.

PR #40186, opened 2026-03-31
[XLA:GPU][oneAPI] Update timestamp mask calculation to support 64-bit kernel timestamps

bhavani-subramanian:
Currently, the timestamp mask calculation supports up to 32-bit kernel timestamps. However, newer Intel GPUs (such as [BMG](https://www.intel.com/content/www/us/en/products/sku/241598/intel-arc-b580-graphics/specifications.html)) use 64-bit kernel timestamps, and this PR updates this calculation accordingly.


allanrenucci (approved):

bhavani-subramanian:
@allanrenucci Thanks for approving this PR! Looks like there is an error in the `import/copybara` stage for which I don't have access. Can you please take a look? Thanks.

allanrenucci (approved):

PR #40199, opened 2026-03-31
[ROCm] Fix profiler leaking stale hipErrorInvalidDevice on ROCm

magaonka-amd:
📝 Summary of Changes
- Replace hipGetDeviceProperties in GetDeviceCapabilities with rocprofiler
  agent data already available from RocmTracer. This eliminates HIP runtime
  calls that fail for non-visible devices when ROCR_VISIBLE_DEVICES restricts
  GPU visibility (e.g. pytest-xdist workers).
- Since ROCm, hipGetLastError() is sticky: it retains errors even after
  subsequent successful HIP API calls. The stale hipErrorInvalidDevice from
  the profiler leaked into unrelated GPU operations, causing flaky test
  failures in JAX's FFI handlers (solver, linalg, prng).
- Agent clock rates are in MHz (vs KHz in hipDeviceProp_t); memory is
  filtered to VRAM-only banks (FRAME_BUFFER_PUBLIC/PRIVATE).
- Add unit test verifying agent data matches hipGetDeviceProperties.

🚀 Kind of Contribution

Since this issue was very interesting to debug adding bit more detail for future readers:

### The Problem

Random JAX tests fail intermittently when running the full test suite with pytest-xdist (parallel workers). The failing tests are always in solver/linalg/prng FFI handlers, with the error:

```
jaxlib/gpu/solver_kernels_ffi.cc:266: operation hipGetLastError() failed: invalid device ordinal
```

or

```
jaxlib/gpu/linalg_kernels.cc:113: operation gpuGetLastError() failed: invalid device ordinal
```

The failures are random,the same test passes on the next run, and a different test fails instead. It only happens with xdist (parallel workers on multi-GPU systems) and only on ROCm 7+.

### How It Happens

**Step 1: xdist restricts GPU visibility per worker**

JAX's `conftest.py` assigns one GPU per xdist worker:

```python
# conftest.py line 82-83
os.environ.setdefault(
    "ROCR_VISIBLE_DEVICES", str(xdist_worker_number % num_rocm_devices)
)
```

So worker gw0 sees GPU 0 only, gw1 sees GPU 1 only, etc.

**Step 2: The profiler counts ALL physical GPUs**

XLA's ROCm profiler initializes its device count from `rocprofiler` agent enumeration:

```cpp
// rocm_tracer.cc line 403-409
num_gpus_ = 0;
for (const auto& agent : GetGpuDeviceAgents()) {
    if (agent.type == ROCPROFILER_AGENT_TYPE_GPU) {
        num_gpus_++;  // counts ALL physical GPUs (8), not visible ones (1)
    }
}
```

`rocprofiler` sees all 8 physical GPUs regardless of `ROCR_VISIBLE_DEVICES`.

**Step 3: The profiler calls `hipGetDeviceProperties` with invalid device IDs**

When the profiler collects data, it iterates all `num_gpus_` devices:

```cpp
// rocm_collector.cc line 556-562
for (int id = 0; id < num_gpus_; id++) {  // loops 0..7
    per_device_collector_[id].GetDeviceCapabilities(id, &device_plane);
}
```

Inside `GetDeviceCapabilities`:

```cpp
// rocm_collector.cc line 408-411
if (hipGetDeviceProperties(&device_properties_, device_ordinal) != hipSuccess) {
    return;  // handles error via return code, but...
}
```

For worker gw0 (which only sees GPU 0), devices 1-7 are invisible. `hipGetDeviceProperties` returns `hipErrorInvalidDevice` for each. The error is handled by the return code, but the HIP runtime also sets a per-thread sticky error.

**Step 4: ROCm 7 changed `hipGetLastError()` behavior**

Since ROCm 7, `hipGetLastError()` is per-thread sticky -- it retains the last error even after subsequent successful HIP API calls. Before ROCm 7, successful calls would clear it (like CUDA's `cudaGetLastError()`).

So the `hipErrorInvalidDevice` from step 3 persists in the thread's error state indefinitely.

**Step 5: A JAX FFI handler picks up the stale error**

Later, when a test runs a solver/linalg/prng operation, the FFI handler launches a GPU kernel and checks for errors:

```cpp
// solver_kernels_ffi.cc line 158
MakeBatchPointersAsync(stream, ...);
JAX_FFI_RETURN_IF_GPU_ERROR(gpuGetLastError());  // picks up stale hipErrorInvalidDevice!
```

The `gpuGetLastError()` call returns the stale `hipErrorInvalidDevice` from the profiler's earlier failed `hipGetDeviceProperties` call -- not from the kernel launch above it.

### Why It Is Rare and Flaky

1. **Timing-dependent**: The profiler's `CollectData` doesn't run after every test. It fires at internal profiler flush points. The stale error only affects whichever test happens to run next on that thread after the flush.

2. **Test ordering matters**: With xdist shuffling tests across 16 workers, the probability that a solver/linalg test runs immediately after a profiler flush on the same worker thread is low but nonzero.

3. **ROCm 7+ only**: On ROCm 6 and earlier, `hipGetLastError()` wasn't sticky across successful calls, so the stale error would be cleared by the next successful HIP API call before any FFI handler could see it.

### The Fix
Populate device property using rocprof-agent instead of hipcall.

### Evidence

- **Debug trap in HIP runtime** (`HIP_RETURN` macro) confirmed the source: In 8 GPU setup :
  ```
  [HIP TRAP] hipErrorInvalidDevice set by hipGetDevicePropertiesR0600
  ```
  18 entries per test run (7 invisible devices × ~2-3 profiler flushes per worker).

- **Symbolized backtrace** confirmed the call chain:
  ```bash
  call 0  hipGetDevicePropertiesR0600 (hip_device.cpp:661)
  call 1  PerDeviceCollector::GetDeviceCapabilities (rocm_collector.cc)
  call 2  RocmTraceCollectorImpl::Export (rocm_collector.cc)
  call 3  GpuTracer::CollectData (device_tracer_rocm.cc)
  call 4  ProfilerSession::CollectData (profiler_session.cc)
  and so on...
  ```
- **`--xla_gpu_autotune_level=0`** (disables autotuning, reduces profiler activity) eliminated flakiness -- consistent with profiler being the trigger.

### DIFF
--- a/xla/backends/profiler/gpu/rocm_collector.cc
+++ b/xla/backends/profiler/gpu/rocm_collector.cc
@@ -163,7 +163,13 @@ OccupancyStats PerDeviceCollector::GetOccupancy(
   }
 
   stats.occupancy_pct = number_of_active_blocks * params.block_size * 100;
-  stats.occupancy_pct /= device_properties_.maxThreadsPerMultiProcessor;
+  // TODO(ROCm): GetOccupancy is currently dead code -- no callers populate
+  // max_waves_per_cu / wave_front_size, so occupancy_pct stays zero.
+  // Wire up agent data when occupancy reporting is enabled.
+  auto max_threads = params.max_waves_per_cu * params.wave_front_size;
+  if (max_threads > 0) {
+    stats.occupancy_pct /= max_threads;
+  }
 
   err = hipOccupancyMaxPotentialBlockSize(
       &stats.min_grid_size, &stats.suggested_block_size,
@@ -404,18 +410,15 @@ void PerDeviceCollector::AddEvent(RocmTracerEvent&& event) {
   events_.emplace_back(std::move(event));
 }
 
-void PerDeviceCollector::GetDeviceCapabilities(int32_t device_ordinal,
-                                               XPlaneBuilder* device_plane) {
+void PerDeviceCollector::GetDeviceCapabilities(
+    const rocprofiler_agent_v0_t& agent, XPlaneBuilder* device_plane) {
   device_plane->AddStatValue(*device_plane->GetOrCreateStatMetadata(
                                  GetStatTypeStr(StatType::kDevVendor)),
                              kDeviceVendorAMD);
 
-  if (hipGetDeviceProperties(&device_properties_, device_ordinal) !=
-      hipSuccess) {
-    return;
-  }
-
-  auto clock_rate_in_khz = device_properties_.clockRate;  // this is also in Khz
+  // Agent clock rates are in MHz; profiler stats expect KHz.
+  auto clock_rate_in_khz =
+      static_cast<int64_t>(agent.max_engine_clk_fcompute) * 1000;
   if (clock_rate_in_khz) {
     device_plane->AddStatValue(
         *device_plane->GetOrCreateStatMetadata(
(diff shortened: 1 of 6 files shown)


i-chaochen (changes requested):
could you add a UT for this profiler leaking check in xla/backends/profiler/gpu/rocm_collector_test.cc

magaonka-amd:
added regression test that ensures leaking hipErrorInvalidDevice into per-thread state won't happen

magaonka-amd:
@i-chaochen  I changed the approach a little bit here and changed test accordingly.

Could you please review at your convinience.

xla/backends/profiler/gpu/rocm_collector.cc
```diff
@@ -404,66 +410,85 @@ void PerDeviceCollector::AddEvent(RocmTracerEvent&& event) {
...
+  // Extract memory info from VRAM (frame buffer) banks only, skipping
+  // system memory, LDS, GDS, scratch, etc.
+  if (agent.mem_banks_count > 0 && agent.mem_banks != nullptr) {
+    uint64_t total_memory = 0;
+    uint32_t vram_clock_mhz = 0;
+    uint32_t vram_bus_width_bits = 0;
+    for (uint32_t i = 0; i < agent.mem_banks_count; ++i) {
+      if (agent.mem_banks[i].heap_type == HSA_HEAPTYPE_FRAME_BUFFER_PUBLIC ||
+          agent.mem_banks[i].heap_type == HSA_HEAPTYPE_FRAME_BUFFER_PRIVATE) {
+        total_memory += agent.mem_banks[i].size_in_bytes;
```
draganmladjenovic, line 447: I think the last one just wins and hip runtime uses that one. They are probably mutually exclusive so the numbers add up.  Please check. And maybe put some todo for APUs. Not sure if anything is different there.

magaonka-amd: They are mutually exclusive. 

Hmm, on APU frame buffer could be different , added TODO.

magaonka-amd:
Hello @dimitar-asenov , Sorry for pinging explicitly, Can I please get feedback on this fix. 
This fix supposed to help with random test crash/fails in JAX CI and having this will greatly help us in stabilizing JAX CI runs.

dimitar-asenov (approved):

PR #40232, opened 2026-04-01
Add missing kAllGatherStart/kAllGatherDone verification to HLO verifier

kredd2506:
## Summary
Fixes #40191

`VerifyAsynchronousInstructionPairs` checks that async Start/Done instructions
are properly connected for all async op types except `kAllGatherStart/kAllGatherDone`.
This adds the missing verification, following the same pattern used for `kAllReduceStart/kAllReduceDone`.

## Changes
- `hlo_verifier.cc`: Add `kAllGatherStart` and `kAllGatherDone` cases to the async pair verification switch
- `hlo_verifier_test.cc`: Add tests for valid AllGather Start/Done pair and invalid multiple-Done case

## Test plan
- [ ] CI verification
- [x] Tests follow existing AllReduce verifier test patterns
- [x] Follow Google C++ style

### DIFF
--- a/xla/service/hlo_verifier_test.cc
+++ b/xla/service/hlo_verifier_test.cc
@@ -1856,6 +1856,45 @@ TEST_F(HloVerifierTest, AllReduceDoneWithoutStart) {
                         "needs to be all-reduce-start, found tuple"));
 }
 
+TEST_F(HloVerifierTest, AllGatherStartAndDone) {
+  const char* const kModuleStr = R"(
+  HloModule test
+  ENTRY entry {
+    p0 = f32[2,3] parameter(0)
+    start = (f32[2,3], f32[4,3]) all-gather-start(p0), dimensions={0},
+      replica_groups={{0,1}}
+    ROOT done = f32[4,3] all-gather-done(start)
+  }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnUnverifiedModule(kModuleStr));
+
+  auto status = verifier().Run(module.get()).status();
+  ASSERT_TRUE(status.ok());
+}
+
+TEST_F(HloVerifierTest, AllGatherStartAndMultipleDone) {
+  const char* const kModuleStr = R"(
+  HloModule test
+  ENTRY entry {
+    p0 = f32[2,3] parameter(0)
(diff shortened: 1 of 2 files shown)


xla/service/hlo_verifier_test.cc
```diff
@@ -1856,6 +1856,45 @@ TEST_F(HloVerifierTest, AllReduceDoneWithoutStart) {
...
+  HloModule test
+  ENTRY entry {
+    p0 = f32[2,3] parameter(0)
+    start = (f32[2,3], f32[4,3]) all-gather-start(p0), dimensions={0},
+      replica_groups={{0,1}}
+    ROOT done = f32[4,3] all-gather-done(start)
+  }
+  )";
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                          ParseAndReturnUnverifiedModule(kModuleStr));
```
akuegel: I think here you can still use ParseAndReturnVerifiedModule?

kredd2506:
Good point, updated to use `ParseAndReturnVerifiedModule` for the valid case. I followed the existing `AllReduceStartAndDone` test which uses `ParseAndReturnUnverifiedModule`, but the verified version is cleaner.

akuegel (approved):
Thank you :)

akuegel:
Just FYI, your new check exposed a test that had invalid HLO. I am fixing that and then your PR will get merged.

kredd2506:
Thanks for fixing that! When I first saw the failing test I was reviewing my own logic thinking I'd broken something — but your fix makes sense. The verifier was assuming shapes were arrays/tuples without actually checking, and the nested tuple structure for `AllGatherStart` adds extra complexity.

Nice catch extending the same fix to ReduceScatter too. Glad the async pair check surfaced these.

Quick question — were these crashes exposed by an existing test that had invalid HLO, or was it previously valid HLO that just wasn't going through this verification path before?

akuegel:
> Quick question — were these crashes exposed by an existing test that had invalid HLO, or was it previously valid HLO that just wasn't going through this verification path before?

The crashes were exposed when I tried to adjust support_test.cc but got it wrong initially (creating invalid HLO that crashed instead of being rejected as invalid). After I fixed it for AllGather/AllGatherStart, I asked the KI to check whether there are more such bugs in this file, and it found ReduceScatter and fixed it for me. Only the test it created did not immediately have valid HLO, so I had to adapt it manually.

PR #40236, opened 2026-04-01
[Rocm]  Support VMM allocator to ROCm/HIP

phambinhfin:
## 📝 Summary of Change
Support  VMM (Virtual Memory Management) allocator to ROCm/HIP, enabling stable virtual address allocations on AMD GPUs. This adds 12 new files and modifies 7 existing files across `xla/stream_executor/rocm/` and `xla/pjrt/gpu/`.

Four layers matching the upstream structure:
- **Layer 1** — `RocmRawMemoryAllocation`: RAII wrapper for `hipMemCreate`/`hipMemRelease`
- **Layer 2** — `RocmMemoryReservation`: RAII wrapper for `hipMemAddressReserve`/`hipMemMap`/`hipMemSetAccess`/`hipMemUnmap`
- **Layer 3** — `RocmVmmAllocator`: All-in-one allocator (create + reserve + map + setAccess)
- **Layer 4** — `RocmDeviceAddressVmmAllocator`: GPU timeline-based deferred deallocation via `hipStreamWriteValue64` + coherent host memory

PJRT integration wires `RocmDeviceAddressVmmAllocator` into the GPU client so `XLA_PYTHON_CLIENT_ALLOCATOR=vmm` works on ROCm (previously returned "VMM allocator is only supported with CUDA").

## 🚀 Kind of Contribution

✨ New Feature, 🧪 Tests

Measured on **8× AMD Instinct MI300X (gfx942), ROCm 7.2.0**, jax `0.10.1.dev20260519`, `jax-rocm7-plugin`. The VMM `NEVER_UPDATE` fixed-VA replay path makes command-buffered collectives cheap by remapping only the physical buffers that change each step (stable `allocation_id` + deterministic address-order reuse), instead of re-tracing the command buffer every step.

### Popular models — MaxText, FSDP=8, 8 decoder layers, 30 steps — steady per-step (ms)

| model | base (no VMM) | +COLL (no VMM) | VMM (no COLL) | VMM +COLL | +COLL speedup |
|---|--:|--:|--:|--:|--:|
| Llama2-7B | 244 | 4543 | 157 | **160** | **~29×** |
| Llama3-8B | 154 | 3830 | 180 | **178** | **~21×** |
| Mixtral-8x1B (MoE) | 680 | 5573 | 710 | **701** | **~8×** |
| Llama3.1-8B | 776 | 754 | 722 | **720** | ~1.05× |
| Gemma2-9B | 1450 | 1451 | 1336 | **1337** | ~1.09× |

`base` = command buffer `FUSION,CUBLAS,CUBLASLT,CUSTOM_CALL`; `+COLL` adds `COLLECTIVES`; `VMM` = `NEVER_UPDATE`. All 20 legs pass the unmap-error gate; no model regresses.

### Synthetic microbenchmark — pmap over 8 GPUs — median per-step

| workload | BFC | VMM/ALWAYS | VMM/NEVER | VMM/CAPTURE_CMD_NEVER |
|---|--:|--:|--:|--:|
| allreduce_only | 1304 μs | 1391 μs | 1355 μs | 1386 μs |
| matmul+allreduce | 1329 μs | 1426 μs | 1390 μs | 1387 μs |
| mlp+2x_allreduce | 1638 μs | 1644 μs | 1499 μs | 1527 μs |
| heavy_matmul+ar | 185788 μs | 182451 μs | **7800 μs** | **5140 μs** |

`VMM/ALWAYS ≈ BFC` confirms the speedup comes from VA-remap replay, not the allocator itself. Small workloads sit at parity because the 8-GPU RCCL clique init dominates when there is almost no graph to replay.

### Follow-up optimization on top of this branch — selective copy-vs-remap (auto threshold)

A follow-up (**ROCm/xla #926**, based on this PR's branch `phambinh/rocm-vmm-allocator`, so it lands here once merged) shaves further per-step cost: the tiny scalar/scale/metric slices that churn their physical address every step are kept in a **stable shadow buffer** refreshed with a cheap device-to-device copy, instead of paying a `hipMemUnmap`/`hipMemMap`/`hipMemSetAccess` round-trip. The cutoff is auto-derived from HBM bandwidth + capacity and currently-free memory — `XLA_VMM_TMP_COPY_THRESHOLD`: `0`=off, `1`=auto, `>1`=fixed. ROCm-only (gated on `DeviceAddressVmmAllocator`); NVIDIA unaffected.

Steady per-step, VMM+COLL leg, disabled vs auto (MI300X×8, 8 layers, 30 steps, median; this A/B uses `scan_layers=true`, so the disabled baseline differs slightly from the table above — the delta is the point):

| model | VMM+COLL (disabled) | + auto copy | saved | speedup |
|---|--:|--:|--:|--:|
| Llama2-7B | 133.5 ms | **121.0 ms** | 12.5 ms | **9.4%** |
| Llama3-8B | 168.0 ms | **156.5 ms** | 11.5 ms | **6.8%** |
| Mixtral-8x1B | 690.5 ms | **677.5 ms** | 13.0 ms | **1.9%** |
| Llama3.1-8B | 713.5 ms | **704.5 ms** | 9.0 ms | **1.3%** |
| Gemma2-9B | 1319.5 ms | **1313.0 ms** | 6.5 ms | **0.5%** |

The saving is a roughly fixed per-step cost (~6–13 ms; it is the remap of the small churning slices, whose count is set by command-buffer structure, not model size), with identical loss and zero unmap errors on every run.

## 🧪 Unit Tests

4 test targets (26 tests total) mirroring the CUDA VMM test structure:

- `rocm_raw_memory_allocation_test` (3 tests) — `CreateAllocation`, `AddressReflectsHandle`, `SizeIsAtLeastRequested`
- `rocm_memory_reservation_test` (6 tests) — `CreateReservation`, `MapToWrongType`, `MapToSingleAllocation`, `ScopedMappingUnmapsOnDestruction`, `MapToMultipleAllocations`, `TwoReservationsDifferentAddresses`
- `rocm_vmm_allocator_test` (3 tests) — `AllocateAndFree`, `AllocateZeroBytes`, `MemcpyRoundTrip`
- `rocm_device_address_vmm_allocator_test` (14 tests) — Covers allocate/deallocate, memory read/write, stream accessors, deferred deallocation, VA reuse, destructor safety, and error handling

All tests use `GTEST_SKIP()` when a ROCm executor is not available, allowing graceful skip in sandboxed CI environments.

## 🧪 Execution Tests

- `VmmAllocatorCanBeSet` — Verifies that the PJRT client correctly instantiates `DeviceAddressVmmAllocator` when `GpuAllocatorConfig::Kind::kVmm` is set (shared test for CUDA and ROCm)
- `VmmAllocatorE2ETest` — End-to-end test: creates a VMM-backed PJRT client, compiles and executes an HLO add program, asserts correctness of the result
- 2 multi-GPU tests in `rocm_device_address_vmm_allocator_test` — Verify cross-device allocation isolation (requires 2 GPUs)

All execution tests run with at most 2 GPUs.


i-chaochen (commented):
~

xla/pjrt/gpu/tfrt/utils.cc
```diff
@@ -115,7 +115,9 @@ limitations under the License.
 
 #if GOOGLE_CUDA
 #include "xla/stream_executor/cuda/cuda_device_address_vmm_allocator.h"
-#endif  // GOOGLE_CUDA
+#elif TENSORFLOW_USE_ROCM
```
PatriosTheGreat: Could you rebase? This code was removed

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1649,10 +1650,20 @@ GetStreamExecutorGpuDeviceAllocator(
...
           platform, allocator_config.memory_fraction,
           allocator_config.gpu_system_memory_size, executor_streams);
+#elif TENSORFLOW_USE_ROCM
+      std::vector<std::pair<se::StreamExecutor*, se::Stream*>> executor_streams;
+      executor_streams.reserve(addressable_devices.size());
+      for (const auto& [ordinal, device] : addressable_devices) {
+        executor_streams.push_back(
+            {device->executor(), device->compute_stream()});
+      }
+      return se::gpu::RocmDeviceAddressVmmAllocator::Create(
```
PatriosTheGreat: This code is basically the same as in #if GOOGLE_CUDA branch with the difference that RocmDeviceAddressVmmAllocator::Create called instead of CudaDeviceAddressVmmAllocator.

Could you add Create method to DeviceAddressVmmAllocator which will call one or another depending on the platform and call that method here?

PatriosTheGreat (changes requested):

xla/pjrt/gpu/tfrt/utils.cc
phambinhfin: Done! thanks

xla/pjrt/gpu/se_gpu_pjrt_client.cc
phambinhfin: Sure thanks, @PatriosTheGreat  could you review again, i just measure performance and test , it works great

xla/service/gpu/gpu_executable.cc
```diff
@@ -1535,8 +1536,17 @@ absl::Status GpuExecutable::ExecuteThunksWithVaRemapping(
     // allocation up to the allocation granularity.
     uint64_t total_va_size = 0;
     for (BufferAllocation::Index i : command_buffer_allocation_indexes_) {
-      const uint64_t size = buffer_allocations.GetDeviceAddress(i).size();
-      total_va_size += round_up_to_granularity(size);
+      se::DeviceAddressBase buf = buffer_allocations.GetDeviceAddress(i);
```
PatriosTheGreat, line 1539: Everything related to last_mapping_state and changes in xla/service/gpu/gpu_executable can be extracted to separate PR which should be submitted first since it touches infrastructure shared between ROCM and CUDA, so it can be covered with already existing tests.

xla/service/gpu/gpu_executable.cc
```diff
@@ -1589,72 +1597,152 @@ absl::Status GpuExecutable::ExecuteThunksWithVaRemapping(
...
-    // Batch-map all command buffer allocations into the reserved VA range in
-    // a single call. This maps the contiguous range formed by the descriptors
-    // and enables device access before returning.
+    // Apply the mapping. The first successful execution uses MapTo; subsequent
+    // executions transfer the existing ScopedMapping through Remap so only
+    // changed slices incur driver round-trips (hipMemUnmap/hipMemMap/
+    // hipMemSetAccess). Synchronize the unmap event before Remap so the GPU is
+    // done using the prior mapping before we touch it.
     if (!mapping_descriptors.empty()) {
-      ASSIGN_OR_RETURN(se::MemoryReservation::ScopedMapping scoped_mapping,
```
PatriosTheGreat, line 1653: This remap optimization also deserves a dedicated PR

PatriosTheGreat (changes requested):
This PR seems to contain at least 2 different unrelated changes which should be reviewed and onboarded separately:

1. Adding ROCM version of the VMM device address allocator. This should contain all changes under stream_executor/rocm and the vmm_device_address_allocator_factory.
2. Adding remap functionality to memory_reservation and it's usage in gpu_executable. This also should contain all changes in vmm_device_address_allocator.

xla/service/gpu/gpu_executable.cc
phambinhfin: Thanks, i will make another PR for it

xla/service/gpu/gpu_executable.cc
phambinhfin: Thanks, i will make another PR for it

phambinhfin:
@PatriosTheGreat  @i-chaochen  Thanks for the review — agreed, splitting this up makes sense. Here's how I'll break
it down along the lines you described:
**1. Shared command-buffer remap infra (submitted first).**
I'll extract the `last_mapping_state` / VA-remap changes in `xla/service/gpu/gpu_executable`
together with `MemoryReservation::Remap` into a standalone PR. As you noted, this is
infrastructure shared between CUDA and ROCm and is exercised by the existing
command-buffer tests, so it can land independently of the ROCm-specific work.

**2. ROCm VMM device-address allocator** (everything under `stream_executor/rocm` +
`vmm_device_address_allocator_factory`). To keep each piece small and reviewable, this is
itself a short stack of 4 PRs, bottom-up:
- PR 1/4 — `RocmRawMemoryAllocation` (`hipMemCreate`/`hipMemRelease`): #43942
- PR 2/4 — `RocmMemoryReservation` (reserve/map/setAccess/unmap): #43943
- PR 3/4 — `RocmVmmAllocator` (all-in-one allocate): #43946
- PR 4/4 — `RocmDeviceAddressVmmAllocator` + vendor-neutral factory + PJRT enablement: #43947

The `DeviceAddressVmmAllocator::Create()` platform-dispatch factory you asked for is
already in place and used by the PJRT client (no `#if GOOGLE_CUDA / #elif ROCM` branch in
`se_gpu_pjrt_client.cc`).

**3. Remap optimization (dedicated PR).**
The per-slice remap-skip + deterministic VMM reuse (`allocation_id` / `GetAllocationInfo`,
smallest-address `TryReusePendingDeallocation`, `ScopedMapping::Remap` skip-unchanged) will
be its own PR stacked on top of (1), so the base remap and the optimization are reviewed
separately.

I'll close #40236 in favor of these focused PRs and link them here so the history is easy
to follow. Thanks again for the guidance.

PR #40310, opened 2026-04-02
[xla:gpu] Fix UndefinedBehaviorSanitizer for thunk_executor test

ezhulenev:
Don't pass default constructed params marked as absl_nonnull, we don't need prepare state anyway in this test.


thomasjoerg (approved):
Wow, that was quick. Thank you!

PR #18838, opened 2024-10-29
[NVIDIA GPU] Support multi-operand collective-permute

terryysun:
For collective-permutes with small message sizes, it is beneficial to combine them into a single collective because
1. it gets rid of some kernel launch overhead, and allows NCCL to do some message fusion;
2. fewer collectives make it easier for LHS to make better decision.

In order to support combining collective-permutes, we need to support multi-operand collective-permute first, a.k.a. the combined collective-permute. This PR extends the existing CP interface by overloading it, so that a CP can have multiple operands.

PR #40316, opened 2026-04-02
[Nvidia GPU] Do aliasing only when collective permute is in-place

Tixxx:
📝 Summary of Changes
alias analysis currently treats cp as in-place when the number of operands is 4 which is not correct as we can have combined cp that have 4 inputs but are not in-place.
This pr only considers cp as in place when the inplace property is true.

### DIFF
--- a/xla/hlo/analysis/alias_info.cc
+++ b/xla/hlo/analysis/alias_info.cc
@@ -185,16 +185,21 @@ AliasInfo::GetInPlaceInputOutputPairs(const HloInstruction* user) const {
   }
   if (user->opcode() == HloOpcode::kCollectivePermuteStart &&
       user->operands().size() == 4) {
-    if (user->operand(1)->shape().IsTuple()) {
-      std::vector<std::pair<HloOperandIndex, ShapeIndex>> in_place_pairs(
-          {{HloOperandIndex{1, {}}, {1}}});
-      for (int i = 0; i < user->operand(1)->shape().tuple_shapes().size();
-           i++) {
-        in_place_pairs.push_back({HloOperandIndex{1, {i}}, {1, i}});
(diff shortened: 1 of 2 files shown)


akuegel:
Thanks for this change. I was not familiar with the inplace property and now checked that it was added in https://github.com/openxla/xla/pull/18838. However it looks to me that it might be buggy. E.g. the default for CollectivePermute with 1 operand in XlaBuilder is set to false, and I believe there is no caller of this function that sets it to true. Then, there is code in hlo_dataflow_analysis.cc that checks for ```!Cast<HloCollectivePermuteInstruction>(collective_permute_start)
           ->inplace()``` and then does the same thing as before where we assumed it is in-place. So I fear this inplace flag needs to be cleaned up first before this change can land. Also, the inplace flag is not added to HLO text (so when dumping, we don't see it, but also the HloParser cannot create a CollectivePermute that is inplace).
Can you or someone else at Nvidia please clean this up?

akuegel:
Also, how can we make sure that the inplace attribute is set correctly when creating the XlaOp? How can we distinguish whether in-place will work or not? It seems somewhat strange that this decision is moved outside of the compiler to the builder of the XLA op. I am not familiar with CollectivePermute op, so maybe I am missing something.

Tixxx:
Sync'd offline, we will change how the inplace is detected. waiting for https://github.com/openxla/xla/pull/40601 to reland

akuegel:
> Sync'd offline, we will change how the inplace is detected. waiting for #40601 to reland

https://github.com/openxla/xla/commit/988dddb78ed11d2a0e6dfe8494c0b7fa1e11f058 has relanded, so I think this is good to go. Will approve now.

akuegel (approved):

PR #40320, opened 2026-04-02
[XLA:CPU] Expand the scope of eligible transposes for folding

akhilgoe:
This PR brings the `gemv_rewriter` pass into the CPU pipeline (it was previously GPU-only), which increases the set of transpose patterns that can be folded during CPU compilation.

What this PR changes:

1. Moves `gemv_rewriter` pass and and its tests from `xla/backends/gpu/transforms` to `xla/hlo/transforms/simplifiers` so the pass can be shared across backends.
2. Adds an `is_layout_sensitive` flag to the `gemv_rewrier` pass to control behavior based on whether it runs before or after layout assignment.
3. Expands test coverage to include cases where `is_layout_sensitive` is false.
4. Integrates `gemv_rewriter` to the CPU pipeline.

Enabling this rewrite on CPU improves optimization opportunities and can significantly improve end-to-end performance. Eg: `xla/backends/cpu/benchmarks/e2e/gemma2` shows more than 2x speedup on Intel Xeons (for F32 and F16) and close to 4x speedup for BF16.


xla/hlo/transforms/simplifiers/gemv_rewriter.cc
```diff
@@ -88,6 +90,15 @@ class GemvRewriterVisitor : public DfsHloRewriteVisitor {
       return absl::OkStatus();
     }
 
+    // Skip if the layout is not normalized and the pass is not layout
+    // sensitive.
+    if (!is_layout_sensitive_ &&
+        (!LayoutUtil::IsMonotonicWithDim0Major(lhs->shape().layout()) ||
+         !LayoutUtil::IsMonotonicWithDim0Major(rhs->shape().layout()) ||
+         !LayoutUtil::IsMonotonicWithDim0Major(dot->shape().layout()))) {
```
penpornk: Nit: Prefer
```suggestion
        !(LayoutUtil::IsMonotonicWithDim0Major(lhs->shape().layout()) &&
         LayoutUtil::IsMonotonicWithDim0Major(rhs->shape().layout()) &&
         LayoutUtil::IsMonotonicWithDim0Major(dot->shape().layout()))) {
```

xla/hlo/transforms/simplifiers/gemv_rewriter_test.cc
penpornk, line 1: This file should be marked as branched from the existing gemv_rewriter_test.cc

xla/hlo/transforms/simplifiers/gemv_rewriter.h
```diff
@@ -22,23 +23,30 @@ limitations under the License.
...
 
  protected:
   absl::StatusOr<bool> RunImpl(
       HloModule* module,
       const absl::flat_hash_set<absl::string_view>& execution_threads) override;
+
+ private:
+  // Whether the pass should be layout sensitive. This indicates the position of
+  // the pass with respect to the layout assignment and normalization passes in
+  // pipelines.
```
penpornk: The "This indicates ..." part is hard to understand for me. It's probably better to just explain the relation ship of the bool with other passes. Give examples, e.g., that if the pass comes after layout assignment, then `is_layout_sensitive_` should be set, etc.

penpornk (changes requested):

Feedback for future PRs:
PRs like this can be split. In this case it could be:
1. Move the pass to hlo/transforms/simplifiers without changing anything
2. Add the layout bool and make CPU call the pass.

But for this PR we can proceed like this. If things break and we need to rollback, then we'll split the PR.

dimitar-asenov:
@akhilgoe There has been no activity on this PR for a while. If you still intend on merging it, could you please rebase and address the review comments? Thanks.

xla/hlo/transforms/simplifiers/gemv_rewriter.h
akhilgoe: Done

xla/hlo/transforms/simplifiers/gemv_rewriter_test.cc
akhilgoe: I believe the diff is larger than the github's threshold to identify a relocation. Let me look into this a bit more. We might loose the tracking history of the original file if github thinks that this is a new file.

akhilgoe:
Apologies for the delay @penpornk @dimitar-asenov! I have added a commit to address the review comments. I am still looking if it is possible to make github know that `gemv_rewriter_test` has moved to the new location instead of it thinking that it is a new file.

xla/hlo/transforms/simplifiers/gemv_rewriter.cc
akhilgoe: Done

akhilgoe:
@penpornk @dimitar-asenov I split this PR into 2. First part (moving `gemv_rewriter` to the new location) is merged. I am linking the second one below. Please review this instead. I'll close this PR once you start reviewing the new one.

https://github.com/openxla/xla/pull/43045

dimitar-asenov:
> @penpornk @dimitar-asenov I split this PR into 2. First part (moving `gemv_rewriter` to the new location) is merged. I am linking the second one below. Please review this instead. I'll close this PR once you start reviewing the new one.
> 
> #43045

Both of the PRs that this one was split into are now merged, so I'll close this one.

PR #40334, opened 2026-04-02
Add descriptor destructor callback to CopyToRemoteDevice C API.

mcuiaws:
The CopyToRemoteDevice extension was directly deleting descriptor pointers allocated by the client across the C API boundary, violating memory ownership semantics. This could cause undefined behavior when the client and extension use different allocators.

Add a PJRT_Transfers_DescriptorDestructor callback to the CopyToRemoteDevice args struct so the client provides its own cleanup function. Bump the cross-host transfers extension version to 6.

📝 Summary of Changes
- Add PJRT_Transfers_DescriptorDestructor callback type to the cross-host transfers extension header.
- Add descriptor_destructor field to PJRT_Transfers_PJRT_Buffer_CopyToRemoteDevice_Args.
- Client (pjrt_c_api_client.cc) now provides a destructor callback that frees the descriptor pointers it allocated.
- Extension implementation checks struct_size before accessing descriptor_destructor, falling back to delete for backwards compatibility with version 5 clients.
- Fix PJRT_DEFINE_STRUCT_TRAITS to correctly use descriptor_destructor as the last field.
- Bump PJRT_API_CROSS_HOST_TRANSFERS_EXTENSION_VERSION from 5 to 6.

🎯 Justification
The previous implementation had the extension side directly delete-ing pointers allocated by the client. Across a C API boundary, the client and extension may use different allocators (e.g., different shared libraries with separate heaps), making this undefined behavior. The 
destructor callback ensures each side manages its own memory.

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
No new unit tests added. The existing pjrt_c_api_cross_host_transfers_extension_gpu_test covers the CopyToRemoteDevice path. The change is a mechanical fix to memory ownership — the observable behavior is unchanged when client and extension share the same allocator.

### DIFF
--- a/xla/pjrt/c_api_client/pjrt_c_api_client.cc
+++ b/xla/pjrt/c_api_client/pjrt_c_api_client.cc
@@ -3974,6 +3974,13 @@ void PjRtCApiBuffer::CopyToRemoteDevice(
       pjrt::CppCrossHostRemoteSendCallbackToC(pjrt_c_api(), std::move(on_done));
   args.on_done = on_done_info;
 
+  // Destructor callback for freeing descriptor data
+  auto descriptor_destructor = [](char** descriptor_data, size_t* descriptor_size) {
+    delete descriptor_data;
+    delete descriptor_size;
+  };
+  args.descriptor_destructor = descriptor_destructor;
+
 #if PJRT_API_CROSS_HOST_TRANSFERS_EXTENSION_VERSION < 5
   absl::StatusOr<std::string> descriptor = serialized_descriptor.Await();
   CHECK_OK(descriptor) << "Failed to copy buffer to remote device: "
--- a/xla/pjrt/c_api_client/pjrt_c_api_client.cc
+++ b/xla/pjrt/c_api_client/pjrt_c_api_client.cc
@@ -3974,12 +3974,14 @@ void PjRtCApiBuffer::CopyToRemoteDevice(
       pjrt::CppCrossHostRemoteSendCallbackToC(pjrt_c_api(), std::move(on_done));
   args.on_done = on_done_info;
 
+#if PJRT_API_CROSS_HOST_TRANSFERS_EXTENSION_VERSION >= 6
   // Destructor callback for freeing descriptor data
-  auto descriptor_destructor = [](char** descriptor_data, size_t* descriptor_size) {
+  args.descriptor_destructor = [](char** descriptor_data,
+                                  size_t* descriptor_size) {
     delete descriptor_data;
     delete descriptor_size;
   };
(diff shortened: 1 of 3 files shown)


xla/pjrt/c_api_client/pjrt_c_api_client.cc
```diff
@@ -3974,6 +3974,13 @@ void PjRtCApiBuffer::CopyToRemoteDevice(
       pjrt::CppCrossHostRemoteSendCallbackToC(pjrt_c_api(), std::move(on_done));
   args.on_done = on_done_info;
 
+  // Destructor callback for freeing descriptor data
```
emilyfertig, line 3978: This block should be guarded by `#if PJRT_API_CROSS_HOST_TRANSFERS_EXTENSION_VERSION >= 6` because of the new `descriptor_destructor` field.

xla/pjrt/extensions/cross_host_transfers/pjrt_c_api_cross_host_transfers_extension.cc
```diff
@@ -329,24 +329,34 @@ void PJRT_Transfers_PJRT_Buffer_CopyToRemoteDevice(
   xla::Future<std::string> future(std::move(serialized_descriptor));
 #else
   auto [promise, future] = xla::MakePromise<std::string>();
+  // Use client-provided destructor if available (version 6+), otherwise fall
+  // back to delete for backwards compatibility with version 5 clients.
+  PJRT_Transfers_DescriptorDestructor destructor =
```
emilyfertig: Could you put this behind a version guard too, like this?

```
#if PJRT_API_CROSS_HOST_TRANSFERS_EXTENSION_VERSION < 6
auto destructor = [](char** d, size_t* s) {delete d; delete s; };
#else
PJRT_Transfers_DescriptorDestructor destructor = args->descriptor_destructor;
#endif
```

emilyfertig (changes requested):
Thank you!

emilyfertig:
Hi @mcuiaws , it would be great to get this merged, do you think you can address it in the next few days? (Otherwise I'll patch it and send a coauthored PR.) Thanks!

mcuiaws:
> Hi @mcuiaws , it would be great to get this merged, do you think you can address it in the next few days? (Otherwise I'll patch it and send a coauthored PR.) Thanks!

Yeah I can do it today.

emilyfertig (approved):
Thank you!

@derdrdirk could you please trigger the workflows for this?

derdrdirk (approved):

derdrdirk:
Approved.

PR #40383, opened 2026-04-03
[XLA:GPU] [oneAPI] Use symlinks to preserve RUNPATHS

mraunak:
Update copy_file in emitters/test to create symlink instead of copying. Using copy results in the copied executable not finding the shared libraries in rpaths at runtime. (e.g., with oneAPI build, //xla/codegen/emitters/tests:dynamic_update_slice/vectorize_x1_too_small.hlo.test, fails with error "gpu_fusion_to_mlir: error while loading shared libraries: libumf.so.0: cannot open shared object file."


penpornk (approved):
Sorry for the delay!
I'm not sure if this will break something else. We'll test more internally.

PR #40443, opened 2026-04-06
[xla:gpu] Add LoopbackCollectives + a flag to select implementation

ezhulenev:
Add `LoopbackCollectives` — a GPU collectives implementation that simulates collective ops via local memcpy instead of NCCL. Add `--xla_gpu_collectives_implementation` flag to select collectives implementation by name at runtime.

This is different from `mock_collectives`, in short: `mock_collectives` produces undefined output (skips ops), `LoopbackCollectives` produces correct single-rank output (executes ops with memcpy semantics). Loopback is better for testing model correctness on a single device, while mock is lighter-weight for testing graph structure/scheduling.

Mock collectives flag will be removed in followup PRs as it is too dangerous for any real testing.

```
// Semantic summary of each collective operation:
//
//   AllReduce:         memcpy(recv, send) — other ranks contribute identity
//   Broadcast:         memcpy(recv, send)
//   ReduceScatter:     memcpy this rank's chunk from send to recv
//   AllGather:         replicate send into every rank's slot in recv
//   AllToAll:          memcpy each send buffer to the corresponding recv buffer
//   CollectivePermute: memcpy(recv, send) — data loops back to self
//   Send:              no-op — data "sent" successfully
//   Recv:              zero-fill recv buffer to avoid non-finite values
```


xla/backends/gpu/collectives/loopback_collectives_test.cc
```diff
@@ -0,0 +1,334 @@
...
+  std::vector<RankId> keys(existing_comms.size());
+  std::vector<const Communicator*> existing_comms_ptrs(existing_comms.size());
+  for (size_t i = 0; i < existing_comms.size(); ++i) {
+    keys[i] = RankId(i);
+    existing_comms_ptrs[i] = existing_comms[i].get();
+  }
+
+  GpuCollectives::Config config;
+
+  TF_ASSIGN_OR_RETURN(auto comms, collectives->SplitCommunicators(
```
ermilovmaxim: s/TF_ASSIGN_OR_RETURN/ASSIGN_OR_RETURN/
We are in process of removing TF_* version

ermilovmaxim (approved):

xla/backends/gpu/collectives/loopback_collectives_test.cc
ezhulenev: Done! Pushed.

ermilovmaxim (approved):

ermilovmaxim (approved):

ermilovmaxim (approved):

ermilovmaxim (approved):

PR #40499, opened 2026-04-07
[ROCm] Add waves_per_eu support to Triton GEMM config

nurmukhametov:
Add the ROCm-specific `waves_per_eu` hint to the Triton GEMM config. This parameter specifies the minimum number of wavefronts per execution unit on AMD GPUs. The LLVM backend uses this to:
1) limit the number of SGPRs and VGPRs available per wave, which affects
   register allocation;
2) set register pressure thresholds for the instruction scheduler.

The parameter flows through: TritonGemmKey proto -> TritonGemmConfig struct -> BlockLevelFusionConfig proto -> BlockLevelParameters -> xtile_compiler, where it is applied as the "amdgpu-waves-per-eu" LLVM function attribute on the kernel, matching Triton's own AMD backend behavior. Default value of 0 means no restriction.

The autotuner search space is extended with values {0,1,2,4} for ROCm targets.

### DIFF
--- a/xla/autotuning.proto
+++ b/xla/autotuning.proto
@@ -85,6 +85,7 @@ message AutotuneResult {
     int64 num_ctas = 7;
     bool is_tma_allowed = 8;
     bool is_warp_specialization_allowed = 9;
+    int64 waves_per_eu = 10;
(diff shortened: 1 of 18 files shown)


vwbaker (approved):
LGTM. Thanks!

shyamliagrawal (approved):

PR #40500, opened 2026-04-07
[XLA:GPU] Remove mutable resource uses from Command; pass extras at construction

shawnwang18:
📝 Summary of Changes

  Remove mutable resource use tracking from Command and replace it with immutable construction-time injection.

  Specifically:
  - Remove resource_uses_ member, add_resource_use(), and resource_uses() from Command
  - Remove CollectResourceUses from CommandOperation in CommandExecutor
  - Add extra_resources parameter to CommandExecutor::Create() so callers supply per-command resource uses upfront
  - In command_buffer_cmd_emitter.cc, collect control-dependency tokens into ConversionContext::extra_resources and pass them to CommandExecutor::Create instead of mutating commands post-construction
  - In CreateCommandOperationsWithLHSMode, pre-compute LHS scheduling resource uses before constructing CommandOperation objects, eliminating all post-construction mutation

  🎯 Justification

  Command objects were previously mutable after construction via add_resource_use(). The emitter and executor would call this method after constructing commands to inject control and scheduling dependencies. This makes the lifecycle of Command objects hard to reason about — callers must know when it is safe to read resource_uses() and construction of the execution graph depends on mutation order.

  By pre-computing all resource uses before constructing CommandOperation objects and passing them in at instantiation time, Command objects become effectively immutable with respect to resource tracking. This simplifies the data flow, makes dependencies explicit at the call site, and removes the need for CollectResourceUses walking the command tree.

  🚀 Kind of Contribution

  ♻️  Cleanup

  N/A — this is a refactoring change with no algorithmic impact on the execution graph construction.

  🧪 Unit Tests

  Existing tests in xla/backends/gpu/runtime/command_executor_test.cc cover the execution graph construction logic for all synchronization modes (kConcurrent, kLHS, kConcurrentRegions, kSerialize). No new test cases are required as the observable behavior (dependency graph structure) is unchanged.

  🧪 Execution Tests

  Existing command buffer end-to-end tests in xla/pjrt/gpu/se_gpu_pjrt_client_test.cc (e.g., CommandBufferWithGemm, CommandBufferWithConditional, CommandBufferWithWhile) exercise the full path from emitter → command executor → command buffer recording and validate correctness of the dependency tracking.

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
+++ b/xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
@@ -54,12 +54,12 @@ limitations under the License.
 #include "xla/stream_executor/device_description.h"
 #include "xla/stream_executor/semantic_version.h"
 #include "xla/tsl/platform/errors.h"
+#include "xla/tsl/platform/status_macros.h"
 #include "xla/util.h"
 #include "xla/xla.pb.h"
 #include "tsl/platform/platform.h"
 #include "tsl/profiler/lib/profiler_lock.h"
 #include "tsl/profiler/lib/traceme.h"
-#include "xla/tsl/platform/status_macros.h"
 
 namespace xla {
 namespace gpu {
@@ -436,6 +436,13 @@ ConvertThunksToCommandBuffer(
       !thunks_to_convert.empty()
           ? absl::StrCat("_", thunks_to_convert.front()->thunk_info().thunk_id)
           : "");
+
+  if (VLOG_IS_ON(2)) {
+    auto graph = cmd_executor.RenderExecutionGraph();
+    if (graph.ok()) {
+      VLOG(2) << command_buffer_profile_annotation << " graph: " << *graph;
+    }
+  }
 
   Thunk::ThunkInfo thunk_info;
   thunk_info.profile_annotation = command_buffer_profile_annotation;
--- a/xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
+++ b/xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
@@ -437,13 +437,6 @@ ConvertThunksToCommandBuffer(
           ? absl::StrCat("_", thunks_to_convert.front()->thunk_info().thunk_id)
           : "");
 
-  if (VLOG_IS_ON(2)) {
-    auto graph = cmd_executor.RenderExecutionGraph();
-    if (graph.ok()) {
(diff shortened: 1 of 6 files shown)


ezhulenev (commented):
Thank you!

ezhulenev (approved):

ddunl (approved):

dimitar-asenov (approved):

xla/backends/gpu/runtime/command_executor.cc
```diff
@@ -724,8 +723,9 @@ absl::StatusOr<std::string> CommandExecutor::RenderExecutionGraph() {
         "concurrent/LHS synchronization mode");
   }
 
-  TF_ASSIGN_OR_RETURN(auto operations, CreateCommandOperations(
-                                           commands_, synchronization_mode_));
+  TF_ASSIGN_OR_RETURN(auto operations,
+                      CreateCommandOperations(commands_, synchronization_mode_,
+                                              /*extra_resources=*/{}));
```
thomasjoerg: This PR broke `RenderExecutionGraph`. The graph does not contain any extra_resources, because they are not provided here.

shawnwang18: I put the graph rendering code just after the orignal create operations. please take a look.

xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
```diff
@@ -437,13 +437,6 @@ ConvertThunksToCommandBuffer(
           ? absl::StrCat("_", thunks_to_convert.front()->thunk_info().thunk_id)
           : "");
 
-  if (VLOG_IS_ON(2)) {
-    auto graph = cmd_executor.RenderExecutionGraph();
-    if (graph.ok()) {
-      VLOG(2) << command_buffer_profile_annotation << " graph: " << *graph;
```
thomasjoerg, line 443: Note that this LOGs both, the command buffer's name (profile annotation) and the rendered graph. This is very useful for debugging. Can we please restore this?

thomasjoerg (changes requested):

xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
shawnwang18: Ah, I see. I made the fix to make RenderExecutionGraph() works correctly, can you review it?

thomasjoerg (approved):

PR #40502, opened 2026-04-07
[ROCm] Fix Group-gemm e2e tests to meet gfx950 hip backend requirements

mfrancepillois:
📝 Summary of Changes
The datatypes of group-gemm e2e tests have been changed from bf16 to fp16, as FP16 is currently the only data type supported by HipBLAS for group-GEMM on the gfx950.

🎯 Justification
The Hipblast Group-Gemm backend does not support as many configurations on the gfx950 as it does on the gfx942 (as mentioned https://github.com/openxla/xla/blob/645d29fb62ed2f433af48dc5cf505ac9e992cb3c/xla/hlo/transforms/expanders/ragged_dot_rewriter.cc#L360-L363).
The e2e tests have therefore been adapted to meet these requirements.

🚀 Kind of Contribution
Please remove what does not apply: 🐛 Bug Fix,  🧪 Tests

### DIFF
--- a/xla/backends/gpu/transforms/cublas_gemm_rewriter_test.cc
+++ b/xla/backends/gpu/transforms/cublas_gemm_rewriter_test.cc
@@ -3743,10 +3743,10 @@ TEST_F(GroupedGemmRewriteTest, CustomCallTargetGroupedGemm) {
 HloModule GroupedGemm
 
 ENTRY AddRaggedDotsFunc {
-    p0 = bf16[64,9]{1,0} parameter(0)
-    p1 = bf16[2,9,8]{2,1,0} parameter(1)
+    p0 = f16[64,9]{1,0} parameter(0)
+    p1 = f16[2,9,8]{2,1,0} parameter(1)
     p2 = s32[2] constant({16, 48})
-    ROOT ragged-dot = bf16[64,8]{1,0} ragged-dot(p0, p1, p2),
+    ROOT ragged-dot = f16[64,8]{1,0} ragged-dot(p0, p1, p2),
                       lhs_contracting_dims={1}, rhs_contracting_dims={1},
                       lhs_ragged_dims={0}, rhs_group_dims={0}
 }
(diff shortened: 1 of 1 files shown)


akuegel (changes requested):
I don't think we want to replace all tests that use bf16 with tests that use f16. Maybe instead we should make these tests parametric so that the type can be passed, and then bf16 type can be skipped on ROCM?

mfrancepillois:
> I don't think we want to replace all tests that use bf16 with tests that use f16. Maybe instead we should make these tests parametric so that the type can be passed, and then bf16 type can be skipped on ROCM?

@akuegel Thanks for your feedback.
These tests run only on ROCM target as they tests the ROCm Group-Gemm Hipblaslt support. The fact is that the Hipblaslt implementation currently supports only a few data types, FP16 is supported on both target MI300 and MI350, but BF16 is only supported on MI300.
As these tests already take quite a long time to run and the XLA code paths being tested are identical whether FP16 or BF16 is used, I chose to replace the types rather than use parametric types in order to save execution time.
But if you have a strong preference for parametric formats, we could use them, but we would limit the data type to a single one, depending on the target: bf16 for the MI300 and fp16 for the MI355. WDYT?

akuegel:
> > I don't think we want to replace all tests that use bf16 with tests that use f16. Maybe instead we should make these tests parametric so that the type can be passed, and then bf16 type can be skipped on ROCM?
> 
> @akuegel Thanks for your feedback. These tests run only on ROCM target as they tests the ROCm Group-Gemm Hipblaslt support. The fact is that the Hipblaslt implementation currently supports only a few data types, FP16 is supported on both target MI300 and MI350, but BF16 is only supported on MI300. As these tests already take quite a long time to run and the XLA code paths being tested are identical whether FP16 or BF16 is used, I chose to replace the types rather than use parametric types in order to save execution time. But if you have a strong preference for parametric formats, we could use them, but we would limit the data type to a single one, depending on the target: bf16 for the MI300 and fp16 for the MI355. WDYT?

Thanks, I was not aware those tests run on ROCM only. I guess ideally those tests could be extracted into a different test target, to make it more obvious? But that is then independent of this PR.

akuegel (approved):

PR #40505, opened 2026-04-07
[xla:gpu] Use connected components to resolve CollectivePermute replica groups

ezhulenev:
Today for collective permute with these send-recv pairs: `{{0,1},{2,3}...},{{8,9}, {10,11},...}` we end up creating one GPU clique that spans all devices. NCCL performance degrades A LOT when send/recv operations issued on a clique that includes large number of devices (surprising, but it is a fact!).

Instead use connected components to create smaller cliques that capture only the communicating devices. It does increase the total number of cliques in the job, but for each individual device we almost always have the same number of communicators. For the example above we will create cliques with just two communicating devices: `{0,1}`, `{2,3}`, etc...

This makes collective permute that happens within NVLINK domain 10-20x times faster!


wsmoses:
cc @giordano @pangoraw @romanlee @gbaraldi

pifon2a (approved):

PR #40529, opened 2026-04-08
[XLA:GPU] Dockerized Hermetic Tests for oneAPI

mraunak:
This PR updates the ci_test_xla.sh script for oneAPI and introduces ci_docker_test.sh to run all XLA test cases inside a fully hermetic Docker environment.


build_tools/sycl/ci_docker_test.sh
```diff
@@ -0,0 +1,58 @@
...
+  sudo tee /etc/apt/sources.list.d/intel-gpu-$(lsb_release -cs).list
+
+# Step 3: Install Intel GPU Runtime Packages
+apt-get update && \
+  apt-get install -y --no-install-recommends \
+    intel-opencl-icd intel-ocloc libze1 libze-dev xpu-smi && \
+  apt-get clean && rm -rf /var/lib/apt/lists/*
+
+cd /xla
+bash build_tools/sycl/ci_test_xla.sh
```
MichaelHudgins: This seems basically the same code as is in the if condition if we are already in a container.  I recommend we break this out into its own bash script so we can prevent this file writing in the docker container (because the code is presumably checked out in the container) 

That would allow us to not write an inline file and simplify the if else logic in this script.

build_tools/sycl/ci_test_xla.sh
```diff
@@ -15,11 +15,75 @@
...
+
+jobs=256
+no_of_gpus=$(ls /dev/dri/ | fgrep render | wc -l)
+if [[ "${no_of_gpus}" -eq 0 ]]; then
+  echo "unknown number of gpus."
+  exit 1
+fi
+
+local_test_jobs=$((no_of_gpus * 8))
+if [[ ${jobs} -lt 1 || ${jobs} -ge 512 ]]; then
```
MichaelHudgins: This if condition doesn't seem needed.  The value of jobs looks hard set to 256 above.

build_tools/sycl/ci_test_xla.sh
```diff
@@ -15,11 +15,75 @@
...
+fi
+if [[ ${local_test_jobs} -lt 1 || ${local_test_jobs} -ge 64 ]]; then
+  echo "local_test_jobs invalid setting ${local_test_jobs}"
+  exit 1
+fi
+
+TEST_TARGETS="${TEST_TARGETS:-//xla/stream_executor/...}"
+TEST_TO_SKIP="${TEST_TO_SKIP:-}"
+TEST_TO_ADD="${TEST_TO_ADD:-}"
+OPTIONAL_BAZEL_TEST_OPTIONS="${OPTIONAL_BAZEL_TEST_OPTIONS:-}"
```
MichaelHudgins: Could you please add some usage instructions or expectations to the script?   

There are a lot of envs possible to be used by the caller. 

Are these in practice used or can this logic be simplified?

MichaelHudgins (changes requested):

mraunak:
Hi @MichaelHudgins, simplified the changes, just kept the script needed for the public CI run

MichaelHudgins (approved):

PR #40531, opened 2026-04-08
Fix sort comparison type mismatch causing tf.random.shuffle XLA failure

ssam18:
createSortOp was setting TOTALORDER whenever any element type was float, but the sort comparator only ever compares the first element type (the sort key). When shuffling a float tensor, the keys are i32 and the values are f32 — the loop found f32 and set TOTALORDER on the i32 comparison, which the HLO verifier rejects with "Expected comparison type SIGNED, actual: TOTALORDER".

The fix narrows the check to the first element type only, so integer keys correctly fall back to the default SIGNED comparison.


allanrenucci (approved):

allanrenucci (changes requested):
LGTM. But could you add a test for this? We have existing tests under /xla/mlir_hlo/tests/

PR #40585, opened 2026-04-08
Cross Host Data Transfers: Unify Prepared{Send,Receive} into PreparedTransfer

rao-ashish:
📝 Summary of Changes

This PR is the first in a sequence of PRs that will refactor cross-host data transfer implementations to eventually rely on a shared helper function `CrossHostTransferBuffers`. `CrossHostTransferBuffers` is planned to eventually be integrated into the PJRT APIs to enable receiving data into preallocated receive buffers (this feature is being planned in collaboration with @gspschmid, @emilyfertig, and @pschuh). As a first step, this PR unifies `Prepared{Send,Receive}` structs into a single `PreparedTransfer` struct.

This PR also moves waiting for dependency events out of NCCL group calls since the benefit of a NCCL group section is to aggregate the collectives launched inside of it (unrelated to waiting on dependency events).

🎯 Justification

It is difficult to achieve good comm/compute overlap with cross-host data transfers as the current implementation always allocates receive-buffers 'just-in-time', and because the GPU memory allocator blocks on the compute stream. `CrossHostTransferBuffers` will enable users to receive into preallocated receive buffers, making it easier to avoid the allocator blocking issue. This PR is a first step towards implementing `CrossHostTransferBuffers`.

🚀 Kind of Contribution

🧪 Unit Tests:
This PR only refactors the implementation of `CrossHost{Send/Receive}Buffers`, so the pre-existing unit tests for those methods already test this PR.

🧪 Execution Tests:

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -418,79 +418,41 @@ using AcquiredCliqueAndCommunicator =
     std::pair<std::shared_ptr<gpu::LockableGpuClique::Lock>,
               gpu::GpuCommunicator*>;
 
-class PreparedSend {
+class PreparedTransfer {
  public:
   StreamExecutorGpuClient* client_;
   gpu::GpuCliqueKey clique_key_;
-  tsl::RCReference<CommonPjRtRawBuffer> raw_buffer_;
-  std::vector<tsl::RCReference<tsl::AsyncValue>> definition_events_;
-  tsl::AsyncValueRef<BufferSequencingEvent> usage_event_;
+  tsl::RCReference<PjRtRawBuffer> raw_buffer_;
+  tsl::AsyncValueRef<BufferSequencingEvent> transfer_event_;
   AcquiredCliqueAndCommunicator clique_and_communicator_;
-  std::shared_ptr<Promise<>> promise_;
+  bool is_sender_;
 
-  PreparedSend(StreamExecutorGpuClient* client, gpu::GpuCliqueKey clique_key,
-               tsl::RCReference<CommonPjRtRawBuffer> raw_buffer,
-               std::vector<tsl::RCReference<tsl::AsyncValue>> definition_events,
-               tsl::AsyncValueRef<BufferSequencingEvent> usage_event,
-               AcquiredCliqueAndCommunicator clique_and_communicator,
-               std::shared_ptr<Promise<>> promise)
+  PreparedTransfer(StreamExecutorGpuClient* client,
+                   gpu::GpuCliqueKey clique_key,
+                   tsl::RCReference<PjRtRawBuffer> raw_buffer,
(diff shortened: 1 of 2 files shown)


emilyfertig:
Thanks @rao-ashish ! This looks good to me. I'll let @pschuh take a look too and approve.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -648,83 +562,64 @@ absl::StatusOr<PreparedReceive> PrepareReceive(
...
+  grouped.reserve(prepared_transfers.size());
+  for (auto&& prepared_transfer : prepared_transfers) {
+    grouped[prepared_transfer.clique_key_].push_back(
+        std::move(prepared_transfer));
   }
   return grouped;
 }
 
-absl::Status FulfillDeviceEvent(
+void FulfillDeviceEvent(
```
pschuh: No longer returning the status? Can you justify this change a little?

rao-ashish: Previously, if `FulfillDeviceEvent` returned an error, we were using this to also set the futures returned by `CrossHostSendBuffers` to be errors (see [here](https://github.com/openxla/xla/blob/d1fdcd474cfbea9a50dc60468442d58d3309e20f/xla/pjrt/gpu/se_gpu_pjrt_client.cc#L935)). Now, we are registering a callback on `usage_event` to fulfill those promises [here](https://github.com/openxla/xla/pull/40585/changes#diff-d908640f705270793348b89d82d4df0d433b496664846ba18253f73159dc6c3eR695). This makes it no longer necessary for `FulfillDeviceEvent` to return an error; when `FulfillDeviceEvent` sets `usage_event` to an error state, the callback on `usage_event` will automatically also set the promises to an error state.

On the receiver side, we were previously just [asserting](https://github.com/openxla/xla/blob/d1fdcd474cfbea9a50dc60468442d58d3309e20f/xla/pjrt/gpu/se_gpu_pjrt_client.cc#L1162) that the output of `FulfillDeviceEvent` is ok with `CHECK_OK`. This isn't really necessary because the error still propagates to the caller through the `definition_event` on which `FulfillDeviceEvent` was called.

Since in both cases the error is already being propagated through the device events, I figured we could just remove the returned status.

pschuh (approved):

emilyfertig (approved):

emilyfertig:
Hi @derdrdirk , @pschuh and I approved this PR but we don't have the right permissions to trigger the workflows, could you please help us get this merged? 

This happens every time we review PRs from external collaborators, so I've been tagging the OpenXLA Github oncall each time and asking for help, is that the right thing to do or is there another process we should follow? (Parker is an owner of PJRT in Google3, maybe he could be granted permission to trigger the workflows?)

hawkinsp (approved):

PR #40593, opened 2026-04-08
[XLA:GPU] Ensure that the host and target platforms match for AOT compilation

akhilgoe:
XLA compiler library assumes that the host platform matches the requested target platform for deviceless compilation. This PR:
1. Adds a check to ensure that the two platforms are the same and proceeds with AOT compilation only when that is the case.
2. Adds a oneAPI compatible spec file to the deviceless compilation test
3. Adds a test case to verify that compilation is rejected when the host and target platforms don't match.


xla/tools/xla_compile_lib.cc
```diff
@@ -115,6 +115,14 @@ static absl::StatusOr<std::string> CompileGpuExecutable(
       .set_xla_gpu_experimental_aot_compiled_thunks(true);
 
   if (aot) {
+    TF_ASSIGN_OR_RETURN(se::Platform::Id target_platform_id,
+                        xla::PlatformUtil::GetPlatformIdFromCanonicalName(
+                            (*target_config).platform_name));
+    if (platform_id != target_platform_id) {
+      LOG(FATAL) << "Attempting to AOT compile for "
+                 << (*target_config).platform_name
+                 << ", but the current platform is " << platform_name << ".";
```
beckerhe: Can you make this return an error instead of crashing?

This also avoids the need for a death test below.

xla/tools/xla_gpu_compile_lib_test.cc
```diff
@@ -119,6 +121,34 @@ TEST_F(XlaCompileLibTest, CompilesForGpuWithoutDevice) {
...
+TEST_F(XlaCompileLibTest, FatalCrashForIncorrectPlatformAOTCompile) {
+  // Running death test in a single-threaded environment
+  GTEST_FLAG_SET(death_test_style, "threadsafe");
+  // Incorrect spec file assignment to trigger an AOT compilation failure.
+  const std::string incorrect_spec_file =
+      test_runner().HasProperty(HloRunnerPropertyTag::kUsingGpuRocm)
+          ? "h100_sxm.txtpb"
+      : test_runner().HasProperty(HloRunnerPropertyTag::kUsingGpuOneAPI)
+          ? "mi200.txtpb"
+          : "bmg_g21.txtpb";
```
beckerhe: This is a bit hard to grasp on first sight. Could you make this a bit more readable?

An immediately invoked lambda expression with early returns would work well in my opinion:

```cpp
const std::string incorrect_spec_file = [&] {
  if (test_runner().HasProperty(HloRunnerPropertyTag::kUsingGpuRocm)) {
    return "h100_sxm.txtpb";
  }
  // ...
}();
```

xla/tools/xla_gpu_compile_lib_test.cc
```diff
@@ -100,6 +100,8 @@ TEST_F(XlaCompileLibTest, CompilesForGpuWithoutDevice) {
   const std::string spec_file =
       test_runner().HasProperty(HloRunnerPropertyTag::kUsingGpuRocm)
           ? "mi200.txtpb"
+      : test_runner().HasProperty(HloRunnerPropertyTag::kUsingGpuOneAPI)
+          ? "bmg_g21.txtpb"
           : "h100_sxm.txtpb";
```
beckerhe: This becomes a bit hard to read. Could you maybe convert this to an immediately invoked lambda expression? (Example in my other comment below)

akuegel:
@beckerhe Can you please take a look at this PR? I quickly checked and it seems to make sense to me, just wanted to double-check whether you agree that we shouldn't use AOT for a different platform than the one we are running on? Different device is ok, but not different platform?

akhilgoe (commented):
Adding a bit more context on why I think this check should be a good idea: 

LLVM uses the host platform backend for code generation. So if someone tries to AOT compile using a Hopper config on say a MI200 series machine, LLVM would complain as it cannot select nvvm intrinsics with gcn backend. 

@akuegel @beckerhe

beckerhe (commented):
In general makes sense to me, but it would be great if you could avoid the death test. (Added some suggestions below)

xla/tools/xla_gpu_compile_lib_test.cc
akhilgoe: Done

xla/tools/xla_gpu_compile_lib_test.cc
akhilgoe: Sounds good, done!

xla/tools/xla_compile_lib.cc
akhilgoe: Done

akhilgoe:
@beckerhe I added a commit to address your comments. Please take a look. Also I resolved the dwyu errors for `xla_gpu_compile_lib_test` but did not touch `xla_compile_lib` because those seemed unrelated to the changes I made.

akhilgoe:
A gentle ping! @beckerhe

beckerhe (approved):
Sry for the delay. LGTM!

akhilgoe:
Thanks @beckerhe! I took the DWYU recommendation and removed the gtest dependency but that seems to be causing trouble with the L4 build. Added it back, please take a look.

dimitar-asenov (approved):

akhilgoe:
@beckerhe @dimitar-asenov @xla-rotation 

A recent commit (https://github.com/openxla/xla/commit/4d732da794e0915f0bd5f926587050b12a83a4ed) expanded the `aot_compile` condition to include cases when compiling for the attached device. This caused a UT regression with this PR. I have handled the case and all the tests should pass now. Please take a look.

akuegel (approved):

PR #40601, opened 2026-04-09
[GPU] Infer collective-permute type by slice_sizes

terryysun:
📝 Summary of Changes
Removed `inplace_` attribute in collective-permute. Rely on `slice_sizes` to infer if a collective-permute is in-place, since this attribute is specific to in-place.

🎯 Justification
HLO parser is not aware of the `inplace_` attribute, so it only works end-to-end but won't work when run from HLO. Its functionality can be substituted by inferring from a in-place specific attribute `slice_sizes`.

🚀 Kind of Contribution
♻️ Cleanup

N/A

🧪 Unit Tests:
No functionality change, existing tests.

🧪 Execution Tests:
No functionality change, existing tests.

### DIFF
--- a/xla/hlo/builder/xla_builder.cc
+++ b/xla/hlo/builder/xla_builder.cc
@@ -283,19 +283,17 @@ XlaOp XlaBuilderFriend::BuildCopyDone(XlaBuilder* builder, const XlaOp operand,
 XlaOp XlaBuilderFriend::BuildCollectivePermuteStart(
     XlaBuilder* builder, XlaOp operand,
     const std::vector<std::pair<int64_t, int64_t>>& source_target_pairs,
-    const std::optional<ChannelHandle>& channel_id, const bool inplace) {
+    const std::optional<ChannelHandle>& channel_id) {
   return builder->CollectivePermuteImpl(operand, source_target_pairs,
-                                        channel_id, /*async=*/true, inplace);
+                                        channel_id, /*async=*/true);
 }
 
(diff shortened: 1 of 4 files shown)


akuegel (approved):
Thank you :)

dimitar-asenov:
We had to revert this because it was breaking a test. Will post more info later.

terryysun:
Hi @dimitar-asenov could you share more info about the test failure? Thanks!

dimitar-asenov:
It's an internal test that fails with `SIGILL` but only under ASAN with an optimized build. I'm fairly certain that the failure is outside of GPU code. Since it's only detected with ASAN, it might be some undefined behavior to do with uninitialized fields. Though I can't quite see where the issue is coming from and the internal test failure doesn't come with sufficient logging so I couldn't get quickly get better information.

Will dig some more today or tomorrow.

terryysun:
Hi @dimitar-asenov did you get a chance to look into it? Thanks!

dimitar-asenov:
@terryysun Sorry for the delay on this. I think I found the issue: `xla/python/ops.cc` also needed to be adjusted. This failed in a highly non-intuitive way and it took a while longer to debug.

I will patch the change internally and re-merge the PR. There's no need for any action on your end.

dimitar-asenov:
The change has been successfully merged:
https://github.com/openxla/xla/commit/1444242f205550694b9785fa3255d7053f54d453

PR #40603, opened 2026-04-09
[XLA:GPU][oneAPI] Add target config spec file for PVC

akhilgoe:
This PR adds a target config spec file for Intel's Data Center GPU Max, codenamed Ponte Vecchio (PVC). It also adds a corresponding entry for target config test.


derdrdirk (approved):
LGTM

PR #40605, opened 2026-04-09
[XLA:GPU] Merge CommandType into Thunk::Kind

shawnwang18:
Now that Command is a subclass of Thunk, the separate CommandType enum is redundant. Every Command instance already carries a Thunk::Kind via the inherited kind() accessor — there is no reason to maintain a parallel type identifier.

  This CL removes CommandType entirely and folds its values into Thunk::Kind as Command-specific entries (e.g. kEmptyCmd, kGemmCmd, kAllGatherCmd), following the same go/keep-sorted ordering used by the rest of the enum.

  Changes:
  - Remove CommandType enum, XLA_GPU_COMMAND_LIST macro, and CommandTypeString() from command.h/command.cc
  - Remove kCommand from Thunk::Kind; add 30 Command-specific Kind entries in sorted order
  - Command constructor now takes Thunk::Kind directly; cmd_type_ member and command_type() accessor removed — callers use the inherited kind()
  - IsCollectiveCommand now operates on Thunk::Kind instead of CommandType
  - KindToString and KindToProto in thunk.cc updated for the new kinds (Command kinds map to THUNK_KIND_UNSPECIFIED in proto, same as kCommand did)
  - TracedCommandBufferCmd and CollectiveCmd constructor signatures updated accordingly

  Note: CommandType::kAsyncDone was renamed to Thunk::Kind::kAsyncDoneCmd to avoid collision with the existing Thunk::Kind::kAsyncDone used by AsyncDoneThunk.

### DIFF
--- a/xla/backends/gpu/runtime/command.cc
+++ b/xla/backends/gpu/runtime/command.cc
@@ -15,31 +15,19 @@ limitations under the License.
 
 #include "xla/backends/gpu/runtime/command.h"
 
-#include <string>
-
 namespace xla::gpu {
 
-std::string CommandTypeString(CommandType type) {
-  switch (type) {
-#define CASE_CMD_STRING(enum_name, cmd_name, ...) \
-  case CommandType::enum_name:                    \
-    return cmd_name;
-    XLA_GPU_COMMAND_LIST(CASE_CMD_STRING)
-#undef CASE_CMD_STRING
-  }
-}
-
-bool IsCollectiveCommand(CommandType type) {
-  switch (type) {
-    case CommandType::kAllGatherCmd:
-    case CommandType::kAllReduceCmd:
-    case CommandType::kAllToAllCmd:
-    case CommandType::kCollectiveBroadcastCmd:
-    case CommandType::kCollectivePermuteCmd:
-    case CommandType::kRaggedAllToAllCmd:
-    case CommandType::kReduceScatterCmd:
-    case CommandType::kRecvCmd:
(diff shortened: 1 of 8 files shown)


ezhulenev:
I don't like it because it's too disruptive and adds too much stuff at one, how about start from `EmptyCmd` and show how proper migratiuon will look like:

1. empty_thunk.{h.cc}
2. class EmptyThunk : Command
3. Execute - does nothing at all, Record - records
4. Add kEmtpy (kNoOp?) kind to thunk.

command_buffer_cmd.h should be gone by the end of this process, and each Thunk (Command) should live in its own thunk file.

Then do the same to kKernel, kCustomKernel. Just get it working with a few thunks/commands first to see how it "feels" and how hard is the migration process.

shawnwang18:
> I don't like it because it's too disruptive and adds too much stuff at one, how about start from `EmptyCmd` and show how proper migratiuon will look like:
> 
> 1. empty_thunk.{h.cc}
> 2. class EmptyThunk : Command
> 3. Execute - does nothing at all, Record - records
> 4. Add kEmtpy (kNoOp?) kind to thunk.
> 
> command_buffer_cmd.h should be gone by the end of this process, and each Thunk (Command) should live in its own thunk file.
> 
> Then do the same to kKernel, kCustomKernel. Just get it working with a few thunks/commands first to see how it "feels" and how hard is the migration process.

We can do the thunk migration first, then remove the CommandId from Command. After all commands are migrated, that seems will havec less modifications.

PR #40609, opened 2026-04-09
[XLA:GPU] Migrate ComputationIdCmd to ReplicaOrPartitionIdThunk

shawnwang18:
As part of the ongoing effort to consolidate the Command and Thunk hierarchies, this CL migrates ComputationIdCmd by folding its Record() implementation directly into ReplicaOrPartitionIdThunk (and by extension its ReplicaIdThunk / PartitionIdThunk subclasses).

  Changes:

  - ReplicaOrPartitionIdThunk now inherits from Command instead of Thunk, implementing Record() with the logic previously in ComputationIdCmd.
  - ComputationIdCmd is deleted from command_buffer_cmd.h/.cc.
  - A new protected Command constructor is introduced that accepts an explicit Thunk::Kind and ThunkInfo, allowing existing Thunk subclasses to self-describe via their native kind rather than using Kind::kCommand. This means the command-buffer conversion pass and emitter switch cases require no changes.
  - command_buffer_cmd_emitter.cc is updated to dispatch directly to the thunk's Record() instead of constructing a ComputationIdCmd.

  Why:

  The goal is to eliminate the parallel *Cmd hierarchy for thunks that already carry all the state needed to record themselves into a command buffer. Thunks that implement Command can be recorded directly, reducing indirection and the maintenance burden of keeping two representations in sync.


xla/backends/gpu/runtime/replica_id_thunk.cc
```diff
@@ -44,6 +50,41 @@ absl::Status ReplicaOrPartitionIdThunk::ExecuteOnStream(
...
+ReplicaOrPartitionIdThunk::Record(const Thunk::ExecuteParams& execute_params,
+                                  const RecordParams& record_params,
+                                  RecordAction record_action,
+                                  se::CommandBuffer* command_buffer) {
+  se::DeviceAddressBase dst =
+      execute_params.buffer_allocations->GetDeviceAddress(dest_);
+
+  GlobalDeviceId global_device_id =
+      execute_params.collective_params->global_device_id;
+  TF_ASSIGN_OR_RETURN(
```
ezhulenev: prefer regular ASSIGN_OR_RETURN in new code

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
```diff
@@ -260,14 +260,12 @@ static absl::StatusOr<std::unique_ptr<Command>> Convert(
 
 static absl::StatusOr<std::unique_ptr<Command>> Convert(
     const PartitionIdThunk& thunk) {
-  return std::make_unique<ComputationIdCmd>(thunk.dest(),
-                                            ComputationIdCmd::Kind::kPartition);
+  return std::make_unique<PartitionIdThunk>(thunk.thunk_info(), thunk.dest());
 }
 
 static absl::StatusOr<std::unique_ptr<Command>> Convert(
     const ReplicaIdThunk& thunk) {
```
ezhulenev: This is a problem, this API works only for copyable thunks, I think we need a wrapper, somethiung like


```
CommandWrapperThunk(Command* command)
```

that will forward everything to `command`.

ezhulenev: Or just call it `CommandWrapper`, command buffer thung guarantees that original sequence stays alive, so we can easily capture a pointer and will be safe!

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
```diff
@@ -260,12 +260,19 @@ static absl::StatusOr<std::unique_ptr<Command>> Convert(
...
+      const_cast<PartitionIdThunk*>(&thunk));
 }
 
 static absl::StatusOr<std::unique_ptr<Command>> Convert(
     const ReplicaIdThunk& thunk) {
-  return std::make_unique<ReplicaIdThunk>(thunk.thunk_info(), thunk.dest());
+  // ReplicaIdThunk implements Command directly; wrap a pointer to it rather
+  // than copying. Safe because the original thunk sequence is kept alive by
+  // CommandBufferThunk for the duration of the command buffer.
+  return std::make_unique<CommandWrapper>(const_cast<ReplicaIdThunk*>(&thunk));
```
ezhulenev: const_casts are banned at google, I'd recommend update APU to take non-const refs, or pointers

ezhulenev (approved):

ezhulenev:
👍

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
shawnwang18: I think the CommandWrapper is a temporal class during the migration, after all Commands are migrated, can we change the CommandSequence to have shared pointer?  then we do not need the wraper.

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
shawnwang18: fixed the API.

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
ezhulenev: Yes, after everything is migrated, CommandBuffeThunk should be able to pass owned thunk sequence to command executor, basically command sequence will be a wrapper for ThunkSequence were each Thunk has type Command. I think in the end it should be pretty clean.

xla/backends/gpu/runtime/replica_id_thunk.cc
shawnwang18: fixed

akuegel (approved):

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
```diff
@@ -362,10 +353,13 @@ static absl::Status AppendCommands(ConversionContext& ctx,
       return append(Convert<RecvThunk>(thunk));
     case Thunk::Kind::kSend:
       return append(Convert<SendThunk>(thunk));
+    // These thunks implement Command directly; append borrowed pointers.
     case Thunk::Kind::kPartitionId:
-      return append(Convert<PartitionIdThunk>(thunk));
+      cmd_sequence.Append(&static_cast<PartitionIdThunk&>(thunk));
```
akuegel: Our lint checker complains:

`Are you taking an address of a cast?  This is dangerous: could be a temp var.  Take the address before doing the cast, rather than after`

Same in line 361 below.

shawnwang18: Oh, thanks very much. Just push the fix.

dimitar-asenov (approved):

PR #40656, opened 2026-04-09
[GPU] Wire deadlock prevention for async collective multi-streaming

terryysun:
📝 Summary of Changes
Wire the deadlock prevention for async collective multi-streaming and fix a cast issue.

🎯 Justification
Deadlock prevention is necessary for multi-streaming.

🚀 Kind of Contribution
🐛 Bug Fix, ♻️ Cleanup

N/A

🧪 Unit Tests:
Reactivated existing unit test and added new unit test for util function.

🧪 Execution Tests:
Existing tests.

### DIFF
--- a/xla/service/collective_ops_utils.cc
+++ b/xla/service/collective_ops_utils.cc
@@ -205,10 +205,6 @@ absl::StatusOr<std::unique_ptr<CollectiveDeviceListBase>> GetAsyncReplicaGroups(
       instruction->opcode() == HloOpcode::kAllReduceStart) {
     return instruction->device_list()->Clone();
   }
-  // Handle unwrapped inner collectives.
-  if (DynCast<HloCollectiveInstruction>(instruction) != nullptr) {
-    return instruction->device_list()->Clone();
-  }
   return InvalidArgument(
       "Unexpected instruction type: %s is not an async collective "
       "instruction",
--- a/xla/service/collective_ops_utils.cc
+++ b/xla/service/collective_ops_utils.cc
@@ -205,6 +205,10 @@ absl::StatusOr<std::unique_ptr<CollectiveDeviceListBase>> GetAsyncReplicaGroups(
       instruction->opcode() == HloOpcode::kAllReduceStart) {
     return instruction->device_list()->Clone();
   }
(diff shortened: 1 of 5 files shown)


kredd2506:
A couple of questions as I'm trying to understand this area better:

1. Is the "1 overlap safe, 2+ unsafe" threshold because a deadlock cycle needs at least 2 shared ranks to form (A waits for B via one rank, B waits for A via another)?

2. The `DynCast<HloCollectiveInstruction>` fallback in `GetAsyncReplicaGroups` — which pass unwraps async collectives before scheduling?

Also noticed CI is failing on multiple targets (XLA Linux x86 CPU, ARM64 CPU, TF GPU L4) — is the re-enabled `OverlappingRanksPreventOverlappingCollectives` test surfacing something, or unrelated?

akuegel (changes requested):
As mentioned by @kredd2506 there are test failures with this PR. E.g. GpuLatencyHidingSchedulerBaseTest.ScheduleMultipleParallelResourcesShouldNotCrash (in gpu_latency_hiding_scheduler_test.cc)
Can you please check and fix them?

terryysun:
> A couple of questions as I'm trying to understand this area better:
> 
> 1. Is the "1 overlap safe, 2+ unsafe" threshold because a deadlock cycle needs at least 2 shared ranks to form (A waits for B via one rank, B waits for A via another)?
> 2. The `DynCast<HloCollectiveInstruction>` fallback in `GetAsyncReplicaGroups` — which pass unwraps async collectives before scheduling?
> 
> Also noticed CI is failing on multiple targets (XLA Linux x86 CPU, ARM64 CPU, TF GPU L4) — is the re-enabled `OverlappingRanksPreventOverlappingCollectives` test surfacing something, or unrelated?

Thanks for the comments! regarding the questions above,
1. Yes.
2. Async collectives are not unwrapped before scheduling, it's the implementation limitation of the overlap limit util and should not be spilled to other utils. Refactored to keep it contained to the overlap limit calculation util.

Some of the failures may not be relevant, but the one @akuegel mentioned is caused by changes in this PR. Fixed.

akuegel (approved):

PR #40702, opened 2026-04-10
[ROCm] Label FP8 fast accumulation as supported on ROCm 7

amd-jianli12:
📝 Summary of Changes
Set is_algorithm_supported to true since FP8 fast accumulation is supported on ROCm 7.

🎯 Justification
HLOs in test cases `dot_any_f8_any_f8_f32_fast_accum_with_lhs_f8e4m3fn_rhs_f8e4m3fn_output_f8e5m2_from_cc_8_9_rocm_63_no_restriction_*` run successfully on ROCm 7, so label them as supported. Requirement of minimum ROCm version with FP8 fast accumulation support should be 6.3.

🚀 Kind of Contribution
🐛 Bug Fix

### DIFF
--- a/xla/service/gpu/dot_algorithm_support_test.cc
+++ b/xla/service/gpu/dot_algorithm_support_test.cc
@@ -167,6 +167,17 @@ TEST_P(DotAlgorithmSupportTest, AlgorithmIsSupportedFromCudaCapability) {
   if (const auto* ccc = gpu_cc.cuda_compute_capability()) {
     is_algorithm_supported =
         ccc->SupportsAllFeaturesOf(params.min_cuda_capability);
+
+    // CublasLt does not support FP8 fast accumulation.
+    DebugOptions debug_options = GetDebugOptionsForTest();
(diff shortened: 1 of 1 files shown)


xla/service/gpu/dot_algorithm_support_test.cc
```diff
@@ -188,17 +188,6 @@ TEST_P(DotAlgorithmSupportTest, AlgorithmIsSupportedFromCudaCapability) {
     }
   }
 
-  // CublasLt does not support FP8 fast accumulation.
```
dimitar-asenov, line 191: Doesn't this code affect both ROCm and CUDA? Wouldn't removing this cause failure on the CUDA path? Did you test this with Nvidia hardware?

amd-jianli12: Great catch, absolutely I should keep this change ROCm only. Let me update.

amd-jianli12:
Hi @dimitar-asenov, I have updated the PR, would you kindly have another look?

dimitar-asenov (approved):

PR #40722, opened 2026-04-10
[XLA:GPU] Migrate Memset32Cmd into Memset32BitValueThunk

shawnwang18:
Eliminates the separate `Memset32Cmd` class by making `Memset32BitValueThunk`
  implement the `Command` interface directly (`MemcpyDeviceToDeviceCmd` → `DeviceToDeviceCopyThunk`).

  ### Changes

  - **`Memset32BitValueThunk`** now inherits from `Command` (via the protected
    dual-inheritance constructor), adding a `Record()` method for direct
    command buffer recording.
  - **`Memset32Cmd`** deleted from `command_buffer_cmd.h/.cc`.
  - **Emitter** (`command_buffer_cmd_emitter.cc`): the `kMemset32BitValue` case
    now borrows a raw pointer to the thunk (same pattern as `kReplicaId` /
    `kPartitionId`) instead of constructing an owned `Memset32Cmd` wrapper.
  - **Tests**: updated `command_buffer_thunk_test.cc` and
    `command_buffer_cmd_test.cc` to use the thunk directly; added
    `RecordCommandBuffer` and `RecordCommandBufferUpdate` GPU tests to
    `memset_thunk_test.cc`.

  ### Design notes

  - `ExecuteOnStream` calls `stream->Memset32(dest, value, byte_size)` while
    `Record` calls `command_buffer->CreateMemset(dest, value, num_elements)`.
    The APIs differ enough that no shared helper is warranted.
  - The 0-byte early-return guard from `Memset32Cmd::Record` is preserved in
    the new `Record()` implementation.
  - Profiling tests previously used different fill values to distinguish the
    thunk path from the command path; after this change the thunk *is* the
    command, so both paths use the same value.


xla/backends/gpu/runtime/memset_thunk_test.cc
```diff
@@ -33,6 +47,17 @@ namespace {
...
 using ::tsl::proto_testing::ParseTextProtoOrDie;
 
+static se::StreamExecutor* GpuExecutor() {
+  auto name =
+      absl::AsciiStrToUpper(PlatformUtil::CanonicalPlatformName("gpu").value());
+  auto* platform = se::PlatformManager::PlatformWithName(name).value();
+  return platform->ExecutorForDevice(0).value();
+}
+
+// ===========================================================================
```
ezhulenev: ditto, delimiters

ezhulenev (approved):

xla/backends/gpu/runtime/memset_thunk_test.cc
shawnwang18: fixed.

akuegel (approved):

PR #40738, opened 2026-04-10
[xla:gpu] Optimize hot path for requesting GpuCliques

ezhulenev:
I found that for large model with thousands of collectives we spend very significant time in requesting collective cliques and delay launching kernels on device. Optimize hot path by caching replica groups and cliques in collective thunk.

Avoid expensive copies of `CollectiveDeviceList device_list` (for large device groups it can be significant), and replace linear search with hash map lookup.


akuegel (approved):

ezhulenev:
Rebased + fixed conflicts. Needs re-approval.

pifon2a (approved):

ermilovmaxim (approved):

PR #40750, opened 2026-04-11
Add overflow checks in shape inference dimension computations

mohammadmseet-hue:
## Summary
- Add `OverflowSafeMultiply` checks to four locations where unchecked `int64` multiplication can silently wrap on crafted inputs:
  - `InferPadShape`: `(dim-1) * interior_padding`
  - `DilatedBound`: `(bound-1) * dilation`
  - `InferAllGatherShape`: `shard_count * dimension`
  - `InferAllToAllShape`: `dimension * split_count`
- Returns `InvalidArgument` (or `LOG(FATAL)` for `DilatedBound`) on overflow instead of producing a corrupted shape.
- Adds `//xla:util` dep to `window_util` BUILD target for `overflow_util.h`.

## Test plan
- Existing shape inference tests continue to pass (no behavioral change for valid inputs).
- Overflow paths trigger on dimensions near `INT64_MAX`, which are not reachable in normal compilation.


akuegel (approved):

xla/service/shape_inference.cc
```diff
@@ -2726,10 +2736,15 @@ ShapeInference::InferScalarBroadcastShape(absl::Span<const Shape> shapes) {
       IsUnboundedDynamicSize(new_dimensions[split_dimension])
           ? Shape::kUnboundedSize
           : new_dimensions[split_dimension] / split_count;
+  auto [ata_product, ata_overflow] =
```
akuegel: This causes test failures for tests with dynamic dimensions. In that case, you could have new_dimensions[concat_dimension] be negative, in which case we would run into an error.
You should only call OverflowSafeMultiply if both values are positive.

akuegel (changes requested):

xla/service/shape_inference.cc
mohammadmseet-hue: Thanks! You're right — `kUnboundedSize` is `int64_t::min()`, so `OverflowSafeMultiply` correctly reports overflow for what's actually a valid dynamic dimension sentinel.

I've guarded the overflow checks to only run when dimensions are non-negative. This fixes both the AllToAll case you flagged and the AllGather case (same pattern). The padding case was already inside the `!is_unbounded_dynamic_dimension` else-branch so it's fine.

Pushed the fix.

akuegel (approved):

akuegel:
@mohammadmseet-hue can you please rebase? We have hit a problem importing this PR and I think a rebase will fix it.

akuegel:
@mohammadmseet-hue friendly ping? Unfortunately we cannot merge this without another rebase from your side. Sorry for the inconvenience.

mohammadmseet-hue:
Rebased onto current main (c535b12432). Two commits, clean cherry-pick, no conflicts in the changed files. Ready for re-import, @akuegel.

PR #40751, opened 2026-04-11
[xla:gpu] Delete special case for calls to command buffers

ezhulenev:
This is leftovers from the era when calls were ONLY command buffers. Today we use a command buffer pass that uses thunks, and calls can be anything!


akuegel (approved):

ezhulenev:
Rebased + fixed.

pifon2a (approved):

PR #40761, opened 2026-04-12
[xla:gpu] One-sided collective permute for XLA:GPU

ezhulenev:
Add a collectives execution mode that tells how to execute collective operation. For now it's configured globally for collective operation type, but longer term it should be controlled by a frontend attribute for each collective instance.

1. Implemented one-sided collective permute using latest NCCL APIs.
2. Migrated P2P collective permute to collective memory APIs
3. Added test coverage for all 3 supported modes.
4. Extend collectives backend assigner pass to assign a backends + mode for all collective operations.
5. Move collectives backends assigner pass after all collective optimizations to correctly assign backend to already combined collectives.

```
  // Memory mode for XLA:GPU collective operations.
  //
  // COLLECTIVES_PRIVATE_MEMORY:
  //   Collective operations working on private device memory that lives in a
  //   per-device memory space. This is also called two-sided collectives
  //   because it requires both sides of the collective operation to exchange
  //   local device pointers before they can make progress.
  //
  // COLLECTIVES_SYMMETRIC_MEMORY:
  //   Collective operations working on symmetric device memory, which gives
  //   a global view of device memory across all participating ranks. With
  //   symmetric memory, collective operations can be implemented as one-sided
  //   operations where data can be written directly to a remote buffer, or as
  //   device-initiated collectives that run as device kernels and do not
  //   require any device-to-host synchronization.
  //
  // COLLECTIVES_PEER_MEMORY:
  //   Collective operations implemented directly on top of peer memory access,
  //   either by reading/writing peer memory directly or by mapping all
  //   participating buffers to the multimem. This is typically supported only
  //   within a single process, where all ranks can use rendezvous to exchange
  //   device buffers.
  //
  enum CollectivesMode {
    COLLECTIVES_PRIVATE_MEMORY = 0;
    COLLECTIVES_SYMMETRIC_MEMORY = 1;
    COLLECTIVES_PEER_MEMORY = 3;
  }
```


akuegel:
@PatriosTheGreat can you please review this PR?

xla/backends/gpu/transforms/collectives/BUILD
```diff
@@ -86,7 +88,6 @@ xla_cc_test(
         "@com_google_absl//absl/status:status_matchers",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings:string_view",
-        "@com_google_googletest//:gtest",
```
PatriosTheGreat, line 89: Could you revert this line? Otherwise I'm getting build errors

xla/xla.proto
```diff
@@ -338,6 +338,39 @@ message DebugOptions {
   // go/keep-sorted start newline_separated=yes skip_lines=2 ignore_prefixes=["optional AutotuneCacheMode","optional bool","optional float","optional int32","optional int64","optional LibNvJitLinkMode","map<string, string>","optional PGLEStrictnessLevel","optional PipelineParallelismOptLevel","repeated CollectiveOpType","repeated CommandBufferCmdType","repeated string","optional ShapeChecks","optional string","optional WhileLoopUnrolling","repeated GenericTritonEmitterFeature","optional CommandBufferSchedulingMode", "repeated AutotuneBackend", "optional CommandBufferUpdateMode"] // NOLINT
   // clang-format on
 
+  // Execution mode for XLA:GPU collective operations.
+  //
+  // COLLECTIVES_HOST_INITIATED:
```
PatriosTheGreat: Host initiated collectives might also require symmetric memory. For example NCCL symmetric kernels does this. Currently there is a custom logic which registers buffers for these Thunks with NCCL: https://github.com/openxla/xla/blob/main/xla/backends/gpu/runtime/collective_thunk.cc#L250.
I guess we need to migrate them to your createsymmetricmemory API. I'll do this later.

My question here -- should we consider NCCL symmetric kernels here as a device initiated collective or host initiated? Or maybe even peer access one

xla/service/gpu/gpu_executable.cc
```diff
@@ -114,6 +114,7 @@ limitations under the License.
 #include "xla/stream_executor/device_description.h"
 #include "xla/stream_executor/event_based_timer.h"
 #include "xla/stream_executor/kernel_stats.h"
+#include "xla/stream_executor/memory_allocation.h"
```
PatriosTheGreat: Could you add a corresponding dependency to BUILD file?

xla/service/gpu/gpu_executable.cc
```diff
@@ -125,6 +126,7 @@ limitations under the License.
 #include "xla/tsl/platform/env.h"
 #include "xla/tsl/platform/env_time.h"
 #include "xla/tsl/platform/logging.h"
+#include "xla/tsl/platform/status_macros.h"
```
PatriosTheGreat: Here also. Need dependency in BUILD file.

xla/service/gpu/gpu_executable.cc
```diff
@@ -802,31 +806,26 @@ absl::Status ExecuteThunksImpl(const DebugOptions* debug_options,
 }
 
 namespace {
-// Wrap RunId into a unique struct to guarantee we do not accidentally try to
-// run multiple unrelated rendezvous for a same key.
-struct InitializationKey {
+// Wrap RunId and stage name into a unique struct to guarantee we do not
+// accidentally try to run multiple unrelated rendezvous for a same key.
+struct ExecuteThunksRendezvousKey {
```
PatriosTheGreat: Is this refactoring tied to the current change?
Can this be done as a separate PR?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -2164,6 +2164,16 @@ StreamExecutorGpuClient::RunAsync(
   absl::Span<const BufferAllocation* const> allocations =
       gpu_exec->GetAllocations();
 
+  // Build a map of per-color allocation granularity. Collective memory requires
+  // larger alignment than the BFC allocator guarantees (256 bytes).
+  absl::flat_hash_map<LogicalBuffer::Color, int64_t> allocate_granularity;
+  if (auto* collectives =
```
PatriosTheGreat, line 2154: This change also looks like a standalone change. If you could send this as a separate PR it would simplify submition/potential reverting.

PatriosTheGreat (changes requested):

xla/service/gpu/gpu_executable.cc
ezhulenev: Done

xla/service/gpu/gpu_executable.cc
ezhulenev: Removed from this one.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
ezhulenev: I need it for testing this PR e2e, but go ahead and bundle with your changes if you need to.

xla/xla.proto
ezhulenev: `RegisterBufferOnce` is a legacy thing that has to be cleaned up, the new modern way is to request symmetric memory explicitly when needed.

Also as discussed offline, updated this enum to focus on memory kinds and not host/device initiated: (1) private memory (2) symmetric memory (3) peer-memory (special case of private memory in same process)

ezhulenev:
@PatriosTheGreat pushed a new version with new enum + rebased to latest HEAD and resolved conflicts.

Tixxx:
I have a general comment on the collective memory interface. Nccl is trying to abstract the different memory exposures under window handle and communicator, while the CollectiveMemory interface seems to reverse that abstraction. The multimemMemory can be retrieved using ncclGetMultimemPointer while the peer memory seems to be arbitrary cuda allocations with p2p access enabled. Are you planning to consolidate these different memory types under a symmetric memory and remove the registered handle mechanism? My understanding is that for collective operations, we have 2 two types:
1. arbitrary cuda allocation, p2p would be okay with a p2pAccess check
2. symmetric memory which assumes multicast registration, this can be used for both device and host initiated communications, this is only used by collectives

type 1 is what bfc does by default and for operations that need direct peer access, they perform the p2p check.
type 2 is currently exposed through the handle mechanism.
The two types can be hidden under a common interface and pass this around to custom kernels or xla's communicator calls. XLA's communicator should decide which api to call based on how the memory is allocated. The thunks have become fairly complex with all the different backend impl details. 
btw the integration of host put api was supposed to replace the memcpy impl as memcpy only works under a single process.

ezhulenev:
Peer memory == we don't use any fancy libraries and just resolve peer pointers using local rendezvous, kind of a special case, which I think can be removed assuming nccl + symmetric memory works well.

Yes, I want to consolidate everything under symmetric memory API, and remove the old handler registration.

Regarding `multimemMemory`, I also think it should be possible to delete after porting to latest NCCL APIs. In the end I think it would be nice to see just PRIVATE_MEMORY for two-sided communication and SYMMETRIC_MEMORY for all new fancy APIs, like ncclDevComm + Put/Signal. NCCL should hide the details from us, multimem is a super low level concept that leaked all the way up to thunks.

`The thunks have become fairly complex with all the different backend impl details.` +1, I hope we can get the number of implementations down to 2.

ezhulenev:
@Tixxx is nccl window always a multimem behind the scene? Can we drop the special case of multimem construction then, and always rely on NCCL to do it for us automatically? That would significantly simplify thunks/runtime and StreamExecutor.

Tixxx:
> @Tixxx is nccl window always a multimem behind the scene? Can we drop the special case of multimem construction then, and always rely on NCCL to do it for us automatically? That would significantly simplify thunks/runtime and StreamExecutor.

Yes, that was also my point above. ncclMemAlloc makes sure that the memory is allocated based on multimem granularity requirement and the registration makes sure the memory is bound to an multicast object that can be used with multimem instructions. Symmetric memory is just a nice VA on top to make sure we dont need to deal with importing and exporting IPC handles across processes. Frankly i didnt find how xla handles mc memory currently when doing 1 process per gpu as the memory needs to be exported using ipc handle. The current code seems to work under 1 process multi-device.

ezhulenev:
It doesn't 🙂 what happens when symmetric memory is across hosts? Multimem handle will be null or it will cover local members only?

Tixxx:
> It doesn't 🙂 what happens when symmetric memory is across hosts? Multimem handle will be null or it will cover local members only?

Depends on if it's multinode nvlink or not. If yes, it will cover all devices within the nvlink domain. If not, then it will only be local devices. Cross node comm will be handled either by proxy host thread or GIN.

xla/backends/gpu/runtime/collective_memory_cache.cc
```diff
@@ -18,59 +18,76 @@ limitations under the License.
...
   absl::MutexLock lock(mutex_);
-  multicast_memories_.erase(
-      std::remove_if(
-          multicast_memories_.begin(), multicast_memories_.end(),
-          [](tsl::TiedRef<stream_executor::gpu::MulticastMemory>&
-                 multicast_memory) { return multicast_memory.Expired(); }),
-      multicast_memories_.end());
-  multicast_memories_.push_back(std::move(multicast_memory));
+  auto it = sym_memories_.find(Key{clique, addr});
+  if (it == sym_memories_.end()) return nullptr;
```
PatriosTheGreat: Formatter complain:
if (it == sym_memories_.end()) {
   return nullptr;
}

xla/backends/gpu/runtime/collective_memory_cache.cc
```diff
@@ -18,59 +18,76 @@ limitations under the License.
...
+      Key{clique, addr}, std::move(multicast_memory));
+  return it->second.Lock();
+}
+
+std::shared_ptr<stream_executor::gpu::MulticastMemory>
+CollectiveMemoryCache::FindMulticastMemory(
+    const GpuCliqueKey& clique, stream_executor::DeviceAddressBase addr) {
+  absl::MutexLock lock(mutex_);
+  auto it = multicast_memories_.find(Key{clique, addr});
+  if (it == multicast_memories_.end()) return nullptr;
```
PatriosTheGreat: Also formatter complain:

if (it == multicast_memories_.end()) { 
   return nullptr;
}

xla/backends/gpu/runtime/collective_permute_thunk.cc
```diff
@@ -38,32 +38,33 @@ limitations under the License.
...
 #include "xla/future.h"
 #include "xla/hlo/ir/collective_op_group_mode.h"
 #include "xla/hlo/ir/hlo_instruction.h"
 #include "xla/hlo/ir/hlo_instructions.h"
+#include "xla/runtime/buffer_use.h"
 #include "xla/runtime/device_id.h"
 #include "xla/service/buffer_assignment.h"
 #include "xla/service/computation_placer.h"
-#include "xla/service/gpu/backend_configs.pb.h"
+#include "xla/service/gpu/buffer_allocations.h"
```
PatriosTheGreat: Is this used?

xla/backends/gpu/transforms/collectives/collective_backend_assigner.cc
```diff
@@ -72,55 +78,92 @@ absl::StatusOr<GPUCommunicationType> GetCommunicationType(
...
+    HloModule* module, int num_visible_devices_per_process,
+    const se::GpuComputeCapability& gpu_version, int64_t threshold_in_bytes,
+    int64_t slice_size) {
   bool changed = false;
   for (HloComputation* comp : module->computations()) {
     for (HloInstruction* instr : comp->instructions()) {
-      if (!IsCollectiveOp(instr)) {
-        continue;
-      }
+      if (!IsCollectiveOp(instr)) continue;
```
PatriosTheGreat: Also auto formatter complains, same as above

xla/backends/gpu/transforms/collectives/collective_backend_assigner.cc
```diff
@@ -72,55 +78,92 @@ absl::StatusOr<GPUCommunicationType> GetCommunicationType(
...
-      if (!use_nvshmem) {
-        continue;
-      }
-
-      TF_ASSIGN_OR_RETURN(GpuBackendConfig gpu_config,
-                          instr->backend_config<GpuBackendConfig>());
+           (slice_size > 0 &&
+            IsIntraNVLinkDomain(module->config(), slice_size))) &&
+          (!IsAllReduceOp(instr) || shape_size < threshold_in_bytes);
+      if (!use_nvshmem) continue;
```
PatriosTheGreat: Same as above

xla/backends/gpu/transforms/collectives/collective_backend_assigner.cc
```diff
@@ -72,55 +78,92 @@ absl::StatusOr<GPUCommunicationType> GetCommunicationType(
...
   }
   return changed;
 }
 
+// Assigns collectives_mode for collective-permute operations based on the
+// xla_gpu_collective_permute_mode debug option.
+absl::StatusOr<bool> AssignCollectivePermuteMode(HloModule* module) {
+  const auto mode =
+      module->config().debug_options().xla_gpu_collective_permute_mode();
+  if (mode == DebugOptions::COLLECTIVES_PRIVATE_MEMORY) return false;
```
PatriosTheGreat: Same as above

xla/backends/gpu/transforms/collectives/collective_backend_assigner.cc
```diff
@@ -72,55 +78,92 @@ absl::StatusOr<GPUCommunicationType> GetCommunicationType(
...
+
+  bool changed = false;
+  for (HloComputation* comp : module->computations()) {
+    for (HloInstruction* instr : comp->instructions()) {
+      if (!IsCollectivePermuteOp(instr)) continue;
```
PatriosTheGreat: Same as above

xla/xla.proto
```diff
@@ -338,6 +338,35 @@ message DebugOptions {
...
+  //   Collective operations implemented directly on top of peer memory access,
+  //   either by reading/writing peer memory directly or by mapping all
+  //   participating buffers to the multimem. This is typically supported only
+  //   within a single process, where all ranks can use rendezvous to exchange
+  //   device buffers.
+  //
+  enum CollectivesMode {
+    COLLECTIVES_PRIVATE_MEMORY = 0;
+    COLLECTIVES_SYMMETRIC_MEMORY = 1;
+    COLLECTIVES_PEER_MEMORY = 3;
```
PatriosTheGreat: = 2 I guess

xla/service/hlo_buffer.h
```diff
@@ -25,6 +25,7 @@ limitations under the License.
 #include "absl/log/check.h"
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
+#include "absl/strings/str_format.h"
```
PatriosTheGreat, line 28: The dependency should be added to BUILD file

xla/hlo/transforms/collectives/BUILD
```diff
@@ -455,7 +455,7 @@ cc_library(
         "@com_google_absl//absl/log",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/status:statusor",
-        "@com_google_absl//absl/strings",
+        "@com_google_absl//absl/strings:string_view",
```
PatriosTheGreat, line 458: Is this change needed?

xla/backends/gpu/transforms/collectives/collective_backend_assigner_test.cc
```diff
@@ -246,6 +259,99 @@ TEST_F(CollectiveBackendAssignerTest, NonIntraNvlinkDomainUsesDefault) {
...
+    HloModule m
+
+    add {
+      lhs = f32[] parameter(0)
+      rhs = f32[] parameter(1)
+      ROOT add = f32[] add(lhs, rhs)
+    }
+
+    ENTRY main {
+      p0 = f32[1024,1024] parameter(0)
```
PatriosTheGreat: Can we use smaller shapes here like 8, 8? Zeroing input buffers on a device increases test-case time quite significantly

xla/service/gpu/buffer_allocations.cc
```diff
@@ -89,7 +89,7 @@ std::optional<BufferAllocation::Index> BufferAllocations::FindAllocationIndex(
   for (BufferAllocation::Index i = 0; i < buffers_.size(); ++i) {
     auto* buf = static_cast<char*>(buffers_[i].opaque());
     auto* ptr = static_cast<char*>(addr.opaque());
-    if (ptr >= buf && ptr <= buf + buffers_[i].size()) {
+    if (ptr >= buf && ptr < buf + buffers_[i].size()) {
```
PatriosTheGreat, line 92: Is it possible to add a unit test for this? IIUC BufferAllocations class can be easily unit-tested standalone

PatriosTheGreat (changes requested):

xla/backends/gpu/runtime/collective_memory_cache.cc
ezhulenev: Done

xla/backends/gpu/runtime/collective_memory_cache.cc
ezhulenev: Done

xla/backends/gpu/transforms/collectives/collective_backend_assigner.cc
ezhulenev: Done

xla/backends/gpu/transforms/collectives/collective_backend_assigner.cc
ezhulenev: Done

xla/xla.proto
ezhulenev: Done

xla/service/hlo_buffer.h
ezhulenev: Done

xla/backends/gpu/transforms/collectives/collective_backend_assigner_test.cc
ezhulenev: Done

xla/service/gpu/buffer_allocations.cc
ezhulenev: Added a test

xla/hlo/transforms/collectives/BUILD
ezhulenev: Nope, str_format is needed though. Fixed.

PatriosTheGreat (approved):

ermilovmaxim (approved):

ermilovmaxim (approved):

ezhulenev:
@akuegel can I please get this one after rebase as well

akuegel:
> @akuegel can I please get this one after rebase as well

I think @derdrdirk has a cloned version of this PR already that he tries to land. We had to adjust some dependencies where build_cleaner complained, and a proto best practices warning. So I think it makes sense to get this adjusted version landed.

ezhulenev:
👍  SG, I just rebased, no new changes here

akuegel:
@ezhulenev We consistently hit a failing test in JAX: https://github.com/jax-ml/jax/blob/main/tests/pallas/gpu_pallas_distributed_test.py

Can you please check whether you can reproduce the error? We get `INTERNAL: CUDA error: : CUDA_ERROR_INVALID_VALUE: invalid argument`

ezhulenev:
@akuegel @PatriosTheGreat I cherry picked a fix for gpu memory space assigment + also bumped the alignemnt requirement, and now all tests pass. The alignment requirement will be revisited in my next PR that migrates to single BFC allocator for both memory spaces.

PatriosTheGreat (approved):

apivovarov:
This PR is quite large. I recommend to split it into several smaller PRs

1. Infrastructure and proto
- changes to proto files
- Add xla_gpu_collective_permute_mode flag
- Add collectives_mode_ and helper methods to CollectiveThunk class

2. Memory Management/Caching
- CollectiveMemoryCache refactoring
- CollectiveMemory API - FindPeerAddress, FindSymmetricMemory, etc
- new checks in buffer_allocations.cc

3. Compiler Passes
- CollectiveBackendAssigner logic
- GpuMemorySpaceAssignment logic
- reorder in GpuCompiler pipeline

4. One-Sided Implementation
- RunOneSidedPermute logic and NcclCollectives checks.
- Refactor P2P memcpy to use CollectiveMemory API
- End-to-End tests / roundtrip tests

PR #40826, opened 2026-04-13
Add PJRT_Buffer_CopyRawFromHost API for asynchronous data transfer from host

neudinger:
## 📝 Summary of Changes

Add `PJRT_Buffer_CopyRawFromHost` — a first-class C API entry point for asynchronously copying raw host data into an existing on-device `PJRT_Buffer`, complementing the existing `PJRT_Buffer_CopyRawToHost` (device→host).

**Files changed (10 files, +159/−2):**

| Layer | Files | Change |
|-------|-------|--------|
| C API | pjrt_c_api.h | New `PJRT_Buffer_CopyRawFromHost_Args` struct, typedef, `PJRT_Api` field; bump `PJRT_API_MINOR` 103→104 |
| C++ interface | pjrt_client.h | Add `virtual Future<> CopyRawFromHost()` to `PjRtBuffer` with default `UNIMPLEMENTED` |
| Backend impl | `xla/pjrt/common_pjrt_client.{h,cc}` | Override in `CommonPjRtBufferImpl` — acquires scoped raw buffer, validates bounds, delegates to `CopyRawHostToDeviceAndReturnEvent()` (works for both CPU and GPU backends) |
| C wrapper | `xla/pjrt/c/pjrt_c_api_wrapper_impl.{h,cc}` | Implement `PJRT_Buffer_CopyRawFromHost` wrapper, wire into default `PJRT_Api` table |
| C API client | `xla/pjrt/c_api_client/pjrt_c_api_client.{h,cc}` | `PjRtCApiBuffer::CopyRawFromHost` with version gate (requires ≥0.104) |
| TF wrapper | tf_pjrt_client.h | Forward delegation in `TfPjRtBuffer` |
| ABI tests | pjrt_c_api_test.cc | Add field to struct size computation and offset verification tests |

---

## 🎯 Justification

Currently, updating an existing device buffer's contents from host requires going through the `PJRT_RawBuffer` extension API (`PJRT_RawBuffer_CopyRawHostToDevice`), which requires first creating a raw alias via `PJRT_RawBuffer_CreateRawAliasOfBuffer`. This is:

1. **Inconvenient** — two API calls instead of one for the common "overwrite buffer with new host data" pattern
2. **Asymmetric** — `PJRT_Buffer_CopyRawToHost` (device→host) exists as a first-class API, but the reverse direction requires the extension
3. **Extension-only** — the RawBuffer extension is explicitly marked experimental with no ABI stability guarantees

Workloads that benefit:
- **Model weight loading / checkpointing** — directly update device buffers from host without re-allocating
- **Online learning / fine-tuning** — push updated parameters to device buffers in-place
- **Custom PJRT plugin authors** — symmetric API that doesn't require implementing the experimental RawBuffer extension

---

## 🚀 Kind of Contribution

✨ New Feature

---

## 🧪 Unit Tests

The ABI layout test in pjrt_c_api_test.cc (`PjrtCAbiTestBase.FieldOffsetsAndSizes`) is updated to verify the new `PJRT_Buffer_CopyRawFromHost` field:
- **Struct size computation**: Added version-gated entry for `minor_version >= 104`
- **Field offset/size verification**: Added `offsetof`/`sizeof` entry ensuring stable ABI layout

**Additional tests needed before merge** (not yet written):

```
# C API level round-trip test (create buffer → CopyRawFromHost → CopyRawToHost → verify)
bazel test //xla/pjrt/c:pjrt_c_api_cpu_test --test_filter="*CopyRawFromHost*"

# CPU backend test
bazel test //xla/pjrt/cpu:pjrt_client_test_cpu --test_filter="*CopyRawFromHost*"
```

Suggested test case: Create a buffer with known data, call `CopyRawFromHost` to overwrite it, read back with `CopyRawToHost`, and assert the data matches the new values. Edge cases: partial writes with offset > 0, out-of-bounds validation.

### DIFF
--- a/warnings.bazelrc
+++ b/warnings.bazelrc
@@ -5,10 +5,12 @@
 
 # Treat warnings as errors...
 build:warnings --copt=-Werror --host_copt=-Werror
+build:warnings --host_copt=-Wno-deprecated-declarations
 # ...and silence them outside of the workspace.
 build:warnings --per_file_copt=external/.*@-w
-# ...and silence them on host builds.
+# ...and silence them on host builds outside the workspace.
 build:warnings --host_per_file_copt=external/.*@-w
+build:warnings --host_per_file_copt=xla/tsl/.*@-w
 
 build:warnings --copt=-Wall
 build:warnings --copt=-Werror
--- a/xla/pjrt/c/CHANGELOG.md
+++ b/xla/pjrt/c/CHANGELOG.md
@@ -1,5 +1,12 @@
 # PJRT C API changelog
 
+## 0.105
+
+* Add `PJRT_Buffer_CopyRawFromHost` for asynchronous raw data transfer from a
+  host buffer into a device buffer. The API mirrors `PJRT_Buffer_CopyRawToHost`:
+  callers supply a source pointer, byte offset, and transfer size; an event is
+  returned that becomes ready when the transfer completes.
+
 ## 0.104
 
 * Added `total_allocation_bytes`, `indefinite_allocations` and `peak_unpadded_heap_bytes` to GetCompiledMemoryStats
--- a/xla/pjrt/c/pjrt_c_api.h
+++ b/xla/pjrt/c/pjrt_c_api.h
@@ -111,7 +111,7 @@ PJRT_DEFINE_STRUCT_TRAITS(PJRT_Extension_Base, next);
 // Changes include:
 // * Adding a new field to the PJRT_Api or argument structs
 // * Renaming a method or argument (doesn't affect ABI)
-#define PJRT_API_MINOR 104
+#define PJRT_API_MINOR 105
 
 // The plugin should set the major_version and minor_version of
 // PJRT_Api.pjrt_api_version to be the `PJRT_API_MAJOR` and `PJRT_API_MINOR` in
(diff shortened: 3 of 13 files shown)


jparkerh (changes requested):
Hi Kevin - can we have some more info on the application here? Maybe a test case at the c_api_client? I want to make sure that we're enforcing the correct memory semantics with this API, I think this API as is has the possibility to introduce bad behavior (we're not intended to mutate the on device buffer without a donation hold)

neudinger:
> Hi Kevin - can we have some more info on the application here? Maybe a test case at the c_api_client? I want to make sure that we're enforcing the correct memory semantics with this API, I think this API as is has the possibility to introduce bad behavior (we're not intended to mutate the on device buffer without a donation hold)

Hi @jparkerh, thanks for the review!

I am trying to avoid memory allocation overhead on the critical path when streaming data directly into a preallocated PJRT_Buffer during pipeline execution.

Is this kind of zero-allocation copy already possible with the current API ?
I saw "PJRT_RawBuffer_CreateRawAliasOfBuffer" and "PJRT_RawBuffer_CopyRawHostToDevice" but it seems more dangerous to use it than if a "PJRT_Buffer_CopyRawFromHost" exist.

what specific safety checks or state validations can I add to this PR to enforce the rules and make it safe ?

neudinger:
I will work more on it to enhance the safety.
And avoid the possible bad behavior you mentioned.

dimitar-asenov:
@neudinger From the comments it seems that this PR is still work in progress. I've marked it as a draft. Once you've updated it, please mark it as ready for review again. Thanks!

PR #40838, opened 2026-04-13
Canonicalize custom element sizes on tuple leaf arrays in Literal::SetShape

Omar-Khaleel:
This change recursively canonicalizes custom `layout.element_size_in_bits` values on tuple-contained array leaves in `Literal::SetShape`.

Previously, root array shapes had their custom element sizes cleared, but tuple-contained leaf arrays could retain non-zero custom element sizes. This allowed an inconsistent state where literal buffer sizing honored the custom bit width while later typed materialization assumed a dense native representation.

By applying the canonicalization consistently across all leaf arrays, this change restores the invariant expected by literal allocation and typed population paths.

### DIFF
--- a/xla/literal.cc
+++ b/xla/literal.cc
@@ -254,19 +254,21 @@ void Literal::SetShape(const Shape& shape) {
     shape_ = intered_shape_ptr;
     return;
   }
+
   auto owning_shape_ptr = std::make_unique<Shape>(shape);
-  if (!LayoutUtil::HasLayout(*owning_shape_ptr)) {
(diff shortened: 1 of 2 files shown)


Omar-Khaleel:
Opened the corresponding tracking issue:

https://github.com/openxla/xla/issues/40839

akuegel (commented):

Omar-Khaleel:
Added a regression test covering tuple-contained array leaves with custom `layout.element_size_in_bits`, so the fix is now exercised directly in the literal shape canonicalization path.

akuegel (approved):

PR #40844, opened 2026-04-13
[XLA:simplifier] Skip f64 A/sqrt(B) => A*rsqrt(B) rewrite (correctly-rounded composition)

blasphemetheus:
## TL;DR

XLA's algebraic simplifier unconditionally rewrites `divide(A, sqrt(B))` to `multiply(A, rsqrt(B))`. For f64, this silently downgrades a program built from two correctly-rounded StableHLO primitives (`kSqrt` and `kDivide`) into one that contains `kRsqrt`, an implementation-defined-precision op. On CPU+avx512f the resulting emit lands on a different in-range f64 than `sqrt+fdiv` for 29.1% of regular-range f64 inputs, with a 2-ULP tail at 0.2%.

This PR adds an element-type guard to that single simplifier rewrite, skipping it for f64. f32 is unaffected — ML workloads explicitly want the fast rsqrt path there.

## Motivating report

The elixir-nx/nx community filed [elixir-nx/nx#1704](https://github.com/elixir-nx/nx/issues/1704) after a `1.0/sqrt(3.0)` doctest started returning `0.5773502691896257` instead of the IEEE round-to-nearest-even value `0.5773502691896258`:

```python
import jax, jax.numpy as jnp, numpy as np
x = jnp.float64(3.0)

# On CPU+avx512f, before this PR:
print(repr(jax.jit(lambda x: 1.0 / jnp.sqrt(x))(x).item()))  # 0.5773502691896257
print(repr(1.0 / np.sqrt(3.0)))                              # 0.5773502691896258
```

The JAX spelling `1.0 / jnp.sqrt(x)` lowers to `divide(constant(1), sqrt(x))` in StableHLO. The user's expectation — and the spec's guarantee for a composition of correctly-rounded ops — is that this stays a correctly-rounded-composition computation. The simplifier rewrite breaks that guarantee.

## Spec citation

From the [StableHLO spec](https://openxla.org/stablehlo/spec):

- Correctly-rounded bucket: `add`, `sub`, `mul`, `div`, **`sqrt`**.
- Implementation-defined precision bucket: **`rsqrt`**, `exp`, `log`, `tanh`, `erf`, `pow`, …

A program written as `divide(constant(1), sqrt(x))` is built entirely from the correctly-rounded bucket and carries the ≤1 ULP bound that comes from composing two correctly-rounded operations. The simplifier silently moves that program into the implementation-defined bucket. The StableHLO spec does not require the two buckets to produce the same bit pattern, and indeed on some emitters they don't.

## The fix

One element-type guard on the simplifier rewrite:

```diff
   // A/sqrt(B) => A*rsqrt(B).
+  //
+  // Skip f64: kSqrt and kDivide are in StableHLO's correctly-rounded bucket,
+  // but kRsqrt is in the implementation-defined-precision bucket and several
+  // emitters lower it to a fast approximation that is up to 1–2 ULP off the
+  // IEEE-correctly-rounded value. Rewriting silently moves the user's
+  // explicit 1/sqrt(x) into the approximate bucket they did not opt into.
+  // f32 is unaffected: ML workloads explicitly want the fast rsqrt path.
-  if (Match(divide, m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse()))) {
+  if (Match(divide,
+            m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse())
+                .WithPredicate([](const HloInstruction* inst) {
+                  return inst->shape().element_type() != F64;
+                }))) {
```

The inverse rewrite at the next block (`A/rsqrt(B) => A*sqrt(B)`) is intentionally left unchanged — it goes from the implementation-defined bucket to the correctly-rounded bucket, **improving** precision, so it should fire as widely as possible.

## What this PR is NOT

For an earlier revision of this PR I also modified `xla/codegen/intrinsic/rsqrt.cc` to unconditionally take the `llvm.sqrt + fdiv` fallback for f64 (dropping the `!avx512f` condition), so that explicit `rsqrt(x)` calls also benefited. Benchmarks on this branch (see `xla/backends/cpu/benchmarks/rsqrt_benchmark_test.cc`) show that change costs ~29% of the f64 rsqrt throughput on avx512f at N=4096. Given `kRsqrt` is spec'd as implementation-defined precision, and given the user-facing bug from #1704 is fully resolved by the simplifier guard alone, the intrinsic change is **not included in this PR** and is available as a separate follow-up if XLA project policy prefers the tighter emit. The benchmarks make the cost quantitative either way.

### Failing → passing end-to-end test

`xla/tests/divide_sqrt_f64_test.cc` (added in this PR):

```cpp
TEST_F(DivideBySqrtF64Test, MatchesIeeeCorrectlyRoundedComposition) {
  // HLO: divide(constant(1), sqrt(parameter(0)))
  // For each input, assert bit-exact equality with 1.0 / std::sqrt(x).
  for (double x : {2.0, 3.0, 0.5, 1.5, 7.0}) {
    TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(kModule));
    auto input = LiteralUtil::CreateR0<double>(x);
    TF_ASSERT_OK_AND_ASSIGN(auto result, Execute(std::move(module), {&input}));
    EXPECT_EQ(result, LiteralUtil::CreateR0<double>(1.0 / std::sqrt(x)));
  }
}
```

Bit-exact `EXPECT_EQ`, not within-tolerance: both paths are ≤1 ULP from the real-number ideal, but land on *different* in-range values, so any tolerance that accepts one accepts the other and hides the bug. Verified red on `main`, green with this PR.

### ULP-distance distribution

`DiagnosticSweepAgainstReference` in the same file sweeps 1000 log-uniform f64 inputs over `[1e-200, 1e200]` and prints a histogram. On an Intel/AMD avx512f host with the simplifier guard reverted (i.e., pre-PR behavior):

    0 ULP : 709 / 1000 (70.9%) — matches reference exactly
    1 ULP : 289 / 1000 (28.9%) — rewrite lands on the neighbouring f64
    2 ULP :   2 / 1000 (0.2%)  — Newton-Raphson refinement + reference
                                  rounding compound to the theoretical
                                  worst case

With this PR applied: 1000/1000 (100%) match reference exactly.

The 2-ULP tail is not clustered at any exponent range — the two specific inputs are at `x ≈ 2e-99` and `x ≈ 2.4e-13`, scattered across the domain.

### HLO-level structural regression test

`RecipSqrtF64NotRewritten` in `xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc` parses a minimal f64 `divide(p1, sqrt(p0))` module and asserts the root remains a `Divide(Parameter(1), Sqrt(Parameter(0)))` after simplification. Paired with the existing `RecipSqrt` f32 test, which continues to assert the f32 rewrite *does* fire, as an explicit regression guard against the f64 guard overshooting.

### Invariant guard against future rewrites of the same class

`DoesNotIntroduceApproximateOpsFromCorrectlyRoundedInputsF64` in the same file. Builds an f64 HLO module from only correctly-rounded primitives (`add`, `sub`, `mul`, `div`, `sqrt`) arranged in patterns the simplifier might try to rewrite, runs the simplifier, and fails if any implementation-defined-precision op appears in the result. Catches this specific class of "downgrade correctly-rounded inputs into an approximate op" rewrite, not just the `A/sqrt(B)` case.

### Benchmark (reference data)

`xla/backends/cpu/benchmarks/rsqrt_benchmark_test.cc` with `+avx,+avx512f,+avx512vl`:

| Size | Post-PR (sqrt+fdiv fallback, i.e. what f64 `rsqrt` would lower to if we also took the intrinsic change) | Pre-PR (vrsqrt+Newton on avx512f) |
|------|---|---|
| 128  | 993 ns | 1056 ns |
| 256  | 1280 ns | 1145 ns |
| 512  | 1865 ns | 1551 ns |
| 1024 | 2864 ns | 2233 ns |
| 4096 | 9048 ns | 6391 ns |

The SIMD+Newton path is 10–30% faster at larger batch sizes. Included as reference for evaluating the separable intrinsic-side follow-up; not the measurement target of this PR's fix.

## Tested

- [x] `bazel test //xla/tests:divide_sqrt_f64_test_cpu` — 2 tests pass; DiagnosticSweep reports 0 / 100 divergences with guard applied (reverted locally to 291 / 1000, 29.1%, on the pre-guard state).
- [x] `bazel test //xla/hlo/transforms/simplifiers:algebraic_simplifier_test` — all existing `*Rsqrt*` and `*Sqrt*` tests unchanged; added `RecipSqrtF64NotRewritten` and `DoesNotIntroduceApproximateOpsFromCorrectlyRoundedInputsF64` pass.
- [x] Red-direction verification: with the simplifier guard reverted, `RecipSqrtF64NotRewritten` fails (simplifier rewrites the f64 module) and `DivideBySqrtF64Test.MatchesIeeeCorrectlyRoundedComposition` fails for inputs 3.0, 1.5, 7.0 with the exact 1-ULP bit mismatch.
- [x] JAX reproducer at top of PR: returns `0.5773502691896258` after, `0.5773502691896257` before.

### DIFF
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
@@ -2448,7 +2448,15 @@ absl::Status AlgebraicSimplifierVisitor::HandleDivide(HloInstruction* divide) {
   }
 
   // A/sqrt(B) => A*rsqrt(B).
-  if (Match(divide, m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse()))) {
+  //
+  // Skip f64: kSqrt and kDivide are in StableHLO's correctly-rounded bucket,
+  // but kRsqrt is in the implementation-defined-precision bucket and several
+  // emitters lower it to a fast approximation that is up to 1 ULP off the
+  // IEEE-correctly-rounded value. Rewriting silently moves the user's
+  // explicit 1/sqrt(x) into the approximate bucket they did not opt into.
+  // f32 is unaffected: ML workloads explicitly want the fast rsqrt path.
+  if (Match(divide, m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse())) &&
+      divide->shape().element_type() != F64) {
     auto* rsqrt = divide->mutable_operand(1)->AddInstruction(
         HloInstruction::CreateUnary(divide->shape(), HloOpcode::kRsqrt, b));
     return ReplaceWithNewInstruction(
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
@@ -2455,8 +2455,11 @@ absl::Status AlgebraicSimplifierVisitor::HandleDivide(HloInstruction* divide) {
   // IEEE-correctly-rounded value. Rewriting silently moves the user's
   // explicit 1/sqrt(x) into the approximate bucket they did not opt into.
   // f32 is unaffected: ML workloads explicitly want the fast rsqrt path.
-  if (Match(divide, m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse())) &&
-      divide->shape().element_type() != F64) {
+  if (Match(divide,
+            m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse())
+                .WithPredicate([](const HloInstruction* inst) {
+                  return inst->shape().element_type() != F64;
+                }))) {
     auto* rsqrt = divide->mutable_operand(1)->AddInstruction(
         HloInstruction::CreateUnary(divide->shape(), HloOpcode::kRsqrt, b));
     return ReplaceWithNewInstruction(
--- a/xla/backends/cpu/benchmarks/BUILD
+++ b/xla/backends/cpu/benchmarks/BUILD
@@ -295,6 +295,28 @@ xla_cc_test(
 )
 
 xla_cc_test(
+    name = "rsqrt_benchmark_test",
+    srcs = ["rsqrt_benchmark_test.cc"],
+    fail_if_no_test_linked = False,  # NOLINT=This contains benchmarks only, no tests.
+    fail_if_no_test_selected = False,  # NOLINT=This contains benchmarks only, no tests.
+    deps = [
+        ":hlo_benchmark_runner",
+        ":multi_benchmark_config",
+        "//xla:literal",
+        "//xla:literal_util",
+        "//xla:shape_util",
+        "//xla:xla_data_proto_cc",
+        "//xla/service/cpu:cpu_aot_compilation_result",
+        "//xla/tsl/platform:logging",
+        "//xla/tsl/platform:test_benchmark",
+        "//xla/tsl/platform:test_main",
+        "@com_google_absl//absl/strings",
+        "@com_google_absl//absl/strings:string_view",
+        "@com_google_absl//absl/types:span",
+    ],
+)
+
+xla_cc_test(
     name = "fusion_benchmark_test",
     srcs = ["fusion_benchmark_test.cc"],
     fail_if_no_test_linked = False,  # NOLINT=This contains benchmarks only, no tests.
--- a/xla/backends/cpu/benchmarks/rsqrt_benchmark_test.cc
+++ b/xla/backends/cpu/benchmarks/rsqrt_benchmark_test.cc
@@ -1,0 +1,141 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+// Benchmarks for the CPU emit of `rsqrt`, supporting the performance
+// discussion on openxla/xla#40844. The f64 variants are the point of
+// interest: with the PR applied, f64 rsqrt lowers to llvm.sqrt + fdiv
+// regardless of AVX support; without the PR, it takes the SIMD
+// vrsqrt + Newton-Raphson refinement path when avx512f is available.
+// The delta between those two emit strategies is what the
+// BM_RsqrtF64Avx512 benchmark targets.
+
+#include <cstdint>
+#include <memory>
+#include <random>
+#include <vector>
+
+#include "absl/strings/str_cat.h"
+#include "absl/strings/string_view.h"
+#include "absl/types/span.h"
+#include "xla/backends/cpu/benchmarks/hlo_benchmark_runner.h"
+#include "xla/backends/cpu/benchmarks/multi_benchmark_config.h"
+#include "xla/literal.h"
+#include "xla/literal_util.h"
+#include "xla/service/cpu/cpu_aot_compilation_result.h"
+#include "xla/shape_util.h"
+#include "xla/tsl/platform/logging.h"
+#include "xla/tsl/platform/test_benchmark.h"
+#include "xla/xla_data.pb.h"
+
+namespace xla::cpu {
+
+static void BM_RsqrtF32(benchmark::State& state, HloBenchmarkOptions options) {
+  int64_t d0 = state.range(0);
+
+  absl::string_view hlo = R"(
+    HloModule rsqrt_f32_$d0
+
+    ENTRY e {
+      input = f32[$d0] parameter(0)
+      ROOT output = rsqrt(input)
+    }
+  )";
+
+  std::minstd_rand0 engine;
+
+  auto input_shape = ShapeUtil::MakeShape(F32, {d0});
+  auto p0 =
+      *LiteralUtil::CreateRandomLiteral<F32>(input_shape, &engine, 2.0f, 0.5f);
+  std::vector<const Literal*> args = {&p0};
+  CHECK_OK(
+      RunHloBenchmark(state, hlo, args, {{"$d0", absl::StrCat(d0)}}, options));
+}
+
+static void BM_RsqrtF64(benchmark::State& state, HloBenchmarkOptions options) {
+  int64_t d0 = state.range(0);
+
+  absl::string_view hlo = R"(
+    HloModule rsqrt_f64_$d0
+
+    ENTRY e {
+      input = f64[$d0] parameter(0)
+      ROOT output = rsqrt(input)
+    }
+  )";
+
+  std::minstd_rand0 engine;
+
+  auto input_shape = ShapeUtil::MakeShape(F64, {d0});
+  auto p0 =
+      *LiteralUtil::CreateRandomLiteral<F64>(input_shape, &engine, 2.0, 0.5);
+  std::vector<const Literal*> args = {&p0};
+  CHECK_OK(
+      RunHloBenchmark(state, hlo, args, {{"$d0", absl::StrCat(d0)}}, options));
+}
+
+// Forces AVX512F + vector-width 512 so the f64 rsqrt emit actually hits
+// the SIMD vrsqrt + Newton-Raphson path on pre-PR builds. With the PR
+// applied, the intrinsic falls back to llvm.sqrt + fdiv even in this
+// config — so the BM_RsqrtF64Avx512 number is the direct measurement
+// of the emit-strategy change.
+static void BM_RsqrtF64Avx512(benchmark::State& state) {
+  int64_t d0 = state.range(0);
+  HloBenchmarkOptions options;
+  options.aot_options = std::make_unique<CpuAotCompilationOptions>(
+      /*triple=*/"x86_64-unknown-linux-gnu", /*cpu_name=*/"skylake-avx512",
+      /*features=*/"+avx,+avx512f,+avx512vl",
+      /*entry_point_name=*/"rsqrt_f64",
+      /*relocation_model=*/CpuAotCompilationOptions::RelocationModel::Static);
+  options.aot_options->mutable_debug_options()->set_xla_cpu_prefer_vector_width(
+      512);
+
+  absl::string_view hlo = R"(
+    HloModule rsqrt_f64_$d0
+
+    ENTRY e {
+      input = f64[$d0] parameter(0)
+      ROOT output = rsqrt(input)
+    }
+  )";
+  std::minstd_rand0 engine;
+
+  auto input_shape = ShapeUtil::MakeShape(F64, {d0});
+  auto p0 =
+      *LiteralUtil::CreateRandomLiteral<F64>(input_shape, &engine, 2.0, 0.5);
+  std::vector<const Literal*> args = {&p0};
+  CHECK_OK(
+      RunHloBenchmark(state, hlo, args, {{"$d0", absl::StrCat(d0)}}, options));
+}
+
+#define REGISTER_RSQRT_BENCHMARK(NAME) \
+  XLA_CPU_BENCHMARK(NAME)              \
+      ->MeasureProcessCPUTime()        \
+      ->Arg(128)                       \
+      ->Arg(256)                       \
+      ->Arg(512)                       \
+      ->Arg(1024)                      \
+      ->Arg(4096);
+
+REGISTER_RSQRT_BENCHMARK(BM_RsqrtF32);
+REGISTER_RSQRT_BENCHMARK(BM_RsqrtF64);
+BENCHMARK(BM_RsqrtF64Avx512)
+    ->MeasureProcessCPUTime()
+    ->Arg(128)
+    ->Arg(256)
+    ->Arg(512)
+    ->Arg(1024)
+    ->Arg(4096);
+
+}  // namespace xla::cpu
--- a/xla/codegen/intrinsic/accuracy/accuracy_budget.h
+++ b/xla/codegen/intrinsic/accuracy/accuracy_budget.h
@@ -157,10 +157,10 @@ constexpr AccuracyBudget kErfF64Budget = {
 
 // Sqrt
 constexpr AccuracyBudget kSqrtF32Budget = {
-    /*cpu=*/{/*regular=*/1,
+    /*cpu=*/{/*regular=*/0,
              /*subnormal=*/1000000},
     /*gpu=*/
-    {/*regular=*/1,
+    {/*regular=*/0,
      /*subnormal=*/1000000},
 };
 
--- a/xla/codegen/intrinsic/accuracy/accuracy_budget.h
+++ b/xla/codegen/intrinsic/accuracy/accuracy_budget.h
@@ -105,7 +105,7 @@ constexpr AccuracyBudget kRsqrtF32Budget = {
 };
 
 constexpr AccuracyBudget kRsqrtF64Budget = {
-    /*cpu=*/{/*regular=*/0,
+    /*cpu=*/{/*regular=*/1,
              /*subnormal=*/1000000,
              /*special_values=*/4},
     /*gpu=*/
--- a/xla/codegen/intrinsic/accuracy/accuracy_budget.h
+++ b/xla/codegen/intrinsic/accuracy/accuracy_budget.h
@@ -105,7 +105,7 @@ constexpr AccuracyBudget kRsqrtF32Budget = {
 };
 
 constexpr AccuracyBudget kRsqrtF64Budget = {
-    /*cpu=*/{/*regular=*/1,
+    /*cpu=*/{/*regular=*/0,
              /*subnormal=*/1000000,
              /*special_values=*/4},
     /*gpu=*/
--- a/xla/codegen/intrinsic/accuracy/accuracy_budget.h
+++ b/xla/codegen/intrinsic/accuracy/accuracy_budget.h
@@ -157,10 +157,10 @@ constexpr AccuracyBudget kErfF64Budget = {
 
 // Sqrt
 constexpr AccuracyBudget kSqrtF32Budget = {
-    /*cpu=*/{/*regular=*/0,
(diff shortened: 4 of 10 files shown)


blasphemetheus:
### Audit result: this is the only precision-downgrading rewrite in the XLA simplifier

To verify this PR's fix is comprehensive rather than a spot-patch, I did an exhaustive sweep of `algebraic_simplifier.cc` for rewrites that take inputs built from StableHLO's correctly-rounded ops (`add`, `sub`, `mul`, `div`, `sqrt`) and produce an output containing an implementation-defined-precision op (`rsqrt`, `exp`, `log`, `tanh`, `erf`, `pow`, ...).

**Result:** only one such rewrite exists — the `divide(A, sqrt(B)) => multiply(A, rsqrt(B))` pattern this PR fixes. Every other transcendental rewrite in the simplifier is either neutral or a precision **improvement**:

| Rewrite | Direction |
|---|---|
| `A/sqrt(B) => A*rsqrt(B)` | **⬇ downgrade (f64)** ← this PR |
| `A/rsqrt(B) => A*sqrt(B)` | ⬆ improvement (rsqrt → sqrt) |
| `rsqrt(1/A) => sqrt(A)` | ⬆ improvement |
| `rsqrt(pow(A,-2)) => A` | ⬆ improvement (full reduction) |
| `pow(A, 0.5) => sqrt(A)` | ⬆ improvement (pow → sqrt) |
| `pow(A, -1) => 1/A` | ⬆ improvement |
| `pow(A, 2) => A*A` | ⬆ improvement |
| `ln(exp(A)) => A` | ⬆ improvement (full reduction) |
| `ln(sqrt(A)) => 0.5*ln(A)` | neutral (log on both sides) |
| `ln(rsqrt(A)) => -0.5*ln(A)` | ⬆ improvement (rsqrt removed) |
| `ln(pow(A,B)) => B*ln(|A|)` | neutral |
| `exp(A)/exp(B) => exp(A-B)` | neutral |
| `A/exp(B) => A*exp(-B)` | neutral |
| `exp(A)*exp(B) => exp(A+B)` | neutral |
| `sqrt(x)*sqrt(x) => x` | ⬆ improvement (double-rounding removed) |
| `rsqrt(x)*rsqrt(x) => 1/x` | ⬆ improvement (rsqrt → divide) |

I also grepped all of `xla/` for production files that both pattern-match transcendentals and produce them — the search returns exactly one file: `algebraic_simplifier.cc`. The only other file that introduces `rsqrt` is `batchnorm_expander.cc`, which legitimately lowers `stablehlo.batch_norm_*` (spec says implementation-defined precision).

**Consequence for review:** the `DoesNotIntroduceApproximateOpsFromCorrectlyRoundedInputsF64` invariant test is valuable as a **regression guard** against any new rewrite of this class being introduced in the future, not because there are other existing ones it might catch today. Just pushed a follow-up commit that broadens the test's input module with more patterns (`sqrt(A)/sqrt(B)`, nested divides, `p/(sqrt(A)*sqrt(A))`, scalar-broadcast constants, sqrt differences) so the invariant surfaces future offenders as widely as possible, and extracts the opcode classifier into a reusable static helper.

xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
```diff
@@ -2448,7 +2448,15 @@ absl::Status AlgebraicSimplifierVisitor::HandleDivide(HloInstruction* divide) {
...
   // A/sqrt(B) => A*rsqrt(B).
-  if (Match(divide, m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse()))) {
+  //
+  // Skip f64: kSqrt and kDivide are in StableHLO's correctly-rounded bucket,
+  // but kRsqrt is in the implementation-defined-precision bucket and several
+  // emitters lower it to a fast approximation that is up to 1 ULP off the
+  // IEEE-correctly-rounded value. Rewriting silently moves the user's
+  // explicit 1/sqrt(x) into the approximate bucket they did not opt into.
+  // f32 is unaffected: ML workloads explicitly want the fast rsqrt path.
+  if (Match(divide, m::Divide(m::Op(&a), m::Sqrt(m::Op(&b)).WithOneUse())) &&
```
akuegel, line 2458: nit: this can be rewritten with matcher code on the matched Divide:
`WithShape(m::Shape().WithElementType(PrimitiveType::F64))`

akuegel:
@blasphemetheus can you please do the minor adjustment that I suggested?

@seantalts can you please review this PR as well?

seantalts (changes requested):
Thanks @akuegel. 

Hi @blasphemetheus, thank you for your PR! I have a few questions about the context here to make sure I understand. Where in the function domain do you see that we are off by many (how many?) ULPs? In [accuracy_budget](https://github.com/openxla/xla/blob/main/xla/codegen/intrinsic/accuracy/accuracy_budget.h) we test that in f64 rsqrt for non-subnormals we are off by only 1 ULP. Are you talking about being off by many ULPs just for subnormals? Can you describe your use case for rsqrt for subnormals, and can you provide benchmark results for this change as well? Much appreciated!

blasphemetheus:
Thanks for the review! @seantalts. The "many ULPs" framing in the original PR body was imprecise — it traced back to the old subnormal budget slack and doesn't apply to the regular range. I've rewritten the PR body and also narrowed the PR scope based on your performance concern; details below, answering your four questions first.

## 1. Where in the domain are we off, and by how many ULPs?

Regular-range f64 inputs, on CPU+avx512f. The divergence between `divide(1, sqrt(x))` (what the user wrote) and `1.0 / std::sqrt(x)` (the correctly-rounded composition they expect) on the pre-PR emit is ≤2 ULP. Committed a diagnostic sweep in `xla/tests/divide_sqrt_f64_test.cc`; on a 1000-sample log-uniform sweep over `[1e-200, 1e200]`:

    0 ULP : 709 / 1000 (70.9%) — matches reference exactly
    1 ULP : 289 / 1000 (28.9%) — rewrite lands on the neighbouring f64
    2 ULP :   2 / 1000 (0.2%)  — Newton-Raphson refinement + reference
                                  rounding compound to the theoretical
                                  worst case

The 2-ULP tail is scattered across the domain (`x ≈ 2e-99` and `x ≈ 2.4e-13`), not clustered. With the PR applied, 1000/1000 match the reference exactly.

## 2. Are you talking about being off by many ULPs just for subnormals?

No — the motivating bug is regular-range ≤2 ULP, not subnormals. The PR body's original "many ULPs" language was tracing the old subnormal budget (1000000 ULPs slack) and shouldn't have been the headline framing. Apologies for the confusion. I've rewritten the PR body to lead with the regular-range sweep numbers.

## 3. Use case for f64 rsqrt on subnormals?

I don't have one, and I'm no longer claiming subnormals are the motivation. The subnormal budget is untouched by this PR.

## 4. Benchmark results

Committed `xla/backends/cpu/benchmarks/rsqrt_benchmark_test.cc`. AOT compile forcing `+avx,+avx512f,+avx512vl`, `--benchmark_min_time=0.5s`, on a local Intel/AMD avx512f host. The before/after measures the emit strategy of the *intrinsic* itself (pre-PR `vrsqrt + Newton-Raphson` vs the alternative `llvm.sqrt + fdiv`):

| Batch size | `vrsqrt + NR` | `sqrt + fdiv` | `sqrt + fdiv` slower by |
|---|---|---|---|
| 128  | 1056 ns | 993 ns  | noise |
| 256  | 1145 ns | 1280 ns | 11.8% |
| 512  | 1551 ns | 1865 ns | 20.2% |
| 1024 | 2233 ns | 2864 ns | 28.3% |
| 4096 | 6391 ns | 9048 ns | 41.6% |

Your concern about the perf cost of the intrinsic change is well-founded. Taking the alternative emit costs 10–40% of the f64 rsqrt throughput on this hardware at non-trivial batch sizes.

## What I've done with this

An earlier revision of this PR bundled two source changes: the algebraic simplifier guard *and* a one-line change in `xla/codegen/intrinsic/rsqrt.cc` that would make the `rsqrt` intrinsic always fall back to `sqrt + fdiv` for f64 (dropping the `!avx512f` condition). The benchmarks above are the cost of the second change.

I've now dropped the intrinsic change from this PR (commit [a0bda9e89d](https://github.com/openxla/xla/pull/40844/commits/a0bda9e89d)). The narrowed PR has a much simpler story:

- **Scope**: one element-type guard on the simplifier's `A/sqrt(B) => A*rsqrt(B)` rewrite.
- **What it fixes**: the user-visible bug from elixir-nx/nx#1704. `divide(constant(1), sqrt(x))` built from the correctly-rounded bucket stays in the correctly-rounded bucket rather than being silently downgraded to the implementation-defined-precision bucket.
- **What it doesn't change**: the `rsqrt` intrinsic emit. Users who explicitly call `rsqrt(x)` keep the fast approximate path — which is exactly the contract StableHLO spec promises them for that op.
- **Perf cost of this PR**: zero. The `rsqrt` emit path is unchanged; only the simplifier's behavior on the specific HLO pattern `divide(A, sqrt(B))` over f64 changes.
- **Spec argument (unchanged)**: `kSqrt` and `kDivide` are correctly-rounded per StableHLO; `kRsqrt` is implementation-defined precision. The user's source, composed entirely from correctly-rounded ops, carries that spec bound. The rewrite silently moves it.

## Option B — if you want to tighten the intrinsic too

If XLA project policy prefers to also close the gap on explicit `rsqrt(x)` callers, the intrinsic change is a separable one-line follow-up:

```diff
   if (options.disable_platform_dependent_math ||
-      (type.element_type() == F64 && !options.Contains("+avx512f")) ||
+      type.element_type() == F64 ||
       !options.Contains("+avx")) {
```

The benchmarks in this branch quantify the cost (table above), the `DoesNotIntroduceApproximateOpsFromCorrectlyRoundedInputsF64` invariant would still protect against future rewrites of the same class, and the existing `intrinsic_accuracy_test` contract at `kRsqrtF64Budget.cpu.regular = 1` would be retained (i.e. the fallback path is still ≤1 ULP, same as the claimed budget — composition of two correctly-rounded ops isn't itself correctly rounded, so we don't tighten the budget to 0). 

## Other changes addressed

- akuegel's matcher-code nit: the `!= F64` check is now expressed via `WithPredicate` on the Divide matcher (commit [d38d5f89b0](https://github.com/openxla/xla/pull/40844/commits/d38d5f89b0)), mirroring the existing `WithPredicate` usage at `algebraic_simplifier.cc:1398`. Used `WithPredicate` over a literal `WithShape(m::Shape().WithElementType(F64))` because the matcher library has no "NotElementType" combinator and a single predicate lambda reads better than two `Match` calls.

Happy to revise any of this. If you prefer the narrower scope I can rebase to a single-commit history if that's better as well.

blasphemetheus:
ah, ignore that, I put the wrong comment in the wrong place :( sorry, let me edit that back.

seantalts:
> Spec citation
From the [StableHLO spec](https://openxla.org/stablehlo/spec):

> Correctly-rounded bucket: add, sub, mul, div, sqrt.
Implementation-defined precision bucket: rsqrt, exp, log, tanh, erf, pow, …
A program written as divide(constant(1), sqrt(x)) is built entirely from the correctly-rounded bucket and carries the ≤1 ULP bound that comes from composing two correctly-rounded operations. The simplifier silently moves that program into the implementation-defined bucket. The StableHLO spec does not require the two buckets to produce the same bit pattern, and indeed on some emitters they don't.

Can you point me to where in the spec it says that some ops have more guarantees around rounding or accuracy than others? 

XLA doesn't typically make accuracy guarantees. We often want to mostly be within a few ULPs of the correct answer and enjoy the extra performance that leniency grants us.

blasphemetheus:
Right!

Hmm.

## Easy answer

So according to [the accuracy section](https://openxla.org/stablehlo/spec#accuracy):

> At the moment, StableHLO does not provide guarantees about numerical accuracy, but this may change in the future (#1156).

So it's clear what the answer is at this point. There is no location _within the spec_ to cite which contains the different buckets for rounding or accuracy.

But

## Quick Aside

Say if a cash register's calculation was off to the tens place somehow and you wanted to calculate a total as the seller (It's money, so you want precision to the hundredth place). You overpay or underpay up to ten cents sometimes. If you're keeping track of the register, you would want more precision to know if you have the correct amount of cents (did you count correctly, is all the money there, etc). Wouldn't you want a guarantee of numerical accuracy in this case? So there must be a standard of precision for a cash register, and for a cash register, it's whatever leads to a result of an accurate hundreds place.

## Assumption 1

XLA isn't a cash register.

Let's make an assumption: There is some situation where one would reasonably want some level of precision while using stableHLO. The spec claims that no guarantee is made (yet). Both cannot be true, so lets ignore the spec's claim. There is some standard of precision which would be unreasonable in some circumstances.

So now what is the guide for accuracy and precision? 

## Guide for precision

I do notice the IEEE-754 is often cited as a source of truth. For example:

> f16, f32 and f64 types corresponding to respectively binary16 ("half precision"), binary32 ("single precision") and binary64 ("double precision") formats described in [the IEEE 754 standard](https://ieeexplore.ieee.org/document/8766229).

or looking at [rsqrt](https://openxla.org/stablehlo/spec#rsqrt)

>  Semantics
> 
>  Performs element-wise reciprocal square root operation on operand tensor and produces a result tensor. Depending on the element type, does the following:
> 
> - For floats: rSqrt from IEEE-754.
> - For complex numbers: complex reciprocal square root.
> - For quantized types: dequantize_op_quantize(rsqrt, operand, type(result)).

and others. So [IEEE-754](https://ieeexplore.ieee.org/document/8766229) is often referred as a source of truth. That's different from saying that IEEE-754 _is always_ the source of truth.

## Assumption 2

In this PR I assume that IEEE-754 is the guide (second assumption).

in the 5.1 Overview section:

> Unless otherwise specified, each of the computational operations specified by this standard that returns a numeric result shall be performed as if it first produced an intermediate result correct to infinite precision and with unbounded range, and then rounded that intermediate result, if necessary, to fit in the destination’s format ...

So this is the "correctly rounded" bucket. Arithmetic operations are listed in 5.4.1 (including addition, subtraction, multiplication, division, squareRoot, fusedMultiplyAdd, etc). So squareRoot and division should be correctly-rounded.

Then there is the "Recommended operations" in section 9. The header states

> Clause 5 completely specifies the operations required for all supported arithmetic formats.
> This clause specifies additional operations that are recommended. 

Then in "9.2 Additional mathematical operations"

> In this subclause the domain of an operation is that subset of the affinely extended reals for which the corresponding mathematical function is well defined.
> A conforming operation shall return results correctly rounded for the applicable rounding direction for all operands in its domain.

This is where I got the second bucket. `rSqrt` is in this category.

Also "correct rounding" is defined.

> **correct rounding**: This standard’s method of converting an infinitely precise result to a floating-point number, as determined by the applicable rounding direction. A floating-point number so obtained is said to be correctly rounded

## Difficult Answer

There's an idea of a reproducible operation: in 11:

> A reproducible operation is one of the operations described in Clause 5 or is a supported operation from 9.2, 9.3, 9.5 or 9.6

A 9.2 op is "reproducible" (carrying IEEE-754's guarantees) only when supported as a 9.2-conforming op (meets the "shall return results correctly rounded" condition).

XLA's f64 `rsqrt` emit does not meet that condition, declaring a 1-ULP accuracy budget, not 0, so it isn't a 9.2 op, it's an approximation.

A user's program `divide(constant(1), sqrt(x))` is from two clause 5 ops, so it is reproducible. The simplifier rewrites it to contain `rSqrt` which is not correctly rounded for f64, so it's no longer reproducible per 11.

The improvement here is for the simplifier not to silently cross such a line in a program the user did not write a certain way. 

You can still write `rsqrt(x)` and keep the fast approximate path. The improvement is the simplifier won't rewrite `A/sqrt(B)` into the fast path for the one element type where the difference is visible.

blasphemetheus:
tldr: I'm using ieee-754 as the source for these buckets, so not the spec

seantalts:
Thanks for the detailed response. I'm going to be direct, because the thread has drifted from where it needs to be.
First, on the perf claim. "Perf cost of this PR: zero" is wrong, and the benchmark table directly above that sentence in your own comment shows why. The simplifier rewrite exists because f64 A/sqrt(B) benefits from the fast path — that's its whole point. And it's not only an rsqrt→sqrt swap: the rewrite also turns the outer op from a multiply into a divide. So guarding it means every f64 A/sqrt(B) site now takes sqrt+fdiv instead of rsqrt+mul, which is a substantially worse emit on both halves of the expression. Per your own numbers that's 10–40% slower at realistic batch sizes, and dropping the intrinsic change didn't eliminate that cost, it just moved it to a surface you stopped benchmarking. The measurement that speaks to this PR is a before/after on a compiled f64 divide(A, sqrt(B)) module. You have not produced that measurement, and the "zero" framing papers over its absence.
Second, on the framing. Your audit table ranks rewrites as "downgrade" or "improvement" on the accuracy axis and leaves the performance axis out entirely. That's the axis XLA actually optimizes for. We accept a few ULPs of slack in transcendentals in exchange for meaningful throughput — this is a deliberate project stance, not an oversight, and it's especially load-bearing for ML workloads. The divergence you're closing is 1 ULP in 28.9% of cases and 2 ULP in 0.2%. Paying a double-digit-percent regression on a hot pattern to close a 1-ULP gap is not a tradeoff this project makes, and framing it as a correctness fix rather than an accuracy/perf tradeoff is itself part of the problem.
Third, on the spec framing in the original PR body. The "correctly-rounded bucket" vs "implementation-defined precision bucket" distinction is presented as if it comes from the StableHLO spec. It doesn't — as you acknowledged two comments up, the spec explicitly disclaims accuracy guarantees and there is no such bucketing in it. The buckets are your own synthesis from IEEE-754, which is fine as an argument but is not what the PR body says. Leading a PR with a spec citation that isn't actually in the spec is the kind of thing that wastes reviewer time, because I read it, took it at face value, and had to work backwards from your later comment to figure out that the original framing was load-bearing and wrong.
Fourth, and I'll be plain about it: several replies in this thread read as AI-drafted and are not being checked against the PR's own evidence or the sources they cite. The "zero perf cost" claim contradicts a benchmark table in the same comment. The spec framing cites a document that doesn't say what it's quoted as saying. The audit table frames a project-norms question on the wrong axis. These are the kinds of errors a re-read catches, and their presence is making the review expensive — I'm spending my time disentangling claims that a pre-post pass would have caught. I don't object to drafting tools, but the posted text has to be your own checked position, and right now it isn't.
I'm closing this out. The downstream bug is real, but the fix trades a cost XLA has explicitly chosen to pay for a precision XLA has explicitly chosen not to guarantee, and the PR as written has leaned on framings that don't survive contact with either the benchmark table or the spec it cites. If there's a specific workload where the accuracy matters enough to justify the cost, that's a different conversation and would need to be framed around the workload.

PR #40846, opened 2026-04-13
[XLA:GPU][oneAPI][Bug Fix] Fix SPIR-V Extensions

mdfaijul:
This PR enables fixed set of common SPIR-V extensions. Previously, almost all extensions were enabled that caused legalization issues in the LLVM SPIR-V backend. In the future, more extensions will be enabled based on the readiness of SPIR-V backend. In order to set a fixed extension set, SPIR-V translation has been refactored in this PR.


xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
```diff
@@ -20,15 +20,19 @@ limitations under the License.
 #include <vector>
 
 #include "absl/base/no_destructor.h"
+#include "llvm/Analysis/TargetLibraryInfo.h"
```
akuegel, line 23: You are adding a few headers here but no dependencies to the BUILD file. I haven't checked but it could be that some of those headers need a new direct dependency on some BUILD target. We should not rely on getting these dependencies transitively.

mdfaijul: @akuegel Added required dependencies. Thanks!

akuegel (approved):

PR #40849, opened 2026-04-14
[xla:gpu] Remove unused IsNestedCommandBuffer API

shawnwang18:
The API was declared but never called anywhere in the codebase.


akuegel (approved):
Thanks for the cleanup. As far as I can tell this was temporarily used for implementing a workaround, and it looks like when the workaround was removed this was forgotten to be removed.

PR #40854, opened 2026-04-14
[xla:gpu] Use fast fingerprint for scheduling XLA module

ezhulenev:
This saves tens of minutes of string serialization for large modules!

### DIFF
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -55,6 +55,7 @@ limitations under the License.
 #include "xla/hlo/transforms/simplifiers/hlo_memory_scheduler.h"
 #include "xla/hlo/utils/hlo_query.h"
 #include "xla/layout.h"
+#include "xla/printer.h"
 #include "xla/service/buffer_value.h"
 #include "xla/service/gpu/alias_info.h"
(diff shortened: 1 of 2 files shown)


pifon2a (approved):
Awesomeness

xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -449,14 +450,21 @@ absl::Status RunP2PSchedulePreparation(HloModule* module) {
...
+  module->Print(&printer, HloPrintOptions::Canonical()
+                              .set_print_backend_config(true)
+                              // The backend config can be a json string,
+                              // and the order of keys in json is not
+                              // guaranteed. So we need to sort the keys
+                              // to make the fingerprint deterministic.
+                              .set_sort_backend_config(true));
+  tsl::Fprint128 fp128 = printer.ToFingerprint128();
   std::string fingerprint =
-      module->GetFingerprint128(HloPrintOptions::Canonical()
```
allanrenucci, line 453: Out of curiosity, why not update `HloModule::GetFingerprint128` to use `HighwayHashPrinter` instead?

ezhulenev: Yes, that would be the right approach, but from outside I'm afraid I'll break some TPU tests. You should definitely try to do that, this is the right fix!

PR #40919, opened 2026-04-15
Cross Host Data Transfers: Refactor ScheduleSendsOnLocalDevice into ScheduleTransfersOnLocalDevice

rao-ashish:
📝 Summary of Changes
This PR builds on [XLA #40585](https://github.com/openxla/xla/pull/40585), and is the next in a sequence of PRs that will refactor cross-host data transfer implementations to eventually rely on a shared helper function `CrossHostTransferBuffers`. `CrossHostTransferBuffers` is planned to eventually be integrated into the PJRT APIs to enable receiving data into preallocated receive buffers (this feature is being planned in collaboration with @gspschmid, @emilyfertig, and @pschuh). As part of implementing `CrossHostTransferBuffers`, this PR introduces a `CrossHostTransferSpec` struct and refactors `ScheduleSendsOnLocalDevice` into a more general `ScheduleTransfersOnLocalDevice`.

This PR also cleans up some of the error handling around the `prepare_transfers` closure inside `ScheduleTransfersOnLocalDevice` (formerly the `setup_sends` closure inside `ScheduleSendsOnLocalDevice`). Previously, if we got an error when we tried to extract the `LocalDeviceState`, we failed to set the `transfer_event` as an error. The current changes make sure that the error from `prepare_transfers` is always propagated through the transfer event.

🎯 Justification
It is difficult to achieve good comm/compute overlap with cross-host data transfers as the current implementation always allocates receive-buffers 'just-in-time', and because the GPU memory allocator blocks on the compute stream. `CrossHostTransferBuffers` will enable users to receive into preallocated receive buffers, making it easier to avoid the allocator blocking issue. This PR is a step towards implementing `CrossHostTransferBuffers`.

🚀 Kind of Contribution

🧪 Unit Tests:
This PR only refactors the implementation of `CrossHost{Send/Receive}Buffers`, so the pre-existing unit tests for those methods already test this PR.

🧪 Execution Tests:

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -714,167 +714,187 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
 
   // Schedule sends.
   for (auto& [device, send_idxs] : sends_by_device) {
+    GlobalDeviceId src_global_device_id = device->global_device_id();
+
     // Create a transfer event for transfers on this device.
     tsl::AsyncValueRef<BufferSequencingEvent> transfer_event =
         BufferSequencingEvent::Create(this->async_work_runner());
 
-    std::vector<tsl::RCReference<PjRtRawBuffer>> curr_raw_buffers;
-    curr_raw_buffers.reserve(send_idxs.size());
+    // Extract transfer dependencies, form transfer specs, and fulfill
+    // usage_event_promises.
     std::vector<tsl::RCReference<tsl::AsyncValue>> curr_transfer_dependency_avs;
-    std::vector<GlobalDeviceId> curr_dst_ids;
-    curr_dst_ids.reserve(send_idxs.size());
+    std::vector<CrossHostTransferSpec> transfer_specs;
+    transfer_specs.reserve(send_idxs.size());
 
     for (int idx : send_idxs) {
-      curr_raw_buffers.push_back(std::move(raw_buffers[idx]));
       for (tsl::RCReference<tsl::AsyncValue>& event :
            transfer_dependency_avs[idx]) {
         curr_transfer_dependency_avs.push_back(std::move(event));
       }
-      curr_dst_ids.push_back(dst_global_device_ids[idx]);
+      transfer_specs.push_back(CrossHostTransferSpec{
+          src_global_device_id, dst_global_device_ids[idx],
+          std::move(raw_buffers[idx])});
 
       usage_event_promises[idx]->Set(PjRtDeviceEventRef(transfer_event));
     }
 
-    ScheduleSendsOnLocalDevice(
-        device, std::move(transfer_event), std::move(curr_raw_buffers),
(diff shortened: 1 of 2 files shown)


xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -714,167 +714,187 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
+
+  auto prepare_transfers = [&]() -> absl::Status {
     gpu::GpuCollectives* gpu_collectives =
         gpu::GpuCollectives::Default(stream->parent()->GetPlatform()->Name());
 
     gpu::AcquiredCliquesMap acquired_cliques_map;
-    for (int i = 0; i < raw_buffers.size(); ++i) {
-      absl::StatusOr<PreparedTransfer> prepared_send =
+    for (int i = 0; i < transfer_specs.size(); ++i) {
+      bool is_sender = device_id == transfer_specs[i].src_global_device_id;
```
emilyfertig: Nit: Could you use `const` where it would help with interpretability (like here IMO) and in function signatures? (e.g. `PrepareTransfer`, I should have noted that on the last PR) https://google.github.io/styleguide/cppguide.html#Use_of_const

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -714,167 +714,187 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
-          /*dtype=*/U8,
-          /*count=*/mem->mem().size(),
-          /*peer=*/RankId(1),
-          /*executor=*/gpu::GpuCollectives::On(*stream)));
+      if (prepared_transfer.is_sender_) {
+        TF_RETURN_IF_ERROR(gpu_communicator->LaunchSend(
+            /*send_buffer=*/mem->mem(),
+            /*dtype=*/U8,
+            /*count=*/mem->mem().size(),
+            /*peer=*/RankId(1),
```
emilyfertig, line 824: Could you add a brief comment about why it's ok to hard-code RankId 0 and 1?

emilyfertig (changes requested):

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Added a comment about this.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Marked some variables as const. It becomes tricky to mark more variables / function parameters as const here because we're often using `std::move` and because some pre-existing helpers used unexpectedly require things to not be const (e.g. `client->gpu_run_options()` requires the `client` to not be const, meaning we can't mark it as const in `AcquireCliqueAndCommunicator` and `PrepareTransfer`).

emilyfertig (approved):
Thanks!

hawkinsp (approved):

PR #40920, opened 2026-04-15
[xla] Add a separate pass to propagate metadata across kCall instruction

ezhulenev:
- New `PropagateCallMetadata` HLO pass that propagates metadata (op_name prefix and stack_frame_id) from `kCall` instructions into their called computations, recursing through nested control-flow (while, conditional) but not into embedded computations (reduce's to_apply, etc.)
- Removes metadata propagation from `CallInliner` — the `RecursivelyUpdateMetadata` helper and per-instruction metadata update during inlining are no longer needed since the new pass handles this as a separate concern
- Wired into GPU compiler pre-SPMD pipeline before `CallInliner`, so metadata is propagated while `kCall` ops still exist (including non-inlinable calls)
- Test coverage for op_name propagation, stack frame concatenation, overflow protection, redundant prefix detection, nested calls, and idempotency

#### Motivation

The `CallInliner` only updates metadata for calls it actually inlines. Non-inlinable calls (e.g. calls with inlineable="false") were skipped entirely, leaving their callee instructions with incomplete metadata context. Extracting this into a standalone pass ensures all calls get metadata propagation regardless of inlining decisions.


xla/service/propagate_call_metadata.cc
```diff
@@ -0,0 +1,161 @@
...
+    return false;
+  }
+  // op_name is a substring of prefix — already captured.
+  if (!clean_name.empty() && absl::StrContains(clean_prefix, clean_name)) {
+    return false;
+  }
+  // Overflow guard.
+  if (clean_prefix.size() + clean_name.size() + 1 >= kMaxOpNameSize) {
+    return false;
+  }
```
thomasjoerg: Would it maybe be a better user experience to cap at kMaxOpNameSize rather than bailing out completely?
op_names are not handled very consistently, but let's try to avoid adding to the confusion.

xla/hlo/transforms/propagate_call_metadata_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  EXPECT_EQ(add->metadata().op_name(), "outer/inner");
+
+  // Second run should be a no-op.
+  TF_ASSERT_OK_AND_ASSIGN(bool changed2,
+                          PropagateCallMetadata().Run(module.get()));
+  EXPECT_FALSE(changed2);
+  EXPECT_EQ(add->metadata().op_name(), "outer/inner");
+}
+
+}  // namespace
```
thomasjoerg, line 357: Nice set of unit tests. Thank you!

thomasjoerg (approved):

xla/service/propagate_call_metadata.cc
ezhulenev: Good point! Added truncation to `kMaxOpNameSize` + a test of it.

akuegel (commented):
I think we prefer having general HLO passes inside xla/hlo/transforms directory.

ezhulenev:
@akuegel  👌  moved to `hlo/transforms`.

akuegel (approved):

ezhulenev:
Rebased

ermilovmaxim (approved):

thomasjoerg (approved):

mkuperst:
I think this will have to be rolled back. The reason metadata updates happen inside the inliner is that in case a function has multiple call-sites (i.e. the graph isn't flattened), we need to update the inlined code based on the context into which it's inlined, so it can't be done correctly before inlining.

I'm not sure what the right solution is for non-inlineable calls. If there's a single callsite then we could do something like this only for them, but if that can't be guaranteed, then I think this should be the consumer's responsibility.

ezhulenev:
Hmmm, can we keep the pass and paramterize inliner? For GPU we do this, because we flatten, other pipelines keep old behavior? I can send a PR later today.

mkuperst:
Do GPU always flatten before the first time it inlines anything, and do we want to keep doing it forever? If not, can we rethinkg this? One alternative is to keep this behavior in the inliner and then run the pass *late* rather than early, and propagate metadata into whatever's left

Another, which seems more correct in general is to not propagate into calls that are never inlined, and teach tooling that consumes metadata to deal with un-inlined calls. I agree this is more difficult, but it seems conceptually more correct. (@hawkinsp also feels strongly about this, I think :) )

Anyway, I'd prefer we roll back first, and figure out the right way to handle this later.

ezhulenev:
Yeah, go ahead and rollback, I'll send a patched version. On GPU the motivation is that we actually have non-inlined calls that dropped metadata.

Re: non-updating metadata - yes, that indeed feels a lot better, but it's monumental change, the way how we query backend configs and frontend attributes in all the places, they all will have to follow the call chain, which is a ton of complexity. For now I think I'll just keep it flexible.

PR #40948, opened 2026-04-15
[ROCm] Cleanup GEMM Rewriter Tests

ScXfjiang:
There are three tests in GemmRewriter that are skipped for ROCm, which are addressed in this PR.

The following two are due to the different padding strategies for ROCm and CUDA:
- GemmRewriteTest.Int8GemmRankGreaterThanTwo
- ParameterizedGemmRewriteTest.Int8GemmNotMultipleOfFour

The following one is due to the lack of support of S8 x S8 -> F32 on ROCm:
- ParameterizedGemmRewriteTest.GemmTypeCombinationCheck

### DIFF
--- a/xla/backends/gpu/transforms/gemm_rewriter.cc
+++ b/xla/backends/gpu/transforms/gemm_rewriter.cc
@@ -2535,8 +2535,6 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
          PrimitiveType::BF16, DataType::kBF16},
         {ComputationType::kF32, DataType::kFloat, PrimitiveType::F16,
          PrimitiveType::F16, DataType::kHalf},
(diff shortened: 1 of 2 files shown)


xla/backends/gpu/transforms/gemm_rewriter.cc
```diff
@@ -2566,6 +2564,14 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
       return true;
     }
 
+    // S8 x S8 -> F32 is supported by cuBLAS/cuBLASLt but not by
```
akuegel: Can this be added to supported_cublas_type_combinations instead?

ScXfjiang: Hi @akuegel, thanks! I have updated it.

akuegel (approved):

PR #40978, opened 2026-04-16
[XLA:GPU] Pin CuTeDSL runtime artifact for FFI test

mgoldfarb-nvidia:
📝 Summary of Changes

- Pin the CUTLASS 4.6.1 CuTeDSL runtime release artifacts for x86_64 and aarch64, with CUDA 12 and CUDA 13 selected from the configured target platform.
- Expose CuteDSLRuntime.h and the released shared runtime as ordinary Bazel dependencies through both WORKSPACE and Bzlmod.
- Update the existing test-only static registration shim to the runtime-owned v2 handlers and enable the existing vector-add FFI execution test in OSS.
- Intentionally exclude production runtime discovery or loading and all collective FFI implementation. Those remain follow-on work.

🎯 Justification

This is the minimal dependency and test prerequisite for the CuTeDSL collective FFI follow-on. Keeping the non-collective handler implementation in the released CuTeDSL runtime preserves the ownership boundary: its Python package and runtime library can evolve their ABI together, while OpenXLA only provides test-only static registration here.

The dependency-distribution blocker from the earlier discussion is now resolved: CUTLASS 4.6.1 publishes the runtime artifacts independently from the Python wheel. This update preserves the original review conversation while replacing the old wheel-based implementation with the current minimal artifact integration.

🚀 Kind of Contribution

✨ New Feature, 🧪 Tests

📊 Benchmark

Not applicable; this change adds dependency plumbing and enables an existing execution test.

🧪 Unit Tests

Buildifier check passed for all changed Bazel and Starlark files. MODULE.bazel retains an unrelated pre-existing buildifier ordering warning.

Bzlmod repository resolution passed on ComputeLab:

    bazel query --config=bzlmod @cutlass_cutedsl_runtime//:headers

🧪 Execution Tests

Passed on one ComputeLab B200 with hermetic CUDA 12.8:

    bazel test //xla/backends/gpu/libraries/cutedsl:ffi_test --test_output=errors --verbose_failures

Result: ffi_test_b200 passed in 8.2 seconds.


gflegar:
I merged half of this in https://github.com/openxla/xla/pull/41848 to the extent that it works internally for us. Haven't figured out what the best way is to make it work in OSS because the CuTe DSL dependency setup is out of date. I think OpenXLA moved to managing its dependencies through a separate GH repo a few weeks ago, so maybe it should be added there. Not sure yet what the policy for that is. @dimitar-asenov do you have any thoughts on how the external dependency setup could be done?

@mgoldfarb-nvidia I haven't gotten through reading the doc with the new runtime design you shared on chat with us yet, but was just wondering if we would still need to depend on the CuTe DSL wheel to make the FFI registration work if we upstreamed the things you wanted to OpenXLA? If not, then maybe that would be the cleanest solution, to have all the runtime components for CuTe DSL in xla/backends/gpu/libraries/cutedsl directly, and then we don't need to deal with pulling in additional dependencies into OpenXLA. CC @pifon2a

dimitar-asenov:
I don't really know much about the external dependency management, perhaps @MichaelHudgins might shed some light.

mgoldfarb-nvidia:
> @mgoldfarb-nvidia I haven't gotten through reading the doc with the new runtime design you shared on chat with us yet, but was just wondering if we would still need to depend on the CuTe DSL wheel to make the FFI registration work if we upstreamed the things you wanted to OpenXLA? If not, then maybe that would be the cleanest solution, to have all the runtime components for CuTe DSL in xla/backends/gpu/libraries/cutedsl directly, and then we don't need to deal with pulling in additional dependencies into OpenXLA. CC @pifon2a

Currently for OSS the where is the only way to get access to the necessary libraries. We can investigate an alternative hosting path for the underlying runtime library.

Open-souring the runtime code is an option for us to consider. I think if we stick with the older runtime path i.e. only load compiled ELF binary we should not have too much roadblock here since its basically OrcJit. For the newer runtime code that might be harder due to the desire to support partial compilation (e.g. of LLVM). In that path we have to solve static linking and deliver the archive through github or other binary delivery process.

akuegel:
What is the state of this PR, is it already clear in which direction this needs to go? What are the next steps?

gflegar:
Not really, I'm struggling to figure out who can make a decision on how the dependencies should be managed. Would you know who that could be @akuegel?

The latest is that Nvidia could provide a separate repo for the (static) runtime we use internally, and I assume it could be possible to package the DSO in that repo as well @mgoldfarb-nvidia? This would avoid pulling in the entirety of the XLA, and we only get the runtime component with the FFI handlers that we actually need.

Then we would need to depend on that repo from somewhere. Would that be XLA itself, or https://github.com/google-ml-infra/rules_ml_toolchain? (or alternatively copy the binaries directly into XLA).

mgoldfarb-nvidia:
> The latest is that Nvidia could provide a separate repo for the (static) runtime we use internally, and I assume it could be possible to package the DSO in that repo as well @mgoldfarb-nvidia? This would avoid pulling in the entirety of the XLA, and we only get the runtime component with the FFI handlers that we actually need.

I am discussing with the team how we can host binaries on https://github.com/nvidia/cutlass for distribution. We can share both static archive intended for internal use as well as the DSO for general public use. I would prefer we build with DSO as much as possible since it will have less likelihood of issues. If we start publishing it will hopefully also be in the next 4.6 release which will begin next month.

dimitar-asenov:
It's a been a while since this PR has seen any activity on it. Seeing the latest comment it seems that it may not be ready for merging yet. I'll mark it as draft.

@mgoldfarb-nvidia Please mark as ready for review once there's more clarity what should be done (and after a rebase).

mgoldfarb-nvidia:
> It's a been a while since this PR has seen any activity on it. Seeing the latest comment it seems that it may not be ready for merging yet. I'll mark it as draft.
> 
> @mgoldfarb-nvidia Please mark as ready for review once there's more clarity what should be done (and after a rebase).

Thanks, yes we are planning to release static and dynamic runtime and compiler libraries starting around 4.6 cutlass release. These should be accessible separate from the Python wheel.

mgoldfarb-nvidia:
Updated using new GitHub artifacts

xla/backends/gpu/libraries/cutedsl/vector_add.hlo
chsigg, line 1: Looks like an unrelated change.

third_party/cutlass_cutedsl_runtime/BUILD.bazel
```diff
@@ -11,4 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# =============================================================================
+# ==============================================================================
```
chsigg, line 14: I'm aware that there are some counter examples, but generally we don't add license headers to starlark files.

chsigg (approved):

third_party/cutlass_cutedsl_runtime/BUILD.bazel
chsigg: Actually, the majority of starlark files do have license headers. Please ignore my above comment.

PR #41021, opened 2026-04-16
[XLA:GPU] add cuDNN AOT compilation support

Cjkkkk:
📝 Summary of Changes
Enable cuDNN AOT deviceless compilation in XLA.

* Allow cudnn fusion compiler and cudnn custom call compiler to take in null dnn support and device description.
* When dnn support is null, use device description to query/prepare/build cudnn graph without initializing cudnn handles.
* Add a conversion method to convert XLA deviceDescription to cuDNN deviceProperty.
* Now in `RunBackend`, `RunCudnnCompilerPasses` can be run even if se is null.

### DIFF
--- a/xla/backends/gpu/codegen/cudnn_test.cc
+++ b/xla/backends/gpu/codegen/cudnn_test.cc
@@ -134,7 +134,8 @@ class CuDnnFusionFileCheckTest : public CuDnnFusionTest {
     const std::string root_name(
         module->entry_computation()->root_instruction()->name());
     BinaryMap dnn_compiled_graphs;
-    CuDnnFusionCompiler cudnn_compiler(*stream_executor()->AsDnn(),
+    CuDnnFusionCompiler cudnn_compiler(stream_executor()->AsDnn(),
+                                       se::DeviceDescription(),
                                        dnn_compiled_graphs);
     // Run filecheck even if CuDnnFusionCompiler failed.
     cudnn_compiler.Run(module.get()).IgnoreError();
@@ -280,8 +281,8 @@ ENTRY e {
   ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
                        ParseAndReturnVerifiedModule(kHloText));
   BinaryMap dnn_compiled_graphs;
-  CuDnnFusionCompiler cudnn_compiler(*stream_executor()->AsDnn(),
-                                     dnn_compiled_graphs);
+  CuDnnFusionCompiler cudnn_compiler(
+      stream_executor()->AsDnn(), se::DeviceDescription(), dnn_compiled_graphs);
   ASSERT_OK_AND_ASSIGN(bool changed, cudnn_compiler.Run(module.get()));
   EXPECT_TRUE(changed);
   EXPECT_THAT(module->entry_computation()->root_instruction(),
@@ -314,8 +315,8 @@ e {
   ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
                        ParseAndReturnVerifiedModule(kHloText));
   BinaryMap dnn_compiled_graphs;
-  CuDnnFusionCompiler cudnn_compiler(*stream_executor()->AsDnn(),
-                                     dnn_compiled_graphs);
+  CuDnnFusionCompiler cudnn_compiler(
+      stream_executor()->AsDnn(), se::DeviceDescription(), dnn_compiled_graphs);
   EXPECT_THAT(cudnn_compiler.Run(module.get()),
               absl_testing::IsOkAndHolds(false));
   // Single dot is not supported by cuDNN, so Triton should be used.
@@ -363,8 +364,8 @@ ENTRY e {
   ROOT r = tuple(f0, f1)
 })"));
   BinaryMap dnn_compiled_graphs;
-  CuDnnFusionCompiler cudnn_compiler(*stream_executor()->AsDnn(),
-                                     dnn_compiled_graphs);
+  CuDnnFusionCompiler cudnn_compiler(
+      stream_executor()->AsDnn(), se::DeviceDescription(), dnn_compiled_graphs);
   ASSERT_OK_AND_ASSIGN(bool changed, cudnn_compiler.Run(module.get()));
   EXPECT_TRUE(changed);
   EXPECT_THAT(module->entry_computation()->root_instruction(),
@@ -476,6 +477,112 @@ CHECK: "stride": [{{[[:space:]]*}}1,{{[[:space:]]*}}1,{{[[:space:]]*}}256{{[[:sp
   )"));
 
   EXPECT_TRUE(RunAndCompare(kHloText, ErrorSpec{/*aabs=*/1e-3, /*arel=*/1e-3}));
+}
+
+TEST_F(CuDnnFusionExecutionTest, DotF32DevicelessCompilationSucceeds) {
+  if (!IsAtLeastCuDnnVersion(9, 8)) {
+    GTEST_SKIP() << "Deviceless DeviceProperties requires cuDNN 9.8+.";
+  }
+  constexpr absl::string_view kHlo = R"(
+fusion1 {
+  p0 = f32[32,96] parameter(0)
+  p1 = f32[96,64] parameter(1)
+  ROOT r = f32[32,64] dot(p0, p1),
+    lhs_contracting_dims={1}, rhs_contracting_dims={0}
+}
+
+ENTRY e {
+  p0 = f32[32,96] parameter(0)
+  p1 = f32[96,64] parameter(1)
+  ROOT _ = f32[32,64] fusion(p0, p1), kind=kCustom, calls=fusion1,
+    backend_config={"fusion_backend_config": {kind: "__cudnn$fusion"}}
+})";
+
+  // Verify that CuDnnFusionCompiler succeeds with null dnn_support (deviceless
+  // mode), driven solely by the DeviceDescription — no live cuDNN handle.
+  {
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
+                         ParseAndReturnVerifiedModule(kHlo));
+    const se::DeviceDescription& device_description =
+        this->device_description();
+    BinaryMap dnn_compiled_graphs;
+    CuDnnFusionCompiler cudnn_compiler(/*dnn_support=*/nullptr,
+                                       device_description, dnn_compiled_graphs);
+    ASSERT_OK_AND_ASSIGN(bool changed, cudnn_compiler.Run(module.get()));
+    EXPECT_TRUE(changed);
+  }
+
+  // Now compile the same fusion end-to-end with deviceless cuDNN compilation
+  // enabled and actually execute it, comparing the result against the reference
+  // backend. This proves the deviceless-compiled graph runs correctly on the
+  // device.
+  ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
+                       ParseAndReturnVerifiedModule(kHlo));
+  DebugOptions& debug_options =
+      module->mutable_config().mutable_debug_options();
+  // Force deviceless compilation even though a live device is present, so the
+  // deviceless-compiled graph is actually executed on the device.
+  debug_options.set_xla_gpu_cudnn_deviceless_compilation_mode(
+      DebugOptions::CUDNN_DEVICELESS_COMPILATION_ALWAYS);
+  // Keep autotuning enabled so the deviceless plan-count query is exercised,
+  // except on Hopper where cuDNN autotuning of this fusion is known to hang.
+  debug_options.set_xla_gpu_autotune_level(get_cuda_cc().IsHopper() ? 0 : 4);
+  EXPECT_TRUE(RunAndCompare(std::move(module),
+                            ErrorSpec{/*aabs=*/1e-3, /*arel=*/1e-3}));
+}
+
+TEST_F(CuDnnFusionExecutionTest, DotF32DevicelessBinaryMatchesLive) {
+  if (!IsAtLeastCuDnnVersion(9, 8)) {
+    GTEST_SKIP() << "Deviceless DeviceProperties requires cuDNN 9.8+.";
+  }
+  constexpr absl::string_view kHlo = R"(
+fusion1 {
+  p0 = f32[32,96] parameter(0)
+  p1 = f32[96,64] parameter(1)
+  r = f32[32,64] dot(p0, p1),
+    lhs_contracting_dims={1}, rhs_contracting_dims={0}
+  neg = f32[32,64] negate(r)
+  ROOT a = f32[32,64] add(neg, neg)
+}
+
+ENTRY e {
+  p0 = f32[32,96] parameter(0)
+  p1 = f32[96,64] parameter(1)
+  ROOT _ = f32[32,64] fusion(p0, p1), kind=kCustom, calls=fusion1,
+    backend_config={"fusion_backend_config": {kind: "__cudnn$fusion"}}
+})";
+
+  se::StreamExecutor* executor = stream_executor();
+  const se::DeviceDescription& device_description =
+      executor->GetDeviceDescription();
+
+  // Compile with live dnn_support.
+  ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module_live,
+                       ParseAndReturnVerifiedModule(kHlo));
+  BinaryMap binary_map_live;
+  CuDnnFusionCompiler live_compiler(executor->AsDnn(), se::DeviceDescription(),
+                                    binary_map_live);
+  ASSERT_OK_AND_ASSIGN(bool changed_live, live_compiler.Run(module_live.get()));
+  ASSERT_TRUE(changed_live);
+
+  // Compile deviceless (null dnn_support, same DeviceDescription).
+  ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module_deviceless,
+                       ParseAndReturnVerifiedModule(kHlo));
+  BinaryMap binary_map_deviceless;
+  CuDnnFusionCompiler deviceless_compiler(
+      /*dnn_support=*/nullptr, device_description, binary_map_deviceless);
+  ASSERT_OK_AND_ASSIGN(bool changed_deviceless,
+                       deviceless_compiler.Run(module_deviceless.get()));
+  ASSERT_TRUE(changed_deviceless);
+
+  // Both maps must have the same keys and identical serialized binaries,
+  // proving the deviceless path is equivalent to the live path.
+  ASSERT_EQ(binary_map_live.size(), binary_map_deviceless.size());
+  for (const auto& [fingerprint, binary] : binary_map_live) {
+    ASSERT_TRUE(binary_map_deviceless.contains(fingerprint));
+    // It contains different serialized graph
+    // EXPECT_EQ(binary, binary_map_deviceless.at(fingerprint));
+  }
 }
 
 TEST_F(CuDnnFusionExecutionTest, DotBF16WithCopyExecutesCorrectly) {
--- a/xla/backends/gpu/target_config/cudnn_device_props.cc
+++ b/xla/backends/gpu/target_config/cudnn_device_props.cc
@@ -1,0 +1,80 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/backends/gpu/target_config/cudnn_device_props.h"
+
+#include <memory>
+#include <vector>
+
+#include "absl/status/statusor.h"
+#include "absl/strings/str_format.h"
+#include "third_party/cudnn_frontend/include/cudnn_frontend.h"
+#include "xla/stream_executor/cuda/cudnn_frontend_helpers.h"
+#include "xla/stream_executor/device_description.h"
+
+namespace xla::gpu {
+
+absl::StatusOr<std::shared_ptr<cudnn_frontend::DeviceProperties>>
+BuildDeviceProperties(const stream_executor::DeviceDescription& desc) {
+  const auto* cc = desc.gpu_compute_capability().cuda_compute_capability();
+  int device_ver = cc ? (cc->major * 100 + cc->minor * 10) : 0;
+  int driver_ver = desc.driver_version().major_version() * 1000 +
+                   desc.driver_version().minor_version() * 10;
+  std::string json = absl::StrFormat(
+      R"json({
+  "deviceVer":%d,
+  "multiProcessorCount":%d,
+  "warpSize":%d,
+  "maxSharedMemoryPerBlock":%d,
+  "maxSharedMemoryPerBlockOptin":%d,
+  "reservedSharedMemoryPerBlock":%d,
+  "maxRegistersPerSM":%d,
+  "maxCtasPerSM":%d,
+  "maxThreadsPerBlock":%d,
+  "maxThreadsPerSM":%d,
+  "regsPerBlock":%d,
+  "totalGlobalMem":%d,
+  "smClockRateKHz":%d,
+  "l2CacheSize":%d,
+  "maxBlockSize":[%d,%d,%d],
+  "memClockRateKHz":%d,
+  "maxGridSize":[%d,%d,%d],
+  "supportCoopLaunch":%d,
+  "pciDeviceId":0,
+  "isTccDriver":0,
+  "cudaDeviceId":0,
+  "driverVer":%d,
+  "deviceName":"%s"
+})json",
+      device_ver, desc.core_count(), desc.threads_per_warp(),
+      desc.shared_memory_per_block(), desc.shared_memory_per_block_optin(),
+      desc.reserved_shared_memory_per_block(), desc.registers_per_core_limit(),
+      desc.max_blocks_per_multiprocessor(), desc.threads_per_block_limit(),
+      desc.threads_per_core_limit(), desc.registers_per_block_limit(),
+      desc.device_memory_size(),
+      static_cast<int64_t>(desc.clock_rate_ghz() * 1e6), desc.l2_cache_size(),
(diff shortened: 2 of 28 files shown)


xla/backends/gpu/codegen/cudnn_test.cc
```diff
@@ -427,6 +431,90 @@ CHECK: "stride": [{{[[:space:]]*}}1,{{[[:space:]]*}}1,{{[[:space:]]*}}64{{[[:spa
...
+
+  se::StreamExecutor* executor = backend().default_stream_executor();
+  const se::DeviceDescription& device_description =
+      executor->GetDeviceDescription();
+
+  // Compile with live dnn_support.
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module_live,
+                          ParseAndReturnVerifiedModule(kHlo));
+  BinaryMap binary_map_live;
+  CuDnnFusionCompiler live_compiler(executor->AsDnn(), device_description,
```
sergachev, line 491: nit: empty se::DeviceDescription() instead of real device_description here to be sure

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -419,6 +419,11 @@ void PreloadCudnnSubLibsHelper(dnn::ConvolutionKind kind) {
 CudnnSupport::CudnnSupport(StreamExecutor* parent) : parent_(parent) {}
 
 absl::Status CudnnSupport::Init() {
+  if (parent_ == nullptr) {
+    LOG(INFO) << "No steam executor provided, initialize cudnnSupport with "
```
sergachev, line 423: This would spam a lot, should be VLOG I think.

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6847,36 +6865,97 @@ absl::StatusOr<std::unique_ptr<dnn::DnnGraph>> CudnnSupport::DeserializeGraph(
...
+      R"("cudaDeviceId":0,"driverVer":%d,"deviceName":"%s"})",
+      device_ver, desc.core_count(), desc.threads_per_warp(),
+      desc.shared_memory_per_block(), desc.shared_memory_per_block_optin(),
+      /*desc.reserved_shared_memory_per_block()*/ 1024,
+      desc.registers_per_core_limit(),
+      /*desc.max_blocks_per_multiprocessor()*/ 32,
+      desc.threads_per_block_limit(), desc.threads_per_core_limit(),
+      desc.registers_per_block_limit(), desc.device_memory_size(),
+      static_cast<int64_t>(desc.clock_rate_ghz() * 1e6), desc.l2_cache_size(),
+      3996000, /*desc.supports_coop_launch()*/ 1, driver_ver, desc.name());
```
sergachev, line 6891: 3996000 shouldn't be hardcoded

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6847,36 +6865,97 @@ absl::StatusOr<std::unique_ptr<dnn::DnnGraph>> CudnnSupport::DeserializeGraph(
...
+  int driver_ver =
+      desc.driver_version().major() * 1000 + desc.driver_version().minor() * 10;
+  return absl::StrFormat(
+      R"({"deviceVer":%d,"multiProcessorCount":%d,"warpSize":%d,)"
+      R"("maxSharedMemoryPerBlock":%d,"maxSharedMemoryPerBlockOptin":%d,)"
+      R"("reservedSharedMemoryPerBlock":%d,"maxRegistersPerSM":%d,)"
+      R"("maxCtasPerSM":%d,"maxThreadsPerBlock":%d,"maxThreadsPerSM":%d,)"
+      R"("regsPerBlock":%d,"totalGlobalMem":%d,"smClockRateKHz":%d,)"
+      R"("l2CacheSize":%d,"maxBlockSize":[1024,1024,64],"memClockRateKHz":%d,)"
+      R"("maxGridSize":[2147483647,65535,65535],)"
```
sergachev, line 6880: block_dim_limit_x/y/z

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6847,36 +6865,97 @@ absl::StatusOr<std::unique_ptr<dnn::DnnGraph>> CudnnSupport::DeserializeGraph(
...
+  int device_ver = cc ? (cc->major * 100 + cc->minor * 10) : 0;
+  int driver_ver =
+      desc.driver_version().major() * 1000 + desc.driver_version().minor() * 10;
+  return absl::StrFormat(
+      R"({"deviceVer":%d,"multiProcessorCount":%d,"warpSize":%d,)"
+      R"("maxSharedMemoryPerBlock":%d,"maxSharedMemoryPerBlockOptin":%d,)"
+      R"("reservedSharedMemoryPerBlock":%d,"maxRegistersPerSM":%d,)"
+      R"("maxCtasPerSM":%d,"maxThreadsPerBlock":%d,"maxThreadsPerSM":%d,)"
+      R"("regsPerBlock":%d,"totalGlobalMem":%d,"smClockRateKHz":%d,)"
+      R"("l2CacheSize":%d,"maxBlockSize":[1024,1024,64],"memClockRateKHz":%d,)"
```
sergachev, line 6879: thread_dim_limit_x/y/z

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6847,36 +6865,97 @@ absl::StatusOr<std::unique_ptr<dnn::DnnGraph>> CudnnSupport::DeserializeGraph(
   return std::make_unique<CudnnGraph>(std::move(graph));
 }
 
-absl::Status CudnnGraph::Prepare(dnn::DnnSupport& dnn_support,
+std::string BuildCudnnDevicePropsJson(const DeviceDescription& desc) {
+  const auto* cc = desc.gpu_compute_capability().cuda_compute_capability();
+  int device_ver = cc ? (cc->major * 100 + cc->minor * 10) : 0;
```
sergachev, line 6870: Is the "a" suffix relevant?

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6847,36 +6865,97 @@ absl::StatusOr<std::unique_ptr<dnn::DnnGraph>> CudnnSupport::DeserializeGraph(
...
+        graph_.create_execution_plans({cudnn_frontend::HeurMode_t::A}));
+    if (engine_options.require_determinism) {
+      graph_.deselect_numeric_notes(
+          {cudnn_frontend::NumericalNote_t::NONDETERMINISTIC});
+    }
+    if (engine_options.require_command_buffer) {
+      graph_.select_behavior_notes(
+          {cudnn_frontend::BehaviorNote_t::SUPPORTS_CUDA_GRAPH_NATIVE_API});
+    }
+    RETURN_CUDNN_FRONTEND_STATUS(graph_.check_support(cudnn_handle));
```
sergachev, line 6914: These lines shouldn't need duplication

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6847,36 +6865,97 @@ absl::StatusOr<std::unique_ptr<dnn::DnnGraph>> CudnnSupport::DeserializeGraph(
   return std::make_unique<CudnnGraph>(std::move(graph));
 }
 
-absl::Status CudnnGraph::Prepare(dnn::DnnSupport& dnn_support,
+std::string BuildCudnnDevicePropsJson(const DeviceDescription& desc) {
```
EusebioDM, line 6868: I think it would be quite valuable to have a tests ensuring that for [all TargetConfigs](https://github.com/openxla/xla/tree/main/xla/backends/gpu/target_config/specs) (which includes the DeviceDescriptor), this function produces a cudnnDeviceProps that is equivalent to the one we'd query from the actual device (e.g. [doing this](https://github.com/NVIDIA/cudnn-frontend/blob/main/samples/cpp/misc/deviceless_aot_compilation.cpp#L42-L43)).

WDYT about extracting this standalone function into a new file/target under `backends/gpu/target_config`, and then creating a test similar to [embedded_target_config_test.cc](https://github.com/openxla/xla/blob/main/xla/backends/gpu/target_config/embedded_target_config_test.cc).cc to verify this behavior.

xla/service/gpu/gpu_compiler.cc
```diff
@@ -2673,19 +2673,17 @@ absl::StatusOr<std::unique_ptr<Executable>> GpuCompiler::RunBackend(
...
-                                           &dnn_compiled_graphs));
-  }
-
   const DebugOptions& debug_opts = module->config().debug_options();
   ASSIGN_OR_RETURN(GpuTopology gpu_topology,
                    InferGpuTopology(module->config(), stream_exec, options,
                                     debug_opts, platform_id_));
 
+  BinaryMap dnn_compiled_graphs;
+  se::dnn::DnnSupport* dnn = stream_exec ? stream_exec->AsDnn() : nullptr;
```
EusebioDM, line 2682: For cuDNN versions that don't support AOT mode, I think we'd like to keep the current behavior, and probably log a message.

xla/stream_executor/cuda/cuda_dnn.cc
EusebioDM, line 423: I think that we should also return an error here if the cuddn version doesn't support AOT mode.

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -6847,36 +6865,97 @@ absl::StatusOr<std::unique_ptr<dnn::DnnGraph>> CudnnSupport::DeserializeGraph(
...
+  const auto* cc = desc.gpu_compute_capability().cuda_compute_capability();
+  int device_ver = cc ? (cc->major * 100 + cc->minor * 10) : 0;
+  int driver_ver =
+      desc.driver_version().major() * 1000 + desc.driver_version().minor() * 10;
+  return absl::StrFormat(
```
EusebioDM, line 6873: Nit: I think you can do something like this here:
```
R"json(
{deviceVer":%d,"multiProcessorCount":%d,"warpSize":%d, ... }
)json"
```

That way you don't need the `R("` in every line, and the code highlighter and formatter should recognize it as JSON.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: I dont think we can derive this memory clock rate from xla device description currently. I use this hard coded value for testing. 
1. Use dummy value if cudnn doesnt actually use this value
2. modify xla to add similar attribute in device description if cudnn needs it

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: Right, block_dim_limit_x/y/z would be safer, I think these values are fixed after CC 3.0 but could change in the future.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: I think no, when initialized directly using cudnn [API](https://github.com/NVIDIA/cudnn-frontend/blob/c6011ef5cebfcb3d6ad785a7baccf170d27c3245/samples/cpp/misc/deviceless_aot_compilation.cpp#L43), cudnn query cuda device property to get the hardware compute capability, it is no the compilation target.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: sounds good, we should definitely do this.

xla/stream_executor/cuda/cuda_dnn.cc
sergachev: 2 is doable by wiring through the device description, the value is already queried by XLA here https://github.com/openxla/xla/blob/141bbb6539642e1c2a46ff748bd330313bad6bb4/xla/stream_executor/cuda/cuda_executor.cc#L1629

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: since `CudnnSupport::Init()` is only called with non null StreamExecutor in `dnn::DnnSupport* CudaExecutor::AsDnn()`. I think these lines never run.

Cjkkkk: since `CudnnSupport::Init()` is only called with non null StreamExecutor in `dnn::DnnSupport* CudaExecutor::AsDnn()`. I think these lines never run.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: thanks, I added the `mem_clock_khz`, could you take another look?

xla/stream_executor/cuda/cuda_dnn.cc
sergachev: If this is never expected to run, shouldn't it be returning an error instead?

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: I added the test. could you take a look

xla/backends/gpu/target_config/cudnn_device_props_test.cc
```diff
@@ -0,0 +1,156 @@
...
+#include "xla/tsl/platform/test.h"
+
+namespace xla::gpu {
+namespace {
+
+namespace se = ::stream_executor;
+
+// Fields that BuildDeviceProperties() hardcodes (or that don't affect cuDNN
+// plan selection), so they are not expected to match the live device.
+constexpr const char* kIgnoredFields[] = {
```
EusebioDM: Nit: use absl::string_view instead of char* here, and below too

xla/service/gpu/nvptx_compiler.cc
```diff
@@ -323,7 +323,8 @@ absl::Status NVPTXCompiler::OptimizeHloPostLayoutAssignment(
 }
 
 absl::Status NVPTXCompiler::RunCudnnCompilerPasses(
```
EusebioDM, line 329: We should also avoid attempting to do cudnn deviceless compilation for versions that don't support it yet. 

So we could do an early return if dnn_support is null and the cudnn version doesn't support deviceless.

xla/backends/gpu/transforms/cudnn_fusion_compiler.cc
```diff
@@ -985,6 +992,7 @@ class CuDnnFusionVisitor : public DfsHloRewriteVisitor {
                               {fingerprint_without_workspace, workspace_size});
       TF_RETURN_IF_ERROR(add_workspace(workspace_size));
       TF_ASSIGN_OR_RETURN(const std::string serialized, serialize_graph(graph));
+      std::cerr << serialized << "\n";
```
EusebioDM: We shouldn't forget to remove this submitting :)

xla/stream_executor/cuda/cuda_dnn.cc
EusebioDM: Thanks for adding the test, it looks great.

xla/stream_executor/cuda/cuda_dnn.cc
Cjkkkk: fixed, added a guard to make sure in deviceless mode, we never reach this code.

xla/service/gpu/gpu_compiler.cc
Cjkkkk: added, PTAL

xla/backends/gpu/target_config/cudnn_device_props_test.cc
Cjkkkk: fixed, PTAL

xla/service/gpu/nvptx_compiler.cc
Cjkkkk: fixed, PTAL

sergachev:
LGTM

EusebioDM:
Adding some context here for anyone reading the PR, based on my offline chats with @Cjkkkk .

The main open question is determining the exact feature gap between these two modes. Ideally, we would use the deviceless API exclusively to avoid maintaining two separate code paths. However, @Cjkkkk  pointed out that deviceless cuDNN mode does not support all engines yet. If this gap is minor or if support is coming soon, we might still be able to go entirely deviceless.

There is a known issue with a microbenchmark hanging (at least internally at Google). I don't think this needs to block the PR since these changes are currently hidden behind a feature flag.

xla/stream_executor/cuda/cuda_dnn.cc
```diff
@@ -419,6 +420,9 @@ void PreloadCudnnSubLibsHelper(dnn::ConvolutionKind kind) {
 CudnnSupport::CudnnSupport(StreamExecutor* parent) : parent_(parent) {}
 
 absl::Status CudnnSupport::Init() {
+  if (parent_ == nullptr) {
+    return absl::InternalError("No steam executor provided.");
```
EusebioDM: Typo: Should be "stream" here

xla/xla.proto
```diff
@@ -562,6 +562,9 @@ message DebugOptions {
   // but potentially higher the performance.
   optional int32 xla_gpu_cudnn_gemm_max_plans = 318;
 
+  // Enable cuDNN deviceless compilation.
+  optional bool xla_gpu_enable_cudnn_deviceless_compilation = 503;
```
EusebioDM: Given the issues in cudnn versions < 9.21 would prevent us from always doing deviceless, WDYT about a 3-state flag (Disabled / Auto / Always)?

"Auto" would trigger during deviceless compilation, and "Always" gives us a testing path for eventual deprecation of the device-dependent path. 

This would be similar to the CUB deviceless flag https://github.com/openxla/xla/blob/main/xla/xla.proto#L408-L421

xla/backends/gpu/codegen/cudnn_test.cc
```diff
@@ -428,6 +432,92 @@ CHECK: "stride": [{{[[:space:]]*}}1,{{[[:space:]]*}}1,{{[[:space:]]*}}64{{[[:spa
   EXPECT_TRUE(RunAndCompare(kHloText, ErrorSpec{/*aabs=*/1e-3, /*arel=*/1e-3}));
 }
 
+TEST_F(CuDnnFusionExecutionTest, DotF32DevicelessCompilationSucceeds) {
```
EusebioDM, line 482: Would it be possible to actually run the dot that we compiled using the deviceless mode here? 
Ideally also with the autotuner not disabled.

That way we make sure things are actually running properly. You might need to disable that for H100s if you run into the hanging problem that I was having.

EusebioDM (changes requested):
Sorry for all the back and forth! 

I did some investigation and it seems like the hanging issue on H100 is likely coming from a bug in cudnn that was fixed in version 9.21. I'm no longer able to reproduce it in that version.

Because of this issue I don't think we can have the deviceless path be the default path, so I left some comments regarding that.

But I think after those changes we should be in a good state to merge this, and build upon that if needed. Also adding @beckerhe and @derdrdirk for them to take a look.

derdrdirk (approved):
LGTM

@EusebioDM what is missing to merge this?

EusebioDM:
> LGTM
> 
> @EusebioDM what is missing to merge this?

[This comment](https://github.com/openxla/xla/pull/41021/changes#r3274894727) would be the only blocker for me. 

I left another comment about adding adding another test, but thats not a blocker.

xla/backends/gpu/codegen/cudnn_test.cc
Cjkkkk: Done.

xla/xla.proto
Cjkkkk: Sounds good, fixed

Cjkkkk:
I addressed the comments and make the flag 3 staged. Could someone take another look? Another thing is that I couldnt quite reproduce the hang issue with cudnn 9.19 + H100.

xla/service/gpu/nvptx_compiler.cc
```diff
@@ -327,22 +327,58 @@ absl::Status NVPTXCompiler::OptimizeHloPostLayoutAssignment(
...
+  bool use_deviceless_cudnn = false;
+  switch (debug_options.xla_gpu_cudnn_deviceless_compilation_mode()) {
+    case DebugOptions::CUDNN_DEVICELESS_COMPILATION_DISABLED:
+      use_deviceless_cudnn = false;
+      break;
+    case DebugOptions::CUDNN_DEVICELESS_COMPILATION_ALWAYS:
+      // Force deviceless even if we have a device.
+      use_deviceless_cudnn = true;
+      break;
+    case DebugOptions::CUDNN_DEVICELESS_COMPILATION_UNSET:
```
EusebioDM: Nit: I think it would sense to go for disabled if this isn't set (which shouldn't happen, but just in case).

xla/debug_options_flags.cc
```diff
@@ -529,6 +529,8 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_experimental_aot_compiled_thunks(true);
   opts.set_xla_gpu_deviceless_cub_mode(
       DebugOptions::DEVICELESS_CUB_WITH_FALLBACK);
+  opts.set_xla_gpu_cudnn_deviceless_compilation_mode(
```
EusebioDM, line 530: I think we should keep this in the `CUDNN_DEVICELESS_COMPILATION_DISABLED` mode by default for now, until we figure out the potential hangin issue.

EusebioDM (changes requested):
Thanks for the changes! LGTM just the minor comment about the default flag value, we can merge after that.

xla/debug_options_flags.cc
Cjkkkk: fixed.

xla/service/gpu/nvptx_compiler.cc
Cjkkkk: fixed.

EusebioDM (approved):
LGTM, thanks!

EusebioDM (changes requested):
Looks like there are a couple presubmit tests broken, can you take a look?

* https://btx.cloud.google.com/invocations/37efb8f9-9df3-43ee-9106-8df09dbfde1d/targets Image

* https://github.com/openxla/xla/actions/runs/27240603849/job/80443182437?pr=41021

The other red ones look unrelated

Cjkkkk:
looks like hard coded reserved_shared_memory_per_block and max_blocks_per_multiprocessor is not true on old arch.

Cjkkkk:
@EusebioDM  Hi, could you please take another look to see if the issue is fixed? Thanks.

xla/backends/gpu/codegen/cudnn_test.cc
```diff
@@ -420,20 +416,127 @@ ENTRY e {
...
+  p0 = f32[32,96] parameter(0)
+  p1 = f32[96,64] parameter(1)
+  ROOT _ = f32[32,64] fusion(p0, p1), kind=kCustom, calls=fusion1,
+    backend_config={"fusion_backend_config": {kind: "__cudnn$fusion"}}
+})";
+
+  // Verify that CuDnnFusionCompiler succeeds with null dnn_support (deviceless
+  // mode), driven solely by the DeviceDescription — no live cuDNN handle.
+  {
+    TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
```
EusebioDM: You should use `ASSERT_OK_AND_ASSIGN` here, and in the other usages you added. The TF_ variants are now deprecated.

This should fix https://github.com/openxla/xla/actions/runs/28042212458/job/83011106348?pr=41021

xla/backends/gpu/target_config/cudnn_device_props.cc
```diff
@@ -0,0 +1,81 @@
...
+      desc.max_blocks_per_multiprocessor(),
+      desc.threads_per_block_limit(), desc.threads_per_core_limit(),
+      desc.registers_per_block_limit(), desc.device_memory_size(),
+      static_cast<int64_t>(desc.clock_rate_ghz() * 1e6), desc.l2_cache_size(),
+      desc.thread_dim_limit().x, desc.thread_dim_limit().y,
+      desc.thread_dim_limit().z,
+      static_cast<int64_t>(desc.mem_clock_ghz() * 1e6),
+      desc.block_dim_limit().x, desc.block_dim_limit().y,
+      desc.block_dim_limit().z,
+      /*desc.supports_coop_launch()*/ 1, driver_ver, desc.name());
```
EusebioDM, line 73: I believe you need to re-format this file to fix this [presubmits ](https://github.com/openxla/xla/actions/runs/28042212721/job/83011107145?pr=41021)

xla/backends/gpu/target_config/cudnn_device_props_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  }
+}
+
+std::string Dump(const Json::Value& v) {
+  Json::StreamWriterBuilder builder;
+  builder["indentation"] = "  ";
+  return Json::writeString(builder, v);
+}
+
+TEST(CudnnDevicePropsTest, MatchesLiveDevice) {
```
EusebioDM, line 109: Looks like these are still failing for p100 and v100 on [presubmit ](https://btx.cloud.google.com/invocations/42bda811-1514-4f65-bbae-a31512047404/targets) can you take a look?

EusebioDM (changes requested):
Thank you! Looks good, but we just need to get the presubmits green.

xla/backends/gpu/target_config/cudnn_device_props.cc
Cjkkkk: Fixed

xla/backends/gpu/codegen/cudnn_test.cc
Cjkkkk: Fixed

xla/backends/gpu/target_config/cudnn_device_props_test.cc
Cjkkkk: Ignored 2 fields that are not always serialized by cudnn. It is fine when the json converted from xla device description has more fields than what cudnn actually needs.

EusebioDM:
Thank you for the changes! LGTM, but can you rebase again? 

The Copybara import is failing due to an unrelated change that should be fixed after you rebase.

Cjkkkk:
Rebased, could you take a look? Thanks!

beckerhe (approved):

EusebioDM:
Sorry @Cjkkkk, would you able to squash all the commits together? 

Our import tool is having some problems with the PR.

Cjkkkk:
Hi @EusebioDM , it is squashed, could you take another look?

EusebioDM:
> Hi @EusebioDM , it is squashed, could you take another look?

It looks good! We are currently trying to get it past the internal CI which is being a bit flaky at the moment. We should be able to get it submitted today.

EusebioDM:
I had to fix some brittle internal tests (not in XLA), so waiting for approval on those changes internally.

PR #41055, opened 2026-04-16
Cross Host Data Transfers: Add CrossHostTransferBuffers; modify CrossHostSendBuffers to call into it

rao-ashish:
📝 Summary of Changes
This PR builds on XLA #40919, and is the next in a sequence of PRs that will refactor cross-host data transfer implementations to eventually rely on a shared helper function CrossHostTransferBuffers. CrossHostTransferBuffers is planned to eventually be integrated into the PJRT APIs to enable receiving data into preallocated receive buffers (this feature is being planned in collaboration with @gspschmid, @emilyfertig, and @pschuh).

This PR implements `CrossHostTransferBuffers` as a helper function inside `StreamExecutorGpuClient`, and modifies `StreamExecutorGpuClient::CrossHostTransferBuffers` to call into it. This PR also ensures that if async dispatch is used (`PJRT_GPU_ENABLE_ASYNC_DISPATCH=1`), `CrossHostTransferBuffers` schedules transfers through the async_dispatch_thread to prevent deadlocks.

🎯 Justification
It is difficult to achieve good comm/compute overlap with cross-host data transfers as the current implementation always allocates receive-buffers 'just-in-time', and because the GPU memory allocator blocks on the compute stream. CrossHostTransferBuffers will enable users to receive into preallocated receive buffers, making it easier to avoid the allocator blocking issue. This PR introduces CrossHostTransferBuffers as a StreamExecutorGpuClient helper function; it will eventually be exposed through the PJRT APIs.

🚀 Kind of Contribution

🧪 Unit Tests:
This PR only refactors the implementation of CrossHost{Send/Receive}Buffers, so the pre-existing unit tests for those methods already test this PR.

🧪 Execution Tests:

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -663,10 +663,7 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
   std::vector<tsl::RCReference<PjRtRawBuffer>> raw_buffers;
   raw_buffers.reserve(buffers.size());
 
-  std::vector<std::vector<tsl::RCReference<tsl::AsyncValue>>>
-      transfer_dependency_avs;
-  transfer_dependency_avs.reserve(buffers.size());
-
+  std::vector<tsl::RCReference<tsl::AsyncValue>> transfer_dependency_avs;
   std::vector<tsl::RCReference<PjRtDeviceEventPromise>> usage_event_promises;
   usage_event_promises.reserve(buffers.size());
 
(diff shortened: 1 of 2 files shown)


xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -693,67 +690,145 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
 
     // Get the local_device_state and use it to schedule transfers. Fail
     // transfers early if we cannot get the local_device_state.
-    absl::StatusOr<LocalDeviceState*> local_device_state =
+    absl::StatusOr<LocalDeviceState*> maybe_local_device_state =
         tensorflow::down_cast<PjRtStreamExecutorDevice*>(device)
             ->GetLocalDeviceState();
-    if (!local_device_state.ok()) {
-      SetEventAsError(transfer_event, local_device_state.status());
+    if (!maybe_local_device_state.ok()) {
```
emilyfertig: nit: Prefer keeping the statusor and dereferencing it vs creating a new variable: https://abseil.io/tips/181

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -693,67 +690,145 @@ StreamExecutorGpuClient::CrossHostSendBuffers(
...
-        *local_device_state, src_global_device_id, std::move(transfer_event),
-        std::move(curr_transfer_dependency_avs), std::move(transfer_specs));
+    LocalDeviceState* local_device_state = *maybe_local_device_state;
+
+    // Launch ScheduleTransfersOnLocalDevice on either the async dispatch thread
+    // of the calling thread.
+    if (local_device_state->async_dispatch_thread()) {
+      local_device_state->async_dispatch_thread()->Schedule(
+          tsl::WithCurrentContext(
+              [this, local_device_state, device_id, transfer_dependency_avs,
```
emilyfertig, line 816: std::move the transfer_dependency_avs? (and below)

Also, to make sure I understand, in this PR each buffer newly waits on all transfer dependency AVs, not just its own, right?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: Thanks, modified to keep it as a statusor

xla/pjrt/gpu/se_gpu_pjrt_client.cc
rao-ashish: We can't move `transfer_dependency_avs` here because they are reused in each iteration of the loop over `transfers_by_device`.

Yes, we're now waiting on all transfer dependencies instead of per-group dependencies. If per-transfer dependencies become necessary later on, we could add a field into `CrossHostTransferSpec` to represent those, but for now the user could split up the arrays passed into `device_put` so that if different arrays are known to become ready at different times, they are batched into separate `device_put` calls.

emilyfertig: Oops I lost track of the loop. Got it, thanks!

emilyfertig (approved):

hawkinsp (approved):

PR #40979, opened 2026-04-16
[tsl] Make BFCAllocator respect requested alignment

ezhulenev:
Add a first test for `BFCAllocator`! 😮 

`BFCAllocator` must respect required alignment, because we want to use it to allocate memory for symmetric collective operation that require 4096 byte alignment. Today we simply round up ALL allocation to 4096 bytes, and hope that BFC will return properly aligned pointer. This is very dangerous.

After this fix it will be possible to share one BFC allocator for regular XLA allocations and symmetric XLA allocation.

- Renamed `unused_alignment` → `alignment` throughout the allocation path
- `FindChunkPtr`: accepts `alignment` parameter. When a free chunk's pointer isn't aligned, computes the required padding, splits off the unaligned prefix as a free chunk, and returns the aligned remainder
- Extend: ensures sub-allocator is asked for at least `kMinAllocationSize` alignment; `CHECKs` that the returned memory meets this minimum

PR #41069, opened 2026-04-16
[xla:gpu] Unify CUDA allocators under cuMemCreate allocator

ezhulenev:
One allocator to rule them ALL 🔥 

Consolidate multiple CUDA allocator implementations into a single `CudaDeviceAllocator` backed by the CUDA VMM API, and make `BFCAllocator` respect user-specified alignment. This allows a single BFC pool per GPU to serve both default (256-byte aligned) and collective (VMM-granularity aligned) memory spaces, eliminating wasted memory from separate pools.

## Motivation

Previously, XLA maintained **separate BFC allocator pools** per GPU for default and collective memory, plus several independent allocator implementations (`CudaVmmAllocator`, `CudaCollectiveAllocator`, `NcclAllocator`, `NvshmemAllocator`). The collective pool had to be sized conservatively, and memory couldn't flow between pools. Additionally, `BFCAllocator` ignored the `alignment` parameter in `AllocateRaw`, making it impossible to share a single pool across memory spaces with different alignment requirements.

### BFCAllocator alignment support (`xla/tsl/framework/`)

Done in https://github.com/openxla/xla/pull/40979

### Unified CUDA VMM allocator (`xla/stream_executor/cuda/`)

- Consolidated `CudaVmmAllocator`, `CudaCollectiveAllocator`, `NcclAllocator` into `CudaDeviceAllocator`
- `CudaDeviceAllocator` uses `cuMemCreate`/`cuMemAddressReserve`/`cuMemMap` with configurable options: peer access, RDMA, fabric export
- Symmetric memory alignment (`GetSymmetricMemoryAlignment`) correctly returns `cuMemGetAllocationGranularity(RECOMMENDED)`

### MultiDeviceAdapter and TfAllocatorAdapter (`xla/stream_executor/integrations/`)

- `TfAllocatorAdapter` accepts `min_alignment` parameter, passes it to `BFCAllocator::AllocateRaw`
- `MultiDeviceAdapter::AllocatorInfo` is a documented aggregate struct with per-entry `min_alignment`
- `AddMemorySpaceAlias` allows routing multiple memory spaces to the same underlying allocator with different alignment
- Moved `MultiDeviceAdapter` implementation from header to `.cc` file

### Client integration (`xla/pjrt/gpu/`)

- Single BFC allocator per GPU serves both `kDefault` (256-byte aligned via `kXlaAllocatedBufferAlignBytes`) and `kCollective` (VMM-granularity aligned) memory spaces
- NVSHMEM remains a separate BFC when enabled (backed by `nvshmem_malloc`)

### Collectives cleanup (`xla/backends/gpu/collectives/`)

- Removed `GpuCollectives::Allocate/Deallocate` pure-virtual methods (now default to `Unimplemented`)
- Deleted `NcclCollectives` and `RcclCollectives` allocator methods and `SymmetricMemoryAlignment`
- Removed `ResolveGpuCollectives` and `allocate_granularity` from `gpu_executable.cc`


xla/stream_executor/cuda/cuda_device_allocator.cc
```diff
@@ -15,102 +15,244 @@ limitations under the License.
...
+// Creates a physical VMM allocation, trying fabric handle type first if
+// requested and falling back to non-fabric on failure.
+static absl::StatusOr<CUmemGenericAllocationHandle> CreatePhysicalAllocation(
+    const CUmemAllocationProp& properties, uint64_t padded_size,
+    bool enable_fabric_export) {
+  CUmemGenericAllocationHandle handle;
+
+  if (enable_fabric_export) {
+    CUmemAllocationProp fabric_properties = properties;
+    fabric_properties.requestedHandleTypes =
```
Tixxx: this is exactly what nccl is doing though. Why not just calling ncclMemAlloc? You get all these handle checks for free

ezhulenev: Some people build XLA without NCCL, and AFAIK it also adds to linking hell in Tensorflow, so it's easier to duplicate code here than to deal with that.

xla/stream_executor/cuda/cuda_executor.h
```diff
@@ -37,15 +37,14 @@ limitations under the License.
 #include "absl/synchronization/mutex.h"
 #include "absl/time/time.h"
 #include "absl/types/span.h"
+#include "third_party/gpus/cuda/include/cuda.h"
```
PatriosTheGreat, line 40: Is this needed?

xla/stream_executor/cuda/cuda_executor.cc
```diff
@@ -766,44 +772,49 @@ CudaExecutor::VmmMemoryHandle& CudaExecutor::VmmMemoryHandle::operator=(
...
+  if (gpu_collectives == nullptr) {
+    return absl::InternalError("Failed to get NVSHMEM collectives");
   }
+  return gpu_collectives;
+}
 
+// Deprecated: nvshmem backward compatibility path.
+static absl::StatusOr<void*> NvshmemCollectiveMemoryAllocate(
+    StreamExecutor* executor, uint64_t bytes) {
+  if (bytes == 0) return nullptr;
```
PatriosTheGreat: Auto-formatter complains:

if (bytes == 0) {
   return nullptr;
}

xla/stream_executor/cuda/cuda_device_allocator.cc
```diff
@@ -15,102 +15,254 @@ limitations under the License.
...
+      &granularity, &properties, CU_MEM_ALLOC_GRANULARITY_RECOMMENDED)));
+
+  size_t effective_alignment = std::max(options.alignment, granularity);
+  uint64_t padded_size = xla::RoundUpTo<uint64_t>(size, effective_alignment);
+
+  ASSIGN_OR_RETURN(CUmemGenericAllocationHandle handle,
+                   CreatePhysicalAllocation(properties, padded_size));
+
+  absl::Cleanup release_handle = [&] {
+    cuda::ToStatus(cuMemRelease(handle)).IgnoreError();
```
PatriosTheGreat: Can we log an error in this case?

xla/stream_executor/cuda/cuda_device_allocator.cc
```diff
@@ -15,102 +15,254 @@ limitations under the License.
...
+  };
+
+  // Reserve virtual address range and map the physical allocation into it.
+  CUdeviceptr ptr;
+  RETURN_IF_ERROR(cuda::ToStatus(
+      cuMemAddressReserve(&ptr, padded_size, effective_alignment, 0, 0)));
+  absl::Cleanup free_address = [&] {
+    cuda::ToStatus(cuMemUnmap(ptr, padded_size)).IgnoreError();
```
PatriosTheGreat: Also here, can we log errors?

xla/stream_executor/cuda/cuda_device_allocator_test.cc
```diff
@@ -92,5 +95,39 @@ TEST(CudaDeviceAllocatorTest, MemcpyRoundTrip) {
   EXPECT_EQ(host_src, host_dst);
 }
 
+TEST(CudaDeviceAllocatorTest, AllocateWithFabricHandle) {
```
PatriosTheGreat, line 100: This test case fails for me with an error:
cuda_device_allocator_test.cc:110
allocator.Allocate(4096) returned error: INTERNAL: CUDA error: : CUDA_ERROR_INVALID_VALUE: invalid argument

On H100 machine. Can you reproduce this?

xla/stream_executor/cuda/cuda_executor.cc
```diff
@@ -29,7 +29,6 @@ limitations under the License.
 #include <utility>
```
PatriosTheGreat, line 29: These changes seems to break cuda_executor_multigpu_test.cc specifically 
CudaExecutorMultiGpuTest.CudaMulticastMemoryUsingNonVmmMemory case

PatriosTheGreat, line 29: These changes seems to break cuda_executor_multigpu_test.cc specifically 
CudaExecutorMultiGpuTest.CudaMulticastMemoryUsingNonVmmMemory case

xla/stream_executor/cuda/cuda_platform.cc
```diff
@@ -29,9 +28,8 @@ limitations under the License.
 #include "third_party/gpus/cuda/include/cuda.h"
 #include "third_party/gpus/cuda/nvml/include/nvml.h"
 #include "third_party/gpus/cudnn/cudnn_version.h"
-#include "xla/debug_options_flags.h"
+#include "cub/version.cuh"
```
PatriosTheGreat, line 31: Is this needed?

PatriosTheGreat (changes requested):

xla/stream_executor/cuda/cuda_executor.h
ezhulenev: Yes, to silence warnings from `CUdevice` members

xla/stream_executor/cuda/cuda_executor.cc
ezhulenev: Done

xla/stream_executor/cuda/cuda_device_allocator.cc
ezhulenev: Done

xla/stream_executor/cuda/cuda_device_allocator.cc
ezhulenev: Done

xla/stream_executor/cuda/cuda_platform.cc
ezhulenev: Yes, we do use constants from that theader.

xla/stream_executor/cuda/cuda_executor.cc
ezhulenev: Fixed

xla/stream_executor/cuda/cuda_device_allocator_test.cc
ezhulenev: Added a check for fabric support + for next test added a check for presence of 2 GPUs and peer access actually working.

ezhulenev:
@PatriosTheGreat pushed new version, now it passes all tests on my machine

xla/stream_executor/cuda/cuda_device_allocator.cc
```diff
@@ -15,102 +15,266 @@ limitations under the License.
...
+
+  XLA_VLOG_DEVICE(3, executor->device_ordinal())
+      << "Deallocating " << ptr << " padded size: " << padded_size;
+
+  // Unlike cuMemFree which defers until in-flight kernels complete, cuMemUnmap
+  // immediately invalidates the virtual address mapping. Synchronize to ensure
+  // no kernels are still accessing this memory. This is fine, because we never
+  // use this allocator on a hot path and always wrap it into BFCAllocator that
+  // does arena-based allocation.
+  absl::Status status = cuda::ToStatus(cuCtxSynchronize());
```
shawnwang18, line 194: I don't get why it is not on a hot path.  The allocator might be used for nccl symmetric memory, or default allocation if update-free cuda-graph is enabled. 


VMM Deallocation is happened after kernels on each step, if it is blocked here, then all later  operations will be blocked:  
1. ) all later deallocations if it is not the last. 
2. ) the allocations for next step.
3. ) the cuda-graph update for next step which should be performed during kernel running of current step. 

All above operations are assumed to run during the previous step's kernel running window. 

I think this will create large bubbles across steps, right?  We should not introduce a synchronization point for memory allocator.  otherwise it is very likely that CPU's overhead is exposed on GPU.

ezhulenev: Memory allocation happens once when SE initialized and feed when SE destroyed, on hot path we create buffers using BFC on top of physical allocation. Se::Allocate should never be called directly without wrapping in something else. And XLA carefully avoids it

shawnwang18: Ah, you mean it just happened once for the BFC Allocator's memory pool, for VMM allocator, then this will mean that the VMM allocation's physical->virtual mapping is fixed the whole step, you will have no way to remap a physical address to another virtual address, right?  

Because `cuMemMap` does not support non-0 physical memory offset, I was thinking that if you use VMM to allocate memory, then the motivation is to support the re-mapping, which is required for cuda graph or nccl symmetric memory.

ezhulenev: Hmmm, nccl symmetric memory works just fine, not sure how exactly it is implemented.  cuda graph remapping - yeah, this will be a problem with 0 offset constraint

PatriosTheGreat (approved):

akuegel:
@ezhulenev We have a test that fails with OOM: https://github.com/tensorflow/tensorflow/blob/d59cb911faecdf8fe9da2687083dbf42f3d157f5/tensorflow/compiler/tests/concat_ops_test.py#L307

Can you check what is going on, and if this can be avoided? It seems this test was specifically added to ensure that we don't run out of memory with too many arguments.

ezhulenev:
@akuegel I can reproduce it locally, this is bad interaction of VMM granualrity and TF allocators, working on a fix.

ezhulenev:
Will split it into pieces and send separate PRs.

PR #41114, opened 2026-04-17
[ROCm] Use rocprim::segmented_inclusive_scan for batched row scan

magaonka-amd:
📝 Summary of Changes
- Replace hand-rolled BlockScanKernel with rocprim::segmented_inclusive_scan using counting + transform iterators to generate segment offsets without device memory allocation
- Remove dependency on rocprim::detail::default_scan_config_base

🎯 Justification
This fixes build breakage in TheROCK 7.13 and helps XLA to move away from rocprim::detail

🚀 Kind of Contribution


magaonka-amd:
will close this because I opened duplicate PR.

PR #41116, opened 2026-04-17
[ROCm] Enable usage of dpp and swizzle for gpu.shuffle.

zoranjovanovic-ns:
📝 Summary of Changes
Intoduce new PromoteShuffleDownToDPP pass to lower gpu.shuffle with amdgpu.dpp and amdgpu.swizzle when beneficial.

🎯 Justification
dpp and swizzle are faster then bpermute which was generated until now

🚀 Kind of Contribution
Please remove what does not apply: 🐛 ⚡️ Performance Improvement

🧪 Unit Tests:
xla/backends/gpu/codegen/emitters/transforms/tests/promote_shuffle_to_dpp.mlir

🧪 Execution Tests:


zoranjovanovic-ns:
Converted to draft because of performance regressions on some kernels.

olegshyshkov (approved):

xla/backends/gpu/codegen/emitters/transforms/promote_shuffle_to_dpp.cc
```diff
@@ -0,0 +1,184 @@
...
+    std::optional<int64_t> offset = getConstantIntValue(op.getOffset());
+    if (!offset || *offset != 16) {
+      return rewriter.notifyMatchFailure(
+          op, "only offset 16 is supported for swizzle promotion");
+    }
+
+    mlir::Location loc = op.getLoc();
+    mlir::Type result_type = op.getShuffleResult().getType();
+
+    mlir::Value swizzle = mlir::amdgpu::SwizzleBitModeOp::create(
```
draganmladjenovic, line 225: For gfx950 it is V_PERMLANE16_SWAP_B32 and for gfx10+ V_PERMLANEX16_B32 is a way to go. Offset 32 can use bpermute with idx threadIdx.x * 4 + 128 which should provide better codegen than default also. So that can be handled. One just need to make sure that the block sizes are N * warpsize.

xla/backends/gpu/codegen/emitters/transforms/promote_shuffle_to_dpp.cc
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// Promotes gpu.shuffle down ops with constant offsets in [1, 15] to
+// amdgpu.dpp row_shr ops. DPP row_shr shifts data right within 16-lane rows,
+// which is correct for the reduction use case where only lane 0's result
+// matters and all relevant source lanes are within the same row.
+//
+// The pattern only fires when the validity predicate (second result) has no
+// users, which is always the case for XLA's shuffle-reduce lowering.
+struct PromoteShuffleDownToDPP
```
draganmladjenovic, line 115: You can make producer emit shuffle with subgroup of 16. That exactly matches dpp semantics. So there is no need to worry if we match the right thing. Same can be done for swizzle (32).

xla/backends/gpu/codegen/emitters/transforms/promote_shuffle_to_dpp.cc
zoranjovanovic-ns: Improvement for gfx950 will be in separate PR.

zoranjovanovic-ns:
@xla-rotation can someone take a look?

dimitar-asenov (approved):

xla/backends/gpu/codegen/emitters/transforms/promote_shuffle_to_dpp.cc
```diff
@@ -0,0 +1,259 @@
...
+    std::optional<int64_t> width = getConstantIntValue(op.getWidth());
+    if (!width) {
+      return rewriter.notifyMatchFailure(op,
+                                         "width must be a constant integer");
+    }
+
+    // Skip DPP promotion when the enclosing function is VALU-bound due to
+    // special-function ops; leave the shuffle as ds_bpermute in that case.
+    auto func = op->getParentOfType<mlir::FunctionOpInterface>();
+    if (func && FuncHasSpecialFunctionOps(func)) {
```
draganmladjenovic, line 151: Then why not use ds_swizzle or better ds_bpermute indexing than just bailing on dpp? Do you have some exaples of thease "valu bound" kernels?

PR #41131, opened 2026-04-17
[xla:gpu] Pass non-const collective cliques and memory to FFI

ezhulenev:
FFI handlers need non-cost pointers to be able to call `Tie` API on them. This is compatible with how cliques and memory passed to regular thunks

### DIFF
--- a/xla/backends/gpu/ffi.h
+++ b/xla/backends/gpu/ffi.h
@@ -29,7 +29,7 @@ limitations under the License.
 #include "xla/backends/gpu/runtime/collective_params.h"
 #include "xla/ffi/api/c_api.h"
 #include "xla/ffi/api/c_api_internal.h"  // IWYU pragma: keep
-#include "xla/ffi/ffi.h"  // IWYU pragma: export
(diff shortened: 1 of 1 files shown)


xla/backends/gpu/ffi.h
```diff
@@ -29,7 +29,7 @@ limitations under the License.
 #include "xla/backends/gpu/runtime/collective_params.h"
 #include "xla/ffi/api/c_api.h"
 #include "xla/ffi/api/c_api_internal.h"  // IWYU pragma: keep
-#include "xla/ffi/ffi.h"  // IWYU pragma: export
+#include "xla/ffi/ffi.h"                 // IWYU pragma: export
```
ermilovmaxim, line 32: Seems accidental

ermilovmaxim (approved):

xla/backends/gpu/ffi.h
ezhulenev: it is a clang-format difference between g3 and oss

PR #41132, opened 2026-04-17
[GPU] Consolidate GPU device memory to a unified CUDA-graph-stable pool

terryysun:
📝 Summary of Changes
Consolidated GPU device memory allocation under VMM backed allocator, added NCCL symmetric memory compatible logics and SPMD tag for reuse management.

🎯 Justification
Unifying device memory gets rid of the manual memory fraction, and completely lifts the dependency to `cudaMemAlloc`. Unifying under `DeviceAddressVmmAllocator` keeps address stable and avoid CUDA graph updates across iterations.

🚀 Kind of Contribution
✨ New Feature

N/A.

🧪 Unit Tests:
Added.

🧪 Execution Tests:
Added.


terryysun:
Hi @PatriosTheGreat could you take a look at this PR? the VMM consolidation is behind the VMM allocator flag and off by default, it should not break any existing workflow.

xla/backends/cpu/ynn_support.h
PatriosTheGreat, line 77: Is this related?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -2201,6 +2203,10 @@ StreamExecutorGpuClient::RunAsync(
         gpu::MemorySpaceColor::kCollective)] = collective_memory_alignment;
   }
 
+  // Detect SPMD: multi-device computations need deterministic VMM TryReuse.
+  auto* da = run_options->run_options().device_assignment();
```
PatriosTheGreat: We usually prefer to use full names for variable names and in this case it's better not to use auto since it's not obvious from the right side

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -2201,6 +2203,10 @@ StreamExecutorGpuClient::RunAsync(
...
+  bool spmd = da && da->replica_count() * da->computation_count() > 1;
```
PatriosTheGreat: How this will work in case when we have 2 JAX programs: one works on 2 local devices and the next one works on 4 of them? Like this:

```

def matmul(x):
  return x @ x.T

x = jnp.arange(16 * 16, dtype=jnp.float32).reshape((16, 16))
two_devices = jax.devices()[:2]
mesh = jax.sharding.Mesh(two_devices, ['x'])
two_devices_result = jax.jit(
    shard_map.shard_map(
        matmul,
        mesh=mesh,
        in_specs=P('x'),
        out_specs=P('x'),
        check_rep=False,
    ))(x).block_until_done()

four_devices = jax.devices()[:4]
mesh = jax.sharding.Mesh(four_devices, ['x'])
four_devices_result = jax.jit(
    shard_map.shard_map(
        matmul,
        mesh=mesh,
        in_specs=P('x'),
        out_specs=P('x'),
        check_rep=False,
    ))(x).block_until_done()
```


Wouldn't vmm device address allocator try to reuse the memory from 2 devices run, which is not setup properly for 4 devices?

xla/service/gpu/gpu_executable.cc
```diff
@@ -1195,6 +1197,9 @@ absl::StatusOr<BufferAllocations> GpuExecutable::GenerateBufferAllocations(
         collectives->SymmetricMemoryAlignment();
   }
 
+  auto* da = run_options->run_options().device_assignment();
+  bool spmd = da && da->replica_count() * da->computation_count() > 1;
```
PatriosTheGreat: I guess SPMD here means an HLO executed on several devices (aka collective module). Can we call it everywhere like that?

xla/stream_executor/cuda/cuda_raw_memory_allocation.cc
```diff
@@ -31,20 +31,36 @@ limitations under the License.
 #include "tsl/platform/statusor.h"
 
 namespace stream_executor::gpu {
-namespace {
 
-CUmemAllocationProp BuildAllocationProperties(CUdevice device) {
+// Builds allocation props with shareable handles for NCCL symmetric memory.
+CUmemAllocationProp BuildVmmAllocationProperties(CUdevice device) {
```
PatriosTheGreat: This logic duplicates  compiler/xla/stream_executor/cuda/cuda_vmm_allocator.cc. Can you use that one here instead or unify them?

xla/stream_executor/cuda/cuda_raw_memory_allocation.cc
```diff
@@ -31,20 +31,36 @@ limitations under the License.
...
+                           CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED,
+                           device) == CUDA_SUCCESS &&
+      fabric_supported) {
+    handle_types |= CU_MEM_HANDLE_TYPE_FABRIC;
+  }
+#endif
   props.requestedHandleTypes =
-      static_cast<CUmemAllocationHandleType>(CU_MEM_HANDLE_TYPE_NONE);
+      static_cast<CUmemAllocationHandleType>(handle_types);
+  int rdma_supported = 0;
```
PatriosTheGreat: Is this currently support cross-host usage with MNNVL? If so is there memory handlers exchange and export/import somewhere?

xla/stream_executor/vmm_device_address_allocator.cc
```diff
@@ -448,6 +453,7 @@ DeviceAddressVmmAllocator::TryReusePendingDeallocation(PerDeviceState& state,
         reused_mem.opaque(), it->mem.size(), size, rounded_size,
         state.executor->device_ordinal());
     state.pending_deallocations.erase(it);
+    if (spmd) state.spmd_allocations.insert({reused_mem.opaque(), true});
```
PatriosTheGreat: Simply checking if the memory was for collective might not be enough, since JAX and XLA allows users to run modules with different amount of devices. Also adding spmd flag (or any information regarding collective usage of memory) to allocator doesn't seems to be correct.
Ideally allocator should just allocate a physical memory. Then a runtime should be able to setup that memory with VA in a way that it can be used with NCCL or any collective kernels.

PatriosTheGreat (changes requested):

xla/backends/cpu/ynn_support.h
terryysun: merge issue, fixed.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
terryysun: renamed.

xla/stream_executor/cuda/cuda_raw_memory_allocation.cc
terryysun: Thanks! consolidated.

xla/service/gpu/gpu_executable.cc
terryysun: Yea I was also trying to come up with a better naming. Collective is a bit ambiguous since not all those allocations are for collectives. Renamed to multi_device, does it sound better?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
terryysun: If NCCL symmetric memory is not enable, allocations in 2-device collective can be safely reused by allocations in 4-device collective. If NCCL symmetric memory is enabled, the second executable will trigger fresh registration because of CollectiveMemoryCache cache miss because each executable has its own CollectiveMemoryCache. Even if the two collectives are in the same executable, thus share the same CollectiveMemoryCache, because CollectiveMemoryCache is key'd by clique, it's still a cache miss thus triggering new registration.

xla/stream_executor/cuda/cuda_raw_memory_allocation.cc
terryysun: Yes, multi-host/MNNVL are supported, handle exchange/import/export are all handled by NCCL.

xla/stream_executor/vmm_device_address_allocator.cc
terryysun: Running modules with different devices would trigger new symmetric window registration, the spmd filter is to guarantee the correctness of the allocations that go into the same symmetric window.
We can have runtime handle the setup by keying CollectiveMemoryCache with allocation index of slice, but that would also affect the default code path. The current approach is trying to keep the blast radius minimal before we are confident that vmm allocator is as reliable as bfc.

terryysun:
Hi @PatriosTheGreat could you take another look? Thanks!

xla/stream_executor/vmm_device_address_allocator.cc
PatriosTheGreat: But then what's the problem of adding single-device memory to the "release queue"?
Let's say we launched module on 1 device. It allocated/deallocated 4 kb of memory for scratch buffer. Next we launched a module on 2 devices which also using 4 kb scratch memory. What's the problem of getting the previous chunk of memory from the first module and configuring it with nccl register window?

I thought all this separation between spmd/non-spmd memory is needed to prevent memory re-configuration during the next run, but I guess I was wrong.


Overall my concern here is that since we have a mechanism which can configure memory buffer for any slice of devices (nccl window register/deregister) then why allocator should know which type of memory it allocates?

xla/service/gpu/gpu_executable.cc
PatriosTheGreat: I guess multi-device makes more sense here

PatriosTheGreat (changes requested):

xla/pjrt/gpu/se_gpu_pjrt_client.cc
PatriosTheGreat: But then why can't single device allocation also be reused for multi-device run and trigger registration?

terryysun: NCCL symmetric window registration contains a bootstrap all-gather, so every rank must enter registration together. Taking the example below, i.e. 2-device allocation then a single device allocation with the same size. The first iteration will be fine, but the next iteration if the 2-device allocation picks up the allocation from the 1-device allocation in the first iteration, rank ) will enter registration while rank 1 can pick up the 2-device allocation from last iteration and doesn't enter registration, then they will deadlock.

xla/stream_executor/vmm_device_address_allocator.cc
terryysun: Thanks! The problem is with the registration process, asymmetric allocation sequence -> asymmetric VAs -> asymmetric cache hit/miss behavior on different ranks -> asymmetric registration all-gather calls across ranks -> deadlock, so we need a mechanism to separate non-multi-device ones.

PatriosTheGreat: Let's continue this discussion in other thread

xla/pjrt/gpu/se_gpu_pjrt_client.cc
PatriosTheGreat: Hmm not sure I understood this. The run_options->run_options().device_assignment() is an HLO level constant, so all allocations within a single HLO execution will either be spmd (multi-device) or non-spmd (single-device), am I correct here?

If the above is true then in your sample I guess you have a multi-device module followed by single device module which are using allocations of the same size. But since these modules are different from each other isn't this situation fully equals to the situation when we have 2 devices module followed by 4 devices module (also which are using allocations of the same size)? I though that in this case since it's a different gpu_executable it doesn't matter which memory XLA picked up it still should re-register it. I guess I'm missing something here.

terryysun: Right, they are either spmd or non-spmd. The filter targets cross-executable interference. In practice we don't see workloads alternating executables with different participant sets, the typical pattern is single-device side computations (LR scheduler, PRNG) between full-mesh training steps, which the filter covers. Without the current setup, existing benchmarks will deadlock, and if we see workloads that require repeated mix-sized collectives, we need to re-key the cache.

PatriosTheGreat: Thanks for explanation, I think I understood the problem now.
I guess in theory you might need to have a dedicated release queue for a any possible clique, but for now you are good with single-device and multi-device queues.

I still think that adding a new parameter to allocator interface is not the best place to propagate this information, since:

1. Allocators should not care how the memory they allocated is going to be used.

2. The spmd/non-spmd property is a module-level property, and it's incorrect to set it differently for different allocations within a single module.

I think instead of additional parameter we would need to have some mechanism here which will allow you to pass device_assignment to vmm_device_address_allocator from se_gpu_pjrt_client. In order to make this change smaller I think you can add a new setter method to vmm_device_address_allocator like "set_device_assignment" and in  se_gpu_pjrt_client you can check if the allocator type is vmm_device_address_allocator then call this setter. I think for this case dynamic downcast is actually better then adding a parameter for all allocators.

terryysun: Thanks for the detailed suggestions! Refactored the multi_device filtering to a module level context with `DeviceAddressVmmAllocator::DeviceAssignmentScope`. The per-allocation multi_device filter is dropped.

PatriosTheGreat (approved):
Thanks!

xla/pjrt/gpu/se_gpu_pjrt_client.cc
PatriosTheGreat: Thank! Approved this change

thcmbs:
Hi, I'm working on merging the PR, sorry for the delay. Canyou please rebase and please use RETURN_IF_ERROR/ASSIGN_OR_RETURN over of TF_... variants.

terryysun:
> Hi, I'm working on merging the PR, sorry for the delay. Canyou please rebase and please use RETURN_IF_ERROR/ASSIGN_OR_RETURN over of TF_... variants.

Thanks! rebased and updated the return variants.

dimitar-asenov (approved):

PR #41165, opened 2026-04-18
[xla:gpu] Map whole allocation range to symmetric memory

ezhulenev:
When the BFC allocator carves sub-allocations from a pre-allocated collective memory buffer, different runs can get sub-allocations at different offsets (e.g. if allocator state diverges across devices). Previously, `AcquireSymmetricMemory` registered only the sub-allocation as a symmetric window, keyed by its address. This caused `CollectiveMemoryCache` misses on subsequent runs (different sub-offset → different key), leading to redundant `CreateSymmetricMemory` calls and deadlocks when ranks disagreed on whether to call the collective registration.

Fix: Resolve each buffer allocation to its full underlying device address range (via `GetMemoryAddressRange` / `cuMemGetAddressRange`) and register the entire range as the symmetric window. The cache key is now the full range, so all sub-allocations within the same BFC chunk share one symmetric window. Per-allocation offsets from the window
  base are pre-computed and stored in SymmetricMemoryEntry.

  - `AcquireSymmetricMemory`: Three-pass structure — resolve ranges, deduplicate, then populate with offsets. Multiple sub-allocations from the same range are registered once
  - `CollectiveMemory`: Add `SymmetricMemoryEntry` struct (symmetric_memory + offset)

Added e2e test that used to deadlock without the fix.


PatriosTheGreat:
This solution might have 2 potential problems:

1. The memory on different devices during module execution might be different, since some of devices might hold output buffers of previous modules. Consider the following sample:

```
@jax.jit
  def f(x):
    return jax.lax.batch_matmul(x, x)

x = ...
gpus = jax.devices('gpu')
two_devices = gpus[:2]
result_2_devices = jax.pmap(f, axis_name='i', devices=two_devices)(x)

four_devices = gpus[:4]
result_4_devices = jax.pmap(f, axis_name='i', devices=four_devices)(x)
```

Here during the second module execution first 2 devices would hold a buffer for result_2_devices.

2. Some JAX clients are not using BFC preallocated memory to run several modules in parallel. For example this is used in JAX tests to run them in parallel. In this mode there might be situation when two input parameters are allocated with a single physical allocation in one device and with two allocations on another. In this case we can't use a single symmetrical memory to map them.

ezhulenev:
True, this is not a viable approach.

PR #41167, opened 2026-04-18
Fix shape-dependent erf precision by adding saturation select

### LINKED ISSUE #41122: `erf` produces different fp32 result for the same input based on tensor shape
## Summary

XLA's `erf` lowering is not shape-invariant: the same fp32 input produces a different output depending on the innermost axis size of the enclosing tensor. For `x = 4.5`:

- Shape with innermost dim **= 1** (e.g. `[1]`, `[7, 64, 1]`) → `erf(x) = 0.9999998212`
- Shape with innermost dim **≥ 2** (e.g. `[2]`, `[7, 64, 5]`) → `erf(x) = 1.0` (correctly rounded)

The two outputs differ by ~3 fp32 ULPs. A single downstream `floor` / `ceil` / `>=` call amplifies this into an integer-valued, silent divergence.

The bug reproduces identically in JAX, which confirms the root cause lives in XLA's `erf` HLO lowering (not in TensorFlow's XLA bridge).

## Versions tested

| stack | version | reproduces? |
|---|---|---|
| TensorFlow | 2.21.0 (stable, 2026-03-06) | ✓ |
| TensorFlow | tf-nightly `2.22.0-dev20260417` | ✓ |
| JAX | 0.9.2 | ✓ |

Linux x86_64; reproduced on both CPU-only and CUDA 12 / NVIDIA T4 configurations.

## Minimal reproducer (TensorFlow, self-contained)

```python
import os
os.environ['TF_CPP_MIN_LOG_LEVEL'] = '2'
import tensorflow as tf

@tf.function(jit_compile=True)
def xla_erf(x):
    return tf.math.erf(x)

print(f"TF version: {tf.__version__}")
for shape in [(1,), (2,), (7, 1), (7, 64, 1), (7, 64, 5), (1, 1, 1)]:
    x = tf.fill(shape, tf.constant(4.5, dtype=tf.float32))
    e  = tf.math.erf(x).numpy()
    xl = xla_erf(x).numpy()
    print(f"shape {str(shape):>13}  eager={float(e.flat[0]):.10f}  XLA={float(xl.flat[0]):.10f}  match={float(e.flat[0]) == float(xl.flat[0])}")
```

(The same result holds for `tf.raw_ops.Erf`.)

Expected (what `math.erf(4.5)` rounds to in fp32): `1.0` for every shape.
Observed on TF 2.21.0 and tf-nightly `2.22.0-dev20260417`:

```
shape          (1,)  eager=1.0000000000  XLA=0.9999998212  match=False
shape          (2,)  eager=1.0000000000  XLA=1.0000000000  match=True
shape        (7, 1)  eager=1.0000000000  XLA=0.9999998212  match=False
shape    (7, 64, 1)  eager=1.0000000000  XLA=0.9999998212  match=False
shape    (7, 64, 5)  eager=1.0000000000  XLA=1.0000000000  match=True
shape     (1, 1, 1)  eager=1.0000000000  XLA=0.9999998212  match=False
```

Exactly the same pattern appears when we bypass TF entirely and go through JAX:

## Cross-framework reproducer (JAX, confirms the bug is in XLA)

```python
import jax, jax.numpy as jnp
from jax import lax
import numpy as np

@jax.jit
def jit_erf(x):
    return lax.erf(x)

print(f"jax: {jax.__version__}")
for shape in [(1,), (2,), (7, 1), (7, 64, 1), (7, 64, 5), (1, 1, 1)]:
    x = jnp.full(shape, 4.5, dtype=jnp.float32)
    e  = np.asarray(lax.erf(x)).flat[0]
    xl = np.asarray(jit_erf(x)).flat[0]
    print(f"shape {str(shape):>13}  eager={float(e):.10f}  jit={float(xl):.10f}")
```

Output on jax 0.9.2:

```
shape          (1,)  eager=0.9999998212  jit=0.9999998212
shape          (2,)  eager=1.0000000000  jit=1.0000000000
shape        (7, 1)  eager=0.9999998212  jit=0.9999998212
shape    (7, 64, 1)  eager=0.9999998212  jit=0.9999998212
shape    (7, 64, 5)  eager=1.0000000000  jit=1.0000000000
shape     (1, 1, 1)  eager=0.9999998212  jit=0.9999998212
```

Every XLA-backed path (TF+jit, JAX-eager, JAX+jit) agrees on the shape-dependent output. Only TensorFlow's non-XLA Eigen kernel (used in TF eager) returns the correctly-rounded `1.0` for every shape.

| shape | TF eager (Eigen) | TF `jit_compile=True` (XLA) | JAX `lax.erf` (XLA) | JAX `jit(lax.erf)` (XLA) |
|---|---|---|---|---|
| `(1,)`        | `1.0000000000` | `0.9999998212` | `0.9999998212` | `0.9999998212` |
| `(2,)`        | `1.0000000000` | `1.0000000000` | `1.0000000000` | `1.0000000000` |
| `(7, 64, 1)`  | `1.0000000000` | `0.9999998212` | `0.9999998212` | `0.9999998212` |
| `(7, 64, 5)`  | `1.0000000000` | `1.0000000000` | `1.0000000000` | `1.0000000000` |

This rules out TF's XLA bridge as the source: same XLA, different frontend, same bug.

## Suggested fix

Apply the same `erf` polynomial that XLA already uses for innermost-dim ≥ 2 when the innermost dim is 1. There is no algorithmic reason the accuracy should drop for size-1 innermost tensors — the same approximation that produces `1.0` for shape `(2,)` can be applied to shape `(1,)`.

kredd2506:
## Summary
Fixes #41122

The scalar erf lowering in `expand_float_ops.cc` clamps input `x` to `[-kErfInvOneMinusHalfULP, +kErfInvOneMinusHalfULP]` and evaluates the rational polynomial, producing `0.9999998212` at the boundary instead of `1.0` (3 f32 ULPs off). The vector path in [`xla/codegen/intrinsic/erf.cc`](https://github.com/openxla/xla/blob/main/xla/codegen/intrinsic/erf.cc) already handles saturation correctly via a `select`; this adds the matching select to the scalar path.

## Context

This was diagnosed collaboratively in #41122:

- @wuyii8941 reported the shape-dependent precision issue and provided a side-by-side LLVM IR dump that isolated the divergence to the scalar-vs-vector lowering paths
- Follow-up investigation narrowed down that the polynomial coefficients are identical across both paths — only the saturation strategy differs
- @wuyii8941 confirmed a saturation select after [line 107](https://github.com/openxla/xla/blob/main/xla/codegen/emitters/transforms/expand_float_ops.cc#L107) as the correct fix direction

The [comment on line 83-84](https://github.com/openxla/xla/blob/main/xla/codegen/emitters/transforms/expand_float_ops.cc#L83-L84) documents the intended behavior (*"outside of which x should be +/-1"*) but the implementing code never enforced the `±1` fallback.

## Approach

Compute the saturation check on the original unclamped `x` before the clamp is applied:

```cpp
Value abs_x = math::AbsFOp::create(b, original_x);
Value saturates = ma::CmpFOp::create(b, CmpFPredicate::OGE, abs_x,
                                     c(kErfInvOneMinusHalfULP));
Value saturated_value = math::CopySignOp::create(b, c(1.0f), original_x);
// ... existing clamp + polynomial evaluation ...
rewriter.replaceOpWithNewOp<SelectOp>(op, saturates, saturated_value,
                                      poly_result);
```

For `x = 4.5`: `|4.5| >= 3.7439` → select returns `copysign(1.0, 4.5) = 1.0`.
For in-range `x`: select picks the polynomial result, unchanged from before.

## Testing

- Added CHECK lines to `expand_float_ops.mlir` verifying the lowering contains `math.absf`, `arith.cmpf oge`, `math.copysign`, and `arith.select`
- All 19 tests in `//xla/codegen/emitters/transforms/tests:tests` pass locally (`bazel test //xla/codegen/emitters/transforms/tests:tests`)

cc @wuyii8941 — you flagged this and confirmed the fix direction in the issue, feel free to take a look if interested.

### DIFF
--- a/xla/codegen/emitters/transforms/expand_float_ops.cc
+++ b/xla/codegen/emitters/transforms/expand_float_ops.cc
@@ -98,13 +98,26 @@ struct RewriteErf32Pattern : public mlir::OpRewritePattern<mlir::math::ErfOp> {
       return r;
     };
 
-    Value x = op.getOperand();
-    x = ma::MaximumFOp::create(b, x, c(-kErfInvOneMinusHalfULP));
+    Value original_x = op.getOperand();
+    // For |x| >= kErfInvOneMinusHalfULP, erf(x) rounds to ±1 in f32, so
+    // return copysign(1, x) instead of evaluating the polynomial on the
+    // clamped input. Without this the polynomial evaluated at the clamp
+    // boundary gives ~0.9999998 (3 ULPs short of 1.0) for large x.
+    Value abs_x = mlir::math::AbsFOp::create(b, original_x);
+    Value saturates = ma::CmpFOp::create(b, ma::CmpFPredicate::OGE, abs_x,
+                                         c(kErfInvOneMinusHalfULP));
+    Value saturated_value =
+        mlir::math::CopySignOp::create(b, c(1.0f), original_x);
+
+    Value x = ma::MaximumFOp::create(b, original_x,
+                                     c(-kErfInvOneMinusHalfULP));
     x = ma::MinimumFOp::create(b, x, c(kErfInvOneMinusHalfULP));
     Value x2 = ma::MulFOp::create(b, x, x);
+    Value poly_result = ma::DivFOp::create(
+        b, ma::MulFOp::create(b, x, poly(x2, kAlpha)), poly(x2, kBeta));
 
-    rewriter.replaceOpWithNewOp<ma::DivFOp>(
-        op, ma::MulFOp::create(b, x, poly(x2, kAlpha)), poly(x2, kBeta));
+    rewriter.replaceOpWithNewOp<SelectOp>(op, saturates, saturated_value,
+                                          poly_result);
 
     return mlir::success();
   }
--- a/xla/codegen/emitters/transforms/tests/expand_float_ops.mlir
+++ b/xla/codegen/emitters/transforms/tests/expand_float_ops.mlir
@@ -28,6 +28,13 @@ module {
 
 // CHECK-LABEL: @erf
 // CHECK-NOT: erf
+// The lowering must include a saturation select so that erf(x) returns
+// exactly ±1 for |x| >= kErfInvOneMinusHalfULP, instead of evaluating the
+// polynomial on the clamped input (which gives ~0.9999998 for large x).
+// CHECK: math.absf
+// CHECK: arith.cmpf oge
+// CHECK: math.copysign
+// CHECK: arith.select
 
 // -----
 


akuegel (approved):

kredd2506:
Trying to build a mental model of the CI signal here — could use a sanity check. The `JAX Linux x86 AMD Instinct GPU` check is failing on the Copybara import (#41202), but the log shows the failures are Triton autotuner errors on f64 GEMMs (`No configs could be compiled`) in tests like `CustomJVPTest.test_vmap_inside_defjvp` and `EighTest.testEighIdentity*`, none of which look related to erf. I also checked a few other recent PRs (#41205, #41200, #41198, #41196) and the same check is failing there too, so I'm guessing this is a pre-existing flake or missing f64 Triton config on ROCm rather than something I broke. But I'm new to this CI setup — is that reading right, or is there something I should dig into on my end?

akuegel:
> Trying to build a mental model of the CI signal here — could use a sanity check. The `JAX Linux x86 AMD Instinct GPU` check is failing on the Copybara import (#41202), but the log shows the failures are Triton autotuner errors on f64 GEMMs (`No configs could be compiled`) in tests like `CustomJVPTest.test_vmap_inside_defjvp` and `EighTest.testEighIdentity*`, none of which look related to erf. I also checked a few other recent PRs (#41205, #41200, #41198, #41196) and the same check is failing there too, so I'm guessing this is a pre-existing flake or missing f64 Triton config on ROCm rather than something I broke. But I'm new to this CI setup — is that reading right, or is there something I should dig into on my end?

It seems likely it is a pre-existing flake. Regarding triton and f64 gemms, this is being investigated internally.

PR #41222, opened 2026-04-20
Integrate Triton up to [6ea516a6e4](https://github.com/triton-lang/tr…

nurmukhametov:
…iton/commits/6ea516a6e4cb1878fb6969d222e23294712b3955)

[List of integrated
commits](https://github.com/triton-lang/triton/compare/6a0b5461eb439bdb3ef0c721837c64dc2c345fcb...6ea516a6e4cb1878fb6969d222e23294712b3955)

Adapt to upstream API changes:
* `DialectInferLayoutInterface::inferReshapeOpEncoding` gained an `allowReorder` parameter in triton-lang/triton#9997.
* Rebase `mma_limit_pred.patch` and `no_accelerate_through_broadcast.patch` against the upstream rename of `DescriptorLoadOp` to `DescriptorLoadLikeOpInterface` in `AccelerateMatmul.cpp`. Patch semantics are unchanged.


vwbaker:
Hi Aleksei, can you give me some context for this change? Is there a specific commit you're waiting for?

We will not be accepting triton integration updates to openxla. When we update Triton for openxla, we update it for all of our projects and is a process that requires internal updates and reviews. 

However, we do update fairly often aiming to do so every 1-2 weeks. Is this cadence not working for AMD?

vwbaker:
This is our current integration if you'd like to follow it: https://github.com/openxla/xla/pull/41282

nurmukhametov:
> Hi Aleksei, can you give me some context for this change? Is there a specific commit you're waiting for?

There are two commits I need: 6ea516a6e and 23aecad95. Both will be picked up in the next integration: https://github.com/openxla/xla/pull/41282.

> We will not be accepting triton integration updates to openxla. When we update Triton for openxla, we update it for all of our projects and is a process that requires internal updates and reviews.
> 
> However, we do update fairly often aiming to do so every 1-2 weeks. Is this cadence not working for AMD?

Thanks for explaining. It works for us. Sorry for the noise, I wasn't aware of the process.

vwbaker:
Ok just as an FYI for the LLVM update: that is similarly another integration happening internally at a large scale so whether those updates get included in the Triton update is out of our control and it's possible that the llvm update does not land with the triton integration, but they are also updating quite often.

PR #41293, opened 2026-04-21
[ROCm] Add cuda-only tags for cuda specific tests

hsharsha:
📝 Summary of Changes
These are cuda specific tests filtered out because of -no-oss tags. Add cuda-only tags to them

🎯 Justification
It breaks ROCm builds.

🚀 Kind of Contribution: 🐛 Bug Fix, ♻️ Cleanup


hsharsha:
@xla-rotation  Can you please prioritize to review and submit this patch. It affects ROCm builds when we do not use -no-oss tag for filtering

EusebioDM (approved):
Sorry about the breakage!

alekstheod:
backends/gpu/tests/with_xla_gpu_cuda_data_dir_mi200_sorting.hlo there is one more test which started to fail but we will handle it internally.

beckerhe (approved):

PR #41348, opened 2026-04-22
[xla:gpu] Add analysis + pass to annotate compile time DS/DUS

ezhulenev:
- Add `DynamicSliceAnalysis` — analyzes dynamic-slice (DS) and dynamic-update-slice (DUS) instructions whose offsets are a linear function of a parent while loop's induction variable. Evaluates offset expressions via `HloEvaluator`, verifies constant stride, and returns `{loop_index, byte_offset, byte_stride}`. Also supports finding all DS/DUS operations on the same buffer (`FindDynamicSliceChain`) and checking non-overlapping access (`IsNonOverlapping`).
- Add `DynamicSliceAnnotator` — an `HloModulePass` that runs the analysis on every DS/DUS in while loop bodies and stamps successfully-analyzed instructions with a `DynamicSliceConfig` backend config (offset + stride * iteration).
- Add `DynamicSliceConfig` proto to `GpuBackendConfig` to carry the annotation through the pipeline.

This is groundwork for the v2 dynamic-slice fusion, which replaces the legacy offset-table approach with host-side loop state for offset computation, enabling support for arbitrary hero computations inside dynamic-slice fusions.

### DIFF
--- a/xla/backends/gpu/transforms/dynamic_slice_analysis.cc
+++ b/xla/backends/gpu/transforms/dynamic_slice_analysis.cc
@@ -1,0 +1,610 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/backends/gpu/transforms/dynamic_slice_analysis.h"
+
+#include <climits>
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <utility>
+#include <vector>
+
+#include "absl/algorithm/container.h"
+#include "absl/container/flat_hash_map.h"
+#include "absl/container/inlined_vector.h"
+#include "absl/log/check.h"
+#include "absl/log/log.h"
+#include "absl/status/statusor.h"
+#include "absl/types/span.h"
+#include "xla/tsl/platform/status_macros.h"
+#include "xla/hlo/evaluator/hlo_evaluator.h"
+#include "xla/hlo/ir/hlo_casting_utils.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_instructions.h"
+#include "xla/hlo/ir/hlo_opcode.h"
+#include "xla/literal.h"
+#include "xla/literal_util.h"
+#include "xla/service/gpu/ir_emission_utils.h"
+#include "xla/service/pattern_matcher.h"
+#include "xla/shape.h"
+#include "xla/shape_util.h"
+#include "xla/util.h"
+#include "xla/xla_data.pb.h"
+
+namespace xla::gpu {
+
+//===-----------------------------------------------------------------------===/
+// DynamicSliceDescriptor
+//===-----------------------------------------------------------------------===/
+
+static std::optional<absl::InlinedVector<int64_t, 4>>
+ComputeByteStridesForSubByte(const Shape& shape) {
+  absl::InlinedVector<int64_t, 4> strides(shape.dimensions().size());
+  int64_t stride_in_bits = ShapeUtil::ElementSizeInBits(shape);
+  for (int32_t i : shape.layout().minor_to_major()) {
+    if (stride_in_bits % CHAR_BIT != 0) {
+      strides[i] = -1;
+    } else {
+      strides[i] = stride_in_bits / CHAR_BIT;
+    }
+    stride_in_bits *= shape.dimensions(i);
+  }
+  return strides;
+}
+
+static std::optional<absl::InlinedVector<int64_t, 4>> ComputeByteStrides(
+    const Shape& shape) {
+  auto strides = ShapeUtil::ByteStrides(shape);
+  if (strides) {
+    return strides;
+  }
+  return ComputeByteStridesForSubByte(shape);
+}
+
+static bool IsZeroOffset(const HloInstruction* slice, int32_t dim) {
+  return GetSliceSize(slice, dim) == slice->operand(0)->shape().dimensions(dim);
+}
+
+int32_t GetFirstOffsetOperandIndex(const HloInstruction* slice) {
+  CHECK(slice->opcode() == HloOpcode::kDynamicSlice ||
+        slice->opcode() == HloOpcode::kDynamicUpdateSlice);
+  return slice->opcode() == HloOpcode::kDynamicSlice ? 1 : 2;
+}
+
+int64_t GetSliceSize(const HloInstruction* slice, int32_t dim) {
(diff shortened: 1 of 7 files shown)


xla/backends/gpu/transforms/dynamic_slice_analysis.cc
```diff
@@ -0,0 +1,610 @@
...
+  return StaggeredVariable{while_loop, operand, *init_value};
+}
+
+// Counts the number of intervening while loops between `instr` and
+// `while_loop`. Returns 0 if `instr` is directly inside `while_loop`'s body.
+static int64_t ComputeLoopIndex(const HloInstruction* instr,
+                                const HloInstruction* while_loop) {
+  int64_t depth = 0;
+  for (const HloComputation* comp = instr->parent(); comp != nullptr;) {
+    auto callers = comp->caller_instructions(HloOpcode::kWhile);
```
akuegel: As far as I can tell, we are calling this in cases with `while_loop` identified either by `TryResolveStaggeredVariable()` or by `ResolveFunctionalDependencyOnInductionVariable`. As far as I can tell, in the first case, we know that the caller of `instr` parent computation is `while_loop`. In the second case we know that there is a chain of unique callers that are CallLike ended by `while_loop` caller. It looks to me we should account for other callers than While here? Maybe there is also an opportunity for simplification here: It looks to me as if the loop index we are looking for is either 0 (for staggered variable case) or required_parameters.size() from the returned InductionVariableFunctionalDependency struct.

xla/backends/gpu/transforms/dynamic_slice_analysis.cc
```diff
@@ -0,0 +1,610 @@
...
+    const HloInstruction* instr) {
+  if (instr->opcode() != HloOpcode::kDynamicSlice &&
+      instr->opcode() != HloOpcode::kDynamicUpdateSlice) {
+    return Internal("FindDynamicSliceChain: expected DS or DUS, got %s",
+                    HloOpcodeString(instr->opcode()));
+  }
+
+  const HloComputation* computation = instr->parent();
+  auto callers = computation->caller_instructions(HloOpcode::kWhile);
+  if (callers.size() != 1 || callers.front()->while_body() != computation) {
```
akuegel: Simpler:

```
auto maybe_caller = computation->GetUniqueCaller(HloOpcode::kWhile);
if (!maybe_caller.has_value() || maybe_caller->while_body() != computation) {
```

xla/backends/gpu/transforms/dynamic_slice_analysis.cc
```diff
@@ -0,0 +1,610 @@
...
+  if (!Match(gte, m::GetTupleElement(m::Parameter()))) {
+    return std::nullopt;
+  }
+  const HloInstruction* param = gte->operand(0);
+  int64_t staggered_index = gte->tuple_index();
+
+  // Step 2: The parameter's parent computation must be the body of exactly
+  // one while loop.
+  const HloComputation* body = param->parent();
+  auto callers = body->caller_instructions(HloOpcode::kWhile);
```
akuegel: Simpler:

```
auto maybe_while_loop = computation->GetUniqueCaller(HloOpcode::kWhile);
if (!maybe_while_loop.has_value()) {
  return std::nullopt;
}
const HloInstruction* while_loop = *maybe_while_loop;
```

xla/backends/gpu/transforms/dynamic_slice_analysis.cc
ezhulenev: Done,

xla/backends/gpu/transforms/dynamic_slice_analysis.cc
ezhulenev: Done.

xla/backends/gpu/transforms/dynamic_slice_analysis.cc
ezhulenev: Actually in practice it's always 0 today, runtime (coming next PR) supports arbitrary nesting. Removed this function and updated documentation.

xla/backends/gpu/transforms/dynamic_slice_analysis.cc
```diff
@@ -320,10 +301,13 @@ absl::StatusOr<std::optional<DynamicSliceDescriptor>> AnalyzeDynamicSlice(
     auto functional_dependency =
         ResolveFunctionalDependencyOnInductionVariable(operand);
     if (functional_dependency) {
+      // ResolveFunctionalDependencyOnInductionVariable guarantees exactly one
```
akuegel, line 304: Right, there is exactly one while loop in the chain. I guess you only care about the number of while loops and not about other call-like ops in the chain?

akuegel (approved):

xla/backends/gpu/transforms/dynamic_slice_analysis.cc
ezhulenev: Yes, in theory we can have:

```
while i = 0 ... N {
  while j = 0 ... M {
    DUS(i, ....)
  }
}
  
```

Where we apply update to buffer indexed by parent loop (microbatching + scan loop), but analysis doesn't support it yet. I plan to look at it as a follow up because I'm pretty sure it can happen in practice.

ezhulenev: FWIW while loop runtime support tracks loop nests here: https://github.com/openxla/xla/blob/main/xla/backends/gpu/runtime/while_loop.h#L92-L100

ermilovmaxim (approved):

akuegel:
@ezhulenev 
After this library became used, we are hitting this error on a benchmark: `UNIMPLEMENTED: HloEvaluator does not implement the partition-id opcode`. Can you maybe add logic to guard against that?

ezhulenev:
Let me send a quick patch, I'll see if it's easy to just evaluate it

sfvaroglu:
@ezhulenev MaxText llama3-8b training NaNs at step 1 after this PR landed. Disabling `pipeline.AddPass<DynamicSliceAnnotator>()` in gpu_compiler.cc restores correct numerics. Can you please check?

ezhulenev:
It must be https://github.com/openxla/xla/pull/42396, can you add vlogs to dynamic memcpy thunk to see if it started copying data with wrong offesets

sfvaroglu:
Offsets seem correct, but for ~168 thunks the per-thunk offset table has only 15 entries when the loop body actually runs 16 times.

ezhulenev:
Hm, not sure how it happened, we do rely or trip count and and must not change after we run the pass

hawkinsp:
I'm just following up on this: we're trying to release JAX but don't want to release with the maxtext NaN problem. What's the current status? Should we locally revert this PR? Should it be reverted?

ezhulenev:
I'm migrating memcpy fusion to new thunk in https://github.com/openxla/xla/pull/42656, @sfvaroglu what's the simplest command to reproduce the error? I want to check if my PR fixes the issue.

sfvaroglu:
> I'm migrating memcpy fusion to new thunk in #42656, @sfvaroglu what's the simplest command to reproduce the error? I want to check if my PR fixes the issue.

This should do it
`python3 -m maxtext.trainers.pre_train.train /opt/maxtext/src/maxtext/configs/base.yml run_name=repro model_name=llama3-8b scan_layers=true remat_policy=minimal_with_context steps=2 per_device_batch_size=0.5 enable_checkpointing=false dataset_path=local dataset_type=synthetic attention=cudnn_flash_te max_target_length=8192 tokenizer_type=tiktoken tokenizer_path=/opt/maxtext/src/maxtext/assets/tokenizers/tokenizer_llama3.tiktoken hardware=gpu ici_fsdp_parallelism=8`

PR #41349, opened 2026-04-22
[xla:gpu] Add dynamic-slice defuser pass

ezhulenev:
- Add `DynamicSliceDefuser` — an `HloModulePass` that unwraps trivial `kLoop` fusions whose root is a `DynamicSlice` or `DynamicUpdateSlice` and whose only other non-parameter instructions are bitcasts and constants.
- These trivial fusions are created by priority-fusion absorbing bitcasts into DS/DUS instructions. They hide the DS/DUS from downstream passes (`DynamicSliceAnnotator`, `DynamicSliceFusionRewriterV2`) which need to see them in raw form. `FusionWrapper` re-wraps them later if needed.

This is part of the v2 dynamic-slice fusion work, ensuring DS/DUS instructions are visible to the analysis and annotation passes.

### DIFF
--- a/xla/backends/gpu/transforms/BUILD
+++ b/xla/backends/gpu/transforms/BUILD
@@ -1385,6 +1385,32 @@ xla_cc_test(
 )
 
 cc_library(
+    name = "dynamic_slice_defuser",
+    srcs = ["dynamic_slice_defuser.cc"],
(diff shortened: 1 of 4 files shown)


ezhulenev:
After discussion with @pifon2a I'm choosing a different direction

PR #41381, opened 2026-04-22
[XLA:GPU] Allow swap of operands for some non commutative ops in cuDNN flex attention

Cjkkkk:
📝 Summary of Changes
cuDNN flex attention currently has restrictions that the first operand should be virtual (parameter isn't virtual, parameter 0 is an exception because it is managed by cuDNN). 
For commutative ops, operands can just be swapped. For non commutative ops like subtract, operands are swapped and a negate is added after. For divide, we error out because swapping operands with reciprocal are not safe when first operand has value 0.


akuegel (approved):

Cjkkkk:
@akuegel Hi Adrian, seems like there are internal test failures, would you be able to share some infos on that? Thanks.

akuegel:
> @akuegel Hi Adrian, seems like there are internal test failures, would you be able to share some infos on that? Thanks.

It was an unrelated failure. This has been fixed, now we were able to merge this PR.

PR #41399, opened 2026-04-22
Gpu profiler deadlock ending up in DelayKernel timeout when autotuning in parallel

Finistere:
📝 Summary of Changes
This PR improves parallel compilation with `PJRT_Cient_Compile` when autotuning is enabled.

Autotuning relies on a DelayKernel to ensure that the start event is recorded just before the profiled kernel. It's a simple spin loop that keeps the GPU busy until the host could schedule the start & end recording and the profiled kernel launches. Once that's done the host changes a semaphore to signal that the spin loop can stop. This works fine when a single `PJRT_Cient_Compile` is called. But when calling in parallel, we're seeing DelayKernel timing out all the time.

After some investigation, it's due to CUDA calls such as `cuMemcpyDtoHAsync` that block while the DelayKernel spins in the same stream. The problem is that this then blocks further calls like `cuRecordEvent`, and now the start/end event recording cannot be scheduled and we end up in a deadlock between a thread that wants to profile and another one that intends to check input or output buffers. The deadlock is only "solved" through the timeout mechanism.

The first commit takes the gpu lock in shared mode to avoid the checks to execute simultaneously with the profiling. That solves almost all deadlocks. But I still noticed a few deadlocks because of `cuModuleUnload`within the `ScopedModuleHandle` desctructor. It's more complicated to prevent any conflicts here. So I propose here one fairly minimal solution that ensures we have a shared lock before calling the destructors, but I'm not sure whether that's acceptable.

🎯 Justification
Today compiling multiple program in parallel with `PJRT_Cient_Compile` is slower for us than compiling them sequentially. Given that auto-tuning takes a significant amount of time and needs to happen at start-up we want to minimize it as much as possible.

I've tested on the 5090 (CUDA 13.1 / 590.48.01) and 4090 (CUDA 13 / 580.95.05).

🚀 Kind of Contribution

### DIFF
--- a/xla/service/gpu/gpu_executable.cc
+++ b/xla/service/gpu/gpu_executable.cc
@@ -462,6 +462,38 @@ GpuExecutable::GpuExecutable(
 }
 
 GpuExecutable::~GpuExecutable() {
+  // Take reader locks on the per-device GPU mutex for all executors this
+  // executable has initialized modules on. This prevents cuModuleUnload
+  // (triggered by kernel/thunk destruction below) from racing with the
+  // DelayKernel used during autotuning profiling. The profiler holds the
+  // writer lock while the DelayKernel is active; taking reader locks here
+  // ensures we wait for profiling to finish before unloading modules.
+  std::vector<se::StreamExecutor*> executors;
+  {
+    absl::MutexLock lock(&module_handle_mutex_);
+    executors.reserve(module_handles_.size());
+    for (const auto& module_handle : module_handles_) {
+      executors.push_back(module_handle.first);
+    }
+  }
+
+  std::vector<std::unique_ptr<absl::ReaderMutexLock>> gpu_locks;
+  gpu_locks.reserve(executors.size());
+  for (se::StreamExecutor* executor : executors) {
+    gpu_locks.push_back(
+        std::make_unique<absl::ReaderMutexLock>(&GetGpuMutex(executor)));
+  }
+
+  // Explicitly destroy thunks and module handles under the reader locks.
+  // Thunk destruction triggers kernel unloading (CudaExecutor::UnloadKernel ->
+  // cuModuleUnload). This must happen while we hold reader locks to prevent
+  // cuModuleUnload from racing with the DelayKernel on another thread.
+  thunk_executor_.reset();
+  {
+    absl::MutexLock lock(&module_handle_mutex_);
+    module_handles_.clear();
+  }
+
   if (has_module() && enable_debug_info_manager_) {
     XlaDebugInfoManager::Get()->UnregisterModule(module().unique_id());
   }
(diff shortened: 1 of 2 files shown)


xla/service/gpu/gpu_executable.cc
```diff
@@ -462,6 +462,31 @@ GpuExecutable::GpuExecutable(
...
+  // Take reader locks on the per-device GPU mutex for all executors this
+  // executable has been run on. This prevents cuModuleUnload (triggered by
+  // kernel/thunk destruction below) from racing with the DelayKernel used
+  // during autotuning profiling. The profiler holds the writer lock while the
+  // DelayKernel is active; taking reader locks here ensures we wait for
+  // profiling to finish before unloading modules.
+  std::vector<std::unique_ptr<absl::ReaderMutexLock>> gpu_locks;
+  {
+    absl::MutexLock lock(&module_handle_mutex_);
+    for (se::StreamExecutor* executor : executors_) {
```
shyamliagrawal: Can you use the keys of module_handles_, it already tracks every executor the executable has been initialized on? You will have to create a new local container of executors to release the module_handle_mutex_ before locking the GPU devices.

shyamliagrawal (commented):
LGTM otherwise.

xla/service/gpu/gpu_executable.cc
Finistere: Fixed, definitely cleaner. Thanks

shyamliagrawal (approved):
Thank you!

bchetioui:
This change causes deadlocks in JAX tests, so we are reverting it! JAX's `gpu_pallas_distributed_test` when run on 2xB200 reproduces the issue.

xla/service/gpu/gpu_executable.cc
```diff
@@ -462,6 +462,38 @@ GpuExecutable::GpuExecutable(
...
+  // executable has initialized modules on. This prevents cuModuleUnload
+  // (triggered by kernel/thunk destruction below) from racing with the
+  // DelayKernel used during autotuning profiling. The profiler holds the
+  // writer lock while the DelayKernel is active; taking reader locks here
+  // ensures we wait for profiling to finish before unloading modules.
+  std::vector<se::StreamExecutor*> executors;
+  {
+    absl::MutexLock lock(&module_handle_mutex_);
+    executors.reserve(module_handles_.size());
+    for (const auto& module_handle : module_handles_) {
```
sohaibiftikhar, line 475: The order of iteration over a hashmap is not deterministic.
Differing orders will cause locks on line 484 to take reader mutexes in differing order.

Finistere: I think we could solve this by first accumulating the executors in an array and then sorting it by `device_ordinal`, no?

sohaibiftikhar: Or simply by pointer address should be enough for instance.

Finistere:
I've made a new PR with a fix for the GPU deadlock: https://github.com/openxla/xla/pull/42347. If you get a chance @sohaibiftikhar, I'd appreciate a review.

PR #41430, opened 2026-04-22
[xla] Forward known trip count as constant to while users

ezhulenev:
If while loop has trivial iteration bounds we can replace the GTE from while loop that reads loop counter with a constant value.


ermilovmaxim (approved):

draganmladjenovic:
@ezhulenev This breaks on while loops using u32 as an counter https://github.com/openxla/xla/actions/runs/24956978017/job/73076855099?pr=41591

PR #41476, opened 2026-04-23
[xla:gpu] Make CollectiveThunk a Command and retire AllReduceCmd

shawnwang18:
CollectiveThunk now inherits from Command and implements Record() directly, removing the need for a parallel *Cmd wrapper hierarchy for collective operations. This PR migrates AllReduceCmd as the first consumer of the new path — the emitter now borrows AllReduceThunk* directly into CommandSequence, matching the KernelThunk / GemmThunk / GpublasLtMatmulThunk pattern. Subsequent PRs will migrate ReduceScatterCmd,
  AllToAllCmd, AllGatherCmd, etc.

  Motivation

  The *Cmd wrappers (e.g. AllReduceCmd) duplicate state already carried by the corresponding thunks (config, reduction kind, buffers) and duplicate behavior that the thunk already implements (comm resolution, RunAllReduce invocation). Keeping both in lock-step is a maintenance tax whenever a new collective feature is added (e.g. symmetric buffer support, rendezvous semantics). Letting the thunk play both roles —
  eager execution and command-buffer recording — eliminates the wrapper entirely.

  Key changes

  Core (collective_thunk.{h,cc})

  - CollectiveThunk : public Command. Keeps IsTracedCommand() == true and requires_initialization() == true.
  - New Record(ExecuteParams&, RecordParams&, RecordAction, se::CommandBuffer*) traces RunCollective into a nested CUDA graph via se::TraceCommandBufferFactory::Create, then attaches it as a child command (CreateChildCommand) or updates an existing child (UpdateChildCommand).
  - New private helpers to avoid duplication with ExecuteOnStream:
    - FirstCallRendezvous(params, clique_key, label, flag) — the pre/post first-call rendezvous previously inlined in ExecuteOnStream.
    - RunWithCommAndRendezvous(params, fn) — resolves clique_key + Communicator*, runs the pre-rendezvous, invokes fn(clique_key, comm), then the post-rendezvous. Both ExecuteOnStream and Record are now thin wrappers over it.
  - Doc comment now explicitly calls out the hybrid Thunk+Command role, mirroring KernelThunk/CuDnnThunk.

  Emitter (command_buffer_cmd_emitter.cc)

  - Deleted the Convert(const AllReduceThunk&) overload.
  - Thunk::Kind::kAllReduce arm now appends a borrowed AllReduceThunk* to the CommandSequence, matching kKernel / kGemm / kCublasLtMatmul.

  *Cmd cleanup

  - Deleted AllReduceCmd class (command_buffer_cmd.{h,cc} — ctor, Record, buffer_uses, ~85 lines).
  - Removed V(kAllReduceCmd, …) from XLA_GPU_COMMAND_LIST (command.h).
  - In IsCollectiveCommand (command.cc): removed kAllReduceCmd and added kCollectiveCmd so execute_params.mock_collectives continues to short-circuit collective commands — CollectiveThunk's constructor uses CommandType::kCollectiveCmd.

  Command::ExecuteOnStream docstring (command.h)

  Softened the "not supported" wording to reflect reality: Thunk/Command hybrids (KernelThunk, CollectiveThunk, CuDnnThunk) routinely override it to support eager execution alongside recording.

  Behavioral notes

  - Rendezvous semantics are preserved, not regressed. AllReduceCmd previously relied on CollectiveCmd::RecordTracedCommand which performed no rendezvous. The migrated path performs the pre/post first-call rendezvous around the trace, which is the same policy ExecuteOnStream has always used.
  - NCCL lazy init interacts with stream capture. The multi-GPU test performs an eager warm-up ExecuteOnStream before the first Record call because NCCL's first-call initialization issues sync primitives that CUDA stream capture rejects. The rendezvous alone doesn't drive that init. This is the same warm-up pattern the production initialization path uses — worth tracking as a follow-up to fold the warm-up into
  CollectiveThunk::Initialize so Record can be called cold.
  - No proto / serializer changes. AllReduceCmd had no proto representation; the emitter is the only consumer.

  Tests

  - all_reduce_thunk_test (single-GPU, CUDA 12.9+): RecordCommandBufferCreate, RecordCommandBufferUpdate. Uses a NoOpAllReduceThunk that overrides Record with a trivial memset trace to exercise the command-buffer plumbing without NCCL.
  - all_reduce_thunk_multigpu_test (new, multi-GPU, CUDA 12.9+, 2 GPUs): RecordCommandBufferCreate, RecordCommandBufferUpdate. Uses a DirectAllReduceThunk + tsl::thread::ThreadPool to drive two ranks concurrently through Setup → Prepare → AcquireCollectiveCliques → Initialize → warm-up ExecuteOnStream → Record(Create) → Submit → verify dst == 3.0 → Record(Update) with a fresh buffer set → Submit → verify dst ==
  30.0, and confirms Update returns the same command-node pointer.


ezhulenev (commented):
@apivovarov please also take a look, I skimmed it and it look good to me

ezhulenev (approved):

akuegel (approved):

xla/backends/gpu/runtime/BUILD
```diff
@@ -2051,16 +2051,88 @@ cc_library(
...
+        ":collective_thunk",
+        ":command",
+        ":command_state",
+        ":scratch_memory_requests",
+        ":thunk",
+        "//xla:future",
+        "//xla:shape_util",
+        "//xla:xla_data_proto_cc",
+        "//xla/backends/gpu/collectives:gpu_clique_key",
+        "//xla/backends/gpu/runtime:collective_execution",
```
akuegel: build_cleaner claims this dependency is unused.

shawnwang18: Thanks, fixed.

akuegel (approved):

PR #41492, opened 2026-04-23
[xla:gpu] Add one-sided AllGather implementation

ezhulenev:
Similar to collective permute support all-gather using Put/Signal APIs.


apivovarov (approved):
Great! Thank you! Excellent example for transitioning one-shot RaggerAllToAll to the symmetrical memory.

ermilovmaxim (approved):

apivovarov:
We also need to add a setter of `xla_gpu_all_gather_mode` to xla/debug_options_flags.cc

ezhulenev:
Oops, pushed a fix with falg parsing! @apivovarov

apivovarov (approved):

xla/backends/gpu/runtime/all_gather_thunk.cc
```diff
@@ -125,16 +160,28 @@ absl::StatusOr<ThunkProto> AllGatherThunk::ToProto() const {
...
                                            const GpuCliqueKey& clique_key,
                                            se::Stream& stream,
                                            Communicator& comm) {
+  int device_ordinal = stream.parent()->device_ordinal();
+
   ASSIGN_OR_RETURN(std::vector<DeviceBufferPair> device_buffers,
                    ConvertToDeviceBuffers(params.buffer_allocations, buffers(),
                                           config_.config.operand_element_type));
+
+  if (use_symmetric_memory() && clique_key.is_local()) {
```
apivovarov, line 177: `is_local` will return false on multihost setup such as GB200 2x4.
we can use `fabric_safe` instead of `is_local`
```
  FabricHomogeneity homogeneity =
      CheckFabricHomogeneity(stream.parent(), clique_key);
  bool fabric_safe = (homogeneity == FabricHomogeneity::kHomogeneous);
```
It guaranties that the participating devices are interconnected with fast and healthy network (Means the devices belongs to a single fabricUUID/clique_id nvmlDeviceGetGpuFabricInfo)

nvmlDeviceGetGpuFabricInfo API is available starting from Driver 545+ (released on Oct 2023) , otherwise CheckFabricHomogeneity returns kUnknown

From the other side, RunOneSidedAllGather uses NCCL Signal and Put API available starting from NCCL 2.29 (first released on Dec 23, 2025). 

Highly unlikely that new NCCL 2.29 will be used with very old driver 535.
Means that it is safe to use `CheckFabricHomogeneity` with strict check `homogeneity == FabricHomogeneity::kHomogeneous`.
RunOneSidedAllGather users supposed to use quite fresh driver and NCCL version

ezhulenev: Can you do it on your side, I don't have a way to test it

ezhulenev: And for collective-permute as well please. In our internal fork I simply removed that check, but if there is a canoncial API for this check, let's do that!

xla/backends/gpu/runtime/all_gather_thunk.cc
```diff
@@ -161,5 +208,129 @@ absl::Status RunAllGather(std::vector<DeviceBufferPair>& buffers,
...
+              clique_key);
+        }
+
+        size_t offset = dest_offset + current_rank * chunk_size;
+
+        XLA_VLOG_DEVICE(3, device_ordinal)
+            << "OneSidedAllGather: Put " << chunk_size << " bytes to peer "
+            << peer_rank << " at offset " << offset;
+
+        RETURN_IF_ERROR(c->LaunchPut(buf.source_buffer, sym_mem, offset,
```
apivovarov, line 296: Nit: maybe use `comm->` instead? if I need to search for all places where the communicator is used I might search for "comm", so this place can be easily overlooked

xla/backends/gpu/runtime/all_gather_thunk.cc
apivovarov: Sure, we can do it.

xla/backends/gpu/runtime/all_gather_thunk.cc
```diff
@@ -71,32 +85,53 @@ absl::Status CheckImplementableInst(const HloAllGatherInstruction* inst) {
...
 
-/*static*/ CollectiveOpGroupMode AllGatherThunk::GetGroupMode(
+CollectiveOpGroupMode AllGatherThunk::GetGroupMode(
     const HloAllGatherInstruction* inst) {
   return GetAllGatherConfig(inst).config.group_mode;
 }
 
+absl::Status AllGatherThunk::PrepareCollective(const PrepareParams& params,
+                                               const GpuCliqueKey& clique_key) {
+  if (use_symmetric_memory() && clique_key.is_local()) {
```
akuegel, line 123: This seems to be somewhat redundant with the recent code addition in the caller of PrepareCollective. Can you please check the lines before calling PrepareCollective in collective_thunk.cc, how it fits together with that?

xla/backends/gpu/runtime/all_gather_thunk.cc
akuegel: If we want to use that check for fabric_safe, we need access to the stream. So I guess the PrepareCollective method would need this as parameter as well?

xla/xla.proto
```diff
@@ -514,6 +514,10 @@ message DebugOptions {
   // transformation.
   optional int64 xla_gpu_collective_permute_decomposer_threshold = 237;
 
+  // Memory mode for the all-gather operation. Determines which memory
+  // space all-gather buffers are allocated in. See CollectivesMode for details.
+  optional CollectivesMode xla_gpu_all_gather_mode = 503;
```
akuegel: Id should be 476, and "Next id: 476" line updated accordingly.

xla/backends/gpu/runtime/all_gather_thunk.cc
ezhulenev: It's super confusing, but we have two symmetric memory flags now, the one in base CollectiveThunk has different meaning :) talked about it with @PatriosTheGreat earlier, it's unfortunate state, but it will be resolved soon

xla/xla.proto
ezhulenev: Fixed proto. Regarding `fabric_safe` - yes, that would be the right thing, but it needs some rethinking how it all works together, the idea of `Prepare` is that you should not be able to run anything in it (no stream, no buffer). For now I'm using is_local just to make tests pass and gradually build right APIs.

xla/debug_options_flags.cc
```diff
@@ -275,6 +276,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_enable_all_gather_combine_by_dim(false);
   opts.set_xla_gpu_collective_permute_mode(
       DebugOptions::COLLECTIVES_PRIVATE_MEMORY);
+  opts.set_xla_gpu_all_gather_mode(DebugOptions::COLLECTIVES_PRIVATE_MEMORY);
```
apivovarov, line 280: Lets decide on the order of xla_gpu_all_gather_mode and xla_gpu_collective_permute_mode records and use it consistently in both xla.proto and debug_options_flags.cc. e.g. xla_gpu_all_gather_mode should go first.

xla/backends/gpu/runtime/all_gather_thunk.cc
```diff
@@ -71,32 +85,53 @@ absl::Status CheckImplementableInst(const HloAllGatherInstruction* inst) {
...
 }
 
 AllGatherThunk::AllGatherThunk(ThunkInfo thunk_info, CollectiveConfig config,
-                               std::vector<Buffer> buffers)
-    : CollectiveThunk(Thunk::kAllGather, thunk_info, std::move(buffers)),
+                               std::vector<Buffer> buffers,
+                               CollectivesMode collectives_mode)
+    : CollectiveThunk(Thunk::kAllGather, thunk_info, std::move(buffers),
+                      CommunicationId(0), collectives_mode),
       config_(AllGatherConfig{std::move(config)}) {}
```
apivovarov, line 107: Should we add the following here too, similar to the first ctor?
```
CHECK_EQ(config_.config.operand_element_type.size(), this->buffers().size());
```

sfvaroglu:
> Great! Thank you! Excellent example for transitioning one-shot RaggerAllToAll to the symmetrical memory.

@apivovarov Please see https://github.com/openxla/xla/pull/41552

xla/backends/gpu/runtime/all_gather_thunk.cc
```diff
@@ -161,5 +208,129 @@ absl::Status RunAllGather(std::vector<DeviceBufferPair>& buffers,
...
+// writes its source chunk into every peer's destination buffer at the correct
+// offset using symmetric memory.
+//
+// Synchronization protocol:
+//   1. Signal all peers "recv buffer ready" (gates peers' Puts until we have
+//      consumed the previous data).
+//   2. WaitSignal from all peers (all destination buffers are ready).
+//   3. Put our source chunk into each peer's destination buffer.
+//   4. WaitSignal from all peers for their PutSignals (data has arrived).
+static absl::Status RunOneSidedAllGather(
```
Tixxx, line 221: nccl already does this through copy engine batch requests, https://github.com/NVIDIA/nccl/blob/1933fdd6360a8bfccaa0166bd71bce363d32e5b6/src/ce_coll.cc#L526.
I dont think it's necessary to replicate it here.
The logic to request collective memory is good to have.

ezhulenev: Where nccl decides to call `ncclCeAllGather`? Or it's a new user API?

Tixxx: if it's symmetric memory and NCCL_CTA_POLICY_ZERO is set for the comm, it will use CE by default. Right now, NCCL_CTA_POLICY_ZERO is based on the symmetric buffer flag, we need to make it aware of the modes you added.

ezhulenev: I'm trying to find code path in nccl where it detects it, we pass regular device pointers to it, how can nccl know to what window it's mapped? What if device pointer mapped to multiple windows at the same time? What if it finds non-symmetric windows? I have so many questions :)

Tixxx: the symm check is here(https://github.com/NVIDIA/nccl/blob/1933fdd6360a8bfccaa0166bd71bce363d32e5b6/src/enqueue.cc#L2954), it will take the comm and the pointer to see if there's a window associated with it.
I dont think you can register a device pointer with multiple windows though. And what's a non-symm window? Are you referring to some memory that's not registered?

ezhulenev: AI anaysis: (a longer version https://gist.github.com/ezhulenev/b272d925bb2f18d9aaca95d1651b248e)

```
But if you have a pattern like: "allocate a large buffer, register it as window A, then register a slice of it as window B" — and different ranks do this with slightly different sizing or ordering — the
   sorted-array lookup will silently pick one window, and different ranks might pick differently.
```

I'm thinking how to make it bulletproof in presence of BFC, unified allocator, etc. With the way how we do caching I'm not convinced any of the approaches will not lead to corruption. With explicit puts and signals in XLA we can keep arbitrary number of symmetric memoris owned and managed by XLA, and as long as we can agree what windows to use for operation, we will be safe (this is TBD, I'll send PR soon),

FWIW I also like explicit Put/Signal APIs, as it doesn't hide any magic and hard to miss if it silentky didn't use CE.

apivovarov:
# HLO Crash

The hlo listed below crashes with the following error when executed with `--xla_gpu_all_gather_mode=symmetric`

### Error:
```
INTERNAL: NCCL operation ncclCommWindowRegister( comm, addr.opaque(), addr.size(), &win, 0x01) failed: invalid argument (run with NCCL_DEBUG=WARN for details). Last NCCL warning(error) log entry (may be unrelated) 'Window address must be suitably aligned.'. [executable_name='ag']
```

### HLO:
```
HloModule ag, input_output_alias={ {}: (1, {}) }, replica_count=8
// HloModule ag, replica_count=8   // Workaround

while_condition {
  params = (s32[], f32[4], f32[32]) parameter(0)
  loop_counter = s32[] get-tuple-element(params), index=0
  limit = s32[] constant(20)
  ROOT result = pred[] compare(loop_counter, limit), direction=LT
}

while_body {
  params = (s32[], f32[4], f32[32]) parameter(0)

  loop_counter = s32[] get-tuple-element(params), index=0
  input = f32[4] get-tuple-element(params), index=1
  output = f32[32] get-tuple-element(params), index=2

  ag = f32[32] all-gather(input), replica_groups={{0,1,2,3,4,5,6,7}}, dimensions={0}
  new_loop_counter = s32[] add(loop_counter, s32[] constant(1))
  ROOT result = tuple(new_loop_counter, input, ag)
}

ENTRY entry_computation {
  init_loop_counter = s32[] constant(0)

  input = f32[4] parameter(0)
  output = f32[32] parameter(1)

  while_init = tuple(init_loop_counter, input, output)
  while = (s32[], f32[4], f32[32]) while(while_init), condition=while_condition, body=while_body

  ROOT result = get-tuple-element(while), index=2
}
```

### Workaround
To make it work - remove `input_output_alias` attr on Line 1.

### Analysis:
Optimized hlo dump shows
module_0001.ag.cl_905114880.sm_9.0a_gpu_after_optimizations.txt
```
ENTRY %entry_computation (input.1: f32[4], output.1: f32[32]) -> f32[32] {
  %init_loop_counter_1 = s32[] constant(0)
  %copy.9 = s32[] copy(%init_loop_counter_1)
  %output.1 = f32[32]{0:S(1)} parameter(1), metadata={scheduling_name="output.1"}
  %input.1 = f32[4]{0} parameter(0), metadata={scheduling_name="input.1"}
  %copy.10 = f32[4]{0:S(1)} copy(%input.1)
  %tuple = (s32[], f32[4]{0:S(1)}, f32[32]{0:S(1)}) tuple(%copy.9, %copy.10, %output.1), metadata={scheduling_name="tuple"}
  %while = (s32[], f32[4]{0:S(1)}, f32[32]{0:S(1)}) while(%tuple), condition=%while_condition, body=%while_body, metadata={scheduling_name="while"}, backend_config={"known_trip_count":{"n":"20"},           "known_init_step":{"init":"0","step":"1"},"known_induction_variable":{"tuple_index":"0"},"dynamic_variable_tuple_indices":[]}
  ROOT %result.2.0 = f32[32]{0:S(1)} get-tuple-element(%while), index=2, metadata={scheduling_name="result.2.0"}
}
```

### Problem
parameter(1) is colored as S(1). But in XLA GPU input parameters must be S(0).

### Why parameter(1) is colored as S(1).
ENTRY output is S(1) because ENTRY Output is transitive AG output which is S(1). The output is an alias of parameter(1).  This is why parameter(1) is also colored as S(1).

### How to solve the issue
If ENTRY output is S(1) - insert copy operator at the end of the ENTRY to air-gap transitive AG S1 output from ENTRY S0 output.
BTW, It should automatically solve the issue with parameter(1). Copy op for parameter(1) should be automatically inserted once ENTRY output is colored as S0. Currently copy of for parameter(1) is already inserted if we run hlo without `input_output_alias` attr.

### Correct optimized HLO
Correct optimized HLO should be
```
ENTRY %entry_computation (input.1: f32[4], output.1: f32[32]) -> f32[32] {
  %init_loop_counter_1 = s32[] constant(0)
  %copy.9 = s32[] copy(%init_loop_counter_1)
  %output.1 = f32[32]{0} parameter(1), metadata={scheduling_name="output.1"}   // CHANGE: S1 is removed
  %copy.11 = f32[32]{0:S(1)}  copy(%input.1)   // CHANGE: copy op is added
  %input.1 = f32[4]{0} parameter(0), metadata={scheduling_name="input.1"}
  %copy.10 = f32[4]{0:S(1)} copy(%input.1)
  %tuple = (s32[], f32[4]{0:S(1)}, f32[32]{0:S(1)}) tuple(%copy.9, %copy.10, %copy.11), metadata={scheduling_name="tuple"}    // CHANGE: %copy.11 is used now
  %while = (s32[], f32[4]{0:S(1)}, f32[32]{0:S(1)}) while(%tuple), condition=%while_condition, body=%while_body, metadata={scheduling_name="while"}, backend_config={"known_trip_count":{"n":"20"},           "known_init_step":{"init":"0","step":"1"},"known_induction_variable":{"tuple_index":"0"},"dynamic_variable_tuple_indices":[]}
  %result.2.0 = f32[32]{0:S(1)} get-tuple-element(%while), index=2, metadata={scheduling_name="result.2.0"}
  ROOT %copy.13 =  f32[32]{0} copy(%result.2.0)   // CHANGE: copy op is added
}
```
Changes are tagged with `// CHANGE:` - total 4 changes


### More info
I bit more info

One of the many places where copy might be inserted/removed is GpuCompiler::RunPostSchedulingPipelines
GpuCompiler::RunPostSchedulingPipelines calls
RunPostSchedulingCopyInsertion calls
copy_insertion.CopyInsertion::AddSpecialCaseCopies
it has block which inserts copy op for ENTRY params because they are considered as read-only
```
  // Iterate through values of all constants and entry parameters. These values
  // are special because they are held in read-only buffers. If any of these
  // values share a buffer with other values (for example, the init value of a
  // while is a constant) then copy the value at its definition and replace all
  // its uses with the copy.
```
However, if we use `input_output_alias` then ENTRY params participating in the aliasing are not read-only anymore and copy is not inserted for them.

The result of adding/not_adding copy op for params is visible in dump `post-scheduling-passes.after_pipeline-start.before_async-to-sync-converter.txt`

### Challenge
None of the "regular" hlo dumps contains coloring information.
We start seeing S(1) info only in the following hlo dumps:
- gpu_after_optimizations-buffer-assignment.txt
- gpu_after_optimizations-buffer-assignment-values.txt
- gpu_after_optimizations.txt

My understanding that running additional hlo passes after optimizations-buffer-assignment.txt to insert copy op is not really possible.

If we have coloring info in "regular" hlo dumps it would be possible then to create a pass which inserts copy ops for  ENTRY function params and the result in case they are colored as S(1)

### Appendix A
#### cmd
```
XLA_FLAGS="--xla_gpu_all_gather_mode=symmetric --xla_dump_to=/tmp/dump_gpu_ag --xla_dump_hlo_as_text=true --xla_dump_hlo_pass_re=.*" \
blaze-bin/xla/tools/multihost_hlo_runner/hlo_runner_main \
--hlo_argument_mode=use_device_id_as_input \
--device_type=gpu \
--num_partitions=1 --num_replicas=8 \
--logtostderr \
--log_output=false \
$F \
--vmodule=nvptx_compiler=1,gpu_compiler=1,all_gather_thunk=5,functional_hlo_runner=1,cuda_executor=5
```

akuegel:
@ezhulenev can you please rebase?
Also, is the analysis that @apivovarov posted something that helps tracking down the problem? Or do you need more information?

ezhulenev:
Rebased and fixed a conflicts. The problem identified by @apivovarov is a general one for all ops that require symmetric memory, and I believe he o @PatriosTheGreat have a plan to address it

PatriosTheGreat:
> Rebased and fixed a conflicts. The problem identified by @apivovarov is a general one for all ops that require symmetric memory, and I believe he o @PatriosTheGreat have a plan to address it

@apivovarov is working on fixing this issue in parallel

apivovarov:
To address S1/S0 Mem space HLO Crash issue described above I opened

#42137 - [GPU] Use Collective ops Buffer Analysis in special copy insertion

### The problem:
The old value-based copy insertion (`ShouldAddCopyForCollectiveMemorySpace`) was blind to control flow boundaries. If an S1 collective op was inside a while loop and used an input_output_alias, the old logic only looked at direct HloUse chains. It couldn't see that the entry parameter and the loop's collective op shared the same physical buffer, which resulted in missing S0/S1 air-gap copies.

More details are in the PR description

ezhulenev:
Rebased. Needs a re-approval

dimitar-asenov (approved):

apivovarov:
To isolate S1/S0 Mem space in HLO which uses RequiresCollectiveSymmetricMemorySpace (e.g. --xla_gpu_collective_permute_mode=symmetric) I opened #42653

ezhulenev:
Rebased + fixed.

ezhulenev:
Rebased one more time.

ermilovmaxim (approved):

ermilovmaxim (approved):

PR #41497, opened 2026-04-23
[xla] Add a separate pass to propagate metadata across kCall instructions

ezhulenev:
Attempt 2, firs one was in: https://github.com/openxla/xla/pull/40920

The diff is that we keep metadata propagation inside inliner and add a separate pass for calls that are not inlined. The right approach is to teach our profiling infrastructure to look through the control flow ops in HLO and does this merging at run time, and also teach all HLO passes to look for attributes up in the call chain, but this change is too big and too hard to do from OSS side.

- New `PropagateCallMetadata` HLO pass that propagates metadata (op_name prefix and stack_frame_id) from `kCall` instructions into their called computations, recursing through nested control-flow (while, conditional) but not into embedded computations (reduce's to_apply, etc.)
- Removes metadata propagation from `CallInliner` — the `RecursivelyUpdateMetadata` helper and per-instruction metadata update during inlining are no longer needed since the new pass handles this as a separate concern
- Wired into GPU compiler pre-SPMD pipeline before `CallInliner`, so metadata is propagated while `kCall` ops still exist (including non-inlinable calls)
- Test coverage for op_name propagation, stack frame concatenation, overflow protection, redundant prefix detection, nested calls, and idempotency

#### Motivation

The `CallInliner` only updates metadata for calls it actually inlines. Non-inlinable calls (e.g. calls with inlineable="false") were skipped entirely, leaving their callee instructions with incomplete metadata context. Extracting this into a standalone pass ensures all calls get metadata propagation regardless of inlining decisions.


xla/service/cpu/cpu_compiler.cc
```diff
@@ -624,6 +626,7 @@ absl::Status CpuCompiler::RunHloPassesThroughLayoutAssn(
       spmd_pipeline.AddPass<RecognizeReduceWindow>();
       spmd_pipeline.AddPass<CollectivePermuteCSE>();
     }
+    spmd_pipeline.AddPass<PropagateCallMetadata>();
```
akuegel: Now that CallInliner does still propagate metadata, I guess we only need the new pass at the start of the CPU pipeline?

mkuperst: Another alternative is actually to defer this to the end. This way the inliner will do the right thing for everything that gets inlined, and then we can make this fixup for whatever didn't get inlined for whatever reason.

ezhulenev: Good point, move the propagate metadata to the very end to annotate whatever calls survive all earlier passes.

akuegel (approved):

xla/hlo/transforms/propagate_call_metadata.h
```diff
@@ -0,0 +1,68 @@
...
+#include "xla/hlo/pass/hlo_pass_interface.h"
+#include "xla/xla_data.pb.h"
+
+namespace xla {
+
+// Propagates metadata (op_name prefix and stack_frame_id) from kCall
+// instructions into their called computations, recursing through nested
+// calls and control-flow structures.
+//
+// This pass should run before call inlining, while kCall ops still exist.
```
mkuperst: I think this comment is now stale.

xla/hlo/transforms/propagate_call_metadata.cc
```diff
@@ -0,0 +1,170 @@
...
+        PropagateMetadataToInstruction(instruction, prefix, parent_frame_id);
+      }
+    }
+  }
+
+  OpMetadata metadata = hlo->metadata();
+  if (UpdateOpName(metadata, prefix)) {
+    hlo->set_metadata(metadata);
+  }
+  UpdateStackFrame(hlo, parent_frame_id);
```
mkuperst, line 107: This adds a metadata copy compared to the original code, right? (Maybe not a big deal, though?)

xla/hlo/transforms/propagate_call_metadata.cc
```diff
@@ -0,0 +1,170 @@
...
+  std::string op_name = metadata.op_name();
+  absl::string_view clean_name = absl::StripPrefix(op_name, "/");
+  clean_name = absl::StripSuffix(clean_name, "/");
+
+  if (absl::StartsWith(clean_name, clean_prefix)) {
+    return false;
+  }
+  if (!clean_name.empty() && absl::StrContains(clean_prefix, clean_name)) {
+    return false;
+  }
```
mkuperst: This looks a bit broad (compared with the StartsWith). What case does this cover?

xla/hlo/transforms/propagate_call_metadata.cc
ezhulenev: Yes. With how we keep it not, I think it's not big deal, quality of life improvement from better perf profiles is bigger.

xla/hlo/transforms/propagate_call_metadata.cc
ezhulenev: Removed, indeed I don't think it' useful

mkuperst (approved):

penpornk (approved):

Issue #37117, opened 2026-02-01
Logic Flaw: IsAlwaysDuplicable violates may_duplicate_ semantics and ignores index-computation costs


red1239109-cmd:
InstructionFusion currently allows certain instructions (e.g., kBroadcast, widening kConvert) to be duplicated even when may_duplicate_ == false, via the IsAlwaysDuplicable exception logic.

This implementation weakens the semantic contract of may_duplicate_, introduces hidden policy exceptions, and ignores non-trivial index-computation and register-pressure costs—especially for high-rank broadcasts. While this may have originated as a performance heuristic, it now represents a structural logic flaw that complicates reasoning, maintenance, and future optimization decisions.

**Problem 1: Semantic Violation of may_duplicate_**

From a reader and maintainer perspective, the meaning of may_duplicate_ is straightforward:
"may_duplicate_ = false" implies that instruction duplication must not occur.

However, the current implementation violates this expectation:

```cpp
bool IsAlwaysDuplicable(const HloInstruction& instruction) {
  return (instruction.opcode() == HloOpcode::kConvert && widening) ||
         instruction.opcode() == HloOpcode::kBroadcast;
}

```

This creates a policy exception path where duplication occurs even when explicitly disabled by the caller. The code itself acknowledges the inconsistency:

```cpp
// TODO(jlebar): Duplicating instructions when we have a variable called "may
// duplicate" that's equal to false is not pretty.

```

This is not merely cosmetic:

1. It breaks the semantic contract of the flag.
2. It forces maintainers to remember undocumented exceptions.
3. It makes fusion behavior harder to reason about, debug, and evolve.

In practice, may_duplicate_ no longer means "no duplication," but rather "no duplication except for a hardcoded whitelist."

**Problem 2: Ignoring Index-Generation and Register Costs**

The rationale for IsAlwaysDuplicable appears to be primarily memory-centric:
"Duplicating these ops reduces memory traffic."

This assumption is incomplete. "Memory-cheap" does not imply "compute-cheap."

For kBroadcast in particular, duplication can significantly increase compute cost:

* High-rank broadcasts require repeated index reconstruction at every consumer.
* Index computation often involves div/mod or equivalent arithmetic operations per element.
* These operations increase register pressure, instruction count, and can reduce occupancy.

[...]

PR #41501, opened 2026-04-23
xla/profiler: Support CUPTI 13.2 multi-subscriber tracing

zhenying-liu:
📝 Summary of Changes
Add XLA support for CUPTI 13.2 multi-subscriber tracing. The change uses CUPTI V2 subscribe/activity APIs when available, falls back to the existing CUPTI V1 path when needed, and tracks the runtime subscription mode so activity handling and teardown stay consistent.

🎯 Justification
This allows XLA profiling to coexist with other CUPTI subscribers in the same process on CUDA/CUPTI 13.2. It benefits workloads that run XLA profiling alongside another CUPTI-based tool.

🚀 Kind of Contribution

N/A

🧪 Unit Tests:
Extended xla/backends/profiler/gpu/cupti_error_manager_test.cc to cover runtime V1/V2 subscription selection and the matching activity setup/error-handling paths.

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/backends/profiler/gpu/cupti_interface.h
+++ b/xla/backends/profiler/gpu/cupti_interface.h
@@ -19,8 +19,10 @@ limitations under the License.
 #include <cstddef>
 #include <cstdint>
 
-#include "third_party/gpus/cuda/extras/CUPTI/include/cupti.h"
+#include "third_party/gpus/cuda/extras/CUPTI/include/cupti_activity.h"
+#include "third_party/gpus/cuda/extras/CUPTI/include/cupti_callbacks.h"
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_profiler_target.h"
+#include "third_party/gpus/cuda/extras/CUPTI/include/cupti_result.h"
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_target.h"
 #include "third_party/gpus/cuda/include/cuda.h"
 
@@ -58,6 +60,15 @@ struct CUpti_PmSampling_CounterData_GetSampleInfo_Params;
 namespace xla {
 namespace profiler {
 
+// CUPTI 13.2 V2 callback metadata parameters are intentionally erased to void*
+// so this common interface can compile with older CUPTI headers.
+using CuptiBuffersCallbackRequestFuncV2 =
+    void(CUPTIAPI*)(uint8_t** buffer, size_t* size, size_t* max_num_records,
+                    void* buffer_request_info);
+using CuptiBuffersCallbackCompleteFuncV2 =
+    void(CUPTIAPI*)(uint8_t* buffer, size_t size, size_t valid_size,
+                    void* buffer_complete_info);
+
 // Provides a wrapper interface to every single CUPTI API function. This class
 // is needed to create an easy mock object for CUPTI API calls. All member
 // functions are defined in the following order: activity related APIs, callback
@@ -65,9 +76,9 @@ namespace profiler {
 // the order in the original CUPTI documentation.
 class CuptiInterface {
  public:
-  CuptiInterface() {}
+  CuptiInterface() = default;
 
-  virtual ~CuptiInterface() {}
+  virtual ~CuptiInterface() = default;
 
   // CUPTI activity API
   virtual CUptiResult ActivityDisable(CUpti_ActivityKind kind) = 0;
@@ -90,6 +101,30 @@ class CuptiInterface {
   virtual CUptiResult ActivityRegisterCallbacks(
       CUpti_BuffersCallbackRequestFunc func_buffer_requested,
       CUpti_BuffersCallbackCompleteFunc func_buffer_completed) = 0;
+
+  // CUDA 13.2 CUPTI V2 multi-subscriber variants.
+  // These must be used alongside cuptiSubscribe_v2; mixing V1 and V2 activity
+  // APIs is unsupported and causes undefined behavior.
+  virtual CUptiResult ActivityRegisterCallbacksV2(
+      CUpti_SubscriberHandle subscriber,
+      CuptiBuffersCallbackRequestFuncV2 func_buffer_requested,
+      CuptiBuffersCallbackCompleteFuncV2 func_buffer_completed) = 0;
+
+  virtual CUptiResult ActivityEnableV2(CUpti_SubscriberHandle subscriber,
+                                       CUpti_ActivityKind kind, void* cfg) = 0;
+
+  virtual CUptiResult ActivityDisableV2(CUpti_SubscriberHandle subscriber,
+                                        CUpti_ActivityKind kind, void* cfg) = 0;
+
+  virtual CUptiResult ActivitySetAttributeV2(CUpti_SubscriberHandle subscriber,
+                                             CUpti_ActivityAttribute attr,
+                                             size_t* valueSize,
+                                             void* value) = 0;
+
+  virtual CUptiResult ActivityUseSystemThreadIdV2(
+      CUpti_SubscriberHandle subscriber) = 0;
+
+  virtual CUptiResult ActivityUsePerThreadBufferV2() = 0;
 
   virtual CUptiResult ActivityUsePerThreadBuffer() = 0;
 
@@ -114,6 +149,10 @@ class CuptiInterface {
   virtual CUptiResult Subscribe(CUpti_SubscriberHandle* subscriber,
                                 CUpti_CallbackFunc callback,
                                 void* userdata) = 0;
+
+  virtual CUptiResult SubscribeV2(CUpti_SubscriberHandle* subscriber,
+                                  CUpti_CallbackFunc callback,
+                                  void* userdata) = 0;
 
   virtual CUptiResult Unsubscribe(CUpti_SubscriberHandle subscriber) = 0;
 
--- a/xla/backends/profiler/gpu/cupti_tracer.cc
+++ b/xla/backends/profiler/gpu/cupti_tracer.cc
@@ -45,11 +45,8 @@ limitations under the License.
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti.h"
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_activity.h"
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_callbacks.h"
-// Note: can not include cupti_driver_cbid.h because it is not once guarded in
-// cuda 11. Remove comment here once cuda 11 is no longer supported.
-// #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_driver_cbid.h"
+#include "third_party/gpus/cuda/extras/CUPTI/include/cupti_driver_cbid.h"
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_result.h"
-#include "third_party/gpus/cuda/extras/CUPTI/include/cupti_target.h"
 #include "third_party/gpus/cuda/include/cuda.h"
 #include "xla/backends/profiler/gpu/cuda_version_variants.h"
 #include "xla/backends/profiler/gpu/cupti_buffer_events.h"
@@ -316,6 +313,32 @@ void CUPTIAPI ProcessCuptiActivityBuffer(CUcontext context, uint32_t stream_id,
   absl::Status status =
       CuptiTracer::GetCuptiTracerSingleton()->ProcessActivityBuffer(
           context, stream_id, buffer, valid_size);
+  if (!status.ok()) {
+    LOG(ERROR) << status;
+  }
+}
+
+// V2 activity callbacks for cuptiSubscribe_v2 (allowMultipleSubscribers=1).
+
+void CUPTIAPI RequestCuptiActivityBuffer_v2(uint8_t** buffer, size_t* size,
+                                            size_t* maxNumRecords,
+                                            void* /*info*/) {
+  CuptiTracer::GetCuptiTracerSingleton()->RequestActivityBuffer(buffer, size);
+  VLOG(3) << "Requested CUPTI V2 Buffer, buffer=" << std::hex
+          << reinterpret_cast<uintptr_t>(*buffer) << std::dec
+          << " size=" << *size;
+  *maxNumRecords = 0;
+}
+
+void CUPTIAPI ProcessCuptiActivityBuffer_v2(uint8_t* buffer, size_t size,
+                                            size_t valid_size, void* /*info*/) {
+  VLOG(3) << "Processing CUPTI V2 Buffer, buffer:" << std::hex
+          << reinterpret_cast<uintptr_t>(buffer) << std::dec
+          << " size: " << size << " valid_size: " << valid_size;
+  // V2 mode has no per-subscriber context/stream_id; pass nullptr/0.
+  absl::Status status =
+      CuptiTracer::GetCuptiTracerSingleton()->ProcessActivityBuffer(
+          /*context=*/nullptr, /*stream_id=*/0, buffer, valid_size);
   if (!status.ok()) {
     LOG(ERROR) << status;
   }
@@ -1043,8 +1066,12 @@ class CuptiDriverApiHookWithActivityApi : public CuptiDriverApiHook {
 
  private:
   void TrackContext(CUpti_CallbackId cbid, CUcontext ctx) {
-    if (!option_.sync_devices_before_stop) return;
-    if (ctx == nullptr) return;
+    if (!option_.sync_devices_before_stop) {
+      return;
+    }
+    if (ctx == nullptr) {
+      return;
+    }
     absl::MutexLock lock(mutex_);
     if (cbid == CUPTI_DRIVER_TRACE_CBID_cuCtxDestroy_v2 ||
         cbid == CUPTI_DRIVER_TRACE_CBID_cuCtxDestroy) {
@@ -1200,8 +1227,16 @@ void CuptiTracer::Disable() {
     pm_sampling_enabled_ = false;
   }
 
-  DisableApiTracing().IgnoreError();
-  DisableActivityTracing().IgnoreError();
+  // ActivityDisableV2 needs the subscriber handle still valid; disable
+  // activities before API tracing (which calls Unsubscribe) only in the V2
+  // subscriber path.
+  if (using_v2_subscriber_api_) {
+    DisableActivityTracing().IgnoreError();
+    DisableApiTracing().IgnoreError();
+  } else {
+    DisableApiTracing().IgnoreError();
+    DisableActivityTracing().IgnoreError();
+  }
   cupti_interface_->CleanUp();
   Finalize().IgnoreError();
   cupti_driver_api_hook_->SyncAndFlush().IgnoreError();
@@ -1234,6 +1269,7 @@ void CuptiTracer::Disable() {
   collector_ = nullptr;
   option_.reset();
   cupti_driver_api_hook_.reset();
+  using_v2_subscriber_api_ = false;
   tsl::profiler::AnnotationStack::Enable(false);
 }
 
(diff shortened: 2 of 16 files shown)


zhenying-liu:
The CI ROCm / JAX Linux x86 AMD failure seems unrelated with my change:
```
RuntimeError: Unable to initialize backend 'rocm': FAILED_PRECONDITION: No visible GPU devices. (you may need to uninstall the failing plugin package, or set JAX_PLATFORMS=cpu to skip this backend.)
RuntimeError: Unknown backend: 'gpu' requested, but no platforms that are instances of gpu are present. Platforms present are: cpu 
```

xla/backends/profiler/gpu/cupti_interface.h
```diff
@@ -115,6 +145,8 @@ class CuptiInterface {
                                 CUpti_CallbackFunc callback,
                                 void* userdata) = 0;
 
+  virtual bool SubscribeUsedV2() const = 0;
```
zhangdaleig: This seems from name keep some state here? If so could you make the state out?

xla/backends/profiler/gpu/cupti_interface.h
```diff
@@ -20,10 +20,17 @@ limitations under the License.
...
 
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti.h"
+#include "third_party/gpus/cuda/extras/CUPTI/include/cupti_activity.h"
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_profiler_target.h"
 #include "third_party/gpus/cuda/extras/CUPTI/include/cupti_target.h"
 #include "third_party/gpus/cuda/include/cuda.h"
 
+// XLA_CUPTI_HAS_MULTI_SUBSCRIBER is defined when the CUPTI V2 multi-subscriber
+// APIs are available (CUPTI_API_VERSION >= 130200, i.e. CUDA 13.2+).
+#if CUPTI_API_VERSION >= 130200
```
zhangdaleig: #if is not suggested, please refer in the build file if_cuda_newer_than

xla/backends/profiler/gpu/cupti_interface.h
```diff
@@ -20,10 +20,17 @@ limitations under the License.
...
+#define XLA_CUPTI_HAS_MULTI_SUBSCRIBER
```
zhangdaleig: No #define #if needed, add all those api interface is fine here.
Also in the cupti_wrapper, refer the the weak function pointer there to control we the real cupti api exists or not.

xla/backends/profiler/gpu/device_tracer_cuda.cc
```diff
@@ -102,8 +102,11 @@ absl::Status GpuTracer::DoStart() {
                      &options_.enable_activity_hardware_tracing)
       .IgnoreError();
 
-// CUDA/CUPTI 10 have issues (leaks and crashes) with CuptiFinalize.
-#if CUDA_VERSION >= 11000
+// CUDA/CUPTI 10 have issues (leaks and crashes) with cuptiFinalize.
+// On builds with CUPTI V2 multi-subscriber support, the runtime V2 path must
+// also avoid cuptiFinalize because it tears down global CUPTI state needed
+// by subsequent V2 subscriptions.
+#if CUDA_VERSION >= 11000 && !defined(XLA_CUPTI_HAS_MULTI_SUBSCRIBER)
```
zhangdaleig: also here, do not add more !defined in new code.

xla/backends/profiler/gpu/cupti_tracer.cc
```diff
@@ -321,6 +321,35 @@ void CUPTIAPI ProcessCuptiActivityBuffer(CUcontext context, uint32_t stream_id,
   }
 }
 
+#ifdef XLA_CUPTI_HAS_MULTI_SUBSCRIBER
```
zhangdaleig: You could add interface like IsCuptiSupportSubscribeV2(uint8 * support_subscribe_v2) to avoid this #ifdef

xla/backends/profiler/gpu/cupti_interface.h
zhenying-liu: Done. I removed `SubscribeUsedV2()` and the wrapper-owned `subscribe_used_v2_` state. `Subscribe() `now reports the selected path through a `bool* used_v2` out-param, so the state stays with the caller/tracer instead of inside `CuptiWrapper`.

xla/backends/profiler/gpu/cupti_interface.h
zhenying-liu: Done. The V2 APIs are now part of the common interface without the macro. CUPTI 13.2-only callback metadata types are erased behind wrapper-local aliases, and `CuptiWrapper::Subscribe() `attempts V2 first, then falls back to V1 when V2 is unsupported or unavailable from the loaded CUPTI library.

xla/backends/profiler/gpu/cupti_tracer.cc
zhenying-liu: I avoided adding a separate `IsCuptiSupportSubscribeV2()` query because the tracer needs to know whether the actual subscription used V2, not only whether V2 appears available. `Subscribe()` now attempts V2, falls back to V1 when needed, and returns the selected path through `used_v2`. This removes the #ifdef from cupti_tracer.cc while keeping the decision tied to the real subscription result.

xla/backends/profiler/gpu/device_tracer_cuda.cc
zhenying-liu: Done. I removed the !defined(XLA_CUPTI_HAS_MULTI_SUBSCRIBER) compile-time condition. device_tracer_cuda.cc now keeps the existing CUDA >= 11 `cupti_finalize `setting, and `CuptiTracer::Finalize()` decides at runtime to skip `cuptiFinalize()` when the V2 subscriber path was actually used.

xla/backends/profiler/gpu/cupti_interface.h
zhenying-liu: Done. I removed the custom XLA_CUPTI_HAS_MULTI_SUBSCRIBER macro and the conditional declarations from cupti_interface.h. The wrapper interface now always declares the V2 methods, and cupti_wrapper_v2.cc contains the CUPTI 13.2-specific calls plus fallback implementations for older headers. I did not use if_cuda_newer_than here because the code needs to compile based on whether the CUPTI headers define these symbols, which may not always match the configured CUDA version.

zhenying-liu:
The CI ROCm failure was due to:
`rocm_platform.cc:49] failed call to hipInit: HIP_ERROR_NoDevice`

xla/backends/profiler/gpu/cupti_wrapper_v2.cc
zhangdaleig, line 1: I think that #if CUPTI_API_VERSION >= 130200 is not prefered as #if CUDA_VERSION...

Seems this cupti_wrapper_v2.cc is not needed. As mentioned early, you could declar weak function pointers like PM Sampling Interface cupti ptrs, like:

[[gnu::weak]] CUptiResult cuptiProfilerHostInitialize(
    CUpti_Profiler_Host_Initialize_Params* params);

Declare:
[[gnu::weak]] CUptiResult cuptiActivityEnable_v2(
    CUpti_SubscriberHandle subscriber,
    CUpti_ActivityKind kind, void* cfg);

It will be null if cupti shared object or static lib using during compiling/linking do not have it.
In the cupti_wrapper calls, check these are null or not first, call it if not null.

zhenying-liu: Done. I removed cupti_wrapper_v2.cc and folded the V2 CUPTI wrappers into cupti_wrapper.cc.

The optional V2 CUPTI APIs are now declared as [[gnu::weak]] symbols, following the existing PM Sampling / Profiler Host API pattern. Each wrapper checks the weak symbol for nullptr and returnsCUPTI_ERROR_NOT_SUPPORTED when the symbol is unavailable, so Subscribe falls back to V1.

I kept CUDA_VERSION >= 13020 guards only around code that depends on CUDA 13.2 header types such as CUpti_SubscriberParams and CUpti_ActivityConfig.

xla/backends/profiler/gpu/cupti_wrapper.cc
```diff
@@ -31,6 +32,30 @@ limitations under the License.
 namespace xla {
 namespace profiler {
 
+// CUPTI V2 symbols are optional across CUDA installations. Weak declarations
+// let the wrapper detect missing symbols at runtime and fall back to V1.
+#if CUDA_VERSION >= 13020
```
zhangdaleig: Please remove #if CUDA_VERSION

xla/backends/profiler/gpu/cupti_wrapper.cc
```diff
@@ -67,6 +92,90 @@ CUptiResult CuptiWrapper::ActivityRegisterCallbacks(
...
+      reinterpret_cast<CUpti_BuffersCallbackCompleteFunc_v2>(
+          func_buffer_completed));
+#else
+  return CUPTI_ERROR_NOT_SUPPORTED;
+#endif  // CUDA_VERSION >= 13020
+}
+
+CUptiResult CuptiWrapper::ActivityEnableV2(CUpti_SubscriberHandle subscriber,
+                                           CUpti_ActivityKind kind, void* cfg) {
+#if CUDA_VERSION >= 13020
```
zhangdaleig: If worry  that the API is not stable even exists, make the checking on dynamic returned cuda version. If not just remove the #if CUDA_VERSION.

xla/backends/profiler/gpu/cupti_wrapper.cc
```diff
@@ -112,10 +221,49 @@ CUptiResult CuptiWrapper::EnableDomain(uint32_t enable,
   return cuptiEnableDomain(enable, subscriber, domain);
 }
 
+CUptiResult CuptiSubscribeV2(CUpti_SubscriberHandle* subscriber,
+                             CUpti_CallbackFunc callback, void* userdata,
+                             bool* used_v2) {
+  if (used_v2 != nullptr) {
+    *used_v2 = false;
+  }
+#if CUDA_VERSION >= 13020
```
zhangdaleig: all these CUDA_VERSION should be removed.

xla/backends/profiler/gpu/cupti_interface.h
```diff
@@ -112,8 +146,8 @@ class CuptiInterface {
                                    CUpti_CallbackDomain domain) = 0;
 
   virtual CUptiResult Subscribe(CUpti_SubscriberHandle* subscriber,
```
zhangdaleig, line 149: Do not let the API handle preference on v2 or v1. Leave this api working on v1. But in tracer, you could using a option with default value to choose the preference.

zhangdaleig: check type: CuptiTracerOptions

xla/backends/profiler/gpu/cupti_wrapper.cc
zhenying-liu: Done. I removed the CUDA_VERSION >= 13020 guards from cupti_wrapper.cc.

xla/backends/profiler/gpu/cupti_wrapper.cc
zhenying-liu: Done. I removed the CUDA_VERSION guard instead of adding a dynamic CUDA-version check.

The wrapper now relies on weak-symbol availability: if a CUPTI V2 entry point is not present, the weak symbol is null and the wrapper returns CUPTI_ERROR_NOT_SUPPORTED. This avoids tying the decision to the CUDA runtime version, which may not match the actual CUPTI symbols available to the process.

xla/backends/profiler/gpu/cupti_wrapper.cc
zhenying-liu: Done. I removed the CUDA_VERSION checks from cupti_wrapper.cc.

The V2 CUPTI symbols are now weak-declared unconditionally in that file. The wrapper checks each weak symbol for null before calling it and returns CUPTI_ERROR_NOT_SUPPORTED when the symbol is unavailable. For older CUPTI headers that do not define the CUDA 13.2 V2 parameter types, cupti_wrapper.cc uses local ABI-compatible aliases/structs so the code still builds while runtime availability is decided by the weak symbols.

xla/backends/profiler/gpu/cupti_interface.h
zhenying-liu: Done. I restored Subscribe() to be the V1 API only and added a separate SubscribeV2() API for the multi-subscriber path.

The V1/V2 preference is now handled in CuptiTracer through CuptiTracerOptions::prefer_cupti_v2, which defaults to true. The tracer tries SubscribeV2() first when that option is enabled and falls back to V1 Subscribe() only when V2 is unavailable/not supported.

zhenying-liu:
//xla/backends/profiler/gpu:cupti_error_manager_test_nvgpu_any failed in XLA Linux x86 GPU L4. It seemed to be a timeout. I tried locally on GB200 with arm and it passed.

zhangdaleig (approved):
LGTM. One thing to mention. too many unrelated changes on formatting.

zhenying-liu:
> LGTM. One thing to mention. too many unrelated changes on formatting.

Thanks. The formatting changes are to please the Clang Tidy, otherwise, it will report errors.

zhenying-liu:
Thanks for approving CUPTI 13.2 multi-subscriber support in XLA. 

Please note that I also had a JAX PR ([#37117](https://github.com/jax-ml/jax/pull/37117)) to support CUPTI multi-subscriber in JAX/Mosaic, but it exposed a CUDA 13.2 [TokaMax failure](https://github.com/openxla/tokamax/blob/ad75b704b1aaf8702c5b6206844479da085ea619/tokamax/_src/benchmarking_test.py#L145) because CUPTI does not support mixing XLA’s legacy V1 path with JAX/Mosaic’s V2 APIs. CUPTI reported CUPTI_ERROR_MULTIPLE_SUBSCRIBERS_NOT_SUPPORTED, so the JAX change was rolled back.

Now that the XLA CUPTI V2 support is approved, would it be possible to coordinate landing the XLA PR #41501 and the corresponding JAX PR ([#37117](https://github.com/jax-ml/jax/pull/37117)) together, so JAX/XLA does not sit in a mixed V1/V2 state and avoid the TokaMax failure?

allanrenucci:
Please fix CI failures before we can land this.

zhenying-liu:
> Please fix CI failures before we can land this.

Thanks. I was able to reproduce. Will send the fix ASAP.

zhenying-liu:
There are a lot of triton failures due to  third_party/triton/common/llvm_cl910324404.patch which was introduced by upstream/base commit 458f765ee0:
```
Error in patch: Error applying patch /__w/xla/xla/third_party/triton/common/llvm_cl910324404.patch: in patch applied to /__w/.cache/bazel/bazel/_bazel_root/81ae3cd958120fe313fca06cacea3f67/external/triton/test/Conversion/tritongpu_to_llvm_blackwell.mlir: could not apply patch due to CONTENT_DOES_NOT_MATCH_TARGET, error applying change near line 2427
```

allanrenucci (approved):

xla/backends/profiler/gpu/mock_cupti.h
```diff
@@ -16,16 +16,17 @@ limitations under the License.
...
 #define XLA_BACKENDS_PROFILER_GPU_MOCK_CUPTI_H_
 
-#include <stddef.h>
-#include <stdint.h>
-
+#include <cstddef>
 #include <cstdint>
 
-#include "third_party/gpus/cuda/extras/CUPTI/include/cupti.h"
+#include <gmock/gmock.h>
```
allanrenucci, line 22: The BUILD target is missing the corresponding dependency: `@com_google_googletest//:gtest_for_library`

allanrenucci (changes requested):
There seems to be legitimate build failures: https://github.com/openxla/xla/actions/runs/25498398398/job/74824225582?pr=41501

zhenying-liu:
@allanrenucci For the CI/XLA Linux x86 GPU L4 failure:
  I tried to reproduce the CI failure locally but could not reproduce it so far. I ran `bazel test //xla/backends/profiler/
  gpu:cupti_error_manager_test_nvgpu_any` on both a GB200 machine and an Intel host with an RTX6000, using CUDA 13.0 / hermetic CUDA 13.0, and the test passed locally.

  The CI failure still looks like a remote build failure rather than a test runtime failure: ResultStore reports `SPAWN(NON_ZERO_EXIT) CppCompile returned a non-zero exit code when running remotely`. I cannot read the protected bytestream logs from my session. Could someone with access extract the compiler stderr/stdout around the failing `CppCompile` action?

  Action stdout blob:
  `bytestream://remotebuildexecution.googleapis.com/projects/tensorflow-testing/instances/default_instance/
  blobs/8acb05e9a941d7ce9c4b112d038b8f86bd9497d4babad814058681b81aa4ca58/1532`

xla/backends/profiler/gpu/mock_cupti.h
zhenying-liu: > The BUILD target is missing the corresponding dependency: `@com_google_googletest//:gtest_for_library`

@allanrenucci I think I fixed this issue in the latest commit but I could be wrong or there might be some other build failures. Could you please show me all the current error messages of the build failures as I don't have access?

allanrenucci: Do you not have access to the [CI / XLA Linux x86 GPU L4 (pull_request)](https://github.com/openxla/xla/actions/runs/25592498697/job/75132475983?pr=41501#step:6:639) logs?

```
xla/backends/profiler/gpu/profile_with_cuda_kernels_test.cc:113:3: error: use of undeclared identifier 'cuptiGetVersion'; did you mean 'cuDriverGetVersion'?
  113 |   cuptiGetVersion(&cupti_version);
      |   ^~~~~~~~~~~~~~~
      |   cuDriverGetVersion
external/cuda_cudart/include/cuda.h:5128:18: note: 'cuDriverGetVersion' declared here
 5128 | CUresult CUDAAPI cuDriverGetVersion(int *driverVersion);
      |                  ^
xla/backends/profiler/gpu/profile_with_cuda_kernels_test.cc:113:19: error: cannot initialize a parameter of type 'int *' with an rvalue of type 'uint32_t *' (aka 'unsigned int *')
  113 |   cuptiGetVersion(&cupti_version);
      |                   ^~~~~~~~~~~~~~
external/cuda_cudart/include/cuda.h:5128:42: note: passing argument to parameter 'driverVersion' here
 5128 | CUresult CUDAAPI cuDriverGetVersion(int *driverVersion);
      |                                          ^
xla/backends/profiler/gpu/profile_with_cuda_kernels_test.cc:115:45: error: use of undeclared identifier 'CUPTI_API_VERSION'; did you mean 'CUPTI_API_EXIT'?
  115 |             << " and CUPTI_API_VERSION " << CUPTI_API_VERSION;
      |                                             ^~~~~~~~~~~~~~~~~
      |                                             CUPTI_API_EXIT
external/cuda_cupti/include/cupti_callbacks.h:96:3: note: 'CUPTI_API_EXIT' declared here
   96 |   CUPTI_API_EXIT                  = 1,
      |   ^
xla/backends/profiler/gpu/profile_with_cuda_kernels_test.cc:118:53: error: use of undeclared identifier 'CUPTI_API_VERSION'; did you mean 'CUPTI_API_EXIT'?
  118 |       enable_pm_sampling && (cupti_version >= 24 && CUPTI_API_VERSION >= 24);
      |                                                     ^~~~~~~~~~~~~~~~~
      |                                                     CUPTI_API_EXIT
external/cuda_cupti/include/cupti_callbacks.h:96:3: note: 'CUPTI_API_EXIT' declared here
   96 |   CUPTI_API_EXIT                  = 1,
      |   ^
4 errors generated.
```

zhenying-liu: @allanrenucci Apologies for the oversight, and thank you for flagging that. I do have access to the logs. I'll address the CI failures first, then tackle the DWYU errors.

zhenying-liu:
> PR #41501 additions: 759, deletions: 222, total: 981
> 
> ⚠️ This PR has a large delta of over 500. Consider breaking the PR down into smaller PRs for a faster code review.

@allanrenucci Please advise on the next step for the Zizmor / GitHub Actions Security Analysis failure. It appears to be about existing workflow config in .github/workflows/pr_size_check.yml, not this PR’s source changes.

If it can be ignored/overridden, could you please proceed with Copybara/import for the current PR head? Thanks!

allanrenucci (approved):

zhenying-liu:
Could anyone please share the copybara failures?

apivovarov:
For target xla/backends/profiler/gpu:cupti_utils:
'deps' attribute:
  Remove platform:logging

For target xla/backends/profiler/gpu:cupti_buffer_events_test:
'deps' attribute:
  Remove platform:test

zhenying-liu:
> For target xla/backends/profiler/gpu:cupti_utils: 'deps' attribute: Remove platform:logging
> 
> For target xla/backends/profiler/gpu:cupti_buffer_events_test: 'deps' attribute: Remove platform:test

Thanks. I checked these against the parent commit. Both reported deps were already present before this PR, and this PR does not modify `cupti_utils.cc` or `cupti_buffer_events_test.cc`. 

That said, I’ll try a small separate cleanup commit to satisfy the import/DWYU checks. If it does not fix DWYU or causes another Copybara/import issue, I can revert that cleanup commit cleanly.

PR #41504, opened 2026-04-24
[xla] Don't forget memory copies for custom calls with collective operands

ezhulenev:
Don't forget to correctly handle custom calls that request collective memory operands.


bchetioui (approved):
Is there a test that can be added to avoid regressions?

ezhulenev:
@bchetioui test is coming next!

PR #41515, opened 2026-04-24
[xla:gpu] Resolve multimem from SymmetricMemory

ezhulenev:
Use symmetric memory to create multimem instead of using low level APIs directly.


PatriosTheGreat (approved):
I'll onboard this PR since it require changes on our side

Tixxx:
probably need ncclGetLsaMultimemDevicePointer in nccl.symbols file

PR #41561, opened 2026-04-24
[tsl:concurrency] Add automatic Future::Map flattening

ezhulenev:
Add support for automatic mapping of Future<Future<T>> to Future<T>

### DIFF
--- a/xla/tsl/concurrency/future.h
+++ b/xla/tsl/concurrency/future.h
@@ -553,10 +553,21 @@ template <>
 struct IsMappable<void, absl::Status> : public std::true_type {};
 template <typename R, typename U>
 struct IsMappable<R, absl::StatusOr<U>> : public std::is_constructible<R, U> {};
+template <typename R>
+struct IsMappable<R, Future<>> : public IsMappable<R, absl::Status> {};
+template <typename R, typename U>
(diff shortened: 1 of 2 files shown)


ermilovmaxim (approved):

ermilovmaxim (approved):

xla/tsl/concurrency/future.h
```diff
@@ -553,24 +553,38 @@ template <>
 struct IsMappable<void, absl::Status> : public std::true_type {};
 template <typename R, typename U>
 struct IsMappable<R, absl::StatusOr<U>> : public std::is_constructible<R, U> {};
+template <typename R>
+struct IsMappable<R, Future<>> : public std::true_type {};
```
seantalts: I think this would allow us to convert a Future<void> to Future<T> and then promise<T != void>.Set(absl::OkStatus) would fail?
AI suggests this:
```
template <typename R>
struct IsMappable<R, Future<>> : public IsMappable<R, absl::Status> {};
```

ezhulenev: yeah, AI seems to be correct, also updated the type predicate for Future<U>, it must be mappable from StatucOr<U> for consistency

akuegel (approved):

PR #35304, opened 2025-12-15
[XLA:GPU] [plugin] Adding sycl/oneapi plugin support for Intel GPU

ashiqimranintel:
This PR will enable sycl/oneapi plugin support for Intel GPU

PR #41563, opened 2026-04-25
[XLA:GPU] [oneAPI] Plugin Part2: oneAPI plugin registration

ashiqimranintel:
This PR registers oneAPI plugin support. Part 2 of PR https://github.com/openxla/xla/pull/35304


ashiqimranintel:
@akuegel , can you review it?

xla/pjrt/gpu/BUILD
```diff
@@ -597,6 +616,8 @@ cc_library(
         ":se_gpu_pjrt_compiler_cuda_registration",
     ]) + if_rocm([
         ":se_gpu_pjrt_compiler_rocm_registration",
+    ]) + if_sycl([
```
akuegel, line 619: I believe this part is the one that caused the revert. Apparently even if the dependency is behind the if_sycl guard, it will still be analyzed by Bazel and causes a build metric regression on our side.

We should try if we can add this dependency is implementation_deps dependency:

https://bazel.build/versions/7.7.0/reference/be/c-cpp

It looks like this requires setting the bazel build flag `experimental_cc_implementation_deps`

ashiqimranintel: @akuegel I tested bazel xla deps, it's passing, I suggest let's test it on your side, if it is causing problem, we can go towards experimental_cc_implementation_deps path

akuegel: Ok, we can give it another try.

akuegel (approved):

xla/pjrt/gpu/se_gpu_pjrt_compiler_sycl_registration.cc
```diff
@@ -0,0 +1,35 @@
...
+
+STREAM_EXECUTOR_REGISTER_MODULE_INITIALIZER(pjrt_register_se_gpu_compiler, {
+  PjRtRegisterDefaultCompiler(
+      OneapiName(), std::make_unique<StreamExecutorGpuCompiler>(
+                        OneapiId(), stream_executor::sycl::kSyclPlatformId));
+  CHECK_OK(StreamExecutorPlatformIdMapping::Global().AddMapping(
+      stream_executor::sycl::kSyclPlatformId, OneapiId()));
+});
+
+}  // namespace xla
```
akuegel: The linter complains about a missing newline at the end of the file.

ashiqimranintel: added the change

PR #41602, opened 2026-04-27
[xla:se] Delete unused collective memory allocators

ezhulenev:
These are unused and will be replaced by unified VMM allocator.


xla/stream_executor/cuda/cuda_platform.cc
```diff
@@ -130,13 +127,7 @@ absl::StatusOr<StreamExecutor*> CudaPlatform::FindExisting(int ordinal) {
 
 absl::StatusOr<std::unique_ptr<StreamExecutor>>
 CudaPlatform::GetUncachedExecutor(int ordinal) {
-  // TODO(b/468297040): We should not be using DebugOptions here.
-  xla::DebugOptions debug_options = xla::GetDebugOptionsFromFlags();
-  auto executor = std::make_unique<CudaExecutor>(
-      this, ordinal,
-      debug_options.xla_gpu_experimental_enable_nvshmem()
```
akuegel, line 137: I think this change also requires deleting two tests that set this flag to true:

https://github.com/openxla/xla/blob/55236acee8a141681916e04a66bfd3a7a076f54e/xla/pjrt/gpu/BUILD#L373

https://github.com/openxla/xla/blob/55236acee8a141681916e04a66bfd3a7a076f54e/xla/pjrt/gpu/BUILD#L424

Also isn't nccl memory allocator used by default?

akuegel: Ah, sorry, the flag controls not only the allocator. So I guess we can keep the tests, but doesn't it mean the allocator is used when nvshem is turned on?

ezhulenev: This code was never properly plugged into SE, today we allocate nvshmem memory via NvshmemCollectives::Allocate. It will stay for now this way until we can remove it for good.

ezhulenev:
rebased

akuegel (approved):

PR #41620, opened 2026-04-27
[XLA:GPU] autotune inflight deduplication

Finistere:
📝 Summary of Changes
When compiling programs in parallel with `PJRT_Client_Compile`, we would end up re-autotuning the same kernels, which would lead to parallel compilation being slower than sequential.

This patch halved the compilation time of our models as they're split into multiple programs. I've changed the Autotune to use a default global in-flight de-duplication map, a bit similar to the `AutotuneCacheMap` in `xla/service/gpu/autotuning/autotuner_cache.cc`. 

I'm not sure whether I took the best approach. So I'd appreciate any feedback, and I'm happy to make the appropriate changes.

🎯 Justification
We split models into multiple programs, which leads to some duplication of auto-tuning. With this PR, we halved the wall time for compilation.

🚀 Kind of Contribution
Please remove what does not apply: ⚡️ Performance Improvement

🧪 Unit Tests:
I've added a simple unit-test that verifies we're not going through auto-tuning twice with in-flight deduplication.

### DIFF
--- a/xla/backends/autotuner/autotuner.cc
+++ b/xla/backends/autotuner/autotuner.cc
@@ -74,6 +74,12 @@ limitations under the License.
 
 namespace xla {
 
+/*static*/ std::shared_ptr<InFlightTuningMap> InFlightTuningMap::Global() {
+  static auto* global = new std::shared_ptr<InFlightTuningMap>(
+      std::make_shared<InFlightTuningMap>());
+  return *global;
+}
+
 namespace {
 
 tsl::Fprint128 GetFingerprint(const HloInstruction* instr) {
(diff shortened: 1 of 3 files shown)


shyamliagrawal:
Hi Benjamin, can you move this in-flight entry to a class implementing the cache interface? I believe we can create a new layered hierarchical cache implementing the same interface which then can take one in-flight and one persistent cache.

IIUC, in-flight cache with the current interface can be achieved if the first lookup returns a miss while the subsequent ones just wait for the cache insert to happen.

Finistere:
Ok, thank you. I'll rework the PR in that direction

dimitar-asenov:
@Finistere  I'll mark this PR as a draft. Once you rework it, please mark it as ready for review again.

PR #41626, opened 2026-04-27
[XLA:GPU] Skip and coalesce VA remapping for unchanged physical handles.

shawnwang18:
This PR improves GPU command-buffer VA remapping by preserving the existing reserved VA mapping across executions and only remapping the allocation slices whose backing physical allocation changed.

  Previously, the VA remapping path rebuilt the mapping too broadly on each execution. That is correct, but it does unnecessary driver work when most command-buffer allocations keep the same backing memory across steps. This PR adds enough allocation identity tracking to detect unchanged mappings safely, then introduces a selective remap path that skips unchanged slices and coalesces access updates for changed
  ranges.

  ## Motivation

  Command-buffer execution uses a stable virtual address range so command buffers can be reused even when XLA buffer allocations move between executions. However, not every allocation changes on every execution. Reissuing unmap/map/access calls for unchanged slices adds overhead, especially because `cuMemSetAccess` can be expensive.

  A naive optimization would compare raw `MemoryAllocation*` pointers or virtual addresses and skip remapping when they match. That is not safe enough because allocator reuse can create ABA-style cases: a pointer/address can appear unchanged while actually referring to a new allocation lifetime. This PR fixes that by tracking allocator-owned allocation IDs and comparing those IDs, not just raw pointers.

  ## Design

  The VMM allocator now exposes `DeviceAddressVmmAllocator::GetAllocationInfo`, which returns:

  - the current physical `MemoryAllocation*`
  - a stable `allocation_id`
  - the mapped allocation size

  The `allocation_id` is stable when a pending deallocation is immediately reused, since that is the same allocator entry and stream ordering makes reuse safe. When an allocation is fully freed and later recreated, the allocator assigns a new ID, even if the virtual address or pointer value is reused. This gives the VA remapping path an ABA-safe identity check.

  `GpuExecutable` stores the previous successful mapping state for command-buffer allocations. On each execution it compares the current `(allocation_id, mapped_size)` against the prior state:

  - unchanged slices are preserved
  - changed slices are marked for remap
  - adjacent changed slices are grouped so `SetAccess` can be coalesced

  The first execution still uses `MemoryReservation::MapTo` to build the full mapping. Later executions call the new `MemoryReservation::Remap` API, passing the existing `ScopedMapping` plus per-slice remap descriptors.

  ## MemoryReservation Remap Behavior

  `MemoryReservation::Remap` takes ownership of the existing full-range `ScopedMapping`. It validates that descriptors are contiguous and non-null, then walks the range in order.

  For each unchanged slice, it leaves the current mapping intact. For each changed slice, it performs the required unmap/map operation. Consecutive changed slices are treated as one run for `SetAccess`, reducing driver calls while preserving the per-allocation map/unmap granularity required by CUDA VMM.

  Failure handling is intentionally conservative. If a remap fails after consuming the previous `ScopedMapping`, `Remap` unmaps every slice described by the request. This avoids leaving a partially mapped VA range without an RAII owner. `GpuExecutable` clears its cached mapping state on remap failure so recovery can rebuild from a fresh full mapping.

  ## Logging and Debuggability

  The VA remap path now logs whether each command-buffer allocation was skipped or remapped, including allocation ID transitions and a short allocation type label. The allocation type labeling logic now lives on `BufferAllocation::AllocationTypeLabel()` instead of as a GPU-local helper, since it is derived entirely from `BufferAllocation` state.

  The summary log reports how many allocations were skipped and how many remap runs were coalesced.

  ## Tests

  This PR adds coverage for:

  - unchanged remap descriptors issuing no driver calls
  - partial remaps only touching changed slices
  - adjacent changed slices coalescing `SetAccess`
  - invalid remap descriptors being rejected
  - remap failure cleanup unmapping all slices
  - allocation ID stability across pending-deallocation reuse
  - command-buffer VA remapping logs confirming the skip/remap path is used

  ## Testing

  - `git diff --check`
  - `bazel --batch test //xla/stream_executor/cuda:cuda_memory_reservation_test //xla/stream_executor/cuda:cuda_device_address_vmm_allocator_test`
  - `bazel --batch build //xla/service/gpu:gpu_executable`
  - `bazel --batch build //xla/service:buffer_assignment //xla/service/gpu:gpu_executable`

### DIFF
--- a/xla/service/buffer_assignment.cc
+++ b/xla/service/buffer_assignment.cc
@@ -461,6 +461,28 @@ static const HloInstruction* GetOutputInstruction(
   return nullptr;
 }
 
+absl::string_view BufferAllocation::AllocationTypeLabel() const {
+  if (is_constant()) {
+    return "constant";
+  }
+  if (is_entry_computation_parameter()) {
+    return "parameter";
+  }
+  if (is_thread_local()) {
+    return "thread_local";
+  }
+  if (is_tuple()) {
+    return "tuple";
+  }
+  if (maybe_live_out()) {
+    return "maybe_live_out";
+  }
+  if (IsPreallocatedTempBuffer()) {
+    return "temp";
+  }
+  return "other";
+}
+
 std::string BufferAllocation::ToShortString(bool human_readable_size) const {
   std::string output;
   if (human_readable_size) {
--- a/xla/service/buffer_assignment.cc
+++ b/xla/service/buffer_assignment.cc
@@ -464,20 +464,15 @@ static const HloInstruction* GetOutputInstruction(
 absl::string_view BufferAllocation::AllocationTypeLabel() const {
   if (is_constant()) {
     return "constant";
-  }
-  if (is_entry_computation_parameter()) {
+  } else if (is_entry_computation_parameter()) {
     return "parameter";
-  }
-  if (is_thread_local()) {
+  } else if (is_thread_local()) {
     return "thread_local";
-  }
-  if (is_tuple()) {
+  } else if (is_tuple()) {
     return "tuple";
-  }
-  if (maybe_live_out()) {
+  } else if (maybe_live_out()) {
     return "maybe_live_out";
-  }
-  if (IsPreallocatedTempBuffer()) {
+  } else if (IsPreallocatedTempBuffer()) {
     return "temp";
   }
   return "other";
--- a/xla/stream_executor/cuda/cuda_device_address_vmm_allocator_test.cc
+++ b/xla/stream_executor/cuda/cuda_device_address_vmm_allocator_test.cc
@@ -31,7 +31,6 @@ limitations under the License.
 #include "xla/stream_executor/stream.h"
 #include "xla/stream_executor/stream_executor.h"
 #include "xla/stream_executor/vmm_device_address_allocator.h"
-#include "xla/tsl/platform/statusor.h"
 
 namespace stream_executor {
 namespace {
@@ -77,12 +76,12 @@ class DeviceAddressVmmAllocatorTest : public ::testing::Test {
 };
 
 TEST_F(DeviceAddressVmmAllocatorTest, AllocateAndDeallocate) {
-  TF_ASSERT_OK_AND_ASSIGN(
+  ASSERT_OK_AND_ASSIGN(
       auto allocator,
       gpu::CudaDeviceAddressVmmAllocator::Create(executor_, stream_.get()));
 
   // Allocate memory.
-  TF_ASSERT_OK_AND_ASSIGN(
+  ASSERT_OK_AND_ASSIGN(
       auto scoped_address,
       allocator->Allocate(executor_->device_ordinal(), 1024,
                           /*retry_on_failure=*/true,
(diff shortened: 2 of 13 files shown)


xla/stream_executor/memory_reservation.h
```diff
@@ -99,6 +104,27 @@ class MemoryReservation {
...
+  };
+
+  // Re-maps a contiguous reservation range described by `mappings`.
+  // `existing` must cover the same full range. For each mapping where
+  // remap_required is false, the current mapping is preserved. For the rest,
+  // the slice is unmapped and mapped to `allocation`; SetAccess is invoked
+  // once per maximal run of consecutive remapped mappings. On failure,
+  // Remap unmaps all slices described by `mappings`, leaving the reservation
+  // with no active mapping for the range.
+  absl::StatusOr<ScopedMapping> Remap(
```
ezhulenev: This looks like would be a better added as:

```
ScopedMapping::Remap(absl::Span<const RemapDescriptor> mappings) &&;
```

xla/stream_executor/memory_reservation.h
```diff
@@ -99,6 +104,27 @@ class MemoryReservation {
   absl::StatusOr<ScopedMapping> MapTo(
       absl::Span<const MappingDescriptor> mappings);
 
+  // Describes a desired post-Remap mapping for a slice of the reservation.
+  struct RemapDescriptor {
```
ezhulenev: nit: `MappingDescriptor` and `RemapDescriptor` consistency.

I'd call both either Mapping/Remapping or Map/Remap

xla/service/buffer_assignment.cc
```diff
@@ -461,6 +461,28 @@ static const HloInstruction* GetOutputInstruction(
   return nullptr;
 }
 
+absl::string_view BufferAllocation::AllocationTypeLabel() const {
+  if (is_constant()) {
+    return "constant";
+  }
+  if (is_entry_computation_parameter()) {
```
ezhulenev: `else if` chain? it will a little shorter

xla/stream_executor/cuda/cuda_device_address_vmm_allocator_test.cc
```diff
@@ -287,6 +296,40 @@ TEST_F(DeviceAddressVmmAllocatorTest,
...
+TEST_F(DeviceAddressVmmAllocatorTest,
+       AllocationInfoIdStableAcrossPendingDeallocationReuse) {
+  TF_ASSERT_OK_AND_ASSIGN(
+      auto allocator,
+      gpu::CudaDeviceAddressVmmAllocator::Create(executor_, stream_.get()));
+
+  const int ordinal = executor_->device_ordinal();
+  constexpr uint64_t kSize = 1024;
+
+  TF_ASSERT_OK_AND_ASSIGN(
```
ezhulenev: prefer regular ASSERT_OK_AND_ASSIGN in all touched files

xla/stream_executor/memory_reservation.h
shawnwang18: Good point, putting it into ScopedMapping seems more intuitive.  fixed.

xla/stream_executor/memory_reservation.h
shawnwang18: Fixed. using Mapping/Remapping

xla/stream_executor/cuda/cuda_device_address_vmm_allocator_test.cc
shawnwang18: fixed.

xla/service/buffer_assignment.cc
shawnwang18: Fixed.

shawnwang18:
Current design has an issue that keep the physical-virtual mapping across steps will pin the physical memory, so will be more easy to get OOM. Will think idea if we can resolve the issue but still keeps the mapping across steps.

PR #41647, opened 2026-04-27
[XLA:GPU] Deterministic autotuner selection without a trusted backend

and-ivanov:
📝 Summary of Changes
Make the autotuner's pass/fail outcome deterministic when no codegen backend provides a trusted reference output.
Every successful profile's output is assigned to a cluster: either an existing one whose representative matches (under relative_tolerance) or a new one. Each cluster tracks has_trusted_member: set iff at least one member comes from a backend with CanProduceWrongResults() = false. After the profile loop, the winning cluster is picked: among clusters containing a trusted member, the largest one wins. If no cluster has a trusted member, the largest overall wins (earliest insertion breaks ties). Non-winning cluster members are demoted with kWrongResults. The existing PickBestConfig then selects the fastest survivor, i.e. the fastest member of the winning cluster.

**Assumption:** most configs implement the same algorithm and most of them produce correct output. The "majority" cluster is therefore correct. The additional memory/compute cost is low: number of cluster is small.

🎯 Justification
  The flake we observed: `TritonTest.FuseChannelDequantization` drives an int4 GEMM fusion through `AutotunerPass`. For that fusion only the Triton backend emits any configs (CUDNN / cuBLAS / cuBLAS-LT produce zero candidates, and
  CUBLASLT_FISSION is disabled in the test). Every Triton config has `CanProduceWrongResults() == true`, so post-profile the autotuner has no trustworthy reference to compare against. The selection rule reduces to "pick the fastest config", and when two configs with differing outputs have near-equal timings (the repro had configs at ~12 µs where one produced correct output and one produced wrong output) measurement noise decides which one wins.

🚀 Kind of Contribution
🐛 Bug Fix

N/A — correctness/determinism fix. With `check_buffers = true`, per-iteration work goes from one buffer comparison to between 1 and C output comparisons (C = number of clusters seen so far, expected to be small with stated assumptions).
memory overhead is C-1 extra output buffers held for the pass duration. The `check_buffers = false` path is unchanged.
**Possible extension:** We can add parameter to cap C. This will still guarantee determinism but not majority selection. With max_C=1 it resolves to the "first config is a reference" strategy, removing overhead but sacrificing correctness under the assumption of correct majority.

🧪 Unit Tests:
  `xla/backends/autotuner/autotuner_test.cc`:
  - `AutotuneClustersOutputsWhenAllBackendsUntrusted`: three untrusted configs split into a 2-member majority cluster and a 1-member minority cluster whose member is the globally fastest. Verifies the majority cluster's fastest member is picked, not the global fastest.
  - `AutotuneTrustedClusterWinsOverLargerUntrustedCluster`: one trusted config vs three agreeing-but-untrusted configs (3× larger cluster and faster). Verifies the trusted cluster wins.
  - Existing `AutotuneWithBufferCheckFiltersWrongResults` updated to the unified clustering path.

### DIFF
--- a/xla/backends/autotuner/autotuner.cc
+++ b/xla/backends/autotuner/autotuner.cc
@@ -825,7 +825,9 @@ int Autotuner::AssignToOutputCluster(std::vector<OutputCluster>& clusters,
       return c;
     }
   }
-  if (!allow_new_cluster) return -1;
+  if (!allow_new_cluster) {
+    return -1;
+  }
   int new_idx = clusters.size();
   clusters.push_back(OutputCluster{std::move(output), /*count=*/1,
                                    /*has_trusted_member=*/is_trusted_config});
@@ -835,15 +837,21 @@ int Autotuner::AssignToOutputCluster(std::vector<OutputCluster>& clusters,
 void Autotuner::DemoteNonWinningClusterConfigs(
     std::vector<ConfigResult>& results,
     const std::vector<OutputCluster>& clusters) {
-  if (clusters.empty()) return;
+  if (clusters.empty()) {
+    return;
+  }
   // Prefer clusters with a trusted member; among the preferred set, pick the
   // largest count (earliest insertion wins on tie).
   const bool any_trusted = absl::c_any_of(
       clusters, [](const OutputCluster& c) { return c.has_trusted_member; });
   int winner = -1;
   for (int c = 0; c < clusters.size(); ++c) {
-    if (any_trusted && !clusters[c].has_trusted_member) continue;
-    if (winner < 0 || clusters[c].count > clusters[winner].count) winner = c;
+    if (any_trusted && !clusters[c].has_trusted_member) {
+      continue;
+    }
+    if (winner < 0 || clusters[c].count > clusters[winner].count) {
+      winner = c;
+    }
   }
   VLOG(2) << "Output clustering formed " << clusters.size()
           << " cluster(s); selected cluster " << winner << " with "
--- a/xla/backends/autotuner/autotuner.cc
+++ b/xla/backends/autotuner/autotuner.cc
@@ -22,6 +22,7 @@ limitations under the License.
 #include <iterator>
 #include <limits>
 #include <memory>
+#include <numeric>
 #include <optional>
 #include <string>
 #include <utility>
@@ -636,8 +637,7 @@ Autotuner::CompileAll(const HloInstruction* instr,
 
 absl::StatusOr<std::vector<Autotuner::ConfigResult>> Autotuner::ProfileAll(
     std::vector<ExecutableCandidate> candidates, const HloInstruction* instr) {
-  std::vector<ConfigResult> config_results;
-  config_results.reserve(candidates.size());
+  std::vector<ConfigResult> config_results(candidates.size());
 
   absl::MutexLock lock(profiler_m_);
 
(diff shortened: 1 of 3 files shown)


xla/backends/autotuner/autotuner.cc
```diff
@@ -632,13 +632,15 @@ absl::StatusOr<std::vector<Autotuner::ConfigResult>> Autotuner::ProfileAll(
...
-    reference_output = GetReferenceOutput(candidates, *input_buffers);
-    if (!reference_output.has_value()) {
-      LOG(WARNING) << "No reference output found even though buffer checking ";
-    }
+    VLOG(2) << "Validating outputs via clustering across " << candidates.size()
+            << " config(s).";
+  }
+
+  std::vector<OutputCluster> clusters;
+  std::vector<int> cluster_of_result;
```
shyamliagrawal: nit: cluster_of_results is a bit confusing, some suggestions: cluster_indices_of_results or result_to_cluster_map

shyamliagrawal: Or rather, you can add cluster_index in ConfigResult struct, it will simplify a lot of things.

and-ivanov: done

shyamliagrawal:
Hi Andrei,

Apologies for the delay, I was trying benchmarks and I noticed (~50%) regression in compile time in our GeMMs microbenchmarks. Did you try to compile time benchmarks and observe any similar regression?

Maybe we can adapt the solution to avoid all this over-work when the reference is actually available, which is true most of the time.
Can you also please sync to the head and resolve the conflict?

and-ivanov:
Thanks @shyamliagrawal! I updated the algorithm to avoid the extra clustering work when a trusted backend/reference is available.

The new flow profiles trusted configs first. If they produce a valid reference cluster, untrusted configs are only compared against the existing trusted cluster(s) and are not allowed to create new output clusters. This restores the original behavior in the common trusted-reference case: we use the trusted output as the single reference cluster instead of doing consensus clustering across all untrusted candidates.

If multiple trusted configs disagree (which should not happen if they are indeed trusted), untrusted configs can still contribute counts to the existing trusted clusters, and the largest trusted cluster wins after all candidates are profiled. If no trusted reference is available (i.e. trusted config with compile error), we fall back to the new clustering behavior among untrusted configs, which preserves the determinism fix.

I also added tests covering order preservation, untrusted voting for trusted clusters, and the fallback behavior when no trusted reference is available.

xla/backends/autotuner/autotuner.cc
```diff
@@ -760,40 +811,58 @@ absl::Status Autotuner::DumpHlo(HloInstruction* instr, const Config& config) {
...
+  int new_idx = clusters.size();
+  clusters.push_back(OutputCluster{std::move(output), /*count=*/1,
+                                   /*has_trusted_member=*/is_trusted_config});
+  return new_idx;
+}
+
+void Autotuner::DemoteNonWinningClusterConfigs(
+    std::vector<ConfigResult>& results,
+    const std::vector<OutputCluster>& clusters) {
+  if (clusters.empty()) return;
```
shyamliagrawal: Can you please move this and below single statement if blocks under braces?

shyamliagrawal (approved):
The compile time on the benchmarks looks good now, thanks!

xla/backends/autotuner/autotuner.cc
and-ivanov: done

PR #41652, opened 2026-04-27
[xla:se] Make CudaVmmAllocator compatible with NCCL

ezhulenev:
Instead of always linking NCCL make our own NCCL-compatible allocator:

Rewrite `CudaVmmAllocator` for NCCL-compatible VMM allocation. Replace the simple bool-based constructor with an `Options` struct controlling alignment, peer access, FABRIC handle support, and RDMA
  capability. Allocation now uses `POSIX_FILE_DESCRIPTOR` handle type as baseline (matching NCCL's `ncclMemAlloc`), with optional `FABRIC` handle that gracefully falls back if not supported. Deallocation now synchronizes via `cuCtxSynchronize()` before unmapping since `cuMemUnmap` immediately invalidates
  virtual address mappings unlike `cuMemFree`.

Add `is_fabric_supported` and `vmm_granularity_` to `CudaExecutor`. Replace `is_vmm_supported_` bool with cached `vmm_granularity_` (zero means unsupported). VMM support is now mandatory — `Init()` returns an error if the device doesn't support it. Query fabric handle support at init time via `CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED`. Detect peer access across all devices and pass it to VMM allocator options.

### DIFF
--- a/xla/stream_executor/cuda/BUILD
+++ b/xla/stream_executor/cuda/BUILD
@@ -1274,11 +1274,14 @@ cc_library(
         "//xla/stream_executor:memory_allocator",
         "//xla/stream_executor:stream_executor_h",
         "//xla/tsl/platform:status_macros",
+        "@com_google_absl//absl/base",
+        "@com_google_absl//absl/base:core_headers",
+        "@com_google_absl//absl/cleanup",
+        "@com_google_absl//absl/log",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings:str_format",
         "@local_config_cuda//cuda:cuda_headers",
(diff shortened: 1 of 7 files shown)


ezhulenev:
Fixed conflicts and rebased

akuegel (approved):

akuegel:
@ezhulenev We get some failing internal tests with this change, therefore we need to roll back. The errors reported are: "platform_util.cc:270] Failed to create stream executor for device CUDA:0: CUDA error: : CUDA_ERROR_INVALID_VALUE: invalid argument

ezhulenev:
@akuegel do you know the CUDA version for the test that failed?

ezhulenev:
Is it with MIG?

PR #41660, opened 2026-04-27
[XLA:GPU] Fix SolLatencyEstimator cost for nested Triton GEMM fusions

sfvaroglu:
📝 Summary of Changes
- Hoist IsTritonGemm into xla/backends/gpu/codegen/triton/support.{cc,h}, which matches both kTritonGemmFusionKind and kTritonNestedGemmFusionKind with a single kDot. 
- In gpu_hlo_schedule.cc::GetLatencyEstimator, wrap ProfileGuidedLatencyEstimator with SolLatencyEstimator (when supported) instead of GpuLatencyEstimator's T-shirt sizes; fall back to GpuLatencyEstimator if SOL can't be built.
- Add MatmulInterpolatorTest.SupportsDotTritonNestedGemmFusion.

🎯 Justification
- IsTritonGemm predicates missed __triton_nested_gemm_fusion, so the SOL cost model assigned no reification cost to nested Triton GEMMs and produced inaccurate latency estimates.
- When PGLE coverage is partial, the wrapped fallback drives un-profiled ops; using SOL there is much more precise than coarse T-shirt sizing.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

N/A

🧪 Unit Tests:
MatmulInterpolatorTest.SupportsDotTritonNestedGemmFusion.

🧪 Execution Tests:
Tested with Llama3.1-405b on B200 (2 x 8)

### DIFF
--- a/xla/backends/gpu/codegen/triton/support.cc
+++ b/xla/backends/gpu/codegen/triton/support.cc
@@ -858,5 +858,18 @@ bool IsTritonFusedComputation(const HloComputation& computation) {
                  .kind() == kTritonGemmFusionKind;
 }
 
+bool IsTritonGemm(const HloInstruction& instr) {
+  if (instr.opcode() != HloOpcode::kFusion ||
+      instr.called_computations().size() != 1) {
+    return false;
+  }
+  if (!IsGpuFusionKind(instr, kTritonGemmFusionKind) &&
+      !IsGpuFusionKind(instr, kTritonNestedGemmFusionKind)) {
+    return false;
+  }
+  return absl::c_count_if(instr.fused_instructions(),
+                          HloPredicateIsOp<HloOpcode::kDot>) == 1;
+}
+
 }  // namespace gpu
 }  // namespace xla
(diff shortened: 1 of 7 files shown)


xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -481,9 +481,38 @@ std::unique_ptr<LatencyEstimator> GetLatencyEstimator(
...
+            mlir_context, std::move(cost_analysis));
+        if (sol.ok()) {
+          base_estimator = std::move(*sol);
+          VLOG(1) << "PGLE fallback: using SolLatencyEstimator";
+        }
+      }
+    }
+    if (base_estimator == nullptr) {
+      base_estimator =
+          gpu_latency_estimator
```
golechwierowicz: If a check on line 500 fails then accessing `gpu_latency_estimator` is UB. Please fix.

xla/backends/gpu/codegen/triton/support.cc
```diff
@@ -858,5 +858,18 @@ bool IsTritonFusedComputation(const HloComputation& computation) {
                  .kind() == kTritonGemmFusionKind;
 }
 
+bool IsTritonGemm(const HloInstruction& instr) {
```
golechwierowicz, line 861: FYI, this change should've been split to two logical changes: one that changes the implementation of the function, and the second as the one which moves the code to support.cc. Please, keep this in mind for the future.

xla/service/gpu/gpu_hlo_schedule.cc
sfvaroglu: Fixed

golechwierowicz (approved):

PR #41757, opened 2026-04-29
Add ReMap API for memory reservation

shawnwang18:
Adds a `MemoryReservation::ScopedMapping::Remap` API for updating an existing reservation mapping in place. The API takes a list of `RemappingDescriptor`s covering the currently mapped range and remaps only the slices marked as changed, while preserving unchanged slices.

  ## Motivation

  Upcoming VA remapping work needs a lower-level reservation API that can update the physical backing of an existing virtual address range without tearing down and rebuilding the entire mapping. This lets callers skip unchanged slices and reduce expensive driver calls, especially `SetAccess`, when only part of a reservation changes.

  ## Details

  This PR adds:

  - `MemoryReservation::RemappingDescriptor`, describing the desired post-remap mapping for each reservation slice.
  - `ScopedMapping::Remap(...)`, which:
    - validates that descriptors are contiguous and cover the existing mapping range,
    - preserves slices where `remap_required == false`,
    - unmaps/maps only slices where `remap_required == true`,
    - coalesces `SetAccess` over maximal runs of adjacent remapped slices,
    - consumes the original `ScopedMapping` and returns a new one for the remapped range.
  - Unit coverage using a fake `CountingReservation` to verify skip/coalesce behavior and failure cleanup without requiring CUDA driver state.

  ## Testing

  - `git diff --check`
  - `bazel build //xla/stream_executor:memory_reservation --noshow_progress`


xla/stream_executor/memory_reservation.cc
```diff
@@ -143,4 +145,104 @@ absl::StatusOr<MemoryReservation::ScopedMapping> MemoryReservation::MapTo(
...
+    }
+    // Coalesce: extend this run as long as the next descriptor is also
+    // being remapped. Map/UnMap stay per-slice (cuMemMap requires per-handle
+    // calls); SetAccess is the expensive call to coalesce.
+    const size_t run_start = mappings[i].reservation_offset;
+    size_t run_size = 0;
+    size_t j = i;
+    while (j < mappings.size() && mappings[j].remap_required) {
+      const auto& dj = mappings[j];
+      TF_RETURN_IF_ERROR(reservation->UnMap(dj.reservation_offset, dj.size));
```
ezhulenev, line 231: prefer RETURN_IF_ERROR in touched files

xla/stream_executor/memory_reservation.cc
```diff
@@ -143,4 +145,104 @@ absl::StatusOr<MemoryReservation::ScopedMapping> MemoryReservation::MapTo(
...
+        }
+      }
+    }
+  });
+
+  // Walk descriptors in order, processing maximal contiguous runs of slices
+  // that require remapping. For each run we issue per-slice UnMap and Map,
+  // then a single SetAccess covering the whole run.
+  size_t i = 0;
+  while (i < mappings.size()) {
```
ezhulenev, line 218: honestly this loop is unredable 😄  I don't have any actionable suggestions, but you know what happens here better then me, any way to make it simpler? Split into prepare/execute loops maybe? Reduce mutable state, i = j and j = i, I'm just lost

ezhulenev (approved):

ezhulenev:
LGTM, API looks clean, loop in implementation is confusing as hell, but not sure how much improvement is possible.

PR #41659, opened 2026-04-27
[xla] Port XLA Backend to use BFC allocator

ezhulenev:
Port XLA backends to use BFC for memory allocation, this path is used in some legacy Tensorflow integrations and can lead to excessive alloc/free cost at run time and wasted memory when base allocator has to do allocation in 2MB granularity.

Update Backend to use MultiDeviceAllocator, similar to how PJRT client does it.

PR #41761, opened 2026-04-29
[xla] Port XLA Backend to use BFC allocator

ezhulenev:
A second attempt to land https://github.com/openxla/xla/pull/41659, this time we use BFC for defalt memory space, and a passthrough allocator for host memory space.

Port XLA backends to use BFC for memory allocation, this path is used in some legacy Tensorflow integrations and can lead to excessive alloc/free cost at run time and wasted memory when base allocator has to do allocation in 2MB granularity.

Update Backend to use MultiDeviceAllocator, similar to how PJRT client does it.


xla/service/backend.cc
```diff
@@ -88,6 +103,79 @@ struct Backend::IntraOpThreadPool {
   std::unique_ptr<Eigen::ThreadPoolDevice> device;
 };
 
+// Thin tsl::Allocator wrapper around StreamExecutor for a given memory space.
+// Used to plug non-default memory spaces (e.g. host pinned) into
+// MultiDeviceAdapter via TfAllocatorAdapter.
```
akuegel, line 110: This class could be in an anonymous namespace?

ezhulenev: oops, fixed!

akuegel (approved):

PR #41175, opened 2026-04-19
[PJRT & GPU] implement `PJRT_Device_ClearMemoryStats`

fl0pedro:
This implements the ability to reset peak memory tracking for PJRT devices. I needed something like this from JAX, but it didn't seem to be implemented in the API. This is useful for quick benchmarks w/o the need to run the full profiler. 

I only implemented this for **GPUs**, and tested it on NVIDIA 3090s. As for CPUs much of these statistics are not implemented, so I followed suit. I do not have access to TPUs, so this was also skipped. 

The test thoroughly analyzes 3 different tensors, two of which are deleted in between and the baseline memory and peak memory throughout. I also included two tests for resetting the peak memory on CPU both of which as expected fail. All this was run with `bazel test //xla/pjrt:pjrt_memory_stats_test --config=cuda --action_env=TF_CUDA_COMPUTE_CAPABILITIES="8.9" --test_output=all`

PR #41778, opened 2026-04-29
add `ClearMemoryStats` to pjrt_c_api

fl0pedro:
Plumbing to expose `PJRT_Device_ClearMemoryStats` (implemented in #41175) through the `PjRtCApiDevice` and IFRT Device interfaces. This is the necessary C API/IFRT follow-up to support resetting peak memory tracking downstream.

### DIFF
--- a/xla/python/pjrt_ifrt/pjrt_device.cc
+++ b/xla/python/pjrt_ifrt/pjrt_device.cc
@@ -76,5 +76,9 @@ absl::Span<Memory* const> PjRtDevice::Memories() const { return memories_; }
 
 int PjRtDevice::ProcessIndex() const { return process_index_; }
 
+absl::Status PjRtDevice::ClearMemoryStats() const {
+  return pjrt_device_->ClearMemoryStats();
+}
+
 }  // namespace ifrt
 }  // namespace xla
--- a/xla/python/pjrt_ifrt/pjrt_device.cc
+++ b/xla/python/pjrt_ifrt/pjrt_device.cc
@@ -77,6 +77,10 @@ absl::Span<Memory* const> PjRtDevice::Memories() const { return memories_; }
 int PjRtDevice::ProcessIndex() const { return process_index_; }
 
 absl::Status PjRtDevice::ClearMemoryStats() const {
+  if (!IsAddressable()) {
+    return absl::FailedPreconditionError(
+        "ClearMemoryStats is only supported on addressable devices.");
+  }
   return pjrt_device_->ClearMemoryStats();
 }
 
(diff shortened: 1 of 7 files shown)


allanrenucci:
@pschuh Can you have a look at this PR? Thanks

pschuh (approved):

allanrenucci (approved):

xla/python/pjrt_ifrt/pjrt_device.cc
```diff
@@ -76,5 +76,9 @@ absl::Span<Memory* const> PjRtDevice::Memories() const { return memories_; }
 
 int PjRtDevice::ProcessIndex() const { return process_index_; }
 
+absl::Status PjRtDevice::ClearMemoryStats() const {
+  return pjrt_device_->ClearMemoryStats();
```
akuegel, line 84: From AI review:
Issue: Potential nullptr dereference. In PjRtDevice, pjrt_device_ is null for non-addressable devices, as documented in the corresponding header file. This method should check if the device is addressable (e.g., by calling IsAddressable() or checking pjrt_device_ != nullptr) before attempting to call ClearMemoryStats(). If the device is not addressable, it should return an appropriate error, such as absl::FailedPreconditionError or absl::UnimplementedError.

Also we do see a segfault in the test //xla/python/pjrt_ifrt:pjrt_client_impl_test_cpu

fl0pedro: Should be fixed now 👍

akuegel (approved):

xla/python/pjrt_ifrt/pjrt_device.cc
```diff
@@ -76,5 +76,13 @@ absl::Span<Memory* const> PjRtDevice::Memories() const { return memories_; }
 
 int PjRtDevice::ProcessIndex() const { return process_index_; }
 
+absl::Status PjRtDevice::ClearMemoryStats() const {
```
allanrenucci, line 79: Reposting internal review comment from @hyeontaek:
> This does not sound like an API that should be defined at the IFRT level because the memory stats are not supported as an IFRT API. This line of work will need to start from defining what memory stats is and how to expose via an IFRT API.
>
> Also this does not follow the style of the IFRT interface that IFRT top-level APIs use pure virtual methods and any definition of the methods goes to the respective IFRT implementations.
>
> If you need to expose this feature only for PjRtDevice only, a short-term way to do it will be do `llvm::dyncast_or_nonnull` on an IFRT device and interact with `xla::ifrt::PjRtCompatibleDevice::pjrt_device()` directly.

akuegel: @fl0pedro did you see this comment?

allanrenucci:
Closing due to inactivity

PR #41784, opened 2026-04-29
Avoid nonnegative inference for signed integer squares

yetval:
Signed integer multiplication can overflow and wrap to a negative value, so treating `multiply(x, x)` as always non-negative is not sound for signed integral element types.

This can let algebraic simplification fold comparisons such as `(x * x) < 0` or remove `abs(x * x)`, changing observable behavior for overflow-sensitive programs.

This change keeps the existing inference for floating-point and unsigned types, but avoids it for signed integral types unless a stronger no-overflow proof is added later.

Regression coverage added for `s8`:

- preserving `abs(multiply(p, p))`
- preserving `compare(multiply(p, p), 0), direction=LT`

Related TensorFlow issue: tensorflow/tensorflow#117142

Local validation:

- `git diff --check`
- Reproduced tensorflow/tensorflow#117142 with TensorFlow 2.21.0 on macOS arm64 CPU

Not run locally:

- Bazel C++ tests; bazel/bazelisk is not installed in this environment.

### DIFF
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
@@ -535,7 +535,8 @@ bool AlgebraicSimplifierVisitor::IsNonNegative(
   }
   switch (hlo->opcode()) {
     case HloOpcode::kMultiply: {
-      return hlo->operand(0) == hlo->operand(1);
+      return hlo->operand(0) == hlo->operand(1) &&
+             !primitive_util::IsSignedIntegralType(hlo->shape().element_type());
     }
     case HloOpcode::kAbs:
     case HloOpcode::kExp:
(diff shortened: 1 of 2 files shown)


allanrenucci:
Thanks for the contribution!

As in C and C++, signed integer overflow should be treated as undefined behavior (UB), although we don't strictly specify that for HLO. The compiler is then free to perform the simplification.

Considering this change can have performance implication, I believe this is not desirable.

yetval:
@allanrenucci Thanks for the clarification.

That makes sense if signed integer overflow is intended to be UB-style in HLO. My concern is that this does not seem to be clearly specified, and the observable TensorFlow behavior is surprising because the compiled function returns the overflowed `prod` value while optimizing a later observer of that same value as if `prod` were non-negative.

If the intended XLA semantics allow this, I agree this PR is probably not the right fix. Would you prefer that I close this and instead open a smaller PR documenting the signed-overflow assumption for HLO/algebraic simplification? Or is there already a canonical place where this is specified that the TensorFlow issue should point to?

allanrenucci:
Yes, let's close this PR. I don't think the behavior you observed is documented anywhere. I suppose it we wanted to document it, it should be part of the StableHLO spec: https://github.com/openxla/stablehlo/blob/main/docs/spec.md#multiply

PR #41799, opened 2026-04-29
[NVIDIA GPU] Add a knob for scheduling collectives as early as possible

Tixxx:
Add a control knob to tell LHS to always schedule collectives before compute instead of relying on overlap limit flag.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -459,6 +459,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_dot_merger_threshold_mb(64);
   opts.set_xla_enable_fast_math(false);
   opts.set_xla_gpu_experimental_parallel_collective_overlap_limit(1);
+  opts.set_xla_gpu_experimental_collective_start_as_early_as_possible(false);
   opts.set_xla_gpu_experimental_parallel_async_compute_limit(2);
   opts.set_xla_pjrt_allow_auto_layout_in_hlo(false);
   opts.set_xla_gpu_enable_scatter_determinism_expander(false);
@@ -2763,6 +2764,14 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       "This controls how many in-flight collectives "
       "latency hiding scheduler can schedule."));
   flag_list->push_back(tsl::Flag(
+      "xla_gpu_experimental_collective_start_as_early_as_possible",
+      bool_setter_for(
+          &DebugOptions::
+              set_xla_gpu_experimental_collective_start_as_early_as_possible),
+      debug_options
+          ->xla_gpu_experimental_collective_start_as_early_as_possible(),
+      "This controls whether collectives should start as early as possible."));
+  flag_list->push_back(tsl::Flag(
       "xla_gpu_experimental_parallel_async_compute_limit",
       int32_setter_for(
           &DebugOptions::set_xla_gpu_experimental_parallel_async_compute_limit),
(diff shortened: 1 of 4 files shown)


xla/xla.proto
```diff
@@ -887,6 +887,9 @@ message DebugOptions {
   //   coll.2-done = collective(coll.2-start)
   optional int32 xla_gpu_experimental_parallel_collective_overlap_limit = 336;
 
+  // Controls whether collectives should start as early as possible.
+  optional bool xla_gpu_experimental_collective_start_as_early_as_possible = 476;
```
PatriosTheGreat: Is there a trade-off between starting early vs delayed start? My initial expectation is that we either way would benefit from an early start, am I wrong here? Or is this flag needed to migrate to an early start under experiment, and it will be removed later?

Tixxx: Yea, it works well in compute bound cases where a few collectives are exposed, this will help achieve almost perfect overlap. In some extreme communication bound scenarios, even though the scheduling looks good, some compute kernel launches are pushed back during runtime since the collectives are bundled too closely which consume a lot of the resources on GPU. The resource utilization is being addressed in comms library so for now we will need a flag to toggle this scheduling behavior for different scenarios.

xla/service/gpu/gpu_hlo_schedule.cc
```diff
@@ -672,13 +642,11 @@ absl::Status RunLatencyHidingSchedulerPasses(
...
-  // estimator is used we will tell LHS to extend async-done intervals as much
-  // as possible to start collectives as early as possible.
-  bool prioritize_compute_over_async_start = false;
-  if (config.parallel_collective_overlap_limit > 1) {
-    prioritize_compute_over_async_start = true;
-  }
+  // It's more beneficial to prioritize compute over async start when we have
+  // more async ops in parallel.
+  bool prioritize_compute_over_async_start =
+      config.parallel_collective_overlap_limit &&
```
golechwierowicz: Can we keep explicit `> 1`?

xla/debug_options_flags.cc
```diff
@@ -105,7 +105,8 @@ absl::StatusOr<std::vector<RepeatedFlagModifier>> ParseRepeatedEnumModifiers(
 namespace {
 
 template <typename T>
-static auto FindRepeatedFieldValue(google::protobuf::RepeatedField<int>* list, T value) {
+static auto FindRepeatedFieldValue(google::protobuf::RepeatedField<int>* list,
```
golechwierowicz: This change seems unrelated

xla/xla.proto
golechwierowicz: Can we add the rationale you provided to the comment in the xla.proto file?

xla/debug_options_flags.cc
Tixxx: for some reason, clang format kept changing this part. I manually reverted this.

xla/service/gpu/gpu_hlo_schedule.cc
Tixxx: updated

PatriosTheGreat (approved):

golechwierowicz (approved):

golechwierowicz (approved):
It seems like there are some clang-tidy checks which fail.

Tixxx:
> It seems like there are some clang-tidy checks which fail.

Yea i think the errors were from parts of the files that i didnt touch. looks like there was a fix to address it, i just rebased.

seantalts (approved):

seantalts:
would you mind rebasing? ty!

Tixxx:
> would you mind rebasing? ty!

Done.

seantalts (approved):

akuegel:
@Tixxx Can you please rebase again?

Tixxx:
> @Tixxx Can you please rebase again?

done

akuegel (approved):

PR #41832, opened 2026-04-30
add_conv_window_restore_and_amax_support

Cjkkkk:
📝 Summary of Changes
1. the window of dgrad needs to be restored to what fprop is so cudnn can understand it
2. conv assignment should skip any conv that is already assigned with a kind
3. support conv fusion with additional reduce
4. some activation support like Elu


derdrdirk (approved):
LGTM

derdrdirk (approved):
LGTM

PR #41887, opened 2026-05-01
[DOCS] Document symbolic expressions and maps

pavithraes:
**📝 Summary of Changes**

* Add new docs for `SymbolicExpr` and `SymbolicMap`
* Update mentions of `AffineMap` to `SymbolicMap`

**🚀 Kind of Contribution**

📚 Documentation


docs/indexing.md
```diff
@@ -978,16 +982,12 @@ where `s0` refers to the innermost dimension of the input.
 
 For more examples see [indexing_analysis_test.cc](https://github.com/openxla/xla/blob/main/xla/hlo/analysis/indexing_analysis_test.cc).
 
-## Indexing Map Simplifier
-
-The default simplifier for `mlir::AffineMap` upstream cannot make any
-assumptions about the ranges of dimensions/symbols. Therefore, it cannot
-simplify expressions with `mod` and `div`efficiently.
+## Indexing Map simplification
```
pavithraes, line 985: Generalized the heading because [`indexing_map_simplifier` was moved into `indexing_map`](https://github.com/openxla/xla/commit/55c1f87da747afb9ec8be12d6b5dc4d82dda25b2#diff-673323fb64b2a5eb334fd4fe4692f117788ffd454d5a0f29abeb928b449a582e).

`SymbolicExprSimplifier` is also handled in the same file, hence removed the introductory sentences:

>The default simplifier for `mlir::AffineMap` upstream cannot make any
assumptions about the ranges of dimensions/symbols. Therefore, it cannot
simplify expressions with `mod` and `div`efficiently.

Please let me know if you would like to re-phrase and keep this information.

docs/indexing.md
```diff
@@ -1000,27 +1000,27 @@ The simplifier can rewrite the following expressions.
...
+After the composition of indexing maps and their simplification we will get the following map:
 
 `(d0, d1, d2) -> (d0, d1, d2)`.
 
 Indexing map simplification also simplifies the constraints.
 
 1. Constraints of type
-`lower_bound <= affine_expr (floordiv, +, -, *) constant <= upper_bound` are
-rewritten as `updated_lower_bound <= affine_expr <= updated_upped_bound`.
+`lower_bound <= symbolic_expr (floordiv, +, -, *) constant <= upper_bound` are
```
pavithraes: Didn't include the new operations (min, max) here because I didn't see any mentions in the [simplification tests](https://github.com/openxla/xla/blob/main/xla/hlo/analysis/indexing_map_test.cc), please let me know if it should be included.

docs/indexing.md
```diff
@@ -193,7 +197,7 @@ have feasible values between `0` and `tensor_size - slice_size - 1`.
 `<expression> in <range>`, e.g. `d0 + s0 in [0, 20]`. Together with
 `Variable.bounds` they define the "domain" of indexing function.
 
-Let's study-by-example to understand what's all of the above actually means.
+Let's study-by-example to understand what all of the above actually means.
```
pavithraes, line 201: FWICT, the mapping for the Ops in this file does not seem to have changed after migration to SymbolicMap (i.e., none of the new operations like min/max were used). Please let me know if any updates are required.

docs/symbolic_expressions.md
pavithraes, line 1: Currently, this is a high-level explanation. Would you like to expand it to inlcude any implementation specifics, esp. for [SymbolicMap](https://github.com/openxla/xla/blob/18261672d5c2704a6e1017a0b66bd70fabb7a445/xla/hlo/analysis/symbolic_map.h#L69)?

docs/symbolic_expressions.md
```diff
@@ -0,0 +1,30 @@
...
+
+`SymbolicExpr` represents mathematical expressions as an abstract syntax tree, with *dimensions* and *symbols* that are not resolved until later in the compilation pipeline.
+
+Example of a `SymbolicExpr`: `d0 + s0 * 8`
+
+It allows XLA to perform symbolic algebra on tensor shapes, and enables calculating how [tiled data](./tiled_layout.md) changes after specific operations.
+
+Supported operations: `add`, `multiply`, `mod`, `floordiv`, `ceildiv`, `min`, `max`.
+
+*Note: `ceildiv`, `min`, and `max` do not have `AffineExpr` counterparts.*
```
pifon2a: I am pretty sure `ceilDiv` exists in the affineexpr. 

```
enum class AffineExprKind {
  Add,
  /// RHS of mul is always a constant or a symbolic expression.
  Mul,
  /// RHS of mod is always a constant or a symbolic expression with a positive
  /// value.
  Mod,
  /// RHS of floordiv is always a constant or a symbolic expression.
  FloorDiv,
  /// RHS of ceildiv is always a constant or a symbolic expression.
  CeilDiv,

  /// This is a marker for the last affine binary op. The range of binary
  /// op's is expected to be this element and earlier.
  LAST_AFFINE_BINARY_OP = CeilDiv,

  /// Constant integer.
  Constant,
  /// Dimensional identifier.
  DimId,
  /// Symbolic identifier.
  SymbolId,
};
```

but `min` and `max` are new.

pifon2a (changes requested):

docs/symbolic_expressions.md
pavithraes: Thanks for catching, updated.

docs/_toc.yaml
```diff
@@ -37,6 +37,8 @@ toc:
     path: /xla/hermetic_cuda
   - title: HLO Passes
     path: /xla/hlo_passes
+  - title: Symbolic expressions and maps
```
karupayun: This seems to be mostly ordered alphabetically, so it should go lower in the file

karupayun: And I would prefer using capital format (Symbolic Expressions and Maps)

docs/_toc.yaml
```diff
@@ -37,6 +37,8 @@ toc:
...
+    path: /xla/symbolic_expressions
```
karupayun: Nit: symbolic_expression

docs/indexing.md
```diff
@@ -127,53 +127,57 @@ approximately.
...
 ```c
-struct Interval {
- int64_t lower;
- int64_t upper;
-};
-
 class IndexingMap {
    // Variable represents dimension, range or runtime variable.
   struct Variable {
+    // struct Interval represents a closed interval [lower_bound, upper_bound].
```
karupayun, line 149: This is nice, but could you also add it in the original code, since you are updating it?

docs/indexing.md
karupayun: Exactly, for now it's almost a transparent migration. It will change soon, but are you right

docs/indexing.md
karupayun: It's part of the new canonicalization of SymbolicExpr -> https://github.com/openxla/xla/blob/main/xla/hlo/analysis/symbolic_expr_test.cc#L360

The simplification calls the internal canonicalization first, so yes, please include them

docs/symbolic_expressions.md
```diff
@@ -0,0 +1,30 @@
+# Symbolic expressions and maps
```
karupayun: Again, capitalized (*E*xpressions, *M*aps)

docs/symbolic_expressions.md
```diff
@@ -0,0 +1,30 @@
...
+
+Symbolic expressions (`SymbolicExpr`) and maps (`SymbolicMap`) are mathematical abstraction systems that enable symbolic tensor computations and transformations in the compilation pipeline.
```
karupayun: Actually, I would use singular. A symbolic expression is a mathematical ...

docs/symbolic_expressions.md
karupayun: This is great! I will leave some comments but if you prefer we could iterate in a different CL.

docs/symbolic_expressions.md
```diff
@@ -0,0 +1,30 @@
...
+They act as the "mathematical bridge" between a high-level HLO operation and the actual memory addresses accessed by the GPU/CPU.
+
+`SymbolicExpr` and `SymbolicMap` XLA's custom implementations that supersede the legacy `mlir::AffineExpr` and `mlir::AffineMap`.
+
+## `SymbolicExpr`
```
karupayun: What about rewritting it to something like this:
SymbolicExpr
A SymbolicExpr represents a node in an abstract syntax tree (AST). Unlike standard MLIR Affine Expressions, these support a wider range of operations necessary for modern GPU tiling. 

Example of a `SymbolicExpr`: `d0 + s0 * 8`

Supported Operations: add, mul, mod, FloorDiv, CeilDiv, min, and max.

Immutability: Expressions are pointer-like handles to internal storage managed by mlir::MLIRContext. They are automatically deduplicated (uniquified).

Variables: The system treats all variables (dimensions and symbols) as VariableID. The interpretation of a variable depends on its context within a SymbolicMap.

Canonicalization: SymbolicExpr provides a method that applies algebraic simplifications (such as constant folding and identity elimination) to ensure expressions are represented in a standard, minimal form. For example, (d0 + 1) - 1 will be simplified to d0. This is vital for predictable equality checks (operator==) in expressions that are mathematically equivalent but structurally different.

docs/symbolic_expressions.md
```diff
@@ -0,0 +1,30 @@
...
+
+## `SymbolicMap`
+
+`SymbolicMap` is a collection and combination of `SymbolicExprs`. It's a mathematical mapping of transformation between coordinate systems, typically between input and output tensors.
+
+Example of a `SymbolicMap`: `(d0, d1)[s0, s1] -> ((d0 + s0), (d1 * s1))`
+
+`SymbolicMap` forms the mathematical basis for `IndexingMap`, that describes how tensor elements map to each other in [HLO semantics](./operation_semantics.md).
+
+`IndexingMap` consists on symbolic maps with domain-specific constraints. It enables shape and tiling analysis, and collapsing operation chains (like consecutive reshaping, transpose, broadcast, etc.) into optimized indexing calculation.
```
karupayun: Could you add some documentation about key operations?

Something like:

Key Operations:
Composition (Compose): Enables chaining transformations (e.g., $f \circ g$). This is the core mechanism for fusing multiple HLO operations into a single indexing calculation.

Variable Substitution (ReplaceDimsAndSymbols): Allows re-mapping coordinate spaces by substituting dimensions or symbols with other expressions.

Optimization (CompressDims / CompressSymbols): Simplifies maps by identifying and removing unused variables, keeping the generated indexing logic as lean as possible.

karupayun (commented):
Thanks for this PR pavithra, it looks very nice. I left a bunch of comments, feel free to disagree with anything :)

docs/_toc.yaml
pavithraes: Re: Capitalization, the [Google developer documentation style guide](https://developers.google.com/style/highlights) prefers sentence case titles. Although it hasn't been strictly followed throughout the documentation set, I think we can adhere to it for this one and new docs, and consider a separate style-guide based formatting update for all docs to maintain consistecy. Please let me know if you feel strongly otherwise.

Re: ordering in ToC, I did this because we have a proposal in progress to group common topics together. I do agree that we can follow the status-quo in this PR and handle the ToC ordering separately. I'll make the update!

docs/symbolic_expressions.md
pavithraes: Thanks for the detailed notes, updated!

docs/symbolic_expressions.md
pavithraes: Thanks for the notes, updated with some basic examples from `symbolic_map_test.cc`(https://github.com/openxla/xla/blob/main/xla/hlo/analysis/symbolic_map_test.cc).

docs/symbolic_expressions.md
pavithraes: Updated.

docs/symbolic_expressions.md
pavithraes: Haven't made this update yet ([per the style-guide note above](https://github.com/openxla/xla/pull/41887#discussion_r3187468109)). Please let me know if you feel strongly though!

docs/indexing.md
pavithraes: The comment on line 151 was also not in the codebase (`// Name of the variable is used for nicer printing.`)

I've added both of these lines to `indexing_map.h`.

(Just noting that this line 149 comment is already mentioned at the [Interval stuct definition here](https://github.com/openxla/xla/blob/b3b6195466f1afdcfa8663982addb7d373c96815/xla/hlo/analysis/interval.h#L31))

docs/symbolic_expressions.md
pavithraes: Thanks for the comments, we can incorporate them in this PR. :)

pavithraes:
@karupayun Thanks for the detailed review! I've addressed and/or replied to your comments. Please take a look when you get a chance. :)

karupayun (approved):
LGTM! Thanks for addressing the changes

docs/symbolic_expression.md
```diff
@@ -0,0 +1,102 @@
...
+
+*Supported operators: `+`, `-`, `*`, `/` (`floorDiv`), `%` (`mod`)*
+
+### Example usage
+
+```sh
+v0 = CreateSymbolicVariable(0, context); // 0 is the var_id
+v1 = CreateSymbolicVariable(1, context);
+
+SymbolicExpr expr = (((v0 + 42) * v1.min(2).max(0)) / 2).ceilDiv(2);
```
karupayun, line 31: To be honest, this could be clearer (we already support things like (max(min(v1, 2), 0) / 2), but since I haven't updated the tests yet, I guess this is fine for now

PR #38198, opened 2026-02-24
[oneDNN] Upgrade to oneDNN v3.11

othakkar:
This PR updates the oneDNN dependency from v3.7.3 to v3.11. The new version unifies support for both synchronous and asynchronous execution, which is required for compatibility with the thunk runtime.

As a result of this unification, the build flag for enabling asynchronous execution is no longer necessary and has been removed.

PR #41989, opened 2026-05-04
[XLA:CPU][oneDNN] Refactor TSL oneDNN threadpool for async oneDNN support

othakkar:
Part 1 of 3 to replace #38198.

This PR refactors the TSL `onednn_threadpool` integration in XLA:CPU to support asynchronous oneDNN execution (to be enabled later).


xla/tsl/util/onednn_threadpool.h
```diff
@@ -140,6 +143,33 @@ class OneDnnThreadPool : public threadpool_iface {
...
+      }
+      run_jobs(balance, njobs_to_schedule, n, njobs, fn);
+      counter.DecrementCount();
+      counter.Wait();
+    } else {
+      absl::BlockingCounter counter(njobs);
+      std::function<void(int, int)> handle_range = [=, &handle_range, &counter](
+                                                       int first, int last) {
+        while (last - first > 1) {
+          const auto mid = first + (last - first) / 2;
```
penpornk, line 163: This can just be
```suggestion
          const auto mid = (first + last) / 2;
```

xla/tsl/util/onednn_threadpool.h
```diff
@@ -88,14 +88,16 @@ class OneDnnThreadPool : public threadpool_iface {
   bool get_in_parallel() const override {
     return (eigen_interface_->CurrentThreadId() != -1) ? true : false;
   }
+#ifndef ENABLE_ONEDNN_ASYNC
   uint64_t get_flags() const override { return ASYNCHRONOUS; }
-#ifdef ENABLE_ONEDNN_ASYNC
+#else   // ENABLE_ONEDNN_ASYNC
+  uint64_t get_flags() const override { return 0; }
```
akuegel, line 99: I am confused by this change. First, why do we return ASYNCHRONOUS only if ENABLE_ONEDNN_ASYNC is not defined, but 0 if it is defined? When I read the comments in threadpool_iface, it looks like we should return ASYNCHRONOUS if we want asynchronous behavior.

Second, what are you trying to achieve here? To me it looks like the current code should already be fine?
Maybe you wanted to change it to return ASYNCHRONOUS only if ENABLE_ONEDNN_ASYNC is defined, and 0 otherwise? In that case, you likely need something like this:

```
#ifdef ENABLE_ONEDNN_ASYNC
  uint64_t get_flags() const override { return ASYNCHRONOUS; }
  // wait() method for synchronous execution is basically a no-op.
  // But we need to implement it to satisfy the interface.
  // This is the requirement of the new experimental async runtime support
  // in oneDNN.
  virtual void wait() override {}
#else
  uint64_t get_flags() const override { return 0; }
#endif
```

penpornk: I agree with @akuegel here. In addition to the `ASYNCHRONOUS` flag value, the `wait()` function only needs to be defined when oneDNN async is enabled.

xla/tsl/util/onednn_threadpool.h
```diff
@@ -140,6 +143,33 @@ class OneDnnThreadPool : public threadpool_iface {
         }
         counter.DecrementCount();
         run_jobs(balance, first, n, njobs, fn);
+#else   // ENABLE_ONEDNN_ASYNC
```
penpornk: For readability, let's not have an `#ifdef` that breaks `if`-`else` clauses. The `#ifdef` can just be in `if (use_caller_thread) { .... }` since the `else { }` case is still the same.

penpornk (changes requested):

xla/tsl/util/onednn_threadpool.h
othakkar: @akuegel @penpornk This is a synchronous threadpool interface, used only by TensorFlow (XLA has its own under [xla/backends/cpu/runtime/onednn/onednn_threadpool.h](https://github.com/openxla/xla/blob/main/xla/backends/cpu/runtime/onednn/onednn_threadpool.h)).

You're right that when `ENABLE_ONEDNN_ASYNC` is true, we'd normally need to return `ASYNCHRONOUS`, but we return `0` instead, because this threadpool is synchronous. Returning `0` signals to oneDNN that the caller handles synchronization.

When `ENABLE_ONEDNN_ASYNC` is not defined (the default case right now), we return `ASYNCHRONOUS` to preserve the existing behavior — the older oneDNN version (still used by ARM) treated ASYNCHRONOUS as a blocking wait internally, and I didn't want to change that codepath.

Let me know if this helps.

xla/tsl/util/onednn_threadpool.h
othakkar: @penpornk There is a slight difference in the `else { }` case, which is easy to miss. When `ENABLE_ONEDNN_ASYNC` is not defined, the counter is decremented before calling `run_jobs`:
```
counter.DecrementCount();
run_jobs(balance, first, n, njobs, fn);
```
This ordering doesn't seem correct, but it was introduced in one of the ARM PRs, and it currently exists. I didn't want to change it in this patch, since this threadpool code is shared with ARM code.

When `ENABLE_ONEDNN_ASYNC` is defined, oneDNN supports truly async execution, so decrementing before scheduling could cause segmentation faults. The counter must be decremented after the job completes.
```
run_jobs(balance, first, n, njobs, fn);
counter.DecrementCount();
```

Let me know if you'd prefer I add `ifndefs` only around the affected lines instead.

xla/tsl/util/onednn_threadpool.h
akuegel: If it is a behavior change with the new oneDNN version, doesn't it mean we need to land it at the same time as the version upgrade? Otherwise it looks to me that we would change (and potentially break) the case where ENABLE_ONEDNN_ASYNC is defined, as now we return 0 where before we returned ASYNCHRONOUS? Even if it is not defined by default, there could be users that define it?

othakkar: When `ENABLE_ONEDNN_ASYNC` is defined with this PR (and with `main`), it will pull in the [`dev-v3.7-thunk-preview`](https://github.com/openxla/xla/blob/b3b6195466f1afdcfa8663982addb7d373c96815/third_party/mkl_dnn/workspace.bzl#L20) branch from oneDNN, which has complete functional support for the synchronous threadpool that TF uses (with some expected performance drop resolved in the `v3.11.3` version). So, the users won't run into functional issues.

That said, we don't anticipate any other users using this build flag except for us.

This is only temporary until #41991 is merged, after which `v3.11.3` will be pulled in by default in XLA/TF/JAX. We split the original PR (#38198) into 3 smaller, more manageable parts for comprehensive testing.

othakkar:
@penpornk @akuegel thanks for your review comments. I've addressed them. Please let me know if this resolves your concerns.

xla/tsl/util/onednn_threadpool.h
penpornk: > This is a synchronous threadpool interface, used only by TensorFlow

Oh, right. I didn't notice the path and thought this one was for XLA.

> You're right that when ENABLE_ONEDNN_ASYNC is true, we'd normally need to return ASYNCHRONOUS, but we return 0 instead, because this threadpool is synchronous. Returning 0 signals to oneDNN that the caller handles synchronization.

Please document this in the code.

> When ENABLE_ONEDNN_ASYNC is not defined (the default case right now), we return ASYNCHRONOUS to preserve the existing behavior — the older oneDNN version (still used by ARM) treated ASYNCHRONOUS as a blocking wait internally, and I didn't want to change that codepath.

But then setting it to 0 shouldn't change anything for ARM, right? It would be a blocking wait either way. Why do we need `ASYNCHRONOUS` here?

penpornk (changes requested):

othakkar:
> Please document this in the code.

Done.

> But then setting it to 0 shouldn't change anything for ARM, right? It would be a blocking wait either way. Why do we need `ASYNCHRONOUS` here?

When `ENABLE_ONEDNN_ASYNC` is not defined (which will remain the default for ARM after #41991 is merged), `get_flags` will still return `ASYNCHRONOUS`, so there will be no changes for ARM (until the oneDNN version used by ARM is upgraded to v3.11+).
That said, setting `get_flags` to 0 for ARM shouldn't change anything functionally because we add a [blocking wait for synchronization in the `if` block](https://github.com/Intel-tensorflow/xla/blob/0bcec7d0598a92132ad8f9aa86e6754c3d58533e/xla/tsl/util/onednn_threadpool.h#L156-#L158) and [fix the blocking wait in the `else` block](https://github.com/Intel-tensorflow/xla/blob/0bcec7d0598a92132ad8f9aa86e6754c3d58533e/xla/tsl/util/onednn_threadpool.h#L170-#L171).

othakkar:
@penpornk, checking to see if I can help clarify any questions/concerns on this PR.

xla/tsl/util/onednn_threadpool.h
penpornk: This has not been addressed.

xla/tsl/util/onednn_threadpool.h
penpornk: I think the order should be fixed. Let me make a PR swapping the existing order and making `get_flags` return 0. We'll ask Arm to help test this. Then this PR can be simplified.

penpornk (changes requested):

xla/tsl/util/onednn_threadpool.h
penpornk: Done. 
TF PR: https://github.com/tensorflow/tensorflow/pull/118877 (Asked Arm for feedback here.)
XLA PR: https://github.com/openxla/xla/pull/42852

penpornk (changes requested):
PR https://github.com/openxla/xla/pull/42852 is in. Please update your PR.

othakkar:
@penpornk I rebased the PR with the latest `main`. Sorry for the delay!

othakkar:
@penpornk, could you please help re-review the PR?

xla/tsl/util/onednn_threadpool.h
```diff
@@ -94,13 +94,18 @@ class OneDnnThreadPool : public threadpool_iface {
...
+  uint64_t get_flags() const override { return 0; }
   // wait() method for synchronous execution is basically a no-op.
   // But we need to implement it to satisfy the interface.
   // This is the requirement of the new experimental async runtime support
   // in oneDNN.
   virtual void wait() override {}
+#else
+  // Return ASYNCHRONOUS to work with current onednn version.
+  // TODO(intel-tf): remove this ifdefing block after making the async oneDNN default
+  uint64_t get_flags() const override { return ASYNCHRONOUS; }
```
penpornk, line 108: I thought we agreed that this is safe to set to 0 (regardless of `ENABLE_ONEDNN_ASYNC` = removing #ifdef) since this thread pool is handling the wait.

xla/tsl/util/onednn_threadpool.h
```diff
@@ -94,13 +94,18 @@ class OneDnnThreadPool : public threadpool_iface {
...
+  // Return 0 for synchronous execution; caller handles synchronization.
+  uint64_t get_flags() const override { return 0; }
   // wait() method for synchronous execution is basically a no-op.
   // But we need to implement it to satisfy the interface.
   // This is the requirement of the new experimental async runtime support
   // in oneDNN.
   virtual void wait() override {}
+#else
+  // Return ASYNCHRONOUS to work with current onednn version.
+  // TODO(intel-tf): remove this ifdefing block after making the async oneDNN default
```
penpornk: Could you please mention v3.11 instead? Thank you!

xla/tsl/util/onednn_threadpool.h
penpornk: Resolved offline.

penpornk (changes requested):

xla/tsl/util/onednn_threadpool.h
othakkar: Done!

penpornk (approved):
Thank you again and sorry for the delay!

PR #41997, opened 2026-05-04
[ROCm] Stop rocprofiler context in RocmTracer::Disable() to prevent cross-session event leak

magaonka-amd:
📝 Summary of Changes
- `RocmTracer::Enable()` calls `rocprofiler_start_context(context_)` but `RocmTracer::Disable()` never called the matching rocprofiler_stop_context(). 

- The only `stop_context` call lived in `toolFinalize()`, which runs once at process exit. Between sessions the context stayed active, accumulating HIP/kernel events that then drained into the next session's collector.

🎯 Justification
without the fix there is a cross session event leak in XLA profiler for ROCm

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
- `DisableStopsRocprofilerContext` uses the SDK's own rocprofiler_context_is_active() API to assert the context is inactive after Disable() returns.
- `DisableIsolatesNextSession` runs a no-profiler window of HIP work between two profile sessions; the second session does zero HIP activity between its Enable() and Disable(), so its collector must receive zero events. Without the fix the leaked-window events drain into that collector.

### DIFF
--- a/xla/backends/profiler/gpu/BUILD
+++ b/xla/backends/profiler/gpu/BUILD
@@ -549,9 +549,11 @@ xla_cc_test(
         "//xla/tsl/lib/core:status_test_util",
         "@com_google_absl//absl/log",
         "@com_google_absl//absl/strings:string_view",
+        "@com_google_absl//absl/time",
         "@com_google_googletest//:gtest_main",
         "@local_config_rocm//rocm:hip",  # buildcleaner: keep
         "@local_config_rocm//rocm:rocm_headers",
+        "@local_config_rocm//rocm:rocprofiler_sdk",  # buildcleaner: keep
         "@tsl//tsl/profiler/protobuf:xplane_proto_cc",
     ],
 )
--- a/xla/backends/profiler/gpu/rocm_tracer_test.cc
+++ b/xla/backends/profiler/gpu/rocm_tracer_test.cc
@@ -24,7 +24,11 @@ limitations under the License.
 #include <gtest/gtest.h>
 #include "absl/log/log.h"
 #include "absl/strings/string_view.h"
+#include "absl/time/clock.h"
+#include "absl/time/time.h"
 #include "rocm/include/hip/hip_runtime.h"
+#include "rocm/include/rocprofiler-sdk/context.h"
+#include "rocm/include/rocprofiler-sdk/fwd.h"
 #include "xla/backends/profiler/gpu/rocm_collector.h"
 #include "xla/backends/profiler/gpu/rocm_tracer_utils.h"
 #include "xla/tsl/lib/core/status_test_util.h"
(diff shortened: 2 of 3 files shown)


nurmukhametov (commented):
It looks good to me overall, but I would have avoided the function-local #define/#undef HIP_ASSERT_OK.

nurmukhametov (approved):

dimitar-asenov (approved):

xla/backends/profiler/gpu/rocm_tracer_test.cc
```diff
@@ -249,6 +253,89 @@ TEST(RocmTracerTest, CapturesHipEvents) {
...
+
+  active = -1;
+  ASSERT_EQ(rocprofiler_context_is_active(tracer.context_, &active),
+            ROCPROFILER_STATUS_SUCCESS);
+  EXPECT_EQ(active, 0)
+      << "Disable() should call rocprofiler_stop_context(context_)";
+}
+
+TEST(RocmTracerTest, DisableIsolatesNextSession) {
+#define HIP_ASSERT_OK(expr) ASSERT_EQ((expr), hipSuccess) << #expr " failed"
```
dimitar-asenov: Please do not introduce a new macro. You may want to make a new function like, `CheckSuccess(expr) ->status` and call that via `RETURN_IF_ERROR`. Or just make the checks below slightly more verbose without defining any new functions.

dimitar-asenov (changes requested):

xla/backends/profiler/gpu/rocm_tracer_test.cc
magaonka-amd: Thanks for the feedback , fixed it with just regular ASSERT_EQ. removed the macro.

xla/backends/profiler/gpu/rocm_tracer_test.cc
```diff
@@ -24,7 +24,11 @@ limitations under the License.
 #include <gtest/gtest.h>
 #include "absl/log/log.h"
 #include "absl/strings/string_view.h"
+#include "absl/time/clock.h"
```
akuegel, line 27: You need a corresponding dependency for these two includes.

xla/backends/profiler/gpu/rocm_tracer_test.cc
```diff
@@ -24,7 +24,11 @@ limitations under the License.
...
+#include "absl/time/time.h"
 #include "rocm/include/hip/hip_runtime.h"
+#include "rocm/include/rocprofiler-sdk/context.h"
```
akuegel, line 30: Same here.

xla/backends/profiler/gpu/rocm_tracer_test.cc
magaonka-amd: added absl time to build

xla/backends/profiler/gpu/rocm_tracer_test.cc
magaonka-amd: added rocprofsdk to build

akuegel (approved):

xla/backends/profiler/gpu/BUILD
```diff
@@ -550,9 +549,11 @@ xla_cc_test(
         "//xla/tsl/lib/core:status_test_util",
         "@com_google_absl//absl/log",
         "@com_google_absl//absl/strings:string_view",
+        "@com_google_absl//absl/time",
         "@com_google_googletest//:gtest_main",
         "@local_config_rocm//rocm:hip",  # buildcleaner: keep
         "@local_config_rocm//rocm:rocm_headers",
+        "@local_config_rocm//rocm:rocprofiler_sdk",
```
akuegel: Can you please add a comment # buildcleaner: keep
to this dependency? It seems our buildcleaner tool considers this dependency unused internally (but I think it is used in OSS).

magaonka-amd: done added !

dimitar-asenov (approved):

PR #42006, opened 2026-05-04
Re-land [PR #36441] Support all-reduce hoisting with scalar multiplication pattern

sfvaroglu:
📝 Summary of Changes
Re-lands the reverted [PR #36441](https://github.com/openxla/xla/pull/36441) with a fix and unit tests.

🎯 Justification
For the ZeRO-1 gradient-accumulation pattern `AR(g) * AR(s) + b` inside a while body, the gradient `AR(g)` is hoisted out of the loop, while the scalar `AR(s)` is now kept inside the loop body so the per-iteration scaling factor remains the all-reduced value.

🚀 Kind of Contribution

The HLOs in compiler/xla/tools/benchmarks/hlo/ do not have this pattern.

🧪 Unit Tests:
Added to xla/service/while_loop_all_reduce_code_motion_test.cc

🧪 Execution Tests:
N/A

### DIFF
--- a/xla/service/while_loop_all_reduce_code_motion.cc
+++ b/xla/service/while_loop_all_reduce_code_motion.cc
@@ -779,6 +779,9 @@ MovableAllReduceContext IsAllReduceMovable(
           break;
         }
         case HloOpcode::kMultiply: {
+          // Hoisting reduce-scatter through a multiply is unsupported: it
+          // would require rewriting the other operand to the post-scatter
+          // shape, which this pass does not do.
           if (is_reduce_scatter) {
             is_all_reduce_movable = false;
             break;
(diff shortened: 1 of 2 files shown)


xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -778,6 +778,47 @@ MovableAllReduceContext IsAllReduceMovable(
...
+
+          if (current_is_scalar && other_is_all_reduce) {
+            VLOG(4) << "Scalar all-reduce " << all_reduce->name()
+                    << " is used as a factor of a multiply with another "
+                       "all-reduce ("
+                    << unwrapped_other->name()
+                    << "); marking it unmovable so it stays in the loop body "
+                       "and preserves the math of the hoisted all-reduce.";
+            is_all_reduce_movable = false;
+          } else if (other_is_all_reduce && other_is_broadcast) {
```
akuegel: Reading this line is confusing, as it seems to imply that the other operand is both a all-reduce and a broadcast. Maybe rename `other_is_all_reduce` to `unwrapped_other_is_all_reduce`?

Also, I would like a comment to explain why we are treating this as a case where we think the all_reduce is potentially movable? Here some questions I would like to have answered by a comment:
I don't understand why it makes a difference that the other operand is a broadcasted all-reduce. Is there some implicit assumption here, e.g. that the all-reduce or reduce-scatter operand is scalar? Why do we say in line 783 that if `is_reduce_scatter` is true, it is not movable, but here in line 811 if unwrapped_other is a reduce scatter, it is potentially movable?

xla/service/while_loop_all_reduce_code_motion.cc
```diff
@@ -778,6 +778,66 @@ MovableAllReduceContext IsAllReduceMovable(
...
+            // accumulation off by a factor of num_replicas. Keep it in
+            // the body.
+            VLOG(4) << "Scalar all-reduce " << all_reduce->name()
+                    << " is used as a factor of a multiply with another "
+                       "all-reduce ("
+                    << unwrapped_other->name()
+                    << "); marking it unmovable so it stays in the loop body "
+                       "and preserves the math of the hoisted all-reduce.";
+            is_all_reduce_movable = false;
+          } else if (unwrapped_other_is_all_reduce && other_is_broadcast) {
```
akuegel: Thanks for the comments. It seems we should check here that unwrapped_other has a scalar shape? `const bool unwrapped_other_is_scalar = ShapeUtil::IsScalar(unwrapped_other->shape());`

It turns out adding this additional check makes the test that caused the rollback pass. It seems the failure case is something like this:
`multiply(broadcast1(AR_non_scalar_1), broadcast2(AR_non_scalar_2))`

Can you please also add a test with that pattern?

akuegel (approved):

PR #42028, opened 2026-05-05
[xla:se] Make CudaVmmAllocator compatible with NCCL

ezhulenev:
Second attempt to land: https://github.com/openxla/xla/pull/41652. In addition to previous version add fallbacks + `LOG(WARNING)` when some of the VMM capabilities are not accessible (i.e. fabric handles, POSIX FD handle, etc.).

#### Original PR:

Instead of always linking NCCL make our own NCCL-compatible allocator:

Rewrite `CudaVmmAllocator` for NCCL-compatible VMM allocation. Replace the simple bool-based constructor with an `Options` struct controlling alignment, peer access, FABRIC handle support, and RDMA
  capability. Allocation now uses `POSIX_FILE_DESCRIPTOR` handle type as baseline (matching NCCL's `ncclMemAlloc`), with optional `FABRIC` handle that gracefully falls back if not supported. Deallocation now synchronizes via `cuCtxSynchronize()` before unmapping since `cuMemUnmap` immediately invalidates
  virtual address mappings unlike `cuMemFree`.

Add `is_fabric_supported` and `vmm_granularity_` to `CudaExecutor`. Replace `is_vmm_supported_` bool with cached `vmm_granularity_` (zero means unsupported). VMM support is now mandatory — `Init()` returns an error if the device doesn't support it. Query fabric handle support at init time via `CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED`. Detect peer access across all devices and pass it to VMM allocator options.

### DIFF
--- a/xla/stream_executor/cuda/BUILD
+++ b/xla/stream_executor/cuda/BUILD
@@ -1319,9 +1319,7 @@ cc_library(
         "//xla/stream_executor:stream_executor_h",
         "//xla/tsl/platform:status_macros",
         "@com_google_absl//absl/base",
-        "@com_google_absl//absl/base:core_headers",
         "@com_google_absl//absl/cleanup",
-        "@com_google_absl//absl/log",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings:str_format",
@@ -1631,6 +1629,7 @@ cc_library(
         "@com_google_absl//absl/synchronization",
         "@com_google_absl//absl/time",
         "@com_google_absl//absl/types:span",
+        "@local_config_cuda//cuda:cuda_headers",
     ],
 )
 
--- a/xla/stream_executor/cuda/BUILD
+++ b/xla/stream_executor/cuda/BUILD
@@ -1318,11 +1318,14 @@ cc_library(
         "//xla/stream_executor:memory_allocator",
         "//xla/stream_executor:stream_executor_h",
         "//xla/tsl/platform:status_macros",
+        "@com_google_absl//absl/base",
+        "@com_google_absl//absl/base:core_headers",
+        "@com_google_absl//absl/cleanup",
+        "@com_google_absl//absl/log",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings:str_format",
         "@local_config_cuda//cuda:cuda_headers",
-        "@tsl//tsl/platform:numbers",
     ],
 )
 
--- a/xla/stream_executor/cuda/cuda_executor.cc
+++ b/xla/stream_executor/cuda/cuda_executor.cc
@@ -62,6 +62,7 @@ limitations under the License.
 #include "xla/stream_executor/cuda/cuda_compute_capability.h"
 #include "xla/stream_executor/cuda/cuda_context.h"
 #include "xla/stream_executor/cuda/cuda_core_info_table.h"
+#include "xla/stream_executor/cuda/cuda_device_allocator.h"
 #include "xla/stream_executor/cuda/cuda_event.h"
 #include "xla/stream_executor/cuda/cuda_host_allocator.h"
 #include "xla/stream_executor/cuda/cuda_kernel.h"
@@ -111,10 +112,8 @@ limitations under the License.
 #include "xla/tsl/platform/statusor.h"
 #include "xla/tsl/platform/threadpool.h"
 #include "xla/util.h"
-#include "tsl/platform/casts.h"
 #include "tsl/platform/fingerprint.h"
 #include "tsl/platform/numa.h"
-#include "tsl/platform/numbers.h"
 
 namespace stream_executor {
 namespace gpu {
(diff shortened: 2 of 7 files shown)


akuegel (approved):
Approving to test internally what happens with the tests that caused the rollback before.

xla/stream_executor/cuda/cuda_executor.cc
```diff
@@ -532,24 +559,66 @@ absl::StatusOr<bool> IsMulticastSupported(CUdevice device) {
...
-                                               bool is_rdma_supported) {
-  CUmemAllocationProp properties = {};
-  properties.type = CU_MEM_ALLOCATION_TYPE_PINNED;
-  properties.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
-  properties.requestedHandleTypes =
-      static_cast<CUmemAllocationHandleType>(CU_MEM_HANDLE_TYPE_NONE);
-  properties.location.id = device;
-  properties.allocFlags.gpuDirectRDMACapable = is_rdma_supported ? 1 : 0;
-  return properties;
+absl::StatusOr<bool> IsFabricSupported(CUdevice device) {
```
akuegel, line 562: The error happens in this method. With the following changes to this method (suggested by KI), the test passes:

```
absl::StatusOr<bool> IsFabricSupported(CUdevice device) {
  int fabric_supported = 0;
  CUresult result = cuDeviceGetAttribute(
      &fabric_supported, CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED,
      device);

  // If the driver is too old to know about the attribute, it usually returns
  // INVALID_VALUE
  if (result == CUDA_ERROR_INVALID_VALUE) {
    VLOG(1) << "CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED not supported "
               "by driver.";
    return false;
  }

  TF_RETURN_IF_ERROR(cuda::ToStatus(result));
  return fabric_supported > 0;
}
```

xla/stream_executor/cuda/BUILD
```diff
@@ -1318,11 +1318,14 @@ cc_library(
         "//xla/stream_executor:memory_allocator",
         "//xla/stream_executor:stream_executor_h",
         "//xla/tsl/platform:status_macros",
```
akuegel, line 1320: In this build file, the `cuda_executor` target needs a dependency on `cuda_headers`

xla/stream_executor/cuda/BUILD
```diff
@@ -1318,11 +1318,14 @@ cc_library(
...
+        "@com_google_absl//absl/base",
+        "@com_google_absl//absl/base:core_headers",
```
akuegel: This is unused

xla/stream_executor/cuda/BUILD
```diff
@@ -1318,11 +1318,14 @@ cc_library(
...
+        "@com_google_absl//absl/cleanup",
+        "@com_google_absl//absl/log",
```
akuegel: This is unused.

ezhulenev: Done

xla/stream_executor/cuda/BUILD
ezhulenev: Done

xla/stream_executor/cuda/BUILD
ezhulenev: Done

xla/stream_executor/cuda/cuda_executor.cc
ezhulenev: Thanks! Added to the change

akuegel (approved):

xla/stream_executor/cuda/cuda_vmm_allocator.cc
```diff
@@ -33,110 +38,202 @@ limitations under the License.
...
+    }
+    return cuda::ToStatus(result);
+  };
+
+  bool has_fabric = properties.requestedHandleTypes & CU_MEM_HANDLE_TYPE_FABRIC;
+  bool has_posix_fd = properties.requestedHandleTypes &
+                      CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR;
+
+  if (has_fabric) {
+    if (auto r = try_create("FABRIC+POSIX_FD")) {
```
olupton, line 94: @ezhulenev this is too spammy; it prints a bunch of
```
W0521 08:57:26.631656 3843558 cuda_vmm_allocator.cc:80] [0] VMM cuMemCreate with FABRIC+POSIX_FD handle types failed: INTERNAL: CUDA error: : CUDA_ERROR_NOT_PERMITTED: operation not permitted; will retry with simpler handle types.
```
on 8xH100 systems, which is expected because fabric handles require IMEX (https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/virtual-memory-management.html#allocate-and-export) and IMEX is normally only available for MNNVL systems (https://docs.nvidia.com/multi-node-nvlink-systems/imex-guide/overview.html).

I think the try-and-fall-back logic is fine and matches what NCCL does, but it's too noisy as it stands.

PatriosTheGreat: This should be prevented by cuda_executor, but I misused there active links information which is non-zero on H100 and B200. Sent fix here: https://github.com/openxla/xla/pull/43111

With this logs should go away on non-cluster platforms.

PR #37766, opened 2026-02-12
fix: add recursion depth limit to HLO parser

thevilledev:
📝 Summary of Changes

Add recursion depth tracking to HLO parser to prevent stack overflow crashes from deeply nested input. A shared `current_recursion_depth_` counter is added to 6 recursive parse functions. When the configurable limit (default 500) is exceeded, the parser returns a graceful `"maximum recursion depth exceeded"` error instead of crashing.

🎯 Justification

The HLO parser is vulnerable to stack exhaustion from maliciously crafted or malformed deeply nested HLO text. This is reachable anywhere HLO text is accepted (tools, file loading, etc.).

Note the parser is not recommended for production use and can be considered as a debug API. This issue was disclosed & discussed with the maintainers.

Fixes OSS-Fuzz [#390461336](https://issues.oss-fuzz.com/issues/390461336) from Tensorflow.

🚀 Kind of Contribution

🐛 Bug Fix

📊 Benchmark (for Performance Improvements)

N/A -- not a performance change. The added overhead is essentially one integer increment/compare and a trivial RAII guard per recursive parse call.

🧪 Unit Tests:

Five new tests for the HLO parser:

- `DeeplyNestedOperandsExceedsRecursionLimit`: verifies `add(add(add...` nesting triggers the depth error at a reduced limit of 10.
- `DeeplyNestedShapesExceedsRecursionLimit`: verifies `(((f32[])))` shape nesting triggers the depth error.
- `DeeplyNestedTupleLiteralsExceedsRecursionLimit`: verifies `(((0)))` tuple literal nesting triggers the depth error.
- `DeeplyNestedInlineComputationsExceedsRecursionLimit`: verifies nested `call(...), to_apply={...}` inline computations trigger the depth error.
- `DeeplyNestedOperandsDoesNotStackOverflow`: regression test with 50,000 nesting levels using the default limit, confirming no stack overflow crash.
- `DeeplyNestedUntypedTupleLiteralExceedsRecursionLimit`: verifies that deeply nested untyped tuple literals in `literal=` attributes hit the recursion limit instead of overflowing the stack.

Test output:

```
$ docker exec xla bazel-bin/xla/hlo/parser/hlo_parser_test --gtest_filter='*DeeplyNested*' --gtest_print_time=1
Note: Google Test filter = *DeeplyNested*
...
HloParserTest.DeeplyNestedShapesExceedsRecursionLimit                               [       OK ] 
[...]

PR #42055, opened 2026-05-05
Add OSS-Fuzz harnesses for HLO parser and proto deserialization

ricaskew:
Add OSS-Fuzz harnesses for HLO parser and proto deserialization

This PR adds two libFuzzer harnesses to xla/fuzz/ as part of an
OSS-Fuzz integration for openxla/xla. A companion PR in google/oss-fuzz
will be opened concurrently to wire the build and register the project.

The HLO text-format parser and proto deserialization path were selected
as the primary fuzzing surfaces because they accept arbitrary external
input, have significant parsing complexity, and currently have no
OSS-Fuzz coverage.

## Harnesses

**hlo_parser_fuzz** — exercises `xla::ParseAndReturnUnverifiedModule`
against arbitrary text input bytes via `absl::string_view`. Targets
the HLO text-format parser surface.

**hlo_proto_fuzz** — exercises `xla::HloModule::CreateFromProto`
against arbitrary byte sequences. Stage 1 deserializes raw bytes into
`HloModuleProto` via `ParseFromArray`; stage 2 converts the proto into
an `HloModule`. Includes an explicit size guard against integer overflow
on the `ParseFromArray` size argument (guards `size > INT_MAX` before
the cast to `int`, implemented via `std::numeric_limits<int>::max()`
from `<limits>`).

## Build

Both harnesses build via Bazel under `//xla/fuzz/`. The BUILD file uses
`cc_binary` with `fuzz_target` and `nobuilder` tags following standard
OSS-Fuzz practice. Dependencies are minimal and correctly scoped per
target.

The OSS-Fuzz build uses XLA's hermetic LLVM18 toolchain which
dynamic-links libc++. The companion oss-fuzz PR resolves packaging by
embedding `--linkopt=-Wl,-rpath,$ORIGIN` and copying the required
libc++ shared objects into `$OUT/` at build time.

## Testing

Both binaries were smoke-tested inside the OSS-Fuzz Docker base image:
- `hlo_parser_fuzz`: 100 runs, zero crashes, 2,371 coverage PCs with
  growth into `xla::HloLexer` and related parser code paths
- `hlo_proto_fuzz`: 100 runs, zero crashes, 819 coverage PCs with
  growth into `xla::HloModuleConfig` and protobuf `TcParser` family

### DIFF
--- a/xla/fuzz/BUILD
+++ b/xla/fuzz/BUILD
@@ -1,0 +1,42 @@
+# Description:
+#   libFuzzer harnesses for openxla/xla parser and proto-deserialization
+#   surfaces. Built by the OSS-Fuzz integration under projects/xla/ in
+#   google/oss-fuzz, and runnable locally with -fsanitize=fuzzer flags
+#   passed via --copt and --linkopt.
+
+load("//xla/tsl/platform:rules_cc.bzl", "cc_binary")
+
+package(
+    default_visibility = ["//visibility:public"],
+    licenses = ["notice"],
+)
+
+cc_binary(
+    name = "hlo_parser_fuzz",
+    srcs = ["hlo_parser_fuzz.cc"],
+    tags = [
+        "fuzz_target",
+        "nobuilder",
+        "no_oss",
+    ],
+    deps = [
+        "//xla/hlo/parser:hlo_parser",
+        "//xla/service:hlo_module_config",
+        "@com_google_absl//absl/strings:string_view",
+    ],
+)
+
+cc_binary(
+    name = "hlo_proto_fuzz",
+    srcs = ["hlo_proto_fuzz.cc"],
+    tags = [
+        "fuzz_target",
+        "nobuilder",
+        "no_oss",
+    ],
+    deps = [
+        "//xla/hlo/ir:hlo",
+        "//xla/service:hlo_module_config",
+        "//xla/service:hlo_proto_cc",
+    ],
+)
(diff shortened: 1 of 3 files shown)


ricaskew:
Companion OSS-Fuzz integration PR: https://github.com/google/oss-fuzz/pull/15464

seantalts:
@GleasonK are you a good person to look at this? I am curious if we even want a fuzzer on HLO, since we mostly just care about the subset of HLO that JAX emits (AFAIK).

ricaskew:
Two thoughts on this. The text parser's fuzz value comes specifically from exploring inputs outside the JAX-emitted subset — malformed and edge-case inputs are where parser bugs tend to live, and PR #37766's recursion-depth limit is a recent example of that class. An in-repo harness also gives the team a regression signal for future grammar or recursion changes. Separately, the proto harness covers a distinct surface — `HloModule::CreateFromProto`, the binary deserialization path used by any consumer of serialized HLO modules — which is independent of the text-parser scope question entirely.

seantalts:
Isn't the reason bugs still live there because they are never exercised by real users or real code? ;)
But I will let @GleasonK triage further.

DavidKorczynski (changes requested):
From OSS-Fuzz perspective (https://github.com/google/oss-fuzz/pull/15464) we'd be happy to integrate this but am waiting for maintainers to be onboard.

akuegel:
@ricaskew Thank you for this contribution. However I asked around internally, and there was no interest in this fuzzer harnesses for HLO parser and proto deserialization. We actually do have a proto fuzzer internally already.

PR #42108, opened 2026-05-06
[ROCM][NFC] Refactor gpuBlasLt and matmul thunk interfaces to reduce duplication and simplify scale handling

pemeliya:
## Summary

- **Simplified `BlasLt::MatmulPlan` interface** by removing multiple overloaded `ExecuteOnStream` convenience wrappers that manually unpacked arguments into `MemoryArgs`. All callers now use the `MemoryArgs`-based API directly, eliminating ~100 lines of boilerplate forwarding code in `gpu_blas_lt.h`. Note that, one forwarding function is kept for compatibility with **Tensorflow**.

- **Unified regular and grouped matmul plan caching** in `CublasLtMatmulThunk`. Removed the separate `GetCachedGroupedMatmulPlan` method and merged it into `GetCachedMatmulPlan` using `std::visit` over the `GemmConfig`/`GroupedGemmConfig` variant. Also removed the separate `grouped_plan_cache_` and `GetOrCreateGroupedMatmulPlan`; both plan types now share a single plan cache.

- **Moved alpha/beta scale computation from `ExecuteOnStream` into plan creation** for both CUDA and ROCm backends. The `MatmulPlan` now stores pre-computed alpha/beta as fixed-size byte arrays (`kMaxScaleBytes`), determined at plan creation time based on operand types. Indeed, this is enough to run a long chain of **TYPED_MATMUL** macros only once during initialization. This also removes the `DoMatmul` indirection and eliminates the need to look up the `BlasLt` instance from the stream at execution time (plans now hold a `const BlasLt&` reference).

- **Split ROCm's monolithic `MatmulPlan` into `RegularMatmulPlan` and `GroupedMatmulPlan`**. The old single class used `std::optional` fields to represent both plan types, making the code harder to follow. The split gives each plan type its own clean set of members and `ExecuteOnStream`/`GetAlgorithms` implementations.

- **Changed `BlasLt::Get()` to take `StreamExecutor*` instead of `const Stream*`** and made it return `absl::StatusOr<BlasLt*>` for proper error propagation. Updated all callers (autotuners, matmul thunk) accordingly -- this also removes unnecessary stream creation in the autotuner paths that only needed the BlasLt handle.

- **Removed `GetAlgorithms`' dependency on `Stream*`**. The stream parameter was unnecessary for algorithm enumeration; plans now use their stored `BlasLt` reference to access the handle directly.

- **Made `TF_HIPBLASLT` always enabled** by hardcoding the flag to `1` and removing the configure-time `rocm_hipblaslt` toggle from `build_defs.bzl.tpl`, `rocm_configure.bzl`, and `rocm_config.h.tpl`. The `#if TF_HIPBLASLT` guard around `hip_blas_lt.h` is also removed since hipBlasLt is now always available.

- **Minor cleanups**: removed the `CublasLtMatmulThunk` copy constructor, inlined `ExecuteOnStreamInternal` into `ExecuteOnStream`, renamed `parent_` to `executor_` and `blas_lt_` to `handle_` for clarity, and accepted `GroupedGemmConfig` by const-ref.

- **GroupedMatmul**: replaced LOG(FATAL) during matmul object creation with a proper error forwarding with `StatusOr<>`

### DIFF
--- a/third_party/gpus/rocm/build_defs.bzl.tpl
+++ b/third_party/gpus/rocm/build_defs.bzl.tpl
@@ -68,13 +68,11 @@ def is_rocm_configured():
     """
     return %{rocm_is_configured}
 
-def rocm_hipblaslt():
-    return %{rocm_is_configured} and %{rocm_hipblaslt}
-
 def if_rocm_hipblaslt(x):
-    if %{rocm_is_configured} and (%{rocm_hipblaslt} == "True"):
-        return select({"//conditions:default": x})
-    return select({"//conditions:default": []})
+    """ 
+    hipBlasLt is always available: kept for compatibility with Tensorflow.
+    """
+    return select({"//conditions:default": x})
 
 def rocm_library(copts = [], deps = [], **kwargs):
     """Wrapper over cc_library which adds default ROCm options."""
--- a/third_party/gpus/rocm/rocm_config.h.tpl
+++ b/third_party/gpus/rocm/rocm_config.h.tpl
@@ -21,6 +21,7 @@ limitations under the License.
 #define TF_ROCM_VERSION %{rocm_version_number}
 #define TF_MIOPEN_VERSION %{miopen_version_number}
 #define TF_HIPRUNTIME_VERSION %{hipruntime_version_number}
-#define TF_HIPBLASLT %{hipblaslt_flag}
+// NOTE: HipBlasLt is now always available, this flag is deprecated !
+#define TF_HIPBLASLT 1
 
 #endif  // ROCM_ROCM_CONFIG_H_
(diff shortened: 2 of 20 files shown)


dimitar-asenov:
This PR makes a lot of unrelated changes. As a result it's very large with a delta of +540 -852. This makes it hard to review and keep track of what code change corresponds to what logical change.

Please break it down into smaller PRs. In general, please always strive for small, atomic PRs. This makes the review process much easier and faster.

Thanks.

pemeliya:
Ok sure, I will try split it into smaller ones. The point is, the core of my change was refactoring of gpublas_lt.h interface - which as a consequence forced changes in a number of different files

dimitar-asenov:
Thanks for splitting the PR. I will mark it as draft. If you want to reuse it for some part of the remaining code, please trim it and reopen again.

PR #42127, opened 2026-05-06
[XLA:GPU] Only fuse when both inputs to conv are converted

Cjkkkk:
📝 Summary of Changes
Only fuse when both inputs to conv are converted.


derdrdirk (approved):
LGTM

PR #42177, opened 2026-05-07
[ROCM][NFC] cleanup: pass amdgpu targets directly to crosstool, remove hipblaslt macros

pemeliya:
- TF_HIPBLASLT macro is now always enabled, but it is kept for compatibility with TF builds.
- AMDGPU target options are passed directly to the crosstool through templates, no need some fancy command line options

First PR extracted from https://github.com/openxla/xla/pull/42108

### DIFF
--- a/third_party/gpus/rocm/build_defs.bzl.tpl
+++ b/third_party/gpus/rocm/build_defs.bzl.tpl
@@ -16,7 +16,7 @@ def if_rocm(if_true, if_false = []):
 
 def rocm_default_copts():
     """Default options for all ROCm compilations."""
-    return if_rocm(["-x", "rocm"] + %{rocm_extra_copts})
+    return if_rocm(["-x", "rocm"])
 
 def rocm_copts(opts = []):
     """Gets the appropriate set of copts for (maybe) ROCm compilation.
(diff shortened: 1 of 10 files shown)


dimitar-asenov (approved):

akuegel:
@pemeliya can you please rebase?

third_party/gpus/rocm/build_defs.bzl.tpl
```diff
@@ -68,13 +68,11 @@ def is_rocm_configured():
     """
     return %{rocm_is_configured}
 
-def rocm_hipblaslt():
-    return %{rocm_is_configured} and %{rocm_hipblaslt}
-
 def if_rocm_hipblaslt(x):
```
alekstheod, line 71: I think this can be completely removed and also the use of it!

pemeliya: ah I wanted to remove all of it.. but the point is we are cursed to keep Tensorflow build fine and this is used directly in [one place](https://github.com/tensorflow/tensorflow/blob/7c7fdefb932502ad7a92cccc91615e87bc2e15ae/tensorflow/core/kernels/BUILD#L3637).

seantalts (approved):

dimitar-asenov (approved):

PR #42192, opened 2026-05-07
[XLA:CPU][oneDNN] Use if_onednn_async instead of if_onednn in onednn_cc_test bazel functions

othakkar:
A [commit](https://github.com/openxla/xla/commit/b49f469068abe70c0631ebb472992ff8fce99c57) recently added the `manual` tag to oneDNN unit tests, since they were failing when `ENABLE_ONEDNN_ASYNC` is not defined, which does not work when one wants to run these tests via wildcard patterns like `bazel test //xla/service/cpu/tests/....`.

Instead of unconditionally adding the `manual` tag, this PR switches `onednn_cc_test` to use `if_onednn_async` so these targets only build when the oneDNN async runtime is enabled (`--define=build_with_onednn_async=true`), rather than for all oneDNN-enabled platforms.

### DIFF
--- a/xla/service/cpu/tests/BUILD
+++ b/xla/service/cpu/tests/BUILD
@@ -429,7 +429,7 @@ onednn_cc_test(
     ],
 )
 
-xla_cc_test(
+onednn_cc_test(
     name = "onednn_layer_norm_test",
     srcs = ["onednn_layer_norm_test.cc"],
     copts = tsl_copts(),
@@ -449,7 +449,7 @@ xla_cc_test(
     ],
 )
 
-xla_cc_test(
+onednn_cc_test(
     name = "onednn_softmax_test",
     srcs = ["onednn_softmax_test.cc"],
     copts = tsl_copts(),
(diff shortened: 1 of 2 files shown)


xla/service/cpu/tests/BUILD
seantalts: I think this one needs to be changed to onednn_cc_test in order to remain disabled right? Could you verify that none of these tests run if onednn_async is not defined? thank you!

seantalts (changes requested):

xla/service/cpu/tests/BUILD
othakkar: Thanks for noticing this! I missed this one. Updated it now (https://github.com/openxla/xla/pull/42192/changes/02c9d7b1ee9cbe4d6a310fedafe20f845d1bfa4c). 
Yes, I verified that none of these tests run if ENABLE_ONEDNN_ASYNC is not defined.

seantalts (approved):

othakkar:
@seantalts I pushed a change to fix the failing tests. Could you please review the changes again? Thanks.

seantalts (approved):

PR #42195, opened 2026-05-07
[xla:gpu] Allocate default kCollective memory from VMM allocator

ezhulenev:
`CudaVmmAllocator` now returns memory compatible with NCCL symmetric memory requirements, no need for a hack calling into NCCL.

### DIFF
--- a/xla/stream_executor/cuda/cuda_executor_test.cc
+++ b/xla/stream_executor/cuda/cuda_executor_test.cc
@@ -157,7 +157,7 @@ TEST(CudaExecutorTest, CreateCollectiveMemoryAllocatorWorks) {
   TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<MemoryAllocation> allocation,
                           allocator->Allocate(1024));
   EXPECT_NE(allocation->address().opaque(), nullptr);
-  EXPECT_EQ(allocation->address().size(), 1024);
+  EXPECT_GE(allocation->address().size(), 1024);
 }
 
 // TODO: b/420735471 - Enable test once fixed.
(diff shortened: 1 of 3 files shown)


ezhulenev:
@PatriosTheGreat @olegshyshkov friendly ping

xla/stream_executor/cuda/cuda_executor_test.cc
```diff
@@ -157,7 +157,7 @@ TEST(CudaExecutorTest, CreateCollectiveMemoryAllocatorWorks) {
   TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<MemoryAllocation> allocation,
                           allocator->Allocate(1024));
   EXPECT_NE(allocation->address().opaque(), nullptr);
-  EXPECT_EQ(allocation->address().size(), 1024);
+  EXPECT_GE(allocation->address().size(), 1024);
```
PatriosTheGreat, line 160: Did this increase because of alignment?
If so, can we compare the aligned result size?

PatriosTheGreat (changes requested):

xla/stream_executor/cuda/cuda_executor_test.cc
ezhulenev: Because of granularity, we over-allocate and now correctly report the true size so that BFC allocator doesn't waste padded bytes.

PatriosTheGreat (approved):

xla/stream_executor/cuda/cuda_executor_test.cc
PatriosTheGreat: Got it, thanks

PR #42197, opened 2026-05-07
[XLA:GPU] Recompute autotuner reference output per candidate on AMDGPU.

Ruturaj4:
The fusion autotuner exhibits a multi-process race on AMDGPU when several XLA processes share one physical GPU and concurrently run the autotuner pass: the hoisted reference output buffer in Autotuner::ProfileAll can be perturbed between candidate compares, causing every candidate to fail with WRONG RESULTS in a single failed pass and surfacing as

  jax.errors.JaxRuntimeError: INTERNAL: Autotuning failed for HLO ...
    All configs failed during profiling. Failures (N): WRONG RESULTS

under "pytest -n N". The race is at the reference-buffer level (cross- process flock on the pass eliminates the loud failure but introduces other timing pathologies; --xla_gpu_fusion_autotune_top_k_configs=1 does not cure since the race fires with even one candidate).

Add an AutotuneConfig knob, recompute_reference_per_candidate, that moves the GetReferenceOutput call inside Autotuner::ProfileAll's candidate loop, shrinking the contamination window to a single compute+compare pair. Enabled in AutotunerPass::Create when the StreamExecutor's platform is ROCm; default-off for CUDA and SYCL. Cost on AMDGPU is up to ~2x autotuner profile time, paid once at compile.

Verified on gfx950 / ROCm 7.2: 0 autotuner WRONG RESULTS across 10,079 tests in a CI-density pytest -n 16 sweep (vs ~3 per 8-proc run before the patch).

📝 Summary of Changes

🎯 Justification
change.

🚀 Kind of Contribution

🧪 Unit Tests:
transformed HLO.

🧪 Execution Tests:

### DIFF
--- a/xla/backends/autotuner/autotuner.cc
+++ b/xla/backends/autotuner/autotuner.cc
@@ -645,8 +645,11 @@ absl::StatusOr<std::vector<Autotuner::ConfigResult>> Autotuner::ProfileAll(
       std::unique_ptr<InputBuffers> input_buffers,
       profiler_->CreateInputBuffers(candidates[0].executable.get(), instr));
 
+  // Compute the reference output once, unless the caller asked to recompute
+  // it per candidate (see AutotuneConfig::recompute_reference_per_candidate).
   std::optional<ScopedShapedBuffer> reference_output;
-  if (autotune_config_.check_buffers) {
+  if (autotune_config_.check_buffers &&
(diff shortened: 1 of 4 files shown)


pemeliya:
@Ruturaj4, does a similar issue happen on CUDA platform?

PR #42237, opened 2026-05-08
Autotuner: tolerate losing races to upload results

olupton:
📝 Summary of Changes
If multiple compilations are proceeding in parallel, different shards of different compilations might be autotuning the same fusion at the same time and racing each other to upload results to the KV store. This was flaky, because the upload-if-it-doesn't-already-exist logic was not atomic.

🎯 Justification
This fixes flakiness in the JAX test //tests/multiprocess:pjit_test_gpu, which triggers multiple parallel compilations that race each other to upload sharded autotuner results to the KV store.

🚀 Kind of Contribution
🐛 Bug Fix, 🧪 Tests

n/a

🧪 Unit Tests:
Tests added to verify one kind of error code is tolerated and others are not.

🧪 Execution Tests:
n/a


dimitar-asenov (approved):

olupton:
@dimitar-asenov let me know if you need anything from my side

PR #42259, opened 2026-05-08
[xla:ffi] When setting opaque state require TypeInfo

ezhulenev:
When setting FFI state via `TypeId` require passing `TypeInfo` as well. Templated APIs are supposed to be used in statically linked binaries, and when different FFI handlers can be in separate object files we rely on opaque type ids and state, and always lookup them through FFI API.

This is one more fix required for making "internal" FFI headers "safe" when linked as multiple object files. This is still not safe in general and not recommended, but this is an escape hatch that makes it work correctly when XLA and FFI handler compiled with the same toolchain.

### DIFF
--- a/xla/ffi/execution_state.cc
+++ b/xla/ffi/execution_state.cc
@@ -31,19 +31,14 @@ limitations under the License.
 
 namespace xla::ffi {
 
-absl::Status ExecutionState::Set(TypeId type_id, void* state) {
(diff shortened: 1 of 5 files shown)


seantalts (approved):

seantalts (approved):

ezhulenev:
Pushed a fix for a test.

seantalts (approved):

dykescorey4-netizen (approved):

PR #42306, opened 2026-05-11
Avoid normalizing batched gathers with overlapping batching dims and start_index_map

zeph1912:
**Summary of Changes**

Fixes BatchedGatherScatterNormalizer so it does not rewrite a valid batched gather when operand_batching_dims overlaps start_index_map.

The normalizer rewrites a batched gather into an ordinary gather by prepending operand_batching_dims to start_index_map. If the same operand dimension is present in both lists, the rewritten ordinary gather has duplicate start_index_map entries, for example {0, 0}. Ordinary gather validation requires those entries to be distinct, so the pass can emit invalid HLO.

The patch adds a small disjointness guard before matching batched gathers for normalization. When the overlap case appears, the pass now leaves the original valid batched gather unchanged.

**Justification**

This is a correctness fix for invalid IR generation in an HLO pass. A legal batched gather can encode one dimension as both an explicit batching dimension and an indexed dimension. XLA's batched gather semantics handle that case by giving explicit batching dims precedence, but the current normalization loses that distinction and constructs an ordinary gather whose metadata is rejected by the ordinary gather verifier.

A minimal poc:

```hlo
ROOT %gather = s32[4,40,2]{2,1,0}
  gather(s32[4,2]{1,0} %pad, s32[4,40,1]{2,1,0} %start_indices),
  offset_dims={2}, collapsed_slice_dims={}, start_index_map={0},
  operand_batching_dims={0}, start_indices_batching_dims={0},
  index_vector_dim=2, slice_sizes={1,2}
```

Before this patch, normalization prepends operand_batching_dims={0} to start_index_map={0}, producing start_index_map={0,0} for an ordinary gather. After this patch, the pass recognizes that this normalization would be
ill-formed and skips the rewrite.

**Kind of Contribution**

Bug Fix, Tests

**Unit Tests:**

Added
`BatchedGatherScatterNormalizerTest.DoNotNormalizeBatchGatherWhenBatchingDimsOverlapStartIndexMap`.

The test builds a minimal HLO module with overlapping `operand_batching_dims={0}` and `start_index_map={0}`, runs
`BatchedGatherScatterNormalizer`, and asserts that the pass reports no change and preserves the original batched gather metadata.

**Execution Tests:**

No new execution test. This is a pass-level IR validity fix covered by the unit test above.

### DIFF
--- a/xla/service/BUILD
+++ b/xla/service/BUILD
@@ -6090,6 +6090,7 @@ xla_cc_test(
     srcs = ["batched_gather_scatter_normalizer_test.cc"],
     deps = [
         ":batched_gather_scatter_normalizer",
+        "//xla/hlo/ir:hlo",
         "//xla/hlo/testlib:hlo_hardware_independent_test_base",
         "//xla/tests:xla_internal_test_main",
         "@com_google_absl//absl/strings:string_view",
--- a/xla/service/batched_gather_scatter_normalizer.cc
+++ b/xla/service/batched_gather_scatter_normalizer.cc
@@ -38,9 +38,24 @@ limitations under the License.
 namespace xla {
 
 namespace {
+bool HasOverlap(absl::Span<const int64_t> lhs, absl::Span<const int64_t> rhs) {
+  return absl::c_any_of(lhs, [&](int64_t value) {
+    return absl::c_linear_search(rhs, value);
+  });
+}
+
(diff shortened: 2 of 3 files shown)


zeph1912:
Gently pinging

PatriosTheGreat (approved):
Thanks

PR #42349, opened 2026-05-11
[XLA:GPU] Disable transfer buffer coalescing

Finistere:
📝 Summary of Changes

Disable coalescing in `TransferBufferSubAllocator`. Transfer buffers are backed by pinned host allocations and are passed directly to D2H/H2D copies. Allowing BFC to coalesce adjacent regions can produce transfer ranges that do not correspond to a single underlying pinned allocation, which can trigger intermittent `CUDA_ERROR_INVALID_VALUE` failures during host offloading (`_xla_compute_type=host`).

This seems to match the CUDA host-allocation documentation, which notes that the driver tracks the virtual memory ranges allocated by `cuMemHostAlloc`. Keeping each transfer buffer within a single `HostMemoryAllocate` region avoids constructing coalesced ranges that were never explicitly allocated as one pinned host region.

🎯 Justification
Host offloading memory transfers would fail with `CUDA_ERROR_INVALID_VALUE`. I'm seeing this when offloading on the 5090 with fairly large buffers.

🚀 Kind of Contribution

### DIFF
--- a/xla/backends/gpu/host_offloading/BUILD
+++ b/xla/backends/gpu/host_offloading/BUILD
@@ -60,13 +60,17 @@ xla_test(
     backends = ["gpu"],
     deps = [
         ":gpu_host_offloading_allocator",
+        "//xla/core/host_offloading:host_offloading_allocator",
         "//xla/service:platform_util",
+        "//xla/stream_executor:device_address",
         "//xla/stream_executor:platform_manager",
+        "//xla/stream_executor:stream",
         "//xla/stream_executor:stream_executor_h",
         "//xla/tsl/platform:statusor",
         "@com_google_absl//absl/status",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings",
         "@com_google_googletest//:gtest_main",
+        "@tsl//tsl/platform:status_matchers",
     ],
 )
--- a/xla/backends/gpu/host_offloading/gpu_host_offloading_allocator.cc
+++ b/xla/backends/gpu/host_offloading/gpu_host_offloading_allocator.cc
@@ -124,7 +124,11 @@ class TransferBufferSubAllocator : public tsl::SubAllocator {
   void* Alloc(size_t alignment, size_t num_bytes, size_t* bytes_received) final;
   void Free(void* ptr, size_t num_bytes) final;
 
-  bool SupportsCoalescing() const final { return true; }
+  // Transfer buffers are passed directly to D2H/H2D copies. Coalescing
+  // adjacent HostMemoryAllocate regions into one logical BFC range can trigger
+  // CUDA_ERROR_INVALID_VALUE during transfers, so keep each chunk within a
+  // single HostMemoryAllocate region.
+  bool SupportsCoalescing() const final { return false; }
 
  private:
   stream_executor::StreamExecutor* executor_;
--- a/xla/backends/gpu/host_offloading/gpu_host_offloading_allocator_test.cc
+++ b/xla/backends/gpu/host_offloading/gpu_host_offloading_allocator_test.cc
@@ -15,14 +15,23 @@ limitations under the License.
 
 #include "xla/backends/gpu/host_offloading/gpu_host_offloading_allocator.h"
 
+#include <cstddef>
+#include <memory>
+#include <utility>
+#include <vector>
+
 #include <gtest/gtest.h>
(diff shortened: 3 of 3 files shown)


xla/backends/gpu/runtime/host_execute_thunk.cc
```diff
@@ -627,14 +636,25 @@ HostExecuteStartThunk::GetAsyncEventsUniqueId() const {
...
     Thunk::ThunkInfo thunk_info,
+    absl::InlinedVector<HostExecuteStartThunk::SliceAndShape, 4> results,
     std::shared_ptr<HostExecuteAsyncEvents> async_events)
     : HostAsyncThunk(Thunk::Kind::kHostExecuteDone, std::move(thunk_info)),
+      results_(std::move(results)),
       async_events_(std::move(async_events)) {
   CHECK(async_events_) << "async_events must not be null";
 }
 
 std::string HostExecuteDoneThunk::ToString(int indent) const { return ""; }
```
ezhulenev, line 638: FWIW other async done thunk have empty buffer uses (AsyncDoneThunk: https://github.com/openxla/xla/blob/main/xla/backends/gpu/runtime/async_thunk.h#L112-L114), I think this is more correct as done is purely even sync without any actual data movements happening at run time.

ExecutionGraph doesn't handle it correctly today as it doesn't have a notion of "async" operation, but this is not a problem yet because it's not used for thunk scheduling. I want to make a proper fix that adds sone "async reasoning" to runtime.

ezhulenev: AsyncStart/Done pairs dependencies are printed in thunk sequence via async_event, host async needs a little bit modernization to become more consistent with device async. Ideally I'd love to standardize all async execution under one or two thunks (Async and HostAsync?)

basioli-k: Agree with Eugene.
Would prefer to be consistent here

Finistere: Sorry for the delay. Thanks for the context! I was initially a bit confused when hitting the `CUDA_ERROR_INVALID_VALUE` as to why those thunks didn't have any dependencies. I'm removing this from the PR, definitely makes sense to be consistent

Finistere: Removed the buffer uses changes, the PR now only contains the disabling of the buffer coalescing.

Finistere:
To be honest, I'm not 100% certain that's the best fix. I didn't see any issue since this patch, but maybe there's a better solution. Maybe I should have created an issue instead of a PR?

basioli-k:
> To be honest, I'm not 100% certain that's the best fix. I didn't see any issue since this patch, but maybe there's a better solution. Maybe I should have created an issue instead of a PR?

Would it be possible to add a unit test so that we can repro the issue?

dimitar-asenov:
@Finistere Please address the open comment about adding a unit test, or close this PR if you no longer wish to merge it.

Finistere:
Hey @basioli-k, @dimitar-asenov, sorry for the late response. I've added a unit test that reproduces the behavior I'm seeing:
```sh
bazel test //xla/backends/gpu/host_offloading:gpu_host_offloading_allocator_coalescing_test
```

fails without disabling the coaslescing (RTX 5090 driver 595.84):
```
Expected: is OK
  Actual: INTERNAL: CUDA error: : CUDA_ERROR_INVALID_VALUE: invalid argument
xla/stream_executor/cuda/cuda_status.cc:50
xla/stream_executor/cuda/cuda_stream.cc:129
 (of type absl::lts_20260526::Status)
```

xla/backends/gpu/host_offloading/gpu_host_offloading_allocator_coalescing_test.cc
basioli-k, line 1: move test to xla/backends/gpu/host_offloading/gpu_host_offloading_allocator_test.cc

basioli-k (changes requested):

xla/backends/gpu/host_offloading/gpu_host_offloading_allocator_coalescing_test.cc
Finistere: fixed

dimitar-asenov (approved):

PR #42361, opened 2026-05-11
[TSL]: Internal Retries for GCE Metadata Authentication

EmilNichita:
### 📝 Summary of Changes

This PR introduces an optional retry loop directly inside the C++ GCE authentication flow (GoogleAuthProvider::GetToken).
The retry behavior is fully configurable via environment variables (with sensible defaults):
- TF_GCS_AUTH_MAX_REQUESTS: Maximum number of attempts to contact the GCE metadata server (default: 1).
- TF_GCS_AUTH_RETRY_DELAY_SEC: Delay in seconds between retry attempts (default: 5).

### 🎯 Justification
When running TensorFlow workloads on GKE (or GCE), the local metadata server (`gke-metadata-server` or `metadata.google.internal`) can experience delayed startup or temporary unresponsiveness under heavy node load.

Currently, if GoogleAuthProvider::GetToken attempts to contact the metadata server during this transient window and fails, it would:
- Log a warning about GCE metadata failure.
- Cache an empty bearer token ("").
- Disable future GCE checks for the lifetime of the process (expiration_timestamp_sec_ = 0 or UINT64_MAX depending on flags).

As a result, transient node startup delays caused permanent PermissionDenied ("Anonymous caller does not have storage.objects.get access...") errors for any subsequent GCS read/write operations, forcing entire pipeline tasks to fail and restart. **For example**, this essentially bricks any container using a tensorflow `epath` backend that has one bad auth try.

### 🚀 Kind of Contribution
🐛 Bug Fix, ⚡️ Performance Improvement

### 📊 Benchmark (for Performance Improvements)
Not applicable.

### 🧪 Unit Tests:
Added a unit test to `google_auth_provider_test.cc`

### 🧪 Execution Tests:

Not applicable.

### DIFF
--- a/xla/tsl/platform/cloud/google_auth_provider.cc
+++ b/xla/tsl/platform/cloud/google_auth_provider.cc
@@ -37,6 +37,7 @@ limitations under the License.
 #include <utility>
 
 #include "absl/strings/match.h"
+#include "absl/strings/numbers.h"
 #include "absl/synchronization/mutex.h"
 #include "json/json.h"
 #include "xla/tsl/platform/env.h"
@@ -63,8 +64,8 @@ constexpr char kCloudSdkConfig[] = "CLOUDSDK_CONFIG";
 // the GCE metadata service.
 constexpr char kNoGceCheck[] = "NO_GCE_CHECK";
 
-// The environment variable to set the maximum number of retries for GCE auth.
-constexpr char kGcsAuthMaxRetries[] = "TF_GCS_AUTH_MAX_RETRIES";
+// The environment variable to set the maximum number of requests for GCE auth.
+constexpr char kGcsAuthMaxRequests[] = "TF_GCS_AUTH_MAX_REQUESTS";
 
 // The environment variable to set the delay in seconds between GCE auth retries.
 constexpr char kGcsAuthRetryDelaySec[] = "TF_GCS_AUTH_RETRY_DELAY_SEC";
@@ -139,6 +140,24 @@ absl::Status GetWellKnownFileName(std::string* filename) {
   return absl::OkStatus();
 }
 
+absl::Status ParseNonNegativeIntEnvVar(const char* env_var_name,
+                                       int default_value, int* result) {
+  if (!result) {
+    return absl::FailedPreconditionError("'result' cannot be nullptr.");
+  }
+  const char* env_value = std::getenv(env_var_name);
+  if (!env_value) {
+    *result = default_value;
+    return absl::OkStatus();
+  }
+  if (!absl::SimpleAtoi(env_value, result) || *result < 0) {
+    return absl::InvalidArgumentError(
+        absl::StrCat("Failed to parse $", env_var_name,
+                     " as a non-negative integer: ", env_value));
+  }
+  return absl::OkStatus();
+}
+
 }  // namespace
 
 GoogleAuthProvider::GoogleAuthProvider(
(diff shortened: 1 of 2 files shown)


xla/tsl/platform/cloud/google_auth_provider.cc
```diff
@@ -180,7 +186,22 @@ absl::Status GoogleAuthProvider::GetToken(std::string* t) {
                      absl::StrCat("GCE check skipped due to presence of $",
                                   kNoGceCheck, " environment variable."));
   } else {
-    token_from_gce_status = GetTokenFromGce();
+    const char* max_retries_env = std::getenv(kGcsAuthMaxRetries);
+    const char* retry_delay_env = std::getenv(kGcsAuthRetryDelaySec);
+    int max_retries = max_retries_env ? std::atoi(max_retries_env) : 1;
```
dimitar-asenov: Here and below, please use absl::SimpleAtoi, and return an error if the environment variable was provided but it could not be parsed.

xla/tsl/platform/cloud/google_auth_provider.cc
```diff
@@ -180,7 +186,22 @@ absl::Status GoogleAuthProvider::GetToken(std::string* t) {
...
+    int retry_delay_sec = retry_delay_env ? std::atoi(retry_delay_env) : 5;
+
+    for (int i = 0; i < max_retries; ++i) {
+      token_from_gce_status = GetTokenFromGce();
+      if (token_from_gce_status.ok()) {
+        break;
+      }
+      if (i < max_retries - 1) {
+        LOG(INFO) << "GCE auth failed with status: " << token_from_gce_status.ToString()
+                  << ". Retrying in " << retry_delay_sec << " seconds... (Attempt " << i + 1 << " of " << max_retries << ")";
```
dimitar-asenov: The numbers reported here aren't quite right. E.g. for max_retries 3, the code will print two things:
(Attempt 1 of 3)
(Attempt 2 of 3) 

and then stop. It would be nice if there was (Attempt 3 of 3).

Either print `i+2`, or `max_retries-1` to make sure that the numbers match. I think `i+2` is a bit better, because it is the 2nd and 3rd attempts (even though it's the 1st and 2nd retry, because the original attempt was not a *re*-try).

dimitar-asenov (changes requested):

dimitar-asenov (approved):

xla/tsl/platform/cloud/google_auth_provider.cc
EmilNichita: Moved the check to its own function, let me know if that's an appropriate place for it 🙌

xla/tsl/platform/cloud/google_auth_provider.cc
EmilNichita: Good catch! On second thought, renamed `TF_GCS_AUTH_MAX_REQUESTS` closer to how it's specified for ex. in the standard `requests` python package - i.e. max total number of requests attempted. Printing as you suggested i+2. 
(skipping the first log to preserve identical behavior for users from before this PR)

xla/tsl/platform/cloud/google_auth_provider.cc
dimitar-asenov: Looks good. Thanks.

xla/tsl/platform/cloud/google_auth_provider.cc
dimitar-asenov: Thanks.

dimitar-asenov (approved):

PR #42383, opened 2026-05-11
[ROCm] xla/pjrt: honor XLA_PYTHON_CLIENT_ALLOCATOR=platform

magaonka-amd:
📝 Summary of Changes
Make GpuAllocatorConfig::Kind::kPlatform actually deliver a synchronous passthrough allocator instead of returning nullptr. Previously the env var XLA_PYTHON_CLIENT_ALLOCATOR=platform was silently overridden by the BFC allocator that Backend::Backend() builds unconditionally for CUDA/ROCm platforms.

🎯 Justification
without this fix 

🚀 Kind of Contribution
🐛 Bug Fix, 

🧪 Execution Tests:
setting this XLA_PYTHON_CLIENT_ALLOCATOR= flag from JAX now working after the fix and honoring the flag value.

### DIFF
--- a/xla/pjrt/gpu/BUILD
+++ b/xla/pjrt/gpu/BUILD
@@ -140,6 +140,7 @@ cc_library(
         "//xla/stream_executor:memory_space",
         "//xla/stream_executor:platform",
         "//xla/stream_executor:stream",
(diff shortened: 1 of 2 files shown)


magaonka-amd:
pushed better fix hence closing this.

PR #42396, opened 2026-05-12
[xla:gpu] Use DynamicSliceAnnotator configs in dynamic memcpy rewriter

ezhulenev:
- Simplify `FusionDynamicMemcpyRewriter` by reading pre-annotated `DynamicSliceConfig` backend configs from DS/DUS instructions instead of performing complex runtime analysis
- Remove `DynamicOffsetEvaluator` class and all helper functions (`EvaluateDynamicOffsets`, `GetLoopConfig`, `ComputeOffsetsForLoop`, `SetLoopMemcpyConfig`) that walked call stacks and used `HloEvaluator` to compute per-iteration offsets
- Instead, read the `DynamicSliceConfig` (byte\_offset + byte\_stride) already placed on DS/DUS by the `DynamicSliceAnnotator` pass, and compute per-iteration offsets as `clamp(byte_offset + i * byte_stride, 0, max_offset)`
- Add `DynamicSliceAnnotator` to `RunDynamicSliceFusionPasses` in gpu\_compiler so annotations are available before fusion dispatch

This is one step toward unifying all DS/DUS-like fusion on top of new shared infra.


ermilovmaxim (approved):

akuegel (commented):
It looks like this test is newly failing with this change:

//xla/backends/gpu/tests:offload_scan_output.hlo.test_h100_sxm
(and for other GPU configs as well)

Can you please check whether you can reproduce the failure?
`FAILED_PRECONDITION: Failed to evaluate instruction (param_2) since it depends on infeed or parameters to its parent computation (async_computation). [EvalErrorDetailUrl='\x00\x00\x00\x00']`

ezhulenev:
@akuegel pushed a fix + test for `DynamicSliceAnnotator`

ermilovmaxim (approved):

Fleinsopp:
File change diff

PR #42415, opened 2026-05-12
[XLA:GPU] Take shared lock on GPU when checking buffers during autotune

Finistere:
📝 Summary of Changes
When compiling multiple programs with PJRT_Client_Compile in parallel we see a lot of DelayKernels timing out. From my investigation and understanding, it's due to CUDA calls like cuMemcpyDtoHAsync done during the buffer checks that block while the DelayKernel is spinning on the same stream. The problem is that DelayKernel spins until the thread that launched it could finish the profiling setup which requires a few CUDA calls (cuRecordEvent, etc.). And those can't be executed because cuMemcpyDtoHAsync is blocked, so everything only ends up unblocked once the DelayKernel times out.

Taking the shared lock before doing any checks solves almost all the problems I've seen with parallel autotuning and also avoids any impact on the timings.

🎯 Justification
Allows efficient parallel auto-tuning.

🚀 Kind of Contribution


sohaibiftikhar (commented):
Please fix the CI errors and then ping me for approval.

Finistere:
@sohaibiftikhar the two remaining failing checks don't seem related to this PR, especially the zizmor one. Should I take a look at the DWYU ones?
```
buildozer 'add deps :cublas' //xla/backends/gpu/autotuner:fission_backend_test
buildozer 'add deps :rocblas' //xla/backends/gpu/autotuner:fission_backend_test
```

sohaibiftikhar (approved):
zizmor is fine. I am not even sure what it does.

PR #42431, opened 2026-05-12
Fix std::min template deduction on darwin arm64

seantalts:
## Summary

On macOS arm64, `size_t` (`unsigned long`) and `uint64_t` (`unsigned long long`) are distinct integer types. On Linux they both alias to `unsigned long`, so the template argument deduction for `std::min` succeeds. Apple-clang rejects the call site at `buffers_float_check_thunk.cc:176` with:

```
no matching function for call to 'min(uint64_t, size_t)'
```

Pinning the explicit template argument (`std::min<uint64_t>(...)`) fixes the build on Mac and is a no-op on Linux.

## Test plan

- [x] Builds cleanly on darwin_arm64 (Mac Studio M-series, macOS 26, apple-clang 21).
- [x] Build on Linux x86_64 is unchanged (template arg deduction would have succeeded with the same result type).
- [x] No behavior change — same call, same arguments, just explicit on the template parameter.


seantalts:
Closing — opened in error against upstream. Branch lives on seantalts/xla; not for upstream submission yet.

PR #42432, opened 2026-05-12
[XLA:CPU] Add xtile.extract_aligned / xtile.insert_aligned

seantalts:
## Summary

Adds two new ops to the xtile dialect — `xtile.extract_aligned` and `xtile.insert_aligned` — that mirror `xtile.extract` / `xtile.insert` but carry an explicit precondition: for every dim `i`,

```
offsets[i] + (full_tile_shape[i] - 1) * strides[i] < buffer_shape[i]
```

Under this precondition the partial-tile case cannot occur, so bufferization emits the unguarded subview chain — no `scf.if`, no padded-buffer alloc on the boundary path. This lets LLVM SROA/DSE fold the per-iteration scratch buffers in the common case.

The loop emitter (`emitter_helpers.cc::EmitParameterExtract` and the `InsertTileOp::create` site in `fusion_emitter.cc`) chooses the aligned variant when `buffer_shape[i]` is statically divisible by `padded_tile_size[i] * tile_stride[i]` on every non-reduced dim. Under the standard `step = padded_tile_size * tile_stride` iteration pattern this implies every tile access is in-range.

## Measured impact

On a JAX-emitted `f32[1024] add` fusion:

| | baseline | this PR |
|---|---|---|
| pre-opt LLVM IR allocas | 5 | 3 |
| post-opt LLVM IR allocas | 2 | **0** |
| post-opt LLVM IR memcpys | 3 | **0** |
| loop body instructions/iter | ~25 | ~10 |

Generated loop body becomes the canonical SIMD add: 4× `ldp q,q`, 4× `fadd.4s`, 2× `stp q,q`, branch.

CPU microbenchmarks (Mac Studio M-series, 3-rep median, 7-rep for outlier re-check):

| Benchmark family | Δ vs baseline |
|---|---|
| `BM_AddF32/{128,1024,8192,32768}` | **−24% to −27%** |
| `BM_FusionF32/*` | flat to −5% |
| `BM_ReduceAddF32/*` | within noise (±2%) |
| `BM_BatchedDot_F32_F32_*` | within noise (−3.5% to +2.4%) |

## Dependency note

This PR's head branch was developed off `mac-fix/std-min-uint64-deduction` for Mac build access; it includes that one-line commit as a prerequisite. See #42431 (the same author's std::min fix). Once #42431 lands, this PR's diff will reduce to just the xtile changes.

## Follow-ups (not in this PR)

- `ShapedType::isDynamic()` guard in `TileInfoIsAlwaysFull` for dynamic-shape robustness.
- `xla/codegen/xtile/codegen/experimental_fusion_emitter.cc` has a matching `InsertTileOp` site (line ~1037) used by the Triton/GPU `xtile_compiler` path; intentionally not touched here (different emitter, would need separate validation).
- The layout-fixup alloc still fires inside `ExtractAlignedTile` when the subview has a non-identity (`strided<[1], offset:?>`) layout from a dynamic offset. Relaxing `to_tensor`'s identity-layout precondition (or moving to `vector.transfer_read` directly) would eliminate it. See accompanying design doc.

## Test plan

- [x] New lit cases in `xla/codegen/xtile/ir/transforms/tests/bufferize.mlir` cover both ops:
  - `@extract_aligned_dynamic_offset` asserts `CHECK-NOT: scf.if` + alloc+copy fixup for non-identity layout
  - `@insert_aligned_dynamic_offset` asserts `CHECK-NOT: scf.if` + direct subview/copy
- [x] `bazel test //xla/codegen/xtile/ir/transforms/tests:bufferize.mlir.test` passes
- [x] Existing tests in the same file (which exercise the unconstrained ops with bare offsets) continue to pass
- [x] Spot-check of CPU benchmarks (dot, reduce, fusion, elementwise) shows no regressions on darwin_arm64
- [ ] CI sweep on Linux x86_64

### DIFF
--- a/xla/backends/gpu/runtime/buffers_float_check_thunk.cc
+++ b/xla/backends/gpu/runtime/buffers_float_check_thunk.cc
@@ -173,7 +173,7 @@ absl::Status CheckFloatsAndLog(
   // results of the previous FloatCheck operation are available.
   TF_RETURN_IF_ERROR(reduce_append_kernel.Launch(
       thread_dim, se::BlockDim(1, 1, 1), stream, tmp_ptr,
-      std::min(tmp_ptr.ElementCount(), num_blocks), entry_id,
+      std::min<uint64_t>(tmp_ptr.ElementCount(), num_blocks), entry_id,
       buffer_debug_log.GetDeviceHeader(), buffer_debug_log.GetDeviceEntries()));
 
   return absl::OkStatus();
--- a/xla/codegen/xtile/codegen/emitter_helpers.cc
+++ b/xla/codegen/xtile/codegen/emitter_helpers.cc
@@ -15,6 +15,8 @@ limitations under the License.
 
 #include "xla/codegen/xtile/codegen/emitter_helpers.h"
 
+#include <algorithm>
+#include <cstddef>
 #include <cstdint>
 #include <optional>
 #include <utility>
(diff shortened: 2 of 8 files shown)


seantalts:
Closing — opened in error against upstream. Branch lives on seantalts/xla; not for upstream submission yet.

PR #42455, opened 2026-05-12
[XLA:GPU] Use generic SM100 keys for op profiles

and-ivanov:
📝 Summary of Changes
This changes SM100 HLO math op profiles to use the generic sm_100 profile key instead of distinguishing
B200 and GB200.

For profile tables where GPU-level/device-level characteristics matter, this keeps device-specific
profile keys via GetDeviceSpecificProfileName(). Matmul and collective profile lookup/generation now use
that path, preserving keys such as sm_100_B200 and sm_100_GB200.

🎯 Justification
For scalar/math HLO op profiles, B200 and GB200 should not need separate entries within the same SM100 architecture.

Matmul profiles are different: they are measured for the whole GPU rather than a single SM, so sharing
one model across B200 and GB200 is likely inaccurate because the devices differ in SM count, DRAM
bandwidth, and other hardware details.

Collective profiles also remain device-specific because topology matters, and topology depends on the
GPU/system variant.

🧪 Unit Tests:
//xla/service/gpu/model:hlo_op_profiles_test `TEST_F(HloOpProfilesTest, GetProfileSm100)`

### DIFF
--- a/xla/service/gpu/model/collective_interpolator.cc
+++ b/xla/service/gpu/model/collective_interpolator.cc
@@ -396,7 +396,7 @@ HloOpcode AsyncToSyncOpcode(const HloCollectiveInstruction& instr) {
 
 absl::StatusOr<HloInstructionProfileList> ReadDefaultProfiles(
     const se::DeviceDescription& device_info) {
-  std::string key = HloOpProfiles::GetProfileName(device_info);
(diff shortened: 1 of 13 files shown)


akuegel (approved):
Thank you :)

PR #40557, opened 2026-04-08
[ROCm] Enable scaled dot for ROCm.

zoranjovanovic-ns:
📝 Summary of Changes
Enable scaled dot on ROCm, updated corresponding unit tests.

🎯 Justification
supported by triton and hipblaslt

🚀 Kind of Contribution
Please remove what does not apply: ✨ New Feature, 🧪 Tests

📊 Benchmark (for Performance Improvements)
Please measure and include speedups for one of the public HLOs in
`compiler/xla/tools/benchmarks/hlo/`.

🧪 Unit Tests:
//xla/backends/gpu/autotuner:triton_test_amdgpu_any
//xla/backends/gpu/autotuner:hipblaslt_test_amdgpu_any

🧪 Execution Tests:
What execution tests were added? For example, a new optimization should be
tested with an end-to-end execution test triggering the optimization and
asserting correctness. Please provide test cases running with at most 2 GPUs.

PR #42457, opened 2026-05-12
make mx type unit tests shared by triton and hipblaslt backends

ScXfjiang:
After https://github.com/openxla/xla/pull/40557, ROCm has two backends that supports scaled dot, so we adjust the test to be shared by both backends.

In addition:
- I disable Triton backend for ROCm for mixed type, i.e., FP8 * FP4
- I revert the hipblaslt limitation for the batch size

### DIFF
--- a/xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
+++ b/xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
@@ -81,9 +81,8 @@ class MxScaledDotExecutionTest : public HloPjRtGpuTestBase {
     TF_ASSERT_OK_AND_ASSIGN(auto test_optimized,
                             GetOptimizedModule(hlo_string, test_config));
     EXPECT_THAT(
-        RunFileCheck(
-            test_optimized->ToString(),
-            "CHECK: {{(__cublas\\$lt\\$matmul\\$mx)|(scaled-dot)}}"),
+        RunFileCheck(test_optimized->ToString(),
+                     "CHECK: {{(__cublas\\$lt\\$matmul\\$mx)|(scaled-dot)}}"),
         absl_testing::IsOkAndHolds(true));
   }
 };
--- a/xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
+++ b/xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
@@ -29,7 +29,7 @@ limitations under the License.
 namespace xla::gpu {
 namespace {
 
-class HipblasLtMxExecutionTest : public HloPjRtGpuTestBase {
+class MxScaledDotExecutionTest : public HloPjRtGpuTestBase {
  protected:
   void SetUp() override {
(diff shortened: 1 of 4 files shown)


xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
```diff
@@ -39,8 +39,6 @@ class HipblasLtMxExecutionTest : public HloPjRtGpuTestBase {
     }
   }
 
-  // Runs numerical correctness (hipBLASLt MX vs decomposed reference) and
-  // verifies that the optimized HLO uses the expected custom call target.
```
zoranjovanovic-ns, line 43: Can you modify instead of removing the comment?

zoranjovanovic-ns (approved):

xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
ScXfjiang: I have updated the comment.

ScXfjiang:
Hi @xla-rotation, could you take a look please? Thanks!

xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
```diff
@@ -39,8 +39,8 @@ class HipblasLtMxExecutionTest : public HloPjRtGpuTestBase {
...
+  // Runs numerical correctness and verifies that the optimized HLO uses the
+  // expected custom call target.
```
akuegel, line 43: You are removing the part that the optimized HLO uses the expected custom call. So the comment is not correct anymore.

xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
```diff
@@ -64,27 +64,6 @@ class HipblasLtMxExecutionTest : public HloPjRtGpuTestBase {
     EXPECT_TRUE(RunAndCompareTwoModules(std::move(test_module),
                                         std::move(reference_module), error_spec,
                                         /*run_hlo_passes=*/true));
-
```
akuegel, line 67: I can see that this is somewhat redundant to what we do above, but not entirely. We are checking whether one of the custom calls is used. Maybe you can refactor it like this:

Keep the code below, delete the code above, and add a line at the end of the test: ```EXPECT_TRUE(RunAndCompareTwoModules(std::move(test_optimized),
                                        std::move(ref_optimized), error_spec,
                                        /*run_hlo_passes=*/false));```

ScXfjiang: Thanks for the feedback. I have updated the code.

xla/backends/gpu/autotuner/mx_scaled_dot_execution_test.cc
ScXfjiang: updated

akuegel (approved):

PR #42581, opened 2026-05-14
[XLA:CPU] Worker::Parallelize: decouple num_partitions from num_work_items

### LINKED ISSUE #42518: [XLA:CPU] Possible perf regression in `Worker::Parallelize`
Apologies if I'm misreading the code but I'd appreciate someone with context confirming or telling me I've missed something.

## Observation

At [`work_queue.h:240`](https://github.com/openxla/xla/blob/main/xla/backends/cpu/runtime/work_queue.h#L240):

```cpp
work_queue(num_work_items, /*num_partitions=*/this->count_down.count())
```

In the original [9d95a71119](https://github.com/openxla/xla/commit/9d95a71119), `count_down` counted **workers**, so this line correctly meant `num_partitions = num_workers`. Then [5db7c22b15](https://github.com/openxla/xla/commit/5db7c22b15) ("Make Worker::Parallelize deadlock-proof") repurposed `count_down` to count **items**:

```diff
-  tsl::CountDownAsyncValueRef<tsl::Chain> count_down(num_workers);
+  tsl::CountDownAsyncValueRef<tsl::Chain> count_down(num_tasks);
```

which makes sense as a correctness fix, but unless I'm missing something the partition-count line wasn't updated. `count_down.count()` now returns `num_tasks`, so we end up with one `ABSL_CACHELINE_ALIGNED` partition per item: 64B each, `FixedArray<Partition, 32>` spills to heap for any job > 32 items, and every `Pop` walks to a fresh cache line via work-stealing wrap-around.

Could be deliberate for stealing granularity, but the line reads like it intended the old semantics.

## Impact

Added a small `BM_WorkerParallelize` micro (noop body, AMD EPYC 9634 SMT, 3 trials × 20 reps × min-time 0.3 s):

| `num_work_items` | `partitions = items` (current) | `partitions = workers` | diff |
|---:|---:|---:|---:|
| 128 | 45,412 ns | 16,251 ns | −64% |
| 256 | 65,045 ns | 22,824 ns | −65% |
| 512 | 74,796 ns | 35,227 ns | −53% |
| 1024 | 82,599 ns | 52,293 ns | −37% |

## Proposed Fix

Substitute `num_workers` within `ParallelizeContext`:

```diff
       count_down(std::move(count_down)),
-      work_queue(num_work_items, /*num_partitions=*/this->count_down.count()),
+      work_queue(num_work_items, /*num_partitions=*/num_workers),
       parallel_work(std::forward<ParallelWork>(parallel_work)) {}
```

`WorkQueue` already supports `num_partitions != num_work_items` - [`work_queue_test.cc:43-81`](https://github.com/openxla/xla/blob/main/xla/backends/cpu/runtime/work_queue_test.cc#L43) covers `(14 items, 4 partitions)` etc., so I don't believe this breaks anything.

Happy to send a PR with the fix and the micro if this is wanted. Filing the issue first in case I've misread the design intent.

Perhaps @ezhulenev best positioned / most familiar with change?

zdryan:
📝 **Summary of Changes**

`Worker::ParallelizeContext` was constructing its `WorkQueue` with `num_partitions = count_down.count()`, which equals `num_work_items` at that point. That made every work item its own `ABSL_CACHELINE_ALIGNED` partition. The inline `FixedArray<Partition, 32>` spilled to heap for any sequence > 32 items, and every worker's `Pop` hit a different cache line via the wrap-around steal loop.

Change decouples `num_partitions` from `num_work_items` and floors partition size at 8 items:

```
num_partitions = max(1, min(num_workers, ceil(items / 8)))
num_workers    = min(num_workers, num_partitions)
```

Each partition holds >=8 items (last partition can have fewer). When `ceil(items / 8) == 1` we fall through to the existing serial short-circuit. `WorkQueue` itself already supported `num_partitions != num_work_items` (`work_queue_test.cc:43-81` exercises it); only the production init chose the degenerate config.

🎯 **Justification**

Fixes openxla/xla#42518. Any CPU workload that calls `Worker::Parallelize` with `num_work_items > 32` (or with `num_work_items < num_workers` on wide thread pools) was paying the cost of heap-allocated partitions and cross-cacheline `Pop` atomics on every dispatch.

🚀 **Kind of Contribution**

- 🐛 Bug Fix
- ⚡️ Performance Improvement

📊 **Benchmark (for Performance Improvements)**

`compiler/xla/tools/benchmarks/hlo/` contains GPU HLOs. This change is CPU runtime only. The included microbenchmark (`BM_WorkerParallelize`, 336-thread host) shows the deltas.

  | Items | Before    | After     | diff    |
  |-------|-----------|-----------|------|
  | 128   | 45,412 ns | 16,251 ns | -64% |
  | 256   | 65,045 ns | 22,824 ns | -65% |
  | 512   | 74,796 ns | 35,227 ns | -53% |
  | 1024  | 82,599 ns | 52,293 ns | -37% |

### DIFF
--- a/xla/backends/cpu/runtime/work_queue.h
+++ b/xla/backends/cpu/runtime/work_queue.h
@@ -34,6 +34,7 @@ limitations under the License.
 #include "xla/tsl/concurrency/async_value_ref.h"
 #include "xla/tsl/concurrency/chain.h"
 #include "xla/tsl/platform/logging.h"
+#include "xla/util.h"
 
 #define EIGEN_USE_THREADS
 #include "unsupported/Eigen/CXX11/Tensor"
@@ -221,7 +222,8 @@ template <typename ParallelWork>
 struct Worker::ParallelizeContext {
   ParallelizeContext(Eigen::ThreadPoolInterface* thread_pool,
                      tsl::CountDownAsyncValueRef<tsl::Chain> count_down,
-                     size_t num_work_items, ParallelWork&& parallel_work);
+                     size_t num_work_items, size_t num_partitions,
+                     ParallelWork&& parallel_work);
 
   Eigen::ThreadPoolInterface* thread_pool;
   tsl::CountDownAsyncValueRef<tsl::Chain> count_down;
@@ -234,10 +236,10 @@ template <typename ParallelWork>
 Worker::ParallelizeContext<ParallelWork>::ParallelizeContext(
     Eigen::ThreadPoolInterface* thread_pool,
     tsl::CountDownAsyncValueRef<tsl::Chain> count_down, size_t num_work_items,
-    ParallelWork&& parallel_work)
+    size_t num_partitions, ParallelWork&& parallel_work)
     : thread_pool(thread_pool),
       count_down(std::move(count_down)),
-      work_queue(num_work_items, /*num_partitions=*/this->count_down.count()),
+      work_queue(num_work_items, num_partitions),
       parallel_work(std::forward<ParallelWork>(parallel_work)) {}
 
 template <typename ParallelWork>
@@ -331,6 +333,10 @@ template <typename ParallelWork>
 ABSL_ATTRIBUTE_ALWAYS_INLINE tsl::AsyncValueRef<tsl::Chain> Worker::Parallelize(
     Eigen::ThreadPoolInterface* thread_pool, size_t num_workers,
     size_t num_work_items, ParallelWork&& parallel_work) {
+  if (ABSL_PREDICT_FALSE(num_work_items == 0)) {
+    return tsl::MakeAvailableAsyncValueRef<tsl::Chain>();
+  }
+
   // Short-circuit single-threaded execution.
   if (ABSL_PREDICT_FALSE(num_workers == 1)) {
     if (absl::Status status = ExecuteInline(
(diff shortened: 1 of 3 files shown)


xla/backends/cpu/runtime/work_queue_test.cc
```diff
@@ -28,6 +28,7 @@ limitations under the License.
 #include "xla/tsl/platform/test.h"
 #include "xla/tsl/platform/test_benchmark.h"
 #include "xla/tsl/platform/threadpool.h"
+#include "tsl/platform/cpu_info.h"
```
seantalts, line 31: should this be prefixed by xla/?

xla/backends/cpu/runtime/work_queue.h
```diff
@@ -345,15 +346,22 @@ ABSL_ATTRIBUTE_ALWAYS_INLINE tsl::AsyncValueRef<tsl::Chain> Worker::Parallelize(
...
-  // Ensure we don't launch more workers than work items. Extra workers would be
-  // idle or cause out-of-bounds partition access.
-  num_workers = std::min(num_work_items, num_workers);
+
+  // Floor partition size so each Pop amortises across multiple items (one cache
+  // line) before stealing. Cap num_workers to num_partitions to avoid idle
+  // Eigen tasks and out-of-bounds partition access in Worker(worker_index, ...).
+  constexpr size_t kMinItemsPerPartition = 8;
+  const size_t num_partitions =
+      std::max<size_t>(1, std::min(num_workers, (num_work_items +
```
seantalts: num_workers could have been 0 before - I'm not sure what the implications are of not allowing 0 now. If there are no work items do we still schedule a task on the eigen threadpool? Maybe handle num_work_items == 0 explicitly?

seantalts:
ty @zdryan !

xla/backends/cpu/runtime/work_queue_test.cc
zdryan: The header lives at `third_party/tsl/tsl/platform/cpu_info.h` - so should be correct as is.

Similar pattern used ubiquitously elsewhere for including `cpu_info.h`, for instance:
https://github.com/openxla/xla/blob/9db19f8005c83cf594d2d65baa14169a60358a49/xla/backends/cpu/runtime/kernel_test.cc#L34

xla/backends/cpu/runtime/work_queue.h
zdryan: Good point - we will needlessly allocate a parallelization context + work queue + worker construction.

This case actually seems to be UB (currently) outside of debug  builds given:
https://github.com/openxla/xla/blob/9db19f8005c83cf594d2d65baa14169a60358a49/xla/backends/cpu/runtime/work_queue.h#L246

Will add a short circuit for this particular case to avoid the allocation and mitigate against UB.

xla/backends/cpu/runtime/work_queue_test.cc
seantalts: Ah, I guess we're not done with the migration yet. Thanks.

seantalts (approved):

akuegel (approved):
Approving again for reimport.

akuegel:
@zdryan Can you please rebase, it seems our import does not work anymore with this old version of head.

zdryan:
> @zdryan Can you please rebase, it seems our import does not work anymore with this old version of head.

Done - now rebased @akuegel.

akuegel (approved):

zdryan:
Reproduced the gemma2 regression locally - was a consequence of unconditional partition cap dropping `num_workers` below the value `kernel.cc` had already sized. Left idle cores for small/medium kernels.

Amended change and refreshed benchmark numbers in commit message.

| Items | Before | After | Diff |
  |------:|---------:|----:|------:|
  | 16    |   16,565 | 5,537 | -67% |
  | 32    |   20,840 | 7,785 | -63% |
  | 64    |   28,386 | 11,588 | -59% |
  | 128   |   41,239 | 16,002 | -61% |
  | 256   |   59,833 | 22,647 | -62% |
  | 512   |   93,335 | 76,402 | -18% |
  | 1024  |  171,905 | 83,917 | -51% |

*`num_workers` = 336 similarly in this table*.

CC @seantalts @akuegel

xla/backends/cpu/runtime/work_queue.h
```diff
@@ -345,15 +351,32 @@ ABSL_ATTRIBUTE_ALWAYS_INLINE tsl::AsyncValueRef<tsl::Chain> Worker::Parallelize(
...
-  // idle or cause out-of-bounds partition access.
-  num_workers = std::min(num_work_items, num_workers);
+
+  // Only pack items into fewer partitions when the caller oversupplied workers
+  // otherwise num_workers is already the right cap and reducing it leaves cores
+  // idle.
+  size_t num_partitions;
+  if (num_workers > num_work_items) {
+    constexpr size_t kMinItemsPerPartition = 8;
+    num_partitions = CeilOfRatio(num_work_items, kMinItemsPerPartition);
```
akuegel, line 361: Right, it should be `CeilOfRatio` and not `FloorOfRatio`. I tried to address the review comment of Eugene https://github.com/openxla/xla/pull/42683#discussion_r3260958277
But then took FloorOfRatio without thinking.

akuegel (approved):

PR #42627, opened 2026-05-14
[ROCm] xla/pjrt: honor XLA_PYTHON_CLIENT_ALLOCATOR=platform

magaonka-amd:
📝 Summary of Changes
Make `GpuAllocatorConfig::Kind::kPlatform` actually deliver a synchronous passthrough allocator instead of returning nullptr. Previously the env var XLA_PYTHON_CLIENT_ALLOCATOR=platform was silently overridden by the BFC allocator.

🎯 Justification
Lets take below simple python code 
```bash

import jax, jax.numpy as jnp

dev = jax.devices()[0]
print("before:", dev.memory_stats())

x = jnp.ones(1 << 30, dtype=jnp.float32).block_until_ready()  # 4 GiB
print("after :", dev.memory_stats())
```
when you run this 
```bash
XLA_PYTHON_CLIENT_ALLOCATOR=platform \
  XLA_PYTHON_CLIENT_PREALLOCATE=false \
  TF_CPP_MIN_LOG_LEVEL=0 \
  TF_CPP_VMODULE=bfc_allocator=1,se_gpu_pjrt_client=1 \
    python3.12 /workspace/scripts/pr856_repro/prove_bfc_grabs_huge.py 2>&1 \
    | grep -E "XLA_backend_|GPU_[0-9]+_bfc|Using platform allocator|memory_stats|VRAM"
I0514 16:45:00.591301 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_0_bfc
I0514 16:45:00.601878 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_1_bfc
I0514 16:45:00.612506 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_2_bfc
I0514 16:45:00.623345 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_3_bfc
I0514 16:45:00.634821 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_4_bfc
I0514 16:45:00.646178 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_5_bfc
I0514 16:45:00.657589 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_6_bfc
I0514 16:45:00.668910 1890407 bfc_allocator.cc:91] Creating new BFCAllocator named: XLA_backend_7_bfc
I0514 16:45:01.647998 1890407 se_gpu_pjrt_client.cc:1536] Using platform allocator.
I0514 16:45:01.697490 1890407 bfc_allocator.cc:182] Extending allocation by 2.00MiB bytes for XLA_backend_0_bfc.
I0514 16:45:01.759927 1890407 bfc_allocator.cc:182] Extending allocation by 4.00GiB bytes for XLA_backend_0_bfc.
```
you can see BFC allocation is growing despite user explicitly selecting platform allocator.
first few Creating new BFCAllocator make sesne due to how `Backend::Backend()` constructs them unconditionally.

with the fix we will not have BFC allocation silently getting picked up like below.
```bash
I0514 16:45:01.697490 1890407 bfc_allocator.cc:182] Extending allocation by 2.00MiB bytes for XLA_backend_0_bfc.
I0514 16:45:01.759927 1890407 bfc_allocator.cc:182] Extending allocation by 4.00GiB bytes for XLA_backend_0_bfc.
```
🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
Adds a regression test PlatformAllocatorIsSynchronousPassthrough that asserts the resulting client's allocator is a StreamExecutorAddressAllocator and not a MultiDeviceAdapter (which would indicate the silent BFC fallback).


magaonka-amd:
addressed one DWYU comment but rest look like DWYU couldn't figure out conditional dependency.

reichlfl (approved):
@pschuh to weigh in if you have any concerns about this, but this looks reasonable to me.

pschuh:
Who is creating the bfc allocator? I thought it was supposed to fallback to Backend::memory_allocator() and skip the bfc allocator if it was nullptr (from GpuAllocatorConfig::Kind::kPlatform)

magaonka-amd:
> Who is creating the bfc allocator? I thought it was supposed to fallback to Backend::memory_allocator() and skip the bfc allocator if it was nullptr (from GpuAllocatorConfig::Kind::kPlatform)

IIUC https://github.com/openxla/xla/blob/main/xla/service/backend.cc#L250 here BFC is built.

pschuh:
Oh wow, broken recently: https://github.com/openxla/xla/pull/41761 thanks for the fix. That original change that broke things seems risky because now we have two BFC allocators floating around...

magaonka-amd:
ummm, oh yeah good point. what do you prefer
- I rework this PR to fix the BFC allocator getting built twice
- or I can follow up with another PR to address the BFC allocator problem?
I need to read bit more about it to make that fix.

right now this PR helps jax unit tests for rocm because we rely on platform allocator for pytests.

pschuh:
I think its just a mess, but I feel like platform has a clear meaning that it shares the allocator with the LocalClient for old code which may reuse the LocalClient between pjrt and a direct user of LocalClient. We probably should have a separate option for constructing a StreamExecutorAddressAllocator and ignoring the BFC allocator below at the pjrt level. Its feeling like we should just allow the allocator to be configured process-wide and avoid having two allocators, but I don't think we want to solve that in this CL.

magaonka-amd:
@pschuh if this fix is okay with you can we get this merged? if anything needs to be done from my side please do let me know.

ezhulenev (approved):

pschuh (approved):

magaonka-amd:
hey folks looks like copybara is not progressing much on this PR , is there anything I need to do?

thcmbs:
Hi folks, i'm looking into merging the PR. However we have some seemingly real internal test failures, so it can take a bit of time to debug. 

One of the failures is from https://github.com/jax-ml/jax/blob/main/jax/experimental/jax2tf/tests/primitives_test.py

```
0604 03:46:03.138700    8508 cuda_context.cc:190] Check failed: cuda::ToStatus(result, "Failed setting context") is OK (INTERNAL: CUDA error: Failed setting context: CUDA_ERROR_NOT_PERMITTED: operation not permitted
[third_party/tensorflow/compiler/xla/stream_executor/cuda/cuda_status.cc:50](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/stream_executor/cuda/cuda_status.cc?l=50&ws=tap-presubmit-server/616880884&snapshot=2)
) 
E0604 03:46:03.727037    8508 process_state.cc:708] RAW: Raising signal 6 with default behavior
I0604 03:46:03.727043    8508 process_state.cc:1342] RAW: FailureSignalHandler() exiting
/build/work/aff014c224a5fefc4109b4e7522f1e34f838/google3/tools/test/remote-runtest.sh: line 256:  7981 Aborted                 "$@"
-- 2026-06-04 03:46:05 PDT Forge runner: Test killed by SIGABRT while running on royx4.prod.google.com
```

pschuh:
Ah, that is what I was concerned about above. I think we need a different solution. platform means that it shares the allocator with tf which is why the jax2tf tests are failing. We cannot change the meaning of platform and either need to allow configuring the underlying allocator to be the address allocator or a separate option for pjrt which changes the allocator.

magaonka-amd:
Hmmm.  I need to go back and rework then. thanks for the insight. let me see how to fix this.

magaonka-amd:
Im closing this PR , I took suggestion from reviewers and implemented https://github.com/openxla/xla/pull/44335
and it is merged , thanks everyone.

PR #42710, opened 2026-05-15
[ROCm] Enable hlo-opt lit test suite on gfx1250

nurmukhametov:
Adds a gfx1250 target spec to xla/backends/gpu/target_config/specs and wires it into the hlo_lit_tests suite in xla/backends/gpu/tests/BUILD. Generalizes the existing mi200 special-case in xla/lit.bzl into a small rocm_gpus list so rocm-only tag apply to any ROCm target, not just mi200.

Add hlo lit test for Triton TDM intrinsic lowering xla/backends/gpu/tests/triton_tdm_elemwise.hlo


xla/backends/gpu/tests/triton_tdm_elemwise.hlo
```diff
@@ -0,0 +1,47 @@
+// RUN: %if IS_ROCM %{ hlo-opt %s --platform=gpu --stage=llvm-before-optimizations --xla_gpu_target_config_filename=%S/../target_config/specs/%{GPU}.txtpb | FileCheck %s --check-prefixes=CHECK-%{GPU} %}
```
akuegel: I think you can do it in a single line without the %if IS_ROCM condition. `--check-prefixes=CHECK-%{GPU},CHECK-%{PTX}`
And use CHECK-PTX in the lines where you currently use CHECK-CUDA

nurmukhametov: Thanks for suggesting this, though it required me to add `--allow-unused-prefixes`. I hope that's OK.

akuegel (approved):

PR #42740, opened 2026-05-17
[XLA:GPU] Fix async collective kernel command buffer's empty capture error

shawnwang18:
📝 Summary of Changes
  Fixes the async collective-kernel path in AllReduceThunk::RunCollective() so command-buffer recording launches the collective kernel on the stream passed by Record() tracing.

  - Added an explicit-stream CollectiveKernelThunk::ExecuteOnStream(params, stream) overload.
  - Updated AllReduceThunk::RunCollective() to use the resolved stream for collective-kernel support checks and execution.
  - Added command-buffer create/update coverage for the async CollectiveKernelThunk path in all_reduce_thunk_test.cc.
  - Updated BUILD deps for the new test coverage.

  🎯 Justification
  When AllReduceThunk::RunCollective() is reached through Record(), the traced graph must capture work launched on the stream provided by TraceCommandBufferFactory. The previous async collective-kernel path redirected execution to the async collective stream, leaving the traced command buffer empty.

  🚀 Kind of Contribution
  🐛 Bug Fix, 🧪 Tests

  N/A. This is a correctness fix for command-buffer capture behavior.

  🧪 Unit Tests:
  Added AllReduceThunkTest.RecordCommandBufferCreateUpdateAsyncCollectiveKernel, covering command-buffer create/update for the async collective-kernel path.

  🧪 Execution Tests:
  Ran git diff --check: passed.

  Attempted:

  bazel build //xla/backends/gpu/runtime:all_reduce_thunk_test
  bazel test //xla/backends/gpu/runtime:all_reduce_thunk_test --test_filter=AllReduceThunkTest.RecordCommandBufferCreateUpdateAsyncCollectiveKernel


ezhulenev (approved):

shawnwang18:
@golechwierowicz Could you take look at this PR?

xla/backends/gpu/runtime/collective_kernel_thunk.cc
```diff
@@ -429,11 +429,24 @@ absl::Status CollectiveKernelThunk::Initialize(const InitializeParams& params) {
 
 absl::Status CollectiveKernelThunk::ExecuteOnStream(
     const ExecuteParams& params) {
-  se::Stream* stream = params.stream;
   if (is_async_) {
-    stream = params.collective_params->async_streams.at(
-        static_cast<int64_t>(AsyncStreamKind::ASYNC_STREAM_KIND_COLLECTIVE));
+    return ExecuteOnStream(
```
sohaibiftikhar: A better way imho would be to simply lift the `is_async` check out into the all_reduce thunk and use `WithComputeStream` from there.

No branching logic would exist in this file then for async streams. Which makes sense since the task of this class is mostly to prepare buffers/request cliques etc.

Also no override or Impl methods would be needed.

sohaibiftikhar (approved):

xla/backends/gpu/runtime/collective_kernel_thunk.cc
shawnwang18: Thanks for the suggestion, your pattern is better, I changed to your design.

sohaibiftikhar (approved):

golechwierowicz:
Looks like there's some test failing.

https://github.com/tensorflow/tensorflow/actions/runs/26097050432/job/76737488639

shawnwang18:
> Looks like there's some test failing.
> 
> https://github.com/tensorflow/tensorflow/actions/runs/26097050432/job/76737488639

It seems that the failure is un-related with this PR, I do not see any cause from my PR. 

 Error:

  tensorflow.python.framework.errors_impl.InvalidArgumentError:
  Default MaxPoolingGradOp only supports NHWC on device type CPU
  [[{{node MaxPoolGrad_1}}]]

PR #42779, opened 2026-05-18
Increase timeout for dot_operation_test

hsharsha:
📝 Summary of Changes

🎯 Justification
change.

🚀 Kind of Contribution

🧪 Unit Tests:
transformed HLO.

🧪 Execution Tests:

### DIFF
--- a/xla/tests/BUILD
+++ b/xla/tests/BUILD
@@ -1095,6 +1095,7 @@ xla_test(
 xla_test(
     name = "dot_operation_test",
     srcs = ["dot_operation_test.cc"],
+    timeout = "long",
     aot_compile_test = False,
     backend_tags = {
         # TODO(b/445172709): Re-enable once fixed.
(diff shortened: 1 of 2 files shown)


xla/tests/BUILD
```diff
@@ -1095,6 +1095,7 @@ xla_test(
 xla_test(
     name = "dot_operation_test",
     srcs = ["dot_operation_test.cc"],
+    timeout = "long",
```
akuegel, line 1098: We prefer not to have any tests with timeout long. Can you determine which of the tests takes too long, and possibly disable it on ROCM? Or is it just the amount of tests here that causes the timeout?

akuegel (commented):
Sorry for late reply, forgot to hit "Submit review".

xla/tests/BUILD
akuegel: @hsharsha Is this still failing on ROCM side?

hsharsha: @akuegel It still timesout with 300s limit on ROCm, It is taking around 400-500s to run this test. We are still debugging. I will keep this as draft for now. If it helps I can close it and reopen a new PR

akuegel: Thanks for the reply, keeping this as a draft sounds good. If really needed, I guess we can make it timeout long with a ROCM guard.

dimitar-asenov:
Marking as draft per the last comment.

PR #40889, opened 2026-04-14
Migrate XLA to Bazel 8.6.0 (WORKSPACE based)

meteorcloudy:
Updates include:
- .bazelversion to 8.6.0
- .bazelrc to enable workspace and disable bzlmod
- rules_cc and rules_proto patches for Bazel 8 compatibility
- python toolchain template fixes
- def_file_filter fixes to use rules_cc helpers

PR #42789, opened 2026-05-18
Migrate XLA to Bazel 8.7.0 (Workspace based)

cerisier:
Quite an important change but hopefully we can finally get xla to bazel 8 and hopefully even bzlmod soon 

## Summary

- Applies the Bazel 8 migration commit from PR #40889 onto current main.
- And various related migration changes.
- Updates the migration to Bazel 8.7.0.
- Updates rules_ml_toolchain to HEAD now that https://github.com/google-ml-infra/rules_ml_toolchain/pull/261 is merged
- Keeps lower Bzlmod dependency versions from the Bazel 8 migration where they still work with the current module graph.


cerisier:
Cc @meteorcloudy and @beckerhe

beckerhe (commented):
Thanks a lot for this change. This looks good to me, but I'm not a Bazel expert. @meteorcloudy do you have any concerns?

WORKSPACE
```diff
@@ -16,6 +16,9 @@ load(":workspace3.bzl", "xla_workspace3")
 
 xla_workspace3()
 
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
```
meteorcloudy, line 19: Note that downstream projects like JAX and TF have to also add this to WORKSPACE when upgrading XLA to this version.

meteorcloudy (approved):

meteorcloudy:
Looks good, but historically, we do bazel version bump for XLA, JAX, TF at the same time since they are somewhat coupled. @beckerhe Not sure if the XLA team is fine with doing this only for XLA?

cerisier:
> Looks good, but historically, we do bazel version bump for XLA, JAX, TF at the same time since they are somewhat coupled. @beckerhe Not sure if the XLA team is fine with doing this only for XLA?

If needed, I can look in doing the same port on both TF and JAX. Let me know!
Bazel 8 support for XLA would be a big move towards ease of contribution and integration!

WORKSPACE
akuegel: If that is the only thing needed for TF and JAX, I think we can take care of that while merging the PR.

akuegel:
> Looks good, but historically, we do bazel version bump for XLA, JAX, TF at the same time since they are somewhat coupled. @beckerhe Not sure if the XLA team is fine with doing this only for XLA?

According to your experience, does it seem possible to do it for XLA first (with the small adjustment to TF and JAX that you mentioned)? I think as long as it works, doing the version bump for XLA first should be ok from XLA standpoint.

workspace2.bzl
```diff
@@ -261,29 +262,13 @@ def _tf_repositories():
         urls = tf_mirror_urls("https://pypi.python.org/packages/source/s/six/six-1.16.0.tar.gz"),
     )
 
-    filegroup_external(
-        name = "astunparse_license",
-        licenses = ["notice"],  # PSFL
```
akuegel, line 268: Why is this removed?

workspace2.bzl
```diff
@@ -261,29 +262,13 @@ def _tf_repositories():
...
     tf_http_archive(
         name = "absl_py",
         sha256 = "8a3d0830e4eb4f66c4fa907c06edf6ce1c719ced811a12e26d9d3162f8471758",
         strip_prefix = "abseil-py-2.1.0",
         urls = tf_mirror_urls("https://github.com/abseil/abseil-py/archive/refs/tags/v2.1.0.tar.gz"),
     )
 
-    filegroup_external(
-        name = "org_python_license",
-        licenses = ["notice"],  # Python 2.0
```
akuegel, line 283: Why is this removed?

meteorcloudy:
> According to your experience, does it seem possible to do it for XLA first (with the small adjustment to TF and JAX that you mentioned)? I think as long as it works, doing the version bump for XLA first should be ok from XLA standpoint.

Yes, I think it's possible. XLA also has downstream tests against TF and JAX.

workspace2.bzl
cerisier: This was taken from @meteorcloudy commit

workspace2.bzl
cerisier: Same

cerisier:
Any reason why Copybara dropped some of the changes ? It's red because most `thid_party` changes are dropped and still use old incompatible loads.

akuegel:
> Any reason why Copybara dropped some of the changes ? It's red because most `thid_party` changes are dropped and still use old incompatible loads.

I think it is related to the setup with tsl. It is complicated. We will have to re-add those changes manually internally.

cerisier:
Understood, let me know if anything needs to be done on my end to help

third_party/rules_proto.patch
```diff
@@ -0,0 +1,13 @@
...
+ """Lovely workaround to be able to expose native constants pretending to be Starlark."""
+
++load("@com_google_protobuf//bazel/common:proto_info.bzl", "ProtoInfo")
++
+ # buildifier: disable=native-proto
+ NativeProtoInfo = ProtoInfo
+
+ # buildifier: disable=native-proto
+-native_proto_common = proto_common
++native_proto_common = struct()
```
akuegel: Setting native_proto_common to an empty struct will break any rules that call methods on it (like compile). If native.proto_common is gone in Bazel 8, you should load the Starlark replacement from @com_google_protobuf or @rules_proto instead of using an empty struct.

workspace1.bzl
```diff
@@ -1,8 +1,8 @@
 """TensorFlow workspace initialization. Consult the WORKSPACE on how to use it."""
 
 load("@com_github_grpc_grpc//bazel:grpc_deps.bzl", "grpc_deps")
-load("@io_bazel_rules_closure//closure:defs.bzl", "closure_repositories")
 load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
+load("@rules_cc//cc:extensions.bzl", "compatibility_proxy_repo")
```
akuegel, line 5: The PR description says rules_cc 0.2.0 is used to avoid the need for the compatibility proxy repository. However, this line adds a load and call for that exact proxy.

Moreover, rules_cc 0.2.0 likely doesn't have cc/extensions.bzl, so this might break WORKSPACE builds.

akuegel (commented):
Internal AI reviewer flagged some issues, these two are possibly relevant (though I lack the bazel knowledge to be 100% sure).

workspace1.bzl
cerisier: PR description doesn't say that ? also rules_cc 0.2.0 does have cc/extensions.bzl so I think it's misguided.

third_party/rules_proto.patch
cerisier: This came from @meteorcloudy commit. My understanding is that this is done as a transient state while update to bzlmod and higher versions of rulesets. native_proto_common doesn't seem to be used transitively at all. I can try loading the symbol from @protobuf but I'm not sure it's really required here.

cerisier:
Let's see if this is enough for internal review ?

cerisier:
@akuegel @beckerhe can we retry internal checks ? Would be really great to make progress on this.

cerisier:
Gentle ping ?

WORKSPACE
Wyverald: if this is needed for downstream projects too, why don't we just put it in workspace3.bzl or workspace4.bzl?

cerisier: We could. However, someone already ported that to downstream.

Wyverald:
Hey @cerisier, while importing the team ran into TensorFlow CI failures (see https://github.com/tensorflow/tensorflow/pull/120927). Could you take a look over there and see if you can figure out the errors?

cerisier:
@Wyverald sure I'll take a look!

cerisier:
@Wyverald so the failure is:

```While resolving toolchains for target @@rules_cc//cc/private/toolchain:compiler (b6e707a): No matching toolchains found for types @@bazel_tools//tools/cpp:toolchain_type.```

This happens because of 2 reasons:
1. Tensorflow still uses `cc_toolchain_suite` and `crosstool_top`
2. `compiler_flag` in rules_cc@0.2.0 (which we now use) now requires a C++ toolchain (`use_cc_toolchain` now has `mandatory=True` by default)

Before:
- tensorflow provides a `cc_toolchain_suite`
- `compiler_flag` from prior `rules_cc` version has `use_cc_toolchain mandatory=False`)
- Toolchain gets picked up because `find_cpp_toolchain` has a legacy cc toolchain attr fallback

Now:
- Tensorflow provides a `cc_toolchain_suite`
- `compiler_flag` from `rules_cc@0.2.0` has `use_cc_toolchain(mandatory=True)`
- Fails during toolchain resolution beause Tensorflow doesn't provide any platform cc toolchain

I see 2 solutions here:
- Use a patch when loading rules_cc in XLA or in Tensorflow to revert `use_cc_toolchain(mandatory=True)` to `False`
- Migrate Tensorflow out of crosstool_top but that seems quite involved

Do you see any other solution ?

Also since my next step is to work on bzlmod support for XLA, we'll face a different issue that'll need disussion around patch management if we go this way.

Wyverald:
> Use a patch when loading rules_cc in XLA or in Tensorflow to revert `use_cc_toolchain(mandatory=True)` to `False`

Let's do this for now, I think. Since it looks like you have all of XLA building fine with rules_cc 0.2.0, maybe we only need to do this patching in TensorFlow? I'm not too sure how that would work, so if that's too hard, let's just patch it out for XLA too for now.

cerisier:
Actually tensorflow had a cross compile platform cc toolchain somewhere so i'm trying registering them on the cross compile config there so that we don't have to patch anything here.

See https://github.com/tensorflow/tensorflow/pull/121302

akuegel:
> Actually tensorflow had a cross compile platform cc toolchain somewhere so i'm trying registering them on the cross compile config there so that we don't have to patch anything here.
> 
> See [tensorflow/tensorflow#121302](https://github.com/tensorflow/tensorflow/pull/121302)

So this has been merged, but it seems we still hit the same error on presubmit: https://github.com/tensorflow/tensorflow/actions/runs/27812239258/job/82305044577

cerisier:
@akuegel I sent https://github.com/tensorflow/tensorflow/pull/121612

I had this unstaged but somehow mistakenly thought it wasn't needed since the original failure was aarch64 related...
Sorry about that. Hopefully we can get it merged quickly.

cerisier:
@akuegel so the new tensorflow PR commits have been merged already. Can we retry internal checks ?

PR #42808, opened 2026-05-19
[GPU] Eliminate redundant DUS-accumulator initialization

terryysun:
📝 Summary of Changes
Added new pass `dus_accumulator_zero_init_elimination`: When a `while` loop's body writes every read slot of its accumulator buffer via `dynamic-update-slice`, the pre-loop init is dead. If the init is a `broadcast` of a scalar constant, the pass replaces it with `AllocateBuffer`, saving a buffer-wide write per loop entry. Soundness is guarded by dead-input check.  Gated by `--xla_gpu_enable_dus_accumulator_zero_init_elimination` (default off).

🎯 Justification
Pre-zeroing then fully overwriting an accumulator via DUS is a common pattern. The pre-fill is pure overhead.

🚀 Kind of Contribution
✨ New Feature

N/A

🧪 Unit Tests:
`dus_accumulator_zero_init_elimination_test.cc` runs the pass at HLO level over three scaffolds (raw, fusion-wrapped DUS, fusion-wrapped init), a multi-accumulator case, and a negative case that doesn't fully cover its read slots.   

🧪 Execution Tests:
`dus_accumulator_zero_init_elimination_execution_test.cc` uses `RunAndCompareTwoModules` to run the same HLO with the flag off and on, asserting bitwise-identical outputs. Two scenarios: single-accumulator loop and two-accumulator loop.


xla/service/while_loop_unroller.cc
```diff
@@ -415,18 +415,10 @@ bool IsLoopInductionVar(const HloInstruction* instr,
 
 // Recursively checks if the given instruction inside a while loop can be
 // expressed as a value range, possibly depending on the loop induction variable
-// of that while loop.
+// of that while loop. RecursivelyIdentifyRange follows fusion-Parameter edges
```
akuegel, line 418: I think this part of the PR could be extracted into its own PR?

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+namespace gpu {
+namespace {
+
+bool IsBroadcastOfScalarConstant(const HloInstruction* h) {
+  while (h->opcode() == HloOpcode::kBitcast ||
+         h->opcode() == HloOpcode::kCopy ||
+         h->opcode() == HloOpcode::kReshape) {
+    h = h->operand(0);
+  }
+  if (h->opcode() != HloOpcode::kBroadcast) return false;
```
akuegel: Please use { } after `if`.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+#include "xla/service/gpu/ir_emission_utils.h"
+#include "xla/service/value_range.h"
+#include "xla/service/while_loop_unroller.h"
+#include "xla/shape.h"
+#include "xla/shape_util.h"
+#include "xla/tsl/platform/errors.h"
+#include "xla/tsl/platform/statusor.h"
+#include "xla/xla.pb.h"
+
+namespace xla {
```
akuegel: nit: use `namespace xla::gpu` instead of nested namespaces

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+
+bool IsBroadcastOfScalarConstant(const HloInstruction* h) {
+  while (h->opcode() == HloOpcode::kBitcast ||
+         h->opcode() == HloOpcode::kCopy ||
+         h->opcode() == HloOpcode::kReshape) {
+    h = h->operand(0);
+  }
+  if (h->opcode() != HloOpcode::kBroadcast) return false;
+  const HloInstruction* src = hlo_query::StripCastLike(h->operand(0));
+  return src->opcode() == HloOpcode::kConstant &&
```
akuegel: Please use `hlo_query::IsScalarConstant()`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    return std::nullopt;
+  }
+  return r;
+}
+
+// Generalizes AdvancedMatchShapeCoveringDynamicIndexInstruction to indices
+// that are arithmetic expressions over fused parameters (e.g. descending DUS).
+bool IsKillingDus(const HloInstruction* dus, const HloInstruction* operand,
+                  const WhileLoopConfig& cfg) {
+  if (dus->opcode() != HloOpcode::kDynamicUpdateSlice) return false;
```
akuegel: Please use {} around `return false;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  if (dus->operand(0) != operand) return false;
```
akuegel: same here or maybe even combine it into one if:

```
if (dus->opcode() != HloOpcode::kDynamicUpdateSlice || dus->operand(0) != operand) {
  return false;
}
```

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    const HloInstruction* idx = dus->operand(2 + d);
+    std::optional<Range> r = ResolveIndexRangeAsFunctionOfIv(idx, cfg);
+    if (r.has_value() && r->IsBounded() && r->IsSingleValue() &&
+        r->min().GetSignedValue() == 0) {
+      if (slice_shape.dimensions(d) != input_shape.dimensions(d)) {
+        return false;
+      }
+      continue;
+    }
+    if (dyn_dim_opt.has_value()) return false;  // more than one dynamic dim
```
akuegel: Please use {} around `return false`. Also combine with if below.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    if (!r.has_value()) return false;
+    dyn_dim_opt = d;
+    dyn_dim_range = r;
+  }
+  if (!dyn_dim_opt.has_value() || !dyn_dim_range.has_value()) return false;
```
akuegel: Please use {} around `return false`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  int64_t dyn_dim = *dyn_dim_opt;
+  std::optional<Range> idx_range = dyn_dim_range;
+
+  const int64_t dim_size = input_shape.dimensions(dyn_dim);
+  const int64_t slice_size = slice_shape.dimensions(dyn_dim);
+  if (slice_size <= 0 || slice_size > dim_size) return false;
```
akuegel: Please use {} around `return false;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  std::vector<bool> covered(dim_size, false);
```
akuegel: This looks like a naive way to check whether the whole dimension is covered.
I believe there is a better algorithm for this. `min()` should be <= 0, `max()` should be >= `dim_size - 1`, and `slice_size >= step()`. No need to iterate over the whole range.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  }
+  return true;
+}
+
+// TODO: scatter killing-write support; needs reasoning over scatter index
+// expressions and update_window_dims.
+
+absl::StatusOr<bool> FusionParamIsKilled(const HloInstruction* fusion,
+                                         int64_t param_idx, WalkCtx& ctx) {
+  if (++ctx.depth > WalkCtx::kMaxDepth) return false;
```
akuegel: Please use {} around `return false;`

Also, is it intentional that depth is not decreased here when we return false?

Alternatively, like mentioned below, I would prefer it like this:

```
if (ctx.depth >= WalkCtx::kMaxDepth) {
  return false;
}
++ctx.depth;
```

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  HloComputation* fc = fusion->fused_instructions_computation();
+  HloInstruction* fused_param = fc->parameter_instruction(param_idx);
+  bool ok = false;
+  TF_ASSIGN_OR_RETURN(ok, AllUsersKillBuffer(fused_param, ctx));
```
akuegel: We prefer ASSIGN_OR_RETURN now (no TF prefix).

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      // each receiving branch's parameter_0 must be killed.
+      bool every_branch_kills = true;
+      bool flows_into_some_branch = false;
+      for (int b = 0; b < user->branch_count(); ++b) {
+        const HloInstruction* branch_input = user->operand(b + 1);
+        if (branch_input != operand) continue;
+        flows_into_some_branch = true;
+        HloComputation* branch = user->branch_computation(b);
+        HloInstruction* branch_param = branch->parameter_instruction(0);
+        TF_ASSIGN_OR_RETURN(bool branch_ok,
```
akuegel: TF_ASSIGN_OR_RETURN -> ASSIGN_OR_RETURN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    if (u->opcode() == HloOpcode::kGetTupleElement &&
+        u->tuple_index() == slot) {
+      if (slot_gte != nullptr) return false;  // multiple GTEs at slot
+      slot_gte = u;
+    }
+  }
+  if (slot_gte == nullptr) return true;  // never read; trivially dead
+
+  WalkCtx ctx{cfg};
+  TF_ASSIGN_OR_RETURN(bool all_kill, AllUsersKillBuffer(slot_gte, ctx));
```
akuegel: TF_ASSIGN_OR_RETURN -> ASSIGN_OR_RETURN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      HloInstruction* init = init_tuple->mutable_operand(slot);
+      std::optional<Shape> alloc_shape_opt = ClassifyReplaceableInit(init);
+      if (!alloc_shape_opt.has_value()) continue;
+      if (init->has_sharding()) continue;
+      if (init->user_count() != 1) continue;  // init must feed only this while
+      const Shape& alloc_shape = *alloc_shape_opt;
+      if (alloc_shape.dimensions_size() == 0) continue;
+      if (alloc_shape.dimensions(0) != cfg.trip_count) continue;
+
+      TF_ASSIGN_OR_RETURN(bool dead, SlotIsDeadInput(slot, cfg));
```
akuegel: TF_ASSIGN_OR_RETURN -> ASSIGN_OR_RETURN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      if (!dead) continue;
+
+      HloInstruction* alloc =
+          while_op->parent()->AddInstruction(HloInstruction::CreateCustomCall(
+              alloc_shape, /*operands=*/{}, "AllocateBuffer"));
+      alloc->set_metadata(init->metadata());
+      alloc->set_frontend_attributes(init->frontend_attributes());
+      alloc->set_statistics_viz(init->statistics_viz());
+      TF_RETURN_IF_ERROR(init_tuple->ReplaceOperandWith(slot, alloc));
```
akuegel: TF_RETURN_IF_ERROR -> RETURN_IF_ERROR

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  TF_ASSIGN_OR_RETURN(ok, AllUsersKillBuffer(fused_param, ctx));
+  --ctx.depth;
+  return ok;
+}
+
+absl::StatusOr<bool> AllUsersKillBuffer(const HloInstruction* buffer,
+                                        WalkCtx& ctx) {
+  if (buffer->user_count() == 0) return true;  // trivially dead
+  for (const HloInstruction* user : buffer->users()) {
+    TF_ASSIGN_OR_RETURN(bool ok, UserKillsBuffer(user, buffer, ctx));
```
akuegel: TF_ASSIGN_OR_RETURN -> ASSIGN_OR_RETURN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+};
+
+class DusRawDusParam : public DusAccumulatorZeroInitEliminationTest,
+                       public ::testing::WithParamInterface<RawDusCase> {};
+
+TEST_P(DusRawDusParam, ParameterisedScan) {
+  const RawDusCase& c = GetParam();
+  SCOPED_TRACE(c.name);
+  const std::string hlo = MakeRawDusHlo(c);
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo));
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+  EnableFlag(module.get());
+  TF_ASSERT_OK_AND_ASSIGN(
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+
+class DusFusionDusParam : public DusAccumulatorZeroInitEliminationTest,
+                          public ::testing::WithParamInterface<FusionDusCase> {
+};
+
+TEST_P(DusFusionDusParam, ParameterisedScan) {
+  const FusionDusCase& c = GetParam();
+  SCOPED_TRACE(c.name);
+  const std::string hlo = MakeFusionDusHlo(c);
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo));
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+  EnableFlag(module.get());
+  TF_ASSERT_OK_AND_ASSIGN(
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+
+class DusFusionInitParam
+    : public DusAccumulatorZeroInitEliminationTest,
+      public ::testing::WithParamInterface<FusionInitCase> {};
+
+TEST_P(DusFusionInitParam, ParameterisedScan) {
+  const FusionInitCase& c = GetParam();
+  SCOPED_TRACE(c.name);
+  const std::string hlo = MakeFusionInitHlo(c);
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo));
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+  EnableFlag(module.get());
+  TF_ASSERT_OK_AND_ASSIGN(
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+  zi1 = bf16[4,16] broadcast(c0), dimensions={}
+  init = (s32[], bf16[4,8], bf16[4,16]) tuple(z, zero_init, zi1)
+  w = (s32[], bf16[4,8], bf16[4,16]) while(init), condition=cond, body=body,
+      backend_config={"known_trip_count":{"n":"4"},
+                      "known_init_step":{"init":"0","step":"1"},
+                      "known_induction_variable":{"tuple_index":"0"}}
+  ROOT r = bf16[4,8] get-tuple-element(w), index=1
+}
+)";
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(kHlo));
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+  EnableFlag(module.get());
+  TF_ASSERT_OK_AND_ASSIGN(
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+  zero_init = bf16[8,8] broadcast(c0), dimensions={}
+  init = (s32[], bf16[8,8]) tuple(z, zero_init)
+  w = (s32[], bf16[8,8]) while(init), condition=cond, body=body,
+      backend_config={"known_trip_count":{"n":"4"},
+                      "known_init_step":{"init":"0","step":"1"},
+                      "known_induction_variable":{"tuple_index":"0"}}
+  ROOT r = bf16[8,8] get-tuple-element(w), index=1
+}
+)";
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(kHlo));
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+  EnableFlag(module.get());
+  TF_ASSERT_OK_AND_ASSIGN(
```
akuegel: TF_ASSERT_OK_AND_ASSIGN -> ASSERT_OK_AND_ASSIGN

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    if (!ok) return false;
```
akuegel: Please use {} around `return false;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  HloInstruction* fused_param = fc->parameter_instruction(param_idx);
+  bool ok = false;
+  TF_ASSIGN_OR_RETURN(ok, AllUsersKillBuffer(fused_param, ctx));
+  --ctx.depth;
+  return ok;
+}
+
+absl::StatusOr<bool> AllUsersKillBuffer(const HloInstruction* buffer,
+                                        WalkCtx& ctx) {
+  if (buffer->user_count() == 0) return true;  // trivially dead
```
akuegel: Please use {} around `return true;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      }
+    }
+  }
+  return std::nullopt;
+}
+
+struct WalkCtx {
+  const WhileLoopConfig& cfg;
+  int depth = 0;
+  static constexpr int kMaxDepth = 8;
```
akuegel: I think it would help to explain why we only allow MaxDepth of 8 here? How was it chosen?

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    TF_ASSIGN_OR_RETURN(bool ok, UserKillsBuffer(user, buffer, ctx));
+    if (!ok) return false;
+  }
+  return true;
+}
+
+absl::StatusOr<bool> UserKillsBuffer(const HloInstruction* user,
+                                     const HloInstruction* operand,
+                                     WalkCtx& ctx) {
+  if (++ctx.depth > WalkCtx::kMaxDepth) {
```
akuegel: What about:

```
if (ctx.depth >= WalkCtx::kMaxDepth) {
  return false;
}
++ctx.depth;
```

This avoids increase + decrease when we have reached maximum depth.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      break;
+
+    case HloOpcode::kFusion: {
+      int64_t op_idx = user->operand_index(operand);
+      if (op_idx < 0) {
+        result = false;
+        break;
+      }
+      result = FusionParamIsKilled(user, op_idx, ctx).value_or(false);
+      if (result) break;
```
akuegel: Please use {} around `break;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+
+      // DSF: in-place semantics live in output_to_operand_aliasing, not in
+      // the fused body — walk the alias map instead.
+      if (user->fusion_kind() != HloInstruction::FusionKind::kCustom) break;
+      auto bcfg = user->backend_config<GpuBackendConfig>();
+      if (!bcfg.ok()) break;
```
akuegel: Please use {} around `break;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      if (op_idx < 0) {
+        result = false;
+        break;
+      }
+      result = FusionParamIsKilled(user, op_idx, ctx).value_or(false);
+      if (result) break;
+
+      // DSF: in-place semantics live in output_to_operand_aliasing, not in
+      // the fused body — walk the alias map instead.
+      if (user->fusion_kind() != HloInstruction::FusionKind::kCustom) break;
```
akuegel: Please use {} around `break;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+                   output_idx.size() == 1 &&
+                   output_idx[0] < fused_root->operand_count()) {
+          aliased = fused_root->operand(output_idx[0]);
+        } else {
+          continue;
+        }
+        while (aliased->opcode() == HloOpcode::kBitcast) {
+          aliased = aliased->operand(0);
+        }
+        if (aliased->opcode() != HloOpcode::kDynamicUpdateSlice) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    case HloOpcode::kConditional: {
+      // Scope: operand passed straight to a branch (not inside a tuple);
+      // each receiving branch's parameter_0 must be killed.
+      bool every_branch_kills = true;
+      bool flows_into_some_branch = false;
+      for (int b = 0; b < user->branch_count(); ++b) {
+        const HloInstruction* branch_input = user->operand(b + 1);
+        if (branch_input != operand) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      auto bcfg = user->backend_config<GpuBackendConfig>();
+      if (!bcfg.ok()) break;
+      const FusionBackendConfig& fc = bcfg->fusion_backend_config();
+      if (fc.kind() != kCustomFusionKind || !fc.has_custom_fusion_config() ||
+          fc.custom_fusion_config().name() !=
+              kDynamicSliceFusionWithDynamicAddressComputationConfigName) {
+        break;
+      }
+      for (const auto& alias : user->output_operand_aliasing()) {
+        if (alias.second.first != op_idx) continue;
```
akuegel: Please use {} around `continue`. And combine with condition below.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  return result;
+}
+
+// Two checks: (1) body_param[slot] is overwritten before any read, and
+// (2) body_root[slot] carries the writer's output, not a passthrough.
+absl::StatusOr<bool> SlotIsDeadInput(int64_t slot, const WhileLoopConfig& cfg) {
+  HloComputation* body = cfg.while_instr->while_body();
+  HloInstruction* body_param = body->parameter_instruction(0);
+  HloInstruction* body_root = body->root_instruction();
+  if (body_root->opcode() != HloOpcode::kTuple) return false;
```
akuegel: Please use {} around `return false;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+
+  HloInstruction* slot_gte = nullptr;
+  for (HloInstruction* u : body_param->users()) {
+    if (u->opcode() == HloOpcode::kGetTupleElement &&
+        u->tuple_index() == slot) {
+      if (slot_gte != nullptr) return false;  // multiple GTEs at slot
```
akuegel: Please use {} around `return false;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      slot_gte = u;
+    }
+  }
+  if (slot_gte == nullptr) return true;  // never read; trivially dead
```
akuegel: Please use {} around `return true;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+
+  WalkCtx ctx{cfg};
+  TF_ASSIGN_OR_RETURN(bool all_kill, AllUsersKillBuffer(slot_gte, ctx));
+  if (!all_kill) return false;
```
akuegel: Please use {} around `return false;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+
+  // If the body passes slot through unchanged (root traces to slot_gte via
+  // bitcast/copy), post-loop uses would see garbage if we elided the init.
+  const HloInstruction* root_v = body_root->operand(slot);
+  while (root_v->opcode() == HloOpcode::kBitcast ||
+         root_v->opcode() == HloOpcode::kCopy) {
+    root_v = root_v->operand(0);
+  }
+  if (root_v == slot_gte) return false;  // unmodified passthrough
```
akuegel: Please use {} around `return false;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    }
+  }
+
+  bool changed = false;
+  for (HloInstruction* while_op : whiles) {
+    if (while_op->has_sharding()) continue;
+    auto loop_cfg = while_op->backend_config<WhileLoopBackendConfig>();
+    if (!loop_cfg.ok() || !loop_cfg->has_known_trip_count()) continue;
+    int64_t trip_count = loop_cfg->known_trip_count().n();
+    if (trip_count <= 0) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    for (HloInstruction* ins : comp->instructions()) {
+      if (ins->opcode() == HloOpcode::kWhile) whiles.push_back(ins);
+    }
+  }
+
+  bool changed = false;
+  for (HloInstruction* while_op : whiles) {
+    if (while_op->has_sharding()) continue;
+    auto loop_cfg = while_op->backend_config<WhileLoopBackendConfig>();
+    if (!loop_cfg.ok() || !loop_cfg->has_known_trip_count()) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+  for (HloComputation* comp :
+       module->MakeNonfusionComputations(execution_threads)) {
+    for (HloInstruction* ins : comp->instructions()) {
+      if (ins->opcode() == HloOpcode::kWhile) whiles.push_back(ins);
+    }
+  }
+
+  bool changed = false;
+  for (HloInstruction* while_op : whiles) {
+    if (while_op->has_sharding()) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    int64_t iv_tuple_idx;
+    if (loop_cfg->has_known_induction_variable()) {
+      iv_tuple_idx = loop_cfg->known_induction_variable().tuple_index();
+    } else {
+      std::optional<int64_t> iv_idx_opt = GetLoopInductionVarTupleIdx(while_op);
+      if (!iv_idx_opt.has_value()) continue;
+      iv_tuple_idx = *iv_idx_opt;
+    }
+
+    if (!loop_cfg->has_known_init_step()) continue;
```
akuegel: Please use {} around `continue`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+    if (!loop_cfg.ok() || !loop_cfg->has_known_trip_count()) continue;
+    int64_t trip_count = loop_cfg->known_trip_count().n();
+    if (trip_count <= 0) continue;
+
+    int64_t iv_tuple_idx;
+    if (loop_cfg->has_known_induction_variable()) {
+      iv_tuple_idx = loop_cfg->known_induction_variable().tuple_index();
+    } else {
+      std::optional<int64_t> iv_idx_opt = GetLoopInductionVarTupleIdx(while_op);
+      if (!iv_idx_opt.has_value()) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      continue;
+    }
+
+    WhileLoopConfig cfg{/*while_instr=*/while_op,
+                        /*init=*/0,
+                        /*trip_count=*/trip_count,
+                        /*induction_var_idx=*/iv_tuple_idx};
+
+    HloInstruction* init_tuple = while_op->mutable_operand(0);
+    if (init_tuple->opcode() != HloOpcode::kTuple) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+
+    int64_t n_slots = init_tuple->operand_count();
+    for (int64_t slot = 0; slot < n_slots; ++slot) {
+      if (slot == cfg.induction_var_idx) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+
+      HloInstruction* init = init_tuple->mutable_operand(slot);
+      std::optional<Shape> alloc_shape_opt = ClassifyReplaceableInit(init);
+      if (!alloc_shape_opt.has_value()) continue;
```
akuegel: Please use {} around `continue`, and combine with conditions below

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+      if (init->has_sharding()) continue;
+      if (init->user_count() != 1) continue;  // init must feed only this while
+      const Shape& alloc_shape = *alloc_shape_opt;
+      if (alloc_shape.dimensions_size() == 0) continue;
+      if (alloc_shape.dimensions(0) != cfg.trip_count) continue;
+
+      TF_ASSIGN_OR_RETURN(bool dead, SlotIsDeadInput(slot, cfg));
+      if (!dead) continue;
```
akuegel: Please use {} around `continue;`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.h
```diff
@@ -23,31 +23,23 @@ limitations under the License.
 #include "xla/hlo/pass/hlo_pass_interface.h"
 
 namespace xla {
```
akuegel: nit: please use `namespace xla::gpu`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_execution_test.cc
```diff
@@ -0,0 +1,132 @@
...
+==============================================================================*/
+
+#include <gtest/gtest.h>
+#include "absl/strings/string_view.h"
+#include "xla/error_spec.h"
+#include "xla/service/hlo_module_config.h"
+#include "xla/tests/hlo_pjrt_test_base.h"
+#include "xla/xla.pb.h"
+
+namespace xla {
```
akuegel: nit: please use `namespace xla::gpu`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
```diff
@@ -0,0 +1,610 @@
...
+#include "absl/strings/string_view.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/hlo/ir/hlo_opcode.h"
+#include "xla/hlo/testlib/hlo_hardware_independent_test_base.h"
+#include "xla/hlo/utils/hlo_query.h"
+#include "xla/tsl/platform/statusor.h"
+
+namespace xla {
```
akuegel: nit: please use `namespace xla::gpu`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+           .debug_options()
+           .xla_gpu_enable_dus_accumulator_zero_init_elimination()) {
+    return false;
+  }
+
+  std::vector<HloInstruction*> whiles;
+  for (HloComputation* comp :
+       module->MakeNonfusionComputations(execution_threads)) {
+    for (HloInstruction* ins : comp->instructions()) {
+      if (ins->opcode() == HloOpcode::kWhile) whiles.push_back(ins);
```
akuegel: Please use {} around `whiles.push_back(ins);`

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,435 @@
...
+absl::StatusOr<bool> DusAccumulatorZeroInitElimination::RunImpl(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  if (!module->config()
+           .debug_options()
+           .xla_gpu_enable_dus_accumulator_zero_init_elimination()) {
+    return false;
+  }
+
+  std::vector<HloInstruction*> whiles;
```
akuegel: In general I would prefer if you only collect those while instructions that you want to process down below, instead of collecting all whiles and having lots of `continue` statements in the while loop below.
Also, the conditons what needs to hold for whether we want to process a while loop could go into a helper method, so that this method RunImpl does not get too big. So at least some part of the checks could be done earlier while filtering for while loops, and you could potentially also return std::optional<WhileLoopConfig> from your helper method and store that config together with the loop you have extracted.

xla/service/while_loop_unroller.cc
terryysun: Fair point, but introducing two dependent PRs will make the upstreaming take much longer, considering this is only ~10 lines, shall we keep them here?

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks! fixed.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks! fixed.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: updated.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Make sense, refactored.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks! fixed.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Good point! Updated to the simpler approach.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Code style fixed. It's not intentional, and I just realized that we don't even need the cap since there cannot be infinite recursion in HLO so we no longer need this part.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks! fixed.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: It was initially introduced as a defensive guard, but I just realized that HLO cannot have infinite recursion so we don't need the cap. Removed.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks! removed that max cap

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks! fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.h
terryysun: updated

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_execution_test.cc
terryysun: updated

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination_test.cc
terryysun: updated

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: fixed

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thank you for the suggestions! Refactored to a helper function and now it only collects loops that pass the filter.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,475 @@
...
+         h->opcode() == HloOpcode::kCopy ||
+         h->opcode() == HloOpcode::kReshape) {
+    h = h->operand(0);
+  }
+  return h->opcode() == HloOpcode::kBroadcast &&
+         hlo_query::IsScalarConstant(hlo_query::StripCastLike(h->operand(0)));
+}
+
+// Init value is irrelevant; soundness comes from the dead-input check.
+std::optional<Shape> ClassifyReplaceableInit(const HloInstruction* init) {
```
akuegel: Suggestion: just return a boolean here, as essentially you either return std::nullopt or init->shape(). I checked the call site, and there you can check whether this method returns true, and if yes, assign init->shape() to alloc_shape.
Also the method name could be improved, what about `IsInitReplaceable`?

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,475 @@
...
+Range LoopIterationRange(const WhileLoopConfig& cfg) {
+  return Range{ConstantValue::GetSigned(0, /*bitwidth=*/64),
+               ConstantValue::GetSigned(cfg.trip_count - 1, /*bitwidth=*/64),
+               ConstantValue::GetSigned(1, /*bitwidth=*/64),
+               /*is_linear=*/true};
+}
+
+// RecursivelyIdentifyRange handles the recursion into fused parameters, so
+// this works inside or outside fusions.
+std::optional<Range> ResolveIndexRangeAsFunctionOfIv(
```
akuegel: It looks like there is just one caller of this function. Could you just always return a `Range` instead of `std::optional<Range>`, and have the check from line 119 at the call site?

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,475 @@
...
+    const HloInstruction* idx = dus->operand(2 + d);
+    std::optional<Range> r = ResolveIndexRangeAsFunctionOfIv(idx, cfg);
+    if (r.has_value() && r->IsBounded() && r->IsSingleValue() &&
+        r->min().GetSignedValue() == 0) {
+      if (slice_shape.dimensions(d) != input_shape.dimensions(d)) {
+        return false;
+      }
+      continue;
+    }
+    // Reject if we already saw a dynamic dim, or if no range was resolved.
```
akuegel: I think this check could be before the one in line 147 (where has_value() would be replaced by the check that is currently inside the function ResolveIndexRangeAsFunctionOfIv). Then, in line 147 no check for `r.has_value()` would be needed.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
```diff
@@ -0,0 +1,475 @@
...
+  }
+  if (root_v == slot_gte) {
+    return false;  // unmodified passthrough
+  }
+  return true;
+}
+
+// Filters a single while op against the pass's loop-shape preconditions and
+// returns a WhileLoopConfig if it's a candidate, nullopt otherwise.
+std::optional<WhileLoopConfig> ClassifyCandidateWhile(
```
akuegel: I think you could define your own struct here instead of reusing WhileLoopConfig. I noticed below that you need a const_cast because this struct has the `while_instr` as const. I think it would be better to have a custom struct and avoid the const_cast.

xla/xla.proto
```diff
@@ -651,6 +651,10 @@ message DebugOptions {
   // Rewrite layer norm patterns into cuDNN library calls.
   optional bool xla_gpu_enable_cudnn_layer_norm = 262;
 
+  // Replaces broadcast(0) scan-accumulator init with AllocateBuffer when the
+  // body's DUS covers every slot.
+  optional bool xla_gpu_enable_dus_accumulator_zero_init_elimination = 485;
```
akuegel: This will need another rebase.

akuegel (commented):
Sorry for the late reply, as it is a rather large PR I didn't get to the review last Friday, and on Monday there was a public holiday here.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks for the detailed suggestion! Renamed and it returns boolean now.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Updated, always return range now and checked at call site.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Good point! Hoisted the check and it's simpler now.

xla/backends/gpu/transforms/dus_accumulator_zero_init_elimination.cc
terryysun: Thanks! Refactored with a local struct.

xla/xla.proto
terryysun: Rebased.

akuegel (approved):

PR #42879, opened 2026-05-19
[ROCm] Iterate full activity vector in ApiActivityInfoExchange to fix HIP-graph kernel loss

magaonka-amd:
📝 Summary of Changes
- The second half of ApiActivityInfoExchange in `xla/backends/profiler/gpu/rocm_collector.cc` keyed activities by  correlation_id and read only `activity_iter.second.front()`. 
- This is problem with when hipGraphLaunch is involved because single graph launch can have multiple kernels all with same correlation ID. and all that gets recorded only once in xplane because it only reads front()

🎯 Justification
without this change whenever command buffer is ON xla profiler drops events and events in final trace will be missing.
- Iterate the entire activity vector instead of only .front().
- Hoist the API-event lookup out of the inner loop (invariant across the vector since all activities share the correlation_id).

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
Added `RocmCollectorTest.MultipleActivitiesPerCorrelationIdAllExported` in rocm_collector_test.cc. Inserts one api_event + three activity events all sharing one correlation_id (the hipGraphLaunch shape), runs Flush() + Export(), and asserts all three kernel records appear in the xplane device plane.

### DIFF
--- a/xla/backends/profiler/gpu/rocm_collector.cc
+++ b/xla/backends/profiler/gpu/rocm_collector.cc
@@ -623,7 +623,11 @@ std::vector<RocmTracerEvent> RocmTraceCollectorImpl::ApiActivityInfoExchange() {
   */
 
   std::vector<RocmTracerEvent> aggregated_events;
-  aggregated_events.reserve(api_events_map_.size());
+  size_t total_activities = 0;
+  for (const auto& [_, v] : activity_ops_events_map_) {
+    total_activities += v.size();
+  }
+  aggregated_events.reserve(api_events_map_.size() + total_activities);
 
   // Copy info from activity events to API callback events
   for (auto& [key, api_event] : api_events_map_) {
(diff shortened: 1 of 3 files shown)


xla/backends/profiler/gpu/rocm_collector.cc
```diff
@@ -666,65 +670,77 @@ std::vector<RocmTracerEvent> RocmTraceCollectorImpl::ApiActivityInfoExchange() {
...
-
-      if (api_event == auxiliary_api_events_map_.end()) {
-        OnEventsDropped(
-            "An event from activity was discarded."
-            "Could not find the counterpart HIP API.",
-            activity_event.correlation_id);
-        PrintRocmTracerEvent(activity_event, ". Dropped!");
-        continue;
-      }
+    if (activity_iter.second.empty()) continue;
```
akuegel: Please use {} around `continue;`

magaonka-amd: My bad I should unlearn this habit. fixed it thanks.

akuegel (approved):

PR #42887, opened 2026-05-19
[xla:gpu] Add DynamicSliceFusionRewriterV2 targeting V2 DUS-fusions

ezhulenev:
Add `DynamicSliceFusionRewriterV2` that targets the V2 of the dynamic-slice-fusion library, it wraps a hero instruction that reads from sliced buffers (via Slice or DynamicSlice) and/or writes into sliced buffers (via DynamicUpdateSlice) into a custom fusion.

In contrast to previous (legacy, to-be-deleted) version it allows flexible configs for what slices/update-slices must become a part of the DUS-fusion, in some cases we don't want to pull particular slices into the fusion body as it would create unwanted buffer dependencies and limit scheduler ability to overlap computation/communication.

This is not wired into gpu-compiler, it will come in the followup PRs.

### DIFF
--- a/xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
+++ b/xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
@@ -1,0 +1,646 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.h"
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "absl/algorithm/container.h"
+#include "absl/container/flat_hash_map.h"
+#include "absl/container/flat_hash_set.h"
+#include "absl/container/inlined_vector.h"
+#include "absl/log/check.h"
+#include "absl/log/log.h"
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/str_cat.h"
+#include "absl/strings/string_view.h"
+#include "absl/types/span.h"
+#include "xla/tsl/platform/status_macros.h"
+#include "xla/hlo/ir/hlo_casting_utils.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_instructions.h"
+#include "xla/hlo/ir/hlo_opcode.h"
+#include "xla/service/gpu/backend_configs.pb.h"
+#include "xla/service/gpu/gpu_constants.h"
+#include "xla/service/gpu/ir_emission_utils.h"
+#include "xla/shape.h"
+#include "xla/shape_util.h"
+
+namespace xla::gpu {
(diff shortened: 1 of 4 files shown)


xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -0,0 +1,646 @@
...
+  const Shape& slice_shape = instr->opcode() == HloOpcode::kDynamicUpdateSlice
+                                 ? instr->operand(1)->shape()
+                                 : instr->shape();
+
+  auto byte_strides = ShapeUtil::ByteStrides(slice_shape);
+  if (!byte_strides.has_value()) {
+    return false;
+  }
+
+  int64_t slice_bytes = ShapeUtil::ByteSizeOf(slice_shape);
```
akuegel: Not sure whether it matters here, and it is a pre-existing problem, but ByteSizeOf seems to call into ByteSizeOfPrimivitveType eventually and that assumes PRED type is packed (but it isn't).

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -0,0 +1,646 @@
...
+
+  // All instructions to clone into the fusion body (topological order):
+  // slices, noops, hero, result noops, DUS instructions. The last instruction
+  // becomes the fusion root (may be a single DUS, or the hero itself when
+  // there are no DUS results). When there are multiple results,
+  // CreateFusionBody appends a tuple instruction as the root.
+  std::vector<HloInstruction*> instructions;
+};
+
+std::optional<DynamicSliceFusionPlan> BuildFusionPlan(
```
akuegel, line 402: Very nice :)

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -0,0 +1,646 @@
...
+                cloned_hero, result.result_number)));
+      }
+    }
+    builder.AddInstruction(HloInstruction::CreateTuple(tuple_operands));
+  }
+
+  return module->AddComputationAndUnifyNamesAndIds(builder.Build(), false);
+}
+
+absl::Status SetDsfBackendConfig(HloInstruction* fusion) {
```
akuegel: I stumbled when reading this, it is very close to `SetDfsBackendConfig`. Maybe no abbreviation here and use `SetDynamicSliceFusionBackendConfig`

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -0,0 +1,646 @@
...
+//===----------------------------------------------------------------------===//
+// RunImpl
+//===----------------------------------------------------------------------===//
+
+absl::StatusOr<bool> DynamicSliceFusionRewriterV2::RunImpl(
+    HloModule* module,
+    const absl::flat_hash_set<absl::string_view>& execution_threads) {
+  bool changed = false;
+
+  std::vector<HloComputation*> computations(module->computations().begin(),
```
akuegel: Why not initialize it to `MakeNonFusionComputations()`? Then we don't need the continue below.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
ezhulenev: Done

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
ezhulenev: Done

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
ezhulenev: changed to `ByteSizeOfPrimivitveType` to `CHECK ` that slice is an array. For PRED I think it should correctly return 1 byte, I'm a bit lost of all of the packing rules, but looking at the code I think it's correct. In worst case we'll just not rewrite it.

akuegel (approved):

PR #42976, opened 2026-05-20
Testing modification to arc-xla runners [Changes Should Not Be Merged]

Eliasj42:
📝 Summary of Changes
Testing a modification to arc-xla runners. confirming expected results before changing production configs
the rocm-ci workflow will target a temporary runner with the new configuration

🎯 Justification
The current runner is sometimes experiencing unclean shutdowns which is causing our backend nodes to stall for up to several hours in the worst cases

🚀 Kind of Contribution
⚡️ Performance Improvement, 🧪 Tests

`.github/workflows/rocm_ci.yml`

🧪 Unit Tests:
N/A

🧪 Execution Tests:
N/A

### DIFF
--- a/.github/workflows/rocm_ci.yml
+++ b/.github/workflows/rocm_ci.yml
@@ -43,7 +43,7 @@ jobs:
     needs: rocm-config
     if: ${{ github.event_name != 'pull_request' || github.base_ref == 'main' }}
     name: JAX Linux x86 AMD Instinct GPU
-    runs-on: linux-x86-64-1gpu-amd
+    runs-on: linux-x86-64-1gpu-amd.test
     env:
       DOCKER_IMAGE: ${{ needs.rocm-config.outputs.docker-image }}
     container: &rocm_container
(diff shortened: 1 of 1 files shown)


akuegel:
@Eliasj42 can you please turn this into a draft PR? Also, what exactly do you want to achieve with this PR if it should not be merged?

Eliasj42:
> @Eliasj42 can you please turn this into a draft PR? Also, what exactly do you want to achieve with this PR if it should not be merged?

I made changes to the runner backend And I want to confirm that there are no unintended side effects so I want to run the rocm_ci tests with the new runners. This PR points the rocm_ci workflow to a test runner  that contains all the changes we want to move into prod. 

It is highly unlikely that the new runners will have issues but I just wanted to run this last test to be 100% sure.

I currently have a runner created with the `.test` suffix that contains all the changes I am looking to add. if the workflow runs successfully I will move those changes into the prod runner on our end.

PR #42989, opened 2026-05-20
[xla:gpu] Support simple scalar expressions in DynamicSliceFusion

ezhulenev:
A second part after https://github.com/openxla/xla/pull/43046

Now that DynamicSliceFusion supports small scalar expressions to compute offsets, wire support for run time evaluation to Thunk


xla/backends/gpu/runtime/thunk.proto
```diff
@@ -188,13 +188,40 @@ message DynamicSliceThunkProto {
...
+  }
+
   // Mirrors DynamicSliceFusion::Offset.
   message OffsetProto {
     int64 dimension_number = 1;
-    oneof value {
-      int64 constant_offset = 2;
-      int64 runtime_parameter_number = 3;
-    }
+    OffsetExprProto offset = 2;
```
akuegel: I think we should not change the type of a proto tag. Can you instead mark the other ones reserved or something like that, and use tag number 4 here?

akuegel (commented):
Would it be possible to split this PR? E.g. the changes in dynamic_slice_fusion in a separate PR, and later on the runtime adjustments?

ezhulenev:
@akuegel extracted the `dynamic_slice_fusion` changes into https://github.com/openxla/xla/pull/43046, will update this PR once that one lands

ezhulenev:
Rebased it to latest main branch after part 1 was merged. Now this PR contains only runtime changes in a DUS thunk.

akuegel (approved):

PR #42996, opened 2026-05-21
[XLA:GPU] Run command buffer warmup without requiring collectives

shawnwang18:
📝 Summary of Changes
  CommandBufferThunk now keeps the warmup iteration enabled whenever fallback
  thunks are available, instead of only when commands require initialization
  through collective commands.

  Added a regression test covering initialized, non-collective command buffers to
  verify the first execution runs through the fallback thunk and subsequent
  execution uses the command buffer.

  🎯 Justification
  Some custom call library operators also require warmup before command buffer
  execution.  The concrete case is the TE/JAX `CollectiveGemm` external custom call when using
  the cuBLASMp backend. The custom call is marked `kCmdBufferCompatible` and is
  intended to be CUDA graph safe, but cuBLASMp internally uses NCCL for its
  symmetric-memory path. In the failing trace, graph capture was invalidated by an
  `ncclPutSignal` call inside cuBLASMp, producing
  `CUDA_ERROR_STREAM_CAPTURE_INVALIDATED`.

Previously warmup was effectively limited to collective commands via
  `Command::requires_initialization()`. This change enables warmup by default for
  all command-buffered operators with fallback thunks, covering custom call
  library operators and other non-collective cases.

  🚀 Kind of Contribution
  🐛 Bug Fix

  N/A

  🧪 Unit Tests:
  Added `CommandBufferThunkTest.InitializedThunkRunsWarmupWithoutCollectives`.

  🧪 Execution Tests:
  Ran:

  `bazel test //xla/backends/gpu/runtime:command_buffer_thunk_test --test_filter=CommandBufferThunkTest.InitializedThunkRunsWarmupWithoutCollectives`

### DIFF
--- a/xla/backends/gpu/runtime/BUILD
+++ b/xla/backends/gpu/runtime/BUILD
@@ -760,6 +760,7 @@ xla_test(
     backends = ["gpu"],
     tags = ["cuda-only"],
     deps = [
+        ":collective_params",
         ":command",
         ":command_buffer_cmd",
         ":command_buffer_thunk",
@@ -771,6 +772,7 @@ xla_test(
         "//xla:types",
         "//xla:xla_data_proto_cc",
         "//xla/runtime:buffer_use",
+        "//xla/runtime:device_id",
         "//xla/service:buffer_assignment",
         "//xla/service:executable",
         "//xla/service:platform_util",
--- a/xla/backends/gpu/runtime/BUILD
+++ b/xla/backends/gpu/runtime/BUILD
@@ -762,7 +762,6 @@ xla_test(
     deps = [
         ":collective_params",
         ":command",
-        ":command_buffer_cmd",
         ":command_buffer_thunk",
         ":command_executor",
         ":cudnn_thunk",
@@ -787,19 +786,16 @@ xla_test(
         "//xla/stream_executor:kernel_spec",
         "//xla/stream_executor:platform",
         "//xla/stream_executor:platform_manager",
+        "//xla/stream_executor:stream_executor_address_allocator",
         "//xla/stream_executor:stream_executor_h",
-        "//xla/stream_executor:stream_executor_memory_allocator",
         "//xla/stream_executor/cuda:cuda_compute_capability",
         "//xla/stream_executor/cuda:cudnn_plugin",
-        "//xla/tsl/lib/core:status_test_util",
-        "//xla/tsl/platform:statusor",
         "@com_google_absl//absl/status:status_matchers",
         "@com_google_absl//absl/status:statusor",
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/types:span",
         "@com_google_googletest//:gtest_main",
         "@cudnn_frontend_archive//:cudnn_frontend",
-        "@local_config_cuda//cuda:cuda_headers",
     ],
 )
 
--- a/xla/backends/gpu/runtime/command_buffer_thunk.cc
+++ b/xla/backends/gpu/runtime/command_buffer_thunk.cc
@@ -206,9 +206,8 @@ absl::Status CommandBufferThunk::Initialize(const InitializeParams& params) {
       GetOrCreateCommandBuffer(params.executor, *params.buffer_allocations));
   absl::MutexLock lock(cmd_buffer->mutex);
 
-  // If there are no thunks, or command buffer does not require initialization,
-  // we can mark warm up as done immediately.
-  if (!thunks_ || !commands_.requires_initialization()) {
+  // If there are no thunks, there is no fallback path to use for warmup.
+  if (!thunks_) {
     cmd_buffer->warmup_done = true;
   }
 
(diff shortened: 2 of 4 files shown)


akuegel:
I think @ezhulenev knows this code better than I do, so I will wait for him to take a look first.

ezhulenev (approved):

akuegel (approved):

golechwierowicz:
It looks like `xla/backends/gpu/runtime:cuda_command_buffer_thunk_test_b200` test fails.

shawnwang18:
> It looks like `xla/backends/gpu/runtime:cuda_command_buffer_thunk_test_b200` test fails.

Thanks,the test should be fixed.

akuegel (approved):

akuegel:
@shawnwang18 we got a 161% regression in the prefill step of some model. We will rollback for now. If I understand correctly, currently the only other case besides collectives where we would potentially need the warmup is for CustomCallThunk? If yes, wouldn't the easiest fix be to add an override in CustomCallThunk and LegacyCustomCallThunk that returns true for `requires_initialization()` (same as is done for CollectiveThunk)?

shawnwang18:
> @shawnwang18 we got a 161% regression in the prefill step of some model. We will rollback for now. If I understand correctly, currently the only other case besides collectives where we would potentially need the warmup is for CustomCallThunk? If yes, wouldn't the easiest fix be to add an override in CustomCallThunk and LegacyCustomCallThunk that returns true for `requires_initialization()` (same as is done for CollectiveThunk)?

Strange, I do not understand why there are such huge perf difference. The change in this PR is that first step is now running without cuda-graph, and cuda-graph is started on the 2nd step.  future steps will remain the same as before.  

We cannot directly use require_initialization() to implement that, because this is a special requirement for nccl operator where even local node does not have memory pointer change, XLA still does graph update. This is not required for custom_call I think.

akuegel:
> > @shawnwang18 we got a 161% regression in the prefill step of some model. We will rollback for now. If I understand correctly, currently the only other case besides collectives where we would potentially need the warmup is for CustomCallThunk? If yes, wouldn't the easiest fix be to add an override in CustomCallThunk and LegacyCustomCallThunk that returns true for `requires_initialization()` (same as is done for CollectiveThunk)?
> 
> Strange, I do not understand why there are such huge perf difference. The change in this PR is that first step is now running without cuda-graph, and cuda-graph is started on the 2nd step. future steps will remain the same as before.
> 
> We cannot directly use require_initialization() to implement that, because this is a special requirement for nccl operator where even local node does not have memory pointer change, XLA still does graph update. This is not required for custom_call I think.

This is something I don't understand yet. Currently we use the warmup if there is any collective (by `CollectiveThunk` returning true for `requires_initialization()`), and `CommandExecutor`s `requires_initialization()` iterating over all commands and checking if any of them returns true for `requires_initialization()`. So as far as I understand, if `CustomCallThunk` returns true for `requires_initialization()`, we would also use it if in the commands attached to `CommandExecutor` there is any CustomCallThunk. So what other case exists in which we need warmup that is not caught with that?

Regarding huge perf difference: Apparently we have many small kernels in that benchmark, so running with cuda graph or without makes a big difference. It is just the prefill step that becomes slower, not sure how much that translates into end2end regression.

shawnwang18:
> This is something I don't understand yet. Currently we use the warmup if there is any collective (by `CollectiveThunk` returning true for `requires_initialization()`), and `CommandExecutor`s `requires_initialization()` iterating over all commands and checking if any of them returns true for `requires_initialization()`. So as far as I understand, if `CustomCallThunk` returns true for `requires_initialization()`, we would also use it if in the commands attached to `CommandExecutor` there is any CustomCallThunk. So what other case exists in which we need warmup that is not caught with that?

making the custom_call's requires_initialization() can resolve the warmup issue, it do will trigger warmup for this operator.  

The issue is the the side effects triggered by requires_initialization, command buffer runtime will always do graph update on every step, even if current step's memory pointers do not change for operators who has requires_initialization().  

The always_update behavior is required by NCCL operator, NCCL needs all rank's memory pointer persistent in order to skip the update, but this should not be the requirement for custom_call.

shawnwang18:
>  Regarding huge perf difference: Apparently we have many small kernels in that benchmark, so running with cuda graph or without makes a big difference. It is just the prefill step that becomes slower, not sure how much that translates into end2end regression.

I was thinking that we does not need care too much about the perf for warmup step.

akuegel:
Ok, with Gemini's help I was able to understand what the problem with my idea is (you mentioned it as well, but as I don't really know the runtime it was hard for me to understand). So `requires_initialization` controls not just the warmup, but also the graph capture. While we do potentially need warmup for custom calls, we do not need graph capture. Gemini suggests this: "To fix the performance regression properly, the runtime needs to separate these two concepts. We need a new flag or method (e.g., requires_warmup()) that:

Triggers the one-time fallback warmup run on the first iteration.
Does not force the command buffer to re-record on every subsequent step."

shawnwang18:
> Ok, with Gemini's help I was able to understand what the problem with my idea is (you mentioned it as well, but as I don't really know the runtime it was hard for me to understand). So `requires_initialization` controls not just the warmup, but also the graph capture. While we do potentially need warmup for custom calls, we do not need graph capture. Gemini suggests this: "To fix the performance regression properly, the runtime needs to separate these two concepts. We need a new flag or method (e.g., requires_warmup()) that:
> 
> Triggers the one-time fallback warmup run on the first iteration. Does not force the command buffer to re-record on every subsequent step."

Yes, that is what happened. 

Do we really cares about the warmup step's perf?  If not, doing the warmup for all operators by default make sense to me, there are currently too many flags in Thunk, and warmup maybe required for many implementations. 

@ezhulenev @akuegel what you guys think? 
adding a new need_warmup() flag for thunk, or always do warmup regardless of operator types?

PR #43001, opened 2026-05-21
[ROCm] Make rocm-smi logging less verbose

Eetusjo:
📝 Summary of Changes
Change from LOG(WARNING) to VLOG(2) in recently introduced rocm-smi integration and related code.

🎯 Justification
The existing warnings are too verbose in environments where metrics queries fail for one reason another, resulting in log spam, and potentially mislead e.g. in diagnosing failing tests. Give the user control over whether to show them.

### DIFF
--- a/xla/stream_executor/rocm/rocm_executor.cc
+++ b/xla/stream_executor/rocm/rocm_executor.cc
@@ -1161,9 +1161,8 @@ RocmExecutor::CreateDeviceDescription(int device_ordinal) {
     if (pcie_bw.has_value()) {
       desc.set_pcie_bandwidth(*pcie_bw);
     } else {
-      LOG(WARNING) << "Could not determine PCIe bandwidth for device "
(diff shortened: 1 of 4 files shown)


pemeliya:
@Eetusjo are you sure we shall just mute all of these rocm-smi warnings? I mean, all of them are not informative for us?
After that, no one can figure out that we need to set proper values in TF_CPP_VMODULE to get them back

Eetusjo:
> @Eetusjo are you sure we shall just mute all of these rocm-smi warnings? I mean, all of them are not informative for us? After that, no one can figure out that we need to set proper values in TF_CPP_VMODULE to get them back

Yeah, talked about this and I'm going to restore some of the warnings esp. for failing rocm-smi API calls. I'll see if I could leave some as VLOGs to still reduce the verbosity. Didn't have time to change this yet but I'll do it asap

PR #43046, opened 2026-05-21
[xla:gpu] Support simple scalar expressions in DynamicSliceFusion part #1

ezhulenev:
Previously `DynamicSliceFusion` required offset to be a known constant value or a parameter of the fusion. This is suboptimal as it leaves offset calculation outside of fusion and it means that XLA is forced to compile a kernel for it, and it is super wasteful as the kernel operates on scalar value. 

 - Add `DynamicSliceFusion::Offset::Expr`, a small scalar offset expression AST for constants, fusion parameters, arithmetic, comparisons, and selects.
- Extend dynamic-slice fusion analysis and thunk serialization to carry offset expressions instead of only constant or runtime parameter offsets.

Thunk still supports only constant or runtime offsets, runtime support will come in next PR.

### DIFF
--- a/xla/backends/gpu/transforms/dynamic_slice_fusion.h
+++ b/xla/backends/gpu/transforms/dynamic_slice_fusion.h
@@ -19,6 +19,8 @@ limitations under the License.
 #include <cstdint>
 #include <optional>
 #include <ostream>
+#include <type_traits>
+#include <utility>
 #include <variant>
 #include <vector>
 
@@ -26,6 +28,8 @@ limitations under the License.
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_format.h"
 #include "absl/strings/str_join.h"
+#include "absl/types/span.h"
+#include "xla/comparison_util.h"
 #include "xla/hlo/ir/hlo_computation.h"
 #include "xla/hlo/ir/hlo_instruction.h"
 #include "xla/service/gpu/backend_configs.pb.h"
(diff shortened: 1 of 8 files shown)


akuegel (approved):

xla/backends/gpu/transforms/dynamic_slice_fusion.h
```diff
@@ -183,43 +205,79 @@ struct DynamicSliceFusion {
...
+                       const DynamicSliceFusion::Offset::Expr::Compare& b) {
+  return a.direction == b.direction && a.args == b.args;
+}
+
+inline bool operator==(const DynamicSliceFusion::Offset::Expr::Select& a,
+                       const DynamicSliceFusion::Offset::Expr::Select& b) {
+  return a.args == b.args;
+}
+
+inline bool operator==(const DynamicSliceFusion::Offset::Expr& a,
```
golechwierowicz, line 261: It looks like this part does not compile internally

```
[src/libcxx/include/__algorithm/comp.h:26]:16: error: invalid operands to binary expression ('const xla::gpu::DynamicSliceFusion::Offset::Expr' and 'const xla::gpu::DynamicSliceFusion::Offset::Expr')
   26 |     return __x == __y;
      |            ~~~ ^  ~~~
...
[xla/backends/gpu/transforms/dynamic_slice_fusion.h:235]:17: note: in instantiation of function template specialization 'std::__u::operator==<xla::gpu::DynamicSliceFusion::Offset::Expr, std::allocator<xla::gpu::DynamicSliceFusion::Offset::Expr>>' requested here
  235 |   return a.args == b.args;
[xla/backends/gpu/transforms/dynamic_slice_fusion.h:223] note: candidate function not viable: no known conversion from 'const xla::gpu::DynamicSliceFusion::Offset::Expr' to 'const DynamicSliceFusion::Offset::Expr::Constant' for 1st argument
  223 | inline bool operator==(const DynamicSliceFusion::Offset::Expr::Constant& a,
      |             ^          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[xla/backends/gpu/transforms/dynamic_slice_fusion.h:228]():13: note: candidate function not viable: no known conversion from 'const xla::gpu::DynamicSliceFusion::Offset::Expr' to 'const DynamicSliceFusion::Offset::Expr::Parameter' for 1st argument
  228 | inline bool operator==(const DynamicSliceFusion::Offset::Expr::Parameter& a,
      |             ^          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[xla/backends/gpu/transforms/dynamic_slice_fusion.h:233]():13: note: candidate function not viable: no known conversion from 'const xla::gpu::DynamicSliceFusion::Offset::Expr' to 'const DynamicSliceFusion::Offset::Expr::Add' for 1st argument
  233 | inline bool operator==(const DynamicSliceFusion::Offset::Expr::Add& a,
      |             ^          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ezhulenev: Added forward declaration, I think it must be enough to compile it.

pifon2a (approved):

PR #43249, opened 2026-05-27
[XLA:GPU] Enable VMM-allocator backed VA remapping for GPU command buffers

shawnwang18:
# Command-Buffer VA Remapping: the issues with the current approach and how the new idea resolves them

This document explains the **new idea** for supporting GPU command buffers, the
**problems with the current approach** that motivated it, and **how the new idea
resolves each problem**. It stays at the level of design intent, not code.

---

## 1. Background: why command buffers need "VA remapping"

A GPU command buffer (a CUDA graph) records the **device addresses** of the
buffers it touches at the moment it is captured. Replaying the graph is cheap —
that is the whole point of command buffers. But if the allocator hands out
*different* addresses on the next run, the recorded graph is stale and must be
**updated** node-by-node before it can replay. For large graphs that update is a
real, recurring cost.

The way to avoid it is **VA remapping**, built on CUDA Virtual Memory
Management, which separates *virtual addresses* from *physical memory*:

> Record the command buffer once against a **fixed virtual address range**, and
> on each execution point those fixed addresses at whatever physical memory the
> run actually uses. The graph always sees the same addresses, so it never needs
> updating.

Both the current approach and the new idea pursue this. They differ in **how the
fixed addresses are kept valid across runs**, and in **which allocations get this
treatment** — and that difference is the source of the problems below.

---

## 2. The current approach, and the issues it creates

The current approach keeps the addresses fixed by *"re-mapping everything, every
time."* A virtual address range is reserved once, and on **every execution** the
entire range is re-pointed at the current run's physical memory. Which
allocations to remap is a **static choice** decided ahead of time (e.g. "remap
all of them" or "remap a structurally-chosen subset").

This creates three issues:

### Issue 1 — Constant re-mapping cost

Every allocation is re-mapped on every run, even allocations whose physical
address never actually changed. The driver-level remap work is paid continuously,
run after run, regardless of whether anything moved.

### Issue 2 — Double-buffering overhead

You cannot re-map a virtual address range while the GPU is still reading it. To
work around this, the current approach **double-buffers**: it alternates between
two reservation sets so the CPU can re-map one while the GPU executes on the
other, synchronized by an event. This costs two reservation sets, two copies of
the graph per device, extra plumbing to alternate between them, and a
cross-execution synchronization point.

### Issue 3 — A static, all-or-nothing guess

The choice of which allocations are stable enough to pin is made **statically**,
before the program has run. It is a guess: it cannot tell which allocations are
*genuinely* stable at runtime, so it must either over-commit (pin allocations
that actually move, forcing fallback work) or under-commit (leave stable
allocations updating needlessly). The policy also lives **inside the
command-buffer thunk**, coupling the thunk tightly to virtual addresses and remap
modes.

---

## 3. The new idea

The new idea reframes the goal as *"don't re-map when nothing changed, and only
pin the allocations that are actually stable."* It rests on three shifts, each of
which targets one of the issues above.

### Shift 1 — Reuse persistent mappings *(resolves Issue 1)*

Instead of tearing down and rebuilding the address-to-memory mapping every run,
the mapping is made **persistent and reusable**. In steady state an allocation
usually lands at the same physical memory each iteration, so the existing mapping
is simply kept — most runs do **no driver-level remapping at all**. Re-mapping
happens only when an address genuinely changes. When teardown is needed it is
deferred and ordered against the GPU stream, rather than forcing the host to
wait.

→ The continuous remap cost of Issue 1 disappears: work is paid only on real
change.

### Shift 2 — Drop double-buffering *(resolves Issue 2)*

Once mappings are reused rather than rebuilt, there is no longer any reason to
overlap "CPU re-mapping" with "GPU executing" — the common path performs no
re-mapping to overlap. So the entire double-buffering apparatus (two reservation
sets, the alternating index, the synchronization event) is **removed**, replaced
by a single persistent reservation per device that is created once and reused for
the executable's lifetime.

→ The structural overhead and synchronization point of Issue 2 are eliminated.

### Shift 3 — Learn the policy adaptively *(resolves Issue 3)*

Rather than statically guessing which allocations are stable, the executable
**watches the first couple of executions** and then freezes a decision:

- Allocations whose address proves **stable** are pinned to a fixed virtual
  address — recorded once, never updated.
- Allocations that **move** are left alone — they keep using ordinary
  command-buffer updates.

The stable/dynamic split is therefore *learned from real behavior* instead of
assumed. The remap-vs-update **policy is also decided centrally** (by the
executable) and simply handed to the command-buffer thunk for each execution, so
the thunk no longer needs to know about remap modes or virtual addresses — it
just applies the policy it is given. (The old static modes still exist, expressed
as special cases of this same general mechanism.)

→ The mis-guessing and the tight coupling of Issue 3 are both resolved: the
decision is data-driven and made in one place.

---

## 4. Summary

| Issue with current approach | How the new idea resolves it |
|---|---|
| Re-maps every allocation on every run, even unchanged ones | Reuses persistent mappings; re-maps only on real change |
| Double-buffering: two reservation sets, duplicate graphs, sync event | Single persistent reservation; no overlap needed, so no double-buffering |
| Static, all-or-nothing guess of which allocations to pin, baked into the thunk | Adaptive split learned from observed stability; policy centralized and handed to the thunk |

### In one sentence

> The current approach keeps command-buffer addresses fixed by **re-mapping
> everything on every run** (and double-buffers to hide the cost); the new idea
> keeps them fixed by **reusing persistent mappings** and pins only the
> allocations it has **observed to be stable**, leaving the rest to normal
> updates.

# Benchmark results
image

  🧪 Unit Tests:
  Added and updated coverage for:
  - `//xla/stream_executor/cuda:cuda_device_address_vmm_allocator_test`
    - VMM allocation/deallocation behavior
    - mapped allocation and reservation alias handling
    - stale allocation/mapping reuse
    - deferred unmap/deallocation batching
    - PA-budget reclaim behavior
    - multi-device allocator isolation
  - `//xla/backends/gpu/runtime:command_buffer_thunk_test`
    - verifies command-buffer update policy skips commands whose allocations are
      VA-remapped.
  - `//xla/service/gpu:gpu_executable_test`
    - verifies command-buffer allocation index collection skips MLIR constants.

  🧪 Execution Tests:
  Added CUDA/Hopper-gated PJRT execution coverage in
  `//xla/pjrt/gpu:se_gpu_pjrt_client_test` for:
  - VMM allocator setup and basic E2E execution
  - command-buffer VA remapping for fusion, GEMM/cuBLAS, conditional, while, and
    dynamic-slice fusion workloads
  - repeated single-range VA reuse
  - `CAPTURE_CMD_NEVER_UPDATE` custom library command execution
  - interleaved execution of multiple executables with independent VA reservations


phambinhfin:
Hi @shawnwang18, thanks for the great work on this.
One thing I want to raise about dropping the double buffer for a single persistent
reservation: it really only wins when the physical addresses stay the same across
steps, so the remap almost never happens.  So for static-shape models it's totally fine. But this stability is not guaranteed in general.. With a
single reservation, every remap step has to synchronize first (wait for the GPU
to finish before we touch the mapping), so we lose the CPU/GPU overlap that the
double buffer gave us. On those workloads it could be a regression vs double buffer.
So my suggestion: maybe keep the double buffer for now, and just add the "only
remap the slices that actually changed, skip the ones that didn't" optimization
on top of it. That way we still cut most of the remap cost, but we don't lose the
overlap safety net when addresses do move.

shawnwang18:
@phambinhfin Thanks for the comments, you raised the key point: 

```
every remap step has to synchronize first (wait for the GPU
to finish before we touch the mapping), so we lose the CPU/GPU overlap that the
double buffer gave us
```

we definitely want to avoid the synchronization introduced by remap. keeping both double buffering and reused-based remapping may achieve this, but still have challenges: 
1. One downside of keeping both is that the design is too complex, I want the design to be clean. 
2.  We may still can not hide the remapping cost with VA-range multiplexing, because cuda VMM APIs are required to set all peer device memory's VMM access, the cost is very high, it has a high chance that VMM cost is larger than the device run time. 

While, actually we can achieve this even without double buffering, below is the key point: 

1.  we will not try to do the VA-remapping for all buffers, instead we will allow some buffers to have non-persisitent addresses,  so for some command, if all its buffer allocations are from VA-remapped addresses, then it will alway skip command buffer update.  and if the command has any allocation are non-persisitent, then the command will require UPDATE.  

2. We will put those allocations to the non-persisitent set if we have no way to avoid VA-remapping across steps ( for exampke, output buffers that are not freed by the caller when entering the next step). 

3. For real workloads, I observed that the allocations address that have no way to preserve on the next step is very few, usually only the output buffers that that are not freed by the caller when entering the next step), for those output buffers, we can have below options: 

3.1 .  we do a copy of those buffers, and return the copied buffer to the caller (this PR implements this optimization), this approach works because this buffers are usually metrics buffer, and just a scalar, so copying them is low cost.  With copying, we can have a stable address for the source buffer of the copying, so single reservation range will not introduce VA-remap that needs CPU/GPU synchronization. 

3.2 . And there are not many of these buffers usually, update the command buffer nodes that access these buffers should be fine, we can just exclude them from the VA-remap.

phambinhfin:
Hi @shawnwang18, this resonates a lot — I actually tried a very similar copy-based
approach befor.
The idea was the same as yours: map a small set of physical "pool" slots to
permanent VA addresses once at startup, record the graph against those stable
VAs, and then per iteration just D2D-copy the param data into the pool before
replay and copy the live-outs back after. Since the graph always sees the same VA,
there's no per-iteration map/unmap/setaccess at all — we map once, then it's pure
replay + copy, so no VA-remap and no CPU/GPU sync.
In micro-benchmarks it was a big win — ~23x ,  transformer step and
~24x for  pmap+allreduce — exactly because the copied buffers were small
and the D2D copy (~2.9us/MB at HBM bandwidth for the devices i tested) is way cheaper than the NCCL re-trace
it avoids.
The reason I didn't push it further: in my version I copied ALL command-buffer
allocations (params + temps + live-outs). That's fine when buffers are small, but
once the model's inputs/outputs get big, the D2D copy cost stops being negligible
and starts eating the benefit (and the pool memory grows with the number of slots
too). So copying everything doesn't scale.
But your point that these are usually just small metric/scalar buffers is the key.
If we copy ONLY the small buffers — add a size check / threshold, copy the ones
below it and leave the large buffers on the normal path — we keep the cheap-copy
benefit and avoid the expensive copies for big I/O. Great idea!

PR #43375, opened 2026-05-28
[ROCm] Fix dot_bf16.hlo test

nurmukhametov:
kDot (bf16 x bf16 -> f32) is unsupported by hipBLASLt on gfx90a (MI200) and relied on the legacy cublas fallback removed in 115a93ed. Disable the test on mi200 and switch the ROCm RUN line to %{GPU}.txtpb so gfx1250 (and future ROCm targets) use their own spec instead of mi200's.

### DIFF
--- a/xla/backends/gpu/tests/BUILD
+++ b/xla/backends/gpu/tests/BUILD
@@ -808,6 +808,7 @@ lit_test_suite_for_gpus(
             "pdl.hlo",
         ],
         "mi200": [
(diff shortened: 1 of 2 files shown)


draganmladjenovic (approved):
LGTM

nurmukhametov:
@xla-rotation could I get a review on this, please?

dimitar-asenov (approved):

PR #43381, opened 2026-05-28
[xla:gpu] Migrate to DynamicSliceFusionRewriterV2

ezhulenev:
Migrate `xla_gpu_enable_dynamic_slice_fusion` flag to `DynamicSliceFusionRewriterV2`. This change intentionally drops support for `ReduceScatter` wrapping, it will be added separately. This flag is disabled by default, this PR tests that users that do enable it explicitly passes all the test. Enabling this flag by default will be in the followup PR.

In contrast to "legacy" DUS fusion new (V2) can wrap arbitrary HLO and doesn't need any special casing to handle custom calls or collectives. However the capturing of collectives is tricky as it can extend the lifetime of DUS buffers, as they are captured by asynct-start and async-done pairs. Support for wrapping collectives + correct "asyncification" will come in followup PRs after migrating to new mechanism.


akuegel (commented):
Can you please rebase?

ermilovmaxim (approved):

ermilovmaxim (approved):

PR #43390, opened 2026-05-28
Cross Host Data Transfers: Ensure one communicator used and one clique acquired per pair of communicating devices

rao-ashish:
📝 Summary of Changes
`PrepareTransfer` currently acquires a different clique and communicator for each unique `(src, dst)` pair of devices involved in a cross host transfer. However, if a batch of transfers involves both transfers `a->b` and `a<-b` for some devices `a` and `b`, this means that `a->b` and `a<-b` will be executed on different communicators, and will not be batched into a single NCCL group call. This PR modifies `PrepareTransfers` to ensure that sends and receives between the same pair of devices are batched into the same NCCL group.

🎯 Justification
Batching transfers with NCCL group calls can reduce launch overhead and provide substantial speedups for small message sizes. This PR allows for sends and receives executed on the same device to be executed as part of the same NCCL group.

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
This change is already tested by previously added unit tests to `se_gpu_pjrt_client_multi_gpu_test.cc`, which continue to pass.

🧪 Execution Tests:

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -448,18 +448,23 @@ class PreparedTransfer {
   AcquiredCliqueAndCommunicator clique_and_communicator_;
   bool is_sender_;
 
+  // The rank inside the communicator which corresponds to the remote device
+  // we need to transfer data to/from.
+  RankId peer_communicator_rank_;
+
   PreparedTransfer(StreamExecutorGpuClient* client,
                    gpu::GpuCliqueKey clique_key,
                    tsl::RCReference<PjRtRawBuffer> raw_buffer,
                    tsl::AsyncValueRef<BufferSequencingEvent> transfer_event,
                    AcquiredCliqueAndCommunicator clique_and_communicator,
-                   bool is_sender)
+                   bool is_sender, RankId peer_communicator_rank)
       : client_(client),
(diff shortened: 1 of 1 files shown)


xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -563,25 +568,35 @@ absl::StatusOr<PreparedTransfer> PrepareTransfer(
                            dst_device);
   });
 
-  // Form the GPU clique key.
+  // Form the GPU clique key. We use sorted([src_device, dst_device]) as the
+  // device list because GPU clique key is by default order-sensitive, and
+  // we want to use the same communicator for transfers i->j and i<-j.
   // TODO(asrao, mwhittaker): Supply correct incarnations when creating the
   // clique key.
+  std::vector<GlobalDeviceId> devices = {src_device, dst_device};
```
pschuh, line 576: absl::InlinedVector<GlobalDeviceId, 2>.

rao-ashish: This gets passed into `gpu::GpuCliqueKey()` and `AcquireCliqueAndCommunicator` below, which expect an `std::vector`, so it's probably better to just keep this as an `std::vector`?

pschuh (approved):

PR #43459, opened 2026-05-29
[XLA:GPU] Fix file check with the new cudnn FE 1.22

Cjkkkk:
📝 Summary of Changes
With recent cuDNN frontend upgrade to 1.22, the json dumps changes. Fixed the file check accordingly.


dimitar-asenov:
Hi @Cjkkkk. Thank you for the fix. We haven't yet imported 1.22 internally, so this fix will break our internal tests. We can merge this as soon as do the internal import.

PR #43461, opened 2026-05-29
[xla:gpu] Add RegisteredMemory API to XLA collectives

ezhulenev:
This is a revival of registered memory deleted in https://github.com/openxla/xla/commit/51ba59ee3cf4e42c5bb503aa236494988ab44d6e, the old API was useless, but the registered memory vs symmetric memory distinction is critical and we need both at run time. This PR restores registered memory in XLA, but in a new modernized form consistent with `SymmetricMemory`.

`RegisteredMemory` is purely a host-side optimization hint and lifetime handle: it lets NCCL set up IPC/NVLS/network handles for the range ahead of time so subsequent collectives run zero-copy, without changing collective semantics or results. Because no symmetry is assumed, it deliberately exposes no usable APIs.

### DIFF
--- a/xla/backends/gpu/collectives/BUILD
+++ b/xla/backends/gpu/collectives/BUILD
@@ -302,6 +302,7 @@ xla_cc_test(
         "//xla/core/collectives:communicator",
         "//xla/core/collectives:rank_id",
         "//xla/core/collectives:reduction_kind",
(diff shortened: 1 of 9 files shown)


PatriosTheGreat (approved):
Thanks!

ermilovmaxim (approved):

PR #43528, opened 2026-06-01
[XLA:CPU] Add TORAX iterhybrid HLO as in-tree reference workload

seantalts:
## Summary

Lands the [TORAX](https://github.com/google-deepmind/torax) ITER hybrid scenario (predictor-corrector solver) jitted simulation step as an in-tree HLO under `xla/tools/benchmarks/hlo/`.

- 524 inputs, 268 outputs, ~40k lines, 752 fusion/while/conditional ops on 25/26/101-cell radial grids. Differentiable 1D tokamak plasma transport simulator in JAX, f64. Represents a small-grid implicit-solve shape with many parallel scalar ops, not previously covered by the in-tree HLO corpus (which is matmul/attention-dominated).
- Size is rank-5 of existing `.hlo` files in the repo, between `hlo_mixtral_8x7b_bf16_2x8` and `hlo_mixtral_8x7b_bf16_1x8`.

## LAPACK FFI stubbing

The dumped HLO contained 7 LAPACK custom-calls (3 `lapack_dgetrf_ffi` on 4×4 and 24×3×3 matrices, 4 `lapack_dtrsm_ffi`) emitted by JAX's `linalg` lowering. XLA upstream does not register these handlers on CPU — they live in `jaxlib/cpu/_lapack.so` and are wired into XLA's runtime FFI registry by Python at `import jax` (`jax/_src/lax/linalg.py: register_module_custom_calls(lapack)`). Standalone XLA C++ runners (`run_hlo_module`, `multihost_hlo_runner`) cannot execute JAX HLO that touches `jax.numpy.linalg` as-is.

Each call has been rewritten in-place to a shape-equivalent stub:

```
dgetrf(A) -> tuple(copy(A), iota, 0)   # LU = A, perm = identity, info = 0
dtrsm(A, B) -> copy(B)                  # solve(A, B) = B (passthrough)
```

The stubs are O(n) instead of O(n³), but the affected matrices are 4×4 and 24×3×3 — negligible vs the surrounding fusion-heavy compute on the 25/26/101-cell grids. The HLO carries a top-of-file note and a per-site `// LAPACK stub:` comment at each replacement so the rewrite is self-documenting.

```
bazel run //xla/tools:run_hlo_module -- \
  --platform=Host --input_format=hlo --iterations=3 \
  $(pwd)/xla/tools/benchmarks/hlo/hlo_torax_iterhybrid_predictor_corrector_f64.hlo
```


seantalts:
@penpornk I wanted to get a TORAX benchmark in our OSS repo as well. This PR was created on my personal machine with public info.

penpornk (approved):
Thank you, Sean!

PR #43554, opened 2026-06-01
[XLA:GPU] Cleaning up command buffer command emitter

shawnwang18:
📝 Summary of Changes
  Cleaning up command_buffer_cmd_emitter.cc to simplify thunk-to-command conversion:

  - Uses the shared Command base class for direct command appends.
  - Keeps control-flow, async, sequential, and command-buffer cases explicit.
  - Splits concurrent-region collection, scheduling, and dependency insertion into smaller helpers.
  - Removes unused bookkeeping and prunes now-unused BUILD deps/includes.

  🎯 Justification
  This reduces boilerplate in command buffer emission and makes the concurrent-region path easier to read and maintain without changing supported behavior.

  🚀 Kind of Contribution
  ♻️  Cleanup

  N/A, this is a cleanup.

  🧪 Unit Tests:
  bazel test //xla/backends/gpu/runtime:command_buffer_cmd_emitter_test

  🧪 Execution Tests:
  N/A. No runtime behavior change intended.


xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
```diff
@@ -498,70 +387,114 @@ class ConcurrentRegionScheduler {
   std::vector<Thunk*> scheduled_thunks_;
   absl::flat_hash_map<const Thunk*, std::vector<Thunk*>> scheduled_successors_;
   std::vector<Thunk*> scheduled_source_thunks_;
-  std::vector<Thunk*> scheduled_sink_thunks_;
+  std::array<Thunk*, kMaxConcurrentOps> scheduled_sink_thunks_ = {};
```
ezhulenev, line 390: is ` = {}` really needed here? forgot what are the initialization rules for members

ezhulenev (approved):

xla/backends/gpu/runtime/command_buffer_cmd_emitter.cc
shawnwang18: Yes, I think it is required. 

Becuase the original design is using std::vector.resize(), which will initialize to null for unused slots, ` = {} ` is doing the same initialization .

dimitar-asenov (approved):

PR #43630, opened 2026-06-02
[ROCm] Enable HLO XLA transform pass registration on GPU

mminutoli:
## Summary

- **`gpu_compiler.cc`**: Add `ApplyXlaTransforms` calls at the end of `RunOptimizationPasses` (PRE_SCHEDULER) and at the start of `RunPostSchedulingPipelines` (POST_SCHEDULER). The GPU compiler never called this pass, so registered transforms were silently ignored.
- **`pjrt_c_api_gpu_internal.cc`**: Add `PJRT_Xla_Transform_Extension` to the GPU plugin's extension chain (between `shardings_extension` and `abi_version_extension`). Without it, `FindExtension` returned null and registration via the PJRT C API was silently skipped.
- **`pjrt_c_api_xla_transform_internal.cc`**: Fix a bug in `CApiXlaTransformAdapter`: it previously stored `PJRT_XlaTransform_Callbacks` by value and passed the address of that copy to `transform_hlo_module`. The JAX-side callback (`CApiTransformHloModuleCallback`) uses `offsetof` arithmetic to recover its `CApiCallbackState` from that pointer, so it must receive the original registered address. Fix by storing a pointer instead.

### DIFF
--- a/xla/pjrt/c/BUILD
+++ b/xla/pjrt/c/BUILD
@@ -655,6 +655,8 @@ cc_library(
         ":pjrt_c_api_triton_extension_hdrs",
         ":pjrt_c_api_triton_internal",
         ":pjrt_c_api_wrapper_impl",
+        ":pjrt_c_api_xla_transform_extension_hdrs",
+        ":pjrt_c_api_xla_transform_internal",
         "//xla/backends/cpu:target_machine_options",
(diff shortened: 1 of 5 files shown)


andydavis1 (approved):
Thanks! Just wanted to comment that this infrastructure is really new, so expect changes and issues as we iterate on this.

hawkinsp (approved):

PR #43638, opened 2026-06-03
[xla:gpu] Automatically register all allocations with GPU communicators

ezhulenev:
- Add `AllocatorMemoryRegistration` to track allocator-owned GPU ranges reported by BFC suballocator visitors.
- Register recorded ranges with newly-created GPU cliques, using only the communicator on the same local device ordinal.
- Tie NCCL registered-memory lifetimes to the clique and release registrations when allocator ranges are freed.
- Enable this path in SE GPU PJRT only for BFC with preallocation.

We deliberately do pre-registration only when XLA uses BFC with preallocation as in this case we can guarantee that all memory is allocated before the first communicator is created. Supporting dynamic memory allocation is out of scope op this PR, but can be added later if needed.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -47,6 +47,7 @@ limitations under the License.
 #include "absl/types/span.h"
 #include "xla/tsl/platform/status_macros.h"
 #include "xla/backends/cpu/target_machine_options.h"
+#include "xla/backends/gpu/collectives/allocator_memory_registration.h"
 #include "xla/backends/gpu/collectives/gpu_clique.h"
 #include "xla/backends/gpu/collectives/gpu_clique_key.h"
 #include "xla/backends/gpu/collectives/gpu_cliques.h"
@@ -238,7 +239,8 @@ StreamExecutorGpuClient::StreamExecutorGpuClient(
     std::shared_ptr<KeyValueStoreInterface> kv_store,
     bool abort_collectives_on_failure,
     std::shared_ptr<const GpuTopology> gpu_topology,
-    std::optional<int> num_nodes)
+    std::optional<int> num_nodes,
+    std::shared_ptr<gpu::AllocatorMemoryRegistration> memory_registration)
     : xla::PjRtStreamExecutorClient(
           platform_name, client, std::move(devices), process_index,
           /*memory_spaces=*/{},  // Initialized below.
@@ -246,6 +248,7 @@ StreamExecutorGpuClient::StreamExecutorGpuClient(
           should_stage_host_to_device_transfers, std::move(gpu_run_options)),
       num_nodes_(num_nodes),
       abort_collectives_on_failure_(abort_collectives_on_failure),
+      memory_registration_(std::move(memory_registration)),
       kv_store_(std::move(kv_store)) {
   VLOG(1) << absl::StreamFormat(
       "Constructed StreamExecutor GPU client: #devices=%d #num_nodes=%d",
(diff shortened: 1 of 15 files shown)


xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1797,10 +1809,26 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
...
                                           options.max_inflight_computations));
   EnablePeerAccess(xla_client->backend().stream_executors());
+
+  GpuAllocatorConfig allocator_config = options.allocator_config;
+
+  // Add allocator visitors to track allocator-owned ranges and register them
+  // with newly-created same-device clique communicators.
+  auto memory_registration = MaybeCreateAllocatorMemoryRegistration(options);
+  if (memory_registration != nullptr) {
+    gpu::RegisterOnGpuCliqueCreatedCallback(
```
PatriosTheGreat: Nit: This looks like an initialization logic, could you move this to MaybeCreateAllocatorMemoryRegistration?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1286,6 +1289,15 @@ BuildLocalDeviceStates(LocalClient* xla_client, bool schedule_async,
   return std::move(addressable_devices);
 }
 
+std::shared_ptr<gpu::AllocatorMemoryRegistration>
+MaybeCreateAllocatorMemoryRegistration(const GpuClientOptions& options) {
+  if (!options.allocator_config.preallocate ||
+      options.allocator_config.kind != GpuAllocatorConfig::Kind::kBFC) {
```
PatriosTheGreat: Should this also be disabled to kVMM?

PatriosTheGreat (changes requested):

xla/pjrt/gpu/se_gpu_pjrt_client.cc
ezhulenev: It's not a problem for NCCL, each transport layer can just ignore a registration hint if memory allocation type is not compatible, just don't want to forget to revert it once everything will be vmm.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
ezhulenev: Done.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
ezhulenev: Now that VMM is the default device allocator, it should always work when we use BFC. The second VMM allocator is a different beast, not even sure it will respect sub allocator callbacks, but it's not enabled by default, so I don't worry about it.

ezhulenev: Removed the check completely, as this is just a performance optimizatuon hint not required for correctness, and non-compliant allocators simply ignore callbacks.

PatriosTheGreat (approved):

dimitar-asenov:
The constructor whose signature you're changing is also used in Tesorflow:

- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/eager/context_distributed_manager.cc#L333
- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/gpu/gpu_device.cc#L1945

The current PR breaks those calls internally. Could you please evolve the code in such a way so that we can submit? If you prepare a corresponding PR for Tensorflow, I could merge the two internally.

ezhulenev:
Rebased + fixed. Moved new arg to the end of the constructor arg list and made default to nullptr.

ezhulenev:
I'd not fix TF side at all, memory registration is a performance hint not needed for correctness, and I'm not sure how exactly TF allocates and manages memory pools, and I don't want to know :)

ermilovmaxim (approved):

PR #43686, opened 2026-06-03
[XLA:GPU] Adapt SolLatencyEstimator for Triton and Custom kernel intra-node AllReduce backend

mfrancepillois:
📝 Summary of Changes
This change introduces a three-layer fix:

1. CollectiveKernelStrategy proto annotation (backend_configs.proto)
Add CollectiveKernelStrategy enum to CollectiveBackendConfig with KERNEL_STRATEGY_DEFAULT (NCCL), KERNEL_STRATEGY_TRITON_CUSTOM_ONE_SHOT, KERNEL_STRATEGY_TRITON_CUSTOM_TWO_SHOT. Both the Triton codegen kernel and the built-in custom C++ kernel receive the same annotation values, because they implement identical NVLink data-movement patterns: one-shot performs (N-1) direct full-buffer reads from peers; two-shot performs reduce-scatter + all-gather, each phase reading (N-1)/N of the buffer from peers. A single annotation therefore correctly covers both backends.

2. New HLO pass: CollectiveKernelStrategyAnnotator
Runs after CollectiveBackendAssigner, before scheduling. Calls BuildAllReduceInfo() (the same eligibility check used by thunk_emitter) to determine whether the one-shot or two-shot custom kernel will be used at runtime — for both the Triton codegen and the built-in C++ backend — and annotates the HLO instruction's backend_config. Only registered when xla_gpu_unsupported_use_all_reduce_one_shot_kernel is enabled, keeping the NCCL path completely unchanged otherwise.

3. SolGPUCostModel / SolLatencyEstimator updates
SolGPUCostModel::Config: add nvlink_bw_per_lane_gbps (populated from CudaBandwidthSettings / RocmBandwidthSettings via GpuPerformanceWithCollectiveModel::GetIciBandwidthPerLaneGbps) and nvlink_barrier_latency.
SolGPUCostModel::IntraNodeAllReduceLatency: NVLink-based formula for one-shot (≤ 256 KB) and two-shot (≤ 4 MB) strategies, applicable to both the Triton codegen kernel and the built-in custom C++ kernel. Strategy is derived internally from size_bytes via GetAllReduceStrategy(), matching the runtime threshold exactly.
SolLatencyEstimator::DispatchEstimation SINGLE_PARTITION path: when kernel_strategy is KERNEL_STRATEGY_TRITON_CUSTOM_ONE_SHOT or KERNEL_STRATEGY_TRITON_CUSTOM_TWO_SHOT, use IntraNodeAllReduceLatency (NVLink formula + HBM time) instead of CollectiveInterpolator (NCCL empirical table). A single code path covers both backend variants because they perform identical NVLink traffic: (N-1)×size_bytes for one-shot, 2×(N-1)/N×size_bytes for two-shot.
SolLatencyEstimator::NodeCost: when a sync custom kernel AllReduce (Triton codegen or built-in C++) runs on the compute stream and fully blocks compute, return ComputeCollectiveTime rather than kLowCost=1.0 µs, so the latency lands on the critical path.

🎯 Justification
Intra-node custom collective kernels — both the Triton codegen variant and the built-in custom C++ variant (selected by xla_gpu_unsupported_use_all_reduce_one_shot_kernel) — use GPU-side P2P NVLink symmetric memory access rather than the NCCL/RCCL ring algorithm, so the existing NIC-bandwidth-based cost model is wrong for those collectives.

Without this change, enabling either custom AllReduce kernel causes two independent categories of incorrect scheduling decisions:

1. Async custom kernel AllReduce: wrong GetLatencyBetween budget
SolLatencyEstimator::GetLatencyBetween(AllReduceStart, AllReduceDone) returns the value of ComputeCollectiveTime(), which routes intra-node (SINGLE_PARTITION) collectives to CollectiveInterpolator. CollectiveInterpolator uses NCCL-calibrated empirical bandwidth tables. The Latency Hiding Scheduler uses this budget to decide how much compute to interleave between AllReduceStart and AllReduceDone. An over-estimated budget causes the scheduler to try to fill a large overlapping window that does not exist at runtime, potentially reordering compute operations in ways that hurt rather than help. An under-estimated budget would prevent profitable overlap from being exploited.

2. Sync custom kernel AllReduce: NodeCost returns kLowCost instead of the true blocking latency
The sync custom kernel AllReduce is dispatched on the compute stream and fully blocks all further compute until it finishes. NodeCost() currently returns kLowCost=1 µs for any IsAsyncCollectiveStartOp() instruction regardless of is_sync. This causes the scheduler to treat a ~5 µs blocking op as if it were free on the critical path. For a Transformer training step with many small AllReduces (e.g. bias/norm gradient synchronisations), this error compounds: the scheduler builds incorrect priority orders and may produce a schedule that is far from optimal.

🚀 Kind of Contribution

🧪 Execution Tests:
- collective_kernel_strategy_annotator_test.cc: verifies small (ONE_SHOT), medium (TWO_SHOT) and large (DEFAULT/NCCL) AllReduces are annotated correctly.
- sol_gpu_cost_model_test.cc: IntraNodeAllReduceLatency formula validation for one-shot and two-shot strategies.
- sol_latency_estimator_test.cc: scheduler integration — async custom kernel ONE_SHOT uses the NVLink formula and differs from NCCL; sync custom kernel ONE_SHOT ComputeCollectiveTime exceeds kLowCost; async NCCL NodeCost remains kLowCost=1 µs.


dimitar-asenov:
Hi @mfrancepillois . This PR is rather large and it would help with the review if you break it down into smaller ones, e.g. one PR per layer.

mfrancepillois:
This PR has been split into 3 smaller PRs:
PR 1/3: https://github.com/openxla/xla/pull/43742
PR 2/3: https://github.com/openxla/xla/pull/43743
PR 3/3: https://github.com/openxla/xla/pull/43745
This PR can therefore be closed.

PR #43701, opened 2026-06-03
  [XLA:GPU] Lower DynamicSliceFusionV2 to command buffers

shawnwang18:
📝 Summary of Changes
  Adds GPU command buffer lowering support for DynamicSliceFusionV2Thunk.

 DynamicSliceFusionV2Thunk is lowered to a command buffer when:

  - DYNAMIC_SLICE_FUSION is enabled in xla_gpu_enable_command_buffer.
  - It is in a convertible thunk sequence that meets xla_gpu_graph_min_graph_size.
  - verify_offsets is disabled.
  - All embedded thunks inside the DSF thunk are also command-buffer convertible.
  - If command-buffer mode is NEVER_UPDATE, the DSF thunk must not have loop-dependent offsets.
  - If it is inside a WhileThunk and the whole while is being lowered, loop-dependent DSF offsets require loop unrolling to be enabled and the while loop to have a known trip count.

  This change:
  - Makes DynamicSliceFusionV2Thunk implement Command.
  - Records DynamicSliceFusionV2Thunk as a child command buffer.
  - Propagates command-buffer prepare/initialize/update requirements through nested command executors.
  - Forces command buffer updates when loop-dependent dynamic slice offsets can change without allocation address changes.
  - Blocks command-buffer lowering when runtime offset verification is enabled.
  - Avoids unsafe NEVER_UPDATE lowering for loop-dependent offsets unless the containing while loop can be unrolled.

  🎯 Justification
  DynamicSliceFusionV2Thunk is currently excluded from command buffer lowering, which prevents graphs containing dynamic slice fusion from being captured as command buffers.

  This enables command buffer execution for dynamic slice fusion while preserving correctness for loop-dependent offsets and unsupported runtime offset verification paths.

  🚀 Kind of Contribution
  ✨ New Feature, 🧪 Tests

  N/A. This change enables command buffer lowering support and does not make a standalone performance claim.

  🧪 Unit Tests:
  Added coverage in command_buffer_conversion_pass_test.cc for:
  - Converting DynamicSliceFusionV2Thunk into a CommandBufferThunk.
  - Rejecting DynamicSliceFusionV2Thunk when verify_offsets is enabled.
  - Rejecting loop-dependent DynamicSliceFusionV2Thunk in NEVER_UPDATE mode.
  - Handling loop-dependent DynamicSliceFusionV2Thunk inside while loops with and without loop unrolling.

  🧪 Execution Tests:
  Added DynamicSliceFusionV2ThunkTest.CommandBufferUpdatesLoopDependentSliceOffset, which executes a command buffer containing DynamicSliceFusionV2Thunk and verifies that loop-dependent slice offsets are updated across executions.


xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
```diff
@@ -185,11 +190,39 @@ bool ThunkSequenceIsConvertible(const ThunkSequence& thunks,
...
                         const CommandBufferConfig& config);
 
+bool HasLoopDependentDynamicSliceFusionV2(const ThunkSequence& thunks) {
+  bool has_loop_dependent_dsf = false;
+  for (const std::unique_ptr<Thunk>& thunk : thunks) {
+    if (has_loop_dependent_dsf) {
+      break;
+    }
+    thunk
+        ->Walk([&](const Thunk* nested) -> absl::Status {
```
ezhulenev: I'm pretty sure Walk should support a callback that returns void, then you don't need to `IgnoreError`

xla/backends/gpu/runtime/command_executor.h
```diff
@@ -162,6 +162,14 @@ class CommandExecutor {
     return requires_initialization;
   }
 
+  bool requires_update() const {
+    bool requires_update = false;
+    commands_.Walk([&](const Command* command) {
+      requires_update |= command->requires_update();
```
ezhulenev, line 176: this is an example of void Walk callback

xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
```diff
@@ -119,6 +123,31 @@ DynamicSliceFusionV2Thunk::DynamicSliceFusionV2Thunk(
       executor_(std::move(embedded_thunks)),
       verify_offsets_(verify_offsets) {}
 
+static bool IsLoopDependent(const std::optional<DynamicSliceConfig>& config) {
+  return config.has_value() && config->has_loop_index() &&
+         config->byte_stride() != 0;
+}
+
+bool DynamicSliceFusionV2Thunk::HasLoopDependentOffsets() const {
+  for (const DynamicSliceFusion::Parameter& parameter : parameters_) {
```
ezhulenev: absl::c_any_of should make this shorter

```
bool param_loop_dependent = absl::c_any_of(..., IsLoopDependent)
bool res_loop_dependent = 
return param_loop_dependent | res_loop_dependent
```

xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
```diff
@@ -119,6 +123,31 @@ DynamicSliceFusionV2Thunk::DynamicSliceFusionV2Thunk(
...
+  }
+  for (const DynamicSliceFusion::Result& result : results_) {
+    if (IsLoopDependent(result.update_config)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+absl::Status DynamicSliceFusionV2Thunk::SetOrUpdateCommandBufferExecutor(
```
ezhulenev, line 147: I don't like it, but I think we should live with it for now, ideally all thunks with nested thuink executor should automatically in constructor build a corresponmding command execuitor if they are compatible.

Can you add a TODO and cleanup it later for all thunks with nested thunk executors. I think I already clened up all thunks and we don't have any thunks that "execute" SequentialThunk, everyuthing should be handled via executors.

xla/backends/gpu/runtime/command_buffer_conversion_pass.cc
shawnwang18: Fixed

xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
shawnwang18: Fixed

xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
shawnwang18: I did think about this,  because command buffer conversion pass happens after Thunk was constructed, I do not have a good idea on doing this.  Let me re-think.

ezhulenev: I think each individual thunk can just call conversion in constructor (optioanllyu, only if all nested commands are valid).

But for this we need to migrate all thunks to static factories instead of constructors (similar to thunks in CPU backend), because today we can't return an error from thunk constructor

ezhulenev (approved):

dimitar-asenov (approved):

dimitar-asenov (approved):

dimitar-asenov:
@shawnwang18 

It seems this code requires CUDA 12.9. Please make sure that the functions you change and the tests you add do not fail with 12.8. E.g. you can do a version check to keep the old logic or to skip a test:

```
  if (executor->GetDeviceDescription().runtime_version() <
      stream_executor::SemanticVersion(12, 9, 0)) { ... }
```

shawnwang18:
> @shawnwang18
> 
> It seems this code requires CUDA 12.9. Please make sure that the functions you change and the tests you add do not fail with 12.8. E.g. you can do a version check to keep the old logic or to skip a test:
> 
> ```
>   if (executor->GetDeviceDescription().runtime_version() <
>       stream_executor::SemanticVersion(12, 9, 0)) { ... }
> ```

Thanks the info, I just added the cuda guard for the feature. Is there any google internal tests failed with un-supported cuda version?

dimitar-asenov (approved):

PR #43722, opened 2026-06-04
[XLA:GPU][oneAPI] Enable Intel Triton XPU build

akhilgoe:
This PR integrates Intel Triton XPU backend into XLA


vwbaker:
Thanks for the PR Akhil,

We have put in a lot of effort to simplify our patching structure and not overcomplicate our dependence on Triton and this is going in the opposite direction, so I'd like to make sure we've tried everything else first.

(1) as for the third_party/triton/intel_xpu/intel_forward_fix.patch, some look like they can be upstreamed to triton, in which case we suggest doing that rather than creating patches here. Some of them look like an llvm discrepancy rather than an actual error - we have these patches already in the triton/common/ folder.

(2) Is there any way you can create something that is standalone rather than using patches like this? - it makes it very difficult to review and will likely cause issues updating from OSS because we want no differences between XLA's triton in OSS/internally. 

Triton also announced the possibility of extensions/plugins that can be loaded at runtime - https://docs.google.com/presentation/d/1dnm8uhvabdwqsQAsaPM7IRpEh2tktQ91E1d40r91n1M/edit?slide=id.gfc5f9321c5_2_472#slide=id.gfc5f9321c5_2_472. This I believe would be ideal for your use case if you want to try and get things added to a triton version without Triton's approval. https://github.com/triton-lang/triton-ext has some more information.

Alternatively, individual passes for added to XLA rather than patch files would be preferable.

akhilgoe:
Thanks @vwbaker for the review! 

For (1), yes, these changes are mainly to keep Intel XPU Triton compatible with the rest of the XLA environment. We can fold them into the XPU backend commit hash itself and remove this patch file entirely. 
With that, we should only need one remaining patch file, `intel_build.patch`, to add build rules for Intel XPU backend targets. The `oss/build_files.patch` updates should be a one-time change.

We also have some DPAS encoding (like MMA) related changes in shared/common code so using Triton extensions to create custom passes might not work. The Intel Triton team evaluated a plugin-based approach for some custom passes but did not continue with it because it does not work on Windows.

These changes, and future XPU backend hash updates, should remain NFC for non-Intel GPU backends in XLA.

akhilgoe:
I pushed a commit to remove `intel_forward_fix.patch` patch file. Please take a look and let us know what you think. Thanks!

vwbaker:
When was the last time you talked to Triton about adding a backend for intel? If they would accept this, then we will have no issue.

What is the plan here? As far as I can see, the build file is attempting to build targets that currently don't exist in XLA or Triton. Is the plan to have a patch file or a massive patch file with all intel's code that would live under triton/third_party/intel?

Why is the workspace file so complicated? It seems like all you need is an "if ENABLE_INTEL_XPU_TRITON += intel patches"? It also makes all the backends go through intel's definition of it, and I'd prefer to keep all the common parts of the code in the top workspace file. You could for example just call intel_patches() which could return an empty list if ENABLE_INTEL_XPU_TRITON is not set.

akhilgoe:
This PR should be a NFC for CUDA/ROCM. When `ENABLE_INTEL_XPU_TRITON` is set (for Intel GPU backends only), the Triton source is [Intel XPU Backend for Triton](https://github.com/intel/intel-xpu-backend-for-triton), otherwise the source is OpenAI Triton. So, no patch file needed for `third_party/intel`. Eg: In the presubmit checks for this PR, [CI / XLA Linux X86 GPU ONEAPI (pull_request)](https://github.com/openxla/xla/actions/runs/27261232814/job/80507183266?pr=43722) ran with the Intel XPU Backend for Triton as its Triton source, while all other CI jobs used OpenAI Triton.

vwbaker:
Has there been a recent attempt to talk to the Triton Compiler about getting intel as a third party backend?

akhilgoe:
@vwbaker I have added a commit to address your comments and marked this PR ready for review. Please take a look whenever you get time. Thanks!

xla/backends/gpu/codegen/triton/BUILD
```diff
@@ -155,7 +156,10 @@ cc_library(
         "compilation_pipeline.cc",
         "compilation_pipeline_cuda.cc",
         "compilation_pipeline_rocm.cc",
-    ],
+    ] + if_sycl_is_configured([
+        # TODO(intel-tf): Enable the oneAPI pipeline once it is ready in a future PR.
+        # "compilation_pipeline_oneapi.cc",
```
vwbaker: is this necessary? it is only a comment at the moment.

third_party/triton/workspace.bzl
```diff
@@ -1,15 +1,60 @@
...
+    if use_xpu_triton(repository_ctx):
+        sha256 = XPU_TRITON_SHA256
+        patch_files = oss_only_patch_list + [
+            "//third_party/triton:intel_xpu/intel_build.patch",
+        ]
+        strip_prefix = "intel-xpu-backend-for-triton-" + XPU_TRITON_COMMIT
+        urls = tf_mirror_urls("https://github.com/intel/intel-xpu-backend-for-triton/archive/{}.tar.gz".format(XPU_TRITON_COMMIT))
+
+    for patch_file in patch_files:
+        repository_ctx.path(Label(patch_file))
```
vwbaker, line 33: what does this do?

vwbaker (commented):
Hi Akhil, thanks for the changes. It looks good, just a couple comments.

xla/backends/gpu/codegen/triton/BUILD
akhilgoe: Not necessary, I just wanted to show how the oneAPI sources and dependencies would be added conditionally but I can remove it you prefer.
Once this PR lands, I plan to create a follow up PR to add the compilation pipeline and close all remaining gaps needed to enable Triton codegen for the oneAPI backend.

third_party/triton/workspace.bzl
akhilgoe: This resolves the patch_file label names. I borrowed this from the `tf_http_archive` rule which includes a comment referencing a Bazel issue where failing to resolve labels before `download_and_extract` triggers re-download of the archive.

https://github.com/openxla/xla/blob/main/third_party/repo.bzl#L57-L64

vwbaker: interesting; would you mind adding a comment about that?

xla/backends/gpu/codegen/triton/BUILD
vwbaker: yes please remove as it's not getting used yet.it looks like you already have an example that is real below.

third_party/triton/workspace.bzl
vwbaker: However, given that what you're doing is just overwriting the variables and then doing what tf_http_archive does, would you be able to overwrite the variables and then call tf_http_archive instead of manually handling the patches?

akhilgoe: Added the comment.
I was trying to go for something similar but the problem is that at the repository fetch time (when the decision has to be made to switch to Intel XPU Triton depending on the value of the `ENABLE_INTEL_XPU_TRITON` repo_env variable) the only runtime context available is through the [`environ`](https://github.com/openxla/xla/pull/43722/changes#diff-3b84c89ab4381930eb53b8ca1045f3b09ee677e699f3380ba0c57e70acbca2ecR51) field which is accessible only from within a `repository_rule` implementation. The existing `tf_http_archive` cannot access it. Hence, I could not think of a cleaner way to go about this.

xla/backends/gpu/codegen/triton/BUILD
akhilgoe: Done

third_party/triton/workspace.bzl
vwbaker: Thanks for clarifying!

vwbaker (approved):

PR #43729, opened 2026-06-04
  [xla:gpu] Split command buffer warmup from initialization and add warmup for custom_call thunk

shawnwang18:
📝 Summary of Changes
  Adds a separate requires_warmup() predicate for GPU command-buffer commands, distinct from requires_initialization(). Command buffer warmup now uses requires_warmup(), while initialization/update behavior still uses requires_initialization().

  Marks command-buffer-compatible FFI custom calls as requiring warmup without forcing initialization semantics. NCCL collective commands continue to require initialization, and therefore warmup by default.

  🎯 Justification
  Some commands need a one-time fallback execution before command-buffer capture, but should not force repeated initialization/update behavior. This separates the two concepts so kCmdBufferCompatible custom calls can warm up correctly without being treated like NCCL collectives.

  🚀 Kind of Contribution
  🐛 Bug Fix, ♻️  Cleanup, 🧪 Tests

  N/A. This is not a performance improvement.

  🧪 Unit Tests:
  Added coverage for:

  - Command::requires_warmup() defaulting to requires_initialization()
  - Commands requiring warmup independently from initialization
  - kCmdBufferCompatible custom calls requiring warmup
  - Owned custom-call handlers not requiring warmup by default

  🧪 Execution Tests:
  Validated with:

  bazel test //xla/backends/gpu/runtime:command_executor_test //xla/backends/gpu/runtime:custom_call_thunk_test --test_filter=CommandExecutorTest.RequiresWarmupDefaultsToRequiresInitialization:CommandExecutorTest.RequiresWarmupCanBeSeparateFromInitialization:CustomCallThunkTest.CommandBufferCompatibleCustomCallRequiresWarmup:CustomCallThunkTest.CustomCallWithOwnedHandlers

  Also previously smoke-tested command-buffer execution with:

  bazel test //xla/backends/gpu/runtime:command_buffer_thunk_test --test_filter=CommandBufferThunkTest.Memset32Cmd

### DIFF
--- a/xla/backends/gpu/runtime/collective_thunk.h
+++ b/xla/backends/gpu/runtime/collective_thunk.h
@@ -127,6 +127,7 @@ class CollectiveThunk : public Command {
 
   bool IsTracedCommand() const override { return true; }
   bool requires_initialization() const override { return true; }
+  bool requires_warmup() const override { return true; }
 
   absl::Status Prepare(const PrepareParams& params) override;
   absl::Status Initialize(const InitializeParams& params) override;
--- a/xla/backends/gpu/runtime/command.h
+++ b/xla/backends/gpu/runtime/command.h
@@ -160,6 +160,10 @@ class Command : public Thunk {
   // ensure that all ranks execute NCCL command update.
   virtual bool requires_initialization() const { return false; }
 
+  // Returns true if command requires a one-time fallback execution before it is
+  // recorded into a command buffer.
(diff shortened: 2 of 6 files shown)


olupton:
See also https://github.com/openxla/xla/pull/42996, which was a less targeted version of this that landed + got reverted. cc: also @ezhulenev

dimitar-asenov (approved):

PR #43742, opened 2026-06-04
[XLA:GPU] Adapt SolLatencyEstimator for Triton and Custom kernel intra-node AllReduce backend (1/3)

mfrancepillois:
📝 Summary of Changes

This PR is the first in a series of three.
PR 1/3: https://github.com/openxla/xla/pull/43742    <= This PR
PR 2/3: https://github.com/openxla/xla/pull/43743
PR 3/3: https://github.com/openxla/xla/pull/43745

This PR adds a new HLO pass that annotates AllReduce/AllReduceStart instructions with a CollectiveKernelStrategy field in their CollectiveBackendConfig, recording which custom kernel backend (ONE_SHOT, TWO_SHOT, or DEFAULT/NCCL) will be used at runtime.

This series of PRs introduces a three-layer fix:

1. CollectiveKernelStrategy proto annotation (backend_configs.proto)
Add CollectiveKernelStrategy enum to CollectiveBackendConfig with KERNEL_STRATEGY_DEFAULT (NCCL), KERNEL_STRATEGY_TRITON_CUSTOM_ONE_SHOT, KERNEL_STRATEGY_TRITON_CUSTOM_TWO_SHOT. Both the Triton codegen kernel and the built-in custom C++ kernel receive the same annotation values, because they implement identical NVLink data-movement patterns: one-shot performs (N-1) direct full-buffer reads from peers; two-shot performs reduce-scatter + all-gather, each phase reading (N-1)/N of the buffer from peers. A single annotation therefore correctly covers both backends.

2. New HLO pass: CollectiveKernelStrategyAnnotator
Runs after CollectiveBackendAssigner, before scheduling. Calls BuildAllReduceInfo() (the same eligibility check used by thunk_emitter) to determine whether the one-shot or two-shot custom kernel will be used at runtime — for both the Triton codegen and the built-in C++ backend — and annotates the HLO instruction's backend_config. Only registered when xla_gpu_unsupported_use_all_reduce_one_shot_kernel is enabled, keeping the NCCL path completely unchanged otherwise.

3. SolGPUCostModel / SolLatencyEstimator updates
SolGPUCostModel::Config: add nvlink_bw_per_lane_gbps (populated from CudaBandwidthSettings / RocmBandwidthSettings via GpuPerformanceWithCollectiveModel::GetIciBandwidthPerLaneGbps) and nvlink_barrier_latency.
[...]

PR #43745, opened 2026-06-04
[XLA:GPU] Adapt SolLatencyEstimator for Triton and Custom kernel intra-node AllReduce backend (3/3)

mfrancepillois:
📝 Summary of Changes

This PR is the third in a series of three.
PR 1/3: https://github.com/openxla/xla/pull/43742
PR 2/3: https://github.com/openxla/xla/pull/43743
PR 3/3: https://github.com/openxla/xla/pull/43745     <= This PR

Wire the kernel_strategy annotation (PR 1) and NVLink cost formula (PR 2) into the scheduler to fix two categories of incorrect scheduling decisions when xla_gpu_unsupported_use_all_reduce_one_shot_kernel is enabled:

1. Async AllReduce: GetLatencyBetween() budget. When kernel_strategy is TRITON_ONE/TWO_SHOT, route SINGLE_PARTITION AllReduces to TritonAllReduceLatency() (NVLink P2P formula) instead of CollectiveInterpolator (NCCL empirical bandwidth tables). This gives the Latency Hiding Scheduler a correct overlap budget.

2. Sync AllReduce: NodeCost() on the critical path. When is_sync=true and kernel_strategy is TRITON_ONE/TWO_SHOT, return ComputeCollectiveTime() instead of kLowCost=1us, so the blocking compute-stream AllReduce lands correctly on the critical path.

- sol_latency_estimator.cc: read kernel_strategy from CollectiveBackendConfig, dispatch accordingly in DispatchEstimation and NodeCost.
- sol_latency_estimator_test.cc: verify async Triton ONE_SHOT uses NVLink formula and differs from NCCL; sync Triton ONE_SHOT NodeCost exceeds kLowCost; async NCCL NodeCost remains 1us unchanged.

This series of PRs introduces a three-layer fix:

1. CollectiveKernelStrategy proto annotation (backend_configs.proto)
Add CollectiveKernelStrategy enum to CollectiveBackendConfig with KERNEL_STRATEGY_DEFAULT (NCCL), KERNEL_STRATEGY_TRITON_CUSTOM_ONE_SHOT, KERNEL_STRATEGY_TRITON_CUSTOM_TWO_SHOT. Both the Triton codegen kernel and the built-in custom C++ kernel receive the same annotation values, because they implement identical NVLink data-movement patterns: one-shot performs (N-1) direct full-buffer reads from peers; two-shot performs reduce-scatter + all-gather, each phase reading (N-1)/N of the buffer from peers. A single annotation therefore correctly covers both backends.

2. New HLO pass: CollectiveKernelStrategyAnnotator
Runs after CollectiveBackendAssigner, before scheduling. Calls BuildAllReduceInfo() (the same eligibility check used by thunk_emitter) to determine whether the one-shot or two-shot custom kernel will be used at runtime — for both the Triton codegen and the built-in C++ backend — and annotates the HLO instruction's backend_config. Only registered when xla_gpu_unsupported_use_all_reduce_one_shot_kernel is enabled, keeping the NCCL path completely unchanged otherwise.

3. SolGPUCostModel / SolLatencyEstimator updates
SolGPUCostModel::Config: add nvlink_bw_per_lane_gbps (populated from CudaBandwidthSettings / RocmBandwidthSettings via GpuPerformanceWithCollectiveModel::GetIciBandwidthPerLaneGbps) and nvlink_barrier_latency.
SolGPUCostModel::IntraNodeAllReduceLatency: NVLink-based formula for one-shot (≤ 256 KB) and two-shot (≤ 4 MB) strategies, applicable to both the Triton codegen kernel and the built-in custom C++ kernel. Strategy is derived internally from size_bytes via GetAllReduceStrategy(), matching the runtime threshold exactly.
SolLatencyEstimator::DispatchEstimation SINGLE_PARTITION path: when kernel_strategy is KERNEL_STRATEGY_TRITON_CUSTOM_ONE_SHOT or KERNEL_STRATEGY_TRITON_CUSTOM_TWO_SHOT, use IntraNodeAllReduceLatency (NVLink formula + HBM time) instead of CollectiveInterpolator (NCCL empirical table). A single code path covers both backend variants because they perform identical NVLink traffic: (N-1)×size_bytes for one-shot, 2×(N-1)/N×size_bytes for two-shot.
SolLatencyEstimator::NodeCost: when a sync custom kernel AllReduce (Triton codegen or built-in C++) runs on the compute stream and fully blocks compute, return ComputeCollectiveTime rather than kLowCost=1.0 µs, so the latency lands on the critical path.

🎯 Justification
Intra-node custom collective kernels — both the Triton codegen variant and the built-in custom C++ variant (selected by xla_gpu_unsupported_use_all_reduce_one_shot_kernel) — use GPU-side P2P NVLink symmetric memory access rather than the NCCL/RCCL ring algorithm, so the existing NIC-bandwidth-based cost model is wrong for those collectives.

Without this change, enabling either custom AllReduce kernel causes two independent categories of incorrect scheduling decisions:

1. Async custom kernel AllReduce: wrong GetLatencyBetween budget
SolLatencyEstimator::GetLatencyBetween(AllReduceStart, AllReduceDone) returns the value of ComputeCollectiveTime(), which routes intra-node (SINGLE_PARTITION) collectives to CollectiveInterpolator. CollectiveInterpolator uses NCCL-calibrated empirical bandwidth tables. The Latency Hiding Scheduler uses this budget to decide how much compute to interleave between AllReduceStart and AllReduceDone. An over-estimated budget causes the scheduler to try to fill a large overlapping window that does not exist at runtime, potentially reordering compute operations in ways that hurt rather than help. An under-estimated budget would prevent profitable overlap from being exploited.

2. Sync custom kernel AllReduce: NodeCost returns kLowCost instead of the true blocking latency
The sync custom kernel AllReduce is dispatched on the compute stream and fully blocks all further compute until it finishes. NodeCost() currently returns kLowCost=1 µs for any IsAsyncCollectiveStartOp() instruction regardless of is_sync. This causes the scheduler to treat a ~5 µs blocking op as if it were free on the critical path. For a Transformer training step with many small AllReduces (e.g. bias/norm gradient synchronisations), this error compounds: the scheduler builds incorrect priority orders and may produce a schedule that is far from optimal.

🚀 Kind of Contribution

🧪 Execution Tests:
- sol_latency_estimator_test.cc: scheduler integration — async custom kernel ONE_SHOT uses the NVLink formula and differs from NCCL; sync custom kernel ONE_SHOT ComputeCollectiveTime exceeds kLowCost; async NCCL NodeCost remains kLowCost=1 µs.

### DIFF
--- a/xla/service/gpu/model/sol_latency_estimator.cc
+++ b/xla/service/gpu/model/sol_latency_estimator.cc
@@ -59,6 +59,12 @@ namespace {
 
 using ::mlir::MLIRContext;
 
+bool IsTritonCollectiveKernel(
+    CollectiveBackendConfig::CollectiveKernelStrategy ks) {
+  return ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_ONE_SHOT ||
+         ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_TWO_SHOT;
+}
+
 bool IsSupportedCollectiveOp(const HloInstruction& instr) {
   return HloPredicateIsOp<HloOpcode::kAllReduceStart, HloOpcode::kAllReduce,
                           HloOpcode::kReduceScatter, HloOpcode::kAllGatherStart,
@@ -253,15 +259,13 @@ absl::StatusOr<absl::Duration> DispatchEstimation(
     }
     case GPUCommunicationType::SINGLE_PARTITION: {
       // If an intra-node collective kernel will be used (Triton codegen kernel
-      // or the built-in custom C++ kernel — both use the same NVLink-based
-      // cost formula), apply the NVLink model instead of the NCCL-calibrated
-      // CollectiveInterpolator.
-      auto backend_cfg = instr.backend_config<GpuBackendConfig>();
+      // — uses the NVLink-based cost formula), apply the NVLink model instead
+      // of the NCCL-calibrated CollectiveInterpolator.
+      absl::StatusOr<GpuBackendConfig> backend_cfg =
+          instr.backend_config<GpuBackendConfig>();
       if (backend_cfg.ok()) {
-        const auto ks =
-            backend_cfg->collective_backend_config().kernel_strategy();
-        if (ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_ONE_SHOT ||
-            ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_TWO_SHOT) {
+        if (IsTritonCollectiveKernel(
+                backend_cfg->collective_backend_config().kernel_strategy())) {
           SolGPUCostModel sol_model(sol_flags);
           const int num_gpus = num_groups_and_devices->second;
           const int active_links =
@@ -474,18 +478,17 @@ LatencyEstimator::TimeCost SolLatencyEstimator::GetLatencyBetween(
     return kLowLatency;
   }
 
-  // For sync intra-node AllReduce (Triton codegen or built-in custom C++
-  // kernel — both annotated as TRITON_ONE_SHOT / TRITON_TWO_SHOT) the kernel
-  // runs on the compute stream and fully blocks further compute — no overlap
-  // is possible between start and done.  NodeCost() already reports the full
-  // blocking latency on the critical path, so return kLowLatency here to
-  // avoid double-counting.
-  if (IsGPUSyncCollective(from.GetInstr())) {
-    auto cfg = from.GetInstr().backend_config<GpuBackendConfig>();
-    if (cfg.ok()) {
-      const auto ks = cfg->collective_backend_config().kernel_strategy();
-      if (ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_ONE_SHOT ||
-          ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_TWO_SHOT) {
+  // For sync intra-node AllReduce (Triton codegen kernel annotated as
+  // TRITON_ONE_SHOT / TRITON_TWO_SHOT) the kernel runs on the compute stream
+  // and fully blocks further compute — no overlap is possible between start
+  // and done.  NodeCost() already reports the full blocking latency on the
+  // critical path, so return kLowLatency here to avoid double-counting.
+  {
+    absl::StatusOr<GpuBackendConfig> cfg =
+        from.GetInstr().backend_config<GpuBackendConfig>();
+    if (cfg.ok() && cfg->collective_backend_config().is_sync()) {
+      if (IsTritonCollectiveKernel(
+              cfg->collective_backend_config().kernel_strategy())) {
         VLOG(10) << "GetLatencyBetween: Returning kLowLatency for sync "
                     "intra-node AllReduce (cost already in NodeCost): "
                  << from.GetInstr().name();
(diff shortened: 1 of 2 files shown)


xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -247,6 +252,33 @@ absl::StatusOr<absl::Duration> DispatchEstimation(
           gpu_device_info, sol_flags, analysis);
     }
     case GPUCommunicationType::SINGLE_PARTITION: {
+      // If an intra-node collective kernel will be used (Triton codegen kernel
+      // or the built-in custom C++ kernel — both use the same NVLink-based
```
sohaibiftikhar: nit: The c++ kernel has been removed.

xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -247,6 +252,33 @@ absl::StatusOr<absl::Duration> DispatchEstimation(
...
+      // cost formula), apply the NVLink model instead of the NCCL-calibrated
+      // CollectiveInterpolator.
+      auto backend_cfg = instr.backend_config<GpuBackendConfig>();
+      if (backend_cfg.ok()) {
+        const auto ks =
```
sohaibiftikhar: auto -> explicit type
google cpp guidelines regulate that auto is only used when the type is clear from the RHS (for example GpuBackendConfig upstairs is clear).

xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -442,6 +474,26 @@ LatencyEstimator::TimeCost SolLatencyEstimator::GetLatencyBetween(
     return kLowLatency;
   }
 
+  // For sync intra-node AllReduce (Triton codegen or built-in custom C++
```
sohaibiftikhar: nit: same here with cpp.

xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -442,6 +474,26 @@ LatencyEstimator::TimeCost SolLatencyEstimator::GetLatencyBetween(
...
+  // kernel — both annotated as TRITON_ONE_SHOT / TRITON_TWO_SHOT) the kernel
+  // runs on the compute stream and fully blocks further compute — no overlap
+  // is possible between start and done.  NodeCost() already reports the full
+  // blocking latency on the critical path, so return kLowLatency here to
+  // avoid double-counting.
+  if (IsGPUSyncCollective(from.GetInstr())) {
+    auto cfg = from.GetInstr().backend_config<GpuBackendConfig>();
```
sohaibiftikhar: can we parse the config only once and use it in both places? This not quite free I think. Since IsGPUSyncCollective also parses the config. The same upstairs and downstairs.

xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -442,6 +474,26 @@ LatencyEstimator::TimeCost SolLatencyEstimator::GetLatencyBetween(
...
+    if (cfg.ok()) {
+      const auto ks = cfg->collective_backend_config().kernel_strategy();
+      if (ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_ONE_SHOT ||
+          ks == CollectiveBackendConfig::KERNEL_STRATEGY_TRITON_TWO_SHOT) {
```
sohaibiftikhar: Instead of repeating this condition in three places we could create a single method. IsTritonCollectiveKernel for instance.

xla/service/gpu/model/sol_latency_estimator_test.cc
```diff
@@ -880,5 +879,227 @@ TEST_F(IsSolLatencyEstimatorEnabledTest, DisabledForHopperWithHostOffloaded) {
...
+        /*collective_interpolator=*/nullptr);
+  }
+
+  // Creates a SolLatencyEstimator and calls NodeCost.
+  absl::Duration NodeCost(const HloInstruction& instr,
+                          const HloComputation* computation) {
+    auto estimator = *SolLatencyEstimator::Create(
+        scheduler_config_, std::make_unique<DummyLatencyEstimator>(),
+        gpu_device_info_, HloCostAnalysis::DefaultShapeSize, computation);
+    // Patch sol_flags_ into the estimator via Create's sol_config derivation:
```
sohaibiftikhar: this comment looks incorrect?

mfrancepillois: Thanks for noticing. The comment has been removed.

xla/service/gpu/model/sol_latency_estimator.cc
```diff
@@ -474,6 +530,29 @@ LatencyEstimator::TimeCost SolLatencyEstimator::NodeCost(
...
+  // Sync intra-node AllReduce kernels (Triton codegen, annotated as
+  // TRITON_ONE_SHOT / TRITON_TWO_SHOT) run on the compute stream and block
+  // compute entirely.  Their full latency is on the critical path, so we must
+  // report it here as NodeCost rather than as a hidden GetLatencyBetween.
+  if (hlo_query::IsAsyncCollectiveStartOp(instr, /*include_send_recv=*/true)) {
+    absl::StatusOr<GpuBackendConfig> cfg =
+        instr->backend_config<GpuBackendConfig>();
+    if (cfg.ok() && cfg->collective_backend_config().is_sync()) {
+      const auto ks = cfg->collective_backend_config().kernel_strategy();
+      if (IsTritonCollectiveKernel(ks)) {
```
sohaibiftikhar: nit: can also just inline this now (same upstairs).

sohaibiftikhar (approved):

PR #43774, opened 2026-06-05
[xla:gpu] Track SPMD-partitioner provenance on generated collectives

ezhulenev:
**NFC**:  only improving the quality of debug-only metadata on SPMD-partitioned XLA programs

### Motivation

The SPMD partitioner synthesizes collectives (all-reduce, all-gather, collective-permute, all-to-all) when resharding tensors across partitions. Downstream GPU passes — scheduling, backend assignment, and profiling — often need to treat these partitioner-generated collectives differently from collectives that originate in user code. Today there is no way to tell them apart after partitioning: once the collective is in the HLO graph, its origin is lost.

This change records that provenance so later passes can act on it.

###  What it does

- Adds a platform-independent frontend attribute `is_spmd_generated`, set to "true" on every collective the SPMD partitioner creates.
- Threads it through the SPMD partitioner's collective creators via a small helper that, in addition to tagging the collective, propagates the merged frontend attributes and metadata of the operand(s) the collective is derived from. This keeps source-location and user annotations attached to the synthesized op, which matters for debugging partitioned programs.
- Mirrors the signal into a GPU-specific `is_spmd_generated` field on `CollectiveBackendConfig`, and teaches the GPU collective combiner to OR-merge it across combined collectives (alongside the existing `is_pipelined` merge), so provenance survives collective combining.
- Adds `IsSpmdGenerated(const HloInstruction&)` to the GPU collective utils, which checks both the frontend attribute and the backend-config field.

Overall this is quality of life improvement for debugging partitioned JAX programs, as some collectives are explicit in the JAXPR, and some are added "magically" by XLA, debugging where magic comes from is super handy!


xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils.cc
```diff
@@ -100,17 +100,22 @@ absl::Status MergeCollectiveBackendConfig(
...
+    // IsSpmdGenerated checks both the frontend attribute (set by the SPMD
+    // partitioner) and the backend config field (set on earlier combines).
+    any_spmd_generated |= IsSpmdGenerated(*inst);
   }
-  if (any_pipelined) {
-    config.mutable_collective_backend_config()->set_is_pipelined(true);
-  }
+
+  config.mutable_collective_backend_config()->set_is_pipelined(any_pipelined);
+  config.mutable_collective_backend_config()->set_is_spmd_generated(
```
PatriosTheGreat, line 116: Could it also be useful to distinguish between combined collective where all operations are emitter by partitioner and the mixed one?

For example we can have a enum `collective_emitter` instead of is_spmd_generated attribute with 3 values:
user -- collective came from the JAX code
partitioner -- emitted by partitioner
compiler -- some pass in XLA (such as all_reduce_decomposer) emitted collective 
mixed -- combined collective with both types

PatriosTheGreat (commented):
Thanks, this looks very useful. I have only one nit/question.

xla/backends/gpu/transforms/collectives/gpu_collective_combiner_utils.cc
ezhulenev: Though about it, but it's primarily for end-user consumption who looks at xprof traces, and decided to keep it simple, I for example want to simply know if any of the collectives are SPMD generated or not, as we do mostly rely on shard map and want to spot cases when things mixed together.

PatriosTheGreat (approved):

dimitar-asenov:
Hi @ezhulenev, could you please rebase the PR on `HEAD`?

ezhulenev:
Rebased.

ermilovmaxim (approved):

PR #43897, opened 2026-06-08
[ROCm] Don't use async deallocaton in MIOpen autotuner backend

draganmladjenovic:
📝 Summary of Changes
Introduce file local OwningScratchAllocator implementation that does deallocation on destruction.

🎯 Justification
OwningScratchAllocator implementation moved to async deallocaton model which doesn't work for miopen backend.

🚀 Kind of Contribution
🐛 Bug Fix

### DIFF
--- a/xla/backends/gpu/autotuner/miopen.cc
+++ b/xla/backends/gpu/autotuner/miopen.cc
@@ -64,6 +64,32 @@ using MIOpenBackendConfig = stream_executor::dnn::AlgorithmProto;
 
 namespace {
 
+struct OwningScratchAllocator : public se::ScratchAllocator {
+  OwningScratchAllocator(int device_ordinal,
+                         se::DeviceAddressAllocator* allocator)
+      : device_ordinal_(device_ordinal), allocator_(allocator) {}
+
+  int64_t GetMemoryLimitInBytes() override { return -1; }
+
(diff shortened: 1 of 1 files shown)


xla/backends/gpu/autotuner/miopen.cc
```diff
@@ -64,6 +64,32 @@ using MIOpenBackendConfig = stream_executor::dnn::AlgorithmProto;
...
+struct OwningScratchAllocator : public se::ScratchAllocator {
+  OwningScratchAllocator(int device_ordinal,
+                         se::DeviceAddressAllocator* allocator)
+      : device_ordinal_(device_ordinal), allocator_(allocator) {}
+
+  int64_t GetMemoryLimitInBytes() override { return -1; }
+
+  absl::StatusOr<se::DeviceAddress<uint8_t>> AllocateBytes(
+      int64_t byte_size) override {
+    if (byte_size < 0) {
```
derdrdirk, line 76: Nit: Do you ever want to allocate 0 bytes?

derdrdirk (approved):
LGTM

xla/backends/gpu/autotuner/miopen.cc
draganmladjenovic: Don't think so. This can be TF_CHECK I guess.

PR #43907, opened 2026-06-08
[xla:gpu] Capture DUS fusion offset expressions

ezhulenev:
Adds offset-expression cloning to `DynamicSliceFusionRewriterV2`. The rewriter now clones supported scalar integer offset computations into the dynamic-slice fusion body instead of always capturing computed offsets as opaque fusion operands.

- Added `DynamicSliceFusion::Offset::IsExpr(const HloInstruction*)` as the shared predicate for offset expression compatibility.
- Updated V2 planning to clone offset expression interiors for dynamic-slice and dynamic-update-slice offsets.
- Keeps constants sunk into the fusion body and external scalar leaves as fusion parameters.
- Added tests covering DS/DUS offset cloning, select/compare offsets, shared offset expressions, unsupported offset roots, nested dynamic-slice boundaries, and `Offset::IsExpr` eligibility.

Leaving scalar offset computations outside of fusion body forces XLA to compile and launch tiny device kernels which is super wasteful.

### DIFF
--- a/xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
+++ b/xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
@@ -399,20 +399,18 @@ void AddOffsetExpressionInstructions(
     return;
   }
 
-  if (clone_instruction_set.contains(instr)) {
+  if (!clone_instruction_set.insert(instr).second) {
     return;
   }
 
+  offset_instruction_set.insert(instr);
   for (HloInstruction* operand : instr->operands()) {
     AddOffsetExpressionInstructions(operand, clone_instructions,
                                     clone_instruction_set,
                                     offset_instruction_set);
   }
 
-  if (clone_instruction_set.insert(instr).second) {
-    offset_instruction_set.insert(instr);
-    clone_instructions.push_back(instr);
-  }
+  clone_instructions.push_back(instr);
 }
 
 void AddDynamicSliceOffsetExpressionInstructions(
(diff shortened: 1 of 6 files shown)


xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -378,25 +384,67 @@ std::vector<SlicedResult> ResolveSlicedResults(
...
+    return;
+  }
+
+  for (HloInstruction* operand : instr->operands()) {
+    AddOffsetExpressionInstructions(operand, clone_instructions,
+                                    clone_instruction_set,
+                                    offset_instruction_set);
+  }
+
+  if (clone_instruction_set.insert(instr).second) {
```
akuegel: Shouldn't this be before the recursive call in line 407?
Otherwise we might visit the same instruction twice?

Also, I guess we could then use the `else` branch for the early return and don't need to use `contains()` call. And the `clone_instructions.push_back(instr);` line could still be after the for loop (unconditionally) to preserve post-order.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -410,63 +458,111 @@ std::optional<DynamicSliceFusionPlan> BuildFusionPlan(
...
+      return;
+    }
+    // Only cloned offset expressions are transparent for this traversal; other
+    // cloned instructions are already represented inside the fusion body.
+    if (!offset_instruction_set.contains(operand)) {
+      return;
+    }
+    // Recurse through offset expressions to find scalar leaves.
+    for (HloInstruction* offset_operand : operand->operands()) {
+      self(self, offset_operand);
```
akuegel, line 556: Do we need to be careful with this recursive call that we don't visit the same HloInstruction twice?

ezhulenev: Fixed recursive traversal and added a test that we can safely capture offset exprs with multiple uses of same hlo instruction.

akuegel (approved):

PR #43910, opened 2026-06-08
[ROCm] Fix missing ROCm branches in GpuFloatSupport::IsSupported for BF16

mminutoli:
## Problem

`GpuFloatSupport::IsSupported` had CUDA-only capability checks for several BF16 elementwise opcodes. On ROCm these always returned `false`, causing unnecessarily promote BF16 ops to F32 even on hardware that supports them natively.

All three bugs follow the same pattern: a CUDA capability check with no corresponding ROCm branch, so the function falls through to `return false`.

### `kAdd` / `kMultiply` / `kSubtract` + BF16

**Before:** `cuda_compute_capability->IsAtLeastHopper()` with no ROCm branch.

**Impact:** The scatter reducer body grew from 2 to 6 ops, causing `GetAtomicModifierParameters` to return `nullopt` and falling back to a 32-bit CAS spin-loop instead of `global_atomic_pk_add_bf16`. This produced a **30–66× slowdown** in `jax.ops.segment_sum` with BF16 on MI300+ hardware.

**Fix:** Return `rocm_compute_capability->has_packed_bf16_atomics_support()` (true on gfx942 / MI300, gfx950 / MI350, gfx12).

### `kExp` / `kLog` + BF16

**Before:** `compute_capability_.IsCuda()` with no ROCm branch.

**Impact:** BF16 exp/log always promoted to F32 on ROCm.

**Fix:** Return `rocm_compute_capability->has_bf16_dtype_support()` (true on MI100+, gfx11, gfx12).

### `kAbs` / `kMaximum` / `kMinimum` / `kNegate` + BF16

**Before:** `cuda_compute_capability->IsAtLeastAmpere()` with no ROCm branch.

**Impact:** BF16 elementwise unary/binary ops always promoted to F32 on ROCm.

**Fix:** Return `rocm_compute_capability->has_bf16_dtype_support()` (true on MI100+, gfx11, gfx12).

## Tests

New tests added in `gpu_float_support_test.cc`:

| Test | What it checks |
|---|---|
| `BF16LogAndExpOnRocmMI300IsNotNormalized` | gfx942: exp/log must not normalize |
| `Bf16UnaryOpRocmTest/IsNotNormalizedOnMI300` | gfx942: kNegate/kAbs must not normalize |
| `Bf16MinMaxOnRocmMI300IsNotNormalized` | gfx942: kMinimum/kMaximum must not normalize |
| `BF16AddOnRocmMI300IsNotNormalized` | gfx942: kAdd must not normalize (scatter fix) |
| `BF16AddOnRocmMI200IsNormalized` | gfx90a: kAdd correctly still normalizes (MI200 lacks packed BF16 atomics) |

The existing `BF16LogAndExpOnRocmIsNormalized` test (default/unknown ROCm arch) continues to pass, confirming no regression for pre-MI100 devices.

### DIFF
--- a/xla/service/gpu/gpu_float_support.cc
+++ b/xla/service/gpu/gpu_float_support.cc
@@ -141,7 +141,13 @@ bool GpuFloatSupport::IsSupported(const HloInstruction& hlo) const {
     case HloOpcode::kExp:
     case HloOpcode::kLog:
       if (LowPrecisionType() == BF16) {
-        return compute_capability_.IsCuda();
+        if (compute_capability_.IsCuda()) {
+          return true;
+        }
+        if (auto* rocm_compute_capability =
+                compute_capability_.rocm_compute_capability()) {
+          return rocm_compute_capability->has_bf16_dtype_support();
+        }
       }
       return false;
     case HloOpcode::kAbs:
@@ -149,20 +155,28 @@ bool GpuFloatSupport::IsSupported(const HloInstruction& hlo) const {
     case HloOpcode::kMinimum:
     case HloOpcode::kNegate:
       if (LowPrecisionType() == BF16) {
-        auto* cuda_compute_capability =
-            compute_capability_.cuda_compute_capability();
-        return cuda_compute_capability != nullptr &&
-               cuda_compute_capability->IsAtLeastAmpere();
+        if (auto* cuda_compute_capability =
+                compute_capability_.cuda_compute_capability()) {
+          return cuda_compute_capability->IsAtLeastAmpere();
+        }
+        if (auto* rocm_compute_capability =
+                compute_capability_.rocm_compute_capability()) {
+          return rocm_compute_capability->has_bf16_dtype_support();
+        }
       }
       return false;
     case HloOpcode::kAdd:
     case HloOpcode::kMultiply:
     case HloOpcode::kSubtract: {
       if (LowPrecisionType() == BF16) {
-        auto* cuda_compute_capability =
-            compute_capability_.cuda_compute_capability();
-        return cuda_compute_capability != nullptr &&
-               cuda_compute_capability->IsAtLeastHopper();
+        if (auto* cuda_compute_capability =
+                compute_capability_.cuda_compute_capability()) {
+          return cuda_compute_capability->IsAtLeastHopper();
+        }
+        if (auto* rocm_compute_capability =
+                compute_capability_.rocm_compute_capability()) {
+          return rocm_compute_capability->has_packed_bf16_atomics_support();
+        }
       }
       return false;
     }
--- a/xla/service/gpu/gpu_float_support_test.cc
+++ b/xla/service/gpu/gpu_float_support_test.cc
@@ -633,6 +633,7 @@ ENTRY main {
 }
 
 TEST_F(FloatSupportTest, BF16LogAndExpOnRocmIsNormalized) {
+  // Default/unknown ROCm arch has no bf16 dtype support → normalize.
   auto cc = se::RocmComputeCapability();
   constexpr absl::string_view kHloModule = R"(
 HloModule module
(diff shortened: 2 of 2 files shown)


i-chaochen (approved):

xla/service/gpu/gpu_float_support.cc
```diff
@@ -141,28 +141,42 @@ bool GpuFloatSupport::IsSupported(const HloInstruction& hlo) const {
     case HloOpcode::kExp:
     case HloOpcode::kLog:
       if (LowPrecisionType() == BF16) {
-        return compute_capability_.IsCuda();
+        if (compute_capability_.IsCuda()) {
+          return true;
+        }
+        if (auto* rocm_compute_capability =
+                compute_capability_.rocm_compute_capability()) {
+          return rocm_compute_capability->has_bf16_dtype_support();
```
draganmladjenovic, line 149: None of this make sense. Every rocm target has emulated BF16. Only gfx1250 so far has real instructions for some of these operations. I'd rather focus on making the optimization that is failing be able to unwrap bf16 -> fp32 -> bf16 conversions.

draganmladjenovic (changes requested):

xla/service/gpu/gpu_float_support.cc
```diff
@@ -141,28 +141,42 @@ bool GpuFloatSupport::IsSupported(const HloInstruction& hlo) const {
...
       if (LowPrecisionType() == BF16) {
-        auto* cuda_compute_capability =
-            compute_capability_.cuda_compute_capability();
-        return cuda_compute_capability != nullptr &&
-               cuda_compute_capability->IsAtLeastAmpere();
+        if (auto* cuda_compute_capability =
+                compute_capability_.cuda_compute_capability()) {
+          return cuda_compute_capability->IsAtLeastAmpere();
+        }
+        if (auto* rocm_compute_capability =
```
draganmladjenovic, line 162: For abs and negate is true for anything under the sun, including cuda targets. It is just bit twidling. Minimum and maximum are a bit complicated llvm backend has a chance to do better than bf16 -> fp32 / min/ fp32 -> bf16, but it used to fail to do so.

i-chaochen (changes requested):
Thanks Dragan's check!

Yes, after internal discusion/investigation, I would suggest to keep the existing bf16 intact and to narrow down the actual performance issue.

Otherwise, this may simply push the BF16→F32→BF16 expansion later into LLVM/ROCm codegen, rather than removing it. That may be neutral, worse, or backend-dependent.


IIUC this motivation is to have the native bf16 arithmetic instruction instead of `atomicAdd` as a CAS loop, which I guess it's why you could see the perf boost with this patch? but I believe we shouldn't misunderstand others to claim we have this native bf16 instruction on https://github.com/openxla/xla/pull/43910/changes#diff-b7c8a31fea54817f336e75b51e61e09440b347fa4c37397a6fddf54cc668a5f7R140-R179

- Although we indeed have bf16 MFMA/WMMA matrix instructions https://clang.llvm.org/docs/AMDGPUBuiltinReference.html 

Essentially, you should only reference [has_packed_bf16_atomics_support()](https://github.com/openxla/xla/blob/main/xla/stream_executor/rocm/rocm_compute_capability.h#L180-L182)  for bf16 path, this `has_bf16_dtype_support` just means BF16 is a legal value type; individual arithmetic Ops still need target-specific native support, otherwise they are normalized or lowered through F32, so this is why we shall target to the actual performance issue.

We have bf16 exp on gfx1250 https://lists.llvm.org/pipermail/all-commits/Week-of-Mon-20250714/235310.html

i-chaochen:
As a side note, I would suggest next time to put PR and against `rocm/xla:main` first, which is same periodically synced mirror main branch, and you can put `claude-review` label to have local review with our review skills there. 

https://github.com/ROCm/xla/tree/rocm-dev-infra/.claude/skills

https://github.com/ROCm/xla#opt-in-pr-labels

draganmladjenovic:
https://github.com/openxla/xla/blob/main/xla/codegen/emitters/transforms/atomic_rmw_utils.cc#L107

xla/service/gpu/gpu_float_support.cc
mminutoli: well, but that needs to be sorted in `has_bf16_dtype_support` according to the device, doesn't it?

Arech8: @draganmladjenovic , but even with that, if that works faster than before, why not use it? Why can't it stay until a better fix is developed, tested and merged?

draganmladjenovic:
https://github.com/openxla/xla/pull/44428

PR #43912, opened 2026-06-09
[tsl] Add support for spatial partitioning of BFC-managed address range

ezhulenev:
Adds an opt-in spatial partitioning mode to `BFCAllocator` that serves allocations from lower and upper ends of a single preallocated address range. `AllocationEnd::kLower` requests grow up from the low address,
  `AllocationEnd::kUpper` requests grow down from the high address, and both ends share one central gap.

```
  low address                                             high address
  |-------------------------------------------------------------------|

   lower-end owned --->        central gap        <--- upper-end owned
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

  This is not a static partition. The split is dynamic: both ends grow and shrink over time, freed space at either end of the central gap coalesces back into the gap, and either end may reclaim it later. The invariant is spatial:
  lower-end-owned chunks and upper-end-owned chunks never overlap, and allocation fails only when the central gap is exhausted.

Upper-end placement must be reproducible for symmetric collective buffers. Because lower-end and upper-end allocations never reuse each other’s interior holes, upper-end placement is a pure function of the upper-end request stream and is not perturbed by lower-end activity, except for true exhaustion of the central gap. This lets symmetric buffers land at identical offsets across ranks while still drawing from the same preallocated BFC pool as ordinary device buffers.

Each BFC allocator `Chunk` carries a `ChunkTag`:

  - `kLower`: lower-end-owned chunk, either in use or a lower-end reusable interior hole.
  - `kUpper`: upper-end-owned chunk, either in use or an upper-end reusable interior hole.
  - `kCentralGap`: the single central gap between lower-end and upper-end ownership.

Free interior holes remain in the existing size-class bins and are searched only by matching tag. The central gap is tracked separately by `central_gap_` and is not inserted into bins. Allocation first tries same-tag binned holes, then carves from the central gap. Lower-end allocations carve upward; upper-end allocations align downward and carve from the high end.

Coalescing preserves same-tag interior holes, but boundary frees can rejoin the central gap. Adjacent same-tag holes then cascade back into the gap, keeping the split dynamic without allowing one end to consume the other end’s interior holes.

The GPU PJRT path gates this behind `xla_gpu_enable_allocator_spatial_partitioning` and requires BFC preallocation. When enabled, one BFC allocator backs both default memory and collective memory: default allocations use the lower end, while collective/symmetric allocations use the upper end.

### DIFF
--- a/xla/tsl/framework/BUILD
+++ b/xla/tsl/framework/BUILD
@@ -232,8 +232,8 @@ tsl_cc_test(
         "//xla/tsl/platform:test_benchmark",
         "//xla/tsl/platform:test_main",
         "@com_google_absl//absl/base",
+        "@com_google_absl//absl/base:no_destructor",
         "@com_google_absl//absl/synchronization",
-        "@tsl//tsl/platform:platform_port",
     ],
 )
 
--- a/xla/tsl/framework/bfc_allocator.cc
+++ b/xla/tsl/framework/bfc_allocator.cc
@@ -60,11 +60,16 @@ BFCAllocator::BFCAllocator(std::unique_ptr<SubAllocator> sub_allocator,
                            size_t total_memory, const std::string& name,
                            const Options& opts)
     : opts_(opts),
+      free_chunk_tag_(opts.enable_spatial_partitioning ? ChunkTag::kCentralGap
+                                                       : ChunkTag::kLower),
       coalesce_regions_(sub_allocator->SupportsCoalescing()),
       sub_allocator_(std::move(sub_allocator)),
       name_(name),
-      free_chunks_list_(kInvalidChunkHandle),
+      unused_chunk_handle_head_(kInvalidChunkHandle),
       next_allocation_id_(1) {
+  CHECK(!opts.enable_spatial_partitioning || !opts.allow_growth)  // Crash OK
+      << "Spatial partitioning requires a single fixed address range "
+         "(allow_growth=false).";
   if (opts.allow_growth) {
     // 2MiB smallest initial allocation, unless total memory available
     // is less.
@@ -82,6 +87,13 @@ BFCAllocator::BFCAllocator(std::unique_ptr<SubAllocator> sub_allocator,
   // Allocate the requested amount of memory.
   memory_limit_ = total_memory;
   stats_.bytes_limit = static_cast<int64_t>(total_memory);
+
+  // Cap on how much a chunk may exceed the requested size before we split it.
+  // If the user did not set a fraction, default to 128MB.
+  max_internal_fragmentation_bytes_ =
+      (opts.fragmentation_fraction > 0.0)
+          ? opts.fragmentation_fraction * memory_limit_
+          : 128 << 20;
 
   // Create a bunch of bins of various good sizes.
 
@@ -210,6 +222,7 @@ bool BFCAllocator::Extend(size_t alignment, size_t rounded_bytes) {
   c->prev = kInvalidChunkHandle;
   c->next = kInvalidChunkHandle;
   c->freed_at_count = 0;
+  c->tag = free_chunk_tag_;
 
   region_manager_.set_handle(c->ptr, h);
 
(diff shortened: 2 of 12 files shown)


ermilovmaxim (approved):
preliminary PR import, to test against google3

xla/xla.proto
```diff
@@ -636,6 +636,11 @@ message DebugOptions {
   // Whether to dump llvm ir when compiling to ptx.
   optional bool xla_gpu_dump_llvmir = 155;
 
+  // Enables spatial partitioning of the GPU BFC allocator so default and
+  // collective allocations share one fixed address range. Requires BFC
+  // preallocation.
+  optional bool xla_gpu_enable_allocator_spatial_partitioning = 493;
```
PatriosTheGreat: This should be sorted, could you place this after xla_gpu_enable_all_gather_combine_by_dim ?

xla/tsl/framework/BUILD
```diff
@@ -232,6 +232,7 @@ tsl_cc_test(
         "//xla/tsl/platform:test_benchmark",
         "//xla/tsl/platform:test_main",
         "@com_google_absl//absl/base",
+        "@com_google_absl//absl/base:no_destructor",
         "@com_google_absl//absl/synchronization",
         "@tsl//tsl/platform:platform_port",
```
PatriosTheGreat: Since include "tsl/platform/mem.h" was removed this is not needed anymore

xla/tsl/framework/bfc_allocator.cc
```diff
@@ -816,9 +1011,54 @@ void BFCAllocator::DeleteChunk(ChunkHandle h) {
...
+void BFCAllocator::InsertFreeChunk(BFCAllocator::ChunkHandle h) {
+  Chunk* c = ChunkFromHandle(h);
+  CHECK(!c->in_use() && (c->bin_num == kInvalidBinNum));  // Crash OK
+  if (c->tag == ChunkTag::kCentralGap) {
+    CHECK_EQ(central_gap_, kInvalidChunkHandle)  // Crash OK
+        << "spatial partitioning expects one central gap";
+    central_gap_ = h;
+    return;
+  }
+  InsertFreeChunkIntoBin(h);
```
PatriosTheGreat: I guess this is now the only place where InsertFreeChunkIntoBin is getting called. Can we simply inline the method here?

dimitar-asenov:
@ezhulenev could you please fix these two linter issues:

```
CheckLint found errors.
These lines are out of order.
	xla/xla.proto:349-1222
Redundant blank line at the end of a code block should be deleted.  [whitespace/blank_line] [3]
	xla/debug_options_flags.cc:3271
```

xla/tsl/framework/bfc_allocator.cc
```diff
@@ -580,115 +653,222 @@ void BFCAllocator::AddTraceMe(absl::string_view traceme_name,
...
+void* BFCAllocator::AllocateChunkFromHighEnd(ChunkHandle h,
+                                             size_t rounded_bytes,
+                                             size_t num_bytes,
+                                             size_t alignment) {
+  Chunk* chunk = ChunkFromHandle(h);
+
+  uintptr_t chunk_start = absl::bit_cast<uintptr_t>(chunk->ptr);
+  const uintptr_t aligned_start =
+      HighEndAlignedStart(chunk_start, chunk->size, rounded_bytes, alignment);
+  const size_t prefix_size = aligned_start - chunk_start;
```
PatriosTheGreat, line 808: Should we check here that aligned_start is >= chunk_start?
I guess this should be guaranteed by the caller, but having additional check might be useful

PatriosTheGreat (changes requested):

xla/xla.proto
ezhulenev: Done

xla/tsl/framework/bfc_allocator.cc
ezhulenev: Done

xla/tsl/framework/bfc_allocator.cc
ezhulenev: Done as well, agree it's better to crash loudly then to debug models producing garbage.

xla/tsl/framework/BUILD
ezhulenev: Deleted.

dimitar-asenov (approved):

dimitar-asenov:
@ermilovmaxim @PatriosTheGreat @olegshyshkov Could you please look at the internal CL and provide some feedback based on it?

xla/tsl/framework/bfc_allocator.cc
```diff
@@ -344,6 +367,31 @@ size_t BFCAllocator::RoundedBytes(size_t bytes) {
   return rounded_bytes;
 }
 
+uintptr_t BFCAllocator::AlignUp(uintptr_t ptr, size_t alignment) {
```
PatriosTheGreat, line 370: Nit: Might be useful to add:

```
  CHECK(absl::has_single_bit(alignment))
      << "alignment must be a power of 2, got " << alignment;
```

Or mention in the comment this requirement

xla/xla.proto
```diff
@@ -640,6 +640,11 @@ message DebugOptions {
   // dimension.
   optional bool xla_gpu_enable_all_gather_combine_by_dim = 254;
 
+  // Enables spatial partitioning of the GPU BFC allocator so default and
+  // collective allocations share one fixed address range. Requires BFC
+  // preallocation.
+  optional bool xla_gpu_enable_allocator_spatial_partitioning = 493;
```
PatriosTheGreat: You mighty need to rebase and change the number of this property since another one was added recently.

PatriosTheGreat (approved):

ezhulenev:
Rebased

xla/xla.proto
ezhulenev: Done

xla/tsl/framework/bfc_allocator.cc
ezhulenev: Done

ermilovmaxim (approved):

olegshyshkov (approved):

PR #43945, opened 2026-06-09
[GPU] Add non-invariant annotation mechanism for kernel arguments.

sergachev:
📝 Summary of Changes
Introduces `xla.no_invariant_operands` frontend attribute and wires it through GPU kernel argument handling so that marked read-only operands can opt out of the invariant path (non-coherent `ld.global.nc` loads).

This change alone is not expected to change any observable compiler behavior.

🎯 Justification
This is required to use PDL launch instructions which can introduce non-trivial overlaps of kernels. When kernels with data dependencies overlap, the existing analysis of kernels in isolation is not sufficient to correctly derive invariant arguments.

🚀 Kind of Contribution
✨ New Feature

🧪 Unit Tests:
yes

🧪 Execution Tests:
no


xla/codegen/emitters/kernel_arguments.cc
```diff
@@ -152,10 +154,19 @@ absl::StatusOr<KernelArguments> CreateKernelArguments(
     const HloInstruction* hlo_instruction,
     absl::Span<const Shape> unmanaged_arguments) {
   std::vector<KernelArgument> kernel_arguments;
+  absl::flat_hash_set<int> no_invariant_operands =
+      NonInvariantOperands(*hlo_instruction);
+
+  int op_idx = 0;
```
akuegel: You can use `llvm::enumerate()` for iterating through operands. Or in the loop use kernel_arguments.size() instead of `op_idx`.

xla/codegen/emitters/kernel_api_builder.cc
```diff
@@ -171,10 +172,14 @@ absl::StatusOr<mlir::func::FuncOp> EmitKernelApi(
     attrs.push_back(builder.getNamedAttr(
         mlir::LLVM::LLVMDialect::getDereferenceableAttrName(),
         builder.getIndexAttr(arg.slice().size())));
-    if (!arg.written()) {
+    if (!arg.written() && arg.invariant()) {
```
akuegel: Personally I would prefer to have the check for !arg.written in an outer if block, and inside use `if (arg.invariant()) { ... } else { ...}`.

xla/codegen/emitters/kernel_arguments_test.cc
```diff
@@ -383,5 +383,39 @@ TEST_F(KernelArgumentsTest, UnmanagedArguments) {
                           kUnmanaged, kUnmanaged));
 }
 
+TEST_F(KernelArgumentsTest,
+       NoInvariantFrontendAnnotationPropagatesToArguments) {
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
```
akuegel: Please use `ASSERT_OK_AND_ASSIGN` (without TF prefix). Also below in line 399 and 406.

xla/codegen/emitters/kernel_api_builder_test.cc
```diff
@@ -78,5 +91,34 @@ TEST(DefaultWorkItemIndexingMap, MultiDimensionTile) {
   EXPECT_EQ(symbolic_map.GetResult(3), tile_sym_z);
 }
 
+using KernelApiBuilderTest = HloHardwareIndependentTestBase;
+
+TEST_F(KernelApiBuilderTest, NoInvariantOperandsFrontendAttribute) {
+  TF_ASSERT_OK_AND_ASSIGN(std::unique_ptr<VerifiedHloModule> module,
```
akuegel: Please use ASSERT_OK_AND_ASSIGN (without TF prefix). Also below in line 105 and 115.

xla/frontend_attributes.h
```diff
@@ -33,6 +33,13 @@ void SetDisjointReadWriteRegionsAttr(HloInstruction* instruction);
 // frontend attribute set (returns false otherwise).
 bool HasDisjointReadWriteRegionsAttr(HloInstruction* instruction);
 
+// Comma-separated list of operand indices that should not be assumed invariant.
```
akuegel: This comment is a bit confusing. This is actually just the name for the frontend attribute, and the attribute itself is the comma-separated list, right?

akuegel (approved):

PR #44044, opened 2026-06-10
[XLA:GPU] Emit dynamic slice copy thunks from analysis

shawnwang18:
📝 Summary of Changes
Adds a read-only `dynamic_slice_copy` analysis for static slice and raw DS/DUS copy fusions, and uses that analysis from `ThunkEmitter` to emit specialized D2D copy thunk sequences without materializing temporary HLO copy instructions.

This also removes `FusionDynamicMemcpyRewriter` and its fusion-dispatch pipeline wiring now that the emitter can recognize the copy cases directly.

🎯 Justification
This keeps the dynamic-slice copy classification as analysis-only and keeps the thunk emitter responsible for choosing the specialized thunk sequence. It avoids creating temporary IR solely to match copy patterns.

🚀 Kind of Contribution
♻️ Cleanup, 🧪 Tests

Not applicable. This PR is the analysis/thunk-emission cleanup layer; async overlap support is split into the stacked follow-up PR.

🧪 Unit Tests:
- `//xla/backends/gpu/transforms:dynamic_slice_copy_test`

🧪 Execution Tests:
- `//xla/backends/gpu/tests:gpu_copy_test` with filter `GpuCopyTest.UseMemcpyForTrivialStaticSliceFusion:GpuCopyTest.UseMemcpyForDynamicSlice:GpuCopyTest.UseMemcpyForDynamicUpdateSlice:GpuCopyTest.UseMemcpyForDynamicUpdateSliceWithBitcasts:GpuCopyTest.UseDynamicMemcpyIntegrationTest`


xla/backends/gpu/transforms/dynamic_slice_copy_fusion_async_wrapper_test.cc
```diff
@@ -0,0 +1,99 @@
...
+    })";
+
+  ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> module,
+                       ParseAndReturnVerifiedModule(kHlo));
+
+  EXPECT_THAT(FusionDynamicMemcpyRewriter().Run(module.get()),
+              absl_testing::IsOkAndHolds(true));
+
+  DynamicSliceCopyFusionAsyncWrapper wrapper;
+  EXPECT_THAT(wrapper.Run(module.get()), absl_testing::IsOkAndHolds(true));
```
ezhulenev: I find CHECK-based tests easier to read, jsut outline the new strucutre of IR with an asyhnc root

xla/service/gpu/gpu_compiler.cc
```diff
@@ -3121,6 +3121,16 @@ absl::Status GpuCompiler::RunPreSchedulingPasses(
...
   HloPassPipeline pipeline("pre-scheduling-passes");
   pipeline.AddPass<FusionWrapper>(gpu_device_info);
+  const auto* cuda_cc =
+      gpu_device_info.gpu_compute_capability().cuda_compute_capability();
+  if (cuda_cc != nullptr && cuda_cc->IsAtLeastAmpere()) {
+    // FusionDispatchPipeline also runs FusionDynamicMemcpyRewriter after
+    // scheduling for fusions created by post-scheduling passes. Run it here as
+    // well so copy-hero dynamic slice fusions exist before LHS and can be
+    // wrapped as async memcpy.
+    pipeline.AddPass<FusionDynamicMemcpyRewriter>();
```
ezhulenev: One problem with this approach is that I want to delete `FusionDynamicMemcpyRewriter`, see https://github.com/openxla/xla/pull/43854, we don't really need a pass to tag fusiosn that look like mempcyh with a custom type, we can simply pattern-match it in thunk emitter and emit copy thunks.

My goal is to intercept everything that looks like memcpyu in think emitter and emit either a regular copy thunk or a copyu wrapped into DUSV2.

What would it mean for `DynamicSliceCopyFusionAsyncWrapper`? Should the host offloading directly insert dus fusions? Can it?

xla/backends/gpu/transforms/dynamic_slice_copy_fusion_async_wrapper_test.cc
shawnwang18: Fixed.

xla/service/gpu/gpu_compiler.cc
shawnwang18: Made the fix, now the AsyncStart body is the DUS + COPY instruction.  It still requires `FusionDynamicMemcpyRewriter`  to lower the DUS + COPY into DynamicSliceV2Thunk in this PR. 

Next PR we can remove   `FusionDynamicMemcpyRewriter`  and use pattern matching when emitting the thunk.

xla/backends/gpu/transforms/dynamic_slice_copy_fusion_async_wrapper.cc
```diff
@@ -0,0 +1,71 @@
...
+        continue;
+      }
+
+      // Use the same async-start context shape as the existing generic async
+      // wrapper. LHS classifies this pair as async memcpy, not async compute.
+      ASSIGN_OR_RETURN(
+          HloInstruction * async_done,
+          computation->CreateAsyncInstructions(
+              instruction, {ShapeUtil::MakeScalarShape(U32)},
+              HloInstruction::kMainExecutionThread, /*replace=*/true));
```
akuegel: `kMainExecutionThread` for `async_execution_thread` and `true` for `replace` are the defaults, so not needed to pass them here.

xla/backends/gpu/transforms/dynamic_slice_copy_fusion_async_wrapper.cc
```diff
@@ -0,0 +1,71 @@
...
+    std::vector<HloInstruction*> instructions =
+        computation->MakeInstructionPostOrder();
+    for (HloInstruction* instruction : instructions) {
+      if (!IsDynamicSliceMemcpyFusion(instruction)) {
+        continue;
+      }
+
+      // Use the same async-start context shape as the existing generic async
+      // wrapper. LHS classifies this pair as async memcpy, not async compute.
+      ASSIGN_OR_RETURN(
```
akuegel: Given you are not using the variable `async_done`, you can use `RETURN_IF_ERROR` instead. Full code:

`RETURN_IF_ERROR(computation->CreateAsyncInstructions(instruction, {ShapeUtil::MakeScalarShape(U32)}).status());`

xla/backends/gpu/transforms/dynamic_slice_fusion.cc
```diff
@@ -204,6 +205,153 @@ static std::optional<DynamicSliceConfig> ExtractDynamicSliceConfig(
...
+    return MemcpyFusionCandidate{ds_or_dus, ds_or_dus->operand(1)};
+  }
+
+  return std::nullopt;
+}
+
+static bool CanLowerMemcpyFusionCandidate(
+    const DynamicSliceFusion::MemcpyFusionCandidate& candidate) {
+  auto resolve_copy_hero_parameters = [](const HloInstruction* operand) {
+    std::unique_ptr<HloInstruction> copy =
```
akuegel: I think this is somewhat of a hack. As far as I can tell, you are creating this copy simply to be able to reuse `ResolveParameter()` method in case that `candidate.copy_operand` is a parameter itself. Can the method `ResolveParameter()` be adjusted to handle that? @ezhulenev what do you think?

xla/backends/gpu/transforms/dynamic_slice_copy_fusion_async_wrapper.cc
shawnwang18: Fixed

xla/backends/gpu/transforms/dynamic_slice_copy_fusion_async_wrapper.cc
shawnwang18: Fixed.

xla/backends/gpu/transforms/dynamic_slice_fusion.cc
ezhulenev: Yes, I also don;t like it and in my earlier rewriter tried very hard NOT to do this.

xla/backends/gpu/transforms/dynamic_slice_fusion.h
```diff
@@ -206,6 +206,21 @@ struct DynamicSliceFusion {
   static absl::StatusOr<std::vector<Result>> ResolveResults(
       const HloInstruction* hero);
 
+  // Raw DS/DUS-root fusion that can be converted to a copy-hero
+  // dynamic-slice fusion.
+  struct MemcpyFusionCandidate {
```
ezhulenev: Let's extract the the `MemcpyFusionCandidate` + all the machinery to pattern match copy-like DUS-fusions into `dynamic_slice_copy.h` library and rewrite existing rewriter on top of it as step 1. Then `dynamic_slice_copy_async_wrapper` can use this new analysi to detect fusions that are copy fusions and wrap them into async.

Also with a memcpy fusion rewriter we don't need to add any copies to the body, thunk emitter should use `dynamic_slice_copy` library to detect copy-like fusions and special-handle it like we do static slices, simply treturn a thunk sequence with one D2D copy + buffer allocations trick.

shawnwang18: Done with that, check  again?

xla/service/gpu/thunk_emitter.h
```diff
@@ -154,8 +156,11 @@ class ThunkEmitter {
 
   AsyncThunkSequence EmitFusion(const HloFusionInstruction* instr);
 
-  absl::StatusOr<std::optional<ThunkSequence>> TryEmitTrivialSliceFusion(
-      const HloFusionInstruction* instr);
+  AsyncThunkSequence EmitDynamicSliceCopyFusion(
+      const HloFusionInstruction* instr, DynamicSliceCopyFusion copy);
+
+  absl::StatusOr<ThunkSequence> EmitStaticSliceCopyFusion(
```
ezhulenev: Let's also return AsyncThunkSequence here for consistency, I missed it in my original PR.

ezhulenev (approved):

ezhulenev (approved):

akuegel (approved):

PR #44047, opened 2026-06-10
[ROCm] redirect C64/C128 matmuls to rocBLAS at runtime

magaonka-amd:
📝 Summary of Changes
After XLA removed the cublas/rocblas backend, ROCm has no way to compute C64/C128 GEMMs because hipBLASLt does not support complex GEMMs yet.

This PR redirects complex GEMM calls to rocBLAS at runtime. hipBLASLt support for complex GEMMs is in progress in ROCm; once that lands, I'll version-gate this code.

🎯 Justification
This  fix will only act as stopgap between now and when hipblasLT complex number support rolls out ( which is coming in ROCK 7.14 , approx in a month ). In that period we are very inclined to still support this. We are also fully retiring rocblas soon so there is very little benefit to coming up with rocblas custom calls exclusively for ROCm.

🚀 Kind of Contribution
 🐛 Bug Fix

🧪 Unit Tests:
just need to re-enable tests with complex gemm once this PR merges.

Testing:
- Tested 222 complex GEMM tests in JAX pytest and 7 bazel tests using complex GEMM.
- From the XLA side only tested locally. Once this PR merges we will need to re-enable the skipped complex GEMM tests in XLA.

### DIFF
--- a/xla/backends/gpu/transforms/gemm_rewriter.cc
+++ b/xla/backends/gpu/transforms/gemm_rewriter.cc
@@ -2592,6 +2592,13 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
         {ComputationType::kF64, DataType::kDouble, PrimitiveType::F64,
          PrimitiveType::F64, DataType::kDouble},
 
+        // Complex types: hipBLASLt has no complex GEMM kernels, so the ROCm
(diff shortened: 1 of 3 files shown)


derdrdirk (approved):
LGTM

PR #44083, opened 2026-06-11
Fix: validate string element count against shape in MakeArrayFromHost…

arunmanni-ai:
## Summary of Changes

Add validation that the number of deserialized string elements matches
`shape.num_elements()` before passing to `MakeArrayFromHostBuffer`.
Returns `InvalidArgumentError` on mismatch.

Apply the same check in `ParseMakeArraysFromHostBufferShardsSpecHostBufferProto`
for the `kString` branch.

## Justification

Without this check, a caller can supply a host buffer containing N string
elements alongside a shape with `num_elements() > N`, causing the consuming
loop in `pjrt_client.cc` to read past the end of the deserialized vector.

## Kind of Contribution

Bug Fix

## Unit Tests

Existing tests in `ifrt_backend_test.cc` cover the success path with matching
shape and buffer sizes and continue to pass. New mismatch cases return
`InvalidArgumentError` as expected.

## Reference

Issue Tracker- 522501454

### DIFF
--- a/xla/python/ifrt_proxy/server/ifrt_backend.cc
+++ b/xla/python/ifrt_proxy/server/ifrt_backend.cc
@@ -110,7 +110,7 @@ absl::StatusOr<IfrtArrayRef> MakeStringArrayFromHostBuffer(
   if (static_cast<size_t>(num_elements) != string_host_buffer.size()) {
     return absl::InvalidArgumentError(absl::StrCat(
         "String host buffer has ", string_host_buffer.size(),
-        " element(s) but shape requires ", num_elements, " element(s)"));
+        " elements but shape requires ", num_elements, " elements"));
   }
   const void* data = string_host_buffer.data();
 
@@ -163,13 +163,13 @@ ParseMakeArraysFromHostBufferShardsSpecHostBufferProto(
   if (dtype.kind() == DType::kString) {
     ASSIGN_OR_RETURN(std::vector<absl::Cord> string_host_buffer,
                      DeserializeStringHostBufferFromString(*host_buffer));
-    const int64_t num_elements_shards = shape.num_elements();
-    if (static_cast<size_t>(num_elements_shards) !=
+    const int64_t num_elements = shape.num_elements();
+    if (static_cast<size_t>(num_elements) !=
             string_host_buffer.size()) {
       return absl::InvalidArgumentError(absl::StrCat(
           "String host buffer has ", string_host_buffer.size(),
-          " element(s) but shape requires ", num_elements_shards,
-          " element(s)"));
+          " elements but shape requires ", num_elements,
+          " elements"));
     }
     data = string_host_buffer.data();
     on_done_with_host_buffer = [host_buffer = std::move(host_buffer),
--- a/xla/python/ifrt_proxy/server/ifrt_backend.cc
+++ b/xla/python/ifrt_proxy/server/ifrt_backend.cc
@@ -107,8 +107,7 @@ absl::StatusOr<IfrtArrayRef> MakeStringArrayFromHostBuffer(
   ASSIGN_OR_RETURN(std::vector<absl::Cord> string_host_buffer,
                    DeserializeStringHostBufferFromString(*host_buffer));
   const int64_t num_elements = shape.num_elements();
-  if (num_elements < 0 ||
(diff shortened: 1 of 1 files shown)


arunmanni-ai:
Please let me know if any changes or improvements are needed. I’m happy to update the implementation.

arunmanni-ai:
Hi @chsigg , Do you want me to change or improve anything ?

xla/python/ifrt_proxy/server/ifrt_backend.cc
```diff
@@ -106,6 +106,13 @@ absl::StatusOr<IfrtArrayRef> MakeStringArrayFromHostBuffer(
     ShardingRef sharding) {
   ASSIGN_OR_RETURN(std::vector<absl::Cord> string_host_buffer,
                    DeserializeStringHostBufferFromString(*host_buffer));
+  const int64_t num_elements = shape.num_elements();
+  if (num_elements < 0 ||
```
chsigg: I would skip the non-negative check.

xla/python/ifrt_proxy/server/ifrt_backend.cc
```diff
@@ -157,6 +164,15 @@ ParseMakeArraysFromHostBufferShardsSpecHostBufferProto(
   if (dtype.kind() == DType::kString) {
     ASSIGN_OR_RETURN(std::vector<absl::Cord> string_host_buffer,
                      DeserializeStringHostBufferFromString(*host_buffer));
+    const int64_t num_elements_shards = shape.num_elements();
+    if (num_elements_shards < 0 ||
```
chsigg: Ditto

chsigg (changes requested):
Thanks! Just a little nit, I will approve afterwards.

xla/python/ifrt_proxy/server/ifrt_backend.cc
arunmanni-ai: Done, removed the non-negative check. The static_cast<size_t> comparison already handles that case implicitly.

chsigg (approved):

chsigg:
Thanks!

xla/python/ifrt_proxy/server/ifrt_backend.cc
```diff
@@ -157,6 +163,14 @@ ParseMakeArraysFromHostBufferShardsSpecHostBufferProto(
   if (dtype.kind() == DType::kString) {
     ASSIGN_OR_RETURN(std::vector<absl::Cord> string_host_buffer,
                      DeserializeStringHostBufferFromString(*host_buffer));
+    const int64_t num_elements_shards = shape.num_elements();
```
hyeontaek: This is a bit unusual naming because we usually use `shard_` to indicate a per-shard property. There is no confusion in this function, so we can simply use `num_elements`.

xla/python/ifrt_proxy/server/ifrt_backend.cc
```diff
@@ -106,6 +106,12 @@ absl::StatusOr<IfrtArrayRef> MakeStringArrayFromHostBuffer(
     ShardingRef sharding) {
   ASSIGN_OR_RETURN(std::vector<absl::Cord> string_host_buffer,
                    DeserializeStringHostBufferFromString(*host_buffer));
+  const int64_t num_elements = shape.num_elements();
+  if (static_cast<size_t>(num_elements) != string_host_buffer.size()) {
+    return absl::InvalidArgumentError(absl::StrCat(
+        "String host buffer has ", string_host_buffer.size(),
+        " element(s) but shape requires ", num_elements, " element(s)"));
```
hyeontaek: Nit: Please just use `elements` instead of `element(s)` in error messages for brevity (other places, too).

hyeontaek (changes requested):
Thank you!

xla/python/ifrt_proxy/server/ifrt_backend.cc
arunmanni-ai: Done. Renamed to `num_elements` for consistency.

xla/python/ifrt_proxy/server/ifrt_backend.cc
arunmanni-ai: Updated. Simplified error messages to use "elements".

hyeontaek (approved):

PR #44097, opened 2026-06-11
[ROCm] Condition the ragged-dot rewriting on the xla_gpu_experimental_use_ragged_dot_grouped_gemm flag

mfrancepillois:
📝 Summary of Changes
Extend the condition in `HandleRaggedDot` to check that the flag `xla_gpu_experimental_use_ragged_dot_grouped_gemm` is set before allowing the rewriting of `ragged-dot` into `kCublasLtGroupedMatmulCallTarget`

🎯 Justification
Even though this flag is already tested before in the pipeline (`ragged-dot-rewriter`), the `gemm-rewriter` pass can be run on an HLO that contains `ragged-dot` ops. In this case, the pass could rewrite the ragged-dot even if users explicitly disable hipblaslt group-gemm by setting `xla_gpu_experimental_use_ragged_dot_grouped_gemm=false`. And this case does indeed occur when the fission backend calls this pass.
Checking the flag in this pass avoids this problem.

🚀 Kind of Contribution

### DIFF
--- a/xla/backends/gpu/transforms/gemm_rewriter.cc
+++ b/xla/backends/gpu/transforms/gemm_rewriter.cc
@@ -799,15 +799,21 @@ class GemmRewriterVisitor : public DfsHloRewriteVisitor {
 
   absl::Status HandleRaggedDot(HloInstruction* instr) override {
     // The cuDNN ragged dot fusion is only available on NVIDIA/CUDA devices.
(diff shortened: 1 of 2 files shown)


akuegel (commented):
Can you please add a test for this?

akuegel (approved):

PR #44127, opened 2026-06-11
Fix FissionBackend to propagate user DebugOptions into extracted modules

mfrancepillois:
📝 Summary of Changes
Fix: copy the backend's stored DebugOptions (which originate from the user's HloModule config) into the extracted module's config before running the rewriter pipeline.

🎯 Justification
ExtractComputationIntoNewModule creates a new HloModule with a bare HloModuleConfig{}, whose DebugOptions are initialized to DefaultDebugOptionsIgnoringFlags() rather than the user-configured values. As a result, every pass run inside GetFissionedAndRewrittenModule (GemmRewriter, PriorityFusion, etc.) was reading default flag values instead of the user's.
This could be problematic in certain cases, for example as the default value for `xla_gpu_experimental_use_ragged_dot_grouped_gemm` is `true`, the fission backend could choose to lower `ragged-dot` using hipblaslt group-gemm even though users have explicitly disable this option by setting `xla_gpu_experimental_use_ragged_dot_grouped_gemm=false`.

🚀 Kind of Contribution

🧪 Unit Tests:
A test has been added to `fission_backend_test.cc` to verify that user defined parameter are taken into account in the passoes run by the Fission backend. (The test uses the falg  `xla_gpu_gemm_rewrite_size_threshold` and does not impment the `ragged-dot` issue described above, because the ragged-dot rewriting only applies to AMD targets.)

### DIFF
--- a/xla/backends/gpu/autotuner/fission_backend.cc
+++ b/xla/backends/gpu/autotuner/fission_backend.cc
@@ -186,6 +186,13 @@ FissionBackend::GetFissionedAndRewrittenModule(
   const auto* fusion = Cast<HloFusionInstruction>(&fusion_instr);
   std::unique_ptr<HloModule> hlo_module =
       ExtractComputationIntoNewModule(*fusion->called_computation());
+  // ExtractComputationIntoNewModule creates a new HloModule with a default
+  // HloModuleConfig, whose DebugOptions are initialized to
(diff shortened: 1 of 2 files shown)


derdrdirk (approved):
LGTM

PR #44130, opened 2026-06-11
[ROCm] Cache hipStream_t handles in ROCm executor to avoid hipStreamCreate cost

mfrancepillois:
📝 Summary of Changes
This PR implements a process-level `HipStreamHandleCache` singleton directly in rocm_stream.cc. Cache key: (device_ordinal, creation_flags, creation_priority_int).

On destruction (RocmStream::~RocmStream):
  1. BlockHostUntilDone() already ran -- stream is idle.
  2. hipStreamQuery() confirms idleness; on error the handle is destroyed rather than cached (no poisoning).
  3. hipStreamGetFlags / hipStreamGetPriority are called to build the exact cache key, ensuring a retrieved handle always matches the flags and priority the new stream would have used -- even if XLA later switches to hipStreamNonBlocking.
  4. Idle handle is stored; hipStreamDestroy is skipped.

On creation (RocmStream::Create via CreateStream):
  The cache is checked first; on hit the cached handle is returned
  directly and hipStreamCreate is skipped. On miss the cold path
  calls hipStreamCreate as before.

The LocalDeviceState and RocmStream wrapper objects are still created and destroyed normally on every client instantiation. DNN state is cleaned up via DeallocateStream as usual. Only the underlying HIP queue (hipStream_t) is reused.

Also fix a latent use-after-free in `LocalDeviceState::~LocalDeviceState`:
  The fix adds `callback_thread_->Drain()` between `SynchronizeAllActivity()`
  and the explicit stream/event clears. After `Drain()` the callback thread
  is idle and `compute_events_` can be safely cleared.

🎯 Justification
On the ROCm side, creating and destroying the pjrt client takes a long time due to the creation and destruction of HIP streams.
```
Per-test overhead (8 GPU ROCm, iota_test):

CREATION (~406ms):
  Phase1 GetGpuXlaClient:      0.2ms    (negligible, singleton)
  Phase2 hipStreamCreate ×112: 385ms    ← dominant creation cost
  Phase3 EnablePeerAccess:     0.4ms    (negligible, cached)
  Phase4 BFC Allocator:        0.2ms    (negligible, no prealloc)
  Phase5 BuildDistributed:      20ms    (RCCL topology)

DESTRUCTION (~513ms):
  dtor body:                   0.05ms
  thread pool shutdown:       138ms 
  hipStreamDestroy ×112:      375ms    ← dominant destruction cost
  SyncAllActivity:              1.5ms   (device 0 only)
```
For tests, such as `iota_test` that create and destroy pjrt client 4500 times, the long delay associated with these creation and destruction of pjrt clients is a major problem.

🚀 Kind of Contribution
Please remove what does not apply: 🐛 Bug Fix, ⚡️ Performance Improvement

🧪 Unit Tests:
Tests have been added to xla/stream_executor/rocm/rocm_stream_test.cc to verify that the the cache is working correctly.
Tests have been added in xla/pjrt/worker_thread_test.cc to check the stream is correctly drained when it is released.

### DIFF
--- a/xla/pjrt/worker_thread_test.cc
+++ b/xla/pjrt/worker_thread_test.cc
@@ -107,22 +107,5 @@ TEST(WorkerThreadTest, ConsecutiveDrainsOnIdleThread) {
   thread.Drain();  // First drain on empty queue — must not deadlock.
   thread.Drain();  // Second drain on still-empty queue — must not deadlock.
 }
-
-// Drain() guarantees *prior* closures have completed.  A closure scheduled
-// *after* Drain() starts may or may not have run when Drain() returns; only
-// the work enqueued before the sentinel is guaranteed.
-TEST(WorkerThreadTest, DrainGuaranteesPriorClosuresComplete) {
-  WorkerThread thread(tsl::Env::Default(), "test");
-
-  std::atomic<int> before{0};
-
-  thread.Schedule(
-      [&before]() { before.fetch_add(1, std::memory_order_relaxed); });
-
-  thread.Drain();  // The closure above is guaranteed to have completed.
-
-  EXPECT_EQ(before.load(std::memory_order_relaxed), 1);
-}
-
 }  // namespace
 }  // namespace xla
--- a/xla/pjrt/worker_thread_test.cc
+++ b/xla/pjrt/worker_thread_test.cc
@@ -1,0 +1,128 @@
+/* Copyright 2026 The OpenXLA Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "xla/pjrt/worker_thread.h"
+
+#include <atomic>
+#include <vector>
+
+#include "absl/synchronization/mutex.h"
+#include "xla/hlo/testlib/test.h"
+#include "xla/tsl/platform/env.h"
+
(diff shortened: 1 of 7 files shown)


xla/pjrt/worker_thread_test.cc
```diff
@@ -0,0 +1,128 @@
...
+  WorkerThread thread(tsl::Env::Default(), "test");
+
+  thread.Drain();  // First drain on empty queue — must not deadlock.
+  thread.Drain();  // Second drain on still-empty queue — must not deadlock.
+}
+
+// Drain() guarantees *prior* closures have completed.  A closure scheduled
+// *after* Drain() starts may or may not have run when Drain() returns; only
+// the work enqueued before the sentinel is guaranteed.
+TEST(WorkerThreadTest, DrainGuaranteesPriorClosuresComplete) {
```
dimitar-asenov: Given the tests you have above (e.g. `DrainWaitsForAllScheduledClosures`), does this one contribute any additional coverage?

xla/stream_executor/rocm/rocm_stream.cc
```diff
@@ -51,8 +55,67 @@ limitations under the License.
...
+// Thread safety: mu guards all map accesses.
+struct HipStreamHandleCache {
+  absl::Mutex mu;
+  // Key: (device_ordinal, flags, priority_int)
+  absl::btree_map<std::tuple<int, unsigned int, int>, std::vector<hipStream_t>>
+      handles ABSL_GUARDED_BY(mu);
+};
+
+HipStreamHandleCache& GetHipStreamHandleCache() {
+  static auto* cache = new HipStreamHandleCache();  // Intentional leak.
```
dimitar-asenov: Instead of `new` please use:
https://github.com/abseil/abseil-cpp/blob/master/absl/base/no_destructor.h

dimitar-asenov (changes requested):

xla/pjrt/worker_thread_test.cc
mfrancepillois: You are right. This test does not contribute any additional coverage. It has been removed.

dimitar-asenov (approved):

PR #44145, opened 2026-06-11
[NVIDIA GPU] Change CollectiveBroadcast op to support dynamic roots

Tixxx:
The current CollectiveBroadcast op hardcodes the root to be the rank 0 in the replica group, this doesn't provide the right level of flexibility. Almost all communication libraries support having dynamic root in the broadcast API. This PR extends the current CB op to take dynamic root determined through an attribute.
1. added an optional "has_dynamic_root" attribute to the CB op
2. once it's set to true, CB op will expect at least 2 operands: the last operand will be a flat array of s32 values which indicate the roots, all other operands will be the inputs. The number of roots should be the same as the number of non-root inputs.
```
  input = f32[128,32]{0,1} parameter(0)
  root = f32[1]{0} parameter(1)
  cb = f32[128,32]{0,1} collective-broadcast(input, root), replica_groups={{1,0},{2,3}}, has_dynamic_root=true
```
This is the PR for the frontend changes including hlo parser and xla builder.
Runtime support is in another pr.

### DIFF
--- a/xla/hlo/builder/xla_builder.cc
+++ b/xla/hlo/builder/xla_builder.cc
@@ -4586,40 +4586,16 @@ XlaOp XlaBuilder::AllToAllTupleWithDeviceList(
 XlaOp XlaBuilder::CollectiveBroadcast(
     XlaOp operand, absl::Span<const ReplicaGroup> replica_groups,
     const std::optional<ChannelHandle>& channel_id) {
-  return CollectiveBroadcastImpl(operand, CollectiveDeviceList(replica_groups),
-                                 channel_id);
+  return CollectiveBroadcastImpl({operand},
+                                 CollectiveDeviceList(replica_groups),
+                                 channel_id, /*has_dynamic_root=*/false);
 }
 
 XlaOp XlaBuilder::CollectiveBroadcastWithDeviceList(
     XlaOp operand, const CollectiveDeviceListBase& replica_groups,
     const std::optional<ChannelHandle>& channel_id) {
-  return CollectiveBroadcastImpl(operand, replica_groups, channel_id);
-}
-
-XlaOp XlaBuilder::CollectiveBroadcastImpl(
-    XlaOp operand, absl::Span<const ReplicaGroup> replica_groups,
-    const std::optional<ChannelHandle>& channel_id) {
-  return CollectiveBroadcastImpl(operand, CollectiveDeviceList(replica_groups),
-                                 channel_id);
-}
-
-XlaOp XlaBuilder::CollectiveBroadcastImpl(
-    XlaOp operand, const CollectiveDeviceListBase& replica_groups,
-    const std::optional<ChannelHandle>& channel_id) {
-  return ReportErrorOrReturn([&]() -> absl::StatusOr<XlaOp> {
-    ASSIGN_OR_RETURN(const Shape* operand_shape, GetShapePtr(operand));
-    HloInstructionProto instr;
-    ASSIGN_OR_RETURN(Shape shape, ShapeInference::InferCollectiveBroadcastShape(
-                                      {operand_shape}));
-    *instr.mutable_shape() = shape.ToProto();
-    PopulateDeviceList(&instr, replica_groups);
-    if (channel_id.has_value()) {
-      instr.set_channel_id(channel_id->handle());
-    }
-
-    return AddInstruction(std::move(instr), HloOpcode::kCollectiveBroadcast,
-                          {operand});
-  });
+  return CollectiveBroadcastImpl({operand}, replica_groups, channel_id,
+                                 /*has_dynamic_root=*/false);
 }
 
 XlaOp XlaBuilder::CollectiveBroadcast(
@@ -4656,11 +4632,14 @@ XlaOp XlaBuilder::CollectiveBroadcastImpl(
       ASSIGN_OR_RETURN(const Shape* operand_shape, GetShapePtr(operand));
       operand_shapes.push_back(operand_shape);
     }
-    CHECK_GT(operand_shapes.size(), 1);
+    if (has_dynamic_root) {
+      TF_RET_CHECK(operand_shapes.size() > 1);
+    }
     HloInstructionProto instr;
     Shape shape;
-    ASSIGN_OR_RETURN(shape, ShapeInference::InferCollectiveBroadcastShape(
-                                operand_shapes, /*has_dynamic_root=*/true));
+    ASSIGN_OR_RETURN(
+        shape, ShapeInference::InferCollectiveBroadcastShape(
+                   operand_shapes, /*has_dynamic_root=*/has_dynamic_root));
     *instr.mutable_shape() = shape.ToProto();
     PopulateDeviceList(&instr, replica_groups);
     if (channel_id.has_value()) {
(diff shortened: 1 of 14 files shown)


pifon2a:
Changes to HLO instructions should be justified. Could you share the design doc for what you're planning to implement and why you need the dynamism here? If the `root` operand is always a constant, then we can probably just permute the replica ids. If it is not, it would be nice to understand an example for that.

Tixxx:
> Changes to HLO instructions should be justified. Could you share the design doc for what you're planning to implement and why you need the dynamism here? If the `root` operand is always a constant, then we can probably just permute the replica ids. If it is not, it would be nice to understand an example for that.

Sounds good, I started a discussion with the proposed design [here](https://github.com/openxla/xla/discussions/44548). Can you take a look at it? Thanks

pifon2a (approved):

akuegel (approved):

xla/hlo/builder/xla_builder.cc
```diff
@@ -4622,6 +4622,56 @@ XlaOp XlaBuilder::CollectiveBroadcastImpl(
...
+    std::vector<const Shape*> operand_shapes;
+    for (const auto& operand : operands) {
+      ASSIGN_OR_RETURN(const Shape* operand_shape, GetShapePtr(operand));
+      operand_shapes.push_back(operand_shape);
+    }
+    CHECK_GT(operand_shapes.size(), 1);
+    HloInstructionProto instr;
+    Shape shape;
+    ASSIGN_OR_RETURN(shape, ShapeInference::InferCollectiveBroadcastShape(
+                                operand_shapes, /*has_dynamic_root=*/true));
```
akuegel: Is it intended to use `has_dynamic_root` hardcoded to true and not pass `has_dynamic_root` that is passed to CollectiveBroadcastImpl?

xla/hlo/parser/hlo_parser.cc
```diff
@@ -2027,6 +2027,9 @@ HloInstruction* HloParserImpl::CreateInstruction(  // NOLINT
           *shape, operands, std::move(device_list), channel_id));
     }
     case HloOpcode::kCollectiveBroadcast: {
+      bool has_dynamic_root;
```
akuegel: This value should be initialized. What should be the default value if the attribute is not there?

xla/service/shape_inference.cc
```diff
@@ -2784,7 +2784,24 @@ ShapeInference::InferScalarBroadcastShape(absl::Span<const Shape> shapes) {
 
 /* static */ absl::StatusOr<Shape>
 ShapeInference::InferCollectiveBroadcastShape(
-    absl::Span<const Shape* const> operand_shapes) {
+    absl::Span<const Shape* const> operand_shapes, bool has_dynamic_root) {
+  if (has_dynamic_root) {
+    RETURN_IF_ERROR(
+        ExpectArray(*(operand_shapes.back()),
+                    "last operand of collective-broadcast with dynamic root"));
+    const Shape& dynamic_root = *operand_shapes.back();
```
akuegel: nit: this could be moved above to the beginning of the if block. Then in ExpectArray, we could already use `dynamic_root` instead of *(operand_shapes.back())

xla/service/shape_inference.cc
```diff
@@ -2784,7 +2784,24 @@ ShapeInference::InferScalarBroadcastShape(absl::Span<const Shape> shapes) {
...
+    TF_RET_CHECK(dynamic_root.IsArray() && dynamic_root.element_type() == S32 &&
+                 dynamic_root.dimensions().size() == 1 &&
+                 ShapeUtil::ElementsIn(dynamic_root) ==
+                     (operand_shapes.size() - 1))
+        << "The last operand of collective-broadcast with dynamic root must be "
+           "a 1-D array of S32 with the same number of elements as the number "
+           "of "
+           "non-root operands, but got "
+        << ShapeUtil::HumanString(dynamic_root);
+    return ShapeUtil::MakeTupleShapeWithPtrs(operand_shapes);
```
akuegel: In your example in the PR description, the output shape of the CollectiveBroadcast op does not contain the dynamic root operand shape. So shouldn't you only pass a slice of operand_shapes without the last operand?

xla/hlo/builder/xla_builder.cc
```diff
@@ -6385,6 +6435,14 @@ XlaOp CollectiveBroadcast(const XlaOp operand,
                                                 channel_id);
 }
 
+XlaOp CollectiveBroadcast(const absl::Span<const XlaOp> operands,
+                          absl::Span<const ReplicaGroup> replica_groups,
+                          const std::optional<ChannelHandle>& channel_id,
+                          bool has_dynamic_root) {
+  return operands.at(0).builder()->CollectiveBroadcast(
```
akuegel, line 6422: We do not want exceptions to be thrown. Please add a CHECK that operands is non-empty (e.g. like for AllToAllTuple) and then you can directly index into operands[0].

xla/service/shape_inference.cc
```diff
@@ -2784,7 +2784,24 @@ ShapeInference::InferScalarBroadcastShape(absl::Span<const Shape> shapes) {
 
 /* static */ absl::StatusOr<Shape>
 ShapeInference::InferCollectiveBroadcastShape(
-    absl::Span<const Shape* const> operand_shapes) {
+    absl::Span<const Shape* const> operand_shapes, bool has_dynamic_root) {
+  if (has_dynamic_root) {
+    RETURN_IF_ERROR(
```
akuegel, line 2791: Based on the PR description, shouldn't you verify that `operand_shape` contains at least 2 elements, and return InvalidArgumentError if not?

xla/hlo/builder/xla_builder.cc
```diff
@@ -4622,6 +4622,56 @@ XlaOp XlaBuilder::CollectiveBroadcastImpl(
...
+    absl::Span<const XlaOp> operands,
+    const CollectiveDeviceListBase& replica_groups,
+    const std::optional<ChannelHandle>& channel_id, bool has_dynamic_root) {
+  return ReportErrorOrReturn([&]() -> absl::StatusOr<XlaOp> {
+    std::vector<const Shape*> operand_shapes;
+    for (const auto& operand : operands) {
+      ASSIGN_OR_RETURN(const Shape* operand_shape, GetShapePtr(operand));
+      operand_shapes.push_back(operand_shape);
+    }
+    CHECK_GT(operand_shapes.size(), 1);
```
akuegel: This should be a TF_RET_CHECK

xla/hlo/builder/xla_builder.cc
Tixxx: yea good point, this is override for the multi-operand collectiveBroadcast since the old one only takes a single operand, i didnt consolidate all the constructors yet so the new constructor is only for cb with dynamic roots. I can consolidate it now.

xla/service/shape_inference.cc
Tixxx: yea im checking them in the callers. i will add them here too

xla/service/shape_inference.cc
Tixxx: yes you are right, i thought collectivebroadcast was an in-place op where xla'd automatically alias all input and output operands. Just checked that it's not inplace yet, i will remove the last operand.

xla/hlo/parser/hlo_parser.cc
Tixxx: changed it to optional<bool>

akuegel (approved):

PR #44154, opened 2026-06-11
[xla] Add ObjectPool::Preallocate + cleanup it a bit

ezhulenev:
Adds `ObjectPool::Preallocate(size_t, Args...),` which eagerly creates a requested number of additional pooled objects using the existing builder and returns them to the pool as available objects. If object creation fails, `Preallocate` returns the first builder error without adding partially-created objects from that call to the pool.

Also updates `ObjectPool` internals to store available objects in a mutex-protected `std::vector<std::unique_ptr<T>>` instead of a linked list wrapper, adds `Get()` for borrowing only already-available objects, and exposes `num_available()` alongside `num_created()`.


xla/runtime/object_pool_test.cc
```diff
@@ -59,6 +69,105 @@ TEST(ObjectPoolTest, GetOrCreate) {
...
+
+  auto destroy =
+      [](ObjectPool<NonDefaultConstructible, int32_t>::BorrowedObject obj) {};
+  destroy(std::move(obj0));
+
+  TF_ASSERT_OK_AND_ASSIGN(auto obj1, pool.Get());
+  EXPECT_EQ(obj1->value, 42);
+  EXPECT_EQ(pool.num_created(), 1);
+}
+
```
seantalts, line 210: Could we add a case where the builder fails on the Nth call and asserting num_created()/num_available() are unchanged (and earlier objects from that call are not leaked into the pool).

seantalts (approved):

xla/runtime/object_pool.h
```diff
@@ -17,143 +17,169 @@ limitations under the License.
...
+  // borrowed object is returned to the pool when this handle is destroyed or
+  // overwritten by move assignment.
+  using BorrowedObject = std::unique_ptr<T, ReturnToPool>;
 
-    T& operator*() const { return *entry_->object; }
-    T* operator->() const { return &*entry_->object; }
+  // Creates `num_objects` additional objects with the builder and returns them
+  // to the pool as available objects. If the builder fails, returns the first
+  // error without adding any object from this call to the pool.
+  absl::Status Preallocate(size_t num_objects, Args... args);
```
ermilovmaxim, line 65: Could you Preallocate() for GpuCompiler?

ermilovmaxim (approved):

xla/runtime/object_pool.h
ezhulenev: Not that nicely, today object pool assumes that builder can fail and putting that into constructor looks not great (.IngoreError()? ugh). I want to support failable/non-failable object pool builders, but can't figure out a nice API for that.

ezhulenev: Added new constructor to obj pool and updated gpu_compiler

ermilovmaxim (approved):

PR #44222, opened 2026-06-12
[GPU] Fix family-conditional PTX compilation.

sergachev:
📝 Summary of Changes
Let XLA properly fall back to compute capabilities it knows when it sees hardware it doesn't know.

🎯 Justification
Allows compiling programs for newer hardware with older software.

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
yes

🧪 Execution Tests:
no


EusebioDM (approved):
LGTM!

I don't think I have write permissions, so you might need to also approve @derdrdirk

derdrdirk (approved):

PR #44235, opened 2026-06-12
[NVIDIA GPU] Thunk change for dynamic root collective bcast

Tixxx:
Follow-up pr of [this](https://github.com/openxla/xla/pull/44145).
This adds a metadata to the collectivfe broadcast thunk to host the cpu buffers for root ranks.
Ranks will be copied to host for each invocation.
I will push the e2e test after the above pr is merged so it doesn't fail all the CIs.

### DIFF
--- a/xla/backends/gpu/runtime/collective_broadcast_thunk.cc
+++ b/xla/backends/gpu/runtime/collective_broadcast_thunk.cc
@@ -24,6 +24,7 @@ limitations under the License.
 #include "absl/base/casts.h"
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
+#include "absl/synchronization/mutex.h"
 #include "absl/types/span.h"
 #include "xla/tsl/platform/status_macros.h"
 #include "xla/backends/gpu/collectives/gpu_clique_key.h"
@@ -41,6 +42,7 @@ limitations under the License.
 #include "xla/hlo/ir/hlo_instructions.h"
 #include "xla/service/buffer_assignment.h"
 #include "xla/stream_executor/device_address.h"
+#include "xla/stream_executor/memory_allocation.h"
 #include "xla/stream_executor/stream.h"
 #include "xla/xla_data.pb.h"
 
@@ -48,10 +50,12 @@ namespace xla::gpu {
 
 CollectiveBroadcastThunk::CollectiveBroadcastThunk(ThunkInfo thunk_info,
                                                    CollectiveConfig config,
-                                                   std::vector<Buffer> buffers)
+                                                   std::vector<Buffer> buffers,
+                                                   bool has_dynamic_root)
     : CollectiveThunk(Thunk::kCollectiveBroadcast, thunk_info,
                       std::move(buffers)),
-      config_(config) {}
+      config_(config),
+      has_dynamic_root_(has_dynamic_root) {}
 
 CollectiveBroadcastThunk::CollectiveBroadcastThunk(
     ThunkInfo thunk_info, const HloCollectiveBroadcastInstruction* instr,
@@ -59,7 +63,7 @@ CollectiveBroadcastThunk::CollectiveBroadcastThunk(
     : CollectiveThunk(Thunk::kCollectiveBroadcast, thunk_info,
                       std::move(buffers)),
       config_(GetCollectiveConfig(instr, std::nullopt)),
-      cb_metadata_({has_dynamic_root, {}}) {}
+      has_dynamic_root_(has_dynamic_root) {}
 
 /*static*/ absl::Status CollectiveBroadcastThunk::CheckImplementable(
     const HloInstruction* instr, int64_t replica_count,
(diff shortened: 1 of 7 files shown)


xla/backends/gpu/runtime/collective_broadcast_thunk.cc
```diff
@@ -106,28 +122,58 @@ absl::StatusOr<ThunkProto> CollectiveBroadcastThunk::ToProto() const {
...
   return proto;
 }
+absl::Status LoadCollectiveBroadcastRoots(
+    se::Stream& stream, DeviceBufferPair roots_device_buffer,
+    CollectiveBroadcastMetadata& cb_metadata) {
+  RETURN_IF_ERROR(stream.Memcpy(cb_metadata.bcast_roots->address().opaque(),
+                                roots_device_buffer.source_buffer,
+                                roots_device_buffer.source_buffer.size()));
+  // Wait for the copies to complete.
+  if (absl::Status blocked = stream.BlockHostUntilDone(); !blocked.ok()) {
```
PatriosTheGreat: Use RETURN_IF_ERROR() << "..." instead

xla/backends/gpu/runtime/collective_broadcast_thunk.cc
```diff
@@ -106,28 +122,58 @@ absl::StatusOr<ThunkProto> CollectiveBroadcastThunk::ToProto() const {
 
   return proto;
 }
+absl::Status LoadCollectiveBroadcastRoots(
```
PatriosTheGreat: This method can be in-lined since it just calls memcpy

xla/backends/gpu/runtime/collective_broadcast_thunk.cc
```diff
@@ -106,28 +122,58 @@ absl::StatusOr<ThunkProto> CollectiveBroadcastThunk::ToProto() const {
...
+}
 absl::Status CollectiveBroadcastThunk::RunCollective(
     const ExecuteParams& params, const GpuCliqueKey& clique_key,
     se::Stream& stream, Communicator& comm) {
   ASSIGN_OR_RETURN(std::vector<DeviceBufferPair> device_buffers,
                    ConvertToDeviceBuffers(params.buffer_allocations, buffers(),
                                           config_.operand_element_type));
-  return ::xla::gpu::RunCollectiveBroadcast(device_buffers, stream, comm);
+  if (cb_metadata_.has_dynamic_root) {
+    RETURN_IF_ERROR(LoadCollectiveBroadcastRoots(stream, device_buffers.back(),
```
PatriosTheGreat: Could you move this to RunCollectiveBroadcast so all code related to has_dynamic_root at lives in a single method?

xla/backends/gpu/runtime/collective_broadcast_thunk.cc
```diff
@@ -71,6 +72,21 @@ CollectiveBroadcastThunk::CollectiveBroadcastThunk(
   return GetCollectiveConfig(inst, std::nullopt).group_mode;
 }
 
+absl::Status CollectiveBroadcastThunk::Initialize(
+    const InitializeParams& params) {
+  if (cb_metadata_.has_dynamic_root && !initialized_) {
```
PatriosTheGreat: Can you use cb_metadata_.bcast_roots != nullptr here instead of initialized_ parameter? Also I guess cb_metadata_ should be GPU specific property, so you need something like a map from stream_executor* to cb_metadata_ as a thunk field.

PatriosTheGreat (changes requested):

penumallas1261 (approved):

PatriosTheGreat (approved):

Tixxx:
fixed the failure for the multigpu test

PatriosTheGreat:
Can you please add a few test-cases to xla/backends/gpu/runtime/collective_broadcast_thunk_multigpu_test.cc checking a new logic?

Tixxx:
> Can you please add a few test-cases to xla/backends/gpu/runtime/collective_broadcast_thunk_multigpu_test.cc checking a new logic?

sorry forgot to include all the changes after re-opening the pr from a saved commit. pushed the test changes too

Tixxx:
@PatriosTheGreat  gentle ping on the review

xla/backends/gpu/runtime/collective_broadcast_thunk.cc
```diff
@@ -113,21 +137,52 @@ absl::Status CollectiveBroadcastThunk::RunCollective(
...
+      return absl::InternalError(
+          absl::StrFormat("Failed to copy dynamic roots on stream %p: %s",
+                          &stream, blocked.message()));
+    }
+  }
   auto* gpu_comm = absl::down_cast<GpuCommunicator*>(&comm);
   Future<> future = gpu_comm->GroupExecute([&]() -> absl::Status {
-    for (auto buffer : buffers) {
+    RankId root = RankId(0);
+    for (int64_t i = 0; i < buffers.size(); ++i) {
```
akuegel: This for loop should not go until buffers.size() - 1 for the `has_dynamic_root && cb_metadat` case.
As far as I understand, when has_dynamic_root is true, the last buffer in buffers (buffers.back()) is the root rank buffer, not a data payload buffer. `roots_ptr` is also only allocated for buffers.size() - 1 elements (line 88 above).

`const int64_t num_broadcasts =
      (has_dynamic_root && cb_metadata) ? buffers.size() - 1 : buffers.size();`

And then let the loop go until `< num_broadcast`

Can you please also add a test case that exercises this case?

Tixxx: was waiting for the cb hlo change to land. pushed the e2e test and emitter change.

akuegel (approved):

PR #44244, opened 2026-06-12
Update oneAPI CI tag filters 

mraunak:
This PR updates the oneAPI GPU CI build and test tag filters to align with the broader GPU test selection.


build_tools/ci/build.py
```diff
@@ -480,18 +480,17 @@ def nvidia_gpu_build_with_compute_capability(
 )
 
 oneapi_build_tag_filter = (
-    "oneapi-only",
```
dimitar-asenov, line 483: What happens to the many existing usages of "oneapi-only" when this tag is removed. Is this change safe?

mraunak: @dimitar-asenov, thank you for the review, added it back

dimitar-asenov (approved):

PR #44260, opened 2026-06-13
ifrt_proxy: validate peer-supplied string count before writing into preallocated host buffer

evilgensec:
## Summary

`DeserializeFromCordIntoPreallocatedStringHostBuffer` (`xla/python/ifrt_proxy/common/array_util.cc`) writes one `absl::Cord` per string element of a `StringArrayContents` proto into `preallocated_buffer`, advancing a raw pointer with no bounds check. The buffer is sized by the caller `Array::CopyToStringHostBuffer` from the local array shape (`shape_.num_elements()`), while the element count is read from the proto returned by the IFRT proxy server. The two were never reconciled: the header documented "callers must ensure that the `preallocated_buffer` consists of `N` ... objects" as an unenforced invariant, and the call site did not enforce it.

A proxy server that returns a `StringArrayContents` with more `strings` than the client buffer holds, or an attacker injecting a response on an insecure transport (`GetClientCredentialsPossiblyInsecure`), causes an out-of-bounds heap write of peer-controlled `absl::Cord` objects past the destination buffer.

The peer function `DeserializeStringHostBufferFromString` is not affected because it returns a freshly grown vector.

## Fix

Pass the destination element count into `DeserializeFromCordIntoPreallocatedStringHostBuffer` and reject any serialized count that does not match before writing. The caller threads `shape_.num_elements()`, and the unit test is updated for the new signature.

## Testing

Exercised the deserialize routine with the real `StringArrayContents` proto and `absl::Cord` under `-DNDEBUG -fsanitize=address,undefined`. Before the change, a serialized buffer carrying more strings than the preallocated destination triggers `heap-buffer-overflow` in `absl::Cord::operator=` inside the write loop. After the change the mismatched count is rejected with `InvalidArgumentError`, and a well-formed buffer round-trips unchanged.

### DIFF
--- a/xla/python/ifrt_proxy/client/array.cc
+++ b/xla/python/ifrt_proxy/client/array.cc
@@ -722,7 +722,8 @@ tsl::Future<> Array::CopyToStringHostBuffer(
   auto on_ready = [promise = std::move(promise),
                    host_buffer_store = rpc_helper_->host_buffer_store(),
                    host_buffer_handle,
-                   dst_buffer = static_cast<absl::Cord*>(data)](
+                   dst_buffer = static_cast<absl::Cord*>(data),
+                   num_elements = static_cast<size_t>(shape_.num_elements())](
                       absl::StatusOr<std::shared_ptr<CopyToHostBufferResponse>>
                           resp) mutable {
     if (!resp.ok()) {
@@ -731,7 +732,7 @@ tsl::Future<> Array::CopyToStringHostBuffer(
     }
     host_buffer_store->Lookup(host_buffer_handle)
         .OnReady([promise = std::move(promise), dst_buffer, host_buffer_store,
-                  host_buffer_handle](
+                  host_buffer_handle, num_elements](
                      absl::StatusOr<absl::Cord> array_contents) mutable {
           absl::Cleanup cleanup = [&]() {
             host_buffer_store->Delete(host_buffer_handle)
@@ -751,7 +752,7 @@ tsl::Future<> Array::CopyToStringHostBuffer(
           }
           auto deserialization_status =
               DeserializeFromCordIntoPreallocatedStringHostBuffer(
-                  *array_contents, dst_buffer);
+                  *array_contents, num_elements, dst_buffer);
           promise.Set(deserialization_status);
         });
   };
--- a/xla/python/ifrt_proxy/client/array.cc
+++ b/xla/python/ifrt_proxy/client/array.cc
@@ -723,7 +723,7 @@ tsl::Future<> Array::CopyToStringHostBuffer(
                    host_buffer_store = rpc_helper_->host_buffer_store(),
                    host_buffer_handle,
                    dst_buffer = static_cast<absl::Cord*>(data),
-                   num_elements = static_cast<size_t>(shape_.num_elements())](
+                   num_elements = shape_.num_elements()](
                       absl::StatusOr<std::shared_ptr<CopyToHostBufferResponse>>
                           resp) mutable {
     if (!resp.ok()) {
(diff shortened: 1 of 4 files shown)


xla/python/ifrt_proxy/common/array_util.cc
```diff
@@ -206,6 +206,19 @@ absl::Status DeserializeFromCordIntoPreallocatedStringHostBuffer(
         "Failed to parse serialized string buffer");
   }
 
+  // The element count is supplied by the peer (the proxy server, or an attacker
```
hyeontaek: This is not clear from the code alone, but we ask the IFRT Proxy user to build a single trust domain for IFRT Proxy client and server, with their RPC hardened using a mechanism the user provides. Thus, we do not expect either side to act as an attacker, and I do not think this comment is sketching an actual scenario, unless the user has violated the contract and put IFRT Proxy client and server on separate trust domains or put their RPCs without additional secure networking.

Regardless, this is a useful code hardening. Please see a related change #44083 and see if the change can be more coherent.

hyeontaek (changes requested):
Thanks!

xla/python/ifrt_proxy/common/array_util.cc
evilgensec: Done in 6cfa81de. Dropped the attacker / trust-boundary framing from both the .cc and .h comments — given the single-trust-domain contract between client and server, the check is now presented as robustness against an inconsistent buffer rather than an attack scenario. Also aligned the error message with the server-side element-count validation in #44083 ("String host buffer has N elements but shape requires M elements") so the two paths read coherently.

xla/python/ifrt_proxy/client/array.cc
```diff
@@ -722,7 +722,8 @@ tsl::Future<> Array::CopyToStringHostBuffer(
   auto on_ready = [promise = std::move(promise),
                    host_buffer_store = rpc_helper_->host_buffer_store(),
                    host_buffer_handle,
-                   dst_buffer = static_cast<absl::Cord*>(data)](
+                   dst_buffer = static_cast<absl::Cord*>(data),
+                   num_elements = static_cast<size_t>(shape_.num_elements())](
```
hyeontaek: Please use `int64_t` for counting the number of elements that is calculated from a shape, which is a convention in this codebase (also generally, the Google style guide recommends using signed types if unsigned int semantics is required).

xla/python/ifrt_proxy/common/array_util.h
```diff
@@ -90,11 +90,12 @@ absl::StatusOr<std::unique_ptr<std::string>> SerializeStringHostBuffer(
...
-// Callers must ensure that the `preallocated_buffer` consists of `N`
-// `absl::Cord` objects, where N is the number of string elements in the
-// `serialized_string_buffer`.
+// `preallocated_buffer` points to `num_elements` default-constructed
+// `absl::Cord` objects, sized from the array shape. The number of string
+// elements encoded in `serialized_string_buffer` is validated against
+// `num_elements` before any element is written; a mismatch returns an error.
 absl::Status DeserializeFromCordIntoPreallocatedStringHostBuffer(
-    const absl::Cord& serialized_string_buffer,
+    const absl::Cord& serialized_string_buffer, size_t num_elements,
```
hyeontaek: Also use `int64_t` here because this quantity is originally calculated from a shape, rather than computed from a C++ container.

hyeontaek (changes requested):

evilgensec:
Thanks @hyeontaek, addressed in f3954a1.

The count is now `int64_t` throughout: in `array.cc` the lambda capture keeps the `int64_t` returned by `Shape::num_elements()` directly (dropped the `size_t` cast), and in `array_util.{h,cc}` the `num_elements` parameter and the `strings_size()` comparison are signed `int64_t` (the protobuf `int` from `strings_size()` widens cleanly). The test passes the container size cast to `int64_t`.

hyeontaek (approved):

akuegel (approved):

PR #44279, opened 2026-06-14
[xla:gpu] Allow S(0) buffers to use S(1) allocations

ezhulenev:
- Add `BufferAssigner::CanUseColor`, a construction-time predicate that decides whether a buffer with one HLO color can use an allocation with another color.
  - Keep input/output allocations strict: they must match the raw buffer color, since caller-provided/live-out storage cannot have its runtime requirements strengthened during buffer assignment.
  - Use `CanUseColor` in GPU buffer assignment to allow `S(0)` buffers to use `S(1)` temp allocations, while still rejecting `S(1)` buffers in `S(0)` allocations.

  In XLA:GPU, `S(1)` is used for symmetric collective operations. Without this change, a collective operation with large parameters/results can effectively reserve a whole allocation that is unused most of the time. By allowing `S(0)` temp buffers to use `S(1)` allocations, XLA can significantly reduce memory usage.

### DIFF
--- a/xla/service/buffer_assignment.cc
+++ b/xla/service/buffer_assignment.cc
@@ -347,10 +347,10 @@ absl::Status BufferAllocation::AddAssignment(const HloValue& buffer,
   CHECK_LE(offset + size, size_)
       << "LogicalBuffer " << buffer
       << " size out of range at offset: " << offset << " with size: " << size;
-  if (!(IsPreallocatedTempBuffer() && color() != 0)) {
+  if (IsInputOrOutput()) {
     TF_RET_CHECK(buffer.color() == color())
         << "Buffer color " << buffer.color() << " for buffer " << buffer
-        << " does not match allocation color " << color()
+        << " does not match input/output allocation color " << color()
         << ". This could be caused by an unusable donated buffer.";
   }
   OffsetSize offset_size;
@@ -518,6 +518,20 @@ std::string BufferAllocation::ToString() const {
                   " value: %s (size=%d,offset=%d): %s\n",
                   buffer->ToShortString(), offset_size.size, offset_size.offset,
                   ShapeUtil::HumanStringWithLayout(buffer->shape())));
+  }
+  // Report any buffers that reused this allocation's free space despite having
+  // a different color. They are already listed above; this makes the
+  // cross-color reuse explicit for humans reading the dump.
+  if (!cross_color_buffers_.empty()) {
+    std::vector<const HloValue*> sorted_reused(cross_color_buffers_.begin(),
+                                               cross_color_buffers_.end());
+    absl::c_sort(sorted_reused, &CompareHloValuesById);
+    StrAppend(&output, " reused by buffers of a different color:\n");
+    for (const HloValue* buffer : sorted_reused) {
+      StrAppend(&output,
+                absl::StrFormat("  value: %s (color=%d)\n",
+                                buffer->ToShortString(), buffer->color()));
+    }
   }
   return output;
 }
@@ -875,6 +889,13 @@ absl::Status BufferAssignment::CombineTempAllocations(
           temp_allocation.peak_buffers_.begin(),
           temp_allocation.peak_buffers_.end());
     }
+
+    // Carry over the cross-color reuse reporting info from the absorbed
+    // allocation so the combined allocation's dump still reflects it.
+    combined_allocation->cross_color_buffers_.insert(
+        combined_allocation->cross_color_buffers_.end(),
+        temp_allocation.cross_color_buffers_.begin(),
+        temp_allocation.cross_color_buffers_.end());
 
     if (temp_buffer_color.has_value()) {
       if (combined_allocation->color() == 0) {
@@ -1595,7 +1616,21 @@ absl::StatusOr<bool> BufferAssigner::MaybeAssignBuffer(
           << " to allocation: " << *allocation;
 
   ASSIGN_OR_RETURN(auto buffer_color, hlo_buffer.color());
-  if (buffer_color != allocation->color()) {
+
+  // Input/output allocations are backed by caller-provided or live-out storage,
+  // so buffer assignment cannot strengthen their runtime allocation
+  // requirements after the fact. Temporary allocations can use the backend's
+  // directional color-assignment predicate below.
+  if (allocation->IsInputOrOutput() && buffer_color != allocation->color()) {
+    VLOG(4) << "Cannot assign: input/output allocation has color "
+            << allocation->color() << ", but buffer has color " << buffer_color
+            << ".";
+    return false;
+  }
+
+  // This is a storage-compatibility check, not recoloring: the buffer keeps its
+  // HLO color even when it can use an allocation with a different color.
+  if (!opts_.can_use_allocation(buffer_color, allocation->color())) {
     VLOG(4) << "Can't assign: buffer has color " << buffer_color
             << " and allocation has color " << allocation->color() << ".";
     return false;
@@ -1736,7 +1771,7 @@ absl::StatusOr<bool> BufferAssigner::MaybeAssignBuffer(
       assignment->AddAssignment(allocation, hlo_buffer, /*offset=*/0,
                                 assignment->HloBufferSize(hlo_buffer)));
   return true;
-}  // namespace xla
+}
 
 absl::Status BufferAssigner::AssignSingleHloBuffer(
     const HloBuffer* hlo_buffer, bool is_thread_local,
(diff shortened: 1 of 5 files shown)


xla/service/buffer_assignment_test.cc
```diff
@@ -691,6 +698,184 @@ TEST_F(BufferAssignmentTest, AliasedParamCanBeReused) {
...
+  EXPECT_TRUE(param_buffer.is_parameter_aliased_with_output());
+  EXPECT_EQ(param_buffer.color(), 1);
+  EXPECT_EQ(buffers->dataflow_analysis().GetUniqueValueAt(neg_1, {}).color(),
+            0);
+
+  EXPECT_NE(param_buffer.index(), neg_1_buffer.index());
+  EXPECT_EQ(param_buffer.index(), neg_2_buffer.index());
+}
+
+TEST_F(BufferAssignmentTest, CanUseColorReusesFixedHeapSimulationColor) {
```
PatriosTheGreat: This test-case fails internally with error:

INTERNAL: RET_CHECK failure ([xla/hlo/ir/hlo_schedule.cc:388] nonfusion_computations.size() <= sequence_size For thread main, schedule has 1 sequences, but module has 2 non-fusion computations for thread main

PatriosTheGreat (changes requested):

xla/service/buffer_assignment.cc
```diff
@@ -2021,6 +2056,227 @@ BufferAssigner::SplitBuffersByPrivateStackComputation(
...
+    }
+  }
+  if (destinations.empty()) {
+    return 0;
+  }
+
+  // Step 2: Collect movable candidates.
+  std::vector<ReuseCandidate> candidates;
+  absl::flat_hash_set<const HloBuffer*> seen_buffers;
+  for (const HloValue* value : *values) {
```
PatriosTheGreat, line 2187: clang-tidy complains that iteration order is non-deterministic and suggests to use:
for (const HloValue* value : tsl::SortedRange(*values))

xla/service/buffer_assignment.cc
```diff
@@ -2021,6 +2056,227 @@ BufferAssigner::SplitBuffersByPrivateStackComputation(
...
+  // `allocation`, or nullopt if it does not fit. A gap is free if no buffer
+  // occupying it is live at the same time as the candidate.
+  auto find_placement = [&](const ReuseCandidate& candidate,
+                            BufferAllocation* allocation) {
+    // Collect the byte ranges that block the candidate: only buffers whose live
+    // range overlaps the candidate's. Non-overlapping buffers are invisible
+    // because the candidate can safely reuse their storage.
+    std::vector<BufferAllocation::OffsetSize> blocking;
+    for (const auto& [assigned_value, offset_size] :
+         allocation->assigned_buffers()) {
```
PatriosTheGreat: Same here with non-deterministic loop. Clang-tidy suggests:
tsl::SortedRange(allocation->assigned_buffers())) {

xla/service/buffer_assignment_test.cc
```diff
@@ -206,11 +209,14 @@ class BufferAssignmentTest : public HloHardwareIndependentTestBase {
 
   std::unique_ptr<BufferAssignment> RunBufferAssignmentWithInstructionSequence(
       HloModule* module, absl::Span<HloInstruction* const> instruction_sequence,
-      int64_t alignment = 1) {
+      int64_t alignment = 1,
+      BufferAssigner::CanUseColor can_use_color =
+          BufferAssigner::DefaultCanUseColor()) {
     HloSchedule schedule(module);
     schedule.set_sequence(module->entry_computation(), instruction_sequence);
```
PatriosTheGreat, line 217: Adding CHECK_OK(schedule.Update()); after this line fixes the test

PatriosTheGreat (changes requested):

xla/service/buffer_assignment_test.cc
```diff
@@ -691,6 +698,184 @@ TEST_F(BufferAssignmentTest, AliasedParamCanBeReused) {
...
+  ROOT neg2 = f32[100]{0:S(1)} negate(neg1)
+}
+)";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo_text));
+
+  std::unique_ptr<BufferAssignment> buffers = RunBufferAssignment(
+      module.get(), /*alignment=*/1,
+      [](BufferValue::Color buffer_color, BufferValue::Color allocation_color) {
+        return buffer_color == allocation_color ||
```
PatriosTheGreat: This lambda is used in several places. Maybe it worth to add it near to the DefaultCanUseColor and call it something like "CollectiveMemorySpaceCanBeUsedForDefaultColor"?

xla/service/buffer_assignment.cc
```diff
@@ -2021,6 +2056,227 @@ BufferAssigner::SplitBuffersByPrivateStackComputation(
...
+      if (candidate_interferes_with(candidate, assigned_value)) {
+        blocking.push_back(offset_size);
+      }
+    }
+    int64_t alignment = assignment->color_alignment_(allocation->color());
+    return FindBestFitGap(allocation, candidate.size, alignment,
```
PatriosTheGreat, line 2250: allocation here can't be nullptr since otherwise allocation->color will fail.

Could you add null check at the beginning of lambda and also pass reference to the FindBestFitGap

PatriosTheGreat (changes requested):

xla/service/buffer_assignment.cc
ezhulenev: Done.

xla/service/buffer_assignment_test.cc
ezhulenev: Done

xla/service/buffer_assignment.cc
ezhulenev: `values` are pointers, so sorting doesn't buy any determinism, the overall loop is still deterministic as we sort candidates after collecting all of them, I think clang-tidy is wrong here

xla/service/buffer_assignment.cc
ezhulenev: Same here, sorting pointers is useless, also blocking is sorted by size after it's collected, so this whole thing is already determenistic anyway

xla/service/buffer_assignment_test.cc
ezhulenev: Did a rename, now the predicate called `CanUseAllocation` (buffer color can use allocation color), and added a static helper `AllowCrossColorReuse` for use in tests and GPU backend.

xla/service/buffer_assignment.cc
```diff
@@ -1736,7 +1771,7 @@ absl::StatusOr<bool> BufferAssigner::MaybeAssignBuffer(
       assignment->AddAssignment(allocation, hlo_buffer, /*offset=*/0,
                                 assignment->HloBufferSize(hlo_buffer)));
   return true;
-}  // namespace xla
+}
```
akuegel, line 1774: Thanks for catching this, however I believe the right fix is to end the anonymous namespace with
```}  // namespace```

File: xla/service/buffer_assignment.cc [hunk 8]
```diff
@@ -2027,6 +2062,228 @@ BufferAssigner::SplitBuffersByPrivateStackComputation(
...
+
+  // Sweep left to right. `free_start` is the first byte not covered by a
+  // blocking range seen so far; [free_start, blocking_range.offset) is the next
+  // free gap. We keep the gap with the least leftover slack.
+  std::optional<ReusePlacement> best;
+  int64_t free_start = 0;
+  for (const BufferAllocation::OffsetSize& blocking_range : blocking) {
+    int64_t offset = RoundUpTo(free_start, alignment);
+    if (offset + size <= blocking_range.offset) {
+      int64_t slack = blocking_range.offset - (offset + size);
```
akuegel: optional: what about computing slack outside of the if block and changing the if condition to `if (slack >= 0)`?

xla/service/buffer_assignment_test.cc
```diff
@@ -664,19 +671,20 @@ TEST_F(BufferAssignmentTest, AliasedParamCanBeReused) {
...
-  module->AddEntryComputation(builder.Build());
+ENTRY main {
+  p0 = f32[100]{0} parameter(0)
+  neg1 = f32[100]{0} negate(p0)
+  ROOT neg2 = f32[100]{0} negate(neg1)
+}
+)";
 
-  TF_ASSERT_OK(module->input_output_alias_config().SetUpAlias({}, 0, {}));
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo_text));
```
akuegel: nit: use `ASSERT_OK_AND_ASSIGN`

xla/service/buffer_assignment_test.cc
```diff
@@ -692,6 +700,177 @@ TEST_F(BufferAssignmentTest, AliasedParamCanBeReused) {
...
+HloModule test, input_output_alias={ {}: (0, {}, may-alias) }
+
+ENTRY main {
+  p0 = f32[100]{0:S(1)} parameter(0)
+  neg1 = f32[100]{0} negate(p0)
+  ROOT neg2 = f32[100]{0:S(1)} negate(neg1)
+}
+)";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo_text));
```
akuegel: nit: use `ASSERT_OK_AND_ASSIGN`

xla/service/buffer_assignment_test.cc
```diff
@@ -692,6 +700,177 @@ TEST_F(BufferAssignmentTest, AliasedParamCanBeReused) {
...
+  neg0 = f32[2048]{0:S(1)} negate(p0)
+  zero = f32[] constant(0)
+  r1 = f32[] reduce(neg0, zero), dimensions={0}, to_apply=add_s
+  neg1 = f32[1024]{0} negate(p1)
+  r0 = f32[] reduce(neg1, zero), dimensions={0}, to_apply=add_s
+  ROOT out = f32[] add(r1, r0)
+}
+)";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo_text));
```
akuegel: nit: use `ASSERT_OK_AND_ASSIGN`

xla/service/buffer_assignment_test.cc
```diff
@@ -692,6 +700,177 @@ TEST_F(BufferAssignmentTest, AliasedParamCanBeReused) {
...
+  neg0 = f32[1024]{0:S(1)} negate(p0)
+  zero = f32[] constant(0)
+  r1 = f32[] reduce(neg0, zero), dimensions={0}, to_apply=add_s
+  neg1 = f32[2048]{0} negate(p1)
+  r0 = f32[] reduce(neg1, zero), dimensions={0}, to_apply=add_s
+  ROOT out = f32[] add(r1, r0)
+}
+)";
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module, ParseAndReturnVerifiedModule(hlo_text));
```
akuegel: nit: use `ASSERT_OK_AND_ASSIGN`

PatriosTheGreat (approved):

xla/service/buffer_assignment.cc
PatriosTheGreat: Anonymous namespace is closed at line 2145. I guess this comment was a leftover from some previous change.

akuegel: Ah, good point. I did not notice that this is the end of the function body, only noticed that a namespace comment is removed. So this comment can be resolved.

olegshyshkov (approved):

PatriosTheGreat:
Some internal tests are failing with this change on this check:
https://github.com/openxla/xla/blob/main/xla/pjrt/compiled_memory_stats.cc#L64

I'll try to put together a reproducer to share.

xla/service/buffer_assignment.cc
ezhulenev: Done

ezhulenev:
I relaxed `compiled_memory_stats.cc` to check that host/device buffers are in separate allocations, but allow multiple device colors to share an allocation.

ermilovmaxim (approved):

PR #44295, opened 2026-06-15
[XLA:GPU] Achieve command buffer persisitent address through DeviceAddressVmmAllocator va-remapping APIs

shawnwang18:
📝 Summary of Changes
This PR is the command-buffer VMM allocation step in the stack:

Previous PR: https://github.com/openxla/xla/pull/44355
Next PR: https://github.com/openxla/xla/pull/44296

This change routes command-buffer execution allocation through `GpuExecutableBufferAllocator` using `se::DeviceAddressVmmAllocator`:

- Creates executable/executor-scoped VMM remapping state for command-buffer allocation indexes.
- Reserves a stable virtual address range and assigns deterministic offsets for command-buffer-visible allocations.
- Allocates command-buffer temp/live-out buffers through the VMM allocator when VA remapping is active.
- Maps entry-parameter buffers into reservation-backed aliases for command-buffer execution, then unmaps those aliases after execution.
- Synchronizes pending VMM operations when the executable-scoped allocator is destroyed.
- Keeps returned-output/live-out policy inside `GpuExecutableBufferAllocator` so returned buffers expose the expected physical allocation address instead of a reservation alias.
- Moves command-buffer update-mode/index collection and returned-output allocation-index collection into `GpuExecutableBufferAllocator`, keeping `GpuExecutable` and PJRT call sites focused on scope creation, allocation generation, execution, and teardown.
- Leaves PJRT teardown behavior and existing `buffer_allocations` naming unchanged.

🎯 Justification
Command-buffer replay is sensitive to device address changes. VMM reservations let XLA present stable virtual addresses for buffers captured by command buffers while still allowing the physical backing allocations, including parameter buffers, to vary between executions.

Centralizing this policy in `GpuExecutableBufferAllocator` keeps the VMM reservation, remapping, returned-output handling, and command-buffer allocation-index collection in one owner. That makes the `GpuExecutable` and PJRT execution paths smaller and prepares the follow-up command-buffer update-info runtime change in #44296.

🚀 Kind of Contribution
⚡️ Performance Improvement, ♻️ Cleanup

📊 Benchmark (for Performance Improvements)
Not run locally. This PR adds the runtime allocation/remapping plumbing expected to reduce command-buffer update overhead for VMM-capable command-buffer executions, especially with `xla_gpu_command_buffer_update_mode=NEVER_UPDATE` or `CAPTURE_CMD_NEVER_UPDATE`.

🧪 Unit Tests:
No new unit tests were added in this PR. Existing focused coverage was run:

[...]

PR #44296, opened 2026-06-15
[XLA:GPU] A generic XLA runtime to support stable allocation addresses for command buffer.

shawnwang18:
📝 Summary of Changes
Add allocation address policy plumbing for GPU command-buffer execution on top of #44295.

Core idea: `AllocationAddressInfo` is the small runtime contract that tells command-buffer code which buffer allocation addresses are stable for the current execution.

The flow works as follows:
- `GpuExecutableBufferAllocator` builds the execution `BufferAllocations` and decides which allocation indices have persistent command-buffer-visible addresses. This can include global constants, VMM reservation-backed allocations, and any other allocation whose visible address is stable for the execution.
- `GpuExecutable` passes that information to thunks as `AllocationAddressInfo`. The `address_info_ready` flag tells consumers whether this policy is finalized; before it is ready, command-buffer code keeps the conservative update behavior.
- Once `address_info_ready` is true, `persistent_alloc_indices` is the source of truth for stable allocation addresses. Any allocation index not listed there is treated as dynamic by command-buffer consumers.
- `CommandBufferThunk` uses this policy to limit address-change checks to dynamic allocations when deciding whether a command buffer needs recording or updating.
- `CommandExecutor` uses the same policy at command granularity: if every allocation referenced by a command is persistent, that command can skip update recording because its command-buffer-visible addresses are stable.

This gives the runtime a generic way to expose persistent addresses to command-buffer execution without tying the command-buffer update logic to one specific stabilization mechanism.

This PR:
- Adds `Thunk::AllocationAddressInfo` as the runtime contract for allocation address stability.
- Threads optional allocation address info through `GpuExecutable`, PJRT GPU execution, thunk initialization, and thunk execution.
- Computes persistent allocation indices in `GpuExecutableBufferAllocator`, including global constant buffer allocations.
- Treats allocations outside `persistent_alloc_indices` as dynamically allocated for command-buffer update decisions.
- Updates `CommandBufferThunk` to check only dynamic allocation addresses when deciding whether a command buffer needs recording or updating.
- Updates `CommandExecutor` to skip command updates when every allocation referenced by a command is persistent.
- Caches command-buffer dynamic-allocation checks once the address info is ready.
- Adds and updates unit coverage for allocation-address policy behavior and allocation-index filtering.

🎯 Justification
#44295 moves command-buffer VA remapping into VMM-backed execution `BufferAllocations`, giving some command-buffer-visible addresses that can stay stable across executions.

This PR generalizes that stability into allocation address information. Commands that only reference persistent allocations do not need command-buffer updates just because underlying physical allocations changed. Commands that reference any non-persistent allocation still go through the existing dynamic-address update checks.

🚀 Kind of Contribution
✨ New Feature, ♻️ Cleanup, 🧪 Tests

N/A. This PR introduces runtime address-policy plumbing and cleanup; it does not claim a measured performance improvement.

🧪 Unit Tests:
Validated locally with:
- `bazel test --test_output=errors //xla/backends/gpu/runtime:command_buffer_thunk_test`
- `bazel test --test_output=errors //xla/backends/gpu/runtime:dynamic_slice_fusion_v2_thunk_test`
- `bazel test --test_output=errors --test_filter='CommandBufferThunkTest.UpdatePolicyIgnoresVaRemappedAllocations|GpuExecutableTest.CommandBufferAllocationIndexesIncludeMlirConstants' //xla/backends/gpu/runtime:command_buffer_thunk_test //xla/service/gpu:gpu_executable_test`

Additional validation:
- `bazel build --config=clang-tidy-noerrors //xla:debug_options_flags_test //xla/service/gpu:gpu_executable_buffer_allocator //xla/pjrt/gpu:se_gpu_pjrt_client`

🧪 Execution Tests:
Existing command-buffer execution coverage is exercised through the command-buffer thunk tests above. No additional multi-GPU execution test is added in this PR.

---

**Stack**

Previous PR: https://github.com/openxla/xla/pull/44295
Next PR: N/A

### DIFF
--- a/xla/backends/gpu/runtime/command_buffer_thunk.cc
+++ b/xla/backends/gpu/runtime/command_buffer_thunk.cc
@@ -15,6 +15,7 @@ limitations under the License.
 
 #include "xla/backends/gpu/runtime/command_buffer_thunk.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <iterator>
 #include <memory>
--- a/xla/backends/gpu/runtime/command_buffer_thunk.cc
+++ b/xla/backends/gpu/runtime/command_buffer_thunk.cc
@@ -53,18 +53,6 @@ namespace xla::gpu {
 using tsl::profiler::TraceMe;
 using tsl::profiler::TraceMeEncode;
 
-static bool CommandBufferUpdatesEnabled(
-    const CommandExecutor& commands,
-    const Thunk::AllocationAddressInfo* address_info) {
-  if (address_info == nullptr || !address_info->address_info_ready) {
-    return true;
-  }
-  DCHECK(absl::c_is_sorted(commands.allocs_indices()));
-  DCHECK(absl::c_is_sorted(address_info->persistent_alloc_indices));
-  return !absl::c_includes(address_info->persistent_alloc_indices,
-                           commands.allocs_indices());
-}
-
 //===----------------------------------------------------------------------===//
 // CommandBufferThunk
 //===----------------------------------------------------------------------===//
@@ -72,6 +60,24 @@ static bool CommandBufferUpdatesEnabled(
 CommandBufferThunk::ExecutorCommandBuffer::ExecutorCommandBuffer(
     std::unique_ptr<se::CommandBuffer> command_buffer)
     : command_buffer(std::move(command_buffer)) {}
+
+bool CommandBufferThunk::ExecutorCommandBuffer::HasDynamicAllocations(
+    const CommandExecutor& commands,
+    const Thunk::AllocationAddressInfo* address_info) {
+  if (address_info == nullptr || !address_info->address_info_ready) {
+    return true;
+  }
+
+  absl::call_once(
+      has_dynamic_allocations_once, [&]() ABSL_NO_THREAD_SAFETY_ANALYSIS {
+        DCHECK(absl::c_is_sorted(commands.allocs_indices()));
+        DCHECK(absl::c_is_sorted(address_info->persistent_alloc_indices));
+        has_dynamic_allocations = !absl::c_includes(
+            address_info->persistent_alloc_indices, commands.allocs_indices());
+      });
+
+  return has_dynamic_allocations;
+}
 
 CommandBufferThunk::CommandBufferThunk(
     CommandExecutor commands, ThunkInfo thunk_info,
@@ -243,11 +249,11 @@ absl::Status CommandBufferThunk::Initialize(const InitializeParams& params) {
   // not enabled), we also record them into the command buffer before execution.
   // This is required to guarantee that collective commands are recorded on all
   // participating ranks to avoid deadlocks.
-  bool updates_enabled =
-      CommandBufferUpdatesEnabled(commands_, params.allocation_address_info);
+  bool has_dynamic_allocations = cmd_buffer->HasDynamicAllocations(
+      commands_, params.allocation_address_info);
   if (cmd_buffer->command_buffer->state() ==
           se::CommandBuffer::State::kCreate ||
-      (updates_enabled && commands_.requires_update_on_initialize())) {
+      (has_dynamic_allocations && commands_.requires_update_on_initialize())) {
     VLOG(3) << "Initialize command buffer on device #"
             << params.executor->device_ordinal()
             << " by recoding command buffer cmd sequence"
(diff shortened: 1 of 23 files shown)


xla/backends/gpu/runtime/command_executor.cc
```diff
@@ -57,6 +57,20 @@ limitations under the License.
 namespace xla::gpu {
 
 namespace {
+bool IsSubsetOfSorted(absl::Span<const BufferAllocation::Index> subset,
```
akuegel: Why is this helper needed? Could we use absl::c_includes instead?

akuegel (approved):

xla/backends/gpu/runtime/command_executor.cc
shawnwang18: make sense. My previous understanding was wrong, I was thinking the custom helper is bit faster.  

Just fix to use absl::c_includes

xla/backends/gpu/runtime/command_buffer_thunk.cc
```diff
@@ -60,17 +61,33 @@ CommandBufferThunk::ExecutorCommandBuffer::ExecutorCommandBuffer(
...
 
+bool CommandBufferThunk::ExecutorCommandBuffer::HasDynamicAllocations(
+    const CommandExecutor& commands,
+    const Thunk::AllocationAddressInfo* address_info) {
+  if (address_info == nullptr || !address_info->address_info_ready) {
+    return true;
+  }
+
+  absl::call_once(
+      has_dynamic_allocations_once, [&]() ABSL_NO_THREAD_SAFETY_ANALYSIS {
```
ezhulenev: can you avoid `ABSL_NO_THREAD_SAFETY_ANALYSIS`? I think you can simply annotate with waht mutex is needed and it will pass static analysis

xla/backends/gpu/runtime/thunk.h
```diff
@@ -200,6 +200,26 @@ class Thunk {
...
+    // or incomplete for initial executions.
+    bool address_info_ready = false;
+
+    // Allocation indices whose device addresses are stable for this execution.
+    // This can include global constant allocations, allocations backed by
+    // reusable VMM reservation addresses, and any other allocations whose
+    // addresses are known to be persistent. Allocations not listed here should
+    // be treated as dynamically allocated by consumers that need address-change
+    // checks.
+    absl::Span<const BufferAllocation::Index> persistent_alloc_indices;
```
ezhulenev: I don't think that all this extra complexity with absl::call_once is needed, assuming that all allocs are sorted by index it's very cheap to recompute allocs that changed at run time. Set intersection of sorted containers is super cheap.

Also mutable `address_info_ready` has a very bad smell.

`AllocationAddressInfo != nullptr` - this should be the check on hot path, if you don't have VA remapping just pass nullptr, if you have VA remapping pass a non-null ptr value and let everyone figyure out things they need at run time.

Can you add a C++ micrbenchmark for utility function that does alloc intersection. I'm pretty sure the cost will be in microseconds, and a lot of state and absl::call_once flags could be removed. I'm pretty sure this extra compelcity will not have any impact on e2e perf.

xla/backends/gpu/runtime/command_buffer_thunk.cc
shawnwang18: fixed

xla/backends/gpu/runtime/thunk.h
shawnwang18: That makes sense, tried a small tests: 
```
 | Operation        | Case            | Commands | Persistent | Diff size | Cached hot (ns) | Recompute (ns) |
  |------------------|-----------------|----------|------------|-----------|-----------------|----------------|
  | c_includes       | all persistent  |        8 |          8 |         0 |            1.6  |           7.1  |
  | c_includes       | all persistent  |      128 |        128 |         0 |            1.6  |         219.6  |
  | c_includes       | all persistent  |      512 |        512 |         0 |            1.6  |         956.1  |
  | c_includes       | mostly dynamic  |      512 |         32 |       480 |            1.6  |          31.4  |
  | c_set_difference | all persistent  |        8 |          8 |         0 |            1.7  |           8.0  |
  | c_set_difference | all persistent  |      128 |        128 |         0 |            1.6  |         198.5  |
  | c_set_difference | all persistent  |      512 |        512 |         0 |            1.6  |         957.5  |
  | c_set_difference | half dynamic    |      512 |        512 |       256 |            1.6  |        1114.8  |
  | c_set_difference | mostly dynamic  |      512 |         32 |       480 |            1.6  |         342.3  |
```

The save is small, and I am removing the call_once design.

ezhulenev (approved):

akuegel (approved):

PR #44322, opened 2026-06-15
Clean up CUPTI V2 subscriber fallback handling

zhenying-liu:
📝 Summary of Changes
 Make CUPTI V2 subscriber fallback handling more robust.

This change adds CUPTI wrapper/interface coverage for subscriber-scoped V2 timestamp and activity-record APIs, updates tracer timestamp selection for V2 subscribers, and preserves legacy fallback behavior when V2 APIs are unavailable. It also keeps the V2 subscriber alive until cached activity buffers are parsed, supports optional V2 subscriber reuse across profiling sessions, and adds targeted tests for timestamp fallback, retained subscriber reuse, V2 activity callback reuse, and tolerated V2 API errors.

🎯 Justification
  This is important for profiler compatibility across CUPTI versions while preserving CUPTI V2 subscriber support.

 In CUPTI 13.2, using the legacy `cuptiGetTimestamp()` with a V2 subscriber is allowed. In CUPTI 13.3, `cuptiGetTimestamp_v2()` is introduced and required for the V2 subscriber path. The implementation therefore probes `cuptiGetTimestamp_v2()` when available, uses it on success, and falls back to legacy `cuptiGetTimestamp()` for compatibility when CUPTI reports `CUPTI_ERROR_NOT_SUPPORTED` or `CUPTI_ERROR_UNKNOWN`.

 `CUPTI_ERROR_NOT_COMPATIBLE` is handled separately because it indicates that the legacy timestamp path should not be probed in the current V2 subscriber state. In that case, the tracer returns `0` rather than mixing timestamp APIs in an unsupported state.

 This benefits GPU profiling workloads that run with newer CUPTI V2 multi-subscriber support, while keeping profiler behavior compatible with CUDA/CUPTI installations that do not yet expose the full 13.3 V2 timestamp API.

🚀 Kind of Contribution
🐛 Bug Fix, ♻️ Cleanup, 🧪 Tests

Not applicable.

🧪 Unit Tests:
  Added/updated `//xla/backends/profiler/gpu:cupti_error_manager_test` coverage for:
  - V2 timestamp unavailable fallback paths
  - V2 timestamp `NOT_COMPATIBLE` handling
  - retained V2 subscriber reuse
  - V2 activity callback registration reuse
  - V2 attribute errors that should not disable CUPTI
  - V2 activity-record parsing fallback behavior

🧪 Execution Tests:
  No new execution tests were added.

### DIFF
--- a/xla/backends/profiler/gpu/cupti_tracer.cc
+++ b/xla/backends/profiler/gpu/cupti_tracer.cc
@@ -1702,6 +1702,10 @@ uint64_t CuptiTracer::GetLegacyTimestamp() const {
       cupti_interface_->GetTimestamp(&tsc) == CUPTI_SUCCESS) {
     return tsc;
   }
+  LOG_FIRST_N(WARNING, 1)
+      << "CUPTI GetTimestamp failed; returning timestamp 0. Events that "
+         "depend on this timestamp may be dropped during time normalization.";
+  // Return 0 to mark the timestamp unavailable so profiling can continue.
   return 0;
 }
 
@@ -1740,7 +1744,12 @@ uint64_t CuptiTracer::GetTimestampForCurrentSubscriber() const {
   if (use_legacy_timestamp_with_v2_subscriber_) {
     return GetLegacyTimestamp();
   }
-  // Return 0 rather than mixing V1 and V2 timestamp domains.
+  LOG_FIRST_N(WARNING, 1)
+      << "CUPTI timestamp is unavailable for the current V2 subscriber state; "
+         "returning timestamp 0.";
+  // Return 0 to mark the timestamp unavailable for this V2 subscriber state so
+  // profiling can continue. Events that depend on this timestamp may be dropped
+  // during time normalization.
   return 0;
 }
 
--- a/xla/backends/profiler/gpu/cupti_tracer.cc
+++ b/xla/backends/profiler/gpu/cupti_tracer.cc
@@ -1037,15 +1037,16 @@ class CuptiDriverApiHookWithActivityApi : public CuptiDriverApiHook {
                                 CUpti_CallbackId cbid,
                                 const CUpti_CallbackData* cbdata) override {
     // Stash away the current Cupti timestamp into cbdata.
-    *cbdata->correlationData =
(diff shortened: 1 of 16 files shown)


xla/backends/profiler/gpu/cupti_tracer.cc
```diff
@@ -1651,6 +1696,54 @@ absl::Status CuptiTracer::Finalize() {
...
+    return CuptiTracer::GetTimestamp();
+  }
+  if (using_v2_timestamp_api_ &&
+      cupti_interface_->GetTimestampV2(subscriber_, &tsc) == CUPTI_SUCCESS) {
+    return tsc;
+  }
+  if (use_legacy_timestamp_with_v2_subscriber_) {
+    return GetLegacyTimestamp();
+  }
+  // Return 0 rather than mixing V1 and V2 timestamp domains.
```
Tixxx: what does 0 time stamp represent here? it'd be better if there's a warning to users to explain that here.

xla/backends/profiler/gpu/cupti_tracer.cc
```diff
@@ -1651,6 +1696,54 @@ absl::Status CuptiTracer::Finalize() {
   return 0;
 }
 
+uint64_t CuptiTracer::GetLegacyTimestamp() const {
+  uint64_t tsc;
+  if (cupti_interface_ != nullptr &&
+      cupti_interface_->GetTimestamp(&tsc) == CUPTI_SUCCESS) {
+    return tsc;
+  }
```
Tixxx, line 1704: same here, warning or info msg would be helpful

Tixxx (approved):

xla/backends/profiler/gpu/cupti_tracer.cc
zhenying-liu: Done.

xla/backends/profiler/gpu/cupti_tracer.cc
zhenying-liu: Done.

dimitar-asenov (approved):

PR #44335, opened 2026-06-15
[ROCm] pjrt: add kAddress synchronous passthrough allocator

magaonka-amd:
📝 Summary of Changes
Since https://github.com/openxla/xla/pull/41761 , we don't have option to select stream address allocator because xla backend now defaults to BFC
I originally made https://github.com/openxla/xla/pull/42627 this PR under the assumption that platfrom allocator always means address allocator. after comments/suggestions from G I now have followup change.

"I think we need a different solution. platform means that it shares the allocator with tf which is why the jax2tf tests are failing. We cannot change the meaning of platform and either need to allow configuring the underlying allocator to be the address allocator or a separate option for pjrt which changes the allocator."

based on this: I decided to go with second option i,e separate option for pjrt which changes the allocator.

Add a new allocator kind `GpuAllocatorConfig::Kind::kAddress`, exposed via
`XLA_PYTHON_CLIENT_ALLOCATOR=address`. It constructs a dedicated synchronous
passthrough `se::StreamExecutorAddressAllocator` at the PJRT level and bypasses
the BFC allocator entirely.

🎯 Justification

On JAX pytests we rely on a passthrough (non-BFC) allocator for testing  because BFC can cause OOM issues quickly, also address allocator is fairly simple this makes it easy to use while debugging too. That capability used to be reachable
via `platform`, 

🚀 Kind of Contribution
✨ New Feature

🧪 Unit Tests:
`AddressAllocatorIsSynchronousPassthrough` asserts that with `kind = kAddress`
the resulting client's allocator is a `StreamExecutorAddressAllocator` and not a
`MultiDeviceAdapter` (i.e. the BFC allocator is bypassed).

This PR requires pairing JAX changes to make it effective from jax side I tested locally that address allocator can be set from JAX. with my local changes. Ill open JAX PR soon.


magaonka-amd:
@pemeliya  this is approved in downstream so opening it up here .

@pschuh  this is the follow up to the suggestion you had for my prev PR to restore stream address allocator 
https://github.com/openxla/xla/pull/42627

I went with making separate pjrt option for the allocator and gave it address allocator as name.
at your convenience can you please provide feedback on this approach.

pschuh (approved):

magaonka-amd:
FYI, pushed small edit to test file, it didn't have "address" string in it. sorry about that.

pschuh (approved):

magaonka-amd:
Hello Copybara seem to have issue with picking this change up. 

also Tensorflow pre-submit failed with unrelated issue to this PR
```bash
E0000 00:00:1781560826.656115      66 session.cc:95] Failed to create session: INTERNAL: cudaSetDevice() on GPU:0 failed. Status: uncorrectable ECC error encountered
E0000 00:00:1781560826.656177      66 c_api.cc:2243] INTERNAL: cudaSetDevice() on GPU:0 failed. Status: uncorrectable ECC error encountered
```

can I get help in merging this PR if everything else is okay.

PR #44338, opened 2026-06-15
[XLA:CPU] Fix binary add post-op rank mismatch for oneDNN ndims requirement

qgao007:
📝 Summary of Changes
Fix rank mismatch in oneDNN binary add post-ops.

### DIFF
--- a/xla/service/cpu/onednn_contraction_rewriter.cc
+++ b/xla/service/cpu/onednn_contraction_rewriter.cc
@@ -375,6 +375,16 @@ absl::StatusOr<Shape> AdjustAddendShape(const HloInstruction* contraction,
           },
           addend->shape());
     }
+
+    // oneDNN requires binary post-op source dimensions to match
+    // the destination dimensions. Prepend dimensions with size 1 if needed.
+    auto addend_rank = addend->shape().dimensions().size();
+    auto output_rank = contraction->shape().dimensions().size();
+    if (addend_rank < output_rank) {
+      std::vector<int64_t> ones(output_rank - addend_rank, 1);
+      return ShapeUtil::InsertDimensionsAtIndex(addend->shape(), 0, ones);
+    }
+
     return addend->shape();
   }
   if (broadcast_instr->opcode() != HloOpcode::kBroadcast) {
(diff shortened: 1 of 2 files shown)


othakkar:
@penpornk Could you please help review this PR? Without it, we see a `core dumped` issue with binary add post-ops. Thanks!

xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -886,6 +886,21 @@ class OneDnnContractionRewriteVisitor : public DfsHloRewriteVisitor {
...
+      } else if (kind == OneDnnFusionConfig::BINARY_ADD) {
+        // oneDNN's binary post-op operand must match the output rank. Expand a
+        // lower-rank addend with leading size-1 dimensions via a Bitcast.
+        const Shape& out_shape = contraction->shape();
+        int64_t missed_rank = out_shape.dimensions().size() -
+                              addend->shape().dimensions().size();
+        if (missed_rank > 0) {
+          std::vector<int64_t> ones(missed_rank, 1);
+          Shape expanded_shape =
+              ShapeUtil::InsertDimensionsAtIndex(addend->shape(), 0, ones);
```
penpornk, line 898: The missing rank doesn't always need to be index 0, right?
If we are broadcasting along dim 1 instead, we will have the same problem again. E.g.,
```
const.0 = f32[32,128] constant({...})
broadcast.0 = f32[32,128,128] broadcast(const.0), dimensions={0,2}
```

penpornk (commented):
Sorry for the delay!

xla/service/cpu/onednn_contraction_rewriter.cc
othakkar: Yes, the missing dims are always at index 0. [`IsOperandFusible`](https://github.com/openxla/xla/blob/bdbeb1eb4d50206d3431cb38c440ef9b86f7c6b6/xla/service/cpu/onednn_contraction_rewriter.cc#L542) enforces right-alignment, so by the time we reach this block, any rank gap is guaranteed to be leading. 

The prepended 1s are purely to satisfy oneDNN's requirement that binary post-op operand rank equals the output rank.

For your example, [`AdjustAddendShape`](https://github.com/openxla/xla/blob/bdbeb1eb4d50206d3431cb38c440ef9b86f7c6b6/xla/service/cpu/onednn_contraction_rewriter.cc#L459) produces a shape like `fp32[32, 1, 128]`. If the dot is already rank 3, there's no rank mismatch, and this code is a no-op.

penpornk: Thank you for the clarification!

penpornk (approved):

PR #44408, opened 2026-06-16
ifrt_proxy: validate peer-supplied RemapPlan and shard specs before use

PwntGambit:
Following the input-validation hardening in #44083 and #44260, validate two more peer-supplied structures in the IFRT proxy server:

- HandleRemapArraysRequest: call RemapPlan::Validate() after FromProto so a plan with inconsistent in_array/out_array indices or shard intervals is rejected before RemapArrays acts on it.
- CheckRange: also verify the last stepped interval index stays within num_shards. The existing `end <= num_shards + step - 1` bound is necessary but not sufficient for large steps; ComputeDeviceListFromIntervals already guards its loop the same way.
- HandleMakeArraysFromHostBufferShardsRequest: require addressable_shard_ indices and host_buffers to have equal length before the parallel loop.

No functional change for valid inputs; inconsistent protos now return InvalidArgument instead of relying on downstream behavior.

📝 Summary of Changes

🎯 Justification
change.

🚀 Kind of Contribution

🧪 Unit Tests:
transformed HLO.

🧪 Execution Tests:

### DIFF
--- a/xla/python/ifrt/remap_plan.cc
+++ b/xla/python/ifrt/remap_plan.cc
@@ -145,15 +145,12 @@ absl::Status CheckRange(int64_t num_shards,
                            interval.end);
   }
   // The `end` bound above is necessary but not sufficient: with a large `step`,
-  // the last stepped index `start + (num_steps - 1) * step` can exceed
-  // `num_shards` while still satisfying `index < end`, which would lead to
-  // out-of-bounds indexing of the per-shard buffers. Verify the last stepped
-  // index explicitly.
+  // the last stepped index can exceed `num_shards` while still satisfying
+  // `index < end`, which would lead to out-of-bounds indexing of the per-shard
+  // buffers. Verify the last stepped index explicitly.
   if (interval.end > interval.start) {
-    const int64_t num_steps =
-        (interval.end - interval.start + interval.step - 1) / interval.step;
     const int64_t last_index =
-        interval.start + (num_steps - 1) * interval.step;
+        interval.end - 1 - (interval.end - 1 - interval.start) % interval.step;
     if (last_index >= num_shards) {
       return InvalidArgument(
           "interval addresses shard %d, which is out of range [0, %d)",
--- a/xla/python/ifrt/remap_plan.cc
+++ b/xla/python/ifrt/remap_plan.cc
@@ -144,6 +144,22 @@ absl::Status CheckRange(int64_t num_shards,
                            num_shards + interval.step - 1, interval.step,
                            interval.end);
   }
+  // The `end` bound above is necessary but not sufficient: with a large `step`,
+  // the last stepped index `start + (num_steps - 1) * step` can exceed
+  // `num_shards` while still satisfying `index < end`, which would lead to
+  // out-of-bounds indexing of the per-shard buffers. Verify the last stepped
+  // index explicitly.
+  if (interval.end > interval.start) {
+    const int64_t num_steps =
+        (interval.end - interval.start + interval.step - 1) / interval.step;
+    const int64_t last_index =
+        interval.start + (num_steps - 1) * interval.step;
+    if (last_index >= num_shards) {
+      return InvalidArgument(
+          "interval addresses shard %d, which is out of range [0, %d)",
+          last_index, num_shards);
+    }
+  }
   return absl::OkStatus();
 }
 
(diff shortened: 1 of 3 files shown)


PwntGambit:
@googlebot I signed it!

xla/python/ifrt/remap_plan.cc
```diff
@@ -144,6 +144,22 @@ absl::Status CheckRange(int64_t num_shards,
...
                            interval.end);
   }
+  // The `end` bound above is necessary but not sufficient: with a large `step`,
+  // the last stepped index `start + (num_steps - 1) * step` can exceed
+  // `num_shards` while still satisfying `index < end`, which would lead to
+  // out-of-bounds indexing of the per-shard buffers. Verify the last stepped
+  // index explicitly.
+  if (interval.end > interval.start) {
+    const int64_t num_steps =
+        (interval.end - interval.start + interval.step - 1) / interval.step;
```
akuegel: I believe computing `last_index` can be simplified:

`const int64_t last_index = interval.end - 1 - (interval.end - 1 - interval.start) % interval.step`.

PwntGambit: Thanks @akuegel — done in the latest commit; simplified `last_index` to the direct form you suggested (equivalent result).

akuegel (approved):

PwntGambit:
agreed, simplified and sent

xla/python/ifrt_proxy/server/ifrt_backend.cc
```diff
@@ -1092,6 +1101,11 @@ IfrtBackend::HandleRemapArraysRequest(ArrayStore::Reservation& asr,
                    array_store_.Find(remap_request.array_handles()));
   ASSIGN_OR_RETURN(RemapPlan plan,
                    RemapPlan::FromProto(client_.get(), remap_request.plan()));
+  // The plan is deserialized from an untrusted peer and `FromProto` performs no
```
hyeontaek: This part of the change is not desirable for the expected setup for IFRT Proxy.

(1) We ask the users to build a single trust domain for both IFRT Proxy client and server (using their own way to secure RPCs and prevent unauthorized access). Thus, having a potential untrusted peer is a failure to follow this requirement and the user would change their setup.

(2) We should not do `plan.Validate()` for every `RemapArrays` call because it is very slow. It has been the caller's responsibility to construct a valid `RemapPlan` and run validation as needed, especially when the same plan is used across multiple remap calls.

(3) If a bad range is used, the IFRT implementations used by IFRT Proxy server (e.g., pjrt_ifrt) can check it. If you think their handling of bad ranges is insufficient, please address the issue there instead of adding this validation.

xla/python/ifrt/remap_plan.cc
```diff
@@ -144,6 +144,19 @@ absl::Status CheckRange(int64_t num_shards,
                            num_shards + interval.step - 1, interval.step,
                            interval.end);
   }
+  // The `end` bound above is necessary but not sufficient: with a large `step`,
+  // the last stepped index can exceed `num_shards` while still satisfying
+  // `index < end`, which would lead to out-of-bounds indexing of the per-shard
+  // buffers. Verify the last stepped index explicitly.
+  if (interval.end > interval.start) {
```
hyeontaek, line 151: Do you mind adding a test that confirms the effectiveness of this hardened check? That will prevent a future backsliding. Thanks!

xla/python/ifrt_proxy/server/ifrt_backend.cc
```diff
@@ -966,6 +966,15 @@ IfrtBackend::HandleMakeArraysFromHostBufferShardsRequest(
   for (const auto& spec_proto : make_arrays_request->specs()) {
     xla::ifrt::Client::MakeArraysFromHostBufferShardsSpec::Buffers buffers;
     buffers.reserve(spec_proto.host_buffers_size());
+    if (spec_proto.addressable_shard_indices_size() !=
+        spec_proto.host_buffers_size()) {
+      return absl::InvalidArgumentError(absl::StrCat(
+          "MakeArraysFromHostBufferShardsSpec has ",
+          spec_proto.addressable_shard_indices_size(),
+          " addressable_shard_indices but ", spec_proto.host_buffers_size(),
+          " host_buffers; these parallel repeated fields must have equal "
```
hyeontaek: This later part ("; these...") can be omitted because this is already implied in the earlier part of the error message

hyeontaek (changes requested):
Thanks!

xla/python/ifrt_proxy/server/ifrt_backend.cc
PwntGambit: @hyeontaek have a look, let me know if that's what you're looking for, and if there are any changes I need to make.

hyeontaek (approved):

allanrenucci (approved):

PR #44428, opened 2026-06-17
[ROCm] Emit packed bf16 atomic add for scatter/segment_sum by matchin…

zoranjovanovic-ns:
…g FloatNormalization conversions.

📝 Summary of Changes
Make atomic-RMW matcher (GetAtomicModifierParameters) to look through the extf → addf(f32) → truncf body that FloatNormalization emits for bf16, recovering the narrow bf16 modifier so scatter-add lowers to packed atomicrmw fadd <2 x bf16> (global_atomic_pk_add_bf16) instead of a CAS loop. GpuFloatSupport/FloatNormalization are unchanged; targets without a native bf16 atomic still fall back to CAS.

🎯 Justification
bf16 segment_sum/scatter-add result in slow CAS loop on MI300/MI350 despite the HW having a packed bf16 atomic, making bf16 ~7x slower than f16.

🚀 Kind of Contribution
🧪 Tests

🧪 Unit Tests:
Added direct_atomic_rmw_fadd_bf16_widened + a gfx942 CHECK-GFX942-MI300 RUN line to lower_tensors.mlir, asserting the packed atomicrmw fadd <2 x bf16> with no CAS. All 9 RUN-line prefixes pass.

🧪 Execution Tests:

### DIFF
--- a/xla/codegen/emitters/transforms/atomic_rmw_utils.cc
+++ b/xla/codegen/emitters/transforms/atomic_rmw_utils.cc
@@ -84,6 +84,16 @@ std::optional<ml::AtomicBinOp> GetAtomicBinOp(Operation* modifier_op,
       .Default([](Operation* op) { return std::nullopt; });
 }
 
+// Looks through an arith.extf widening cast and returns the narrower source
+// value. Low-precision floating-point reductions (e.g. bf16) are computed in a
+// wider type, so the atomic modifier appears in the body as extf(modifier).
+Value LookThroughExtF(Value value) {
+  if (auto ext = value.getDefiningOp<arith::ExtFOp>()) {
+    return ext.getIn();
+  }
+  return value;
+}
+
 }  // namespace
 
 // Returns atomic op modifier and the atomic bin op kind.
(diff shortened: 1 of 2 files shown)


pifon2a (approved):

xla/codegen/emitters/transforms/atomic_rmw_utils.cc
```diff
@@ -103,19 +112,67 @@ std::optional<std::pair<Value, ml::AtomicBinOp>> GetAtomicModifierParameters(
...
   if (!kind.has_value()) {
     return std::nullopt;
   }
-  // Find the modifier arg that does not match the argument of `atomic_rmw`
-  // body.
-  Value block_arg = op.getBody()->getArgument(0);
-  Value modifier_arg = modifier_op->getOperand(0) == block_arg
-                           ? modifier_op->getOperand(1)
-                           : modifier_op->getOperand(0);
+  Value lhs = LookThroughExtF(modifier_op->getOperand(0));
```
draganmladjenovic, line 161: Both should have extf and from the same type? Not sure if we can have case where rhs is constant. In that case one should ensure that it is exactly representable as bf16.

PR #44429, opened 2026-06-17
[xla] Propagate HangWatchdog execution timeouts through CoordinationService

### LINKED ISSUE #38503: Multi-host job hangs forever when one rank is alive-but-stuck in an NCCL collective
## Description

On a multi-host GPU job, if a single process stops entering a collective while staying **alive**
(process up, JAX coordination heartbeat still beating), every other process blocks **forever** inside
the NCCL collective. The job makes no progress and prints **no error and no warning** — it hangs until
something external kills it. I can't find any native timeout that aborts this, which makes an
orchestrator-driven auto-restart impossible.

These xla flags do not seem to work:
- `heartbeat_timeout_seconds` only catches a **dead** process. the stuck rank is alive and keeps
  heartbeating, so it's never flagged.
- `initialization_timeout` only controls the startup handshake, not a mid-run hang.
- `xla_gpu_nccl_termination_timeout_seconds` (the one flag that sounds right — *"Timeout in seconds
  before terminating jobs stuck in NCCL Rendezvous"*) does **not** fire for this (see matrix below).

### Minimal reproducer (model-free)
One process per host (≥2 hosts). At step 5, `process_index == 1` stops entering the all-reduce, so
every other rank blocks in `jax.lax.psum` forever — no error, no timeout. No model, no data.

```python
import os, time
import jax, jax.numpy as jnp, numpy as np
from jax import shard_map
from jax.sharding import Mesh, PartitionSpec as P

STALL_STEP, STALL_RANK = 5, 1
ELEMS = 64 * 1024 * 1024  # fp32 per-device shard = 256 MiB, enough to engage real NCCL traffic

jax.distributed.initialize(
    coordinator_address=os.environ["COORDINATOR_ADDRESS"],
    num_processes=int(os.environ["NUM_PROCESSES"]),
    process_id=int(os.environ["PROCESS_ID"]),
)
rank = jax.process_index()
mesh = Mesh(np.array(jax.devices()), ("dev",))

# Array sharded across ALL devices of ALL hosts, so the psum is a genuine cross-host collective.
x = jax.make_array_from_process_local_data(
    jax.sharding.NamedSharding(mesh, P("dev")),
    jnp.ones(jax.local_device_count() * ELEMS, jnp.float32),
    (jax.device_count() * ELEMS,),
)
all_reduce = jax.jit(lambda v: shard_map(
    lambda a: jax.lax.psum(a, "dev"), mesh=mesh, in_specs=P("dev"), out_specs=P("dev"))(v))

for step in range(1000):
    if step == STALL_STEP and rank == STALL_RANK:
        print(f"rank {rank}: stalling at step {step} — peers will block forever", flush=True)
        time.sleep(1e9)
    x = all_reduce(x)
    x.block_until_ready()
    print(f"rank {rank}: step {step} done", flush=True)
```

### How to run (≥2 hosts, one process each)
```
# host 0:  COORDINATOR_ADDRESS=<host0>:1234 NUM_PROCESSES=2 PROCESS_ID=0  python nccl_stall_repro.py
# host 1:  COORDINATOR_ADDRESS=<host0>:1234 NUM_PROCESSES=2 PROCESS_ID=1  python nccl_stall_repro.py
```

### What I observe
All ranks print `step 0 done` … `step 4 done`; rank 1 then prints `stalling at step 5` and sleeps, and
every other rank blocks at step 5 in the all-reduce — indefinitely, with **no error and no warning**.

### Native flags I tried (all on jax 0.9.0.1, stall injected at step 5, each run ≫ its timeout)
| Flag (value) | Recognized by XLA_FLAGS? | Result |
|---|---|---|
| *(none — control)* | — | **Hangs indefinitely**, silent |
| `xla_gpu_nccl_termination_timeout_seconds=120` | yes (loads) | **Hangs** — never fires for an alive-but-stuck rank |
| `xla_gpu_first_collective_call_terminate_timeout_seconds=120` | yes | **Hangs** — only covers the *first* collective; stall is mid-run |
| `xla_gpu_nccl_termination_timeout_seconds=120` + `xla_gpu_nccl_terminate_on_error=true` | yes | **Hangs** — no NCCL *error* is raised, just a clean hang |
| all three above combined | yes | **Hangs** — even all recognized catchers together never abort |
| `xla_gpu_executable_terminate_timeout=120` | yes | **Hangs** — this guards only the process-local executable rendezvous, not a cross-host in-flight collective |
| `xla_gpu_executable_warn_stuck_timeout=30` | yes | **Hangs** — warn-only, process-local |

### Expected vs actual
- **Expected:** a native, configurable timeout that aborts the process when a collective makes no
  progress for N seconds.
- **Actual:** nothing aborts it; the job hangs indefinitely and silently (no error, no warning). The
  timeouts in 0.9.0.1 only cover collective/rendezvous *init* or the process-local rendezvous — not an
  in-flight cross-host collective.

### Workaround
We use a small Python watchdog thread that `os.abort()`s if no step completes within a deadline. It
works (it fires even while a rank is blocked in `block_until_ready`), but it's external and coarse.

For reference, PyTorch's NCCL backend has this natively: `ProcessGroupNCCL` runs a watchdog that
times out a collective which doesn't complete within `timeout` and aborts (`TORCH_NCCL_*` knobs). We're
looking for the JAX/XLA equivalent.

### Questions
1. Are `xla_gpu_execution_progress_tracking` / `xla_gpu_execution_terminate_timeout` (in XLA `main`)
   the intended fix? Do they abort a host-side / on-GPU stuck-but-alive rank in an **in-flight**
   collective, not just an NCCL-rendezvous-init stall?
2. Should `xla_gpu_nccl_termination_timeout_seconds` cover this case? If not, what layer is expected
   to catch a rank that is alive but never enters the collective?
3. Is there any supported way on 0.9.x to get a per-execution progress watchdog? (The
   `xla_gpu_executable_terminate_timeout` flag exists but only guards the process-local rendezvous, so
   it doesn't cover a cross-host in-flight collective.)

## System info (python version, jaxlib version, accelerator, etc.)
Output of `import jax; jax.print_environment_info()` on a worker:
```
jax:    0.9.0.1
jaxlib: 0.9.0.1
numpy:  2.1.3
python: 3.12.12
device info: NVIDIA GB200-4, 4 local devices per host
platform: Linux 6.17.0 aarch64
nvidia-smi: Driver 580.126.20, CUDA 13.2, NVIDIA GB200 (waiting ranks at 100% GPU util during the hang)
```

Liauuu:
Fixes https://github.com/jax-ml/jax/issues/38503
Related to JAX Issue #38503

**## 📋 Context & Problem Statement**

This PR directly addresses the cross-host deadlock issue described in JAX #38503 

In a distributed multi-host setup (as reproduced on NVIDIA GB200 clusters under JAX 0.9.0.1), if a single rank or host becomes "alive-but-stuck" (e.g., encountering a hardware/driver lag or a simulated stall via `time.sleep`), the remaining healthy nodes block indefinitely inside the NCCL collective layer (such as cross-host `all_reduce` / `jax.lax.psum`). 

Because the stuck rank remains technically alive, the JAX coordination heartbeat continues to beat, meaning native cluster-level timeouts fail to detect the failure. Consequently, the healthy hosts hang permanently at 100% GPU utilization without raising any native errors or warnings, preventing orchestrator-driven automatic restarts.

**### 📊 Expected Execution Flow Verification (Before vs After)**

❌ BEFORE (As Observed in Issue Tracker) * **Host 0 (or any stalling rank)**: Hits a local timeout or code freeze (e.g., `time.sleep(1e9)` or hardware lag). * **Host 1 (and all other peers)**: Print `step 4 done` and proceed to step 5. However, because Host 0 never joins the step 5 collective, all other alive ranks block indefinitely inside the NCCL `all_reduce` execution layer. * **Result**: The cluster encounters a silent, permanent hang with no error messages or native warnings raised, requiring an external manual watchdog or `SIGKILL`. 

🟢 AFTER 
* **Host 0 (or timing-out rank)**: The `HangWatchdog` detects the lack of execution progress and immediately invokes the newly wired `ExecutionTimeoutHandler`.
* **RPC Error Reporting**:Host 0 uses the RPC layer to invoke `CoordinationServiceAgent::ReportError`, safely transmitting the `DeadlineExceededError` payload to the central coordination cluster.
* **Global Broadcast & Unwind**: The central Coordination Service receives this RPC notification and instantly broadcasts the task failure flag to all remaining healthy nodes in the distributed network.
* **Result**: Upon receiving the broadcast, Host 1 (and all other peers) invoke `AbortCollectivesOnTaskFailure`, unblocking the stalled NCCL transport layer and allowing the entire multi-host job to exit cleanly with a clear stack trace rather than freezing permanently.

**##### 🎯 Architectural Decision: Why Reuse CoordinationService?

1. **Limitation of Local Watchdogs in Distributed Environments**: 
   In a single-host environment (e.g., a single node with 8 GPUs), when one GPU stalls and causes a collective NCCL hang, the local `HangWatchdog::Abort()` safely teardowns the single process to resolve the issue. However, in a multi-host setup, each `HangWatchdog` operates strictly within its own process boundary and lacks the RPC capability to communicate with remote peers. If a local watchdog simply cancels or aborts locally, remote hosts remain entirely oblivious and block indefinitely inside the NCCL collective transport layer.

2. **Preventing Infrastructure Bloat and Network Pollution**:
   While we could theoretically extend the watchdog's internal logic to forcefully terminate remote peers, doing so introduces severe trade-offs. Forcefully breaking in-flight collective networks without a structured protocol can pollute the distributed cluster state and lead to cascading infrastructure failures. Furthermore, embedding complex network orchestration into a component originally designed for simple timeout tracking violates the single-responsibility principle and leads to code maintainability issues.

3. **Leveraging Production-Proven Infrastructure**:
   Therefore, this PR respectfully reuses XLA's established `CoordinationService` fault-tolerance infrastructure. Instead of inventing a new signaling mechanism, the local watchdog simply uses the existing RPC layer to report the execution timeout to the central coordination agent. This allows the coordination service to gracefully broadcast the failure and trigger a clean, synchronized collective unwind via the `CancellationToken` across all participating hosts.


xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -1141,4 +1141,38 @@ absl::Status UpdateGlobalProcessInfo(absl::Span<coordination::TaskInfo> infos) {
...
+                                           const absl::Status& error) {
+  ProcessGpuCliques& state = GetProcessGpuCliques();
+  absl::MutexLock lock(state.mu);
+  if (state.task_state_infos.empty()) {
+    return absl::OkStatus();
+  }
+
+  std::vector<coordination::TaskInfo> updated = state.task_state_infos;
+  bool found = false;
+  for (coordination::TaskInfo& info : updated) {
```
ezhulenev: prefer this, to loops with mutable state

```
auto task = absl::c_find(updated, predicate)
if (task == updated.end()) errpr
```

xla/service/gpu/gpu_executable.cc
```diff
@@ -683,10 +684,30 @@ absl::Status GpuExecutable::ExecuteThunksImpl(
...
 
-    guard = HangWatchdog::Global().Watch(
-        watchdog_name, watchdog_timeout,
-        HangWatchdog::Abort(watchdog_name, watchdog_timeout,
-                            std::move(pre_abort)));
+    const gpu::GpuExecutableRunOptions* gpu_run_options =
+        run_options->run_options().gpu_executable_run_options();
+
+    HangWatchdog::CancelCallback on_timeout;
+    if (gpu_run_options != nullptr &&
```
ezhulenev: nit: `if (gpu_run_options && gpu_run_options->`

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1912,6 +1914,37 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
     gpu_run_options->set_enable_mock_collectives();
   }
 
+  if (options.abort_collectives_on_failure) {
+    int process_index = options.node_id;
+    std::shared_ptr<DistributedRuntimeClient> distributed_client =
+        options.distributed_client;
+    gpu_run_options->set_execution_timeout_handler(
+        [process_index,
```
ezhulenev: Just capture fields you need in lambda, should be shorter than declaring local variables

`[process_index = options.node_id, distributed_client= options.distributed_client, ...`

ezhulenev (approved):

ezhulenev:
@Liauuu FWIW in practice I think when hang watchdog fires, it's often because GPU state is borked and even if you cancel collectives it's a pretty high chance that you will not be able to continue. I'd consider having a second hang watchdog at +30 seconds (or +1 min) that will abort the process if collective cancellation didn't help. 

My original motivation was to kill the task and let higher level batch scheduler (i.e. Slurm) to restart the job, because the  chance of successful recovery is pretty slim. But given that it's under a flag, we can give a try, maybe I'm too pessimistic.

Liauuu:
Hi @ezhulenev,

Thank you very much for the detailed architectural feedback and for taking the time to review this — I really appreciate it.

Your point about the “borked GPU state” and the need for a hard kill fallback for higher-level schedulers like Slurm made a lot of sense, and I agree that it’s an important safety net for production robustness.

Following your suggestion, I’ve added a second-stage hang watchdog:

Inside the execution timeout handler, a detached background thread is spawned immediately.
The fallback timeout is set to 60 seconds to allow more time for the first-stage collective cancellation to complete cleanly.
If the process is still alive after the timeout (i.e., the cancellation path itself is stuck), it logs an error and triggers std::abort() as a last resort.

I’ve also addressed the other style nits you pointed out, including simplifying the pointer check, using absl::c_find_if, and inlining lambda captures. All necessary headers have been verified as well.

Thanks again for the helpful guidance — I’d really appreciate another look when you have time.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1915,13 +1917,19 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
...
-    int process_index = options.node_id;
-    std::shared_ptr<DistributedRuntimeClient> distributed_client =
-        options.distributed_client;
     gpu_run_options->set_execution_timeout_handler(
-        [process_index,
-         distributed_client = std::move(distributed_client)](
+        [process_index = options.node_id,
+         distributed_client = options.distributed_client](
             absl::string_view action, absl::Duration timeout) {
+          std::thread([action = std::string(action)] {
```
ezhulenev: std::threads are banned in XLA codebase, instead of doing it here you can re-assign `guard` in gpu_executable to new watchdog, essentially replace already fired watchdog with a new one. Be carefult to do it before cancelling collectives to avoid data races.

Liauuu: Hi @ezhulenev,

I’ve updated the implementation to remove std::thread, in order to better align with the XLA codebase guidelines, and refactored the design to use HangWatchdog::Guard re-assignment as suggested.

To address potential dangling reference issues during async execution (since the lambda may outlive the original stack frame where the guard is created, and the guard is not fully assigned until after the lambda is constructed), I introduced a heap-backed std::shared_ptr holder in gpu_executable.cc.

This allows the holder to be safely captured by value, extending its lifetime appropriately, while still enabling the execution timeout handler in se_gpu_pjrt_client.cc to re-arm the existing guard with a 60-second fallback watchdog before initiating collective cancellation.

The changes build cleanly and appear to follow the intended robustness guidelines. Thanks again for the helpful guidance and direction

xla/service/gpu/gpu_executable.cc
```diff
@@ -691,23 +691,24 @@ absl::Status GpuExecutable::ExecuteThunksImpl(
...
+                    guard_holder]() mutable {
         LOG(ERROR) << absl::StreamFormat("%s failed to finish in %v.",
                                          watchdog_name, watchdog_timeout);
         if (pre_abort) {
           std::move(pre_abort)();
         }
         gpu_run_options->execution_timeout_handler()(watchdog_name,
-                                                     watchdog_timeout);
+                                                     watchdog_timeout,
+                                                     *guard_holder);
```
ezhulenev: you can simply capture guard by reference and re-assign it here, `gpu_run_options->execution_timeout_handler` doesn't need to know about `HangWatchdog`

ezhulenev: if you execute this callback guards is guaranteed to be alive, otherwise timeout callback would not be called, so it's ok to access it via ref

ezhulenev:
I'm not sure extending the guard lifetime beyond the stack scope is safe, because hang watchdog uses it to decide if timeout happened or not, I think you might accidenteally trigger callbacks

Liauuu:
Hi @ezhulenev,

Thanks so much for highlighting the concern around the guard inside the callback — that helped me rethink the design more carefully.

To keep things simple and reliable, I’ve refactored the code so that:

The fallback_guard and all nested watchdog/thread logic are removed.

The on_timeout lambda now just logs/pre‑aborts, calls execution_timeout_handler() to safely report the error to the coordination service, and then immediately invokes std::abort().

Really appreciate your thoughtful guidance.

xla/service/gpu/gpu_executable.cc
```diff
@@ -683,10 +685,29 @@ absl::Status GpuExecutable::ExecuteThunksImpl(
...
+        gpu_run_options->execution_timeout_handler()) {
+      on_timeout = [watchdog_name, watchdog_timeout,
+                    pre_abort = std::move(pre_abort),
+                    gpu_run_options]() mutable {
+        if (pre_abort) {
+          std::move(pre_abort)();
+        }
+        gpu_run_options->execution_timeout_handler()(watchdog_name,
+                                                     watchdog_timeout);
+        std::abort();
```
ezhulenev: you can do this:

```
guard = HangWatchdog::Global().Watch("post-abort ...", absl::Minute(1),
                                         HangWatchdog::Abort("post-abort ..."
, [watchdog_timeout](absl::Minute(1))));
```

If cancellation will work, guard will be destroyed and control flow passed back to jax, if not, process will be killed.,

Liauuu: Thanks again for your guidance — really appreciate it.

penumallas1261 (approved):

Liauuu:
Hi @ezhulenev,

Thank you again for taking the time to review and approve my PR — I really appreciate it.

I noticed that the downstream TensorFlow PR (#121746) has been moved to Draft, so I took a look at it today. It seems to be blocked by a Windows CI linker error (lld-link.exe failed on _pywrap_tensorflow_common.dll), which appears to be an infrastructure or memory issue.

If you happen to have a moment, would you mind pointing me in the right direction on what I should do next?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1912,6 +1915,34 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
...
+          if (absl::Status s =
+                  gpu::AbortCollectivesOnTaskFailure(process_index, error);
+              !s.ok()) {
+            LOG(WARNING) << s;
+          }
+
+          if (distributed_client != nullptr) {
+            absl::StatusOr<CoordinationServiceAgent*> agent =
+                distributed_client->GetCoordinationServiceAgent();
+            if (agent.ok()) {
```
PatriosTheGreat, line 1830: Could you add warning in else branch?

xla/pjrt/gpu/se_gpu_pjrt_client_multi_gpu_test.cc
```diff
@@ -1108,6 +1108,7 @@ absl::StatusOr<PreparedCrossHostTransferTest> PrepareCrossHostTransferTest(
   options.num_nodes = 2;
   options.kv_store =
       GetDistributedKeyValueStore(distributed_client, /*key_prefix=*/"cross:");
+  options.distributed_client = distributed_client;
```
PatriosTheGreat, line 1280: Can you also add a few unit tests here similar to DistributedInit?

You can use some mock in memory coordination_service for this purposes.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1912,6 +1915,34 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
...
+        [process_index = options.node_id,
+         distributed_client = options.distributed_client](
+            absl::string_view action, absl::Duration timeout) {
+          absl::Status error = absl::DeadlineExceededError(absl::StrFormat(
+              "%s failed to finish in %v", action, timeout));
+
+          if (absl::Status s =
+                  gpu::AbortCollectivesOnTaskFailure(process_index, error);
+              !s.ok()) {
+            LOG(WARNING) << s;
```
PatriosTheGreat, line 1824: Shouldn't we stop execution after this error?

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1912,6 +1915,34 @@ absl::StatusOr<std::unique_ptr<PjRtClient>> GetStreamExecutorGpuClient(
...
+          }
+
+          if (distributed_client != nullptr) {
```
PatriosTheGreat, line 1827: Could you add info logging in else branch?

xla/pjrt/distributed/coordination/coordination_service_agent.cc
```diff
@@ -990,4 +990,53 @@ absl::StatusOr<tsl::Env*> CoordinationServiceAgent::GetEnv() {
...
+          "Coordination service agent must be initialized first before "
+          "reporting error."));
+    }
+    if (state_ == xla::coordination::TaskState::ERROR) {
+      return MakeCoordinationError(FailedPrecondition(
+          "Coordination service agent is already in error state."));
+    }
+  }
+  SetError(MakeCoordinationError(error, task_id_, /*is_reported_error=*/true));
+  LOG(INFO) << "Reporting error to coordination service: " << error;
```
PatriosTheGreat: VLOG?

PatriosTheGreat (changes requested):

xla/pjrt/distributed/coordination/coordination_service_agent.cc
Liauuu: Thanks for catching that. I've updated it to VLOG(5) as suggested.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
Liauuu: Thanks! Added the info log as suggested.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
Liauuu: Done. Added the warning log.

xla/pjrt/gpu/se_gpu_pjrt_client_multi_gpu_test.cc
Liauuu: I've added a few unit tests using the in-memory coordination service. Thanks for the suggestion!

xla/pjrt/gpu/se_gpu_pjrt_client.cc
Liauuu: Good point. We don't stop execution immediately here because we first try to cancel the running collectives and let them unwind cleanly via the CancellationToken. If that hangs, the 1-minute watchdog in gpu_executable.cc will hard-abort the process.

xla/service/gpu/gpu_executable.cc
```diff
@@ -659,10 +660,33 @@ absl::Status GpuExecutable::ExecuteThunksImpl(
...
-                            std::move(pre_abort)));
+    const gpu::GpuExecutableRunOptions* gpu_run_options =
+        run_options->run_options().gpu_executable_run_options();
+
+    HangWatchdog::CancelCallback on_timeout;
+    if (gpu_run_options &&
+        gpu_run_options->execution_timeout_handler()) {
+      on_timeout = [watchdog_name, watchdog_timeout,
+                    pre_abort = std::move(pre_abort),
+                    gpu_run_options, &guard]() mutable {
```
terryysun: `guard` is a local variable and will be released when `ExecuteThunksImpl` returns, for async execution that means right after the enqueue. Have you verified that `guard` won't be released too early for the watchdog to trigger?

Liauuu: Thanks for catching this! You're right that the local guard could be released when ExecuteThunksImpl returns in the async path. I moved the watchdog lifetime into ExecutionWatchdogScope, and ExecuteThunksImpl now only arms the watchdog. I'll also add Bazel tests to verify that the timeout callback still fires after ExecuteThunksImpl returns and that mock NCCL is properly unblocked in the deadlock case. Thanks again!

Liauuu:
@terryysun @PatriosTheGreat

Thank you very much for your reviews.
I've addressed the comments you pointed out. Since I wanted to make sure the changes were actually working as intended, I also added and ran the following tests. While writing the tests, I found a few issues and fixed them as well.

**1. execution_watchdog_test**
Verified that ExecutionWatchdogScope is created only when a timeout is configured, that the scope remains alive after the asynchronous task returns so that timeouts can still be detected, and that the timeout handler is invoked correctly.
(I found a shared_ptr reference cycle while testing, changed it to weak_ptr, and the test passed.)
```
Target //xla/service/gpu:execution_watchdog_test up-to-date:
bazel-bin/xla/service/gpu/execution_watchdog_test
INFO: Elapsed time: 9.547s, Critical Path: 8.86s
INFO: 5 processes: 1 internal, 4 processwrapper-sandbox.
INFO: Build completed successfully, 5 total actions
//xla/service/gpu:execution_watchdog_test PASSED in 0.8s
```
**2. abort_collectives_on_task_failure_test**
Verified that the GPU clique is aborted when a task fails, and that the aborted clique is marked as stale so it cannot be reacquired.
(I found that the future still held the lock after Await(), preventing it from being released. I fixed this by adding futures.clear(). I also updated the test so that both ranks reacquire simultaneously according to the rendezvous protocol. )

`Target //xla/backends/gpu/collectives:abort_collectives_on_task_failure_test up-to-date: bazel-bin/xla/backends/gpu/collectives/abort_collectives_on_task_failure_test INFO: Elapsed time: 11.767s, Critical Path: 11.14s INFO: 4 processes: 1 internal, 3 processwrapper-sandbox. INFO: Build completed successfully, 4 total actions //xla/backends/gpu/collectives:abort_collectives_on_task_failure_test PASSED in 2.2s Executed 1 out of 1 test: 1 test passes.`

**3. se_gpu_pjrt_client_multi_gpu_test (ExecutionTimeoutHandlerReportsErrorToCoordinationService)**
Verified that manually invoking execution_timeout_handler on one node reports the error to the coordination service and propagates IsError() to the peer node. 
(I found that connecting the two nodes sequentially caused a deadlock at the barrier, so I changed the test to initialize them in parallel.)

`Target //xla/pjrt/gpu:se_gpu_pjrt_client_multi_gpu_test_nvgpu_any up-to-date: bazel-bin/xla/pjrt/gpu/se_gpu_pjrt_client_multi_gpu_test_nvgpu_any INFO: Build completed successfully, 4 total actions //xla/pjrt/gpu:se_gpu_pjrt_client_multi_gpu_test_nvgpu_any PASSED in 14.8s Executed 1 out of 1 test: 1 test passes.`

Liauuu:
Hi @PatriosTheGreat,

Thanks again for your review. I've addressed the issues I found from the latest CI run (including the missing BUILD dependencies, macro/style updates, and rebasing onto the latest main) and pushed the fixes.

For the remaining failures, they seem to be related to the `@@llvm` Bazel issue, the ROCm/GPU environment, and broader dependency/DWYU checks rather than the changes in this PR, although I could certainly be mistaken.

If you have a chance, I'd really appreciate it if you could rerun the CI. Thank you very much.

Liauuu:
I took a look at the CI failures and found a couple of issues related to my changes (check-contents and clang-tidy), so I've addressed those and rebased the branch. I appreciate the time you took to review.

xla/pjrt/gpu/se_gpu_pjrt_client_multi_gpu_test.cc
```diff
@@ -264,6 +265,177 @@ TEST(StreamExecutorGpuClientTest, DistributedInit) {
...
+
+  for (const absl::Status& status : statuses) {
+    ASSERT_OK(status);
+  }
+
+  auto* gpu_client0 =
+      tsl::down_cast<StreamExecutorGpuClient*>(pjrt_clients[0].get());
+  ExecuteOptions exec_options;
+  const gpu::GpuExecutableRunOptions* run_options =
+      gpu_client0->gpu_run_options(exec_options);
```
PatriosTheGreat: Could you rebase? This method doesn't get execute options anymore

xla/pjrt/gpu/se_gpu_pjrt_client_multi_gpu_test.cc
```diff
@@ -264,6 +265,177 @@ TEST(StreamExecutorGpuClientTest, DistributedInit) {
...
+          xla::CoordinationServiceImpl::Options{num_nodes}));
+
+  std::vector<absl::Status> statuses(num_nodes);
+  {
+    tsl::thread::ThreadPool thread_pool(tsl::Env::Default(),
+                                        "DistributedInitAbortCollectives",
+                                        num_nodes);
+    for (int i = 0; i < num_nodes; ++i) {
+      thread_pool.Schedule([i, num_nodes, kServiceAddress, &statuses]() {
+      DistributedRuntimeClient::Options distributed_options;
```
PatriosTheGreat: Additional tab needed for this block

PatriosTheGreat (changes requested):

xla/pjrt/gpu/se_gpu_pjrt_client_multi_gpu_test.cc
Liauuu: I've rebased onto the latest main and updated the test to use the current gpu_run_options() API.

xla/pjrt/gpu/se_gpu_pjrt_client_multi_gpu_test.cc
Liauuu: Fixed the indentation. I appreciate your continued review.

PatriosTheGreat (approved):

thcmbs:
Thanks for the contribution, we have some jax tests failing, for example:
```
//third_party/py/jax/experimental/jax2tf/tests:jax_primitives_coverage_test_gpu_b200
```
```
JaxPrimitiveTest.test_jax_implemented_dot_general_preferred_lhs_int8_4_3_rhs_int8_3_6_dimensionnumbers_1_0_preferred_int32_enable_xla_True
```
```
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/absl/testing/parameterized.py", [line 327](https://cs.corp.google.com/piper///depot/google3/third_party/py/absl/testing/parameterized.py?l=327&ws=tap-presubmit-server/666202634&snapshot=2), in bound_param_test
    return test_method(self, **testcase_params)
  File "/<embedded stdlib>/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py", [line 80](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py?l=80&ws=tap-presubmit-server/666202634&snapshot=2), in test_jax_implemented
    raise e
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py", [line 71](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py?l=71&ws=tap-presubmit-server/666202634&snapshot=2), in test_jax_implemented
    harness.dyn_fun(*harness.dyn_args_maker(self.rng()))
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/internal_test_util/test_harnesses.py", [line 213](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/internal_test_util/test_harnesses.py?l=213&ws=tap-presubmit-server/666202634&snapshot=2), in dyn_fun
    return self.fun(*all_args)
           ~~~~~~~~^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/lax/lax.py", [line 2542](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/lax/lax.py?l=2542&ws=tap-presubmit-server/666202634&snapshot=2), in dot_general
    return dot(lhs, rhs, dimension_numbers=dimension_numbers, precision=precision,
               preferred_element_type=preferred_element_type, out_sharding=out_sharding)
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/lax/lax.py", [line 2625](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/lax/lax.py?l=2625&ws=tap-presubmit-server/666202634&snapshot=2), in dot
    return dot_general_p.bind(lhs, rhs,
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^
                              dimension_numbers=(cdims, bdims),
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                              precision=canonicalize_precision(precision),
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                              preferred_element_type=preferred_element_type,
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                              out_sharding=out_sharding)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/core.py", [line 720](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/core.py?l=720&ws=tap-presubmit-server/666202634&snapshot=2), in bind
    return self.bind_with_trace(prev_trace, args, avals, params)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/core.py", [line 728](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/core.py?l=728&ws=tap-presubmit-server/666202634&snapshot=2), in bind_with_trace
    return trace.process_primitive(self, args, params)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/core.py", [line 1308](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/core.py?l=1308&ws=tap-presubmit-server/666202634&snapshot=2), in process_primitive
    return primitive.impl(*args, **params)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/dispatch.py", [line 88](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/dispatch.py?l=88&ws=tap-presubmit-server/666202634&snapshot=2), in apply_primitive
    outs = fun(*args)
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/traceback_util.py", [line 195](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/traceback_util.py?l=195&ws=tap-presubmit-server/666202634&snapshot=2), in reraise_with_filtered_traceback
    return fun(*args, **kwargs)  # pyrefly: ignore[not-callable]
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/pjit.py", [line 264](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/pjit.py?l=264&ws=tap-presubmit-server/666202634&snapshot=2), in cache_miss
    executable, pgle_profiler, const_args) = _run_python_pjit(
                                             ~~~~~~~~~~~~~~~~^
        p, args_flat, fun, args, kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/pjit.py", [line 140](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/pjit.py?l=140&ws=tap-presubmit-server/666202634&snapshot=2), in _run_python_pjit
    out_flat, compiled, profiler, const_args = _pjit_call_impl_python(
                                               ~~~~~~~~~~~~~~~~~~~~~~^
        *args_flat, **p.params)
        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/pjit.py", [line 1148](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/pjit.py?l=1148&ws=tap-presubmit-server/666202634&snapshot=2), in _pjit_call_impl_python
    compiled = computation.compile()
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/interpreters/pxla.py", [line 1257](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/interpreters/pxla.py?l=1257&ws=tap-presubmit-server/666202634&snapshot=2), in compile
    executable = UnloadedMeshExecutable.from_hlo(
        self._name, self._hlo, **self.compile_args,
        compiler_options_kvs=compiler_options_kvs,
        device_list=compilation_device_list)
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/interpreters/pxla.py", [line 1754](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/interpreters/pxla.py?l=1754&ws=tap-presubmit-server/666202634&snapshot=2), in from_hlo
    xla_executable = _cached_compilation(
        hlo, name, tuple_args, allow_prop_to_inputs, allow_prop_to_outputs,
        tuple(host_callbacks), backend, device_list, compiler_options_kvs,
        pgle_profiler)
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/interpreters/pxla.py", [line 1534](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/interpreters/pxla.py?l=1534&ws=tap-presubmit-server/666202634&snapshot=2), in _cached_compilation
    xla_executable = compiler.compile_or_get_cached(
        backend, computation, dev, compile_options, host_callbacks,
        da, pgle_profiler)
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 505](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=505&ws=tap-presubmit-server/666202634&snapshot=2), in compile_or_get_cached
    return _compile_and_write_cache(
        backend,
    ...<5 lines>...
        cache_key,
    )
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 770](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=770&ws=tap-presubmit-server/666202634&snapshot=2), in _compile_and_write_cache
    executable = backend_compile_and_load(
        backend, computation, executable_devices, compile_options, host_callbacks
    )
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/profiler.py", [line 420](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/profiler.py?l=420&ws=tap-presubmit-server/666202634&snapshot=2), in wrapper
    return func(*args, **kwargs)
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 401](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=401&ws=tap-presubmit-server/666202634&snapshot=2), in backend_compile_and_load
    raise e
  File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 391](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=391&ws=tap-presubmit-server/666202634&snapshot=2), in backend_compile_and_load
    return backend.compile_and_load(
           ~~~~~~~~~~~~~~~~~~~~~~~~^
        module,
        ^^^^^^^
        executable_devices=executable_devices,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        compile_options=options,
        ^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
jax.errors.JaxRuntimeError: INTERNAL: RET_CHECK failure ([third_party/tensorflow/compiler/xla/backends/autotuner/config_assigner.cc:412](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/backends/autotuner/config_assigner.cc?l=412&ws=tap-presubmit-server/666202634&snapshot=2)) !candidates.empty() Autotuning failed for HLO: %gemm_fusion_dot_general.1 = s32[4,6]{1,0} fusion(%args_0_.1, %bitcast.7.0), kind=kCustom, calls=%gemm_fusion_dot_general.1_computation, metadata={op_name="jit(dot_general)/dot_general" stack_frame_id=12}, backend_config={"operation_queue_id":"0","fusion_backend_config":{"kind":"__triton_gemm"},"force_earliest_schedule":false,"reification_cost":[],"device_type":"DEVICE_TYPE_INVALID"}. No configs could be compiled.
```


PTAL

Liauuu:
> Thanks for the contribution, we have some jax tests failing, for example:
> 
> ```
> //third_party/py/jax/experimental/jax2tf/tests:jax_primitives_coverage_test_gpu_b200
> ```
> 
> ```
> JaxPrimitiveTest.test_jax_implemented_dot_general_preferred_lhs_int8_4_3_rhs_int8_3_6_dimensionnumbers_1_0_preferred_int32_enable_xla_True
> ```
> 
> ```
> Traceback (most recent call last):
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/absl/testing/parameterized.py", [line 327](https://cs.corp.google.com/piper///depot/google3/third_party/py/absl/testing/parameterized.py?l=327&ws=tap-presubmit-server/666202634&snapshot=2), in bound_param_test
>     return test_method(self, **testcase_params)
>   File "/<embedded stdlib>/contextlib.py", line 85, in inner
>     return func(*args, **kwds)
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py", [line 80](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py?l=80&ws=tap-presubmit-server/666202634&snapshot=2), in test_jax_implemented
>     raise e
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py", [line 71](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/experimental/jax2tf/tests/jax_primitives_coverage_test.py?l=71&ws=tap-presubmit-server/666202634&snapshot=2), in test_jax_implemented
>     harness.dyn_fun(*harness.dyn_args_maker(self.rng()))
>     ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/internal_test_util/test_harnesses.py", [line 213](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/internal_test_util/test_harnesses.py?l=213&ws=tap-presubmit-server/666202634&snapshot=2), in dyn_fun
>     return self.fun(*all_args)
>            ~~~~~~~~^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/lax/lax.py", [line 2542](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/lax/lax.py?l=2542&ws=tap-presubmit-server/666202634&snapshot=2), in dot_general
>     return dot(lhs, rhs, dimension_numbers=dimension_numbers, precision=precision,
>                preferred_element_type=preferred_element_type, out_sharding=out_sharding)
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/lax/lax.py", [line 2625](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/lax/lax.py?l=2625&ws=tap-presubmit-server/666202634&snapshot=2), in dot
>     return dot_general_p.bind(lhs, rhs,
>            ~~~~~~~~~~~~~~~~~~^^^^^^^^^^
>                               dimension_numbers=(cdims, bdims),
>                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>                               precision=canonicalize_precision(precision),
>                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>                               preferred_element_type=preferred_element_type,
>                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>                               out_sharding=out_sharding)
>                               ^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/core.py", [line 720](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/core.py?l=720&ws=tap-presubmit-server/666202634&snapshot=2), in bind
>     return self.bind_with_trace(prev_trace, args, avals, params)
>            ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/core.py", [line 728](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/core.py?l=728&ws=tap-presubmit-server/666202634&snapshot=2), in bind_with_trace
>     return trace.process_primitive(self, args, params)
>            ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/core.py", [line 1308](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/core.py?l=1308&ws=tap-presubmit-server/666202634&snapshot=2), in process_primitive
>     return primitive.impl(*args, **params)
>            ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/dispatch.py", [line 88](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/dispatch.py?l=88&ws=tap-presubmit-server/666202634&snapshot=2), in apply_primitive
>     outs = fun(*args)
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/traceback_util.py", [line 195](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/traceback_util.py?l=195&ws=tap-presubmit-server/666202634&snapshot=2), in reraise_with_filtered_traceback
>     return fun(*args, **kwargs)  # pyrefly: ignore[not-callable]
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/pjit.py", [line 264](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/pjit.py?l=264&ws=tap-presubmit-server/666202634&snapshot=2), in cache_miss
>     executable, pgle_profiler, const_args) = _run_python_pjit(
>                                              ~~~~~~~~~~~~~~~~^
>         p, args_flat, fun, args, kwargs)
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/pjit.py", [line 140](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/pjit.py?l=140&ws=tap-presubmit-server/666202634&snapshot=2), in _run_python_pjit
>     out_flat, compiled, profiler, const_args = _pjit_call_impl_python(
>                                                ~~~~~~~~~~~~~~~~~~~~~~^
>         *args_flat, **p.params)
>         ^^^^^^^^^^^^^^^^^^^^^^^
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/pjit.py", [line 1148](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/pjit.py?l=1148&ws=tap-presubmit-server/666202634&snapshot=2), in _pjit_call_impl_python
>     compiled = computation.compile()
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/interpreters/pxla.py", [line 1257](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/interpreters/pxla.py?l=1257&ws=tap-presubmit-server/666202634&snapshot=2), in compile
>     executable = UnloadedMeshExecutable.from_hlo(
>         self._name, self._hlo, **self.compile_args,
>         compiler_options_kvs=compiler_options_kvs,
>         device_list=compilation_device_list)
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/interpreters/pxla.py", [line 1754](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/interpreters/pxla.py?l=1754&ws=tap-presubmit-server/666202634&snapshot=2), in from_hlo
>     xla_executable = _cached_compilation(
>         hlo, name, tuple_args, allow_prop_to_inputs, allow_prop_to_outputs,
>         tuple(host_callbacks), backend, device_list, compiler_options_kvs,
>         pgle_profiler)
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/interpreters/pxla.py", [line 1534](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/interpreters/pxla.py?l=1534&ws=tap-presubmit-server/666202634&snapshot=2), in _cached_compilation
>     xla_executable = compiler.compile_or_get_cached(
>         backend, computation, dev, compile_options, host_callbacks,
>         da, pgle_profiler)
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 505](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=505&ws=tap-presubmit-server/666202634&snapshot=2), in compile_or_get_cached
>     return _compile_and_write_cache(
>         backend,
>     ...<5 lines>...
>         cache_key,
>     )
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 770](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=770&ws=tap-presubmit-server/666202634&snapshot=2), in _compile_and_write_cache
>     executable = backend_compile_and_load(
>         backend, computation, executable_devices, compile_options, host_callbacks
>     )
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/profiler.py", [line 420](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/profiler.py?l=420&ws=tap-presubmit-server/666202634&snapshot=2), in wrapper
>     return func(*args, **kwargs)
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 401](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=401&ws=tap-presubmit-server/666202634&snapshot=2), in backend_compile_and_load
>     raise e
>   File "/build/work/f881e3af2ec8ab53b368433b40eb05454f19/google3/runfiles/google3/third_party/py/jax/_src/compiler.py", [line 391](https://cs.corp.google.com/piper///depot/google3/third_party/py/jax/_src/compiler.py?l=391&ws=tap-presubmit-server/666202634&snapshot=2), in backend_compile_and_load
>     return backend.compile_and_load(
>            ~~~~~~~~~~~~~~~~~~~~~~~~^
>         module,
>         ^^^^^^^
>         executable_devices=executable_devices,
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>         compile_options=options,
>         ^^^^^^^^^^^^^^^^^^^^^^^^
>     )
>     ^
> jax.errors.JaxRuntimeError: INTERNAL: RET_CHECK failure ([third_party/tensorflow/compiler/xla/backends/autotuner/config_assigner.cc:412](https://cs.corp.google.com/piper///depot/google3/third_party/tensorflow/compiler/xla/backends/autotuner/config_assigner.cc?l=412&ws=tap-presubmit-server/666202634&snapshot=2)) !candidates.empty() Autotuning failed for HLO: %gemm_fusion_dot_general.1 = s32[4,6]{1,0} fusion(%args_0_.1, %bitcast.7.0), kind=kCustom, calls=%gemm_fusion_dot_general.1_computation, metadata={op_name="jit(dot_general)/dot_general" stack_frame_id=12}, backend_config={"operation_queue_id":"0","fusion_backend_config":{"kind":"__triton_gemm"},"force_earliest_schedule":false,"reification_cost":[],"device_type":"DEVICE_TYPE_INVALID"}. No configs could be compiled.
> ```
> 
> PTAL

Hi @thcmbs,

Thanks for pointing this out, and I took a closer look at the traceback.

From what I can tell, the failure occurs during XLA compilation, specifically in the Triton GEMM autotuning path. The `dot_general(..., preferred_element_type=int32)` is lowered to a `__triton_gemm` fusion, and the failure comes from `config_assigner.cc` with:

> `!candidates.empty()`
> `Autotuning failed ... No configs could be compiled.`

Based on the stack trace, it looks like none of the candidate autotuning configurations could be compiled for this HLO on B200, causing compilation to fail before execution begins. Unfortunately, the traceback doesn't include the per-configuration compilation errors, so I can't tell why the Triton compilation itself failed.

My understanding is that this is outside the code path modified by this PR. The changes here are limited to the runtime HangWatchdog / CoordinationService timeout handling, whereas this failure happens earlier during the Triton autotuner / code generation phase. Also, the HangWatchdog is disabled by default unless the corresponding timeout flags are explicitly enabled, although I haven't verified the exact flags used for this particular test run.

That said, I may be missing something. I'm happy to investigate further if you think there's a connection that I've overlooked.

Issue #44228, opened 2026-06-12
[Regression] `CollectivePipeliner` GPU pass present in JAX 0.4.38 is missing from JAX 0.10.1 — pipelining flags accepted but produce no pass stages; class symbols absent from binaries


rsdcastro:
## Summary
The XLA collective pipeliner pass (`xla/service/collective_pipeliner.cc`) was present and runnable in the GPU plugin shipped with `jax==0.4.38` (confirmed via `strings`/`nm` of `jax_plugins/xla_cuda12/xla_cuda_plugin.so` — see "Cross-check on older JAX" below). In `jax==0.10.1` / `jaxlib==0.10.1` (current stable as of June 2026), the pass is no longer invoked on GPU regardless of which `--xla_gpu_enable_pipelined_*` flag is set:

- `--xla_gpu_enable_pipelined_all_reduce=true` (and the `_all_gather` / `_reduce_scatter` variants) are accepted by the flag parser without error
- HLO pass dumps (`XLA_FLAGS=--xla_dump_hlo_pass_re=.*`) contain **no `collective-pipeliner-forward` or `-backward` stages** in any compiled module
- `nm -D` / `strings` on every `.so` shipped in `site-packages/jaxlib/` and `site-packages/jax_plugins/` shows **no `CollectivePipeliner` symbols** — the class is not linked into the binary
- The umbrella flag `--xla_gpu_enable_pipelined_collectives=true` is rejected as "Unknown flag" by the parser
- `--xla_gpu_run_post_layout_collective_pipeliner=true` is also rejected as "Unknown flag" (though it appears as a proto field name string in `libjax_common.so`)

The pass is still referenced in XLA source (`xla/service/collective_pipeliner.cc`), so it exists upstream — but appears not built into the GPU plugin shipped with current JAX stable. The "Cross-check on older JAX" section below shows the same wheels installed via the same `pip install "jax[cuda12]==<version>"` command have the symbols in 0.4.38 but not in 0.10.1, narrowing the regression to a build-config change between those two releases.

## Environment

| Component | Version |
|---|---|
| OS | CentOS Stream 9, Linux 6.13.2 x86_64 |
| GPU | NVIDIA H100, driver 580.82.07 |
| Python | 3.11.15 |
| `jax` | 0.10.1 |
| `jaxlib` | 0.10.1 |
| `jax-cuda12-plugin` | 0.10.1 |
| `jax-cuda12-pjrt` | 0.10.1 |
[...]

PR #44459, opened 2026-06-17
Fix link-time omission of collective_pipeliner in GPU binaries

Liauuu:
## Problem

This PR addresses issue #44228, where the `CollectivePipeliner` symbols and its corresponding GPU pass appear to be missing from the final compiled JAX GPU binaries (such as `libjax_common.so` or `pjrt_c_api_gpu_plugin.so`) in recent releases.

As a result, the `--xla_gpu_enable_pipelined_*` flags are accepted by the parser but appear to have no observable effect on the GPU HLO compilation pipeline.

## Changes

* Added `alwayslink = True` to the `//xla/service:collective_pipeliner` target in `xla/service/BUILD`.
* Added `# buildcleaner: keep` to the production compiler dependency in `xla/service/gpu/BUILD` to prevent automatic removal of the link dependency.

## Rationale

Before applying this change, we investigated whether the absence of the pass was intentional:

1. No conditional compilation guards (`#ifdef` / `#if`) were found around the `CollectivePipeliner` integration points in `gpu_compiler.cc`.
2. The pass remains exposed through user-facing runtime flags and optimization effort levels (see `docs/effort_levels.md`).
3. Existing GPU optimization tests continue to reference workflows involving both `LatencyHidingScheduler` and `CollectivePipeliner`.

Based on this investigation, the implementation appears to remain active and supported within the codebase.

Because the pass does not appear to be retained through direct symbol references from the final shared-library link step, the linker may discard the corresponding archive members during shared-library construction. Adding `alwayslink = True` ensures that the implementation is retained in the final link closure and remains available to the GPU compilation pipeline.

## Testing

* Verified that the build graph remains valid after introducing `alwayslink`.
* Verified that the change is limited to retention of the `collective_pipeliner` target and does not introduce new dependency edges.
* Full GPU runtime validation is left to upstream CI due to local hardware limitations.


golechwierowicz (approved):

allanrenucci (changes requested):
It seems this is addressing a symptom of the problem and not the root cause.

Please figure out why the linker discards `CollectivePipeliner` instead.

Liauuu:
@allanrenucci

Thanks a lot for the feedback — I really appreciate the push to look deeper.

After tracing the registration and build paths, my current understanding is that `CollectivePipeliner` appears more isolated in the production linkage compared to other passes.

It is primarily referenced through `AddPass<CollectivePipeliner>(...)`, while other passes often appear to have additional reference paths or consumers in other translation units.

Because of this, my current hypothesis is that the implementation object (`collective_pipeliner.cc`) may not always be strongly pulled into the final link unit, depending on how transitive dependencies are resolved during static linking/LTO.

The factory change was an attempt to introduce an explicit reference from `gpu_compiler.cc` to the implementation, so the linker has a concrete symbol anchor.

Thanks again!

allanrenucci:
I am going to close this PR. I strongly suspect it is AI generated and our team does not have the bandwidth to review low quality AI generated PRs.

PR #44560, opened 2026-06-19
[ROCm][Triton] All gather thunk dispatch (PR2/4)

mfrancepillois:
📝 Summary of Changes
This PR is the second in a series of four.

PR 1/4: https://github.com/openxla/xla/pull/44559   
PR 2/4: https://github.com/openxla/xla/pull/44560   <--- This PR
PR 3/4: https://github.com/openxla/xla/pull/44562
PR 4/4: https://github.com/openxla/xla/pull/44564

These PRs aim to enable the Triton backend for the All-Gather collective operation based on the experimental tiling strategy.

The PR wires AllGather through CollectiveKernelThunk/AllGatherThunk at runtime, and adds a dedicated flag.

🎯 Justification
Lowering All-Gather through the triton backend significantly improves performance of this operation for AMD/ROCm target (compared to RCCL backend), especially for small tensor, and could allow us to take greater advantage of kernel-level fusion.

Example of 5 Triton Performance Improvements — All-Gather 1D on MI355
 | DType | Shape | Replicas | Triton (µs) | RCCL (µs) | Speedup |
|--------|--------|----------|------------------:|---------------:|--------:|
| s8 | [16384] | 2 | 17.428 | 42.342 | **2.43×** |
| bf16 | [16384] | 2 | 17.708 | 41.733 | **2.36×** |
| f32 | [131072] | 2 | 27.924 | 52.497 | **1.88×** |
| s32 | [4096] | 2 | 24.435 | 44.287 | **1.81×** |
| s8 | [262144] | 4 | 40.459 | 71.119 | **1.76×** |

Example of 5 Triton Performance Improvements — All-Gather 2D on MI355
dtype | shape | replicas | RCCL (µs) | Triton (µs) | speedup
-- | -- | -- | -- | -- | --
bf16 | [32, 32] | 2 | 41.153 | 16.744 | **2.46×**
bf16 | [128, 64] | 4 | 68.079 | 35.498 | **1.92×**
bf16 | [256, 256] | 4 | 68.927 | 36.992 | **1.86×**
f32 | [128, 64] | 4 | 64.403 | 34.928 | **1.84×**
f8e4m3fn | [512, 512] | 4 | 68.342 | 41.506 | **1.65×**

🚀 Kind of Contribution
Please remove what does not apply:⚡️ Performance Improvement, ✨ New Feature

PR #44559, opened 2026-06-19
[ROCm][Triton] AllGather build info (PR1/4)

mfrancepillois:
📝 Summary of Changes
This PR is the first in a series of four.

PR 1/4: https://github.com/openxla/xla/pull/44559   <--- This PR
PR 2/4: https://github.com/openxla/xla/pull/44560
PR 3/4: https://github.com/openxla/xla/pull/44562
PR 4/4: https://github.com/openxla/xla/pull/44564

These PRs aim to enable the Triton backend for the All-Gather collective operation based on the experimental tiling strategy.

This PR adds a dedicated all_gather.cc/.h module with the metadata and tiling logic needed to decide whether and how to run an AllGather via the Triton backend (similar to all_reduce.cc/h).

🎯 Justification
Lowering All-Gather through the triton backend significantly improves performance of this operation for AMD/ROCm target (compared to RCCL backend), especially for small tensor, and could allow us to take greater advantage of kernel-level fusion.

Example of 5 Triton Performance Improvements — All-Gather 1D on MI355
 | DType | Shape | Replicas | Triton (µs) | RCCL (µs) | Speedup |
|--------|--------|----------|------------------:|---------------:|--------:|
| s8 | [16384] | 2 | 17.428 | 42.342 | **2.43×** |
| bf16 | [16384] | 2 | 17.708 | 41.733 | **2.36×** |
| f32 | [131072] | 2 | 27.924 | 52.497 | **1.88×** |
| s32 | [4096] | 2 | 24.435 | 44.287 | **1.81×** |
| s8 | [262144] | 4 | 40.459 | 71.119 | **1.76×** |

Example of 5 Triton Performance Improvements — All-Gather 2D on MI355
dtype | shape | replicas | RCCL (µs) | Triton (µs) | speedup
-- | -- | -- | -- | -- | --
bf16 | [32, 32] | 2 | 41.153 | 16.744 | **2.46×**
bf16 | [128, 64] | 4 | 68.079 | 35.498 | **1.92×**
bf16 | [256, 256] | 4 | 68.927 | 36.992 | **1.86×**
f32 | [128, 64] | 4 | 64.403 | 34.928 | **1.84×**
f8e4m3fn | [512, 512] | 4 | 68.342 | 41.506 | **1.65×**

🚀 Kind of Contribution
Please remove what does not apply: ⚡️ Performance Improvement,✨ New Feature

🧪 Execution Tests:
Tests included in this PR.

PR #44562, opened 2026-06-19
[ROCm][Triton] All gather triton emitter (PR3/4)

mfrancepillois:
📝 Summary of Changes
This PR is the third in a series of four.

PR 1/4: https://github.com/openxla/xla/pull/44559   
PR 2/4: https://github.com/openxla/xla/pull/44560   
PR 3/4: https://github.com/openxla/xla/pull/44562  <--- This PR
PR 4/4: https://github.com/openxla/xla/pull/44564

These PRs aim to enable the Triton backend for the All-Gather collective operation based on the experimental tiling strategy.

The PR adds AllGatherEmitter class and all Triton MLIR emission logic. This activates the full emitter path for AllGather in the collective backend.

🎯 Justification
Lowering All-Gather through the triton backend significantly improves performance of this operation for AMD/ROCm target (compared to RCCL backend), especially for small tensor, and could allow us to take greater advantage of kernel-level fusion.

Example of 5 Triton Performance Improvements — All-Gather 1D on MI355
 | DType | Shape | Replicas | Triton (µs) | RCCL (µs) | Speedup |
|--------|--------|----------|------------------:|---------------:|--------:|
| s8 | [16384] | 2 | 17.428 | 42.342 | **2.43×** |
| bf16 | [16384] | 2 | 17.708 | 41.733 | **2.36×** |
| f32 | [131072] | 2 | 27.924 | 52.497 | **1.88×** |
| s32 | [4096] | 2 | 24.435 | 44.287 | **1.81×** |
| s8 | [262144] | 4 | 40.459 | 71.119 | **1.76×** |

Example of 5 Triton Performance Improvements — All-Gather 2D on MI355
dtype | shape | replicas | RCCL (µs) | Triton (µs) | speedup
-- | -- | -- | -- | -- | --
bf16 | [32, 32] | 2 | 41.153 | 16.744 | **2.46×**
bf16 | [128, 64] | 4 | 68.079 | 35.498 | **1.92×**
bf16 | [256, 256] | 4 | 68.927 | 36.992 | **1.86×**
f32 | [128, 64] | 4 | 64.403 | 34.928 | **1.84×**
f8e4m3fn | [512, 512] | 4 | 68.342 | 41.506 | **1.65×**

🚀 Kind of Contribution
Please remove what does not apply:⚡️ Performance Improvement, ✨ New Feature

### DIFF
--- a/xla/codegen/tiling/experimental/tile_propagation.cc
+++ b/xla/codegen/tiling/experimental/tile_propagation.cc
@@ -1251,11 +1251,9 @@ Tiles PropagateTileToInputForAllGatherOp(const TilingSpace& tiling_space,
       << "Multi-operand AllGather is not yet supported.";
   const Shape& input_shape = hlo.operand(0)->shape();
   int64_t local_size = input_shape.dimensions(gather_dim);
-  int64_t num_replicas = hlo.shape().dimensions(gather_dim) / local_size;
 
   const DimTile& output_dim_tile = output_tile.dim_tiles()[gather_dim];
 
-  SymbolicExpr replica_id = output_dim_tile.offset / local_size;
   SymbolicExpr input_offset = output_dim_tile.offset % local_size;
 
   llvm::SmallVector<DimTile> input_dim_tiles;
@@ -1271,16 +1269,15 @@ Tiles PropagateTileToInputForAllGatherOp(const TilingSpace& tiling_space,
     }
   }
 
-  mlir::MLIRContext* ctx = output_tile.mlir_context();
+  // The input tile does NOT add a new replica_id dimension beyond what the
+  // output tile already carries. The AllGather emitter derives source_rank
+  // from the XTile tile_id function argument:
+  //   tiles_per_rank = local_size_in_gather_dim / tile_size_in_gather_dim
+  //   source_rank    = tile_id / tiles_per_rank
+  // This is computed directly in AllGatherEmitter::EmitAllGather() without
+  // needing a symbolic replica_id expression in the input tile.
   llvm::SmallVector<DimTile> replica_id_dim_tiles =
       llvm::to_vector(output_tile.replica_ids());
-  replica_id_dim_tiles.push_back(DimTile{
-      /*offset=*/replica_id,
-      /*size=*/CreateSymbolicConstant(1, ctx),
-      /*stride=*/CreateSymbolicConstant(1, ctx),
-      /*upper_bound=*/
-      CreateSymbolicConstant(num_replicas, ctx),
-  });
 
   Tile input_tile(output_tile.tiling_space(), std::move(input_dim_tiles),
                   std::move(replica_id_dim_tiles));
--- a/xla/backends/gpu/codegen/triton/BUILD
+++ b/xla/backends/gpu/codegen/triton/BUILD
@@ -729,6 +729,7 @@ cc_library(
         "//xla:util",
         "//xla:xla_data_proto_cc",
         "//xla/backends/gpu/codegen/triton/ir:triton_xla",
+        "//xla/backends/gpu/runtime:all_gather",
         "//xla/backends/gpu/runtime:all_reduce",
         "//xla/backends/gpu/runtime:collective_params",
         "//xla/backends/gpu/transforms/collectives:collective_ops_utils",
--- a/xla/backends/gpu/codegen/triton/collective_emitter.cc
+++ b/xla/backends/gpu/codegen/triton/collective_emitter.cc
@@ -58,6 +58,7 @@ limitations under the License.
 #include "stablehlo/dialect/StablehloOps.h"
 #include "xla/backends/gpu/codegen/triton/ir/triton_xla_ops.h"
 #include "xla/backends/gpu/codegen/triton/lowering_util.h"
+#include "xla/backends/gpu/runtime/all_gather.h"
 #include "xla/backends/gpu/runtime/all_reduce.h"
 #include "xla/backends/gpu/runtime/collective_params.h"
 #include "xla/backends/gpu/transforms/collectives/collective_ops_utils.h"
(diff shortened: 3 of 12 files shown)


xla/codegen/tiling/experimental/tile_propagation.cc
```diff
@@ -1271,16 +1269,15 @@ Tiles PropagateTileToInputForAllGatherOp(const TilingSpace& tiling_space,
...
   }
 
-  mlir::MLIRContext* ctx = output_tile.mlir_context();
+  // The input tile does NOT add a new replica_id dimension beyond what the
+  // output tile already carries. The AllGather emitter derives source_rank
+  // from the XTile tile_id function argument:
+  //   tiles_per_rank = local_size_in_gather_dim / tile_size_in_gather_dim
+  //   source_rank    = tile_id / tiles_per_rank
+  // This is computed directly in AllGatherEmitter::EmitAllGather() without
+  // needing a symbolic replica_id expression in the input tile.
```
sohaibiftikhar, line 1278: We do need replica ids when we want to propagate inputs to producers. Can't just remove replica_ids here.

I need to think on how to support scratch buffer based all-gather along with simply having a zero-copy version. But this sort of breaks the ability to form collective dots.

sohaibiftikhar (changes requested):

xla/codegen/tiling/experimental/tile_propagation.cc
mfrancepillois: Thanks for this feedback. The problem is that when we pass `replica-id` the first arg is no longer a plain buffer, but is "automatically" wrapped into a pointer table. IMO, this behavior is not correct and the wrapping should be conditioned by the arg type. I've created a PR to propose this change: https://github.com/openxla/xla/pull/46352 
If you can take a look and tell me what you think it would be much appreciate. Thanks

sohaibiftikhar: Sorry for the delay in response. You can ping me and we can discuss this offline but we need to make this decision at compile time not at emission. What you did there in that CL is too brittle unfortunately. I have a pending change that adds collective fusions and probably this decision needs to go there.

In addition how hard would it be to just add support for collective memory to the collective kernel thunk (most of support is already there) and we can insert a barrier before and after and would have all-gather support without even needing scratch buffers. Then we can add a scratch buffer version as a followup. There is benefit to using scratch buffers because we get one less sync point but it shouldn't come at the cost of not being able to fuse otherwise. So starting with an all-gather kernel that doesn't depend on scratch buffers is a good start imho.

mfrancepillois: Thanks for the feedback. I created a new PR with All-Gather backend implementation relying on collective memory: https://github.com/openxla/xla/pull/46707
This would replace PR3 and PR4 of this series of PRs

dimitar-asenov:
@mfrancepillois Can we close this PR, it seems you have sent a replacement already?

mfrancepillois:
> @mfrancepillois Can we close this PR, it seems you have sent a replacement already?

TBH, I'm not sure how we are going to proceed with the AG triton backend. It seems that @sohaibiftikhar has some work that he wanted to land that could have an impact the AG implementation strategy. The other PR implements another way to handle AG using collective memory, so it is not totally equivalent. In fact, I was waiting to see exactly what @sohaibiftikhar was going to land and before making a decision on the next step to this PR (and the other).

sohaibiftikhar:
@dimitar-asenov @mfrancepillois 
Sorry for the delay. There has been a lot of back and forth on this PR. To recap we had a chain of 4 PRs to start with. We landed 2 of them now. In the background I also landed some related changes to the thunk and for collective fusion pass that is supposed to be used here.

In general we want to support 2 things with all-gather codegen.
1. one-shot all-gather. This is what this change is working towards. Perhaps we could also fuse in element-wise ops if needed into it.
2. two-shot all-gather using symmetric memory. This is what the current tiling support is at. The goal is to be able to fuse in dots into it.
3. This also lays the foundation for reduce-scatter fusions so we do dot + reduce-scatter.

The current change still needs work. I would like to do this in a principled way so we don't have to redo this when we try to fuse into it. For eg: We cannot simply undo the tiling work that exists. For that we need to make the decision on what type of all-gather we want to emit during collective fusion pass. Maxime also had some alternative approaches (in the other PRs linked here) to this but none of them work just out of the box sadly. For that purpose I am also working towards landing this PR in the background which is not always evident from the state of the PR itself.

There has been some more hiccups as well.
1. Last week the fusion pass had to be rolled back due to failures in some jax tests. I am working towards fixing those and submitting this again. Then the pathway for this change would be green again. I am on the side also towards all tiling infra necessary for the emitter changes in this chain. At the time of writing even one-shot `all-reduce` is not active anymore.
2. We have also had internally a lot of instability in landing symmetric memory infra changes for multi host setups which has also eaten up time in trying to build infrastructure for the thunk. This is somewhat unrelated to this change itself though since we don't strictly need it for single host one-shot. But I would like to have support for multi-host one-shot as well for which we do need that infra.

All in all this is in plan but it will still take some time. If it is not super urgent to land it I would ask to wait a bit on this so we can land it in a fashion where it doesn't need rework as soon as it lands.

sohaibiftikhar:
Thanks @mfrancepillois for the effort here.

Initial one-shot AG support was landed in https://github.com/openxla/xla/pull/48277 while I was out. There are some gaps that I will now fill out but this change list can be closed now.

In the end the approach that was taken was similar to what you had done in the alternative change with the memcpy. Hiding the memcpy behind graphs means that there is no additional overhead and we saw benefits up until 512KB for AG over NCCL at least.

Support is still behind a flag though and I will be testing it out before making it the default.

mfrancepillois:
> Thanks @mfrancepillois for the effort here.
> 
> Initial one-shot AG support was landed in #48277 while I was out. There are some gaps that I will now fill out but this change list can be closed now.
> 
> In the end the approach that was taken was similar to what you had done in the alternative change with the memcpy. Hiding the memcpy behind graphs means that there is no additional overhead and we saw benefits up until 512KB for AG over NCCL at least.
> 
> Support is still behind a flag though and I will be testing it out before making it the default.

Thanks for the update.
I think the merged AG support isn’t working for the ROCm target, as ROCm doesn’t currently support symmetric memory, but this should be fixed soon-ish.
I therefore close this PR.

PR #44653, opened 2026-06-23
Validate buffer slices during deserialization

M0nd0R:
## Summary

- validate `BufferAllocation::Slice::FromProto` rejects negative offsets and sizes
- reject serialized slices whose `[offset, offset + size)` range exceeds the referenced allocation
- add regression coverage for malformed `BufferAllocationSliceProto` values

## Why

`BufferAllocation::Slice::FromProto` reconstructs runtime buffer slices from serialized executable data. It previously validated only `buffer_allocation_index`, while the compile-time assignment path guarantees that `offset >= 0`, `size >= 0`, and `offset + size <= allocation.size()`. This change enforces that same invariant before constructing the slice.

## Testing

- `git diff --check`
- `clang++ -std=c++17 -fsanitize=address -g -DNDEBUG -DAPPLY_FIX poc_primitive_asan.cc -o /tmp/openxla_poc_fixed && /tmp/openxla_poc_fixed` from the local reproduction bundle; the fixed logic rejects the malformed slice
- attempted `bazel test //xla/service:buffer_assignment_test`; local sparse-checkout setup initially missed top-level `third_party` and `tools`, and after adding them the invocation stayed in Bazel repository mapping until interrupted. Retrying with `--noexperimental_repo_remote_exec` fails during repository initialization because `rules_ml_toolchain` requires that experimental flag.

### DIFF
--- a/xla/backends/gpu/runtime/dynamic_slice_thunk_test.cc
+++ b/xla/backends/gpu/runtime/dynamic_slice_thunk_test.cc
@@ -104,10 +104,40 @@ se::StreamExecutor* GpuExecutor() {
 }
 void CheckProtoRoundTrip(const DynamicSliceThunk& thunk,
                          const DynamicSliceThunkProto& proto) {
+  // Size allocations to cover every slice referenced by the thunk. Hardcoding
+  // 1024 used to work when FromProto only checked the allocation index; with
+  // offset/size bounds checks it must also fit workspace slices (e.g. 1MiB).
+  std::vector<int64_t> allocation_sizes(10, 1024);
+  auto ensure_fits = [&](const BufferAllocation::Slice& slice) {
+    const int index = slice.index();
+    if (index >= static_cast<int>(allocation_sizes.size())) {
+      allocation_sizes.resize(index + 1, 1024);
+    }
+    allocation_sizes[index] =
+        std::max(allocation_sizes[index], slice.offset() + slice.size());
+  };
+  for (const std::optional<BufferAllocation::Slice>& arg :
+       thunk.get_arguments()) {
+    if (arg.has_value()) {
+      ensure_fits(*arg);
+    }
+  }
+  for (const std::optional<std::vector<DynamicSliceThunk::Offset>>& offsets :
+       thunk.get_offsets()) {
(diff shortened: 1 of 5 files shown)


akuegel (approved):

akuegel:
It looks like this change is causing test failures in xla/backends/gpu/runtime/dynamic_slice_thunk_test.cc and xla/backends/gpu/runtime/gpublas_lt_matmul_thunk_test.cc

Also in //xla/backends/gpu/runtime:thunk_proto_deserialization_test: https://github.com/openxla/xla/actions/runs/29749387329/job/88375566893

@M0nd0R can you please take a look?

M0nd0R:
Thanks for the heads-up — looked into the failures.

**Root cause:** the new `FromProto` bounds check is correct; several existing test fixtures had allocation sizes that never matched the slices they deserialize (they only worked before because `FromProto` checked the index, not `offset`/`size`).

**What was failing**
1. `ThunkProtoDeserializationTest.CublasLtGroupedMatmulThunk` / `GpuBlasLtMatmulThunkTest.ThunkProtoSerializationGroupedMatmul`  
   Slice for allocation `#4` is `size=1302400` (`f32[2,407,400]`), but the fixture allocated `#4` as `651200` (the non-grouped `f32[407,400]` size).
2. `DynamicSliceThunkTest.*ProtoRoundTrip`  
   `CheckProtoRoundTrip` hard-coded every allocation to `1024`, so the `1MiB` workspace slices (`size=1048576`) failed the new check.

**Fix (pushed):**
- bump grouped-matmul allocation `#4` to `1302400` in both matmul proto tests
- size `CheckProtoRoundTrip` allocations from the thunk’s argument/offset slices so they always cover what `FromProto` reconstructs

Please re-run the affected targets when convenient:
- `//xla/backends/gpu/runtime:thunk_proto_deserialization_test`
- `//xla/backends/gpu/runtime:gpublas_lt_matmul_thunk_test`
- `//xla/backends/gpu/runtime:dynamic_slice_thunk_test`

akuegel (approved):

golechwierowicz (approved):

PR #43947, opened 2026-06-09
[XLA:GPU] ROCm VMM Layer 4/4: RocmDeviceAddressVmmAllocator + factory + tests

phambinhfin:
# [XLA:GPU] ROCm VMM PR 4/4: `RocmDeviceAddressVmmAllocator` + vendor-neutral factory

## Overview — VMM allocator for ROCm/HIP

Last of 4 stacked PRs bringing the **VMM (Virtual Memory Management) allocator** to
ROCm/HIP (split out of #40236 per review feedback). With this PR, `allocator=vmm` works
end-to-end on ROCm.

- **PR 1/4 — `RocmRawMemoryAllocation`** (#43942): physical allocation, `hipMemCreate`/`hipMemRelease`.
- **PR 2/4 — `RocmMemoryReservation`** (#43943): VA reserve / map / setAccess / unmap.
- **PR 3/4 — `RocmVmmAllocator`** (#43946): all-in-one allocator (create + reserve + map + setAccess).
- **PR 4/4 — `RocmDeviceAddressVmmAllocator` + vendor-neutral factory** (this PR, #43947).

> _Update the PR numbers above if they differ._

---

## This PR (PR 4/4): `RocmDeviceAddressVmmAllocator` + factory

The production VMM allocator: a `DeviceAddressVmmAllocator` implementation that adds
**GPU timeline-based deferred deallocation** on top of the lower layers, plus the
**vendor-neutral factory** that lets the PJRT GPU client select it without
platform `#if`s.

### Files
- `xla/stream_executor/rocm/rocm_device_address_vmm_allocator.{h,cc}` — ROCm allocator.
- `xla/stream_executor/rocm/rocm_device_address_vmm_allocator_test.cc` — unit tests.
- `xla/stream_executor/rocm/rocm_event.{h,cc}` — add `RocmEvent::Synchronize()`
  (`hipEventSynchronize`); the base returned `Unimplemented`, which the VMM unmap event needs.
- `xla/stream_executor/vmm_device_address_allocator.{h,cc}` — shared base additions
  (e.g. `GetAllocationInfo`/`AllocationInfo`, `GetReservation`) used by the allocator.
- `xla/stream_executor/vmm_device_address_allocator_factory.cc` — `DeviceAddressVmmAllocator::Create()`
  dispatching to the CUDA or ROCm implementation by platform.
- `xla/stream_executor/BUILD`, `xla/stream_executor/rocm/BUILD` — new targets/deps.

### Design
[...]

PR #44680, opened 2026-06-23
[XLA:GPU]  WIP ROCm VMM enablement : wire DeviceAddressVmmAllocator into PJRT GPU client  optimize

phambinhfin:
## Summary
- Enables the ROCm VMM allocator end-to-end by wiring `RocmDeviceAddressVmmAllocator` into the StreamExecutor PJRT GPU client (`se_gpu_pjrt_client.cc`), so `XLA_PYTHON_CLIENT_ALLOCATOR=vmm` selects it on ROCm (mirroring the existing CUDA path).
- Adds the `rocm_device_address_vmm_allocator` BUILD dependency.
- Extends the existing CUDA `VmmTest` (`VmmAllocatorCanBeSet`) to also run on ROCm, asserting the client's allocator is a `RocmDeviceAddressVmmAllocator`.

This is the final integration layer on top of the ROCm VMM allocator stack. It is stacked on #43947 (Layer 4/4: `RocmDeviceAddressVmmAllocator` + factory). Until #43947 merges, this PR's diff will also show the Layer 4 commit.

## Test plan
- [ ] `se_gpu_pjrt_client_test` `VmmTest.VmmAllocatorCanBeSet` passes on a ROCm (gfx942/gfx950) device.
- [ ] Build `//xla/pjrt/gpu:se_gpu_pjrt_client` under `--config=rocm`.
- [ ] Manual: `XLA_PYTHON_CLIENT_ALLOCATOR=vmm` selects the ROCm VMM allocator.

### DIFF
--- a/xla/pjrt/gpu/se_gpu_pjrt_client.cc
+++ b/xla/pjrt/gpu/se_gpu_pjrt_client.cc
@@ -153,6 +153,7 @@ limitations under the License.
 #include "xla/stream_executor/gpu/gpu_cudamallocasync_allocator.h"
 #elif TENSORFLOW_USE_ROCM
 #include "rocm/rocm_config.h"
+#include "xla/stream_executor/rocm/rocm_device_address_vmm_allocator.h"
 #endif
 
 #include "xla/service/gpu/gpu_executable_run_options.h"
@@ -1423,9 +1424,19 @@ GetStreamExecutorGpuDeviceAllocator(
       return se::gpu::CudaDeviceAddressVmmAllocator::Create(
           platform, allocator_config.memory_fraction,
           allocator_config.gpu_system_memory_size, executor_streams);
+#elif TENSORFLOW_USE_ROCM
+      std::vector<std::pair<se::StreamExecutor*, se::Stream*>> executor_streams;
+      executor_streams.reserve(addressable_devices.size());
+      for (const auto& [ordinal, device] : addressable_devices) {
+        executor_streams.push_back(
+            {device->executor(), device->compute_stream()});
+      }
+      return se::gpu::RocmDeviceAddressVmmAllocator::Create(
+          platform, allocator_config.memory_fraction,
+          allocator_config.gpu_system_memory_size, executor_streams);
 #else
       return absl::UnimplementedError(
-          "VMM allocator is only supported with CUDA.");
+          "VMM allocator is only supported with CUDA or ROCm.");
 #endif  // GOOGLE_CUDA
     }
 
--- a/xla/pjrt/gpu/BUILD
+++ b/xla/pjrt/gpu/BUILD
@@ -211,6 +211,7 @@ cc_library(
         "@local_config_cuda//cuda:cuda_headers",
     ]) + if_rocm([
         # keep sorted
+        "//xla/stream_executor/rocm:rocm_device_address_vmm_allocator",
         "@local_config_rocm//rocm:rocm_headers",
     ]) + if_sycl([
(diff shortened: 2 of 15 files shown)


phambinhfin:
Closing: this PR's diff incorrectly includes the stacked Layer 4 commits (1000+ line delta) and is not a clean standalone change. Will re-open a correctly scoped enablement PR.

phambinhfin:
@dimitar-asenov  Thanks for the reviewing and support for ROCm VMM feature. We add the first step optimization about selecting dynamic and big buffer sizes for mapping only and it should be target on ROCm.  Could you take a look?

dimitar-asenov:
@phambinhfin Do you mean the current PR? If so, please mark it as open and ready for review.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1425,9 +1424,19 @@ GetStreamExecutorGpuDeviceAllocator(
...
+            {device->executor(), device->compute_stream()});
+      }
+      return se::gpu::RocmDeviceAddressVmmAllocator::Create(
+          platform, allocator_config.memory_fraction,
+          allocator_config.gpu_system_memory_size, executor_streams);
 #else
       return absl::UnimplementedError(
-          "VMM allocator is only supported with CUDA.");
+          "VMM allocator is only supported with CUDA or ROCm.");
 #endif  // GOOGLE_CUDA
```
dimitar-asenov, line 1440: Please update this comment.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
```diff
@@ -1425,9 +1424,19 @@ GetStreamExecutorGpuDeviceAllocator(
       return se::gpu::CudaDeviceAddressVmmAllocator::Create(
           platform, allocator_config.memory_fraction,
           allocator_config.gpu_system_memory_size, executor_streams);
+#elif TENSORFLOW_USE_ROCM
+      std::vector<std::pair<se::StreamExecutor*, se::Stream*>> executor_streams;
+      executor_streams.reserve(addressable_devices.size());
+      for (const auto& [ordinal, device] : addressable_devices) {
+        executor_streams.push_back(
+            {device->executor(), device->compute_stream()});
+      }
```
dimitar-asenov, line 1433: This code is identical to the one above for CUDA, please restructure to reuse the logic:

You can have the common code on top, then do the #if check and just put the `Create` inside of it. This means that we'll always execute the common code even if we hit the else branch (not cuda and not rocm). However, since that's an error anyway, I think that's acceptable.

dimitar-asenov: Actually now that I've also read the code below, you should not need any `#if` at all. You can directly call `stream_executor::DeviceAddressVmmAllocator::Create` and that should dispatch (or fail) in the right way. Please do this instead.

xla/pjrt/gpu/se_gpu_pjrt_client.cc
dimitar-asenov: Or remove after you remove the `#if/#else`

phambinhfin:
> @phambinhfin Do you mean the current PR? If so, please mark it as open and ready for review.

Ah, I mean #45070 . Sorry for the confusion . Close this one because, this one was design based on the previous XLA optimization for VMM. However, it was remove so we decided to make some expriments in  #45070 and it works well as well.

PR #44702, opened 2026-06-23
increase default value for kMaxEvents

ScXfjiang:
Increase the default trace-viewer event cap from 1M to 5M so large XPlane profiles are not prematurely truncated when converted to Chrome trace JSON for large-scale models. With this default value, both trace json generation and TraceLens can work for at least 3 steps profiling for 8-node 405B model.

### DIFF
--- a/xla/tsl/profiler/convert/xplane_to_trace_events.cc
+++ b/xla/tsl/profiler/convert/xplane_to_trace_events.cc
@@ -107,7 +107,7 @@ void ConvertXPlaneToTraceEvents(uint32_t device_id, const XPlaneVisitor& xplane,
 }  // namespace
 
 uint64_t GetTraceViewerMaxEvents() {
-  constexpr uint64_t kMaxEvents = 1000000;
+  constexpr uint64_t kMaxEvents = 5000000;
   // Testing only env variable, not recommended for use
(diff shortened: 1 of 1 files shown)


ScXfjiang:
@xla-rotation, could you have a look please? Thanks in advance!

xla/tsl/profiler/convert/xplane_to_trace_events.cc
```diff
@@ -107,7 +107,7 @@ void ConvertXPlaneToTraceEvents(uint32_t device_id, const XPlaneVisitor& xplane,
 }  // namespace
 
 uint64_t GetTraceViewerMaxEvents() {
-  constexpr uint64_t kMaxEvents = 1000000;
+  constexpr uint64_t kMaxEvents = 5000000;
```
i-chaochen, line 110: if we think `kMaxEvents` is too small, we can specific `TF_PROFILER_TRACE_VIEWER_MAX_EVENTS` to get bigger one, isn't?

ScXfjiang: Yes, here I just change the default value so that it can work for more scenarios directly without explicitly overwriting it.

dimitar-asenov (approved):

PR #44738, opened 2026-06-24
Avoid splitting softmax scaled logits across CPU fusions

Liauuu:
📝 Summary of Changes
Prevent the CPU instruction fusion pass from splitting scaled softmax logits across separate fusions.
CpuInstructionFusion::ShouldFuse now rejects fusion patterns that separate the dot × scale computation used by the row-max reduction from the corresponding subtract/exp path, avoiding incorrect numerics caused by inconsistent softmax logits computations.

🎯 Justification
This change is an attempt to address the issue reported in jax-ml/jax#38609.
While investigating the report, I observed that eager execution and JIT matched through the exp computation. The NaNs appeared later in the softmax path under jax.jit on CPU. 

**expm(Q) and scaled logits match between eager and JIT**
Q:          match=True  max_abs_diff=0
logits:     match=True  max_abs_diff=0
logits_max: match=True  max_abs_diff=0

**First eager/JIT divergence occurs in the softmax path**
shifted:      match=False  (eager max=0, JIT max=2557.42)
exp_shifted:  match=False  (JIT: inf ×4)
denom:        match=False  (JIT: inf ×4)
weights:      match=False  (JIT: nan ×9)
out:          JIT   = [nan nan nan nan nan]
              eager = [2.108 0.701 -0.218 0.090 0.078]

I then tested with the CPU fusion pass disabled and found that the NaNs disappeared. 

**Disabling the fusion pass eliminates the issue**

XLA_FLAGS=--xla_disable_hlo_passes=fusion

eager[0]: [ 2.108  0.701 -0.218  0.090  0.078]
jit[0]:   [ 2.108  0.701 -0.218  0.090  0.078]
eager/jit match: True

From the generated HLO, it appeared that the scaled softmax logits computation was being split across separate fusion kernels across the reduction boundary. This could cause the row-max reduction path and the subtract/exp path to use different fused computations of the logits, leading to numerical inconsistencies and, in some cases, NaNs.

**HLO split observed across two fusion kernels**

%multiply_reduce_fusion
  - computes dot * scale for reduce_max

%subtract_exponential_fusion
  - recomputes dot * scale for subtract/exp

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
Added AvoidSplitSoftmaxScaledLogitsFusions in xla/service/cpu/cpu_instruction_fusion_test.cc.
The test builds a minimal softmax HLO, runs CpuInstructionFusion, and verifies that the problematic split-fusion pattern is not introduced.

I intended to validate the change end-to-end by rebuilding jaxlib and confirming that the NaNs disappear when this fusion pattern is prevented. However, I was unable to complete that verification in my environment due to build and resource constraints.

### DIFF
--- a/xla/service/cpu/cpu_instruction_fusion.cc
+++ b/xla/service/cpu/cpu_instruction_fusion.cc
@@ -19,7 +19,9 @@ limitations under the License.
 
 #include "absl/algorithm/container.h"
 #include "absl/log/log.h"
+#include "absl/types/span.h"
 #include "xla/codegen/emitters/elemental_hlo_to_mlir.h"
+#include "xla/hlo/analysis/hlo_reachability.h"
 #include "xla/hlo/ir/hlo_casting_utils.h"
 #include "xla/hlo/ir/hlo_instruction.h"
 #include "xla/hlo/ir/hlo_instructions.h"
--- a/xla/service/cpu/cpu_instruction_fusion.cc
+++ b/xla/service/cpu/cpu_instruction_fusion.cc
@@ -103,6 +103,79 @@ bool BlockSubcomputationFusion(const HloInstruction* instruction,
   return false;
 }
 
+bool FusionInstructionContainsOpcode(const HloInstruction& fusion,
+                                     HloOpcode opcode) {
+  if (!fusion.IsFusion()) {
+    return fusion.opcode() == opcode;
+  }
+  return absl::c_any_of(
+      fusion.fused_instructions(),
+      [opcode](const HloInstruction* instr) { return instr->opcode() == opcode; });
+}
+
+bool MultiplyFeedsReduceAndSubtract(const HloInstruction& multiply) {
+  bool feeds_reduce = false;
+  bool feeds_subtract = false;
+  auto note_user = [&](const HloInstruction& user) {
+    if (user.opcode() == HloOpcode::kReduce) {
+      feeds_reduce = true;
+    }
+    if (user.opcode() == HloOpcode::kSubtract) {
+      feeds_subtract = true;
+    }
+    if (user.IsFusion()) {
+      if (FusionInstructionContainsOpcode(user, HloOpcode::kReduce)) {
+        feeds_reduce = true;
+      }
+      if (FusionInstructionContainsOpcode(user, HloOpcode::kSubtract)) {
+        feeds_subtract = true;
+      }
+    }
+  };
+  for (const HloInstruction* user : multiply.users()) {
+    note_user(*user);
+  }
+  return feeds_reduce && feeds_subtract;
+}
+
+bool WouldSplitSoftmaxScaledMultiply(const HloInstruction* producer,
+                                     const HloInstruction* consumer) {
+  if (producer->opcode() != HloOpcode::kMultiply) {
+    return false;
+  }
+  if (!MultiplyFeedsReduceAndSubtract(*producer)) {
+    return false;
+  }
+  if (consumer->opcode() == HloOpcode::kReduce ||
+      consumer->opcode() == HloOpcode::kSubtract) {
+    return true;
+  }
+  if (consumer->IsFusion() &&
+      (FusionInstructionContainsOpcode(*consumer, HloOpcode::kReduce) ||
+       FusionInstructionContainsOpcode(*consumer, HloOpcode::kSubtract))) {
+    return true;
+  }
+  return false;
+}
+
+bool WouldFuseDotIntoSplitSoftmaxFusion(const HloInstruction* producer,
+                                        const HloInstruction* consumer) {
+  if (producer->opcode() != HloOpcode::kDot || !consumer->IsFusion()) {
+    return false;
+  }
+  if (!FusionInstructionContainsOpcode(*consumer, HloOpcode::kSubtract) ||
+      !FusionInstructionContainsOpcode(*consumer, HloOpcode::kExp)) {
+    return false;
+  }
+  for (const HloInstruction* user : producer->users()) {
+    if (user->opcode() == HloOpcode::kMultiply &&
+        MultiplyFeedsReduceAndSubtract(*user)) {
+      return true;
+    }
+  }
+  return false;
+}
+
 }  // namespace
 
 bool CpuInstructionFusion::IsExpensive(const HloInstruction& instruction) {
@@ -382,6 +455,20 @@ FusionDecision CpuInstructionFusion::ShouldFuse(HloInstruction* consumer,
   if (producer->opcode() != HloOpcode::kFusion && is_expensive(*producer) &&
       ReusesOperandElements(consumer, operand_index)) {
     return FusionDecision::Forbid("Fusion is not profitable.");
+  }
+
+  // Stable softmax lowers as: dot -> multiply(scale) -> reduce(max) and
+  // subtract(multiply, broadcast(max)) -> exponential. Fusing multiply only into
+  // reduce, then fusing dot into subtract+exponential separately, recomputes
+  // scaled logits in a second kernel and can miscompile for large values.
+  if (WouldSplitSoftmaxScaledMultiply(producer, consumer)) {
+    return FusionDecision::Forbid(
+        "Don't split softmax scaled logits across separate fusions.");
+  }
+  if (WouldFuseDotIntoSplitSoftmaxFusion(producer, consumer)) {
+    return FusionDecision::Forbid(
+        "Don't fuse dot into softmax subtract-exponential fusion separately "
+        "from its scaled-logits reduce.");
   }
 
   RETURN_IF_NOT_FUSIBLE(InstructionFusion::ShouldFuse(consumer, operand_index));
--- a/xla/service/cpu/cpu_instruction_fusion.cc
+++ b/xla/service/cpu/cpu_instruction_fusion.cc
@@ -90,7 +90,8 @@ bool IsMaxReduction(const HloInstruction* reduce) {
   return root->opcode() == HloOpcode::kMaximum;
 }
 
-// Matches subtract(shift_value, broadcast(reduce)) used by exponential, i.e.
+// Matches subtract(shift_value, broadcast(reduce)) or the reversed operand
+// order subtract(broadcast(reduce), shift_value), used by exponential, i.e.
 // the numerically sensitive stable-softmax shift pattern.
 bool IsExpShiftCoupledSubtract(const HloInstruction& subtract,
                                const HloInstruction& shift_value,
@@ -349,6 +350,12 @@ CpuInstructionFusion::ComputeGloballyUnfusible(
       InstructionFusion::ComputeGloballyUnfusible(post_order, reachability);
   for (HloInstruction* producer : post_order) {
     if (IsCoupledReductionShiftExpProducer(producer)) {
+      // The base implementation treats effectively-unary elementwise ops as
+      // free to duplicate. For producers feeding both a max-reduce and its
+      // exp-shift subtract, duplication is unsound: FMA contraction can make
+      // the recomputed copy differ from the copy the max was taken over by up
+      // to one ulp, which exp() turns into inf/nan. Force such producers to be
+      // materialized once.
       do_not_duplicate.insert(producer);
     }
   }
--- a/xla/service/cpu/cpu_instruction_fusion.cc
+++ b/xla/service/cpu/cpu_instruction_fusion.cc
@@ -16,6 +16,7 @@ limitations under the License.
 #include "xla/service/cpu/cpu_instruction_fusion.h"
 
 #include <cstdint>
+#include <functional>
 
 #include "absl/algorithm/container.h"
 #include "absl/log/log.h"
@@ -82,6 +83,141 @@ bool CanBeOutputFusedIntoSomeOperand(const HloInstruction* consumer) {
           CanBeOutputFused(consumer->operand(1), consumer));
 }
 
+bool IsMaxReduction(const HloInstruction* reduce) {
+  if (reduce->opcode() != HloOpcode::kReduce) {
+    return false;
+  }
+  const HloInstruction* root = reduce->to_apply()->root_instruction();
+  return root->opcode() == HloOpcode::kMaximum;
+}
+
+bool IsDotLikeOpcode(HloOpcode opcode) {
+  return opcode == HloOpcode::kDot;
+}
+
+// Matches subtract(shift_value, broadcast(reduce)) used by exponential, i.e.
+// the numerically sensitive stable-softmax shift pattern.
+bool IsExpShiftCoupledSubtract(const HloInstruction& subtract,
+                               const HloInstruction& shift_value,
+                               const HloInstruction& reduce) {
+  if (subtract.opcode() != HloOpcode::kSubtract) {
+    return false;
+  }
+  const HloInstruction* op0 = subtract.operand(0);
+  const HloInstruction* op1 = subtract.operand(1);
+  const bool uses_shift_value =
+      op0 == &shift_value || op1 == &shift_value ||
+      (op0->opcode() == HloOpcode::kBroadcast &&
+       op0->operand(0) == &shift_value) ||
+      (op1->opcode() == HloOpcode::kBroadcast &&
+       op1->operand(0) == &shift_value);
+  const bool uses_reduce_broadcast =
+      (op0->opcode() == HloOpcode::kBroadcast &&
+       op0->operand(0) == &reduce) ||
+      (op1->opcode() == HloOpcode::kBroadcast &&
+       op1->operand(0) == &reduce);
+  if (!uses_shift_value || !uses_reduce_broadcast) {
+    return false;
+  }
+  return absl::c_any_of(subtract.users(), [](const HloInstruction* user) {
+    return user->opcode() == HloOpcode::kExp;
+  });
+}
+
+const HloInstruction* FindMaxReduceForShiftValue(
+    const HloInstruction& shift_value) {
+  for (const HloInstruction* user : shift_value.users()) {
+    if (user->opcode() == HloOpcode::kReduce && user->operand(0) == &shift_value &&
+        IsMaxReduction(user)) {
+      return user;
+    }
+  }
+  return nullptr;
+}
+
+// A multiply that feeds both row-max reduction and exp-shift subtraction.
+bool IsCoupledReductionShiftExpProducer(const HloInstruction* instr) {
+  if (instr->opcode() != HloOpcode::kMultiply) {
+    return false;
+  }
+  const HloInstruction* reduce = FindMaxReduceForShiftValue(*instr);
+  if (reduce == nullptr) {
+    return false;
+  }
+  return absl::c_any_of(instr->users(), [&](const HloInstruction* user) {
+    return IsExpShiftCoupledSubtract(*user, *instr, *reduce);
+  });
+}
+
+bool InstructionOrFusionContains(
+    const HloInstruction* instr,
+    const std::function<bool(const HloInstruction*)>& pred) {
+  if (instr->opcode() != HloOpcode::kFusion) {
+    return pred(instr);
+  }
+  return absl::c_any_of(instr->fused_instructions(), pred);
+}
+
+bool IsReduceSideConsumer(const HloInstruction* consumer,
+                            const HloInstruction* shift_value) {
+  return InstructionOrFusionContains(consumer, [&](const HloInstruction* instr) {
+    return instr->opcode() == HloOpcode::kReduce &&
+           instr->operand(0) == shift_value && IsMaxReduction(instr);
+  });
+}
+
+bool IsShiftExpSideConsumer(const HloInstruction* consumer,
+                            const HloInstruction* shift_value,
+                            const HloInstruction* reduce) {
+  return InstructionOrFusionContains(consumer, [&](const HloInstruction* instr) {
+    return IsExpShiftCoupledSubtract(*instr, *shift_value, *reduce);
+  });
+}
+
+bool FusionLooksLikeShiftExpPath(const HloInstruction* consumer) {
+  if (consumer->opcode() != HloOpcode::kFusion) {
+    return false;
+  }
+  return absl::c_any_of(consumer->fused_instructions(),
+                        [](const HloInstruction* instr) {
+                          return instr->opcode() == HloOpcode::kExp ||
+                                 instr->opcode() == HloOpcode::kSubtract;
+                        });
+}
+
+// Returns true if fusing producer into consumer targets one side of a coupled
+// reduce-max / exp-shift producer that must stay numerically consistent.
+bool HasCoupledReductionShiftExpSplitRisk(const HloInstruction* producer,
+                                          const HloInstruction* consumer) {
+  const HloInstruction* shift_value = nullptr;
+  if (IsCoupledReductionShiftExpProducer(producer)) {
+    shift_value = producer;
+  } else if (IsDotLikeOpcode(producer->opcode())) {
+    for (const HloInstruction* user : producer->users()) {
+      if (IsCoupledReductionShiftExpProducer(user)) {
+        shift_value = user;
+        break;
+      }
+    }
+  }
+  if (shift_value == nullptr) {
+    return false;
+  }
+
+  const HloInstruction* reduce = FindMaxReduceForShiftValue(*shift_value);
+  if (reduce == nullptr) {
+    return false;
+  }
+
+  if (producer != shift_value) {
+    return IsShiftExpSideConsumer(consumer, shift_value, reduce) ||
+           FusionLooksLikeShiftExpPath(consumer);
+  }
+
+  return IsReduceSideConsumer(consumer, shift_value) ||
+         IsShiftExpSideConsumer(consumer, shift_value, reduce);
+}
+
 // Should we block the fusion of the subcomputation of the passed instruction?
 bool BlockSubcomputationFusion(const HloInstruction* instruction,
                                const HloModuleConfig& config) {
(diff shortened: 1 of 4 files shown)


seantalts (commented):
Thanks for digging into this — the root-cause analysis (reduce_max seeing different scaled-logit values than the subtract path → positive `shifted` → `exp` overflow → NaN) is convincing and well-documented. A few concerns on approach and robustness before this lands:

1. **The pattern match is much broader than softmax** (see inline on `MultiplyFeedsReduceAndSubtract`). As written it forbids fusion for *any* multiply feeding a reduce and a subtract, which will catch unrelated normalization/variance patterns and regress their fusion with no numerics benefit.
2. **Altitude:** this special-cases a softmax shape in the generic cost model, but the underlying invariant being violated is "recomputed fused sub-expressions are numerically identical across kernels." That can recur for any reduce-then-shift pattern the matcher doesn't enumerate. Worth discussing whether the fix belongs at the recomputation-consistency level (or a general "don't split a reduce-input that's reused post-reduction" rule) rather than a softmax matcher.
3. **End-to-end verification:** the PR notes the NaN-disappears check couldn't be completed, and the unit test only asserts the structural split pattern isn't produced on a hand-written module. Could you confirm the guard actually fires in the real jax lowering and the runtime NaNs are gone? Fusion ordering means the reduce/subtract may already be inside fusions by the time `ShouldFuse` runs, so the structural test passing doesn't guarantee the runtime fix.

xla/service/cpu/cpu_instruction_fusion.cc
```diff
@@ -107,6 +107,79 @@ bool BlockSubcomputationFusion(const HloInstruction* instruction,
...
+                                     HloOpcode opcode) {
+  if (!fusion.IsFusion()) {
+    return fusion.opcode() == opcode;
+  }
+  return absl::c_any_of(
+      fusion.fused_instructions(),
+      [opcode](const HloInstruction* instr) { return instr->opcode() == opcode; });
+}
+
+bool MultiplyFeedsReduceAndSubtract(const HloInstruction& multiply) {
```
seantalts: This matches *any* `multiply` whose users include a reduce and a subtract — it never checks that the multiply consumes a `dot`, that the other operand is a broadcast scalar (the scale), or that the subtract's other operand is `broadcast(reduce_max)`. Patterns like layernorm/variance (`x*w` feeding both a reduction and a residual subtract) will match incidentally and lose vectorized fusion for no numerics gain. If the intent is softmax-specific, consider verifying the dot producer and the scale/broadcast-max operand shapes here.

xla/service/cpu/cpu_instruction_fusion.cc
```diff
@@ -388,6 +461,20 @@ FusionDecision CpuInstructionFusion::ShouldFuse(HloInstruction* consumer,
     return FusionDecision::Forbid("Fusion is not profitable.");
   }
 
+  // Stable softmax lowers as: dot -> multiply(scale) -> reduce(max) and
+  // subtract(multiply, broadcast(max)) -> exponential. Fusing multiply only into
+  // reduce, then fusing dot into subtract+exponential separately, recomputes
+  // scaled logits in a second kernel and can miscompile for large values.
+  if (WouldSplitSoftmaxScaledMultiply(producer, consumer)) {
```
seantalts: These guards treat the symptom (split scaled logits) rather than the underlying issue: XLA recomputes `dot*scale` in two kernels and the recomputations aren't bit-identical. That assumption — recomputed fused expressions are consistent — is violated generally, so near-softmax variants the matcher doesn't enumerate can still NaN. Could we generalize to "don't split a reduce input that's also reused after the reduction," or fix recomputation consistency directly?

xla/service/cpu/cpu_instruction_fusion_test.cc
```diff
@@ -1140,5 +1140,81 @@ TEST_F(InstructionFusionTest, SkipReduceComputationsIfFusionEmitters) {
...
+    %exp = f32[5,5] exponential(%sub)
+    %zero = f32[] constant(0)
+    %reduce_sum = f32[5] reduce(%exp, %zero),
+        dimensions={1}, to_apply=%region_1.2
+    %broadcast_sum = f32[5,5] broadcast(%reduce_sum), dimensions={0}
+    ROOT %div = f32[5,5] divide(%exp, %broadcast_sum)
+  }
+)";
+
+TEST_F(InstructionFusionTest, AvoidSplitSoftmaxScaledLogitsFusions) {
```
seantalts: This asserts the structural split pattern isn't produced, but not that the runtime NaN is actually fixed. Given the PR couldn't verify end-to-end, an execution test (or a note confirming the jax repro now matches eager) would make this much stronger. Also `HasSplitSoftmaxSubtractExponentialFusion` hard-codes `operand(0)==kFusion && operand(1)==kDot` ordering — brittle if operand order shifts.

seantalts (changes requested):

Liauuu:
> Thanks for digging into this — the root-cause analysis (reduce_max seeing different scaled-logit values than the subtract path → positive `shifted` → `exp` overflow → NaN) is convincing and well-documented. A few concerns on approach and robustness before this lands:
> 
> 1. **The pattern match is much broader than softmax** (see inline on `MultiplyFeedsReduceAndSubtract`). As written it forbids fusion for _any_ multiply feeding a reduce and a subtract, which will catch unrelated normalization/variance patterns and regress their fusion with no numerics benefit.
> 2. **Altitude:** this special-cases a softmax shape in the generic cost model, but the underlying invariant being violated is "recomputed fused sub-expressions are numerically identical across kernels." That can recur for any reduce-then-shift pattern the matcher doesn't enumerate. Worth discussing whether the fix belongs at the recomputation-consistency level (or a general "don't split a reduce-input that's reused post-reduction" rule) rather than a softmax matcher.
> 3. **End-to-end verification:** the PR notes the NaN-disappears check couldn't be completed, and the unit test only asserts the structural split pattern isn't produced on a hand-written module. Could you confirm the guard actually fires in the real jax lowering and the runtime NaNs are gone? Fusion ordering means the reduce/subtract may already be inside fusions by the time `ShouldFuse` runs, so the structural test passing doesn't guarantee the runtime fix.

@seantalts Thank you very much for taking the time to review this.

I still think this approach is matching a specific pattern, and I'm not sure yet whether it's the right long-term abstraction. Since I wasn't able to complete the end-to-end verification before, I first wanted to confirm whether this change actually fixes the issue in the real JAX lowering. I also thought about how to make it more general, but I'm not sure of the best approach yet. If this still needs to be generalized further, I'll keep working on it.

Based on your feedback, I narrowed the matching logic so it doesn't affect unrelated patterns as much.

Instead of the previous `MultiplyFeedsReduceAndSubtract` helper, I replaced it with `HasCoupledReductionShiftExpSplitRisk`, which only matches the numerically coupled pattern (max-reduction together with the corresponding subtract+exp path), rather than any multiply feeding both a reduce and a subtract. I also added the same pattern to `ComputeGloballyUnfusible` so that these coupled producers are treated as do-not-duplicate candidates, instead of relying only on the `ShouldFuse` check.

I was also able to complete the end-to-end verification this time. The original JAX reproducer no longer produces NaNs, and the guard does fire on the real JAX lowering.

Before:

```text
========================================================================
Interpretation
========================================================================
Through matrix expm + scaled logits (Q, logits, logits_max):
  all match: True

Softmax numerics (shifted, exp, normalize, output):
  shifted: match=False  max_abs_diff=8192
  out: match=False  max_abs_diff=nan

eager out[0] = [ 2.1076884  0.7011127 -0.21764016  0.08972669  0.07841968]
jit   out[0] = [nan nan nan nan nan]

------------------------------------------------------------

fused_computation.19 (row-max path recomputes dot*scale inside fusion):
  %mul.10 = multiply(...)
  ROOT %reduce_max = reduce(%mul.10)

fused_computation.18 (subtract+exp path recomputes dot*scale again):
  %mul.9 = multiply(...)
  %sub.20 = subtract(%mul.9, ...)
```

After:

```text
========================================================================
Eager vs JIT step-by-step comparison
========================================================================
Q: match=True
logits: match=True
logits_max: match=True
shifted: match=True
exp_shifted: match=True
denom: match=True
weights: match=True
out: match=True

eager out[0] = [ 2.1076884  0.7011127 -0.21764016  0.08972669  0.07841968]
jit   out[0] = [ 2.1076884  0.7011127 -0.21764016  0.08972669  0.07841968]

------------------------------------------------------------

Entry computation:

%broadcast_multiply_fusion.1 = fusion(...)
%subtract_exponential_fusion = fusion(...)

fused_computation.19 (scaled logits isolated):
  ROOT %mul.9 = multiply(...)

fused_computation.18 (subtract+exp reuses the output instead of recomputing):
  %sub.20 = subtract(...)
  ROOT %exp.0 = exponential(...)
```

xla/service/cpu/cpu_instruction_fusion.cc
Liauuu: Thank you very much for the review. Based on your suggestion, I narrowed the matching so it is limited to the softmax case as much as possible. Instead of matching any multiply feeding both a reduce and a subtract, it now only matches the coupled max-reduction and subtract+exp pattern.

xla/service/cpu/cpu_instruction_fusion.cc
Liauuu: Thank you very much for the suggestion. I also applied the same logic in ComputeGloballyUnfusible, in addition to ShouldFuse. I spent some time thinking about a more general solution, but I wasn't confident in one yet. Since I couldn't complete the end-to-end verification before, I first wanted to verify that this approach actually fixes the issue.

xla/service/cpu/cpu_instruction_fusion_test.cc
Liauuu: Thank you very much for the review. I was able to complete the end-to-end verification this time, and fortunately the original JAX reproducer no longer produces NaNs. I also updated the test to avoid depending on a fixed operand order by searching the fusion instructions instead of assuming specific operand positions.

xla/service/cpu/cpu_instruction_fusion.cc
```diff
@@ -390,6 +540,15 @@ FusionDecision CpuInstructionFusion::ShouldFuse(HloInstruction* consumer,
 
   RETURN_IF_NOT_FUSIBLE(InstructionFusion::ShouldFuse(consumer, operand_index));
 
+  // Fusing one side of a reduce-max / exp-shift coupled producer into a
+  // separate kernel can recompute the shift value with different numerics.
+  if (HasCoupledReductionShiftExpSplitRisk(producer, consumer) &&
```
seantalts: Good news, I think: as far as I can trace the base pass, this guard never fires with effect, so you may be able to delete it along with `HasCoupledReductionShiftExpSplitRisk`, `IsReduceSideConsumer`, `IsShiftExpSideConsumer`, `FusionLooksLikeShiftExpPath`, and `InstructionOrFusionContains`, and keep just the matcher + override:

1. **The multiply case is pre-empted by `do_not_duplicate`.** In `InstructionFusion::RunImpl` (instruction_fusion.cc:671), `do_not_duplicate.count(operand)` is checked *before* `ShouldFuse` is consulted, so once your `ComputeGloballyUnfusible` override marks the coupled multiply, regular fusion of it is already blocked and this guard is never reached for it.
2. **The dot case is unreachable.** A `kDot` producer is rejected much earlier in this function by `CanBeLoopFused(*producer)`, well before this check.
3. **The fusion-interior matching can't match.** Inside a fusion, the subtract/reduce operands are fused *parameters*, so the pointer comparisons in `IsExpShiftCoupledSubtract` / `IsReduceSideConsumer` against the outer multiply/reduce always fail, and `InstructionOrFusionContains` with these predicates returns false for any fusion consumer.

For what it's worth, your `ComputeGloballyUnfusible` override is doing the real work here, and it's a nice catch: the base implementation skips this multiply via its `EffectivelyAtMostUnary` early-out (the broadcast-scalar scale operand makes it look free to duplicate), which is exactly how the split arose in the first place.

Suggest verifying by deleting this guard and re-running the JAX repro + the new test. If they still pass, the PR shrinks to the matcher + override.

xla/service/cpu/cpu_instruction_fusion.cc
```diff
@@ -82,6 +83,141 @@ bool CanBeOutputFusedIntoSomeOperand(const HloInstruction* consumer) {
...
+        IsMaxReduction(user)) {
+      return user;
+    }
+  }
+  return nullptr;
+}
+
+// A multiply that feeds both row-max reduction and exp-shift subtraction.
+bool IsCoupledReductionShiftExpProducer(const HloInstruction* instr) {
+  if (instr->opcode() != HloOpcode::kMultiply) {
```
seantalts: The narrowing here looks good and clearly fixes the repro. One thought for a possible follow-up (fine to defer): the invariant being protected (a cheap elementwise shift value duplicated into both the max-reduce kernel and the exp-shift kernel) doesn't actually depend on the op being a multiply. `add(dot, bias)` feeding the same coupled pattern has the identical recompute-divergence risk. Matching any elementwise instruction here (rather than just multiply) would be a natural step toward the generalization we discussed, without broadening beyond the numerically coupled shape.

xla/service/cpu/cpu_instruction_fusion_test.cc
```diff
@@ -1140,5 +1140,144 @@ TEST_F(InstructionFusionTest, SkipReduceComputationsIfFusionEmitters) {
...
+
+  TF_ASSERT_OK_AND_ASSIGN(auto module,
+                          ParseAndReturnVerifiedModule(module_string));
+  TF_ASSERT_OK_AND_ASSIGN(bool changed,
+                          CpuInstructionFusion(&alias_info_).Run(module.get()));
+  EXPECT_TRUE(changed);
+  EXPECT_FALSE(HasSplitCoupledReductionShiftExpFusion(*module));
+}
+
+TEST_F(InstructionFusionTest, StillFusesUncoupledReduceAndSubtract) {
```
seantalts: Thanks for adding a negative test; the idea is right, but as written it doesn't exercise the matcher's discrimination: the module contains no multiply, so `IsCoupledReductionShiftExpProducer` trivially rejects everything, and the assertion is only `changed == true`, which any fusion in the module satisfies. A sharper version: a multiply feeding an *add*-reduce plus subtract (or a max-reduce whose subtract doesn't feed exp), asserting the multiply still gets fused, e.g. that no standalone multiply remains in the entry computation.

seantalts (commented):
Thank you for the thorough follow-up, and for sticking with this one! Root-causing a numerics bug across the eager/JIT boundary and then confirming the fix end-to-end with before/after HLO is exactly the kind of investigation that makes these reviews easy. The narrowed matcher plus the `ComputeGloballyUnfusible` override address the earlier concerns well, and the repro data in the PR description is very helpful.

One structural suggestion that I think is good news: tracing the new code against the base pass, I believe your `ComputeGloballyUnfusible` override is the entire effective fix, and the `ShouldFuse` guard plus its consumer-matching helpers are dead code (details inline on the guard). If that trace is right, this PR can shrink to roughly `IsCoupledReductionShiftExpProducer` + the override (~40 lines), which would make it even easier to review and maintain. Worth verifying by deleting the guard and re-running your JAX repro and the new unit test.

Two smaller things:
- The negative test doesn't exercise the matcher yet (inline).
- A few lines exceed 80 columns (e.g. the lambda in `FindMaxReduceForShiftValue`, `fused_instructions_computation()` in the test), so worth a clang-format pass.

Thanks again for the careful work here.

Liauuu:
> Thank you for the thorough follow-up, and for sticking with this one! Root-causing a numerics bug across the eager/JIT boundary and then confirming the fix end-to-end with before/after HLO is exactly the kind of investigation that makes these reviews easy. The narrowed matcher plus the `ComputeGloballyUnfusible` override address the earlier concerns well, and the repro data in the PR description is very helpful.
> 
> One structural suggestion that I think is good news: tracing the new code against the base pass, I believe your `ComputeGloballyUnfusible` override is the entire effective fix, and the `ShouldFuse` guard plus its consumer-matching helpers are dead code (details inline on the guard). If that trace is right, this PR can shrink to roughly `IsCoupledReductionShiftExpProducer` + the override (~40 lines), which would make it even easier to review and maintain. Worth verifying by deleting the guard and re-running your JAX repro and the new unit test.
> 
> Two smaller things:
> 
> * The negative test doesn't exercise the matcher yet (inline).
> * A few lines exceed 80 columns (e.g. the lambda in `FindMaxReduceForShiftValue`, `fused_instructions_computation()` in the test), so worth a clang-format pass.
> 
> Thanks again for the careful work here.

@seantalts Thank you very much for the detailed review.
Following your suggestion, I removed the ShouldFuse guard and the related consumer-matching helpers. I also expanded the tests with several additional cases as you recommended, so they cover the matcher more thoroughly, and ran clang-format to clean up the formatting.
Below are the test results.
I also re-ran the end-to-end verification:

**Softmax repro unit test**
INFO: Analyzed target //xla/service/cpu:cpu_instruction_fusion_test (0 packages loaded, 21 targets configured). INFO: Found 1 test target... Target //xla/service/cpu:cpu_instruction_fusion_test up-to-date: bazel-bin/xla/service/cpu/cpu_instruction_fusion_test INFO: Elapsed time: 2.206s, Critical Path: 0.20s INFO: 2 processes: 1 internal, 1 local. INFO: Build completed successfully, 2 total actions //xla/service/cpu:cpu_instruction_fusion_test PASSED in 0.2s

**CPU instruction fusion regression tests**
INFO: Analyzed target //xla/service/cpu:cpu_instruction_fusion_test (0 packages loaded, 21 targets configured). INFO: Found 1 test target... Target //xla/service/cpu:cpu_instruction_fusion_test up-to-date: bazel-bin/xla/service/cpu/cpu_instruction_fusion_test INFO: Elapsed time: 2.206s, Critical Path: 0.20s INFO: 2 processes: 1 internal, 1 local. INFO: Build completed successfully, 2 total actions //xla/service/cpu:cpu_instruction_fusion_test PASSED in 0.2s


**HLO Emitters **
***01_multiply_softmax :Protected** 
 %dot = f32[5,5]{1,0} dot(%lhs, %rhs), lhs_contracting_dims={1}, rhs_contracting_dims={0}
  %broadcast_multiply_fusion = f32[5,5]{1,0} fusion(%dot), kind=kLoop, calls=%fused_computation.1
  %wrapped_reduce = f32[5]{0} fusion(%broadcast_multiply_fusion, %neg_inf), kind=kLoop, calls=%wrapped_reduce_computation
  ROOT %subtract_exponential_fusion = f32[5,5]{1,0} fusion(%broadcast_multiply_fusion, %wrapped_reduce), kind=kLoop, calls=%fused_computation

***02_add_bias_softmax: Protected** 

 %dot = f32[5,5]{1,0} dot(%lhs, %rhs), lhs_contracting_dims={1}, rhs_contracting_dims={0}
  %wrapped_add = f32[5,5]{1,0} fusion(%dot, %bias), kind=kLoop, calls=%wrapped_add_computation
  %wrapped_reduce = f32[5]{0} fusion(%wrapped_add, %neg_inf), kind=kLoop, calls=%wrapped_reduce_computation
  ROOT %subtract_exponential_fusion = f32[5,5]{1,0} fusion(%wrapped_add, %wrapped_reduce), kind=kLoop, calls=%fused_computation

***03_sum_reduce_not_coupled: Fused/Split** 
 %dot = f32[5,5]{1,0} dot(%lhs, %rhs), lhs_contracting_dims={1}, rhs_contracting_dims={0}
  %multiply_reduce_fusion = f32[5]{0} fusion(%dot), kind=kLoop, calls=%fused_computation.1
  ROOT %subtract_exponential_fusion = f32[5,5]{1,0} fusion(%multiply_reduce_fusion, %dot), kind=kLoop, calls=%fused_computation

***04_no_exp_not_coupled: Fused/Split** 
 %dot = f32[5,5]{1,0} dot(%lhs, %rhs), lhs_contracting_dims={1}, rhs_contracting_dims={0}
  %multiply_reduce_fusion = f32[5]{0} fusion(%dot), kind=kLoop, calls=%fused_computation.1
  ROOT %subtract_negate_fusion = f32[5,5]{1,0} fusion(%multiply_reduce_fusion, %dot), kind=kLoop, calls=%fused_computation

**jaxlib repro**
=== repro ===
multiply softmax (scale * q@k.T): allclose=True, max_abs_diff=0.000000e+00
  PASS
add softmax (q@k.T + bias): allclose=True, max_abs_diff=0.000000e+00
  PASS

Thank you again for taking the time to trace through the pass so carefully.

xla/service/cpu/cpu_instruction_fusion.cc
Liauuu: Thank you! I removed the dead code you pointed out and verified that both the unit tests and the JAX repro still pass after removing it.

xla/service/cpu/cpu_instruction_fusion.cc
Liauuu: Thank you. Following your suggestion, I changed the matcher to handle elementwise producers instead of only multiply. I also excluded kFusion, since I wanted to avoid treating fusion instructions themselves as elementwise producers.

xla/service/cpu/cpu_instruction_fusion_test.cc
Liauuu: Thank you. Following your suggestion, I expanded the tests to cover the additional cases. I included the add-based coupled pattern, as well as the uncoupled sum-reduce and non-exp cases to make sure they still fuse as expected. I also updated the negative tests to verify that no standalone producer remains, instead of only checking changed == true.

seantalts:
I'm seeing a lot of failing CI stuff, could you take a look?

Liauuu:
> I'm seeing a lot of failing CI stuff, could you take a look?

I fixed dwyu dependency issues and rebased onto latest main and re-checked clang-format on my side.
For many of the remaining failures (e.g. XLA Linux x86 CPU/GPU L4, JAX Linux x86 CPU Bzlmod, clang-tidy-cuda, and the benchmark builds), they all report:
Error: ScriptExecutorError
 Executing the custom container implementation failed.
so it looks like there may be a common issue with the self-hosted runner/custom container. The benchmark jobs also failed to build the binaries, which then caused the artifact upload steps to fail because no benchmark_output_* directory was generated.
The AMD Instinct GPU job (Test XLA [single_gpu]) failed with exit code 3, but I'm not sure whether that's related or a separate issue.
Thanks again for the feedback—I really appreciate it!

xla/service/cpu/cpu_instruction_fusion.cc
```diff
@@ -283,6 +345,20 @@ bool CpuInstructionFusion::IsExpensive(const HloInstruction& instruction) {
   return false;
 }
 
+InstructionFusion::HloInstructionSet
+CpuInstructionFusion::ComputeGloballyUnfusible(
+    absl::Span<HloInstruction* const> post_order,
```
seantalts, line 349: Please add a comment explaining why this insert is needed. Two things a reader can't get from the code:

1. The base implementation skips these producers via `EffectivelyAtMostUnary` (the broadcast-scalar operand makes the multiply look free to duplicate).
2. Duplication is unsafe because the two copies aren't bit-identical: the subtract kernel can contract multiply+subtract into a fused multiply-subtract (unrounded product), while the reduce kernel uses the rounded product. That puts `shifted` up to one ulp above zero, and exp overflows when logits are large. Your repro numbers match this: 2557 and 8192 are about one ulp of logits near 1e10.

Suggested comment:
```
// The base implementation treats effectively-unary elementwise ops as free
// to duplicate. For producers feeding both a max-reduce and its exp-shift
// subtract, duplication is unsound: FMA contraction can make the recomputed
// copy differ from the copy the max was taken over by up to one ulp, which
// exp() turns into inf/nan. Force such producers to be materialized once.
```

xla/service/cpu/cpu_instruction_fusion.cc
```diff
@@ -82,6 +82,68 @@ bool CanBeOutputFusedIntoSomeOperand(const HloInstruction* consumer) {
...
+bool IsMaxReduction(const HloInstruction* reduce) {
+  if (reduce->opcode() != HloOpcode::kReduce) {
+    return false;
+  }
+  const HloInstruction* root = reduce->to_apply()->root_instruction();
+  return root->opcode() == HloOpcode::kMaximum;
+}
+
+// Matches subtract(shift_value, broadcast(reduce)) used by exponential, i.e.
+// the numerically sensitive stable-softmax shift pattern.
```
seantalts, line 97: The comment describes `subtract(shift_value, broadcast(reduce))`, but the checks are order-agnostic, so this also matches `subtract(broadcast(reduce), shift_value)`. That's fine (blocking the reversed form is just conservative), but the comment should mention it matches either operand order.

xla/service/cpu/cpu_instruction_fusion_test.cc
```diff
@@ -1140,5 +1140,240 @@ TEST_F(InstructionFusionTest, SkipReduceComputationsIfFusionEmitters) {
...
+  %p0 = f32[] parameter(0)
+  %p1 = f32[] parameter(1)
+  ROOT %m = f32[] maximum(%p0, %p1)
+}
+
+ENTRY main {
+  %lhs = f32[5,5] parameter(0)
+  %rhs = f32[5,5] parameter(1)
+  %dot = f32[5,5] dot(%lhs, %rhs), lhs_contracting_dims={1}, rhs_contracting_dims={0}
+  %scale = f32[] constant(0.44721359)
```
seantalts, line 1244: Sorry, my earlier wording ("identical recompute-divergence risk") was wrong for add; the known mechanism is FMA contraction of multiply+subtract, which has no add equivalent. Suggested comment:
```
// Conservatively treat any elementwise shift value as coupled. Multiply is
// the case with a known divergence mechanism (FMA contraction); other
// elementwise ops are included defensively.
```

seantalts (commented):
Thanks for the quick turnaround on all three points. This looks close to ready. Remaining items are mostly documentation:

1. **The PR description is stale** (and it usually becomes the commit message on import). It still says the fix is in `ShouldFuse` (it's now the `ComputeGloballyUnfusible` override), names a test that no longer exists, and says end-to-end verification wasn't done (it has been). Please update it.
2. The override needs a comment explaining why it exists (inline).
3. Correcting my own earlier suggestion about add(dot, bias) (inline).

xla/service/cpu/cpu_instruction_fusion.cc
Liauuu: Done. Added the suggested comment. Appreciate the explanation.

xla/service/cpu/cpu_instruction_fusion.cc
Liauuu: Done. Clarified that it matches either operand order.

xla/service/cpu/cpu_instruction_fusion_test.cc
Liauuu: Done. Updated the comment as suggested. Appreciate all the feedback.

Liauuu:
> Thanks for the quick turnaround on all three points. This looks close to ready. Remaining items are mostly documentation:
> 
> 1. **The PR description is stale** (and it usually becomes the commit message on import). It still says the fix is in `ShouldFuse` (it's now the `ComputeGloballyUnfusible` override), names a test that no longer exists, and says end-to-end verification wasn't done (it has been). Please update it.
> 2. The override needs a comment explaining why it exists (inline).
> 3. Correcting my own earlier suggestion about add(dot, bias) (inline).

Done. Addressed all of the remaining comments and updated the PR description. Thanks for taking the time to review this.

seantalts (approved):
Approving to run benchmarks internally

Liauuu:
> Approving to run benchmarks internally

Grateful for all the guidance during the review. Thanks for approving.

Liauuu:
I noticed a few remaining clang-tidy issues from the latest CI, so I fixed them and rebased onto the latest upstream main. Thanks again for taking another look.

seantalts (approved):

seantalts:
Benchmarks look good, we're trying to merge, but heads up our CI internally is a mess rn.

Liauuu:
Thank you for the detailed reviews and guidance. I'm still learning and truly appreciate your feedback . I’ve learned a lot through this process.

PR #44763, opened 2026-06-24
[XLA:GPU][ROCm] Fix rocm_clang_local host toolchain under Bazel 8

nurmukhametov:
Under Bazel 8, rocm_clang_local's host and target C++ compiles fall back to the autodetected gcc, which rejects the Clang-only warning flags from --config=warnings (-Wself-assign, -Wthread-safety-analysis, etc.).

Register the ROCm crosstool via --extra_toolchains so toolchain resolution routes C++ compiles through the ROCm wrapper (clang via CLANG_COMPILER_PATH) while -x rocm still goes to hipcc. Drop the --crosstool_top, which is a no-op now.


nurmukhametov:
could I get a review on this, please?

pifon2a (approved):

PR #44796, opened 2026-06-25
SPMD: Use collective-broadcast for replicated dynamic-slice from sharded operands

shawnwang18:
## Summary

This PR adds an SPMD lowering for a dynamic-slice pattern that appears in the Layered-FSDP forward pass for Muon workloads.

When a replicated result dynamically slices a single layer from an operand sharded along the layer dimension, XLA previously materialized the full operand with an all-gather and then sliced out the requested layer. This PR lowers that pattern to a per-owner `collective-broadcast`: each partition slices its local candidate, a conditional selects the partition that owns the requested layer, and that owner broadcasts only the selected slice.

## Workload

The motivating workload is Layered-FSDP for Muon.

Muon's Newton-Schulz update needs the full weight/gradient matrix for each layer. To preserve full matrices locally, Layered-FSDP shards scanned layer stacks along the layer dimension instead of sharding the matrix row/column dimensions.

A representative weight layout is:

```text
logical shape: [L, K, N]
sharding:      P("lp", None, None)
```

where `L` is the scanned layer dimension and `lp` is the layer-parallel mesh axis.

During the forward pass, the scan body selects the current layer:

```text
W_i = dynamic-slice(W, i, 0, 0), dynamic_slice_sizes={1, K, N}
```

The selected `W_i` is needed replicated across the LP partitions for the layer computation. If `L == LP`, layer `i` is owned by LP partition `i`.

The ideal communication pattern for scan iteration `i` is therefore:

```text
owner = partition that owns layer i
broadcast W_i from owner to all LP partitions
```

## Original Issue

Before this PR, XLA SPMD did not have a specialized lowering for this case:

```text
replicated result = dynamic-slice(tiled operand, dynamic layer index)
```

Because the operand is sharded on the sliced dimension but the dynamic-slice result is replicated, the existing path falls back to replicating the operand first. In practice this produces:

```text
all-gather full W across LP
dynamic-slice W_i from the gathered W
```

That is correct, but inefficient for this workload.

For example, if the scan only needs one layer's weight `[1, K, N]`, the old lowering gathers the full layer stack `[L, K, N]` and then slices out one layer. This defeats the purpose of Layered-FSDP's layerwise ownership for the forward pass, because every scan iteration communicates all layer shards instead of the one layer needed by that iteration.

## Solution In This PR

This PR adds `SpmdPartitioningVisitor::TryDynamicSliceWithCollectiveBroadcast` and calls it from `HandleDynamicSlice` before the generic replicated-output fallback.

The new lowering matches a conservative pattern:

```text
- dynamic-slice output is replicated
- operand is tiled
- exactly one dimension is dynamically sliced
- sliced size is 1
- operand is sharded only along the sliced dimension
- no partial replication
- partitioning is even
- slice index is an S32 scalar
- operand sharding uses the same number of devices as the SPMD partition count
```

For the matched case, XLA lowers:

```text
dynamic-slice(tiled_operand, i) -> replicated result
```

to:

```text
clamped_i   = clamp(i, 0, dim_size - 1)
owner       = clamped_i / shard_size
local_i     = clamped_i - owner * shard_size
local_slice = dynamic-slice(local_operand, local_i)

result = conditional(owner)
  branch 0: collective-broadcast(local_slice), root=partition owning shard 0
  branch 1: collective-broadcast(local_slice), root=partition owning shard 1
  ...
```

Each conditional branch contains a `collective-broadcast` with a static root. The dynamic part is only the branch selection, which is derived from the replicated layer index. This lets the HLO represent the desired "changing owner per scan iteration" pattern without requiring a dynamic-root collective.

The generated communication now moves only the selected layer slice instead of all layer shards.

## Before / After

Before:

```text
all-gather W_local -> W_full
dynamic-slice W_full[i]
```

After:

```text
local dynamic-slice W_local[local_i]
conditional(owner)
  collective-broadcast from owner partition
```

So the forward pass communication changes from:

```text
communicate [L, K, N] per scan iteration
```

to:

```text
communicate [1, K, N] per scan iteration
```

for the matched Layered-FSDP pattern.

## Testing

Added a focused SPMD partitioner test:

```text
DynamicSliceReplicatedResultUsesCollectiveBroadcast
```

The test verifies that a replicated dynamic-slice result from a layer-sharded operand produces:

```text
- no all-gather
- one conditional with one branch per owner partition
- collective-broadcast in each branch
- replica groups whose first member is the branch's broadcast root
```

Validation performed for the motivating workload:

```text
bazel test --test_output=errors \
  --test_filter="DynamicSliceReplicatedResultUsesCollectiveBroadcast" \
  //xla/service/spmd:spmd_partitioner_test
```

Additional workload validation:

```text
- Rebuilt JAX/JAXLIB CUDA artifacts against this XLA checkout.
- Ran lp_fsdp_muon_test.py --mode lp --no-run --dump-hlo with
  JAX_ENABLE_COMPILATION_CACHE=false.
- Confirmed the LP forward HLO contains the expected conditional
  collective-broadcast lowering.
```

Observed HLO counts for the workload:

```text
collective-broadcast-start = 32
collective-broadcast-done  = 32
conditional                = 4
dynamic_slice_collective_broadcast_owner = 16
all-gather-start           = 0
all-gather-done            = 0
```

🚀 Kind of Contribution
🐛 Bug Fix, ⚡️ Performance Improvement, 🧪 Tests

No public HLO benchmark result is included. The validation target was the layer-FSDP forward HLO from the muon_jax repro. With a fresh JAX compile, the LP forward HLO has 0 all-gather-start/all-gather-done ops and contains the expected conditional collective-broadcast lowering.

🧪 Unit Tests:
- bazel test --test_output=errors --test_filter="*DynamicSliceReplicatedResultUsesCollectiveBroadcast*" //xla/service/spmd:spmd_partitioner_test

🧪 Execution Tests:
- Rebuilt JAX/JAXLIB CUDA artifacts against this XLA checkout.
- Ran lp_fsdp_muon_test.py --mode lp --no-run --dump-hlo with JAX_ENABLE_COMPILATION_CACHE=false.
- Confirmed HLO counts: collective-broadcast-start=32, collective-broadcast-done=32, conditional=4, dynamic_slice_collective_broadcast_owner=16, all-gather-start=0, all-gather-done=0.


xla/service/spmd/spmd_partitioner.h
```diff
@@ -107,6 +107,15 @@ struct SpmdPartitionerOptions {
   // Enables windowed einsum for result reduce-scatter.
   bool enable_windowed_einsum_for_reduce_scatter = true;
 
+  // Enables a narrow dynamic-slice lowering that broadcasts a single slice
+  // from its sharded owner instead of all-gathering the full operand first.
+  bool enable_dynamic_slice_collective_broadcast = true;
```
bixia1: I am not very sure about this, as we haven't tested TPU. But we can leave it like this and see.

xla/service/spmd/spmd_partitioner.cc
```diff
@@ -3837,7 +3837,212 @@ absl::Status SpmdPartitioningVisitor::HandleConstant(HloInstruction* hlo) {
   return absl::OkStatus();
 }
 
+// Rewrites a replicated dynamic-slice from an operand that is tiled only along
+// the sliced dimension into a single-slice collective broadcast.
```
bixia1: s/the sliced dimension/a single slicing dimension/

xla/service/spmd/spmd_partitioner.cc
```diff
@@ -3837,7 +3837,212 @@ absl::Status SpmdPartitioningVisitor::HandleConstant(HloInstruction* hlo) {
...
+//     branch 1: collective-broadcast(local_slice),  // source is tile 1
+//     ...)
+//
+// Each branch's replica group lists the owning partition first. Collective
+// broadcast therefore forwards only that partition's local one-element slice to
+// every partition, yielding the same replicated result without all-gathering
+// the full operand.
+//
+// Keep the guard conditions conservative. The rewrite relies on one-to-one
+// ownership between tiles of the sliced dimension and partitions: no tuple
```
bixia1: s/sliced/slicing/

xla/service/spmd/spmd_partitioner.cc
```diff
@@ -3837,7 +3837,212 @@ absl::Status SpmdPartitioningVisitor::HandleConstant(HloInstruction* hlo) {
...
+// shapes, one sliced dimension of size 1, scalar S32 index, no partial
+// replication, no sharding on non-sliced dimensions, an evenly divisible
```
bixia1: s/non-sliced/non-slicing/

xla/service/spmd/spmd_partitioner.cc
```diff
@@ -3837,7 +3837,212 @@ absl::Status SpmdPartitioningVisitor::HandleConstant(HloInstruction* hlo) {
...
+
+  std::vector<int64_t> owner_partition(num_slice_partitions, -1);
+  bool valid_tile_assignment = true;
+  input_sharding.EachTile(
+      [&](absl::Span<const int64_t> indices, int64_t device) {
+        if (indices.size() <= *sliced_dim || device < 0 ||
+            device >= num_partitions_) {
+          valid_tile_assignment = false;
+          return;
+        }
```
bixia1, line 4561: This check whether the sharding is invalid? I wonder whether previous passes can guarantee this won't happen, in which case, maybe we can return a error status instead?

xla/service/spmd/spmd_partitioner.cc
```diff
@@ -3837,7 +3837,212 @@ absl::Status SpmdPartitioningVisitor::HandleConstant(HloInstruction* hlo) {
...
+        for (int64_t dim = 0; dim < rank; ++dim) {
+          if (dim != *sliced_dim && indices[dim] != 0) {
+            valid_tile_assignment = false;
+            return;
+          }
+        }
+        const int64_t owner = indices[*sliced_dim];
```
bixia1: s/owner/slice_index/ ?

xla/service/spmd/spmd_partitioner.cc
bixia1: kCollectiveBroadcast supports multiple replica, can you expand the comment here to briefly explain why the transformation here don't support "partial replication"

bixia1 (approved):
Nice work!

We are working on Shardy partitioner and will eventually retire SPMD partitioner. So we will need to port this work to Shardy at some point.

xla/service/spmd/spmd_partitioner.cc
shawnwang18: Fixed.

xla/service/spmd/spmd_partitioner.cc
shawnwang18: Fixed.

xla/service/spmd/spmd_partitioner.cc
shawnwang18: Done

xla/service/spmd/spmd_partitioner.cc
shawnwang18: done

xla/service/spmd/spmd_partitioner.cc
shawnwang18: Done

xla/service/spmd/spmd_partitioner.cc
shawnwang18: Done

akuegel (approved):

akuegel:
@shawnwang18 This was rolled back because the default was true, and disabled on CPU. This meant it was also enabled on TPU. Can we enable it on GPU, and make the default false?

shawnwang18:
> @shawnwang18 This was rolled back because the default was true, and disabled on CPU. This meant it was also enabled on TPU. Can we enable it on GPU, and make the default false?

Thanks, will do.

akuegel:
Can you please rebase? Also please add flag parsing logic to debug_options_flags.cc

akuegel (approved):

bixia1 (approved):

bixia1 (approved):

PR #44813, opened 2026-06-25
[Bazel] Add explicit loads of external rule symbols

cerisier:
This is a follow up to #42789 

Final batch of missing loads across XLA packages and inside existing patches for 3rdparty rulesets. 

Some loads were commented which I understand might come from google internal requirements.
Not sure if these will require special copybara treatment for internal checks to pass.

After this, only upstream dependencies such as shardy or llvm are left to fix upstream.


cerisier:
cc @beckerhe @akuegel

akuegel (approved):

cerisier:
Pretty sure failure comes from the uncommented loads :( Similar to https://github.com/openxla/shardy/pull/1334 it seems?

akuegel:
> Pretty sure failure comes from the uncommented loads :( Similar to [openxla/shardy#1334](https://github.com/openxla/shardy/pull/1334) it seems?

No, in this case the error is different. It is similar to https://github.com/openxla/xla/pull/44823. Rebase should help to avoid this error.

akuegel (approved):

third_party/triton/oss_only/build_files.patch
```diff
@@ -9,8 +9,8 @@ new file mode 100644
 +# This package imports OpenAI's Triton (https://github.com/triton-lang/triton).
 +
```
akuegel, line 10: Changes in this file are sort of tricky and would require additional work on our side. In the interest of getting the other changes merged sooner, could you please revert the changes in this file and create a separate PR for it?

akuegel (approved):

third_party/triton/oss_only/build_files.patch
cerisier: Done !

PR #44823, opened 2026-06-25
[Bazel] Make XLA CPU selects platform-based with legacy flag mappings

cerisier:
This PR makes legacy non platforms apis not be forced down to downstream users.

XLA now has toolchains, platforms and constraints counterparts for all of its --cpu, --crosstool_top, --compiler usage.

This means that, using `platform_mapping`, we can now make non legacy platforms config settings and selects the default, and let XLA continue to use legacy flags while still being compatible (not sure if internal google usage forces keeping those but they could eventually go away too).

This is mostly an attempt to see potential breakage downstream and will likely require some work on tensorflow and JAX before even knowing if internal checks accept this.

This would be a good quality of life improvements for downstream users using legacy flags in selects for instance can create subtle bugs quite hard to diagnose, especially with selects and `//conditions:defaults` being evaluated if downstream users don't have platform_mappings themselves.


cerisier:
cc @beckerhe @akuegel

akuegel (approved):
Approving so that we can see whether internal presubmits will pass.

cerisier:
Thanks for trying ! One shotting it would have been too good ! Curious what the failure is.

akuegel:
> Thanks for trying ! One shotting it would have been too good ! Curious what the failure is.

I think you need to rebase. The error is that one of the transformations is a no-op, but this is just because your baseline does not yet have the change where the transformation will do something.

akuegel (approved):

PR #44829, opened 2026-06-25
[xla:gpu] Do not sink constants that are not part of offset expr

ezhulenev:
Do not sink arbitrary constants into DUS fusions as they must be handled by `ThunkEmitter`, sink only scalar constants that are part of offset expression and captured by the expression itself and do not need a buffer at run time.


ermilovmaxim (approved):

terryysun (approved):
Thanks for the fix!

akuegel (approved):

PR #44867, opened 2026-06-26
[XLA:GPU] zero initialize the output buffers of cuDNN SDPA to avoid unexpected garbage values

Cjkkkk:
📝 Summary of Changes
Zero initialize the output buffers of cuDNN SDPA so users don't get random values for regions that cuDNN doesn't touch.  This could happen when user specify the padded region of the inputs. cuDNN skips computing output for these regions and keep the initial values for padded regions of the output buffers.

🎯 Justification
It is expected that the SDPA gradient output should be 0 if its corresponding input is in the padded regions. Zero initialize the output buffers so this constraints hold.

🚀 Kind of Contribution


derdrdirk (approved):
LGTM

PR #44868, opened 2026-06-26
[Bazel] Load rules_cc in Triton BUILD patch

cerisier:
Split out from #44813 per reviewer request.

This keeps the Triton BUILD patch rule-load changes separate from the straightforward third-party BUILD load additions, since third_party/triton/oss_only/build_files.patch needs additional handling on the XLA side.

### DIFF
--- a/third_party/triton/oss_only/build_files.patch
+++ b/third_party/triton/oss_only/build_files.patch
@@ -9,8 +9,8 @@ new file mode 100644
 +# This package imports OpenAI's Triton (https://github.com/triton-lang/triton).
 +
 +# copybara:uncomment load("@rules:copybara.bzl", "copybara_config_test")
-+# load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
-+# load("@rules_cc//cc:cc_library.bzl", "cc_library")
++load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
++load("@rules_cc//cc:cc_library.bzl", "cc_library")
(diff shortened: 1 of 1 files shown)


vwbaker:
Yeah this will need to be done internally; I'll prepare the change and post the corresponding PR here.

vwbaker:
Now that I'm looking at it, I'm curious, why is this necessary? We already build Triton in OSS so it doesn't actually need the cc_library / cc_binary loads - these are build into the bazel core. Have you tested whether this is needed?

cerisier:
I'm not sure I understand, bazel 9 requires explicit load as all these are now moved to starlark modules like `rules_cc`. Bazel 8 accepted those with `--incompatible_autoload_externally` set to a bunch of legacy autoloaded rules, but bazel 9 changed that.

vwbaker:
I was just curious why it's building at all (the CI is green) if this is necessary haha. So it sounds like our CI is building on bazel 8 and you're trying to build on bazel 9? I think the change is a good one if this is the case, I'm just. checking. Here's the new PR: https://github.com/openxla/xla/pull/44889. Probably won't get in until Monday as most of us are in Europe.

cerisier:
Sorry I should have mentioned this is for Bazel 9 !

vwbaker:
All good; let me know if the other PR is sufficient or if it's missing something.

akuegel:
@cerisier https://github.com/openxla/xla/pull/44889 got merged, I guess this PR can be closed?

cerisier:
Wonderful thanks !

PR #44916, opened 2026-06-27
[XLA:GPU][codegen] Extend SPIR-V math op lowering to support 16-bit floating point precisions

akhilgoe:
This PR adds 16-bit floating point support to SPIR-V math op lowerings via builtin intrinsics. It fixes the legalization failure errors that many tests like `xla/hlo/builder/lib/math_test.cc` throw for F16 precisions.


dimitar-asenov (approved):

akhilgoe (commented):
@xla-rotation, can you please help merge this PR? Thanks!

dimitar-asenov (approved):

PR #44984, opened 2026-06-29
[XLA:GPU] Fix device association for reordered GPU communicator splits

shawnwang18:
📝 Summary of Changes

Keep the parent communicator, split key, and device rank for each GPU in one
child-rank-ordered sequence when splitting a GPU clique.

Add a two-GPU regression that splits parent clique `[D0, D1]` into reordered
child clique `[D1, D0]`. The test verifies that the split path was used and
that each child communicator reports both the expected rank and the expected
physical `StreamExecutor`.

### Root cause

`SplitCommunicatorsWithCancel` consumes `parent_comms`, `keys`, and `ranks` as
parallel arrays. Before this change, XLA constructed them using two independent
orders:

- `parent_comms` and `keys` were emitted from a map sorted by parent rank.
- `ranks`, which determines the `StreamExecutor`, was sorted by child rank.

For a parent `[D0, D1]` and child `[D1, D0]`, the old inputs were:

| Index | Parent communicator | Split key | Device rank / executor | Result                       |
|------:|---------------------|----------:|------------------------|------------------------------|
|     0 | parent rank 0 / D0  |         1 | child rank 0 / D1      | D0 communicator paired to D1 |
|     1 | parent rank 1 / D1  |         0 | child rank 1 / D0      | D1 communicator paired to D0 |

`ncclCommSplit` still received each valid parent communicator and split key, so
the underlying NCCL child ranks were correct. XLA also stored each returned
communicator under its split key. This made communicator creation appear
successful.

The wrapper was nevertheless constructed with `ranks[i]`'s executor. For a
rank permutation, the resulting `NcclCommunicator` therefore held the NCCL
communicator for one physical GPU and the `StreamExecutor` for another.

Automatic BFC memory registration uses that wrapper executor to find the GPU's
allocator range. It consequently registered D0's arena with D1's communicator
and D1's arena with D0's communicator. In a larger reordered group, unaffected
ranks could use NCCL's registered-buffer path while affected ranks used the
ordinary path. NCCL requires registered-buffer participation to be consistent
across ranks; mixing the paths is undefined and manifested as a collective
hang followed by XLA's rendezvous timeout. See the
[NCCL user-buffer registration requirements](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html).

This change sorts the complete parent-rank/device-rank pairs by child rank and
populates all three arrays in one loop:

| Index | Parent communicator | Split key | Device rank / executor | Result                 |
|------:|---------------------|----------:|------------------------|------------------------|
|     0 | parent rank 1 / D1  |         0 | child rank 0 / D1      | D1 remains paired to D1 |
|     1 | parent rank 0 / D0  |         1 | child rank 1 / D0      | D0 remains paired to D0 |

Child-rank ordering also preserves Loopback's contract that returned
communicator index `i` represents child rank `i`. This does not change
`split_share`; its existing default and behavior remain intact.

🎯 Justification

Any workload that builds communicators over the same devices in different rank
orders can otherwise bind a communicator to the wrong GPU executor. The issue
became visible in the dynamic-slice collective-broadcast workload, where the
broadcast owner changes and reordered communicator groups are created inside a
loop.

Keeping each physical participant's communicator, key, rank, and executor
aligned prevents wrong-device memory registration and the resulting NCCL
deadlock.

🚀 Kind of Contribution

🐛 Bug Fix, 🧪 Tests

Not applicable. This is a correctness fix and does not change collective
algorithms or data movement.

🧪 Unit Tests:

- PASS: `bazel test //xla/backends/gpu/collectives:gpu_cliques_test --test_filter=GpuCliquesTest.SplitCliquesKeepsReorderedRanksOnCorrectExecutors`
- PASS: `bazel test //xla/backends/gpu/collectives:gpu_cliques_test '--test_filter=GpuCliquesTest.Split*'`
- Negative control: with the old ordering temporarily restored, the new test
  failed on both ranks with swapped executors and `0 -> 1`, `1 -> 0` reported
  ranks. Restoring this fix made the same test pass.
- The unfiltered target was also attempted, but the unchanged
  `DifferentCollectivesProduceDifferentCliques` test blocked in its rendezvous
  after all participants arrived. All split-specific tests pass.

🧪 Execution Tests:

- The committed regression uses exactly two GPUs and executes parent-to-child
  clique splitting with reordered ranks through the Loopback GPU collectives
  implementation.
- As additional diagnostic validation, the original four-GPU Muon reproducer
  completed a training step with this ordering fix and source-built NCCL
  2.30.7 under `NCCL_CHECK_MODE=DEBUG_LOCAL`; no wrong-device registration
  diagnostics were reported. This diagnostic workload is not added as a test.

### DIFF
--- a/xla/backends/gpu/collectives/BUILD
+++ b/xla/backends/gpu/collectives/BUILD
@@ -245,6 +245,7 @@ xla_cc_test(
         ":gpu_clique_key",
         ":gpu_cliques",
         ":gpu_collectives",
+        ":gpu_communicator",
         ":loopback_collectives",
         ":nccl_or_rccl_collectives",  # fixdeps: keep
         "//xla:executable_run_options",
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -623,11 +623,29 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
       absl::StrAppend(str, mapping.first.value(), "->", mapping.second.value());
     };
 
-    // Collect parent communicators we'll be splitting from and keys for
-    // creating new communicators.
+    // SplitCommunicators treats parent_comms, keys, and ranks as parallel
+    // arrays. Keep all three in child-rank order so that each parent
+    // communicator stays paired with the device and rank of the same physical
+    // participant. Ordering parent_comms by parent rank while independently
+    // ordering ranks by child rank can bind a split communicator to the wrong
+    // StreamExecutor when the child clique permutes ranks.
+    std::vector<const RankPair*> sorted_rank_pairs(rank_pairs.begin(),
+                                                   rank_pairs.end());
+    absl::c_sort(sorted_rank_pairs, [](const RankPair* a, const RankPair* b) {
+      return a->second.rank < b->second.rank;
+    });
+
+    // Collect parent communicators we'll be splitting from, keys for creating
+    // new communicators, and device ranks in their common child-rank order.
     std::vector<Communicator*> parent_comms;
     std::vector<RankId> keys;
-    for (auto& [parent_rank, split_rank] : rank_mapping) {
+    std::vector<DeviceRank> ranks;
+    parent_comms.reserve(sorted_rank_pairs.size());
+    keys.reserve(sorted_rank_pairs.size());
+    ranks.reserve(sorted_rank_pairs.size());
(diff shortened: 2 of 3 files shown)


shawnwang18:
Superseded by #45000 after renaming the fork head branch to `shawnw/fix-gpu-clique-split-ordering`. GitHub automatically closed this cross-fork PR during the branch rename.

PR #44992, opened 2026-06-29
[ROCm] Serialize rocm_smi access during parallel device init

nurmukhametov:
XLA initializes devices concurrently (PlatformUtil::GetStreamExecutors), so GetRocmPcieBandwidth and GetRocmXgmiTopology issue rsmi_* calls in parallel. rocm_smi only guards state with a per-device mutex, but its gpu_metrics object is a process-global singleton, so concurrent queries on different devices corrupt its shared std::map and SEGV. Hold a global mutex across each rsmi call sequence.


derdrdirk (approved):
LGTM

derdrdirk:
DWYU is failing https://github.com/openxla/xla/actions/runs/28367944608/job/84038349881?pr=44992

nurmukhametov:
> DWYU is failing https://github.com/openxla/xla/actions/runs/28367944608/job/84038349881?pr=44992

I looks unrelated to this PR but I have fixed it.

akuegel (approved):

PR #44993, opened 2026-06-29
[ROCm] Package Tensile fallback libs for rocblas

nurmukhametov:
The per-arch rocblas glob added in PR #43603 (`<arch>`) misses the arch-independent *_fallback.dat Tensile solution libraries, causing a null-deref SEGV in loadPlaceholderLibrary() when a problem falls back (e.g. bf16 grouped conv on gfx90a). Add "*fallback.dat" to the glob.

### DIFF
--- a/third_party/gpus/rocm/BUILD.tpl
+++ b/third_party/gpus/rocm/BUILD.tpl
@@ -222,6 +222,7 @@ rocm_lib_import(
     name = "rocblas",
     data = glob([
         "%{rocm_root}/lib/librocblas.so*",
+        "%{rocm_root}/lib/rocblas/library/*fallback.dat",
     ]) + glob([
         pattern
         for arch in rocm_gpu_architectures()


third_party/gpus/rocm/BUILD.tpl
```diff
@@ -222,6 +222,7 @@ rocm_lib_import(
     name = "rocblas",
     data = glob([
         "%{rocm_root}/lib/librocblas.so*",
+        "%{rocm_root}/lib/rocblas/library/*fallback.dat",
```
draganmladjenovic, line 225: Can you check where these reside on multirach 7.14 ?

nurmukhametov: > Can you check where these reside on multirach 7.14 ?

It has the following layout:

| Location | Architectures | Fallback naming |
|---|---|---|
| Subdirs `library/<arch>/` | gfx900, gfx906, gfx908, gfx90a, gfx942, gfx950, gfx1150–1153 | `..._fallback_<arch>.dat` |
| Top-level `library/` (flat) | gfx1010–1012, gfx1030–1036, gfx1100–1103, gfx1200–1201 | `..._fallback_<arch>.dat` |

Both cases seem to be covered by these two rules:

```bash
            "%{rocm_root}/lib/rocblas/library/*" + arch + "*",
            "%{rocm_root}/lib/rocblas/library/" + arch + "/**/*",
```

draganmladjenovic (approved):

nurmukhametov:
@xla-rotation, could I get a review on this please?

dimitar-asenov (approved):

PR #45000, opened 2026-06-29
[XLA:GPU] Fix device association for reordered GPU communicator splits

shawnwang18:
📝 Summary of Changes

Fix GPU clique splitting when a child clique reorders the parent ranks. XLA now emits each parent communicator, split key, and device/executor together in child-rank order. Add a two-GPU regression for `[D0, D1] → [D1, D0]`.

🎯 Justification

The old code built parallel arrays in different orders, which could pair an NCCL communicator with the wrong GPU `StreamExecutor`. Wrong-device memory registration could make ranks use inconsistent NCCL paths and hang collectives. This fix preserves the communicator/device association.

### Example

For parent clique `[D0, D1]` and reordered child clique `[D1, D0]`, the old code built:

```text
parent communicator: [D0, D1]
split key:           [ 1,  0]
executor:            [D1, D0]
```

The arrays are consumed by index, so the D0 communicator was paired with the D1 executor, and the D1 communicator with the D0 executor. The communicator split could succeed, but later buffer registration used the wrong physical GPU.

The fix builds all fields in child-rank order:

```text
parent communicator: [D1, D0]
split key:           [ 0,  1]
executor:            [D1, D0]
```

Each communicator now remains paired with its matching device and executor.

🚀 Kind of Contribution

🐛 Bug Fix, 🧪 Tests

Not applicable; this is a correctness fix.

🧪 Unit Tests:

- PASS: `bazel test //xla/backends/gpu/collectives:gpu_cliques_test --test_filter=GpuCliquesTest.SplitCliquesKeepsReorderedRanksOnCorrectExecutors`
- PASS: `bazel test //xla/backends/gpu/collectives:gpu_cliques_test --test_filter=GpuCliquesTest.Split*`

🧪 Execution Tests:

The two-GPU Loopback regression exercises reordered clique splitting. The original four-GPU Muon reproducer also completed one training step with source-built NCCL 2.30.7.

### DIFF
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -63,6 +63,7 @@ limitations under the License.
 #include "xla/stream_executor/stream_executor.h"
 #include "xla/tsl/platform/env.h"
 #include "xla/tsl/platform/logging.h"
+#include "xla/tsl/util/sorted_range.h"
 #include "xla/util.h"
 #include "tsl/platform/casts.h"
 #include "tsl/platform/hash.h"
@@ -624,28 +625,21 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
     };
 
     // SplitCommunicators treats parent_comms, keys, and ranks as parallel
-    // arrays. Keep all three in child-rank order so that each parent
+    // arrays. Collect all three in child-rank order so that each parent
     // communicator stays paired with the device and rank of the same physical
-    // participant. Ordering parent_comms by parent rank while independently
-    // ordering ranks by child rank can bind a split communicator to the wrong
-    // StreamExecutor when the child clique permutes ranks.
-    std::vector<const RankPair*> sorted_rank_pairs(rank_pairs.begin(),
-                                                   rank_pairs.end());
-    absl::c_sort(sorted_rank_pairs, [](const RankPair* a, const RankPair* b) {
-      return a->second.rank < b->second.rank;
-    });
-
-    // Collect parent communicators we'll be splitting from, keys for creating
-    // new communicators, and device ranks in their common child-rank order.
+    // participant. Ordering parent_comms by parent rank while ordering ranks by
+    // child rank can bind a split communicator to the wrong StreamExecutor.
     std::vector<Communicator*> parent_comms;
     std::vector<RankId> keys;
     std::vector<DeviceRank> ranks;
-    parent_comms.reserve(sorted_rank_pairs.size());
-    keys.reserve(sorted_rank_pairs.size());
-    ranks.reserve(sorted_rank_pairs.size());
-    for (const RankPair* rank_pair : sorted_rank_pairs) {
-      RankId parent_rank = rank_pair->first;
-      const DeviceRank& device_rank = rank_pair->second;
+    parent_comms.reserve(rank_pairs.size());
+    keys.reserve(rank_pairs.size());
+    ranks.reserve(rank_pairs.size());
+    for (const RankPair* rank_pair :
+         tsl::SortedRange(rank_pairs, [](const RankPair* a, const RankPair* b) {
+           return a->second.rank < b->second.rank;
+         })) {
+      const auto& [parent_rank, device_rank] = *rank_pair;
       auto parent_comm = (*parent_clique)->comm(parent_rank);
       if (!parent_comm.has_value()) {
         return InvalidArgument(
--- a/xla/backends/gpu/collectives/gpu_cliques.cc
+++ b/xla/backends/gpu/collectives/gpu_cliques.cc
@@ -623,11 +623,29 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
       absl::StrAppend(str, mapping.first.value(), "->", mapping.second.value());
     };
 
-    // Collect parent communicators we'll be splitting from and keys for
-    // creating new communicators.
+    // SplitCommunicators treats parent_comms, keys, and ranks as parallel
+    // arrays. Keep all three in child-rank order so that each parent
+    // communicator stays paired with the device and rank of the same physical
+    // participant. Ordering parent_comms by parent rank while independently
+    // ordering ranks by child rank can bind a split communicator to the wrong
+    // StreamExecutor when the child clique permutes ranks.
+    std::vector<const RankPair*> sorted_rank_pairs(rank_pairs.begin(),
+                                                   rank_pairs.end());
+    absl::c_sort(sorted_rank_pairs, [](const RankPair* a, const RankPair* b) {
+      return a->second.rank < b->second.rank;
+    });
+
+    // Collect parent communicators we'll be splitting from, keys for creating
+    // new communicators, and device ranks in their common child-rank order.
     std::vector<Communicator*> parent_comms;
     std::vector<RankId> keys;
-    for (auto& [parent_rank, split_rank] : rank_mapping) {
+    std::vector<DeviceRank> ranks;
+    parent_comms.reserve(sorted_rank_pairs.size());
+    keys.reserve(sorted_rank_pairs.size());
+    ranks.reserve(sorted_rank_pairs.size());
+    for (const RankPair* rank_pair : sorted_rank_pairs) {
+      RankId parent_rank = rank_pair->first;
+      const DeviceRank& device_rank = rank_pair->second;
       auto parent_comm = (*parent_clique)->comm(parent_rank);
       if (!parent_comm.has_value()) {
         return InvalidArgument(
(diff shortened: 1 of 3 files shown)


xla/backends/gpu/collectives/gpu_cliques_test.cc
```diff
@@ -509,7 +569,7 @@ TEST(GpuCliquesTest, DifferentCollectivesProduceDifferentCliques) {
   // Acquire clique with loopback collectives for the same key.
   ASSERT_OK_AND_ASSIGN(Collectives * loopback_base,
                        CollectivesRegistry::Get("GPU", "loopback"));
-  auto* loopback = absl::down_cast<GpuCollectives*>(loopback_base);
+  auto* loopback = tsl::down_cast<GpuCollectives*>(loopback_base);
```
dimitar-asenov: Please undo this change.

xla/backends/gpu/collectives/gpu_cliques_test.cc
```diff
@@ -247,6 +248,65 @@ TEST(GpuCliquesTest, SplitCliques) {
...
+  }
+
+  tsl::thread::ThreadPool pool(tsl::Env::Default(), "collectives", 2);
+  tsl::Executor& exec = *pool.AsExecutor();
+
+  ASSERT_OK_AND_ASSIGN(std::vector<se::StreamExecutor*> executors,
+                       CreateExecutors(platform, 2));
+  ASSERT_OK_AND_ASSIGN(Collectives * loopback_base,
+                       CollectivesRegistry::Get("GPU", "loopback"));
+  auto* loopback = tsl::down_cast<GpuCollectives*>(loopback_base);
```
dimitar-asenov: Please use `absl::dow_ncast`. In general always prefer `absl` alternatives, if they exist.

xla/backends/gpu/collectives/gpu_cliques_test.cc
```diff
@@ -247,6 +248,65 @@ TEST(GpuCliquesTest, SplitCliques) {
...
+                                    child_group, child_acquired_cliques);
+  ASSERT_OK_AND_ASSIGN(auto child_cliques,
+                       WaitCliques(std::move(child_futures)));
+
+  ASSERT_NE((*child_cliques.front())->parent(), nullptr);
+  EXPECT_EQ((*child_cliques.front())->parent()->key(), parent_key);
+  for (size_t rank = 0; rank < 2; ++rank) {
+    auto comm = (*child_cliques[rank])->comm(RankId(rank));
+    ASSERT_TRUE(comm.has_value());
+    auto* gpu_comm = tsl::down_cast<GpuCommunicator*>(*comm);
```
dimitar-asenov: Please use absl::down_cast

dimitar-asenov (changes requested):

ezhulenev:
How exactly we can end up with this weird-looking clique key? I think XLA will also request devices in a sorted global device id order? And all custom calls are encouraged to do the same. Also I'm not sure if `[d1,d0]` is a subset of [`d0,d1]` from XLA collectives point of view.

```
GpuCliqueKey child_key({kD1, kD0}, 2);
```

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -623,11 +623,29 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
...
+    // new communicators, and device ranks in their common child-rank order.
     std::vector<Communicator*> parent_comms;
     std::vector<RankId> keys;
-    for (auto& [parent_rank, split_rank] : rank_mapping) {
+    std::vector<DeviceRank> ranks;
+    parent_comms.reserve(sorted_rank_pairs.size());
+    keys.reserve(sorted_rank_pairs.size());
+    ranks.reserve(sorted_rank_pairs.size());
+    for (const RankPair* rank_pair : sorted_rank_pairs) {
+      RankId parent_rank = rank_pair->first;
```
ezhulenev: prefer structrutal bindings instead of ->first/->second: `auto [parent_rank, device_rank] = ...`

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -623,11 +623,29 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
...
 
-    // Collect parent communicators we'll be splitting from and keys for
-    // creating new communicators.
+    // SplitCommunicators treats parent_comms, keys, and ranks as parallel
+    // arrays. Keep all three in child-rank order so that each parent
+    // communicator stays paired with the device and rank of the same physical
+    // participant. Ordering parent_comms by parent rank while independently
+    // ordering ranks by child rank can bind a split communicator to the wrong
+    // StreamExecutor when the child clique permutes ranks.
+    std::vector<const RankPair*> sorted_rank_pairs(rank_pairs.begin(),
```
ezhulenev: Check tsl::SortedRange, I think it will be a bit shorter and more idiomatic here

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -636,18 +654,10 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
...
     }
 
-    std::vector<DeviceRank> ranks;
-    ranks.reserve(rank_pairs.size());
-    for (auto& rank_pair : rank_pairs) {
-      ranks.emplace_back(rank_pair->second);
-    }
-
-    // Sort device ranks, mainly to get more readable logs below.
-    absl::c_sort(ranks, [](auto& a, auto& b) { return a.rank < b.rank; });
```
ezhulenev, line 649: Technically device order can be arbitrary, but in XLA collectives I tried to keep them sorted because it's easier to follow, and it's natural with SPMD partitioner and all meshes used in real jax programs. I wonde why anyone might want to reorder them?

shawnwang18: > Technically device order can be arbitrary, but in XLA collectives I tried to keep them sorted because it's easier to follow, and it's natural with SPMD partitioner and all meshes used in real jax programs. I wonde why anyone might want to reorder them?

 Sorted device order is preferable when order has no semantic meaning. A concrete exception is collective-broadcast, where the first device in the replica group is the source.

 For example, suppose x[4,8,16] is sharded across [D0,D1,D2,D3] along dimension 0, so Di owns x[i,:,:]. All devices need the replicated result of:

 dynamic-slice(x, index, 0, 0)  // result shape [1,8,16]

 If the replicated index is 2, only D2 owns the correct slice. The broadcast group must therefore be:

[D2, D0, D1, D3]

This makes D2 communicator rank 0, which XLA uses as the broadcast root. The parent clique may still be sorted as [D0,D1,D2,D3]; the child clique is reordered only to encode the source.

 Sorting the child back to [D0,D1,D2,D3] would make D0 the source and broadcast the wrong slice. Thus the user mesh remains sorted—the SPMD partitioner generates this permutation because broadcast order is semantically significant.

dimitar-asenov (approved):

xla/backends/gpu/collectives/gpu_cliques.cc
```diff
@@ -623,11 +624,22 @@ InitializeGpuClique(GpuCollectives* collectives, se::StreamExecutor* device,
...
+    // child rank can bind a split communicator to the wrong StreamExecutor.
     std::vector<Communicator*> parent_comms;
     std::vector<RankId> keys;
-    for (auto& [parent_rank, split_rank] : rank_mapping) {
+    std::vector<DeviceRank> ranks;
+    parent_comms.reserve(rank_pairs.size());
+    keys.reserve(rank_pairs.size());
+    ranks.reserve(rank_pairs.size());
+    for (const RankPair* rank_pair :
+         tsl::SortedRange(rank_pairs, [](const RankPair* a, const RankPair* b) {
```
ezhulenev: nit: sort-by lambda can be defined separately to keep for loop on a single line

ezhulenev (approved):

xla/backends/gpu/collectives/gpu_cliques.cc
ezhulenev: Makes sense... kind of... why rank 2 can't be the root of the broadcast? https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/colls.html#c.ncclBroadcast supports it.

This fix still makes sense because of correctness, but I'd revisit how broadcast works in XLA, keeping device ids sorted maximizes the chance of cache hit and clique reuse and makes life of humans reading logs a lot simpler.

xla/backends/gpu/collectives/gpu_cliques.cc
ezhulenev: or even as static function

shawnwang18: fixed

dimitar-asenov (approved):

penumallas1261 (approved):

PR #45058, opened 2026-06-30
[XLA:GPU][oneAPI] Add ULP budgets for double precision trigonometric functions

akhilgoe:
This PR adds Intel-GPU specific ULP budgets for double precision sin and cos. The OCL implementation shows up to 2 ULP deviation from the double precision mpmath baseline, so Intel GPU-specific tolerances are needed to avoid test failures.


derdrdirk (approved):
LGTM

PR #45070, opened 2026-06-30
[ROCm] [xla:gpu]VMM command-buffer: skip-remap + copy-into-shadow for launch-bound regimes

phambinhfin:
## Summary
This one followed by #44726 as optimization:

Optimize the ROCm VMM command-buffer VA-remapping path (`ExecuteWithVaRemapping`) so that trace-once capture + replay (`NEVER_UPDATE`) is a net win in launch-bound (tensor-parallel) regimes, where the per-step VA remap previously outweighed the benefit.

The optimization is based on two properties of the per-step remap on ROCm:

- Most command-buffer slots (model weights) keep the same backing physical buffer across steps, yet the existing path unmapped and re-mapped the entire reservation every step — and on ROCm `hipMemUnmap` is the dominant per-step cost.
- A small number of tiny slices (scale/scalar/metric buffers) change address every step and are the worst case for remapping (a full unmap/map/SetAccess round-trip costs milliseconds even for a few bytes).

Two ROCm-gated changes; the CUDA path and the default ROCm behavior are unchanged:

1. **Skip-remap** — cache each slot's source (BFC) address and only remap the slots whose backing buffer actually moved, reusing the in-tree, unit-tested `se::MemoryReservation::ScopedMapping::Remap()`. In steady state (nothing moved) it issues zero driver calls and skips the unmap-event sync. Default ON for ROCm; set `--xla_gpu_command_buffer_vmm_skip_remap=false` to force the legacy path for A/B.
2. **Copy-into-shadow** — keep tiny churning slices out of the contiguous reservation and give each a stable shadow buffer (fixed address baked into the replayed command buffer), refreshed each step with a stream-ordered D2D copy instead of a remap. Opt-in via `--xla_gpu_command_buffer_vmm_copy_threshold_bytes` (bytes; `0` = default, disabled).

Both knobs are standard XLA flags (`xla.proto` / `DebugOptions`), not environment variables; they are read from the module's `DebugOptions` and only take effect on ROCm.

## Performance

Measured with MaxText (synthetic data, identical converged loss across configurations). Numbers are **relative per-step speedups** (steady-state median, after warm-up), **normalized to the default baseline configuration**; absolute throughput is omitted. Each model compares the default baseline config against the VMM-optimized config (trace-once replay + collectives captured + per-step remap minimized). Baseline = `1.00×`; higher is faster.

### 8×MI300X

| model       | baseline (default) | VMM-optimized |
| ----------- | :----------------: | :-----------: |
| llama2-7b   | 1.00×              | 1.09×         |
| llama3-8b   | 1.00×              | 1.16×         |
| llama3.1-8b | 1.00×              | 1.17×         |
| mistral-7b  | 1.00×              | 1.14×         |
| gemma2-9b   | 1.00×              | 1.23×         |
| qwen3-8b    | 1.00×              | 1.13×         |
| llama2-13b  | 1.00×              | 1.37×         |

### 8×MI355

| model       | baseline (default) | VMM-optimized |
| ----------- | :----------------: | :-----------: |
| llama2-7b   | 1.00×              | 1.10×         |
| llama3-8b   | 1.00×              | 1.16×         |
| llama3.1-8b | 1.00×              | 1.17×         |
| mistral-7b  | 1.00×              | 1.37×         |
| gemma2-9b   | 1.00×              | 1.53×         |
| llama2-13b  | 1.00×              | 1.15×         |

A consistent end-to-end per-step improvement across popular dense LLMs on both platforms, largest on the most launch-bound models. In the bandwidth-bound FSDP regime the same configuration is performance-neutral (no regression).

## Tests

- **Unit:** `gpu_executable_buffer_allocator_test` covers the flag-decision helpers — ROCm-only gating for `xla_gpu_command_buffer_vmm_skip_remap`, and the `xla_gpu_command_buffer_vmm_copy_threshold_bytes` threshold (ROCm-only, negatives clamped to 0, non-ROCm always 0).
- **End-to-end:** A/B with `--xla_gpu_command_buffer_vmm_skip_remap=true/false` and `--xla_gpu_command_buffer_vmm_copy_threshold_bytes=0/65536` on llama2 / llama3; converged loss matches the non-VMM baseline in every run.


phambinhfin:
@dimitar-asenov  Thanks for the reviewing and support for ROCm VMM feature. We add the first step optimization about selecting dynamic and big buffer sizes for mapping only and it should be target on ROCm.  Could you take a look?

xla/service/gpu/gpu_executable_buffer_allocator.cc
```diff
@@ -72,6 +115,48 @@ uint64_t RoundUpToGranularity(uint64_t size, uint64_t granularity) {
...
+// reset()+MapTo() behavior unchanged. Set XLA_VMM_SKIP_REMAP=0 to force the
+// legacy path on ROCm for A/B measurement.
+bool VmmRemapSkipEnabled(absl::string_view platform_name) {
+  if (platform_name != "ROCM") {
+    return false;
+  }
+  // Cache the (ROCm) decision once; the env var is read a single time per
+  // process. Pure parsing lives in vmm_internal for unit testing.
+  static const bool enabled = vmm_internal::ParseVmmRemapSkipEnabled(
+      "ROCM", std::getenv("XLA_VMM_SKIP_REMAP"));
```
dimitar-asenov: Here you're directly reading an environmental variable. Please don't do that. Instead you should go through the usual XLA flags in xla.proto and debug options. Please adjust the rest of the PR accordingly.

dimitar-asenov (changes requested):

xla/service/gpu/gpu_executable_buffer_allocator.cc
phambinhfin: Sure , i will fix it now

phambinhfin: @dimitar-asenov I fixed the comment, please take a look, thanks !

xla/xla.proto
```diff
@@ -595,6 +595,20 @@ message DebugOptions {
   // See CommandBufferUpdateMode for details.
   optional CommandBufferUpdateMode xla_gpu_command_buffer_update_mode = 469;
 
+  // ROCm-only. In the VMM command-buffer VA-remapping path, skip remapping the
+  // slots whose backing buffer did not move since the previous step (only the
+  // slots that actually changed are remapped). Has no effect on non-ROCm
+  // platforms. Default true (ROCm); set false to force the legacy full-remap
+  // path.
+  optional bool xla_gpu_command_buffer_vmm_skip_remap = 505;
```
dimitar-asenov: What is the intended life-time of these flags? Do we want to eventually phase them out and only have the default behavior? If so, we should add `_experimental_`  in the prefix.

xla/service/gpu/gpu_executable_buffer_allocator.cc
dimitar-asenov, line 353: After the changes below, this function will be 400 lines and I'm getting totally lost in it. Please break it down into at least 4-5 sub-parts.

dimitar-asenov (changes requested):

xla/service/gpu/gpu_executable_buffer_allocator.cc
```diff
@@ -724,6 +464,285 @@ GpuExecutableBufferAllocator::ExecutionScope::ExecuteWithVaRemapping(
...
+    if (!plan.mapping_descriptors.empty()) {
+      ASSIGN_OR_RETURN(se::MemoryReservation::ScopedMapping scoped_mapping,
+                       va_ranges_->va_reservation->MapTo(
+                           absl::MakeSpan(plan.mapping_descriptors)));
+      va_ranges_->scoped_mapping = std::move(scoped_mapping);
+    }
+  }
+
+  va_ranges_->last_mapped_src_addrs = std::move(plan.new_src_addrs);
+#ifndef NDEBUG
```
dimitar-asenov: Is it necessary to guard this? What is the overhead of always doing this? Is it costly to compute? Or does it use too much RAM? Do you have any experiments with/without this guard?

It would be best to avoid macros, but that depends on how large the overhead is. What can we expect here in a typical use-case?

phambinhfin: Thanks for the review!  that block only exists for debug purposes (it feeds the DCHECK, which is compiled out in release anyway), so guarding it with #ifndef NDEBUG wasn't worth the macro. I've removed it and now always populate it — the cost is negligible. Thanks for noting!

dimitar-asenov (approved):

dimitar-asenov (approved):

shawnwang18:
@dimitar-asenov @phambinhfin 
This PR makes a huge regression for va-remap command buffer run!

The skip_remap mode introduced in this PR is already automatically handled by DeviceAddressVmmAllocator, for example, for the temp buffer. 

Original code is allocated with 

```
DeviceAddressVmmAllocator->Allocate(tnt device_ordinal, uint64_t allocation_size, bool retry_on_failure,int64_t memory_space, MemoryReservation* reservation, uint64_t reservation_offset, uint64_t mapping_size,
      bool return_reservation_addres)  
```

In the Allocate API, you can together specify the va-remap va range, in the following run, when you deallocate the allocate, the va-remapping is not released, it is put in a deferred unmap queue. and future Allocate with va-remap will automatically reuse the old maping. 

All the va-remap skip should automatically handled by the DeviceAddressVmmAllocator in the original code. 

This PR is manually doing this with old APIs, and the perf benefits mentioned in this PR is vs old code when DeviceAddressVmmAllocator does not support auto va-remap reuse. 

**This PR is not Rocm only**, it regressed all the va-remapping code path. 

The Copy-into-shadow  is another totally different optimization, where we can apply in original code. 

I strongly suggest to revert this PR,  the original code should work for rocm, any fix should go to the allocator. gpu_exectable_buffer_allocator should be backend agonostic

shawnwang18:
The DeviceAddressVmmAllocator's optimization was just in last week, I think @phambinhfin 's perf benefits is vs old implementation, skip-remap should work out-of-box with current DeviceAddressVmmAllocator's

shawnwang18:
I created a revert PR for this https://github.com/openxla/xla/pull/45405, let's confirm whether the original code works for ROCM, if it is not, the fix should go to the allocator.  @phambinhfin

shawnwang18:
Oh, this PR is based on old implementation https://github.com/openxla/xla/commit/ae872477540f68ac5fe4bd6c92dba0517da83acf#diff-46ef34e24c542944160b839f262e845182c98bdb108e642ad989095d5ab2fbef


then I think when this PR was doing a rebase, which just revert all the updates we made since 22-June.

phambinhfin:
Hi @shawnwang18 — thank you for the work on the allocator, and apologies if this PR affected your changes and caused a regression; that was not intended.

If the VA-remap reuse is already handled in `DeviceAddressVmmAllocator`, then I agree the skip-remap part of this PR is redundant and it should be reverted — happy to go with #45405.

For the D2D copy-into-shadow part, I'll look for a better place for it (likely in the allocator, kept backend-agnostic) and re-baseline against current main before proposing it separately.

Thanks again for catching this.

shawnwang18:
> Hi @shawnwang18 — thank you for the work on the allocator, and apologies if this PR affected your changes and caused a regression; that was not intended.
> 
> If the VA-remap reuse is already handled in `DeviceAddressVmmAllocator`, then I agree the skip-remap part of this PR is redundant and it should be reverted — happy to go with #45405.
> 
> For the D2D copy-into-shadow part, I'll look for a better place for it (likely in the allocator, kept backend-agnostic) and re-baseline against current main before proposing it separately.
> 
> Thanks again for catching this.

Thanks, for the copy-into-shadow optimization, let's do it separately, I think both cuda/rocm can apply the same optimization.

PR #44389, opened 2026-06-16
[ROCm] Enable float and buffer checker

draganmladjenovic:
📝 Summary of Changes
Enables float and buffer xor checker thunks on ROCm

🎯 Justification
Enables more debugging tools on ROCm platform

🚀 Kind of Contribution
✨ New Feature

📊 Benchmark (for Performance Improvements)
N\A

🧪 Unit Tests:
None

🧪 Execution Tests:
Moved cuda specific tests into:
//xla/stream_executor/gpu:buffer_debug_float_check_kernel_test
//xla/stream_executor/gpu:buffer_debug_xor_checksum_kernel_test

PR #45093, opened 2026-06-30
[XLA:GPU][ROCm] Fix absl::bit_cast in device code for GCC 13 compatibility

nurmukhametov:
PR #44389 introduced `buffer_debug_float_check_kernel_rocm.cu.cc`, which uses `absl::bit_cast` in `__device__ constexpr` initializers. This breaks the rocm_clang_local build on Ubuntu 24.04 where GCC 13 is the default.

GCC 13 reimplemented `std::__or_<>` using __or_fn helper functions that are not annotated `__host__ __device__`. This causes clang/HIP to reject instantiation of `std::is_trivially_copyable` (via `__is_complete_or_unbounded`) from device context during overload resolution, even when the type is complete and the slow overload is never actually called.

absl::bit_cast triggers this via its static_assert on is_trivially_copyable. Replace `absl::bit_cast` with `__builtin_bit_cast` in `__device__ code`, which is a compiler intrinsic and bypasses standard library type trait machinery entirely. Also substitute the bfloat16 infinity bit pattern (0x7F80) directly.

### DIFF
--- a/xla/stream_executor/rocm/buffer_debug_float_check_kernel_rocm.cu.cc
+++ b/xla/stream_executor/rocm/buffer_debug_float_check_kernel_rocm.cu.cc
@@ -38,12 +38,11 @@ namespace stream_executor::gpu {
 
 template <>
 __device__ constexpr hip_bfloat16 kInfinity<hip_bfloat16> =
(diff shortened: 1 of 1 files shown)


draganmladjenovic (approved):

derdrdirk (approved):
LGTM

PR #45130, opened 2026-07-01
[XLA:GPU] Trigger command buffers update when persistent allocations set changed.

shawnwang18:
For command buffer run, add the required processing for allowing `persistent_alloc_indices` to transition only from absent to present across execution steps and must remain unchanged once present. 

We may require this transition because the chosen of buffer allocations whose address are in `persistent_alloc_indices` might not be available from the first step, e.g, if persistent allocations are only available after several steps of profiling.  So we need to make sure that command buffer is correctly updated when `persistent_alloc_indices` is changed across steps. 

📝 Summary of Changes

- Document that `persistent_alloc_indices` may transition only from absent to
  present across execution steps and must remain unchanged once present.
- Track only whether persistent-allocation information was valid when each
  executor's command buffer was recorded, avoiding an owned vector copy and
  repeated value comparisons.
- Force a full command-buffer update in `CommandBufferThunk::Initialize` when
  persistent allocation information first becomes valid.
- In `ExecuteOnStream`, recover from a recorded/current information-validity
  mismatch by fully updating the command buffer. This covers a
  profiling-skipped update and interleaving with an older execution initialized
  before policy activation.
- Use the public `ShapedSlice` header and its direct BUILD dependency.

🎯 Justification

The persistent allocation policy controls which recorded addresses a command
buffer can safely retain. When the policy first becomes available, an
allocation-filtered update can skip a command whose allocations are now all
persistent, leaving stale addresses in the command buffer. A full update fixes
that transition.

Although the policy transition is logically one-way, executions initialized on
opposite sides of it can interleave on the shared per-executor command buffer.
An Execute-time validity check therefore forces a full update for either
recorded/current mismatch without comparing policy vectors on every step.

🚀 Kind of Contribution

🐛 Bug Fix

N/A. This is a correctness fix.

🧪 Unit Tests:

Extended `PersistentAllocIndicesBecomingAvailableTriggersUpdate` to verify:

- absent-to-present activation during Initialize triggers exactly one update;
- Execute recovers when the initialization update was skipped;
- an older absent-policy execution and the following present-policy execution
  each restore the correct recorded addresses when they interleave.

Also ran `GpuExecutableTest.CommandBufferAllocationPolicy`.

🧪 Execution Tests:

Ran the focused command-buffer regression and allocation-policy test with
Bazel. All seven configured GPU architecture variants of
`command_buffer_thunk_test` passed, and `gpu_executable_test` passed.


xla/backends/gpu/runtime/command_buffer_thunk_test.cc
```diff
@@ -165,6 +165,30 @@ class RequiresUpdateOnExecuteMemsetThunk : public Memset32BitValueThunk {
   int* record_count_;
 };
 
+class CountingDeviceToDeviceCopyThunk : public DeviceToDeviceCopyThunk {
```
dimitar-asenov: It's a bit heavy to introduce this extra class only so that you count calls to `Record`. Could you explain what that is the better design choice? E.g. what if you added `record_count_` directly to `DeviceToDeviceCopyThunk`, would that be possible and will it make anything worse?

shawnwang18: I would not add `record_count_`  directly to `DeviceToDeviceCopyThunk`, as Command objects must be stateless and thread-safe because the same command can be recorded concurrently on different executors. A plain counter would race; an atomic counter would add production-only test instrumentation and overhead. 

It is only test purpose, another way is to generalize the existing test-only memset helper into a configurable CountingMemsetThunk and reuse it here, this may require to rewrite the test using the  CountingMemsetThunk, but you would lose the D2D-specific coverage of simultaneously rebinding source and destination allocations. So maybe I would like to keep this class in the test, how do you think?

shawnwang18: I've removed class CountingDeviceToDeviceCopyThunk  @dimitar-asenov could you help take a look again?

dimitar-asenov: OK, thanks for explaining. I missed the part that we will lose coverage this way. Perhaps, they old way was better. Let's keep the code with the D2D-specific bit, can you please revert the last commit (or otherwise adapt the code?)

Truly sorry for the hassle and the misunderstanding.

dimitar-asenov (approved):

PR #45218, opened 2026-07-03
[HLO] Keep self-pairs in AllGatherDynamicSlicePermutedOffsetSimplifier

EylonKrause:
### Problem
`AllGatherDynamicSlicePermutedOffsetSimplifier` can miscompile: partitions whose permutation entry is a fixed point (they keep their own shard) end up receiving zeros.

### Root cause
`HandleDynamicSlice` (`all_gather_permuted_ds_simplifier.cc:73-80`) strips self-pairs (`src == dst`) from `permutation_pairs` before building the collective-permute. The pairs are `(src_partition_id, dest_partition_id)`; a fixed point `(p, p)` means partition `p` recovers its own all-gather shard. Per collective-permute semantics (`docs/operation_semantics.md`; `xla_builder.h`), a replica/partition that is **not** a target of any pair outputs a tensor of zeros. Removing the self-pair makes those partitions non-targets, so they get zeros instead of their own data. The transform is only correct for derangements.

This is exercised by the existing tests `AllPartitionsPermutedMultipleReplicaGroupsAlmostMatch` and `AllPartitionsPermutedMultipleReplicaGroups`, but they assert only `source_target_pairs` structure, not numeric output, so the miscompile is not caught.

### Fix
Keep all pairs — a self-pair `(p, p)` is a valid identity copy — so every partition remains a target.

### Verification
Static reasoning from the documented collective-permute contract; no bazel build available on my machine. Happy to extend a test to evaluate numeric output for a permutation containing a fixed point.

### DIFF
--- a/xla/hlo/transforms/simplifiers/all_gather_permuted_ds_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/all_gather_permuted_ds_simplifier.cc
@@ -71,15 +71,15 @@ AllGatherDynamicSlicePermutedOffsetSimplifierVisitor::HandleDynamicSlice(
           /*allow_multiple_users=*/false);
 
   if (offset_spec.has_value() && !offset_spec->permutation_pairs.empty()) {
-    // Remove the duplicated pairs as no collective permute is needed.
-    offset_spec->permutation_pairs.erase(
-        std::remove_if(offset_spec->permutation_pairs.begin(),
-                       offset_spec->permutation_pairs.end(),
-                       [](const std::pair<int64_t, int64_t>& pair) {
-                         return pair.first == pair.second;
-                       }),
-        offset_spec->permutation_pairs.end());
     // Replace the pattern with a collective permute.
+    //
+    // Note: self-pairs (src == dst) must be kept. A pair (p, p) means partition
+    // p reads the all-gather offset holding its own shard, i.e. it keeps its
+    // own data. Removing such pairs would make those partitions non-targets of
+    // the collective-permute, and per its semantics a replica/partition that is
+    // not a target of any pair outputs zeros -- silently replacing the
+    // partition's data with zeros. Keeping the self-pair produces the correct
+    // identity copy.
     HloInstruction* cp =
         dynamic_slice->AddInstruction(HloInstruction::CreateCollectivePermute(
             dynamic_slice->shape(), all_gather->mutable_operand(0),
--- a/xla/hlo/transforms/simplifiers/all_gather_permuted_ds_simplifier_test.cc
+++ b/xla/hlo/transforms/simplifiers/all_gather_permuted_ds_simplifier_test.cc
@@ -126,9 +126,14 @@ TEST_F(AllGatherPermutedDsSimplifierTest,
 
   HloInstruction* root = module->entry_computation()->root_instruction();
   EXPECT_THAT(root, op::CollectivePermute(op::Parameter(0)));
+  // Self-pairs (e.g. {0, 0}) must be kept: a partition that reads its own
+  // all-gather offset keeps its own data. A collective-permute zeroes any
+  // partition that is not a target of some pair, so dropping the self-pairs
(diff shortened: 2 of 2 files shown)


allanrenucci (changes requested):
Thanks for the contribution. Please add a test or extend an existing one

EylonKrause:
Thanks @allanrenucci! Added test coverage in `all_gather_permuted_ds_simplifier_test.cc`:

- **New** `KeepsSelfPairsForIdentityPartitions`: only partitions 0 and 1 swap shards; 2–7 read their own shard. Asserts the self-pairs `{2,2}..{7,7}` are present in the collective-permute.
- **Updated** `AllPartitionsPermutedSingleReplicaGroupAlmostMatch` and `AllPartitionsPermutedMultipleReplicaGroups`: their previous expectations encoded the old behavior of dropping self-pairs. Since a collective-permute zeroes any partition that is not a target of some pair, those expectations described partitions being silently zeroed; they now include the self-pairs.

EylonKrause:

felixwqp:
Thanks for contributing, can you please test a e2e workload using a real hardware, e.g. gpu, to ensure this hlo works well?
You can use docs/tools_multihost_hlo_runner.md to test.

EylonKrause:
Thanks @felixwqp! I'd like to, but I'm developing on a single-GPU workstation (one consumer GPU under WSL2) with no multi-GPU or multi-host access, so I can't run a true multi-partition all-gather workload end-to-end via the multihost HLO runner — this simplification only exercises meaningfully across ≥2 partitions.

What I can stand behind is the HLO-level coverage: the added `KeepsSelfPairsForIdentityPartitions` asserts the self-pairs `{2,2}..{7,7}` are preserved (so partitions reading their own shard are no longer zeroed), and I updated the two existing tests whose expected pairs had encoded the old self-pair-dropping behavior.

If someone with a multi-GPU host could run the e2e check via `docs/tools_multihost_hlo_runner.md`, that would be ideal — and I'm glad to extend the HLO-level test coverage further in the meantime if that helps. Apologies I can't run the hardware validation myself.

EylonKrause:
Rebased onto current `main` — the branch was ~500 commits behind, so the red checks above were all from a stale base. CI is re-running fresh.

@allanrenucci — on the test request, the PR now has a dedicated `KeepsSelfPairsForIdentityPartitions` (partitions 0 and 1 swap shards while 2–7 read their own, so the pass must emit `{0,1}, {1,0}` **and** the self-pairs `{2,2}…{7,7}`), plus the two existing tests extended to assert the self-pairs (`{0,0}`, `{3,3}`…, `{5,5}`) that the old code silently dropped.

@felixwqp — following up on the hardware question with something more concrete than my earlier reply. This is a documented correctness issue rather than a heuristic change; `docs/operation_semantics.md` states, under the restrictions on `source_target_pairs`:

> If a replica id is not a target in any pair, then the output on that replica is a tensor consisting of 0(s) with the same shape as the input.

The pre-change code erased every `(p, p)` pair before constructing the collective-permute. A partition whose only pair was its self-pair is then the target of no pair, so by the rule above its output is zeros — even though the pattern being simplified (that partition dynamic-slicing the all-gather at its own offset) means it must keep its own data. That is what an e2e run would surface: wrong numerics on any sharding where some partition maps to itself, silently, with no crash.

I still can't run the multihost HLO runner here (single consumer GPU, no multi-partition host), so if you'd like the hardware signal before merging I'd appreciate someone with a multi-GPU host running it — or point me at a specific workload/config you'd consider sufficient and I'll get as close as I can.

felixwqp (approved):

felixwqp:
Sounds good, once internal change is created, we will run an internal multi-host test as multi-host coverage.

EylonKrause:
Thanks @felixwqp — glad it's approved. Sounds good on the internal multi-host test; happy to make any adjustments if that coverage turns something up. Appreciate you shepherding it through the internal import.

PR #45263, opened 2026-07-06
Allow dynamic-update-slice into an unbounded-dynamic operand dimension

EylonKrause:
### Problem
`InferDynamicUpdateSliceShape` rejects a valid `dynamic-update-slice` into an **unbounded-dynamic** operand dimension.

### Root cause
```cpp
const int64_t input_dim_size = operand_shape.dimensions(dim);
...
if (update_dim_size > input_dim_size) {   // input_dim_size == kUnboundedSize (INT64_MIN) for an unbounded dim
  return InvalidArgument("Update dim size %d greater than dynamic slice dimension: %d.", ...);
}
```
For an unbounded-dynamic operand dimension, `operand_shape.dimensions(dim)` is `kUnboundedSize` (`INT64_MIN`). Any concrete static update size (e.g. `5`) then satisfies `5 > INT64_MIN` and the op is rejected with a nonsensical bound (`-9223372036854775808`). The sibling `InferDynamicSliceShape` guards the identical comparison with `!IsUnboundedDynamicSize(input_dim_size)`, and this function already handles unbounded *update* sizes just above — so unbounded operands are in scope, but the operand side is unguarded.

### Fix
Guard with `!IsUnboundedDynamicSize(input_dim_size)`, matching `InferDynamicSliceShape`.

### Verification
Static reasoning: e.g. `operand = f32[?, 10]`, `update = f32[3, 5]` is rejected today (`3 > INT64_MIN`) and accepted with the fix; operand-static-with-too-large-update still errors; both-unbounded and operand-static-with-unbounded-update are unchanged. (No local Bazel build set up — happy to add a shape-inference test.)

### DIFF
--- a/xla/service/shape_inference.cc
+++ b/xla/service/shape_inference.cc
@@ -3442,7 +3442,8 @@ ShapeInference::InferCollectivePermuteDoneShape(const Shape& operand_shape) {
           "Size index %d to dynamic update slice must be >= 0.",
           update_dim_size);
     }
-    if (update_dim_size > input_dim_size) {
+    if (!IsUnboundedDynamicSize(input_dim_size) &&
+        update_dim_size > input_dim_size) {
       return InvalidArgument(
           "Update dim size %d greater than dynamic slice dimension: %d.",
           update_dim_size, input_dim_size);


xla/service/shape_inference.cc
allanrenucci, line 3440: Let's move the `IsUnboundedDynamicSize` check out. I.e.
```c++
if (IsUnboundedDynamicSize(update_dim_size)) {
  continue;
}
if (update_dim_size < 0) { ... }
if (update_dim_size > input_dim_size) { ... }
```

allanrenucci (changes requested):

EylonKrause:

EylonKrause:
Done — moved the unbounded-update check to an early `continue` so the `>= 0` check no longer needs its own guard. I kept `!IsUnboundedDynamicSize(input_dim_size)` on the `>` comparison, though: that guard is the actual fix here. An unbounded-dynamic **operand** dimension is `kUnboundedSize` (`INT64_MIN`) independently of the update dim, so without it a concrete update size (e.g. `5`) is always `> INT64_MIN` and a valid DUS into an unbounded operand dim gets rejected with a nonsensical bound. Happy to adjust further if you had a different structure in mind.

allanrenucci (changes requested):
Sorry, I misread your change and thought the `IsUnboundedDynamicSize` predicate was shared between the two if statements. Let's drop the last commit.

EylonKrause:
Done — dropped the redundant input_dim_size guard on the > check. The loop is now exactly the structure you described: early continue on an unbounded update dim, then the plain < 0 and > input_dim_size checks.

allanrenucci:
No. I am suggesting reverting to the state at 1b4c86365393d200d45a372af006ea0a64b94a74

EylonKrause:
Done — reverted to 1b4c863 (the inline IsUnboundedDynamicSize guards on both the < 0 and > checks).

allanrenucci (approved):
Thanks!

PR #45275, opened 2026-07-06
Restructure clang-format check to enable running it locally.

sergachev:
📝 Summary of Changes
Restructure clang-format check to enable running it locally.

🚀 Kind of Contribution
♻️ Cleanup

### DIFF
--- a/build_tools/ci/run_clang_format.sh
+++ b/build_tools/ci/run_clang_format.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# Copyright 2026 The OpenXLA Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


sergachev:
Stacked with https://github.com/openxla/xla/pull/45274

build_tools/ci/run_clang_format.sh
```diff
@@ -0,0 +1,73 @@
...
+  set +x
+  echo "Error: uv is required to run the clang-format check." >&2
+  echo "Install uv from https://docs.astral.sh/uv/" >&2
+  exit 1
+fi
+
+get_merge_base() {
+  if ! git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    set +x
+    echo "Error: This script must be run inside a Git repository." >&2
```
sohaibiftikhar: You may want to use a utility to avoid repetition. Something like:

https://github.com/openxla/xla/blob/main/build_tools/ci/clang_tidy_wrapper.sh.tpl#L4

build_tools/ci/run_clang_format.sh
```diff
@@ -0,0 +1,73 @@
...
+    exit 1
+  fi
+
+  if [ -n "${TARGET_REF:-}" ]; then
```
sohaibiftikhar:
```suggestion
  if [ -n "${TARGET_REF}" ]; then
```

build_tools/ci/run_clang_format.sh
```diff
@@ -0,0 +1,73 @@
...
+    REFERENCE="$TARGET_REF"
```
sohaibiftikhar, line 43: Consider marking varibales that are not used outside of this function as local.

sohaibiftikhar (approved):

Issue #45184, opened 2026-07-02
[XLA:GPU][oneAPI/SPIR-V] LLVM to SPIR-V address-space handling


neudinger:
Hello XLA / INTEL team  @mraunak  @akhilgoe .

In a custom fork of XLA, I got some unexpected behavior when trying to run some kernels on Intel GPU.

I looked on how the lowering work from LLVM IR to SPIR-V, and I try to compare on how work the address-space handling between NVIDIA, ROCm and the LLVM IR to SPIR-V for intel GPU.

For NVPTX and AMDGPU, XLA goes through backend-specific LLVM targets. 

I found that the SPIR-V backend already has an explicit address-space inference pass in `spirv_backend.cc` because the SPIR-V target does not provide the same inference through `TargetPassConfig`:
It also seems that the kernel argument rewrite in `spirv_backend.cc` assumes all kernel arguments are pointers.
But a valid `spir_kernel` can contain scalar arguments mixed with pointer arguments.

- https://github.com/openxla/xla/blob/main/xla/service/gpu/llvm_gpu_backend/spirv_backend.cc#L75-L78

XLA also already has SPIR/SPIR-V-specific address-space handling outside this file.
For example, constants and kernel pointer arguments are explicitly placed in global address space `1` for SPIR/SPIR-V here:

- https://github.com/openxla/xla/blob/main/xla/service/llvm_ir/fused_ir_emitter.cc#L99-L110
- https://github.com/openxla/xla/blob/main/xla/backends/gpu/codegen/fusion_emitter.cc#L159-L168
- https://github.com/openxla/xla/blob/main/xla/backends/gpu/codegen/llvm/llvm_emitter.cc#L750-L766

Is it better to do this in the SPIR-V LLVM address-space handling, or trying to copy the NVIDIA or ROCm lowering path (I do not know how) ?

I looked on Intel’s TritonGEN lowering, because it seems closer to the SPIR-V model. 

In `TritonGENMemorySpace.h` the memory-space identifiers following the SPIR-V storage class convention:

- https://github.com/intel/intel-xpu-backend-for-triton/tree/main/third_party/intel/include/Dialect/TritonGEN
- https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/include/Dialect/TritonGEN/IR/TritonGENMemorySpace.h

I mirror it in XLA as the Khronos SPIR-V LLVM representation mapping:

- `addrspace(0)`: default / Function
[...]

PR #45282, opened 2026-07-06
[XLA:GPU][oneAPI/SPIR-V] Harden kernel argument address-space rewrite

neudinger:
📝 Summary of Changes

Harden SPIR-V kernel argument rewriting so only pointer arguments in the default LLVM address space are rewritten to CrossWorkgroup. Add local names for the SPIR-V LLVM address spaces used here and a regression test for mixed scalar/pointer kernel signatures.

🎯 Justification

Valid `spir_kernel` functions can contain scalar arguments mixed with pointer arguments. The previous rewrite assumed pointer arguments, which could break valid oneAPI/SPIR-V lowering paths.

Addresses #45184.

🚀 Kind of Contribution
🐛 Bug Fix, ♻️ Cleanup, 🧪 Tests

N/A, this is not a performance change.

🧪 Unit Tests:

Added `KernelArgumentRewritePreservesScalarArguments` in `//xla/service/gpu/llvm_gpu_backend:spirv_backend_test`.

Verified with:

`bazel test //xla/service/gpu/llvm_gpu_backend:spirv_backend_test`

🧪 Execution Tests:

No execution test added. This is covered by the focused SPIR-V backend unit test.

@akhilgoe


xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
```diff
@@ -182,16 +187,24 @@ absl::StatusOr<std::string> CompileToSPIRV(
   for (auto old_func : kernel_funcs) {
     if (old_func->getCallingConv() == llvm::CallingConv::SPIR_KERNEL) {
       std::vector<llvm::Type*> new_arg_types;
+      bool needs_rewrite = false;
       for (auto& old_arg : old_func->args()) {
         llvm::Type* old_arg_type = old_arg.getType();
         auto ptr_type = llvm::dyn_cast<llvm::PointerType>(old_arg_type);
-        if (ptr_type->getAddressSpace() == 0) {
-          auto new_arg_type = llvm::PointerType::get(context, 1);
+        if (ptr_type != nullptr &&
```
mdfaijul: Simpler expression `if (ptr_type && ptr_type->getAddressSpace() == kSpirvLlvmDefaultAddressSpace)`

xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
```diff
@@ -213,11 +226,14 @@ absl::StatusOr<std::string> CompileToSPIRV(
       auto new_arg_it = new_func->arg_begin();
       auto old_arg_it = old_func->arg_begin();
       for (; old_arg_it != old_func->arg_end(); ++old_arg_it, ++new_arg_it) {
+        llvm::Value* replacement = &*new_arg_it;
         if (auto old_ptr_type =
-                llvm::dyn_cast<llvm::PointerType>(old_arg_it->getType())) {
-          auto cast = builder.CreateAddrSpaceCast(new_arg_it, old_ptr_type);
-          old_arg_it->replaceAllUsesWith(cast);
+                llvm::dyn_cast<llvm::PointerType>(old_arg_it->getType());
+            old_ptr_type != nullptr &&
```
mdfaijul: We can use `old_ptr_type` instead of `old_ptr_type != nullptr`.

mdfaijul (changes requested):
Thanks for the PR. Overall looks good to me. Minor nits only as commented.

xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
neudinger: Done

xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
neudinger: Done

mdfaijul (approved):

xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
```diff
@@ -213,11 +225,14 @@ absl::StatusOr<std::string> CompileToSPIRV(
...
         if (auto old_ptr_type =
-                llvm::dyn_cast<llvm::PointerType>(old_arg_it->getType())) {
-          auto cast = builder.CreateAddrSpaceCast(new_arg_it, old_ptr_type);
-          old_arg_it->replaceAllUsesWith(cast);
+                llvm::dyn_cast<llvm::PointerType>(old_arg_it->getType());
+            old_ptr_type &&
+            old_ptr_type->getAddressSpace() == kSpirvLlvmDefaultAddressSpace) {
+          replacement = builder.CreateAddrSpaceCast(new_arg_it, old_ptr_type);
         }
+        old_arg_it->replaceAllUsesWith(replacement);
```
akhilgoe: This block change seems functionally equivalent to the prior change. Am I overlooking anything?

akhilgoe (commented):
The current testing feels a bit heavier than necessary. Without the pointer guard, compiling kernels with scalar arguments would fail, so a simple test that compiles a kernel with scalar arguments and checks `EXPECT_OK(CompileToSPIRV(...))` seems sufficient to me. If we also want to validate the generated binary, I’d suggest using a lit test instead. What do you think?

xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
neudinger: I Simplified the tests as requested

xla/service/gpu/llvm_gpu_backend/spirv_backend_test.cc
```diff
@@ -15,11 +15,41 @@ limitations under the License.
...
+std::unique_ptr<llvm::Module> ParseLlvmIr(std::string_view ir,
+                                          llvm::LLVMContext& context) {
+  llvm::SMDiagnostic diagnostic;
+  std::unique_ptr<llvm::Module> module =
+      llvm::parseAssemblyString(ir, diagnostic, context);
+  if (module == nullptr) {
+    ADD_FAILURE() << "Failed to parse LLVM IR: "
+                  << diagnostic.getMessage().str();
+  }
+  return module;
```
allanrenucci, line 53: Let's instead return a `StatusOr<std::unique_ptr<llvm::Module>>`. Callers can then do:

```cc
ASSERT_OK_AND_ASSIGN(module, ...);
```

allanrenucci (changes requested):
One nit. Otherwise LGTM

xla/service/gpu/llvm_gpu_backend/spirv_backend_test.cc
neudinger: It was to avoid the heavy absl/status/status.h
And avoid @com_google_absl//absl/.. in the bazel dependency  for one line

xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
akhilgoe: Thanks for the simplification! Even here I think you can revert to the original code. Since builder returns the source when op and target types are identical, no redundant casts will occur in original code.

xla/service/gpu/llvm_gpu_backend/spirv_backend.cc
```diff
@@ -213,11 +226,18 @@ absl::StatusOr<std::string> CompileToSPIRV(
...
```
akhilgoe: Thanks for refining the test! Even here I think you can revert to the original code. Since builder returns the source when op and target types are identical, no redundant casts will occur in original code.

akhilgoe (commented):
Thanks @neudinger for the PR. I just have an optional simplification nit.

allanrenucci (approved):

akhilgoe (approved):
Thanks again for the PR!

PR #45296, opened 2026-07-06
[XLA:CPU][oneDNN] Add GELU_ERF pattern matching for StableHLO legalized erfc

othakkar:
Expand CPU oneDNN GELU pattern matching to recognize the exact-GELU form emitted by JAX/StableHLO after `erfc` legalization. This handles the piecewise `select(abs(y) < 1, 1 - ..., ...)` structure for `erfc(-x / sqrt(2))`, in addition to the existing `tanh` and `erf` forms. The decomposition happens because exact GELU lowers through `erfc`, and StableHLO legalizes `erfc` into portable primitive ops rather than preserving it as a first-class op. The matcher stays loose on approximation internals and instead matches the stable mathematical skeleton, improving fusion robustness for Keras/JAX workloads.

### DIFF
--- a/xla/service/cpu/onednn_contraction_rewriter.cc
+++ b/xla/service/cpu/onednn_contraction_rewriter.cc
@@ -222,6 +222,12 @@ bool IsScalar(const HloInstruction* instr) {
   return ShapeUtil::IsEffectiveScalar(instr->shape());
 }
 
+// Match pattern, optionally looking through a convert first.
+template <typename Pattern>
+auto OptionalConvertTo(Pattern pattern) {
+  return m::AnyOf<HloInstruction>(m::Convert(pattern), std::move(pattern));
+}
+
 std::optional<float> GetConstantValueAsFloat32(const HloInstruction* inst) {
   if (!IsScalar(inst)) {
     return std::nullopt;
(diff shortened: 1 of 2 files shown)


xla/service/cpu/onednn_contraction_rewriter.cc
```diff
@@ -285,6 +291,84 @@ inline auto MultiplyMultiplyAnyOrder(PatternA a, PatternB b, PatternC c) {
       m::MultiplyAnyOrder(c, m::MultiplyAnyOrder(a, b)));
 }
 
+// Broadcast of 1/sqrt(2) ≈ 0.7071. Multiple approximations are matched
+// because different precisions (F32, BF16, F16) and compiler optimizations
+// produce slightly different constant values:
+//   0.707106769 — F32 nearest representation of 1/√2
+//   0.70703125  — BF16 nearest representation
+//   0.707182348 — F16-derived approximation after convert to F32
+inline auto BcastInvSqrt2() {
```
akuegel, line 300: Can you please wrap these helpers into an anonymous namespace block? Or alternative make them static?

akuegel (commented):
We still have an open internal bug to make Erfc an HLO op. Once/If that happens, the matcher code would become obsolete.

xla/service/cpu/onednn_contraction_rewriter.cc
othakkar: @akuegel, thanks for the review comment. The helpers are already inside an existing anonymous namespace (lines 73 - 590), which gives them internal linkage. Please let me know if that's what you meant.

akuegel: Ah, thanks, I didn't notice. Yes, that was what I meant.

othakkar:
> We still have an open internal bug to make Erfc an HLO op. Once/If that happens, the matcher code would become obsolete.

@akuegel Can we still merge this PR until/if Erfc becomes an HLO op? We can remove it later.

akuegel (approved):

PR #45312, opened 2026-07-07
[ROCm] Allow asynchronous deallocation for StreamExecutorAddressAllocator

Ruturaj4:
### 📝 Summary of Changes
Report `AllowsAsynchronousDeallocation() == true` for
`StreamExecutorAddressAllocator` **on the ROCM platform only**, and drop the
now-obsolete synchronous-only `CHECK` in `GetStream()`. With this,
`OwningScratchAllocator` frees scratch inline instead of deferring it into a
`Stream::DoHostCallback`. CUDA and SYCL keep their prior behavior.

### 🎯 Justification
On ROCm the synchronous free calls `hipFree`, which deadlocks when invoked
from a HIP stream-callback thread. `OwningScratchAllocator` only defers the
free into a host callback for allocators reporting async-dealloc == false, so
any op allocating a work buffer hangs — notably real FFTs
(`rfft`/`irfft`/`hfft`/`ihfft`/`rfft2`) under the address allocator on gfx950,
which blocked the JAX ROCm test suite at the FFT stage. `Deallocate()` uses
`hipFree`, which already synchronizes the device, so deallocation is safe to
issue asynchronously; reporting `true` frees scratch inline (off the callback)
and keeps the path non-blocking, matching the VMM allocators.

This is scoped to the ROCM platform. CUDA's `cuMemFree` is also
device-synchronizing and would be safe, but SYCL's `sycl::free` synchronizes
only the passed queue (not all device queues), so returning `true` there could
free memory still in use on another queue. CUDA and SYCL therefore retain
their existing behavior.

`GetStream()` is also used by the autotuner/profiler regardless of dealloc
mode, so its `!AllowsAsynchronousDeallocation()` CHECK is removed.

### 🚀 Kind of Contribution
🐛 Bug Fix

### 🧪 Unit Tests:
None. The deadlock is ROCm-GPU-specific (`hipFree` in a HIP callback); on the
host/CPU platform `DoHostCallback` runs fine, so a CPU unit test cannot
reproduce it and would pass with or without the change.

### 🧪 Execution Tests:
Covered by existing JAX ROCm FFT tests (`fft_test.py`, `scipy_fft_test.py`)
run under `XLA_PYTHON_CLIENT_ALLOCATOR=address`: real-transform cases hang
without this fix and pass with it (verified on gfx950 / ROCm 7.2.0).

### DIFF
--- a/xla/stream_executor/stream_executor_address_allocator.cc
+++ b/xla/stream_executor/stream_executor_address_allocator.cc
@@ -96,13 +96,18 @@ StreamExecutorAddressAllocator::GetStreamExecutor(int device_ordinal) const {
 }
 
 bool StreamExecutorAddressAllocator::AllowsAsynchronousDeallocation() const {
-  return false;
+  // Only report asynchronous deallocation on ROCm. Deallocate() routes to
+  // hipFree, which synchronizes the whole device, so scratch can be freed
+  // inline instead of being deferred into a stream host callback (where hipFree
+  // deadlocks on ROCm). CUDA's cuMemFree is also device-synchronizing and would
+  // be safe, but SYCL's sycl::free only synchronizes the passed queue, so
+  // returning true there could free memory still in use on another queue.
+  // Scope the behavior to ROCm to keep the other platforms unchanged.
+  return platform()->Name() == "ROCM";
 }
 
 absl::StatusOr<Stream*> StreamExecutorAddressAllocator::GetStream(
     int device_ordinal) {
-  CHECK(!AllowsAsynchronousDeallocation())
-      << "The logic below only works for synchronous allocators";
   ASSIGN_OR_RETURN(StreamExecutor * executor,
                    GetStreamExecutor(device_ordinal));
   absl::MutexLock lock(mutex_);


xla/stream_executor/scratch_allocator.h
```diff
@@ -98,11 +98,12 @@ class OwningScratchAllocator : public ScratchAllocator {
       LOG(ERROR) << "Allocator returned a null stream for asynchronous "
                     "deallocation.";
     } else {
-      if (absl::Status s = (*stream)->DoHostCallback(
-              [buffers = std::move(buffers_)]() mutable { buffers.clear(); });
-          !s.ok()) {
-        LOG(ERROR) << "Failed to schedule scratch allocator cleanup: " << s;
+      // A synchronous free (e.g. hipFree) must not run inside a stream host
```
draganmladjenovic: ⁠Whole point is to make it non blocking for BFC and you are making it blocking there.  Albeit not true on the surface for StreamExecutorAddressAllocator you can set AllowsAsynchronousDeallocation to return true. You technically don't have to wait  for stream to complete. hipFree will do that for you.

draganmladjenovic (changes requested):

magaonka-amd:
@Ruturaj4  can you match the PR description as per your latest approach ?
other than that latest changes looks good to me.

hsharsha:
@xla-rotation Can you please take a look?

xla/stream_executor/stream_executor_address_allocator.cc
```diff
@@ -96,13 +96,15 @@ StreamExecutorAddressAllocator::GetStreamExecutor(int device_ordinal) const {
 }
 
 bool StreamExecutorAddressAllocator::AllowsAsynchronousDeallocation() const {
-  return false;
+  // Deallocate() uses a synchronous free (e.g. hipFree) that already
+  // synchronizes the device, so deallocation is safe to issue asynchronously.
+  // Reporting true also keeps OwningScratchAllocator from deferring the scratch
+  // free into a stream host callback, where hipFree deadlocks on ROCm.
+  return true;
```
allanrenucci: This seems to rely on an implementation detail of the ROCM platform. Can you explain why it is correct to always return true here? I.e. is it also safe on the CUDA or SYCL platforms?

allanrenucci (changes requested):

xla/stream_executor/stream_executor_address_allocator.cc
draganmladjenovic: On cuda it should be. Good point for sycl. Spec does not seem to guarantee it. oneAPI may do. Then again if it where not to wait for outstanding work, this code was then broken for quite some time. @allanrenucci Can you add sycl dev to this PR to confirm.

bhavani-subramanian: Returning `true` would break for SYCL backend. `sycl::free` synchronizes only the passed queue, not all compute queues on the device. So if a kernel runs on a different queue, the memory can get freed while the device is still accessing it.

Having said that, most of the SYCL tests (ex. those under `stream_executor/sycl`) use the PJRT client and don't go through `StreamExecutorAddressAllocator` (which is modified in this PR).

allanrenucci: Let's just return `true` for the ROCM platform.

Ruturaj4: Updated to scope this to the ROCM platform only: `return platform()->Name() == "ROCM";` (pushed as `e7351fd`). CUDA and SYCL now keep their prior behavior (returning `false`), so the SYCL concern is addressed — we no longer rely on `sycl::free`'s per-queue-only synchronization (thanks @bhavani-subramanian for the detail, and @draganmladjenovic / @allanrenucci for the CUDA/SYCL analysis). The ROCm rationale (`hipFree` self-synchronizes the device, and deferring the free into a HIP stream host-callback deadlocks) is in the updated commit message and PR description. PTAL.

xla/stream_executor/scratch_allocator.h
Ruturaj4: The blocking approach in `scratch_allocator.h` was reverted; the final diff only touches `stream_executor/stream_executor_address_allocator.cc` (scopes `AllowsAsynchronousDeallocation()` to the ROCM platform and drops the obsolete `GetStream()` CHECK). `scratch_allocator.h` is unchanged now — resolving this thread.

allanrenucci (approved):

PR #45323, opened 2026-07-07
[XLA:SE] Refactoring/cleaning DeviceAddressVmmAllocator, adding tests for DeviceAddressVmmAllocator 

shawnwang18:
This PR does refactoring of existing code, and no functional behavior change is made.  

To make the review more easily, this PR has 13 commits, which gradually refactoring the code, each commit is kept simple and small. 

📝 Summary of Changes
- Simplify DeviceAddressVmmAllocator allocation-record and reservation-alias bookkeeping.
- Consolidate mapped allocation reuse/creation, Map target resolution, and pending deallocation handling.
- Account physical-address budget using the committed allocation size.
- Add focused tests for pending reclamation, mapped allocation reuse, and physical budget accounting.

🎯 Justification
The allocator had accumulated overlapping helpers and state representations that made ownership, stale mappings, and deferred completion difficult to reason about. This cleanup reduces code size and gives each lifecycle transition one implementation while preserving existing behavior.

🚀 Kind of Contribution
♻️ Cleanup, 🧪 Tests

Not applicable; this is a behavior-preserving cleanup rather than a performance change.

🧪 Unit Tests:
- `bazel test //xla/stream_executor:device_address_vmm_allocator_test`

🧪 Execution Tests:
- `bazel test //xla/stream_executor/cuda:cuda_device_address_vmm_allocator_test`

### DIFF
--- a/xla/stream_executor/device_address_vmm_allocator.cc
+++ b/xla/stream_executor/device_address_vmm_allocator.cc
@@ -95,7 +95,7 @@ DeviceAddressVmmAllocator::AllocationRecord::AllocationRecord(
       CHECK(allocator_address_reservation_ == nullptr);
       break;
   }
-  CHECK(allocator_address_mapping_.has_value());
+  CHECK(!allocator_address_mapping_.is_null());
 }
 
 DeviceAddressVmmAllocator::PendingDeallocationKind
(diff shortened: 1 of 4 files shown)


xla/stream_executor/device_address_vmm_allocator.cc
```diff
@@ -393,137 +358,115 @@ uint64_t DeviceAddressVmmAllocator::GetAllocationGranularity(
...
-      continue;
-    }
-    CHECK(record.has_reservation_address());
-    // The allocator address can be reused for command-buffer update-free
-    // execution only if the external reservation VA is also the same VA the
-    // command buffer captured.
-    if (!record.reservation_matches(request.reservation_address)) {
-      continue;
-    }
-    if (record.raw_allocation()->address().size() < request.allocation_size) {
```
sergachev, line 503: How is the larger request size handled now without this check? Does any test cover this?

shawnwang18: The check here is dead code, in the refactored code, size check is already performed here: 

  bool address_matches =
      separate_address
          ? record->reservation_matches(request.reservation_address)
          : record->allocator_matches(request.reservation_address);

sergachev: OK. Is this code covered by any test?

shawnwang18: This is actually a un-supported pattern, so no tests were added for this pattern. 

The scenario here is we first allocate a smaller size (1G e.g), and map to [VA, VA + 1G], then deallocate it. 

and then follows another allocate with larger size (2G, e.g), and map to [VA, VA + 2G] (the same va base).   

This currently is not a supported scenario, and will report an error.  For the use case of DeviceAddressVmmAllocate (command buffer va-remap run), we do not have such a scenario. 

The complete design might be that the 2nd allocate needs to drain the old mapping first, then do a fresh re-mapping.  The old code was trying to do in this way, but later just make it just return earlier, which leaves this code dead.

PR #45332, opened 2026-07-07
[ROCm] Fall back to library collective for non-uniform replica groups

Ruturaj4:
The collective (one-shot/two-shot) AllReduce kernel bakes the clique size (world_size) as a compile-time constant from the first replica group and sizes its signal/staging buffers the same way. With non-uniform replica groups (e.g. axis_index_groups=[[0,1],[2,3,4,5,6,7]]) the single emitted kernel runs on cliques of different sizes, causing out-of-bounds peer accesses (GPU page fault) and hangs on gfx950.

Reject the collective kernel when replica groups differ in size so the op falls back to the library (RCCL/NCCL) collective, which handles heterogeneous cliques. Uniform replica groups are unaffected.

## 📝 Summary of Changes
Add a uniformity guard to `IsAllReduceKernelSupported` (`xla/backends/gpu/runtime/all_reduce.cc`): if any replica group's size differs from the first group's size, the custom collective AllReduce kernel is not selected and the op falls back to the library (RCCL/NCCL) collective. No device-code changes; uniform replica groups continue to use the kernel unchanged.

## 🎯 Justification
The collective kernel's clique size (`world_size`) is a codegen-time constant taken from the first replica group, and its signal/staging buffers are sized the same way. A single AllReduce with non-uniform replica groups therefore emits one kernel that is then run across cliques of *different* sizes, which reads/writes past the per-rank buffers → GPU page fault (`SQC` client, addr `0x0`) and/or a wedged HSA queue and hang. This is observed on ROCm/gfx950 with e.g. `pmap` + `axis_index_groups=[[0,1],[2,3,4,5,6,7]]`. Any workload that issues an AllReduce over uneven replica groups is affected; the kernel is architecturally single-clique-size, so falling back to the library collective (which handles heterogeneous cliques) is the correct behavior.

## 🚀 Kind of Contribution
🐛 Bug Fix

N/A — this is a correctness fix. It only diverts non-uniform-group AllReduces (which are broken today) to the library collective; uniform-group AllReduces still use the custom kernel, so no performance change for supported cases.

## 🧪 Unit Tests:
No new unit test in this commit. A targeted addition would be an assertion that `IsAllReduceKernelSupported(...)` returns an `Unimplemented` status for a replica-group set with differing group sizes (e.g. `{{0,1},{2,3,4,5}}`) and remains supported for uniform groups — happy to add this if desired.

## 🧪 Execution Tests:
Covered by the existing `tests/pmap_test.py::PythonPmapTest::testPsumUnevenReplicaGroups`, which exercises uneven `axis_index_groups` and previously hung. Note this cannot be expressed within the "at most 2 GPUs" guidance: non-uniform replica groups require at least 3 devices (two groups of different sizes), and the test itself skips when `device_count <= 2`. Validated on 8x MI355X (gfx950) with the kernel enabled: full ROCm UT suite passes — single-accelerator `29118 passed` (1 pre-existing unrelated DLPack waiver), multi-accelerator `854 passed, 0 failed` (previously wedged at this test). Path confirmed via VLOG: uneven groups → 0 kernel launches / all RCCL; even groups → kernel used.

### DIFF
--- a/xla/backends/gpu/runtime/BUILD
+++ b/xla/backends/gpu/runtime/BUILD
@@ -4151,6 +4151,7 @@ xla_cc_test(
         "//xla/service/gpu:gpu_constants",
         "//xla/service/gpu:gpu_device_info_for_tests",
         "//xla/stream_executor:device_description",
+        "//xla/stream_executor:device_description_proto_cc",
         "//xla/stream_executor/gpu:all_reduce_kernel",
         "//xla/stream_executor/host:host_platform",
         "//xla/tsl/lib/gtl:int_type",
--- a/xla/backends/gpu/runtime/all_reduce.cc
+++ b/xla/backends/gpu/runtime/all_reduce.cc
@@ -288,6 +288,17 @@ absl::Status IsAllReduceKernelSupported(
     return absl::UnimplementedError(
         "Replica groups must be explicitly provided for collective kernels.");
   }
+  // The kernel bakes world_size (from the first replica group) at codegen and
+  // sizes its buffers the same way, so it cannot serve cliques of different
+  // sizes. Fall back to the library collective for non-uniform replica groups.
+  for (const ReplicaGroup& group : replica_groups) {
+    if (group.replica_ids_size() != num_devices) {
+      return absl::UnimplementedError(absl::StrCat(
+          "Collective kernel requires all replica groups to have the same "
+          "size. Got a group of size ",
+          group.replica_ids_size(), " but expected ", num_devices, "."));
+    }
+  }
   if (!reduction_kind.has_value()) {
     return absl::UnimplementedError("Could not match reduction computation.");
   }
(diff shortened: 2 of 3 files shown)


magaonka-amd:
to add bit more context issue exist in CUDA too : https://github.com/jax-ml/jax/actions/runs/28558106527/job/84669868652#step:5:792
I see CUDA bazel 8 gpu runs are timing out since july first.
This commit : https://github.com/openxla/xla/commit/626bc4ef7d43ef21404d7e2b54a50c9f630822b7 exposed the bug I believe.

mfrancepillois (approved):
LGTM. 
Nit: The PR description section "Unit Tests" states that no unit tests have been added, whereas the PR does include unit tests for this change.

allanrenucci (approved):

PR #45345, opened 2026-07-07
[GPU] Prepare DSF V2 embedded thunks with fusion-local buffer allocations

terryysun:
📝 Summary of Changes
Thunks inside DSFv2 fusion use slices numbered against a allocation list, which is internal to fusion. Unlike ExecuteOnStrem/Record, the Prepare phase passes the module level indexing instead of building the internal buffer allocation. As a consequence FFI custom calls with Prepare callback will be indexed against the wrong allocation list.

🎯 Justification
FFI customer calls with Prepare callback will crash when module level allocation is smaller, and silently wrong when larger.

🚀 Kind of Contribution
🐛 Bug Fix

N/A.

🧪 Unit Tests:
Behavior scope is too large for a unit test to catch. N/A.

🧪 Execution Tests:
Added.

### DIFF
--- a/xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
+++ b/xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
@@ -158,10 +158,19 @@ std::string DynamicSliceFusionV2Thunk::ToString(int indent) const {
 }
 
 absl::Status DynamicSliceFusionV2Thunk::Prepare(const PrepareParams& params) {
+  // Embedded thunks' slices are relative to `embedded_allocations_`, so
+  // prepare them with the same fusion-local view that ExecuteOnStream builds.
+  std::vector<se::DeviceAddressBase> buffers = BuildDynamicSliceBuffers(
+      *params.buffer_allocations, IsInsideWhileLoopNest());
+  BufferAllocations embedded_allocs(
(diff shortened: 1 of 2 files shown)


ezhulenev (approved):

penumallas1261 (approved):

xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
```diff
@@ -158,10 +158,19 @@ std::string DynamicSliceFusionV2Thunk::ToString(int indent) const {
 }
 
 absl::Status DynamicSliceFusionV2Thunk::Prepare(const PrepareParams& params) {
+  // Embedded thunks' slices are relative to `embedded_allocations_`, so
```
akuegel, line 161: I think the same has to be done for Initialize as well. I have the same change already pending review, I suggest I will merge my own change.

akuegel:
Already landed an extended version of this PR in https://github.com/openxla/xla/commit/8a8c357a9ed41c14c62222a1a6ed47e30a583f6d

xla/backends/gpu/runtime/dynamic_slice_fusion_v2_thunk.cc
terryysun: Got it. Thanks Adrian!

PR #45377, opened 2026-07-08
[PJRT:GPU] Plumb XLA FFI handler traits through the custom call extension

tenhjo:
The PJRT GPU custom call extension has no way to pass XLA FFI handler traits: `PJRT_Gpu_Register_Custom_Call_Args` has no traits field, and jaxlib accordingly rejects `register_custom_call_target(..., traits=...)` with "The plugin does not support custom call traits". The only way to mark a handler as `XLA_FFI_HANDLER_TRAITS_COMMAND_BUFFER_COMPATIBLE` today is to compile the trait into the handler symbol, which makes it a property of the shared library rather than a per-registration choice. Whether a custom call should be captured into command buffers depends on the workload, so registration time is the right place to decide.

Changes:

- Append a `traits` field to `PJRT_Gpu_Register_Custom_Call_Args` and a second entry point, `custom_call_with_traits`, to `PJRT_Gpu_Custom_Call` (extension version 3).
- The existing `custom_call` entry point never reads `traits`, so existing callers — including those recompiled against the new headers with the grown `struct_size` — keep their exact pre-v3 behavior.
- `custom_call_with_traits` reads `traits` and forwards it to `Ffi::RegisterStaticHandler` for `api_version` 1; for the legacy `api_version` 0 registry, non-zero traits return `Unimplemented`.
- Since the extension struct's `struct_size` is filled in by the plugin, callers can detect traits support at runtime instead of relying on the compile-time extension version.

Tests: registering through the new entry point surfaces traits in the FFI registry metadata; the legacy entry point ignores a poisoned traits field; the legacy entry point still accepts the pre-v3 struct size; the new entry point rejects traits for `api_version` 0.

Consumer side: jax-ml/jax#39003 uses the new entry point in jaxlib's `gpu_plugin_extension.cc`, probing for it at runtime and falling back to the existing rejection when the loaded plugin predates it.

### DIFF
--- a/xla/pjrt/c/pjrt_c_api_gpu_extension.h
+++ b/xla/pjrt/c/pjrt_c_api_gpu_extension.h
@@ -17,14 +17,14 @@ limitations under the License.
 #define XLA_PJRT_C_PJRT_C_API_GPU_EXTENSION_H_
 
 #include <stddef.h>
-
+#include <stdint.h>
 #include "xla/pjrt/c/pjrt_c_api.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#define PJRT_API_GPU_EXTENSION_VERSION 2
+#define PJRT_API_GPU_EXTENSION_VERSION 3
 
 struct PJRT_Gpu_Register_Custom_Call_Args {
   size_t struct_size;
@@ -35,8 +35,10 @@ struct PJRT_Gpu_Register_Custom_Call_Args {
   void* handler_prepare;
   void* handler_initialize;
   void* handler_execute;
+  // XLA_FFI_Handler_TraitsBits. Read only by custom_call_with_traits.
+  uint32_t traits;
 };
-PJRT_DEFINE_STRUCT_TRAITS(PJRT_Gpu_Register_Custom_Call_Args, handler_execute);
+PJRT_DEFINE_STRUCT_TRAITS(PJRT_Gpu_Register_Custom_Call_Args, traits);
 
 // Registers a custom call.
 typedef PJRT_Error* PJRT_Gpu_Register_Custom_Call(
@@ -45,8 +47,11 @@ typedef PJRT_Error* PJRT_Gpu_Register_Custom_Call(
 typedef struct PJRT_Gpu_Custom_Call {
   PJRT_Extension_Base base;
   PJRT_NO_DISCARD PJRT_Gpu_Register_Custom_Call* custom_call;
+  // Like custom_call, but honors args->traits. Added in version 3; callers
+  // detect support by checking base.struct_size covers this member.
+  PJRT_NO_DISCARD PJRT_Gpu_Register_Custom_Call* custom_call_with_traits;
 } PJRT_Gpu_Custom_Call;
-PJRT_DEFINE_STRUCT_TRAITS(PJRT_Gpu_Custom_Call, custom_call);
+PJRT_DEFINE_STRUCT_TRAITS(PJRT_Gpu_Custom_Call, custom_call_with_traits);
 
 #ifdef __cplusplus
 }
--- a/xla/pjrt/c/pjrt_c_api_gpu_extension.h
+++ b/xla/pjrt/c/pjrt_c_api_gpu_extension.h
@@ -17,14 +17,14 @@ limitations under the License.
 #define XLA_PJRT_C_PJRT_C_API_GPU_EXTENSION_H_
 
 #include <stddef.h>
-
+#include <stdint.h>
 #include "xla/pjrt/c/pjrt_c_api.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#define PJRT_API_GPU_EXTENSION_VERSION 2
+#define PJRT_API_GPU_EXTENSION_VERSION 3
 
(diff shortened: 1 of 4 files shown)


tenhjo:
Rebased onto current main to fix the unrelated `check-contents` lint failure (a stale `absl::make_unique` in a TSL test file that had already been cleaned up upstream). The new commit `21afdfe` needs CI approval to run. @akuegel could you re-approve the workflows? Thanks!

tenhjo:
Reworked after the first CI round. The original version added `traits` as a struct-size-gated field on the existing `custom_call` entry point. That broke the JAX-at-HEAD GPU jobs: recompiling jax against the new headers grew its `struct_size` to cover `traits`, but jax's code predates the field and left it uninitialized — so the plugin read stack garbage and rejected valid `api_version=0` registrations (31 failures, all `UNIMPLEMENTED: Custom call traits are only supported for api_version 1.`). Every source-level consumer recompiling against the new headers would hit the same trap, silently for `api_version=1`.

The new shape avoids that entirely: `custom_call` keeps its exact pre-v3 behavior and never reads `traits`; a second entry point `custom_call_with_traits` (following `PJRT_FFI_Extension`'s appended-function pattern) is its only reader, and its callers set the field by construction. And since the plugin fills in the extension struct's `struct_size`, consumers can detect traits support at runtime rather than trusting the compile-time header version — so an old plugin fails loudly instead of silently dropping traits.

tenhjo:
The two ROCm AMD Instinct jobs are red for infra reasons unrelated to this change: XLA lost its self-hosted runner mid-job, and JAX's `multibackend_test_gpu` fails with `HIP_ERROR_InvalidImage` (a ROCm runner driver/image issue seen on unrelated PRs too). Neither touches `pjrt/c/` or custom calls. Everything else is green — both CUDA GPU L4 jobs, Hermetic ROCm, all CPU/Windows/ARM/OneAPI, and the lint jobs.

akuegel:
> The two ROCm AMD Instinct jobs are red for infra reasons unrelated to this change: XLA lost its self-hosted runner mid-job, and JAX's `multibackend_test_gpu` fails with `HIP_ERROR_InvalidImage` (a ROCm runner driver/image issue seen on unrelated PRs too). Neither touches `pjrt/c/` or custom calls. Everything else is green — both CUDA GPU L4 jobs, Hermetic ROCm, all CPU/Windows/ARM/OneAPI, and the lint jobs.

Yes, ROCM jobs are broken at head right now.

xla/pjrt/c/pjrt_c_api_gpu_extension.h
```diff
@@ -45,8 +47,11 @@ typedef PJRT_Error* PJRT_Gpu_Register_Custom_Call(
 typedef struct PJRT_Gpu_Custom_Call {
   PJRT_Extension_Base base;
   PJRT_NO_DISCARD PJRT_Gpu_Register_Custom_Call* custom_call;
+  // Like custom_call, but honors args->traits. Added in version 3; callers
+  // detect support by checking base.struct_size covers this member.
+  PJRT_NO_DISCARD PJRT_Gpu_Register_Custom_Call* custom_call_with_traits;
```
pschuh: Two function types sharing the same arg struct is weird. I feel like we should sort out what struct_size didn't work or duplicate the struct or something.

tenhjo: Yes the shared struct is ugly... but the second pointer is the only way to actually detect support.

Direction is the problem: `args->struct_size` is set by the caller, so it only tells the plugin what got passed. it can't tell the caller whether the plugin actually reads traits. old plugin + new caller → traits silently dropped
And `PJRT_Extension_Base` has no version field, just `struct_size`, so the only plugin→caller signal we've got is a new member showing up. Bumping `PJRT_API_GPU_EXTENSION_VERSION` does nothing at runtime. That's why it's a new function pointer and not a version-gated arg field.

The real problem is `traits` sitting in the shared args struct when only one function reads it. You could split it up: `custom_call_with_traits` gets its own `..._With_Traits_Args`, and `traits` comes back out of the original. 
old path stays and traits lives only where it is read.

pschuh: Hmm, but you could just have a bool (*supports_traits)(); method that you add to the PJRT_Gpu_Custom_Call struct and use that to test in the callee?

tenhjo: yeah fair, duplicating the whole args struct for one `uint32` is too heavy. let's just go single `custom_call` reading `args->traits` (gated by `struct_size`), which is basically your direction.

one tweak: support is static here, it is a property of the plugin build, not runtime. So we don't need a bool. The member being present (`struct_size` covering it) is already the signal, and bool (*)() is a bit off from the usual `PJRT_Error* f(Args*)` shape. So I would add a plain capability field rather than a probe. 

So: one custom_call reads traits + one small capability member for detection?

pschuh: Sure, the extra gating field sgtm.

pschuh (approved):

akuegel (approved):

akuegel (approved):

PR #45396, opened 2026-07-08
Drop redundant optional wrapping in send/recv conflict analysis

EylonKrause:
### Change
Cleanup in `GetAbstractReplicaGroups` (the send/recv branch). The source-target-pairs lookup assigned `Map::at()`'s returned `const std::string&` into a `std::optional<std::string>` and then `CHECK`'d `has_value()`, which is always true (the optional is constructed from a live reference). Use the string directly, drop the redundant `CHECK`, and remove the now-unused `<optional>` include.

The fail-loud `Map::at()` is kept: a device send/recv is expected to carry the source-target-pairs attribute here (device send/recv reach this via collective-permute decomposition, which always sets it), matching the sibling `CHECK(!sr->is_host_transfer())` invariant.

No functional change.

---

### DIFF
--- a/xla/service/collective_conflict_analysis.cc
+++ b/xla/service/collective_conflict_analysis.cc
@@ -17,7 +17,6 @@ limitations under the License.
 
 #include <algorithm>
 #include <cstdint>
-#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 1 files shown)


xla/service/collective_conflict_analysis.cc
```diff
@@ -101,11 +101,14 @@ void GetAbstractReplicaGroups(HloInstruction* instr,
...
     CHECK(!sr->is_host_transfer());
-    std::optional<std::string> source_target_pairs_str =
-        sr->frontend_attributes().map().at(kSendRecvSourceTargetPairsAttr);
-    CHECK(source_target_pairs_str.has_value());
+    const auto& attr_map = sr->frontend_attributes().map();
+    auto it = attr_map.find(kSendRecvSourceTargetPairsAttr);
+    // A device send/recv may legitimately carry no source-target-pairs
+    // attribute (the verifier does not require it pre-schedule); skip it rather
+    // than aborting via Map::at() on the missing key.
+    if (it == attr_map.end()) return;
```
allanrenucci: Do you have a concrete example where this is a problem? I am a bit fence with silently skipping the missing attribute as this may hide a bug.

Alternatively, feel free to drop the redundant CHECK and optional wrapping.

allanrenucci (changes requested):

EylonKrause:
You're right. I couldn't find a concrete case where a device send/recv legitimately reaches this without the source-target-pairs attribute — device send/recv here come from collective-permute decomposition, which always sets it, and the sibling `CHECK(!sr->is_host_transfer())` treats the device-only shape as an invariant already. So I dropped the silent skip.

I did keep the small cleanup you pointed at: the original assigned `Map::at()`'s string reference into a `std::optional` and then `CHECK`'d `has_value()`, which is always true. Now it uses the string directly (with the fail-loud `.at()` intact) and drops the redundant `CHECK` and the unused `<optional>` include. Thanks!

allanrenucci (changes requested):
LGTM but can you update the PR description and title accordingly

allanrenucci (approved):

PR #45401, opened 2026-07-08
[HLO] Rewrite a scalar-operand gather to a broadcast in GatherSimplifier

EylonKrause:
### Problem
`GatherSimplifier` crashed on a gather from a scalar (rank-0) operand: such a gather has empty `offset_dims`, which `IsSimplifiedGather` dereferenced out of bounds (and the pass could otherwise loop without making progress).

### Fix
Rather than treating a scalar-operand gather as already-simplified (which just skips it), eliminate it — mirroring `ScatterSimplifier`:
- `IsSimplifiedGather` returns `false` for a rank-0 operand (`operand_rank == 0`), which also avoids the out-of-bounds `offset_dims` dereference.
- `ExpandInstruction` rewrites a rank-0-operand gather into a `broadcast` of the scalar operand to the result shape (matching the existing zero-slice-size → broadcast path), so the gather is removed instead of being left in place.

Behavior is unchanged for all other gathers.

### Test
`RewritesScalarOperandGatherToBroadcast` asserts a scalar-operand gather is rewritten to `broadcast(%operand), dimensions={}`.

---

### DIFF
--- a/xla/hlo/transforms/simplifiers/gather_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/gather_simplifier.cc
@@ -49,6 +49,13 @@ absl::StatusOr<HloInstruction*> GatherSimplifier::ExpandInstruction(
   auto* operand = gather->operands()[0];
   auto* start_indices = gather->operands()[1];
 
+  // A gather from a scalar (rank-0) operand selects that same scalar for every
+  // index, so it is equivalent to broadcasting the operand to the result shape.
+  if (operand_rank == 0) {
+    return gather->AddInstruction(
+        HloInstruction::CreateBroadcast(gather->shape(), operand, {}));
+  }
+
   // Make the start_indices a two-dimensional tensor.
   ASSIGN_OR_RETURN(start_indices, TransformStartIndices(
                                       start_indices, dims.index_vector_dim()));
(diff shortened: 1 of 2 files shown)


allanrenucci (changes requested):
While the PR fixes the immediate out-of-bounds read and avoids an infinite loop, returning `true` passes the buck. To genuinely mirror `ScatterSimplifier` and uphold the purpose of the `GatherSimplifier` pass, the PR should instead:

* return `false` from `GatherSimplifier::IsSimplifiedGather` if `operand_rank == 0` (or `offset_dims` is empty).
* modify `GatherSimplifier::ExpandInstruction` to explicitly check for `operand_rank == 0`, completely eliminating the Gather by rewriting it into a `Broadcast` of the scalar operand.

EylonKrause:
Thanks — that's much better. Done:

- `IsSimplifiedGather` now returns `false` for a scalar (rank-0) operand instead of treating it as already simplified.
- `ExpandInstruction` rewrites a rank-0-operand gather into a `broadcast` of the operand to the result shape, mirroring the existing zero-slice-size → broadcast path, so the gather is eliminated rather than skipped.

Returning `false` for the scalar case also naturally avoids the out-of-bounds `offset_dims` dereference. Updated the test to check the gather is rewritten to `broadcast(%operand), dimensions={}`.

allanrenucci (commented):
Thanks. LGTM now but please update the PR description accordingly

EylonKrause:
Thanks! Updated the title and description to describe the broadcast rewrite.

allanrenucci (approved):

PR #45444, opened 2026-07-09
spmd-enzyme: Be more conservative when walking constant DUS chain

Pangoraw:
📝 Summary of Changes

This fixes a bug in enzyme-comms-opts gated (cc @wsmoses) spmd partitioner where it would too eagerly follow a constant dus chain. If a dus with constant starts had an update which is a constant dus as well, it would take that dus update as the update but that is only valid is that dus covers the whole update.

🎯 Justification

this caused a bug where the update value taken for the dus was wrong.

🚀 Kind of Contribution

🐛 Bug Fix

🧪 Unit Tests:

added a test where the update taken was wrong because the update dus was wrong.


Pangoraw:
Done in https://github.com/openxla/xla/commit/5aee800e63da39020c86da95e28f83caddc17f76

PR #45491, opened 2026-07-10
Fix ARM NEON GCC compat: replace vcalt_f32 with vclt_f32(vabs_f32())

### LINKED ISSUE #8247: ef57.cc does not compile for ARM with GCC
https://github.com/openxla/xla/commit/49f45a12bcca8edc7f215887dbb112b2639d8a97 introduced an implementation of `xla::ConvertF64ToEf57` for arm64 neon that does not compile with GCC: https://godbolt.org/z/aYfds9Y15. cc: @majnemer.

Functionhx:
Fixes #8247.

### DIFF
--- a/xla/ef57.cc
+++ b/xla/ef57.cc
@@ -92,7 +92,7 @@ void ConvertF64ToEf57(absl::Span<const double> input,
     float32x2_t x_lo_f32 = vcvt_f32_f64(x_lo_f64);
 
     uint32x2_t x_is_finite =
-        vcalt_f32(x_hi_f32, vdup_n_f32(std::numeric_limits<float>::infinity()));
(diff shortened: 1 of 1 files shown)


allanrenucci (commented):
It seems to me the original error was fixed in bdb4346. It is not clear to me what problem this PR addresses. Can you share the error if this code still does not compile with GCC?

Functionhx:
Closing this. The linked GCC failure is the `float32x2_t` argument passed to `vand_u32`, which was fixed by bdb43468. Current main compiles the reproducer with ARM64 GCC while retaining `vcalt_f32`, so this replacement isn't needed. Sorry for the noise.

PR #45514, opened 2026-07-10
[WIP] Fix autotuner backends const-correctness and thread-safety

nurmukhametov:
- Enforce const with mechanical changes across already immutable backends.
- FissionBackend: per-call rewriter pipeline via factory
- Triton/FissionBackend: per-call MLIRContext borrowed from pool


allanrenucci:
My understanding is that `mlir::MLIRContext` is mostly thread safe and has a mechanism to detect data races: https://github.com/llvm/llvm-project/blob/908b5931ce3d9630ae3b897bd4c523945038fdcd/mlir/lib/IR/MLIRContext.cpp#L154-L159

Do you know if we use non thread safe `MLIRContext` APIs? Maybe we should leave out `mlir::MLIRContext` as is, document that the code relies on `mlir::MLIRContext` being thread safe, use C++ `mutable` when needed for const correctness. WDYT?

PR #45534, opened 2026-07-11
[XLA:GPU] Add deviceless cuDNN convolution plan probing

kasper0406:
📝 Summary of Changes
Add CUDA stream-executor APIs for constructing and probing cuDNN convolution graphs without creating a GPU executor or cuDNN handle:

- `BuildConvolutionGraph()` translates convolution descriptors to the cuDNN frontend `graph::Graph` API.
- `CudnnSupportsConvolutionDeviceless()` probes whether cuDNN reports a supported plan for a target `DeviceDescription`.

PR #41021 enabled deviceless cuDNN graph preparation for fusion, attention, and custom-call graphs, while convolution runner enumeration remained handle-bound. This change provides the corresponding convolution probing primitive.

It is a prerequisite for compile-time convolution fallback decisions, such as PR #40021, during both JIT and AOT compilation. Compiler fallback integration remains separate work.

🎯 Justification
Some FP8 grouped convolutions - particularly configurations with fewer than 16 channels per group - have no valid cuDNN plans on RTX 5090 systems and can cause JAX failures. Deviceless probing allows the compiler to detect these configurations before execution (#40021).

🚀 Kind of Contribution
🐛 Bug Fix, ✨ New Feature

Not applicable

🧪 Unit Tests:

- Probe plain F32 forward convolutions using H100 and V100 target descriptions.
- Verify FP8 `FORWARD_GRAPH` support on H100 and rejection on V100.
- Verify that unmodeled true-convolution mode returns `Unimplemented`.
- Compare deviceless and live-GPU plan enumeration for F32/BF16 forward and grouped FP8 forward-graph convolutions.

🧪 Execution Tests:

No convolution kernels are executed. Some parity tests require a live GPU to enumerate handle-bound cuDNN plans.

### DIFF
--- a/xla/backends/gpu/transforms/BUILD
+++ b/xla/backends/gpu/transforms/BUILD
@@ -758,7 +758,6 @@ xla_cc_test(
     ],
     deps = [
         ":cublas_pad_for_gemms",
-        "//xla:xla_data_proto_cc",
         "//xla:xla_proto_cc",
         "//xla/hlo/ir:hlo",
         "//xla/hlo/testlib:hlo_hardware_independent_test_base",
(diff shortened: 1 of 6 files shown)


derdrdirk:
Instead of converting the convolution custom call to a cudnn graph could you rely on the experimental —xla_gpu_enable_conv_fusion pipeline?

xla/backends/gpu/transforms/cudnn_fusion_compiler_deviceless_test.cc
```diff
@@ -0,0 +1,433 @@
...
+    absl::string_view name;
+    absl::string_view hlo;
+    ConvolutionKind kind;
+  };
+  const KindCase kKindCases[] = {
+      {"fprop", kF32ForwardConvHlo, CONVOLUTION_KIND_FPROP},
+      {"wgrad", kF32BackwardFilterConvHlo, CONVOLUTION_KIND_WGRAD},
+      {"dgrad", kF32BackwardInputConvHlo, CONVOLUTION_KIND_DGRAD},
+  };
+  TF_ASSERT_OK_AND_ASSIGN(GpuTargetConfig construction_config,
```
EusebioDM: Nit: Use `ASSERT_OK_AND_ASSIGN` in this file instead. The TF_ prefixed macros are mostly deprecated.

xla/backends/gpu/transforms/BUILD
```diff
@@ -996,22 +1012,70 @@ xla_test(
     ],
 )
 
+xla_cc_test(
```
EusebioDM: As it is this test in the CI will be run with whatever GPU the worker has, which i think is L4 in OSS. 

I think in the `DevicelessSupportMatchesLivePlanEnumeration` test we want to test this against all the NVIDIA GPUs we support (instead of a random one).

To do this we can do something like this 
```
xla_test(
    name = "my_test",
    srcs = ["my_test.cc"],
    backends = ["gpu"],
    tags = ["cuda-only"],
    deps = [...],
)
```
which will run the test for all supported NVIDIA GPUs.

EusebioDM (changes requested):
Thanks for adding this, and specially the comprehensive tests!

The PR LGTM, just the comment about the test.

EusebioDM (approved):
Thanks for the changes!

derdrdirk (approved):
LGTM! Thanks!

EusebioDM:
FYI @kasper0406 @derdrdirk  I'm talking to NVIDIA wrt some of these crashes, and it looks like there are some faulty engines in versions < 9.19. 

We can use a errata filter (e.g. CudnnExecutionPlanEngineFilterStatic in cuda_dnn.cc) to filter those out, but we might need a separate one for deviceless. 

In any case I think we should get this PR merged, but I think we'll be able to get rid of the probing logic you added here once I we get the rid filters to use from the cudnn team.

PR #45559, opened 2026-07-12
[HLO] Return parse errors instead of CHECK-failing on malformed replica_groups=mesh[...]

EylonKrause:
### Problem
The HLO text parser aborts (via `CHECK`/`CHECK_EQ`) instead of returning a parse error on a malformed `replica_groups=mesh[...]` collective-device list. Because these come straight from user-provided HLO text (`ParseAndReturnUnverifiedModule` and friends), a malformed axis reference or device-id list crashes the process rather than producing a diagnostic. Three sites in `ParseAxisRef`/`ParseMesh`:

1. **Invalid axis ref** — `CHECK(absl::SimpleAtoi(axis_name_or_index, &axis_index))` fires when the axis reference is neither a known mesh axis name nor an integer:
   ```
   replica_groups=mesh['axis_0'=2,'axis_1'=2] {'bogus'}
   ```
2. **Axis index out of bounds** — `CHECK(axis_index >= 0 && axis_index < mesh.axis_names().size())` fires on an out-of-range (or negative) index:
   ```
   replica_groups=mesh['axis_0'=2,'axis_1'=2] {'5'}
   ```
3. **device_ids count mismatch** — `CHECK_EQ(devices.size(), device_ids_array.num_elements())` fires when the `device_ids=(...)` count doesn't match the mesh size:
   ```
   replica_groups=mesh['axis_0'=2,'axis_1'=2, device_ids=(0,1,2)] {'axis_0'}
   ```
   The `Array<int64_t>` is also constructed (allocating the product of the axis sizes) *before* the check, so a huge mesh with a mismatched device list can attempt an unbounded allocation.

### Fix
Return a `TokenError` from each site instead of `CHECK`-failing, matching how the rest of the parser reports malformed input. For (3), the device-id count is validated against an overflow-checked product of the axis sizes *before* the `Array` is allocated, so a mismatched or overflowing mesh yields a parse error rather than a bad-alloc or abort.

Valid `mesh[...]` lists are unaffected (the existing `AllReduceWithMeshAxesReplicaGroupList` round-trip test still passes).

### Tests
Three `HloParserTest` cases assert the malformed inputs above now return an error status instead of crashing.

---

### DIFF
--- a/xla/hlo/parser/hlo_parser.cc
+++ b/xla/hlo/parser/hlo_parser.cc
@@ -4190,11 +4190,24 @@ bool HloParserImpl::ParseMesh(std::optional<Mesh>& mesh) {
     if (!ParseToken(TokKind::kRparen, "expected ')' to end device_ids")) {
       return false;
     }
+    // Validate the device_ids count against the mesh size before allocating the
+    // device-id array, returning a parse error instead of CHECK-failing (or
+    // attempting an unbounded allocation for a huge, mismatched mesh).
+    int64_t num_elements = 1;
+    for (int64_t axis_size : axis_sizes) {
+      if (axis_size < 0 ||
+          (axis_size != 0 &&
+           num_elements > std::numeric_limits<int64_t>::max() / axis_size)) {
+        return TokenError("mesh size overflows int64");
+      }
+      num_elements *= axis_size;
+    }
+    if (static_cast<int64_t>(devices.size()) != num_elements) {
+      return TokenError(absl::StrFormat(
+          "Expected %d device_ids based on mesh axes, but got %d", num_elements,
+          devices.size()));
+    }
     Array<int64_t> device_ids_array(axis_sizes);
-    CHECK_EQ(devices.size(), device_ids_array.num_elements())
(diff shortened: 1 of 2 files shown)


xla/hlo/parser/hlo_parser.cc
```diff
@@ -4190,11 +4190,24 @@ bool HloParserImpl::ParseMesh(std::optional<Mesh>& mesh) {
     if (!ParseToken(TokKind::kRparen, "expected ')' to end device_ids")) {
       return false;
     }
+    // Validate the device_ids count against the mesh size before allocating the
+    // device-id array, returning a parse error instead of CHECK-failing (or
+    // attempting an unbounded allocation for a huge, mismatched mesh).
```
thcmbs: nit: please keep this concise, no need to mention what the previous impl. was doing.

thcmbs (commented):
Thanks.

EylonKrause:
Done — shortened the comment to just describe what the validation does.

thcmbs (approved):

PR #45560, opened 2026-07-12
Gate unsound inverse-function folds behind fast-math

### LINKED ISSUE #42732: XLA algebraic simplifier folds inverse function pairs, ignoring IEEE 754 intermediate overflow/underflow
## Summary

XLA's algebraic simplifier folds inverse function compositions (`log(exp(x)) → x`, `sqrt(x²) → |x|`) even when the intermediate value overflows or underflows in float32. This changes the observable result: eager mode produces `inf`/`-inf` (correct per IEEE 754), while XLA returns a finite value.

```python
import tensorflow as tf

# exp(89.0) overflows float32 → inf. log(inf) = inf.
x = tf.constant([89.0])

# Eager: inf (correct — intermediate overflows)
print(tf.math.log(tf.exp(x)).numpy())  # [inf]

# XLA: 89.0 (WRONG — simplifier bypasses overflow)
print(tf.function(lambda x: tf.math.log(tf.exp(x)),
                  jit_compile=True)(x).numpy())  # [89.]
```

Observed on TensorFlow 2.22.0-dev20260429.

## Scope

Three inverse-pair simplifications are affected:

### 1. `log(exp(x)) → x` when exp(x) overflows (x ≥ 89 for float32)

| x | exp(x) | Eager log(exp(x)) | XLA log(exp(x)) |
|---|---|---|---|
| 88.0 | 1.65e38 (normal) | 88.0 | 88.0 | OK |
| **89.0** | **inf** (overflow) | **inf** | **89.0** | BUG |
| 100.0 | inf | inf | 100.0 | BUG |
| 200.0 | inf | inf | 200.0 | BUG |

### 2. `sqrt(x²) → |x|` when x² overflows (|x| > ~1.84e19 for float32)

| x | x² | Eager sqrt(x²) | XLA sqrt(x²) |
|---|---|---|---|
| 1e19 | 1e38 (normal) | 1e19 | 1e19 | OK |
| **1e20** | **inf** (overflow) | **inf** | **1e20** | BUG |
| -1e20 | inf | inf | 1e20 | BUG |

### 3. `exp(log(x)) → x` when x is subnormal (GPU FTZ flushes intermediate to 0)

| x | log(x) | Eager exp(log(x)) | XLA exp(log(x)) |
|---|---|---|---|
| 1e-45 | -inf (after FTZ) | 0 | 1e-45 | BUG (GPU) |
| 1e-40 | -inf (after FTZ) | 0 | 1e-40 | BUG (GPU) |

### Gradient impact

The simplification also changes gradients:

```python
# Gradient of log(exp(x)) should be 1, but at x=89:
# - Eager: NaN (exp overflows → inf, log chain rule → 0*inf = NaN)
# - XLA: 1 (simplified to identity, gradient = 1)
```

## Practical Impact

The `log(exp(x))` pattern appears when computing `log_softmax` or `logsumexp` manually:

```python
# Manual log-sum-exp without numerical stabilization:
def unstable_logsumexp(x):
    return tf.math.log(tf.reduce_sum(tf.exp(x)))

x = tf.constant([89.0, 1.0, 2.0])
# Eager: inf (correctly signals overflow in exp)
# XLA: 89.0 (simplifier hides the overflow)
```

Code that checks for inf/NaN to detect numerical issues will miss the overflow under XLA. The overflow is real — it affects the precision of the result — but XLA's simplifier silently papers over it.

The `sqrt(x²)` pattern is common in norm computations:

```python
# Euclidean norm of a vector with large elements
x = tf.constant([1e20, 1e20, 1e20])
norm = tf.sqrt(tf.reduce_sum(x ** 2))
# Eager: inf (x² overflows)
# XLA: may simplify differently
```

## Root Cause

XLA's algebraic simplifier (in `xla/service/algebraic_simplifier.cc`) applies mathematical identities:
- `log(exp(x)) → x` (since log and exp are inverses)
- `sqrt(x²) → |x|` (since sqrt and square are inverses for non-negative values)

These identities hold over the reals, but not in IEEE 754 float32 when intermediate values overflow or underflow. The simplifier should check whether the intermediate computation would overflow before applying the transformation, or guard the transformation with a range check.

## Cross-framework Check

JAX (0.9.2) has the **same bug**: `jax.jit(lambda x: jnp.log(jnp.exp(x)))(89.0) = 89.0` while eager returns `inf`. This confirms the root cause is in XLA core's algebraic simplifier.

```python
import jax, jax.numpy as jnp
f = lambda x: jnp.log(jnp.exp(x))
print(f(jnp.float32(89.0)))           # inf (eager)
print(jax.jit(f)(jnp.float32(89.0)))  # 89.0 (jit — same bug)
```

## Environment

- TensorFlow 2.22.0-dev20260429
- JAX 0.9.2: **same bug** (confirms XLA core)
- Affects both CPU and GPU XLA
- Grappler: NOT affected (does not apply this simplification)
- Forward pass: diverges (eager=inf, XLA=finite)
- Gradient: diverges (eager=NaN, XLA=1)

## Recommendation

File in `openxla/xla`. The algebraic simplifier should not fold `log(exp(x)) → x` or `sqrt(x²) → |x|` when the intermediate value provably overflows or underflows in the target float type. The simplifier already has access to the element type and could check against the overflow threshold (`log(FLT_MAX) ≈ 88.72` for float32) before applying the transformation.

Sumu004:
`ln(exp(A)) => A` and `sqrt(A*A) => |A|` in `AlgebraicSimplifierVisitor` hold over the reals but not under IEEE 754: if `exp(A)` overflows to `inf`, eager evaluation gives `log(inf) = inf`, but the unconditional fold silently returns the finite `A` instead, changing both the value and the gradient. Reproduces identically under JAX and TF-XLA, confirming it's an XLA-level bug rather than either framework's own code.

This file already gates this class of transform behind fast-math elsewhere (e.g. `A - A => 0` is unsound for `NaN` and sits behind `enable_fast_math()`), so both folds now use that same guard rather than a new mechanism.

The originating report also mentions `exp(log(x))` losing precision under GPU flush-to-zero — couldn't find that fold in `HandleExp`, so it's out of scope here.

Fixes #42732

### DIFF
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
@@ -5229,9 +5229,19 @@ absl::Status AlgebraicSimplifierVisitor::HandleOr(HloInstruction* logical_or) {
 
 absl::Status AlgebraicSimplifierVisitor::HandleLog(HloInstruction* log) {
   // ln(exp(A)) => A
+  //
+  // This identity holds over the reals, but not for IEEE 754 floating point:
+  // if exp(A) overflows to inf (A >= ~88.7 for f32, ~709.8 for f64), eager
+  // evaluation gives log(inf) = inf, while the folded form silently returns
+  // the finite A instead -- changing both the forward value and the gradient
+  // (chain rule through the unfolded form gives NaN at the overflow point,
+  // vs. 1 for the folded identity). Restrict to fast-math mode, matching how
+  // this file already gates other identities that are unsound at special
+  // values (e.g. "A - A => 0" above, unsound for NaN).
   VLOG(10) << "trying transform [ln(exp(A)) => A]: " << log->ToString();
   HloInstruction *a, *b;
-  if (Match(log, m::Log(m::Exp(m::Op(&a)))) &&
+  if (options_.enable_fast_math() &&
+      Match(log, m::Log(m::Exp(m::Op(&a)))) &&
       ReplaceInstructionIfCompatible(log, a)) {
     return absl::OkStatus();
   }
@@ -9551,7 +9561,14 @@ absl::Status AlgebraicSimplifierVisitor::HandleSort(HloInstruction* sort) {
 absl::Status AlgebraicSimplifierVisitor::HandleSqrt(HloInstruction* sqrt) {
   VLOG(10) << "trying transform [sqrt(A*A) => |A|] " << sqrt->ToString();
   HloInstruction* sqrt_operand = sqrt->mutable_operand(0);
-  if (sqrt_operand->opcode() == HloOpcode::kMultiply &&
+  // sqrt(A*A) => |A| holds over the reals, but not for IEEE 754 floating
+  // point: if A*A overflows to inf (|A| > ~1.84e19 for f32), eager evaluation
+  // gives sqrt(inf) = inf, while the folded form silently returns the finite
+  // |A| instead. Restrict to fast-math mode, matching HandleLog's identical
+  // ln(exp(A)) => A guard immediately above in this file for the same
+  // reason.
+  if (options_.enable_fast_math() &&
+      sqrt_operand->opcode() == HloOpcode::kMultiply &&
       sqrt_operand->operand(0) == sqrt_operand->operand(1)) {
     PrimitiveType element_type = sqrt_operand->shape().element_type();
     // For 'A' of type C{64,128}, |A| has type F{32,64}, and the transformation
--- a/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
+++ b/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
@@ -5228,20 +5228,10 @@ absl::Status AlgebraicSimplifierVisitor::HandleOr(HloInstruction* logical_or) {
 }
 
 absl::Status AlgebraicSimplifierVisitor::HandleLog(HloInstruction* log) {
-  // ln(exp(A)) => A
-  //
-  // This identity holds over the reals, but not for IEEE 754 floating point:
-  // if exp(A) overflows to inf (A >= ~88.7 for f32, ~709.8 for f64), eager
-  // evaluation gives log(inf) = inf, while the folded form silently returns
-  // the finite A instead -- changing both the forward value and the gradient
-  // (chain rule through the unfolded form gives NaN at the overflow point,
-  // vs. 1 for the folded identity). Restrict to fast-math mode, matching how
-  // this file already gates other identities that are unsound at special
-  // values (e.g. "A - A => 0" above, unsound for NaN).
+  // ln(exp(A)) => A optimization is gated behind fast math flag.
   VLOG(10) << "trying transform [ln(exp(A)) => A]: " << log->ToString();
   HloInstruction *a, *b;
-  if (options_.enable_fast_math() &&
-      Match(log, m::Log(m::Exp(m::Op(&a)))) &&
+  if (options_.enable_fast_math() && Match(log, m::Log(m::Exp(m::Op(&a)))) &&
       ReplaceInstructionIfCompatible(log, a)) {
     return absl::OkStatus();
   }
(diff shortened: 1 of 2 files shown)


xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
```diff
@@ -5229,9 +5229,19 @@ absl::Status AlgebraicSimplifierVisitor::HandleOr(HloInstruction* logical_or) {
...
   // ln(exp(A)) => A
+  //
+  // This identity holds over the reals, but not for IEEE 754 floating point:
+  // if exp(A) overflows to inf (A >= ~88.7 for f32, ~709.8 for f64), eager
+  // evaluation gives log(inf) = inf, while the folded form silently returns
+  // the finite A instead -- changing both the forward value and the gradient
+  // (chain rule through the unfolded form gives NaN at the overflow point,
+  // vs. 1 for the folded identity). Restrict to fast-math mode, matching how
+  // this file already gates other identities that are unsound at special
+  // values (e.g. "A - A => 0" above, unsound for NaN).
```
thcmbs: This seemingly AI generated comment is extremely verbose. Please keep it concise and direct.

xla/hlo/transforms/simplifiers/algebraic_simplifier.cc
```diff
@@ -9551,7 +9561,14 @@ absl::Status AlgebraicSimplifierVisitor::HandleSort(HloInstruction* sort) {
 absl::Status AlgebraicSimplifierVisitor::HandleSqrt(HloInstruction* sqrt) {
   VLOG(10) << "trying transform [sqrt(A*A) => |A|] " << sqrt->ToString();
   HloInstruction* sqrt_operand = sqrt->mutable_operand(0);
-  if (sqrt_operand->opcode() == HloOpcode::kMultiply &&
+  // sqrt(A*A) => |A| holds over the reals, but not for IEEE 754 floating
+  // point: if A*A overflows to inf (|A| > ~1.84e19 for f32), eager evaluation
+  // gives sqrt(inf) = inf, while the folded form silently returns the finite
+  // |A| instead. Restrict to fast-math mode, matching HandleLog's identical
+  // ln(exp(A)) => A guard immediately above in this file for the same
+  // reason.
```
thcmbs: Same as above.

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -2957,12 +2957,44 @@ TEST_F(AlgebraicSimplifierTest, LnExp) {
...
+  AlgebraicSimplifier simplifier(options);
   ASSERT_TRUE(simplifier.Run(m.get()).value());
 
   EXPECT_EQ(computation->root_instruction(), param0);
 }
 
+// Regression test: ln(exp(A)) => A must NOT fire under default (non-fast-math)
+// options, since it silently hides f32 overflow in exp(A) for A >= ~88.7
+// (log(exp(89.0)) is inf under IEEE 754, but the fold would return the
+// finite 89.0).
```
thcmbs: Same as above. There is absolutely no reason to specify "since it silently hides f32 overflow in exp(A) for A >= ~88.7
// (log(exp(89.0)) is inf under IEEE 754, but the fold would return the
// finite 89.0)." here.

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -2957,12 +2957,44 @@ TEST_F(AlgebraicSimplifierTest, LnExp) {
...
+      HloInstruction::CreateUnary(r0f32, HloOpcode::kExp, param0));
+  builder.AddInstruction(
+      HloInstruction::CreateUnary(r0f32, HloOpcode::kLog, exp0));
+
+  auto computation = m->AddEntryComputationWithLayouts(builder.Build());
+
+  AlgebraicSimplifier simplifier(default_options_);
+  // The simplifier may return true/false depending on whether *other*
+  // instructions were simplified elsewhere in the graph; what matters is
+  // that this specific pattern is left untouched.
```
thcmbs: Please remove.

xla/hlo/transforms/simplifiers/algebraic_simplifier_test.cc
```diff
@@ -12714,7 +12750,11 @@ TEST_F(AlgebraicSimplifierTest, SquaredComplexSqrtIsFloat) {
 
   TF_ASSERT_OK_AND_ASSIGN(auto m, ParseAndReturnVerifiedModule(kModuleStr));
   SCOPED_TRACE("Before rewrite\n" + m->ToString());
+  // sqrt(A*A) => |A| is unsound under IEEE 754 when A*A overflows (eager
+  // gives sqrt(inf) = inf, the fold gives the finite |A|), so it's gated
+  // behind fast-math (see HandleSqrt).
```
thcmbs: "sqrt(A*A) => |A| optimization is gated behind fast math flag" would be enough.

thcmbs (commented):
Thanks for the contribution. 

I left some comments, and the same applies basically everywhere in the PR. 
Please cleanup the inline comments from your agent, or teach it to leave concise, useful inline comments.

akuegel (approved):

akuegel (changes requested):
I think you also need to adjust the test AlgebraicSimplifierTest.SqrtOfSelfMultiply

akuegel:
@Sumu004 in case you have not seen my latest reply, using direct mention

akuegel (approved):

Sumu004:
Fixed, SqrtOfSelfMultiply now uses fast-math options too. Pushed.

PR #45561, opened 2026-07-12
[HLO] Return parse errors instead of OOB on malformed sharding/window/conv attributes

EylonKrause:
### Problem
Three attribute handlers in the HLO text parser read or write out of bounds (one overflows the heap) on malformed-but-lexable HLO text, instead of returning a parse error. Because these come straight from user text (`ParseAndReturnUnverifiedModule`, `xla::ParseSharding`, `xla::ParseWindow`, `xla::ParseConvolutionDimensionNumbers`), malformed input crashes / corrupts memory rather than producing a diagnostic.

1. **Sharding tile assignment — heap-buffer-overflow.** `ParseSingleSharding` allocates the tile-assignment `Array` sized to the product of the bracketed tile dimensions, then copies the full device list with no count check:
   ```cpp
   auto tiles = std::make_shared<Array<int64_t>>(tile_assignment_dimensions);
   absl::c_copy(devices, tiles->begin());   // writes devices.size() into product(dims) slots
   ```
   `sharding={devices=[2,2]0,1,...,31}` allocates 4 slots but copies 32 devices — a heap-buffer-overflow write. (The analogous `mesh[... device_ids=...]` path is guarded; this classic `devices=[...]` path is not.)

2. **Window `rhs_reversal` — OOB read.** `ParseWindow` length-checks `stride`, `lhs_dilate`, `rhs_dilate` and `pad` against `size`, but not `rhs_reversal`, then reads `rhs_reversal[i]` for every dimension:
   ```cpp
   window->mutable_dimensions(i)->set_window_reversal(
       rhs_reversal.empty() ? false : (rhs_reversal[i] == 1));
   ```
   `window={size=2x2 rhs_reversal=1}` reads `rhs_reversal[1]` out of bounds.

3. **Conv `dim_labels` — OOB write.** `ParseConvolutionDimensionNumbers` sizes the spatial-dimensions repeated field to the number of digit chars, but bounds each digit by the *full* label length:
   ```cpp
   } else if (c < '0' + lhs.size() && c >= '0') {
     dnums->set_input_spatial_dimensions(c - '0', i);   // index can exceed the field size
   ```
   `dim_labels=9?????????_io->bf` (one spatial dim, digit value 9) writes at index 9 of a size-1 field. Same for the rhs and output labels.

### Fix
- Sharding: check `devices.size() == tiles->num_elements()` and return an error on mismatch before the copy.
- Window: add the missing `rhs_reversal` size check alongside the others.
- Conv: bound each digit by the spatial-dimensions field size (`input/kernel/output_spatial_dimensions_size()`) instead of the label length, and report the accurate valid range.

All three are behavior-preserving on valid HLO (the printer always emits matching counts) and convert the corruption/OOB into a returned error `Status`.

### Tests
Three `HloParserTest` cases assert the malformed inputs above now return an error instead of crashing.

---

### DIFF
--- a/xla/hlo/parser/hlo_parser.cc
+++ b/xla/hlo/parser/hlo_parser.cc
@@ -4846,6 +4846,12 @@ bool HloParserImpl::ParseSingleSharding(std::optional<HloSharding>& sharding,
             "non-maximal shardings must have more than one device assigned");
       }
       auto tiles = std::make_shared<Array<int64_t>>(tile_assignment_dimensions);
+      if (devices.size() != tiles->num_elements()) {
+        return Error(
+            loc, absl::StrFormat("sharding device count %d does not match the "
+                                 "tile assignment size %d",
+                                 devices.size(), tiles->num_elements()));
+      }
       absl::c_copy(devices, tiles->begin());
       sharding =
           subgroup_types.empty()
@@ -6545,6 +6551,9 @@ bool HloParserImpl::ParseWindow(Window* window, bool expect_outer_curlies) {
   if (!pad.empty() && pad.size() != size.size()) {
     return Error(loc, "expects 'pad=' has the same size as 'size='");
   }
+  if (!rhs_reversal.empty() && rhs_reversal.size() != size.size()) {
+    return Error(loc, "expects 'rhs_reversal=' has the same size as 'size='");
+  }
 
   for (int i = 0; i < size.size(); i++) {
     window->add_dimensions()->set_size(size[i]);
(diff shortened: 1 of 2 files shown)


xla/hlo/parser/hlo_parser.cc
```diff
@@ -4727,6 +4727,12 @@ bool HloParserImpl::ParseSingleSharding(std::optional<HloSharding>& sharding,
             "non-maximal shardings must have more than one device assigned");
       }
       auto tiles = std::make_shared<Array<int64_t>>(tile_assignment_dimensions);
+      if (devices.size() != tiles->num_elements()) {
+        return Error(
+            loc, absl::StrFormat("sharding device count %d does not match the "
```
thcmbs: nit: consider using StrCat s.t. we don't have to handle int types casting.

thcmbs (commented):
Thanks.

EylonKrause:
Done — switched to absl::StrCat so there is no format-specifier/int-type handling.

thcmbs:
Thanks, PTAL at the lint failure

EylonKrause:
Fixed the clang-format failure (reformatted the StrCat call to break after Error().

akuegel:
@EylonKrause can you please rebase?

EylonKrause:
Rebased onto current `main` (now at 3656b9b8).

Note for the reviewer: the rebase had a conflict in `hlo_parser_test.cc` because the negative tests from #45559 / #45821 / #45871 landed at the same insertion point. I resolved it by keeping all of those tests and appending this PR's three (`ShardingDeviceCountExceedsTileAssignment`, `WindowRhsReversalWrongSize`, `ConvDimLabelDigitExceedsSpatialDims`) — no upstream test was dropped or altered. The `hlo_parser.cc` changes merged cleanly and are unchanged.

akuegel (approved):

PR #45567, opened 2026-07-13
[XLA:GPU] Split command buffer VA remapping into its own allocator class

shawnwang18:
## 📝 Summary of Changes

This PR separates the command-buffer virtual-address remapping implementation from `GpuExecutableBufferAllocator` into a dedicated `GpuExecutableVaRemapAllocator` subclass. It is a structural refactor of the existing `SKIP_TEMP` behavior; no functional change is intended.

- `GpuExecutableBufferAllocator` is now the common base implementation for allocation resolution and the `ALWAYS_UPDATE` policy. It also remains the runtime fallback for `SKIP_TEMP` when VA remapping is unavailable.
- `ExecutionScope` is now polymorphic and is returned as `std::unique_ptr<ExecutionScope>`. Protected hooks such as `PrepareReservation` and `AllocateTransientBuffer` provide the extension points needed by address-policy-specific subclasses.
- The new `GpuExecutableVaRemapAllocator` owns the `SKIP_TEMP` implementation:
  - per-executor remapping state and synchronization;
  - deterministic VA reservation sizing and layout;
  - VMM-backed allocation of command-buffer-referenced temporary buffers;
  - persistent allocation indices combining constants with VA-remapped temporary allocations; and
  - synchronization of deferred VMM operations during allocator teardown.
- `GpuExecutableBufferAllocator::Create` selects the base allocator for `ALWAYS_UPDATE` and the VA-remapping subclass for `SKIP_TEMP`.
- GPU executable and PJRT call sites now use the factory and pointer-owned execution scopes.
- The BUILD target includes the new source/header pair and retains the direct dependencies required by both implementations.
- VMM test logging now enables both allocator translation units so existing assertions continue to observe the VA-reservation messages after the code move.

This PR is stacked on #45130, **[XLA:GPU] Update command buffers when allocation policy changes**. The follow-up policy is implemented separately in #45568.

## 🎯 Justification

The base allocator previously mixed general buffer-allocation mechanics with one specific VMM address-stability policy. Splitting those responsibilities makes the default path easier to reason about and gives additional update policies a narrow place to implement reservation, allocation, and execution behavior without further expanding the base class.

This cleanup is primarily preparation for profiled command-buffer allocation policies. Existing GPU command-buffer workloads should retain the same `ALWAYS_UPDATE` and `SKIP_TEMP` behavior.

## 🚀 Kind of Contribution

♻️ Cleanup

Not applicable. This is a behavior-preserving refactor and makes no performance claim.

## 🧪 Unit Tests:

- Updated `GpuExecutableTest.CommandBufferAllocationPolicy` to construct allocators through the new factory and exercise the pointer-owned `ExecutionScope`.
- Existing assertions continue to cover the persistent-allocation sets for `ALWAYS_UPDATE` and `SKIP_TEMP`, including the non-VMM fallback path.

## 🧪 Execution Tests:

- Existing `VmmTest` command-buffer tests continue to cover the two-GEMM `SKIP_TEMP` path, VA reservation reuse, and fallback when the configured allocator does not provide VMM remapping.
- `ScopedBufferAllocatorVLog` now captures logs from `gpu_executable_va_remap_allocator.cc`, preserving the execution-test assertions whose log statements moved with the implementation.


dimitar-asenov (approved):

ddunl (approved):

allanrenucci (approved):

xla/pjrt/gpu/BUILD
```diff
@@ -135,6 +135,8 @@ cc_library(
         "//xla/service:transfer_manager",
         "//xla/service/gpu:buffer_allocations",
         "//xla/service/gpu:gpu_constants",
+        "//xla/service/gpu:gpu_executable",
```
akuegel: Can you please revert the changes in this file?
Internally we need it this way to workaround some tagging issue. It "works" if the dependency is inside a "restricted" block (like if_cuda_or_rocm).

shawnwang18: thanks for the info, done with the revert

akuegel (approved):

PR #45683, opened 2026-07-14
pjrt: record Prepare failure before the two-phase barrier

nbhargava:
## The bug

`CommonPjRtLoadedExecutable::Execute`'s two-phase launch runs Prepare on every device in parallel, then each device reads a shared `failed` counter to decide whether to proceed to Launch. A `preparing==0` barrier ensures every device finishes Prepare before any device reads `failed` — but each device writes `failed++` *after* passing that barrier. So a device whose Prepare succeeded can read `failed == 0` (a failing peer has passed the barrier but not yet written), proceed to ExecuteLaunch, and block forever at the collective rendezvous waiting for a peer that never launches.

This happens during multi-GPU bringup: a transient CUDA allocation failure on one device during Prepare wedges the whole process at the NCCL rendezvous instead of returning an error.

## The fix

Move each device's `failed++` write to before its `preparing--`, still under the same mutex. Now `preparing==0` implies every device has already recorded its Prepare outcome, so any device that passes the barrier observes the final `failed` count. No new barrier is needed; the existing one now also covers the failure write.

## Commits

1. **pjrt: record Prepare failure before the two-phase barrier** — the fix proper. Self-contained (only reorders the failure-recording block in `Execute`) and safe to cherry-pick on its own.
2. **pjrt: add ExecuteLaunch hook and two-phase barrier regression test** — test-only. Adds `SetExecuteLaunchHook` (a null-checked `std::function` invoked at the top of ExecuteLaunch; never set in production) plus the regression test. No behavior change when the hook is unset.

## Tests

`TwoPhaseExecutePrepareFailureSkipsLaunch` (new, in `pjrt_stream_executor_client_test.cc`): a 2-device Host-platform StreamExecutor client runs Execute with one device given a wrong-size buffer so `CheckBufferCompatibilities` fails its Prepare. The new `SetExecuteLaunchHook` test hook counts how many devices reach ExecuteLaunch (phase 2). The test asserts that across 50 iterations with the failing device alternating, (a) Execute returns an error within 60s, and (b) **zero** devices reach ExecuteLaunch. (b) is the regression assertion — on a regressed barrier, a succeeding device reaches phase 2 on some schedules.

Covers barrier exit-path correctness without accelerator hardware; does not reproduce the original GPU-collective hang directly (that needs real collectives and timing control).

### On the testing approach (feedback welcome)

We're not fully settled on the hook being the right seam, so flagging the reasoning for reviewers. The invariant under test — no device enters Launch when any peer's Prepare failed — is not observable through the public API on the Host platform: `Execute` returns the same error status whether or not a device slipped into phase 2, and the real symptom (a collective rendezvous hang) only manifests with real collectives on multi-accelerator hardware. So the test observes phase-2 entry directly via a settable hook on the production class, following the ModuleHook pattern in `xla/service/llvm_compiler.h` (see also `GpuCompiler::SetAsmHook` and `DistributedRuntimeClient::Options::missed_heartbeat_callback`).

Alternatives considered: reproducing the actual hang (needs real collectives plus schedule control, and a hang-based oracle); mocking `ExecuteLaunch` (not virtual, and a test double would bypass the production barrier path under test); a friend-class test peer (phase-2 entry is a transient event, not inspectable state). If reviewers prefer, we're happy to rename the setter to `SetExecuteLaunchHookForTest` and/or add mutex-guarded set-once semantics like `LLVMCompiler`'s hooks.

### DIFF
--- a/xla/pjrt/se/pjrt_stream_executor_client_test.cc
+++ b/xla/pjrt/se/pjrt_stream_executor_client_test.cc
@@ -15,6 +15,7 @@ limitations under the License.
 
 #include "xla/pjrt/se/pjrt_stream_executor_client.h"
 
+#include <atomic>
 #include <cstddef>
 #include <cstdint>
 #include <memory>
@@ -30,7 +31,10 @@ limitations under the License.
 #include "absl/status/status.h"
 #include "absl/status/status_matchers.h"
 #include "absl/status/statusor.h"
+#include "absl/strings/str_format.h"
 #include "absl/synchronization/mutex.h"
+#include "absl/synchronization/notification.h"
+#include "absl/time/time.h"
 #include "xla/tsl/platform/status_macros.h"
 #include "xla/backends/cpu/target_machine_options.h"
 #include "xla/client/client_library.h"
@@ -43,12 +47,14 @@ limitations under the License.
 #include "xla/literal_comparison.h"
 #include "xla/literal_util.h"
 #include "xla/pjrt/abstract_tracked_device_buffer.h"
+#include "xla/pjrt/common_pjrt_client.h"
 #include "xla/pjrt/pjrt_client.h"
 #include "xla/pjrt/pjrt_compiler.h"
 #include "xla/pjrt/pjrt_executable.h"
 #include "xla/pjrt/plugin/xla_cpu/cpu_topology.h"
 #include "xla/pjrt/plugin/xla_cpu/cpu_topology_description.h"
 #include "xla/pjrt/se/local_device_state.h"
+#include "xla/service/computation_placer.h"
 #include "xla/service/platform_util.h"
 #include "xla/shape.h"
 #include "xla/shape_util.h"
@@ -60,6 +66,7 @@ limitations under the License.
 #include "xla/tsl/platform/env.h"
 #include "xla/tsl/platform/statusor.h"
 #include "xla/xla_data.pb.h"
+#include "tsl/platform/casts.h"
 #include "tsl/platform/path.h"
 
 namespace xla {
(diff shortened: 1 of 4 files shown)


xla/pjrt/se/pjrt_stream_executor_client_test.cc
```diff
@@ -349,5 +397,102 @@ TEST(PjRtStreamExecutorClientTest, MakeAllocationReadyEventAsync) {
...
+  // The pre-fix race is schedule-dependent (it fires only when a succeeding
+  // device is last to the barrier), so repeat with the failing device
+  // alternating. On Host-SE each iteration is microseconds.
+  constexpr int kIterations = 50;
+
+  auto client_or = GetClientWithDevices(kNumDevices);
+  if (absl::IsUnavailable(client_or.status())) {
+    GTEST_SKIP() << client_or.status();
+  }
+  TF_ASSERT_OK_AND_ASSIGN(auto client, std::move(client_or));
```
akuegel: Please use ASSERT_OK_AND_ASSIGN

xla/pjrt/se/pjrt_stream_executor_client_test.cc
```diff
@@ -349,5 +397,102 @@ TEST(PjRtStreamExecutorClientTest, MakeAllocationReadyEventAsync) {
...
+  Shape wrong_shape = ShapeUtil::MakeShape(F32, {2});
+
+  CompileOptions compile_options;
+  compile_options.executable_build_options.set_num_replicas(kNumDevices);
+  DeviceAssignment assignment(kNumDevices, /*computation_count=*/1);
+  for (int i = 0; i < kNumDevices; ++i) {
+    assignment(i, 0) = i;
+  }
+  compile_options.executable_build_options.set_device_assignment(assignment);
+  TF_ASSERT_OK_AND_ASSIGN(
```
akuegel: Please use ASSERT_OK_AND_ASSIGN

xla/pjrt/se/pjrt_stream_executor_client_test.cc
```diff
@@ -349,5 +397,102 @@ TEST(PjRtStreamExecutorClientTest, MakeAllocationReadyEventAsync) {
...
+  tensorflow::down_cast<CommonPjRtLoadedExecutable*>(executable.get())
+      ->SetExecuteLaunchHookForTesting([&](PjRtDevice*) {
+        launch_calls.fetch_add(1, std::memory_order_relaxed);
+      });
+
+  std::vector<std::unique_ptr<PjRtBuffer>> ok_bufs(kNumDevices);
+  std::vector<std::unique_ptr<PjRtBuffer>> wrong_bufs(kNumDevices);
+  for (int d = 0; d < kNumDevices; ++d) {
+    auto* mem = *client->addressable_devices()[d]->default_memory_space();
+    TF_ASSERT_OK_AND_ASSIGN(ok_bufs[d],
```
akuegel: Please use ASSERT_OK_AND_ASSIGN

xla/pjrt/se/pjrt_stream_executor_client_test.cc
```diff
@@ -349,5 +397,102 @@ TEST(PjRtStreamExecutorClientTest, MakeAllocationReadyEventAsync) {
...
+                            client->CreateUninitializedBuffer(shape, mem));
+    TF_ASSERT_OK_AND_ASSIGN(
```
akuegel: Please use ASSERT_OK_AND_ASSIGN

akuegel (approved):

xla/pjrt/se/pjrt_stream_executor_client_test.cc
```diff
@@ -15,11 +15,13 @@ limitations under the License.
 
 #include "xla/pjrt/se/pjrt_stream_executor_client.h"
 
+#include <atomic>
 #include <cstdint>
 #include <memory>
 #include <numeric>
 #include <optional>
 #include <string>
+#include <thread>
```
golechwierowicz: We cannot include <thread> internally. Can you remove it or rewrite the test such that this include is not needed?

nbhargava: I took a pass using `tsl::Thread` which seems to be used elsewhere in the project? Please let me know if this is ill-advised.

akuegel (approved):

akuegel:
@nbhargava Can you please rebase?

golechwierowicz (approved):

golechwierowicz:
This causes tests on CPU to fail, please fix: https://github.com/openxla/xla/actions/runs/30252937637/job/89934988240

akuegel (approved):

PR #45735, opened 2026-07-15
[ROCm] Pass through hsa env variables through the lit config

alekstheod:
📝 Summary of Changes
Pass through rocm related hsa env variables in lit.cfg.py script.

🎯 Justification
Some rocm ci jobs are using env variables to customize the test execution.
lit tests are hermetic and do not leak env variables used by rocm. 
We explicitly pass through these env variables now.

🚀 Kind of Contribution

Not relevant

🧪 Unit Tests:
CI

🧪 Execution Tests:
CI

### DIFF
--- a/xla/lit.cfg.py
+++ b/xla/lit.cfg.py
@@ -52,6 +52,13 @@ for env in [
     "LLVM_COVERAGE_FILE",
     "GCOV_PREFIX",
     "GCOV_PREFIX_STRIP",
+    # ROCm related env variables
+    "HSA_MODEL_TOPOLOGY",
+    "HSA_MODEL_LIB",
+    "HSA_KMT_MODEL_GPUVM_BASE",
+    "HSA_ENABLE_SDMA",
+    "HSA_ENABLE_INTERRUPT",
+    "LD_LIBRARY_PATH",
 ] + extra_env_flags:
   value = os.environ.get(env)
   if value:


chsigg:
Can you explain what these do and why they are needed in your CI jobs?
Is LD_LIBRARY_PATH really needed? That might be asking for trouble because users have it set.

alekstheod:
> Can you explain what these do and why they are needed in your CI jobs? Is LD_LIBRARY_PATH really needed? That might be asking for trouble because users have it set.

We do use a gpu emulator which works with the api interception mechanism. For that we need to set certain env variables. This is used for our preparation for a next gpu release.

alekstheod:
> Can you explain what these do and why they are needed in your CI jobs? Is LD_LIBRARY_PATH really needed? That might be asking for trouble because users have it set.

LD_LIBRARY_PATH Is required unfortunately we need to point where to load the libs from (emulator dir).

chsigg:
I'm not satisfied with picking up LD_LIBRARY_PATH. I asked Gemini and it focused on config.environment. Maybe you can also set the LD_LIBRARY_PATH as part of the RUN command.

> I'm reviewing https://github.com/openxla/xla/pull/45735. It seems there must be a better way to pass LD_LIBRARY_PATH to the lit RUN commands without blindly copying from the run environment.

You're right to flag that. Blindly copying `LD_LIBRARY_PATH` (or similar host environment variables) into a Lit test environment is a known anti-pattern in Bazel-based projects like OpenXLA.

It breaks hermeticity. When you pull `LD_LIBRARY_PATH` from the host, the test outcomes become dependent on the local machine's state. This can poison Bazel's remote cache or allow tests to pass locally by picking up system libraries while failing in CI (or vice-versa).

Depending on how the ROCm libraries are being provisioned in the XLA build, here are three better ways to handle this that you can suggest in your review:

## 1. Use RPATH / RUNPATH (Most Robust)

If the binaries being executed in the Lit `RUN` commands (e.g., `xla-opt`, `llc`) and the shared libraries they depend on are both built or managed by Bazel, the best approach is to avoid `LD_LIBRARY_PATH` entirely.

The binaries should be linked with a relative `RPATH` (e.g., `-Wl,-rpath,'$ORIGIN/../lib'`) that points to the relative location of the shared libraries in the Bazel runfiles tree. This makes the binaries self-locating and inherently hermetic without any Lit configuration changes.

## 2. Construct the Path via Bazel Runfiles

If modifying `RPATH` isn't feasible (which is common for precompiled, third-party vendor binaries), you should construct the `LD_LIBRARY_PATH` dynamically from the test's sandboxed runfiles, rather than the host system.

Ensure the required ROCm `.so` files are included in the test's `data` attribute. Then, in `lit.cfg.py`, build the path using Bazel's `RUNFILES_DIR`:

```python
runfiles_dir = os.environ.get('RUNFILES_DIR')
if runfiles_dir:
    # Construct the path relative to the hermetic Bazel workspace
    hermetic_lib_path = os.path.join(runfiles_dir, 'rocm_repository_name/lib') 
    config.environment['LD_LIBRARY_PATH'] = hermetic_lib_path

```

## 3. Explicit Injection via the BUILD File

If the project currently relies on a system-installed ROCm (e.g., hardcoded in `/opt/rocm/lib`) rather than a hermetic Bazel repository, you still shouldn't slurp the entire host `LD_LIBRARY_PATH`, as it pulls in unrelated host clutter.

Instead, isolate the specific path requirement in the `BUILD` file using the `env` attribute, and have Lit read only that explicit variable:

**In the BUILD file:**

```starlark
# Pass the explicit dependency path, not the whole host environment
env = {"ROCM_LIB_PATH": "/opt/rocm/lib"},

```

**In lit.cfg.py:**

```python
# Map the explicit variable to the Lit environment
if 'ROCM_LIB_PATH' in os.environ:
    config.environment['LD_LIBRARY_PATH'] = os.environ['ROCM_LIB_PATH']

```

**Review recommendation:** I'd suggest pushing back on the direct `os.environ` copy and asking if the required HSA/ROCm libraries can be resolved via `RUNFILES_DIR` (Approach 2). If they are strictly system dependencies right now, Approach 3 restricts the blast radius.

alekstheod:
> I'm not satisfied with picking up LD_LIBRARY_PATH. I asked Gemini and it focused on config.environment. Maybe you can also set the LD_LIBRARY_PATH as part of the RUN command.
> 
> > I'm reviewing #45735. It seems there must be a better way to pass LD_LIBRARY_PATH to the lit RUN commands without blindly copying from the run environment.
> 
> You're right to flag that. Blindly copying `LD_LIBRARY_PATH` (or similar host environment variables) into a Lit test environment is a known anti-pattern in Bazel-based projects like OpenXLA.
> 
> It breaks hermeticity. When you pull `LD_LIBRARY_PATH` from the host, the test outcomes become dependent on the local machine's state. This can poison Bazel's remote cache or allow tests to pass locally by picking up system libraries while failing in CI (or vice-versa).
> 
> Depending on how the ROCm libraries are being provisioned in the XLA build, here are three better ways to handle this that you can suggest in your review:
> ## 1. Use RPATH / RUNPATH (Most Robust)
> 
> If the binaries being executed in the Lit `RUN` commands (e.g., `xla-opt`, `llc`) and the shared libraries they depend on are both built or managed by Bazel, the best approach is to avoid `LD_LIBRARY_PATH` entirely.
> 
> The binaries should be linked with a relative `RPATH` (e.g., `-Wl,-rpath,'$ORIGIN/../lib'`) that points to the relative location of the shared libraries in the Bazel runfiles tree. This makes the binaries self-locating and inherently hermetic without any Lit configuration changes.
> ## 2. Construct the Path via Bazel Runfiles
> 
> If modifying `RPATH` isn't feasible (which is common for precompiled, third-party vendor binaries), you should construct the `LD_LIBRARY_PATH` dynamically from the test's sandboxed runfiles, rather than the host system.
> 
> Ensure the required ROCm `.so` files are included in the test's `data` attribute. Then, in `lit.cfg.py`, build the path using Bazel's `RUNFILES_DIR`:
> 
> ```python
> runfiles_dir = os.environ.get('RUNFILES_DIR')
> if runfiles_dir:
>     # Construct the path relative to the hermetic Bazel workspace
>     hermetic_lib_path = os.path.join(runfiles_dir, 'rocm_repository_name/lib') 
>     config.environment['LD_LIBRARY_PATH'] = hermetic_lib_path
> ```
> 
> ## 3. Explicit Injection via the BUILD File
> 
> If the project currently relies on a system-installed ROCm (e.g., hardcoded in `/opt/rocm/lib`) rather than a hermetic Bazel repository, you still shouldn't slurp the entire host `LD_LIBRARY_PATH`, as it pulls in unrelated host clutter.
> 
> Instead, isolate the specific path requirement in the `BUILD` file using the `env` attribute, and have Lit read only that explicit variable:
> 
> **In the BUILD file:**
> 
> ```python
> # Pass the explicit dependency path, not the whole host environment
> env = {"ROCM_LIB_PATH": "/opt/rocm/lib"},
> ```
> 
> **In lit.cfg.py:**
> 
> ```python
> # Map the explicit variable to the Lit environment
> if 'ROCM_LIB_PATH' in os.environ:
>     config.environment['LD_LIBRARY_PATH'] = os.environ['ROCM_LIB_PATH']
> ```
> 
> **Review recommendation:** I'd suggest pushing back on the direct `os.environ` copy and asking if the required HSA/ROCm libraries can be resolved via `RUNFILES_DIR` (Approach 2). If they are strictly system dependencies right now, Approach 3 restricts the blast radius.

Ok I will investigate, maybe I can pass a custom rpath instead.

dimitar-asenov:
@alekstheod, marking this PR as draft for now. Please mark as ready for review once you make the requested changes.

PR #45753, opened 2026-07-15
[XLA:GPU] Enforce the LHS scheduler memory limit as a soft-capped skip filter

shawnwang18:
# Enforce the LHS scheduler memory limit as a soft-capped skip filter

## Problem

The latency-hiding scheduler's `memory_limit` only participates in pairwise candidate-comparison tie-breakers (`kDecreaseMemoryOverLimit`, `kMemoryPeakOverLimit` in `ReadySetLt::MemoryPressurePolicy`); it is never enforced as a constraint on which node may be scheduled. On large unrolled graphs the aggressive latency-hiding policies can commit many long-lived buffers as overlap filler before the comparator rules engage, producing schedules whose estimated peak far exceeds the configured limit.

Concretely, on MaxText Llama3-8B with `scan_layers=false` (32 unrolled layers, context parallelism 8, sequence length 128K, 8xH200), the scheduler defers the weight-gradient GEMMs of the unrolled backward pass to the end of the schedule while advancing through hundreds of data-gradient GEMMs. Each deferred weight gradient keeps its saved forward activation and incoming activation gradient alive, so temporary memory grows roughly linearly with the number of unrolled layers:

| Configuration | Temporary memory | Result |
|---|---|---|
| scan=false, LHS enabled | 290 GB | OOM |
| scan=true, LHS enabled | 62 GB | fits |
| scan=false, LHS disabled | 65 GB | fits |

The scheduler itself estimated a peak of ~329 GB against a configured limit of ~95.5 GB and accepted the schedule anyway.

## Fix

Add `SchedulerConfig.enforce_memory_limit`, gated behind a new debug option `xla_gpu_experimental_lhs_enforce_memory_limit` (default off):

- In `DefaultSchedulerCore::FindAndExtractBestNodeAvailable`, skip any ready node whose projected net memory-pressure increase would push tracked usage above `memory_limit`, as long as any alternative node is available (new `SkipNodeReason::kExceedsMemoryLimit`, mirroring the existing overlap-limit skip).
- **Soft cap / progress guarantee:** if no node was selectable only because of memory skips, the selection is retried once with the filter disabled, falling back to the existing over-limit min-increase comparator rules. Scheduling always makes progress and can never fail due to the filter.
- The projected increase reuses `ReadySetLt::GetMemoryPressureChanges` (moved to public), so async-done nodes are judged with their paired async-start folded in, consistent with the comparator. Net-negative or neutral candidates are never skipped, so the filter cannot block a memory-reducing schedule.
- Nodes scheduled through the annotation-group path are not filtered (documented limitation).

Under memory pressure this stops the scheduler from grabbing far-away weight-gradient GEMMs as overlap filler; each one is instead scheduled at its dataflow-forced position (the shared incoming-gradient operand forces `WG_i` before the climb past layer i), which reproduces the memory-friendly per-layer interleaving. With headroom, the filter never fires and latency hiding is unaffected.

## Testing

- `EnforceMemoryLimitAvoidsOverlapAboveLimit`: reuses the `RerunWithSmallerMemoryLimit` module; with the flag on, the scheduler reaches the memory-friendly order (peak 88 bytes <= 110-byte limit) in a single pass, where the flag-off schedule peaks at 136 bytes.
- `EnforceMemoryLimitNoEffectWithHeadroom`: with a generous limit the filter never fires and the collective still overlaps independent work.
- `EnforceMemoryLimitTinyLimitStillSchedules`: with a 1-byte limit every candidate is skipped; the soft-cap fallback still produces a complete schedule.
- `EnforceMemoryLimitDelaysWgradLikeSinks`: miniature two-layer unrolled backward pass (GPU-like aggressive policies); under enforcement the two saved activations' live ranges must not overlap.
- Full `//xla/service:latency_hiding_scheduler_test` suite passes (98/98). No behavior change with the flag off.

### DIFF
--- a/xla/debug_options_flags.cc
+++ b/xla/debug_options_flags.cc
@@ -515,6 +515,7 @@ DebugOptions DefaultDebugOptionsIgnoringFlags() {
   opts.set_xla_gpu_experimental_thunk_buffer_debug_module_outputs(false);
   opts.set_xla_gpu_enable_gxl_ragged_all_to_all(false);
   opts.set_xla_gpu_gxl_scratch_size_bytes(64 * 1024 * 1024);
+  opts.set_xla_gpu_experimental_lhs_enforce_memory_limit(false);
 
   // Disable float checks.
   opts.set_xla_gpu_detect_nan(DebugOptions::DETECTION_MODE_NONE);
@@ -3266,6 +3267,14 @@ void MakeDebugOptionsFlags(std::vector<tsl::Flag>* flag_list,
       int64_setter_for(&DebugOptions::set_xla_gpu_gxl_scratch_size_bytes),
       debug_options->xla_gpu_gxl_scratch_size_bytes(),
       "Size in bytes of the scratch buffer for GXL collectives."));
+  flag_list->push_back(tsl::Flag(
+      "xla_gpu_experimental_lhs_enforce_memory_limit",
+      bool_setter_for(
+          &DebugOptions::set_xla_gpu_experimental_lhs_enforce_memory_limit),
+      debug_options->xla_gpu_experimental_lhs_enforce_memory_limit(),
+      "If true, the latency-hiding scheduler avoids scheduling nodes that "
+      "would push memory usage above the scheduler memory limit when any "
+      "alternative node is available."));
   flag_list->push_back(tsl::Flag(
       "xla_gpu_experimental_use_ragged_dot_grouped_gemm",
       bool_setter_for(
--- a/xla/service/gpu/gpu_hlo_schedule.cc
+++ b/xla/service/gpu/gpu_hlo_schedule.cc
@@ -657,6 +657,8 @@ absl::Status RunLatencyHidingSchedulerPasses(
       memory_limit,
       options.xla_gpu_experimental_parallel_collective_overlap_limit(),
       options.xla_gpu_experimental_parallel_async_compute_limit());
+  config.enforce_memory_limit =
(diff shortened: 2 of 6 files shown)


shawnwang18:
The proposal in this PR is not working well. 


  The first design skips a ready candidate only when:

  current_memory + candidate_immediate_delta > limit

  This is too reactive. The important wrong operator can have a small or even favorable immediate memory delta. Its harmful effect appears only after additional scheduling decisions.

  If every candidate exceeds the limit, the scheduler also returns to the old chooser to guarantee progress. Consequently, the limit is guidance rather than a global memory guarantee.

  ### Reset-and-replay backtracking

  We considered resetting to an earlier scheduling state, excluding the previous choice, and replaying the schedule.

  Finding the relevant decision could require exploring combinations of choices across a very large DAG. The search space can grow exponentially, making compiler latency unacceptable. We therefore dropped this approach.

  ### Five decreasing memory-limit attempts

  We tried five complete LHS runs with guidance limits:

  79.59 GB → 71.63 GB → 64.47 GB → 58.02 GB → 52.22 GB

  Every run produced the same approximately 339.07 GB schedule.

  The crucial choices happened while all relevant candidates still appeared safe. They were selected by latency heuristics such as kAsyncDepth, so tightening the immediate limit did not alter them.

  After all attempts failed validation, we restored the pre-LHS schedule. That reduced compiled temporary memory to 64.95 GB and completed a real training step without OOM. However, the success came from discarding the LHS schedule, not from finding good memory-safe overlap.

PR #45757, opened 2026-07-15
[ROCm] Added support for waves_per_eu function attribute

zoranjovanovic-ns:
📝 Summary of Changes
Parses an optional waves_per_eu value from the serialized_metadata JSON on Triton custom calls and threads it through to the ROCm amdgpu-waves-per-eu LLVM function attribute when emitting the kernel.

🎯 Justification
Makes possible to tune kernel occupancy (waves_per_eu), form ROCm/JAX Pallas, for Triton kernels using the existing metadata mechanism

🚀 Kind of Contribution

🧪 Unit Tests:
triton_call_test.cc

### DIFF
--- a/xla/service/gpu/triton_call.cc
+++ b/xla/service/gpu/triton_call.cc
@@ -73,16 +73,18 @@ TritonCall TritonCall::Parse(absl::string_view backend_config,
   int64_t waves_per_eu = 0;
   if (auto attr = attrs.getAs<mlir::StringAttr>("serialized_metadata")) {
     llvm::Expected<llvm::json::Value> parsed =
-        llvm::json::parse(attr.getValue().str());
+        llvm::json::parse(attr.getValue());
     if (parsed) {
       if (llvm::json::Object* obj = parsed->getAsObject()) {
         if (std::optional<llvm::StringRef> v = obj->getString("waves_per_eu")) {
           int64_t tmp = 0;
-          if (llvm::to_integer(*v, tmp)) {
+          if (llvm::to_integer(*v, tmp) && tmp > 0) {
             waves_per_eu = tmp;
           }
         } else if (std::optional<int64_t> n = obj->getInteger("waves_per_eu")) {
-          waves_per_eu = *n;
+          if (*n > 0) {
+            waves_per_eu = *n;
+          }
         }
       }
     } else {
--- a/xla/service/gpu/triton_call.cc
+++ b/xla/service/gpu/triton_call.cc
@@ -16,10 +16,15 @@ limitations under the License.
 #include "xla/service/gpu/triton_call.h"
 
 #include <cstdint>
+#include <optional>
 #include <utility>
 #include <vector>
 
 #include "absl/strings/string_view.h"
+#include "llvm/ADT/StringExtras.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Error.h"
+#include "llvm/Support/JSON.h"
 #include "mlir/AsmParser/AsmParser.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/MLIRContext.h"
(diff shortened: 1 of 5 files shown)


xla/service/gpu/triton_call_test.cc
```diff
@@ -0,0 +1,75 @@
+/* Copyright 2026 The TensorFlow Authors. All Rights Reserved.
```
i-chaochen: The OpenXLA Authors

thcmbs:
@superbobry Can you please have a look? I'm not sure if this is the right path for AMD to use (and whether it aligns with `jax-triton` path.

superbobry:
This custom call is only used by Pallas Triton, which is deprecated and will be removed soon. 

Perhaps, you want to contribute a similar change to jax-triton instead?

zoranjovanovic-ns:
> This custom call is only used by Pallas Triton, which is deprecated and will be removed soon.
> 
> Perhaps, you want to contribute a similar change to jax-triton instead?

I am aware of the fact that 'USE_NEW_CUSTOM_CALL' will be set to True in the future and that afterwards we will need different solution - but until then, this is the simplest and probably only mechanism that we have for setting this function attribute.

When USE_NEW_CUSTOM_CALL is set to True, call will be handled through pjrt, but it will require passing waves_per_eu value there and extending **PJRT_Triton_Compile_Args** C ABI struct with waves_per_eu field - requiring a struct-size/version bump.

We do have plan to require changes in **PJRT_Triton_Compile_Args** C ABI - not only for waves_per_eu field, plan is to do that before USE_NEW_CUSTOM_CALL is switched to True by default - but before this happens, we would like to provide an option for setting waves_per_eu attribute.

zoranjovanovic-ns:
> This custom call is only used by Pallas Triton, which is deprecated and will be removed soon.

@superbobry Do you mean that "Pallas Triton" or "custom call" is deprecated ?

akuegel:
> > This custom call is only used by Pallas Triton, which is deprecated and will be removed soon.
> 
> @superbobry Do you mean that "Pallas Triton" or "custom call" is deprecated ?

Pallas Triton is deprecated: https://github.com/jax-ml/jax/blob/main/jax/_src/pallas/pallas_call.py#L956-L963

Custom call is not deprecated

xla/service/gpu/triton_call.cc
```diff
@@ -61,11 +66,35 @@ TritonCall TritonCall::Parse(absl::string_view backend_config,
...
     }
   }
+  // The optional `serialized_metadata` attribute carries a JSON object emitted
+  // from JAX's `pl.pallas_call(..., metadata=...)`. It is used to pass the
+  // ROCm `waves_per_eu`. Any parse failure leaves `waves_per_eu`
+  // unset (0).
+  int64_t waves_per_eu = 0;
+  if (auto attr = attrs.getAs<mlir::StringAttr>("serialized_metadata")) {
+    llvm::Expected<llvm::json::Value> parsed =
+        llvm::json::parse(attr.getValue().str());
```
akuegel: The `.str()` call should not be necessary. `llvm:json::parse` handles StringRef just fine.

xla/service/gpu/triton_call.cc
```diff
@@ -61,11 +66,35 @@ TritonCall TritonCall::Parse(absl::string_view backend_config,
...
+    if (parsed) {
+      if (llvm::json::Object* obj = parsed->getAsObject()) {
+        if (std::optional<llvm::StringRef> v = obj->getString("waves_per_eu")) {
+          int64_t tmp = 0;
+          if (llvm::to_integer(*v, tmp)) {
```
akuegel: Maybe also check that `tmp > 0`?

xla/service/gpu/triton_call.cc
```diff
@@ -61,11 +66,35 @@ TritonCall TritonCall::Parse(absl::string_view backend_config,
...
+            waves_per_eu = tmp;
+          }
+        } else if (std::optional<int64_t> n = obj->getInteger("waves_per_eu")) {
+          waves_per_eu = *n;
```
akuegel: Maybe also check that `*n > 0`?

akuegel (commented):
As far as I understand, the main concern with this PR is that its functionality will be broken once Pallas Triton is removed. Given that internal workloads are unaffected by this change, we can likely still accept the change, but please be aware that this might break at any time.

zoranjovanovic-ns:
@akuegel Updated PR.

akuegel (approved):

superbobry:
Sorry for the delay.

The custom call will be removed 6 months after Pallas Triton is removed.

PR #45767, opened 2026-07-15
[ROCm] Add support of data files dependency to a wheel, required for rocm he…

alekstheod:
…rmetic builds

📝 Summary of Changes
Add support of transitive data dependency to the wheel targets

🎯 Justification
Rocm builds require rocm libs in its runfiles dir to be able to
be built executed in a hermetic manner. Therefore we propagate
data dependency from the rocm libs

🚀 Kind of Contribution

Not relevant

🧪 Unit Tests:
CI

🧪 Execution Tests:
CI

### DIFF
--- a/third_party/py/python_wheel.bzl
+++ b/third_party/py/python_wheel.bzl
@@ -237,9 +237,12 @@ def _collect_data_files_impl(ctx):
     for symlink_dep in ctx.attr.symlink_deps:
         for f in symlink_dep[FilePathInfo].files.to_list():
             files[f] = True
-    return [DefaultInfo(files = depset(
-        files.keys(),
-    ))]
+
+    files_depset = depset(files.keys())
+    return [DefaultInfo(
+        files = files_depset,
+        runfiles = ctx.runfiles(files = files_depset.to_list()),
+    )]
 
 collect_data_files = rule(
     implementation = _collect_data_files_impl,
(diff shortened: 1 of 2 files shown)


akuegel (approved):

alekstheod:
Adjasted based on @ybaturina request, using collect_data_files instead of adapting py_import rule.

third_party/py/python_wheel.bzl
```diff
@@ -237,9 +237,12 @@ def _collect_data_files_impl(ctx):
...
         for f in symlink_dep[FilePathInfo].files.to_list():
             files[f] = True
-    return [DefaultInfo(files = depset(
-        files.keys(),
-    ))]
+
+    files_depset = depset(files.keys())
+    return [DefaultInfo(
+        files = files_depset,
+        runfiles = ctx.runfiles(files = files_depset.to_list()),
```
akuegel: Suggestion: replace this line with `runfiles = ctx.runfiles(transitive_files = files_depset)`

According to AI tool:

Rationale:
In Bazel / Starlark rules, when you already hold a depset of files (files_depset), passing it directly to the transitive_files parameter of ctx.runfiles(...) avoids calling .to_list(). Calling .to_list() flattens the depset DAG into an in-memory Python/Starlark list during Bazel's analysis phase, which can increase memory usage and degrade build performance on large dependency graphs. Preserving the depset via transitive_files keeps analysis fast and memory-efficient.

alekstheod: Done

akuegel (approved):

akuegel (approved):

PR #45799, opened 2026-07-16
[GPU] Reroute external hero users to the fusion output in DSFv2 rewriter

terryysun:
📝 Summary of Changes
Reroute other users of DSF hero to the fusion result, so the original hero can be removed.

🎯 Justification
DSFv2 moved hero into the fusion and removes the original hero, but the original hero cannot be removed if it has other users outside the chain. The will cause duplicated hero operations.

🚀 Kind of Contribution
🐛 Bug Fix

N/A

🧪 Unit Tests:
Added.

🧪 Execution Tests:
Added.

### DIFF
--- a/xla/backends/gpu/tests/dynamic_slice_fusion_v2_test.cc
+++ b/xla/backends/gpu/tests/dynamic_slice_fusion_v2_test.cc
@@ -28,6 +28,10 @@ limitations under the License.
 #include "xla/backends/gpu/runtime/while_loop.h"
 #include "xla/backends/gpu/tests/hlo_pjrt_gpu_test_base.h"
 #include "xla/ffi/ffi.h"
+#include "xla/hlo/ir/hlo_computation.h"
+#include "xla/hlo/ir/hlo_instruction.h"
+#include "xla/hlo/ir/hlo_module.h"
+#include "xla/hlo/ir/hlo_opcode.h"
 #include "xla/literal.h"
 #include "xla/literal_util.h"
 #include "xla/shape_util.h"
@@ -103,6 +107,7 @@ class DynamicSliceFusionV2Test : public HloPjRtGpuTestBase {
  protected:
   DebugOptions GetDebugOptionsForTest() const override {
     DebugOptions debug_options = HloPjRtGpuTestBase::GetDebugOptionsForTest();
+    debug_options.set_xla_gpu_enable_dynamic_slice_fusion(true);
     debug_options.set_xla_gpu_experimental_dynamic_slice_fusion_verify_offsets(
         true);
     return debug_options;
@@ -170,6 +175,71 @@ TEST_F(DynamicSliceFusionV2Test, SingleOutputOneDUS) {
       Literal result, Execute(std::move(*ParseAndReturnVerifiedModule(hlo)), {},
                               /*run_hlo_passes=*/false));
   EXPECT_TRUE(LiteralTestUtil::Equal(expected, result));
+}
+
+TEST_F(DynamicSliceFusionV2Test, HeroWithExternalUserRunsOncePerIteration) {
+  // A duplicate hero surviving the rewrite would show up in the optimized
+  // HLO as a second custom-call and skew the accumulated result.
+  const char* hlo = R"(
+    HloModule test
+
+    body {
+      param = (s32[], f32[64], f32[4,64]) parameter(0)
+      i = s32[] get-tuple-element(param), index=0
+      res = f32[64] get-tuple-element(param), index=1
+      buf = f32[4,64] get-tuple-element(param), index=2
+      hero = f32[64] custom-call(res),
+        custom_call_target="__xla_test$$memset_const",
+        api_version=API_VERSION_TYPED_FFI,
+        backend_config="{value = 7.0 : f32}"
+      hero_2d = f32[1,64] reshape(hero)
+      zero = s32[] constant(0)
+      updated = f32[4,64] dynamic-update-slice(buf, hero_2d, i, zero)
+      new_res = f32[64] add(res, hero)
+      one = s32[] constant(1)
+      next_i = s32[] add(i, one)
+      ROOT tuple = (s32[], f32[64], f32[4,64]) tuple(next_i, new_res, updated)
+    }
+
+    cond {
+      param = (s32[], f32[64], f32[4,64]) parameter(0)
+      i = s32[] get-tuple-element(param), index=0
+      limit = s32[] constant(4)
+      ROOT cmp = pred[] compare(i, limit), direction=LT
+    }
+
+    ENTRY main {
+      zero = s32[] constant(0)
+      init_res = f32[64] broadcast(f32[] constant(0)), dimensions={}
+      init_buf = f32[4,64] broadcast(f32[] constant(0)), dimensions={}
+      init = (s32[], f32[64], f32[4,64]) tuple(zero, init_res, init_buf)
+      ROOT while = (s32[], f32[64], f32[4,64])
+        while(init), condition=cond, body=body
+    }
+  )";
+
+  ASSERT_OK_AND_ASSIGN(
+      Literal result, Execute(std::move(*ParseAndReturnVerifiedModule(hlo)), {},
+                              /*run_hlo_passes=*/true));
+
+  Literal expected = LiteralUtil::MakeTupleOwned(
+      LiteralUtil::CreateR0<int32_t>(4),
+      LiteralUtil::CreateFullWithDescendingLayout<float>({64}, 28.0f),
+      LiteralUtil::CreateFullWithDescendingLayout<float>({4, 64}, 7.0f));
+  EXPECT_TRUE(LiteralTestUtil::Equal(expected, result));
+
+  // A duplicate of the hero surviving the rewrite is visible in the optimized
+  // HLO as a second custom-call next to the one inside the fusion.
+  ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> optimized,
+                       GetOptimizedModule(hlo));
+  int64_t hero_count = 0;
+  for (const HloComputation* computation : optimized->computations()) {
+    for (const HloInstruction* instr : computation->instructions()) {
+      hero_count += instr->opcode() == HloOpcode::kCustomCall &&
+                    instr->custom_call_target() == "__xla_test$$memset_const";
+    }
+  }
+  EXPECT_EQ(hero_count, 1);
 }
 
 TEST_F(DynamicSliceFusionV2Test, SingleOutputOneDUSWithOffsetExpression) {
--- a/xla/backends/gpu/tests/dynamic_slice_fusion_v2_test.cc
+++ b/xla/backends/gpu/tests/dynamic_slice_fusion_v2_test.cc
@@ -15,6 +15,7 @@ limitations under the License.
 
 #include <cstddef>
 #include <cstdint>
+#include <memory>
 #include <utility>
 
 #include <gmock/gmock.h>
@@ -178,8 +179,8 @@ TEST_F(DynamicSliceFusionV2Test, SingleOutputOneDUS) {
 }
 
 TEST_F(DynamicSliceFusionV2Test, HeroWithExternalUserRunsOncePerIteration) {
-  // A duplicate hero surviving the rewrite would show up in the optimized
-  // HLO as a second custom-call and skew the accumulated result.
+  // A surviving duplicate computes identical values -- a pure performance
+  // defect, visible as a second custom-call in the optimized HLO.
   const char* hlo = R"(
     HloModule test
 
@@ -228,8 +229,6 @@ TEST_F(DynamicSliceFusionV2Test, HeroWithExternalUserRunsOncePerIteration) {
       LiteralUtil::CreateFullWithDescendingLayout<float>({4, 64}, 7.0f));
   EXPECT_TRUE(LiteralTestUtil::Equal(expected, result));
 
-  // A duplicate of the hero surviving the rewrite is visible in the optimized
-  // HLO as a second custom-call next to the one inside the fusion.
   ASSERT_OK_AND_ASSIGN(std::unique_ptr<HloModule> optimized,
                        GetOptimizedModule(hlo));
   int64_t hero_count = 0;
--- a/xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
+++ b/xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
@@ -16,6 +16,7 @@ limitations under the License.
 #include "xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.h"
 
 #include <cstdint>
+#include <memory>
 #include <optional>
 #include <string>
 #include <utility>
@@ -34,6 +35,7 @@ limitations under the License.
 #include "absl/types/span.h"
 #include "xla/tsl/platform/status_macros.h"
 #include "xla/backends/gpu/transforms/dynamic_slice_fusion.h"
+#include "xla/hlo/analysis/hlo_reachability.h"
 #include "xla/hlo/ir/hlo_casting_utils.h"
 #include "xla/hlo/ir/hlo_computation.h"
 #include "xla/hlo/ir/hlo_instruction.h"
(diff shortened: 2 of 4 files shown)


xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2_test.cc
```diff
@@ -640,6 +640,64 @@ TEST_F(DynamicSliceFusionRewriterV2Test, DUSOnlyNoSlicedInput) {
...
+  // No custom-call outside the fusion; the add reads back via dynamic-slice.
+  const char* expected = R"(
+    ; CHECK:     %dynamic-slice-fusion{{.*}} {
+    ; CHECK:       {{.*}} custom-call(
+    ; CHECK:              custom_call_target="fake_target"
+    ; CHECK:       ROOT {{.*}} dynamic-update-slice(
+    ; CHECK:     }
+    ; CHECK:     body
+    ; CHECK-NOT:   custom-call(
+    ; CHECK:       {{.*}} = f32[1,64]{{.*}} dynamic-slice(
```
ezhulenev, line 692: I wonder should we even support it? It hero has any other users, shouldn't we skip the rewrite? Or at least for heroes that DUSed + real-used, because in the end we still have one DUS operation that we are forced to materialize, it added a lot of complexity but no real perf wins.

If we fuse one DS that feeds the hero, then it's safe to re-route all users.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+    for (const SlicedResult& sliced_result : sliced_results) {
+      if (!sliced_result.noops.empty())
+        canonical[sliced_result.result_number] = sliced_result.noops.front();
+    }
+    std::vector<HloInstruction*> users(hero->users().begin(),
+                                       hero->users().end());
+    for (HloInstruction* user : users) {
+      auto* gte = Cast<HloGetTupleElementInstruction>(user);
+      auto [it, inserted] = canonical.try_emplace(gte->tuple_index(), gte);
+      if (inserted || it->second == gte) continue;
```
akuegel: nit: please use { } block for if statements

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+    if (absl::c_any_of(hero->users(), [](const HloInstruction* user) {
+          return user->opcode() != HloOpcode::kGetTupleElement;
+        })) {
+      VLOG(2) << "Not rerouting " << hero->name() << ": non-GTE tuple user";
+      return absl::OkStatus();
+    }
+    // Fold duplicate GTEs into the resolved one so each result has one head.
+    absl::flat_hash_map<int64_t, HloInstruction*> canonical;
+    for (const SlicedResult& sliced_result : sliced_results) {
+      if (!sliced_result.noops.empty())
```
akuegel: nit: please use { } block for if statements

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+      if (inserted || it->second == gte) continue;
+      RETURN_IF_ERROR(gte->ReplaceAllUsesWith(it->second));
+      RETURN_IF_ERROR(parent->RemoveInstruction(gte));
+    }
+  }
+
+  // Rerouting a user that (transitively) feeds the fusion would create a
+  // cycle; such users keep reading the original hero.
+  std::unique_ptr<HloReachabilityMap> reachability =
+      HloReachabilityMap::Build(parent);
```
akuegel: I think HloDfsReachability would be better here. It seems unlikely that we need full reachability information, so "on-demand" dfs search should be faster.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+    absl::Span<const SlicedResult> sliced_results) {
+  HloComputation* parent = fusion->parent();
+  bool tuple_hero = hero->shape().IsTuple();
+
+  if (tuple_hero) {
+    // Non-GTE consumers need the raw tuple, which the fusion lacks; skip.
+    if (absl::c_any_of(hero->users(), [](const HloInstruction* user) {
+          return user->opcode() != HloOpcode::kGetTupleElement;
+        })) {
+      VLOG(2) << "Not rerouting " << hero->name() << ": non-GTE tuple user";
```
akuegel: If we cannot reroute, I guess it means we will have the duplicated hero, right? Should we possibly modify the rewriter to not rewrite in such cases where we cannot reroute, and return an error here?

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+    }
+    HloInstruction* chain_user =
+        tail.empty() ? sliced_result.update_slice : tail.front();
+    std::vector<HloInstruction*> external_users;
+    for (HloInstruction* user : head->users()) {
+      if (user != chain_user && !reachability->IsReachable(user, fusion)) {
+        external_users.push_back(user);
+      }
+    }
+    if (external_users.empty()) continue;
```
akuegel: nit: please use { } block for if statements

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+
+    HloInstruction* slot =
+        fusion->shape().IsTuple()
+            ? parent->AddInstruction(HloInstruction::CreateGetTupleElement(
```
akuegel: Should we check whether such a GetTupleElement already exists?

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+      HloReachabilityMap::Build(parent);
+
+  for (const SlicedResult& sliced_result : sliced_results) {
+    if (sliced_result.update_slice == nullptr) {
+      continue;  // RewriteHero's GTE replacement covers these.
+    }
+    // Noops are single-user, so exactly one user of head continues the chain.
+    HloInstruction* head = hero;
+    absl::Span<HloInstruction* const> tail = sliced_result.noops;
+    if (tuple_hero) {
```
akuegel, line 748: Can we assume that noops is non-empty if the hero has tuple shape?
Also, further down it looks as if we want the shape of head to be a non-tuple? If yes, shouldn't we have a while loop here with `while (head->shape().IsTuple()`?

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,87 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+
+  if (tuple_hero) {
+    // Non-GTE consumers need the raw tuple, which the fusion lacks; skip.
+    if (absl::c_any_of(hero->users(), [](const HloInstruction* user) {
+          return user->opcode() != HloOpcode::kGetTupleElement;
+        })) {
+      VLOG(2) << "Not rerouting " << hero->name() << ": non-GTE tuple user";
+      return absl::OkStatus();
+    }
+    // Fold duplicate GTEs into the resolved one so each result has one head.
```
akuegel, line 720: Where does the duplication come from? Can we possibly avoid that there are duplicates?

xla/backends/gpu/tests/dynamic_slice_fusion_v2_test.cc
```diff
@@ -171,6 +176,71 @@ TEST_F(DynamicSliceFusionV2Test, SingleOutputOneDUS) {
   EXPECT_TRUE(LiteralTestUtil::Equal(expected, result));
 }
 
+TEST_F(DynamicSliceFusionV2Test, HeroWithExternalUserRunsOncePerIteration) {
+  // A duplicate hero surviving the rewrite would show up in the optimized
+  // HLO as a second custom-call and skew the accumulated result.
```
akuegel: Just for my understanding: "skew the accumulated result". Does it mean the duplicate custom call is not only a performance issue, but also a correctness issue? If yes, see my other comment that we should possibly not rewrite if we cannot reroute the other users of the hero.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
ezhulenev: I'd completely bail on cases where hero used multiple times, just to keep it simple. In practice we need to optimize jax scan loops that "accumulate" residuals and then read them back from DUSed buffers, everything else adds a ton of complexity and minimal perf improivement.

In this particular PR re-routing will need fixed point iteration to optimize away all slices.

Correct approach would be to rework fusion pipeline to make DUS fusion more of a first class citizen in XLA, then DS+add (example from test) would be fused by "codegen-fusion", but this is a much bigger change with many open questions.

akuegel: I agree, let's keep it simple, especially since we have not yet re-enabled the feature by default yet, so we would not get any extended test coverage unless we run presubmits with the feature turned on.

@terryysun Can you please turn this PR into a draft and create another one where you just disallow heroes with multiple users? If later on we have a use case (and benchmarks showing that it is beneficial) for this reroute feature + allowing multiple users, we can reactivate the PR.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2_test.cc
terryysun: Here's a microbenchmark where DUS fusion is still beneficial with other users rerouted
```
import os, time, jax, jax.numpy as jnp

M, K, N, T = 8192, 64, 8192, 32 
w = jax.random.normal(jax.random.key(0), (K, N), jnp.bfloat16)

@jax.jit
def step(res):
  def body(carry, _):
    out = carry @ w # hero
    return carry + 1e-3 * out[:, :K], out # external user
  new_res, stack = jax.lax.scan(body, res, None, length=T)
  return new_res, stack.sum()

res, s = step(jnp.ones((M, K), jnp.bfloat16))
```
and profile:
unfused step time ~4.07ms:
Screenshot 2026-07-16 at 22 34 32
fused step time ~2.67ms:
Screenshot 2026-07-16 at 22 34 49
The microbenchmark setup is probably in favor of fusion but we also observed ~1-2% speedup on a e2e model we are optimizing.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: Thanks! fixed.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: Fixed.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: Good point! Switched to HloDfsReachability

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: Hi @akuegel I just posted a microbenchmark above showing the perf benefit. And we do have a e2e model that can benefit from this. 

Agreed that first class DUS fusion @ezhulenev mentioned is the better long term solution, this PR is more of a incremental optimization on top of the status quo. BTW the fusion pass after this pass can already absorb the the slices into the consumer.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: Fixed.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: Added a helper to check and reuse.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
akuegel: Thanks for providing the microbenchmark. I see you also updated the comment for the end2end test and made it clear it is currently (at head) a performance defect but not causing wrong results. I still think it would be best if we merge this PR only after we have enabled dynamic slice fusion by default. I am still waiting for feedback from an internal customer whether the latest correctness fix has fixed the issues they have faced. Once that is done, I will re-enable the feature by default. Then we can have more confidence in presubmit tests when this PR gets merged.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: > Can we assume that noops is non-empty if the hero has tuple shape?

Yes, a tuple hero's value only reaches a DUS through GTE so it's never empty.

> Also, further down it looks as if we want the shape of head to be a non-tuple? If yes, shouldn't we have a while loop here with while (head->shape().IsTuple()?

Nested tuple results are rejected at matching, so GTE always yields non-tuple.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: Yes this can wait until DSF is default, we all spent quite some time on that one :)

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: The duplication are from some passes right before this pass and after the last time we run CSE. Running CSE right before this pass would be the global way to avoid duplication, this is just a lightweight local resolution.

xla/backends/gpu/tests/dynamic_slice_fusion_v2_test.cc
terryysun: Sorry for the confusion, fixed the wording.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2_test.cc
ezhulenev: Is it because gemm ends up computed twice? what if only DS is fused? Or the re-routed slice actually get's fused into the addition fusion?

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
ezhulenev: +1, I'd love to re-enable it and make sure it works, it took too much time to fix all edge cases

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2_test.cc
terryysun: Sorry I missed this thread earlier! It's purely from the fusion, there's no duplicated gemm in those profiles. DS-only fusion has no target in this setup, so it's essentially no fusion. The re-routed slice does get fused into the addition fusion.

akuegel:
@terryysun @ezhulenev FYI, I am still waiting for confirmation whether everything is ok now to flip the flag. The person offering to check this was out of office for a few days, but promised to look into it now. Hopefully we can flip the flag on Monday.

akuegel:
> @terryysun @ezhulenev FYI, I am still waiting for confirmation whether everything is ok now to flip the flag. The person offering to check this was out of office for a few days, but promised to look into it now. Hopefully we can flip the flag on Monday.

Finally we could flip the flag. If it is not rolled back again, I think we can give this PR a try. @terryysun can you please rebase?

terryysun:
> > @terryysun @ezhulenev FYI, I am still waiting for confirmation whether everything is ok now to flip the flag. The person offering to check this was out of office for a few days, but promised to look into it now. Hopefully we can flip the flag on Monday.
> 
> Finally we could flip the flag. If it is not rolled back again, I think we can give this PR a try. @terryysun can you please rebase?

Thanks! Rebased.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,122 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+  if (hero->shape().IsTuple()) {
+    for (HloInstruction* user : hero->users()) {
+      if (user->opcode() != HloOpcode::kGetTupleElement) {
+        return true;
+      }
+      heads.push_back(user);
+    }
+  } else {
+    heads.push_back(hero);
+  }
```
akuegel, line 697: For the common case that there are obviously no external users (only the single DUS chain user):

```
bool has_external_users = absl::c_any_of(heads, [](const HloInstruction* head) {
  return head->user_count() > 1;
});
if (!has_external_users) {
  return false;
}
```

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
```diff
@@ -665,6 +667,122 @@ absl::Status SetDynamicSliceFusionBackendConfig(HloInstruction* fusion) {
...
+  std::unique_ptr<HloDfsReachability> reachability =
+      HloDfsReachability::Build(hero->parent());
+  for (HloInstruction* head : heads) {
+    for (HloInstruction* user : head->users()) {
```
akuegel, line 701: Would it be possible to skip the DUS chain user here?

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2_test.cc
```diff
@@ -2546,5 +2604,55 @@ TEST_F(DynamicSliceFusionRewriterV2Test, O2LooksThroughOptBarrier) {
...
+
+  const char* expected = R"(
+    ; CHECK-NOT: dynamic-slice-fusion
+    ; CHECK:     %hero = {{.*}} custom-call(
+    ; CHECK:     {{.*}} = f32[4,64]{{.*}} broadcast(%hero)
+    ; CHECK-NOT: dynamic-slice-fusion
+  )";
+  RunAndFilecheckHloRewrite(hlo, MakePipeline(), expected);
+}
+
```
akuegel, line 2663: Suggested test addition: A test with a tuple hero having external users on both a sliced tuple element and a non-sliced (passthrough) tuple element to verify that canonical GTE replacement and dynamic-slice rerouting coexist cleanly on the same fusion.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: user_count() == 1 may still form a cycle. For example say we have
```
hero = tuple-producing custom call
out0 = GTE(hero, 0)
out1 = GTE(hero, 1)
dus0 = DUS(buffer, out0)
copy = copy(dus0)
dus1 = DUS(copy, out1)
ROOT tuple(dus0, dus1)
```
out0 and out1 each have one user, but fusing both will create cycle. (Probably not a practical example but just wanted to show it's possible in HLO.)

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2.cc
terryysun: We need to check for potential cycles like `DUS → copy → fusion input` so probably have to keep it.

xla/backends/gpu/transforms/dynamic_slice_fusion_rewriter_v2_test.cc
terryysun: Good point! Extended `TupleOutputOneDUS` to cover the scenario.

akuegel (approved):

terryysun:
Hi @akuegel could you help take a look at why this is rolled back? Thanks!

akuegel:
> Hi @akuegel could you help take a look at why this is rolled back? Thanks!

I didn't get a reproducer yet, but it was an issue with wrong results, and they did ask the AI to analyze what might be the reason. Here is what we got:

"If the DUS write indices are data-dependent (dynamic) and the write doesn't cover the full output shape, the DynamicSlice readback may read stale data from the destination buffer rather than the hero's output. In your training scan, the DUS indices come from the scan iteration variable, so there could be edge cases with aliasing or buffer reuse."

Whether that makes any sense, I don't know. But the fact is that the PR did cause problems. I guess it was a good idea that we did the flag flip before, otherwise it would have been quite a bit harder to attribute the problems to this PR.

terryysun:
> > Hi @akuegel could you help take a look at why this is rolled back? Thanks!
> 
> I didn't get a reproducer yet, but it was an issue with wrong results, and they did ask the AI to analyze what might be the reason. Here is what we got:
> 
> "If the DUS write indices are data-dependent (dynamic) and the write doesn't cover the full output shape, the DynamicSlice readback may read stale data from the destination buffer rather than the hero's output. In your training scan, the DUS indices come from the scan iteration variable, so there could be edge cases with aliasing or buffer reuse."
> 
> Whether that makes any sense, I don't know. But the fact is that the PR did cause problems. I guess it was a good idea that we did the flag flip before, otherwise it would have been quite a bit harder to attribute the problems to this PR.

Thanks @akuegel, did you get a chance to make a sharable repro?

PR #45816, opened 2026-07-16
[XLA:GPU] Don't let rank-0 dimension-size scalars inherit op sharding

nurmukhametov:
📝 Summary of Changes
Stop rank-0 dimension-size scalars from inheriting the enclosing op's sharding.

🎯 Justification
Shardy-partitioned MoE models (MaxText Mixtral, DeepSeek2, GPT-OSS) fail to compile. Their bf16 top-k routing lowers into an op that, under a tiled sharding, passes a multi-dimensional sharding onto a scalar, which is invalid and aborts compilation. This has been latent and only surfaced after this #44538

🚀 Kind of Contribution
🐛 Bug Fix

🧪 Unit Tests:
New case in `mhlo_to_hlo/tests/sharding.mlir`

### DIFF
--- a/xla/hlo/translate/mhlo_to_hlo/tests/sharding.mlir
+++ b/xla/hlo/translate/mhlo_to_hlo/tests/sharding.mlir
@@ -496,3 +496,16 @@ func.func @main() -> tensor<4xf32> {
   return %1#0 : tensor<4xf32>
 }
 
+// -----
+// CHECK-LABEL: ENTRY %main{{.*}} (Arg_0.1: s32[128,8]) -> s32[128,8] {
+// CHECK: %[[ARG:.*]] = s32[128,8] parameter(0)
+// CHECK-NOT: s32[] constant(8), sharding={devices=[8,1]<=[8]}, metadata=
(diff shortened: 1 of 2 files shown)


xla/hlo/translate/mhlo_to_hlo/tests/sharding.mlir
```diff
@@ -496,3 +496,15 @@ func.func @main() -> tensor<4xf32> {
   return %1#0 : tensor<4xf32>
 }
 
+// -----
```
thcmbs, line 499: fme: should we explicitly check // CHECK-NOT: sharding= or similar here?

nurmukhametov: Added it, but it isn't really useful for FileCheck here. CHECK-NOT is only evaluated between its surrounding CHECKs, and the following CHECK can't match in faulty case, so FileCheck fails there before reaching CHECK-NOT. Also, the translate tool has validation enabled and aborts before printing any IR, so there is nothing to scan. Harmless though, and it documents what we don't want on scalars.

thcmbs (approved):

PR #45821, opened 2026-07-16
[HLO] Validate the dimensions attribute in the HLO parser to avoid crashes

EylonKrause:
### Problem
Several opcode handlers extract or use a value from the `dimensions={...}` attribute without validating it, so malformed-but-lexable HLO text crashes the parser (a `std::out_of_range` throw, an out-of-bounds read, or a `CHECK`-fail) instead of returning an error `Status`.

**Empty `dimensions={}`** — `all-gather`, `all-gather-start`, `reduce-scatter`, and `get-dimension-size` / `set-dimension-size` extract `dimensions->at(0)`, but the parser only checks that the required attribute is *present*, not non-empty. An empty list therefore throws `std::out_of_range`:
```
all-gather(input), replica_groups={}, dimensions={}
reduce-scatter(input), replica_groups={}, to_apply=add, dimensions={}
get-dimension-size(p), dimensions={}
```

**Out-of-range dimension value** — `scan` checks `dimensions->size() == 1` but then uses the value as `operands[0]->shape().dimensions(scan_dim)` with no range check, so an out-of-range value reads out of bounds (and the downstream `ShapeUtil::InsertDimensionAtIndex` `CHECK`-fails):
```
scan(input, init), dimensions={-1}, num_carries=0, to_apply=add
```

### Fix
- Require `dimensions->size() == 1` in the all-gather / reduce-scatter / get-dimension-size / set-dimension-size handlers before extracting `dimensions->at(0)`, matching the checks the sibling `kSort` and `kAllToAll` handlers already have. (`all-reduce`/`all-reduce-start` share the reduce-scatter parse block and don't take a `dimensions` attribute, so the new check is guarded by `opcode == kReduceScatter`.)
- Range-check `scan`'s dimension against the operand rank and return a parse error if it is out of range.

### Tests
Four `HloParserTest` cases assert that `all-gather`, `reduce-scatter`, and `get-dimension-size` with `dimensions={}`, and `scan` with an out-of-range dimension, now return an error instead of crashing.

---

### DIFF
--- a/xla/hlo/parser/hlo_parser.cc
+++ b/xla/hlo/parser/hlo_parser.cc
@@ -1937,7 +1937,8 @@ HloInstruction* HloParserImpl::CreateInstruction(  // NOLINT
       attrs["use_global_device_ids"] = {/*required=*/false, AttrTy::kBool,
                                         &use_global_device_ids};
       if ((!preset_operands && !ParseOperands(&operands, builder)) ||
-          !ParseAttributes(attrs, allow_attributes, shape)) {
+          !ParseAttributes(attrs, allow_attributes, shape) ||
+          dimensions->size() != 1) {
         return nullptr;
       }
       if (opcode == HloOpcode::kAllGather) {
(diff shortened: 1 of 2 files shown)


thcmbs (approved):

derdrdirk:
This PR causes internal test failure that relied on the previous HloParser behavior. Did you encounter a specific problem that made you want to fix this or was this suggested by AI?

PR #45836, opened 2026-07-16
[xla:gpu] Fine grained control of collectives pipelining

ezhulenev:
Add a `CollectivePipeliningMode` to control what collective operations instances are allowed to be pipelined:

* `DEFAULT` - default pipelining at given opt level
* `OFF` - no pipelining at all
* `ON` - pipeline all operations that could be pipelined (legacy behavior of `true` flag)
* `EXPLICIT` - only pipeline operations marked with a frontend attribute

Explicit pipelining attribute gives more control to end users (JAX) on what operations are ok to pipeline. Simplified example below shows a very simple case with independent collectives, in practice in JAX users might want to statically unroll a small loop inside a scan loop, and pipelining all iterations of unrolled loop is too aggressive and unnecessary increases the buffer assignment and keeps to many collective operations in flight. 

**IMPORTANT:** legacy boolean flags parsed into `AUTO` mode.

### Example

Assuming independent collectives, N iterations, and simplified async all-gather operations, the original loop is:

```
  for i = 0 .. N-1:
    a0 = all-gather(x0[i])
    a1 = all-gather(x1[i])
    a2 = all-gather(x2[i])
    compute(i, a0, a1, a2)
```

#### Pipelining all three

All three collectives are prefetched one iteration ahead:

```
  # Prologue: initiate iteration 0 collectives.
  p0 = all-gather-start(x0[0])
  p1 = all-gather-start(x1[0])
  p2 = all-gather-start(x2[0])

  # Steady state.
  for i = 0 .. N-2:
    a0 = all-gather-done(p0)
    a1 = all-gather-done(p1)
    a2 = all-gather-done(p2)

    # Initiate collectives for the next iteration.
    p0 = all-gather-start(x0[i + 1])
    p1 = all-gather-start(x1[i + 1])
    p2 = all-gather-start(x2[i + 1])

    # Ideally overlaps with the next collectives.
    compute(i, a0, a1, a2)

  # Epilogue: consume the final prefetched values.
  a0 = all-gather-done(p0)
  a1 = all-gather-done(p1)
  a2 = all-gather-done(p2)
  compute(N - 1, a0, a1, a2)
```

The loop carries three outstanding collective handles and needs prologue/epilogue plumbing for all three.

#### Pipelining only all-gather-0

In explicit mode:

```
  all-gather-0, is_pipelineable=true
  all-gather-1
  all-gather-2
```

Conceptually becomes:

```
  # Prologue.
  p0 = all-gather-start(x0[0])

  # Steady state.
  for i = 0 .. N-2:
    a0 = all-gather-done(p0)

    # These remain in their original iteration.
    a1 = all-gather(x1[i])
    a2 = all-gather(x2[i])

    # Only all-gather-0 is prefetched.
    p0 = all-gather-start(x0[i + 1])

    compute(i, a0, a1, a2)

  # Epilogue.
  a0 = all-gather-done(p0)
  a1 = all-gather(x1[N - 1])
  a2 = all-gather(x2[N - 1])
  compute(N - 1, a0, a1, a2)
```

Only one collective crosses the iteration boundary, so the loop carries one additional handle instead of three. This reduces loop state, buffering, code-size growth, and the number of outstanding communications.


penumallas1261 (approved):

ezhulenev:
Rebased.

docs/flags_guidance.md
```diff
@@ -78,10 +78,7 @@ Flag                                            | Type                 | Notes
 ### GPU XLA flags
 
 The `-O1` optimization level enables advanced compiler passes for improved GPU
-performance, including several categories of flags below: pipelining of
-data-parallel collectives (`xla_gpu_enable_pipelined_all_gather`,
-`xla_gpu_enable_pipelined_all_reduce`,
-`xla_gpu_enable_pipelined_reduce_scatter`), while loop unrolling
+performance, including several categories of flags below: while loop unrolling
```
akuegel: Why changing the O1 behavior? If I understand correctly, now with -O1 it will be off by default? Wouldn't "auto" be the better replacement mode?

xla/xla.proto
```diff
@@ -754,18 +781,10 @@ message DebugOptions {
   // (on top of basic use of PDL without them enabled by xla_gpu_enable_pdl).
   optional bool xla_gpu_enable_pdl_launch = 504;
 
-  optional bool xla_gpu_enable_pipelined_all_gather = 227;
```
akuegel, line 766: Merging this change will be a bit difficult, as we have internal users which are setting these flags. Would it be possible to keep these flags first and remove them in a second step once the new flags exist? Then we can start migrating users to the new flags before they are removed.

ezhulenev: Parser has a backward compat mode where it parses the flags. Or users have hardcoded protos? As I remember it was usually `XLA_FLAGS`.

docs/flags_guidance.md
ezhulenev: Yes, my bad. To make it easier to understand I changed enum to 4 levels:

1. Default - select on/off based on opt level
2. Off - turn it off for all ops
3. On - turn it on for all ops
4. Explicit - turn it on for annotated ops only

akuegel (approved):

ezhulenev:
Rebased + fixed.

pifon2a (approved):

ezhulenev:
Rebased + fixed proto conflicts.

ezhulenev:
Rebased again.

akuegel:
> Rebased again.

Thanks. In this case, I am handling necessary rebases myself, as I did have to adjust internal code. While the old flags are still supported, they are now marked as deprecated, and we do have users that test that they only set stable flags.

PR #45877, opened 2026-07-17
[XLA:GPU][oneAPI/SPIR-V] Skip compilation and loading of empty constants modules

neudinger:
Hello, xla and intel team, in my journey to understand and improve the SPIR-V compilation pipeline in PJRT, I found that XLA always passes the constants LLVM module to the target compiler, even when the module is empty.
Screenshot 2026-07-17 at 18 24 17

I observed 2 time the `oneAPI::zeModuleCreate_SPIRVToNative` function being called, once for the kernel and once for the empty constants module.

This can be seen by a [perfetto](https://perfetto.dev/) screenshot.

I think the empty constants module can be skipped entirely, avoiding unnecessary compilation and module loading.

## 📝 Summary of Changes

* Skip target-binary compilation when the emitted constants LLVM module contains no functions or globals.
* Skip device-module loading whenever the constants binary is empty, instead of handling this case only for CUDA.
* Verify that constants metadata is not present without a corresponding constants binary.
* Add a regression test covering resolution of an empty constants binary.

## 🎯 Justification

Previously, XLA always passed the constants LLVM module to the target compiler, even when the module was empty. \
At runtime, loading an empty constants module was skipped only for CUDA.

This unnecessarily requires other GPU backends to support compilation and loading of empty modules. \
In particular, it blocks SPIR-V-oriented GPU compilation paths for HLO modules that do not emit device constants.

The change makes the empty-constants case platform-independent. Constant-free GPU workloads benefit from avoiding an invalid or unnecessary constants compilation and module-loading step.

## 🚀 Kind of Contribution

🐛 Bug Fix, 🧪 Tests

## 📊 Benchmark

This is mainly a correctness and backend-portability fix, but the purpose is to avoid unnecessary compilation and module loading for empty constants modules so it may improve performance in such cases.

## 🧪 Unit Tests

Added `GpuModuleGlobalsTest.EmptyBinarySkipsModuleLoading` to `gpu_executable_test.cc`.

The test uses mock stream and stream-executor objects to verify that resolving an empty constants binary:

* does not query the platform;
* does not call `LoadModule`; and
* returns an empty globals map.

Test target:

```bash
ONEAPI_DEVICE_SELECTOR=level_zero:0
bazel test //xla/service/gpu:gpu_executable_test \
  --test_filter=GpuModuleGlobalsTest.EmptyBinarySkipsModuleLoading
```

```bash
bazel build //xla/pjrt/c:pjrt_c_api_gpu_plugin.so \
  --config=sycl_hermetic \
  --repo_env=ONEAPI_VERSION=2026.0 \
  --repo_env=OS=ubuntu_24.10
```

## 🧪 Execution Tests

None added. The changed behavior concerns empty constants-module handling and is covered by the focused hardware-independent unit test above.

### DIFF
--- a/xla/service/gpu/BUILD
+++ b/xla/service/gpu/BUILD
@@ -948,6 +948,8 @@ xla_cc_test(
         "//xla/stream_executor/abi:executable_abi_version_proto_cc",
         "//xla/stream_executor/cuda:cuda_compute_capability",
         "//xla/stream_executor/gpu:tma_metadata",
+        "//xla/stream_executor:mock_stream",
+        "//xla/stream_executor:mock_stream_executor",
         "//xla/tsl/lib/core:status_test_util",
         "//xla/tsl/platform:env",
         "//xla/tsl/platform:status_macros",
--- a/xla/service/gpu/compile_module_to_llvm_ir.cc
+++ b/xla/service/gpu/compile_module_to_llvm_ir.cc
@@ -265,7 +265,9 @@ absl::StatusOr<CompileModuleResults> CompileModuleToLlvmIr(
       thunk_emitter.EmitHloEntryComputation(hlo_module);
 
   llvm::Module* constants_module = thunk_emitter.constants_module();
-  if (!constants_module->empty() || !constants_module->global_empty()) {
+  const bool has_constants_module =
+    !constants_module->empty() || !constants_module->global_empty();
+  if (has_constants_module) {
     ASSIGN_OR_RETURN(
         results.constants_binary,
         compiler->CompileToTargetBinary(thunk_emitter.ConsumeConstantsModule())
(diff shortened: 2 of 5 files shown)


akuegel (approved):

golechwierowicz (commented):
`xla/backends/gpu/tests:gpu_copy_test_gpu` test fails, please fix

mdfaijul:
oneAPI side the PR looks good to me. Thanks! @neudinger for the PR.

akuegel (approved):

neudinger:
> `xla/backends/gpu/tests:gpu_copy_test_gpu` test fails, please fix

I did not find the target gpu_copy_test_gpu

I suppose it is

bazel test \
--config=sycl_hermetic \
//xla/backends/gpu/tests:gpu_copy_test_intelgpu_any \
--test_env=ONEAPI_DEVICE_SELECTOR=level_zero:0,1

akuegel:
> > `xla/backends/gpu/tests:gpu_copy_test_gpu` test fails, please fix
> 
> I did not find the target gpu_copy_test_gpu
> 
> I suppose it is
> 
> bazel test --config=sycl_hermetic //xla/backends/gpu/tests:gpu_copy_test_intelgpu_any --test_env=ONEAPI_DEVICE_SELECTOR=level_zero:0,1

You are right, the suffix depends on the specific backend. I think in your case intelgpu_any sounds right.

